PR #6532, opened 2025-07-22
cc: Enable Zero-copy rasterizer on ATV

yell0wd0g:
This CL adds three flags to configure two pieces of functionality:

- It sets `--enable-zero-copy` to instruct the Compositor's
rasterizer to write directly to GPU resources (i.e. textures).
This was added a while back to 3P (#5958) but for some reason
I forgot AndroidTV. I just stumbled upon it in chrome:gpu.

- This PR also disables GPU accelerated video decoding and
encoding: these shouldn't be used on Cobalt anyway; the flags
allow us to avoid e.g. capability enumeration (I can see on
my Sabrina and Kirkwood's chrome:gpu that encoding is 
enumerated as supported).

Verified locally on my kirkwood pro.

Bug: 420684984

PR #6550, opened 2025-07-23
cobalt: Disable accelerated video deco/encoding

yell0wd0g:
So that the GPU "process" doesn't try to enumerate supported 
formats. Note that Cobalt/Starboard doesn't use these paths, 
all accelerated video decoding is done under Starboard.

Result of this is always empty chrome:gpu 
hardware-accelerated enco/decoding profiles.

Similar to #6532 but for 3P/Linux.

Bug: 420684984

PR #6549, opened 2025-07-22
cobalt: Disable accelerated video deco/encoding

yell0wd0g:
So that the GPU "process" doesn't try to enumerate supported formats. 

Result of this is always empty chrome:gpu hardware-accelerated 
enco/decoding profiles.

Similar to #6532 but for 3P/Linux.

Bug: 420684984

PR #6551, opened 2025-07-23
cobalt: Disable accelerated video deco/encoding

yell0wd0g:
So that the GPU "process" doesn't try to enumerate supported
formats. Note that Cobalt/Starboard doesn't use these paths,
all accelerated video decoding is done under Starboard.

Result of this is always empty chrome:gpu
hardware-accelerated enco/decoding profiles.

Similar to #6532 but for 3P/Linux.

Bug: 420684984


yell0wd0g:
@kaidokert hopefully this is the last iteration of #6549 #6550 :rofl:

kaidokert (approved):

PR #6552, opened 2025-07-23
Fix fallocate API leak and avoid depending on posix_fallocate

niranjanyardi:
posix: Remove instances of the LINUX API "fallocate" and avoid depending on posix_fallocate

Avoid API symbol leak for "fallocate"
Avoid depending on posix_fallocate as the API is not mandated by the POSIX standard, see https://pubs.opengroup.org/onlinepubs/009604299/functions/posix_fallocate.html

Issue: 432313858

### DIFF
--- a/third_party/sqlite/BUILD.gn
+++ b/third_party/sqlite/BUILD.gn
@@ -220,7 +220,7 @@ component("chromium_sqlite3") {
       # mremap is unsupported for Starboard platforms.
       "HAVE_MREMAP=0",
 
-      # Avoid using posix_fallocate for starboard. Refer to b/432313858
+      # posix_fallocate is unsupported for Starboard platforms
       "HAVE_POSIX_FALLOCATE=0",
     ]
   }


third_party/sqlite/BUILD.gn
```diff
@@ -178,6 +178,11 @@ component("chromium_sqlite3") {
     "src/amalgamation/sqlite3.h",
   ]
 
+  if (is_starboard) {
+    # Avoid using fallocate for starboard. Refer to b/432313858
+    sources -= [ "sqlite3_shim.c" ]
```
niranjanyardi: @andrewsavage1 let's discuss this in our 1:1 tomorrow

base/files/file_util_posix.cc
```diff
@@ -996,7 +996,7 @@ bool AllocateFileRegion(File* file, int64_t offset, size_t size) {
   // space. It can fail because the filesystem doesn't support it. In that case,
   // use the manual method below.
 
-#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS)
+#if (BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS)) && !BUILDFLAG(IS_STARBOARD)
```
niranjanyardi, line 999: above comment suggest that a platform specific implementation is not needed. so safe to add this conditional.
I copied comment below for easier visibility


  // First try the more effective platform-specific way of allocating the disk
  // space. It can fail because the filesystem doesn't support it. In that case,
  // use the manual method below.

y4vor (approved):

third_party/sqlite/BUILD.gn
```diff
@@ -216,8 +216,13 @@ component("chromium_sqlite3") {
   }
 
   if (is_starboard) {
-    # mremap is unsupported for Starboard platforms.
-    defines += [ "HAVE_MREMAP=0" ]
+    defines += [
+      # mremap is unsupported for Starboard platforms.
+      "HAVE_MREMAP=0",
+
+      # Avoid using posix_fallocate for starboard. Refer to b/432313858
```
andrewsavage1: Please remove the bug number and just state that fallocate is unsupported for Starboard platforms like above, the bug number won't be helpful for most people reading this

niranjanyardi: done

PR #6567, opened 2025-07-24
Cobalt: Disable BackForwardCache for both ATV and 3P

yell0wd0g:
Back/forward cache is used to speed up backwards navigation. As such
it's not used by YTLR Webapps, which are Single Page Apps (SPAs) [1]. 
This CL/PR disables that functionality via command line flag. 

I expect this CL/PR to be net neutral or positive, because the bfcache
scaffolding won't be initiated.

[1] https://web.dev/articles/bfcache#spa

Bug: 420684984


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -78,7 +78,9 @@ public static List<String> getDefaultCommandLineOverridesList() {
         paramOverrides.add("--disable-accelerated-video-encode");
         // Rasterize Tiles directly to GPU memory.
         paramOverrides.add("--enable-zero-copy");
-
+        // Backforwardcache keeps a copy of the current page when navigating away to
+        // speed up backwards navigation. YTLR Webapps are SPAs and don't use it.
+        paramOverrides.add("--disable-back-forward-cache");
```
zhongqiliang, line 80: curious how often user rewind the video?

yell0wd0g: Is rewinding considered a backward navigation?

zhongqiliang: Sorry, I understood it wrongly. 
For the browser page 's backwards navigation, there is the back button on the remote, do we know if that triggers backwards navigation?

yell0wd0g: Hmm I'm not sure of all the use cases, but I can see the back
button used to navigate within Kabuki's page, or to go from 
the player "page" back to the browse "page (or entirely out
of the YT app, which wouldn't affect bfcaching).

Most of the back button uses should not cause navigations, but
rather just DOM modifications by JS, since it's a SPA.

Maybe the player->browse could be a same-origin navigation? 
If yes I thought it would not benefit from bfcache because
it'd be a same origin navigation but now I see [that changed
for Android](https://groups.google.com/a/chromium.org/g/bfcache-dev/c/0zTVPni5F9g#:~:text=We%20want%20to%20add%20support,independent%20after%20cross%2Dsite%20navigations.). But this would only affect the whole page,
the individual resources HTML/CSS/JS/fonts/images would still
be cached in either the network cache, image caches etc. So
I think it would have some effect but not necessarily a 
massive one. 

WDYT?

zhongqiliang: Do you think it would be better to submit this after Kimono on C26 is fully released? So we can do A/B testing on it.

Awallky: I would prefer that if we are really unsure of the effects. It will help us minimize the blast radius of this speculative removal.

kaidokert:
>  As such it's not used by YTLR Webapps, which are Single Page Apps (SPAs)

This is mostly true, but we do have a few navigation events ( sign-in, profile change). We may actually benefit in that case.

Do we know what the absolute resource saving from this is ?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
yell0wd0g: I was using DevTools and on both a Chrome-desktop session and RaspPi, the Devtools "Application" tab tells me that youtube.com/tv is not eligible for using BFCache
(because of a series of reason I can elaborate (example desktop [DevTools reasons](https://screenshot.googleplex.com/9e7xGZTt6YYYzoV)).

As such, disabling this would have no perceivable effect on Kabuki, and would
save some memory footprint. Thoughts? 

@kaidokert feel free to comment here too (re you [comment](https://github.com/youtube/cobalt/pull/6567#issuecomment-3146789792)).

yell0wd0g:
@zhongqiliang @kaidokert WDYT?

zhongqiliang:
I am fine with it. C25 does not do cache for back forward.

zhongqiliang (approved):

johnxwork (approved):

PR #6569, opened 2025-07-24
posix: Add more tests for stat

madhurajayaraman:
Bug: 432487521

### DIFF
--- a/starboard/nplb/posix_compliance/posix_stat_test.cc
+++ b/starboard/nplb/posix_compliance/posix_stat_test.cc
@@ -47,7 +47,7 @@ class PosixStatTest : public ::testing::Test {
     file_path_ = test_dir_ + "/the_file.txt";
     int fd = open(file_path_.c_str(), O_CREAT | O_WRONLY, user_rw);
     ASSERT_NE(fd, -1) << "Failed to create test file: " << strerror(errno);
-    int write_res = write(fd, kTestFileContent, strlen(kTestFileContent));
+    ssize_t write_res = write(fd, kTestFileContent, strlen(kTestFileContent));
     ASSERT_EQ(static_cast<unsigned long>(write_res), strlen(kTestFileContent));
     ASSERT_EQ(0, close(fd));
 


starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,227 @@
...
+
+TEST_F(PosixStatTest, PathComponentNotDirectoryFails) {
+  struct stat statbuf;
+  std::string path = file_path_ + "/some_other_dir";
+  errno = 0;
+  EXPECT_EQ(stat(path.c_str(), &statbuf), -1);
+  EXPECT_EQ(errno, ENOTDIR);
+}
+
+// TODO: b/432312684 : Remove this test if we don't allow chmod.
```
madhurajayaraman: Will link this PR in the bug so we can get rid of this test depending on where the chmod investigation lands. Also whoever removes it should add a comment to the top of the file adding this to the list of untested scenarios -- will add comment to bug once this goes in/gets approval.

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,227 @@
...
+ protected:
+  void SetUp() override {
+    char template_name[] = "/tmp/stat_test_XXXXXX";
+    char* dir_name = mkdtemp(template_name);
+    ASSERT_NE(dir_name, nullptr) << "mkdtemp failed: " << strerror(errno);
+    test_dir_ = dir_name;
+
+    // Create a file within the test directory.
+    file_path_ = test_dir_ + "/the_file.txt";
+    int fd = open(file_path_.c_str(), O_CREAT | O_WRONLY, 0644);
```
andrewsavage1: Why do group/others need read permission for this file?

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,227 @@
...
+    ASSERT_NE(fd, -1) << "Failed to create test file: " << strerror(errno);
+    ASSERT_EQ(static_cast<unsigned long>(
```
andrewsavage1: Let's put the call to `write` on a separate line so the ASSERT is more readable

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,227 @@
...
+                  write(fd, kTestFileContent, strlen(kTestFileContent))),
+              strlen(kTestFileContent));
+    close(fd);
```
andrewsavage1: ASSERT this returns 0

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,227 @@
...
+  errno = 0;
+  ASSERT_EQ(stat(file_path_.c_str(), &statbuf), 0)
+      << "stat failed: " << strerror(errno);
+
+  // Get parent directory info to compare device ID.
+  struct stat parent_statbuf;
+  ASSERT_EQ(stat(test_dir_.c_str(), &parent_statbuf), 0);
+
+  // Check device and inode numbers.
+  EXPECT_GT(statbuf.st_dev, static_cast<unsigned long>(0));
```
andrewsavage1: Instead of casting a value, use integer literal suffices (`0u`)

starboard/nplb/posix_compliance/posix_stat_test.cc
andrewsavage1: I think we will allow chmod, let's remove this comment

andrewsavage1 (commented):
I saw Holden commented in another similar PR about combining added tests with existing tests for the symbol. If we decide to do that, we should do it here as well

starboard/nplb/posix_compliance/posix_stat_test.cc
madhurajayaraman: U right I updated to 0600

madhurajayaraman:
> I saw Holden commented in another similar PR about combining added tests with existing tests for the symbol. If we decide to do that, we should do it here as well

Yes but I disagree with separating them, I actually think we should have a folder for "old starboard" or something along those lines and move the starboard scenario tests in there so the posix_compliance folder has unit tests that actually test posix compliance.

madhurajayaraman:
We ultimately decided to update the older sb functionality specific files -- in the case where the older files are pretty obsolete or unnecessarily complex, I'll replace the old files with the new ones. Other PRs could just have the old tests augmented.

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,225 @@
...
+  std::string path = file_path_ + "/some_other_dir";
+  errno = 0;
+  EXPECT_EQ(stat(path.c_str(), &statbuf), -1);
+  EXPECT_EQ(errno, ENOTDIR);
+}
+
+TEST_F(PosixStatTest, PermissionDeniedFails) {
+  struct stat statbuf;
+  std::string protected_dir = test_dir_ + "/protected";
+  ASSERT_EQ(mkdir(protected_dir.c_str(), 0755), 0);
```
andrewsavage1: similarly here, why do anything with the group/other bits?

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,225 @@
...
+ protected:
+  void SetUp() override {
+    char template_name[] = "/tmp/stat_test_XXXXXX";
+    char* dir_name = mkdtemp(template_name);
+    ASSERT_NE(dir_name, nullptr) << "mkdtemp failed: " << strerror(errno);
+    test_dir_ = dir_name;
+
+    // Create a file within the test directory.
+    file_path_ = test_dir_ + "/the_file.txt";
+    int fd = open(file_path_.c_str(), O_CREAT | O_WRONLY, 0600);
```
andrewsavage1: The code is easier to read if you replace the file permission numbers with the equivalent flags—i.e. I don't know offhand which 2 permission bits 0600 corresponds to, but it would be fairly easy to guess what `S_IRUSR | S_IWUSR` means

starboard/nplb/posix_compliance/posix_stat_test.cc
madhurajayaraman: 0755 was just the default, but I'll update to 0700 since that's what we use everywhere. Can also just use the permission bits and save it for 0700 on top of the file

starboard/nplb/posix_compliance/posix_stat_test.cc
```diff
@@ -0,0 +1,230 @@
...
+    char template_name[] = "/tmp/stat_test_XXXXXX";
+    char* dir_name = mkdtemp(template_name);
+    ASSERT_NE(dir_name, nullptr) << "mkdtemp failed: " << strerror(errno);
+    test_dir_ = dir_name;
+
+    // Create a file within the test directory.
+    file_path_ = test_dir_ + "/the_file.txt";
+    int fd = open(file_path_.c_str(), O_CREAT | O_WRONLY, user_rw);
+    ASSERT_NE(fd, -1) << "Failed to create test file: " << strerror(errno);
+    int write_res = write(fd, kTestFileContent, strlen(kTestFileContent));
```
andrewsavage1: Doesn't `write` return a ssize_t meaning you wouldn't have to cast it directly below?

andrewsavage1 (approved):

starboard/nplb/posix_compliance/posix_stat_test.cc
madhurajayaraman: ssize_t is a long, not an unsigned long. So yes I do need to change the type of write_res, but I unfortunately do still need to cast

PR #6575, opened 2025-07-24
Copy ContentViewRendererView to cobalt/shell

johnxwork:
This is to make customizations to the class.

Bug: 433592095


haozheng-cobalt:
Could you summarize in the PR description what customization you made to ContentViewRendererView, and what issue it would address?

johnxwork:
> Could you summarize in the PR description what customization you made to ContentViewRendererView, and what issue it would address?

I didn't make any customizations. It's copied over.

cobalt/shell/android/content_view_render_view.cc
```diff
@@ -0,0 +1,138 @@
...
+#include "ui/android/window_android.h"
+#include "ui/gfx/android/java_bitmap.h"
+#include "ui/gfx/geometry/size.h"
+
+using base::android::JavaParamRef;
+using base::android::ScopedJavaLocalRef;
+
+namespace embedder_support {
+
+ContentViewRenderView::ContentViewRenderView(JNIEnv* env,
```
Awallky, line 29: I think we could ultimately just subclass these classes (C++ & Java) as well and reduce the amount of code we need to maintain ourselves. I don't know if we need to copy this over. Each side could be overridden as needed.

johnxwork: I don't want to change any implementation in this PR. We could subclass it after the copy? It's perfectly fine if after the copying we decide to replace this cc file with a subclass implementation.

kaidokert (approved):
This seems fine, it's a code move/reorg refactor. We'll follow up with changes later

PR #6586, opened 2025-07-25
Add subtree importing logic to the rebase script

SortaCookie:
Automation for subtree merging to the linearize script.

### DIFF
--- a/main.py
+++ b/main.py
@@ -2,20 +2,41 @@ import argparse
 import git
 import glob
 import json
+import re
+import sys
 import os
 import shutil
 import subprocess
+
+import common
+
+if os.path.isdir('./cobalt/tools'):
+    sys.path.append(os.path.abspath('./cobalt/tools'))
+    import convert_dep_to_squashed
+else:
+    raise ModuleNotFoundError('cobalt/tools is not found')
+
+COBALT_IMPORTED_MODULES = [
+    'buildtools/third_party/libc++/trunk',
+    'net/third_party/quiche/src',
+    'third_party/angle',
+    'third_party/boringssl/src',
+    'third_party/googletest/src',
+    'third_party/icu',
+    'third_party/webrtc',
+    'v8',
+]
 
 
 def find_deepest_common_dir(paths):
     if not paths:
         return '/'
-    
+
     # Obtain directory paths for each file path
     dir_paths = [os.path.dirname(p) for p in paths]
     # Split directory paths into components
     split_paths = [p.split('/') for p in dir_paths]
-    
+
     # Handle absolute paths correctly
     if split_paths and split_paths[0] and split_paths[0][0] == '':
         # For absolute paths, the first element is an empty string.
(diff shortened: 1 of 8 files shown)


main.py
```diff
@@ -2,20 +2,76 @@
...
 import subprocess
 
+if os.path.isdir('./cobalt/tools'):
+    sys.path.append(os.path.abspath('./cobalt/tools'))
+    import convert_dep_to_squashed
+else:
+    raise ModuleNotFoundError('third_party/depot_tools is not found')
+
+# Not perfect because it'll match a mismatched quotes
+_DEPS_FIELD_REGEX = re.compile(r'(\s*)[\'\"](src\/)?(\w+)[\'\"]\s*:')
```
kaidokert: What does it match ? Would be good to have a small sample set in a test somewhere to show a couple examples.

main.py
```diff
@@ -2,20 +2,76 @@
...
+SUBTREE_SCRATCH_DIR = '/tmp/cobalt/subtree_scratch'  # Temporary directory to stage repo clones
```
kaidokert: Do a `os.path.join(tempfile.gettempdir(),"cobalt","subtree_scratch")` please just to be polite. I know this will work 100% of the cases but .. we never know when we want to turn this code into a bigger tool.

main.py
```diff
@@ -2,20 +2,76 @@
...
+COBALT_IMPORTED_MODULES = [
```
kaidokert, line 19: Do we need a hardcoded list here ? I thought we'd be able to identify those from walking the commits on the main branch and extract them.

It's .. probably okay as a stopgap for the tool though

main.py
```diff
@@ -2,20 +2,76 @@
...
+    'third_party/angle',
+    # 'third_party/boringssl/src',
+    # 'third_party/googletest/src',
+    # 'third_party/icu',
+    # 'third_party/webrtc',
+    # 'v8',
+]
+
+
+def get_field_def_pos(deps_path, module):
```
kaidokert: This one has quite a bit of parsing logic: probably a good time to break this out into a separate python module ( a `common.py` or something ) and add a test set for it, e.g. `common_test.py` as well. Makes debugging and following the logic easier.

Ask Gemini to do this, it'll happily oblige

main.py
SortaCookie: Simply removed it. It was vestigial from the diff patches that I was messing around with.

main.py
SortaCookie: Moved to a new file.

main.py
SortaCookie: Moved to common and is tested.

main.py
SortaCookie: I extracted the logic to a helper function so it's easier to use while walking the commits on the main. I'll leave it future cls to get that functionality.

main.py
```diff
@@ -2,20 +2,41 @@
...
 import shutil
 import subprocess
 
+import common
+
+if os.path.isdir('./cobalt/tools'):
+    sys.path.append(os.path.abspath('./cobalt/tools'))
+    import convert_dep_to_squashed
+else:
+    raise ModuleNotFoundError('third_party/depot_tools is not found')
```
sideb0ard: Log message has wrong directory name

main.py
```diff
@@ -24,30 +45,53 @@ def find_deepest_common_dir(paths):
...
+
         return '/'.join(common_path) if common_path else ''
 
 
+def import_subtree(repo, deps_file, module, git_source, git_rev):
+    start, end = common.get_field_def_pos(deps_file, 'src/' + module)
+    print('Commenting out the DEPS files for module %s' % module)
+    with open(deps_file, 'r+', encoding='utf-8') as f:
+        lines = f.readlines()
+        for i in range(start, end):
```
sideb0ard, line 79: I think there is an off-by-one error here. 
When i run this locally, it does not comment out the Var line: e.g.
```
-  'src/net/third_party/quiche/src':
+# Cobalt: imported
+#  'src/net/third_party/quiche/src':
     Var('quiche_git') + '/quiche.git' + '@' +  Var('quiche_revision'),
```

I believe `get_field_def_pos` returns the correct line numbers, but they dont all get commented out here. If i add a `+ 1` it correctly comments all 3 lines:
```
-        for i in range(start, end):
+        for i in range(start, end + 1):
```

```
-  'src/net/third_party/quiche/src':
-    Var('quiche_git') + '/quiche.git' + '@' +  Var('quiche_revision'),
+# Cobalt: imported
+#  'src/net/third_party/quiche/src':
+#    Var('quiche_git') + '/quiche.git' + '@' +  Var('quiche_revision'),
```

which is the same as how it looks on https://github.com/youtube/cobalt/blob/main/DEPS
```
# Cobalt: imported
# 'src/net/third_party/quiche/src':
#   Var('quiche_git') + '/quiche.git' + '@' +  Var('quiche_revision'),
```

SortaCookie: There was a bug with the parsing. Thanks for the catch!

PR #6587, opened 2025-07-25
Fix API leak for "posix_fadvise"

niranjanyardi:
posix: Remove usages of the POSIX API "posix_fadvise"

This fixes an API leak for the symbol "posix_fadvise" .
According to https://pubs.opengroup.org/onlinepubs/000095399/functions/posix_fadvise.html
"The posix_fadvise() function is part of the Advisory Information option and need not be provided on all implementations."

Issue: 432314709

### DIFF
--- a/third_party/fontconfig/BUILD.gn
+++ b/third_party/fontconfig/BUILD.gn
@@ -84,10 +84,6 @@ if (use_bundled_fontconfig) {
         "FC_ATTRIBUTE_VISIBILITY_EXPORT=__attribute((visibility(\"hidden\")))",
       ]
     }
-    if (is_starboard) {
-      # posix_fadvise is unsupported for Starboard platforms
-      defines += [ "HAVE_POSIX_FADVISE=0" ]
-    }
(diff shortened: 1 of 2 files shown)


third_party/fontconfig/BUILD.gn
```diff
@@ -84,6 +84,10 @@ if (use_bundled_fontconfig) {
         "FC_ATTRIBUTE_VISIBILITY_EXPORT=__attribute((visibility(\"hidden\")))",
       ]
     }
+    if (is_starboard) {
+      # posix_fadvise is unsupported for Starboard platforms
+      defines += [ "HAVE_POSIX_FADVISE=0" ]
```
andrewsavage1: Add a TODO here to see if we should revert this depending on what we do with fontconfig (i.e. if we use a solution for fonts that doesn't use this route at all). The TODO should probably be the existing fontconfig bug I asked you to look at

andrewsavage1 (approved):

PR #5978, opened 2025-06-03
Copy content/shell/browser/ to cobalt/shell/browser

haozheng-cobalt:
This PR copies everything from content/shell/browser/ to cobalt/shell/browser.
Will remove unrelated parts from the cobalt/shell/browser dir  in follow-up CLs.

Updated below to match Cobalt:
- Header guards.
- Namespaces. [X]. Can not change namespaces to Cobalt yet, we can change it after dependencies are resolved.
- Copyrights
- Imports

Had to copy over the source files cobalt/shell/app/shell_main_delegate.h and cobalt/shell/android/shell_manager.h in this PR too, as they depend on the files from cobalt/shell/browser.
Also updated a few cobalt files to include cobalt/shell/browser headers.

Had to add back conditions for `shell_use_toolkit_views` for now, because it causes [this](https://paste.googleplex.com/4737457423384576#l=6) error without the definition.

Bug: 422209762

PR #6588, opened 2025-07-25
Fix missing VideoGeometrySetterService dependency

abhijeetk:
The VideoGeometrySetterService was added in commit d82b8dacf0a but the content shell was missing the dependency on //cobalt/media/service, causing linker errors during the cobalt_apk build.

This adds the missing dependency to content/shell/BUILD.gn to resolve the undefined symbol errors for VideoGeometrySetterService constructor, destructor, and GetBindCallback methods.

TEST:
python cobalt/build/gn.py -c debug -p android-cobalt-arm64 --no-check
autoninja -C out/android-cobalt-arm64_debug cobalt_apk -j12

### DIFF
--- a/content/shell/BUILD.gn
+++ b/content/shell/BUILD.gn
@@ -318,6 +318,7 @@ static_library("content_shell_lib") {
     "//base:base_static",
     "//build:chromeos_buildflags",
     "//cc/base",
+    "//cobalt/media/service",
     "//cobalt/media/service/mojom",
     "//components/cdm/renderer",
     "//components/custom_handlers",


johnxwork (commented):
Thanks for the patch. But I'm quite sure cobalt_apk does not use content/shell:content_shell_lib anymore. It recently moved to its counterparts in cobalt/shell

abhijeetk:
cc: @yell0wd0g

abhijeetk:
> Thanks for the patch. But I'm quite sure cobalt_apk does not use content/shell:content_shell_lib anymore.

For this branch, The cobalt apk build still transitively depends on content_shell_lib
https://github.com/youtube/cobalt/blob/experimental/chrobalt_m132/cobalt/android/BUILD.gn#L168

>  It recently moved to its counterparts in cobalt/shell

Where I can refer code that you are suggesting.

johnxwork:
You can look at https://github.com/youtube/cobalt/pull/5978. Since this PR to duplicate content_shell_lib, we've been working in the cobalt/shell directory.

rakuco:
Is this PR still necessary with M138?

PR #6590, opened 2025-07-25
Add support for __stack_chk_guard

y4vor:
Added support for __stack_chk_guard based on SbSystemGetRandomUInt64().

Refactored the musl initialization code to include the stack
guard initialization.

Issue: 424003429


cobalt/app/cobalt.cc
```diff
@@ -37,7 +37,7 @@
 #include "ui/ozone/platform/starboard/platform_event_source_starboard.h"
 
 #if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
-#include "third_party/musl/src/starboard/internal/hwcap_impl.h"
+#include <init_musl.h>
```
Awallky, line 40: //cobalt LGTM.
Just curious what change enabled this header to be treated as a system library instead of a local library include?

Awallky (approved):
//cobalt LGTM.

cobalt/app/cobalt.cc
andrewsavage1: We have a number of musl include directories: https://source.corp.google.com/h/github/youtube/cobalt/+/main:third_party/musl/BUILD.gn;l=44

PR #6594, opened 2025-07-25
[POSIX] Remove references to timerfd_* functions

andrewsavage1:
Note that perfetto falls back to PostDelayedTask if timerfd functions are unsupported.

Fixed: 412653597


andrewsavage1:
> Yes, this is probably fine, but something to keep in mind if we ever care a lot about the performance when perfetto runs because using timerfds is probably a good chunk more performant than using the task queues.

I figured it wouldn't matter because I wouldn't expect us to use perfetto in production, but yes that's good to know

kaidokert (approved):

PR #6602, opened 2025-07-28
Fix NSS and Root Store in non-starboard configs

arjungm:
In order to build Chromedriver from the linux-x64x11-no-starboard config, we need to port the following customizations made in the linux-x64x11 config over. This involves disabling NSS from being built into the targets and ensuring it only uses the Chrome Root Store.

Bug: 432077378

### DIFF
--- a/net/BUILD.gn
+++ b/net/BUILD.gn
@@ -1633,11 +1633,10 @@ component("net") {
   # We use `is_cobalt` here because non-starboard configs builds Chromedriver
   # which needs to use the Chrome Root Store only.
   if (is_cobalt && (!chrome_root_store_optional || chrome_root_store_only)) {
-      sources += [
-        "cert/test_root_certs_builtin.cc",
-        "cert/internal/system_trust_store_cobalt.cc"
-      ]
-    }
+    sources += [
+      "cert/test_root_certs_builtin.cc",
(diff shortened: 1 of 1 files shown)


net/BUILD.gn
```diff
@@ -1630,11 +1630,15 @@ component("net") {
...
   }
 
-  if(is_starboard) {
-    sources += [
-      "cert/test_root_certs_builtin.cc",
-      "cert/internal/system_trust_store_starboard.cc"
-    ]
+  # We use `is_cobalt` here because non-starboard configs builds Chromedriver
+  # which needs to use the Chrome Root Store only.
+  if(!chrome_root_store_optional || chrome_root_store_only) {
```
hlwarriner: Since this `if(!chrome_root_store_optional || chrome_root_store_only)` conditional is also part of our customization vs. upstream, please invert the conditionals so the full scope of the customization is guarded by `is_cobalt`/`is_starboard`.

I.e.,
```
if (is_cobalt) {
  if(!chrome_root_store_optional || chrome_root_store_only) {
    sources += [
        "cert/test_root_certs_builtin.cc",
        "cert/internal/system_trust_store_cobalt.cc"
    ]
  }
}
```

arjungm: done

PR #6609, opened 2025-07-28
Register fdatasync in Starboard API 

niranjanyardi:
Register fdatasync in Starboard API , as it doesn't need a wrapper.

The unit tests are generated by gemini-cli.

Issue: 432314054

### DIFF
--- a/starboard/elf_loader/exported_symbols.cc
+++ b/starboard/elf_loader/exported_symbols.cc
@@ -232,6 +232,7 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_SYMBOL(aligned_alloc);
   REGISTER_SYMBOL(calloc);
   REGISTER_SYMBOL(close);
+  REGISTER_SYMBOL(fdatasync);
   REGISTER_SYMBOL(dup);
   REGISTER_SYMBOL(dup2);
   REGISTER_SYMBOL(epoll_create);
@@ -317,7 +318,6 @@ ExportedSymbols::ExportedSymbols() {
     REGISTER_SYMBOL(__errno_location);
   }
   REGISTER_WRAPPER(eventfd);
-  REGISTER_SYMBOL(fdatasync);
   REGISTER_WRAPPER(fstat);
   REGISTER_WRAPPER(freeaddrinfo);
   REGISTER_WRAPPER(ftruncate);


starboard/elf_loader/exported_symbols.cc
```diff
@@ -317,6 +317,7 @@ ExportedSymbols::ExportedSymbols() {
     REGISTER_SYMBOL(__errno_location);
   }
   REGISTER_WRAPPER(eventfd);
+  REGISTER_SYMBOL(fdatasync);
```
borongc: nit: `REGISTER_SYMBOL` should be in the lines with above (L232 - L297).

niranjanyardi: done

starboard/nplb/posix_compliance/posix_fdatasync_test.cc
```diff
@@ -0,0 +1,100 @@
...
+  const size_t kDataSize = sizeof(kData) - 1;
+
+  int fd = open(temp_file.filename().c_str(), O_WRONLY);
+  ASSERT_NE(fd, -1) << "Failed to open temp file: " << strerror(errno);
+
+  ssize_t bytes_written = write(fd, kData, kDataSize);
+  ASSERT_EQ(bytes_written, static_cast<ssize_t>(kDataSize));
+
+  // fdatasync() should succeed.
+  EXPECT_EQ(fdatasync(fd), 0);
```
andrewsavage1, line 41: Are you sure this is testing anything? I would think that this would be trivially true without threading

niranjanyardi: It's just writing data to a file and ensuring the write operation is completed by reading that . so yes, it's testing a trivial case if fdatasync works on the platform(could be any platform).

Do you want me to add more tests ? testing a specific use case ? 
i think the tests are generally trivial as in they test basic functionality and the 2 error codes associated with the method

starboard/nplb/posix_compliance/posix_fdatasync_test.cc
```diff
@@ -0,0 +1,100 @@
...
+  char read_buffer[kDataSize];
+  ssize_t bytes_read = read(fd, read_buffer, kDataSize);
+  EXPECT_EQ(bytes_read, static_cast<ssize_t>(kDataSize));
+  EXPECT_EQ(memcmp(kData, read_buffer, kDataSize), 0);
+
+  EXPECT_EQ(close(fd), 0);
+}
+
+TEST(PosixFdatasyncTest, FailureInvalidFileDescriptor) {
+  // fdatasync() on an invalid file descriptor should fail with EBADF.
```
madhurajayaraman, line 58: Not necessary to change, but a lot of these comments seem superfluous.

starboard/nplb/posix_compliance/posix_fdatasync_test.cc
niranjanyardi: Also, looking at function signature: https://pubs.opengroup.org/onlinepubs/009696799/functions/fdatasync.html

the function itself atleast doesn't take a thread id as input . also function seems thread safe based on the standard: https://pubs.opengroup.org/onlinepubs/009695399/functions/xsh_chap02_09.html

jellefoks (approved):
I think this looks good. This is a hard to test functionality. Maybe you could test that after a large write, fdatasync takes some time because it has to wait for the data to be written, but then what is a reasonable maximum write speed to assume? You could also maybe do a `echo 3 > /proc/sys/vm/drop_caches` from a thread in separate tests right before and after the fdatasync to check  loss / no loss of data, but you need to be root for that and it's very Linux specific.

So... in summary, I think this suffices for now.

starboard/nplb/posix_compliance/posix_fdatasync_test.cc
andrewsavage1: fwiw neither of those things have to be true for us to test with threads, but Jelle's right that it would be quite involved to write a reliable test for this, so let's ignore this for now

PR #6610, opened 2025-07-28
Update all reference of content shell pak to cobalt shell pak

haozheng-cobalt:
This PR simply updates all reference of content shell pak to cobalt shell pak, so that we can start to trim cobalt shell pak next.

Piper side change: https://critique.corp.google.com/cl/788196288

Bug: 433977010,
432782498

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -78,8 +78,8 @@ if (!is_android) {
     }
 
     data_deps = [
+      "//cobalt/shell:pak",
       "//components/crash/core/app:chrome_crashpad_handler",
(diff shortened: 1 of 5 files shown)


Awallky (approved):

Awallky (commented):
Could you add b/432782498 to the Bug footer? It seems relevant, though a bit broader than what this intends to do. This still partly addresses it.

johnxwork (approved):

haozheng-cobalt:
> Could you add b/432782498 to the Bug footer? It seems relevant, though a bit broader than what this intends to do. This still partly addresses it.

Sure, done.

PR #6611, opened 2025-07-28
cobalt/infra: Add Kokoro package steps to GitHub

Awallky:
It is possible for GitHub to pass without detecting issues in the packager script run in Kokoro builds. This adds the same steps that runs in Kokoro so that GitHub will not pass bad builds for the packager script.

Fixed: 434759528

### DIFF
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -130,7 +130,7 @@ runs:
           upload_on_device_test_artifacts: ${{ inputs.upload_on_device_test_artifacts }}
           test_targets_json_file: ${{ inputs.test_targets_json_file }}
       - name: Run Package Script Test
-        if: ${{ (matrix.config == 'qa' || matrix.config == 'gold') && !contains(matrix.platform, 'chromium') && !contains(matrix.platform, 'evergreen') }}
+        if: ${{ (matrix.config == 'qa' || matrix.config == 'gold') && !contains(matrix.platform, 'chromium') && !contains(matrix.platform, 'evergreen') && !contains(matrix.platform, 'modular') }}
         run: |
           # Use the real variable from your environment
           MATRIX_PLATFORM="${{ matrix.platform }}"


.github/workflows/main.yaml
```diff
@@ -336,6 +336,37 @@ jobs:
...
+      - name: Restore CI Essentials
+        uses: actions/download-artifact@v4
+        with:
+          name: ci-essentials-${{ inputs.platform }}
+      - name: Parse Test Target JSON
+        id: test-target-json
+        uses: ./.github/actions/test_targets_json
+      - name: Run Package Script Test
+        run: |
+          ${GITHUB_WORKSPACE}/cobalt/build/packager.py --name=${{ matrix.platform }}_${{ matrix.config }} --json_path=/tmpfs/src/git/src/cobalt/build/android/package.json --out_dir=/tmpfs/src/git/src/out/${{ matrix.platform }}_${{ matrix.config }} --package_dir=/tmpfs/src/git/src/package/${{ matrix.platform }}_${{ matrix.config }}
```
kaidokert: style comment: lets do this:
```
${GITHUB_WORKSPACE}/cobalt/build/packager.py \
   --name=${{ matrix.platform }}_${{ matrix.config }} \
    --json_path=/tmpfs/src/git/src/cobalt/build/android/package.json \
    --out_dir=/tmpfs/src/git/src/out/${{ matrix.platform }}_${{ matrix.config }} \
    --package_dir=/tmpfs/src/git/src/package/${{ matrix.platform }}_${{ matrix.config }}
```

It's a lot more readable with line end breaks

.github/workflows/main.yaml
```diff
@@ -336,6 +336,37 @@ jobs:
         uses: ./src/.github/actions/api_leak_detector
         if: inputs.run_api_leak_detector
 
+  package-script-test:
```
kaidokert: This whole step may need to be at the end of `.github/actions/build/action.yaml` actually. Right after the " Upload Test Artifacts" step.

.github/workflows/main.yaml
```diff
@@ -336,6 +336,37 @@ jobs:
...
+    needs: [initialize, build]
+    if: always() &&
+        (
+          (inputs.platform == 'android-arm' && github.event.pull_request.draft == false)
```
kaidokert: I wouldn't limit it to android. Lets just run it for all platforms, and for all gold + qa configs. If the script doesn't know what to do with Evergreen or linux, we'll fix or stub those things in the script.

.github/workflows/main.yaml
Awallky: Done.

.github/workflows/main.yaml
Awallky: Done.

.github/workflows/main.yaml
Awallky: Ack.

.github/actions/build/action.yaml
```diff
@@ -129,3 +129,13 @@ runs:
           upload_on_host_test_artifacts: ${{ inputs.upload_on_host_test_artifacts == 'true' || inputs.upload_web_test_artifacts == 'true' }}
           upload_on_device_test_artifacts: ${{ inputs.upload_on_device_test_artifacts }}
           test_targets_json_file: ${{ inputs.test_targets_json_file }}
+      - name: Run Package Script Test
+        if: (matrix.config == 'qa' || matrix.config == 'gold')
+        run: |
+          ${GITHUB_WORKSPACE}/cobalt/build/packager.py \
+            --name=${{ matrix.platform }}_${{ matrix.config }} \
+            --json_path=/tmpfs/src/git/src/cobalt/build/android/package.json \
```
kaidokert: All of this looks really good and clean, except that you probably need to strip `/tmpfs/src/git/src` and leave it in Githubs workspace directory ( GITHUB_WORKSPACE i believe ) or just leave relative paths here and put the packaged files in `out/test-package` or something.

Otherwise this looks good

Awallky: I changed to relative paths in the `run` command.

Awallky: > put the packaged files in out/test-package or something

Is that for the `package_dir` field? If so, done. If not, please LMK which field to modify.

kaidokert (approved):
If it builds and passes, LGTM

.github/actions/build/action.yaml
kaidokert: it looks correct now

.github/actions/build/action.yaml
```diff
@@ -129,3 +129,13 @@ runs:
...
           test_targets_json_file: ${{ inputs.test_targets_json_file }}
+      - name: Run Package Script Test
+        if: (matrix.config == 'qa' || matrix.config == 'gold')
+        run: |
+          cobalt/build/packager.py \
+            --name=${{ matrix.platform }}_${{ matrix.config }} \
+            --json_path=cobalt/build/android/package.json \
+            --out_dir=out/${{ matrix.platform }}_${{ matrix.config }} \
+            --package_dir=out/test-package
+          exit 1
```
kaidokert: Actually not sure you need the `exit 1` here.

Awallky: Removed.

arjungm:
@Awallky I was taking a look at the failing checks and I see that it's running the packaging script on `linux-x64x11-modular` but we do not run this configuration. Additionally, chromedriver is not being built for this configuration. If you _want_ to package it, then it needs to be in the list of build targets.

https://github.com/youtube/cobalt/blob/main/.github/config/linux-modular.json#L9-L14

arjungm:
Actually, it seems like the other failures are for platforms that typically have not built chromedriver for us. This could mean you need to find a way to exclude this build step from those platforms.

FWIW, I would **not** exclude chromedriver for linux-modular because the eventual goal is to replace linux-monolithic with it.

Awallky:
BTW @arjungm , I filed a couple of bus, since this was blocked due to incompatibility with evergreen and linux-modular builds and `cobalt/build/linux/package.json`. The JSON lists files that these platforms do not generate. We may need to either add new JSON files specific to those platforms or modify the logic inside the presubmit.

PR #6613, opened 2025-07-29
evergreen: Create an updater_sandbox target

yuying-y:
The target builds for evergreen-x64, linux-x64x11-modular and
linux-x64x11 platforms, but it only runs for linux-x64x11 platform
at the moment. It's a starting point for future iterations.

Issue: 434794270

### DIFF
--- a/chrome/updater/BUILD.gn
+++ b/chrome/updater/BUILD.gn
@@ -339,8 +339,21 @@ if (is_win || is_mac || is_linux) {
         "//components/named_mojo_ipc_server",
       ]
     }
-
-    if (is_linux) {
+    if (is_cobalt_hermetic_build) {
+      # TODO: b/434794270, Cobalt - prototype an updater_sandbox target for evergreen-x64. Will iterate on the list of source files.
+      sources += [
+        "activity_impl_util_linux.cc",
+        "app/server/posix/app_server_posix_linux.cc",
+        "device_management/dm_storage_linux.cc",
+        "installer_linux.cc",
+        "ipc/ipc_security_linux.cc",
+        "linux/setup/setup.cc",
+        "linux/systemd_util.h",
+        "setup_linux.cc",
+        "update_service_internal_impl_qualifying_linux.cc",
+        "update_usage_stats_task_linux.cc",
+      ]
+    } else if (is_linux) {
       sources += [
         "activity_impl_util_linux.cc",
         "app/server/posix/app_server_posix_linux.cc",
--- a/chrome/updater/cobalt/BUILD.gn
+++ b/chrome/updater/cobalt/BUILD.gn
@@ -1,0 +1,42 @@
+# Copyright 2025 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import("//cobalt/build/modular_executable.gni")
+
+if (is_cobalt_hermetic_build) {
+  cobalt_target_type = "modular_executable"
+} else {
+  cobalt_target_type = "executable"
+}
+
+target(cobalt_target_type, "updater_sandbox") {
+  testonly = true
+  sources = [ "updater_sandbox.cc" ]
+
+  configs += [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
+
+  deps = [
+    "//chrome/updater:base",
+    "//chrome/updater:constants_prod",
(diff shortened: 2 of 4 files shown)


chrome/updater/BUILD.gn
```diff
@@ -341,22 +341,38 @@ if (is_win || is_mac || is_linux) {
...
-        "linux/systemd_util.h",
-        "lock_linux.cc",
-        "net/network_fetcher_linux.cc",
-        "setup_linux.cc",
-        "update_service_internal_impl_qualifying_linux.cc",
-        "update_usage_stats_task_linux.cc",
-      ]
-      public_configs = [ "linux:libcurl" ]
+      if (is_cobalt_hermetic_build) {
+        # TODO: prototype an updater_sandbox target for evergreen-x64. Will iterate on the list of source files (b/434794270).
```
andrewsavage1: Use go/todo-format of `# TODO: b/123, Cobalt - Description`

chrome/updater/BUILD.gn
```diff
@@ -341,22 +341,38 @@ if (is_win || is_mac || is_linux) {
...
-        "linux/systemd_util.cc",
-        "linux/systemd_util.h",
-        "lock_linux.cc",
-        "net/network_fetcher_linux.cc",
-        "setup_linux.cc",
-        "update_service_internal_impl_qualifying_linux.cc",
-        "update_usage_stats_task_linux.cc",
-      ]
-      public_configs = [ "linux:libcurl" ]
+      if (is_cobalt_hermetic_build) {
```
andrewsavage1: Avoid changing this many lines of upstream code by i.e. having a separate conditional block for hermetic builds, i.e. you could do
```
if (is_cobalt_hermetic_build) {
  sources += [
    "activity_impl_util_linux.cc",
     ...
  ]
} else if (is_linux) {
  sources += [ ...
}
```
Where you keep the is_linux block the same as it is upstream

chrome/updater/cobalt/BUILD.gn
```diff
@@ -0,0 +1,28 @@
+import("//cobalt/build/modular_executable.gni")
```
andrewsavage1, line 15: Add copyright header

chrome/updater/cobalt/updater_sandbox.cc
```diff
@@ -0,0 +1,31 @@
+// Copyright 2025 The Cobalt Authors
```
andrewsavage1: This is not our fully copyright header

chrome/updater/cobalt/updater_sandbox.cc
```diff
@@ -0,0 +1,31 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "chrome/updater/constants.h"
+#include "chrome/updater/updater.h"
+#include "starboard/common/log.h"
+#include "starboard/event.h"
+
+namespace {
+int g_argc = 0;
```
andrewsavage1, line 21: Why do you need globals for the flags?

chrome/updater/cobalt/updater_sandbox.cc
```diff
@@ -0,0 +1,31 @@
...
+char** g_argv = nullptr;
+}  // namespace
+
+void SbEventHandle(const SbEvent* event) {
+  if (event->type == kSbEventTypeStart) {
+    updater::UpdaterMain(g_argc, g_argv);
+  }
+}
+
+int main(int argc, char* argv[]) {
```
andrewsavage1, line 32: Guard the `main` function behind a `#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)`

chrome/updater/cobalt/updater_sandbox.cc
yuying-y: They need to be saved and passed to updater::UpdaterMain().

chrome/updater/cobalt/updater_sandbox.cc
yuying-y: But this target also builds and only runs for linux-x64x11 platform atm, which is used for iterations and investigation.

andrewsavage1: Right, and on the linux-x64x11 platform `BUILDFLAG(IS_COBALT_HERMETIC_BUILD)` is false, so that would work fine

chrome/updater/cobalt/updater_sandbox.cc
```diff
@@ -0,0 +1,41 @@
+#Copyright 2025 The Cobalt Authors.All Rights Reserved.
```
andrewsavage1: Fix the copyright header commented

chrome/updater/cobalt/updater_sandbox.cc
yuying-y: I may have missed something. But the hermetic platforms also need this main func, don't they? I was referring to cobalt/app/cobalt.cc, which has a main func unguarded. Is the cobalt.cc not built for the hermetic platforms?

andrewsavage1: Hermetic builds have a main entry point at the starboard layer which then routes to the SbEventHandle in the target app layer. So if it's unguarded in cobalt.cc, that's accidental

andrewsavage1 (approved):

PR #6615, opened 2025-07-29
build: Refactor build files to use is_cobalt_hermetic_build

jellefoks:
This change replaces the `is_cobalt` and `!is_android` flags adjusting the libc and ICU builds with the `is_cobalt_hermetic_build` flag.

The `is_cobalt_hermetic_build` flag is used to indicate that the build should not have any dependencies on the host system's libraries, and instead use the versions of libraries that are checked into the repository.

This change affects the build configuration for libc and ICU.

Bug: 434866723

### DIFF
--- a/cobalt/common/libc/BUILD.gn
+++ b/cobalt/common/libc/BUILD.gn
@@ -38,7 +38,9 @@ static_library("libc") {
 source_set("libc_test") {
   testonly = true
 
-  sources = [ "time/tzset_test.cc" ]
+  if (is_cobalt_hermetic_build) {
+    sources = [ "time/tzset_test.cc" ]
+  }
 
   deps = [
     ":libc",


cobalt/common/libc/BUILD.gn
```diff
@@ -17,7 +17,7 @@ static_library("libc") {
     "no_destructor.h",
     "syslog/syslog.cc",
   ]
-  if (!is_android) {
+  if (is_cobalt_hermetic_build) {
```
andrewsavage1, line 20: jsyk this means that non-modular linux won't have these changes (as it's not a hermetic build)—if you want to include that, use `is_starboard`

andrewsavage1 (approved):

jellefoks:
With this change by itself, the non-hermetic build does not use the hermetic tzset anymore, and starts failing the `SetsIcuDefaultCorrectly` because it no longer sets the default timezone in the ICU to match the TZ environment variable...

```
2025-07-31T18:39:00.6386215Z [  FAILED  ] 22 tests, listed below:
2025-07-31T18:39:00.6386718Z [  FAILED  ] Tzset/Timezone.SetsIcuDefaultCorrectly/IanaUTC, where GetParam() = { test_name: "IanaUTC", tz: "UTC", id: "UTC", offset: 0, std_start: { year: 2023, month: 0, day: 1 } }
[snip]
```

PR #6617, opened 2025-07-29
starboard: Remove deprecated SbMedia APIs

kjyoun:
This change removes the following deprecated SbMedia APIs:
- SbMediaGetBufferAlignment
- SbMediaGetBufferPadding
- SbMediaGetBufferStorageType
- SbMediaIsBufferUsingMemoryPool

These APIs were deprecated in Starboard 16 and are no longer used per [change log](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/CHANGELOG.md;l=303-312;drc=a24e4490d336d7421a8fc11c7a03a9de6ec3d29d)
#vide-coded

Bug: 434955864


starboard/nplb/media_buffer_test.cc
```diff
@@ -160,23 +160,6 @@ TEST(SbMediaBufferTest, MediaTypes) {
   }
 }
 
-TEST(SbMediaBufferTest, Alignment) {
```
xiaomings, line 163: We should keep the function and the test for at least one Starboard version.  Imagining a partner has a previous `SbMediaGetBufferAlignment()` implementation that returns `sizeof(void*) * 2`.  With this test in place, their implementation will fail the test so they get a chance to refine it.  If we don't include this test, their implementation may work most of the time so they may not notice it, but still encounter occasional crash or weird behavior.

kjyoun: these APIs are deprecated in  version 16. Hence starboard version 16 has this API definition and tests, though it's not using it.

Updated `changelog.md` in `starboard 17 section.` 

>  Removed APIs deprecated in version 16

xiaomings: Sorry that I didn't describe it correctly.  C25 is backwards compatible for three Starboard versions, so some partners may release C25 with Starboard 15, or they may skip C25 release completely and still using C24 + Starboard 15.  For partners skip Starboard 16 completely, and upgrade their Starboard 15 implementation to Starboard 17, they may encounter the issue mentioned above.

On the other hand, Starboard 17 is special as there is a gap of two years.  If almost all partners will go through Starboard 16 during the two years window, then we can remove the function in Starboard 17.  @y4vor probably has a better idea on this.

kaidokert: We should discuss the deprecation plan for Starboard 17 in more detail perhaps in a separate quick document. Given Chrobalt is not backwards compatible with any previous versions of Starboard, our previous deprecation policy does not really apply - and we could perhaps make a "clean start"

starboard/CHANGELOG.md
```diff
@@ -72,6 +72,12 @@ Their Sb prefixes were removed and renamed to `MediaIsAudioSupported()`,
 `MediaIsSupported()`, and `MediaIsVideoSupported()`.  They are also moved from
 the global namespace into ::starboard::shared::starboard::media.
 
+### Removed APIs deprecated in version 16
```
y4vor, line 75: All these functions are already deprecated and are captured in the file.

starboard/nplb/media_buffer_test.cc
y4vor: The path a partner can take is hard to predict. I think we should add a check for the correct alignment at the place where we get access to the allocated buffer.

kjyoun: we have `starboard/nplb/posix_compliance/posix_posix_memory_allocate_aligned_test.cc` to check `posix_memalign`  works properly.

@y4vor 
do you mean that you need CHECK on actual code that calls it?

PR #6618, opened 2025-07-29
posix: Augment tests for posix_memalign

madhurajayaraman:
Bug: 432487521


starboard/nplb/posix_compliance/posix_posix_memory_allocate_aligned_test.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  EXPECT_EQ(result, EINVAL);
+  // memptr should be unmodified or NULL on failure.
+  EXPECT_EQ(memptr, nullptr);
+}
+
+TEST(PosixMemalignTest, FailsWithInvalidAlignmentNotMultipleOfVoidPointer) {
+  void* memptr = NULL;
+  // Alignment must be a multiple of sizeof(void*). 1 may not be.
+  if (sizeof(void*) > 1) {
```
andrewsavage1: I don't think we should be this generous, `void*` should always be larger than 1 (also update the comment above to reflect the change you make here)

PR #6621, opened 2025-07-29
posix: Augment tests for free

madhurajayaraman:
Bug: 432487521

### DIFF
--- a/starboard/nplb/posix_compliance/posix_free_test.cc
+++ b/starboard/nplb/posix_compliance/posix_free_test.cc
@@ -27,17 +27,15 @@ Calling free() on a pointer not returned by an allocation function
 namespace starboard::nplb {
 namespace {
 
-// The POSIX spec states that if ptr is a null pointer, no action shall occur.
+// If ptr is a null pointer, no action shall occur.
 TEST(PosixFreeTest, FreeNullDoesNothing) {
-  free(NULL);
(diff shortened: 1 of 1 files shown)


starboard/nplb/posix_compliance/posix_free_test.cc
```diff
@@ -0,0 +1,72 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace starboard::nplb {
+namespace {
+
+// The POSIX spec states that if ptr is a null pointer, no action shall occur.
+TEST(PosixFreeTest, FreeNullDoesNothing) {
+  free(NULL);
+  // The test passes if the above line does not crash.
+  SUCCEED();
```
andrewsavage1: Do the SUCCEED calls in these tests actually do anything?

andrewsavage1 (approved):

starboard/nplb/posix_compliance/posix_free_test.cc
madhurajayaraman: no but it felt suspicious to have a test that just had a single line of code, which as i type it is bad rationale to add it...I'll remove the succeed calls since they're redundant & inconsistent w the repo. i think it also adds explicit intent around a single line test, but the comment achieves the same effect.

PR #6625, opened 2025-07-29
[POSIX] Remove all syscalls

andrewsavage1:
Fixed: 424002414

### DIFF
--- a/components/crash/core/app/BUILD.gn
+++ b/components/crash/core/app/BUILD.gn
@@ -53,7 +53,7 @@ static_library("app") {
     "crashpad.h",
   ]
 
-  if (is_linux && !is_starboard || is_chromeos) {
+  if (is_linux && !is_cobalt_hermetic_build || is_chromeos) {
     sources += [ "crashpad_linux.cc" ]
   }
 
--- a/content/utility/BUILD.gn
+++ b/content/utility/BUILD.gn
@@ -143,13 +143,6 @@ source_set("utility") {
     deps += [ "//components/services/screen_ai:screen_ai_sandbox_hook" ]
   }
 
-  if (is_starboard) {
-    deps -= [ 
-      "//content/utility/speech:speech_recognition_sandbox_hook",
-      "//components/services/screen_ai:screen_ai_sandbox_hook",
-    ]
-  }
-
   if (enable_accessibility_service) {
     deps += [ "//services/accessibility:lib" ]
   }


base/files/dir_reader_posix.h
```diff
@@ -17,15 +17,19 @@
 // seems worse than falling back to enumerating all file descriptors so we will
 // probably never implement this on the Mac.
 
-#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS) || BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_STARBOARD)
+#include "base/files/dir_reader_starboard.h"
```
jellefoks, line 21: What is wrong with using `dir_reader_linux.h` or `dir_reader_fallback.h`?

jellefoks (commented):
```
2025-07-29T21:01:58.4608603Z ../../base/trace_event/trace_log.cc:320:55: error: implicit conversion changes signedness: 'PlatformThreadId' (aka 'int') to 'base::PlatformThreadId' (aka 'unsigned long') [-Werror,-Wsign-conversion]
```

base/files/dir_reader_posix.h
andrewsavage1: dir_reader_fallback.h is just a stub, dir_reader_linux.h uses a syscall for getdirents. It was more straightforward to create a starboard-specific implementation using readdir instead of changing the way dir_reader_linux.h worked imo

v8/src/base/platform/platform-posix.cc
```diff
@@ -812,7 +812,7 @@ int OS::GetCurrentThreadId() {
 #elif V8_OS_SOLARIS
   return static_cast<int>(pthread_self());
 #else
-  return static_cast<int>(reinterpret_cast<intptr_t>(pthread_self()));
+  return static_cast<int>(static_cast<intptr_t>(pthread_self()));
```
andrewsavage1: Should just add a separate block for STARBOARD

base/threading/platform_thread_posix.cc
```diff
@@ -268,7 +264,7 @@ PlatformThreadId PlatformThread::CurrentId() {
 #elif BUILDFLAG(IS_POSIX) && BUILDFLAG(IS_AIX)
   return pthread_self();
 #elif BUILDFLAG(IS_POSIX) && !BUILDFLAG(IS_AIX)
-  return reinterpret_cast<int64_t>(pthread_self());
+  return static_cast<int64_t>(pthread_self());
```
andrewsavage1: Just add separate starboard block

jellefoks: This is a fragile change. The function itself calls that out with `Pthreads doesn't have the concept of a thread ID` at the top.

See the code above, on Linux, it calls the syscall for `gettid()`, and on mac a function `pthread_mach_thread_np` is used to get the thread ID.

That means that there is not a portable way to get the thread ID. That is why we still have `REGISTER_SYMBOL(SbThreadGetId);` in `exported_symbols.cc`

This `#elif` is the 'last resort' option and should be avoided if at all possible. The `pthread_self` value has no uniqueness guarantees (it doesn't even have a value guarantee, you can only use it opaquely as a parameter passed into pthread function calls). In a pthreads implementation, it can  (usually will) simply be a pointer to a struct. A new thread can get (re-use) the same pointer as a previously stopped thread. That means that using this version can (will) result in threads being confused with older threads that have already stopped. That is why POSIX has a `pthread_equal()`  ( https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_equal.html ), because you can't use `==` to guarantee that two `pthread_t` values represent the same thread.

We need to return a `PlatformThreadId` here, but `pthread_self()` returns a `pthread_t`, defined as `typedef uintptr_t pthread_t` in musl (aka a `void *` but as an integer value).

The best solution here is to keep using what we did in C25, using `SbThreadGetId`: for `OS_STARBOARD`, make it use `typedef SbThreadGetId PlatformThreadId` in `base/threading/platform_thread.h` (also include `starboard/thread.h` above it), and here `#if OS_STARBOARD` then return `SbThreadGetId()`.

third_party/musl/src/starboard/misc/syscall.c
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "syscall.h"
+
+#include "starboard/thread.h"
+
+#undef syscall
+
+long syscall(long n, ...) {
```
andrewsavage1, line 22: I could also add routing from getrandom to SbSystemGetRandomData, but it would only work when flags aren't set (it would work for rand_util_posix.cc but not the boringssl usage)

base/linux_util.cc
```diff
@@ -20,6 +20,7 @@
 #include "base/files/dir_reader_posix.h"
 #include "base/files/file_util.h"
 #include "base/files/scoped_file.h"
+#include "base/logging.h"
```
y4vor, line 23: Is this still needed?

components/crash/core/app/BUILD.gn
```diff
@@ -53,7 +53,7 @@ static_library("app") {
     "crashpad.h",
   ]
 
-  if (is_linux || is_chromeos) {
+  if (is_linux && !is_cobalt_hermetic_build || is_chromeos) {
```
y4vor, line 56: Maybe is_starboard is OK?

components/crash/core/app/crashpad_linux.cc
```diff
@@ -197,13 +197,15 @@ bool PlatformCrashpadInitialization(
     int fd = base::GlobalDescriptors::GetInstance()->Get(kCrashDumpSignal);
 
     pid_t pid = 0;
+#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
```
y4vor: If we exclude this file form the monolithic build do we still need the hack?
If still need it please file a ticket.

y4vor (approved):

base/linux_util.cc
andrewsavage1: yeah, this previously was included transitively through dir_reader_posix.h

components/crash/core/app/BUILD.gn
andrewsavage1: Might make testing with crashpad a bit harder @hlwarriner but I think it's okay now that he's got it working with hermetic builds

components/crash/core/app/crashpad_linux.cc
andrewsavage1: Removed

jellefoks:
Builds are failing during linking from missing symbols.

components/crash/core/app/BUILD.gn
andrewsavage1: Going to put this back to is_cobalt_hermetic_build, causes monolithic linux to fail

PR #6627, opened 2025-07-29
[Raspi] Fix EGL context initialization failure

sherryzy:
The evergreen qa build on Raspberry Pi is crashing due to a GL context initialization failure. This is caused by the addition of EGL 1.5 attributes that are not supported on Raspi2.

This change fixes the crash by checking the EGL version and only adding the attributes if the version is greater than 1.4.

Bug: 433739466

### DIFF
--- a/ui/gl/gl_context_egl.cc
+++ b/ui/gl/gl_context_egl.cc
@@ -9,6 +9,10 @@
 #include "base/command_line.h"
 #include "base/logging.h"
 #include "base/ranges/algorithm.h"
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#include "base/strings/string_split.h"
+#include "base/strings/string_number_conversions.h"
+#endif
 #include "base/trace_event/trace_event.h"
 #include "build/build_config.h"
 #include "third_party/khronos/EGL/egl.h"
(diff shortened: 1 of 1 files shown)


ui/gl/gl_context_egl.cc
```diff
@@ -151,11 +152,24 @@ bool GLContextEGL::Initialize(GLSurface* compatible_surface,
     }
   }
 
+  // Get EGL version
```
andrewsavage1, line 159: These changes should be guarded by a BUILDFLAG

sherryzy: I actually think these changes can be pushed upstream. But sure, I'll guard this change by BUILDFLAG(IS_COBALT_HERMETIC_BUILD) first.

andrewsavage1: We should generally guard even upstreamable changes as they may change before they're submitted upstream, this looks good now

You could consider switching to BUILDFLAG(IS_COBALT) to show this is a more general change, but this is also fine

ui/gl/gl_context_egl.cc
```diff
@@ -151,12 +154,31 @@ bool GLContextEGL::Initialize(GLSurface* compatible_surface,
...
   }
 
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+  // Get EGL version
+  const char* version_str =
+      eglQueryString(gl_display_->GetDisplay(), EGL_VERSION);
+  int egl_major = 0;
+  int egl_minor = 0;
+  if (version_str) {
+    sscanf(version_str, "%d.%d", &egl_major, &egl_minor);
```
yell0wd0g: Nit: Could we use eg `CStringTokenizer` or `SplitStringPiece()`
here ISO `sscanf`?

sherryzy: Yeah, used base::SplitString instead.

PR #6631, opened 2025-07-30
POSIX: Add utimes/futimes support to Starboard

johnedocampo:
This PR introduces utimes and futimes support for Starboard. This is done by wrapping the function utimensat, and translating any MUSL specific values to their platform-specified versions.

Since the musl implementations of utimes and futimes eventually call utimensat, we can indirectly support both functions by adding utimensat to Starboard and adding a wrapper to it. To ensure that utimes/futimes call our wrapped utimensat function, we edit some `third_party/musl` files, adding/disabling some code through `#if defined(STARBOARD)` and adding additional files to musl's build.gn.

Bug: 432315159


starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h
```diff
@@ -93,12 +93,26 @@ struct musl_stat {
 #endif
 };
 
+// Special flag for utimensat.
+#define MUSL_AT_SYMLINK_NOFOLLOW 0x100
+
```
johnedocampo, line 111: fix spacing here

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -60,3 +76,29 @@ int __abi_wrap_stat(const char* path, struct musl_stat* musl_info) {
...
+                         int musl_flag) {
+  fildes = (fildes == MUSL_AT_FDCWD) ? AT_FDCWD : fildes;
+
+  const int NO_FLAG = 0;
+  if (musl_flag != MUSL_AT_SYMLINK_NOFOLLOW && musl_flag != NO_FLAG) {
+    errno = EINVAL;
+    return -1;
+  }
+  int flag =
+      (musl_flag == MUSL_AT_SYMLINK_NOFOLLOW) ? AT_SYMLINK_NOFOLLOW : NO_FLAG;
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -60,3 +76,29 @@ int __abi_wrap_stat(const char* path, struct musl_stat* musl_info) {
...
+  if (musl_times) {
+    struct timespec times[2];
+    for (int i = 0; i < 2; i++) {
+      times[i].tv_nsec = musl_nsec_to_platform_nsec(musl_times[i].tv_nsec);
+      times[i].tv_sec = musl_times[i].tv_sec;
+    }
+    return utimensat(fildes, path, times, flag);
+  } else {
+    return utimensat(fildes, path, NULL, flag);
+  }
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/tools/api_leak_detector/evergreen/manifest
```diff
@@ -34,7 +34,6 @@ fseek
 fseeko
 ftell
 ftello
-futimes
```
y4vor, line 27: Also update the api leak detector tool:
https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:starboard/tools/api_leak_detector/api_leak_detector.py;l=91

johnedocampo: Done. Ahhh that's what I was missing. Thanks Yavor!

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -14,6 +14,10 @@
 
 #include "starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h"
 
+#include <errno.h>
+#include <fcntl.h>
+
+namespace {
```
johnedocampo: Is there a general ruling as to whether or not to surround these helper functions with `namespace {}`? I've seen some starboard_layer files use the nameless namespace and other files go without it.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -60,3 +88,26 @@ int __abi_wrap_stat(const char* path, struct musl_stat* musl_info) {
...
+                         const struct musl_timespec musl_times[2],
+                         int musl_flag) {
+  fildes = (fildes == MUSL_AT_FDCWD) ? AT_FDCWD : fildes;
+
+  int flag = musl_flag_to_platform_flag(musl_flag);
+  if (flag == -1) {
+    return -1;
+  }
+
+  const struct timespec* platform_times = NULL;
```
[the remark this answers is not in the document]
johnedocampo: Similar code seen in stat_helper (L24) uses NULL; will keep this as is to preserve a common pattern in the file.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
y4vor: Some of these translation functions could be shared so anonymous namespace isn't a good fit. 
Potentially we can use a named namespace, but most already have musl_foo in the name.

johnedocampo: Removed the anonymous namespace. Thank you for the insight Yavor!

y4vor (approved):

y4vor (approved):

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -47,7 +50,6 @@ static_assert(S_IWOTH == 02,
 static_assert(S_IXOTH == 01,
               "The Starboard layer wrapper expects this value from musl");
 
-namespace {
```
andrewsavage1, line 50: Why'd we remove the anon namespace?

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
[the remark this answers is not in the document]
andrewsavage1: We should prefer nullptr in new code, see i.e. https://engdoc.corp.google.com/eng/doc/devguide/cpp/styleguide.md?cl=head#0_and_nullptr/NULL (NULL can be ambiguous)

starboard/tools/api_leak_detector/api_leak_detector.py
```diff
@@ -118,6 +118,7 @@
     'fstat',
     'fsync',
     'ftruncate',
+    'futimes',
```
andrewsavage1: I don't think futimes and utimes should be added here as they're defined in musl to call utimensat

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
johnedocampo: I noticed that some of the `starboard_layer_posix_X` files were using the anon namespace, while others did not. I left a comment earlier in this PR (https://github.com/youtube/cobalt/pull/6631#discussion_r2255473095) wondering if we should prefer to have them or not. From @y4vor:

> Some of these translation functions could be shared so anonymous namespace isn't a good fit.
> Potentially we can use a named namespace, but most already have musl_foo in the name.

I interpreted this as we should remove the anon namespace, but if you feel that this should be done in a separate ticket, I can put the anon namespace back in.

starboard/tools/api_leak_detector/api_leak_detector.py
johnedocampo: Nice catch. Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
andrewsavage1: Okay sounds good!

PR #6633, opened 2025-07-30
Provide build option to disable xr in blink

kaidokert:
This change introduces a build argument 'disable_xr' that allows disabling XR modules in Blink, reducing the binary size for lightweight builds.

Change-Id: Iab6d36de5fccbe78f8051a2100fe60718be1d537


kaidokert:
Author @yunsik-jung  
Originally uploaded in https://github.com/youtube/cobalt_sandbox/pull/789

PR #6635, opened 2025-07-30
Import cpuinfo from commit beb46ca031988

andrewsavage1:
Import cpuinfo as squashed subtree from DEPS

Bug: 424002414


dahlstrom-g (approved):
with minor adjustment for consistency

PR #6640, opened 2025-07-30
cleanup: Remove out-of-date comments in StarboardRenderer

borongc:
As we moved StarboardRenderer as a MojoRenderer, `StarboardRenderer::GetMediaTime()` is called from `MojoRendererService` and it's on a thread from Gpu thread pool.

Remove out-of-date comments and add DCHECK for `StarboardRenderer::GetMediaTime()`.

Issue: 376328722
Issue: 435225475

### DIFF
--- a/media/starboard/starboard_renderer.cc
+++ b/media/starboard/starboard_renderer.cc
@@ -390,9 +390,8 @@ void StarboardRenderer::SetVolume(float volume) {
   }
 }
 
-// Note: Renderer::GetMediaTime() could be called on both main and media
-// threads.
 TimeDelta StarboardRenderer::GetMediaTime() {
+  DCHECK(task_runner_->RunsTasksInCurrentSequence());
   base::AutoLock auto_lock(lock_);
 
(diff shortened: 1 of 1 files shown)


media/starboard/starboard_renderer.cc
```diff
@@ -390,9 +390,8 @@ void StarboardRenderer::SetVolume(float volume) {
   }
 }
 
-// Note: Renderer::GetMediaTime() could be called on both main and media
-// threads.
 TimeDelta StarboardRenderer::GetMediaTime() {
+  DCHECK(task_runner_->RunsTasksInCurrentSequence());
```
kjyoun, line 394: CHECK?

kjyoun (approved):
LGTM with one suggestion.

media/starboard/starboard_renderer.cc
borongc: Keep it as `DCHECK` to stay the same as the rests. We can open a new ticket to clean up them.

PR #6366, opened 2025-07-10
[Raspi] Prevent GpuMemoryBuffer crash on Raspberry Pi 2

sherryzy:
This change prevents a crash on the Raspberry Pi 2 that occurs during the creation of a GpuMemoryBuffer. The platform's graphics driver appears to have limited or buggy support for certain buffer configurations, leading to instability.

The fix involves two parts:

1.  In host_gpu_memory_buffer_manager.cc, the logic is now forced to always use the GPU service path. This avoids a fallback to a host-side shared memory buffer, which was the initial cause of the crash.

2.  In gpu_service_impl.cc, the function is stubbed out to immediately return a null handle. This prevents the service from making a native buffer allocation call to the driver, which would also cause a crash.

Together, these changes guide the buffer allocation request to a safe, no-op path, ensuring application stability on this platform at the cost of disabling the specific feature that required the buffer.

Bug: 430968699

PR #6646, opened 2025-07-30
[Raspi] Prevent GpuMemoryBuffer crash on modular_devel build

sherryzy:
This change prevents a crash on the Raspberry Pi 2 that occurs during the creation of a GpuMemoryBuffer. By commenting out the two DCHECK, everything seems to work fine.

Bug: 430968699

### DIFF
--- a/components/viz/host/host_gpu_memory_buffer_manager.cc
+++ b/components/viz/host/host_gpu_memory_buffer_manager.cc
@@ -190,7 +190,9 @@ void HostGpuMemoryBufferManager::AllocateGpuMemoryBuffer(
                                                                  format)) {
     buffer_handle = gpu::GpuMemoryBufferImplSharedMemory::CreateGpuMemoryBuffer(
         id, size, format, usage);
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
     DCHECK_EQ(gfx::SHARED_MEMORY_BUFFER, buffer_handle.type);
+#endif
     AllocatedBufferInfo buffer_info(buffer_handle, size, format);
     allocated_buffers_[client_id].insert(
         std::make_pair(buffer_handle.id, buffer_info));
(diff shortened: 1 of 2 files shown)


components/viz/host/host_gpu_memory_buffer_manager.cc
```diff
@@ -190,7 +190,9 @@ void HostGpuMemoryBufferManager::AllocateGpuMemoryBuffer(
                                                                  format)) {
     buffer_handle = gpu::GpuMemoryBufferImplSharedMemory::CreateGpuMemoryBuffer(
```
yell0wd0g, line 191: Looking at the [code](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:gpu/ipc/common/gpu_memory_buffer_impl_shared_memory.cc;l=72-89;drc=ac02e200288bef9c817fad9b06763463e4e9493d) there seems to be three cases when 
`GpuMemoryBufferImplSharedMemory::CreateGpuMemoryBuffer()` would not
return `SHARED_MEMORY_BUFFER`: 
- `IsUsageSupported(usage)` is false
- `BufferSizeForBufferFormatChecked()` fails 
- `UnsafeSharedMemoryRegion::Create()` returns invalid

Do you know which one is failing at the moment? Shouldn't be
the first one `IsUsageSupported()` since it seems checked here
in l.188. So it's either a case of an unsupported size (perhaps
alignment for bi/tri planar format?) or the creation itself.

The latter path goes to PlatformSharedMemoryRegion::Create() 
which I guess follows the posix implementation and in turn 
seems to end up in OS-dependent stuff ([here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/memory/platform_shared_memory_region_posix.cc;l=169-269;drc=d1a68d591a2afd65f70ebc3cbf9f2e678d834a51)); perhaps
that's where this path fails?

sherryzy: Yes, it fails at this statement ```base::UnsafeSharedMemoryRegion::Create(buffer_size);```.

sherryzy:
@yell0wd0g Yes, it supersedes https://github.com/youtube/cobalt/pull/6366. Seems like just removed the two Checks, the crash is gone.

PR #6653, opened 2025-07-31
posix: Implement access in Starboard API

borongc:
This PR implements the `access` function as a direct Starboard wrapper around the native platform API, as its structure is ABI-compatible across target platforms.

The unit tests were generated with assistance from `gemini-cli` and refined for correctness.

Issue: 432312684

### DIFF
--- a/starboard/nplb/posix_compliance/posix_unistd_access_test.cc
+++ b/starboard/nplb/posix_compliance/posix_unistd_access_test.cc
@@ -1,0 +1,109 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
andrewsavage1, line 1: I feel something is strange with these two files below, every line is changed even when it looks unchanged. Maybe different line endings (CRLF vs LF)? We should do our best to fix it to avoid this huge diff

borongc: Fixed.

starboard/nplb/posix_compliance/posix_unistd_access_test.cc
```diff
@@ -0,0 +1,109 @@
...
+#include "starboard/common/log.h"
+#include "starboard/configuration_constants.h"
+#include "starboard/system.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace starboard {
+namespace nplb {
+namespace {
+
+class PosixAccessTest : public ::testing::Test {
```
borongc, line 31: `PosixUnistdAccessTest` or `PosixAccessTest` here?

borongc: `PosixAccessTest` according to https://github.com/youtube/cobalt/pull/6654#discussion_r2245956937.

jellefoks (approved):

PR #6654, opened 2025-07-31
posix: Implement chmod in Starboard API

borongc:
This PR implements the `chmod` function as a direct Starboard wrapper around the native platform API, as its structure is ABI-compatible across target platforms.

The unit tests were generated with assistance from `gemini-cli` and refined for correctness.

Issue: 432312684

### DIFF
--- a/starboard/nplb/posix_compliance/posix_sys_chmod_test.cc
+++ b/starboard/nplb/posix_compliance/posix_sys_chmod_test.cc
@@ -1,0 +1,132 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include <errno.h>
+#include <string.h>
+
+#include "starboard/configuration_constants.h"
+#include "starboard/system.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace starboard {
+namespace nplb {
+namespace {
+
+TEST(PosixChmodTest, SunnyDay) {
+  char cache_path[kSbFileMaxPath];
+  ASSERT_TRUE(
+      SbSystemGetPath(kSbSystemPathCacheDirectory, cache_path, kSbFileMaxPath));
+
+  char temp_file_path[kSbFileMaxPath];
+  int result =
+      snprintf(temp_file_path, kSbFileMaxPath, "%s/chmod_test.tmp", cache_path);
+  ASSERT_GT(result, 0);
+  ASSERT_LT(result, static_cast<int>(kSbFileMaxPath));
+
+  FILE* fp = fopen(temp_file_path, "w");
+  ASSERT_NE(nullptr, fp) << "Failed to open temp file.";
+  fclose(fp);
+
+  struct stat stat_buf;
+  ASSERT_EQ(stat(temp_file_path, &stat_buf), 0);
+
+  // Set read-only for owner
+  ASSERT_EQ(chmod(temp_file_path, S_IRUSR), 0);
+  ASSERT_EQ(stat(temp_file_path, &stat_buf), 0);
+  EXPECT_EQ(static_cast<int>(stat_buf.st_mode & S_IRWXU), S_IRUSR);
+
+  // Set read-write for owner
+  ASSERT_EQ(chmod(temp_file_path, S_IRUSR | S_IWUSR), 0);
+  ASSERT_EQ(stat(temp_file_path, &stat_buf), 0);
+  EXPECT_EQ(static_cast<int>(stat_buf.st_mode & S_IRWXU), S_IRUSR | S_IWUSR);
+
+  remove(temp_file_path);
+}
+
+TEST(PosixChmodTest, RainyDayInvalidPath) {
+  const char* invalid_path = "/this/path/does/not/exist/12345";
+  errno = 0;
+  int result = chmod(invalid_path, S_IRUSR);
+
+  // On failure, chmod should return -1 and set errno.
+  EXPECT_EQ(result, -1);
+  EXPECT_EQ(errno, ENOENT);
+}
+
+TEST(PosixChmodTest, SunnyDayReadOnly) {
+  if (geteuid() == 0) {
+    GTEST_SKIP() << "Test is not meaningful when run as root.";
+  }
+  char cache_path[kSbFileMaxPath];
+  ASSERT_TRUE(
+      SbSystemGetPath(kSbSystemPathCacheDirectory, cache_path, kSbFileMaxPath));
+
+  char temp_file_path[kSbFileMaxPath];
+  int result =
+      snprintf(temp_file_path, kSbFileMaxPath, "%s/chmod_test.tmp", cache_path);
+  ASSERT_GT(result, 0);
+  ASSERT_LT(result, static_cast<int>(kSbFileMaxPath));
+
+  FILE* fp = fopen(temp_file_path, "w");
+  ASSERT_NE(nullptr, fp) << "Failed to open temp file.";
+  fclose(fp);
+
+  // Set read-only for owner
+  ASSERT_EQ(chmod(temp_file_path, S_IRUSR), 0);
+
+  // Attempt to open for writing, which should fail.
+  FILE* fp_write = fopen(temp_file_path, "w");
+  EXPECT_EQ(nullptr, fp_write);
+
+  remove(temp_file_path);
+}
+
+TEST(PosixChmodTest, SunnyDayCtime) {
+  char cache_path[kSbFileMaxPath];
+  ASSERT_TRUE(
+      SbSystemGetPath(kSbSystemPathCacheDirectory, cache_path, kSbFileMaxPath));
+
+  char temp_file_path[kSbFileMaxPath];
+  int result =
+      snprintf(temp_file_path, kSbFileMaxPath, "%s/chmod_test.tmp", cache_path);
+  ASSERT_GT(result, 0);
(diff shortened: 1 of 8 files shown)


starboard/nplb/posix_compliance/posix_sys_chmod_test.cc
```diff
@@ -0,0 +1,73 @@
...
+
+#include "starboard/configuration_constants.h"
+#include "starboard/system.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace starboard {
+namespace nplb {
+namespace {
+
+TEST(PosixSysChmodTest, SunnyDay) {
```
andrewsavage1: Why PosixSysChmodTest instead of PosixChmodTest?

borongc: Which would be a prefer name?

Others like `PosixUnistdAccessTest` or `PosixAccessTest`?

It looks like most of our test naming are the former.

andrewsavage1: When I look through it I only see a few recently added tests that use the subdirectory name in the test (Unistd/Sys)—I think better to just use the symbol name instead, unless there's more widespread use that I'm missing

borongc: If that's the case, should we make a ticket to clean them up?

andrewsavage1: Yes, let's do that

borongc: Renamed to `PosixChmodTest`.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h
```diff
@@ -93,12 +122,16 @@ struct musl_stat {
 #endif
 };
 
+typedef unsigned int musl_mode_t;
```
andrewsavage1: This is repeated

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -43,6 +44,74 @@ int stat_helper(int retval,
   return retval;
 }
 
+mode_t musl_mode_to_platform_mode(musl_mode_t musl_mode) {
+  mode_t platform_mode = 0;
+
+  // File type
+  if ((musl_mode & MUSL_S_IFMT) == MUSL_S_IFDIR) {
```
andrewsavage1, line 84: Why not do `if (musl_mode & MUSL_S_IFDIR) { ...` like below?

andrewsavage1 (approved):

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h
```diff
@@ -30,6 +30,35 @@
...
+#define MUSL_S_IFSOCK 0140000
+
+// Musl permission constants
+#define MUSL_S_ISUID 04000
+#define MUSL_S_ISGID 02000
+#define MUSL_S_ISVTX 01000
+#define MUSL_S_IRUSR 0400
+#define MUSL_S_IWUSR 0200
+#define MUSL_S_IXUSR 0100
+#define MUSL_S_IRWXU 0700
```
jellefoks: Many (but not all) of these defines and their corresponding wrappings should not be necessary. A large number of these have predefined values in the spec (*1). Instead, you can add `static_assert` statements to verify their values.

*1: https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_stat.h.html

starboard/nplb/posix_compliance/posix_sys_chmod_test.cc
```diff
@@ -0,0 +1,73 @@
...
+  ASSERT_LT(result, static_cast<int>(kSbFileMaxPath));
+
+  FILE* fp = fopen(temp_file_path, "w");
+  ASSERT_NE(nullptr, fp) << "Failed to open temp file.";
+  fclose(fp);
+
+  struct stat stat_buf;
+  ASSERT_EQ(stat(temp_file_path, &stat_buf), 0);
+
+  // Set read-only for owner
```
jellefoks, line 48: I think you should also add tests for this:
- check that opening the same file for writing actually fails after the file is made read-only.
- check that that ctime is updated reflecting the "Time of last status change" (*1). (needs a delay after file creation to be testable).

Technically, for full compliance test, the test should include testing for group and other permissions, and check disabling reading and enabling/disabling executing, but it's unlikely that we ever need that functionality...

*1: https://pubs.opengroup.org/onlinepubs/009696899/functions/chmod.html  and https://pubs.opengroup.org/onlinepubs/007904875/basedefs/sys/stat.h.html and 
https://pubs.opengroup.org/onlinepubs/009696699/functions/fstat.html

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h
borongc: Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h
borongc: Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
borongc: Removed `musl_mode_to_platform_mode` and used `static_assert` for each value.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -60,3 +113,7 @@ int __abi_wrap_stat(const char* path, struct musl_stat* musl_info) {
   int retval = stat(path, &stat_info);
   return stat_helper(retval, &stat_info, musl_info);
 }
+
+int __abi_wrap_chmod(const char* path, musl_mode_t mode) {
```
andrewsavage1, line 131: Is it necessary to have a wrapper at all anymore?

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -14,6 +14,57 @@
 
 #include "starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.h"
 
+#include <sys/stat.h>
+
+// The following mode flags are specified by POSIX.1-2017.
+static_assert(S_IFMT == 0170000,
```
andrewsavage1: Looking at https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_stat.h.html, it didn't look like the macros for S_IFMT to S_IFSOCK were well defined, only S_ISUID and other file mode bits

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
andrewsavage1: See comment above, I think we still need translation for 

```
S_IFMT
S_IFBLK
S_IFCHR
S_IFIFO
S_IFREG
S_IFDIR
S_IFLNK
S_IFSOCK
```

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
borongc: From `gemini`:

> The file type is a set of mutually exclusive values. A file can be a directory, a regular file, or a symbolic link, but it cannot be more than one of these at the same time. S_IFMT is a "mask" that isolates the bits representing the file type. The check (musl_mode & MUSL_S_IFMT) == MUSL_S_IFDIR first uses the & operator to extract the file type bits from musl_mode, and then uses == to compare that value to a specific file type constant.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
borongc: Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
borongc: Done.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -43,6 +77,74 @@ int stat_helper(int retval,
...
+  }
+  if ((musl_mode & MUSL_S_IFMT) == MUSL_S_IFLNK) {
+    platform_mode |= S_IFLNK;
+  }
+  if ((musl_mode & MUSL_S_IFMT) == MUSL_S_IFSOCK) {
+    platform_mode |= S_IFSOCK;
+  }
+
+  // Permissions
+  if (musl_mode & S_ISUID) {
```
andrewsavage1: Instead of this whole block you should be able to do `platform_mode |= (musl_mode & 07777)`

borongc: Done.

starboard/nplb/posix_compliance/posix_sys_chmod_test.cc
borongc: Added.

starboard/shared/modular/starboard_layer_posix_stat_abi_wrappers.cc
```diff
@@ -43,6 +77,39 @@ int stat_helper(int retval,
...
+  }
+  if ((musl_mode & MUSL_S_IFMT) == MUSL_S_IFLNK) {
+    platform_mode |= S_IFLNK;
+  }
+  if ((musl_mode & MUSL_S_IFMT) == MUSL_S_IFSOCK) {
+    platform_mode |= S_IFSOCK;
+  }
+
+  // Permissions
+  platform_mode |= (musl_mode & 07777);
```
y4vor, line 107: We should also translate the permissions.
Take a look at S_IRUSR, S_IWUSR, S_IXUSR, S_IRGRP, etc...
I see that you added asserts, but if those don't mach the partner would have to do the translations themself.

borongc: This is from our previous discussion (https://github.com/youtube/cobalt/pull/6654#discussion_r2246506673) due to they are in spec.

y4vor: OK, cool

PR #6658, opened 2025-07-31
Run all on-device unit tests and e2e tests without label

isarkis:
This will effectively make on-device unit tests and e2e always-on for non draft PRs.

Issue: 409815433


kaidokert (approved):
lgtm. Should send a followup email as well about the change as it goes out

Libzu (approved):

PR #1337, opened 2023-08-21
Fix ANR from long device provisioning time.

jellefoks:
This avoids the application from blocking during device provisioning.

This separates creation of the media crypto session into its own function, which is then called from a thread to allow it to complete asynchronously off the main thread.

This also adds a 10 second timeout to the http post for provisioning.

b/296283608

PR #6671, opened 2025-07-31
android: Make DrmSystem support app-assisted provisioning

kjyoun:
This change introduces support for application-assisted device provisioning in the Android DrmSystem. This allows the application to handle DRM provisioning directly, which is necessary for playback on unprovisioned devices where a MediaDrm session cannot be created immediately.

Key changes include:
- A `DrmSessionIdMapper` is introduced to manage the lifecycle mismatch between the EME session (created client-side) and the MediaDrm session (created platform-side, post-provisioning).
- The session request and update flow is now asynchronous to accommodate the provisioning process.
- Session ID terminology has been clarified, renaming the client-side ID to `EmeSessionId` to distinguish it from the platform `MediaDrmSessionId`, improving code clarity.

Bug: 79941850

### DIFF
--- a/starboard/android/shared/drm_system.cc
+++ b/starboard/android/shared/drm_system.cc
@@ -330,15 +330,13 @@ void DrmSystem::OnSessionUpdate(int ticket,
                                 SbDrmSessionRequestType request_type,
                                 std::string_view session_id,
                                 std::string_view content) {
-  std::string eme_session_id_str;
   std::string_view eme_session_id;
   if (kEnableAppProvisioning) {
     std::lock_guard lock(mutex_);
     if (session_id_mapper_->IsMediaDrmSessionIdForProvisioningRequired()) {
       session_id_mapper_->RegisterMediaDrmSessionIdForProvisioning(session_id);
     }
-    eme_session_id_str = session_id_mapper_->GetEmeSessionId(session_id);
-    eme_session_id = eme_session_id_str;
+    eme_session_id = session_id_mapper_->GetEmeSessionId(session_id);
   } else {
     eme_session_id = session_id;
   }


starboard/android/shared/drm_system.cc
```diff
@@ -68,11 +73,28 @@ DrmSystem::DrmSystem(
...
   Start();
+
+  if (kEnableAppProvisioning) {
+    if (std::unique_lock lock(mutex_);
+        !job_queue_created_.wait_for(lock, std::chrono::seconds(1), [this] {
+          return job_queue_ != nullptr;
+        })) {
+      SB_LOG(WARNING) << "job_queue is not created in time";
+    }
+  }
```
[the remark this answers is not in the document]
kjyoun: for now, warning is ok

starboard/android/shared/drm_system.cc
```diff
@@ -173,14 +342,52 @@ void DrmSystem::OnSessionUpdate(int ticket,
...
+    // Grabs first valid ticket id from pending requests.
+    for (auto& request : deferred_session_update_requests_) {
+      if (request->ticket() == kSbDrmTicketInvalid) {
+        continue;
+      }
+
+      ticket = request->ticket();
+      request->ResetTicket();
+      break;
+    }
```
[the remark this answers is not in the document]
kjyoun: this is an intended change.

starboard/android/shared/drm_system.cc
```diff
@@ -140,18 +236,91 @@ void DrmSystem::UpdateSession(int ticket,
...
+  std::unique_ptr<SessionUpdateRequest> request;
+  {
+    std::lock_guard scoped_lock(mutex_);
+    if (deferred_session_update_requests_.empty()) {
+      return;
+    }
+
+    request = std::move(deferred_session_update_requests_.front());
+    deferred_session_update_requests_.erase(
+        deferred_session_update_requests_.begin());
```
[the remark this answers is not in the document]
kjyoun: the size is very small, so performance impact is minimal

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +232,91 @@ void DrmSystem::UpdateSession(int ticket,
...
```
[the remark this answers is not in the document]
kjyoun: This list is very short (< 10), so that changing container type would not affect performance

starboard/android/shared/drm_system.cc
```diff
@@ -67,11 +70,28 @@ DrmSystem::DrmSystem(
...
 
   Start();
+
+  if (kEnableAppProvisioning) {
+    if (std::unique_lock lock(mutex_);
+        !job_queue_created_.wait_for(lock, std::chrono::seconds(1), [this] {
+          return job_queue_ != nullptr;
+        })) {
+      SB_LOG(WARNING) << "job_queue is not created in time";
+    }
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/android/shared/drm_system.cc
```diff
@@ -67,11 +70,28 @@ DrmSystem::DrmSystem(
   if (!media_drm_bridge_->is_valid()) {
     return;
   }
+  SB_LOG(INFO) << "Creating DrmSystem: key_system=" << key_system
+               << ", enable_app_provisioning="
+               << (kEnableAppProvisioning ? "true" : "false");
```
xiaomings: nit: consider using `std::boolalpha`.

starboard/android/shared/drm_system.cc
```diff
@@ -67,11 +70,28 @@ DrmSystem::DrmSystem(
...
 
   Start();
+
+  if (kEnableAppProvisioning) {
+    if (std::unique_lock lock(mutex_);
+        !job_queue_created_.wait_for(lock, std::chrono::seconds(1), [this] {
```
xiaomings: We probably don't have to add a timeout here.  If adding one is preferred, let's make the timeout to 10s.

starboard/android/shared/drm_system.h
```diff
@@ -123,6 +142,22 @@ class DrmSystem : public ::SbDrmSystemPrivate,
...
+
+  // Manages the mapping between the CDM session ID in the C++ layer and the
+  // MediaDrm session ID in the Java layer. Most of the time, we can use the
+  // MediaDrm session ID as the CDM session ID. However, there are some cases
+  // where we cannot, as the lifecycle of the CDM session ID can diverge from
+  // the MediaDrm session ID's lifecycle. For example, when a device isn't
+  // provisioned, we can't get a MediaDrm session ID (which can be generated
+  // only after provisioning). In such scenarios, `DrmSystem` still needs a CDM
+  // session ID to interact with the Cobalt CDM module.
+  const std::unique_ptr<DrmSessionIdMapper>
```
osagie98, line 154: `const std::unique_ptr<DrmSessionIdMapper> session_id_mapper_ GUARDED_BY(mutex_);`

starboard/android/shared/drm_system.cc
```diff
@@ -67,11 +70,28 @@ DrmSystem::DrmSystem(
...
     return;
   }
+  SB_LOG(INFO) << "Creating DrmSystem: key_system=" << key_system
+               << ", enable_app_provisioning="
+               << (kEnableAppProvisioning ? "true" : "false");
 
   Start();
+
+  if (kEnableAppProvisioning) {
+    if (std::unique_lock lock(mutex_);
```
osagie98: Instead of waiting for another thread to create the JobQueue before returning, why not create the JobQueue here in a detached state, then later attach it to the thread calling `Run()`?

starboard/android/shared/drm_system.cc
```diff
@@ -88,9 +108,28 @@ void DrmSystem::Run() {
...
+  job_queue_.reset();
+}
+
 DrmSystem::~DrmSystem() {
   ON_INSTANCE_RELEASED(AndroidDrmSystem);
+  if (job_queue_ != nullptr) {
+    job_queue_->StopSoon();
+  }
   Join();
+  SB_DCHECK_EQ(job_queue_, nullptr);
```
osagie98: If there's no `SB_CHECK_EQ`, would it make sense here to have `SB_CHECK(job_queue_ == nullptr)` with a log?

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +232,91 @@ void DrmSystem::UpdateSession(int ticket,
...
+  {
+    std::lock_guard lock(mutex_);
+    media_drm_session_id =
+        session_id_mapper_->GetMediaDrmSessionId(cdm_session_id);
+  }
+
+  MediaDrmBridge::OperationResult result;
+  if (media_drm_session_id.empty()) {
+    SB_LOG(INFO) << __func__
+                 << " >  Handle the given key as provision response, since "
```
osagie98: nit: Handles the given key as a provision response

starboard/android/shared/drm_system.h
osagie98: `GUARDED_BY(mutex_)`

starboard/android/shared/drm_system.h
```diff
@@ -123,6 +142,22 @@ class DrmSystem : public ::SbDrmSystemPrivate,
...
+      session_id_mapper_;  //  Guarded by |mutex_|.
+
+  std::unique_ptr<JobQueue> job_queue_;
```
osagie98: Are jobs placed in the queue run on a single thread? When running JobQueue functions, can you add `SB_CHECK(job_queue_->BelongsToCurrentThread())`?

starboard/android/shared/drm_system.h
kjyoun: Done

starboard/android/shared/drm_system.h
kjyoun: AFAIK, we cannot use GUARD_BY here, since it's for base::Lock, while we are using std::mutex.

I think that it's not recommended to use base::Lock under starboard/, even though this is for Android.
Current code commented about it, instead

starboard/android/shared/drm_system.h
kjyoun: Good point.

Added it to HandlePendingRequests, the only method running on JobQueue.

starboard/android/shared/drm_system.cc
kjyoun: changed this to SB_CHECK(!job_queue_) for the idiomatic null check.

starboard/android/shared/drm_system.cc
kjyoun: This wait is intended to guarantee that `job_queue_` is initialized by the new thread before the constructor returns.
This prevents a race condition where other methods could be called on the object after construction but before `job_queue_` is ready to be used.

Added comment about it

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.cc
```diff
@@ -231,4 +445,12 @@ void DrmSystem::CallKeyStatusesChangedCallbackWithKeyStatusRestricted_Locked() {
   }
 }
 
+bool DrmSystem::IsReady() {
+  if (kEnableAppProvisioning) {
+    return is_key_provided_;
```
xiaomings: This could introduce extra delay if the `OnKeyStatusChange()` is called with a latency.  It's probably ok if the underlying MediaDrm doesn't introduce an latency between updating the license and this callback, which can be verified by its own experiment.

starboard/android/shared/drm_system.cc
```diff
@@ -67,11 +70,30 @@ DrmSystem::DrmSystem(
...
+        !job_queue_created_.wait_for(lock, std::chrono::seconds(1), [this] {
+          return job_queue_ != nullptr;
+        })) {
+      SB_LOG(FATAL) << "job_queue is not created in time";
+    }
+  }
 }
 
 void DrmSystem::Run() {
+  if (kEnableAppProvisioning) {
```
xiaomings: This thread is introduced in `https://github.com/youtube/cobalt/pull/1337` as an optimization, and isn't needed for app assisted individualization.  So we may just remove the `job_queue_` and process the pending requests in `UpdateSessionWithAppProvisioning()`.

starboard/android/shared/drm_system.cc
```diff
@@ -172,14 +341,55 @@ void DrmSystem::OnSessionUpdate(int ticket,
                                 SbDrmSessionRequestType request_type,
                                 std::string_view session_id,
                                 std::string_view content) {
+  std::string cdm_session_id_str;
+  std::string_view cdm_session_id;
```
xiaomings: Consider duplicating the call to `update_request_callback_()`, so we don't need `cdm_session_id_str`.

starboard/android/shared/drm_system.cc
```diff
@@ -172,14 +341,55 @@ void DrmSystem::OnSessionUpdate(int ticket,
...
+  SB_LOG(INFO) << __func__;
+  std::string cdm_session_id;
+  int ticket = kSbDrmTicketInvalid;
+  {
+    std::lock_guard lock(mutex_);
+    cdm_session_id = session_id_mapper_->GetBridgeCdmSessionId();
+    SB_DCHECK(!cdm_session_id.empty());
+
+    // Grabs first valid ticket id from pending requests.
+    for (auto& request : deferred_session_update_requests_) {
```
xiaomings: We can probably assume that the request is always the first one, and CHECK() its ticket shouldn't be `kSbDrmTicketInvalid`.

starboard/android/shared/drm_system.h
```diff
@@ -95,17 +100,31 @@ class DrmSystem : public ::SbDrmSystemPrivate,
     ~SessionUpdateRequest() = default;
 
     void Generate(const MediaDrmBridge* media_drm_bridge) const;
+    MediaDrmBridge::OperationResult GenerateWithAppProvisioning(
+        const MediaDrmBridge* media_drm_bridge) const;
+
+    void ResetTicket();
+    int ticket() const { return ticket_; }
```
xiaomings: Consider merging them into `GetAndResetTicket()`.

starboard/android/shared/drm_session_id_mapper.h
```diff
@@ -0,0 +1,52 @@
...
+
+// Manages the mapping between a CDM session ID and a MediaDrm session ID.
+// This class is thread-compatible. It is not internally thread-safe. The owner
+// of this class should handle locking to ensure thread-safety when the same
+// instance is accessed from multiple threads.
+class DrmSessionIdMapper {
+ public:
+  DrmSessionIdMapper();
+
+  std::string_view GetCdmSessionId(std::string_view media_drm_session_id) const;
```
xiaomings: Consider naming it `GetSbDrmSessionId()`, as cdm can be ambiguous in the context of this class.  The same to other places using cdm in the class.

starboard/android/shared/drm_session_id_mapper.h
```diff
@@ -0,0 +1,52 @@
...
+  std::string_view GetMediaDrmSessionId(std::string_view cdm_session_id) const;
+
+  std::string GetBridgeCdmSessionId();
```
xiaomings: GetSbDrmSessionIdForProvisioning()?

starboard/shared/starboard/media/drm_session_id_mapper.h
```diff
@@ -0,0 +1,52 @@
...
+// instance is accessed from multiple threads.
+class DrmSessionIdMapper {
+ public:
+  DrmSessionIdMapper();
+
+  std::string_view GetCdmSessionId(std::string_view media_drm_session_id) const;
+  std::string_view GetMediaDrmSessionId(std::string_view cdm_session_id) const;
+
+  std::string GetBridgeCdmSessionId();
+  void RegisterMediaDrmSessionIdIfNotSet(std::string_view media_drm_session_id);
```
xiaomings: Consider breaking this into two functions: IsMediaDrmSessionIdForProvisioningRequired() and RegisterMediaDrmSessionIdForProvisioning().

starboard/android/shared/drm_system.h
osagie98: Ack

starboard/android/shared/drm_system.cc
kjyoun: Done. used to_string method in sb/common/string.h

starboard/android/shared/drm_system.cc
kjyoun: I want to avoid the case that Ctor is blocked forever when the thread did not start for some reason. Instead, Ithe app had better crash.

Added comments about it

starboard/android/shared/drm_system.h
kjyoun: Done. and renamed to ConsumeTicket

starboard/shared/starboard/media/drm_session_id_mapper.h
kjyoun: Done

starboard/android/shared/drm_session_id_mapper.h
kjyoun: `DrmSessionIdMapper` is used by drm_system during provlsioning,
but `DrmSessionIdMapper` itself is agnostic to provisioning process.

I want to avoid mentioning Provisioning here.

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: This is to fix the not-correct behavior of exising code.

Apparently[1], this method should return true only after key is provided. 
I suspect that this is the reason  for b/418275303
DrmSystem::IsReady() returns true too early before the key is actually provided.

---
[1]
https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/media_decoder.cc;l=515-518;drc=b46247fc926e917b721734a0fb182c8f99eb7b33

```

  if (drm_system_ && !drm_system_->IsReady()) {
    // Drm system initialization is asynchronous. If there's a drm system, we
    // should wait until it's initialized to avoid errors.
    status = MEDIA_CODEC_NO_KEY;
```

starboard/android/shared/drm_system.cc
kjyoun: Done

kjyoun: Technically, it's possible to have multiple session update in flight.

I added warning log instead.

starboard/android/shared/drm_system.cc
```diff
@@ -125,11 +198,42 @@ void DrmSystem::GenerateSessionUpdateRequest(int ticket,
...
+  switch (status.status) {
+    case DRM_OPERATION_STATUS_SUCCESS:
+      return;
+    case DRM_OPERATION_STATUS_NOT_PROVISIONED:
+      SB_LOG(INFO) << "Device is not provisioned. Generating provision request";
+      {
+        std::lock_guard scoped_lock(mutex_);
+        deferred_session_update_requests_.push_back(std::move(request));
+      }
+      media_drm_bridge_->GenerateProvisionRequest();
```
xiaomings: The code could be simplified if `GenerateProvisionRequest()` returns the provision request instead of sending it via a callback.

starboard/android/shared/drm_system.cc
```diff
@@ -125,11 +198,42 @@ void DrmSystem::GenerateSessionUpdateRequest(int ticket,
   // |onSessionMessage|.
 }
 
+void DrmSystem::GenerateSessionUpdateRequestWithAppProvisioning(
+    std::unique_ptr<SessionUpdateRequest> request) {
+  SB_CHECK(kEnableAppProvisioning);
+
+  SB_LOG(INFO) << __func__;
+  MediaDrmBridge::OperationResult status =
```
xiaomings: nit: status => result

starboard/android/shared/media_drm_bridge.h
```diff
@@ -56,8 +56,8 @@ class MediaDrmBridge {
   };
 
   struct OperationResult {
-    const DrmOperationStatus status;
-    const std::string error_message;
+    DrmOperationStatus status = DRM_OPERATION_STATUS_SUCCESS;
+    std::string error_message;
```
xiaomings, line 60: Just to double check, do we remove the const so they can be modified?

starboard/android/shared/drm_system.cc
kjyoun: Yes, you are right. We can make this method work synchronoulsy.

Done

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/media_drm_bridge.h
kjyoun: Yes, that's correct. This allows the object to be modified after declaration, enabling logic like this:

```
OperationResult result;
if (kEnableAppProvisioning) {
    Log("Provisioning");
    result = doA();
} else {
    result = doB();
}
```

This change makes OperationResult a more flexible container. If immutability is desired after initialization, the caller can simply declare the object const:

```
const OperationResult result = doC();
```

starboard/android/shared/drm_session_id_mapper.h
kjyoun: Here Cdm apparenly means starboard cdm that starboard layer is using.

Is there any other Cdm that make things ambiguous?

starboard/android/shared/drm_system.cc
kjyoun: Yes, you are right. 

We don't need to run there here. and it make code simpler.

starboard/android/shared/drm_system.cc
xiaomings: This no longer applies as we no longer use the thread when `kEnableAppProvisioning` is true.

starboard/android/shared/drm_system.cc
xiaomings: A device not provisioned is a rare case compared to device provisioned.  I am not sure if this would introduce extra latency, but if it would, we will see metrics regressions for encrypted playbacks.  Note that the regression may not be obvious on general metrics as the WT on encrypted playbacks is low compared to clear playbacks.

Just try to bring this up, so feel free to keep this as is.

starboard/android/shared/drm_system.cc
xiaomings: In `GenerateSessionUpdateRequestWithAppProvisioning()`, where `deferred_session_update_requests_` isn't empty, consider assuming that there is a provisioning request in progress and directly enqueuing the request to `deferred_session_update_requests_` without `GenerateWithAppProvisioning()`.  In this case, the first request will always be the provisioning request.

starboard/android/shared/drm_system.h
xiaomings: Let's add some comments to `ConsumeTicket()`.

starboard/android/shared/drm_session_id_mapper.h
xiaomings: It can be ambiguous as the interface in Starboard is named `SbDrm`.  Technically in our ATV codebase `MediaDrm` is the cdm, and `SbDrm` is the integration.  So using `SbDrm` and `MediaDrm` explicitly is more explicit here.

starboard/android/shared/drm_session_id_mapper.h
xiaomings: If we intend to make this cross platform, then "provisioning" is a better name than "bridge", since we can assume many implementations (especially WV based) would need provisioning, but not every platform has a bridge.

starboard/android/shared/drm_system.h
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: Done. changed the logic to not create another provisioning request, if there is deferred session update request

starboard/android/shared/drm_session_id_mapper.h
kjyoun: This logic  can be used for other pre-session requests, like UpdateServerCertificate() in [drm_system_widevine.h](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/shared/widevine/drm_system_widevine.h;l=76-85;drc=e03a1ed6b1a2681ebf6ee068783d06a293910c40)

I think that `GetBridgeCdmSessionId` as a more generic name to cover any case where we need a temporary ID to "bridge" communication before the platform DRM session is created, not just provisioning.

starboard/android/shared/drm_session_id_mapper.h
kjyoun: My reasoning for using `Cdm` was to align with the terminology of the respective platforms. Within the Chromium/Starboard ecosystem, `CDM` is a well-defined term for the Content Decryption Module (link 
  (https://chromium.googlesource.com/chromium/cdm/)). Conversely, in the Android world, the term "CDM" isn't commonly used; the official component is MediaDrm[1]

  This creates a clear distinction in the code:
   * CdmSessionId: The ID used on the Chromium/Starboard side.
   * MediaDrmSessionId: The ID used on the Android platform side.

I felt this pairing was more intuitive than `SbDrm` vs. `MediaDrm`. In that pairing, "Sb" and "Media" aren't natural counterparts, which could create its own confusion.

Let me know if you see it differently.

--
[1] Top google search result for "Android Cdm" is "Companion  Device Manager"

https://screenshot.googleplex.com/9EaEhUrd2tKvHRd

starboard/android/shared/drm_system.cc
kjyoun: Yes, indeed this approach would little bit latency.

BEFORE: data begins to be fed after MediaCrypto is created
AFTER: data begins to be fed after key is provided, which may happen at least several hundreds msec after.

Changed IsReady() to return true when media crypto is created

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +223,89 @@ void DrmSystem::UpdateSession(int ticket,
...
+void DrmSystem::HandlePendingRequests() {
+  SB_LOG(INFO) << __func__;
+  std::unique_ptr<SessionUpdateRequest> request;
+  {
+    std::lock_guard scoped_lock(mutex_);
+    if (deferred_session_update_requests_.empty()) {
+      return;
+    }
+
+    request = std::move(deferred_session_update_requests_.front());
```
xiaomings: We can actually process all pending requests in one go.

starboard/android/shared/drm_system.cc
```diff
@@ -172,24 +327,67 @@ void DrmSystem::OnSessionUpdate(int ticket,
...
+                             "is no pending session update request.";
+          return kSbDrmTicketInvalid;
+        }
+
+        auto& request = deferred_session_update_requests_.front();
+        std::optional<int> valid_ticket = request->ConsumeTicket();
+        SB_CHECK(valid_ticket)
+            << "The provisioning request must have a valid ticket.";
+        return *valid_ticket;
+      }();
```
xiaomings: Can this be
```
std::lock_guard lock(mutex_);

SB_CHECK(!deferred_session_update_requests_.empty());

std::string cdm_session_id = session_id_mapper_->GetBridgeCdmSessionId();
SB_CHECK(!cdm_session_id.empty());

ticket = request->ConsumeTicket();
```

And we can CHECK in `request->ConsumeTicket()` that it should always return a valid one, and maybe give it a better name.

starboard/android/shared/drm_session_id_mapper.h
```diff
@@ -0,0 +1,54 @@
...
+// Manages the mapping between a CDM session ID and a MediaDrm session ID.
+// This class is thread-compatible. It is not internally thread-safe. The owner
+// of this class should handle locking to ensure thread-safety when the same
+// instance is accessed from multiple threads.
+class DrmSessionIdMapper {
+ public:
+  DrmSessionIdMapper();
+
+  std::string_view GetCdmSessionId(std::string_view media_drm_session_id) const;
+  std::string_view GetMediaDrmSessionId(std::string_view cdm_session_id) const;
```
xiaomings: `GetMediaDrmSessionId()` will return empty string when the MediaDrm session id hasn't been set.  Let's add a comment to make this explicit.

starboard/android/shared/drm_session_id_mapper.h
```diff
@@ -0,0 +1,54 @@
...
+#include <string>
+#include <string_view>
+
+namespace starboard {
+
+// Manages the mapping between a CDM session ID and a MediaDrm session ID.
+// This class is thread-compatible. It is not internally thread-safe. The owner
+// of this class should handle locking to ensure thread-safety when the same
+// instance is accessed from multiple threads.
+class DrmSessionIdMapper {
```
xiaomings, line 28: As discussed offline, let's move this class to //starboard/android/shared.  We can revisit the naming below after the move.

kjyoun: Done

starboard/android/shared/drm_session_id_mapper.h
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: Done and renamed to ReleaseTicket

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.h
```diff
@@ -20,16 +20,21 @@
...
 #include <string>
 #include <string_view>
 #include <unordered_map>
 #include <vector>
 
+#include "starboard/android/shared/drm_session_id_mapper.h"
 #include "starboard/android/shared/media_common.h"
 #include "starboard/android/shared/media_drm_bridge.h"
 #include "starboard/common/thread.h"
+#include "starboard/shared/starboard/player/job_queue.h"
```
xiaomings: nit: no longer used.  This is the same to some of the other new includes.

Also, let's remove the mention of JobQueue in the commit message in the github PR.

starboard/android/shared/drm_system.h
```diff
@@ -114,15 +134,27 @@ class DrmSystem : public ::SbDrmSystemPrivate,
...
+      deferred_session_update_requests_;  // Guarded by |mutex_|.
 
-  std::mutex mutex_;
   std::unordered_map<std::string, std::vector<SbDrmKeyId>> cached_drm_key_ids_;
   bool hdcp_lost_;
   std::atomic_bool created_media_crypto_session_{false};
 
   std::unique_ptr<MediaDrmBridge> media_drm_bridge_;
+
+  // Manages the mapping between the CDM session ID in the C++ layer and the
```
xiaomings: Let's move some of this comment to the class `DrmSessionIdMapper` too, so the user of that class knows that the MediaDrm session is used as is most of the time.  It's ok to keep it here (i.e. they are duplicated).

starboard/android/shared/drm_system.cc
```diff
@@ -67,11 +70,20 @@ DrmSystem::DrmSystem(
...
   }
+  SB_LOG(INFO) << "Creating DrmSystem: key_system=" << key_system
+               << ", enable_app_provisioning="
+               << to_string(kEnableAppProvisioning);
 
   Start();
 }
 
 void DrmSystem::Run() {
+  if (kEnableAppProvisioning) {
```
xiaomings: nit:  we can use `if (!kEnableAppProvisioning) Start();` above, and use `SB_CHECK(!kEnableAppProvisioning)` here.

starboard/android/shared/drm_system.cc
```diff
@@ -125,11 +163,53 @@ void DrmSystem::GenerateSessionUpdateRequest(int ticket,
...
+          << "A provisioning request is in progress. Defer this request.";
+      deferred_session_update_requests_.push_back(std::move(request));
+      return;
+    }
+  }
+
+  SB_LOG(INFO) << __func__;
+  MediaDrmBridge::OperationResult result =
+      request->GenerateWithAppProvisioning(media_drm_bridge_.get());
+  switch (result.status) {
```
xiaomings, line 187: nit: in the very end of `GenerateSessionUpdateRequest()` above, there is comment
```
// |update_request_callback_| will be called by Java calling into
// |onSessionMessage|.
```

Let's mention it somewhere in this function too, since the above function will be removed after the experiment.

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +219,88 @@ void DrmSystem::UpdateSession(int ticket,
...
+
+  std::string_view media_drm_session_id;
+  {
+    std::lock_guard lock(mutex_);
+    media_drm_session_id =
+        session_id_mapper_->GetMediaDrmSessionId(cdm_session_id);
+  }
+
+  MediaDrmBridge::OperationResult result;
+  if (media_drm_session_id.empty()) {
```
xiaomings: Consider checking `!deferred_session_update_requests_.empty()`, which is more straight-forward and the information used is local to this class.

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +219,88 @@ void DrmSystem::UpdateSession(int ticket,
...
 }
 
+void DrmSystem::UpdateSessionWithAppProvisioning(int ticket,
+                                                 const void* key,
+                                                 int key_size,
+                                                 const void* session_id,
+                                                 int session_id_size) {
+  SB_CHECK(kEnableAppProvisioning);
+  const std::string_view key_view(static_cast<const char*>(key), key_size);
+  const std::string_view cdm_session_id(static_cast<const char*>(session_id),
```
xiaomings: Consider making the names consistent, e.g. either call them `key_view` and `session_id_view`, or `cdm_key` and `cdm_session_id`.

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +219,88 @@ void DrmSystem::UpdateSession(int ticket,
...
+    SB_LOG(ERROR) << "UpdateSession failed: " << result;
+  }
+
+  session_updated_callback_(
+      this, context_, ticket,
+      result.ok() ? kSbDrmStatusSuccess : kSbDrmStatusUnknownError,
+      result.error_message.c_str(), cdm_session_id.data(),
+      cdm_session_id.size());
+
+  HandlePendingRequests();
```
xiaomings: Shall we add `if (result.ok) ...`?

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +219,88 @@ void DrmSystem::UpdateSession(int ticket,
...
+                 << cdm_session_id;
+    result = media_drm_bridge_->ProvideProvisionResponse(key_view);
+  } else {
+    result = media_drm_bridge_->UpdateSession(ticket, key_view,
+                                              media_drm_session_id);
+  }
+
+  if (!result.ok()) {
+    SB_LOG(ERROR) << "UpdateSession failed: " << result;
+  }
```
xiaomings: nit: SB_LOG_IF()

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +219,88 @@ void DrmSystem::UpdateSession(int ticket,
...
+}
+
+void DrmSystem::HandlePendingRequests() {
+  SB_LOG(INFO) << __func__;
+  std::vector<std::unique_ptr<SessionUpdateRequest>> pending_requests;
+  {
+    std::lock_guard scoped_lock(mutex_);
+    if (deferred_session_update_requests_.empty()) {
+      return;
+    }
```
xiaomings: Feel free to keep this as is so it's explicit that an empty `deferred_session_update_requests_` isn't processed.  On the other hand, the logic later can handle an empty one so it's ok to remove the check too.

starboard/android/shared/drm_system.cc
```diff
@@ -139,18 +219,88 @@ void DrmSystem::UpdateSession(int ticket,
...
+  if (kEnableAppProvisioning) {
+    std::lock_guard lock(mutex_);
+    media_drm_session_id =
+        session_id_mapper_->GetMediaDrmSessionId(session_id_as_string);
+  } else {
+    media_drm_session_id = session_id_as_string;
+  }
+
+  if (media_drm_session_id.empty()) {
+    return;
```
xiaomings, line 307: Consider adding comment "skip, as it's a provisioning session being processed, and a corresponding session in MediaDrm hasn't created" (please rephrase).  And maybe move it into the if block for `if (kEnableAppProvisioning)` so the non app provisioning path is completely unchanged,

starboard/android/shared/drm_system.cc
```diff
@@ -172,24 +322,59 @@ void DrmSystem::OnSessionUpdate(int ticket,
...
+  SB_LOG(INFO) << __func__;
+  std::string cdm_session_id;
+  int ticket;
+  {
+    std::lock_guard lock(mutex_);
+    if (deferred_session_update_requests_.empty()) {
+      SB_LOG(WARNING) << "Provisioning request is sent, even though there "
+                         "is no pending session update request.";
+      return;
+    }
```
xiaomings: Shall we SB_CHECK() this instead?

starboard/android/shared/drm_session_id_mapper.h
xiaomings: To clarify, [CDM](https://en.wikipedia.org/wiki/Encrypted_Media_Extensions#Content_Decryption_Modules) usually refers to the low level module/library that manages the licensing and decryption of the encrypted data.  There is usually integration code on top of it to handle callbacks, state management, etc..

The Chromium cdm folder shown above seems to be a software implementation of such CDM, so it has code like `aes_decryptor.h`.

The top google search results for "Android Cdm" are "Companion Device Manager" related probably due to that the abbreviation of "Companion Device Manager" also happens to be "cdm".  The results below (e.g. "implications of different CDM types") show that "MediaDrm" is a CDM.

starboard/android/shared/drm_system.h
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: Done. updated the signature of the method, so that we don' t need to define interim string_view any more

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: done

starboard/android/shared/drm_system.cc
kjyoun: Good point.

Done

starboard/android/shared/drm_system.cc
kjyoun: Good point. We don't need code that make no difference.

I prefer concise code,

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.cc
kjyoun: Done

starboard/android/shared/drm_system.h
kjyoun: Done

starboard/android/shared/drm_session_id_mapper.h
kjyoun: For the similar purpose, Chromium has media/base/android/java/src/org/chromium/media/MediaDrmSessionManager.java

They use
eme Id (Chromium) vs drm session id (Android)

Either side doesn't is Cdm. (Yes, Cdm could be drm module of both sides)

Do you want follow the same fashion?

Eme session id vs MediaDrm session id.

xiaomings: Using eme here looks good to me.

kjyoun: Done

kjyoun:
Renamed cdm session id -> eme sesson id.

https://github.com/youtube/cobalt/pull/6671/commits/b1bc7448bb69f94917411374f98ac9ab72404c67

starboard/android/shared/drm_session_id_mapper.cc
```diff
@@ -0,0 +1,74 @@
...
+#include <atomic>
+#include <string_view>
+
+#include "starboard/common/log.h"
+
+namespace starboard {
+
+DrmSessionIdMapper::DrmSessionIdMapper() = default;
+
+std::string DrmSessionIdMapper::GenerateBridgeSessionId() {
```
xiaomings: nit: add `// static` in the line above, and move the function to the bottom of the file, so it's in the same order of declaration.

starboard/android/shared/drm_system.cc
```diff
@@ -176,24 +324,57 @@ void DrmSystem::OnSessionUpdate(int ticket,
...
+
+  SB_LOG(INFO) << __func__;
+  std::string cdm_session_id;
+  int ticket;
+  {
+    std::lock_guard lock(mutex_);
+    SB_CHECK(!deferred_session_update_requests_.empty())
+        << "Provisioning request is sent, even though there is no pending "
+           "session update request.";
+    cdm_session_id = session_id_mapper_->GetBridgeEmeSessionId();
```
xiaomings: nit: `cdm_session_id` => `eme_session_id`, and consider replacing all "cdm" in the code to "eme".

starboard/android/shared/drm_session_id_mapper_test.cc
```diff
@@ -0,0 +1,90 @@
...
+  DrmSessionIdMapper mapper;
+  EXPECT_EQ(mapper.GetEmeSessionId("test_id"), "test_id");
+  EXPECT_EQ(mapper.GetMediaDrmSessionId("test_id"), "test_id");
+}
+
+TEST(DrmSessionIdMapperTest, GetBridgeEmeSessionId) {
+  DrmSessionIdMapper mapper;
+  std::string eme_session_id = mapper.GetBridgeEmeSessionId();
+  EXPECT_FALSE(eme_session_id.empty());
+  EXPECT_NE(eme_session_id, "test_id");
```
xiaomings: nit: Consider removing this check.  While "test_id" isn't a good name, it is technically a valid name for bridge session id, i.e. if the implementation always returns "test_id" for bridge session id, it's still correct.

starboard/android/shared/drm_session_id_mapper_test.cc
```diff
@@ -0,0 +1,90 @@
...
+  // The media drm session id should be empty initially.
+  EXPECT_TRUE(mapper.GetMediaDrmSessionId(eme_session_id).empty());
+}
+
+TEST(DrmSessionIdMapperTest, GetEmeSessionId) {
+  DrmSessionIdMapper mapper;
+  std::string eme_session_id = mapper.GetBridgeEmeSessionId();
+
+  const std::string media_drm_session_id = "media_drm_session_id";
+  if (mapper.IsMediaDrmSessionIdForProvisioningRequired()) {
```
xiaomings: ASSERT_TRUE(mapper.IsMediaDrmSessionIdForProvisioningRequired())?

starboard/android/shared/drm_session_id_mapper_test.cc
```diff
@@ -0,0 +1,90 @@
...
+TEST(DrmSessionIdMapperTest, GetMediaDrmSessionId) {
+  DrmSessionIdMapper mapper;
+  std::string eme_session_id = mapper.GetBridgeEmeSessionId();
+
+  const std::string media_drm_session_id = "media_drm_session_id";
+  if (mapper.IsMediaDrmSessionIdForProvisioningRequired()) {
+    mapper.RegisterMediaDrmSessionIdForProvisioning(media_drm_session_id);
+  }
+
+  EXPECT_EQ(mapper.GetMediaDrmSessionId(eme_session_id), media_drm_session_id);
```
xiaomings, line 41: Consider moving this line to the above test case, and name it "BridgeSessionIds".

starboard/android/shared/drm_session_id_mapper_test.cc
```diff
@@ -0,0 +1,90 @@
...
+  std::string eme_session_id1 = mapper.GetBridgeEmeSessionId();
+  std::string eme_session_id2 = mapper.GetBridgeEmeSessionId();
+  EXPECT_EQ(eme_session_id1, eme_session_id2);
+}
+
+TEST(DrmSessionIdMapperTest, IsMediaDrmSessionIdForProvisioningRequired) {
+  DrmSessionIdMapper mapper;
+  EXPECT_FALSE(mapper.IsMediaDrmSessionIdForProvisioningRequired());
+
+  std::string eme_id = mapper.GetBridgeEmeSessionId();
```
xiaomings: By reading this test case, it makes feel that we should rename `GetBridgeEmeSessionId()` to `CreateBridgeSession()`, and check inside that it can only be called once.

Because it's more intuitive for `IsMediaDrmSessionIdForProvisioningRequired()` to return true after a call to `CreateBridgeSession()`, but less intuitive for it to return false after a call to `GetBridgeEmeSessionId()`.

starboard/android/shared/drm_session_id_mapper_test.cc
```diff
@@ -0,0 +1,90 @@
...
+  EXPECT_TRUE(mapper.IsMediaDrmSessionIdForProvisioningRequired());
+
+  mapper.RegisterMediaDrmSessionIdForProvisioning("media_id_1");
+  EXPECT_FALSE(mapper.IsMediaDrmSessionIdForProvisioningRequired());
+}
+
+TEST(DrmSessionIdMapperTest, RegisterMediaDrmSessionIdForProvisioning) {
```
xiaomings: This test case seems to be covered by the above test cases.

starboard/android/shared/drm_system.h
```diff
@@ -116,17 +130,29 @@ class DrmSystem : public ::SbDrmSystemPrivate,
   // TODO: Update key statuses to Cobalt.
   SbDrmSessionKeyStatusesChangedFunc key_statuses_changed_callback_;
 
+  std::mutex mutex_;
+
   std::vector<std::unique_ptr<SessionUpdateRequest>>
-      deferred_session_update_requests_;
+      deferred_session_update_requests_;  // Guarded by |mutex_|.
```
xiaomings, line 136: Not in the scope of this PR. Since we have added more enforcement of ThreadChecker for this class, we can investigate further whether we can no longer guard `deferred_session_update_requests_` using `lock_`.

starboard/android/shared/drm_system.cc
```diff
@@ -141,19 +226,82 @@ void DrmSystem::UpdateSession(int ticket,
...
+  SB_LOG_IF(ERROR, !result.ok()) << "UpdateSession failed: " << result;
+
+  session_updated_callback_(
+      this, context_, ticket,
+      result.ok() ? kSbDrmStatusSuccess : kSbDrmStatusUnknownError,
+      result.error_message.c_str(), session_id.data(), session_id.size());
+
+  if (result.ok()) {
+    HandlePendingRequests();
+  }
```
xiaomings: Consider somehow only call `HandlePendingRequests()` when `ProvideProvisionResponse()` is called above, as it's non-trivial.

starboard/android/shared/drm_session_id_mapper_test.cc
kjyoun: Done

starboard/android/shared/drm_session_id_mapper_test.cc
kjyoun: Done

PR #6422, opened 2025-07-15
third_party/blink: prevent Context Menu from being rendered

ckim1346:
Bug: 430584909

Crash at: http://primes.corp.google.com/clusters/1000000001572446471

RenderFrameHostImpl::ShowContextMenu being called when it shouldn't -- the only UI we should have is the Native "No Network" error message. However, it was being called from what I suspect are touch-screen device inputs. This should prevent the ContextMenu from being called. This PR also disables some of the associated Context Menu unit tests.

PR #6675, opened 2025-08-01
third_party/blink: disabling ContextMenuController

ckim1346:
Crash at: http://primes.corp.google.com/clusters/1000000001572446471

RenderFrameHostImpl::ShowContextMenu being called when it shouldn't -- the only UI we should have is the Native "No Network" error message. However, it was being called from touch-screen device inputs. I previously had an "easy" fix to disable a boolean in #6422 but Miguel pointed out that we should disable the context menu controller entirely as we don't want context menus at all. This PR disables ContextMenuController and places it's called from as well as the associated Context Menu unit tests. This fixes the crash and prevents any context menu from being shown on a Samsung tablet (however, this was tested on an older commit 6.09.300DF as currently touch screen inputs are not working). 

Bug: 430584909

### DIFF
--- a/third_party/blink/renderer/core/exported/web_plugin_container_test.cc
+++ b/third_party/blink/renderer/core/exported/web_plugin_container_test.cc
@@ -845,8 +845,13 @@ TEST_F(WebPluginContainerTest, GestureLongPressReachesPlugin) {
       WebCoalescedInputEvent(event, ui::LatencyInfo()));
   RunPendingTasks();
 
+#if BUILDFLAG(IS_COBALT)
+  EXPECT_EQ(WebInputEvent::Type::kUndefined,
+            test_plugin->GetLastInputEventType());
+#else
   EXPECT_EQ(WebInputEvent::Type::kGestureLongPress,
             test_plugin->GetLastInputEventType());
+#endif
 }
 
 TEST_F(WebPluginContainerTest, MouseEventButtons) {
--- a/third_party/blink/renderer/core/exported/web_view_test.cc
+++ b/third_party/blink/renderer/core/exported/web_view_test.cc
@@ -3295,8 +3295,14 @@ TEST_F(WebViewTest, FinishComposingTextDoesNotDismissHandles) {
 
   EXPECT_TRUE(SimulateGestureAtElementById(
       WebInputEvent::Type::kGestureLongPress, target));
+#if BUILDFLAG(IS_COBALT)
+  // With context menu disabled, long press does not trigger selection.
+  EXPECT_EQ("", frame->SelectionAsText().Utf8());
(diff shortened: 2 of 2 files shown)


zhongqiliang:
Bug: 430584909 should be at the end of the description

zhongqiliang:
It does not seem to be right to have so many files touched with the cobalt build macro.

I think the right way to do this is to disconnect the android input event. 

Is this from a long press on the tablet? can you find the input event in this method?
https://source.corp.google.com/h/github/youtube/cobalt/+/main:content/browser/renderer_host/render_widget_host_view_android.cc;l=1349-1368?q=RenderWidgetHostViewAndroid::OnGestureEvent&sq=git:github%2Fyoutube%2Fcobalt@main, or a method in this file? Then directly return, something like this.

```
  if (event.GetType() == blink::WebInputEvent::kGestureLongPress) {
      // This prevents the event from being forwarded to the renderer.
      return;
  }
```

third_party/blink/renderer/core/page/page.h
```diff
@@ -207,9 +207,11 @@ class CORE_EXPORT Page final : public GarbageCollected<Page>,
   DragController& GetDragController() const { return *drag_controller_; }
   FocusController& GetFocusController() const { return *focus_controller_; }
   SpatialNavigationController& GetSpatialNavigationController();
+  #if !BUILDFLAG(IS_COBALT)
   ContextMenuController& GetContextMenuController() const {
     return *context_menu_controller_;
   }
+#endif
```
yell0wd0g, line 214: Haven't dug into all call sites but it seems like the core 
of this CL/PR is to not return a ContextMenuController here,
and cut out callsites to it via preprocessor, right? If so,
could this class have a different ContextMenuController, e.g.
CobaltContextMenuController, that implements CMC, doing
something sensible with the methods?

CMC is labelled final, so we'd need to change that and make
their methods virtual. It'd be a large change to be sure,
but perhaps more concentrated that the current CL?

A side comment is that most of CMC methods end up touching
ContextMenuProvider; would it be enough to replace the current
one with a CobaltContextMenuProvider to avoid the crashes
that brought us here? To be sure this path would still need
to change the declaration of said CMP, because it's also a
final class, not derivable.

WDYT?

ckim1346: I'd have to look into CMP more closely -- I do think it would be possible but I think the rationale is that we don't need/want a context menu at all in the app so adding a separate Cobalt CMC/CMP wouldn't be necessary.

yell0wd0g: I think in this case those CobaltBla classes would do nothing.
WDYT?

I'm not sure how urgent this CL is, we could consider landing
it to avoid a common crash and then fix it with a smaller
intervention later.

PR #6680, opened 2025-08-01
Remove hacks in boringssl

niranjanyardi:
Remove threading related hacks in boringssl

Issue: 398292397


third_party/boringssl/BUILD.gn
```diff
@@ -209,7 +207,7 @@ component("boringssl") {
       "//starboard/common",
     ]
   }
-  if (enable_cobalt_hermetic_hacks) {
+  if (is_starboard) {
```
niranjanyardi: Q: Is third_party/boringssl/src/crypto/thread_starboard.c a hack ?

From the code, third_party/boringssl/src/crypto/thread_starboard.c seems like a hack - it seems like its a replacement for thead_pthread.c  

thread_pthread.c vs thread_starboard.c :  https://diff.googleplex.com/#key=qIFGNKd915gd the diff looks non-trivial - so  this might not be easy to replace(or is not a hack)


wdyt ?

third_party/boringssl/src/crypto/internal.h
```diff
@@ -154,12 +148,13 @@
 #endif
 #endif
 
-#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
```
niranjanyardi, line 157: this and the below 2 chnages are related 

1) #if BUILDFLAG(IS_STARBOARD)
typedef pthread_once_t CRYPTO_once_t;
#define CRYPTO_ONCE_INIT PTHREAD_ONCE_INIT

2) #if BUILDFLAG(IS_STARBOARD)
#define CRYPTO_once pthread_once 

I'm not sure if this is ahack or done intentionally

y4vor: Why do we need to keep this customization? Can't we let it use pthread.h?

third_party/boringssl/src/crypto/internal.h
```diff
@@ -526,7 +520,7 @@ static inline int constant_time_declassify_int(int v) {
 
 // Thread-safe initialisation.
 
-#if BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
```
y4vor, line 529: can't we just use the OPENSSL_THREADS branch?

third_party/musl/src/starboard/include/pthread.h
```diff
@@ -32,6 +32,8 @@
 #define PTHREAD_PROCESS_PRIVATE 0
 #define PTHREAD_PROCESS_SHARED 1
 
+#define PTHREAD_RWLOCK_INITIALIZER {{{1, 0}}}
```
y4vor: This is strange. The pthread_rwlock_t type in this file is defined as an union. I don't see how this initializer would be used with it. Also the values don't make sense.

niranjanyardi: yea, this is the weird part. I just moved the definition from third_party/boringssl/src/crypto/internal.h , this was already there

y4vor (approved):

PR #6681, opened 2025-08-01
Add default agent prompts

Awallky:
Upstream Chromium added default prompts for gemini-cli.
This adds to the Chrobalt repo to enable more agent development options for the team.

Fixed: 435746813

### DIFF
--- a/agents/OWNERS
+++ b/agents/OWNERS
@@ -1,0 +1,4 @@
+agrieve@chromium.org
+erikchen@chromium.org
+estaab@chromium.org
+smaier@chromium.org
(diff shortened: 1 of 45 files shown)


agents/OWNERS
```diff
@@ -0,0 +1,4 @@
+agrieve@chromium.org
```
borongc, line 1: Should this PR be cherry picked from Chromium upstream instead?

Awallky: I thought about it, but the commit history shows ~20-30 CLs. This doesn't seem worth the effort to preserve the commit history temporarily when we will (eventually) get it back with our rebase efforts.

borongc: In the case like this, the PR should involve infra team members (e.g., @dahlstrom-g ), so we can better understand what's the long term way to do this.

PR #6582, opened 2025-07-25
media: Create AndroidOverlayConfig and RequestOverlayInfo

jxqu2048:
This PR creates an AndroidOverlay via AndroidOverlayConfig, which notifies StarboardRenderer when a SurfaceView is ready and lets StarboardRenderer access the SurfaceView with `overlay_->GetJavaSurface()`.

In result, StarboardRenderer will be able to pass SurfaceView to Starboard.

Issue: 429435008

PR #6682, opened 2025-08-01
media: Acquire initial viewport size for StarboardRenderer

borongc:
Pass viewport size from `ContentRendererClient` to `StarboardRenderer`, so `StarboardRenderer` can use it to set as an initial size for `AndroidOverlay`, which will be used in https://github.com/youtube/cobalt/pull/6582.

The viewport size will be updated later once `StarboardRenderer` reads video config from demuxer.

Issue: 435741374

### DIFF
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -20,6 +20,8 @@
 #include "mojo/public/cpp/bindings/generic_pending_receiver.h"
 #include "starboard/media.h"
 #include "starboard/player.h"
+#include "third_party/blink/public/web/web_view.h"
+#include "ui/gfx/geometry/size_conversions.h"
 
 namespace cobalt {
 
@@ -100,6 +102,12 @@ void CobaltContentRendererClient::RenderFrameCreated(
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   new js_injection::JsCommunication(render_frame);
   new CobaltRenderFrameObserver(render_frame);
+  if (render_frame->GetWebView()) {
+    viewport_size_ =
+        gfx::ToCeiledSize(render_frame->GetWebView()->VisualViewportSize());
+  } else {
+    LOG(WARNING) << "RenderFrameCreated is called with no webview.";
+  }
 }
 
 #if BUILDFLAG(IS_ANDROID)
@@ -188,6 +196,7 @@ void CobaltContentRendererClient::GetStarboardRendererFactoryTraits(
       base::Microseconds(kSbPlayerWriteDurationLocal);
   renderer_factory_traits->audio_write_duration_remote =
       base::Microseconds(kSbPlayerWriteDurationRemote);
+  renderer_factory_traits->viewport_size = viewport_size_;
   // TODO(b/405424096) - Cobalt: Move VideoGeometrySetterService to Gpu thread.
   renderer_factory_traits->bind_host_receiver_callback =
       base::BindPostTaskToCurrentDefault(
--- a/cobalt/renderer/cobalt_content_renderer_client.h
+++ b/cobalt/renderer/cobalt_content_renderer_client.h
@@ -13,6 +13,7 @@
 #include "cobalt/media/audio/cobalt_audio_device_factory.h"
 #include "content/public/renderer/content_renderer_client.h"
 #include "media/base/starboard/renderer_factory_traits.h"
+#include "ui/gfx/geometry/size.h"
 
 namespace content {
 class RenderFrame;
@@ -60,11 +61,12 @@ class CobaltContentRendererClient : public content::ContentRendererClient {
   // Registers a custom content::AudioDeviceFactory
   ::media::CobaltAudioDeviceFactory cobalt_audio_device_factory_;
 
-  base::WeakPtrFactory<CobaltContentRendererClient> weak_factory_{this};
-
   base::ScopedClosureRunner unregister_thread_closure;
 
+  gfx::Size viewport_size_;
+
   THREAD_CHECKER(thread_checker_);
+  base::WeakPtrFactory<CobaltContentRendererClient> weak_factory_{this};
 };
 
 }  // namespace cobalt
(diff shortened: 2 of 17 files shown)


cobalt/renderer/cobalt_content_renderer_client.h
```diff
@@ -60,11 +61,12 @@ class CobaltContentRendererClient : public content::ContentRendererClient {
   // Registers a custom content::AudioDeviceFactory
   ::media::CobaltAudioDeviceFactory cobalt_audio_device_factory_;
 
-  base::WeakPtrFactory<CobaltContentRendererClient> weak_factory_{this};
-
   base::ScopedClosureRunner unregister_thread_closure;
 
+  gfx::Size viewport_size_ = gfx::Size();
```
kjyoun: `gfx::Size viewport_size_;` isn't enough?

I think it will also create by default params

media/starboard/starboard_renderer_unittest.cc
```diff
@@ -181,9 +181,10 @@ class StarboardRendererTest : public testing::Test {
           /*overlay_plane_id=*/base::UnguessableToken::Create(),
           /*audio_write_duration_local=*/base::Seconds(1),
           /*audio_write_duration_remote=*/base::Seconds(1),
-          /*max_video_capabilities=*/""
+          /*max_video_capabilities=*/"",
+          /*viewport_size=*/gfx::Size()
 #if BUILDFLAG(IS_ANDROID)
-          ,
+              ,
```
kjyoun, line 187: is this expected?

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -100,6 +102,10 @@ void CobaltContentRendererClient::RenderFrameCreated(
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   new js_injection::JsCommunication(render_frame);
   new CobaltRenderFrameObserver(render_frame);
+  if (render_frame->GetWebView()) {
+    viewport_size_ =
+        gfx::ToCeiledSize(render_frame->GetWebView()->VisualViewportSize());
+  }
```
kjyoun, line 110: is it expected that `RenderFrameCreated` is called with no web view?

If not, you may want to log

else  {
LOG(WARNING)  << "RenderFrameCreated is called with no web view";
}

kjyoun (approved):

cobalt/renderer/cobalt_content_renderer_client.h
```diff
@@ -60,11 +61,12 @@ class CobaltContentRendererClient : public content::ContentRendererClient {
...
+
   THREAD_CHECKER(thread_checker_);
+  base::WeakPtrFactory<CobaltContentRendererClient> weak_factory_{this};
```
Awallky, line 69: nit: Did the pre-commit move this?  Seems like an unnecessary delta.

borongc: It's not, but `weak_factory_` should be the last variable in the class.

cobalt/renderer/cobalt_content_renderer_client.cc
borongc: It could be, but it's ok for the use case to not have |viewport_size_|, i.e., `0x0`, as it will be updated later.

media/starboard/starboard_renderer_unittest.cc
borongc: By `pre-commit`.

cobalt/renderer/cobalt_content_renderer_client.h
borongc: Done.

Awallky: I don't think that explicitly initializing the member is necessary, since the default gfx::Size ctor is used in both cases.

borongc: We need to init it still? https://source.corp.google.com/h/github/youtube/cobalt/+/main:ui/gfx/geometry/size.h;l=27

cobalt/renderer/cobalt_content_renderer_client.cc
borongc: Done.

cobalt/renderer/cobalt_content_renderer_client.h
Awallky: Ah, yeah makes sense after looking into the reasoning. Could you just call out why this is necessary in the description? Something like `WeakPtrFactory should be the last member in a class to avoid destruction ordering issues`.

cobalt/renderer/cobalt_content_renderer_client.h
borongc: Changed to `gfx::Size viewport_size_;`.

cobalt/renderer/cobalt_content_renderer_client.h
borongc: See https://source.chromium.org/chromium/chromium/src/+/main:base/memory/weak_ptr.h;l=22.

media/starboard/starboard_renderer.cc
```diff
@@ -128,7 +129,8 @@ StarboardRenderer::StarboardRenderer(
       buffering_state_(BUFFERING_HAVE_NOTHING),
       audio_write_duration_local_(audio_write_duration_local),
       audio_write_duration_remote_(audio_write_duration_remote),
-      max_video_capabilities_(max_video_capabilities)
+      max_video_capabilities_(max_video_capabilities),
+      viewport_size_(viewport_size)
```
xiaomings, line 133: It seems that `viewport_size_` isn't used, is it supposed to be used in a future PR?  Also, the commit message mentions that this is used to set as an initial position for `AndroidOverlay`, does it only involve position or does it also set the size?

borongc: This is for https://github.com/youtube/cobalt/pull/6582.

Note it's for `AndroidOverlay` as an initial position, assume x/y is (0, 0) and the size is from `viewport_size_`. But, this is updated once we have VideoConfig.

PR #6688, opened 2025-08-04
starboard: Add tests for Starboard Extension for Features

johnedocampo:
This PR adds testing for the Starboard Extension that adds Chromium Features to the Starboard level.

An extension_test is added to check that the extension api itself is retrievable. We also add `feature_list_test.cc`, which tests the usage of the starboard::features::FeatureList class. This class is what intializes and stores features and parameters that are used in the Starboard level. The parameter test was created by Gemini.

Currently, these tests do not run on CI as their test build target, starboard_platform_tests, is not built. However, these tests were checked locally and are verified that they passed locally.

Bug: 435467391


starboard/extension/extension_test.cc
```diff
@@ -573,5 +574,25 @@ TEST(ExtensionTest, StarboardSystemInfoExtension) {
       << "Extension struct should be a singleton";
 }
 
+TEST(ExtensionTest, StarboardFeaturesExtension) {
+  typedef StarboardExtensionFeaturesApi ExtensionApi;
```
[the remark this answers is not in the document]
johnedocampo: We'll keep this as typedef to follow the same pattern set as the other extension tests in this file.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,122 @@
...
+
+class FeatureExtensionTest : public ::testing::Test {
+ protected:
+  static void SetUpTestSuite() {
+    const StarboardExtensionFeaturesApi* extension_api =
+        static_cast<const StarboardExtensionFeaturesApi*>(
+            SbSystemGetExtension(kStarboardExtensionFeaturesName));
+
+    extension_api->InitializeStarboardFeatures(features, std::size(features),
+                                               params, std::size(params));
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/extension/extension_test.cc
```diff
@@ -573,5 +574,25 @@ TEST(ExtensionTest, StarboardSystemInfoExtension) {
...
+  const char* kExtensionName = kStarboardExtensionFeaturesName;
+
+  const ExtensionApi* extension_api =
+      static_cast<const ExtensionApi*>(SbSystemGetExtension(kExtensionName));
+  if (!extension_api) {
+    return;
+  }
```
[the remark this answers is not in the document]
johnedocampo: This pattern is established from other tests in this file, so we'll keep the pattern the same for now.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+class FeatureExtensionTest : public ::testing::Test {
+ protected:
+  static void SetUpTestSuite() {
+    const StarboardExtensionFeaturesApi* extension_api =
+        static_cast<const StarboardExtensionFeaturesApi*>(
+            SbSystemGetExtension(kStarboardExtensionFeaturesName));
+
+    // Ensure that that the extension_api is obtainable. If it is not available,
+    // we fail here to ensure that the test suite isn't ran.
+    ASSERT_NE(extension_api, nullptr)
```
jasonzhangxx: If we use `ASSERT`, when a platform doesn't support this extension, would it fail?

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+
+struct ParamTestCase {
+  std::string test_name;
+  SbFeatureParam param_definition;
+  ParamValue expected_value;
+};
+
+}  // namespace
+
+class FeatureExtensionTest : public ::testing::Test {
```
jasonzhangxx: The tests are not for feature extension, `StarboardFeatureTest` might be a better name for the test and the cc file.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+#include <iterator>
+#include <variant>
+
+#include "starboard/shared/starboard/feature_list.h"
+#include "starboard/shared/starboard/features.h"
+#include "starboard/system.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+#if BUILDFLAG(IS_ANDROID)
+#include "starboard/android/shared/features_extension.h"
```
jasonzhangxx: My understanding is this is part of platform test, but the code can be reused cross platforms, so if that's right, `BUILDFLAG(IS_ANDROID)` and any file under `starboard/android` should not be used. We should use `SbSystemGetExtension` to get the extension api.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+
+using ParamValue = std::variant<bool, int, double, std::string, int64_t>;
+
+const SbFeature FeatureTestBoolTrue = {"FeatureTestBoolTrue", true};
+const SbFeature FeatureTestBoolFalse = {"FeatureTestBoolFalse", false};
+
+const SbFeatureParam ParamTestBool = {"FeatureTestBoolTrue",
+                                      "ParamBool",
+                                      SbFeatureParamTypeBool,
+                                      {.bool_value = true}};
```
jasonzhangxx: I recommend to add one more bool test for false value.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+const SbFeatureParam ParamTestInt = {"FeatureTestBoolTrue",
+                                     "ParamInt",
+                                     SbFeatureParamTypeInt,
+                                     {.int_value = 10}};
```
jasonzhangxx: I recommend to add more int tests, like for 0 and negative value.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+const SbFeatureParam ParamTestDouble = {"FeatureTestBoolTrue",
+                                        "ParamDouble",
+                                        SbFeatureParamTypeDouble,
+                                        {.double_value = 3.14}};
```
jasonzhangxx: I recommend to add one more tests, like for 0, negative value and super large value.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+const SbFeatureParam ParamTestString = {"FeatureTestBoolTrue",
+                                        "ParamString",
+                                        SbFeatureParamTypeString,
+                                        {.string_value = "Sunny day"}};
```
jasonzhangxx: I recommend to add one more tests, like empty string, string with special symbol ("width=1920; height=1080;").

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+namespace {
+using features::FeatureList;
+using features::SbFeatureParamExt;
+
+using ParamValue = std::variant<bool, int, double, std::string, int64_t>;
+
+const SbFeature FeatureTestBoolTrue = {"FeatureTestBoolTrue", true};
+const SbFeature FeatureTestBoolFalse = {"FeatureTestBoolFalse", false};
+
+const SbFeatureParam ParamTestBool = {"FeatureTestBoolTrue",
```
jasonzhangxx: We can have two features {"StarboardFeatureTest", true}" and {"StarboardFeatureTestDisabled", false}, and reuse "StarboardFeatureTest" for param tests.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+#endif
+
+namespace starboard::shared::starboard {
+namespace {
+using features::FeatureList;
+using features::SbFeatureParamExt;
+
+using ParamValue = std::variant<bool, int, double, std::string, int64_t>;
+
+const SbFeature FeatureTestBoolTrue = {"FeatureTestBoolTrue", true};
```
jasonzhangxx: Const value should start with `k`, like `kFeatureTestBoolTrue`.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -0,0 +1,128 @@
...
+
+        EXPECT_EQ(actual_value, expected_value);
+      },
+      test_case.expected_value);
+}
+
+INSTANTIATE_TEST_SUITE_P(
+    ParamTests,
+    ParamExtensionTest,
+    ::testing::Values(
```
jasonzhangxx: It's kind of a duplicate of `params[]`. I would recommend to keep only one list, so they can be always consistent.

I think it's actually ok to use TEST_F for param test by traversing `params[]`.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: If the extension_api returns nullptr, gtest will just skip the tests in the test suite instead of failing. However, this behavior doesn't seem well documented, so this may not always be the case. I've decided to replace this method using GTEST_SKIP() instead.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Renamed the cc file to `starboard_feature_test.cc` and the test suite to `StarboardFeatureTest`.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Removed. I used the include `starboard/system.h`, which defines the SbSystemGetExtension function. If I understand correctly, this should find the correct platform for the extension.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Added int tests for 0, large positive value, negative value, and large negative value.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Added tests for large positive value, 0, negative value, and large negative value.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Added tests for empty string, and string with special symbol using the given example.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Renamed the two features to `StarboardFeatureTestEnabled` and `StarboardFeatureTestDisabled`.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Added `k` to all const values.

starboard/shared/starboard/feature_list_test.cc
johnedocampo: Done. Replaced TEST_P with TEST_F, and using `params[]` to iterate through the features.

starboard/shared/starboard/starboard_feature_test.cc
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  void SetUp() override {
+    if (!kExtension_api_) {
+      GTEST_SKIP() << "Failed to get Starboard Features API; skipping all "
+                      "tests in this suite.";
+      ;
+    }
+  }
+};
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/shared/starboard/starboard_feature_test.cc
```diff
@@ -0,0 +1,171 @@
...
+    }
+  }
+
+ private:
+  // If the extension cannot be received, all tests in this test suite are
+  // skipped.
+  static const StarboardExtensionFeaturesApi* kExtension_api_;
+};
+
+const StarboardExtensionFeaturesApi* StarboardFeatureTest::kExtension_api_ =
```
jasonzhangxx: It should be ok to only have a static 'kExtension_api` in the cc file. It doesn't have to be a static member of StarboardFeatureTest.

johnedocampo: Done.

starboard/shared/starboard/starboard_feature_test.cc
```diff
@@ -0,0 +1,165 @@
...
+  }
+};
+
+TEST_F(StarboardFeatureTest, CanAccessFeatures) {
+  EXPECT_TRUE(FeatureList::IsEnabled(kStarboardFeatureTestEnabled));
+  EXPECT_FALSE(FeatureList::IsEnabled(kStarboardFeatureTestDisabled));
+}
+
+TEST_F(StarboardFeatureTest, CanAccessParams) {
+  for (const auto& param : kParams) {
```
kjyoun, line 128: looks like you are running parameterized test.

Then how about using TEST_P?

go/googletest-testing#TEST_P

[starboard/nplb/posix_compliance/posix_file_pathconf_test.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/nplb/posix_compliance/posix_file_pathconf_test.cc;l=51;drc=d9609f1f8cf6d7e9ce07c00851386972eb03ea38)

PR #6692, opened 2025-08-04
DEBUGGING: broken CI nightly

arjungm:


cobalt/devinfra/kokoro/bin/dind_build.sh
```diff
@@ -56,6 +56,7 @@ pipeline () {
   if [[ "${TARGET_PLATFORM}" =~ "android" ]]; then
     echo "target_os=['android']" >> .gclient
   fi
+  git status
```
kaidokert: Make this a `git status -v -u` please, just in case

kaidokert: And this would fail too because this isn't in git dir :)

PR #6699, opened 2025-08-05
cobalt/shell: Remove origin trials deps

Awallky:
Origin Trials are for experimenting with Web features that are not yet ready for production. Cobalt has no need for such a feature, since it is not delivering new features for apps, but delivering a base for browsing and media playback.
    
Test: out/android-arm_devel/bin/run_cobalt_browsertests -v
Test: out/android-arm_devel/cobalt_unittests -v
Fixed: 431035934

### DIFF
--- a/cobalt/shell/common/DEPS
+++ b/cobalt/shell/common/DEPS
@@ -1,5 +1,5 @@
 include_rules = [
   "+gin/public",
   "-content/shell/common",
-  "-components/origin_trial",
+  "-components/origin_trials",
 ]


cobalt/shell/common/DEPS
```diff
@@ -1,4 +1,5 @@
 include_rules = [
   "+gin/public",
   "-content/shell/common",
+  "-components/origin_trial",
```
zhongqiliang: The files are under "components/origin_trial**s**", is this a mistake?

Awallky: Good catch! Fixed.

Awallky:
Size of libchrobalt on this branch:
`-rwxr-xr-x  1 avvall primarygroup  62399984 Aug  5 02:13 libchrobalt.so`

Size of `libchrobalt` on the main branch:
-rwxr-xr-x  1 avvall primarygroup  62428720 Aug  5 16:07 libchrobalt.so

Delta:
28736 B or 28.7 kB.

haozheng-cobalt (approved):

PR #6700, opened 2025-08-05
Update .gitignore entries for flattened repositories

rakuco:
After cpuinfo was recently imported into the main repository, I was having some issues on my Linux checkout with some files always being shown as changed in third_party/cpuinfo/src/src/arm/windows/, partly due to CRLF issues and partly because cpuinfo/src was still in third_party/.gitignore.

Comment out the .gitignore entries for most repositories that have been flattened into the tree.

While here, also add a message about this to convert_dep_to_squashed.py.

Bug: 424002414


rakuco:

kaidokert (approved):

rakuco:
@kaidokert does this PR need a second approval or can it be merged? I always forget what the actual requirements are.

yell0wd0g:
> @kaidokert does this PR need a second approval or can it be merged? I always forget what the actual requirements are.

I think for these type of low risk changes one owner+committer approval 
should be enough.

PR #6707, opened 2025-08-05
cobalt/test: Disable Tzset/Timezone Tests

Awallky:
Currently, all but a single test case fail on Cobalt ATV. This disables those tests and files a bug for investigation.

Test: out/android-arm_devel/cobalt_unittests -v
Bug: 436371274

### DIFF
--- a/cobalt/testing/filters/android-arm/cobalt_unittests_filter.json
+++ b/cobalt/testing/filters/android-arm/cobalt_unittests_filter.json
@@ -1,6 +1,5 @@
 {
   "failing_tests": [
-    "UserAgentStringTest.*",
-    "Tzset/Timezone.*"
+    "UserAgentStringTest.*"
   ]
 }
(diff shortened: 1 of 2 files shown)


cobalt/common/libc/time/tzset_test.cc
```diff
@@ -330,7 +331,13 @@ class Timezone : public TzsetTest,
   };
 };
 
-TEST_P(Timezone, SetsIcuDefaultCorrectly) {
+// TODO(b/436371274): Investigate this test failure.
```
jellefoks, line 334: We have filters for `cobalt_unittests` tests in:

cobalt/testing/filters/android-arm/cobalt_unittests_filter.json
cobalt/testing/filters/android-arm64/cobalt_unittests_filter.json

Let's keep all platform specific filtering in one place.

Awallky: As discussed offline, I have removed the JSON filter for these tests. 
I have a separate PR to remove the UserAgentStringTest.

Awallky:
@jellefoks  I created https://github.com/youtube/cobalt/pull/6727 to remove the leftover test filter for UserAgentStringTest.

PR #6708, opened 2025-08-05
Fix broken workflow diff_chromium_branches.

SortaCookie:
Removed changed the fromJson logic to where it is used after being populated. Right now it resolves to '' which causes it to error out.

Bug: 432508006


arjungm:
hey @SortaCookie I'm getting a bit spammed with update notifications, so I'm just gonna un-assign myself. Feel free to re-assign me when it's time to review it!

arjungm (approved):

PR #6711, opened 2025-08-05
Rename Cobalt shell targets names

haozheng-cobalt:
This PR updates the target names  `//cobalt/shell:content_shell_app` and `//cobalt/shell:content_shell_lib`

Test: ran app locally and played video.
Bug: 432786200

### DIFF
--- a/cobalt/shell/android/BUILD.gn
+++ b/cobalt/shell/android/BUILD.gn
@@ -32,8 +32,8 @@ shared_library("libcobalt_native_test") {
   deps = [
     "//base",
     "//base:base_javatests_lib",
-    "//cobalt/shell:content_shell_app",
-    "//cobalt/shell:content_shell_lib",
+    "//cobalt/shell:cobalt_shell_app",
+    "//cobalt/shell:cobalt_shell_lib",
     "//content/public/test/android:content_native_test_support",
   ]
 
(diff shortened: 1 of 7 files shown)


cobalt/shell/android/BUILD.gn
```diff
@@ -32,8 +32,8 @@ shared_library("libcobalt_native_test") {
   deps = [
     "//base",
     "//base:base_javatests_lib",
-    "//cobalt/shell:content_shell_app",
-    "//cobalt/shell:content_shell_lib",
+    "//cobalt/shell:cobalt_shell_app",
+    "//cobalt/shell:cobalt_shell_lib",
```
Awallky, line 36: Lower in this file, we have
`resources_package = "org.chromium.content_shell_apk"` and `content_javatests_aidl `. Can we change these to cobalt? This will probably break the cobalt_browsertests intially.
If it's too much work to make this compatible, feel free to file a bug and defer those changes to another PR. :)

haozheng-cobalt: I've noticed `resources_package = "org.chromium.content_shell_apk"` too, there are more work to do for b/432786200. I'm simply updating the 2 targets names in this PR to ease size comparison analysis.

haozheng-cobalt:
> I'd suggest for this and future PRs to test against the cobalt_browsertests target for a slight bit of extra time, but much more additional test coverage.
> 
> **Note:** Running these on a real Android device will be quite slow, but running them on emulators is actually very fast.

Thanks for the tips!

PR #6729, opened 2025-08-06
Fix(net): Return immediately when either IPv4 or IPv6 succeeds

sherryzy:
On Raspberry Pi, Cobalt QA builds were observed to hang due to IPv6 requests not failing immediately. This change modifies the TransportConnectJob to release the unsuccessful job as soon as either the IPv4 or IPv6 connection attempt succeeds. This prevents the system from waiting unnecessarily and resolves the hanging issue.

Bug: 435455755

### DIFF
--- a/net/socket/transport_connect_job.cc
+++ b/net/socket/transport_connect_job.cc
@@ -16,6 +16,7 @@
 #include "base/notreached.h"
 #include "base/task/single_thread_task_runner.h"
 #include "base/time/time.h"
+#include "build/build_config.h"
 #include "net/base/features.h"
 #include "net/base/host_port_pair.h"
 #include "net/base/ip_endpoint.h"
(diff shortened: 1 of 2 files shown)


jellefoks:
Do we understand why this is only seen on evergreen and is not an issue on Kimono on the same network? We may be fixing a symptom instead of a root cause here.

sherryzy:
@jellefoks Upon further investigation, I traced the root cause of the hang to a deep-level function within the BoringSSL library. The specific line of code is ```UniquePtr<SSL> ssl = MakeUnique<SSL>(ctx);```, which appears to be causing a deadlock. The system is likely waiting for a lock on ctx, resulting in an indefinite hang. I've also noted several unimplemented POSIX functions, such as OPENSSL_free, as documented in bug report b/412461180. It's currently unclear if this is related to the deadlock, but it's worth further exploration.

sherryzy:
But it may be good to ship this fix first to unblock other development work.

PR #6738, opened 2025-08-06
Make cobalt shell build target not testonly

haozheng-cobalt:
We can not simply remove all the testonly targets and dependencies from Cobalt Shell target yet, as it completely fails all the browsertest. The right way to do this is creating another Cobalt Shell test support target for browsertest to depend on.

Bug: 428999732


yell0wd0g (approved):
\o/

PR #6739, opened 2025-08-07
cobalt/shell: Remove services test_support

Awallky:
This change removes services/device/public/cpp:test_support, since it is unnecessary for production behavior. This probably breaks some tests in content_browsertests that will eventually be ported to Cobalt, but those have yet to be fully ported to Cobalt and any relevant test code should be refactored into a test-specific target. It should not be in the release binary.

Note: location_system_permission_status.h is the include that makes the `services/device/public/cpp` dependency necessary (for content shell). Removing this, builds work, tests pass, and user journeys seem to be working. So, it might just be a leftover artifact that is more difficult to catch than unused Java imports.

Test: out/android-x86_devel/bin/run_cobalt_browsertests -v
Test: out/android-x86_devel/cobalt_unittests -v
Fixed: 436872870

### DIFF
--- a/cobalt/shell/browser/shell_content_browser_client.cc
+++ b/cobalt/shell/browser/shell_content_browser_client.cc
@@ -85,7 +85,6 @@
 #include "mojo/public/cpp/bindings/remote.h"
 #include "mojo/public/cpp/bindings/self_owned_receiver.h"
 #include "net/ssl/client_cert_identity.h"
-#include "services/device/public/cpp/geolocation/location_system_permission_status.h"
 #include "services/network/public/cpp/features.h"
 #include "services/network/public/cpp/network_service_buildflags.h"
 #include "services/network/public/mojom/network_context.mojom.h"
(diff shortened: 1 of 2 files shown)


cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -85,7 +85,6 @@
 #include "mojo/public/cpp/bindings/remote.h"
 #include "mojo/public/cpp/bindings/self_owned_receiver.h"
 #include "net/ssl/client_cert_identity.h"
-#include "services/device/public/cpp/geolocation/location_system_permission_status.h"
```
yell0wd0g, line 88: This removal is not in the description, is it intended?

Awallky: It is. This is the include that makes the dependency necessary (for content shell). Removing this, builds work, tests pass, and user journeys seem to be working. So, it might just be a leftover artifact that is more difficult to catch than unused Java imports.

yell0wd0g: Then let's add that explanation to the description plz : )

yell0wd0g (approved):
RS LGTM with extended description.

cobalt/shell/browser/shell_content_browser_client.cc
Awallky: Done.

PR #6743, opened 2025-08-07
refactor(libc): Refactor tzset to use IcuTimeSupport

jellefoks:
This commit refactors the tzset function to improve modularity and maintainability.

The key changes are:

*   **New Architecture**: The tzset implementation has been refactored to use a new, two-tiered architecture for handling time zones. This new architecture separates concerns into:
    *   `TimeZoneState`: Manages the current time zone state from the `TZ` environment variable.
    *   `IcuTimeSupport`: A singleton that acts as the core engine, using ICU to perform time calculations.
*   **Unified Timezone Logic**: The distinction between POSIX and IANA timezones within the core implementation has been removed. This simplifies the code by removing dual-path logic and special-casing for POSIX timezones.
*   **New Tests**: Unit tests for `IcuTimeSupport` and `TimeZoneState` have been added to ensure the correctness of the new implementation. A simple test for `tzset` itself has also been added.

This is a minimal version of a larger refactoring, containing only the changes necessary for tzset.cc.

This change updates the IANA timezone tests to align with the version of the ICU data currently used in the build. Several timezones, such as 'Australia/West' and 'Mexico/BajaSur', have recently abolished daylight saving time. The test expectations are being updated to reflect the older data, which still considers these timezones to have DST.

Bug: 406083355

### DIFF
--- a/cobalt/common/libc/time/icu_time_support.cc
+++ b/cobalt/common/libc/time/icu_time_support.cc
@@ -35,14 +35,9 @@ namespace time {
 
 IcuTimeSupport* IcuTimeSupport::GetInstance() {
   static NoDestructor<IcuTimeSupport> instance;
+  cobalt::common::icu_init::EnsureInitialized();
   return instance.get();
 }
-
-IcuTimeSupport::IcuTimeSupport() {
-  cobalt::common::icu_init::EnsureInitialized();
-}
-
-IcuTimeSupport::~IcuTimeSupport() = default;
 
 void IcuTimeSupport::GetPosixTimezoneGlobals(long& out_timezone,
                                              int& out_daylight,
--- a/cobalt/common/libc/time/icu_time_support.h
+++ b/cobalt/common/libc/time/icu_time_support.h
@@ -15,7 +15,6 @@
 #ifndef COBALT_COMMON_LIBC_TIME_ICU_TIME_SUPPORT_H_
 #define COBALT_COMMON_LIBC_TIME_ICU_TIME_SUPPORT_H_
 
-#include <time.h>
 #include "cobalt/common/libc/no_destructor.h"
 #include "cobalt/common/libc/time/time_zone_state.h"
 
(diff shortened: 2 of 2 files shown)


cobalt/common/icu_init/init.cc
```diff
@@ -141,6 +143,18 @@ bool IcuInit() {
...
+  // Even though IcuInit() is called before main() is called, when the static
+  // initializers are called, the order of execution of static initializers is
+  // undefined. This function allows functions that may use ICU from static
+  // initializers themselves to guarantee that ICU is initialized first.
+  // Note: This is thread-safe because spurious calls to IcuInit() are
+  // thread-safe.
+  if (!g_icu_is_initialized) {
+    g_icu_is_initialized = IcuInit();
+  }
+}
```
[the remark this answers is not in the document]
jellefoks: This is actually safe without a mutex because it's safe to call IcuInit() from multiple threads, and this is more efficient because it avoids mutex locking in the common path.

jellefoks:
This is in preparation for followups #6748 and  #5847, to deprecate eztime and implement a fully hermetic and thoroughly tested localtime and friends.

cobalt/common/libc/time/icu_time_support.cc
```diff
@@ -0,0 +1,56 @@
...
+namespace libc {
+namespace time {
+
+IcuTimeSupport* IcuTimeSupport::GetInstance() {
+  static NoDestructor<IcuTimeSupport> instance;
+  return instance.get();
+}
+
+IcuTimeSupport::IcuTimeSupport() {
+  cobalt::common::icu_init::EnsureInitialized();
```
y4vor, line 38: Maybe call it from the GetInstance static method? It is strange to call it from the ctor when the call is supposed to be only from static initializers.

cobalt/common/libc/time/icu_time_support.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace cobalt {
+namespace common {
+namespace libc {
+namespace time {
+
+// Singleton class that provides all time conversion and timezone services.
+class IcuTimeSupport {
+ public:
+  IcuTimeSupport();
```
y4vor: We should make the ctor private.

y4vor (approved):

cobalt/common/libc/time/icu_time_support.cc
jellefoks: Done.

cobalt/common/libc/time/icu_time_support.h
jellefoks: Done.

PR #6746, opened 2025-08-07
posix: add tests for mprotect

madhurajayaraman:
Bug: 432487521

### DIFF
--- a/starboard/nplb/posix_compliance/posix_memory_protect_tests.cc
+++ b/starboard/nplb/posix_compliance/posix_memory_protect_tests.cc
@@ -19,15 +19,9 @@
   - ENOTSUP: Hard to reliably create combination of flags that are unsupported.
 */
 
-#include <fcntl.h>
 #include <sys/mman.h>
-#include <unistd.h>
 #include <cerrno>
-#include <cstring>
-#include <string>
-#include <vector>
 
-#include "starboard/common/file.h"
 #include "starboard/configuration_constants.h"
 #include "starboard/nplb/file_helpers.h"
 #include "testing/gtest/include/gtest/gtest.h"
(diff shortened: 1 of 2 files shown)


y4vor (approved):

starboard/nplb/posix_compliance/posix_memory_protect_tests.cc
```diff
@@ -0,0 +1,129 @@
...
+  void* mapped_memory_ = MAP_FAILED;
+  ScopedRandomFile scoped_file_;
+};
+
+TEST_F(PosixMprotectTest, SuccessSetsProtNone) {
+  int result = mprotect(mapped_memory_, kSbMemoryPageSize, PROT_NONE);
+  EXPECT_EQ(result, 0) << "mprotect failed: " << strerror(errno);
+}
+
+TEST_F(PosixMprotectTest, SuccessSetsProtRead) {
```
jellefoks, line 65: Can we test reading and writing after calling mprotect? e.g. when PROT_READ, we should be able to read, etc.

And when a block is protected and we access it anyway, is the error a catchable signal? In that case, let's expand the test to verify that we're getting that signal.

madhurajayaraman: added checks for reading after PROT_READ etc, but when we access a block that's protected, we crash and we don't support death tests so we can't check for that.

jellefoks: This 'crash', is that a signal we can intercept with sigaction, so that it doesn't become a death test, but we actually test that we get the signal?

https://man7.org/linux/man-pages/man2/sigaction.2.html

PR #6748, opened 2025-08-07
Add hermetic localtime, localtime_r, gmtime, gmtime_r, mktime, and timegm

jellefoks:
This adds `localtime`, `localtime_r`, `gmtime`, `gmtime_r`, `mktime`, and `timegm` implementations that use ICU. These are functions that convert between `time_t` and `struct tm` including applying the correct timezone rules, short name and time offset lookups for `localtime`, `localtime_r`, and `mktime`.

 It also adds extensive tests.

After this eztime can be deprecated. In order to not futher increase the size of this PR, that will be done in a followup PR.

Bug: 406083355

### DIFF
--- a/starboard/nplb/posix_compliance/posix_time_explosion_test.cc
+++ b/starboard/nplb/posix_compliance/posix_time_explosion_test.cc
@@ -110,7 +110,7 @@ TEST_P(PosixTimeExplosionOverflowTest, HandlesTimeOverflow) {
   const auto& function_info = std::get<0>(GetParam());
   const auto& test_data = std::get<1>(GetParam());
   // TODO: b/390675141 - Remove this after non-hermetic linux build is removed.
-#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
   if (test_data.hermetic_only) {
     GTEST_SKIP() << "Non-hermetic builds fail this test for "
                  << test_data.case_name;
--- a/starboard/nplb/posix_compliance/posix_time_implosion_test.cc
+++ b/starboard/nplb/posix_compliance/posix_time_implosion_test.cc
@@ -75,7 +75,7 @@ TEST_P(PosixTimeImplosionTest, HandlesTime) {
   const auto& function_info = std::get<0>(GetParam());
   const auto& test_data = std::get<1>(GetParam());
   // TODO: b/390675141 - Remove this after non-hermetic linux build is removed.
-#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
   if (test_data.hermetic_only) {
     GTEST_SKIP() << "Non-hermetic builds fail this test for "
                  << test_data.case_name;


jellefoks:
This builds on top of #6743 . Only the second commit is new.

cobalt/common/libc/time/icu_time_support_test.cc
```diff
@@ -40,7 +40,54 @@ class IcuTimeSupportTest : public ::testing::Test {
   const char* original_tz_ = nullptr;
 };
 
-TEST_F(IcuTimeSupportTest, GetPosixTimezoneGlobals) {
+TEST_F(IcuTimeSupportTest, GmTimeConversion) {
+  IcuTimeSupport* time_support = IcuTimeSupport::GetInstance();
+  // The timestamp being tested is:
```
andrewsavage1: If you do this test the other way, i.e. by imploding then exploding, the random number will probably need less explanation

cobalt/common/libc/time/icu_time_support.cc
```diff
@@ -33,6 +33,105 @@ namespace common {
 namespace libc {
 namespace time {
 
+namespace {
+// The maximum time_t value that doesn't overflow 'struct tm'.
+const time_t kMaxTimeValForStructTm =
```
andrewsavage1: constexpr?

cobalt/common/libc/time/icu_time_support.cc
```diff
@@ -33,6 +33,105 @@ namespace common {
...
+  out_exploded->tm_mon = calendar->get(UCAL_MONTH, status);
+  out_exploded->tm_mday = calendar->get(UCAL_DATE, status);
+  out_exploded->tm_hour = calendar->get(UCAL_HOUR_OF_DAY, status);
+  out_exploded->tm_min = calendar->get(UCAL_MINUTE, status);
+  out_exploded->tm_sec = calendar->get(UCAL_SECOND, status);
+  out_exploded->tm_wday = calendar->get(UCAL_DAY_OF_WEEK, status) - UCAL_SUNDAY;
+  out_exploded->tm_yday = calendar->get(UCAL_DAY_OF_YEAR, status) - 1;
+  out_exploded->tm_isdst = calendar->inDaylightTime(status) ? 1 : 0;
+  if (U_FAILURE(status)) {
+    status = U_ZERO_ERROR;
```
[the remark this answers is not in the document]
jellefoks: For this particular call, it is valid to pass through since we are signaling with -1 that the member value can not be determined. For the calls above though, a failure should result in a failure to convert. Check added.

cobalt/common/libc/time/icu_time_support.cc
```diff
@@ -33,6 +35,113 @@ namespace common {
...
+      new icu::GregorianCalendar(*cloned_zone, status));
+  if (!calendar || U_FAILURE(status)) {
+    return -1;
+  }
+
+  calendar->clear();
+  calendar->set(exploded->tm_year + 1900, exploded->tm_mon, exploded->tm_mday,
+                exploded->tm_hour, exploded->tm_min, exploded->tm_sec);
+
+  // Update the exploded struct with the normalized values.
```
andrewsavage1, line 123: Why are we updating the struct? Does `calendar->set` change it?

andrewsavage1 (commented):
Didn't review and of the nplb tests yet, but the actual implementation seems good other than comments

cobalt/common/libc/time/icu_time_support.cc
jellefoks: `mktime` and `timegm` require the input time to be 'normalized', which is what the comment alludes to. For `mktime` that requirement is in POSIX. For `timegm` it is in C23 (ISO/IEC  9899/2024, see draft N3220 section 7.27.2.4), which doesn't use the term 'normalization' but uses other wording for the same effect. 

This is tested in `HandlesNormalization` in posix_compliance. I'll make those tests more thorough because they didn't test all normalized members, or overflow directions. I'll also add two here in icu_time_support_test.cc.

cobalt/common/libc/time/icu_time_support.cc
jellefoks: Done.

cobalt/common/libc/time/icu_time_support_test.cc
jellefoks: Done.

starboard/nplb/posix_compliance/posix_time_gmtime_r_test.cc
andrewsavage1, line 1: I'm sorry in advance for this ask, but it seems like many of these tests have similar bodies and should be parameterized, both in this file and a number of the other added test families

jellefoks: I was torn between what the readability of a test is when a test is parametrized versus copied code since parametrization does replace those plain literals with `param.foo` which may not be considered equally readable. I can go either way. I'm not a fan of copy-pasted code, but I'm also not a fan of a unit test where reading the code for a failing test requires reading too many lines of code from outside of the test body.

Some of the `Handles*Time*` tests differ only in values, as well the `Handles*Overflow`. Those should be straight-forward enough to parametrize.

The tests also have similarities across files, e.g. `PosixGmtimeTests.HandlesEpochTime` and `PosixGmtimeRTests, HandlesEpochTime` differ only in whether `gmtime` is called or `gmtime_r`.

Are you also suggesting making the operation itself a parameter, so that `localtime` + `localtime_r` and `gmtime` + `gmtime_r` all become covered by a single `posix_time_exploding_test.cc` (and similar for the imploding `mktime` and `timegm`).

andrewsavage1: For me it'll be the right balance of readable/not too repetitive if we just go one level of parameterization deep, i.e. just making the time_t and struct tm values the parameters. Something like the following

```
    testing::Values(
        TimeTestParams{
            "HandlesEpochTime",
            0,
            CreateExpectedTm(0, 0, 0, 1, 0, 70, 4, 0, 0)
        },
        ...
```
(or using designated initializers for struct tm instead)

jellefoks: The latest commit implements this.

starboard/nplb/posix_compliance/posix_time_explosion_test.cc
```diff
@@ -0,0 +1,524 @@
...
+  } else {
+    EXPECT_STREQ(result_tm_ptr->tm_zone, test_data.expected_zone);
+  }
+}
+
+TEST_P(PosixTimeExplosionOverflowTest, HandlesTimeOverflow) {
+  const auto& function_info = std::get<0>(GetParam());
+  const auto& test_data = std::get<1>(GetParam());
+  // TODO: b/390675141 - Remove this after non-hermetic linux build is removed.
+#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
```
andrewsavage1: Use the hermetic hacks flag so we know to remove this section/field once we remove non-hermetic builds

andrewsavage1 (approved):

starboard/nplb/posix_compliance/posix_time_explosion_test.cc
jellefoks: Done

PR #6750, opened 2025-08-07
posix: add tests for realloc

madhurajayaraman:
Bug: 432487521

### DIFF
--- a/starboard/nplb/posix_compliance/posix_memory_reallocate_test.cc
+++ b/starboard/nplb/posix_compliance/posix_memory_reallocate_test.cc
@@ -26,22 +26,23 @@
 #include <limits>
 #include <vector>
 
+#include "starboard/configuration_constants.h"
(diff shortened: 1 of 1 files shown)


starboard/nplb/posix_compliance/posix_memory_reallocate_test.cc
```diff
@@ -0,0 +1,106 @@
...
+#include <limits>
+#include <vector>
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace starboard::nplb {
+namespace {
+
+TEST(PosixReallocTest, NullPtrIsLikeMalloc) {
+  void* mem = realloc(NULL, 128);
```
jellefoks: Let's use larger sizes, multiples of the system page size for these tests, since page faults resulting from allocations not being complete can only be detected when it goes over a page boundary.

jellefoks (approved):

PR #6754, opened 2025-08-08
cobalt/shell: Remove unused components/services/storage/test_api

Awallky:
This seems unnecessary in Cobalt and unused. This removes these deps from content_shell_lib.

Test: out/android-x86_devel/bin/run_cobalt_browsertests
Test: out/android-x86_devel/cobalt_unittests
Test: basic user journey for browse and playback

Bug: 437204307


Awallky (commented):
Per Hao and I's offline conversation, we will want to use preprocessor directives to compile these out of production builds but keep them in for various test suites.
Will resolve this when I have added these changes.

Awallky:
No longer necessary now that the utility files have been moved to browser-test only builds.

PR #6755, opened 2025-08-08
cobalt/shell: Remove unused content_test_mojo_bindings

Awallky:
This seems unnecessary in Cobalt and unused. This removes these deps from content_shell_lib.

Test: out/android-x86_devel/bin/run_cobalt_browsertests
Test: out/android-x86_devel/cobalt_unittests
Test: basic user journey for browse and playback

Bug: 437205827

### DIFF
--- a/starboard/android/shared/player_components_factory.cc
+++ b/starboard/android/shared/player_components_factory.cc
@@ -74,12 +74,6 @@ constexpr bool kForceSecurePipelineInTunnelModeWhenRequired = true;
 // video distortion on some platforms.
 constexpr bool kForceResetSurfaceUnderTunnelMode = true;
 
-// By default, Cobalt restarts MediaCodec after stops/flushes during
-// Reset()/Flush(). Set the following variable to > 0 to force it to
-// wait during Reset()/Flush().
-constexpr int64_t kResetDelayUsecOverride = 0;
-constexpr int64_t kFlushDelayUsecOverride = 0;
-
 // This class allows us to force int16 sample type when tunnel mode is enabled.
 class AudioRendererSinkAndroid : public ::starboard::shared::starboard::player::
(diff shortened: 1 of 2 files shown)


Awallky:
Per Hao and I's offline conversation, we will want to use preprocessor directives to compile these out of production builds but keep them in for various test suites.
Will resolve this when I have added these changes.

cobalt/shell/BUILD.gn
```diff
@@ -109,6 +108,10 @@ group("cobalt_shell_lib_deps") {
     "//v8",
   ]
 
+  if (run_browser_tests) {
```
haozheng-cobalt: You could remove this.  This dependency is already in `browsertests_sources`.

And please add all test dependencies in the `browsertests_sources` in the fure.

haozheng-cobalt (changes requested):

cobalt/shell/BUILD.gn
Awallky: Good catch! Done.

cobalt/shell/utility/shell_content_utility_client.cc
```diff
@@ -38,7 +38,6 @@
 #include "content/public/child/child_thread.h"
```
haozheng-cobalt: Please do not modify this file, this file is not compiled into official build anymore.

Any dependencies in `browsertests_sources` ([link](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/shell/BUILD.gn;l=190?q=cobalt%2Fshell%2Fbuild.gn)) will not be included in official build ([link](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/shell/BUILD.gn;l=286?q=cobalt%2Fshell%2Fbuild.gn)).

PR #6758, opened 2025-08-08
This change can avoid crashes caused by exiting the process due to poor network and URL leaks

Jimmy-ge:
avoid networkmodule crash in gold version


Jimmy-ge:
https://partnerissuetracker.corp.google.com/u/0/issues/409716671

PR #6759, opened 2025-08-08
net: Restrict Cobalt trust store code to Linux

rakuco:
The changes in #6602 broke the tvOS build for a few reasons:

* `test_root_certs_builtin.cc` is also added to the build in an `if (is_mac)` block, which causes GN to complain.
* `SystemTrustStoreChromeOnly::chrome_root_store_version()` does not exist on tvOS because `CHROME_ROOT_STORE_SUPPORTED` is not set.

Since this code seems to be trying to solve a Linux-specific problem, restrict the changes to `net/BUILD.gn` to `is_linux`.

Bug: 432077378

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
@@ -40,16 +40,16 @@ public class AudioTrackBridge {
   // Also used by AudioOutputManager.
   static final int AV_SYNC_HEADER_V1_SIZE = 16;
 
-  private AudioTrack audioTrack;
-  private AudioTimestamp audioTimestamp = new AudioTimestamp();
-  private final Object positionLock = new Object();
-  @GuardedBy("positionLock")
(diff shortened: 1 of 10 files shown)


rakuco:

rakuco:
The `FileTest.GetInfoForCreationTime` failures look unrelated, this PR only touches `net`.

arjungm (approved):

arjungm:
good catch, sorry that I missed checking in on tvOS builds for this one

rakuco:
@yell0wd0g I don't have access to the list of members of the cobalt-3p-repository-owners team. Could you CC someone for the missing approval here?

yell0wd0g:
Weird, I thought when are team (cobalt-3p-repository-owners in this 
case) is added as reviewer, it automatically picks up one member.
Maybe it's because @y4vor is part of that team.

yell0wd0g (approved):
LGTM - let's see if my +1 is enough.

rakuco:
> LGTM - let's see if my +1 is enough.

Looks like it was, but I force-pushed earlier today and it disabled the auto-merge feature :/

PR #6760, opened 2025-08-08
refactor: Replace starboard::ConditionVariable with std::condition_variable

kjyoun:
This is part of efforts to deprecate `starboard::Mutex` and `starboard::ConditionVariable`.
 
#vibe-coded
   
Bug: 390503926


y4vor (approved):

kjyoun:
reverted changes in semaphore.h and .cc. It has side effect of breaking nplb test.

Will create a separate PR to address it

PR #6761, opened 2025-08-08
Evergreen: replace tts_linux to tts_starboard stub

johnedocampo:
This PR adds tts_starboard.cc, which stubs in the
TtsPlatformImplBackgroundWorker and TtsPlatformImplLinux classes from tts_linux.cc. The tts_starboard.cc class was created with Gemini.

This allows us to remove our hermetic hack in tts_platform_impl.cc, which was causing crashes on our evergreen builds. This hack originally just returned nullptr when the instance of the tts implementation was called, causing issues most likely due to code assuming that this value isn't a nullptr.

Follow up PRs will properly implement this file.

Bug: 420913744
Bug: 437117010

### DIFF
--- a/content/browser/BUILD.gn
+++ b/content/browser/BUILD.gn
@@ -2439,11 +2439,19 @@ source_set("browser") {
   }
 
   if (is_linux) {
+    if (is_starboard) {
+      sources += [ "speech/tts_starboard.cc" ]
+      deps += [
+        "//ui/linux:linux_ui",
+      ]
+    }
+    else {
     sources += [ "speech/tts_linux.cc" ]
     deps += [
       "//third_party/speech-dispatcher",
       "//ui/linux:linux_ui",
     ]
+    }
   }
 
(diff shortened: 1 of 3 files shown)


content/browser/BUILD.gn
```diff
@@ -2439,11 +2439,19 @@ source_set("browser") {
   }
 
   if (is_linux) {
+    if (is_cobalt_hermetic_build) {
```
johnedocampo: Is `is_cobalt_hermetic_build` correct here? Or should it be `is_starboard`?

y4vor: try using `is_starboard`. Only if the monolithic build `linux-x64x11` has issues we would consider `is_cobalt_hermetic_build`.

y4vor (approved):

content/browser/BUILD.gn
johnedocampo: Done.

PR #6764, opened 2025-08-08
Define browsertests preprocessor

haozheng-cobalt:
Define browsertests preprocessor 

Bug: 437186043

Test: Set gn args run_browser_tests to true to run browsertests.

### DIFF
--- a/cobalt/testing/browser_tests/BUILD.gn
+++ b/cobalt/testing/browser_tests/BUILD.gn
@@ -188,7 +188,12 @@ test("cobalt_browsertests") {
 
   sources = [
     "content_main_runner_impl_browsertest.cc",
+    "frame_tree_browsertest.cc",
     "navigation_browsertest.cc",
+    "session_history_browsertest.cc",
+    "site_per_process_browsertest.cc",
+    "site_per_process_browsertest.h",
+    "web_contents_impl_browsertest.cc",
   ]
 
   if (is_android) {
--- a/.github/workflows/gemini-commit-message.yml
+++ b/.github/workflows/gemini-commit-message.yml
@@ -1,0 +1,118 @@
+name: 'Gemini Commit Message Generator'
+
+# This workflow is triggered when a new comment is created on a Pull Request.
+on:
+  issue_comment:
+    types: [created]
+
+# This block defines the job that will run.
+jobs:
+  generate_commit_message:
+    # This condition ensures the workflow only runs if the comment is on a PR
+    # and the comment body is exactly '/gemini-commit-message'.
+    if: github.event.issue.pull_request && github.event.comment.body == '/gemini-commit-message'
+
+    # The workflow will run on the latest version of Ubuntu.
+    runs-on: ubuntu-latest
+
+    # We need to grant the workflow permission to write comments on pull requests.
+    permissions:
+      pull-requests: write
+
+    steps:
+      # Step 1: Get the code changes (the "diff") from the pull request.
+      - name: 'Get PR Diff'
+        id: diff
+        run: |
+          set -eux
+          # Fetch the diff content directly from the PR's diff URL.
+          diff_content=$(curl -fsSL -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" -H "Accept: application/vnd.github.v3.diff" ${{ github.event.issue.pull_request.url }})
+
+          # The following lines save the diff content into a GitHub Actions output variable.
+          echo "diff_content<<EOF" >> $GITHUB_OUTPUT
+          echo "$diff_content" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+      # Step 2: Call the Gemini API directly with a detailed prompt.
+      # This approach avoids third-party actions and gives us full control over the request.
+      - name: 'Generate Commit Message with Gemini'
+        id: gemini
+        env:
+          # Store the multiline prompt in an environment variable to handle special characters safely.
+          PROMPT: |
+            Act as an expert software engineer specializing in the Cobalt codebase, which is a fork of Chromium. Your task is to generate a professional and informative Git commit message based on the provided pull request details. The final output should be only the commit message itself, without any extra conversational text or markdown formatting. Do not use backticks (`) in your response.
+
+            You must strictly adhere to the following rules:
+
+            **Commit Message Structure:**
+            1.  **Tag Prefix:** The subject line MUST be prefixed with a tag followed by a colon (e.g., "media: Add support for AV1"). Prefer component tags over type tags.
+            2.  **Subject:** Capitalize the subject line, use the imperative mood, limit it to 50 characters, and do not end it with a period.
+            3.  **Body:** Separate the subject from the body with a blank line. The body should explain the 'what' and 'why' of the change, not the 'how', and wrap at 72 characters.
+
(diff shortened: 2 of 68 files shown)


cobalt/shell/BUILD.gn
```diff
@@ -276,6 +278,10 @@ static_library("cobalt_shell_lib") {
     defines += [ "SHELL_USE_TOOLKIT_VIEWS=1" ]
   }
 
+  if (support_browser_tests) {
+    defines += [ "SUPPORT_BROWSER_TESTS=1" ]
```
Awallky: Why don't we set this [here](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/testing/browser_tests/BUILD.gn;drc=5fdf801992b94685ddc2aa4bf78399b41cc35f45;l=198)?

cobalt/shell/BUILD.gn
```diff
@@ -26,6 +26,8 @@ shell_use_toolkit_views = toolkit_views && !is_castos
 # Web test support only built for the linux-x64x11-no-starboard target.
 support_web_tests = !is_android && !is_starboard
 
+support_browser_tests = is_android && !is_gold
```
Awallky: This also means that we cannot run cobalt browser tests on gold builds.

Awallky: Let's only set this to true when we are building `cobalt_browsertests`.

cobalt/shell/BUILD.gn
haozheng-cobalt: This preprocessor will act on source files in cobalt_shell_lib, see follow-up PR https://github.com/youtube/cobalt/pull/6767/files

Awallky: There is a standard way to achieve this in GN by using public_configs. The core idea is that you cannot have a setting in a target propagate "up" to its dependencies. Instead, you create a new "bridging" target that applies a special configuration to its dependencies publicly.

Here is the step-by-step solution:

**The GN Solution: Public Configs and a Bridging Target**

You will create a new source_set target within `cobalt/testing/browser_tests/BUILD.gn`. This new target will depend on the real `//cobalt/shell:cobalt_shell_lib` and publicly apply a new configuration that contains the preprocessor define. The `cobalt_browsertests` will then depend on this new bridging target instead of the original library.

  ---

**Step 1: Define a config in cobalt/testing/browser_tests/BUILD.gn**
This config will hold the preprocessor definition that you want to use in your C++ code.

   1 # In //cobalt/testing/browser_tests/BUILD.gn
   2 
   3 config("cobalt_browser_tests_config") {
   4   # This define will be active for any target that has this config applied.
   5   defines = [ "COBALT_FOR_BROWSER_TESTS=1" ]
   6 }

  ---

**Step 2: Create a Bridging source_set**
Still in cobalt/testing/browser_tests/BUILD.gn, create a new source_set target. This target doesn't have any sources itself; its only job is to link the real library with your new test-only configuration.

    1 # In //cobalt/testing/browser_tests/BUILD.gn
    2 
    3 # This target acts as a bridge. It depends on the real cobalt_shell_lib
    4 # and publicly applies our test-only configuration to it.
    5 source_set("cobalt_shell_lib_with_test_hooks") {
    6   # This makes our new target depend on the original library.
    7   public_deps = [
    8     "//cobalt/shell:cobalt_shell_lib",
    9   ]
   10 
   11   # This is the key part. The public_configs from this target will be
   12   # applied to any target that depends on it (i.e., cobalt_browsertests),
   13   # AND to the public_deps listed above.
   14   public_configs = [
   15     ":cobalt_browser_tests_config",
   16   ]
   17 }

  ---

**Step 3: Update the cobalt_browsertests Target**
Now, modify your cobalt_browsertests target to depend on the new `:cobalt_shell_lib_with_test_hooks` instead of the original `//cobalt/shell:cobalt_shell_lib`.

    1 # In //cobalt/testing/browser_tests/BUILD.gn
    2 
    3 test("cobalt_browsertests") {
    4   deps = [
    5     # Replace the direct dependency with our new bridging target.
    6     ":cobalt_shell_lib_with_test_hooks",
    7 
    8     # ... other dependencies ...
    9   ]
   10   # ... other settings ...
   11 }

  ---


**Step 4: Use the Preprocessor Directive in C++**

Finally, in the source files of `cobalt_shell_lib`, you can now use the preprocessor directive. This code will only be compiled when `cobalt_shell_lib` is being built as a dependency of `cobalt_browsertests`.


    1 // In a file inside //cobalt/shell/...
    2 
    3 #ifdef COBALT_FOR_BROWSER_TESTS
    4 // This include and any related code will only be compiled for browser tests.
    5 #include "cobalt/testing/test_class.h"
    6 #endif
    7 
    8 void SomeFunctionInShellLib() {
    9   // ...
   10 #ifdef COBALT_FOR_BROWSER_TESTS
   11   MyTestClass::DoSomethingForTest();
   12 #endif
   13   // ...
   14 }

This approach correctly uses GN's dependency and configuration system to conditionally compile code without needing to modify the original cobalt_shell_lib target directly.

cobalt/shell/BUILD.gn
haozheng-cobalt: Thanks, agree that your proposal is way better. I'll move it to cobalt_browsertests.

cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -179,6 +179,7 @@ if (is_android) {
 
 test("cobalt_browsertests") {
   use_xvfb = use_xvfb_in_this_config
+  run_browser_tests = true
```
haozheng-cobalt: link

cobalt/shell/BUILD.gn
haozheng-cobalt: Defining the MACRO when only building the cobalt_browsertests would involve defining a new template args which is an overkill and really aggressive. I would only define the the param at the args.gn level, and it can be set when updating gn args locally.

cobalt/build/configs/hacks.gni
```diff
@@ -16,10 +16,12 @@ declare_args() {
...
-}
 
-# TODO(b/403589162): Remove this block once we figure out why the build
-# complains about siso.
-declare_args() {
+  # TODO(b/403589162): Remove this block once we figure out why the build
+  # complains about siso.
   use_siso = false
+
+  # TODO(b/428999732): Cobalt shell was started as a minial test shell. We use
```
Awallky: Can you add a TODO here for refactoring the test code out of the shell?

Awallky: And point to the bug you filed for it?

Awallky (approved):
I can see a way to make this work. 
LGTM % minor comment.

cobalt/build/configs/hacks.gni
haozheng-cobalt: Bug was already added.

cobalt/build/configs/initialize_variables.gni
```diff
@@ -32,6 +32,10 @@ declare_args() {
   build_with_separate_cobalt_toolchain = target_os != "android"
 
   use_evergreen = false
+
+  # TODO(b/428999732, b/437999182): Cobalt shell was started as a minial test shell.
```
andrewsavage1, line 36: nit: typo in minimal

haozheng-cobalt: thx, addressed in another PR.

PR #6766, opened 2025-08-08
Proper fix for the DidMeaningfulLayout crash on raspi2

y4vor:
The CobaltRenderFrameObserver didn't check the starboard graphics extension version before calling ReportFullyDrawn() which is only implemented in version 6. raspi2 currently implements version 3.

Issue: 432242322

### DIFF
--- a/cobalt/renderer/cobalt_render_frame_observer.cc
+++ b/cobalt/renderer/cobalt_render_frame_observer.cc
@@ -53,7 +53,7 @@ void CobaltRenderFrameObserver::DidMeaningfulLayout(
             SbSystemGetExtension(kCobaltExtensionGraphicsName));
     if (graphics_extension &&
         strcmp(graphics_extension->name, kCobaltExtensionGraphicsName) == 0 &&
-        graphics_extension->version >= 1) {
+        graphics_extension->version >= 6) {
       graphics_extension->ReportFullyDrawn();
     }
   }
--- a/third_party/blink/renderer/core/frame/web_frame_widget_impl.cc
+++ b/third_party/blink/renderer/core/frame/web_frame_widget_impl.cc
@@ -2987,10 +2987,6 @@ void WebFrameWidgetImpl::AutoscrollEnd() {
 }
 
 void WebFrameWidgetImpl::DidMeaningfulLayout(WebMeaningfulLayout layout_type) {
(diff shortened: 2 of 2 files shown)


cobalt/renderer/cobalt_render_frame_observer.cc
```diff
@@ -53,7 +53,7 @@ void CobaltRenderFrameObserver::DidMeaningfulLayout(
             SbSystemGetExtension(kCobaltExtensionGraphicsName));
     if (graphics_extension &&
         strcmp(graphics_extension->name, kCobaltExtensionGraphicsName) == 0 &&
```
yell0wd0g, line 55: Just a thought unrelated to this CL: if we're asking for the Graphics
extension in l.53, shouldn't this be an SB_CHECK()? Or is there a legit
use case in which `SbSystemGetExtension()` returns a non-null extension
different from the requested one..?

cobalt/renderer/cobalt_render_frame_observer.cc
```diff
@@ -53,7 +53,7 @@ void CobaltRenderFrameObserver::DidMeaningfulLayout(
...
-        graphics_extension->version >= 1) {
+        graphics_extension->version >= 6) {
```
yell0wd0g, line 56: This is a bit of magic numbers :) -- isn't there a way to 
query the extension system to see whether a given extension
supports method Bla()? (Some sort of introspection, perhaps).

yell0wd0g (approved):
LGTM to unblock this which a great improvement!
Some Qs inline.

cobalt/renderer/cobalt_render_frame_observer.cc
y4vor: We don't have that. The practice we have is to gate API calls by the version they are introduced.

cobalt/renderer/cobalt_render_frame_observer.cc
y4vor: I think those checks around the name are a bit paranoia about partners having a misconfiguration. If we want to re-evaluate the pattern we can do it holistically and update all the extensions calls. Having said that I'm not a fan of the CHECKs. Most extensions are not critical and would not crash the runtime if they are missing. Having a warning/error in the logs would be nice.

PR #5565, opened 2025-04-25
[android] Update calculation of max input size

jasonzhangxx:
Update the code of max input buffer size calculation according to the latest exoplayer code cl/467641494.

b/401431027

PR #6769, opened 2025-08-08
media: Add max video size calculation experiment

johnedocampo:
This PR aims to re-add the code from [pr/5565](https://github.com/youtube/cobalt/pull/5565), which was closed due to needing to be guarded by an experiment. With the ability to place feature experiments inside Starboard now possible, we guard the code from pr/5565 with a Starboard experiment.

When the starboard feature is enabled, we will calculate the max video input size according to the latest exoplayer code from cl/467641494. When the feature is disabled, we will use our current method of calculating the max video input size. This will allow us to do some A/B testing to see which implementation is better.

Bug: 413696596

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
@@ -417,6 +417,7 @@ class MediaCodecBridge {
       ColorInfo colorInfo,
       int tunnelModeAudioSessionId,
       int maxVideoInputSize,
+      boolean maxVideoInputSizeExperiment,
       CreateMediaCodecBridgeResult outCreateMediaCodecBridgeResult) {
     MediaCodec mediaCodec = null;
     outCreateMediaCodecBridgeResult.mMediaCodecBridge = null;
@@ -589,7 +590,7 @@ class MediaCodecBridge {
       }
     }
     if (!bridge.configureVideo(
-        mediaFormat, surface, crypto, 0, maxWidth, maxHeight, outCreateMediaCodecBridgeResult)) {
+        mediaFormat, surface, crypto, 0, maxWidth, maxHeight, maxVideoInputSizeExperiment, outCreateMediaCodecBridgeResult)) {
       Log.e(TAG, "Failed to configure video codec.");
       bridge.release();
       // outCreateMediaCodecBridgeResult.mErrorMessage is set inside configureVideo() on error.
@@ -831,6 +832,7 @@ class MediaCodecBridge {
       int flags,
       int maxSupportedWidth,
       int maxSupportedHeight,
+      boolean maxVideoInputSizeExperiment,
       CreateMediaCodecBridgeResult outCreateMediaCodecBridgeResult) {
     try {
       // Since we haven't passed the properties of the stream we're playing down to this level, from
@@ -848,7 +850,7 @@ class MediaCodecBridge {
         format.setInteger(MediaFormat.KEY_MAX_HEIGHT, Math.min(2160, maxSupportedHeight));
       }
 
-      maybeSetMaxVideoInputSize(format);
+      maybeSetMaxVideoInputSize(format, crypto, maxVideoInputSizeExperiment);
       mMediaCodec.get().configure(format, surface, crypto, flags);
       mFrameRateEstimator = new FrameRateEstimator();
       return true;
@@ -884,11 +886,19 @@ class MediaCodecBridge {
     int ceilDivide = (size + alignment - 1) / alignment;
     return ceilDivide * alignment;
   }
-
+  /**
+   * Returns the maximum sample size assuming three channel 4:2:0 subsampled input frames with the
+   * specified {@code minCompressionRatio}
+   * @param pixelCount The number of pixels
+   * @param minCompressionRatio The minimum compression ratio
+   */
+  private static int getMaxSampleSize(int pixelCount, int minCompressionRatio) {
+    return (pixelCount * 3) / (2 * minCompressionRatio);
+  }
   // Use some heuristics to set KEY_MAX_INPUT_SIZE (the size of the input buffers).
   // Taken from ExoPlayer:
   // https://github.com/google/ExoPlayer/blob/8595c65678a181296cdf673eacb93d8135479340/library/src/main/java/com/google/android/exoplayer/MediaCodecVideoTrackRenderer.java
-  private void maybeSetMaxVideoInputSize(MediaFormat format) {
+  private void maybeSetMaxVideoInputSize(MediaFormat format, MediaCrypto crypto, boolean maxVideoInputSizeExperiment) {
     if (format.containsKey(android.media.MediaFormat.KEY_MAX_INPUT_SIZE)) {
       try {
         Log.i(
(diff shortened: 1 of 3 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -907,36 +917,75 @@ private void maybeSetMaxVideoInputSize(MediaFormat format) {
...
     int minCompressionRatio;
-    switch (format.getString(MediaFormat.KEY_MIME)) {
-      case MimeTypes.VIDEO_H264:
-        if ("BRAVIA 4K 2015".equals(Build.MODEL)) {
-          // The Sony BRAVIA 4k TV has input buffers that are too small for the calculated
-          // 4k video maximum input size, so use the default value.
+    // If |maxVideoInputSizeExperiment| is true, we'll calculate the input size
+    // according to the latest exoplayer code from cl/467641494.
+    // If set to false, we'll use our original calculation method.
+    if (maxVideoInputSizeExperiment) {
```
johnedocampo: I thought about refactoring this part of the code to have one switch block with each case performing different actions depending on if the flag was enabled, but I ultimately decided to separate our current implementation from the exoplayer implementation for two reasons:

1. To preserve each implementation; I think it's important to keep the exoplayer code the same as where it was originally adapted from, and refactoring would lose this preservation. Same goes for the current implementation.
2. When we decide which one is better to keep, it should be easy to just delete the block of code we won't use anymore.

Please let me know what you think and whether it would be better to refactor this code or to keep it the same.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -907,36 +917,56 @@ private void maybeSetMaxVideoInputSize(MediaFormat format) {
...
+              return;
+            }
+        }
+        else {
+          if ("BRAVIA 4K 2015".equals(Build.MODEL)) {
+            // The Sony BRAVIA 4k TV has input buffers that are too small for the calculated
+            // 4k video maximum input size, so use the default value.
+            return;
+          }
         }
```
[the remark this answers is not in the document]
johnedocampo: We may want to keep this as-is to maintain the same exact functionality of the original implementation.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -910,36 +920,56 @@ private void maybeSetMaxVideoInputSize(MediaFormat format) {
...
+    // according to the latest exoplayer code from cl/467641494.
+    // If set to false, we'll use our original calculation values.
     switch (format.getString(MediaFormat.KEY_MIME)) {
       case MimeTypes.VIDEO_H264:
-        if ("BRAVIA 4K 2015".equals(Build.MODEL)) {
-          // The Sony BRAVIA 4k TV has input buffers that are too small for the calculated
-          // 4k video maximum input size, so use the default value.
-          return;
+        if (experimentalVideoSizeCalculation) {
+          if ("BRAVIA 4K 2015".equals(Build.MODEL) // Sony Bravia 4K
```
johnedocampo, line 931: @jasonzhangxx I know there were efforts made in the original PR of this to preserve how the code was written from exoplayer. Is this something we still want to preserve in the experiment? If we do, I can rewrite this block of code to have two blocks, with the first block being the original implementation and the second block being exoplayer's implementation.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -910,36 +920,56 @@ private void maybeSetMaxVideoInputSize(MediaFormat format) {
...
+                || ("Amazon".equals(Build.MANUFACTURER)
+                    && ("KFSOWI".equals(Build.MODEL) // Kindle Soho
+                        || ("AFTS".equals(Build.MODEL) && crypto != null)))) { // Fire TV Gen 2
+              // Use the default value for cases where platform limitations may prevent buffers of the
+              // calculated maximum input size from being allocated.
+              return;
+            }
+        }
+        else {
+          if ("BRAVIA 4K 2015".equals(Build.MODEL)) {
```
johnxwork, line 941: The model targeting logic here seems very narrowly focused. Would other models of similar characteristic need similar treatment? If so, is there some spec that's more general than model?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -589,7 +590,7 @@ public static void createVideoMediaCodecBridge(
       }
     }
     if (!bridge.configureVideo(
-        mediaFormat, surface, crypto, 0, maxWidth, maxHeight, outCreateMediaCodecBridgeResult)) {
+        mediaFormat, surface, crypto, 0, maxWidth, maxHeight, experimentalVideoSizeCalculation, outCreateMediaCodecBridgeResult)) {
```
Awallky, line 593: I'd suggest either adding a builder pattern to reduce the number of parameters passed to the call here and/or improve readability by adding comments for some of the params, per Chromium's style guide - https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/java/java.md#parameter-comments.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -831,6 +832,7 @@ private boolean configureVideo(
       int flags,
       int maxSupportedWidth,
       int maxSupportedHeight,
+      boolean experimentalVideoSizeCalculation,
```
Awallky, line 835: Similar comment as the one above for `bridge.configureVideo`

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -910,36 +920,56 @@ private void maybeSetMaxVideoInputSize(MediaFormat format) {
...
-      case MimeTypes.VIDEO_AV1:
-        maxPixels = maxWidth * maxHeight;
+        pixelCount = maxWidth * maxHeight;
         minCompressionRatio = 4;
         break;
+      case MimeTypes.VIDEO_AV1:
+      // For the experiment, assume a min compression of 2 similar to the platform's C2SoftAomDec.cpp.
+      case MimeTypes.VIDEO_H265:
+      // For the experiment, assume a min compression of 2 similar to the platform's C2SoftHevcDec.cpp, but restrict
+      // the minimum size.
```
Awallky, line 963: These cases execute the same thing, which seems at odds with the implementation. Is this accurate? If so, is this better placed elsewhere?

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -285,6 +286,8 @@ std::unique_ptr<MediaCodecBridge> MediaCodecBridge::CreateVideoMediaCodecBridge(
       j_decoder_name, width_hint, height_hint, fps, max_width.value_or(-1),
       max_height.value_or(-1), j_surface_local, j_media_crypto_local,
       j_color_info, tunnel_mode_audio_session_id, max_video_input_size,
+      starboard::features::FeatureList::IsEnabled(
```
Awallky, line 289: Why can't we just get the feature flag value in the java layer? Is that not possible?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: It's ok to keep it as is.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: Let's keep it as is.

jasonzhangxx (approved):

starboard/android/shared/media_codec_bridge.cc
johnedocampo: This is currently not possible as of now.

johnedocampo: Update: b/447651764 has been filed which is handling this functionality. Once that functionality is added, we can go ahead and update this PR to grab the feature from the Java layer.

PR #6782, opened 2025-08-11
android: Disable flaky MediaCodecVideoDecoderTest

borongc:
MediaCodecVideoDecoderTest seems to be flaky. As Cobalt doesn't use it, disable the test.

Issue: 407549902

### DIFF
--- a/cobalt/testing/filters/android-arm/media_unittests_filter.json
+++ b/cobalt/testing/filters/android-arm/media_unittests_filter.json
@@ -12,6 +12,7 @@
     "ImageReaderGLOwnerSecureSurfaceControlTest.*",
     "KeySystemsTest.*",
     "MP4StreamParserTest.*",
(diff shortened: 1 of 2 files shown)


jellefoks (approved):

borongc:
> Can we instead just cut out the whole chrome-media android unittests? E.g. in //media/BUILD.gn [1], just add "&& ! is_android_tv" or some such.
> 
> [1] https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:media/BUILD.gn;l=225-235?q=android_video_decode_accelerator_unittests&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain

In this case, we need to check every BUILD.gn under `/android` and under `//media` for unit-tests.

PR #6793, opened 2025-08-11
Create Cobalt shell test targets

haozheng-cobalt:
Bug: 437186043


cobalt/testing/shell/BUILD.gn
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+static_library("cobalt_shell_test_lib") {
+  testonly = true
+  visibility = [ "//cobalt/testing/*" ]
+
+  sources = [
+    "utility/shell_content_utility_client.cc",
```
haozheng-cobalt, line 20: Closing off this attempt.

Even tho `utility/shell_content_utility_client.cc` is primarily a  test file.
I wanted to create a cobalt shell test wrapper that's one layer above cobalt shell. And have the cobalt shell test wrapper include the test files.

But other files in cobalt shell also depend on this test file, and I have to let cobalt shell depends on cobalt shell test target, which leads to circular dependecy.

PR #6798, opened 2025-08-12
This change can avoid crashes caused by exiting the process due to poor network and URL leaks

Jimmy-ge:
This change can avoid crashes caused by exiting the process due to poor network and URL leaks

Bug:409716671


johnxwork:
Hey, thanks for the PR.

But I think you are trying to contribute to the 25.lts.1+ branch? This PR is made for main branch which doesn't have a release timeline yet.

Jimmy-ge:
Ok, I'll create a PR for the 25.lts.1+ branch, thank you.

PR #6801, opened 2025-08-12
posix: update string function tests

madhurajayaraman:
Bug: 432487521

### DIFF
--- a/starboard/nplb/posix_compliance/posix_string_compare_no_case_n_test.cc
+++ b/starboard/nplb/posix_compliance/posix_string_compare_no_case_n_test.cc
@@ -14,32 +14,31 @@
 
 #include "testing/gtest/include/gtest/gtest.h"
 
-namespace starboard {
-namespace nplb {
+namespace starboard::nplb {
 namespace {
 
-TEST(PosixCompareNoCaseNTest, SunnyDaySelf) {
+TEST(PosixStringCompareNoCaseNTest, SunnyDaySelf) {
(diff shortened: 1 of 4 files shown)


starboard/nplb/posix_compliance/posix_string_compare_no_case_n_test.cc
```diff
@@ -53,6 +52,41 @@ TEST(PosixCompareNoCaseNTest, SunnyDayCase) {
   EXPECT_EQ(0, strncasecmp(kString4, kString3, strlen(kString4) / 2));
 }
 
+TEST(PosixStringCompareNoCaseNTest, SunnyDayOrdering) {
```
andrewsavage1, line 55: Consider splitting this into two test cases (also in the next file), one with all lowercase and one with mixed case

starboard/nplb/posix_compliance/posix_string_duplicate_test.cc
```diff
@@ -0,0 +1,78 @@
...
+  - ENOMEM: Insufficient memory is available to fulfill the request. Forcing
+    the system to run out of memory is not feasible in a unit test.
+*/
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace starboard::nplb {
+namespace {
+
+TEST(PosixStringDuplicateTest, SunnyDayBasic) {
```
andrewsavage1, line 28: Use a TEST_P for these, as the body of the test cases is exactly the same

PR #6806, opened 2025-08-12
posix: add more tests for open

madhurajayaraman:
Bug: 432487521


andrewsavage1:
Looks like tests are failing in real ways?

PR #6808, opened 2025-08-12
[media] Remove GetWriteHead NOTREACHED

osagie98:
This change removes the `NOTREACHED` in `ChunkDemuxer::GetWriteHead()` as the web app may call the function before `|id_to_mime_map_|` is populated. Now the Devel and Debug builds handle this case in the same way as QA and Gold by returning a default `TimeDelta`.

Bug: 431850940

### DIFF
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -1077,8 +1077,6 @@ base::TimeDelta ChunkDemuxer::GetWriteHead(const std::string& id) const {
   if (iter == id_to_streams_map_.end() || iter->second.empty()) {
     // This function may rarely execute before |id| has been inserted into
     // |id_to_streams_map_|. Return a default TimeDelta in this case.
-    LOG(INFO) << "Attempted to access the write head of SourceBuffer " << id
-                 << " before it's been fully initialized.";
     MEDIA_LOG(ERROR, media_log_)
         << "Attempted to access the write head of SourceBuffer " << id
         << " before it's been fully initialized.";


osagie98:
The linter added a lot of formatting changes, which I'll remove in a minute

media/filters/chunk_demuxer.cc
```diff
@@ -1075,8 +1075,10 @@ base::TimeDelta ChunkDemuxer::GetWriteHead(const std::string& id) const {
 
   auto iter = id_to_streams_map_.find(id);
   if (iter == id_to_streams_map_.end() || iter->second.empty()) {
-    // Handled just in case.
-    NOTREACHED();
+    // This function may rarely execute before |id| has been inserted into
+    // |id_to_streams_map_|. Return a default TimeDelta in this case.
+    LOG(WARNING) << "Attempted to access the write head of SourceBuffer "
+                  << id << " before it's been fully initialized.";
```
borongc: What happen if we return 0 as audio write ahead?

osagie98: This should happen only before any media has been buffered to the SourceBuffer. I don't believe anything should happen as a result

borongc: Will return a default value make more sense here (https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/base/media_switches.cc;l=495)?

kjyoun: not sure removing NOTREACHED() is the right approach.

Is NOTREACHED() condition that CreateDemuxerStream(id) is called after GetWriteHead(id) happening because of 1) the app, which we cannot control, or 2) is it the timing issue of the internal component?

If 1), we cannot avoid this and can happen at any moment. hence it's not a warning
If 2), we have to resolve this issue by fixing timing of internal components.

kjyoun: Looks like NOTREACHED() is met since `CreateDemuxerStream` is executed asynchronously

Then is it OK to return default value in such a case?
Any other good value to return?


https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/filters/chunk_demuxer.cc;l=872-877;drc=cb5543afd569a5d1550142fb651e27a2c745c42d
```
  std::unique_ptr<SourceBufferState> source_state =
      std::make_unique<SourceBufferState>(
          std::move(stream_parser), std::move(frame_processor),
          base::BindRepeating(&ChunkDemuxer::CreateDemuxerStream,
                              base::Unretained(this), id),
          media_log_);
```

osagie98: The default value here is 0, as there's no media buffered I don't think we can return any other value. The web app will call this function again multiple times, so this initial value should be ok.

osagie98: For a default value like `kAudioWriteDurationLocal` (1000000 us), I'm not sure if it's more useful to the web app than the current behavior, as the second call to `GetWriteHead` might unexpectedly be lower than the first call.

kjyoun: Can you update comment?

I think that it's not a WARNING. since it can happen per user behavior
Even though we log WARNIG here, it is not visible to the use, and nothing we can do.

How about this?

// App calls GetWriteHead(), before ChunkDemux creation is completed.
// We are returning 0

Then I am good with this change, but want to listen to the options from other reviewers
@borongc, @xiaomings WDYT?

osagie98: I think that's a fair change to make. My thought behind the WARNING level is that it signals that the call to `GetWriteHead` at that time is not useful and ideally should not happen, though it's mostly harmless. Then we'd see if this is happening more than we expect, and might want to see why we're wasting CPU time on it.

borongc: For WARNING message, I saw we do have `media_log_` here, so we can report it to WebApp.

Something like this:
```
MEDIA_LOG(ERROR, media_log_) << "GetWriteHead return 0";
```

It will be visible to us, and we can clean it up later once we verify this is harmless. @kjyoun  WDYT?

osagie98: Added a media log and downgraded the log to INFO

kjyoun: MEDIA_LOG(WARNING) ?

go/logging#severity

> WARING: It describes unexpected events which may indicate a problem.
> ERROR:  It describes unexpected problematic events that the program is able to recover from

Do we need comments and LOG(INFO)?

Do we really need to LOG this additionally? 
and those comment does not look necessary (since the information that it gives overlaps with log message)

kjyoun: In short, comment + LOG + MEDIA_LOG looks overkill.

MEDIA_LOG alone isn't enough?

borongc: +1 MEDIA_LOG is enough.

osagie98: Removed the first LOG

borongc:
nit: title can be `media: Remove GetWriteHead NOTREACHED`. See go/cobalt-style#commit-message.

PR #6811, opened 2025-08-12
posix: add tests for close

madhurajayaraman:
Bug: 432487521

### DIFF
--- a/starboard/nplb/posix_compliance/posix_file_close_test.cc
+++ b/starboard/nplb/posix_compliance/posix_file_close_test.cc
@@ -12,10 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// close is partially tested in posix_file_open_test.cc.
+/*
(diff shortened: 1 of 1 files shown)


starboard/nplb/posix_compliance/posix_file_close_test.cc
```diff
@@ -12,10 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// close is partially tested in posix_file_open_test.cc.
```
andrewsavage1, line 15: Is this no longer true? Or you just think the comment isn't worth much?

madhurajayaraman: It's still tested to an extent in file_open since it's used every time we open a file, but these tests cover enough of the scenarios that it doesn't really matter that close() is used elsewhere too -- our indicator for close() failing in any way should come from these tests, not from file_open.

PR #6814, opened 2025-08-12
Default run_browser_tests to devel/debug build

haozheng-cobalt:
Default run_browser_tests to devel/debug build so that browsertests will run on the CI. In the meanwhile, developers can always test browsertests whenever setting it to true in gn args.

Test: verified devel built locally, and tested the macro is propagated when setting the param in gn args.
Bug: 437981348


cobalt/build/configs/initialize_variables.gni
```diff
@@ -83,3 +79,9 @@ if (starboard_toolchain == "") {
     starboard_toolchain = cobalt_toolchain
   }
 }
+
+declare_args() {
+  # TODO(b/428999732, b/437999182): Cobalt shell was started as a minimal test shell.
+  # We use this param to separate all test logics from the Cobalt shell.
+  run_browser_tests = is_devel == true || is_debug == true
```
andrewsavage1: use run_browser_tests = !is_official_build

haozheng-cobalt: Got it.

andrewsavage1 (approved):

PR #6815, opened 2025-08-12
HeapObserverSet to use a copy for iteration

johnxwork:
ContextLifecycleNotifier breaks the assumption that HeapObserverSet can not be modified during iteration.
Before we identify the proper fix, let us try to iterate with a copy of the observers and relax the write restriction.
Note that observer deletion during iteration might still cause crash but from the top crash we observer it was AddObserver that's causing the trouble.

Bug: 427982512

### DIFF
--- a/third_party/blink/renderer/platform/heap_observer_set.h
+++ b/third_party/blink/renderer/platform/heap_observer_set.h
@@ -5,6 +5,10 @@
 #ifndef THIRD_PARTY_BLINK_RENDERER_PLATFORM_HEAP_OBSERVER_SET_H_
 #define THIRD_PARTY_BLINK_RENDERER_PLATFORM_HEAP_OBSERVER_SET_H_
 
+#include "build/build_config.h"
+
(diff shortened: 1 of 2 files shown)


andrewsavage1 (commented):
This looks fine as a way of testing, but can you make it really obvious by just having a huge `#if BUILDFLAG(IS_COBALT)` block for our temporary approach vs the upstream approach?

johnxwork:
> This looks fine as a way of testing, but can you make it really obvious by just having a huge `#if BUILDFLAG(IS_COBALT)` block for our temporary approach vs the upstream approach?

Done! PTAL

andrewsavage1 (approved):
Unless you think we're going to keep the edits, consider just putting them in a huge separate block, i.e. 
```
#ifndef THIRD_PARTY_...
#define ...

#if BUILDFLAG(IS_COBALT)
// all your code
#else
// all upstream code
#endif

#endif // THIRD_PARTY_...

andrewsavage1 (approved):

PR #6816, opened 2025-08-12
net/socket: Limit excessive error logging from recvmmsg

ckim1346:
On linux/evergreen builds, UDPSocketPosix has been flooding the logs with the error log at https://source.corp.google.com/h/github/youtube/cobalt/+/main:net/socket/udp_socket_posix.cc;drc=d1a03dcc46dafbf26f6a40db843f70eb935ef693;l=956. 

Added a 5-second rate limiter to limit the logs. I also made the errno logging more accurate; it was being incorrectly logged as 0 (no error) when it was really returning 90 (EMSGSIZE "Message too long") in this case. It'll require more investigation but the logs should be limited regardless of whether recvmmsg can be called successfully on linux/evergreen builds. 

Bug: 434195055

### DIFF
--- a/net/socket/udp_socket_posix.cc
+++ b/net/socket/udp_socket_posix.cc
@@ -34,6 +34,7 @@
 #include "base/rand_util.h"
 #include "base/task/current_thread.h"
 #include "base/task/thread_pool.h"
+#include "base/time/time.h"
 #include "build/chromeos_buildflags.h"
 #include "net/base/cronet_buildflags.h"
 #include "net/base/features.h"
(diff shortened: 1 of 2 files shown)


net/socket/udp_socket_posix.cc
```diff
@@ -930,6 +931,7 @@ int UDPSocketPosix::InternalReadMultiplePackets(
  
   int rv = recvmmsg(socket_, msgs, kNumPacketsPerReadMmsgCall,
                     MSG_DONTWAIT, nullptr);
+  const int recvmmsg_errno = errno;
```
hlwarriner, line 934: This is because the value was getting clobbered by some other system call / standard library function call?

ckim1346: Yeah I believe so possibly the `LOG(ERROR)` call?

andrewsavage1 (approved):

PR #6817, opened 2025-08-13
Embed icudata in libcobalt.so

niranjanyardi:
Issue: 437151583

Change the build configuration to ensure libcobalt.so embeds idu data. We are doing this to ensure parity with C25

### DIFF
--- a/cobalt/common/icu_init/init.cc
+++ b/cobalt/common/icu_init/init.cc
@@ -14,59 +14,22 @@
 
 #include "cobalt/common/icu_init/init.h"
 
-#include <fcntl.h>
-#include <pthread.h>
-#include <sys/mman.h>
-#include <sys/stat.h>
-#include <unicode/utypes.h>
-#include <unistd.h>
-
-#include <string>
-#include <vector>
-
-#include "base/i18n/icu_util.h"  // nogncheck
-#include "starboard/common/log.h"
-#include "starboard/configuration_constants.h"
-#include "starboard/file.h"
-#include "starboard/system.h"
-#include "unicode/putil.h"
-#include "unicode/udata.h"
-
 namespace cobalt {
 namespace common {
 namespace icu_init {
-
-namespace {
-
-pthread_once_t g_initialization_once = PTHREAD_ONCE_INIT;
-bool g_initialization_result = false;
-
-void InitializeIcuDatabase() {
-  g_initialization_result = base::i18n::InitializeICU();
-}
-
-bool IcuInit() {
-  pthread_once(&g_initialization_once, &InitializeIcuDatabase);
-  return g_initialization_result;
-}
-}  // namespace
 
 // Initialize ICU with a static initializer to ensure it is initialized
 // before program execution begins. While this is very early, it is not
 // guaranteed to be early enough for ICU to be used by other global
 // initializers.
-static bool g_icu_is_initialized = IcuInit();
+static bool g_icu_is_initialized = false;
 
 void EnsureInitialized() {
-  // Even though IcuInit() is called before main() is called, when the static
-  // initializers are called, the order of execution of static initializers is
-  // undefined. This function allows functions that may use ICU from static
-  // initializers themselves to guarantee that ICU is initialized first.
-  // Note: This is thread-safe because spurious calls to IcuInit() are
-  // thread-safe.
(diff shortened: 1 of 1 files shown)


cobalt/build/configs/linux_common.gn
```diff
@@ -111,3 +111,6 @@ enable_ppapi = false
 
 # Disable D-Bus
 use_dbus = false
+
+# Embed ICU data in the cobalt binary
+icu_use_data_file = false
```
niranjanyardi, line 116: should we do this for android as well ?

y4vor: I don't think it matters for ATV. We should do it for AOSP.

niranjanyardi: gotcha, then its in the right place

cobalt/build/linux/package.json
```diff
@@ -8,7 +8,6 @@
                 "gen/build_info.json",
                 "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt",
-                "icudtl.dat",
```
niranjanyardi, line 11: @kaidokert / @arjungm  im pretty sure this will fix the packaging errorhttps://b.corp.google.com/issues/437151583#comment3

This may potentially break forge tests. I guess i could use arjun's instructions to run forge tests on this local build.. does that sound good ?

hlwarriner:
I'm sorry I missed this. I'll review by Monday but no need to wait for my review before merging.

cobalt/common/icu_init/init.cc
```diff
@@ -37,97 +38,14 @@ namespace icu_init {
...
-  }
-
-  // Inform the OS that the mapped data is accessed randomly.
-  madvise(icu_data, length, MADV_RANDOM);
-
-  if (!SetIcuDataPointer(icu_data)) {
-    PrintIcuNotLoadedWarning();
-    return;
-  }
+  base::i18n::InitializeICU();
```
niranjanyardi: @jellefoks from our discussion yesterday, did you mention that i shoulkd avoid this intitilization as well ?

cobalt/build/linux/package.json
arjungm: This is changing how packaging works. As long as this file is not needed in future binaries of Chrobalt, this should be fine. However, yes please use the test method described here: go/cobalt-forge#local-dev-testing-with-forge

These tests are ideally expected to catch these breaking changes, so it will be a good litmus test that Forge is okay.

cobalt/common/icu_init/init.cc
niranjanyardi, line 1: @jellefoks please review this file

cobalt/build/linux/package.json
niranjanyardi: i dont expect the tests to break, ill keep a revert ready just in case

cobalt/build/modular_executable.gni
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//third_party/icu/config.gni")
```
jellefoks, line 15: This contains `icu_use_data_file = true`, inconsistent with what you added in `cobalt/build/configs/linux_common.gn`

cobalt/common/icu_init/init.cc
jellefoks: Let's leave the code here when `icu_use_data_file = true` and only remove it when it's false.

Niranjan > there's only 1 value - false globally(the overriden value) . For some reason i can't reply on this comment separately

cobalt/build/modular_executable.gni
niranjanyardi: the value is declared in config.gni and overriden in the linux_common.gn 

there's only 1 value at all times- the overriden value

cobalt/common/icu_init/init.cc
jellefoks: Now, if ever a build variant where `icu_use_data_file = false` is set is made and compiled with our time functions, things will break spectacularly in a difficult to diagnose way. This code should remain, or the `icu_use_data_file` flag should be removed.

niranjanyardi: talked offline , will get a PR up with changes we discussed

PR #6819, opened 2025-08-13
Speed up git checkout in CI

isarkis:
Checkouts are timing out on cobalt_sandbox.  

This change reduces fetch-depth to 1 and 2 where possible to speed up checkouts.

Issue: 438309414

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -31,6 +31,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          # GitHub's actions/checkout action, by default, checks out the merge commit when
+          # a pull request event triggers a workflow. This merge commit represents the state
+          # of the code as if the pull request had been merged into the base branch.
+          # Setting fetch-depth to 2 effectively gets a merge commit (i.e. HEAD) and the
+          # HEAD commit from the base branch (i.e. HEAD~1).
           fetch-depth: 2
       - name: Run pre-commit
         run: |
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -43,6 +43,11 @@ runs:
         run: |
           set -x
           cd src
+          # GitHub's actions/checkout action, by default, checks out the merge commit when
+          # a pull request event triggers a workflow. This merge commit represents the state
+          # of the code as if the pull request had been merged into the base branch.
+          # Setting fetch-depth to 2 effectively gets a merge commit (i.e. HEAD) and the
+          # HEAD commit from the base branch (i.e. HEAD~1).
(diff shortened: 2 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -31,11 +31,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
+          fetch-depth: 2
       - name: Run pre-commit
         run: |
           pre-commit install --hook-type pre-commit --hook-type pre-push
-          pre-commit run --show-diff-on-failure --color=always --from-ref ${{ github.event.pull_request.base.sha && github.event.pull_request.base.sha || github.event.before }} --to-ref HEAD
+          pre-commit run --show-diff-on-failure --color=always --from-ref HEAD~1 --to-ref HEAD
```
andrewsavage1, line 43: I don't this is correct if a PR has multiple commits in it

isarkis: local workspace has exactly 2 commits: a HEAD commit from the target branch, and a merge commit (HEAD from target branch + PR changes)

isarkis: Before making this PR I tested this in a sandbox repo.  Looks like actions/checkout merges PR into HEAD (target branch) as part of a checkout.  Here I added `git log` in precommit step, you can see a merge commit: https://github.com/youtube/cobalt_sandbox/actions/runs/16946414577/job/48028407363?pr=793

Effectively using `fetch-depth:2` guarantees to have HEAD being merge commit, and HEAD~1 being current HEAD from the target branch.

andrewsavage1: Could you do a quick test of pushing two commits, one adding trailing whitespace to a file and the other doing nothing (i.e. adding a comment somewhere)? I want to see the pre-commit check fail when HEAD is not the failing commit

andrewsavage1: Just refreshed and saw your most recent comment. Sg in that case, maybe add a comment that this is dependent on the implementation of actions/checkout

isarkis: > Could you do a quick test of pushing two commits, one adding trailing whitespace to a file and the other doing  nothing (i.e. adding a comment somewhere)? I want to see the pre-commit check fail when HEAD is not the failing commit

I actually ran that test in this PR (unintentionally :), see commit history on this PR.  My first commit had a trailing whitespace that pre-commit flagged.

isarkis: >Just refreshed and saw your most recent comment. Sg in that case, maybe add a comment that this is dependent on the implementation of actions/checkout

Added.

andrewsavage1: thanks!

andrewsavage1 (approved):

PR #6830, opened 2025-08-13
Fix typo and repeated field in switches.h/cc

fayev-g:
Fix typo and repeated field in switches.h/cc

Bug: b/438520038


yell0wd0g (approved):
LGTM!

cobalt/browser/switches.cc
```diff
@@ -197,7 +197,7 @@ const char kUserAgentClientHintsHelp[] =
     "--user_agent_client_hints=\"device_type=GAME;"
     "os_name_and_version=Linux armeabi-v7a\\; Android 7.1.2;evergreen_type=\" "
     "The 18 supported UA fields for overriding are: aux_field, brand, "
-    "build_configuration, chipset_model_number, cobalt_build_version_number, "
```
y4vor, line 200: what is the issue with the cobalt_build_version_number?

fayev-g: It is already in the line below it, and so it's repeated. I'm removing one of them here.

PR #6837, opened 2025-08-14
cobalt/test: Fix Navigation Browser Tests for Android

Awallky:
This change corrects missing data files and serving issues for nav browser tests.
This keeps these tests disabled for non-Android builds. Since these tests take a significant amount of time to run, follow-up PRs will enable these for other platforms.

Test: out/android-arm_devel/bin/run_cobalt_browsertests -v
Bug: 432503432


cobalt/testing/browser_tests/navigation_browsertest.cc
```diff
@@ -401,8 +396,13 @@ INSTANTIATE_TEST_SUITE_P(
 
 // Ensure that browser initiated basic navigations work.
 // TODO: b/432503432 - Investigate test failure
+#if !BUILDFLAG(IS_ANDROID)
```
yell0wd0g: Here and elsewhere: Note that in the immediate future
there'll more platforms than just is_android/not (e.g.
tvOS). So it might be better to change this to:
````
#if BUILDFLAG(IS_ANDROID)
#define MAYBE_bla bla
#else 
#define MAYBE_bla DISABLED_bla
#endif 
```

WDYT?
Also should we use `IS_ANDROID_TV`?

cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -346,7 +351,7 @@ test("cobalt_browsertests") {
   ]
 
   data = [
-    "data/",
+    "//cobalt/testing/browser_tests/data/",
```
yell0wd0g: nit: Could we depend on //content/shell//browser_tests/data ?
That way we wouldn't need to copy those files, and ultimately
the dependency would be for testing only.

yell0wd0g (commented):
Looking good, have some questions and/or nit.

cobalt/testing/browser_tests/navigation_browsertest.cc
Awallky: Good idea. Done.

cobalt/testing/browser_tests/BUILD.gn
Awallky: I was debating either what I originally uploaded or what you suggested. The delta is much much smaller this way though. We will fix things if they break by sharing test data with content.
Done.

PR #6845, opened 2025-08-15
Fix: Set initial focus on DidStopLoading

sherryzy:
Issue:
On startup, Cobalt would not respond to keyboard input until the user manually pressed the Tab key or clicked within the window. Although the main window was active at the OS level, no element within the web content had focus.

The root cause was the failure to set the initial focus on the web page after it finished loading. The browser's FocusController was never triggered to select the first focusable element. Content_shell has this same behavior.

This change resolves the issue by implementing the DidStopLoading() method in CobaltWebContentsObserver. When this method is called, it now explicitly calls Focus() on the renderWidgetHostView. This acts as the initial trigger for Blink's focus management system, ensuring the first focusable element is selected automatically.

Note: The unhandled kSbEventTypeFocus event and the application lifecycle are unrelated to the initial loss of focus upon startup, but may still require implementation.

Bug: 433965181

### DIFF
--- a/cobalt/browser/cobalt_web_contents_observer.cc
+++ b/cobalt/browser/cobalt_web_contents_observer.cc
@@ -18,6 +18,7 @@
 #include "cobalt/browser/embedded_resources/embedded_js.h"
 #include "cobalt/browser/migrate_storage_record/migration_manager.h"
 #include "content/public/browser/navigation_handle.h"
+#include "content/public/browser/render_widget_host_view.h"
 #include "content/public/browser/web_contents.h"
 
 #if BUILDFLAG(IS_ANDROIDTV)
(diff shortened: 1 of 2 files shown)


Awallky (commented):
Can we file a follow-up to verify this requirement in a test? We need a test that exercises our cobalt/browser target, which we do not have yet.

sherryzy:
@Awallky  There is a Kabuki forge test for the focusing event is failing, probably due to this bug. We temporarily disabled it, but we can re-enable it now to catch this issue.

Awallky:
I would prefer something to run in our presubmit CI.
That is, something blocks any changes that regress this behavior so we don't have to chase bugs later.

PR #6846, opened 2025-08-15
Enable basic crashpad functionality for evergreen

hlwarriner:
We're still considering some design decisions but for now this
implementation follows the general approach from c25: the handler
process is only launched after a crash and all interaction with the
crashpad client is done from below Starboard.
    
This is an initial step toward enabling crashpad: it may not yet be
portable and some major features, like runtime annotations and the use
of Cobalt's packaged CA root certificates, are still missing. And it's only
wired up for the elf_loader_sandbox (not the loader_app). But crashes
are captured and uploaded and minidumps can be symbolized for
crashes in the cobalt module.
    
Issue: 40651160


third_party/crashpad/crashpad/client/BUILD.gn
```diff
@@ -72,7 +72,8 @@ crashpad_static_library("client") {
     sources += [ "crashpad_client_fuchsia.cc" ]
   }
 
-  if (enable_cobalt_hermetic_hacks) {
+  if (defined(cobalt_toolchain) && current_toolchain == cobalt_toolchain &&
```
andrewsavage1: Can we do `current_toolchain == default_toolchain && enable_cobalt_hermetic_hacks` instead?

starboard/linux/x64x11/run_starboard_main.cc
```diff
@@ -53,9 +53,11 @@ int SbRunStarboardMain(int argc, char** argv, SbEventHandleCallback callback) {
   }
 
 #if !SB_IS(MODULAR)
-  third_party::crashpad::wrapper::InstallCrashpadHandler(ca_certificates_path);
+#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
```
andrewsavage1: Why is this necessary? I would've thought that SB_IS(MODULAR) would be equivalent

starboard/crashpad_wrapper/wrapper.cc
andrewsavage1, line 1: What's the reason for all the BUILDFLAG additions to crashpad_wrapper? If it's just do we don't call our added functions, i.e. InsertAnnotationForHandler, could we just wrap those instead?

third_party/crashpad/crashpad/handler/BUILD.gn
```diff
@@ -163,6 +165,9 @@ source_set("handler_test") {
 
 if (!crashpad_is_ios) {
   crashpad_executable("crashpad_handler") {
+    if (defined(is_native_target_build) && is_native_target_build) {
+      check_includes = false
```
andrewsavage1, line 169: Should we add a todo for this?

third_party/crashpad/crashpad/snapshot/BUILD.gn
```diff
@@ -20,6 +20,14 @@ if (crashpad_is_in_chromium) {
 }
 
 crashpad_static_library("snapshot") {
+  # TODO: b/406511608 - figure out how to remove the is_starboard condition
+  # here; we should only need to depend on this target from the native_target
+  # toolchain.
+  if ((defined(is_native_target_build) && is_native_target_build) ||
+        is_starboard) {
+    check_includes = false
```
andrewsavage1, line 28: Same here

third_party/crashpad/crashpad/handler/BUILD.gn
hlwarriner: I think it's ok. `check_includes` already indicates that there's a problem that should be addressed, and this is also copied over from c25 (so not a regression).

third_party/crashpad/crashpad/snapshot/BUILD.gn
hlwarriner: I think it's ok. `check_includes` already indicates that there's a problem that should be addressed, and this is also copied over from c25 (so not a regression).

starboard/linux/x64x11/run_starboard_main.cc
hlwarriner: Well I don't want this line to be compiled at all right now, even for evergreen, because this PR is only enabling crashpad for evergreen when run using elf_loader_sandbox (as opposed to the loader_app). I'll make that clear in the PR description. Also, this signature isn't supported yet because the certificates feature isn't supported yet.

third_party/crashpad/crashpad/client/BUILD.gn
hlwarriner: That's cleaner. Thanks!

starboard/crashpad_wrapper/wrapper.cc
hlwarriner: Sorry I'm not quite following. I'm using the buildflag there to temporarily avoid building code for features that we don't yet support.

For example,  we haven't yet added the `InsertAnnotationForHandler` declaration and implementation to the crashpad client in the main branch. So I'd like to just avoid building `third_party::crashpad::wrapper::InsertCrashpadAnnotation` and its call sites. We could do something else, like add a stub implementation in the crashpad client, but I don't feel strongly between the two interim approaches.

starboard/linux/x64x11/run_starboard_main.cc
andrewsavage1: SB_IS(MODULAR) is true for evergreen too, so I would think the hack BUILDFLAG would be redundant because it would be the same condition

starboard/crashpad_wrapper/wrapper.cc
andrewsavage1: Yeah sorry for the unclear initial comment—I was looking at the many places we needed to add BUILDFLAGs with this approach and thought it would be better to try to minimize them by i.e. having RecordStatus be a callable stub and then only guarding against the unimplemented calls into crashpad. But yeah, it probably doesn't matter much which we do as it would be pretty easy to revert these changes in the future

hlwarriner:
The evergreen builds are failing because we're now compiling `base/process/internal_linux.h`, which uses `strcmp` without including `<string.h>`.

```
$ gn path out/evergreen-x64_qa "starboard/elf_loader:elf_loader_sandbox(//starboard/linux/x64x11/toolchain:starboard)" "base(//starboard/linux/x64x11/toolchain:starboard)"
...
//starboard/elf_loader:elf_loader_sandbox --[private]-->
//starboard/crashpad_wrapper:crashpad_wrapper --[private]-->
//third_party/crashpad/crashpad/client:client --[public]-->
//third_party/crashpad/crashpad/third_party/mini_chromium:base --[public]-->
//base:base

Showing one of 3 "interesting" non-data paths. 0 of them are public.
```

We should probably just add that `#include` there.

third_party/crashpad/crashpad/handler/BUILD.gn
andrewsavage1: Where is the issue when you remove this? I saw two issues with the snapshot target below that look like they could be fixed by either adding 

```
  if (is_starboard) {
    deps += [ "//starboard/elf_loader:evergreen_info" ]
  }
```
to the snapshot target or by adding `// nogncheck` to the `#include`s in `third_party/crashpad/crashpad/snapshot/linux/process_snapshot_linux.cc`/`.h`. Can we do that instead?

hlwarriner:
> The evergreen builds are failing because we're now compiling `base/process/internal_linux.h`, which uses `strcmp` without including `<string.h>`.
> 
> ```
> $ gn path out/evergreen-x64_qa "starboard/elf_loader:elf_loader_sandbox(//starboard/linux/x64x11/toolchain:starboard)" "base(//starboard/linux/x64x11/toolchain:starboard)"
> ...
> //starboard/elf_loader:elf_loader_sandbox --[private]-->
> //starboard/crashpad_wrapper:crashpad_wrapper --[private]-->
> //third_party/crashpad/crashpad/client:client --[public]-->
> //third_party/crashpad/crashpad/third_party/mini_chromium:base --[public]-->
> //base:base
> 
> Showing one of 3 "interesting" non-data paths. 0 of them are public.
> ```
> 
> We should probably just add that `#include` there.

Added in a new commit.

starboard/crashpad_wrapper/wrapper.cc
hlwarriner: Got it, thanks. SGTM - I'll keep this as is.

starboard/linux/x64x11/run_starboard_main.cc
hlwarriner: Oh...I probably knew that at some point and then forgot it. I think it's confusing because we often use "modular" to describe the particular modular builds that are non-evergreen. For example, this is what the `linux-modular` builds are. But yeah I think you're right that we can remove  the `ENABLE_COBALT_HERMETIC_HACKS` buildflag here - I'll try that.

hlwarriner:
@y4vor @andrewsavage1, I added another commit to enable crashpad for evergreen on raspi. As I mentioned offline, I'll keep this PR up as an e2e draft so people can patch it to run evergreen with crashpad. But it won't be merged and I expect various checks / other builds to fail. Don't worry about reviewing this and I wouldn't pay too much attention to the changes; some of the customizations to //base etc. will be made more narrowly/thoughtfully in the chain of mergeable PRs that I upload soon.

PR #6847, opened 2025-08-15
Add back service device test to browsertests sources

haozheng-cobalt:
This PR reverts the [removal](https://source.corp.google.com/h/github/youtube/cobalt/+/a6b12416b43087a3be2e003d3151802ce271a7e5) of service device test, and add it to the browsertests condition instead.

Notice that this [removal](https://source.corp.google.com/h/github/youtube/cobalt/+/a6b12416b43087a3be2e003d3151802ce271a7e5) PR is is merged in main branch but not C26, and the inconsistency is breaking another [PR](https://github.com/youtube/cobalt/pull/6820/files#diff-7cde5f2eca7e86b63ef76c3a053c9848c326f9e08706d366b970520e07124c8c). 

Test: built gold build and ran CUJ locally.
Bug: 436671511
437981348

### DIFF
--- a/cobalt/shell/BUILD.gn
+++ b/cobalt/shell/BUILD.gn
@@ -197,6 +197,8 @@ source_set("browsertests_sources") {
     "//content/public/utility",
     "//content/test:content_test_mojo_bindings",
     "//content/test:test_support",
+    "//services/test/echo:lib",
+    "//services/test/echo/public/mojom",
   ]
 }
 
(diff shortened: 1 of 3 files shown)


Awallky (commented):
Can we add the #if run browser test to guard against these in non-official builds?

haozheng-cobalt:
> Can we add the #if run browser test to guard against these in non-official builds?

The file `utility/shell_content_utility_client.h|cc` is only included in non-official builds, the logics are already in place. You could search for `run_browser_tests` in the BUILD.gn to see how it's included.

haozheng-cobalt:
Do NOT cherry pick this to C26! 

See PR description.

PR #6848, opened 2025-08-15
android: Move SbAudioSink Init to CobaltActivity

johnedocampo:
Currently, `audio_sink::SbAudioSinkImpl::Initialize()` is called from application_android.cc, which makes it intialize before the cobalt browser client is able to initialize the feature list and field trials parameters (this initialization happens in cobalt_main_delegate.cc). This order of initialization makes it impossible to set up any starboard experiments for AudioTrackAudioSink, as the app will crash due to the AudioTrackAudioSink requesting information from the starboard FeatureList before it is initialized.

This change moves the `audio_sink::SbAudioSinkImpl::Initialize()` function to `CobaltActivity.java` in the `finishInitialization` function. This change is performed by adding a JNI function `InitializePlatformAudioSink` to `StarboardBridge.java` which is called in CobaltActivity. This function is then linked to `starboard_bridge.cc` which calls the initialize method.

Bug: 438001753

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -183,6 +183,12 @@ public abstract class CobaltActivity extends Activity {
 
   // Initially copied from ContentShellActiviy.java
   private void finishInitialization(Bundle savedInstanceState) {
+
+    // Initialize the platform's AudioSinkImpl. This is called here as this must come after the
+    // browser client's feature list and field trials are initialized. The feature list and field
+    // trials are initialized in CobaltContentBrowserClient::CreateFeatureListAndFieldTrials().
+    getStarboardBridge().initializePlatformAudioSink();
+
     // Load an empty page to let shell create WebContents.
     mShellManager.launchShell("");
     // Inject JavaBridge objects to the WebContents.
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
@@ -171,6 +171,8 @@ public class StarboardBridge {
 
     // void closeNativeStarboard(long nativeApp);
 
+    void initializePlatformAudioSink();
+
     void handleDeepLink(String url, boolean applicationStarted);
 
     void setAndroidBuildFingerprint(String fingerprint);
@@ -325,6 +327,12 @@ public class StarboardBridge {
       throw new IllegalArgumentException("args cannot be null");
     }
     return args;
+  }
+
+  // Initialize the platform's AudioTrackAudioSink. This must be done after the browser client
+  // loads in the feature list and field trials.
+  public void initializePlatformAudioSink() {
+    StarboardBridgeJni.get().initializePlatformAudioSink();
   }
 
   /** Sends an event to the web app to navigate to the given URL */
--- a/starboard/android/shared/starboard_bridge.cc
+++ b/starboard/android/shared/starboard_bridge.cc
@@ -120,6 +120,11 @@ extern "C" SB_EXPORT_PLATFORM jlong JNI_StarboardBridge_StartNativeStarboard(
 #endif  // SB_IS(EVERGREEN_COMPATIBLE)
 }
 
+extern "C" SB_EXPORT_PLATFORM void
+JNI_StarboardBridge_InitializePlatformAudioSink(JNIEnv* env) {
+  ::starboard::shared::starboard::audio_sink::SbAudioSinkImpl::Initialize();
+}
+
 extern "C" SB_EXPORT_PLATFORM void JNI_StarboardBridge_HandleDeepLink(
     JNIEnv* env,
     const JavaParamRef<jstring>& jurl,
--- a/starboard/android/shared/application_android.cc
+++ b/starboard/android/shared/application_android.cc
@@ -72,8 +72,6 @@ ApplicationAndroid::ApplicationAndroid(
   // class.
   RuntimeResourceOverlay::GetInstance();
 
-  ::starboard::shared::starboard::audio_sink::SbAudioSinkImpl::Initialize();
-
   JNIEnv* jni_env = base::android::AttachCurrentThread();
   app_start_timestamp_ = starboard_bridge_->GetAppStartTimestamp(jni_env);
 


starboard/android/shared/starboard_bridge.cc
```diff
@@ -120,6 +120,11 @@ extern "C" SB_EXPORT_PLATFORM jlong JNI_StarboardBridge_StartNativeStarboard(
 #endif  // SB_IS(EVERGREEN_COMPATIBLE)
 }
 
+extern "C" SB_EXPORT_PLATFORM void
+JNI_StarboardBridge_InitializePlatformAudioSink(JNIEnv* env) {
```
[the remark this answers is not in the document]
johnedocampo: Similar functions in starboard_bridge.cc do not do this, so will leave as is to preserve the established pattern.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -183,6 +183,11 @@ public void onFailure() {
 
   // Initially copied from ContentShellActiviy.java
   private void finishInitialization(Bundle savedInstanceState) {
+
+    // Initialize the platform's AudioSinkImpl. This is called here as this must come after the
+    // browser client's feature list and field trials are initialized.
+    getStarboardBridge().InitializePlatformAudioSink();
```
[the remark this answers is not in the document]
johnedocampo: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -327,6 +329,13 @@
     return args;
   }
 
+  /** Initialize the platform's AudioTrackAudioSink. This must be done after the browser client
+   *  loads in the Feature List and Field Trials.
+  */
+  public void InitializePlatformAudioSink() {
+    StarboardBridgeJni.get().InitializePlatformAudioSink();
+  }
```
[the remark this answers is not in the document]
johnedocampo: Done.

Awallky (approved):
LGTM!

Two things that are probably out of scope of this change:

1. Testing - Could we file a bug to address how to test this sequencing requirement we have for initialization here? I think we need some end to end test that we just don't have the ability to implement at the moment. 
2. Style - I noticed the Java classes in this PR violate Chromium's Java style guide naming rules - [link](https://source.android.com/docs/setup/contribute/code-style#follow-field-naming-conventions). It states *Non-public, non-static field names start with m.*
The private member variables should be prepended with `m`. This is pre-existing to your change, but could still use fixing. Could you please file a bug to address this at a later point?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -327,6 +329,13 @@ protected String[] getArgs() {
     return args;
   }
 
+  /** Initialize the platform's AudioTrackAudioSink. This must be done after the browser client
```
borongc: nit: `//` for all 3 lines or `/*`

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -183,6 +183,11 @@ public void onFailure() {
 
   // Initially copied from ContentShellActiviy.java
   private void finishInitialization(Bundle savedInstanceState) {
+
+    // Initialize the platform's AudioSinkImpl. This is called here as this must come after the
+    // browser client's feature list and field trials are initialized.
```
borongc: we should specify which function we call the initialization, so ppl can have more info when reading this comment

borongc:
As our browser is initialized by `startBrowserProcessesAsync()`, it should be succeed after `createStarboardBridge()`, then it calls `finishInitialization()`.

I am thinking about if there is a corner case where `finishInitialization()` is called before `createStarboardBridge()` is finished. @jasonzhangxx we may check if this won't be the case.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
johnedocampo: To make sure I understand correctly, we should add that the feature list and field trials are initialized in `cobalt_main_delegate.cc` through the function `CreateFeatureListAndFieldTrials()`?

borongc: Just say that `the feature list is initialized in CobaltContentBrowserClient::CreateFeatureListAndFieldTrials()` is good enough. So the reader can know where you refer to.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
johnedocampo: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
johnedocampo: Done. Thank you for the clarification!

jasonzhangxx:
> As our browser is initialized by `startBrowserProcessesAsync()`, it should be succeed after `createStarboardBridge()`, then it calls `finishInitialization()`.
> 
> I am thinking about if there is a corner case where `finishInitialization()` is called before `createStarboardBridge()` is finished. @jasonzhangxx we may check if this won't be the case.

I think in `CobaltActivity.createContent()`, before calling `startBrowserProcessesAsync()`, there're code to guarantee that starboard bridge has been created.

johnedocampo:
> LGTM!
> 
> Two things that are probably out of scope of this change:
> 
> 1. Testing - Could we file a bug to address how to test this sequencing requirement we have for initialization here? I think we need some end to end test that we just don't have the ability to implement at the moment.
> 2. Style - I noticed the Java classes in this PR violate Chromium's Java style guide naming rules - [link](https://source.android.com/docs/setup/contribute/code-style#follow-field-naming-conventions). It states _Non-public, non-static field names start with m._
>    The private member variables should be prepended with `m`. This is pre-existing to your change, but could still use fixing. Could you please file a bug to address this at a later point?

@Awallky Thanks for the review and feedback! I've filed your bug requests in [b/439640766](https://b.corp.google.com/issues/439640766) and [b/439641816](https://b.corp.google.com/issues/439641816), respectively.

PR #6849, opened 2025-08-15
android: Use Starboard Features for the delay during recreating VideoDecoder

borongc:
Allow to enable configurating the delay between reset/flush VideoDeocder using Starboard features. Examples are:
- `'--enable-features=VideoDecoderDelayUsecOverride:ResetDelayUsec/10ms'`
- `'--enable-features=VideoDecoderDelayUsecOverride:FlushDelayUsec/10ms'`

Issue: 400511693

### DIFF
--- a/starboard/extension/feature_config.h
+++ b/starboard/extension/feature_config.h
@@ -110,6 +110,15 @@ STARBOARD_FEATURE(kForceFlushDecoderDuringReset,
 // Set the following variable to true to force it reset audio decoder
 // during Reset(). This should be enabled with kForceFlushDecoderDuringReset.
 STARBOARD_FEATURE(kForceResetAudioDecoder, "ForceResetAudioDecoder", false)
+
+// By default, Cobalt restarts MediaCodec after stops/flushes during
+// Reset()/Flush(). Set the following variable to true with parameters
+// kResetDelayUsec and kFlushDelayUsec to force it to wait during
+// Reset()/Flush(). For example, with args
+// '--enable-features=VideoDecoderDelayUsecOverride:ResetDelayUsec/10ms'.
+STARBOARD_FEATURE(kVideoDecoderDelayUsecOverride,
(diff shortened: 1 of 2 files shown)


starboard/extension/feature_config.h
```diff
@@ -110,6 +110,15 @@ STARBOARD_FEATURE(kForceFlushDecoderDuringReset,
 // Set the following variable to true to force it reset audio decoder
 // during Reset(). This should be enabled with kForceFlushDecoderDuringReset.
 STARBOARD_FEATURE(kForceResetAudioDecoder, "ForceResetAudioDecoder", false)
+
+// By default, Cobalt restarts MediaCodec after stops/flushes during
+// Reset()/Flush(). Set the following variable to true with parameters
+// kResetDelayUsec and kFlushDelayUsec to force it to wait during
+// Reset()/Flush(). For example, with args
+// '--enable-features=VideoDecoderDelayUsecOverride:ResetDelayUsec/10ms'.
```
johnedocampo, line 118: Do you by chance have a reference to how Chromium parses base::TimeDelta parameters supplied from the command line? I thought we just supplied a number based on microseconds, but it seems here we can specify the unit of time (I assume other forms of time work as well). Neat!

johnedocampo (approved):
LGTM! Just left one question in regards to how we parse base::TimeDelta() parameters in the command line.

starboard/extension/feature_config.h
borongc: It is converted to the unit of `us` by `base::TimeDelta` as `int64_t`. See https://source.chromium.org/chromium/chromium/src/+/main:base/time/time.h;l=357.

As a result, no extra work for this. We just need to be careful when using it in Starboard, it's in the unit of `us`.

johnedocampo: Ah I see, thanks for the insight!

PR #6850, opened 2025-08-15
XR: Fix disable-xr build

kaidokert:
Fixes GN code to exclude all required IDL and header modules, and updates WebGL rendering context to compile as well.


kaidokert:
This is required for builds in #6841 to pass. Adds extra fixes on top of #6633

PR #6851, opened 2025-08-15
Gate browser test logic in non-official builds

haozheng-cobalt:
This PR gates custom protocol register, background sync controller mock and reduce accept language controller mock to browsertests, swap in prod interfaces instead.

Test: this PR does not affect non-offical, ran CUJ for official builds.
Bug: 437981348


Awallky:
@haozheng-cobalt Do we want to cherry pick this?

PR #6853, opened 2025-08-15
Hook up focus events to PlatformWindowStarboard

sherryzy:
This change hooks up the Starboard focus and blur events to the PlatformWindowStarboard, allowing the window to correctly handle activation and deactivation.

Bug: 433965181

### DIFF
--- a/cobalt/app/cobalt.cc
+++ b/cobalt/app/cobalt.cc
@@ -147,24 +147,25 @@ void SbEventHandle(const SbEvent* event) {
     }
     case kSbEventTypeBlur:
     case kSbEventTypeFocus:
+      CHECK(g_platform_event_source);
+      g_platform_event_source->HandleFocusEvent(event);
+      break;
     case kSbEventTypeConceal:
     case kSbEventTypeReveal:
     case kSbEventTypeFreeze:
     case kSbEventTypeUnfreeze:
       break;
     case kSbEventTypeInput:
-      if (g_platform_event_source) {
-        g_platform_event_source->HandleEvent(event);
-      }
+      CHECK(g_platform_event_source);
+      g_platform_event_source->HandleEvent(event);
       break;
     case kSbEventTypeLink:
     case kSbEventTypeVerticalSync:
     case kSbEventTypeScheduled:
     case kSbEventTypeLowMemory:
     case kSbEventTypeWindowSizeChanged:
-      if (g_platform_event_source) {
-        g_platform_event_source->HandleWindowSizeChangedEvent(event);
-      }
+      CHECK(g_platform_event_source);
+      g_platform_event_source->HandleWindowSizeChangedEvent(event);
       break;
     case kSbEventTypeOsNetworkDisconnected:
     case kSbEventTypeOsNetworkConnected:
--- a/ui/ozone/platform/starboard/platform_event_source_starboard.cc
+++ b/ui/ozone/platform/starboard/platform_event_source_starboard.cc
@@ -207,6 +207,7 @@ void DeliverEventHandler(std::unique_ptr<ui::Event> ui_event) {
 }
 
 void PlatformEventSourceStarboard::HandleEvent(const SbEvent* event) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
   if (event->type != kSbEventTypeInput) {
     return;
   }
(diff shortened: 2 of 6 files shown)


ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -308,6 +308,29 @@ void PlatformEventSourceStarboard::HandleEvent(const SbEvent* event) {
...
+}
+
+void PlatformEventSourceStarboard::HandleFocusEvent(const SbEvent* event) {
+  if (event->type == kSbEventTypeFocus) {
+    base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE, base::BindOnce(&DispatchFocusEventHandler, true));
+  } else if (event->type == kSbEventTypeBlur) {
+    base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE, base::BindOnce(&DispatchFocusEventHandler, false));
+  }
```
yell0wd0g: Style nit:

```
  if (event->type != kSbEventTypeFocus && event->type != kSbEventTypeBlur) {
    return;
  }
  const bool is_focused = event->type == kSbEventTypeFocus;
  base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
        FROM_HERE, base::BindOnce(&DispatchFocusEventHandler, is_focused));
```
WDYT?
Also would it look readable with DispatchFocusEventHandler() inlined?

cobalt/app/cobalt.cc
```diff
@@ -147,6 +147,10 @@ void SbEventHandle(const SbEvent* event) {
     }
     case kSbEventTypeBlur:
     case kSbEventTypeFocus:
+      if (g_platform_event_source) {
```
yell0wd0g: Question: Can this ever happen? Consider CHECK().

yell0wd0g (commented):
This LGTM. Apologies in advance for reviewing a draft CL
but I couldn't help myself 😅

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -352,7 +362,8 @@ void PlatformEventSourceStarboard::HandleWindowSizeChangedEvent(
 
   base::SingleThreadTaskRunner::GetCurrentDefault()->PostTask(
       FROM_HERE,
-      base::BindOnce(&DispatchWindowSizeChangedHandler, input_data->size.width,
+      base::BindOnce(&PlatformEventSourceStarboard::DispatchWindowSizeChanged,
+                     base::Unretained(this), input_data->size.width,
```
y4vor: why do w need to pass the `this` pointer? The `base::Uretained` could be dangerous. What guarantees that `PlatformEventSourceStarboard` would not be accessed after being destroyed?

sherryzy: I'm trying to address Miguel's comment to improve the readability by making the function inlined, so I deleted the helper function. In this case, I think PlatformEventSourceStarboard is a singleton that being created at  application start, and destroyed at application stop, so it should be safe? But I can revert this change back if it's safer to write in the original way.

yell0wd0g: If PlatformEventSourceStarboard lives on a single thread (which from
quick inspection seems to be the case albeit adding 
DCHECK_CALLED_ON_VALID_SEQUENCE() would be better), then adding
a WeakPtrFactory [1] and using here `GetWeakPtr()` would be
more semantically correct. FTR Chrome sees crashes during shutdown,
sometimes due to singletons' destruction order - they don't matter
much of course since things are going down anyway, but still.


[1] https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/memory/weak_ptr.h;l=19-25?q=weakptrfactory&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain

sherryzy: Done.

ui/ozone/platform/starboard/platform_event_source_starboard.h
```diff
@@ -45,6 +48,10 @@ class PlatformEventSourceStarboard : public PlatformEventSource {
 
  private:
   base::ObserverList<PlatformEventObserverStarboard>::Unchecked sb_observers_;
```
andrewsavage1: Should this be annotated with GUARDED_BY_CONTEXT?

ui/ozone/platform/starboard/platform_event_source_starboard.h
```diff
@@ -44,7 +47,12 @@ class PlatformEventSourceStarboard : public PlatformEventSource {
   void DispatchWindowSizeChanged(int width, int height);
 
  private:
-  base::ObserverList<PlatformEventObserverStarboard>::Unchecked sb_observers_;
+  base::ObserverList<PlatformEventObserverStarboard>::Unchecked sb_observers_
+      GUARDED_BY_CONTEXT(sequence_checker_);
```
yell0wd0g, line 51: \o/

yell0wd0g (approved):
LGTM! 

Many kudos on the use of WeakPtrFactory and DCHECKs !

PR #6860, opened 2025-08-16
Revert "[POSIX] Remove all syscalls"

andrewsavage1:
Causes ABI leaks for Evergreen.

Reverts youtube/cobalt#6625

Bug: 424002414


components/crash/core/app/BUILD.gn
```diff
@@ -53,7 +53,7 @@ static_library("app") {
     "crashpad.h",
   ]
 
-  if (is_linux && !is_cobalt_hermetic_build || is_chromeos) {
+  if (is_linux || is_chromeos) {
     sources += [ "crashpad_linux.cc" ]
```
andrewsavage1, line 57: This is likely the only thing that needs to be reverted

yuying-y (approved):

PR #6873, opened 2025-08-18
Reland "[POSIX] Remove all syscalls"

andrewsavage1:
Reverts youtube/cobalt#6860

Bug: 424002414


andrewsavage1:
I have a separate PR to improve the api_leak_detector to catch issues like this in the future

PR #6875, opened 2025-08-18
Search all symbols for API leaks

andrewsavage1:
Fixes a bug where symbols with whitespace in them would be filtered from the results of the api_leak_detector, in turn not treating them as leaks.

Bug: 437417239


y4vor (approved):
Maybe add different manifests per build configuration.

PR #6881, opened 2025-08-18
Separate switches for test and gate renderer test logics in non-official builds

haozheng-cobalt:
This pull request effectively gates renderer test logic for non-official builds by moving test-specific switches and helper classes into separate files and guarding their usage with the RUN_BROWSER_TESTS preprocessor macro. 

Test: ran CUJ on gold build locally, ran browsertests locally.
Bug: 437981348

### DIFF
--- a/cobalt/shell/browser/shell_content_browser_client.cc
+++ b/cobalt/shell/browser/shell_content_browser_client.cc
@@ -126,8 +126,8 @@
 #endif
 
 #if defined(RUN_BROWSER_TESTS)
-#include "cobalt/shell/common/shell_test_switches.h"
-#include "components/metrics/test/test_enabled_state_provider.h"
+#include "cobalt/shell/common/shell_test_switches.h"              // nogncheck
+#include "components/metrics/test/test_enabled_state_provider.h"  // nogncheck
 #endif  // defined(RUN_BROWSER_TESTS)
 
 namespace content {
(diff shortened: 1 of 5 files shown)


cobalt/shell/BUILD.gn
```diff
@@ -186,13 +186,23 @@ source_set("browsertests_sources") {
   sources = [
     "common/power_monitor_test_impl.cc",
     "common/power_monitor_test_impl.h",
+
+    #TODO: split up shell_switch files to test and prod.
+    "common/shell_switches.cc",
+    "common/shell_switches.h",
```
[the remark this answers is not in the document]
haozheng-cobalt: addressed in the second commit.

cobalt/shell/browser/shell.cc
```diff
@@ -88,10 +92,15 @@ Shell::Shell(std::unique_ptr<WebContents> web_contents,
     web_contents_->SetDelegate(this);
   }
 
+#if defined(RUN_BROWSER_TESTS)
```
Awallky, line 95: Why don't we just run IsRunWebTestsSwitchPresent and exclude the preprocessor directive here?
This switch and method should be defined in `cobalt/shell/common/shell_switches.h`.

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -142,9 +146,11 @@ void ShellBrowserMainParts::InitializeMessageLoopContext() {
 }
 
 void ShellBrowserMainParts::ToolkitInitialized() {
+#if defined(RUN_BROWSER_TESTS)
```
Awallky, line 149: See previous comment for the same issue.

cobalt/shell/browser/shell.cc
Awallky: Ah, we are removing them. Nvm

Awallky (approved):
LGTM
Thanks for consolidating these into test-only files and targets!

PR #4506, opened 2024-11-26
CoAT changes to support Kimono build

johnxwork:
1. Removed most java dependencies on content/shell/android/ resources;
2. Support dynamically adding "cobalt." java package prefix when invoking JNI;
3. Avoid using hash names in JNI templates.

go/kimono-chrobalt-build
b/372559266

PR #6886, opened 2025-08-19
unresolved base/android/jni_android.cc conflict

briantting:
rebase PR

### DIFF
--- a/.github/config/evergreen-arm-hardfp-raspi.json
+++ b/.github/config/evergreen-arm-hardfp-raspi.json
@@ -7,7 +7,9 @@
     "test_root_target": "//cobalt:gn_all",
     "num_gtest_shards": 1,
     "targets": [
-      "loader_app"
+      "cobalt:gn_all",
+      "loader_app",
+      "nplb"
     ],
     "includes": [
       {
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -33,6 +33,7 @@ group("gn_all") {
   }
   if (is_cobalt_hermetic_build) {
     deps += [
+      ":cobalt_loader",
       "//starboard/elf_loader:elf_loader_sandbox($starboard_toolchain)",
       "//starboard/elf_loader:elf_loader_sys_sandbox($starboard_toolchain)",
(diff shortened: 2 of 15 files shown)


niranjanyardi (commented):
lgtm, i just reviewed one file im a code owner of: cobalt/build/configs/android_common.gn

niranjanyardi (approved):
please get full approval from soneone on the android team

base/android/jni_android.cc
```diff
@@ -110,6 +189,164 @@ void InitVM(JavaVM* vm) {
   DCHECK(g_out_of_memory_error_class);
 }
 
+<<<<<<< HEAD
+=======
+bool IsVMInitialized() {
```
kaidokert: I think @johnxwork may have worked on this in #4506

This is one of the examples where we should have put clear #ifdef COBALT guards around it to make resolution easier. Looking at it, we still as of this commit do need the Cobalt specific hacks, so both blocks of code need to be integrated

base/android/jni_android.cc
```diff
@@ -110,6 +189,164 @@ void InitVM(JavaVM* vm) {
...
+template <MethodID::Type type>
+jmethodID MethodID::LazyGet(JNIEnv* env,
+                            jclass clazz,
+                            const char* method_name,
+                            const char* jni_signature,
+                            std::atomic<jmethodID>* atomic_method_id) {
+  const jmethodID value = atomic_method_id->load(std::memory_order_acquire);
+  if (value)
+    return value;
+#if BUILDFLAG(IS_COBALT)
```
kaidokert: This block has been completely removed in upstream with JNI Zero migration. I'm not sure if we need to retain it for ours. CC @haozheng-cobalt on this as well

base/android/jni_android.cc
johnxwork: These are Chromium code. I think they become merge conflicts due to removal upstream. The Cobalt additions are guarded: https://github.com/youtube/cobalt/pull/4506/files

base/android/jni_android.cc
johnxwork: These are the JNI hack to avoid symbol conflict with cronet, we'd need them before alternative becomes available.

PR #6890, opened 2025-08-19
android: Use Starboard features for use_stub_{audio/video}_decoder

borongc:
As `ApplicationAndroid` is created on the main thread, but `PlayerComponents::Factory::CreateComponents()` is called from the media thread, it is possible `shared::starboard::Application::Get()` is not yet created when the media thread calls the function, resulting in a native crash.

This PR moves `use_stub_{audio/video}_decoder` to use Starboard feature on android, which may fix the crash.

Issue: 415049224

### DIFF
--- a/starboard/extension/feature_config.h
+++ b/starboard/extension/feature_config.h
@@ -115,6 +115,11 @@ STARBOARD_FEATURE(kForcePlatformOpusDecoder, "ForcePlatformOpusDecoder", false)
 // during Reset(). This should be enabled with kForceFlushDecoderDuringReset.
 STARBOARD_FEATURE(kForceResetAudioDecoder, "ForceResetAudioDecoder", false)
 
+// By default, set the following to true to use stub decoder as audio/video
+// decoder.
+STARBOARD_FEATURE(kUseStubAudioDecoder, "UseStubAudioDecoder", false)
+STARBOARD_FEATURE(kUseStubVideoDecoder, "UseStubVideoDecoder", false)
+
 // By default, Cobalt restarts MediaCodec after stops/flushes during
 // Reset()/Flush(). Set the following variable to true with parameters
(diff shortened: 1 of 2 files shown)


kjyoun:
I'm not sure this is the right approach.

This fix is based on the assumption that shared::starboard::Application::Get()->GetCommandLine() returned nullptr for some reason, which should not happen.

I think we first need to confirm if this assumption is correct. If it is, we should fix GetCommandLine() to ensure it never returns nullptr, rather than patching the symptoms.

While this PR could resolve the crash in MediaDecoder, the underlying problem with `Application` instance would still exist and could cause other issues. When something we assume to be impossible actually occurs, it often points to a more significant problem under the hood.

borongc:
> I'm not sure this is the right approach.
> 
> This fix is based on the assumption that shared::starboard::Application::Get()->GetCommandLine() returned nullptr for some reason, which should not happen.
> 
> I think we first need to confirm if this assumption is correct. If it is, we should fix GetCommandLine() to ensure it never returns nullptr, rather than patching the symptoms.
> 
> While this PR could resolve the crash in MediaDecoder, the underlying problem with `Application` instance would still exist and could cause other issues. When something we assume to be impossible actually occurs, it often points to a more significant problem under the hood.

We currently start Browser thread with `startBrowserProcessesAsync` in [CobaltActivity.java](https://github.com/youtube/cobalt/blob/7a1c4256f8f108257d736ea2021a3641c1241e4b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java#L163), while initialing StarboardBridge and ApplicationAndroid on the main thread. In theory, they should be initialed before browser process finishes initiation.

If the crash is caused by this, we should use `startBrowserProcessesSync` to initialize the browser process to ensure it. As this will be a risky change, I prefer to make this change to check if the assumption is correct.

johnxwork (approved):

PR #6896, opened 2025-08-20
Include no-Bluetooth in lightweight build profile

kaidokert:
Add disable_bluetooth in lightweight profile.


kaidokert:
Testing #6756 in lightweight build mode.

PR #6906, opened 2025-08-20
Fix: Align file permissions to Chromium style

SortaCookie:
This is a fix for chromium pre-commit checks to match the same pre-commit checks Chromium would have run in an effort to better align our code to Chromium's. You are being asked to review because you were the last person to touch this file(s). If you think there's someone better to review please add them. Please the review the changes and if they look good please approve the PR.

Precommit error message:

checkperms.py failed:
... Has shebang but not executable bit: git add --chmod=+x ... ... Has executable bit but not shebang or ELF or Mach-O header: git add --chmod=-x ...

Bug: 435503470


SortaCookie:
Let me know if I need to split out all these bit changes to separate PRs.

kaidokert:
> Let me know if I need to split out all these bit changes to separate PRs.

This looks fine as is as it's just a permissions change

PR #6928, opened 2025-08-21
Add raspi-2-modular build to CI

andrewsavage1:
Bug: 441301254

### DIFF
--- a/cobalt/testing/filters/raspi-2-modular/nplb_loader_filter.json
+++ b/cobalt/testing/filters/raspi-2-modular/nplb_loader_filter.json
@@ -3,6 +3,8 @@
     "MaximumPlayerConfigurationExplorerTests*",
     "MultiplePlayerTest.*",
     "PosixAuxvGetauxvalOptionalTests/PosixAuxvGetauxvalOptionalTest.OptionalValuesAreValid/16",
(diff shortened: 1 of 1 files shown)


cobalt/build/gn.py
```diff
@@ -53,12 +53,13 @@
...
     'chromium_android-x86',
 ]
-_COBALT_LINUX_PLATFORMS = [
+_COBALT_STARBOARD_PLATFORMS = [
     'evergreen-arm-hardfp-raspi',
     'evergreen-x64',
     'linux-x64x11',
     'linux-x64x11-modular',
     'linux-x64x11-no-starboard',
+    'raspi-2-modular',
```
niranjanyardi, line 62: Just double checking my understanding : there is zero expectation from partners to build  platform modularly...correct ? 
for ex: we are not telling collabora to build rdk modularly.

andrewsavage1: That's correct, this is just for internal debugging

isarkis:
> lgtm
> 
> will be good to give @isarkis a heads up that a new builder is running on-device tests for raspi. IDK if they need to add more devices in the lab

We should have enough Raspi devices in the lab (70+), so we are good to go

PR #6931, opened 2025-08-21
Revert "Embed icudata in libcobalt.so"

niranjanyardi:
Issue: 437151583

Reverts youtube/cobalt#6817


niranjanyardi:
just keeping this ready if it breaks forge tests for some reason

PR #6941, opened 2025-08-21
Gate test logic in cobalt to official build

haozheng-cobalt:
This PR gates all the rest test logic to non-official build except ui/aura.
Test: ran CUJ on gold build locally, ran browsertests locally.
Bug: 437981348

### DIFF
--- a/cobalt/testing/browser_tests/BUILD.gn
+++ b/cobalt/testing/browser_tests/BUILD.gn
@@ -45,12 +45,6 @@ if (is_ios) {
   import("//build/config/ios/bundle_data_from_filelist.gni")
 }
 
-group("browser_tests") {
-  testonly = true
-
-  deps = [ ":cobalt_browsertests" ]
-}
-
 # browsertest_support can be used by targets that run cobalt_shell based
 # browser tests.
 static_library("browsertest_support") {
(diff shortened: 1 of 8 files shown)


cobalt/testing/browser_tests/media_browsertest.cc
```diff
@@ -12,6 +12,7 @@
 #include "build/build_config.h"
 #include "cobalt/shell/browser/shell.h"
 #include "cobalt/shell/common/shell_switches.h"
+#include "cobalt/shell/common/shell_test_switches.h"  // nogncheck
```
haozheng-cobalt, line 15: link

cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -45,12 +45,6 @@ if (is_ios) {
   import("//build/config/ios/bundle_data_from_filelist.gni")
 }
 
-group("browser_tests") {
```
Awallky, line 48: Why are we deleting the browser_tests group?

haozheng-cobalt: Why do we need this group? I see it simply references the cobalt_browsertests test target, and this group is not referenced anywhere in the code base.

Awallky: It was my (mis)understanding that the top-level target used this as an alias, but it's not really necessary. Builds still work fine without it.

Awallky (approved):

Awallky:
@haozheng-cobalt Do we want to cherry pick this?

haozheng-cobalt:
> @haozheng-cobalt Do we want to cherry pick this?

Thanks for the reminder, I will get it cherry picked.

PR #6949, opened 2025-08-22
evergreen: Apply the customizations to update_client

yuying-y:
These customizations support the essential operations of Cobalt Evergreen, including slot management, persisted data manipulation, etc. Follow-up PRs will utilize these customizations.

Issue: 431862767

### DIFF
--- a/components/update_client/cobalt_slot_management.h
+++ b/components/update_client/cobalt_slot_management.h
@@ -1,0 +1,98 @@
+// Copyright 2020 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef COMPONENTS_UPDATE_CLIENT_COBALT_SLOT_MANAGEMENT_H_
+#define COMPONENTS_UPDATE_CLIENT_COBALT_SLOT_MANAGEMENT_H_
+
+#include <string>
+
+#include "base/files/file_path.h"
+#include "base/version.h"
+#include "starboard/extension/installation_manager.h"
+
+namespace update_client {
+
+// Installation slot management logic for the Cobalt Updater.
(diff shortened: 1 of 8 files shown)


starboard/loader_app/BUILD.gn
```diff
@@ -310,7 +310,10 @@ static_library("slot_management") {
 if (sb_is_evergreen) {
   source_set("slot_management_test") {
     testonly = true
-    build_loader = false
+
+    # Commented out due to GN error: You set the variable "build_loader"
```
andrewsavage1: This is a remnant of c25 GN config—you can remove this whole comment/build_loader

starboard/loader_app/BUILD.gn
```diff
@@ -319,7 +322,7 @@ if (sb_is_evergreen) {
       ":installation_manager",
       ":installation_store_proto",
       ":slot_management",
-      "//starboard:starboard_with_main",
+      "//starboard:starboard_with_main($starboard_toolchain)",
```
andrewsavage1: Instead of doing this, add to the condition for building slot_management_test: `if (sb_is_evergreen && current_toolchain == starboard_toolchain) {`

andrewsavage1 (approved):

components/update_client/cobalt_slot_management.h
```diff
@@ -0,0 +1,98 @@
...
+
+  // Returns the selected index for the installation slot.
+  int GetInstallationIndex();
+
+  // Cleanup all drain files of the current app.
+  void CleanupAllDrainFiles();
+
+ private:
+  int installation_index_ = IM_EXT_INVALID_INDEX;
+  const CobaltExtensionInstallationManagerApi* installation_api_;
```
yell0wd0g, line 59: Drive-by: I took a look at the use of this, because const* as
argument is uncommon in Chromium, and is discouraged as member
variable (use RawPtr<> instead). Since this data type is a 
struct, could we just make a copy here? That would justify
including the header file here as opposed to forward declaring
it.

components/update_client/persisted_data.cc
```diff
@@ -90,6 +90,44 @@ std::string PersistedData::GetPingFreshness(const std::string& id) const {
...
+std::string PersistedData::GetUpdaterChannel(const std::string& id) const {
+  return GetString(id, "updaterchannel");
+}
+std::string PersistedData::GetLatestChannel() const {
+  const base::Value::Dict* dict =
+      &pref_service_->GetDict(kPersistedDataPreference);
+  if (!dict)
+    return std::string();
+  const std::string* result = dict->FindString("latestchannel");
+  return result != nullptr ? *result : std::string();
```
yell0wd0g, line 110: An std::string can be constructed out of a nullptr, so
this could just be
`return dict->FindString("latestchannel");`

PR #6950, opened 2025-08-22
Check that pthread_create() succeeded, not the thread ID, in common code

rakuco:
`pthread_t` is an opaque type. It is an `unsigned long` in glibc and musl when the latter is being used in C++ code, but it is an opaque pointer type in musl in C code as well as in Apple's implementation.

This means that the recent changes from #6869 introduced some failures in the (not-yet-upstreamed) tvOS build, as some SB_DCHECK_NE comparisons ended up comparing a `pthread_t` (a pointer) to 0 (an integer).

Instead of comparing whether `pthread_create()` has changed the thread ID pointer passed to it, it is more portable to just check whether the call to `pthread_create()` itself has succeeded.

This commit does this for cross-platform code; there are other occurrences in Android and Linux-specific code where this is less of a concern and which would require changing other assertions that also compare `pthread_t`'s and integers.

The `SB_DCHECK_*()` calls have also been changed to `SB_CHECK_*()` calls, both because otherwise `success` is not used in non-debug builds but also because these assertions look important enough to cause execution to stop if they fail.

Fixed: 440370504

### DIFF
--- a/starboard/android/shared/BUILD.gn
+++ b/starboard/android/shared/BUILD.gn
@@ -397,6 +397,7 @@ source_set("starboard_platform_tests") {
   sources = media_tests_sources + player_tests_sources + [
               "//starboard/shared/starboard/drm/drm_test_helpers.cc",
               "//starboard/shared/starboard/drm/drm_test_helpers.h",
+              "//starboard/shared/starboard/starboard_feature_test.cc",
               "jni_env_ext_test.cc",
               "model_year_test.cc",
               "player_get_preferred_output_mode_test.cc",
--- a/starboard/android/shared/android_main.cc
+++ b/starboard/android/shared/android_main.cc
@@ -12,212 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <dirent.h>
-#include <fcntl.h>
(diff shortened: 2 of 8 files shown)


rakuco:

rakuco:
As usual, the test failures look unrelated and come from missing shard data AFAICS.

kjyoun:
Can you address those 2 comments from Gemini?

They sound reasonable to me.

Otherwise, LGTM

rakuco:
> Can you address those 2 comments from Gemini?

Done!

kjyoun (approved):

rakuco:
Thanks; it looks like this also needs approval from someone in the cobalt-starboard-owners team. I guess @yell0wd0g would count? :-)

kjyoun:
found that this pattern is used in other places, as well.

I created a new PR to address them, as well
https://github.com/youtube/cobalt/pull/6952

kjyoun:
https://github.com/youtube/cobalt/pull/6952 is approved. 
I will merge it shortly, It covers this PR, so that this PR would not be necessary

kjyoun:
#6952 is now merged

rakuco:
Thank you!

PR #5227, opened 2025-03-27
Add volume button and mute button key events

TyHolc:
Detect volume and mute button presses from the OS and dispatch them for the webapp to handle.

b/378384110

PR #6954, opened 2025-08-22
cleanup: Remove unused JNI volume control methods

kjyoun:
The JNI functions `nativeVolumeChanged` and `nativeMuteChanged` are not used and their C++ implementations were empty. This commit removes the dead code from both the Java and C++ sides.

Volume controls work without these JNI methods by #5227 

Bug: 413418478


zhongqiliang:
Are the volume key events still sent to Kabuki? There are YT revenue related to this.

johnxwork (approved):

kjyoun:
> Are the volume key events still sent to Kabuki? There are YT revenue related to this.

I don't know, but I can say this is NO-OP change. 
https://github.com/youtube/cobalt/pull/5227 makes volume key handled w/o those JNI methods

PR #6956, opened 2025-08-22
Fix: Player bounds not updating subsequent calls

sherryzy:
The previous implementation of PlayerSetBounds in application_x11 only updated the current video bounds the first time it was called for a given player. Subsequent calls would update the bound for the next frame but would not be reflected in the current frame, causing the player's position and size to appear frozen.

This change removes the conditional check that prevented updates to current_video_bounds_ for existing players. Now, on every call to PlayerSetBounds, the player's existing bounds are removed from current_video_bounds_ and the new bounds are inserted in the correct z-order. This ensures that chang to player bounds are reflected immediately.

Bug: 401006482

### DIFF
--- a/starboard/shared/x11/application_x11.cc
+++ b/starboard/shared/x11/application_x11.cc
@@ -850,9 +850,6 @@ void ApplicationX11::PlayerSetBounds(SbPlayer player,
                                      int height) {
   std::lock_guard lock(frame_mutex_);
 
-  bool player_exists =
-      next_video_bounds_.find(player) != next_video_bounds_.end();
-
   FrameInfo& frame_info = next_video_bounds_[player];
   frame_info.player = player;
   frame_info.z_index = z_index;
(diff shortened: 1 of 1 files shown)


starboard/shared/x11/application_x11.cc
```diff
@@ -861,13 +858,16 @@ void ApplicationX11::PlayerSetBounds(SbPlayer player,
...
-    return;
+  // The bounds should only take effect once the UI frame is submitted. But we
+  // also apply the bounds immediately so that there is no flicker.
+  for (auto it = current_video_bounds_.begin();
+       it != current_video_bounds_.end(); ++it) {
+    if (it->player == player) {
+      current_video_bounds_.erase(it);
+      break;
+    }
   }
```
yell0wd0g, line 869: Chromium supports most of C++20 and in particular 
std::vector::erase_if() ([here](https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++-features.md#std_erase_if_for-containers-allowed)) so we could just do here
```
std::erase_if(current_video_bounds_, [player](const auto& entry) {
    return entry.player == player;
});
```

Albeit I'm not sure what is supported by Starboard-Linux. Maybe the
same?

yell0wd0g (approved):
Code looks good, just a C++20 question. 

Leaving an RS LGTM just in case.

starboard/shared/x11/application_x11.cc
sherryzy: From a quick search, I think the default config is c++17 for starboard: https://github.com/youtube/cobalt/blob/513c742970293544e93ab80aeca8d5788c31cc8c/starboard/build/config/BUILD.gn#L238 so maybe safe to just keep it in the original way..

PR #6967, opened 2025-08-22
Add Android apk size check

isarkis:
Bug: 413391308

### DIFF
--- a/.github/workflows/diff_android_binary.yaml
+++ b/.github/workflows/diff_android_binary.yaml
@@ -1,0 +1,141 @@
+name: Android Binary Size Diff
+
+on:
+  workflow_run:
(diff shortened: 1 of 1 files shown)


yell0wd0g:
This is nice! 

The code as-is doesn't prevent landing a CL/PR based on size delta, right?
It only provides a comment in the CL/PR with said delta, IIUC.

isarkis:
> This is nice!
> 
> The code as-is doesn't prevent landing a CL/PR based on size delta, right? It only provides a comment in the CL/PR with said delta, IIUC.

That's is correct, but this PR is not ready and it's not possible to test it without landing it.  I am going to test it on `cobalt_sandbox`

PR #6969, opened 2025-08-23
Enable MemoryPressureMonitor/Listener on CoAT

johnxwork:
This class was not hooked up before.
With some logging, I can confirm that this PR can trigger existing MemoryPressureListener(about 6-8 in total) to try reducing memory at critical memory pressure.

Bug: 440200929


zhongqiliang:
This is good, we could try it out and see if it can reduce the crash rate. 

Meanwhile, Gemini said this:
User-Visible Effect: The page might become unresponsive or "janky" for a moment during a major garbage collection. If you scroll back up a long page, you might see images flicker or reload as they are re-decoded because they were evicted from the cache.

I 'd say, let 's keep an eye on the user feedbacks.

zhongqiliang (approved):

yell0wd0g (approved):
Excellent catch! 

No wonder mem pressure wasn't caught.

PR #6972, opened 2025-08-24
Add temporary draft build instructions for 26.android branch

billy0059:
Adds initial draft version of building Cobalt for 26.android branch. We assume that initial developers are somewhat familiar with either Chromium or Cobalt build processes.

modified from: https://github.com/youtube/cobalt/commit/20af1a1


kaidokert:
Thanks for putting this up !

PR #6975, opened 2025-08-25
starboard: replace COBALT_BUILD_TYPE_GOLD with COBALT_IS_RELEASE_BUILD

kjyoun:
`COBALT_BUILD_TYPE_GOLD`  is not defined, since `//starboard/build/config:base` is never activated. Instead, we should use `COBALT_IS_RELEASE_BUILD`

This PR also removed no-more-used starboard/build/config/BUILDCONFIG.gn.

Bug: 440648273

### DIFF
--- a/base/message_loop/message_pump_ui_starboard.cc
+++ b/base/message_loop/message_pump_ui_starboard.cc
@@ -14,6 +14,7 @@
 
 #include "base/message_loop/message_pump_ui_starboard.h"
 
+#include "build/build_config.h"
 #include "base/logging.h"
 #include "base/notreached.h"
 #include "base/time/time.h"
--- a/cobalt/browser/user_agent/user_agent_platform_info.cc
+++ b/cobalt/browser/user_agent/user_agent_platform_info.cc
@@ -25,6 +25,7 @@
 #include "base/strings/stringprintf.h"
 #include "base/system/sys_info.h"
 #include "base/system/sys_info_starboard.h"
+#include "build/build_config.h"
 #include "starboard/extension/platform_info.h"
 
 #include "cobalt/cobalt_build_id.h"  // Generated
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -18,6 +18,7 @@
 
 #include "base/feature_list.h"
 #include "base/logging.h"
+#include "build/build_config.h"
 #include "media/base/media_switches.h"
 #include "media/base/video_codecs.h"
 #include "media/starboard/decoder_buffer_allocator_strategy.h"
--- a/media/starboard/decoder_buffer_allocator.h
+++ b/media/starboard/decoder_buffer_allocator.h
@@ -22,6 +22,7 @@
 #include "base/synchronization/lock.h"
 #include "base/thread_annotations.h"
 #include "base/time/time.h"
+#include "build/build_config.h"
 #include "media/base/decoder_buffer.h"
 #include "media/starboard/decoder_buffer_memory_info.h"
 #include "starboard/media.h"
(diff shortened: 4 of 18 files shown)


starboard/build/config/BUILD.gn
```diff
@@ -171,8 +125,54 @@ config("target") {
...
+    defines += [
+      "STARBOARD_BUILD_TYPE_GOLD",
+      "COBALT_BUILD_TYPE_GOLD",
+      "NDEBUG",
+      "OFFICIAL_BUILD",
+      "TRACING_DISABLED",
+      "COBALT_FORCE_CSP",
+      "COBALT_FORCE_CORS",
+    ]
+  }
```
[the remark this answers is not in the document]
kjyoun: good suggestion, though it's out to the scope of this PR and it does not build as it is.

starboard/build/config/BUILD.gn
```diff
@@ -68,52 +68,6 @@ if (cobalt_pending_clean_up) {
   }
 }
 
-config("base") {
```
andrewsavage1, line 71: Is this used anywhere right now in the build? If not, why not just delete it?

kjyoun: build is not used, but we need the definitions that this  build config defines.

e.g.) `STARBOARD_BUILD_TYPE_GOLD`


That's why I moved definitions done in base config to starboard config.

andrewsavage1: We don't define them in the build right now so I think we should switch to what we use elsewhere for config differences i.e. in https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/browser/user_agent/user_agent_platform_info.cc;l=281-289?q=COBALT_IS_RELEASE_BUILD%20user_agen&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

kjyoun: great! Thanks for the flag.

I replaced `STARBOARD_BUILD_TYPE_GOLD` with `COBALT_IS_RELEASE_BUILD`

Now the code makes more sense.

```
Java_dev_cobalt_coat_StarboardBridge_isReleaseBuild() {
#if defined(ENABLE_BUILDFLAG_COBALT_IS_RELEASE_BUILD)
  return true;
#else
  return false;
#endif
}
```

base/message_loop/message_pump_ui_starboard.cc
```diff
@@ -56,7 +56,7 @@ void MessagePumpUIStarboard::CancelImmediate() {
 
 void MessagePumpUIStarboard::RunUntilIdle() {
   DCHECK(delegate_);
-#if !defined(COBALT_BUILD_TYPE_GOLD)
+#if !defined(ENABLE_BUILDFLAG_COBALT_IS_RELEASE_BUILD)
```
andrewsavage1: Generally you'll want to use `BUILDFLAG(COBALT_IS_RELEASE_BUILD)` instead of this define directly

But the bigger point is this is a large functional change, as I don't believe COBALT_BUILD_TYPE_GOLD is defined right now, so we're enabling all of these code paths for gold build. That's probably the right thing to do, just want to call it out

kjyoun: I thought that I should not include "build/buildflag.h" from starboard.
but it turns out that #if BUILDFLAG() works w/o including it.

re: many file changes, 
With PR checked-in, those code paths will be enabled on main branch. 
Plan is to sit on it a few days. Then check-pick to 26.android, which is the branch that actually needs this.

I was surprised `COBALT_BUILD_TYPE_GOLD` used across so many files are actually NO-OP. I think that we need to activate it.

kjyoun: Done

andrewsavage1 (approved):
LGTM, but we should probably add the build/build_config.h include for the BUILDFLAG macros for files that don't already explicitly include it

kjyoun:
re: build/build_config.h

it is included transitively, probably through the one included in starboard/configuration,
though each file should include it directly.

Done

jasonzhangxx (approved):

PR #6977, opened 2025-08-25
[A11y] Check null before requesting focus on webContents

haozheng-cobalt:
This PR adds null checks before requesting focus  on webContents when A11y is enabled.

I cannot reproduce locally the scenario when any of the variables return null, and below are the potential scenarios summarized by Gemini CLI:

   1. `webContents` can be null:
       * This is the most likely object to be null in normal operation.
       * cobaltActivity.getActiveWebContents() could return null if the
         CobaltActivity is in a state where no WebContents is currently active. This
          can happen during activity startup or shutdown, if the WebContents is
         destroyed due to a crash or memory pressure, or if the activity is
         transitioning between different WebContents instances.

   2. `delegate` (`ViewAndroidDelegate`) can be null:
       * webContents.getViewAndroidDelegate() could theoretically return null.
       * This would be an unusual state. A WebContents that is part of the view
         hierarchy is expected to have a ViewAndroidDelegate. This could happen if
         the WebContents is not fully initialized or is in the process of being
         torn down.

   3. `webContentsView` can be null:
       * delegate.getContainerView() can return null.
       * This can happen if the view has not been created yet or has already been
         destroyed (e.g., it's not attached to a window).

#VibeCoding
Test: built app and ran CUJs when Talkback is on/off.
Bug: 440812420

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
@@ -1,223 +1,252 @@
 // Copyright 2025 The Cobalt Authors. All Rights Reserved.
- //
- // Licensed under the Apache License, Version 2.0 (the "License");
- // you may not use this file except in compliance with the License.
- // You may obtain a copy of the License at
- //
- //     http://www.apache.org/licenses/LICENSE-2.0
- //
- // Unless required by applicable law or agreed to in writing, software
- // distributed under the License is distributed on an "AS IS" BASIS,
- // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- // See the License for the specific language governing permissions and
- // limitations under the License.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package dev.cobalt.coat;
 
- package dev.cobalt.coat;
+import android.graphics.Rect;
+import android.os.Bundle;
+import android.os.Handler;
+import android.os.Looper;
+import android.util.Log;
+import android.view.KeyEvent;
+import android.view.View;
+import android.view.accessibility.AccessibilityEvent;
+import androidx.core.view.ViewCompat;
+import androidx.core.view.accessibility.AccessibilityNodeInfoCompat;
+import androidx.customview.widget.ExploreByTouchHelper;
+import java.lang.ref.WeakReference;
+import java.util.BitSet;
+import java.util.List;
+import org.chromium.content_public.browser.WebContents;
+import org.chromium.ui.base.ViewAndroidDelegate;
 
- import android.graphics.Rect;
- import android.os.Bundle;
- import android.os.Handler;
- import android.os.Looper;
- import android.view.KeyEvent;
- import android.view.View;
- import android.view.accessibility.AccessibilityEvent;
- import androidx.core.view.ViewCompat;
- import androidx.core.view.accessibility.AccessibilityNodeInfoCompat;
- import androidx.customview.widget.ExploreByTouchHelper;
- import java.lang.ref.WeakReference;
- import java.util.BitSet;
- import java.util.List;
- import org.chromium.content_public.browser.WebContents;
+/**
+ * An ExploreByTouchHelper that create a virtual d-pad grid, so that Cobalt
+ * remains functional when the TalkBack screen reader is enabled (which
+ * otherwise intercepts d-pad events for most applications).
+ */
+class CobaltA11yHelper extends ExploreByTouchHelper {
 
- /**
-  * An ExploreByTouchHelper that create a virtual d-pad grid, so that Cobalt remains functional when
-  * the TalkBack screen reader is enabled (which otherwise intercepts d-pad events for most
-  * applications).
-  */
- class CobaltA11yHelper extends ExploreByTouchHelper {
-   // The fake dimensions for the nine virtual views.
-   // These values are arbitrary as long as the views stay on the screen.
-   private static final int FAKE_VIEW_HEIGHT = 10;
-   private static final int FAKE_VIEW_WIDTH = 10;
+    private static final String TAG = "CobaltA11yHelper";
+    // The fake dimensions for the nine virtual views.
+    // These values are arbitrary as long as the views stay on the screen.
+    private static final int FAKE_VIEW_HEIGHT = 10;
+    private static final int FAKE_VIEW_WIDTH = 10;
 
-   private static final int CENTER_VIEW_ID = 5;
-   private static final int UP_VIEW_ID = 2;
-   private static final int DOWN_VIEW_ID = 8;
-   private static final int LEFT_VIEW_ID = 4;
-   private static final int RIGHT_VIEW_ID = 6;
+    private static final int CENTER_VIEW_ID = 5;
+    private static final int UP_VIEW_ID = 2;
+    private static final int DOWN_VIEW_ID = 8;
+    private static final int LEFT_VIEW_ID = 4;
+    private static final int RIGHT_VIEW_ID = 6;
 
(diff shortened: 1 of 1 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -166,12 +169,29 @@ private void maybeInjectEvent(int currentFocusedViewId) {
...
-     // Request focus on WebContents.
+    // When the accessibility service is active, it can prevent the application from
+    // gaining focus directly. This code is our best-effort attempt to return focus to
+    // the browser's web view.
      CobaltActivity cobaltActivity = mCobaltActivityRef.get();
-     WebContents webContents = cobaltActivity.getActiveWebContents();
-     View webContentsView = webContents.getViewAndroidDelegate().getContainerView();
-     if (webContentsView != null) {
-       webContentsView.requestFocus();
+     if (cobaltActivity != null) {
```
Awallky: Can we also add histograms for each of the potential null values to detect what is going on here? Once we have UMA running, this will be very insightful information and will help detect which of these issues is most prevalent.

See go/cobalt-telemetry#histograms for how to do this.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -166,12 +169,29 @@ private void maybeInjectEvent(int currentFocusedViewId) {
...
+       WebContents webContents = cobaltActivity.getActiveWebContents();
+       if (webContents != null) {
```
Awallky: Can we flatten this nesting structure to improve readability? We could just check for null-ness in each successive if block for the Object they depend on.

haozheng-cobalt: sorry I don't get how this can be further improved. I keep the nested if structure this way so we could print the log for each null scenario.

zhongqiliang: This code need to improve.
https://g3doc.corp.google.com/company/teams/nest-ninja-community/ninja-tips/episodes/6.md?cl=head
gemini should be able to help.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -166,12 +169,29 @@ private void maybeInjectEvent(int currentFocusedViewId) {
...
+           }
+         } else {
+           Log.i(TAG, "ViewAndroidDelegate is null");
+         }
+       } else {
+         Log.i(TAG, "WebContents is null");
+       }
+     } else {
+       Log.i(TAG, "CobaltActivity is null");
      }
```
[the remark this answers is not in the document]
Awallky: I don't agree with use of the `do/while` block, but that is generally the approach I was thinking of.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -166,12 +169,29 @@ private void maybeInjectEvent(int currentFocusedViewId) {
...
+           } else {
+             Log.i(TAG, "WebContentsView is null");
+           }
+         } else {
+           Log.i(TAG, "ViewAndroidDelegate is null");
+         }
+       } else {
+         Log.i(TAG, "WebContents is null");
+       }
+     } else {
```
johnxwork: Can you change the nested if conditions to:
if (condition_1 == null)
  return;
if (condition_2 == null)
  return;

haozheng-cobalt: I considered this, but think it's more safer to proceed than returning when these variables are null. 
because I'm not super sure we should skip the following a11y node builder when web contents is not active.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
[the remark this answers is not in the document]
zhongqiliang: No need to use do while block, refactor this section to a separate method will do the trick.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
Awallky: You could add a private method to move all of this logic and return a `View` object. The caller can just check if the return value is null. If it's null, skip using it. Otherwise, call the focus method. Then, regardless of the return value, continue to the rest of the function.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
haozheng-cobalt: This PR aims to address the urgent canary crash, we will consider to add histogram for this in the future.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
haozheng-cobalt: Thank both for advise, created a private `focusOnWebContentsView` function to include the flatten conditions.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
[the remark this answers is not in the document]
haozheng-cobalt: Thanks, same as above.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
haozheng-cobalt: Thanks @Awallky, this is helpful, I have created function `focusOnWebContentsView` similar to your idea.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -1,223 +1,252 @@
...
+    protected boolean onPerformActionForVirtualView(int virtualViewId, int action, Bundle arguments) {
+        return false;
+    }
+
+    /**
+     * When the accessibility service is active, it can prevent the application
+     * from gaining focus directly. This code is our best-effort attempt to
+     * return focus to the browser's web view.
+     */
+    private void focusOnWebContentsView() {
```
haozheng-cobalt, line 228: link

johnxwork (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -1,223 +1,252 @@
...
+        },
+                INPUT_FOCUS_CHANGE_DELAY);
+    }
+
+    private void maybeInjectEvent(int currentFocusedViewId) {
+        if (!unhandledInput) {
+            return;
+        }
+        CobaltActivity cobaltActivity = mCobaltActivityRef.get();
+        switch (currentFocusedViewId) {
```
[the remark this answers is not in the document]
haozheng-cobalt: mCobaltActivityRef would never be null because CobaltActivity is a weak reference and outlive a11y.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltA11yHelper.java
```diff
@@ -1,223 +1,252 @@
...
+        // is the relative bounds to each other.
+        node.setBoundsInParent(
+                new Rect(
+                        x * FAKE_VIEW_WIDTH,
+                        y * FAKE_VIEW_HEIGHT,
+                        x * FAKE_VIEW_WIDTH + FAKE_VIEW_WIDTH,
+                        y * FAKE_VIEW_HEIGHT + FAKE_VIEW_HEIGHT));
+        node.setText("");
+
+        if (virtualViewId >= 1 || virtualViewId <= 9) {
```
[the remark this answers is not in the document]
haozheng-cobalt: Same as before.

Awallky:
@haozheng-cobalt I filed https://b.corp.google.com/issues/441347831 to track the addition of histograms to understand which of the nullness issues is most prevalent on end-user devices.

haozheng-cobalt:
> @haozheng-cobalt I filed https://b.corp.google.com/issues/441347831 to track the addition of histograms to understand which of the nullness issues is most prevalent on end-user devices.

thanks.

PR #6979, opened 2025-08-25
Remove the custom thread.h from boringssl

y4vor:
This was a left over customization from Cobalt 25
which we don't need anymore.

Issue: 398292397


andrewsavage1 (approved):
Great cleanup!

PR #6981, opened 2025-08-25
media: Add tests to MediaCapabilitiesCache

johnedocampo:
This PR adds performance tests to the MediaCapabilitiesCache class, to ensure that the time it takes to call `MediaCapabilitiesCache` functions are sufficient for uncached and cached values. The timings used for these tests were created by timing the functions calls on a  Chromecast with Google TV device (lower-end device), and then setting the times based on the reported times, with some leeway for slower android devices we support.

`MediaCapabilitiesCache` requires an instance of StarboardBridge to use for its queries. Using the normal constructor for StarboardBridge is not ideal as its implementation is somewhat coupled with CobaltActivity, which isn't needed for these tests. To address this, this PR also adds some new JNI methods for `StarboardBridge.java` to initialize an isolated StarboardBridge instance for testing.

Bug: 433773029

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
@@ -659,16 +659,18 @@ public class StarboardBridge {
   }
 
   // Creates a StarboardBridge instance that can then be used for testing purposes.
+  // This method does not return a StarboardBridge object as the StarboardBridge constructor
+  // initializes the static instance of StarboardBridge.
   @CalledByNative
   public static void initializeForTesting() {
-    Holder<Activity> activityHolder = new Holder<>();
-    Holder<Service> serviceHolder = new Holder<>();
-    new StarboardBridge(ContextUtils.getApplicationContext(),
-      activityHolder,
-      serviceHolder,
-      new ArtworkDownloaderDefault(),
+    new StarboardBridge(
+      ContextUtils.getApplicationContext(),
+      /*activityHolder=*/ new Holder<>(),
+      /*serviceHolder=*/ new Holder<>(),
+      /*artworkDownloader=*/ new ArtworkDownloaderDefault(),
       new String[0],
       "");
+
   }
 
   // Returns the saved app start timestamp.
--- a/starboard/android/shared/media_capabilities_cache_performance_test.cc
+++ b/starboard/android/shared/media_capabilities_cache_performance_test.cc
@@ -18,6 +18,7 @@
 
 #include <gtest/gtest.h>
 #include <chrono>
+#include <mutex>
 
 #include "starboard/common/log.h"
 
@@ -26,27 +27,40 @@ namespace {
 
 // 200 milliseconds is chosen for the time limit on uncached calls, mainly due
 // to the amount of time it potentially could take for a MediaCapabilitiesCache
-// query on a cold start. Once the cache is warmed up, uncached queries are much
-// faster.
+// query on an empty cache.
 //
 // Cached queries are extremely fast, and usually are timed to take 0
 // milliseconds for a call. We set the cache time limit to 5 ms, to allow some
 // leeway if the function call is slowed down due to other external means.
-const int64_t kUncacheTimeLimitMs = 200;
-const int64_t kCacheTimeLimitMs = 5;
+constexpr int64_t kUncacheTimeLimitMs = 200;
+constexpr int64_t kCacheTimeLimitMs = 5;
+
+// Since these tests rely on the same resource, it is possible that one test
+// could mess up the status of one test if ran in parallel. To make sure this
+// doesn't happen, we use a mutex to lock each test when ran.
+std::mutex media_cache_test_mutex;
 
 class MediaCapabilitiesCachePerformanceTest : public ::testing::Test {
  protected:
   static void SetUpTestSuite() {
     JNIEnv* env = base::android::AttachCurrentThread();
     StarboardBridge::InitializeForTesting(env);
+
+    // Make an initial call to MediaCapabilitiesCache, as the first ever call to
+    // the cache can be extremely slow (up to 700 ms). This behavior is
+    // something that we don't want influencing the tests.
+    MediaCapabilitiesCache::GetInstance()->IsWidevineSupported();
   }
 
   void SetUp() override {
+    // Lock the test so that other tests can't start while this one is running.
+    media_cache_test_mutex.lock();
     instance_ = MediaCapabilitiesCache::GetInstance();
     instance_->ClearCache();
     MimeSupportabilityCache::GetInstance()->ClearCachedMimeSupportabilities();
   }
+
+  void TearDown() override { media_cache_test_mutex.unlock(); }
 
   MediaCapabilitiesCache* instance_ = nullptr;
 };
@@ -61,15 +75,16 @@ void TimeFunctionCall(const std::string& test_name, Func func) {
         .count();
   };
 
-  const int64_t uncached_duration = measure_duration_ms();
-  SB_LOG(INFO) << test_name << " - 1st call (uncached): " << uncached_duration
+  const int64_t uncached_duration_ms = measure_duration_ms();
+  SB_LOG(INFO) << test_name
+               << " - 1st call (uncached): " << uncached_duration_ms
                << " milliseconds.";
-  EXPECT_LT(uncached_duration, kUncacheTimeLimitMs);
+  EXPECT_LT(uncached_duration_ms, kUncacheTimeLimitMs);
 
-  const int64_t cached_duration = measure_duration_ms();
-  SB_LOG(INFO) << test_name << " - 2nd call (cached): " << cached_duration
+  const int64_t cached_duration_ms = measure_duration_ms();
+  SB_LOG(INFO) << test_name << " - 2nd call (cached): " << cached_duration_ms
                << " milliseconds.";
-  EXPECT_LT(cached_duration, kCacheTimeLimitMs);
+  EXPECT_LT(cached_duration_ms, kCacheTimeLimitMs);
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, IsWidevineSupported) {
@@ -84,73 +99,67 @@ TEST_F(MediaCapabilitiesCachePerformanceTest, IsCbcsSchemeSupported) {
 
 TEST_F(MediaCapabilitiesCachePerformanceTest,
        IsHDRTransferCharacteristicsSupported) {
-  const SbMediaTransferId smpte_st_2084 = kSbMediaTransferIdSmpteSt2084;
-  TimeFunctionCall(
-      "IsHDRTransferCharacteristicsSupported (SmpteSt2084)", [&]() {
-        instance_->IsHDRTransferCharacteristicsSupported(smpte_st_2084);
-      });
+  TimeFunctionCall("IsHDRTransferCharacteristicsSupported (SmpteSt2084)",
+                   [&]() {
+                     instance_->IsHDRTransferCharacteristicsSupported(
+                         /*transfer_id=*/kSbMediaTransferIdSmpteSt2084);
+                   });
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, IsPassthroughSupported) {
-  const SbMediaAudioCodec ac3_codec = kSbMediaAudioCodecAc3;
-  TimeFunctionCall("IsPassthroughSupported (AC3)",
-                   [&]() { instance_->IsPassthroughSupported(ac3_codec); });
+  TimeFunctionCall("IsPassthroughSupported (AC3)", [&]() {
+    instance_->IsPassthroughSupported(/*codec=*/kSbMediaAudioCodecAc3);
+  });
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, GetAudioConfiguration) {
-  const int index = 0;
   SbMediaAudioConfiguration configuration;
   TimeFunctionCall("GetAudioConfiguration", [&]() {
-    instance_->GetAudioConfiguration(index, &configuration);
+    instance_->GetAudioConfiguration(/*index=*/0, &configuration);
   });
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, HasAudioDecoderFor) {
-  const std::string mime_type = "audio/mp4; codecs=\"mp4a.40.2\"";
-  const int bitrate = 192000;
   TimeFunctionCall("HasAudioDecoderFor", [&]() {
-    instance_->HasAudioDecoderFor(mime_type, bitrate);
+    instance_->HasAudioDecoderFor(
+        /*mime_type=*/"audio/mp4; codecs=\"mp4a.40.2\"",
+        /*bitrate=*/192'000);
   });
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, HasVideoDecoderFor) {
-  const std::string mime_type = "video/avc; codecs=\"avc1.4d401f\"";
-  const bool must_support_secure = false;
-  const bool must_support_hdr = false;
-  const bool must_support_tunnel_mode = false;
-  const int frame_width = 1920;
-  const int frame_height = 1080;
-  const int bitrate = 8000000;
-  const int fps = 30;
   TimeFunctionCall("HasVideoDecoderFor", [&]() {
-    instance_->HasVideoDecoderFor(mime_type, must_support_secure,
-                                  must_support_hdr, must_support_tunnel_mode,
-                                  frame_width, frame_height, bitrate, fps);
+    instance_->HasVideoDecoderFor(
+        /*mime_type=*/"video/avc; codecs=\"avc1.4d401f\"",
+        /*must_support_secure=*/false,
+        /*must_support_hdr=*/false,
+        /*must_support_tunnel_mode=*/false,
+        /*frame_width=*/1'920,
+        /*frame_height=*/1'080,
+        /*bitrate=*/8'000'000,
+        /*fps=*/30);
   });
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, FindAudioDecoder) {
-  const std::string mime_type = "audio/eac-3";
-  const int bitrate = 640000;
-  TimeFunctionCall("FindAudioDecoder",
-                   [&]() { instance_->FindAudioDecoder(mime_type, bitrate); });
+  TimeFunctionCall("FindAudioDecoder", [&]() {
+    instance_->FindAudioDecoder(/*mime_type=*/"audio/eac-3",
+                                /*bitrate=*/640'000);
+  });
 }
 
 TEST_F(MediaCapabilitiesCachePerformanceTest, FindVideoDecoder) {
-  const std::string mime_type = "video/x-vnd.on2.vp9";
-  const bool must_support_secure = false;
-  const bool must_support_hdr = false;
-  const bool require_software_codec = false;
-  const bool must_support_tunnel_mode = false;
-  const int frame_width = 1920;
-  const int frame_height = 1080;
-  const int bitrate = 8000000;
-  const int fps = 30;
   TimeFunctionCall("FindVideoDecoder", [&]() {
-    instance_->FindVideoDecoder(mime_type, must_support_secure,
-                                must_support_hdr, require_software_codec,
-                                must_support_tunnel_mode, frame_width,
-                                frame_height, bitrate, fps);
+    instance_->FindVideoDecoder(
+        /*mime_type=*/"video/x-vnd.on2.vp9",
+        /*must_support_secure=*/false,
+        /*must_support_hdr=*/false,
+        /*require_software_codec=*/false,
+        /*must_support_tunnel_mode=*/false,
+        /*frame_width=*/1'920,
+        /*frame_height=*/1'080,
+        /*bitrate=*/8'000'000,
+        /*fps=*/30);
   });
 }
 
--- a/starboard/android/shared/starboard_bridge.cc
+++ b/starboard/android/shared/starboard_bridge.cc
@@ -173,8 +173,8 @@ void StarboardBridge::Initialize(JNIEnv* env, jobject obj) {
 
 // static
 void StarboardBridge::InitializeForTesting(JNIEnv* env) {
-  SB_DCHECK(env);
-  return Java_StarboardBridge_initializeForTesting(env);
+  SB_CHECK(env);
+  Java_StarboardBridge_initializeForTesting(env);
 }
 
 long StarboardBridge::GetAppStartTimestamp(JNIEnv* env) {


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -663,6 +664,20 @@ protected long getAppStartTimestamp() {
...
+    Holder<Activity> activityHolder = new Holder<>();
+    Holder<Service> serviceHolder = new Holder<>();
+    StarboardBridge bridge =
+        new StarboardBridge(
+            ContextUtils.getApplicationContext(),
+            activityHolder,
+            serviceHolder,
+            new ArtworkDownloaderDefault(),
+            new String[0],
+            "");
```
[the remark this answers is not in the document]
johnedocampo: Done. We did not need the value anyway, just only the function call.

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  end = std::chrono::high_resolution_clock::now();
+  duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+  SB_LOG(INFO) << test_name << " - 2nd call (cached): " << duration.count()
+               << " milliseconds.";
+  EXPECT_LT(duration.count(), kCacheTimeLimitMs);
+}
+
+TEST_F(MediaCapabilitiesCachePerformanceTest, IsWidevineSupported) {
+  TimeFunctionCall("IsWidevineSupported",
+                   [&]() { instance_->IsWidevineSupported(); });
```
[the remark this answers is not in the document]
johnedocampo: I think for the logs we should use the name of the function we are testing. This should be fine to keep for now.

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  const int frame_width_sdr = 1920;
+  const int frame_height_sdr = 1080;
+  const int bitrate_sdr = 8000000;
+  const int fps_sdr = 30;
+  TimeFunctionCall("FindVideoDecoder (1080p SDR)", [&]() {
+    instance_->FindVideoDecoder(
+        mime_type, must_support_secure_sdr, must_support_hdr_sdr,
+        require_software_codec_sdr, must_support_tunnel_mode_sdr,
+        frame_width_sdr, frame_height_sdr, bitrate_sdr, fps_sdr);
+  });
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  const int frame_width_sdr = 1920;
+  const int frame_height_sdr = 1080;
+  const int bitrate_sdr = 8000000;
+  const int fps_sdr = 30;
+  TimeFunctionCall("HasVideoDecoderFor (1080p SDR)", [&]() {
+    instance_->HasVideoDecoderFor(mime_type, must_support_secure_sdr,
+                                  must_support_hdr_sdr,
+                                  must_support_tunnel_mode_sdr, frame_width_sdr,
+                                  frame_height_sdr, bitrate_sdr, fps_sdr);
+  });
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,157 @@
...
+  EXPECT_LT(duration.count(), kUncacheTimeLimitMs);
+
+  start = std::chrono::high_resolution_clock::now();
+  func();
+  end = std::chrono::high_resolution_clock::now();
+  duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
+  SB_LOG(INFO) << test_name << " - 2nd call (cached): " << duration.count()
+               << " milliseconds.";
+  EXPECT_LT(duration.count(), kCacheTimeLimitMs);
+}
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/BUILD.gn
```diff
@@ -397,6 +397,7 @@ source_set("starboard_platform_tests") {
               "//starboard/shared/starboard/drm/drm_test_helpers.h",
               "//starboard/shared/starboard/starboard_feature_test.cc",
               "jni_env_ext_test.cc",
+              "media_capabilities_cache_performance_test.cc",
```
kjyoun, line 385: CI does not run `starboard_platform_tests` 

Can you add comment about it and instructions how to run these tests manually?

johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,158 @@
...
+
+// 200 milliseconds is chosen for the time limit on uncached calls, mainly due
+// to the amount of time it potentially could take for a MediaCapabilitiesCache
+// query on a cold start. Once the cache is warmed up, uncached queries are much
+// faster.
+//
+// Cached queries are extremely fast, and usually are timed to take 0
+// milliseconds for a call. We set the cache time limit to 5 ms, to allow some
+// leeway if the function call is slowed down due to other external means.
+const int64_t kUncacheTimeLimitMs = 200;
```
kjyoun: constexpr

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,158 @@
...
+void TimeFunctionCall(const std::string& test_name, Func func) {
+  auto measure_duration_ms = [&]() {
+    auto start = std::chrono::high_resolution_clock::now();
+    func();
+    auto end = std::chrono::high_resolution_clock::now();
+    return std::chrono::duration_cast<std::chrono::milliseconds>(end - start)
+        .count();
+  };
+
+  const int64_t uncached_duration = measure_duration_ms();
```
kjyoun: can you add "_ms" suffix for readability as you do for other variables?

here and below

starboard/android/shared/starboard_bridge.cc
```diff
@@ -172,6 +172,12 @@ void StarboardBridge::Initialize(JNIEnv* env, jobject obj) {
   j_starboard_bridge_.Reset(env, obj);
 }
 
+// static
+void StarboardBridge::InitializeForTesting(JNIEnv* env) {
+  SB_DCHECK(env);
```
kjyoun: SB_CHECK for new code

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -660,6 +661,19 @@ protected long getAppStartTimestamp() {
     return mAppStartTimestamp;
   }
 
+  // Creates a StarboardBridge instance that can then be used for testing purposes.
+  @CalledByNative
+  public static void initializeForTesting() {
+    Holder<Activity> activityHolder = new Holder<>();
+    Holder<Service> serviceHolder = new Holder<>();
+    new StarboardBridge(ContextUtils.getApplicationContext(),
+      activityHolder,
```
kjyoun: how about this. if you define activityHolder and serviceHolder for readability?

```
/*activityHolder=*/new Holder<>(),
/*serviceHolder=*/new Holder<>(),
```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -660,6 +661,19 @@ protected long getAppStartTimestamp() {
     return mAppStartTimestamp;
   }
 
+  // Creates a StarboardBridge instance that can then be used for testing purposes.
+  @CalledByNative
+  public static void initializeForTesting() {
+    Holder<Activity> activityHolder = new Holder<>();
+    Holder<Service> serviceHolder = new Holder<>();
+    new StarboardBridge(ContextUtils.getApplicationContext(),
```
kjyoun: can you add comments about why you create StarboardBridge instance, even though we don't do anything with it?

e.g.)
// Create StarbordBridge to trigger a side effect (e.g. initalization of certain static variable)

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,158 @@
...
+  const int index = 0;
+  SbMediaAudioConfiguration configuration;
+  TimeFunctionCall("GetAudioConfiguration", [&]() {
+    instance_->GetAudioConfiguration(index, &configuration);
+  });
+}
+
+TEST_F(MediaCapabilitiesCachePerformanceTest, HasAudioDecoderFor) {
+  const std::string mime_type = "audio/mp4; codecs=\"mp4a.40.2\"";
+  const int bitrate = 192000;
```
kjyoun: nit  `192'000;` to improve number readability

and you can use `constexpr` here

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,158 @@
...
+  const SbMediaAudioCodec ac3_codec = kSbMediaAudioCodecAc3;
+  TimeFunctionCall("IsPassthroughSupported (AC3)",
+                   [&]() { instance_->IsPassthroughSupported(ac3_codec); });
+}
+
+TEST_F(MediaCapabilitiesCachePerformanceTest, GetAudioConfiguration) {
+  const int index = 0;
+  SbMediaAudioConfiguration configuration;
+  TimeFunctionCall("GetAudioConfiguration", [&]() {
+    instance_->GetAudioConfiguration(index, &configuration);
```
kjyoun: nit: `/*index=*/0` and remove line 101

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,158 @@
...
+}
+
+TEST_F(MediaCapabilitiesCachePerformanceTest, HasVideoDecoderFor) {
+  const std::string mime_type = "video/avc; codecs=\"avc1.4d401f\"";
+  const bool must_support_secure = false;
+  const bool must_support_hdr = false;
+  const bool must_support_tunnel_mode = false;
+  const int frame_width = 1920;
+  const int frame_height = 1080;
+  const int bitrate = 8000000;
```
kjyoun: nit: `8'000'000`

starboard/android/shared/media_capabilities_cache_performance_test.cc
```diff
@@ -0,0 +1,158 @@
...
+
+TEST_F(MediaCapabilitiesCachePerformanceTest, FindAudioDecoder) {
+  const std::string mime_type = "audio/eac-3";
+  const int bitrate = 640000;
+  TimeFunctionCall("FindAudioDecoder",
+                   [&]() { instance_->FindAudioDecoder(mime_type, bitrate); });
+}
+
+TEST_F(MediaCapabilitiesCachePerformanceTest, FindVideoDecoder) {
+  const std::string mime_type = "video/x-vnd.on2.vp9";
```
kjyoun: nit: How about this?
here and other places

```
TimeFunctionCall("FindVideoDecoder", [this] {
  instance_->FindVideoDecoder(
      /*mime_type=*/"video/x-vnd.on2.vp9",
      /*must_support_secure=*/false,
      /*must_support_hdr=*/false,
      /*require_software_codec=*/false,
      /*must_support_tunnel_mode=*/false,
      /*frame_width=*/1'920,
      /*frame_height=*/1'080,
      /*bitrate=*/8'000'000,
      /*fps=*/30);
});
```

kjyoun: [go/cstyle#Function_Argument_Comments](http://go/cstyle#Function_Argument_Comments)

starboard/android/shared/media_capabilities_cache_performance_test.cc
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
johnedocampo: Done.

starboard/android/shared/starboard_bridge.cc
johnedocampo: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
johnedocampo: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_performance_test.cc
johnedocampo: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -155,6 +157,36 @@ public StarboardBridge(
...
+    nativeApp =
+        StarboardBridgeJni.get()
+            .startNativeStarboard(
+                getAssetsFromContext(),
+                getFilesAbsolutePath(),
+                getCacheAbsolutePath(),
+                getNativeLibraryDir());
+
+    this.isAmatiDevice = appContext.getPackageManager().hasSystemFeature(AMATI_EXPERIENCE_FEATURE);
+  }
```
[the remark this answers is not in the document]
johnedocampo: I think its okay to keep as is for now. It may be likely that future tests that need an isolated test instance of StarboardBridge may need things initialized a different way from the normal constructor.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -177,6 +209,9 @@ long startNativeStarboard(
     void setAndroidPlayServicesVersion(long version);
 
     boolean isReleaseBuild();
+
+    @VisibleForTesting
+    boolean initTestJNI(StarboardBridge starboardBridge);
```
Awallky: nit: The name does not need the JNI post-fix as it is already labelled by the native methods interface.

starboard/android/shared/starboard_bridge.cc
```diff
@@ -173,6 +185,12 @@ void StarboardBridge::Initialize(JNIEnv* env, jobject obj) {
   j_starboard_bridge_.Reset(env, obj);
 }
 
+// static
+void StarboardBridge::InitializeForTesting(JNIEnv* env) {
+  SB_CHECK(env);
+  return Java_StarboardBridge_initializeForTesting(env);
```
Awallky: I don't think this needs a return statement here, since the return value of the method called is void.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -155,6 +157,36 @@ public StarboardBridge(
...
+  // Constructor for using StarboardBridge in isolation. The normal constructor for StarboardBridge
+  // is coupled with other runtime objects like CobaltActivity. This constructor
+  // initializes the bare minimum needed for StarboardBridge to run in a unit testing environment.
+  @VisibleForTesting
+  public StarboardBridge(
+      Context appContext,
+      Holder<Activity> activityHolder,
+      Holder<Service> serviceHolder,
+      String[] args) {
+    StarboardBridgeJni.get().initTestJNI(this);
```
Awallky: Since this is initialized before setting some necessary fields, are we sure there is no race between this ctor and other methods that might depend on those fields being non-null? If not, I'd consider moving this init call to the end of the ctor.

Awallky: Or maybe call the init and then finishing with `StarboardBridgeJni.get().startNativeStarboard`.

starboard/android/shared/starboard_bridge.cc
johnedocampo: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
johnedocampo: Thanks to https://github.com/youtube/cobalt/pull/7532, this function is now identical to the normal `initJNI` function. Thus, this function has been removed.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -657,6 +658,21 @@ protected long getAppStartTimestamp() {
     return mAppStartTimestamp;
   }
 
+  // Creates a StarboardBridge instance that can then be used for testing purposes.
+  // This method does not return a StarboardBridge object as the StarboardBridge constructor
+  // initializes the static instance of StarboardBridge.
+  @CalledByNative
+  public static void initializeForTesting() {
```
kjyoun, line 665: qq: Do we still need this?

Now JNI code does not need Starboard instance.

johnedocampo: This method still needs to be here as we still need to initialize a StarboardBridge instance from the c++ side. The other added functions however are no longer needed, and have been removed from this PR.

johnedocampo:
> The changes look fine as-is % a few comments. I am hesitant to approve simply because performance tests are dependent upon a responsive system, which can lead to test flakes. For example, a shared resource, such as a file, between multiple tests could lead to delays in one, if run in parallel. Do you know if we run any risk of this in these tests?

@Awallky This is a valid concern. With the current code I believe it is possible that in between timing calls for a test, another test could initialize and reset the cache. To address this potential parallelism issue, I've added a mutex that will lock each test. PTAL.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
johnedocampo: Thanks to https://github.com/youtube/cobalt/pull/7532, we can just use the normal ctor for StarboardBridge.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
kjyoun: looks like that it's because of this code.

[src](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/media_capabilities_cache.cc;l=440;drc=d3bed5c8afa34c2263d077636166e07eb811aa20)
```
  ScopedJavaLocalRef<jintArray> j_supported_hdr_types =
      StarboardBridge::GetInstance()->GetSupportedHdrTypes(env);
```

Then isn't it simpler to add test-only method for `MemCapabilitesCache`.

e.g.)
```
class MediaCapabilitiesCache {

 static std::unique_ptr<MediaCapabilitiesCache>  createInstanceForTest(std::set<SbMediaTransferId> hdr_override) ;

private:
  std:optional<std::set<SbMediaTransferId>> hdr_override_for_testing_;

then 
void MediaCapabilitiesCache::UpdateMediaCapabilities_Locked() {
...
    supported_transfer_ids_ = 
         hdr_override_for_testing_ ?
         *hdr_override_for_testing_ : GetSupportedHdrTypes();
  }
}
````
Then we can isolate all test-related logic in MediaCapabilitiesCache
For me, it looks like an overkill to create whole StarboardBridge instance.

Comment (johnedocampo) [reply cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java:665]: I decided to use the construct a whole StarboardBridge instance due to StarboardBridge.java also constructing AudioOutputManager ([src](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java;l=132;drc=d3bed5c8afa34c2263d077636166e07eb811aa20?q=starboardbridge.&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain)), which is needed by [LoadAudioConfiguration_Locked](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/media_capabilities_cache.cc;l=498;drc=d3bed5c8afa34c2263d077636166e07eb811aa20?q=media_capabilities_cache&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) and [GetIsPassthroughSupported](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/media_capabilities_cache.cc;l=145;drc=d3bed5c8afa34c2263d077636166e07eb811aa20?q=media_capabilities_cache&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain). The constructor for [AudioOutputManager](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/audio_output_manager.cc;l=144;drc=d3bed5c8afa34c2263d077636166e07eb811aa20?q=Audio_output_manager.cc&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) relies on a StarboardBridge instance as well.

I also feel that due to this being a performance test, we want to test the production pipeline as close as possible. I do agree though that constructing an entire StarboardBridge instance is a bit overkill for just these tests, but I think discussions about how we address that begin to come out of scope for this PR. WDYT? @kjyoun

**Comment by kjyoun** [comment]
> > The changes look fine as-is % a few comments. I am hesitant to approve simply because performance tests are dependent upon a responsive system, which can lead to test flakes. For example, a shared resource, such as a file, between multiple tests could lead to delays in one, if run in parallel. Do you know if we run any risk of this in these tests?
> 
> @Awallky This is a valid concern. With the current code I believe it is possible that in between timing calls for a test, another test could initialize and reset the cache. To address this potential parallelism issue, I've added a mutex that will lock each test. PTAL.

I think that it's not a good idea to put performance test in the unit tests.
These limits are fragile. They can change depending on the devices.  Even They will change when we run this test alone in a single thread or 1,000 tests in 1'000 thread in parallel (to debug flakiness).

```
// 200 milliseconds is chosen for the time limit on uncached calls, mainly due
// to the amount of time it potentially could take for a MediaCapabilitiesCache
// query on an empty cache.
//
// Cached queries are extremely fast, and usually are timed to take 0
// milliseconds for a call. We set the cache time limit to 5 ms, to allow some
// leeway if the function call is slowed down due to other external means.
constexpr int64_t kUncacheTimeLimitMs = 200;
constexpr int64_t kCacheTimeLimitMs = 5;
```

I am strongly against comparing performance to any prefixed number in the unit test, especially timing.
Sorry,  I have to express this opinion from the beginning. I didn't read this commit thorough enough.

If you want to see the performance, you can use benchmark tests.

[starboard/benchmark/memory_benchmark.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/benchmark/thread_benchmark.cc)

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
kjyoun: Hmm.. it looks like that we need more serious refactoring to create `MediaCapabilitiesCache` w/o `StarboardBridge`.

kjyoun:
Here is what Gemini says.

---

## 🧐 Why Performance Assertions in Unit Tests are Problematic

* **Fragility and Flakiness :** Performance is inherently non-deterministic. A test asserting a fixed time limit like `kCacheTimeLimitMs = 5` is a **"flaky" test** waiting to happen.
    * **Shared Resources:** As the initial reviewer pointed out, parallel execution, shared resources (I/O, network, locks), or simply **OS scheduling** can introduce delays that cause the test to fail one day and pass the next, even with no code change. The introduced mutex only guarantees *correctness* but might actually *increase* the flake potential by introducing contention.
    * **Device/System Dependence:** The performance of the code depends entirely on the machine it's running on (CPU speed, memory bandwidth, current load). A limit that passes on a powerful developer machine might consistently fail on a slower CI server or embedded device.

* **Misleading Results:** The purpose of a **unit test** is to verify **correctness** (i.e., does the code do what it's supposed to do?). The purpose of a **benchmark** is to measure **performance** (i.e., how fast/efficiently does the code do it?). Mixing them confuses the roles. A test failure due to a time limit often doesn't mean the logic is wrong, but that the execution environment changed.

* **Poor Maintainability:** These fixed-number constants (`200ms`, `5ms`) are **magic numbers**. Every time a dependency is updated, the compiler changes, or the build system is tweaked, someone has to re-evaluate and potentially tune these numbers, which is tedious and error-prone.

---

johnedocampo:
> I think that it's not a good idea to put performance test in the unit tests. These limits are fragile. They can change depending on the devices. Even They will change when we run this test alone in a single thread or 1,000 tests in 1'000 thread in parallel (to debug flakiness).
> 
> ```
> // 200 milliseconds is chosen for the time limit on uncached calls, mainly due
> // to the amount of time it potentially could take for a MediaCapabilitiesCache
> // query on an empty cache.
> //
> // Cached queries are extremely fast, and usually are timed to take 0
> // milliseconds for a call. We set the cache time limit to 5 ms, to allow some
> // leeway if the function call is slowed down due to other external means.
> constexpr int64_t kUncacheTimeLimitMs = 200;
> constexpr int64_t kCacheTimeLimitMs = 5;
> ```
> 
> I am strongly against comparing performance to any prefixed number in the unit test, especially timing. Sorry, I have to express this opinion from the beginning. I didn't read this commit thorough enough.
> 
> If you want to see the performance, you can use benchmark tests.
> 
> [starboard/benchmark/memory_benchmark.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/benchmark/thread_benchmark.cc)

@kjyoun Thank you for the feedback. I didn't see this library before, but it looks much better suited for performance testing.

> I am strongly against comparing performance to any prefixed number in the unit test, especially timing.

I think we still would like a way to somehow detect regressions in the performance of the cache. If the performance testing route and comparing times to prefixed numbers is ill-advised, I think a possible alternative would be to compare the two times of the uncached value and cached value, and ensure that the cached time is faster than the uncached time.

johnedocampo:
> I think that it's not a good idea to put performance test in the unit tests. These limits are fragile. They can change depending on the devices. Even They will change when we run this test alone in a single thread or 1,000 tests in 1'000 thread in parallel (to debug flakiness).
 > ```
 > // 200 milliseconds is chosen for the time limit on uncached calls, mainly due
 > // to the amount of time it potentially could take for a MediaCapabilitiesCache
 > // query on an empty cache.
 > //
 > // Cached queries are extremely fast, and usually are timed to take 0
 > // milliseconds for a call. We set the cache time limit to 5 ms, to allow some
 > // leeway if the function call is slowed down due to other external means.
 > constexpr int64_t kUncacheTimeLimitMs = 200;
 > constexpr int64_t kCacheTimeLimitMs = 5;
 > ```
> I am strongly against comparing performance to any prefixed number in the unit test, especially timing. Sorry, I have to express this opinion from the beginning. I didn't read this commit thorough enough.
> If you want to see the performance, you can use benchmark tests.
> [starboard/benchmark/memory_benchmark.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/benchmark/thread_benchmark.cc)
 
 @kjyoun Thank you for the feedback. I didn't see this library before, but it looks much better suited for performance testing.

> I am strongly against comparing performance to any prefixed number in the unit test, especially timing.
 
I think we still would like a way to somehow detect regressions in the performance of the cache (maybe if we can detect if the cache isn't actually caching its values). If the performance testing route and comparing times to prefixed numbers is ill-advised, I think a possible alternative would be to compare the two times of the uncached value and cached value, and ensure that the cached time is faster than the uncached time. This keeps the timings relative to the test environment itself, and should be less error-prone to flakiness.

kjyoun:
I believe our primary goal should be to verify "behavior" of MediaCapabilitiesCache, not necessarily its performance. We
  need to confirm that it effectively caches capabilities rather than querying them every time. As long as it behaves
  correctly, its performance is less critical and outside our direct control.

To properly test this behavior, dependency injection becomes essential, as it allows us to precisely track how many times `MediaCapabilitiesCache` interacts with its underlying provider.

You can see my POC at: https://github.com/youtube/cobalt/pull/7564. While the code does become a bit more complex with this approach, it enables comprehensive testing of `MediaCapabilitiesCache`, which I believe makes it a worthwhile investment. WDYT?

johnedocampo:
Closing this pull request as the ideas from https://github.com/youtube/cobalt/pull/7564 seem to be a better solution in verifying the correctness of the cache.

Awallky:
> I believe our primary goal should be to verify "behavior" of MediaCapabilitiesCache, not necessarily its performance. We
need to confirm that it effectively caches capabilities rather than querying them every time. 

+1 to KJ's comment. In WebView, we always mocked out the system clock, set it to some default start value, executed code under test, stopped to advance the mock clock values as needed, and verified the various cases, but we tended to avoid testing against timing behaviors in unit tests. I think there were probably perfbots that were separate from them.
It's just very likely to introduce non-debuggable test flakes that way.

Typically the way to monitor performance is by adding some histograms of some sort to monitor the performance. We  address poor performance if those timing histograms show poor performance.

PR #6984, opened 2025-08-25
Remove unused egl_swap_buffers.cc and related code

sherryzy:
This file is no longer used, so this change removes it and all related code, including the `starboard_egl_wrapped_functions` static library and the `SwapBuffersBegin` and `SwapBuffersEnd` functions. In Cobalt 25, it's used in cobalt rendering backend, and that is gone entirely in Cobalt 26.

Bug: 401006482


starboard/linux/x64x11/clang/3.9/BUILD.gn
```diff
@@ -31,8 +31,7 @@ static_library("starboard_platform") {
   configs += [ "//starboard/build/config:starboard_implementation" ]
 
   # This is needed at link time, but not directly linked in
-  data_deps =
-      [ "//starboard/linux/x64x11/shared:starboard_egl_wrapped_functions" ]
+  data_deps = []
```
andrewsavage1: You can just remove this now

starboard/linux/x64x11/gcc/6.3/BUILD.gn
```diff
@@ -31,8 +31,7 @@ static_library("starboard_platform") {
   configs += [ "//starboard/build/config:starboard_implementation" ]
 
   # This is needed at link time, but not directly linked in
-  data_deps =
-      [ "//starboard/linux/x64x11/shared:starboard_egl_wrapped_functions" ]
+  data_deps = []
```
andrewsavage1: Remove

andrewsavage1 (approved):

PR #6985, opened 2025-08-25
cobalt/testing: Fix missing symbol for cobalt_browsertests

Awallky:
During linking of cobalt_browsertests, cobalt is missing an undefined symbol: `switches::kExposeInternalsForTesting`. This adds the proper library so it can be linked properly.
This also removes an unused/duplicate declaration of `switches::kExposeInternalsForTesting`.

Test: Build `cobalt_browsertests`
Bug: 441144322

### DIFF
--- a/cobalt/testing/browser_tests/BUILD.gn
+++ b/cobalt/testing/browser_tests/BUILD.gn
@@ -173,6 +173,7 @@ if (is_android) {
   }
 }
 
+# Note: Testing on official builds is not currently supported.
 test("cobalt_browsertests") {
   use_xvfb = use_xvfb_in_this_config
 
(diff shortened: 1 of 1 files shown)


haozheng-cobalt:
I think eventually the right way to handle this is to combine cobalt/browser/switches.h with cobalt/shell/common/shell_test_switches.h and cobalt/shell/common/shell_switches.h.

cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -219,6 +219,7 @@ test("cobalt_browsertests") {
     "//build:chromecast_buildflags",
     "//cc/slim",
     "//cobalt/shell:browsertests_mojom",
+    "//cobalt/shell:browsertests_sources",
```
haozheng-cobalt: Was this causing issue? Was browsertests_sources not pulled in by //cobalt/shell:cobalt_shell_lib?

Awallky: For gold builds, it still fails to link `switches::kExposeInternalsForTesting` in cobalt/testing/browser_tests/media_browsertest.cc.

haozheng-cobalt:
Yea, the reason is that we can only run browser tests in non-official builds now. Gold will fail when running browser tests.

haozheng-cobalt (approved):

PR #7013, opened 2025-08-27
Fix IllegalStateException from PhysicalDisplayAndroid

zhongqiliang:
The Android OS's Display.getMode() method can throw an [IllegalStateException](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/base/core/java/android/view/DisplayInfo.java;l=793-795?q=DisplayInfo.java)
during HDMI hotplug events. This happens when a display reports a mode ID
not in its list of supported modes.

Wrap the call to getMode() in a try-catch block to prevent application
crashes in these scenarios. The system will recover when a valid
onDisplayChanged event occurs, making this approach safer than crashing.

Bug: 441513616

Change-Id: I0d3e4d0ce2682aaf813996ee434a9db293b26a5c


zhongqiliang:
It is submitted to chromium repo. https://chromium-review.googlesource.com/c/chromium/src/+/6892252

PR #7016, opened 2025-08-27
Introduce `enable_lightweight` argument

kaidokert:
This is a group switch that enables all other lightweight mode switches.

With this, the CI config does not need to be changed for every new introduced change.

### DIFF
--- a/build/config/lightweight.gni
+++ b/build/config/lightweight.gni
@@ -7,8 +7,6 @@ declare_args() {
   enable_lightweight = false
 }
 
-
-# Lightweight build configuration. Only Linux based build is considered.
 declare_args() {
   # Disable Bluetooth support.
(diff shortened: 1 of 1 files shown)


build/config/lightweight.gni
```diff
@@ -2,12 +2,18 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+# Lightweight build configuration. Only Linux based build is considered.
+declare_args() {
+  enable_lightweight = false
+}
+
+
 # Lightweight build configuration. Only Linux based build is considered.
```
andrewsavage1: Redundant now

andrewsavage1 (approved):

build/config/lightweight.gni
kaidokert: Thanks, removed

PR #7020, opened 2025-08-28
Fix race condition in TimeZoneState initialization

jellefoks:
A pure virtual function call to `icu::TimeZone::getOffset` would occur when multiple threads tried to access timezone information simultaneously. This lead to a race condition where the `icu::TimeZone` object was not fully constructed before being used.

This was resolved by adding a mutex to `TimeZoneState` to ensure that the creation and access of the `icu::TimeZone` object is thread-safe. The `GetTimeZone()` method has been replaced with `WithTimeZone()` which executes a function within the mutex's protection.

Bug: 406083355

### DIFF
--- a/cobalt/common/libc/time/time_zone_state.h
+++ b/cobalt/common/libc/time/time_zone_state.h
@@ -15,8 +15,7 @@
 #ifndef COBALT_COMMON_LIBC_TIME_TIME_ZONE_STATE_H_
 #define COBALT_COMMON_LIBC_TIME_TIME_ZONE_STATE_H_
 
-#include <pthread.h>
-#include <functional>
+#include <atomic>
 #include <memory>
 #include <string>
 #include "unicode/timezone.h"
@@ -35,8 +34,8 @@ class TimeZoneState {
   TimeZoneState();
   ~TimeZoneState();
 
-  // Returns a reference to the currently active ICU timezone object.
-  void WithTimeZone(std::function<void(const icu::TimeZone&)> func);
+  // Returns the currently active ICU timezone object.
+  std::shared_ptr<const icu::TimeZone> GetTimeZone();
 
   // Calculates the values for the global C variables (timezone, daylight,
   // tzname).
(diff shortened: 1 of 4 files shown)


cobalt/common/libc/time/time_zone_state.h
```diff
@@ -34,7 +36,7 @@ class TimeZoneState {
   ~TimeZoneState();
 
   // Returns a reference to the currently active ICU timezone object.
```
andrewsavage1: This comment is stale

andrewsavage1 (commented):
Would it instead work if CreateTimeZoneFromEnvironment had a ScopedLock, or would that not fix the race condition? I think the previous interface was simpler which was nice, but maybe there isn't a good way to avoid a WithTimeZone function or similar

jellefoks:
> Would it instead work if CreateTimeZoneFromEnvironment had a ScopedLock, or would that not fix the race condition? I think the previous interface was simpler which was nice, but maybe there isn't a good way to avoid a WithTimeZone function or similar

That would only be safe if we never have to call `CreateTimeZoneFromEnvironment()` again, but `tzset()` (which results in a call to this) can be called at any time with new timezone settings, plus `tset()` is also called implicitly from any time zone explosion or implosion POSIX/libc function (`localtime`/`localtime_r`/`mktime`/`asctime`/`ctime`) with the same effect.

There are in fact unit tests for changing time zones. Those tests do not trigger a failure because those tests don't include racing the `tzset()` call with timezone readers, but that's a limitation of the tests not of the technically possible use of the API.

Now... I'm not fond of the locking in the read path here because performance.

I pondered lock-free approaches, but hesitated using a shared_ptr because the style guide requires a specific good reason. A way around that would be to make our own atomic reference counter and atomic pointer update, and spin for the write locking around that when needed. That would be the most performant for readers, because writes are extremely rare. But... shared_ptr wouldn't need to spin and be even more performant as a result...

Neither will be a hands-down obviously noncontroversial piece of code. If I didn't have to worry about the shared_ptr section in the style guide, that would be my choice, because performance....

But without benchmarks to show the difference, I chose locking to not have to argue a case for using shared_ptr. If you're supportive of that approach I'll implement that in an update.

jellefoks:
> > Would it instead work if CreateTimeZoneFromEnvironment had a ScopedLock, or would that not fix the race condition? I think the previous interface was simpler which was nice, but maybe there isn't a good way to avoid a WithTimeZone function or similar
> 
[snip]

> But without benchmarks to show the difference, I chose locking to not have to argue a case for using shared_ptr. If you're supportive of that approach I'll implement that in an update.

Now this latest push is using a shared_ptr. It has a nice bonus performance side-effect of no-longer needing cloned_zone, saving more performance because the timezone no longer has to be cloned during implosion/explosion.

andrewsavage1 (approved):
I'll probably look at this further when I have more time to think about what's the right solution here, but it seems good to get something in now to fix the crash

PR #7021, opened 2025-08-28
[DNS] Clear SurfaceView through Chromium GPU

johnedocampo:
Bug: 429021006

### DIFF
--- a/media/mojo/services/starboard/starboard_renderer_wrapper.cc
+++ b/media/mojo/services/starboard/starboard_renderer_wrapper.cc
@@ -21,17 +21,24 @@
 #include "base/time/time.h"
 #include "media/base/starboard/starboard_rendering_mode.h"
 #include "media/mojo/services/mojo_media_log.h"
-
+#include "ui/gl/gl_bindings.h"
 namespace media {
 
 StarboardRendererWrapper::StarboardRendererWrapper(
+#if BUILDFLAG(IS_ANDROID)
+    StarboardRendererTraits traits,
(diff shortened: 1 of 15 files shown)


media/mojo/services/starboard/starboard_renderer_wrapper.cc
```diff
@@ -212,4 +233,40 @@ void StarboardRendererWrapper::OnRequestOverlayInfoByStarboard(
...
+    // If it is on the gpu thread, post target_function() directly on it.
+    target_function(target_function_context);
+  } else if (provider->gpu_factory_) {
+    // If it is not on the gpu thread, post target_function() with
+    // |gpu_factory_|.
+    base::WaitableEvent done_event(
+        base::WaitableEvent::ResetPolicy::MANUAL,
+        base::WaitableEvent::InitialState::NOT_SIGNALED);
+    provider->gpu_factory_
+        .AsyncCall(&StarboardGpuFactory::RunSbDecodeTargetFunctionOnGpu)
```
borongc, line 263: Discussed with @jasonzhangxx offline, this will now call all egl calls on gpu thread, which blocks player_worker when destroying VideoDecoder (as it calls `ClearVideoWindow`).

We should run two experiments, one is this one, the other is without waiting it to be finished, so it won't increase the latency to destroy SbPlayer.

That is, we can add a flag here to: (1) the current AsyncCall with WaitableEvent, and (2) the current AsyncCall without WaitableEvent.

PR #7023, opened 2025-08-28
Fix  loader_app  crash issue at launch and also input events processing

santoshmahto7:
There were two issue:
1. loader app crash while wl_surface_attach
      solution(commit): [Resolve egl core functions from SbEglInterface first](https://github.com/youtube/cobalt/commit/ba574f5d02792c70a6a0bfb356bae0463bca8cdb)
2. key event are not handled from remote or keyboard.
    solution(commit): [Handle the key event from kSbInputDeviceTypeRemote device type](https://github.com/youtube/cobalt/commit/db526dfda2e8c6bbacdf24d0d1c14b6855265ee3)

PR #7025, opened 2025-08-28
Handle the key event from kSbInputDeviceTypeRemote device type

santoshmahto7:
Right now, only events with device type kSbInputDeviceTypeKeyboard are handled by starboard ozone implementation, but rdk starboard sends the input event with device type kSbInputDeviceTypeRemote.

For now, key events from device kSbInputDeviceTypeRemote are handled in same way as kSbInputDeviceTypeKeyboard.

Bug: 440527928


santoshmahto7:
This supersedes the PR: https://github.com/youtube/cobalt/pull/7023 which have initial review comments

santoshmahto7:
CI fails but the reason seems to be different. I don't see any option to rerun the CI.

y4vor (approved):

PR #7028, opened 2025-08-28
Add OnMemoryPressure logging

johnxwork:
To help us discover memory pressure handling in logcat.

Bug: 440200929

### DIFF
--- a/third_party/blink/renderer/platform/instrumentation/memory_pressure_listener.cc
+++ b/third_party/blink/renderer/platform/instrumentation/memory_pressure_listener.cc
@@ -112,6 +112,9 @@ void MemoryPressureListenerRegistry::OnMemoryPressure(
     base::MemoryPressureListener::MemoryPressureLevel level) {
   TRACE_EVENT1("blink", "MemoryPressureListenerRegistry::onMemoryPressure",
                "level", level);
+#if BUILDFLAG(IS_COBALT)
+  LOG(INFO) << "Blink handling OnMemoryPressure";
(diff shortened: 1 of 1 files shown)


third_party/blink/renderer/platform/instrumentation/memory_pressure_listener.cc
```diff
@@ -112,6 +112,7 @@ void MemoryPressureListenerRegistry::OnMemoryPressure(
     base::MemoryPressureListener::MemoryPressureLevel level) {
   TRACE_EVENT1("blink", "MemoryPressureListenerRegistry::onMemoryPressure",
                "level", level);
+  LOG(INFO)<<"Blink handling OnMemoryPressure";
```
kjyoun: nit: space before and after <<

LOG(INFO) << "Blink handling OnMemoryPressure";

Doesn't pre-commit fix it?

andrewsavage1: pre-commit doesn't format upstream files

andrewsavage1: Wrap in `#if BUILDFLAG(IS_COBALT)`

kaidokert: Correct - we have Eric working on plugging in pre-commit for upstream files too. We are still probably a few weeks away from that though.

PR #7034, opened 2025-08-29
cobalt/android: Diagnostics for startBrowserProcessesAsync crash

ckim1346:
Throw more descriptive Java exceptions in CobaltActivity.java to pinpoint a [JNI crash](https://primes.corp.google.com/products/5063451/metrics/crashes/clusters/1000000001517437853?sortBy=6&sortOrder=2&duration_seconds=2592000&source=8594998787&reportType=JAVA,NATIVE&trendsTab=overview&timeGranularity=daily). Cases include a failed initialization, low memory, pressing the back key, and then a catch-all if it's not one of those cases. 

Bug: 439066169

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -103,15 +103,10 @@ public abstract class CobaltActivity extends Activity {
     if (!CommandLine.isInitialized()) {
       CommandLine.init(null);
 
-      String[] commandLineArgs =
-          getCommandLineParamsFromIntent(
-              getIntent(), COMMAND_LINE_ARGS_KEY);
+      String[] commandLineArgs = getCommandLineParamsFromIntent(getIntent(), COMMAND_LINE_ARGS_KEY);
       CommandLineOverrideHelper.getFlagOverrides(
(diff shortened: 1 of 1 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -171,13 +171,59 @@ protected void createContent(final Bundle savedInstanceState) {
             new BrowserStartupController.StartupCallback() {
               @Override
               public void onSuccess() {
+                if (isFinishing() || isDestroyed()) {
```
johnxwork, line 172: You need to add a comment here explaining that these verbose code are to differentiate different fail reasons by looking at JNI line number on Prime. Otherwise it's quite strange why not just throw the string error directly

zhongqiliang:
ha? What are we doing here? What do you want to do after it? How would this help with solve the crash?

ckim1346:
> ha? What are we doing here? What do you want to do after it? How would this help with solve the crash?

I've narrowed down the crash to be happening in startBrowserProcessesAsync(). After talking with John, this PR is meant to help diagnose what exactly is causing the crash since I can't repro it locally and the crash only provides the line the crash originated from, not the message from the crash. PTAL at the bug for more details: [b/439066169](https://buganizer.corp.google.com/issues/439066169)

PR #7039, opened 2025-08-29
Restore focus on window on resume

haozheng-cobalt:
Request focus on the root view. This is a simple and reliable way to restore navigation control. The system will then find the first focusable view within this root view to give it focus.

Test: Have problem testing deeplink on Cobalt, I think the deeplink is configured to only Kimono. Also the bug is not reproducible anymore maybe due to my Shield OS upgrade. But I believe this is a solid fix, because I logged to see OnResume is called when app is resumed, and calling the rootView.requestFocus(); function works fine.
I will follow up with QA to verify this fix.

Bug: 431281341

### DIFF
--- a/cobalt/browser/user_agent/user_agent_platform_info.cc
+++ b/cobalt/browser/user_agent/user_agent_platform_info.cc
@@ -102,10 +102,6 @@ void GetUserAgentInputMap(
 
 namespace {
 
-static bool isAsciiAlphaDigit(int c) {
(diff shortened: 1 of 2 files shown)


johnxwork (approved):
Make sure to test screen reader before merging.

haozheng-cobalt:
> Make sure to test screen reader before merging.

Sure, and verified that TTS works fine.

PR #7041, opened 2025-09-01
DEPS: Restrict RDK checkout to Linux

rakuco:
This git repository has two branches with the same name that only differ in casing ("Hackathon2025/DACAppChamp" and "hackathon2025/DACAppChamp"), which breaks the `git-cache` tool on file systems like Apple's APFS.

For now, just make sure the repository is checked out only on Linux, which is the expected use case anyway.

Bug: 442335461


rakuco:

yell0wd0g (approved):
RS LGTM!

PR #7042, opened 2025-09-01
Only depend on oom_intervention on Android TV builds

rakuco:
This code was added in #6948 and its usage in shared files is restricted to `BUILDFLAG(IS_ANDROIDTV)` checks, so there is no need to unconditionally depend on this target -- in fact, doing so breaks the tvOS build:

```
../../cobalt/android/oom_intervention/oom_intervention_config.cc:34:19: error: no member named 'swap_total' in 'base::SystemMemoryInfoKB'
  if (memory_info.swap_total == 0) {
      ~~~~~~~~~~~ ^
../../cobalt/android/oom_intervention/oom_intervention_config.cc:44:41: error: no member named 'swap_total' in 'base::SystemMemoryInfoKB'
      static_cast<uint64_t>(memory_info.swap_total * swap_memory_ratio);
                            ~~~~~~~~~~~ ^
```

Bug: 440200929

### DIFF
--- a/cobalt/browser/BUILD.gn
+++ b/cobalt/browser/BUILD.gn
@@ -47,7 +47,6 @@ source_set("browser") {
     ":embed_polyfilled_javascript",
     ":global_features",
     ":metrics",
-    "//cobalt/android/oom_intervention",
     "//cobalt/browser:switches",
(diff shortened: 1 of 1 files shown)


rakuco:

PR #6952, opened 2025-08-22
refactor: Handle pthread_t and pthread_create() in the portable way

kjyoun:
This commit refactors the way `pthread_create` results are checked to improve portability and adhere to POSIX standards.

The `pthread_t` type is opaque, and its value should not be directly compared or read by application code. The previous implementation incorrectly checked `pthread_t` variables against 0 to determine if a thread had been successfully created. This is not a portable or reliable method.

This change replaces the direct checks with `std::optional<pthread_t>`. This approach correctly encapsulates the optional existence of a thread without inspecting the underlying `pthread_t` value, making the implementation safer and more portable across different systems.

Bug: 438955200

PR #7043, opened 2025-09-01
player_worker: Stop constructing |thread_| with an integer

rakuco:
Follow-up to #6952, which did the same to some other files but missed this occurrence.

One of the ideas behind using `std::optional` was to stop using 0 as a sentinel value; in fact, doing so here breaks the tvOS build:

```
../../starboard/shared/starboard/player/player_worker.cc:113:7: error: no matching constructor for initialization of 'std::optional<pthread_t>' (aka 'optional<_opaque_pthread_t *>')
```

Bug: 438955200


rakuco:

yell0wd0g (approved):
RS LGTM since this seems like a low risk change.

PR #7044, opened 2025-09-01
starboard: Remove unwanted headers from application_android.h

abhijeetk:
This cleanup helps keep the codebase leaner and avoids unnecessary
dependencies, which makes future maintenance, rebases, and cherry-picks
easier.

- Removed starboard-related headers that were already included by
  queue_application.h.
- Removed unused standard C++ headers (memory, atomic, vector) as they
  are not used in application_android.h

Locally verified by building for android-arm64

Bug: 443006742


abhijeetk:
cc :  @yell0wd0g

abhijeetk:

yell0wd0g:
Quick note: We ask CL authors to prefix titles with an indication
of the areas touched, in this case "starboard:" would be great.
Also in the CL description, if you could describe why this is 
done, e.g. "Include files are unused. Verified via CI", that'd be
great down the line (e.g. rebases, cherry picks etc).

abhijeetk:
> Quick note: We ask CL authors to prefix titles with an indication of the areas touched, in this case "starboard:" would be great. Also in the CL description, if you could describe why this is done, e.g. "Include files are unused. Verified via CI", that'd be great down the line (e.g. rebases, cherry picks etc).

Thanks! I’ve updated the change as per the review comments.
Not sure if I have the permissions to trigger the bots. I think some of them might need a maintainer’s approval. 🙂
Please let me know if anything else is needed.

PR #7045, opened 2025-09-02
Disable HTMLMediaElement MediaControl

johnxwork:
This feature was designed for mobile client and Cobalt should not need it. In addition, the attempt to show the MediaControl overlay is causing the number 1 crash in canary.

Bug: 427982512


third_party/blink/renderer/modules/media_controls/media_controls_impl.cc
```diff
@@ -176,6 +176,9 @@ bool ShouldShowPictureInPictureButton(HTMLMediaElement& media_element) {
 }
 
 bool ShouldShowCastButton(HTMLMediaElement& media_element) {
+#if BUILDFLAG(IS_COBALT)
+  return false;
```
yell0wd0g, line 181: Would it be possible to do e.g.
```
  HTMLMediaElementRemotePlayback::SetBooleanAttribute(
      *element, html_names::kDisableremoteplaybackAttr, true);
```
in a central place, e.g. `MediaControlsImpl::Create()`? That
way JS could see the `disableRemotePlayback` attribute set.

yell0wd0g (approved):
Great analysis, one suggestion.

Leaving an RS LGTM to speed up the process.
Please leave a `// TODO(b/427982512): Remove this`
or some such next to each `#if BUILDFLAG(...)` ?
Thx.

third_party/blink/renderer/modules/media_controls/media_controls_impl.cc
johnxwork: It's quite hard to get through all the checks rn. So I'm trying not to push another update. I'll try to find the perfect way to disable it as a follow-up.

PR #7051, opened 2025-09-02
Remove __external_threading.

y4vor:
This file is used only when a custom threading API is used e.g. SbThread. However we no longer do that and we use pthread APIs.

Issue: 442645934


cobalt/platform/includes/__external_threading
niranjanyardi, line 1: can we remove the entire folder tree starting with platform/ ?
AFAICT cobalt/platform/includes/__external_threading is the only file in the platform/ directory

niranjanyardi (approved):
apart from open comment , lgtm

cobalt/platform/includes/__external_threading
y4vor: Yes, this change would do it.

andrewsavage1 (approved):

PR #7052, opened 2025-09-03
starboard: Use different pthread_setname_np() on Apple platforms

rakuco:
`pthread_setname_np()` is not a portable function (that's what the "np" part stands for), and its signature differs between glibc/musl and Apple's libc (the latter takes 1 argument, the former takes 2).

Chromium abstracts this via base's PlatformThread class, but Starboard makes direct `pthread_setname_np()` calls that always assume the glibc/musl signature is being used.

Since in bug 432507888 it was mentioned that it is preferred not to ressurect `SbThreadSetName()`, add some `ifdef`s in code used by both Linux-based platforms and tvOS so that the right function signature is used depending on the build target.

Note that there are other occurrences in `//starboard`, but they are Linux-specific and have not been updated.

Bug: 432507888

### DIFF
--- a/starboard/nplb/posix_compliance/posix_thread_helpers.h
+++ b/starboard/nplb/posix_compliance/posix_thread_helpers.h
@@ -183,7 +183,11 @@ class AbstractTestThread {
 
  private:
   static void* ThreadEntryPoint(void* ptr) {
+#if defined(__APPLE__)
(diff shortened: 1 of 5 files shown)


rakuco:
(cc @Gyuyoung @jkim-julie @abhijeetk)

y4vor (approved):

PR #7053, opened 2025-09-03
common: Remove codec libraries pulled from //media

yell0wd0g:
Top level //media transitively depend on many media decoding/encoding
libraries, which are either not needed in Cobalt, or pulled from 
//starboard dependencies. This CL removes all/many of those dependencies.

**Instruction for rebasers** 
Part of this PR landed as an upstream CL, concretely around 143.0.7460.0.
It could not be cleanly backported, so once Cobalt passes said Chrome
version, the modifications in this PR for the files:
- `//media/base/supported_types_unittest.cc`
- `//media/cdm/library_cdm/clear_key_cdm/cdm_video_decoder.cc`
can simply be dropped.

Bug: 421171908 
Bug: 437151722


yell0wd0g:
out/linux-x64x11_devel/ before/after: 1351277600B vs 1344647880B (~6.3MB or ~0.5% smaller). Other builds coming soon.

yell0wd0g:
out/linux-x64x11_gold before/after 793321872 vs 787352608 (~6MB, or ~0.75%)

yell0wd0g:
out/android-arm_qa/apks/Cobalt.apk before/after: 79892780B vs 79794516B (98264B, or 0.1%)

yell0wd0g:
@niranjanyardi -- expanded version of your #7367 , PTAL

@youtube/cobalt-media PTAL in general. These GN flags should only affect //media,
//starboard has other dependency (flags).

yell0wd0g:
Somehow @youtube/cobalt-media didn't get a reviewer auto-assigned.
@xiaomings PTAL since we talked about this briefly.

media/base/supported_types_unittest.cc
```diff
@@ -40,8 +40,10 @@ TEST(SupportedTypesTest, IsSupportedVideoTypeBasics) {
   // Expect support for baseline configuration of known codecs.
   EXPECT_TRUE(IsSupportedVideoType(
       {VideoCodec::kVP8, VP8PROFILE_ANY, kUnspecifiedLevel, kColorSpace}));
+#if defined(ENABLE_LIBVPX)
```
borongc, line 43: Should the gates of `ENABLE_LIBVPX` and `ENABLE_LIBVPX` be upstreamed?

yell0wd0g: Yes - I'm not 100% confident that ENABLE_LIBVPX is the right 
guard here, becuase libvpx supports e.g. VP8. I expect the
discussion upstream will take a minute. I could hold off this
PR but we might need the size reduction sooner rather than
later. Thoughts?

yell0wd0g: Trying to upstream here: https://chromium-review.googlesource.com/c/chromium/src/+/7018429

media/cdm/library_cdm/clear_key_cdm/cdm_video_decoder.cc
```diff
@@ -26,6 +26,8 @@
 #include "media/base/decoder_status.h"
 #include "media/base/media_switches.h"
 #include "media/base/media_util.h"
+#include "media/base/video_decoder.h"
```
niranjanyardi, line 29: is this change needed ? (dont see any other changes to this file) . Im asking as this will need to be upstreamed

yell0wd0g: Yes, it's in https://crrev.com/c/7018429 as well. It's needed
because it's used in this file but only included transitively
via the bla_decoder.h below. When none of them is included,
VideoDecoder is not known. So this is just an IWYU case.

yell0wd0g:
https://crrev.com/c/7018429 is in flight upstream currently.

The patch itself won't apply cleanly here since there's
stuff in media/base/supported_types_unittest.cc that isn't
present here (encoding-related stuff). Let me make this
evident in the CL for future rebasers. Otherwise this PR
is ready! :)

cobalt/build/configs/common.gn
```diff
@@ -48,5 +48,16 @@ v8_enable_atomic_object_field_writes = false
 # This will force V8 to embed snapshot blob in binary.
 v8_use_external_startup_data = false
 
+# Many of these come from media_options.gni.
+enable_dav1d_decoder = false
+enable_hevc_parser_and_hw_decoder = false
+enable_libaom = false
+enable_platform_ac3_eac3_audio = false
```
borongc, line 55: We need to keep this enabled.

PR #7061, opened 2025-09-03
[tvOS] Remove deprecated pkg_resources

fayev-g:
Bug: 442908995

`gclient runhooks` fails with `ModuleNotFoundError: No module named 'pkg_resources'`. Deprecated in most recent versions of python. Replace with working `packaging` in the mean time to unblock building with existing `mac_toolchain.py` before rebase.

### DIFF
--- a/build/mac_toolchain.py
+++ b/build/mac_toolchain.py
@@ -20,7 +20,7 @@ the full revision, e.g. 9A235.
 
 import argparse
 import os
-import pkg_resources
+from packaging import version
 import platform
 import plistlib
 import shutil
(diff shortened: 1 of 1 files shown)


yell0wd0g (approved):
And `packaging` would work in all Python versions, right?
So nobody's workflows would be broken with this change,
and Python 3.12 would be neatly supported. If all that is
true, LGTM

fayev-g:
> And `packaging` would work in all Python versions, right? So nobody's workflows would be broken with this change, and Python 3.12 would be neatly supported. If all that is true, LGTM

"All python versions" meaning the python versions Cobalt supports, yes. (`packaging` is supported in 3.8+). I prefer this change to locking in 3.11 to get pkg_resources working because the error message doesn't make it clear that it's deprecated and you can use 3.11 in venv for a workaround. You still need to pip install `packaging`, but IMO it's a cleaner solution and, yes, more up-to-date.

rakuco:
FWIW, the upstream fix for this might work as well: https://chromium-review.googlesource.com/c/chromium/src/+/4936301

PR #7065, opened 2025-09-04
[tvOS] Add tvos-arm64 to gn.py

fayev-g:
Add _COBALT_TVOS_PLATFORMS with tvos-arm64 to gn.py. Will continue to add more tvOS platforms to gn.py.

Bug: 442923940

### DIFF
--- a/cobalt/build/gn.py
+++ b/cobalt/build/gn.py
@@ -71,11 +71,9 @@ _COBALT_ANDROID_PLATFORMS = [
 ]
 _COBALT_TVOS_PLATFORMS = [
     'tvos-arm64-simulator',
-]
-
-_COBALT_TVOS_PLATFORMS = [
     'tvos-arm64',
 ]
+
 
 def write_build_args(build_args_path, platform_args_path, build_type, use_rbe):
   """ Write args file, modifying settings for config"""


cobalt/build/gn.py
```diff
@@ -68,6 +68,9 @@
     'android-x86',
 ]
 
+_COBALT_TVOS_PLATFORMS = [
```
kaidokert: Would you mind adding `chromium_mac` up in the `CHROMIUM_PLATFORMS` array as well right away, with the matching `args.gn` ?  It doesn't need to be fully tested, lets just have a placeholder.

fayev-g: Done. Added mac-arm64 to mirror https://source.chromium.org/chromium/chromium/src/+/main:build/modules/mac-arm64/BUILD.gn

yell0wd0g (approved):
I thought I LGTMd this CL or a very similar one 🤔 
Still LGTM ofc

kaidokert:
> The `build/mac_toolchain.py` part should probably be dropped from this PR and landed separately as #7061.
> 

Good point, yes.

fayev-g:
> The `build/mac_toolchain.py` part should probably be dropped from this PR and landed separately as #7061.
> 
> In this PR, I'd add the scaffolding for Mac builds because they are simpler and could help @kaidokert get started with the CI side of things in the next few days.
> 
> For tvOS, the situation is a bit trickier and I think it'd be good to land it in another PR. AFAIU we don't want to add the build system integration the C25 way (i.e. bypass all `//build` integration and set all compiler and linker flags in GN files under `//starboard`), but on the other hand we don't have all the required bits and pieces in `//build` in M114 that allow a tvOS build to be done. One thing that @yell0wd0g suggested was to land the tvOS infrastructure we have, but just make it an iOS build for now while we wait for the M138 rebase. I can share what we've done tomorrow.

Thanks, Kubo! Yes, that all makes sense - I was approaching this change in the spirit of "land the tvOS infra...while we wait for M138 rebase" to unblock CI set-up. I'm happy to follow your lead on what should go in and make sure temporary solutions in main aren't unwieldy.

fayev-g:
> > The `build/mac_toolchain.py` part should probably be dropped from this PR and landed separately as #7061.
> 
> Good point, yes.

Apologies that I didn't explicitly say I attempted a diffbase strategy here. #7061 is merged now.

cobalt/build/configs/tvos-arm64/args.gn
```diff
@@ -0,0 +1,5 @@
+import("//cobalt/build/configs/common.gn")
+
```
yell0wd0g, line 2: I think since we're creating this file it'd be legitimate to
add 
```
is_apple = true
is_ios = true
```
since those are commonly used in Chromium (all versions) and
will propagate further to 138 (`tvos` is a `target_platform`
in addition to those two).

Actually `is_apple` is calculated [1] so this file doesn't
need to add it explicitly (and it might even collide). So
just the `is_ios` would be needed here and now.

[1] https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:build/config/BUILDCONFIG.gn;l=344;drc=cb1807e3ee9adce595454cb1567d7a907b05ebef

rakuco: `is_ios` is also calculated by the build system, the right flag to set here is `target_os = "ios"`.

fayev-g: Done. Also added a comment that these args are temporary until the rebase.

fayev-g:
> I still think this should not land as-is because the right GN flags are not being set for tvOS and this part of the change would be overwritten by our work when it gets upstreamed.
> 
> I've uploaded a branch with the minimum amount of backports required to have tvOS support in the build system so that we can build a few files under the `starboard` target (plus #7081 and #7082): [main...rakuco:cobalt:tvos-scaffolding](https://github.com/youtube/cobalt/compare/main...rakuco:cobalt:tvos-scaffolding) that could serve as another way to add the scaffolding required for tvOS.

Yes, the intention here is to just have a placeholder for tvos-arm64 and introduce the platform in gn.py. The args.gn are meant to be updated and over-written eventually. Your backport chain looks great - it looks independent from this PR though since it works with a simulator target.

fayev-g:
jfyi I moved the `chromium_mac-arm64` commit to https://github.com/youtube/cobalt/pull/7094. Consolidated the changes in that PR so that the intentionally temporary changes are here.

rakuco:
> > I still think this should not land as-is because the right GN flags are not being set for tvOS and this part of the change would be overwritten by our work when it gets upstreamed.
> > I've uploaded a branch with the minimum amount of backports required to have tvOS support in the build system so that we can build a few files under the `starboard` target (plus #7081 and #7082): [main...rakuco:cobalt:tvos-scaffolding](https://github.com/youtube/cobalt/compare/main...rakuco:cobalt:tvos-scaffolding) that could serve as another way to add the scaffolding required for tvOS.
> 
> Yes, the intention here is to just have a placeholder for tvos-arm64 and introduce the platform in gn.py. The args.gn are meant to be updated and over-written eventually. Your backport chain looks great - it looks independent from this PR though since it works with a simulator target.

I've just submitted #7107 with the backport chain + the tvos-arm64-simulator configuration. It's also possible to separate them although technically speaking the build config change doesn't build without the backports (i.e. it depends on how relevant it is for each commit to never break the build).

I think adding a tvOS build configuration via #7107 helps more than doing it via this PR: this PR alone allows `gn gen` to work (although I expect it to fail because of how the targets in `cobalt/BUILD.gn` are set up), but a build using those arguments would not work and the arguments being set here would be set differently once #7107 lands. A tvos-arm64 (or tvos-arm64-device) configuration like the one added here would actually be helpful, but IMHO it'd make more sense to land it after the other PR so that it's modeled after the tvos-arm64-simulator one.

fayev-g:
> > > I still think this should not land as-is because the right GN flags are not being set for tvOS and this part of the change would be overwritten by our work when it gets upstreamed.
> > > I've uploaded a branch with the minimum amount of backports required to have tvOS support in the build system so that we can build a few files under the `starboard` target (plus #7081 and #7082): [main...rakuco:cobalt:tvos-scaffolding](https://github.com/youtube/cobalt/compare/main...rakuco:cobalt:tvos-scaffolding) that could serve as another way to add the scaffolding required for tvOS.
> > 
> > 
> > Yes, the intention here is to just have a placeholder for tvos-arm64 and introduce the platform in gn.py. The args.gn are meant to be updated and over-written eventually. Your backport chain looks great - it looks independent from this PR though since it works with a simulator target.
> 
> I've just submitted #7107 with the backport chain + the tvos-arm64-simulator configuration. It's also possible to separate them although technically speaking the build config change doesn't build without the backports (i.e. it depends on how relevant it is for each commit to never break the build).
> 
> I think adding a tvOS build configuration via #7107 helps more than doing it via this PR: this PR alone allows `gn gen` to work (although I expect it to fail because of how the targets in `cobalt/BUILD.gn` are set up), but a build using those arguments would not work and the arguments being set here would be set differently once #7107 lands. A tvos-arm64 (or tvos-arm64-device) configuration like the one added here would actually be helpful, but IMHO it'd make more sense to land it after the other PR so that it's modeled after the tvos-arm64-simulator one.

Thanks for sharing a draft PR of a backporting-scaffolding solution! I agree that it includes more comprehensive changes and we can all take a look if it would help to merge that before the rebase. (Also nit: I equate "submit" with "merge" and so I was initially excited that PR was already submitted. It looks like it was just put up.)

I initially put up this PR to make it at least clear in code that tvos-* platform should be integrating `gn.py`, not using `gn gen...` directly as in documentation I've seen for previous versions of Cobalt. I was actually unsure about this at first and so Kaido suggested the quick addition in this PR. So it is less about functionality and more just about clarity of usage. (I probably should've made this clearer in the PR description.) With that said, I might be following a different code review culture  - I think this PR is very small and still independent of #7107. Generally I don't see such changes blocked, even if there is still work to be done (e.g. aligning tvos-arm64 with tvos-arm64-simulator if #7107 lands). But I'm happy to work with other strong preferences.

y4vor (approved):

cobalt/build/configs/tvos-arm64/args.gn
```diff
@@ -0,0 +1,8 @@
+import("//cobalt/build/configs/common.gn")
```
rakuco, line 1: Since #7107 has landed, could you update this file to follow what tvos-arm64-simulator does? Namely:
- import `tvos_common.gn` instead
- set `target_environment` instead of `target_os`
- the `use_system_xcode` bit is redundant, but if you'd like to keep it then it should go to `tvos_common.gn`

rakuco: Done in #7424

PR #7068, opened 2025-09-04
cobalt: Refactor User Agent creation logic

yell0wd0g:
A few refactors; this is just for backup. 

- Step 1 - remove unnecessary |enable_skia_rasterizer| ctor param. Chrobalt always uses Skia rasterizer
- Step 2 - Move init function to being a member method
- Step 3 - Make setter functions private. This in turn needs moving the users of those setters in the test file to become ForTesting factory methods (https://chromium.googlesource.com/chromium/src/+/main/docs/patterns/fortesting-methods.md).

TODO: Bundle platform-specific bits of UserAgentPI::Initialize()
into a method.

Bug: 442760345


yell0wd0g:
The main points have landed elsewhere.

PR #7069, opened 2025-09-04
cobalt: Clean up UserAgent logic

yell0wd0g:
This CL carries out two cleanups on Cobalt's UA generation logic:
- Removes the unnecessary `enable_skia_rasterizer` ctor param,
associated member variable and getter. ToT doesn't use it and
anyway we should always use Skia rasterization.
- Removes the superfluous `isAsciiAlphaDigit()` [1] since there's 
a `base::IsAsciiAlphaNumeric()` that does precisely the same [2]
(minus the surprising `int` as parameter - those strange 
char-as-int manipulations will be cleaned in a follow up CL).

No new functionality intended, CI passing would be enough.

[1] https://github.com/youtube/cobalt/blob/78d3f5d152f6fb7a8f5c016d345f27ea92e641c8/cobalt/browser/user_agent/user_agent_platform_info.cc#L105-L107
[2] https://github.com/youtube/cobalt/blob/78d3f5d152f6fb7a8f5c016d345f27ea92e641c8/base/strings/string_util.h#L415-L417

Bug: 442760345

### DIFF
--- a/starboard/android/shared/application_android.h
+++ b/starboard/android/shared/application_android.h
@@ -15,19 +15,13 @@
 #ifndef STARBOARD_ANDROID_SHARED_APPLICATION_ANDROID_H_
 #define STARBOARD_ANDROID_SHARED_APPLICATION_ANDROID_H_
 
-#include <atomic>
-#include <memory>
 #include <string>
-#include <vector>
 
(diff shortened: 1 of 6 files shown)


yell0wd0g:
@fayev-g PTAL

PR #7072, opened 2025-09-04
starboard: Add FormatWithDigitSeparators method for logging numbers

kjyoun:
The `FormatWithDigitSeparators` function makes numbers in logs more readable. It's especially useful for large values like timestamps.

In this example, with formatted number, we can easily tell pts increase by ~40 msec

```
BEFORE
Render > Schedule: index=15, pts=175625625 usec
Render > Schedule: index=16, pts=175667333 usec
Render > Schedule: index=17, pts=175709041 usec

AFTER
Render > Schedule: index=15, pts=175'625'625 usec
Render > Schedule: index=16, pts=175'667'333 usec
Render > Schedule: index=17, pts=175'709'041 usec
```

This PR also does some refactoring
- Move implementation code to .cc file
- Add unit tests for string.h

#vibe-coded

Bug: 441561644

### DIFF
--- a/starboard/common/string.h
+++ b/starboard/common/string.h
@@ -58,7 +58,7 @@ std::string HexEncode(const void* data,
 // (e.g., 1'234'567).
 // NOTE: This function is for logging and debugging purposes only. It should not
 // be used in any UI, as it does not handle internationalization (i18n).
-std::string FormatWithDigitSeparators(int64_t);
+std::string FormatWithDigitSeparators(int64_t number);
 
 template <typename CHAR>
 static SB_C_FORCE_INLINE int strlcpy(CHAR* dst, const CHAR* src, int dst_size) {


starboard/common/string_test.cc
```diff
@@ -0,0 +1,176 @@
...
+
+  int len = strlcpy(dest, "non-empty", 0);
+
+  EXPECT_EQ(9, len);
+}
+
+TEST(StrlcpyTest, HandlesDestinationSizeOfOne) {
+  char dest[kDestSize];
+
+  int len = strlcpy(dest, "test", kDestSize);
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/common/string.cc
```diff
@@ -0,0 +1,62 @@
...
+  std::string result;
+  int count = 0;
+  for (int i = str.length() - 1; i >= 0; i--) {
+    if (count > 0 && count % 3 == 0 && str[i] != '-') {
+      result.insert(0, ",");
+    }
+    result.insert(0, 1, str[i]);
+    count++;
+  }
+  return result;
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/common/string.cc
```diff
@@ -0,0 +1,69 @@
...
+
+  std::string result;
+  auto delimiter_size = delimiter ? std::strlen(delimiter) : 0;
+  result.reserve((delimiter_size + 2) * size);
+
+  const uint8_t* data_in_uint8 = static_cast<const uint8_t*>(data);
+
+  for (int i = 0; i < size; ++i) {
+    result += kDecToHex[data_in_uint8[i] / 16];
+    result += kDecToHex[data_in_uint8[i] % 16];
```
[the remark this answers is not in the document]
kjyoun: It's on existing code

kaidokert (approved):
Seems like a good tidy improvement - code doens't seem to have any issues. Lets have @jellefoks express an opinion as well.

starboard/common/string.cc
```diff
@@ -0,0 +1,64 @@
...
+std::string FormatNumber(int64_t number) {
+  if (number == 0) {
+    return "0";
+  }
+
+  std::string str = std::to_string(number);
+  std::string reversed_result;
+  reversed_result.reserve(str.length() + (str.length() - 1) / 3);
+  int digit_count = 0;
+  for (int i = str.length() - 1; i >= 0; --i) {
```
jellefoks, line 48: I love this idea. Logs should be more readable.

With the locale set to a language where there is a thousands-separator, there actually is a standard way to print them:

`setlocale(LC_NUMERIC, ""); printf("%'d\n", 1123456789);`

Wouldn't it be better to use such existing functionality?

For a C++17 method that also does not rely on the locale setting, I found an approach in this snippet (did not try it), which may also work without having to implement our own:

```
#include <iostream>
#include <locale>
#include <string>
#include <memory>

struct custom_thousands_separator : std::numpunct<char> {
    char do_thousands_sep() const override { return ','; }
    std::string do_grouping() const override { return "\3"; }
};

long long large_number = 1234567890LL;
auto custom_locale = std::locale(std::cout.getloc(), new custom_thousands_separator());
std::cout.imbue(custom_locale);
std::cout << "Custom locale with comma: " << large_number << '\n';
```

kjyoun: thanks for the suggestion, though I prefer sticking the exising impl.

I was glad that I could implement the same feature using existing c++ methods 
but the big red flag is that  setlocale() changes the global state.
[1] is the implementation using setlocale.

I checked how other implementations for the same feature.
They uses its own stand-alone methods, not using setlocale.

---
```
std::string FormatNumber(int64_t number) {
  std::lock_guard<std::mutex> lock(g_locale_mutex);

  // Save the current locale so we can restore it.
  char* old_locale = setlocale(LC_NUMERIC, nullptr);
  std::string original_locale_str;
  if (old_locale) {
    original_locale_str = old_locale;
  }

  // Set the system's default locale for numeric formatting.
  setlocale(LC_NUMERIC, "");

  // The ' format specifier is a non-standard POSIX extension.
  // We assume a buffer large enough for a 64-bit integer with commas.
  char buffer[32];
  snprintf(buffer, sizeof(buffer), "%'lld",
           static_cast<long long int>(number));

  // Restore the original locale to avoid side effects.
  if (old_locale) {
    setlocale(LC_NUMERIC, original_locale_str.c_str());
  }

  return buffer;
```

jellefoks: A counter-argument there is that you don't need to call setlocale to have the `%'lld` (or another standard library method) print more readable numbers, because by definition, it will print the numbers readable in what is considered readable in the locale of the user. That is exactly what locale specifies, the readable format depending on the language and region of the user.

E.g. in many locales, the comma is the decimal separator not the thousands separator, so printing one or more commas to separate thousands is actually confusing. In another locale, a readable number has only two digits between separators after the thousands (using lakh and crore). What I mean is that maybe readable for us as `1,000,000`, for others would have to be `1.000.000` and for others again it should be `10.00.000`. And that is exactly what the locale  is for, to make it so you, as the programmer, don't have to care which one it is, you just specify that you want the readable format for the language and region of the user.

Now, if your intention here not to make the numbers as readable as possible for anyone, and wish to ignore the locale and always use the three-digit separator, then it would make sense to not use existing functionality.

But in that case, I would request to not use the comma as the separator. I propose using the single quote, the same as the separator that was chosen for the number literal separator C++14 (*1), where the separator is allowed at any position in the number. The single quote was chosen because (almost no (*2)) locale has expectations for the meaning of the single quote as a separator, minimizing confusion for everyone.

Additionally, by choosing the single quote as the separator instead of a comma or decimal point, it will ensure that any code or script that reads these formatted numbers to convert them back to a number will not interpret the value differently if the locale when reading the value is different. e.g. formatting a number this way and then converting it back with a library function that adheres to the locale will interpret "1,000,000" as value `1` in European languages, causing all sorts of fun bugs.

*1: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3781.pdf
*2: Switzerland and Italy are only locales where the apostrophe is sometimes used as a digit separator, but luckily for this case, it's always in three digit pairs.

kjyoun: update the code to use single-quote as thousand-separator

starboard/common/string.cc
```diff
@@ -0,0 +1,64 @@
...
+      reversed_result.push_back('\'');
+    }
+    reversed_result.push_back(str[i]);
+    digit_count++;
+  }
+  if (number < 0) {
+    reversed_result.push_back('-');
+  }
+  return {reversed_result.rbegin(), reversed_result.rend()};
+}
```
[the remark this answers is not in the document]
kjyoun: it is chosen due to avoid performance overhead.

starboard/common/string.cc
kjyoun: renamed method to `FormatWithDigitSeparators` to clearly show what this method does.

Previous method name of `FormatNumber` sounds too generic and can be even misleading, since it's not format number for the locale.

PR #7073, opened 2025-09-04
test: Add (disabled) tests and stubs for setlocale and related functions.

jellefoks:
Tests only, with the data tested against generated from non-hermetic locale.

The tests are currently disabled on most configurations, because for non-hermetic linux configurations the machines do not have full locale installed, and for hermetic configurations the stubs in this PR are used.

Bug: 403007005

### DIFF
--- a/cobalt/common/libc/locale/locale.cc
+++ b/cobalt/common/libc/locale/locale.cc
@@ -92,7 +92,7 @@ locale_t newlocale(int category_mask, const char* locale, locale_t base) {
 }
 
 locale_t uselocale(locale_t newloc) {
-  // TODO: This is a stub.
+  // TODO: b/403007005 Fill in this stub.
   return nullptr;
 }
 


cobalt/common/libc/locale/locale.cc
```diff
@@ -0,0 +1,250 @@
...
+  }
+  if (loc == (locale_t)0) {
+    return (locale_t)0;
+  }
+  lconv* new_lconv = new lconv;
+  memcpy(new_lconv, reinterpret_cast<lconv*>(loc), sizeof(lconv));
+  return reinterpret_cast<locale_t>(new_lconv);
+}
+
+char* nl_langinfo_l(nl_item item, locale_t locale) {
```
johnedocampo: nit: Can we add a TODO here to properly implement nl_langinfo_l?

johnedocampo (approved):
LGTM from my end, just need to update the title of the PR.

johnedocampo:
> Setlocale #7073

nit: update the title to follow the styling guide: go/cobalt-style#commit-message

cobalt/common/libc/locale/locale.cc
```diff
@@ -0,0 +1,250 @@
...
+#include <cstddef>
+
+namespace {
+thread_local locale_t g_current_locale = reinterpret_cast<locale_t>(0);
+
+// The default locale is the C locale.
+const lconv* GetCLocaleConv() {
+  static const lconv c_locale_conv = {
+      .decimal_point = const_cast<char*>("."),
+      .thousands_sep = const_cast<char*>(""),
```
y4vor, line 31: Are those values supposed to be empty?

cobalt/common/libc/locale/locale.cc
```diff
@@ -0,0 +1,250 @@
...
+    memcpy(new_lconv, reinterpret_cast<lconv*>(base), sizeof(lconv));
+    freelocale(base);
+  } else {
+    memcpy(new_lconv, GetCLocale(), sizeof(lconv));
+  }
+  return reinterpret_cast<locale_t>(new_lconv);
+}
+
+locale_t uselocale(locale_t newloc) {
+  // TODO: This is a stub.
```
y4vor: Are we planning to implement this one?  If yes, please file a ticket. If no please explain in a comment and remove the TODO part.

cobalt/common/libc/locale/locale.cc
jellefoks: (edit): This is the C locale, which is the default locale, and yes, they are supposed to be empty for the C locale.

cobalt/common/libc/locale/locale.cc
jellefoks: It's in the b/403007005 referenced in the message. Added reference to the bug in the comment here.

PR #7079, opened 2025-09-05
 Create the instructions to build RDK starboard for Cobalt 25

jonastsai:
Initial document to build and run RDK starboard. It contains:
  - What is the RDK reference device and where to get one
  - Build starboard using Cobalt mechanism
  - Run and check the log

Bug: b/442480997


jonastsai:
One concern: the image, updating tools and instructions to use the AH212 devices are located in developers.google.com and the users needs to be a cobalt developer to access it.

cobalt/site/docs/development/setup-rdk.md
```diff
@@ -3,6 +3,92 @@ Book: /youtube/cobalt/_book.yaml
...
+sh 20250521_rdk-glibc-x86_64-arm-toolchain-2.0.sh -d ${RDK_SDK} -y
+
+# Step 2: Checkout the source code
+# Ensure you checkout the 25.lts.1+ branch and update the code to the latest version
+cd <cobalt_folder>
+git checkout 25.lts.1+
+
+# Step 3: Build RDK Starboard
+# Generate the output folder for RDK
+gn gen out/rdk-arm_devel '--args=target_platform="rdk-arm" build_type="devel" target_cpu="arm" target_os="linux" is_clang=false using_old_compiler=true rdk_build_with_yocto=false rdk_starboard_root="//starboard/contrib/rdk/src" rdk_enable_securityagent=false'
```
niranjanyardi: You can just run python gn.py -p rdk and -C devel - all these are default arguments in that.
I had double checked that this works 1-2 months ago

cobalt/site/docs/development/setup-rdk.md
```diff
@@ -3,6 +3,92 @@ Book: /youtube/cobalt/_book.yaml
...
+[rdklogic](https://rdklogic.tv/collections/all-products/products/amlogic-s905x4-developer-box-rdk?variant=43178224484592).
+
+2. Download the RDK image
+
+    The prebuilt RDK image for AH212 is available on the YouTube developer page. Please select the
+[RDK Certified Image(2025-02)](https://developers.google.com/youtube/devices/living-room/compliance/ah212#rdk_certified_image2025-02) image.
+
+3. Update images and getting logs
+
+    Visit the [YouTube developer page](https://developers.google.com/youtube/devices/living-room/compliance/ah212)
```
niranjanyardi: Does this mean all instructions are on the external website ?

So the word document you had is no longer needed ?

jonastsai: The gDoc contains more information regarding how to use the reference device. Ex: how to run cobalt as an executable, check key provision status, pairing RCU via commands etc. They will not included in the external documents. We can add more instructions here if something is mandatory and important. Otherwise, I will keep this document to focus on running the cobalt only.

cobalt/site/docs/development/setup-rdk.md
jonastsai: thanks. I update the instruction accordingly.

cobalt/site/docs/development/setup-rdk.md
jonastsai: I remove the document about download/update images as I have issue to provide the image to public. Let's point the partner to access the developer portal instead.

PR #7080, opened 2025-09-05
patch fix  Functional Tests EME Basic getMetrics fail 

weifang-cool:
b/389180717

Problem:
Functional Tests EME Basic getMetrics fail

Solution:
workround for widevine cdm metrics maybe return Null case

A timing issue between the createMediaCryptoSession and getMetricsInBase64 functions in Cobalt's MediaDrmBridge.
If getMetricsInBase64 call first the metrics return null, then we can ignore this case first.

Change-Id: I04d77cced24ce5bc036f975d0907eee66ca6c735


hlwarriner:
@borongc, can you please take a look from the media side?

PR #7081, opened 2025-09-05
starboard: Adjust more pthread_setname_np() invocations

rakuco:
Follow-up to #7052: there were two other occurrences that were left out of the PR but which are still used by tvOS as well as Linux/Android.

Bug: 432507888

### DIFF
--- a/starboard/common/thread.cc
+++ b/starboard/common/thread.cc
@@ -79,7 +79,11 @@ std::atomic_bool* Thread::joined_bool() {
 
 void* Thread::ThreadEntryPoint(void* context) {
   Thread* this_ptr = static_cast<Thread*>(context);
+#if defined(__APPLE__)
+  pthread_setname_np(this_ptr->d_->name_.c_str());
+#else
   pthread_setname_np(pthread_self(), this_ptr->d_->name_.c_str());
+#endif
   this_ptr->Run();
   return NULL;
 }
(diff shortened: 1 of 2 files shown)


rakuco:

starboard/common/thread.cc
```diff
@@ -79,7 +79,11 @@ std::atomic_bool* Thread::joined_bool() {
 
 void* Thread::ThreadEntryPoint(void* context) {
   Thread* this_ptr = static_cast<Thread*>(context);
+#if defined(__APPLE__)
```
yell0wd0g, line 82: QQ - is it better to use this `defined(__APPLE__)` (which I think
is a compiler defined thing) or `BUILDFLAG(IS_APPLE)` (which will
be a Chromium GN-originated thing?

yell0wd0g (approved):
LGTM w/ a question.

starboard/common/thread.cc
rakuco: I'm using `__APPLE__` here because I thought `//starboard` wasn't allowed to depend on `//base` or `//build`, so `BUILDFLAG(IS_APPLE)` wasn't an option.

yell0wd0g: `BUILDFLAG` comes from `build/` but I see your point. `__APPLE__` SGTM.

PR #7082, opened 2025-09-05
[tvOS] Replace mac_toolchain pkg_resources fix with upstream one

rakuco:
This PR essentially replaces the fix from #7061 with the upstream one that does not require manual installation of extra Python packages (and also means one less conflict to deal with in the rebases).

Installing extra dependencies gets extra confusing when one also has Python
from e.g. Homebrew in addition to the system version: in this case it is
very likely that `gn.py` will be invoked with Homebrew's Python while `gn`
itself will invoke the system-wide `python3` binary, so even if one does
install the `packaging` module it is possible that it will still not be
picked up by the right Python interpreter.

Bug: 442908995

---

Original commit description:

mac_toolchain.py used pkg_resources.parse_version to parse and
ultimately compare version numbers. pkg_resources is deprecated per
https://setuptools.pypa.io/en/latest/pkg_resources.html and
pypa/setuptools a1aeda391a0c (2023-03-05), in setuptools 67.5.0.
pkg_resources.parse_version’s implementation has been
packaging.version.Version since pypa/setuptools db23abf7ef84
(2020-12-20), and when available, since pypa/setuptools 9382fa0c05e5
(2014-09-25). However, pkg_resources as installed does’t always depend
directly on the packaging.version module, it often depends on a
pkg_resources._vendor.packaging.version snapshot. This means that the
historical presence of pkg_resources doesn’t necessarily indicate that
packaging.version will be available by that name.

Rather than relying on an external module for version parsing, just
split the version string on dots. It appears that these version strings
only ever take the form 15.0 (for Xcode 15.0), and don’t ever contain
letters or other confounding characters.

Change-Id: Iea7260629030503f0f90ce33da8816e0da9faaca
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/4936301
Cr-Commit-Position: refs/heads/main@{#1209652}

### DIFF
--- a/build/mac_toolchain.py
+++ b/build/mac_toolchain.py
@@ -20,7 +20,6 @@ the full revision, e.g. 9A235.
 
 import argparse
 import os
-import pkg_resources
 import platform
 import plistlib
 import shutil
@@ -156,8 +155,8 @@ def InstallXcodeBinaries():
     current_license_plist = LoadPList(current_license_path)
     xcode_version = current_license_plist.get(
         'IDEXcodeVersionForAgreedToGMLicense')
-    if (xcode_version is not None and pkg_resources.parse_version(xcode_version)
-        >= pkg_resources.parse_version(cipd_xcode_version)):
+    if (xcode_version is not None
+        and xcode_version.split('.') >= cipd_xcode_version.split('.')):
       should_overwrite_license = False
 
   if not should_overwrite_license:


rakuco:
@fayev-g @yell0wd0g PTAL. I ended up doing this while working on a tvOS-scaffolding branch and noticed that multiple Python binaries were being used in the `gn.py` invocation and managing the installation of the `packaging` package was more work than just adopting the upstream version of the fix.

rakuco:

PR #7076, opened 2025-09-05
cobalt: Refactor platform-specific bits of UserAgentPlatformInfo

yell0wd0g:
This CL refactors `UserAgentPlatformInfo` by adding platform-specific 
`InitializePlatformDependentFieldsAndroid()` and
`InitializePlatformDependentFieldsEvergreen()` and moving the
most platform specifics bits there. This clarifies what's expected 
where  and simplifies extending to new architectures, e.g. tvOS.

This CL is rebased after #7104 - also the controversial parts of
folding //base/system/sys_info_starboard* into u_a_platform_info
is punted for the moment.

Tested via CI unit tests.

Bug: 442760345

PR #7085, opened 2025-09-05
Populate the missing fields in the UA string

y4vor:
Populate the missing fields for the 3P UA string
using the starboard system properties.

Remove is_robolectric from starboard/build/config/os_definitions.gni
as it is in conflict with build/toolchain/android/BUILD.gn.

Issue: 437414261

### DIFF
--- a/base/time/time.h
+++ b/base/time/time.h
@@ -1243,7 +1243,7 @@ class BASE_EXPORT ThreadTicks : public time_internal::TimeBase<ThreadTicks> {
   // Returns true if ThreadTicks::Now() is supported on this system.
   [[nodiscard]] static bool IsSupported() {
 #if BUILDFLAG(IS_STARBOARD)
-    return starboard::CurrentMonotonicThreadTime() != 0;
+    return ::starboard::CurrentMonotonicThreadTime() != 0;
 #elif (defined(_POSIX_THREAD_CPUTIME) && (_POSIX_THREAD_CPUTIME >= 0)) || \
     BUILDFLAG(IS_APPLE) || BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_FUCHSIA)
     return true;
--- a/cobalt/browser/user_agent/user_agent_platform_info.cc
+++ b/cobalt/browser/user_agent/user_agent_platform_info.cc
@@ -26,6 +26,7 @@
 #include "base/system/sys_info.h"
 #include "base/system/sys_info_starboard.h"
 #include "build/build_config.h"
+#include "starboard/common/system_property.h"
 #include "starboard/extension/platform_info.h"
 
 #include "cobalt/cobalt_build_id.h"  // Generated
@@ -205,15 +206,14 @@ void InitializeUserAgentPlatformInfoFields(UserAgentPlatformInfo& info) {
   }
 #endif  // ENABLE_DEBUG_COMMAND_LINE_SWITCHES
 
-#if BUILDFLAG(IS_ANDROID)
-  info.set_device_type("ATV");
-#else
-  info.set_device_type("TV");
-#endif  // BUILDFLAG(IS_ANDROID)
+  info.set_device_type(
+      starboard::GetSystemPropertyString(kSbSystemPropertyDeviceType));
 
-// TODO(cobalt, b/374213479): figure out firmware version for other platforms.
 #if BUILDFLAG(IS_ANDROID)
   info.set_firmware_version(base::SysInfo::GetAndroidBuildID());
+#elif BUILDFLAG(IS_STARBOARD)
+  info.set_firmware_version(
+      starboard::GetSystemPropertyString(kSbSystemPropertyFirmwareVersion));
 #endif  // BUILDFLAG(IS_ANDROID)
 
   info.set_model(base::SysInfo::HardwareModelName());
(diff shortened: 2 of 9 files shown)


haozheng-cobalt:
There is a collision with Miguel's recent PR, he's removing //base/system/sys_info_starboard to facilitate //base rebases. 

@yell0wd0g , https://github.com/youtube/cobalt/pull/7076

haozheng-cobalt (approved):
LGTM

jellefoks:
> Popluate the missing fileds for the 3P UA string using the starborad system properties.
> 
> Issue: 437414261

"starborad" is a tpyo :-P

cobalt/browser/user_agent/user_agent_platform_info.cc
```diff
@@ -211,9 +212,11 @@ void InitializeUserAgentPlatformInfoFields(UserAgentPlatformInfo& info) {
   info.set_device_type("TV");
```
jellefoks: In C25 this one also fetches a property. 

```
  result = SbSystemGetProperty(kSbSystemPropertyDeviceType, value,
                               kSystemPropertyMaxLength);
  SB_DCHECK(result);
  info.set_device_type(value);
```

jellefoks (changes requested):

cobalt/browser/user_agent/user_agent_platform_info.cc
y4vor: Done

cobalt/version.h
```diff
@@ -36,6 +36,7 @@
 //                  release is cut.
 //.
 
+#define COBALT_MAJOR_VERSION "26"
 #define COBALT_VERSION "26.master.0"
```
yell0wd0g, line 40: Could we make this e.g.
`#define COBALT_VERSION COBALT_MAJOR_VERSION ## ".master.0" `
or some such? Concatenating ISO repeating the 26.

base/time/time.h
```diff
@@ -1243,7 +1243,7 @@ class BASE_EXPORT ThreadTicks : public time_internal::TimeBase<ThreadTicks> {
   // Returns true if ThreadTicks::Now() is supported on this system.
   [[nodiscard]] static bool IsSupported() {
 #if BUILDFLAG(IS_STARBOARD)
-    return starboard::CurrentMonotonicThreadTime() != 0;
+    return ::starboard::CurrentMonotonicThreadTime() != 0;
```
yell0wd0g, line 1246: This seems unrelated to this change.

cobalt/browser/user_agent/user_agent_platform_info.cc
```diff
@@ -205,15 +206,14 @@ void InitializeUserAgentPlatformInfoFields(UserAgentPlatformInfo& info) {
...
 
-// TODO(cobalt, b/374213479): figure out firmware version for other platforms.
 #if BUILDFLAG(IS_ANDROID)
   info.set_firmware_version(base::SysInfo::GetAndroidBuildID());
+#elif BUILDFLAG(IS_STARBOARD)
+  info.set_firmware_version(
+      starboard::GetSystemPropertyString(kSbSystemPropertyFirmwareVersion));
 #endif  // BUILDFLAG(IS_ANDROID)
 
   info.set_model(base::SysInfo::HardwareModelName());
```
yell0wd0g, line 219: Instead of adding more code to base/sys_info_starboard.cc,
I'd split here
```
#if BUILDFLAG(IS_STARBOARD)
  info.set_model(GetSystemPropertyString(kSbSystemPropertyModelName));
#else.
  info.set_model(base::SysInfo::HardwareModelName());
#endif
```

This is because ideally we'd avoid modifying //base (even for 
net additions), specially for a use case like this, where 
//cobalt can call directly to //starboard. It'd also simplify
my rebasing of #7076 :)

BUILD.gn
```diff
@@ -408,7 +408,9 @@ group("gn_all") {
           "//chrome/browser/android/examples/partner_browser_customizations_provider:partner_browser_customizations_example_apk",
           "//content/shell/android:content_shell_test_apk",
         ]
-
+        if (is_cobalt) {
+          deps -= [ "//chrome/android:chrome_junit_tests" ]
```
yell0wd0g: There seems to be a number of unrelated changes in this
CL, in this file, //chrome/android/BUILD.gn, etc. The CL
description doesn't specify why or whether they were 
needed, are they necessary?

yell0wd0g (changes requested):
Looking good in substance, I'd rather not modify
//base anymore.

cobalt/version.h
y4vor: Initially I did, but we have a python build script which is parsing the current #define COBALT_VERSION "26.master.0".  

So for readability, both human and script parsing I left it as is.

BUILD.gn
y4vor: They are necessary. Starboard and android set the `is_robolectric` flag differently. 
//starboard/build/config/os_definitions.gni
//build/toolchain/android/BUILD.gn

Updated the PR description.

base/time/time.h
y4vor: Build error, so very related.

cobalt/browser/user_agent/user_agent_platform_info.cc
y4vor: The cleanest design here would be to add what we need to `base/system/sys_info.h` interface the way all other platforms do and implement sys_info_starboard.cc for 3P and modify sys_info_android.cc for Kimono. And the BUILDFLAG would be IS_COBALT as Kimono doesn't set IS_STARBOARD.

What you are proposing is to leave  `base::SysInfo::HardwareModelName` and possible other fields hollow for 3P and do a lot of #ifdefs at higher level in the  `user_agent/user_agent_platform_info.cc`.

We already have may changes in //base. The key is to make them fit well in the existing framework.

yell0wd0g: We can have that conversation down the road :-)

BUILD.gn
andrewsavage1: Let's revert this robolectric change too

y4vor: done

PR #7092, opened 2025-09-08
media/starboard: make SbPlayerSetBoundsHelper not ref-counted

yell0wd0g:
This CL/PR makes SbPlayerSetBoundsHelper not-ref counted: Its only 
callsite is StarboardRenderer::OnVideoGeometryChange() which IIUC 
is a Mojo callback; so it could be on a Task Runner different
from |task_runner_|: left a CHECK and if it trips, the real way to 
avoid a problem is via PostTask.

This CL also removes unnecessary expansion in SbPlayerSetBoundsHelper::SetBounds
parameter(s), using gfx::Rect directly.

Fixed: 448749860

### DIFF
--- a/starboard/build/config/base_configuration.gni
+++ b/starboard/build/config/base_configuration.gni
@@ -72,9 +72,7 @@ declare_args() {
   #   "system_gles2" - Use the system implementation of EGL + GLES2. The
   #                    headers and libraries must be on the system include and
   #                    link paths.
-  #   "glimp"  - Cobalt's own EGL + GLES2 implementation. This requires a
-  #              valid Glimp implementation for the platform.
-  #   "angle"  - A DirectX-to-OpenGL adaptation layer. This requires a valid
+  #   "angle"  - Use ANGLE's EGL + GLES2 implementation. This requires a valid
   #              ANGLE implementation for the platform.
   gl_type = "system_gles2"
 
--- a/starboard/egl_and_gles/BUILD.gn
+++ b/starboard/egl_and_gles/BUILD.gn
@@ -22,17 +22,26 @@
 # This decision is predicated on the value of the |gl_type| GN variable defined
 # in the configuration.gni for the current platform.
 
+import("//build/buildflag_header.gni")
+
 group("egl_and_gles") {
-  assert(gl_type != "none", "gl_type must be set to a valid value")
+  assert(gl_type == "angle" || gl_type == "system_gles2")
   public_deps = [ ":egl_and_gles_$gl_type" ]
+}
+
+buildflag_header("buildflags") {
+  header = "buildflags.h"
+
+  enable_angle = gl_type == "angle"
+  enable_system_gles2 = gl_type == "system_gles2"
+  flags = [
+    "STARBOARD_GL_TYPE_ANGLE=$enable_angle",
+    "STARBOARD_GL_TYPE_SYSTEM_GLES2=$enable_system_gles2",
+  ]
 }
 
 group("egl_and_gles_system_gles2") {
   # Use the system-provided implementation of GLES2.
-}
-
-config("egl_and_gles_glimp_public_config") {
-  defines = [ "GL_GLEXT_PROTOTYPES" ]
 }
 
 declare_args() {
@@ -43,12 +52,10 @@ config("egl_and_gles_angle_config") {
   include_dirs = [ "//third_party/angle/include" ]
 }
 
-if (gl_type == "angle") {
-  group("egl_and_gles_angle") {
-    public_deps = [
-      "//third_party/angle:libEGL_static",
-      "//third_party/angle:libGLESv2_static",
-    ]
-    public_configs = [ ":egl_and_gles_angle_config" ]
-  }
+group("egl_and_gles_angle") {
+  public_deps = [
+    "//third_party/angle:libEGL_static",
+    "//third_party/angle:libGLESv2_static",
+  ]
+  public_configs = [ ":egl_and_gles_angle_config" ]
 }
--- a/starboard/nplb/BUILD.gn
+++ b/starboard/nplb/BUILD.gn
@@ -323,6 +323,7 @@ test("nplb") {
     "//starboard:starboard_group",
     "//starboard/common",
     "//starboard/common:file_wrapper",
+    "//starboard/egl_and_gles:buildflags",
     "//starboard/nplb/compiler_compliance:cpp14_supported",
     "//starboard/shared/starboard/media:media_util",
     "//starboard/shared/starboard/player:video_dmp",
--- a/starboard/shared/starboard/player/filter/testing/BUILD.gn
+++ b/starboard/shared/starboard/player/filter/testing/BUILD.gn
@@ -47,6 +47,7 @@ if (current_toolchain == starboard_toolchain) {
     deps = cobalt_platform_dependencies
     deps += [
       "//$starboard_path:starboard_platform",
+      "//starboard/egl_and_gles:buildflags",
       "//starboard/shared/starboard/player/filter:filter_based_player_sources",
     ]
 
--- a/starboard/testing/fake_graphics_context_provider.cc
+++ b/starboard/testing/fake_graphics_context_provider.cc
@@ -18,6 +18,7 @@
 #include <mutex>
 
 #include "starboard/common/log.h"
+#include "starboard/egl_and_gles/buildflags.h"
 
 #if defined(ADDRESS_SANITIZER)
 // By default, Leak Sanitizer and Address Sanitizer is expected to exist
@@ -60,39 +61,16 @@
 #define EGL_WIDTH SB_EGL_WIDTH
 #define EGL_WINDOW_BIT SB_EGL_WINDOW_BIT
 
-#ifndef EGL_ANGLE_platform_angle
-#define EGL_ANGLE_platform_angle 1
+#if BUILDFLAG(STARBOARD_GL_TYPE_ANGLE)
 #define EGL_PLATFORM_ANGLE_ANGLE 0x3202
 #define EGL_PLATFORM_ANGLE_TYPE_ANGLE 0x3203
+
 #define EGL_PLATFORM_ANGLE_DEVICE_TYPE_ANGLE 0x3209
 #define EGL_PLATFORM_ANGLE_DEVICE_TYPE_EGL_ANGLE 0x348E
-#endif /* EGL_ANGLE_platform_angle */
 
-#ifndef EGL_ANGLE_platform_angle_opengl
-#define EGL_ANGLE_platform_angle_opengl 1
 #define EGL_PLATFORM_ANGLE_TYPE_OPENGL_ANGLE 0x320D
 #define EGL_PLATFORM_ANGLE_TYPE_OPENGLES_ANGLE 0x320E
-#endif /* EGL_ANGLE_platform_angle_opengl */
-
-#if !defined(EGL_VERSION_1_5)
-// Lifted verbatim from egl.h.
-#if defined(_WIN32) && !defined(_WIN32_WCE) && !defined(__SCITECH_SNAP__)
-/* Win32 but not WinCE */
-#define KHRONOS_APIENTRY __stdcall
-#else
-#define KHRONOS_APIENTRY
-#endif
-
-#ifndef EGLAPIENTRY
-#define EGLAPIENTRY KHRONOS_APIENTRY
-#endif
-#define EGLAPIENTRYP EGLAPIENTRY*
-
-typedef SbEglDisplay(EGLAPIENTRYP PFNEGLGETPLATFORMDISPLAYEXTPROC)(
-    SbEglEnum platform,
-    void* native_display,
-    const EGLint* attrib_list);
-#endif  // !defined(EGL_VERSION_1_5)
+#endif  // BUILDFLAG(STARBOARD_GL_TYPE_ANGLE)
 
 #define EGL_CALL(x)                                           \
   do {                                                        \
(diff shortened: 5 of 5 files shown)


yell0wd0g:
@kjyoun WDYT? I know you are into refactors of this kind.

Haven't tested this PR on any device - yet. Looking for 
CI to provide that coverage, ideally.

/gemini review

media/starboard/sbplayer_bridge.h
```diff
@@ -302,6 +302,7 @@ class SbPlayerBridge {
   SbWindow window_;
   SbDrmSystem drm_system_ = kSbDrmSystemInvalid;
   Host* const host_;
+  // TODO(mcasas): Convert this to a base::Callback.
```
kjyoun: nit:  [go/todo-style](https://goto.google.com/todo-style)

kjyoun (commented):
LGTM, with a minor style comment.

For this PR's central premise of being single-threaded, borongc@ can likely provide better insight.

yell0wd0g:
@borongc PTAL (I'll address Gemini and @kjyoun feedback together with yours). Thanks!

borongc:
This PR LGTM in general. Although I agree StarboardRenderer is on a single thread right now, it should be safe for this change. However, I would like to do a sanity check after we cherry pick this PR (https://github.com/youtube/cobalt/pull/6662), and ensure everything works fine in production. Then, we can merge this PR to 26.android.

kjyoun:
#6662 is now cherry-picked to 26.android.

Let's see how it goes. If it has no problem, we can pick this PR as well to 26.android.

media/starboard/sbplayer_set_bounds_helper.cc
```diff
@@ -27,17 +28,20 @@ namespace {
 base::AtomicSequenceNumber s_z_index;
 }  // namespace
 
+SbPlayerSetBoundsHelper::SbPlayerSetBoundsHelper()
+    : task_runner_(base::SequencedTaskRunner::GetCurrentDefault()) {}
```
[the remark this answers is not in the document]
yell0wd0g: @borongc or @kjyoun WDYT?

I always find it more confusing to pass the SequenceTaskRunner
to a ctor because I have to search for the callsite to see 
that.

media/starboard/sbplayer_bridge.h
yell0wd0g: Done!

yell0wd0g:
> #6662 is now cherry-picked to 26.android.
> 
> Let's see how it goes. If it has no problem, we can pick this PR as well to 26.android.

I just gave this a run on my kirkwood, QA build. Watched a bunch of videos, stopped them,
did some seeks, all seemed to work fine.

yell0wd0g (commented):
@borongc (or @kjyoun ) PTAL

media/starboard/sbplayer_set_bounds_helper.cc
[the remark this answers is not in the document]
kjyoun: I believe `SbPlayerSetBoundsHelper` must use the same task runner as `StarboardRenderer`.

If that's correct, then passing the task runner into the constructor is the right approach. It makes the dependency explicit and is more robust than assuming `GetCurrentDefault()` will always be called in the correct context.

media/starboard/starboard_renderer.cc
```diff
@@ -479,8 +476,8 @@ void StarboardRenderer::SetStarboardRendererCallbacks(
 }
 
 void StarboardRenderer::OnVideoGeometryChange(const gfx::Rect& output_rect) {
-  set_bounds_helper_->SetBounds(output_rect.x(), output_rect.y(),
-                                output_rect.width(), output_rect.height());
+  CHECK(task_runner_->RunsTasksInCurrentSequence());
+  set_bounds_helper_.SetBounds(output_rect);
```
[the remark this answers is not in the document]
xiaomings: If this function is always called on `task_runner_`, then we probably don't even need `SbPlayerSetBoundsHelper`, i.e. we can move its functionality into SbPlayerBridge.

Also, if this function isn't called on `task_runner_`, we shouldn't re-post it.  We should revive the lock instead, to ensure that this function is delivered to Starboard asap.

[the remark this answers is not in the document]
yell0wd0g: > If this function is always called on `task_runner_`, then we probably don't even need `SbPlayerSetBoundsHelper`, i.e. we can move its functionality into SbPlayerBridge.

Ah, this is beyond the current CL -- @borongc for opinions? (I'm happy to do it as a follow-up).

> 
> Also, if this function isn't called on `task_runner_`, we shouldn't re-post it. We should revive the lock instead, to ensure that this function is delivered to Starboard asap.

I agree - I chose to ignore this suggestion and leave the CHECK to
guarantee there's only on `SequencedTaskRunner` involved, rather
than what Gemini recommends of preventively supporting multiple
task runners.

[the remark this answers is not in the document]
xiaomings: To clarify, while removing `SbPlayerSetBoundsHelper` might be helpful, we don't have to remove it in this PR or in any immediate follow up PR.

media/starboard/sbplayer_set_bounds_helper.cc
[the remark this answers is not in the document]
borongc: Agree with @kjyoun!

media/starboard/starboard_renderer.cc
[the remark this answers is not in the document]
borongc: SbPlayerSetBoundsHelper and StarboardRenderer should be on the same `task_runner_`, so deprecating `SbPlayerSetBoundsHelper` seems be very reasonable.

media/starboard/sbplayer_set_bounds_helper.cc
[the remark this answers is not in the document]
yell0wd0g: Gotcha, will do in an upcoming commit.

media/starboard/starboard_renderer.cc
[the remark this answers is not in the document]
yell0wd0g: Added a TODO().

media/starboard/sbplayer_bridge.h
```diff
@@ -302,6 +302,7 @@ class SbPlayerBridge {
   SbWindow window_;
   SbDrmSystem drm_system_ = kSbDrmSystemInvalid;
   Host* const host_;
+  // TODO(b/448196546): Convert this to a base::Callback.
```
kjyoun: nit: // TODO: b/448196546 -  Convert this to a base::Callback.

[go/todo-style](https://goto.google.com/todo-style)

https://source.corp.google.com/search?q=TODO:.*-%20lang:cc&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

media/starboard/sbplayer_set_bounds_helper.h
```diff
@@ -15,29 +15,35 @@
...
+class SequencedTaskRunner;
+}
+
 namespace media {
 
 class SbPlayerBridge;
 
-class SbPlayerSetBoundsHelper
-    : public base::RefCountedThreadSafe<SbPlayerSetBoundsHelper> {
+// TODO(b/448196546): Fold this class into SbPlayerBridge.
```
kjyoun: nit: TODO: b/... - Fold ..

media/starboard/sbplayer_set_bounds_helper.h
```diff
@@ -15,29 +15,35 @@
...
 
   void SetPlayerBridge(SbPlayerBridge* player_bridge);
-  bool SetBounds(int x, int y, int width, int height);
+  bool SetBounds(const gfx::Rect& rect);
 
  private:
-  base::Lock lock_;
-  SbPlayerBridge* player_bridge_ GUARDED_BY(lock_) = nullptr;
-  std::optional<gfx::Rect> rect_ GUARDED_BY(lock_);
+  const scoped_refptr<base::SequencedTaskRunner> task_runner_;
```
kjyoun, line 43: Can we use const `base::SequencedTaskRunner const *` here instead of a `scoped_refptr`?

StarboardRenderer owns both objects, so the task_runner is guaranteed to outlive this helper. Passing a reference is simpler and avoids the unnecessary ref-counting.

https://www.chromium.org/developers/smart-pointer-guidelines/

> Reference-counted objects make it difficult to understand ownership and destruction order, especially when multiple threads are involved. There is almost always another way to design your object hierarchy to avoid refcounting.

media/starboard/sbplayer_bridge.h
yell0wd0g: That's the Google TODO style; if I limit the link to not include 
third_party/ and not include starboard/, there's only 15 hits:
https://source.corp.google.com/search?q=TODO:.*-%20lang:cc%20-third_party%20-starboard&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

whereas TODO(b/...) has almost 3K hits:
https://source.chromium.org/search?q=%22TODO(b%22%20-third_party%2F&sq=&ss=chromium

See:
https://source.chromium.org/search?q=%22TODO(b%22%20-third_party%2F&sq=&ss=chromium
```
Tracking comments for future improvements: // TODO(crbug.com/40781525): ..., or, less optimally, // TODO(knowledgeable_username): ....
```

media/starboard/sbplayer_set_bounds_helper.h
yell0wd0g: Hmm I see your point; I used this pattern almost automatically:
I haven't encountered keeping a member naked ptr to a relevant
(Sequenced)TaskRunner.

I could not find a single reference to such naked ptr in .h 
files in Chromium (there were a few in third_party/perfetto
though):
https://source.chromium.org/search?q=base::.*TaskRunner.*%5C*.*_;%20%20f:%5C.h%20-base%2F%20-third_party%2F&sq=&ss=chromium


but I found a number of references to the scoped_refptr<>
member idiom:
https://source.chromium.org/search?q=%5C%3Cbase::.*TaskRunner%5C%3E.*_;%20%20f:%5C.h%20-base%2F%20&ss=chromium

media/starboard/sbplayer_set_bounds_helper.h
yell0wd0g: Same as a above, `//TODO (b/123): Do this and that.` is the most common 
idiom in Chromium IIUC (and this CL is //media, not //starboard).

media/starboard/sbplayer_bridge.h
kjyoun: I can see plenty of new style of TODOs. [1]
Apparently,  chromium code base is picking up a new style of TODO
and starboard codes already adopted new style of TODOs
Then I think that it's better to use a new style in a new code.
WDYT?

---
[1]
On chromium code base, excluding third_party/, including only CC files

TODO(b/1234): https://source.chromium.org/search?q=%22TODO(b%2F%22%20-third_party%2F%20lang:cc&sq=&ss=chromium

TODO: b/1234: https://source.chromium.org/search?q=%22TODO:%20b%2F%22%20-third_party%2F%20lang:cc&ss=chromium

[2]
"Java" style guides for chromium recommends to use
> TODO(crbug.com/40192027): Even better to use a bug for context.
https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/java/java.md#todos

On c++ style guide.
https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md

> Chromium follows the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html) unless an exception is listed below.

Confusing thing is that internal c++ style guide mentions a new style, while external version mentions only old style.

media/starboard/sbplayer_set_bounds_helper.h
kjyoun: Hmm. disappointed with the sloppiness of Chomium code base. ;)

Agree that we have to follow the idiomic way.

kjyoun (approved):
LGTM except one last push for new style TODO. :)

I don't want to bikeshed. Now I fully shared my option on this. It's now your call.

media/starboard/sbplayer_set_bounds_helper.h
kjyoun: found the chromium code base use `raw_ptr<SequencedTaskRunner> task_runner_` instead, though still there are not many cases that used this pattern.

https://source.chromium.org/search?q=raw_ptr.*SequencedTaskRunner%20lang:cc%20f:h$&ss=chromium%2Fchromium%2Fsrc

media/starboard/sbplayer_bridge.h
yell0wd0g: Done converting. Should read:
```
  // TODO: b/448196546 - Convert this to a base::Callback.                                                                                                                                                      
```

media/starboard/sbplayer_set_bounds_helper.h
yell0wd0g: Changed to new TODO style, should read now
```
// TODO: b/448196546 - Fold this class into SbPlayerBridge.                                                                                                                                                     
```

yell0wd0g:
> LGTM except one last push for new style TODO. :)
> 
> I don't want to bikeshed. Now I fully shared my option on this. It's now your call.

No worries - I see the # of new style TODOs, makes sense to not introduce old style 
ones. Changed them.

yell0wd0g:
@borongc WDYT? PTAL

yell0wd0g:
> @borongc WDYT? PTAL

Since I'll probably be working on the TODOs and @borongc said
> This PR LGTM in general. 

let me land this as-is? -- if there're comments we'll address them 
subsequently.

PR #7093, opened 2025-09-08
Refactor: Replace SB_DCHECK with SB_CHECK

kjyoun:
Replaced SB_DCHECK with SB_CHECK for pointers that are dereferenced immediately after the check. This ensures that the checks are present in release builds, preventing crashes from null pointers.

This helps us pinpoint where null pointer dereference happens from crash stack trace.

#vibe-coded

Bug: 435425692
Bug: 443341225

### DIFF
--- a/.github/config/android-arm.json
+++ b/.github/config/android-arm.json
@@ -20,6 +20,10 @@
     {
       "target": "web/living_room/kabuki/testing/end2end/browse:infra_smoke_test_custom_coat_tcl_smart_tv",
       "test_attempts": "6"
+    },
+    {
+      "target": "web/living_room/kabuki/testing/end2end/browse:infra_smoke_test_custom_coat_boreal_non_shared",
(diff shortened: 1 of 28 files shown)


starboard/android/shared/media_decoder.h
```diff
@@ -182,7 +182,7 @@ class MediaDecoder final
   ::starboard::shared::starboard::ThreadChecker thread_checker_;
 
   const SbMediaType media_type_;
-  Host* host_;
+  Host* const git host_;
```
borongc: nit: typo for `git`?

kjyoun: my bad. I added last-moment change and didn't check afterward.
Thanks for finding this.
Fixed it and also deleted the code in Dtor to set `host_ = NULL`, which is not necessary in  dtor. and Once deleting it, we can make host_ const.

PR #7094, opened 2025-09-08
Add chromium_{ios,mac}-arm64 to cobalt/build/gn.py

fayev-g:
Bug: 443745131

Start to mirror `chromium_android-*` platforms and provide base chromium platforms for Apple platforms. Rebase is still WIP, but upstream chromium tvOS mirrors iOS the closest. mac build will help set up CI.

### DIFF
--- a/cobalt/build/configs/chromium_mac-arm64/args.gn
+++ b/cobalt/build/configs/chromium_mac-arm64/args.gn
@@ -1,1 +1,2 @@
-import("//cobalt/build/configs/common.gn")
+target_os = "mac"
+target_cpu = "arm64"


cobalt/build/gn.py
```diff
@@ -52,6 +52,8 @@
     'chromium_android-arm',
     'chromium_android-arm64',
     'chromium_android-x86',
+    'chromium_ios-arm64',
```
niranjanyardi, line 55: I see you are adding 2 new platforms. Are you planning on adding the builds to CI ?
If yes: You would have to modify the github files in that case: https://source.corp.google.com/h/github/youtube/cobalt/+/main:.github/ , someone from tools and infra will be able to help more

fayev-g: Yes! Was going to do that in follow up.

fayev-g:
> Leaving a +2 from Cobalt Build group
> 
> It would be good to add the platforms to CI - but that can be done in a later PR if needed

Thanks! I think Kaido was interested in helping set it up also. Either way, we will follow up. I just wanted to get these in.

cobalt/build/configs/chromium_ios-arm64/args.gn
```diff
@@ -0,0 +1,2 @@
+target_os = "ios"
```
rakuco, line 1: One post-merge note: this is a non-Blingium build (i.e. it's a Chromium-on-iOS build that uses WebKit and only a handful of directories like `base`). If you'd like to target something closer to tvOS, you need `use_blink=true` as well: https://chromium.googlesource.com/chromium/src/+/main/docs/ios/build_instructions.md#Building-Blink-for-iOS

rakuco: I've sent #7106 to adjust this.

PR #7104, opened 2025-09-08
cobalt: Reenable UserAgentPlatformInfo unit tests

yell0wd0g:
This CL reenables the UserAgentPlatformInfo unit tests, which are disabled on ToT. This would be a relatively simple operation except for the fact that there're two calls into Starboad (namely SbSystemGetProperty() and SbSystemGetExtension()) which make the unit test not really a unit test (and also crashes on AndroidTV due to JNI StarboardBridge not being properly loaded). This CL/PR just adds a `...for_testing` flag to avoid those calls.

This CL/PR also makes InitializeUserAgentPlatformInfoFields() a member function, which reduces verbosity and allows for the use of the mentioned `...for_testing` flag.

This CL contains the less controversial bits of #7076

Bug: 442760345

### DIFF
--- a/cobalt/browser/user_agent/user_agent_platform_info.h
+++ b/cobalt/browser/user_agent/user_agent_platform_info.h
@@ -26,7 +26,7 @@ void GetUserAgentInputMap(
 
 class UserAgentPlatformInfo {
  public:
-  UserAgentPlatformInfo();
+  explicit UserAgentPlatformInfo(bool for_testing = false);
   ~UserAgentPlatformInfo() = default;
 
   std::string ToString() const;
@@ -107,6 +107,8 @@ class UserAgentPlatformInfo {
   void set_build_configuration(const std::string& build_configuration);
 
  private:
+  void InitializeUserAgentPlatformInfoFields();
+
   std::string starboard_version_;
   std::string os_name_and_version_;
   std::optional<std::string> original_design_manufacturer_;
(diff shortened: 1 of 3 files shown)


yell0wd0g:
@haozheng-cobalt PTAL - @y4vor FYI

cobalt/browser/user_agent/user_agent_platform_info.h
```diff
@@ -129,6 +131,8 @@ class UserAgentPlatformInfo {
   std::string cobalt_version_;
   std::string cobalt_build_version_number_;
   std::string build_configuration_;
+
+  bool avoid_access_to_starboard_for_testing_ = false;
```
haozheng-cobalt, line 135: Great job fixing the UA tests! I guess the UA tests are running and passing on the CI now?

Is that correct that this flag is a temporary workaround until we fix the starboard JNI issue on unittests? If so should we add a TODO here to remind us to remove this flag when the issue is fixed?

yell0wd0g: Well, it is temporary for the time being, but I'm thinking
for unit tests purposes, we should not access the Android OS
-- we could have another target called e.g. `cobalt_tests`,
where we do so.

The catch here is that this flag is not used systematically,
so there're many OS callsites where the code and the unit
tests do query the OS or other platform support. So there's
no purity in these unit tests :) -- but at least Starboard
is kept behind this flag IIUC.

PR #7105, opened 2025-09-09
Remove unneeded Shape Detection 

haozheng-cobalt:
Shape Detection is an isolated service that exists to facilitate safe face/barcode detection. And it not needed in Cobalt.

Test: built gold ATV locally and ran CUJs.
Bug: 443822030

Memory saved: ~1MB 
```
before:
haozheng@haozheng:~/chromium/src$ ls -lh out/chrobalt-android-arm-gold/apks/Cobalt.apk
-rw-r--r-- 1 haozheng primarygroup 77M Sep  8 23:54 out/chrobalt-android-arm-gold/apks/Cobalt.apk

after:
haozheng@haozheng:~/chromium/src$ ls -lh out/chrobalt-android-arm-gold/apks/Cobalt.apk
-rw-r--r-- 1 haozheng primarygroup 76M Sep  9 00:32 out/chrobalt-android-arm-gold/apks/Cobalt.apk
```

### DIFF
--- a/content/public/android/BUILD.gn
+++ b/content/public/android/BUILD.gn
@@ -187,6 +187,12 @@ android_library("content_full_java") {
     "//url:gurl_java",
     "//url:origin_java",
   ]
+
+  # TODO(b/443992661): Add a GN flag to conditionally disable or enable shape detection.
+  if(is_cobalt) {
+    deps -= [ "//services/shape_detection:shape_detection_java" ]
+  }
+  
   srcjar_deps = [
     ":is_ready_to_pay_service_aidl",
     "//content/browser:client_data_json_generated_enum",
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -222,7 +222,6 @@ android_apk("cobalt_apk") {
     "//content/public/android:content_java",
     "//media/capture/video/android:capture_java",
     "//net/android:net_java",
-    "//services/shape_detection:shape_detection_java",
     "//third_party/mesa_headers",
     "//ui/android:ui_java",
   ]


content/public/android/BUILD.gn
```diff
@@ -187,6 +187,11 @@ android_library("content_full_java") {
     "//url:gurl_java",
     "//url:origin_java",
   ]
+
+  if(is_cobalt) {
+    deps -= [ "//services/shape_detection:shape_detection_java" ]
+  }
```
zhongqiliang, line 194: Is it possible to put this in the BUILD file in the cobalt android layer?

haozheng-cobalt: I don't think it's technically possible, add @andrewsavage1 to help comment on this.

andrewsavage1: Depends how we want to do it—it's possible to stub out the shape_detection_java target in GN, or it's possible to make our own copy of this target and edit it there. The former might make sense if we specifically can't use shape_detection_target (sounds like no, this is just size reduction) and the latter might make sense if we plan to edit this target much more and there aren't upstream references to it that we're using. But it seems likely that the current approach of editing the upstream target is the right approach.

yell0wd0g:
I think this should be done upstream in addition to what's done here -- I'll 
file a crbug.com/new and ideally we'd leave a note in the code to remove 
these specifics provision past a certain rebase milestone/date.

yell0wd0g:
> I think this should be done upstream in addition to what's done here -- I'll file a crbug.com/new and ideally we'd leave a note in the code to remove these specifics provision past a certain rebase milestone/date.

https://g-issues.chromium.org/issues/443992661 -- please leave a `// TODO(b/443992661)`
next to the `//content` delta.

content/public/android/BUILD.gn
haozheng-cobalt: Thanks Andrew, the philosophy is well explained.

haozheng-cobalt:
> > I think this should be done upstream in addition to what's done here -- I'll file a crbug.com/new and ideally we'd leave a note in the code to remove these specifics provision past a certain rebase milestone/date.
> 
> https://g-issues.chromium.org/issues/443992661 -- please leave a `// TODO(b/443992661)` next to the `//content` delta.

Thanks for creating the bug, I've added it in the comment.

zhongqiliang:
https://docs.google.com/document/d/1Nd9WtgBfBFSvJpdHPjIvmagJnPFryuvHjmaE29EaSwo/edit?resourcekey=0-eeZ2S-XYWvzyUyKD2gT_Zw&tab=t.0
Not sure if we need to use TextDetector in the future, it is not used today though.

I am good with remove it now.

yell0wd0g:
> [docs.google.com/document/d/1Nd9WtgBfBFSvJpdHPjIvmagJnPFryuvHjmaE29EaSwo/edit?resourcekey=0-eeZ2S-XYWvzyUyKD2gT_Zw&tab=t.0](https://docs.google.com/document/d/1Nd9WtgBfBFSvJpdHPjIvmagJnPFryuvHjmaE29EaSwo/edit?resourcekey=0-eeZ2S-XYWvzyUyKD2gT_Zw&tab=t.0) Not sure if we need to use TextDetector in the future, it is not used today though.
> 
> I am good with remove it now.

Shape detection has three bits: QR, Text and Face. We don't need any of them,
and IIRC Text was never fully launched (it was complicated to define which
alphabet to use for detection, specially for non-roman typographies). So +1
to remove them all.

PR #7106, opened 2025-09-09
cobalt/build: Build chromium_ios-arm64 with Blink

rakuco:
Just setting `target_os="ios"` enables a WebKit-based build, which is very different from most other Chromium ports and does not help with the idea of getting more coverage for something similar to a tvOS build.

Bug: 443745131


rakuco:

PR #7107, opened 2025-09-09
[tvOS] Add minimum set of required backports + tvos-arm64-simulator build config

rakuco:
This commit includes a series of backports from upstream Chromium that make
it possible to use Chromium's toolchain (including the compiler) to target a
tvOS build. This is in no way a complete list of backports, but rather the
minimum amount of changes we need to be able to 1) specify that a Cobalt
build should target tvOS in GN and in the Cobalt scripts 2) pass the right
arguments to the toolchain 3) detect if we are doing a tvOS build from C++.

The last commit in the series then adds the tvos-arm64-simulator Cobalt
build configuration, so that one can then do

    cobalt/build/gn.py -p tvos-arm64-simulator

and build the `starboard` target (which at the moment is just a handful of
files designed to test that the build is working).

Note: as in the upstream build, `.gclient` needs `target_os=['ios']` so that
`gclient sync` fetches all required iOS/tvOS dependencies.

The focus of this commit is on adding all the required scaffolding: setting
the appropriate GN variables to the right values, adding tvOS configurations
to `cobalt/build` and `starboard/build`, and adding the minimum amount of
C++ and GN configuration files under `starboard/tvos` so that we can
hopefully add tvOS to CI.

The dependency on the cobalt targets in `//BUILD.gn` has been temporarily
disabled for tvOS builds because they depend on further changes to build
cobalt shell as a bundle, which will come later.

Bug: 432823234
Bug: 432824509
Bug: 433180638
Bug: 442923940

### DIFF
--- a/starboard/android/shared/audio_decoder.cc
+++ b/starboard/android/shared/audio_decoder.cc
@@ -71,9 +71,10 @@ void* IncrementPointerByBytes(void* pointer, int offset) {
 
 }  // namespace
 
-AudioDecoder::AudioDecoder(const AudioStreamInfo& audio_stream_info,
-                           SbDrmSystem drm_system,
-                           bool enable_flush_during_seek)
+MediaCodecAudioDecoder::MediaCodecAudioDecoder(
+    const AudioStreamInfo& audio_stream_info,
+    SbDrmSystem drm_system,
+    bool enable_flush_during_seek)
     : audio_stream_info_(audio_stream_info),
       sample_type_(GetSupportedSampleType()),
       enable_flush_during_seek_(enable_flush_during_seek),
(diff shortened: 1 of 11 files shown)


rakuco:
@fayev-g @kaidokert @yell0wd0g PTAL. Following our internal conversations as well as the discussion in #7065, this is one possible approach to land the tvOS scaffolding we'd need for CI to work.

Another idea is to split this into 1) commit adding the Cobalt configuration (similar to #7065) and 2) backports.

(cc @abhijeetk @jkim-julie @Gyuyoung)

build/toolchain/ios/compile_xcassets.py
```diff
@@ -125,41 +126,65 @@ def CompileAssetCatalog(output, platform, target_environment, product_type,
...
+      elif target_environment == 'device':
+        command.extend([
+            '--platform',
+            'iphoneos',
+            '--target-device',
+            'iphone',
+            '--target-device',
+            'ipad',
+        ])
+      elif target_environment == 'catalyst':
```
fayev-g, line 175: Is the catalyst environment required or is it just included here for completeness?

rakuco: It was already here, it was just indented with the rest.

rakuco:
@yell0wd0g are you in the cobalt-3p-repository-owners team by any chance?

yell0wd0g:
> @yell0wd0g are you in the cobalt-3p-repository-owners team by any chance?

 I wasn't sure but I think I am!

rakuco:
> QQ - when rebasing to 138, I guess we could drop
> all the BACKPORT patches...? Or would it need
> more work than that?

I think it will be a mixture: most of them will be part of the tree, but there will be some conflicts in cases where we had to manually make the backports apply on top of M114 (in which case just taking the upstream version of the conflict should work).

PR #7110, opened 2025-09-09
starboard: Check pthread_join completes successfully

kjyoun:
The code relies on a fundamental invariant: after pthread_join() returns, the joined thread is guaranteed to have terminated. This means no more callbacks will be made from that thread.

A failure in `pthread_join()` breaches this invariant and leads to an unexpected state where the supposedly-completed thread is still running and calling callbacks. This can lead to use-after-free errors or other race conditions. This change adds a check to enforce this invariant, preventing the application from continuing in a corrupt state.

#vibe-coded

Bug: 443334872

### DIFF
--- a/starboard/android/shared/audio_track_audio_sink_type.cc
+++ b/starboard/android/shared/audio_track_audio_sink_type.cc
@@ -178,7 +178,7 @@ AudioTrackAudioSink::~AudioTrackAudioSink() {
   quit_ = true;
 
   if (audio_out_thread_) {
-    pthread_join(*audio_out_thread_, nullptr);
+    SB_CHECK_EQ(pthread_join(*audio_out_thread_, nullptr), 0);
   }
 }
 
--- a/starboard/android/shared/continuous_audio_track_sink.cc
+++ b/starboard/android/shared/continuous_audio_track_sink.cc
@@ -107,7 +107,7 @@ ContinuousAudioTrackSink::~ContinuousAudioTrackSink() {
   quit_ = true;
 
   if (audio_out_thread_) {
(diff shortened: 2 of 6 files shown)


kaidokert (approved):
This looks good to me

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -178,7 +178,7 @@ AudioTrackAudioSink::~AudioTrackAudioSink() {
   quit_ = true;
 
   if (audio_out_thread_) {
-    pthread_join(*audio_out_thread_, nullptr);
+    SB_CHECK_EQ(pthread_join(*audio_out_thread_, nullptr), 0);
```
xiaomings, line 181: The code is correct and is what we should have implemented in the first place.

Just wanted to bring up that, if under certain situations the function returns a value other than 0, the app may continue to run with the old code, but will crash in production with the new code.

kjyoun: Crashing the app with a CHECK when pthread_join() fails is the intended behavior because it provides an immediate and clear stack trace.

The audio pipeline's destruction sequence relies on the guarantee that pthread_join() succeeds. Its success ensures the audio thread has terminated and will no longer issue callbacks. If this invariant is broken, a callback could be invoked during the destruction process, causing the app to crash in a corrupt state (see b/443334872#comment3).

Crashing immediately with a CHECK pinpoints the exact cause, preventing a more confusing crash later due to a race condition. This gives us a much cleaner picture of the problem at the earliest possible point.

PR #7114, opened 2025-09-09
[Demo] [WIP] remove media capture

haozheng-cobalt:
This is a demo of how we remove media captures dependencies.
Careful surgery from bottom to up will be needed to remove the code path.  
And once accomplished this should save us few hundreds kb.


content/browser/BUILD.gn
```diff
@@ -3167,6 +3174,12 @@ source_set("browser") {
       "//url:gurl_android",
       "//url:origin_android",
     ]
+    if(is_cobalt) {
+      deps -= [
+        "//media/capture/content/android",
+      "//media/capture/video/android",
```
zhongqiliang, line 3180: indentation

haozheng-cobalt: Sorry this is not working, this is more of a draft plan for me, I'll do smaller PRs to trim the media captures from downstream dependencies to upstream.

PR #849, opened 2023-07-10
Keep the Crx package in memory during an update

hlwarriner:
This change adds support to download a Crx to memory, instead of the file system, and also wires up the in-memory unpack/unzip and verification flows.

b/158043520

Change-Id: I746ca397d0b7ac3724ea9c612f3998840e6362d5

PR #7131, opened 2025-09-10
evergreen: more customizations to update_client

yuying-y:
These customizations support in-memory update and installation slot manipulation. They mostly come from C25, with modifications to work with the new code base. Added a new CrxDownloaderFactoryCobalt class. Unit tests will be included in follow-up PRs.

Issue: 431862767

### DIFF
--- a/components/update_client/component.cc
+++ b/components/update_client/component.cc
@@ -44,6 +44,10 @@
 #include "components/update_client/update_engine.h"
 #include "components/update_client/utils.h"
 
+#if BUILDFLAG(IS_STARBOARD)
+#include "components/update_client/cobalt_slot_management.h"
+#endif
+
 // The state machine representing how a CRX component changes during an update.
 //
 //     +------------------------- kNew
@@ -91,6 +95,8 @@
 //          |                              |
 //     kUpdateError                    kCanUpdate
 
+// TODO(b/448186580): consider replacing all LOG with D(V)LOG.
+
 namespace update_client {
 namespace {
 
@@ -101,17 +107,29 @@ void InstallComplete(scoped_refptr<base::SequencedTaskRunner> main_task_runner,
                      InstallOnBlockingTaskRunnerCompleteCallback callback,
                      const base::FilePath& unpack_path,
                      const CrxInstaller::Result& result) {
+#if BUILDFLAG(IS_STARBOARD)
+  DLOG(INFO) << "InstallComplete";
+#endif
   base::ThreadPool::PostTask(
-      FROM_HERE, kTaskTraits,
-      base::BindOnce(
-          [](scoped_refptr<base::SequencedTaskRunner> main_task_runner,
-             InstallOnBlockingTaskRunnerCompleteCallback callback,
-             const base::FilePath& unpack_path,
-             const CrxInstaller::Result& result) {
-            base::DeletePathRecursively(unpack_path);
-            const ErrorCategory error_category =
-                result.error ? ErrorCategory::kInstall : ErrorCategory::kNone;
-            main_task_runner->PostTask(
+       FROM_HERE, kTaskTraits,
+       base::BindOnce(
+           [](scoped_refptr<base::SequencedTaskRunner> main_task_runner,
+              InstallOnBlockingTaskRunnerCompleteCallback callback,
+              const base::FilePath& unpack_path,
+              const CrxInstaller::Result& result) {
+#if BUILDFLAG(IS_STARBOARD)
+              // TODO(b/448186580): Replace LOG with D(V)LOG
+              LOG(INFO) << "Closure kicked off from InstallComplete";
+              // For Cobalt, don't delete the unpack_path, which is not a
+              // temp directory.
+              // Cobalt uses a dedicated installation slot obtained from
+              // the Installation Manager.
+#else
+              base::DeletePathRecursively(unpack_path);
+#endif
+             const ErrorCategory error_category =
+                 result.error ? ErrorCategory::kInstall : ErrorCategory::kNone;
+             main_task_runner->PostTask(
                 FROM_HERE, base::BindOnce(std::move(callback), error_category,
                                           static_cast<int>(result.error),
                                           result.extended_error));
@@ -123,6 +141,12 @@ void InstallOnBlockingTaskRunner(
     scoped_refptr<base::SequencedTaskRunner> main_task_runner,
     const base::FilePath& unpack_path,
     const std::string& public_key,
+#if BUILDFLAG(IS_STARBOARD)
+    int installation_index,
+    PersistedData* metadata,
+    const std::string& id,
+    const std::string& version,
+#endif
     const std::string& fingerprint,
     std::unique_ptr<CrxInstaller::InstallParams> install_params,
     scoped_refptr<CrxInstaller> installer,
@@ -130,9 +154,15 @@ void InstallOnBlockingTaskRunner(
     InstallOnBlockingTaskRunnerCompleteCallback callback) {
   CHECK(base::DirectoryExists(unpack_path));
 
+#if BUILDFLAG(IS_STARBOARD)
+  DLOG(INFO) << "InstallOnBlockingTaskRunner";
+#endif
+
+#if !BUILDFLAG(IS_STARBOARD)
   // Acquire the ownership of the |unpack_path|.
   base::ScopedTempDir unpack_path_owner;
   std::ignore = unpack_path_owner.Set(unpack_path);
+#endif
 
   if (!base::WriteFile(
           unpack_path.Append(FILE_PATH_LITERAL("manifest.fingerprint")),
@@ -145,6 +175,43 @@ void InstallOnBlockingTaskRunner(
     return;
   }
 
+#if BUILDFLAG(IS_STARBOARD)
+  InstallError install_error = InstallError::NONE;
+  const CobaltExtensionInstallationManagerApi* installation_api =
+      static_cast<const CobaltExtensionInstallationManagerApi*>(
+          SbSystemGetExtension(kCobaltExtensionInstallationManagerName));
+  if (!installation_api) {
+    LOG(ERROR) << "Failed to get installation manager api.";
+    install_error = InstallError::GENERIC_ERROR;
+  } else if (installation_index == IM_EXT_INVALID_INDEX) {
+    LOG(ERROR) << "Installation index is invalid.";
+    install_error = InstallError::GENERIC_ERROR;
+  } else {
+    char app_key[IM_EXT_MAX_APP_KEY_LENGTH];
+    if (installation_api->GetAppKey(app_key, IM_EXT_MAX_APP_KEY_LENGTH) ==
+        IM_EXT_ERROR) {
+      LOG(ERROR) << "Failed to get app key.";
+      install_error = InstallError::GENERIC_ERROR;
+    } else if (CobaltFinishInstallation(installation_api, installation_index,
+                                    unpack_path.value(), app_key)) {
+      // Write the version of the unpacked update package to the persisted data.
+      if (metadata != nullptr) {
+        main_task_runner->PostTask(
+            FROM_HERE, base::BindOnce(
+              &PersistedData::SetLastInstalledEgAndSbVersion,
+              base::Unretained(metadata), id, version,
+              std::to_string(SB_API_VERSION)));
+      }
+    } else {
+      LOG(ERROR) << "CobaltFinishInstallation failed.";
+      install_error = InstallError::GENERIC_ERROR;
+    }
+  }
+
+  CrxInstaller::Result result(install_error);
+  InstallComplete(main_task_runner, std::move(callback), unpack_path, result);
+
+#else
   // Ensures that progress callback is not posted after the completion
   // callback. There is a current design limitation in the update client where a
   // poorly implemented installer could post progress after the completion, and
@@ -200,6 +267,7 @@ void InstallOnBlockingTaskRunner(
           callback_checker,
           base::BindOnce(&InstallComplete, main_task_runner,
                          std::move(callback), unpack_path_owner.Take())));
+#endif                        
 }
 
 #if BUILDFLAG(ENABLE_PUFFIN_PATCHES)
@@ -281,16 +349,52 @@ void UnpackCompleteOnBlockingTaskRunner(
 #else
 void UnpackCompleteOnBlockingTaskRunner(
     scoped_refptr<base::SequencedTaskRunner> main_task_runner,
+#if defined(IN_MEMORY_UPDATES)
+    const base::FilePath& installation_dir,
+#else
     const base::FilePath& crx_path,
+#endif
+#if BUILDFLAG(IS_STARBOARD)
+    int installation_index,
+    PersistedData* metadata,
+    const std::string& id,
+    const std::string& version,
+#endif
     const std::string& fingerprint,
     std::unique_ptr<CrxInstaller::InstallParams> install_params,
     scoped_refptr<CrxInstaller> installer,
     CrxInstaller::ProgressCallback progress_callback,
     InstallOnBlockingTaskRunnerCompleteCallback callback,
     const ComponentUnpacker::Result& result) {
+#if BUILDFLAG(IS_STARBOARD)
+  DLOG(INFO) << "UnpackCompleteOnBlockingTaskRunner";
+#if !defined(IN_MEMORY_UPDATES)
+  base::DeleteFile(crx_path);
+#endif  // !defined(IN_MEMORY_UPDATES)
+#else  // BUILDFLAG(IS_STARBOARD)
   update_client::DeleteFileAndEmptyParentDirectory(crx_path);
+#endif  // BUILDFLAG(IS_STARBOARD)
 
   if (result.error != UnpackerError::kNone) {
+#if BUILDFLAG(IS_STARBOARD)
+    // When there is an error unpacking the downloaded CRX, such as a failure to
+    // verify the package, we should remember to clear out any drain files.
+#if defined(IN_MEMORY_UPDATES)
+    if (base::DirectoryExists(installation_dir)) {
+#else  // defined(IN_MEMORY_UPDATES)
+    if (base::DirectoryExists(crx_path.DirName())) {
+#endif  // defined(IN_MEMORY_UPDATES)
+      const auto installation_api =
+        static_cast<const CobaltExtensionInstallationManagerApi*>(
+          SbSystemGetExtension(kCobaltExtensionInstallationManagerName));
+      if (installation_api) {
+        CobaltSlotManagement cobalt_slot_management;
+        if (cobalt_slot_management.Init(installation_api)) {
+          cobalt_slot_management.CleanupAllDrainFiles();
+        }
+      }
+    }
+#endif  // #if BUILDFLAG(IS_STARBOARD)
     main_task_runner->PostTask(
         FROM_HERE,
         base::BindOnce(std::move(callback), ErrorCategory::kUnpack,
@@ -301,7 +405,14 @@ void UnpackCompleteOnBlockingTaskRunner(
   base::ThreadPool::PostTask(
       FROM_HERE, kTaskTraits,
       base::BindOnce(&InstallOnBlockingTaskRunner, main_task_runner,
-                     result.unpack_path, result.public_key, fingerprint,
+                     result.unpack_path, result.public_key, 
+#if BUILDFLAG(IS_STARBOARD)
+                     installation_index,
+                     metadata,
+                     id,
+                     version,
+#endif                     
+                     fingerprint,
                      std::move(install_params), installer, progress_callback,
                      std::move(callback)));
 }
@@ -334,7 +445,18 @@ void StartInstallOnBlockingTaskRunner(
 void StartInstallOnBlockingTaskRunner(
     scoped_refptr<base::SequencedTaskRunner> main_task_runner,
     const std::vector<uint8_t>& pk_hash,
+#if defined(IN_MEMORY_UPDATES)
+    const base::FilePath& installation_dir,
+    const std::string* crx_str,
+#else
     const base::FilePath& crx_path,
+#endif
+#if BUILDFLAG(IS_STARBOARD)
+    int installation_index,
+    PersistedData* metadata,
+    const std::string& id,
+    const std::string& version,
+#endif
     const std::string& fingerprint,
     std::unique_ptr<CrxInstaller::InstallParams> install_params,
     scoped_refptr<CrxInstaller> installer,
@@ -343,12 +465,31 @@ void StartInstallOnBlockingTaskRunner(
     crx_file::VerifierFormat crx_format,
     CrxInstaller::ProgressCallback progress_callback,
     InstallOnBlockingTaskRunnerCompleteCallback callback) {
+#if BUILDFLAG(IS_STARBOARD)
+  DLOG(INFO) << "StartInstallOnBlockingTaskRunner";
+#endif
   auto unpacker = base::MakeRefCounted<ComponentUnpacker>(
-      pk_hash, crx_path, installer, std::move(unzipper_), std::move(patcher_),
+      pk_hash,
+#if defined(IN_MEMORY_UPDATES)
+      crx_str,
+      installation_dir,
+#else
+      crx_path,
+#endif
+      installer, std::move(unzipper_), std::move(patcher_),
       crx_format);
 
   unpacker->Unpack(base::BindOnce(&UnpackCompleteOnBlockingTaskRunner,
-                                  main_task_runner, crx_path, fingerprint,
+                                  main_task_runner,
+#if defined(IN_MEMORY_UPDATES)
+                                  installation_dir,
+#else
+                                  crx_path,
+#endif
+#if BUILDFLAG(IS_STARBOARD)
+                                  installation_index, metadata, id, version,
+#endif
+                                  fingerprint,
                                   std::move(install_params), installer,
                                   progress_callback, std::move(callback)));
 }
@@ -508,8 +649,20 @@ void Component::Handle(CallbackHandleComplete callback_handle_complete) {
       base::BindOnce(&Component::ChangeState, base::Unretained(this)));
 }
 
+#if BUILDFLAG(IS_STARBOARD)
+void Component::Cancel() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  is_cancelled_ = true;
+  state_->Cancel();
+}
+#endif
+
 void Component::ChangeState(std::unique_ptr<State> next_state) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  LOG(INFO) << "Component::ChangeState next_state="
+    << ((next_state)? next_state->state_name(): "nullptr");
+#endif
 
   previous_state_ = state();
   if (next_state)
@@ -646,6 +799,12 @@ void Component::AppendEvent(base::Value::Dict event) {
 void Component::NotifyObservers(UpdateClient::Observer::Events event) const {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
 
+#if BUILDFLAG(IS_STARBOARD)
+  if (is_cancelled_) {
+    LOG(WARNING) << "Component::NotifyObservers: skip callback";
+    return;
+  }
+#endif
   // There is no corresponding component state for the COMPONENT_WAIT event.
   if (update_context_->crx_state_change_callback &&
       event != UpdateClient::Observer::Events::COMPONENT_WAIT) {
@@ -813,9 +972,21 @@ void Component::State::Handle(CallbackNextState callback_next_state) {
   DoHandle();
 }
 
+#if BUILDFLAG(IS_STARBOARD)
+void Component::State::Cancel() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  // Further work may be needed to ensure cancellation during any state results
+  // in a clear result and no memory leaks.
+}
+#endif
+
 void Component::State::TransitionState(std::unique_ptr<State> next_state) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
   CHECK(next_state);
+#if BUILDFLAG(IS_STARBOARD)
+  LOG(INFO) << "Component::State::TransitionState next_state="
+    << ((next_state)? next_state->state_name(): "nullptr");
+#endif
 
   base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
       FROM_HERE,
@@ -881,6 +1052,25 @@ void Component::StateChecking::DoHandle() {
   auto& component = State::component();
   CHECK(component.crx_component());
 
+#if BUILDFLAG(IS_STARBOARD)
+  auto& config = component.update_context_->config;
+  auto metadata = component.update_context_->update_checker->GetPersistedData();
+  if (metadata) {
+    auto task_runner = base::SequencedTaskRunner::GetCurrentDefault();
+    task_runner->PostTask(
+        FROM_HERE, base::BindOnce(&PersistedData::SetUpdaterChannel,
+                                  base::Unretained(metadata), component.id_,
+                                  config->GetChannel()));
+    task_runner->PostTask(
+        FROM_HERE, base::BindOnce(&PersistedData::SetLatestChannel,
+                                  base::Unretained(metadata),
+                                  config->GetChannel()));
+  } else {
+    LOG(WARNING) << "Failed to get the persisted data store to write the "
+                       "updater channel.";
+  }
+#endif
+
   if (component.error_code_) {
     TransitionState(std::make_unique<StateUpdateError>(&component));
     return;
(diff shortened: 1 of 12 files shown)


components/update_client/url_fetcher_downloader.cc
```diff
@@ -13,34 +14,262 @@
...
+
 namespace update_client {
 
+#if BUILDFLAG(IS_STARBOARD)
+UrlFetcherDownloader::UrlFetcherDownloader(
+    scoped_refptr<CrxDownloader> successor,
+    scoped_refptr<Configurator> config)
+    : CrxDownloader(std::move(successor)),
+      network_fetcher_factory_(config->GetNetworkFetcherFactory()),
+      config_(config) {
```
andrewsavage1: Do we use `config_` at all? It doesn't look like it, in which case I think we could use CrxDownloaderFactoryChromium directly

components/update_client/component.cc
```diff
@@ -26,6 +26,11 @@
 #include "base/task/thread_pool.h"
 #include "base/values.h"
 #include "components/update_client/action_runner.h"
+
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: nit: put conditional includes after other includes: go/cstyle#Names_and_Order_of_Includes

components/update_client/component.cc
```diff
@@ -101,17 +106,27 @@ void InstallComplete(scoped_refptr<base::SequencedTaskRunner> main_task_runner,
...
-             const base::FilePath& unpack_path,
-             const CrxInstaller::Result& result) {
-            base::DeletePathRecursively(unpack_path);
-            const ErrorCategory error_category =
-                result.error ? ErrorCategory::kInstall : ErrorCategory::kNone;
-            main_task_runner->PostTask(
+#if BUILDFLAG(IS_STARBOARD)
+    LOG(INFO) << "InstallComplete";
+#endif
+   base::ThreadPool::PostTask(
```
andrewsavage1: Please keep the original whitespace here to minimize the diff from upstream (I want it to be clear this change is not editing upstream code in place)

components/update_client/component.cc
```diff
@@ -236,7 +301,6 @@ void CrxCachePutCompleteOnCrxCacheBlockingTaskRunner(
 
 void UnpackCompleteOnBlockingTaskRunner(
     scoped_refptr<base::SequencedTaskRunner> main_task_runner,
-    const base::FilePath& crx_path,
```
andrewsavage1, line 239: Don't delete upstream code, guard it with !BUILDFLAG(IS_STARBOARD)

components/update_client/component.cc
```diff
@@ -281,16 +345,52 @@ void UnpackCompleteOnBlockingTaskRunner(
...
     const std::string& fingerprint,
     std::unique_ptr<CrxInstaller::InstallParams> install_params,
     scoped_refptr<CrxInstaller> installer,
     CrxInstaller::ProgressCallback progress_callback,
     InstallOnBlockingTaskRunnerCompleteCallback callback,
     const ComponentUnpacker::Result& result) {
+#if BUILDFLAG(IS_STARBOARD)
+  LOG(INFO) << "UnpackCompleteOnBlockingTaskRunner";
+#if !defined(IN_MEMORY_UPDATES)
+  base::DeleteFile(crx_path);
```
andrewsavage1, line 372: We can't use update_client::DeleteFileAndEmptyParentDirectory?

components/update_client/component.cc
```diff
@@ -301,7 +401,14 @@ void UnpackCompleteOnBlockingTaskRunner(
   base::ThreadPool::PostTask(
       FROM_HERE, kTaskTraits,
       base::BindOnce(&InstallOnBlockingTaskRunner, main_task_runner,
-                     result.unpack_path, result.public_key, fingerprint,
+                     result.unpack_path, result.public_key, 
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 409: Not a hard rule, but I think it's much easier to read macro-conditional calls like this if the full call is in a macro, i.e.
```
#if BAR
foo(a, b, c)
#else
foo(a, c)
#endif
```
instead of 
```
foo(a,
#if BAR
b,
#endif
c);
```

components/update_client/component.cc
```diff
@@ -646,13 +795,19 @@ void Component::AppendEvent(base::Value::Dict event) {
...
+  }
+#endif
   // There is no corresponding component state for the COMPONENT_WAIT event.
   if (update_context_->crx_state_change_callback &&
-      event != UpdateClient::Observer::Events::COMPONENT_WAIT) {
-    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
-        FROM_HERE,
-        base::BindRepeating(update_context_->crx_state_change_callback,
-                            GetCrxUpdateItem()));
+        event != UpdateClient::Observer::Events::COMPONENT_WAIT) {
```
andrewsavage1: revert whitespace

components/update_client/url_fetcher_downloader.cc
yuying-y: It's used to set updater status https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/25.lts.1+:components/update_client/url_fetcher_downloader.cc;l=136.

Sorry it's not included in this PR because the definition of SetUpdaterStatus in cobalt/updater/configurator.cc will be included in the next PR.

components/update_client/component.cc
```diff
@@ -881,6 +1048,24 @@ void Component::StateChecking::DoHandle() {
   auto& component = State::component();
   CHECK(component.crx_component());
 
+#if BUILDFLAG(IS_STARBOARD)
+  auto& config = component.update_context_->config;
+  auto metadata = component.update_context_->update_checker->GetPersistedData();
+  if (metadata != nullptr) {
```
andrewsavage1: nit: just `if (metadata)`

components/update_client/crx_downloader.cc
```diff
@@ -60,16 +61,33 @@ CrxDownloader::download_metrics() const {
...
+                                         std::string* dst,
+#endif                                        
                                          DownloadCallback download_callback) {
+#if BUILDFLAG(IS_STARBOARD)
+  LOG(INFO) << "CrxDownloader::StartDownloadFromUrl: url=" << url;
+#endif
   std::vector<GURL> urls;
   urls.push_back(url);
+#if defined(IN_MEMORY_UPDATES)
+  CHECK(dst != nullptr);
```
andrewsavage1: just `CHECK(dst)`

components/update_client/crx_downloader.cc
```diff
@@ -60,16 +61,33 @@ CrxDownloader::download_metrics() const {
...
 
 void CrxDownloader::StartDownload(const std::vector<GURL>& urls,
                                   const std::string& expected_hash,
+#if defined(IN_MEMORY_UPDATES)
+                                  std::string* dst,
+#endif                                  
                                   DownloadCallback download_callback) {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+#if defined(IN_MEMORY_UPDATES)
+  CHECK(dst != nullptr);
```
andrewsavage1: `CHECK(dst)`

components/update_client/crx_downloader.cc
```diff
@@ -109,20 +145,39 @@ void CrxDownloader::OnDownloadComplete(
...
           // Verifies the hash of a CRX file. Returns NONE or BAD_HASH if
           // the hash of the CRX does not match the |expected_hash|. The input
           // file is deleted in case of errors.
           [](const base::FilePath& filepath, const std::string& expected_hash) {
             if (VerifyFileHash256(filepath, expected_hash))
+#endif
               return CrxDownloaderError::NONE;
+#if BUILDFLAG(IS_STARBOARD)
+#if !defined(IN_MEMORY_UPDATES)
+            base::DeleteFile(filepath);
```
andrewsavage1, line 172: Can we use DeleteFileAndEmptyParentDirectory?

components/update_client/crx_downloader_factory.cc
```diff
@@ -11,10 +11,39 @@
 #include "components/update_client/crx_downloader.h"
 #include "components/update_client/network.h"
 #include "components/update_client/url_fetcher_downloader.h"
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 15: nit: add newline between normal includes and macro

components/update_client/request_sender.cc
```diff
@@ -118,6 +121,16 @@ void RequestSender::SendInternal() {
                      base::Unretained(this), url));
 }
 
+#if BUILDFLAG(IS_STARBOARD)
+void RequestSender::Cancel() {
+  LOG(INFO) << "RequestSender::Cancel";
+  // TODO: enable this in a follow-up PR with the Cobalt network fetcher implementation
```
andrewsavage1: Add a bug number for the TODO and a NOTIMPLEMENTED or NOTREACHED call here

components/update_client/update_checker.cc
```diff
@@ -25,6 +25,9 @@
 #include "base/task/sequenced_task_runner.h"
 #include "base/task/thread_pool.h"
 #include "build/build_config.h"
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: Put conditional includes at bottom

components/update_client/update_checker.cc
```diff
@@ -244,7 +304,38 @@ void UpdateCheckerImpl::CheckForUpdatesHelper(
...
+  }
+}
+
+bool UpdateCheckerImpl::SkipUpdate(
+  const CobaltExtensionInstallationManagerApi* installation_api) {
+  auto free_space_ext = static_cast<const CobaltExtensionFreeSpaceApi*>(
+      SbSystemGetExtension(kCobaltExtensionFreeSpaceName));
+  if (!installation_api) {
+    LOG(WARNING) << "UpdaterChecker::SkipUpdate: missing installation api";
+     return false;
```
andrewsavage1: nit: spacing looks wrong here

components/update_client/url_fetcher_downloader.cc
```diff
@@ -4,6 +4,7 @@
 
 #include "components/update_client/url_fetcher_downloader.h"
 
+#include <stack>
```
andrewsavage1: nit: put this in `#if BUILDFLAG(IS_STARBOARD)` as well

components/update_client/url_fetcher_downloader.cc
```diff
@@ -13,34 +14,262 @@
 #include "base/logging.h"
 #include "base/task/sequenced_task_runner.h"
 #include "base/task/thread_pool.h"
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: put below normal includes

components/update_client/url_fetcher_downloader.cc
```diff
@@ -13,34 +14,262 @@
...
       base::BindOnce(&UrlFetcherDownloader::StartURLFetch, this, url));
+#endif  // BUILDFLAG(IS_STARBOARD)
+}
+
+#if BUILDFLAG(IS_STARBOARD)
+void UrlFetcherDownloader::DoCancelDownload() {
+  LOG(INFO) << "UrlFetcherDownloader::DoCancelDownload";
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  is_cancelled_ = true;
+  // TODO: enable this in a follow-up PR with the Cobalt network fetcher implementation
```
andrewsavage1: Add a bug number and call NOTIMPLEMENTED or NOTREACHED

components/update_client/url_fetcher_downloader.cc
```diff
@@ -87,7 +320,7 @@ void UrlFetcherDownloader::OnNetworkFetcherComplete(int net_error,
   // the request and avoid overloading the server in this case.
   // is not accepting requests for the moment.
   int error = -1;
-  if (!net_error && response_code_ == 200)
+  if (!file_path_.empty() && !net_error && response_code_ == 200)
```
andrewsavage1, line 334: Guard this change with a buildflag

components/update_client/url_fetcher_downloader.h
```diff
@@ -23,30 +28,86 @@ class NetworkFetcherFactory;
...
+#if BUILDFLAG(IS_STARBOARD)
+  void ReportDownloadFailure(const GURL& url, CrxDownloaderError error);
+#endif
+
   SEQUENCE_CHECKER(sequence_checker_);
 
   scoped_refptr<NetworkFetcherFactory> network_fetcher_factory_;
   std::unique_ptr<NetworkFetcher> network_fetcher_;
 
+#if !defined(IN_MEMORY_UPDATES)
```
andrewsavage1: nit: it's generally better to use positive conditionals where possible (here and other places)

hlwarriner:
Does this PR add end-to-end support for the in-memory updates feature? Or are more changes from https://github.com/youtube/cobalt/pull/849 needed / will be added in separate PRs?

hlwarriner:
> Does this PR add end-to-end support for the in-memory updates feature? Or are more changes from #849 needed / will be added in separate PRs?

Oh I see, this PR just includes the in-memory changes under `components/update_client/`.

components/update_client/crx_downloader_factory.cc
```diff
@@ -11,10 +11,39 @@
...
+
+scoped_refptr<CrxDownloader> CrxDownloaderFactoryCobalt::MakeCrxDownloader(
+    scoped_refptr<Configurator> config) const {
+  scoped_refptr<CrxDownloader> url_fetcher_downloader =
+      base::MakeRefCounted<UrlFetcherDownloader>(nullptr, config);
+
+  return url_fetcher_downloader;
+}
+
+#else // BUILDFLAG(IS_STARBOARD)
```
y4vor, line 47: The comment should be on line 78 where the #endif is.

components/update_client/request_sender.cc
y4vor: +1

components/update_client/update_checker.cc
```diff
@@ -244,7 +304,38 @@ void UpdateCheckerImpl::CheckForUpdatesHelper(
...
+  }
+  if (!free_space_ext) {
+    LOG(WARNING) << "UpdaterChecker::SkipUpdate: No FreeSpace Cobalt extension";
+    return false;
+  }
+
+  return CobaltSkipUpdate(installation_api, config_->GetMinFreeSpaceBytes(),
+     free_space_ext->MeasureFreeSpace(kSbSystemPathStorageDirectory),
```
y4vor, line 336: With the POSIX APIs we can probably do the measurement directly. 
We should try to get rid of the extension. 
You can add TODO and file a ticket for implement in a follow up PR.

components/update_client/url_fetcher_downloader.cc
y4vor: +1

components/update_client/component.cc
yuying-y: We can't, because we have to keep the empty directory around.

components/update_client/component.cc
yuying-y: I believe the decision in C25 was made based on the length of the argument list. If it's short, your suggested style works best, and many instances used this style. If it's long with many duplicates, the second style is used. I'll keep this as is.

components/update_client/crx_downloader.cc
yuying-y: No, same as before.

components/update_client/crx_downloader_factory.cc
yuying-y: Does it also help with identifying which `#if` matches to this `#else`? I found some references in the code base: https://source.corp.google.com/search?q=%22%23else%20%2F%2F%22&sq=repo:lbshell-internal%2Fcobalt_src%20AND%20branch:main%09. 

Added the comment to `#endif` as well.

components/update_client/component.h
```diff
@@ -158,8 +171,48 @@ class Component {
...
+        case ComponentState::kUpdateError:
+          return "UpdateError";
+        case ComponentState::kUninstalled:
+          return "Uninstalled";
+        case ComponentState::kRun:
+          return "Run";
+        case ComponentState::kLastStatus:
+          return "LastStatus";
+        default:
+          return "Unknown";
```
yell0wd0g: Switch statements on enumerated types should not have a 
"default" case, so that future additions to the enum (class)
would make the compiler fail here, forcing future developers
to update it.

components/update_client/component.h
```diff
@@ -446,7 +506,19 @@ class Component {
   // The error reported by the update checker.
   int update_check_error_ = 0;
 
+#if defined(IN_MEMORY_UPDATES)
+  // To hold the CRX package in memory. The component owns this string
+  // throughout the entire update.
+  std::string crx_str_;
```
yell0wd0g, line 518: Why use std::string instead of a more semantically correct
type like e.g. `std::vector<uint8_t>` ?

components/update_client/crx_downloader.h
```diff
@@ -144,6 +184,10 @@ class CrxDownloader : public base::RefCountedThreadSafe<CrxDownloader> {
   std::vector<GURL>::iterator current_url_;
 
   std::vector<DownloadMetrics> download_metrics_;
+
+#if defined(IN_MEMORY_UPDATES)
+  std::string* dst_str_;  // not owned, can't be null
```
yell0wd0g: At the very least, this should use [base::RawPtr](https://source.chromium.org/chromium/chromium/src/+/main:base/memory/raw_ptr.md?q=rawptr&ss=chromium%2Fchromium%2Fsrc:base%2F), but, could we
not have raw pointers with lifetime conditions defined in 
comments? WeakPtr would be great, otherwise just have this
as unique_ptr<> and pass ownership?

yell0wd0g (changes requested):
Object ownership could be better defined. 
Where are the unit tests ?

yuying-y:
> Object ownership could be better defined. Where are the unit tests ?

Filed b/446745795.

components/update_client/component.h
yuying-y: Good to know. Updated.

components/update_client/component.h
yuying-y: This request falls under b/444006168. Tracking it as a cleanup in b/444006168.

components/update_client/crx_downloader.h
yuying-y: Also tracked in b/444006168. My understanding of the reason not to use unique_ptr and pass the ownership around is to keep the object valid through the life cycle of component object, to make sure it's available for the whole process of updating, and not destroyed by the other objects which manipulate this crx-as-a-string object. Adding the original author @hlwarriner to comment.

components/update_client/component.cc
```diff
@@ -646,13 +795,19 @@ void Component::AppendEvent(base::Value::Dict event) {
...
+  }
+#endif
   // There is no corresponding component state for the COMPONENT_WAIT event.
   if (update_context_->crx_state_change_callback &&
       event != UpdateClient::Observer::Events::COMPONENT_WAIT) {
-    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
-        FROM_HERE,
-        base::BindRepeating(update_context_->crx_state_change_callback,
-                            GetCrxUpdateItem()));
+      base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
```
andrewsavage1: Keep spacing the same as upstream

components/update_client/component.h
yell0wd0g: Makes sense. Please leave here a `// TODO(b/444006168): Use uint8_t` or
some such here.

components/update_client/component.h
```diff
@@ -446,7 +509,19 @@ class Component {
...
+
+  // With in-memory updates the installation directory is still determined in
+  // the download flow even though it isn't needed until the unpack flow. Since
+  // there is no crx_path_ that the installation directory can be derived from,
+  // a dedicated installation_dir_ data member is added.
```
yell0wd0g: nit: Escape variable names.

components/update_client/crx_downloader.h
yell0wd0g: Please leave a `//TODO()` here.

A first step is to just add here and similar naked pointers
a `raw_ptr<>` template wrapper.

https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/memory/raw_ptr.md;l=1?q=base%2Fmemory%2Fraw_ptr.md&sq=&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain

components/update_client/url_fetcher_downloader.cc
yell0wd0g: In this case I'd suggest leaving a //TODO and removing
it in subsequent CLs/PRs.

components/update_client/url_fetcher_downloader.cc
```diff
@@ -18,29 +18,261 @@
...
+#if defined(IN_MEMORY_UPDATES)
+void UrlFetcherDownloader::DoStartDownload(const GURL& url, std::string* dst) {
+#else  // defined(IN_MEMORY_UPDATES)
 void UrlFetcherDownloader::DoStartDownload(const GURL& url) {
+#endif  // defined(IN_MEMORY_UPDATES)
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  LOG(INFO) << "UrlFetcherDownloader::DoStartDownload";
+  if (is_cancelled_) {
+    LOG(ERROR) << "UrlFetcherDownloader::DoStartDownload: Download already cancelled";
```
yell0wd0g, line 152: Two questions here:
- When I read here and elsewhere "Download already cancelled" it
makes me think that the download in progress is trying to be 
cancelled again, and the effect is ... unknown? Likely nothing
happens? 
- I see `is_cancelled_` starts being false, and then can be
set to true if `UrlFetcherDownloader` is ever `Cancel()`ed. 
But what if I try to start another download? Would the 
execution come to this line and not progress anymore? Or is 
the `UrlFetcherDownloader` only created and instantiated for
one single download?

yuying-y:
> Some comments still.
> 
> Please try to remove `LOG()` statements or at least make them `D(V)LOG()` (see #7177 comments). Also leave `//TODO()s` when appropriate.

These logs are valuable information for Cobalt developers, DevOps and external partners, especially at this moment. If I change them to DLOG(), they'll only be available with debug build, which is not always available on the testing pipeline, or Github releases. If I change them to VLOG(), I'm not sure if `--vmodule` works with the loader_app. I searched the code base, and found `LOG()` being used pervasively. We should solve the problem in a thoughtful and consistent manner, like how KJ flattened the Starboard namespace. Filed b/448186580.

Modified some LOG->DLOG, kept some LOG()s due to the concern above, added overarching TODO comments for them

components/update_client/url_fetcher_downloader.cc
yuying-y: b/162114992 explains more. After `Cancel()` is called, the updater is not expected to start another download. If it does, it should stop here and not progress.

yell0wd0g: Could we have a line explaining this in the member variable 
definition, or in the class documentation? Actually, would 
the introduction of this fail-once variable change the 
behaviour of `UrlFetcherDownloader`?

components/update_client/component.h
yell0wd0g: Unresolving since I don't see the requested TODO.

yell0wd0g (changes requested):
I'm not sure what happened but I left review comments in https://github.com/youtube/cobalt/commit/b0e4998d5478ebf35a83ecd07fd6fce3e3eb3daf#r167054307 (I clicked https://screenshot.googleplex.com/BiKMnKj5R5dAHy5 by mistake), but they don't show up in the "files changed" tab (I thought they would propagate automatically.

Tl;dr: a few comments, nothing major. 

Some of the comments don't seem addressed to me (I unresolved one conversation but I think none are addressed). Maybe something odd has happened with the commits?

yuying-y:
> I'm not sure what happened but I left review comments in [b0e4998#r167054307](https://github.com/youtube/cobalt/commit/b0e4998d5478ebf35a83ecd07fd6fce3e3eb3daf#r167054307) (I clicked [screenshot.googleplex.com/BiKMnKj5R5dAHy5](https://screenshot.googleplex.com/BiKMnKj5R5dAHy5) by mistake), but they don't show up in the "files changed" tab (I thought they would propagate automatically?).
> 
> Tl;dr: a few comments, nothing major.
> 
> Some of the comments don't seem addressed to me (I unresolved one conversation but I think none are addressed). Maybe something odd has happened with the commits?

Yeah, for some reason, the last commit push didn't go through. Besides the previous changes, I also went through the comments in https://github.com/youtube/cobalt/commit/b0e4998d5478ebf35a83ecd07fd6fce3e3eb3daf#r167166116, resolved/filed TODOs, and incorporated in the latest pushed commit.

components/update_client/crx_downloader.h
yuying-y: See https://github.com/youtube/cobalt/commit/b0e4998d5478ebf35a83ecd07fd6fce3e3eb3daf#r167165339.

yuying-y: Ok, I got confused about how Github handles these commits. I removed it at some point, but not sure why it's back. Anyway, added the TODO.

components/update_client/url_fetcher_downloader.cc
yell0wd0g: Can we have the comment on this behaviour in either the 
member variable, or the class? E.g. in l.123 of the .h
file:

```
  // This variable tracks a Cancel() being called: all subsequent
  // download requests will be ignored.
  bool is_cancelled_ = false;
```

components/update_client/crx_downloader_factory.cc
```diff
@@ -42,17 +72,23 @@ scoped_refptr<CrxDownloader> CrxDownloaderFactoryChromium::MakeCrxDownloader(
...
 }
+#endif
 
 }  // namespace
 
 scoped_refptr<CrxDownloaderFactory> MakeCrxDownloaderFactory(
     scoped_refptr<NetworkFetcherFactory> network_fetcher_factory) {
+#if BUILDFLAG(IS_STARBOARD)
+  return base::MakeRefCounted<CrxDownloaderFactoryCobalt>(
+      network_fetcher_factory);
```
yell0wd0g: Bringing here a comment from the lost commit review
--- @yell0wd0g said:
nit: Here and elsewhere where we have a ctors or factory method
with a scoped_refptr<> being passed as parameter and then used
to construct something inside, prefer `std::move`, in this case
e.g.:
```
 return base::MakeRefCounted<CrxDownloaderFactoryCobalt>(
      std::move(network_fetcher_factory));
```

Same in l.27, l.42 of this file, etc.

This is as per calling conventions:
> If the function (at least sometimes) takes a ref on a refcounted object, declare the param as scoped_refptr<T>. The caller can decide whether it wishes to transfer ownership (by calling std::move(t) when passing t) or retain its ref (by simply passing t directly).

https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md#object-ownership-and-calling-conventions

-- To which @yuying-y replied
There's no intention to transfer ownership. Kept as scoped_refptr


-- My reply:

There might be a misunderstanding here. For `scoped_reftptr<>`s, 
`transfer ownership` must be understood as "passing the owned
reference", i.e. the current method `MakeCrxDownloaderFactory()`
has a parameter `network_fetcher_factory` or type `scoped_refptr<>`,
which means it owns a reference to `network_fetcher_factory`.
In these two lines
```
  return base::MakeRefCounted<CrxDownloaderFactoryCobalt>(
      network_fetcher_factory);
```
the compiler will make a copy of `network_fetcher_factory`, by
increasing its reference count +1;  at the end of this method 
the parameter `network_fetcher_factory` will go out of scope,
and the reference count will decrease. So the reference count
increases and decreases as the parameter "moves" through the
function/method. This is called "reference churning" and it's 
what Chromium and Google style guide wants to avoid.

IOW, this function would read:
```
scoped_refptr<CrxDownloaderFactory> MakeCrxDownloaderFactory(
    scoped_refptr<NetworkFetcherFactory> network_fetcher_factory) {
#if BUILDFLAG(IS_STARBOARD)
  return base::MakeRefCounted<CrxDownloaderFactoryCobalt>(
      std::move(network_fetcher_factory));                            //////////
#else
  return base::MakeRefCounted<CrxDownloaderFactoryChromium>(
      network_fetcher_factory);
#endif
}
```

Note that the `!BUILDFLAG(IS_STARBOARD)` branch predates the 
Chromium reference-churning avoidance guidelines, but e.g. l.27 
and l.42 in this file, for example, should use std::move() on 
the refptrs that are being passed around. If you think this is
a widespread problem in these files, feel free to file a bug
and leave here a TODO.

yell0wd0g (changes requested):
Most of the comments addressed, thanks! Will
approve once they are all closed.

Please consider getting a gemini review by
adding /gemini review.

components/update_client/crx_downloader_factory.cc
yuying-y: Thanks for the explanation! Very clear and helpful. Done.

components/update_client/component.cc
```diff
@@ -101,17 +107,28 @@ void InstallComplete(scoped_refptr<base::SequencedTaskRunner> main_task_runner,
...
+       base::BindOnce(
+           [](scoped_refptr<base::SequencedTaskRunner> main_task_runner,
+              InstallOnBlockingTaskRunnerCompleteCallback callback,
+              const base::FilePath& unpack_path,
+              const CrxInstaller::Result& result) {
+#if BUILDFLAG(IS_STARBOARD)
+              // TODO(b/448186580): Replace LOG with D(V)LOG
+              LOG(INFO) << "Closure kicked off from InstallComplete";
+              // For Cobalt, don't delete the unpack_path, which is not a temp directory.
+              // Cobalt uses a dedicated installation slot obtained from the Installation
```
yell0wd0g: nit: These two lines seem longer than 80 columns.

components/update_client/url_fetcher_downloader.cc
```diff
@@ -18,29 +18,266 @@
...
+      directories.push(path.value());
+    } else {
+      unlink(path.value().c_str());
+    }
+  }
+  while (!directories.empty()) {
+    rmdir(directories.top().c_str());
+    directories.pop();
+  }
+}
```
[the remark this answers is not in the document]
yell0wd0g: Conversely I'd say if there's a good reason for not using 
`base::DeletePathRecursively()`, please leave a comment, e.g.
```
// Can't simply use base::DeletePathRecursively() because XYZ.
void CleanupDirectory(base::FilePath& dir) {
```

components/update_client/url_fetcher_downloader.cc
```diff
@@ -18,29 +18,266 @@
...
+
 namespace update_client {
 
+#if BUILDFLAG(IS_STARBOARD)
+UrlFetcherDownloader::UrlFetcherDownloader(
+    scoped_refptr<CrxDownloader> successor,
+    scoped_refptr<Configurator> config)
+    : CrxDownloader(std::move(successor)),
+      network_fetcher_factory_(config->GetNetworkFetcherFactory()),
+      config_(config) {
```
yell0wd0g: nit: `config_(std::move(config)) {`

components/update_client/url_fetcher_downloader.cc
```diff
@@ -18,29 +18,266 @@
...
+  Result result;
+  result.error = static_cast<int>(error);
+
+  DownloadMetrics download_metrics;
+  download_metrics.url = url;
+  download_metrics.downloader = DownloadMetrics::kUrlFetcher;
+  download_metrics.error = -1;
+  download_metrics.downloaded_bytes = -1;
+  download_metrics.total_bytes = -1;
+  download_metrics.download_time_ms = 0;
```
yell0wd0g, line 220: nit: Consider designated initializers:
```
  DownloadMetrics download_metrics = {
    .url = url,
    .downloader = DownloadMetrics::kUrlFetcher,
    .error = -1,
    .downloaded_bytes = -1,
    .total_bytes = -1,
    .download_time_ms = 0};
```
To avoid initializing an empty `DownloadMetrics` unnecesarily.

components/update_client/url_fetcher_downloader.cc
```diff
@@ -87,7 +329,11 @@ void UrlFetcherDownloader::OnNetworkFetcherComplete(int net_error,
   // the request and avoid overloading the server in this case.
   // is not accepting requests for the moment.
   int error = -1;
+#if BUILDFLAG(IS_STARBOARD)
+  if (!file_path_.empty() && !net_error && response_code_ == 200)
+#else
   if (!net_error && response_code_ == 200)
+#endif
```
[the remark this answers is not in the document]
yuying-y: This PR doesn't contain all relevant changes for the in_memory_update feature. That's why it's turned off for now. My local workspace contains this particular change (from C25). I'll include more changes for the feature in follow-up PRs. But since it's brought up, I'll just include this one.

components/update_client/crx_downloader.h
```diff
@@ -144,6 +184,11 @@ class CrxDownloader : public base::RefCountedThreadSafe<CrxDownloader> {
   std::vector<GURL>::iterator current_url_;
 
   std::vector<DownloadMetrics> download_metrics_;
+
+#if defined(IN_MEMORY_UPDATES)
+  // TODO(b/449250040): Replace naked pointers
+  std::string* dst_str_;  // not owned, can't be null
```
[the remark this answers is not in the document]
yuying-y: changed to base::raw_ptr.

components/update_client/url_fetcher_downloader.cc
yuying-y: Reverted due to error:
 ../../components/update_client/url_fetcher_downloader.cc:214:19: error: initialization of non-aggregate type 'DownloadMetrics' with a designated initializer list
  DownloadMetrics download_metrics = {
                  ^                  ~
1 error generated.

PR #7141, opened 2025-09-11
Fix gitignore entries

andrewsavage1:
Removed obsolete keyboxes, moved changes that
didn't need to be upstream, and added rdk.

Bug: 444434208

### DIFF
--- a/starboard/.gitignore
+++ b/starboard/.gitignore
@@ -5,3 +5,5 @@ xcuserdata
 .DS_Store
 keyboxes
 keyboxes-dev
+android/apk/.gradle/
+contrib/rdk/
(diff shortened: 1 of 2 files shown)


starboard/.gitignore
```diff
@@ -3,5 +3,5 @@
 *.swp
 xcuserdata
 .DS_Store
-keyboxes
```
niranjanyardi, line 6: isnt this still needed for DRM playback ? ohh i guess this was just for internal platforms , so not needed now

andrewsavage1: I put it back in after the convo today about possibly bringing these back

PR #7143, opened 2025-09-11
Restore cobalt/browser/device_authentication.{cc/h} and unit test

borongc:
Restore device authentication from C25. This will be implemented as the same as C25 for C26 on 3P devices.

Issue: 442607580


borongc:
FYI @sideb0ard @briantting for rebasing Cobalt as this is from C25.

borongc:
> Would you populate the PR description explaining that you are going to implement 3P lightweight device authentication and this module from C25 is needed?

Updated.

PR #7145, opened 2025-09-11
cobalt/build: Clear |media_use_ffmpeg| GN flag for all platforms

yell0wd0g:
`media_use_ffmpeg` is set to false for both Linux and Android. IIUC
it should be false for all platforms. This CL moves that configuration
to the upper level `configs/common.gn` so every current and future
platform (e.g. tvOS) sees the same setting.

Since `media_use_ffmpeg` is set to false, `ffmpeg_branding` does not
apply so this CL/PR removes it.

Bug: 380293712

### DIFF
--- a/.github/actions/api_leak_detector/action.yaml
+++ b/.github/actions/api_leak_detector/action.yaml
@@ -13,4 +13,4 @@ runs:
         else
           SB_API_VERSION_FLAG="--sb_api_version=${{matrix.sb_api_version}}"
         fi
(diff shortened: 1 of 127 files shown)


yell0wd0g:
@kaidokert since we talked about this today by happenstance.

yell0wd0g:
@fayev-g @rakuco @abhijeetk @Gyuyoung @jkim-julie FYI

yell0wd0g:
@borongc PTAL

yell0wd0g:
Bug updated.

PR #7146, opened 2025-09-11
cobalt: Introduce tvOS user_agent and //base logic

yell0wd0g:
Follow up to #7076. 

This CL introduces `InitializePlatformDependentFieldsTvOS()` to
`UserAgentPlatformInfo` and populates the known bits inside to
generate some common parts. set_rasterizer_type() is moved to
the platform-specific methods, and takes the chance to gather
starboard::GetSystemPropertyString() calls together and under
the existing `avoid_access_to_starboard_for_testing_` umbrella.

It also adds the base::SbSysInfo::* functions needed and 
enables the unittest for that platform (note that 
`BUILDFLAG(IS_IOS)` should be replaced with 
`BUILDFLAG(IS_IOS_TVOS` after the rebase, a TODO is left).

Bug: 442760345
Bug: 444510191


yell0wd0g:
@fayev-g PTAL

@rakuco @jkim-julie @abhijeetk @Gyuyoung FYI/PTAL

base/system/sys_info_starboard.cc
```diff
@@ -22,6 +22,13 @@
 #include "base/system/sys_info.h"
 #include "starboard/common/system_property.h"
 using starboard::GetSystemPropertyString;
+#elif BUILDFLAG(IS_APPLE) && BUILDFLAG(IS_IOS)
```
fayev-g: I think if is_ios is true, is_apple is always true, right? Or no? I thought the is_apple check was used if there were nested checks for like is_ios and is_mac.

yell0wd0g: Ah hmm you're very right. I copied this from ToT Chromium somewhere
@rakuco WDYT?

yell0wd0g:
@johnxwork RS plz :)

base/system/sys_info_starboard.cc
rakuco: It's not very usual indeed, as IS_IOS is a subset of IS_APPLE. I'd go with just IS_IOS here.

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +95,59 @@ std::string SbSysInfo::ModelYear() {
...
+  return base::Contains(kChipsets, systemInfo.machine)
+             ? kChipsets[systeminfo.machine]
+             : "";
+}
+
+std::string SbSysInfo::ModelYear() {
+  struct utsname systemInfo;
+  uname(&systemInfo);
+  // Extracted from
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L27-L44
```
rakuco, line 132: If/when we land `system_get_property.mm`, does it mean we could change this into a call to `SbSystemGetProperty()`?

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +95,59 @@ std::string SbSysInfo::ModelYear() {
...
+      {"AppleTV2,1", "Apple A4"},
+      {"AppleTV3,1", "Apple A5"},
+      {"AppleTV3,2", "Apple A5"},
+      {"AppleTV5,3", "Apple A8"},
+      {"AppleTV6,2", "Apple A10X Fusion"},
+      {"AppleTV11,1", "Apple A12 Bionic"},
+      {"AppleTV14,1", "Apple A15 Bionic"},
+  };
+  return base::Contains(kChipsets, systemInfo.machine)
+             ? kChipsets[systeminfo.machine]
```
johnxwork: Is this a typo? systemInfo vs systeminfo

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +95,59 @@ std::string SbSysInfo::ModelYear() {
...
+      {"AppleTV2,1", "2010"},
+      {"AppleTV3,1", "2012"},
+      {"AppleTV3,2", "2013"},
+      {"AppleTV5,3", "2015"},
+      {"AppleTV6,2", "2017"},
+      {"AppleTV11,1", "2021"},
+      {"AppleTV14,1", "2022"},
+  };
+  return base::Contains(kYears, systemInfo.machine)
+             ? kYears[systeminfo.machine]
```
johnxwork: same typo here

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +95,59 @@ std::string SbSysInfo::ModelYear() {
...
+             : "";
```
johnxwork: The old implementation seems to use "2022" as the fallback: https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L124

cobalt/browser/user_agent/user_agent_platform_info.cc
```diff
@@ -198,6 +203,18 @@ void UserAgentPlatformInfo::InitializePlatformDependentFieldsStarboard() {
...
+
+#elif BUILDFLAG(IS_APPLE) && BUILDFLAG(IS_IOS)
+// TODO(b/444510191): Replace BUILDFLAG(IS_IOS) with BUILDFLAG(IS_IOS_TVOS).
+void UserAgentPlatformInfo::InitializePlatformDependentFieldsTvOS() {
+  const std::string os_name = base::SysInfo::OperatingSystemName();
+  const std::string os_version = base::SysInfo::OperatingSystemVersion();
+  set_os_name_and_version(
+      base::StringPrintf("%s %s", os_name.c_str(), os_version.c_str()));
+
+  set_rasterizer_type("metal");
```
johnxwork, line 214: I haven't checked the new tvOS implementation, but it looks from code search that we didn't mention metal in the UA before.

yell0wd0g: Correct, we didn't, but now it's metal. Should we still say gles?
(Also GLES is not a rastering API, it's a rendering API, so we
were not telling the truth anyway...).

base/system/sys_info_starboard.cc
yell0wd0g: Right. Fixed this and the sibling `kUnknownChipset`.

base/system/sys_info_starboard.cc
yell0wd0g: Thanks! Landing #7107 then -- I'll remove the TODO here and use the right
BUILDFLAG directly.

base/system/sys_info_starboard.cc
yell0wd0g: We could but - FWICT that file provides more `kSbSystemProperty...`
than strictly needed. 

I personally don't like ToT's 
`//cobalt/browser/user_agent --> //base/system --> //starboard`
dependency and would rather have
`//cobalt/browser/user_agent --> //base/system` (e.g. for Android vanilla stuff) -or-
`//cobalt/browser/user_agent --> //starboard` (e.g. for Starboard things);
that would be a +1 to bringing back `//starboard/.../system_get_property.mm`.
But again the question is whether all the code in that file is actually
used.

yell0wd0g:
@johnxwork @fayev-g @rakuco all or most of the comments taken in, PTAL again!

Note that currently there's not much IOS_TVOS coverage in CI, so I could be
writing Fortran and still pass CI. The joys ! :P

base/system/sys_info_starboard_unittest.cc
```diff
@@ -22,7 +22,7 @@
 namespace base {
 namespace starboard {
 
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_IOS_TVOS).
```
rakuco:
```suggestion
#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_IOS_TVOS)
```

base/system/sys_info_starboard.cc
rakuco: Ack. I was asking this from the point of view of "if we end up restoring this file, could we use it and stop duplicating the code here"? Looks like the answer is yes :-)

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +94,61 @@ std::string SbSysInfo::ModelYear() {
...
+      {"AppleTV2,1", "Apple A4"},
+      {"AppleTV3,1", "Apple A5"},
+      {"AppleTV3,2", "Apple A5"},
+      {"AppleTV5,3", "Apple A8"},
+      {"AppleTV6,2", "Apple A10X Fusion"},
+      {"AppleTV11,1", "Apple A12 Bionic"},
+      {"AppleTV14,1", "Apple A15 Bionic"},
+  };
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L49
+  constexpr std::string kUnknownChipset = "ChipsetUnknown";
```
rakuco:
```suggestion
  static constexpr std::string_view kUnknownChipset = "ChipsetUnknown";
```

The current form causes the build to fail:

```
../../base/system/sys_info_starboard.cc:120:25: error: constexpr variable 'kUnknownChipset' must be initialized by a constant expression
  constexpr std::string kUnknownChipset = "ChipsetUnknown";
                        ^                 ~~~~~~~~~~~~~~~~
```

with the `static` part coming from https://groups.google.com/a/chromium.org/g/chromium-dev/c/OXxJy5L96jI/m/tRZHIeZwAgAJ

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +94,61 @@ std::string SbSysInfo::ModelYear() {
...
+      {"AppleTV2,1", "2010"},
+      {"AppleTV3,1", "2012"},
+      {"AppleTV3,2", "2013"},
+      {"AppleTV5,3", "2015"},
+      {"AppleTV6,2", "2017"},
+      {"AppleTV11,1", "2021"},
+      {"AppleTV14,1", "2022"},
+  };
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L48
+  constexpr std::string kUnknownYear = 2022;
```
rakuco:
```suggestion
  static constexpr std::string_view kUnknownYear = "2022";
```

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +94,61 @@ std::string SbSysInfo::ModelYear() {
...
+  // Cobalt 25: https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L126
+  return "YouTube";
+}
+
+std::string SbSysInfo::ChipsetModelNumber() {
+  struct utsname systemInfo;
+  uname(&systemInfo);
+  // Extracted from
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L27-L44
+  std::map<std::string, std::string> kChipsets[] = {
```
rakuco:
```suggestion
  const std::map<std::string_view, std::string_view> kChipsets{
```

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +94,61 @@ std::string SbSysInfo::ModelYear() {
...
+      {"AppleTV5,3", "Apple A8"},
+      {"AppleTV6,2", "Apple A10X Fusion"},
+      {"AppleTV11,1", "Apple A12 Bionic"},
+      {"AppleTV14,1", "Apple A15 Bionic"},
+  };
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L49
+  constexpr std::string kUnknownChipset = "ChipsetUnknown";
+  return base::Contains(kChipsets, systemInfo.machine)
+             ? kChipsets[systemInfo.machine]
+             : kUnknownChipset;
```
rakuco: In upstream reviews, I was advised to avoid the double lookup that happens here (contains + access) and do something like

```c++
if (const auto it = kChipsets.find(systemInfo.machine); it != kChipsets.end()) {
  return std::string(it->second);
}
return std::string(kUnknownChipset);
```

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +94,61 @@ std::string SbSysInfo::ModelYear() {
...
+}
+
+std::string SbSysInfo::ModelYear() {
+  struct utsname systemInfo;
+  uname(&systemInfo);
+  // Extracted from
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L27-L44
+  std::map<std::string, std::string> kYears[] = {
```
rakuco:
```suggestion
  const std::map<std::string_view, std::string_view> kYears{
```

base/system/sys_info_starboard.cc
```diff
@@ -88,6 +94,61 @@ std::string SbSysInfo::ModelYear() {
...
+      {"AppleTV5,3", "2015"},
+      {"AppleTV6,2", "2017"},
+      {"AppleTV11,1", "2021"},
+      {"AppleTV14,1", "2022"},
+  };
+  // https://github.com/youtube/cobalt/blob/62c2380b7eb0da5889a387c4b9be283656a8575d/starboard/shared/uikit/system_get_property.mm#L48
+  constexpr std::string kUnknownYear = 2022;
+  return base::Contains(kYears, systemInfo.machine)
+             ? kYears[systemInfo.machine]
+             : kUnknownYear;
```
rakuco: Same thing about contains + access.

base/system/sys_info_starboard.cc
yell0wd0g: Done.

base/system/sys_info_starboard.cc
yell0wd0g: Done.

base/system/sys_info_starboard.cc
yell0wd0g: Done.

base/system/sys_info_starboard.cc
yell0wd0g: Done.

base/system/sys_info_starboard.cc
yell0wd0g: Done.

base/system/sys_info_starboard_unittest.cc
yell0wd0g: Done.

base/system/sys_info_starboard.cc
yell0wd0g: Gotcha, done.

yell0wd0g:
@rakuco  comments taken in, PTAL again

@johnxwork  or @andrewsavage1  RS plz.

yell0wd0g:
> (side note: I am @rakuco, I think you've just pinged a random bystander in your previous comment :-)

Haha whoops ! I typed too fast.

andrewsavage1 (approved):

PR #7151, opened 2025-09-12
cobalt: Call reportFullyDrawn() on browser thread

johnxwork:
Previously we were depending on a legacy Starboard extension to invoke reportFullyDrawn() on Android. This PR skips the extension detour and creates a new mojo call so that the eventual JNI call is executed on browser thread safely.

Bug: 426650306


cobalt/renderer/cobalt_render_frame_observer.cc
```diff
@@ -53,14 +57,11 @@ void CobaltRenderFrameObserver::DidClearWindowObject() {
...
-            SbSystemGetExtension(kCobaltExtensionGraphicsName));
-    if (graphics_extension &&
-        strcmp(graphics_extension->name, kCobaltExtensionGraphicsName) == 0 &&
-        graphics_extension->version >= 6) {
-      graphics_extension->ReportFullyDrawn();
+    if (was_reported_) {
+      return;
     }
+    was_reported_ = true;
+    cobalt_renderer_host_->ReportFullyDrawn();
```
yell0wd0g: Alternative formulation:
```
  if (cobalt_renderer_host_.is_bound()) {
    cobalt_renderer_host_->ReportFullyDrawn();
    cobalt_renderer_host_.reset();
  }
```

And then we don't need `was_reported_`.

cobalt/browser/performance/public/mojom/rendering.mojom
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+module performance.mojom;
+
+// Implemented in the browser process. The renderer process gets a remote to this
+// interface and calls its methods to send messages to the browser.
```
yell0wd0g: Ideally we'd write here why we're interested in this Report, e.g.
```
// Browser process service to get notified by a Renderer of a 
// non-empty DidMeaningfulLayout. A service side can use this
// signal to e.g. do bla and foo.
```
WDYT?

cobalt/browser/cobalt_browser_interface_binders.cc
```diff
@@ -70,6 +72,8 @@ void PopulateCobaltFrameBinders(
       base::BindRepeating(&h5vcc_runtime::H5vccRuntimeImpl::Create));
   binder_map->Add<performance::mojom::CobaltPerformance>(
       base::BindRepeating(&performance::PerformanceImpl::Create));
+  binder_map->Add<performance::mojom::CobaltRendering>(
+      base::BindRepeating(&browser::CobaltRendererHostImpl::Create));
```
yell0wd0g: Just a thought: is this going to be implemented on 3P as well as on 
Android? Currently CobaltRendererHostImpl is only used to forward to
`starboard::android::shared::StarboardBridge::GetInstance()`.

If that forwarding is only needed on Android, then we can do here
something similar to `ForwardToJavaFrame` and create a Java class
implementing mojo's CobaltRendering similar to what we do for
`CrashAnnotator` ([here](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/browser/). 

Perhaps we could even make `StarboardBridge` implement that 
`import org.chromium.performance.mojom.CobaltRendering`, I'm
not sure about this though.

cobalt/browser/performance/public/mojom/rendering.mojom
```diff
@@ -0,0 +1,22 @@
...
+interface CobaltRendering {
```
yell0wd0g, line 20: This file should be called `cobalt_rendering.mojom` IIUC.

cobalt/browser/cobalt_renderer_host_impl.h
```diff
@@ -0,0 +1,51 @@
...
+
+namespace content {
+class RenderFrameHost;
+}  // namespace content
+
+namespace cobalt {
+namespace browser {
+
+class CobaltRendererHostImpl
+    : public content::DocumentService<performance::mojom::CobaltRendering> {
```
yell0wd0g: 1. Do we need to be a `DocumentService` here? The client side sends
a single `ReportFullyDrawn()` signal; and IIUC that cannot be
sent from Renderer if the RenderFrame is destroyed.

2. I like the comment
```
  // The object is bound to the lifetime of the |render_frame_host| and the mojo
  // connection. See DocumentService for details.
```
but I'd like it more here, in the class inheritance : )

3. There's something odd on having a CobaltRendererHostImpl
here only for a RenderFrameHost service implementation: 
RendererHost is a larger thing that RendererFrame, so I think
this class should be `CobaltRenderFrameHostImpl` ? Ah but
`RenderFrameHostImpl` is already a final `//content` class,
so this one cannot pretend to extend that, and would be more
just a `CobaltRenderingImpl`..?

yell0wd0g (changes requested):
Net improvement! Very surprised we still had those calls
to Starboard from within the "Renderer" process :( 

Multiple questions on naming and I think we could 
simplify the code?

cobalt/browser/performance/public/mojom/rendering.mojom
johnxwork: I was following the other file Andrew added in this module, performance.mojom: https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/browser/performance/public/mojom/performance.mojom;drc=1c478c94cfeec6584d5caa80eb9c2d187fa20a9c;l=17

Should we change both?

cobalt/browser/cobalt_browser_interface_binders.cc
johnxwork: I thought about this question.

IMHO the first frame drawn is one of the most comprehensive and reliable signals to measure startup speed. This is why I made the host implementation in native code so that 3P can reuse it.

It's unclear how 3P would use it, maybe it will be an UMA metric or a web interface which doesn't need browser involvement. My best wish is that an unimplemented callback on 3P would encourage some early thinking.

The idea to use java mojom template might be a good one. I didn't think of that. I just thought this implementation to go through a callback posting and then through the legacy JNI function is closer to C25's thread posting so that the comparison on APC is easier(b/426650306): https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/browser/browser_module.cc;drc=06382972bc9b7344d511798b29f2ffac8f5d3fd1;l=856

I could do the cleanup to use java mojom after I see the data on APC. It's a good small practice to compare the two methods.

cobalt/browser/cobalt_renderer_host_impl.h
johnxwork: 1. If I remove the DocumentObserver, the mojo connection can not be established. I guess we would need some other type of binding syntax if we want to replace DocumentObserver;
2. Moved the comment
3. Changed to CobaltRenderingHostImpl instead. Naming is hard indeed

cobalt/browser/performance/public/mojom/rendering.mojom
andrewsavage1: It's under the //cobalt directory, so I'm not sure why we'd need to call the file cobalt_rendering as well?

yell0wd0g: It's just customary for mojom files to be called the snake case
version of the interface name. -Ish, there are many deviations,
but in general I'd say either reconcile the names or just drop
the "Cobalt" from the interface name?

cobalt/browser/cobalt_browser_interface_binders.cc
yell0wd0g: That makes sense. Then I'd suggest leaving a comment like e.g.:
```
  // Nothing prevents us from jumping directly to a Java Mojo
  // implementation, but we chose to bounce on C++ manually to
  // make the JNI jump explicit, like in Cobalt 25 (insert link),
  // and also because in time, 3P implementations should also do
  // something sensible with the ReportFullyDrawn() event.
```

Maybe even file and link to a bug since the last sentence 
sounds like a TODO()...  :=-)

cobalt/browser/cobalt_rendering_host_impl.h
```diff
@@ -0,0 +1,53 @@
...
+namespace browser {
+
+// The object is bound to the lifetime of the |render_frame_host| and the mojo
+// connection.
+class CobaltRenderingHostImpl
+    : public content::DocumentService<performance::mojom::CobaltRendering> {
+ public:
+  static void Create(
+      content::RenderFrameHost* render_frame_host,
+      mojo::PendingReceiver<performance::mojom::CobaltRendering> receiver);
```
yell0wd0g, line 36: If there're no plans to add more complex stuff to this class 
(it's essentially a holder of the connection etc for lifetime
purposes) then I'd just recommend removing the Create() method 
and making the ctor public.

Ah but thinking some more, `Create()` creates a 
`new CobaltRenderingHostImpl` and that is... leaked? This is
odd, but I see all other sibling `Create()`s do the same, also
in //content. So I guess the only request would be to leave a
comment on this function, eg `// For BinderMap creation purposes`,
so it's clear the reason for its existence.

cobalt/browser/cobalt_rendering_host_impl.h
```diff
@@ -0,0 +1,53 @@
...
+
+namespace content {
+class RenderFrameHost;
+}  // namespace content
+
+namespace cobalt {
+namespace browser {
+
+// The object is bound to the lifetime of the |render_frame_host| and the mojo
+// connection.
```
yell0wd0g, line 30: Threading? Leave a comment saying either "This class is
meant to be created/used on a single TaskRunner" or "This 
class has no restrictions re. which TaskRunner is called
from" (because of `base::android::AttachCurrentThread();`).

yell0wd0g (approved):
LGTM with moarrrr comments.

PR #7152, opened 2025-09-12
Fix for linearize_main workflow to use automated/linear_main.

SortaCookie:
This solves the issue of merged side commits entering the automated_commits_m114.json file. Also added a fix for subsequent changes where modify changes didnt work.

Bug: 409339952
Change-Id: I523d2af96e1c8c93e76639dee469b23472b9415b

### DIFF
--- a/.github/workflows/linearize_main.yaml
+++ b/.github/workflows/linearize_main.yaml
@@ -18,6 +18,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           ref: main
+          fetch-tags: true
           fetch-depth: 0
       - name: Checkout rebase_tools
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/linearize_main.yaml
```diff
@@ -35,13 +35,12 @@ jobs:
...
           git config --global user.name "GitHub Release Automation"
           git config --global user.email "github@google.com"
-          echo "REV=$(git rev-parse --verify origin/main)" >> $GITHUB_ENV
       - name: Run linearize
         run: |
           set -eux
           # m114 merge commit=38a06fe8674ad2140ff85c87b5b3a817304e369a
           python main.py linearize --repo-path=${GITHUB_WORKSPACE} --source-branch=main --new-branch-name=automated/linear_main \
-            --start-commit-ref=38a06fe8674ad2140ff85c87b5b3a817304e369a --end-commit-ref=${REV}
+            --start-commit-ref=38a06fe8674ad2140ff85c87b5b3a817304e369a --end-commit-ref="$(git rev-parse --verify origin/main)"
```
kaidokert: I know it's already like this before, but could we tag `38a06fe8674ad2140ff85c87b5b3a817304e369a` and give it a symbolic name ? Then this change becomes immediately more readable

.github/workflows/linearize_main.yaml
```diff
@@ -54,8 +53,11 @@ jobs:
 
           # Create new commits json for m114
           python main.py commits --repo-path=${GITHUB_WORKSPACE} --source-branch=origin/automated/linear_main \
-            --start-commit-ref=38a06fe8674ad2140ff85c87b5b3a817304e369a --end-commit-ref=$REV --output-file=${GITHUB_WORKSPACE}/automated_commits_m114.json
-          git checkout experimental/rebase_tools
+            --start-commit-ref=38a06fe8674ad2140ff85c87b5b3a817304e369a --end-commit-ref="$(git rev-parse --verify origin/automated/linear_main)" \
```
kaidokert: `38a06fe8674ad2140ff85c87b5b3a817304e369a` - same as above. Will it work with a tag ?

PR #7154, opened 2025-09-12
evergreen: Implement lightweight device authentication

borongc:
This PR brings the lightweight device authentication from C25 to C26 for 3P platforms.

The signed cert scope is appended to the startup url if available on 3P platforms, which allows WebFE to recognize certified devices. Note this doesn't include AOSP platforms.

Test the unit test with `./out/linux-x64x11_devel/cobalt_unittests --enable-logging=stderr --v=1 --single-process-tests`:
```
[----------] 4 tests from DeviceAuthenticationTest
[ RUN      ] DeviceAuthenticationTest.ComputeMessageTest
[       OK ] DeviceAuthenticationTest.ComputeMessageTest (0 ms)
[ RUN      ] DeviceAuthenticationTest.ComputeSignatureWithProvidedSecretTest
[       OK ] DeviceAuthenticationTest.ComputeSignatureWithProvidedSecretTest (0 ms)
[ RUN      ] DeviceAuthenticationTest.GetDeviceAuthenticationSignedURLQueryStringFromComponentsTest
[       OK ] DeviceAuthenticationTest.GetDeviceAuthenticationSignedURLQueryStringFromComponentsTest (0 ms)
[ RUN      ] DeviceAuthenticationTest.NoCertSignatureImpliesNoQueryParameters
[       OK ] DeviceAuthenticationTest.NoCertSignatureImpliesNoQueryParameters (0 ms)
[----------] 4 tests from DeviceAuthenticationTest (0 ms total)
```

Issue: 442607580

### DIFF
--- a/cobalt/browser/BUILD.gn
+++ b/cobalt/browser/BUILD.gn
@@ -90,12 +90,17 @@ source_set("browser") {
 }
 
 source_set("switches") {
+  testonly = true
+
   sources = [
     "switches.cc",
     "switches.h",
   ]
 
-  deps = [ "//base" ]
+  deps = [
+    "//base",
+    "//cobalt/shell:cobalt_shell_lib",
+  ]
 }
 
 config("embed_polyfilled_javascript_config") {
--- a/cobalt/renderer/BUILD.gn
+++ b/cobalt/renderer/BUILD.gn
@@ -15,6 +15,8 @@ import("//starboard/build/buildflags.gni")
 import("//testing/test.gni")
 
 source_set("renderer") {
+  testonly = true
+
   sources = [
     "cobalt_content_renderer_client.cc",
     "cobalt_content_renderer_client.h",
--- a/cobalt/shell/BUILD.gn
+++ b/cobalt/shell/BUILD.gn
@@ -323,6 +323,13 @@ static_library("cobalt_shell_lib") {
       "browser/shell_platform_data_aura.h",
     ]
   }
+
+  if (is_starboard) {
+    sources += [
+      "common/device_authentication.cc",
+      "common/device_authentication.h",
+    ]
+  }
 }
 
 # This component provides a ContentMainDelegate for Content Shell and derived
--- a/cobalt/shell/browser/shell_browser_main_parts.cc
+++ b/cobalt/shell/browser/shell_browser_main_parts.cc
@@ -76,6 +76,10 @@
 #include "cobalt/shell/common/shell_test_switches.h"  // nogncheck
 #endif  // defined(RUN_BROWSER_TESTS)
 
+#if BUILDFLAG(IS_STARBOARD)
+#include "cobalt/shell/common/device_authentication.h"
+#endif
+
 namespace content {
 
 namespace {
@@ -89,18 +93,25 @@ GURL GetStartupURL() {
   // Delay renderer creation on Android until surface is ready.
   return GURL();
 #else
+  GURL initial_url = GURL(switches::kDefaultURL);
   const base::CommandLine::StringVector& args = command_line->GetArgs();
-  if (args.empty()) {
-    return GURL("https://www.google.com/");
+  if (!args.empty()) {
+    std::string url_string = args[0];
+    GURL url = GURL(url_string);
+    if (url.is_valid() && url.has_scheme()) {
+      initial_url = url;
+    } else {
+      LOG(WARNING) << "URL \"" << url_string
+                   << "\" from parameter is not valid, open it as a file.";
+      return net::FilePathToFileURL(
+          base::MakeAbsoluteFilePath(base::FilePath(url_string)));
+    }
   }
 
-  GURL url(args[0]);
-  if (url.is_valid() && url.has_scheme()) {
-    return url;
-  }
-
-  return net::FilePathToFileURL(
-      base::MakeAbsoluteFilePath(base::FilePath(args[0])));
+#if BUILDFLAG(IS_STARBOARD)
+  initial_url = GetDeviceAuthenticationSignedURL(initial_url);
+#endif
+  return initial_url;
 #endif
 }
 
--- a/cobalt/shell/common/shell_switches.cc
+++ b/cobalt/shell/common/shell_switches.cc
@@ -47,6 +47,11 @@ const char kContentShellDevToolsTabTarget[] =
 // comma-separated list of origins.
 const char kIsolatedContextOrigins[] = "isolated-context-origins";
 
+// When set, no device authentication parameters will be appended to the initial
+// URL."
+const char kOmitDeviceAuthenticationQueryParameters[] =
+    "omit-device-authentication-query-parameters";
+
 // Use the given address instead of the default loopback for accepting remote
 // debugging connections. Note that the remote debugging protocol does not
 // perform any authentication, so exposing it too widely can be a security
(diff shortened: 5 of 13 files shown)


cobalt/shell/BUILD.gn
```diff
@@ -323,6 +323,13 @@ static_library("cobalt_shell_lib") {
       "browser/shell_platform_data_aura.h",
     ]
   }
+
+  if (is_starboard) {
```
borongc, line 327: `is_cobalt_hermetic_build` should be more accurate here than `is_starboard` for both 3P devices including AOSP.

andrewsavage1:
Are you able to run these unittests for an evergreen platform?

borongc:
> * evergreen-arm-hardfp-raspi / evergreen-arm-hardfp-raspi / arm-hardfp-raspi_on_device_tests (pull_request)

I didn't see we build `cobalt_unittests` on evergreen platforms right now.

cobalt/shell/common/shell_switches.cc
```diff
@@ -47,6 +47,11 @@ const char kContentShellDevToolsTabTarget[] =
 // comma-separated list of origins.
 const char kIsolatedContextOrigins[] = "isolated-context-origins";
 
+// When set, no device authentication parameters will be appended to the initial
+// URL."
+const char kOmitDeviceAuthenticationQueryParameters[] =
+    "omit_device_authentication_query_parameters";
```
y4vor: Please use the same style for the flag name as the existing ones. And I know we have it with underscore in C25 but Chrome is using dashes:

omit-device-authentication-query-parameters

cobalt/browser/BUILD.gn
```diff
@@ -90,12 +90,17 @@ source_set("browser") {
 }
 
 source_set("switches") {
+  testonly = true
```
y4vor, line 93: why do you need the testonly flag?

cobalt/renderer/BUILD.gn
```diff
@@ -15,6 +15,8 @@ import("//starboard/build/buildflags.gni")
 import("//testing/test.gni")
 
 source_set("renderer") {
+  testonly = true
```
y4vor, line 18: same question here

cobalt/shell/browser/shell.cc
```diff
@@ -385,6 +388,14 @@ void Shell::ResizeWebContentForTests(const gfx::Size& content_size) {
   g_platform->ResizeWebContent(this, content_size);
 }
 
+std::string Shell::GetStartupURLWithDeviceAuth(const std::string& startup_url) {
+  std::string initial_url = startup_url;
+#if BUILDFLAG(IS_STARBOARD)
```
y4vor: Is your intention to exclude ATV here?

cobalt/browser/BUILD.gn
borongc: The flag is needed as Cobalt shell is still test only, and this depends on it.

cobalt/renderer/BUILD.gn
borongc: Same answer as above.

cobalt/shell/browser/shell.cc
borongc: Yes, Cobalt on ATV (1st party) doesn't have `IS_STARBOARD`, and this is only for AOSP (checked with @andrewsavage1).

cobalt/shell/common/shell_switches.cc
borongc: Done.

andrewsavage1:
> I didn't see we build `cobalt_unittests` on evergreen platforms right now.

We should on the evergreen-x64 platform, not raspi or rdk yet.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -247,6 +247,8 @@ public void onWebContentsReady() {
             initializeJavaBridge();
             getStarboardBridge().setWebContents(getActiveWebContents());
 
+            mStartupUrl = mShellManager.getActiveShell().getStartupURL(mStartupUrl);
```
kaidokert: I don't think Android/Java changes are needed for this at all. The same mechanism isn't being used

borongc: This keeps an option for AOSP partner to implement `SbSystemGetProperty()` with `kSbSystemPropertyCertificationScope`, so the watch time on AOSP devices can be certified.

Note we have the same code path for device authentication in C25, but it's different in C26.

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -89,18 +93,25 @@ GURL GetStartupURL() {
...
 
-  GURL url(args[0]);
-  if (url.is_valid() && url.has_scheme()) {
-    return url;
-  }
+#if BUILDFLAG(IS_STARBOARD)
+  initial_url = GetDeviceAuthenticationSignedURL(initial_url);
+#endif
 
-  return net::FilePathToFileURL(
```
andrewsavage1, line 102: Does getting rid of this mean we'll no longer be able to automatically load up static html pages? Having that option can be very nice, especially on device when it's harder to connect to a test server. Is it possible to keep this?

cobalt/shell/common/shell_switches.h
```diff
@@ -21,6 +21,8 @@
 
 namespace switches {
 
+constexpr char kDefaultURL[] = "https://www.youtube.com/tv";
```
andrewsavage1, line 24: We're changing the default URL, so can we get rid of the flags we pass with `--url`?

cobalt/shell/browser/shell_browser_main_parts.cc
borongc: Done.

cobalt/shell/common/shell_switches.h
borongc: We can do this in the following PR. For this PR, let's focus on device authentication only.

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -89,14 +93,25 @@ GURL GetStartupURL() {
...
-  if (url.is_valid() && url.has_scheme()) {
-    return url;
+  if (initial_url.is_valid() && initial_url.has_scheme()) {
+#if BUILDFLAG(IS_STARBOARD)
+    initial_url = GetDeviceAuthenticationSignedURL(initial_url);
+#endif
+    return initial_url;
   }
 
   return net::FilePathToFileURL(
```
andrewsavage1: Isn't this code unreachable right now? If `args[0]` is a file path, I think `initial_url` would stay as the default url, meaning we'd never get here

One way to address this is to have `initial_url = GURL(args[0]);` be set regardless of if it's valid, and just log a warning if it's invalid

borongc: Right. Good catch. Fixed.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: As discussed offline, leave AOSP parts as a TODO.

PR #7159, opened 2025-09-13
starboard: Flatten `starboard::shared` namespace used for linux build

kjyoun:
[go/cobalt-flatten-starboard-namespace](https://goto.google.com/cobalt-flatten-starboard-namespace)    

Currently, Linux-specific code uses the `starboard::shared` namespace instead of `starboard::linux`. This workaround exists because `linux` often conflicts with predefined macros or keywords in some build toolchains.

This situation is misleading, as it results in Linux-specific methods residing within the generic shared namespace. Flattening the `starboard` namespace, which this PR contributes to, resolves this confusion by removing the ambiguous nested namespace.

Bug: 441955897

### DIFF
--- a/starboard/android/shared/platform_service.cc
+++ b/starboard/android/shared/platform_service.cc
@@ -157,7 +157,7 @@ Java_dev_cobalt_coat_CobaltService_nativeSendToClient(JNIEnv* env,
 }
 
 const void* GetPlatformServiceApiAndroid() {
-  return &android::shared::kPlatformServiceApi;
+  return &kPlatformServiceApi;
 }
 
 }  // namespace starboard


starboard/android/shared/platform_service.cc
```diff
@@ -41,12 +41,11 @@ typedef struct CobaltExtensionPlatformServicePrivate {
   }
 } CobaltExtensionPlatformServicePrivate;
 
-namespace starboard::android::shared {
+namespace starboard {
+namespace android::shared {
```
andrewsavage1: Why do we keep this namespace?

starboard/linux/x64x11/system_get_property_impl.h
```diff
@@ -17,19 +17,12 @@
...
 #include "starboard/system.h"
 
-// Omit namespace linux due to symbol name conflict.
 namespace starboard {
-namespace x64x11 {
 
-bool CopyStringAndTestIfSuccess(char* out_value,
-                                int value_length,
-                                const char* from_value);
+bool GetSystemPropertyLinuxX64x11(SbSystemPropertyId property_id,
```
andrewsavage1: Does the X64x11 need to be specified?

starboard/linux/shared/system_get_extensions.cc
```diff
@@ -52,7 +52,7 @@ const void* SbSystemGetExtension(const char* name) {
   }
 #endif
   if (strcmp(name, kCobaltExtensionPlatformServiceName) == 0) {
-    return starboard::shared::GetPlatformServiceApi();
+    return starboard::GetPlatformServiceApiLinux();
```
andrewsavage1, line 55: Do we compile in multiple GetPlatformServiceApi*? If not, can we just use the name GetPlatformServiceApi?

starboard/android/shared/platform_service.cc
kjyoun: This is a trick to avoid appending "starboard::shared::android::" prefix 
(e.g. starboard::shared::android::JniNewStringStandardUTFOrAbort)

This will be gone once https://github.com/youtube/cobalt/pull/7162 is merged

starboard/linux/shared/system_get_extensions.cc
kjyoun: It would compile, though I prefer methods to have unique names.

A code search for GetPlatformServiceApi returns several matches. I have to spend mental energy to find the specific one I'm looking for.
https://source.corp.google.com/search?q=GetPlatformServiceApi&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

With `GetPlatformServiceApiLinux`, I can find what I want immediately.

starboard/linux/x64x11/system_get_property_impl.h
kjyoun: ditto. To use unique name

andrewsavage1: So GetSystemPropertyLinux exists as well? If not, I would rather omit just the X64x11

kjyoun: Done

starboard/android/shared/platform_service.cc
kjyoun: Now this `namespace android::shared` is removed.

andrewsavage1 (approved):

PR #7160, opened 2025-09-14
starboard: Flatten starboard::nplb namespace

kjyoun:
[go/cobalt-flatten-starboard-namespace](http://go/cobalt-flatten-starboard-namespace)

Bug: 441955897

### DIFF
--- a/starboard/nplb/player_write_sample_test.cc
+++ b/starboard/nplb/player_write_sample_test.cc
@@ -29,7 +29,7 @@ namespace {
 using ::starboard::testing::FakeGraphicsContextProvider;
 using ::testing::ValuesIn;
 
-typedef SbPlayerTestFixture::GroupedSamples GroupedSamples;
+using GroupedSamples = SbPlayerTestFixture::GroupedSamples;
 
 class SbPlayerWriteSampleTest
     : public ::testing::TestWithParam<SbPlayerTestConfig> {
--- a/starboard/nplb/BUILD.gn
+++ b/starboard/nplb/BUILD.gn
@@ -82,6 +82,7 @@ test("nplb") {
     "player_test_fixture.h",
     "player_test_util.cc",
     "player_test_util.h",
+    "player_write_sample_test.cc",
   ]
 
   sources = [ "//starboard/common/test_main.cc" ]


starboard/nplb/player_write_sample_test.cc
```diff
@@ -23,14 +23,14 @@
...
-namespace starboard {
 namespace nplb {
 namespace {
 
 using ::testing::ValuesIn;
 
 typedef SbPlayerTestFixture::GroupedSamples GroupedSamples;
-typedef testing::FakeGraphicsContextProvider FakeGraphicsContextProvider;
+using ::starboard::testing::FakeGraphicsContextProvider
+    FakeGraphicsContextProvider;
```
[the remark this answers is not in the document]
kjyoun: Done

kjyoun:
> One option to consider is `::nplb`. So `::starboard::npb` or `::nplb` are reasonable choices, however `::starboard` is not.

I chose to use `::nplb`.

With `starboard::nplb`, nplb methods can call starboard methods directly without the `starboard::` prefix, which hides the dependency on starboard.

However, with `::nplb`, nplb methods must use the `starboard::` prefix to access starboard namespace. Although this makes the code more verbose, it is worthwhile because it creates a clear distinction between the `starboard` and `nplb` namespaces.

starboard/nplb/drm_session_test.cc
```diff
@@ -146,13 +145,14 @@ class SbDrmSessionTest : public ::testing::Test {
                                   int session_id_size);
 
   // Queue containing callback events for OnSessionUpdateRequestGeneratedFunc().
-  Queue<SessionUpdateRequestGeneratedCallbackEvent>
+  starboard::Queue<SessionUpdateRequestGeneratedCallbackEvent>
```
y4vor, line 148: starboard/common may benefit fromits own flattened namespace e.g. `starboard_common`.

starboard/nplb/player_creation_param_helpers.cc
```diff
@@ -16,13 +16,12 @@
...
 #include "starboard/shared/starboard/player/video_dmp_reader.h"
 
-namespace starboard {
 namespace nplb {
 namespace {
 
-using shared::starboard::media::AudioStreamInfo;
-using shared::starboard::media::VideoStreamInfo;
-using shared::starboard::player::video_dmp::VideoDmpReader;
+using ::starboard::shared::starboard::media::AudioStreamInfo;
```
y4vor, line 22: This is an example of thins that should not happen in NPLB. 
We probably have an open bug somewhere to figure out how to make these player tests respect the abstractions and not use internal implementation code.

starboard/nplb/drm_session_test.cc
kjyoun: Will think about it.

starboard/nplb/player_creation_param_helpers.cc
kjyoun: Filed b/445228035

PR #7164, opened 2025-09-15
starboard: Remove QueueApplication dependencies from ApplicationAndroid

abhijeetk:
ApplicationAndroid uses a stub SbEventHandle implementation that
completely discards all Starboard events by logging them as errors.
Also, Android does not call Application::Run() which is responsible
for starting RunLoop() process events. This means Android doesn't
utilize Starboard's event handling system.

This change refactors ApplicationAndroid to inherit directly from
Application instead of QueueApplication, since the QueueApplication
layer provides event handling infrastructure that Android doesn't
use.

The stubSbEventHandle function demonstrates that Cobalt on Android
is a media-only platform that bypasses Starboard's event system
entirely and handles platform events through Android's native JNI
bridge system instead.

Bug: 445113404

### DIFF
--- a/starboard/android/shared/application_android.h
+++ b/starboard/android/shared/application_android.h
@@ -15,19 +15,16 @@
 #ifndef STARBOARD_ANDROID_SHARED_APPLICATION_ANDROID_H_
 #define STARBOARD_ANDROID_SHARED_APPLICATION_ANDROID_H_
 
-#include <string>
-
 #include "starboard/android/shared/runtime_resource_overlay.h"
 #include "starboard/android/shared/starboard_bridge.h"
 #include "starboard/common/command_line.h"
-#include "starboard/common/log.h"
-#include "starboard/shared/starboard/queue_application.h"
+#include "starboard/shared/starboard/application.h"
 
 namespace starboard::android::shared {
 
 using ::starboard::CommandLine;
 
-class ApplicationAndroid : public QueueApplication {
+class ApplicationAndroid : public Application {
  public:
   ApplicationAndroid(std::unique_ptr<CommandLine> command_line,
                      base::android::ScopedJavaGlobalRef<jobject> asset_manager,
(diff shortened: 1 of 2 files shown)


abhijeetk:
cc : @yell0wd0g @Gyuyoung @jkim-julie @rakuco

starboard/android/shared/application_android.h
```diff
@@ -15,19 +15,19 @@
 #ifndef STARBOARD_ANDROID_SHARED_APPLICATION_ANDROID_H_
 #define STARBOARD_ANDROID_SHARED_APPLICATION_ANDROID_H_
 
-#include <string>
+#include <limits>
```
kjyoun: you may not need additional header here, 

once those impl. moves into .cc files.

starboard/android/shared/application_android.h
```diff
@@ -43,13 +43,26 @@ class ApplicationAndroid : public QueueApplication {
   int64_t app_start_with_android_fix() { return app_start_timestamp_; }
 
  protected:
-  // --- QueueApplication overrides ---
-  bool MayHaveSystemEvents() override { return false; }
-  Event* WaitForSystemEventWithTimeout(int64_t time) override {
-    SB_NOTIMPLEMENTED();
-    return NULL;
+  // --- Application pure virtual method implementations ---
```
kjyoun, line 46: can ou

starboard/android/shared/application_android.h
```diff
@@ -43,13 +43,26 @@ class ApplicationAndroid : public QueueApplication {
...
+  Event* GetNextEvent() override { return nullptr; }
```
kjyoun: Since these event are not called and should not be called.

can you change this impl. method and below into logging SB_LOG(FATAL) ?

```
Event* GetNextEvent() override {
  SB_LOG(FATAL) << __func__ << " should not be called since Android doesn't
utilize Starboard's event handling"
}

starboard/android/shared/application_android.h
kjyoun, line 43: can you move these impl. methods inside .cc file?

kjyoun:
LGTM

@johnxwork 
WDYT?

starboard/android/shared/application_android.h
abhijeetk: Done.

starboard/android/shared/application_android.h
abhijeetk: Done.

starboard/android/shared/application_android.h
abhijeetk: Done.

kjyoun (approved):

PR #7165, opened 2025-09-15
Enable feature DefaultEnableANGLEValidation in gold build

santoshmahto7:
DefaultEnableANGLEValidation is disabled for gold build. When disabled, EGL attribute EGL_CONTEXT_OPENGL_NO_ERROR_KHR must be supported in EGL implementation, which isn't in current EGL(=1.5), so enable it by default.

Bug: 444042898

### DIFF
--- a/cobalt/app/cobalt_switch_defaults_starboard.cc
+++ b/cobalt/app/cobalt_switch_defaults_starboard.cc
@@ -88,7 +88,12 @@ const base::CommandLine::SwitchMap GetCobaltParamSwitchDefaults() {
         // switch.
         // TODO(mcasas): Ideally configure depending on policy.
         {switches::kForceGpuMemAvailableMb, "32"},
-        {switches::kEnableFeatures, "LimitImageDecodeCacheSize:mb/24"},
+        // When DefaultEnableANGLEValidation is disabled (e.g gold/qa), EGL
+        // attribute EGL_CONTEXT_OPENGL_NO_ERROR_KHR is set during egl context
+        // creation, but egl extension required to support the attribute is
+        // missing and causes errors. So Enable it by default.
+        {switches::kEnableFeatures,
+         "LimitImageDecodeCacheSize:mb/24, DefaultEnableANGLEValidation"},
     // Force some ozone settings.
 #if BUILDFLAG(IS_OZONE)
         {switches::kUseGL, "angle"}, {switches::kUseANGLE, "gles-egl"},


cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -76,6 +76,16 @@ static constexpr auto kCobaltToggleSwitches = std::to_array<const char*>({
 
 // Map of switches with parameters and their defaults.
 const base::CommandLine::SwitchMap GetCobaltParamSwitchDefaults() {
+  std::string enabled_features = "LimitImageDecodeCacheSize:mb/24";
+#if !DCHECK_IS_ON
```
andrewsavage1: Use DCHECK_IS_ON like a function, i.e. `#if !DCHECK_IS_ON()`

Also our main distinguisher between debug/devel and qa/gold is generally IS_OFFICIAL_BUILD

cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -76,6 +76,16 @@ static constexpr auto kCobaltToggleSwitches = std::to_array<const char*>({
...
+  // When DefaultEnableANGLEValidation is disabled (e.g gold/qa), EGL
+  // attribute EGL_CONTEXT_OPENGL_NO_ERROR_KHR is set during egl context
+  // creation, but egl extension required to support the attribute is
+  // missing and causes errors. Thus enable it by default for gold/qa
+  // build. For devel/debug build, DefaultEnableANGLEValidation is already
```
andrewsavage1: Is there any harm to just enabling it regardless of config? It sounds like we always want it, and it would just be redundant for debug/devel, but that's fine

santoshmahto7: yes we always want it,  only issue is that context which will be lost.   
We may want to remove this once required EGL extension support is available in future since
the required extension EGL_CONTEXT_OPENGL_NO_ERROR_KHR seems important for low end devices. As per spec:
```
EGL_CONTEXT_OPENGL_NO_ERROR_KHR indicates whether a faster and lower power
    mode should be enabled for the OpenGL ES context.
 ```
So I would prefer to leave it for future reference. Recommendation are welcome

cobalt/app/cobalt_switch_defaults_starboard.cc
santoshmahto7: Yes, should be used as function. Thanks

DCHECK_IS_ON() is used for sake of consistency  with code https://source.chromium.org/chromium/chromium/src/+/main:ui/gl/gl_features.cc;l=196
and avoid surprises.
e.g  if someone do a debug/devel build and set dcheck_alway_on=false then the feature will not be enabled and cause issue.
Recommendation are welcome.

cobalt/app/cobalt_switch_defaults_starboard.cc
sherryzy: I agree with Andrew that we don't need this config check. If we need to revert this in the future, we should guard it with an explicit flag (e.g., ENABLE_COBALT_HERMETIC_HACKS) rather than implicitly gating it based on the build type (devel/debug vs. gold).

Considering this is already a simple feature switch, I'm also fine with not adding this build flag at all and just keeping the current code comment for context. This simplifies the code and avoids adding another logical difference between build types.

santoshmahto7: Make sense to keep things simple enough. I will enable it for all cases. Thanks

PR #7170, opened 2025-09-15
Use bundled SSL certs for Starboard

andrewsavage1:
Fixed: 444028932

### DIFF
--- a/net/cert/internal/system_trust_store.cc
+++ b/net/cert/internal/system_trust_store.cc
@@ -406,7 +406,7 @@ namespace {
 
 class StarboardSystemCerts {
  public:
-  StarboardSystemCerts() {}
+  StarboardSystemCerts() = default;
 
   TrustStoreInMemoryStarboard* system_trust_store() {
     return &system_trust_store_;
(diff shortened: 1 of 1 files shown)


net/cert/internal/system_trust_store.cc
```diff
@@ -396,6 +400,57 @@ void InitializeTrustStoreAndroid() {}
...
+        trust_anchor);
+  }
+
+#if BUILDFLAG(CHROME_ROOT_STORE_SUPPORTED)
+  int64_t chrome_root_store_version() override { return 0; }
+#endif
+};
+
+std::unique_ptr<SystemTrustStore> CreateSslSystemTrustStore() {
+  LOG(INFO) << "Using STARBOARD SYSTEM TRUST STORE!!";
```
andrewsavage1: remove this

net/cert/internal/system_trust_store.cc
```diff
@@ -396,6 +400,57 @@ void InitializeTrustStoreAndroid() {}
...
+  }
+
+  bool UsesSystemTrustStore() const override { return true; }
+
+  bool IsKnownRoot(const ParsedCertificate* trust_anchor) const override {
+    return g_root_certs_starboard.Get().system_trust_store()->Contains(
+        trust_anchor);
+  }
+
+#if BUILDFLAG(CHROME_ROOT_STORE_SUPPORTED)
```
y4vor: why do we need this part in a Starboard implementation of the interface?

y4vor (approved):

net/cert/internal/system_trust_store.cc
andrewsavage1: It was cruft, removed

PR #7177, opened 2025-09-16
evergreen: add Cobalt updater

yuying-y:
Introduces the initial MVP (Minimum Viable Product) of the Cobalt Evergreen updater. The updater is responsible for checking for new versions of the Cobalt application, downloading them, and preparing them for installation.

Key features of this initial implementation include:
   - A new updater module within the cobalt directory, containing all the core updater logic.
   - Configuration management for the update client, allowing for different update channels and servers.
   - Network fetching capabilities to download update packages.
   - Support for unzipping and patching downloaded updates.
   - Integration with the Cobalt build system and addition of new command-line switches for controlling updater behavior.

This is the foundational implementation of the updater, and will be iterated on in future commits.

Issue: 431862767

### DIFF
--- a/cobalt/updater/BUILD.gn
+++ b/cobalt/updater/BUILD.gn
@@ -1,0 +1,55 @@
+# Copyright 2025 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+static_library("updater") {
+  sources = [
+    "configurator.cc",
+    "configurator.h",
+    "network_fetcher.cc",
+    "network_fetcher.h",
+    "patcher.cc",
+    "patcher.h",
+    "prefs.cc",
+    "prefs.h",
+    "unzipper.cc",
+    "unzipper.h",
+    "updater_module.cc",
+    "updater_module.h",
+    "util.cc",
+    "util.h",
+  ]
+
+  deps = [
+    "//base",
+    "//cobalt/browser:switches",
+    "//components/crx_file",
+    "//components/prefs",
+    "//components/update_client",
+    "//components/update_client:buildflags",
+    "//courgette:bsdiff",
+    "//courgette:courgette_lib",
+    "//crypto",
+    "//net",
+    "//net/third_party/quiche",
+    "//services/network/public/cpp",
+    "//services/network/public/mojom",
+    "//starboard:starboard_headers_only",
+    "//starboard/common",
+    "//starboard/loader_app:app_key_files",
+    "//starboard/loader_app:drain_file",
+    "//third_party/zlib/google:zip",
+    "//url",
+  ]
+}
+# TODO(b/446745795): add unit tests
--- a/cobalt/updater/configurator.cc
+++ b/cobalt/updater/configurator.cc
@@ -1,0 +1,419 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cobalt/updater/configurator.h"
+
+#include <regex>
+#include <set>
+#include <utility>
+#include "base/command_line.h"
+#include "base/version.h"
+#include "cobalt/browser/switches.h"
+#include "cobalt/updater/network_fetcher.h"
+#include "cobalt/updater/patcher.h"
+#include "cobalt/updater/prefs.h"
+#include "cobalt/updater/unzipper.h"
+#include "cobalt/updater/util.h"
+#include "cobalt/version.h"
+#include "components/prefs/pref_service.h"
+#include "components/update_client/crx_downloader_factory.h"
+#include "components/update_client/network.h"
+#include "components/update_client/patcher.h"
+#include "components/update_client/protocol_handler.h"
+#include "components/update_client/unzipper.h"
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+#include "starboard/common/system_property.h"
+#include "starboard/system.h"
+#include "url/gurl.h"
+
+using starboard::kSystemPropertyMaxLength;
+
+namespace {
+
+const char kDefaultUpdaterChannel[] = "prod";
+const char kUpdaterJSONDefaultUrlQA[] =
+    "https://omaha-qa.sandbox.google.com/service/update2/json";
+const char kUpdaterJSONDefaultUrl[] =
+    "https://tools.google.com/service/update2/json";
+
+std::string GetDeviceProperty(SbSystemPropertyId id) {
+  char value[kSystemPropertyMaxLength];
+  if (SbSystemGetProperty(id, value, kSystemPropertyMaxLength)) {
+    return std::string(value);
+  }
+  return std::string();
+}
+
+}  // namespace
+
+namespace cobalt {
+namespace updater {
+
+Configurator::Configurator(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
+    : pref_service_(CreatePrefService()),
+      persisted_data_(
+          std::make_unique<update_client::PersistedData>(pref_service_.get(),
+                                                         nullptr)),
+      network_fetcher_factory_(
+          base::MakeRefCounted<NetworkFetcherFactoryCobalt>(
+              url_loader_factory)),
+      crx_downloader_factory_(
+          update_client::MakeCrxDownloaderFactory(network_fetcher_factory_)),
+      unzip_factory_(base::MakeRefCounted<UnzipperFactory>()),
+      patch_factory_(base::MakeRefCounted<PatcherFactory>()),
+      is_forced_update_(0),
+      min_free_space_bytes_(48 * 1024 * 1024),  // 48MB
+      allow_self_signed_packages_(false),
+      require_network_encryption_(true) {
+  LOG(INFO) << "Configurator::Configurator";
+
+  const std::string persisted_channel = persisted_data_->GetLatestChannel();
+  if (persisted_channel.empty()) {
+    SetChannel(kDefaultUpdaterChannel);
+  } else {
+    SetChannel(persisted_channel);
+  }
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kUseUncompressedUpdates)) {
+    use_compressed_updates_.store(false);
+  } else {
+    use_compressed_updates_.store(true);
+  }
+}
+Configurator::~Configurator() {
+  LOG(INFO) << "Configurator::~Configurator";
+}
+
+base::TimeDelta Configurator::InitialDelay() const {
+  return base::Seconds(0);
+}
+
+base::TimeDelta Configurator::NextCheckDelay() const {
+  return base::Hours(5);
+}
+
+base::TimeDelta Configurator::OnDemandDelay() const {
+  return base::Seconds(0);
+}
+
+base::TimeDelta Configurator::UpdateDelay() const {
+  return base::Seconds(0);
+}
+
+std::vector<GURL> Configurator::UpdateUrl() const {
+#if !defined(COBALT_BUILD_TYPE_GOLD)
+  if (allow_self_signed_packages_ && !update_server_url_.empty()) {
+    return std::vector<GURL>{GURL(update_server_url_)};
+  }
+#endif  // !defined(COBALT_BUILD_TYPE_GOLD)
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kUseQAUpdateServer)) {
+    return std::vector<GURL>{GURL(kUpdaterJSONDefaultUrlQA)};
+  } else {
+    return std::vector<GURL>{GURL(kUpdaterJSONDefaultUrl)};
+  }
+}
+
+std::vector<GURL> Configurator::PingUrl() const {
+  return UpdateUrl();
+}
+
+std::string Configurator::GetProdId() const {
+  return "cobalt";
+}
+
+base::Version Configurator::GetBrowserVersion() const {
+  return base::Version(COBALT_MAJOR_VERSION);
+}
+
+std::string Configurator::GetBrand() const {
+  return {};
+}
+
+std::string Configurator::GetLang() const {
+  const char* locale_id = SbSystemGetLocaleId();
+  if (!locale_id) {
+    // If Starboard failed to determine the locale, return a hardcoded, but
+    // valid BCP 47 language tag as required by
+    // https://html.spec.whatwg.org/commit-snapshots/e2f08b4e56d9a098038fb16c7ff6bb820a57cfab/#language-preferences
+    return "en-US";
+  }
+  std::string locale_string(locale_id);
+  // POSIX platforms put time zone id at the end of the locale id, like
+  // |en_US.UTF8|. We remove the time zone id.
+  int first_dot = locale_string.find_first_of(".");
+  if (first_dot != std::string::npos) {
+    return locale_string.substr(0, first_dot);
+  }
+  return locale_string;
+}
+
+std::string Configurator::GetOSLongName() const {
+  return GetDeviceProperty(kSbSystemPropertyPlatformName);
+}
+
+base::flat_map<std::string, std::string> Configurator::ExtraRequestParams()
+    const {
+  base::flat_map<std::string, std::string> params;
+  params.insert(std::make_pair("SABI", SB_SABI_JSON_ID));
+  params.insert(std::make_pair("sbversion", std::to_string(SB_API_VERSION)));
+
+  // The flag name to force an update is changed to `is_forced_update_` to
+  // cover the cases where update is requested by client but channel stays the
+  // same, but the flag name sent to Omaha remains `updaterchannelchanged` to
+  // isolate the changes to Omaha configs. If it's decided to change this flag
+  // name on Omaha, it'll be done in a separate PR after the Omaha configs are
+  // updated.
+  params.insert(std::make_pair(
+      "updaterchannelchanged",
+      std::atomic_load(&is_forced_update_) == 1 ? "True" : "False"));
+  // Brand name
+  params.insert(
+      std::make_pair("brand", GetDeviceProperty(kSbSystemPropertyBrandName)));
+
+  // Model name
+  params.insert(
+      std::make_pair("model", GetDeviceProperty(kSbSystemPropertyModelName)));
+
+  // Original Design manufacturer name
+  params.insert(
+      std::make_pair("manufacturer",
+                     GetDeviceProperty(kSbSystemPropertySystemIntegratorName)));
+
+  // Chipset model number
+  params.insert(std::make_pair(
+      "chipset", GetDeviceProperty(kSbSystemPropertyChipsetModelNumber)));
+
+  // Firmware version
+  params.insert(std::make_pair(
+      "firmware", GetDeviceProperty(kSbSystemPropertyFirmwareVersion)));
+
+  // Model year
+  params.insert(
+      std::make_pair("year", GetDeviceProperty(kSbSystemPropertyModelYear)));
+
+  // User Agent String
+  params.insert(std::make_pair("uastring", user_agent_string_));
+
+  // Certification scope
+  params.insert(std::make_pair(
+      "certscope", GetDeviceProperty(kSbSystemPropertyCertificationScope)));
+
+  // Compression status
+  params.insert(std::make_pair("usecompressedupdates",
+                               GetUseCompressedUpdates() ? "True" : "False"));
+
+  return params;
+}
+
+std::string Configurator::GetDownloadPreference() const {
+  return {};
+}
+
+// TODO(b/449022872): consider snake case names
+scoped_refptr<update_client::NetworkFetcherFactory>
+Configurator::GetNetworkFetcherFactory() {
+  return network_fetcher_factory_;
+}
+
+scoped_refptr<update_client::CrxDownloaderFactory>
+Configurator::GetCrxDownloaderFactory() {
+  return crx_downloader_factory_;
+}
+
+scoped_refptr<update_client::UnzipperFactory>
+Configurator::GetUnzipperFactory() {
+  return unzip_factory_;
+}
+
+scoped_refptr<update_client::PatcherFactory> Configurator::GetPatcherFactory() {
+  return patch_factory_;
+}
+
+bool Configurator::EnabledDeltas() const {
+  return false;
+}
+
+bool Configurator::EnabledBackgroundDownloader() const {
+  return false;
+}
+
+bool Configurator::EnabledCupSigning() const {
+  return false;
+}
+
+PrefService* Configurator::GetPrefService() const {
+  return pref_service_.get();
+}
+
+update_client::ActivityDataService* Configurator::GetActivityDataService()
+    const {
+  return nullptr;
+}
+
+bool Configurator::IsPerUserInstall() const {
+  return true;
+}
+
+absl::optional<bool> Configurator::IsMachineExternallyManaged() const {
+  return false;
+}
+
+update_client::UpdaterStateProvider Configurator::GetUpdaterStateProvider()
+    const {
+  return base::BindRepeating([](bool /*is_machine*/) {
+    return update_client::UpdaterStateAttributes();
+  });
+}
+
+std::string Configurator::GetAppGuidHelper(const std::string& updater_channel,
+                                           const std::string& version,
+                                           const int sb_version) {
+  if (updater_channel == "ltsnightly" || updater_channel == "ltsnightlyqa") {
+    return kOmahaCobaltLTSNightlyAppID;
+  }
+  if (version.find(".lts.") == std::string::npos &&
+      version.find(".master.") != std::string::npos) {
+    return kOmahaCobaltTrunkAppID;
+  }
+  std::string channel(updater_channel);
+  // This regex matches to all static channels for C25 and newer in the format
+  // of XXltsY.
+  if (std::regex_match(updater_channel,
+                       std::regex("(2[5-9]|[3-9][0-9])lts\\d+"))) {
+    channel = "static";
+  }
+  auto it = kChannelAndSbVersionToOmahaIdMap.find(channel +
+                                                  std::to_string(sb_version));
+  if (it != kChannelAndSbVersionToOmahaIdMap.end()) {
+    return it->second;
+  }
+  LOG(INFO) << "Configurator::GetAppGuidHelper updater channel and starboard "
+            << "combination is undefined with the new Omaha configs.";
+
+  // All undefined channel requests go to prod configs except for static
+  // channel requestsf for C24 and older.
+  // TODO(b/449024263): Replace regex matchers with substring_set_matcher or re2
+  if (!std::regex_match(updater_channel, std::regex("2[0-4]lts\\d+")) &&
+      sb_version >= 14 && sb_version <= 16) {
+    const auto it = kChannelAndSbVersionToOmahaIdMap.find(
+        "prod" + std::to_string(sb_version));
+    if (it != kChannelAndSbVersionToOmahaIdMap.end()) {
+      return it->second;
+    }
+  }
+  LOG(INFO) << __func__ << " starboard version is invalid.";
+  return kOmahaCobaltAppID;
+}
+
+std::string Configurator::GetAppGuid() const {
+  const std::string version(COBALT_VERSION);
+  return GetAppGuidHelper(updater_channel_, version, SB_API_VERSION);
+}
+
+std::unique_ptr<update_client::ProtocolHandlerFactory>
+Configurator::GetProtocolHandlerFactory() const {
+  return std::make_unique<update_client::ProtocolHandlerFactoryJSON>();
+}
+
+void Configurator::CompareAndSwapForcedUpdate(int old_value, int new_value) {
+  is_forced_update_.compare_exchange_weak(old_value, new_value);
+}
+
+// The updater channel is get and set by main web module thread and update
+// client thread. The getter and set use a lock to prevent synchronization
+// issue.
+std::string Configurator::GetChannel() const {
+  base::AutoLock auto_lock(const_cast<base::Lock&>(updater_channel_lock_));
+  return updater_channel_;
+}
+
+void Configurator::SetChannel(const std::string& updater_channel) {
+  LOG(INFO) << "Configurator::SetChannel updater_channel=" << updater_channel;
+  base::AutoLock auto_lock(updater_channel_lock_);
+  updater_channel_ = updater_channel;
+}
+
+// The updater status is get by main web module thread and set by the updater
+// thread. The getter and set use a lock to prevent synchronization issue.
+std::string Configurator::GetUpdaterStatus() const {
+  base::AutoLock auto_lock(const_cast<base::Lock&>(updater_status_lock_));
+  return updater_status_;
+}
+
+void Configurator::SetUpdaterStatus(const std::string& status) {
+  base::AutoLock auto_lock(updater_status_lock_);
+  updater_status_ = status;
+}
+
+void Configurator::SetMinFreeSpaceBytes(uint64_t bytes) {
+  base::AutoLock auto_lock(const_cast<base::Lock&>(min_free_space_bytes_lock_));
+  min_free_space_bytes_ = bytes;
+}
+
+uint64_t Configurator::GetMinFreeSpaceBytes() {
+  base::AutoLock auto_lock(const_cast<base::Lock&>(min_free_space_bytes_lock_));
+  return min_free_space_bytes_;
+}
+
+std::string Configurator::GetPreviousUpdaterStatus() const {
+  base::AutoLock auto_lock(
+      const_cast<base::Lock&>(previous_updater_status_lock_));
+  return previous_updater_status_;
+}
+
+void Configurator::SetPreviousUpdaterStatus(const std::string& status) {
+  base::AutoLock auto_lock(previous_updater_status_lock_);
+  previous_updater_status_ = status;
+}
+
+bool Configurator::GetUseCompressedUpdates() const {
+  return use_compressed_updates_.load();
+}
+
+void Configurator::SetUseCompressedUpdates(bool use_compressed_updates) {
+  use_compressed_updates_.store(use_compressed_updates);
+}
+
+bool Configurator::GetAllowSelfSignedPackages() const {
+  return allow_self_signed_packages_.load();
+}
+
+void Configurator::SetAllowSelfSignedPackages(bool allow_self_signed_packages) {
+  allow_self_signed_packages_.store(allow_self_signed_packages);
+}
+
+std::string Configurator::GetUpdateServerUrl() const {
+  base::AutoLock auto_lock(const_cast<base::Lock&>(update_server_url_lock_));
+  return update_server_url_;
+}
+
+void Configurator::SetUpdateServerUrl(const std::string& update_server_url) {
+  LOG(INFO) << "Configurator::SetUpdateServerUrl update_server_url="
+            << update_server_url;
+  base::AutoLock auto_lock(update_server_url_lock_);
+  update_server_url_ = update_server_url;
+}
+
+bool Configurator::GetRequireNetworkEncryption() const {
+  return require_network_encryption_.load();
+}
+void Configurator::SetRequireNetworkEncryption(
+    bool require_network_encryption) {
+  require_network_encryption_.store(require_network_encryption);
+}
+
+}  // namespace updater
+}  // namespace cobalt
--- a/cobalt/updater/configurator.h
+++ b/cobalt/updater/configurator.h
@@ -1,0 +1,163 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef COBALT_UPDATER_CONFIGURATOR_H_
+#define COBALT_UPDATER_CONFIGURATOR_H_
+
+#include <stdint.h>
+
+#include <atomic>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "base/containers/flat_map.h"
+#include "base/memory/ref_counted.h"
+#include "base/synchronization/lock.h"
+#include "base/thread_annotations.h"
+#include "components/update_client/configurator.h"
+#include "components/update_client/persisted_data.h"
+#include "services/network/public/cpp/shared_url_loader_factory.h"
+
+class GURL;
+class PrefService;
+
+namespace base {
+class Version;
+}  // namespace base
+
+namespace service_manager {
+class Connector;
+}  // namespace service_manager
+
+namespace update_client {
+class ActivityDataService;
+class NetworkFetcherFactory;
+class ProtocolHandlerFactory;
+}  // namespace update_client
+
+namespace cobalt {
+namespace updater {
+// Configurator class for the Cobalt updater.
+// It implements the update_client::Configurator interface and provides
+// Cobalt-specific configuration parameters for the update process.
+//
+// Provides Cobalt-specific configurations to the update client, such as update
+// URLs, version information, and other settings required for the update
+// process.
+// Exists for the duration of the application's runtime, destroyed when app is
+// suspended or exits.
+// TODO: (b/451696797) Adding a thread_checker for each TaskRunner accessing
+// the Configurator instance.
+// Accessed by both the updater thread and the main thread. Synchronization
+// mechanisms (locks, atomics) are used to ensure thread safety.
+class Configurator : public update_client::Configurator {
+ public:
+  explicit Configurator(
+      scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory);
+
+  // Configurator for update_client::Configurator.
+  base::TimeDelta InitialDelay() const override;
+  base::TimeDelta NextCheckDelay() const override;
+  base::TimeDelta OnDemandDelay() const override;
+  base::TimeDelta UpdateDelay() const override;
+  std::vector<GURL> UpdateUrl() const override;
+  std::vector<GURL> PingUrl() const override;
+  std::string GetProdId() const override;
+  base::Version GetBrowserVersion() const override;
+  std::string GetChannel() const override;
+  std::string GetBrand() const override;
+  std::string GetLang() const override;
+  std::string GetOSLongName() const override;
+  base::flat_map<std::string, std::string> ExtraRequestParams() const override;
+  std::string GetDownloadPreference() const override;
+  scoped_refptr<update_client::NetworkFetcherFactory> GetNetworkFetcherFactory()
+      override;
+  scoped_refptr<update_client::CrxDownloaderFactory> GetCrxDownloaderFactory()
+      override;
+  scoped_refptr<update_client::UnzipperFactory> GetUnzipperFactory() override;
+  scoped_refptr<update_client::PatcherFactory> GetPatcherFactory() override;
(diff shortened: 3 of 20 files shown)


third_party/zlib/google/BUILD.gn
```diff
@@ -8,7 +8,7 @@ if (build_with_chromium && is_ios) {
   import("//build/config/ios/bundle_data_from_filelist.gni")
 }
 
-if (build_with_chromium) {
+if (build_with_chromium || is_cobalt_hermetic_build) {
```
andrewsavage1: build_with_chromium is already true, why do we need this change?

andrewsavage1:
It doesn't look like updater files are actually being built in this PR—is the idea to bring in files unmodified from c25 then change them in a later PR?

y4vor (approved):

third_party/zlib/google/BUILD.gn
yuying-y: Hmm, I remember seeing a dep error without this condition, but today not anymore. I'll remove it for now.

yuying-y:
> It doesn't look like updater files are actually being built in this PR—is the idea to bring in files unmodified from c25 then change them in a later PR?

The files are modified and verified to work, as part of my local change for the updater. This PR doesn't contain the code to build the updater with the app, because the updater is still being stabilized and I don't want its occasional crashes cause confusion and hassle for the other contributors. The idea of this PR is to submit the bare bone of the updater, in order to unblock the work of the H5vcc updater APIs.

cobalt/updater/BUILD.gn
```diff
@@ -0,0 +1,57 @@
...
+    "//services/network/public/mojom",
+    "//starboard:starboard_headers_only",
+    "//starboard/common",
+    "//starboard/loader_app:app_key_files",
+    "//starboard/loader_app:drain_file",
+    "//third_party/zlib/google:zip",
+    "//url",
+  ]
+
+  # TODO(b/213388707): resolve the dependency cycle and remove this exception.
```
andrewsavage1: Seems to be me can pretty easily avoid this circular dependency, see https://b.corp.google.com/issues/213388707#comment3

cobalt/updater/updater_module.h
```diff
@@ -0,0 +1,196 @@
...
+
+ private:
+  std::unique_ptr<base::Thread> updater_thread_;
+  scoped_refptr<update_client::UpdateClient> update_client_;
+  std::unique_ptr<Observer> updater_observer_;
+  scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
+  scoped_refptr<Configurator> updater_configurator_;
+  int update_check_count_ = 0;
+  bool is_updater_running_;
+  uint64_t update_check_delay_sec_ = kDefaultUpdateCheckDelaySeconds;
```
yell0wd0g: This can be const FWICT. Can you try and make as many member
variables const as possible? Or, if this is not const because
a later CL would modify it (to decrease it, for example), then
a comment or a TODO here would help.

Also, if this is converted to a base::TimeDelta, I suggest
making the const (which should be constexpr) and this member
var that type.

cobalt/updater/updater_module.h
```diff
@@ -0,0 +1,196 @@
...
+  kUpdating,
+  kUpdated,
+  kRolledForward,
+  kUpToDate,
+  kUpdateError,
+  kUninstalled,
+  kRun
+};
+
+// Mapping a component state to an updater status.
```
yell0wd0g: This comment and l.76's don't add anything. It could provide
more information specifying what this is meant to do, or how
it's supposed to be used.

This particular map seems strange, since it maps two enums
of practically the same meaning. Could we use just one of
them? It would simplify the code.

yell0wd0g (changes requested):
This is a huge CL and the description is very succinct.
It also has no unit tests. Is this intended to be a
playground CL from which smaller CLs would be extracted,
reviewed and landed individually?

yuying-y:
> This is a huge CL and the description is very succinct. It also has no unit tests. Is this intended to be a playground CL from which smaller CLs would be extracted, reviewed and landed individually?

Updated the commit message with AI's help. This is already a smaller CL extracted from the updater prototype. It only contains the code under cobalt/updater, and unblocks the work for H5vcc updater APIs. Tracking unit test in b/446745795.

cobalt/updater/updater_module.h
yuying-y: update_check_delay_sec_ is set by the caller when initializing the UpdaterModule object. Here's just the assignment of a default value.

cobalt/updater/updater_module.h
yuying-y: UpdaterStatus has two distinct states unique to Cobalt updater:  kSlotLocked and kRolledForward. We may add even more in the future to indicate the updater state. It doesn't make sense to add these states to component because they're the states of the updater, not the crx component. Elaborated the comment.

cobalt/updater/BUILD.gn
yuying-y: This will change soon. See https://b.corp.google.com/issues/213388707#comment4

andrewsavage1: The only other dependency I see is here: https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/25.lts.1+:components/update_client/cobalt_slot_management.cc;l=324-325?q=%22chrome%2Fupdater%22&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2F25.lts.1%2B:components%2Fupdate_client%2F, on `cobalt::updater::ReadEvergreenVersion`, which also looks easy to fix. Let's fix this.

cobalt/updater/patcher.cc
```diff
@@ -0,0 +1,102 @@
...
+// we should remove this #if.
+#include "third_party/puffin/src/include/puffin/puffpatch.h"
+#endif
+
+namespace cobalt {
+namespace updater {
+
+namespace {
+
+class PatcherImpl : public update_client::Patcher {
```
hlwarriner, line 29: A couple questions about this.

1. Can we reuse or share the implementations from `components/update_client/patch/in_process_patcher.cc` instead of copying them here?
2. Do we really need these implementations? I would't think we plan to support differential updates.

cobalt/updater/util.cc
```diff
@@ -0,0 +1,324 @@
...
+  }
+  if (!base::CreateDirectory(*path)) {
+    LOG(ERROR) << "Can't create product directory.";
+    return false;
+  }
+  return true;
+}
+
+bool GetProductDirectoryPath(base::FilePath* path) {
+#if defined(OS_WIN)
```
hlwarriner: Now that cobalt has its own implementation of this function, can we remove the preprocessor branching for windows and macos? Since this function should only be built/called for Evergreen, we shouldn't need to support these platforms.

cobalt/updater/patcher.cc
yuying-y: 1. These are place-holders at the moment, and will be change if we decide to investigate the patch update.
2. Why do you think we don't plan to investigate differential updates? I think it's still a potential area to reduce storage.

cobalt/updater/util.cc
yuying-y: Sounds good. Removed.

cobalt/updater/BUILD.gn
yuying-y: Removed the dependency. Refactoring will be done in a follow-up PR.

cobalt/updater/patcher.cc
hlwarriner: 1. Ok
2. I more so meant that we don't have any concrete plans to support it. I shared a prototype of differential updates in b/174699165#comment11 but also noted there a couple reasons why we were skeptical of such an approach. Maybe it's something we can revisit in chrobalt. If you want to keep the code here to help with future investigation/experimentation then sgtm.

andrewsavage1:
> The idea of this PR is to submit the bare bone of the updater, in order to unblock the work of the H5vcc updater APIs.

I'm confused how this unblocks the H5vcc APIs if we're not yet building these files? It seems better to actually add the build for these files, and it seems fine to me to just add it to the list of dependencies of `//cobalt:gn_all` so it's built but not actually used when running Cobalt if you're worried about stability issues.

yuying-y:
> > The idea of this PR is to submit the bare bone of the updater, in order to unblock the work of the H5vcc updater APIs.
> 
> I'm confused how this unblocks the H5vcc APIs if we're not yet building these files? It seems better to actually add the build for these files, and it seems fine to me to just add it to the list of dependencies of `//cobalt:gn_all` so it's built but not actually used when running Cobalt if you're worried about stability issues.

H5vcc APIs call the functions of the updater to actually do anything, e.g. SetUpdaterStatus(), etc. This PR defined these functions. When the h5vcc target depend on the updater target, it builds. Also, I'll have a PR soon to add it as a dependency of cobalt shell.

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,422 @@
...
+
+std::string GetDeviceProperty(SbSystemPropertyId id) {
+  char value[kSystemPropertyMaxLength];
+  bool result;
+  result = SbSystemGetProperty(id, value, kSystemPropertyMaxLength);
+  std::string prop;
+  if (result) {
+    prop = std::string(value);
+  }
+  return prop;
```
yell0wd0g: nit: 
```
  if (SbSystemGetProperty(id, value, kSystemPropertyMaxLength)) {
    return std::string(value);
  }
  return std::string();
```
? 4 lines versus current 7.

cobalt/updater/network_fetcher.cc
```diff
@@ -0,0 +1,231 @@
...
+}
+
+void NetworkFetcher::OnProgressCallback(ProgressCallback progress_callback,
+                                        uint64_t current) {
+  progress_callback.Run(base::saturated_cast<int64_t>(current));
+}
+
+NetworkFetcherFactoryCobalt::NetworkFetcherFactoryCobalt(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
+    : url_loader_factory_(url_loader_factory) {}
```
yell0wd0g: nit: `url_loader_factory_(std::move(url_loader_factory))`

https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md#object-ownership-and-calling-conventions
> If the function (at least sometimes) takes a ref on a refcounted object, declare the param as scoped_refptr<T>. The caller can decide whether it wishes to transfer ownership (by calling std::move(t) when passing t) or retain its ref (by simply passing t directly).

cobalt/updater/configurator.h
```diff
@@ -0,0 +1,148 @@
...
+  ~Configurator() override;
+
+  std::unique_ptr<PrefService> pref_service_;
+  std::unique_ptr<update_client::PersistedData> persisted_data_;
+  scoped_refptr<update_client::NetworkFetcherFactory> network_fetcher_factory_;
+  scoped_refptr<update_client::CrxDownloaderFactory> crx_downloader_factory_;
+  scoped_refptr<update_client::UnzipperFactory> unzip_factory_;
+  scoped_refptr<update_client::PatcherFactory> patch_factory_;
+  std::string updater_channel_;
+  base::Lock updater_channel_lock_;
```
yell0wd0g, line 143: This class has no comment on its threading (has no comment at all), 
is it used from multiple threads? Assuming yes, since there's this
lock, consider adding thread checking annotations ([here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/threading/thread_checker.h;l=44-77?q=thread_checker&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain&start=1)) and
also try not to use a lock and instead use PostTask between any
threads inside?

cobalt/updater/configurator.h
```diff
@@ -0,0 +1,148 @@
...
+  std::atomic<int32_t> is_forced_update_;
+  std::string updater_status_;
+  base::Lock updater_status_lock_;
+  std::string previous_updater_status_;
+  base::Lock previous_updater_status_lock_;
+  std::string user_agent_string_;
+  uint64_t min_free_space_bytes_ = 48 * 1024 * 1024;
```
yell0wd0g: nit: Use `GUARDED_BY(...lock_);` where appropriate
https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/thread_annotations.h;l=49-58;drc=d1a68d591a2afd65f70ebc3cbf9f2e678d834a51

cobalt/updater/configurator.h
```diff
@@ -0,0 +1,148 @@
...
+  base::Lock min_free_space_bytes_lock_;
```
yell0wd0g, line 152: nit: Can we use `std::atomic<uint64_t>`? 

Also in this class there's atomics and there're variable
with locks, but could we just use one big lock? (I only 
took a quick look at the .cc file because is very large),

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,422 @@
...
+
+std::vector<GURL> Configurator::PingUrl() const {
+  return UpdateUrl();
+}
+
+std::string Configurator::GetProdId() const {
+  return "cobalt";
+}
+
+base::Version Configurator::GetBrowserVersion() const {
```
yell0wd0g, line 137: nit: Can we just return `COBALT_MAJOR_VERSION`?

https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/version.h;l=39;drc=1a248c495379f69283e4f19a23a2ccba141d90c3

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,422 @@
...
+  return GetDeviceProperty(kSbSystemPropertyPlatformName);
+}
+
+base::flat_map<std::string, std::string> Configurator::ExtraRequestParams()
+    const {
+  base::flat_map<std::string, std::string> params;
+  params.insert(std::make_pair("SABI", SB_SABI_JSON_ID));
+  params.insert(std::make_pair("sbversion", std::to_string(SB_API_VERSION)));
+
+  // The flag name to force an update is changed to "is_forced_update_" to cover
```
yell0wd0g: nit: Escape variables in comments either with backticks or pipe
symbols (`|`). Code will be indexed that way. Apply throughout.

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,422 @@
...
+
+  return params;
+}
+
+std::string Configurator::GetDownloadPreference() const {
+  return {};
+}
+
+scoped_refptr<update_client::NetworkFetcherFactory>
+Configurator::GetNetworkFetcherFactory() {
```
yell0wd0g, line 227: nit: Simple accessors and mutators (getters and setters) are usually
named with snake case names in Chromium, so this would be 
`get_network_fetcher_factory()`. Same in l. 239, 244 etc. This
conveys eg that there's no lock involved.


> Accessors and mutators (get and set functions) may be named like variables, in snake_case. These often correspond to actual member variables, but this is not required. For example, int count() and void set_count(int count).

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,422 @@
...
+  if (std::regex_match(updater_channel,
+                       std::regex("(2[5-9]|[3-9][0-9])lts\\d+"))) {
+    channel = "static";
+  }
+  auto it = kChannelAndSbVersionToOmahaIdMap.find(channel +
+                                                  std::to_string(sb_version));
+  if (it != kChannelAndSbVersionToOmahaIdMap.end()) {
+    return it->second;
+  }
+  LOG(INFO) << "Configurator::GetAppGuidHelper updater channel and starboard "
```
yell0wd0g, line 303: nit: Instead of quoting the function name, rather use `__func__` ?

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,422 @@
...
+            << "combination is undefined with the new Omaha configs.";
+
+  // All undefined channel requests go to prod configs except for static
+  // channel requestsf for C24 and older.
+  if (!std::regex_match(updater_channel, std::regex("2[0-4]lts\\d+")) &&
```
yell0wd0g, line 309: regex use is banned in Chromium: https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++-features.md#regex_banned

you'll have to use one of the matchers in 
https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/substring_set_matcher/ or re2

cobalt/updater/network_fetcher.cc
```diff
@@ -0,0 +1,231 @@
...
+
+#if defined(IN_MEMORY_UPDATES)
+void NetworkFetcher::DownloadToString(
+    const GURL& url,
+    std::string* dst,
+    ResponseStartedCallback response_started_callback,
+    ProgressCallback progress_callback,
+    DownloadToStringCompleteCallback download_to_string_complete_callback) {
+  LOG(INFO) << "NetworkFetcher::DownloadToString, url = " << url;
+  // TODO(b/444006168): implement this to support in-memory updates
```
yell0wd0g, line 158: nit: Leave a `NOTIMPLEMENTED()`?

cobalt/updater/network_fetcher.cc
```diff
@@ -0,0 +1,231 @@
...
+      std::move(response_started_callback)));
+  simple_url_loader_->SetOnDownloadProgressCallback(base::BindRepeating(
+      &NetworkFetcher::OnProgressCallback, base::Unretained(this),
+      std::move(progress_callback)));
+  simple_url_loader_->DownloadToFile(
+      url_loader_factory_.get(),
+      base::BindOnce(
+          [](const network::SimpleURLLoader* simple_url_loader,
+             DownloadToFileCompleteCallback download_to_file_complete_callback,
+             base::FilePath file_path) {
```
yell0wd0g: nit: is `file_path` needed?

cobalt/updater/network_fetcher.cc
```diff
@@ -0,0 +1,231 @@
...
+  resource_request->method = "GET";
+  resource_request->load_flags = net::LOAD_DISABLE_CACHE;
+  simple_url_loader_ = network::SimpleURLLoader::Create(
+      std::move(resource_request), traffic_annotation);
+  simple_url_loader_->SetRetryOptions(
+      kMaxRetriesOnNetworkChange,
+      network::SimpleURLLoader::RetryMode::RETRY_ON_NETWORK_CHANGE);
+  simple_url_loader_->SetAllowPartialResults(true);
+  simple_url_loader_->SetOnResponseStartedCallback(base::BindOnce(
+      &NetworkFetcher::OnResponseStartedCallback, base::Unretained(this),
```
yell0wd0g, line 183: nit: Use of `base::Unretained()` needs to be documented.
I see for example `::Cancel()` below `reset()`s this 
`simple_url_loader_`, does that mean we would never
get calls back after `simple_url_loader_` is destroyed?

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,133 @@
...
+  static constexpr int kMaxRetriesOnNetworkChange = 3;
+
+  scoped_refptr<base::SequencedTaskRunner> main_thread_task_runner_;
+
+  scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
+  std::unique_ptr<network::SimpleURLLoader> simple_url_loader_;
+};
+
+// Network fetcher factory.
+class NetworkFetcherFactoryCobalt
```
yell0wd0g, line 128: I don't see at first sight much value in having this 
factory class -- could we just have a `Create()` static
function in `NetworkFetcher` above?

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,133 @@
...
+
+  static constexpr int kMaxRetriesOnNetworkChange = 3;
+
+  scoped_refptr<base::SequencedTaskRunner> main_thread_task_runner_;
+
+  scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
+  std::unique_ptr<network::SimpleURLLoader> simple_url_loader_;
+};
+
+// Network fetcher factory.
```
yell0wd0g: nit: This comment doesn't quite add much.

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,133 @@
...
+
+namespace network {
+class SharedURLLoaderFactory;
+class SimpleURLLoader;
+}  // namespace network
+
+namespace cobalt {
+namespace updater {
+
+class NetworkFetcher : public update_client::NetworkFetcher {
```
yell0wd0g, line 57: nit: Please add a class comment detailing a) The reason for
this class existence b) Expected lifetime c) Threading.

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,133 @@
...
+  void OnProgressCallback(ProgressCallback response_started_callback,
+                          uint64_t current);
+
+  static constexpr int kMaxRetriesOnNetworkChange = 3;
+
+  scoped_refptr<base::SequencedTaskRunner> main_thread_task_runner_;
+
+  scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
+  std::unique_ptr<network::SimpleURLLoader> simple_url_loader_;
+};
```
yell0wd0g, line 124: nit: In general, please add ThreadCheckers as documentations
in code to all classes.

cobalt/updater/network_fetcher.cc
```diff
@@ -0,0 +1,231 @@
...
+                     simple_url_loader->GetContentSize());
+          },
+          simple_url_loader_.get(),
+          std::move(download_to_file_complete_callback)),
+      file_path);
+}
+#endif
+
+void NetworkFetcher::Cancel() {
+  LOG(INFO) << "cobalt::updater::NetworkFetcher::Cancel";
```
yell0wd0g, line 205: Overall comment: try not to leave LOG()s in almost every method:

> Remove all logging before checking in code. 

https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++.md#logging

Alternatively, use `D(V)LOG()`.

yell0wd0g (changes requested):
Took a pass at configurator.* and network_fetcher.*

Global comments: define, document and hopefully enforce
threading and lifetimes.

cobalt/updater/network_fetcher.cc
yuying-y: See https://github.com/youtube/cobalt/pull/7131#issuecomment-3349280361.

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+    // time between 1 and 24 hours.
+    kNextUpdateCheckHours = base::RandInt(1, 24);
+  } else {
+    // Update check ran at least twice. The next check will be scheduled in 24
+    // hours.
+    kNextUpdateCheckHours = 24;
+  }
+  updater_thread_->task_runner()->PostDelayedTask(
+      FROM_HERE, base::BindOnce(&UpdaterModule::Update, base::Unretained(this)),
+      base::Hours(kNextUpdateCheckHours));
```
yell0wd0g: Similar to my other comments: try to use base::TimeDelta rather
than integers and conversion.

cobalt/updater/updater_module.h
yell0wd0g: > Also, if this is converted to a base::TimeDelta, I suggest
making the const (which should be constexpr) and this member
var that type.

This still stands: l.215 of the .cc file is:
```
  LOG(INFO) << "Scheduling UpdateCheck with delay " << update_check_delay_sec_
            << " seconds";
  updater_thread_->task_runner()->PostDelayedTask(
      FROM_HERE, base::BindOnce(&UpdaterModule::Update, base::Unretained(this)),
      base::Seconds(update_check_delay_sec_));
```

So it's not necessary to make this variable and constexpr
and integer. Instead, l.97-98 of this file would read:
```
// Default delay to the first update check.
constexpr auto kDefaultUpdateCheckDelay = TimeDelta::FromSeconds(30);
```
(and then remove l.90-91 of the .cc file).

Then in this file you'll have l.181:
```
base::TimeDelta update_check_delay_ = kDefaultUpdateCheckDelaySeconds;
```

And finally l.215 of the .cc file would read:
```
  LOG(INFO) << "Scheduling UpdateCheck with delay " << update_check_delay_;
  updater_thread_->task_runner()->PostDelayedTask(
      FROM_HERE, base::BindOnce(&UpdaterModule::Update, base::Unretained(this)),
      update_check_delay_);
```

If you need the amount of seconds you can do `update_check_delay_.InSeconds();`
or `update_check_delay_.InSecondsF();`

yell0wd0g (changes requested):

cobalt/updater/configurator.cc
yuying-y: Considering the various callers and dependencies, and scope of changes, filed b/449022872 to track.

cobalt/updater/configurator.h
yuying-y: Granular separate locks are needed.

cobalt/updater/network_fetcher.cc
yuying-y: It's declared as scoped_refptr

cobalt/updater/network_fetcher.h
yuying-y: Complaint about unimplemented `NetworkFetcherFactory` if removed.

cobalt/updater/network_fetcher.cc
yuying-y: Added TODO at the top of the file

cobalt/updater/patcher.cc
yuying-y: Added TODO comment

cobalt/updater/configurator.h
```diff
@@ -0,0 +1,164 @@
...
+// It implements the update_client::Configurator interface and provides
+// Cobalt-specific configuration parameters for the update process.
+//
+// Provides Cobalt-specific configurations to the update client, such as update
+// URLs, version information, and other settings required for the update
+// process.
+// Exists for the duration of the application's runtime, destroyed when app is
+// suspended or exits.
+// Accessed by both the updater thread and the main thread. Synchronization
+// mechanisms (locks, atomics) are used to ensure thread safety.
```
yell0wd0g, line 64: Thanks for the comments on threading. If there're only two 
threads involved, do we need that many locks and atomics,
or could we just have on global one? 
Also consider adding a thread_checker for each TaskRunner
accessing this class (or leave a TODO) to do that. 
Note that in Chromium there're no "thread"s per se, but 
either TaskRunners or derivatives, see [Prefer Sequences 
to Physical Threads](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/threading_and_tasks.md#Prefer-Sequences-to-Physical-Threads), so please correct the comment.

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,147 @@
...
+class SharedURLLoaderFactory;
+class SimpleURLLoader;
+}  // namespace network
+
+namespace cobalt {
+namespace updater {
+// The NetworkFetcher class provides an implementation of the
+// update_client::NetworkFetcher interface, tailored for Cobalt's updater.
+//
+//    This class is responsible for performing network operations required by
```
yell0wd0g: nit: Remove extra indenting?

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,147 @@
...
+//    the updater, such as downloading update payloads or posting update
+//    requests. It acts as a bridge between the updater logic and the underlying
+//    network stack, encapsulating the details of network requests and
+//    responses, and providing a consistent interface for the updater.
+//
+//    A NetworkFetcher instance is created for the duration of a single network
+//    operation (e.g., a file download or a POST request). Once the operation
+//    completes or is canceled, the instance is expected to be destroyed.
+//    Ownership is managed via std::unique_ptr, ensuring proper cleanup after
+//    use.
```
yell0wd0g: Thanks!

QQ: `Ownership is managed via std::unique_ptr` <- does this
refer to the way this class is supposed to be owned and its
lifetime managed? If so, please rephrase as suggestion, since
NetworkFetcher has no control over what its callers do. E.g.
just write `This class should be fully owned.`

cobalt/updater/network_fetcher.cc
```diff
@@ -0,0 +1,236 @@
...
+}
+
+}  // namespace
+
+namespace cobalt {
+namespace updater {
+
+NetworkFetcher::NetworkFetcher(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory)
+    : url_loader_factory_(url_loader_factory) {
```
yell0wd0g: nit:
```
    : url_loader_factory_(std::move(url_loader_factory)) {
```

cobalt/updater/patcher.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "base/memory/ref_counted.h"
+#include "components/update_client/patcher.h"
+
+// TODO(b/174699165): Investigate differential updates
+
+namespace cobalt {
+namespace updater {
+
+class PatcherFactory : public update_client::PatcherFactory {
```
yell0wd0g, line 21: This is a surprising class, because it only creates
PatcherImpls (in the .cc) file), and AFAIK that class
itself is just three methods, no member variables.
In an ideal world, they would be just free functions;
however, the base classes make me think there's more
to it? Could you leave a comment on each class stating
why they are needed? 
It can be short, like:
```
// PatcherFactory is a trivial factory class, needed
// by XYZ to operate.
```
And similarly for PatcherImpl.

Same comments on `UnzipperFactory`

cobalt/updater/prefs.h
```diff
@@ -0,0 +1,30 @@
...
+#define COBALT_UPDATER_PREFS_H_
+
+#include <memory>
+
+class PrefService;
+
+namespace cobalt {
+namespace updater {
+
+std::unique_ptr<PrefService> CreatePrefService();
```
yell0wd0g, line 26: Please write a meaningful documentation line here. It
can be brief, e.g.
`// Finds all the pieces to create a fully initialized PrefService.`

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+    case ComponentState::kUpdated:
+      return kCobaltExtensionUpdaterNotificationStateUpdated;
+    case ComponentState::kUpToDate:
+      return kCobaltExtensionUpdaterNotificationStateUpToDate;
+    case ComponentState::kUpdateError:
+      return kCobaltExtensionUpdaterNotificationStateUpdateFailed;
+    default:
+      return kCobaltExtensionUpdaterNotificationStateNone;
+  }
+}
```
yell0wd0g: nit: Don't have a default, and instead enumerate all ComponentStates?
```
    case ComponentState::kBla:
    case ComponentState::kFoo:
      return kCobaltExtensionUpdaterNotificationStateNone;
```
where kBla, kFoo are the ones not handled already. This makes
sure that if someone was to add a ComponentState in the future,
the code here would need to be updated.

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+  updater_configurator_->SetUpdaterStatus(status);
+  if (status != "Downloading update") {
+    LOG(INFO) << "Updater status is " << status;
+  }
+}
+
+UpdaterModule::UpdaterModule(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    base::TimeDelta update_check_delay)
+    : url_loader_factory_(url_loader_factory),
```
yell0wd0g: nit:
`    : url_loader_factory_(std::move(url_loader_factory)),`

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+      update_check_delay_(update_check_delay) {
+  LOG(INFO) << "UpdaterModule::UpdaterModule";
+  updater_thread_.reset(new base::Thread("Updater"));
```
yell0wd0g, line 147: nit: Move this to the ctor initialization list. Probably the
same for l.151.

cobalt/updater/updater_module.h
```diff
@@ -0,0 +1,205 @@
...
+  std::unique_ptr<base::Thread> updater_thread_;
+  scoped_refptr<update_client::UpdateClient> update_client_;
+  std::unique_ptr<Observer> updater_observer_;
+  scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory_;
+  scoped_refptr<Configurator> updater_configurator_;
+  int update_check_count_ = 0;
+  bool is_updater_running_;
+  base::TimeDelta update_check_delay_ = kDefaultUpdateCheckDelaySeconds;
+
+  THREAD_CHECKER(thread_checker_);
```
yell0wd0g: If this is just used for places where code is on `updater_thread_`,
you can just use the latter instead, e.g. 
instead of:
```
```
you should have:
```
  CHECK(updater_thread_->GetDefaultTaskRunner()->BelongsToCurrentThread());
```

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+  std::string status;
+  if (update_client_->GetCrxUpdateState(id, &crx_update_item_)) {
+    auto status_iterator =
+        component_to_updater_status_map.find(crx_update_item_.state);
+    if (status_iterator == component_to_updater_status_map.end()) {
+      status = "Status is unknown.";
+    } else if (crx_update_item_.state == ComponentState::kUpToDate &&
+               updater_configurator_->GetPreviousUpdaterStatus().compare(
+                   updater_status_string_map.at(UpdaterStatus::kUpdated)) ==
+                   0) {
```
yell0wd0g: nit: Compare strings directly.

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+    } else {
+      status = std::string(
+          updater_status_string_map.find(status_iterator->second)->second);
+    }
+    if (crx_update_item_.state == ComponentState::kUpdateError) {
+      // `QUICK_ROLL_FORWARD` update, adjust the message to "Updated locally,
+      // pending restart".
+      if (crx_update_item_.error_code ==
+          static_cast<int>(UpdateCheckError::QUICK_ROLL_FORWARD)) {
+        status = std::string(
```
yell0wd0g: nit: No need for `std::string()` since `updater_status_string_map`
already contains strings, right? Similarly in l.108 and elsewhere
where this pattern is used.

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+              crx_update_item_.state),
+          GetCurrentEvergreenVersion().c_str());
+    }
+  } else {
+    status = "No status available";
+  }
+  updater_configurator_->SetUpdaterStatus(status);
+  if (status != "Downloading update") {
+    LOG(INFO) << "Updater status is " << status;
+  }
```
yell0wd0g: nit: Trick
```
  LOG_IF(INFO, status != "Downloading update") << "Updater status is " << status;
```

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,560 @@
...
+  LOG(INFO) << "UpdaterModule::~UpdaterModule";
+  if (is_updater_running_) {
+    is_updater_running_ = false;
+    {
+      base::ScopedBlockingCall scoped_blocking_call(
+          FROM_HERE, base::BlockingType::MAY_BLOCK);
+      updater_thread_->task_runner()->PostTask(
+          FROM_HERE,
+          base::BindOnce(&UpdaterModule::Finalize, base::Unretained(this)));
+    }
```
yell0wd0g, line 171: Blocking a destruction of a thread is uncommon. @y4vor for comments.
Does this only happen at the end of the browser lifetime? If so
we could have a different pattern (likely involving DeleteSoon).
Can you leave a `// TODO(b/123): Investigate UpdaterModule destruction 
sequence`? Thanks a bunch.

cobalt/updater/util.cc
```diff
@@ -0,0 +1,310 @@
...
+  std::string compressed_library_path = base::StrCat(
+      {installation_path, kSbFileSepString, kCompressedLibraryPath});
+  std::string uncompressed_library_path = base::StrCat(
+      {installation_path, kSbFileSepString, kUncompressedLibraryPath});
+  struct stat file_info;
+  if (stat(compressed_library_path.c_str(), &file_info) == 0) {
+    return "Compressed";
+  } else if (stat(uncompressed_library_path.c_str(), &file_info) == 0) {
+    return "Uncompressed";
+  } else {
```
yell0wd0g: nit: No need for "else" after a a return body, e.g. ISO
```
  if (a) {
    return;
  } else if (b) {
    return
  } else {
    // bla
  }
```
it should read
```
  if (a) {
    return;
  }
  if (b) {
    return
  }
  // bla
```

cobalt/updater/util.cc
```diff
@@ -0,0 +1,310 @@
...
+  base::File source_file(filepath,
+                         base::File::FLAG_OPEN | base::File::FLAG_READ);
+  if (!source_file.IsValid()) {
+    LOG(ERROR) << "GetLibrarySha256(): Unable to open source file: "
+               << filepath.value();
+    return "";
+  }
+
+  const size_t kBufferSize = 32768;
+  std::vector<char> buffer(kBufferSize);
```
yell0wd0g, line 284: This should be `std::vector<uint8_t>` which is what `SecureHash` 
wants for e.g. `Update()` in l.301, but `ReadAtCurrentPos()` in
l.289 wants a `char`. Can you leave a TODO to fix this discrepancy
(likely by using `ReadAtCurrentPosAndCheck()` and `base::span()`) ?
Thanks

yell0wd0g (changes requested):
Almost there! 

Appreciate the patience.

cobalt/updater/configurator.h
yuying-y: We need individual locks because each guarded variable serves a specific functionality and is set by a specific function. Using a global lock hinders efficiency. TODO added for thread_checker. The linked [documentation](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/threading_and_tasks.md#threads) also uses "main thread"/"IO thread", so I assume using the word "thread" in comment is okay to refer to the general idea.

cobalt/updater/network_fetcher.h
yuying-y: The original comment was not accurate. Updated.

cobalt/updater/patcher.h
yuying-y: It's a skeleton for potential investigation into differential updates in the future. There's a comment on top of this file `TODO(b/174699165): Investigate differential updates`. Added another line of comment to clarify.

cobalt/updater/updater_module.cc
yuying-y: I prefer to keep it this way, so that the log prints before the updater_thread_ is initialized. is_updater_running_ needs to be set true after the updater_thread_ starts.

cobalt/updater/updater_module.h
```diff
@@ -0,0 +1,205 @@
...
+    {UpdaterStatus::kUpdated, "Update installed, pending restart"},
+    {UpdaterStatus::kRolledForward, "Updated locally, pending restart"},
+    {UpdaterStatus::kUpToDate, "App is up to date"},
+    {UpdaterStatus::kUpdateError, "Failed to update"},
+    {UpdaterStatus::kUninstalled, "Update uninstalled"},
+    {UpdaterStatus::kRun, "Transitioning..."},
+};
+
+// Default number of seconds to delay the first update check.
+extern const uint64_t kDefaultUpdateCheckDelaySeconds;
```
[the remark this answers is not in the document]
yuying-y: Updated the definition in updater_module.cc. The extern variable will be used by Cobalt shell in follow-up PRs.

cobalt/updater/configurator.cc
```diff
@@ -0,0 +1,415 @@
...
+std::vector<GURL> Configurator::PingUrl() const {
+  return UpdateUrl();
+}
+
+std::string Configurator::GetProdId() const {
+  return "cobalt";
+}
+
+base::Version Configurator::GetBrowserVersion() const {
+  return base::Version(COBALT_MAJOR_VERSION);
```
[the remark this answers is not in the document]
yuying-y: The server expects the major version only.

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,144 @@
...
+#endif
+  void Cancel() override;
+
+ private:
+  // TODO(b/449228491): add thread checker
+  // Empty struct to ensure the caller of |HandleError()| knows that |this|
+  // may have been destroyed and handles it appropriately.
+  struct ReturnWrapper {
+    void InvalidateThis() {}
+  };
```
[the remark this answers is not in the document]
yuying-y: Explained by the comment.

cobalt/updater/network_fetcher.h
```diff
@@ -0,0 +1,144 @@
...
+// This class is responsible for performing network operations required by
+// the updater, such as downloading update payloads or posting update
+// requests. It acts as a bridge between the updater logic and the underlying
+// network stack, encapsulating the details of network requests and
+// responses, and providing a consistent interface for the updater.
+//
+// A NetworkFetcher instance is created when the updater configurator is
+// created, and its lifetime is managed by the updater configurator.
+//
+// NetworkFetcher is used on a single thread - the updater thread when a
```
yell0wd0g, line 55: A note for posterity: Chrome prefers (Sequenced)TaskRunner as opposed
to Threads. The latter are only needed in some limited and very 
special cases, see https://chromium.googlesource.com/chromium/src/+/lkgr/docs/threading_and_tasks.md#prefer-sequences-to-physical-threads.

So in general, try to correct language to TaskRunner, and in particular
to SequencedTaskRunner (which guarantees Tasks being run in Posting
order, and it's the most common case).

cobalt/updater/network_fetcher.h
yell0wd0g: Can we add a comment here, e.g.
`// This class is needed because XYZ class (or file) needs a factory.`

cobalt/updater/prefs.cc
```diff
@@ -0,0 +1,68 @@
...
+
+  const CobaltExtensionInstallationManagerApi* installation_api =
+      static_cast<const CobaltExtensionInstallationManagerApi*>(
+          SbSystemGetExtension(kCobaltExtensionInstallationManagerName));
+  if (!installation_api) {
+    LOG(ERROR) << "Failed to get installation manager api.";
+    return nullptr;
+  }
+  char app_key[IM_EXT_MAX_APP_KEY_LENGTH];
+  if (installation_api->GetAppKey(app_key, IM_EXT_MAX_APP_KEY_LENGTH) ==
```
yell0wd0g: nit: what about
```  
  std::string app_key;
  app.key.reserve(IM_EXT_MAX_APP_KEY_LENGTH);
  if (installation_api->GetAppKey(app_key.data(), app_key.capacity()) ==
```
? 
Since anyway `app_key` is treated as a std::string in l.53 and l.56.

cobalt/updater/unzipper.cc
```diff
@@ -0,0 +1,79 @@
...
+
+  void Unzip(const base::FilePath& zip_path,
+             const base::FilePath& output_path,
+             UnzipCompleteCallback callback) override {
+    int64_t time_before_unzip =
+        (base::TimeTicks::Now() - base::TimeTicks()).InMicroseconds();
+    std::move(callback).Run(zip::Unzip(zip_path, output_path));
+    int64_t time_unzip_took_usec =
+        (base::TimeTicks::Now() - base::TimeTicks()).InMicroseconds() -
+        time_before_unzip;
```
yell0wd0g: nit: Can't you just do
```
    auto time_before_unzip = base::TimeTicks::Now();
    std::move(callback).Run(zip::Unzip(zip_path, output_path));
    auto time_unzip_took_usec = 
        (base::TimeTicks::Now() - time_before_unzip).InMilliseconds();
```
? 
(Note the direct conversion to milliseconds, so it's not needed in
l.45-46.
Same in l.53-58 and l.62-63.

cobalt/updater/updater_module.cc
```diff
@@ -0,0 +1,561 @@
...
+  }
+}
+
+}  // namespace
+
+namespace cobalt {
+namespace updater {
+
+// The delay in seconds before the first update check.
+const uint64_t kDefaultUpdateCheckDelaySeconds = 30;
```
yell0wd0g: nit: This needs to be just `base::TimeDelta` -- can you leave a TODO?
Maybe just create a bug to "use base::Time all around cobalt/updater",
the title should be enough, no need for a bug description body.

cobalt/updater/updater_module.h
yell0wd0g: I'm not sure you saw my comment, reopening.

Actually, I'm seeing a number of PostTask with Unretained
in the .cc file. Could you leave a TODO at this class to
revisit its threading model please? Thanks.

cobalt/updater/network_fetcher.h
yuying-y: Added a TODO b/453693299 to use sequence to replace base::Thread in updater_module.cc. For now, because it's using a thread, I'll keep the comment to be consistent with the code.

cobalt/updater/unzipper.cc
yuying-y: Yeah, this is a weird legacy of the previous codebase. Modified.

cobalt/updater/updater_module.cc
yuying-y: Done.

yuying-y:
> @[yell0wd0g](https://github.com/yell0wd0g) yell0wd0g [3 days ago](https://github.com/youtube/cobalt/pull/7177#discussion_r2441563535)
> I'm not sure you saw my comment, reopening.
> 
> Actually, I'm seeing a number of PostTask with Unretained
> in the .cc file. Could you leave a TODO at this class to
> revisit its threading model please? Thanks.

Yes, I filed b/449223391 on Oct 3, and added the TODO to network_fetcher.cc. Now adding it to updater_module.h as well.

cobalt/updater/updater_module.h
yuying-y: Yes, I filed b/449223391 on Oct 3, and added the TODO to network_fetcher.cc. Now adding it to updater_module.h as well.

Also, the comment about thread_checker_ was resolved in previous commit in updater_module.cc, but I forgot to remove it from updater_module.h. Now it's removed.

PR #4322, opened 2024-10-28
Fix starboard toolchain when multiple are used

andrewsavage1:
b/376237145

PR #7179, opened 2025-09-16
starboard: Do not set RPATH in "starboard" GN config

rakuco:
On tvOS, the syntax is not recognized by LLD, leading to errors like

    ld64.lld: error: unknown argument '-rpath=/foo/bar'

when linking executables both both host and target, and everything works
fine without setting it.

This seems to be the case on other platforms as well, so revert part of
#4322 and stop setting RPATH on all platforms.

Bug: 376237145
Bug: 432642933

### DIFF
--- a/.github/workflows/linearize_main.yaml
+++ b/.github/workflows/linearize_main.yaml
@@ -56,8 +56,14 @@ jobs:
             --start-commit-ref="$(git rev-parse --verify post-chrobalt-tag)" --end-commit-ref="$(git rev-parse --verify origin/automated/linear_main)" \
             --output-file=${GITHUB_WORKSPACE}/automated_commits_m114.json
           git add automated_commits_m114.json
-          git commit -m "Linearization refresh on $(date +'%Y-%m-%d')."
-          export CP_COMMIT="$(git rev-parse --verify HEAD)"
+          git stash
           git checkout experimental/rebase_tools
-          git cherry-pick --strategy=recursive -X theirs $CP_COMMIT
-          git push --force origin experimental/rebase_tools:experimental/rebase_tools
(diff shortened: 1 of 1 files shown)


rakuco:

starboard/build/config/BUILD.gn
```diff
@@ -167,9 +167,15 @@ config("starboard") {
...
   }
 
-  sb_out_dir =
-      get_label_info("//starboard($starboard_toolchain)", "root_out_dir")
-  ldflags = [ "-Wl,-rpath=" + rebase_path("$sb_out_dir") ]
+  # b/432642933: The rpath option causes lld to fail when targeting tvOS
+  # because -rpath=/path/to/something is not a valid linker option (and
+  # everything works fine without it anyway).
+  # Check for |target_os| because this also affects the host (macOS) side.
+  if (target_os != "ios") {
```
andrewsavage1: I think you can just delete this section for all platforms—could you try that out and see if it passes CI?

rakuco: Done with the second commit in this PR. The list of failures looks a bit arbitrary and hard to pass; does any of the failures look real to you?

andrewsavage1: The main error just looks like a node failed—I'll rerun the failed tests and we can see

andrewsavage1: Looks likely that it was an infra flake

rakuco: Everything passed this time, so I've retitled the PR and edited the commit message to indicate that this part of the config is being removed on all platforms.

andrewsavage1 (approved):

PR #7180, opened 2025-09-16
starboard: Move deeply-nested callback definitions into flattened namespace

kjyoun:
[go/cobalt-flatten-starboard-namespace](http://go/cobalt-flatten-starboard-namespace)

This change refactors the definitions for common player filter callbacks like `ErrorCB`, `EndedCB`, and `PrerolledCB`.

- The definitions have been moved from the nested namespace `starboard::shared::starboard::player::filter` to the top-level `starboard` namespace.
- `typedef` has been replaced with the modern `using` syntax.

This removes the need for many local using aliases and typedefs in the decoder and renderer headers, making the code cleaner and more consistent.

Bug: 441955900


sideb0ard (approved):
lgtm!

PR #7181, opened 2025-09-16
POSIX: Register getcwd as Starboard Symbol

johnedocampo:
This PR registers the POSIX `getcwd()` to Starboard as a valid symbol. A wrapper isn't needed for this function as there are no special inputs or outputs that need to be configured.

The tests for this symbol were generated by Gemini, and then edited for correctness.

Bug: 412650854
Bug: 445714820

### DIFF
--- a/starboard/nplb/posix_compliance/posix_getcwd_test.cc
+++ b/starboard/nplb/posix_compliance/posix_getcwd_test.cc
@@ -1,0 +1,80 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/*
+  The following error conditions are not tested due to the difficulty of
+  reliably triggering them in a portable unit testing environment:
+  - EACCES: Read or search permission was denied for a component of the
+            pathname.
+  - ENOMEM: There is insufficient memory to satisfy the request when dynamically
+            allocating a buffer.
+
+  The non-standard `getcwd(NULL, 0)` behavior, where the implementation is
+  expected to `malloc` a buffer, is not supported and therefore not tested.
+*/
+
+#include <errno.h>
+#include <limits.h>
+#include <string.h>
+#include <vector>
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace nplb {
+namespace {
+
+TEST(PosixGetcwdTest, SucceedsWithSufficientBuffer) {
+  char path_buffer[PATH_MAX];
+  errno = 0;
+
+  char* result = getcwd(path_buffer, sizeof(path_buffer));
+  ASSERT_NE(nullptr, result)
+      << "getcwd failed with unexpected error: " << strerror(errno);
+  EXPECT_EQ(path_buffer, result)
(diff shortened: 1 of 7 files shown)


starboard/nplb/posix_compliance/posix_getcwd_test.cc
```diff
@@ -0,0 +1,81 @@
...
+  char* small_buffer = new char[path_len];
+  errno = 0;
+  char* result = getcwd(small_buffer, path_len);
+
+  EXPECT_EQ(nullptr, result)
+      << "getcwd should have returned nullptr for an undersized buffer.";
+  EXPECT_EQ(ERANGE, errno) << "errno should be ERANGE, but was: " << errno
+                           << " (" << strerror(errno) << ")";
+
+  delete[] small_buffer;
```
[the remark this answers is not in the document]
johnedocampo: Done; instead of using a vector however, we just statically allocate another buffer that is one character too short for the current path.

starboard/nplb/posix_compliance/posix_getcwd_test.cc
```diff
@@ -0,0 +1,79 @@
...
+TEST(PosixGetcwdTest, FailsWithErrangeForSmallBuffer) {
+  char full_path[PATH_MAX];
+  ASSERT_NE(nullptr, getcwd(full_path, sizeof(full_path)));
+  size_t path_len = strlen(full_path);
+
+  // Attempt to get the path with a buffer that is one byte too small to hold
+  // the path and the null terminator. A size of |path_len| is too small.
+  char small_buffer[path_len];
+  errno = 0;
+  char* result = getcwd(small_buffer, path_len);
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/nplb/posix_compliance/posix_getcwd_test.cc
```diff
@@ -0,0 +1,80 @@
...
+#include <limits.h>
+#include <string.h>
+#include <vector>
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace nplb {
+namespace {
+
+TEST(PosixGetcwdTest, SucceedsWithSufficientBuffer) {
```
andrewsavage1, line 37: Is there any way to test that we're getting a reasonable value more than checking than checking that it starts with `/`?

starboard/nplb/posix_compliance/posix_getcwd_test.cc
```diff
@@ -0,0 +1,80 @@
...
+  errno = 0;
+
+  char* result = getcwd(path_buffer, sizeof(path_buffer));
+  ASSERT_NE(nullptr, result)
+      << "getcwd failed with unexpected error: " << strerror(errno);
+  EXPECT_EQ(path_buffer, result)
+      << "getcwd should return a pointer to the provided buffer on success.";
+
+  EXPECT_EQ('/', result[0]) << "Path should be absolute.";
+  EXPECT_GT(strlen(result), 0u) << "Path should not be empty.";
```
andrewsavage1: Seems redundant with the check above—seems like we should switch their order and change the EXPECT_GT to an ASSERT_GT to avoid a segfault

starboard/nplb/posix_compliance/posix_getcwd_test.cc
johnedocampo: I'm not quite sure if we have any strong guarantees about what the current directory is when this test is ran. I think the only things we can assume are that a string is returned by `getcwd`, and that the returned string is an absolute path.

The best alternative I thought of would be to use `chdir`, but as far as I can tell we do not support this function. Should this be a function we support?

starboard/nplb/posix_compliance/posix_getcwd_test.cc
johnedocampo: Done.

starboard/nplb/posix_compliance/posix_getcwd_test.cc
andrewsavage1: No, I think it makes sense to keep the tests as they are

PR #4852, opened 2025-02-05
Set is_cobalt_hermetic_build false globally and to true in the toolchain args

niranjanyardi:
b/393160172

Make is_cobalt_hermetic_build a global variable so that it can be easily replaced by is_starboard and we can avoid some unnecessary includes

PR #7187, opened 2025-09-16
[M138 Rebase] Fix gn.py for linux-x64x11 config

sideb0ard:
Bug: 418842688

Gn modifications to fix cobalt/build/gn.py

$ cobalt/build/gn.py -p linux-x64x11 -c devel
gn gen out/linux-x64x11_devel --check
Done. Made 34069 targets from 4145 files in 40055ms

### DIFF
--- a/cobalt/build/gn.py
+++ b/cobalt/build/gn.py
@@ -71,8 +71,11 @@ def write_build_args(build_args_path, platform_args_path, build_type, use_rbe):
   """ Write args file, modifying settings for config"""
   gen_comment = '# Set by gn.py'
   with open(build_args_path, 'w', encoding='utf-8') as f:
+    f.write(f'use_siso = false {gen_comment}\n')
     if use_rbe:
       f.write(f'use_remoteexec = true {gen_comment}\n')
+    f.write(
+        f'rbe_cfg_dir = rebase_path("//cobalt/reclient_cfgs") {gen_comment}\n')
     f.write(f'build_type = "{build_type}" {gen_comment}\n')
     for key, value in _BUILD_TYPES[build_type].items():
       f.write(f'{key} = {value} {gen_comment}\n')


build/toolchain/android/BUILD.gn
```diff
@@ -61,7 +61,7 @@ android_clang_toolchain("android_clang_arm") {
     toolchain_args = {
       current_cpu = "arm"
       starboard_toolchain = "//build/toolchain/android:android_clang_arm"
-      target_platform = "android-arm"
+      starboard_target_platform = "android-arm"
```
kaidokert: I believe theres a PR on main where this was already changed. Perhaps cherry-pick that ?

cobalt/build/configs/common.gn
```diff
@@ -37,9 +37,6 @@ enable_cardboard = false
 enable_openxr = false
 enable_vr = false
 
-# Disable screen/window/tab capture (Note: live capture, not screenshots).
-enable_screen_capture = false
```
kaidokert: Is this temporary ? We'd still want to disable it, right ? 

If it's temporary, i would probably comment it out with a TODO bug to follow up, rather than delete

cobalt/build/configs/linux_common.gn
```diff
@@ -87,11 +86,6 @@ use_pangocairo = false
 # We are not using tflite in cobalt currently.
 build_with_tflite_lib = false
 
-# Overriding from third_party/blink/renderer/modules/ml/webnn/features.gni:10
-# webnn is a machine learning related module: https://chromestatus.com/feature/5176273954144256
-# We are not using webnn in cobalt currently.
-build_webnn_with_xnnpack = false
```
kaidokert, line 93: Same q : as above - is this intended to be temporary removal ? Or upstream doesn't have it anymore at all ?

cobalt/build/gn.py
```diff
@@ -71,11 +71,8 @@ def write_build_args(build_args_path, platform_args_path, build_type, use_rbe):
   """ Write args file, modifying settings for config"""
   gen_comment = '# Set by gn.py'
   with open(build_args_path, 'w', encoding='utf-8') as f:
-    f.write(f'use_siso = false {gen_comment}\n')
     if use_rbe:
       f.write(f'use_remoteexec = true {gen_comment}\n')
-    f.write(
```
kaidokert: I'm pretty sure we need those still. We can't do the Siso switch all at once - please check the m138-lightweight branch, that still builds with the same settings. Also `use_rbe` should already gate this code actually

starboard/android/shared/toolchain/toolchain.gni
```diff
@@ -18,7 +18,6 @@ declare_args() {
   android_sdk_path = getenv("ANDROID_HOME")
   android_ndk_path = ""
   sb_android_ndk_version = "25.2.9519653"
-  android_ndk_api_level = 24
```
kaidokert, line 21: We probably still need to keep the `android_ndk_api_level` somewhere. I'd comment this out and create a bug to clarify where and how do we set the API level. @ckim1346 worked on this in m114.

kaidokert (commented):
I'd almost want to split out the `//nogncheck` into separate PR entirely - only those changes and nothing else.

We need to figure out a better way to deal with this without having to modify upstream code all over the place. Maybe not force `gn --check` in every build config, or perhaps there's something in gn code that we could do.

Splitting _only those changes_ into separate PR will make it easier to revert once we find a better solution, and copy to other branches as needed.

components/cdm/renderer/BUILD.gn
```diff
@@ -11,6 +11,7 @@ static_library("renderer") {
     "//android_webview/*",
     "//chrome/*",
     "//chromecast/*",
+    "//cobalt/*",
```
andrewsavage1: Should be added in an `if (is_cobalt)` block

andrewsavage1:
Took a look at the `gn check` issues—looks like it's an upstream misconfiguration when `build_with_tflite_lib = false` for some and `use_dawn = false` for the rest.

kaidokert:
> Took a look at the `gn check` issues—looks like it's an upstream misconfiguration when `build_with_tflite_lib = false` for some and `use_dawn = false` for the rest.

Good opportunity to send a patch upstream maybe ? Unless it's fixed in Chromium trunk

build/toolchain/android/BUILD.gn
sideb0ard: done.

andrewsavage1:
> > Took a look at the `gn check` issues—looks like it's an upstream misconfiguration when `build_with_tflite_lib = false` for some and `use_dawn = false` for the rest.
> 
> Good opportunity to send a patch upstream maybe ? Unless it's fixed in Chromium trunk

Not fixed in Chromium's trunk, and yes it's a good patch to upstream. The patch would probably just be adding `// nogncheck` as Thor has already done (but let's just do it in a separate PR that we can clearly upstream)

buildtools/third_party/libc++abi/BUILD.gn
```diff
@@ -6,10 +6,6 @@ import("//build/config/android/config.gni")
 import("//build/config/c++/c++.gni")
 import("//build/config/unwind.gni")
 
-if (is_cobalt) {
-  import("//cobalt/build/configs/modular_variables.gni")
```
kaidokert: @andrewsavage1 or @niranjanyardi - i think we'll still need those, no ?

andrewsavage1: This file doesn't exist anymore, it was removed in https://github.com/youtube/cobalt/pull/4852/. Seems to me that should be cherry picked correctly instead of being changed here

cobalt/build/configs/common.gn
sideb0ard: ah, it was moved to [content/public/common/features.gni](https://source.chromium.org/chromium/chromium/src/+/b6bb66af8daac6c1e5de3c8ba26e6b0bf1bd177f:content/public/common/features.gni;dlc=b027f4707eee557425ce72489365b8b5697c5103) but it is not in a declare_args block. I see that mcasas added a change upstream to put it in a declare_args block - but that does not seem to have made it to the new location https://chromium-review.googlesource.com/c/chromium/src/+/6508106 - I've added back the variable in our config and put the content/public/common/features.gni declaration in a declare_args block

cobalt/build/configs/linux_common.gn
sideb0ard: upstream removed it here - https://chromium-review.googlesource.com/c/chromium/src/+/5532639

starboard/android/shared/toolchain/toolchain.gni
sideb0ard: It's now set in build/config/android/config.gni (see https://chromium-review.googlesource.com/c/chromium/src/+/6547915)

components/cdm/renderer/BUILD.gn
sideb0ard: done

buildtools/third_party/libc++abi/BUILD.gn
sideb0ard: Yeah, seems the cherry-pick didn't catch it in these two files - the original PR is here https://github.com/youtube/cobalt/pull/4852/files and the cherrypick here - https://github.com/sideb0ard/cobalt/commit/d22366171eb6299109cf13f411e747615165bb94 - original removes the import from 13 files and the CP removes it from 11 files. Both those files changed massively due to refactoring. Partition Allocator moved location, and the BUILD.gn file went from 500 lines to 24, which is our current theory why they were not applied. Brian is going to go through the commits and compare LOC and files changed, so see if there are any other cases like this.

cobalt/build/gn.py
sideb0ard: I go these recommendations from Brian actually, after he fought with rbe for awhile.

buildtools/third_party/libc++abi/BUILD.gn
andrewsavage1: So what is the plan for changes like this where the cherry pick was incorrect? Do we fix the cherry pick or do we fix it after as is being done here?

starboard/android/shared/toolchain/toolchain.gni
andrewsavage1: Don't we still need to set it to 24 somewhere?

sideb0ard: yeah, we set 'default_min_sdk_version = 24' for Cobalt  here - https://github.com/youtube/cobalt/blob/feature/rebase_m138/build/config/android/config.gni#L77 - which later is used as the value for android_ndk_api_level here - https://github.com/youtube/cobalt/blob/feature/rebase_m138/build/config/android/config.gni#L204

andrewsavage1: Great!

buildtools/third_party/libc++abi/BUILD.gn
sideb0ard: discussed offline - will go back and fix in the original patch and cherrypick. We'll leave this in for the moment and fix patches later.

cobalt/build/gn.py
sideb0ard: removed modifications.

buildtools/third_party/libc++abi/BUILD.gn
andrewsavage1: But we should split these changes out into their own PR to easily track them

sideb0ard: https://github.com/youtube/cobalt/pull/7231

kaidokert:
>  (but let's just do it in a separate PR that we can clearly upstream)

At some point i suggested a `[upstreamable]` tag into our on repo for things that should be usptreamed.

PR #7188, opened 2025-09-16
POSIX: Register realpath as Starboard Symbol

johnedocampo:
This PR registers the POSIX symbol `realpath` to Starboard. No wrapper needed as there are no special inputs or outputs that need to be translated.

Tests for this symbol were generated by Gemini, and then edited for correctness.

Bug: 416342476
Bug: 445717512

### DIFF
--- a/starboard/nplb/posix_compliance/posix_realpath_test.cc
+++ b/starboard/nplb/posix_compliance/posix_realpath_test.cc
@@ -1,0 +1,244 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


starboard/nplb/posix_compliance/posix_realpath_test.cc
```diff
@@ -0,0 +1,240 @@
...
+    ASSERT_NE(fd, -1);
+    ASSERT_EQ(close(fd), 0);
+
+    ASSERT_EQ(
+        symlink(file1_path.c_str(), (test_dir_ + "/symlink_to_file1").c_str()),
+        0);
+    ASSERT_EQ(symlink(subdir1_path.c_str(),
+                      (test_dir_ + "/symlink_to_subdir1").c_str()),
+              0);
+    ASSERT_EQ(
```
andrewsavage1: We should clean this up by moving single-use files like this one into the tests using them instead of a general SetUp step. The only ones that look reused are file1 and subdir1/file2

andrewsavage1 (approved):

starboard/nplb/posix_compliance/posix_realpath_test.cc
johnedocampo: Done. Added a comment as well explaining that the single-use files are initialized in the tests themselves.

PR #7191, opened 2025-09-16
android: Capture top 4 Java stack frames on exception

zhongqiliang:
The previous FindFirstJavaFileAndLine only extracted the first
file and line number from a Java exception, which is often
insufficient for debugging.

This updates the regex to also capture the method name and
iterates to find the top 4 frames of the stack trace.
The resulting string (e.g., method1@file1:10&method2@file2:20)
will provide much richer context for crash analysis.

Bug: 430385989
Change-Id: Ifd005980c49de10e5a257d10e8169026c88449c6

### DIFF
--- a/base/android/jni_android.cc
+++ b/base/android/jni_android.cc
@@ -350,7 +350,7 @@ void CheckException(JNIEnv* env) {
 #if BUILDFLAG(IS_COBALT)
       std::string exception_info = GetJavaExceptionInfo(env, java_throwable);
       base::android::SetJavaException(exception_info.c_str());
-      exception_token = FindFirstJavaFileAndLine(exception_info);
+      exception_token = FindTopJavaMethodsAndFiles(exception_info, 4);
 #else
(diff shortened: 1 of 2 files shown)


base/android/jni_android.cc
```diff
@@ -376,31 +376,32 @@ std::string GetJavaExceptionInfo(JNIEnv* env, jthrowable java_throwable) {
...
-
-    // smatch object will store the results of the search.
-    std::smatch match;
-
-    // Search the input string for the first occurrence of the pattern.
-    if (std::regex_search(stack_trace, match, pattern)) {
-        // The full match is match[0] (e.g., "(CobaltActivity.java:219)").
-        // The first captured group is match[1] (e.g., "CobaltActivity.java:219").
-        // We return the content of the first captured group.
-        return match[1].str();
```
johnxwork, line 399: No action needed any more. I suspect it could have been `match[0]` here before

johnxwork (approved):

PR #7206, opened 2025-09-17
Remove LD_LIBRARY_PATH from CI

andrewsavage1:
Bug: 432642933

### DIFF
--- a/.github/actions/on_host_tests/action.yaml
+++ b/.github/actions/on_host_tests/action.yaml
@@ -34,8 +34,6 @@ runs:
         env
 
         out_dir="${GITHUB_WORKSPACE}/unit_test/out/${{ matrix.platform}}_${{ matrix.config }}"
-        # TODO: LD_LIBRARY_PATH should not include the starboard subdirectory.
-        LD_LIBRARY_PATH="${out_dir}/starboard"
         LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${out_dir}"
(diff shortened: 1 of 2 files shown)


rakuco (commented):
One thing I was wondering is whether libstarboard.so.17 needs to be in the `/starboard` directory for the monolithic Linux build in the first place, which then means wondering if it makes sense to have a `//starboard/linux/x64x11/toolchain:starboard` for this while having `//cobalt/build/toolchain/linux:clang_x64` build the rest.

If both toolchains are really necessary, wouldn't it be possible to just copy it to the top-level output directory? These are all Linux files targeting the same architecture.

andrewsavage1:
> One thing I was wondering is whether libstarboard.so.17 needs to be in the `/starboard` directory for the monolithic Linux build in the first place, which then means wondering if it makes sense to have a `//starboard/linux/x64x11/toolchain:starboard` for this while having `//cobalt/build/toolchain/linux:clang_x64` build the rest.
> 
> If both toolchains are really necessary, wouldn't it be possible to just copy it to the top-level output directory? These are all Linux files targeting the same architecture.

Yeah I was thinking the same thing and tried to make `build_with_separate_cobalt_toolchain = false` for linux-x64x11, but it results in duplicate symbols. I didn't look further than that, but if you'd like to that could be helpful!

And yes, we can and probably should copy libstarboard.so.17 to the top-level out dir if we need separate toolchains

PR #7207, opened 2025-09-17
[Rebase M138] Add 'nogncheck' to address include issues

sideb0ard:
Bug: 418842688

Add nogncheck to fix cobalt/build/gn.py
These are due to upstream misconfiguration when 'build_with_tflite_lib = false' and 'use_dawn = false'

### DIFF
--- a/components/translate/core/language_detection/language_detection_model.cc
+++ b/components/translate/core/language_detection/language_detection_model.cc
@@ -12,8 +12,8 @@
 #include "base/trace_event/trace_event.h"
 #include "build/build_config.h"
 #include "components/language/core/common/language_util.h"
-#include "components/language_detection/core/constants.h"
-#include "components/language_detection/core/language_detection_model.h"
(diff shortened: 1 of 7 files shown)


components/translate/core/language_detection/language_detection_model.cc
```diff
@@ -12,8 +12,9 @@
 #include "base/trace_event/trace_event.h"
 #include "build/build_config.h"
 #include "components/language/core/common/language_util.h"
-#include "components/language_detection/core/constants.h"
-#include "components/language_detection/core/language_detection_model.h"
+// TODO(cobalt, b/377295011): remove the nogncheck annotation.
```
andrewsavage1: I think you should just remove to TODOs and we should upstream this change

sideb0ard: done.

andrewsavage1 (approved):

PR #7214, opened 2025-09-18
Fix the touch panel crash

a8901147:
A crash occurs during vertical scrolling. The crash was caused by the scroll engine receiving a null pointer.

The `FindFirstElementWithScrollType` function in `topmost_event_target.cc` was returning a scrollable element that did not have an associated `UiNavItem` for vertical scroll events. This led to a null pointer being passed to the `scroll_engine`, causing the application to crash. 

This change ensure scrollable element has a valid `UiNavItem` before calling scroll_engine

Bug: 426634950


lamdbui:
@andrewsavage1 , this seems like a reasonable change for more robust handling of a NULL ptr situation. Any thoughts on this?

cobalt/layout/topmost_event_target.cc
```diff
@@ -800,4 +800,4 @@ void TopmostEventTarget::MaybeSendPointerEvents(
 }
 
 }  // namespace layout
-}  // namespace cobalt
+}  // namespace cobalt
```
andrewsavage1: Please add back EOL

cobalt/ui_navigation/scroll_engine/scroll_engine.cc
```diff
@@ -317,6 +317,12 @@ void ScrollEngine::HandleScrollStart(
     const math::Matrix3F& initial_transform) {
   DCHECK(base::SequencedTaskRunner::GetCurrentDefault() ==
          scroll_engine_.task_runner());
+  
+  if (!scroll_container) {
```
andrewsavage1: This seems unnecessary if it's also being added in topmost_event_target.cc

cobalt/layout/topmost_event_target.cc
a8901147: done

cobalt/ui_navigation/scroll_engine/scroll_engine.cc
a8901147: `ScrollEngine::HandleScrollStart` is only called by `TopmostEventTarget::HandleScrollState`.

you are right. Let me remove it. thanks

PR #7215, opened 2025-09-18
Add max cache size in cobalt

andrewsavage1:
Bug: 388317020

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -81,6 +81,14 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
 constexpr base::FilePath::CharType kTrustTokenFilename[] =
     FILE_PATH_LITERAL("Trust Tokens");
 
+base::FilePath GetCacheDir() {
+  base::FilePath cache_path;
+  base::PathService::Get(base::DIR_CACHE, &cache_path);
+  cache_path = cache_path.Append(FILE_PATH_LITERAL("Default"))
+                   .Append(FILE_PATH_LITERAL("HTTP Cache"));
+  return cache_path;
+}
+
 }  // namespace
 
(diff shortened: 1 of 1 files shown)


cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -216,6 +224,8 @@ void CobaltContentBrowserClient::ConfigureNetworkContextParams(
 
   // Always enable the HTTP cache.
   network_context_params->http_cache_enabled = true;
+  network_context_params->http_cache_max_size = 1024 * 1024 * 24;
```
hlwarriner, line 227: Please just add a comment explaining that this value comes from the hardware certification requirements.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -81,6 +81,14 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
 constexpr base::FilePath::CharType kTrustTokenFilename[] =
     FILE_PATH_LITERAL("Trust Tokens");
 
+base::FilePath GetCacheDir() {
```
hlwarriner, line 84: We should probably be more specific and name this function something like `GetHttpCacheDir` to differentiate the returned path from `kCacheDirname`.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -216,6 +224,8 @@ void CobaltContentBrowserClient::ConfigureNetworkContextParams(
 
   // Always enable the HTTP cache.
   network_context_params->http_cache_enabled = true;
+  network_context_params->http_cache_max_size = 1024 * 1024 * 24;
+  network_context_params->http_cache_directory = GetCacheDir();
```
hlwarriner, line 228: This will be overridden below when `!in_memory`? Is that the intended behavior?

PR #7223, opened 2025-09-19
[macOS][Backport] Cherrypick upstream removals of CG_AVAILABLE_BUT_DEPRECATED

fayev-g:
Cobalt chromium_mac-arm64 and chromium m114 checkout currently do not build due to 
`../../third_party/webrtc/modules/desktop_capture/mac/screen_capturer_mac.mm:35:9: error: expected function body after function declarator
        CG_AVAILABLE_BUT_DEPRECATED(`

This macro no longer works on macOS 15 and was removed in Chromium upstream (landed in m119). Cherrypick those changes into Cobalt: [commit 1](https://source.chromium.org/chromium/_/webrtc/src/+/46da472f824aaebfb1352483e9e2145d7bc1483d), [commit 2](https://source.chromium.org/chromium/chromium/src/+/b63463aa7034239144529066d206d34e30cda570).

I attempted to update the webrtc hash to m138 (https://github.com/youtube/cobalt/pull/7172) and m119, but the cascading dependency build errors didn't make it worth it.

Bug: 446024838

### DIFF
--- a/third_party/webrtc/modules/desktop_capture/mac/screen_capturer_mac.mm
+++ b/third_party/webrtc/modules/desktop_capture/mac/screen_capturer_mac.mm
@@ -20,87 +20,6 @@
 #include "rtc_base/trace_event.h"
 #include "sdk/objc/helpers/scoped_cftyperef.h"
 
-// All these symbols have incorrect availability annotations in the 13.3 SDK.
-// These have the correct annotation. See https://crbug.com/1431897.
-// TODO(thakis): Remove this once FB12109479 is fixed and we updated to an SDK
-// with the fix.
-
-static CGDisplayStreamRef __nullable
-    wrapCGDisplayStreamCreate(CGDirectDisplayID display,
-                              size_t outputWidth,
-                              size_t outputHeight,
-                              int32_t pixelFormat,
-                              CFDictionaryRef __nullable properties,
-                              CGDisplayStreamFrameAvailableHandler __nullable handler)
-        CG_AVAILABLE_BUT_DEPRECATED(
-            10.8,
-            14.0,
-            "Please use ScreenCaptureKit API's initWithFilter:configuration:delegate: instead") {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wunguarded-availability-new"
-  return CGDisplayStreamCreate(
-      display, outputWidth, outputHeight, pixelFormat, properties, handler);
-#pragma clang diagnostic pop
-}
-
-static CFRunLoopSourceRef __nullable
-    wrapCGDisplayStreamGetRunLoopSource(CGDisplayStreamRef cg_nullable displayStream)
-        CG_AVAILABLE_BUT_DEPRECATED(10.8,
-                                    14.0,
-                                    "There is no direct replacement for this function. Please use "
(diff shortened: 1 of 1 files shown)


third_party/webrtc/modules/desktop_capture/mac/screen_capturer_mac.mm
```diff
@@ -20,87 +20,6 @@
 #include "rtc_base/trace_event.h"
```
yell0wd0g, line 20: Since we don't support desktop capture whatsoever on Cobalt --
perhaps just setting `rtc_desktop_capture_supported` in args.gn
to false would make modifications to this file unnecessary, and
save some binary size as side benefit?

[rtc_desktop_capture_supported](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/webrtc/webrtc.gni;l=332-334?q=rtc_desktop_capture_supported&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain:third_party%2Fwebrtc%2F)

content/browser/media/capture/desktop_capture_device_mac.cc
```diff
@@ -6,7 +6,6 @@
```
yell0wd0g, line 6: Similarly to what I mention below, since we don't support desktop
capture on Cobalt, we could just not build this file by setting
`enable_screen_capture` to false in args.gn and that'd make
modifications to this file unnecessary (and perhaps save some
bianry size).

[enable_screen_capture](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:content/common/features.gni;l=23-24?q=enable_screen_capture&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain:content%2F)

andrewsavage1:
Can we cherry pick the chromium commit directly?

third_party/webrtc/modules/desktop_capture/mac/screen_capturer_mac.mm
fayev-g: See response in other comment (consolidating the discussion there).

content/browser/media/capture/desktop_capture_device_mac.cc
fayev-g: I just tried setting both flags to false locally to see if gn gen passes. It seems to be fine for the `rtc_desktop_capture_supported` (don't know if it actually compiles fine), but it looks like `enable_screen_capture` tightly couples some build dependencies and I'm getting unrelated //content/browser/... [import errors](https://paste.googleplex.com/6563442504892416).

imo cherry-picking the upstream fixes seems clearer to me that those macros shouldn't exist anywhere in the codebase anymore. I like the idea of saving binary size though. I think I can disable that separately, especially since I can't just set `enable_screen_capture` to false easily.

fayev-g:
> Can we cherry pick the chromium commit directly?

Done and updated the PR. I didn't know if the cherrypicks would work so I initially `--no-commit`'d them to check locally.

fayev-g:
> This LGTM from a process perspective, I assume actual changes are fine if they're from upstream and pass smoke testing

Yes, verified on my local mac that content_shell is building for chromium_mac-arm64 with this branch. Thanks!

PR #7230, opened 2025-09-19
[tvOS] Do not set RPATH in starboard GN config

rakuco:
Basically a reland of the most conservative version of #7179 that does not remove on all platforms, as that still needs to be worked out in a way that does not break linux-x64x11 at runtime at least.

The syntax is not recognized by LLD, leading to errors like

    ld64.lld: error: unknown argument '-rpath=/foo/bar'

when linking executables both both host and target.

Since everything works fine without setting any RPATH, just avoid trying to set it altogether.

Bug: 432642933

### DIFF
--- a/starboard/build/config/BUILD.gn
+++ b/starboard/build/config/BUILD.gn
@@ -167,9 +167,15 @@ config("starboard") {
     defines += [ "STARBOARD_CONFIGURATION_INCLUDE=\"$starboard_path/configuration_public.h\"" ]
   }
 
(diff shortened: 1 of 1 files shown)


rakuco:
@andrewsavage1 could you merge this as well? Or should I get a second lgtm first?

andrewsavage1:
> @andrewsavage1 could you merge this as well? Or should I get a second lgtm first?

Thanks for the reminder!

PR #7232, opened 2025-09-19
Expose precise memory info

andrewsavage1:
This ensures performance.memory is updated frequently.

Fixed: 445761040

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -119,7 +119,7 @@ public final class CommandLineOverrideHelper {
 
         // Enable precise memory info so we can make accurate client-side
         // measurements.
-        paramOverrides.add("PreciseMemoryInfo")
+        paramOverrides.add("PreciseMemoryInfo");
 
         return paramOverrides;
     }


andrewsavage1:
> Changes LGTM. Two questions:
> 
> * Do we need the same flag in [CommandLineOverrideHelperTest.java](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java)?

This is a test file, no?

> * Do we need to update
>   cobalt_switch_defaults_starboard_test.cc ?

I could probably add a test here that when this flag is added `RuntimeEnabledFeatures::PreciseMemoryInfoEnabled()` is true, but I wouldn't be able to add the equivalent test to android I don't think

yell0wd0g:
> > Changes LGTM. Two questions:
> > 
> > * Do we need the same flag in [CommandLineOverrideHelperTest.java](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java)?
> 
> This is a test file, no?

Ah sorry I meant [CommandLineOverrideHelper.java](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java;l=1;drc=d081530720487ae30bf1f565e7c29376fb600a35)

andrewsavage1:
@yell0wd0g  I was looking through the tests you linked, it seems pointless to add a test case for this IMO as it's doing the same thing with a different string—I don't seem to be able to test that the actual feature is enabled, just that the flag is there.

PR #7243, opened 2025-09-19
blink/renderer: Stub Context Menu Controller

ckim1346:
Per the discussion of #6422, this PR creates a new stub Context Menu Controller that essentially does not do anything when a context menu is called. We want to disable context menus for Chrobalt and they were causing crashes on touch-screen devices. 

Bug: 430584909

### DIFF
--- a/third_party/blink/renderer/core/page/context_menu_controller.h
+++ b/third_party/blink/renderer/core/page/context_menu_controller.h
@@ -43,6 +43,7 @@ class MouseEvent;
 class Page;
 struct ContextMenuData;
 
+// TODO() remove buildflags and upstream to Chromium
 #if BUILDFLAG(IS_COBALT)
 class CORE_EXPORT ContextMenuController
 #else
@@ -125,6 +126,7 @@ class CORE_EXPORT ContextMenuController final
 
  protected:
   // Returns whether a Context Menu was actually shown.
+  // TODO(): upstream virtual to Chromium
   virtual bool ShowContextMenu(LocalFrame*,
                                const PhysicalOffset&,
                                WebMenuSourceType,


third_party/blink/renderer/core/page/stub_context_menu_controller.h
```diff
@@ -0,0 +1,49 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#ifndef THIRD_PARTY_BLINK_RENDERER_CORE_PAGE_STUB_CONTEXT_MENU_CONTROLLER_H_
+#define THIRD_PARTY_BLINK_RENDERER_CORE_PAGE_STUB_CONTEXT_MENU_CONTROLLER_H_
+
+#include "third_party/blink/renderer/core/page/context_menu_controller.h"
+
+#if BUILDFLAG(IS_COBALT)
```
yell0wd0g: nit: Instead of this and the corresponding guard in the .cc, 
how about just leaving a build error:
```
#if !BUILDFLAG(IS_COBALT)
#error "StubContextMenuController should only be used for Cobalt builds."
#endif
```

third_party/blink/renderer/core/page/stub_context_menu_controller.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "third_party/blink/public/common/input/web_menu_source_type.h"
+
+#if BUILDFLAG(IS_COBALT)
+namespace blink {
+
+StubContextMenuController::StubContextMenuController(Page* page)
+    : ContextMenuController(page) {}
+
+bool StubContextMenuController::ShowContextMenu(
```
yell0wd0g, line 38: I wonder if there'd be a unit test or similar where the expectations
should be updated if `BUILDFLAG(IS_COBALT)`. I couldn't find any
page_test.cc or similar, perhaps it exists and we're just not 
building/running it?

third_party/blink/renderer/core/page/context_menu_controller.h
```diff
@@ -119,11 +123,12 @@ class CORE_EXPORT ContextMenuController final
  private:
   friend class ContextMenuControllerTest;
 
+ protected:
   // Returns whether a Context Menu was actually shown.
-  bool ShowContextMenu(LocalFrame*,
-                       const PhysicalOffset&,
-                       WebMenuSourceType,
-                       const MouseEvent* mouse_event = nullptr);
+  virtual bool ShowContextMenu(LocalFrame*,
```
yell0wd0g, line 130: I wonder if this could be upstreamed, WDYT? We could file a
crbug.com/new and ask the OWNER of this file, and leave a 
`// TODO()` here.

third_party/blink/renderer/core/page/stub_context_menu_controller.cc
ckim1346: Yeah I'm a bit surprised all of the tests in third_party/blink/renderer/core/page/context_menu_controller_test.cc are passing -- if you look at the previous two PR's (#6422 and #6675) I had to disable a lot of those tests. Should we still go ahead and update the expectations in that file?

third_party/blink/renderer/core/page/stub_context_menu_controller.h
ckim1346: Tried this out but as I suspected, this causes our non-cobalt build actions to fail

third_party/blink/renderer/core/page/stub_context_menu_controller.cc
yell0wd0g: Ah, probably because the two files are included in 
//third_party/blink/renderer/core/page/build.gni
unconditionally. Said gni file doesn't know about gn flags 
like `is_cobalt` so WDYT about adding these two files to
//third_party/blink/renderer/core/BUILD.gn directly with
something like
```
if (is_cobalt) {
  blink_core_sources_page += [  
    "stub_context_menu_controller.cc",
    "stub_context_menu_controller.h",
  ]
}
```
?

It's a bit less clear (because files are added far from
the array definition) but it would be easier to rebase 
and understand. And then you can have the shorter guard
here.

third_party/blink/renderer/core/page/context_menu_controller.h
ckim1346: I think it would make more sense if the class declaration wasn't guarded by a cobalt buildflag: 
```#if BUILDFLAG(IS_COBALT)
class CORE_EXPORT ContextMenuController
#else
class CORE_EXPORT ContextMenuController final
#endif```

Otherwise is there a benefit for the Chromium upstream for the function to be virtual since other classes can't technically inherit and override the function?

Comment (yell0wd0g) [reply third_party/blink/renderer/core/page/context_menu_controller.h:130]: The upstreaming would include removing the `final` and adding the
`virtual`. Part of the negotiation would be to prove that a)
this is not a maintenance burden on upstream and b) helps us
(maybe other downstream embedders) to not have (to carry) this
local modification.

File: third_party/blink/renderer/core/page/context_menu_controller.h [hunk 4]
```diff
@@ -119,11 +124,13 @@ class CORE_EXPORT ContextMenuController final
...
   friend class ContextMenuControllerTest;
 
+ protected:
   // Returns whether a Context Menu was actually shown.
-  bool ShowContextMenu(LocalFrame*,
-                       const PhysicalOffset&,
-                       WebMenuSourceType,
-                       const MouseEvent* mouse_event = nullptr);
+  // TODO(): upstream virtual to Chromium
+  virtual bool ShowContextMenu(LocalFrame*,
```
andrewsavage1, line 130: Why was this not guarded by a BUILDFLAG? It can create an issue in the chromium build, as we now have a virtual method in a final class

PR #7244, opened 2025-09-19
Add a github action for checking binary size on evergreen-arm platforms

niranjanyardi:
Issue: 441531382

Add a github action to the evergreen-rdk workflow to check binary size regressions

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -56,6 +56,9 @@ group("gn_all") {
       "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",
     ]
   }
+  if (use_evergreen) {
+    deps += [ ":cobalt_sizes" ]
+  }
 }
 
 group("cobalt_browsertests") {
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -15,6 +15,7 @@
 import("//build/linux/strip_binary.gni")
 import("//cobalt/build/modular_executable.gni")
 import("//testing/test.gni")
+import("//tools/binary_size/sizes.gni")
 import("//ui/base/ui_features.gni")
 
 # This arg temporarily enables or disables the updater_sandbox target.
@@ -242,5 +243,18 @@ test("cobalt_unittests") {
 
   if (use_ozone) {
     deps += [ "//ui/ozone:ozone_switches" ]
+  }
+}
+
+if (use_evergreen) {
+  sizes_test("cobalt_sizes") {
+    data_deps = [ ":cobalt" ]
+    data = [ "${root_out_dir}/libcobalt.so" ]
+    executable_args = [
+      "--platform",
+      "cobalt-evergreen",
+      "--isolated-script-test-output",
+      "${root_out_dir}/test_results.json",
+    ]
   }
 }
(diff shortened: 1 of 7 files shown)


cobalt/BUILD.gn
```diff
@@ -227,3 +228,16 @@ test("cobalt_unittests") {
...
   }
 }
+
+if (use_evergreen) {
+  sizes_test("cobalt_sizes") {
+    data_deps = [ ":cobalt" ]
+    data = [ "${root_out_dir}/libcobalt.so" ]
+    executable_args = [
+      "--platform",
+      "linux",
```
kaidokert: I think this here is an issue that forces you to modify `tools/binary_size/sizes.py`. A cleaner approach is to probably define a separate platform. e.g. `cobalt-linux` and/or `cobalt-evergreen` or such. 

Your current change would actually break upstream code in sizes.py

tools/binary_size/sizes.py
```diff
@@ -242,12 +242,7 @@ def main_linux(output_directory, results_collector, size_path):
   """
   assert size_path is None
   binaries = [
```
kaidokert, line 244: Add a new entry to `main_map` in this script where we can list our relevant binaries. You'll likely want to add more / different configs for evergreen.

This change as is is upstream-breaking and we shouldn't do this.

cobalt/BUILD.gn
niranjanyardi: yup, sorry i think the reviewers got added automatically, this is still a draft.

My plan is to provide an extra argument called is_cobalt to the script and then make changes in the script accordingly

kaidokert: `is_cobalt` arg doesn't even seem to fit very well. There's already a very clear per-platform flag split in the codepaths there. Just add a different platform and it should click.

tools/binary_size/sizes.py
niranjanyardi: ack

niranjanyardi:
@isarkis mentioned that it might be better to have the build action run this additonal step instead of having an additional action which requires downloading the output of the github build artifacts .

What do you think @kaidokert ?

cobalt/BUILD.gn
```diff
@@ -227,3 +228,16 @@ test("cobalt_unittests") {
     deps += [ "//ui/ozone:ozone_switches" ]
   }
 }
+
+if (use_evergreen) {
```
kaidokert, line 252: Wouldn't you want to do this for modular builds as well as a ground truth / reference ? E.g. sometimes Evergreen builds grow for Evergreen specific features. Modular builds would represent more as to what goes into the lib.

niranjanyardi: no harm in adding , i just added it to the minimum number of platforms i thought will be useful- evergreen-arm.
ill add it for linux-modular as well

niranjanyardi:
> @isarkis mentioned that it might be better to have the build action run this additonal step instead of having an additional action which requires downloading the output of the github build artifacts .
> 
> What do you think @kaidokert ?

talked offline, we are creating a spearate job. this should eventually scale to more metrics on the chrome perf dashboard: https://chromium.googlesource.com/chromium/src/+/HEAD/tools/binary_size/README.md#resource_sizes_py

PR #7248, opened 2025-09-22
Enable support for widevine 

santoshmahto7:
No DRM key system is  registered from cobalt  on RDK due to which play of encrypted video doesn't work.
So enable widevine for all platform.
Some starboard platform may not have DRM support.

Bug: 444462743

### DIFF
--- a/cobalt/build/configs/common.gn
+++ b/cobalt/build/configs/common.gn
@@ -47,3 +47,6 @@ v8_enable_atomic_object_field_writes = false
 
 # This will force V8 to embed snapshot blob in binary.
 v8_use_external_startup_data = false
+
+# Enable widevine for encrypted media playback.
+enable_widevine = true
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -65,7 +65,6 @@ std::string GetMimeFromAudioType(const ::media::AudioType& type) {
   }
 }
 
-#if BUILDFLAG(IS_ANDROID)
 ::media::SupportedCodecs GetStarboardEmeSupportedCodecs() {
   ::media::SupportedCodecs codecs =
       ::media::EME_CODEC_AAC | ::media::EME_CODEC_AVC1 |
@@ -78,7 +77,6 @@ std::string GetMimeFromAudioType(const ::media::AudioType& type) {
   // TODO(b/375232937) Add IAMF
   return codecs;
 }
-#endif
 
 void BindHostReceiverWithValuation(mojo::GenericPendingReceiver receiver) {
   content::RenderThread::Get()->BindHostReceiver(std::move(receiver));
@@ -110,7 +108,6 @@ void CobaltContentRendererClient::RenderFrameCreated(
   }
 }
 
-#if BUILDFLAG(IS_ANDROID)
 void AddStarboardCmaKeySystems(::media::KeySystemInfos* key_system_infos) {
   ::media::SupportedCodecs codecs = GetStarboardEmeSupportedCodecs();
 
@@ -134,16 +131,13 @@ void AddStarboardCmaKeySystems(::media::KeySystemInfos* key_system_infos) {
       ::media::EmeFeatureSupport::ALWAYS_ENABLED,    // Persistent state.
       ::media::EmeFeatureSupport::ALWAYS_ENABLED));  // Distinctive identifier.
 }
-#endif
 
 void CobaltContentRendererClient::GetSupportedKeySystems(
     ::media::GetSupportedKeySystemsCB cb) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   ::media::KeySystemInfos key_systems;
-#if BUILDFLAG(IS_ANDROID)
   AddStarboardCmaKeySystems(&key_systems);
   std::move(cb).Run(std::move(key_systems));
-#endif
 }
 
 bool CobaltContentRendererClient::IsSupportedAudioType(
--- a/third_party/blink/common/BUILD.gn
+++ b/third_party/blink/common/BUILD.gn
@@ -354,7 +354,7 @@ source_set("common") {
     deps += [ "//ui/gfx/" ]
   }
 
-  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia) {
+  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia || is_starboard) {
     defines = [ "ENABLE_PROTECTED_MEDIA_IDENTIFIER_PERMISSION" ]
   }
 }


niranjanyardi:
@santoshmahto7  you can edit the default gn args list : cobalt/build/configs/evergreen-arm-hardfp-rdk/args.gn
and the flag you mentioned : enable_widevine

kjyoun:
Can you fix the broken builds?

```
$ autoninja -C out/linux-x64x11_debug cobalt:gn_all chromedriver content_shell crashpad_handler dump_syms minidump_stackwalk
....
2025-09-22T17:11:38.3259021Z FAILED: cobalt 
2025-09-22T17:11:38.3264925Z "python3" "../../build/toolchain/gcc_link_wrapper.py" --output="./cobalt" -- ../../third_party/llvm-build/Release+Asserts/bin/clang++ -Werror -fuse-ld=lld -Wl,--fatal-warnings -Wl,--build-id -fPIC -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,--icf=all -Wl,--color-diagnostics -Wl,--undefined-version -Wl,--no-call-graph-profile-sort -m64 -no-canonical-prefixes -Wl,-O2 -Wl,--gc-sections -Wl,--gdb-index -rdynamic -Wl,-z,defs -Wl,--as-needed -nostdlib++ --sysroot=../../build/linux/debian_bullseye_amd64-sysroot -Wl,--disable-new-dtags -Wl,-rpath=\$ORIGIN -Wl,-rpath=/__w/cobalt/cobalt/src/out/linux-x64x11_debug/starboard -Wl,-rpath=/__w/cobalt/cobalt/src/out/linux-x64x11_debug/starboard -o "./cobalt" -Wl,--start-group @"./cobalt.rsp" starboard/libstarboard.so.17 ./libtest_trace_processor.so -Wl,--end-group  -ldl -lpthread -lrt -lresolv -lm -lz -luuid 
2025-09-22T17:11:38.3284014Z ld.lld: error: undefined symbol: cdm::WidevineKeySystemInfo::WidevineKeySystemInfo(unsigned int, base::internal::flat_tree<media::EncryptionScheme, base::identity, std::Cr::less<void>, std::Cr::vector<media::EncryptionScheme, std::Cr::allocator<media::EncryptionScheme>>>, base::internal::flat_tree<media::CdmSessionType, base::identity, std::Cr::less<void>, std::Cr::vector<media::CdmSessionType, std::Cr::allocator<media::CdmSessionType>>>, unsigned int, base::internal::flat_tree<media::EncryptionScheme, base::identity, std::Cr::less<void>, std::Cr::vector<media::EncryptionScheme, std::Cr::allocator<media::EncryptionScheme>>>, base::internal::flat_tree<media::CdmSessionType, base::identity, std::Cr::less<void>, std::Cr::vector<media::CdmSessionType, std::Cr::allocator<media::CdmSessionType>>>, cdm::WidevineKeySystemInfo::Robustness, cdm::WidevineKeySystemInfo::Robustness, media::EmeFeatureSupport, media::EmeFeatureSupport)
2025-09-22T17:11:38.3289086Z >>> referenced by cobalt_content_renderer_client.cc:125 (../../cobalt/renderer/cobalt_content_renderer_client.cc:125)
2025-09-22T17:11:38.3291501Z >>>               obj/cobalt/renderer/renderer/cobalt_content_renderer_client.o:(cobalt::AddStarboardCmaKeySystems(std::Cr::vector<std::Cr::unique_ptr<media::KeySystemInfo, std::Cr::default_delete<media::KeySystemInfo>>, std::Cr::allocator<std::Cr::unique_ptr<media::KeySystemInfo, std::Cr::default_delete<media::KeySystemInfo>>>>*))
2025-09-22T17:11:38.3293609Z clang++: error: linker command failed with exit code 1 (use -v to see invocation)
2025-09-22T17:11:40.9620044Z [46965/46968] LINK ./content_shell
2025-09-22T17:11:40.9638654Z ninja: build stopped: subcommand failed.
2025-09-22T17:12:41.0168841Z RBE Stats: Downloaded: 4.95 GB, Uploaded: 0 B
```

santoshmahto7:
@kjyoun  The build fails since enable_widevine=true isn't set. I have added it in default gn args config files for linux

cobalt/build/configs/linux_common.gn
```diff
@@ -110,3 +110,6 @@ use_dbus = false
 
 # Embed ICU data in the cobalt binary
 icu_use_data_file = false
+
+# Enable widevine for encrypted media playback.
+enable_widevine = true
```
niranjanyardi: DRM support is not present/expected for all Linux platforms. For ex: evergreen-arm-hardfp-raspi doesn't have DRM support on starboard side, i believe even evergreen-x64 doesn't 

This should probably be enabled per platform - you would set this argument in the rdk args.gn specifically - not on the cobalt side 

@jasonzhangxx ptal

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -65,7 +65,7 @@ std::string GetMimeFromAudioType(const ::media::AudioType& type) {
   }
 }
 
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_STARBOARD)
```
niranjanyardi: maybe instead of IS_STARBOARD you need some macro related to widevine like ENABLE_DRM_FOR_COBALT
or some other existing macro - which only gets enabled for RDK

niranjanyardi: FYI @borongc

third_party/blink/common/BUILD.gn
```diff
@@ -354,7 +354,7 @@ source_set("common") {
     deps += [ "//ui/gfx/" ]
   }
 
-  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia) {
+  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia || is_starboard) {
```
niranjanyardi, line 357: instead of is_starboard, you would reuse the drm flag you enabled

cobalt/renderer/cobalt_content_renderer_client.cc
santoshmahto7: Right, In the latest push, this is addressed as:
1. At starboard level,  using enable_starboard_media_drm  - means true only when starboard implementation supports drm (e.g rdk)
2.  cobalt  level:  define buildflag  ENABLE_COBALT_MEDIA_DRM - when gn args enable_widevine=true and enable_starboard_media_drm is also true.
3. enable_widevine is enabled in rdk default config

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -65,7 +65,7 @@ std::string GetMimeFromAudioType(const ::media::AudioType& type) {
   }
 }
 
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(ENABLE_COBALT_MEDIA_DRM)
```
borongc: Would use `IS_STARBOARD` work in this case?

cobalt/renderer/cobalt_content_renderer_client.cc
borongc: Is it necessary to avoid Starboard implementation to play DRM videos if it doesn't support DRM?

That is, given all platforms using Starboard in prod will have DRM supports, is it necessary to have an extra flag for our reference platforms (e.g., linux-x64 which has starboard implementations but doesn't support DRM).

cobalt/build/configs/linux_common.gn
borongc: Would put `is_starboard` in [third_party/widevine/cdm/widevine.gni](https://github.com/youtube/cobalt/blob/main/third_party/widevine/cdm/widevine.gni#L15) for `enable_widevine` solve the issue?

Or should we put `enable_widevine=true` for all Cobalt with Starboard by default?

cobalt/renderer/cobalt_content_renderer_client.cc
santoshmahto7: The first version of patch was using IS_STARBOARD flag, but it seemed, not all starboard platform have DRM implementation (e.g raspi) . So we need to use FLAG  =  (starboard supports  DRM ) AND (chromium widevine is enabled)  == ENABLE_COBALT_MEDIA_DRM

cobalt/renderer/cobalt_content_renderer_client.cc
santoshmahto7: Not sure if I understood correctly, but here is how this patch do.
Once all starboard implementation supports DRM, we can remove the `enable_starboard_media_drm`. until then DRM video playback  work only when  both (starboard + chromium )  have DRM supports which is enable via ENABLE_COBALT_MEDIA_DRM.
So ENABLE_COBALT_MEDIA_DRM is defined only when starboard supports DRM and also chromium support for DRM is enabled (using enable_widevine).

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -65,7 +65,7 @@ std::string GetMimeFromAudioType(const ::media::AudioType& type) {
   }
 }
 
-#if BUILDFLAG(IS_ANDROID)
```
niranjanyardi, line 68: we can remove all BUILDFLAG conditions here and below- as we are enabling widevine for all platforms at cobalt level . As this file is in the cobalt/ directory - we don't need to add a BUILDFLAG(IS_COBALT) here

third_party/blink/common/BUILD.gn
niranjanyardi: i think is_starboard makes sense in this place as all other values are OS's , no need to change this

cobalt/build/configs/BUILD.gn
```diff
@@ -99,6 +100,10 @@ config("buildflag_defines") {
     defines += [ "ENABLE_BUILDFLAG_USE_STARBOARD_MEDIA" ]
   }
 
+  if (enable_widevine) {
+    defines += [ "ENABLE_BUILDFLAG_ENABLE_WIDEVINE" ]
```
niranjanyardi: we don't need this define as it's only  used in the cobalt/ directory not in any of the chromium upostream code

cobalt/build/configs/BUILD.gn
```diff
@@ -15,6 +15,7 @@
 import("//build/config/arm.gni")
 import("//cobalt/build/configs/hacks.gni")
 import("//starboard/build/buildflags.gni")
+import("//third_party/widevine/cdm/widevine.gni")
```
niranjanyardi: will be good to remove this as well. same logic as  https://github.com/youtube/cobalt/pull/7248/files#r2383461932

build/build_config.h
```diff
@@ -287,6 +287,12 @@
 #define BUILDFLAG_INTERNAL_USE_STARBOARD_MEDIA() (0)
 #endif
 
+#if defined(ENABLE_BUILDFLAG_ENABLE_WIDEVINE)
```
niranjanyardi: i guess this section is also not needed as the define is unnecessary : https://github.com/youtube/cobalt/pull/7248/files#r2383461932

cobalt/renderer/cobalt_content_renderer_client.cc
niranjanyardi: i think it makes more sense to just remove the buildflag as it applies to all cobalt platforms

cobalt/build/configs/linux_common.gn
niranjanyardi: talked offline to borong, IIUC the decision is to enable widevine for all platforms on cobalt side. fix resulting errors in starboard media pipeline

niranjanyardi:
i took a look at the failures - they are related to the define @santoshmahto7  added 
```
2025-09-26T17:46:32.2536444Z In file included from ../../components/cdm/renderer/android_key_systems.cc:5:
2025-09-26T17:46:32.2536908Z In file included from ../../components/cdm/renderer/android_key_systems.h:13:
2025-09-26T17:46:32.2537614Z gen/third_party/widevine/cdm/buildflags.h:9:9: error: 'BUILDFLAG_INTERNAL_ENABLE_WIDEVINE' macro redefined [-Werror,-Wmacro-redefined]
2025-09-26T17:46:32.2538176Z #define BUILDFLAG_INTERNAL_ENABLE_WIDEVINE() (1)
```

the latest status is to remove that define - so this error would go away with that

cobalt/renderer/cobalt_content_renderer_client.cc
santoshmahto7: Ack

build/build_config.h
santoshmahto7: Ack

cobalt/build/configs/BUILD.gn
santoshmahto7: Ack

cobalt/build/configs/BUILD.gn
santoshmahto7: Ack

cobalt/build/configs/common.gn
```diff
@@ -47,3 +47,6 @@ v8_enable_atomic_object_field_writes = false
 
 # This will force V8 to embed snapshot blob in binary.
 v8_use_external_startup_data = false
+
+# Enable widevine for encrypted media playback
```
borongc: nit: add period

santoshmahto7: Done

niranjanyardi:
none of the errors look related to this PR, retriggering tests
there's an existing issue related to crashpad : https://b.corp.google.com/issues/447430712

borongc (approved):

niranjanyardi:
@andrewsavage1 can you ptal ? you are part of both approval groups

PR #7252, opened 2025-09-22
Remove content shell config from Cobalt CIs

haozheng-cobalt:
We should not build content shell in Chrobalt's CIs anymore, as Chrobalt now depends on a separate cobalt_shell.
This is not removing the content shell build from our chromium CIs

Bug: 446692827

### DIFF
--- a/.github/workflows/linux.yaml
+++ b/.github/workflows/linux.yaml
@@ -54,7 +54,7 @@ jobs:
       nightly: ${{ github.event.inputs.nightly }}
     secrets:
       datadog_api_key: ${{ secrets.DD_API_KEY }}
-  # This is cobalt_shell without starboard
+  # This is content_shell without starboard
   linux-web-tests:
     uses: ./.github/workflows/main.yaml
     permissions:


.github/config/linux-web-tests.json
```diff
@@ -6,7 +6,6 @@
   "web_tests": true,
   "targets": [
     "blink_web_tests",
-    "content_shell",
```
andrewsavage1, line 9: @hlwarriner why doesn't this fail? Does blink_web_tests build cobalt_shell instead?

hlwarriner: The check ran on this PR still used `content_shell`, and `content_shell` was still built:
* https://github.com/youtube/cobalt/actions/runs/17924904705/job/50969840968?pr=7252#step:6:144
* https://github.com/youtube/cobalt/actions/runs/17924904705/job/50968741044?pr=7252#step:6:47516

This is because of a  `blink_web_tests -> //:blink_web_tests_support_data -> //content/shell:content_shell` data dependency path. I found this with
```
gn refs out/linux-x64x11-no-starboard_devel/ content/shell:content_shell
//:blink_web_tests_support_data
//:chromium_builder_asan
//:gn_all
//content/shell:content_shell_crash_test
//content/test:content_browsertests
```

@haozheng-cobalt, we'd need to get Cobalt's web tests working with cobalt_shell if we want to stop building content_shell for the linux-web-tests builder. We should talk - that may be difficult.

haozheng-cobalt: Thanks for catching this, yes we had a ticket for addressing web tests, it's time for it to be prioritized now. b/434218741

haozheng-cobalt: I see, in this case, content_shell is still built for web_tests, I believe the content shell config is redundant and can be removed.

andrewsavage1: Okay so you're fine with content_shell being built here?

haozheng-cobalt: yes, Holden and I are having another conversation about how to make web tests depend on cobalt_shell. 
It does not have to block this PR.

.github/workflows/linux.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
       nightly: ${{ github.event.inputs.nightly }}
     secrets:
       datadog_api_key: ${{ secrets.DD_API_KEY }}
-  # This is content_shell without starboard
+  # This is cobalt_shell without starboard
```
andrewsavage1: I think this comment is inaccurate for now?

haozheng-cobalt: true, yea, correcting it

PR #7253, opened 2025-09-22
[Rebase m138] Fix for renamed fields in hdr_metadata

madhurajayaraman:
Bug: 418842688

Breaking change introduced in : https://source.chromium.org/chromium/chromium/src/+/ef71a52a18075a6a239b076f7bfdc18261c938f4

### DIFF
--- a/media/starboard/starboard_utils.cc
+++ b/media/starboard/starboard_utils.cc
@@ -308,37 +308,37 @@ SbMediaColorMetadata MediaToSbMediaColorMetadata(
   SbMediaMasteringMetadata sb_media_mastering_metadata = {};
 
   if (hdr_metadata) {
-    const auto& color_volume_metadata = hdr_metadata->color_volume_metadata;
+    const auto& color_volume_metadata = hdr_metadata->smpte_st_2086;
 
     sb_media_mastering_metadata.primary_r_chromaticity_x =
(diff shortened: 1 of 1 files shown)


media/starboard/starboard_utils.cc
```diff
@@ -308,37 +308,37 @@ SbMediaColorMetadata MediaToSbMediaColorMetadata(
   SbMediaMasteringMetadata sb_media_mastering_metadata = {};
 
   if (hdr_metadata) {
-    const auto& color_volume_metadata = hdr_metadata->color_volume_metadata;
+    const auto& color_volume_metadata = hdr_metadata->smpte_st_2086;
```
borongc, line 311: Should the variable be aligned with the new naming as well? like `smpte_st_2086` instead of `color_volume_metadata`?

madhurajayaraman: I'll leave this to you/anyone on media to determine! Let me know what you want to do here.

borongc: Please help rename`color_volume_metadata` to `smpte_st_2086`, thanks!

madhurajayaraman: https://github.com/youtube/cobalt/pull/7281

Seems like this went in already! Abandoning this change.

PR #7254, opened 2025-09-22
[Rebase m138] Add gates to fix unused variable warning

madhurajayaraman:
Bug: 418842688

default_software_renderer is unused because of an IS_COBALT flag, so disabled the warning.


ui/gl/init/gl_display_initializer.cc
```diff
@@ -57,14 +57,14 @@ void GetEGLInitDisplays(bool supports_angle_d3d,
   bool force_software_gl =
       IsSoftwareGLImplementation(GetGLImplementationParts());
 
-  std::string requested_renderer =
-      force_software_gl
 #if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: I think this would be more clear if it were at the declaration of the variable on L48 instead of changing the default value below

```
#if BUILDFLAG(IS_STARBOARD)
  const char* default_software_renderer = gles...
#else
  const char* default_software_renderer = angle...
#endif
```

PR #7264, opened 2025-09-22
[Rebase M138] String view fix for AudioDecoder

sideb0ard:
Bug: 418842688

base/strings/string_piece.h was deleted
https://chromium-review.googlesource.com/c/chromium/src/+/5729956
https://chromium-review.googlesource.com/c/chromium/src/+/5272353

### DIFF
--- a/cobalt/media/audio/audio_decoder.cc
+++ b/cobalt/media/audio/audio_decoder.cc
@@ -16,7 +16,6 @@
 
 #include <stdint.h>
 
-#include <string_view>
 #include <utility>
 
 #include "base/logging.h"
@@ -35,8 +34,7 @@ bool DecodeAudioFileData(blink::WebAudioBus* destination_bus,
   }
 
   LOG(INFO) << "Cobalt WAV decoder initializing..";
-  auto handler = media::WavAudioHandler::Create(base::as_bytes(
-      base::span<const char>(data, data_size)));
+  auto handler = media::WavAudioHandler::Create(base::as_bytes(audio_file_data));
 
   if (!handler) {
     LOG(ERROR) << "Failed to create WavAudioHandler.";


cobalt/media/audio/audio_decoder.cc
```diff
@@ -36,8 +36,8 @@ bool DecodeAudioFileData(blink::WebAudioBus* destination_bus,
   }
 
   LOG(INFO) << "Cobalt WAV decoder initializing..";
-  auto handler =
-      media::WavAudioHandler::Create(base::StringPiece(data, data_size));
+  auto handler = media::WavAudioHandler::Create(
+      base::span<const uint8_t>(reinterpret_cast<const uint8_t*>(data), data_size));
```
kjyoun: How about this ? (Gemini suggested this code)
I want to avoid type casting if posssible.
```
  base::as_byte_span(base::StringPiece(data, data_size)))
```

madhurajayaraman: Shouldn't this just be string_view?

sideb0ard: I can't use that suggested code as is, as it uses 'base::StringPiece' which was deleted but gemini did suggest this updated method to avoid doing a reinterpret_cast
+  auto handler = media::WavAudioHandler::Create(base::as_bytes(
+      base::span<const char>(data, data_size)));

madhurajayaraman: https://github.com/youtube/cobalt/pull/7256 for context, chromium CLs are in that description.

sideb0ard: @madhurajayaraman no, i tried that, but compilation fails, and gemini explains as "can't use std::string_view when a base::span<const uint8_t> is expected primarily because they are different types with different underlying element types"

madhurajayaraman: Looking at wav_audio_handler.h in upstream, I think the function signature changed altogether so let's hold off on this one till we can compile whatever has that? Or we can get this in for the interim.

sideb0ard: done.

PR #7266, opened 2025-09-22
[Rebase M138] Fix Cobalt DecoderBuffer

sideb0ard:
Bug: 418842688

This PR comments out a number of  factory functions on DecoderBuffer, which are not compatible with Cobalt's allocator model and need to be conditionally compiled out.

TESTED: compiles fine for Chromium content_shell and compiles for Cobalt with:
$ autoninja -C out/linux-x64x11_devel/  obj/media/base/base/decoder_buffer.o
Proxy started successfully.
ninja: Entering directory `out/linux-x64x11_devel/'
ninja: no work to do.
RBE Stats: down 0 B, up 0 B,

### DIFF
--- a/media/base/decoder_buffer.cc
+++ b/media/base/decoder_buffer.cc
@@ -58,9 +58,6 @@ class ExternalSharedMemoryAdapter : public DecoderBuffer::ExternalMemory {
 #if !BUILDFLAG(USE_STARBOARD_MEDIA)
 DecoderBuffer::DecoderBuffer(size_t size)
     : data_(base::HeapArray<uint8_t>::Uninit(size)) {}
-
-DecoderBuffer::DecoderBuffer(base::span<const uint8_t> data)
-    : data_(base::HeapArray<uint8_t>::CopiedFrom(data)) {}
 #endif // !BUILDFLAG(USE_STARBOARD_MEDIA)
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
@@ -90,15 +87,16 @@ DecoderBuffer::DecoderBuffer(DemuxerStream::Type type,
 
   memcpy(data_, data.data(), data.size());
 }
-#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
+#else // !BUILDFLAG(USE_STARBOARD_MEDIA)
+DecoderBuffer::DecoderBuffer(base::span<const uint8_t> data)
+    : data_(base::HeapArray<uint8_t>::CopiedFrom(data)) {}
 
-#if !BUILDFLAG(USE_STARBOARD_MEDIA)
 DecoderBuffer::DecoderBuffer(base::HeapArray<uint8_t> data)
     : data_(std::move(data)) {}
 
 DecoderBuffer::DecoderBuffer(std::unique_ptr<ExternalMemory> external_memory)
     : external_memory_(std::move(external_memory)) {}
-#endif // !BUILDFLAG(USE_STARBOARD_MEDIA)
+#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
 
 DecoderBuffer::DecoderBuffer(DecoderBufferType decoder_buffer_type,
                              std::optional<ConfigVariant> next_config)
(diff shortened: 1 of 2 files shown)


media/base/decoder_buffer.h
```diff
@@ -107,18 +107,22 @@ class MEDIA_EXPORT DecoderBuffer
   };
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
+#if !BUILDFLAG(USE_STARBOARD_MEDIA)
   // Allocates buffer with |size| > 0. |is_key_frame_| will default to false.
   // If size is 0, no buffer will be allocated.
   // TODO(crbug.com/365814210): Remove this constructor. Clients should use the
   // FromArray constructor instead asking for a writable DecoderBuffer.
   explicit DecoderBuffer(size_t size);
+#endif // !BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun: do you need this and line 120?

Aren't line 110 (opening #if)  and line 125 (closing #if) enough?

media/base/decoder_buffer.h
```diff
@@ -136,7 +141,9 @@ class MEDIA_EXPORT DecoderBuffer
   //
   // Ownership of |data| is transferred to the buffer.
   static scoped_refptr<DecoderBuffer> FromArray(base::HeapArray<uint8_t> data);
+#endif // !BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun: ditto

aren't  line 134 and line 176 enough?

media/base/decoder_buffer.h
```diff
@@ -158,12 +165,15 @@ class MEDIA_EXPORT DecoderBuffer
       base::ReadOnlySharedMemoryRegion region,
       uint64_t offset,
       size_t size);
+#endif // !BUILDFLAG(USE_STARBOARD_MEDIA)
 
+#if !BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun: Do you need these?

media/base/decoder_buffer.cc
```diff
@@ -39,6 +37,8 @@ void DecoderBuffer::Allocator::Set(Allocator* allocator) {
 }
 #endif // BUILDFLAG(USE_STARBOARD_MEDIA)
 
+namespace {
```
kjyoun, line 40: why moving anonymous namespace scope?

Is this an intended change?

kjyoun (approved):
LGTM % comments

media/base/decoder_buffer.cc
sideb0ard: yeah, the rebase conflict resolution placed it in the wrong place, wrapping our '#IF STARBOARD'. 
With our Cobalt version, it should only wrap the DecoderBuffer::Allocator::GetInstance. Within Chromium code they added ExternalSharedMemoryAdapter within an anonymous namespace in 2024. (https://source.chromium.org/chromium/chromium/src/+/main:media/base/decoder_buffer.cc;bpv=1;bpt=1?q=media%2Fbase%2Fdecoder_buffer.cc&ss=chromium%2Fchromium%2Fsrc)

Otherwise compiler errors with:
../../media/base/decoder_buffer.cc:27:53: error: cannot define or redeclare 'GetInstance' here because namespace '' does not enclose namespace 'Allocator'
   27 | DecoderBuffer::Allocator* DecoderBuffer::Allocator::GetInstance() {
      |                           ~~~~~~~~~~~~~~~~~~~~~~~~~~^
../../media/base/decoder_buffer.cc:28:3: error: non-local lambda expression cannot have a capture-default

media/base/decoder_buffer.h
sideb0ard: good catch, thanks! gemini being overly cautious.

media/base/decoder_buffer.h
sideb0ard: done.

media/base/decoder_buffer.h
sideb0ard: Seems not, compiles fine without this change. thanks!

kjyoun:
The those comments by Gemini sound reasonable to me.

Can you take a look?

sideb0ard:
> The those comments by Gemini sound reasonable to me.
> 
> Can you take a look?

awesome, thank you! Local gemini also found a few other spots to simplify and reduce pre-processor ifdefs

PR #7269, opened 2025-09-22
Add lz4 build targets for Evergreen build

andrewsavage1:
Uses our lz4_compress tool to compress Evergreen shared libraries into lz4 archives.

Introduces stripping for modular official builds (for shared libraries). Unstripped libraries are put in lib.unstripped/ in the out directory. We now use the builtin strip option from gcc_toolchain.gni instead of a custom stripping process.

Also clean up unneeded Evergreen toolchain files.

Fixed: 446747351

### DIFF
--- a/cobalt/build/toolchain/linux/BUILD.gn
+++ b/cobalt/build/toolchain/linux/BUILD.gn
@@ -14,7 +14,7 @@
 
 import("//build/toolchain/gcc_toolchain.gni")
 
-if (is_official_build) {
+if (is_official_build && is_cobalt_hermetic_build) {
   if (target_cpu == "arm") {
     strip_executable = "arm-linux-gnueabi-strip"
   } else if (target_cpu == "arm64") {


starboard/tools/lz4_compress/BUILD.gn
```diff
@@ -13,8 +13,13 @@
 # limitations under the License.
 
 if (current_toolchain == host_toolchain) {
+  config("disable_warnings") {
+    cflags = [ "-Wno-sign-compare" ]
```
hlwarriner: We should fix this. I filed b/446972942 and will take care of it. If you can add a todo here pointing to the bug that would be great.

starboard/evergreen/shared/strip_install_target.gni
hlwarriner, line 1: It's nice that we're able to remove this and use `build/toolchain/gcc_toolchain.gni` instead.

hlwarriner (approved):
Nice! I like the way you implemented this.

I think it would be very helpful to explain a bit more about what this PR is doing in the description.
* "Add lz4 build targets" is kind of ambiguous IMO. It would be good to mention that this PR builds our custom `lz4_compress` tool and runs it to generate lz4-compressed binaries.
* It's also significant that this PR introduces stripping of evergreen binaries. This is in part significant because we still release stripped but uncompressed binaries.
* It would help to mention that we now use `build/toolchain/gcc_toolchain.gni` to strip the binaries, which is why we're able to remove some of the old Evergreen toolchain files. It took me some time to read through `build/toolchain/gcc_toolchain.gni` and understand what was going on.

Other than the description this LGTM.

starboard/tools/lz4_compress/BUILD.gn
andrewsavage1: I decided to just fix it as it was a single error

PR #7271, opened 2025-09-23
[Rebase m138] Templating ScriptPromise types in h5vcc metrics

madhurajayaraman:
Bug: 418842688

Chromium change : https://chromium-review.googlesource.com/c/chromium/src/+/5247712

Chromium adds templates to script promise types

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.h
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.h
@@ -38,8 +38,6 @@ class ExecutionContext;
 class LocalDOMWindow;
 class ScriptState;
 
-using IDLVoid = IDLUndefined;
-
 class MODULES_EXPORT H5vccMetrics final
     : public EventTarget,
       public ExecutionContextLifecycleObserver,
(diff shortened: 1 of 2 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.cc
```diff
@@ -38,9 +38,9 @@ void H5vccMetrics::ContextDestroyed() {
   OnCloseConnection();
 }
 
-ScriptPromise H5vccMetrics::enable(ScriptState* script_state,
+ScriptPromise<void> H5vccMetrics::enable(ScriptState* script_state,
```
andrewsavage1: I doubt this is right—I don't see a single instance of `ScriptPromise<void>` upstream. Maybe it should use `IDLUndefined` instead of `void`?

madhurajayaraman: Yes I had what can best be described as a house of cards of errors, so pushed a couple of things prematurely to get everything separated out into their own PRs. Revisiting this soon!

andrewsavage1 (approved):

PR #7273, opened 2025-09-23
test: Increase duration and margin for flaky clock test

jellefoks:
The `PosixTimeClockTests.ClockIncreasesOverTime` test has been observed to be flaky on certain devices. This is suspected to be caused by the device being busy with other tasks, leading to timing inaccuracies.

To improve the robustness of the test, this change:
- Increases the test's work time from 100ms to 500ms.
- Lowers the expectation for measured CPU time from 75% to 60% of the total elapsed time.

These adjustments provide a wider margin for the test to pass, making it more resilient to short process interruptions and less likely to fail due to intermittent device load.

Fixed: 446869137
Bug: 446869137


andrewsavage1 (approved):

andrewsavage1:
FYI if you have 

Fixed: 446869137

you don't need

Bug: 446869137

as well

PR #7279, opened 2025-09-23
[Rebase M138] - add unreachable break to cflags

sideb0ard:
Bug: 418842688

Similar error to the unreachable return:

../../media/starboard/progressive/mp4_map.cc:483:7: error: 'break' will never be executed [-Werror,-Wunreachable-code-break]
  483 |       break;                                                                                                 
      |       ^~~~~                                                                                                  
1 error generated.                                                                                                   
ninja: build stopped: subcommand failed.

### DIFF
--- a/media/starboard/BUILD.gn
+++ b/media/starboard/BUILD.gn
@@ -22,8 +22,10 @@ import("//media/media_options.gni")
 import("//starboard/build/buildflags.gni")
 
 config("disable_unreachable_warnings") {
-  cflags = [ "-Wno-unreachable-code-return",
-             "-Wno-unreachable-code-break" ]
+  cflags = [
+    "-Wno-unreachable-code-return",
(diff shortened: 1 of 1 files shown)


media/starboard/BUILD.gn
```diff
@@ -21,8 +21,9 @@ assert(use_starboard_media, "This file is only for use_starboard_media builds")
 import("//media/media_options.gni")
 import("//starboard/build/buildflags.gni")
 
-config("disable_unreachable_return") {
-  cflags = [ "-Wno-unreachable-code-return" ]
+config("disable_unreachable_warnings") {
+  cflags = [ "-Wno-unreachable-code-return",
```
andrewsavage1: This doesn't look correctly formatted

sideb0ard: reformatted, thanks.

PR #7280, opened 2025-09-23
Enable basic crash reporting for evergreen-x64

hlwarriner:
This enables Crashpad components to be built and to snapshot+upload crashes for evergreen-x64.

This implementation does not yet have feature parity with c25; most of the significant features and customizations yet to be implemented are tracked by child issues under b/406511608.

The remaining open questions from go/crashpad-design-questions would not have a major impact on this implementation and so should not prevent us from merging this PR.

Significant changes vs. c25:
* The Crashpad client and handler are built on top of //base rather than mini_chromium. This follows the approach used in Chromium.
* The Crashpad handler is installed from the elf_loader_sandbox and loader_app. This obviates the need for partners to do the installation in their ports and also removes a //starboard -> //base -> //starboard dependency cycle.

Issue: 406511608

### DIFF
--- a/third_party/crashpad/crashpad/handler/BUILD.gn
+++ b/third_party/crashpad/crashpad/handler/BUILD.gn
@@ -45,12 +45,10 @@ static_library("handler") {
   }
 
   if (crashpad_is_linux) {
-    if (!defined(is_native_target_build) || !is_native_target_build) {
-      sources += [
-        "linux/cros_crash_report_exception_handler.cc",
-        "linux/cros_crash_report_exception_handler.h",
-      ]
-    }
+    sources += [
+      "linux/cros_crash_report_exception_handler.cc",
+      "linux/cros_crash_report_exception_handler.h",
+    ]
     # TODO(https://crbug.com/1420445): Remove this config when M115 branches.
     configs += [
       "../build:crashpad_is_in_chromium",


base/process/launch_posix.cc
```diff
@@ -48,7 +48,9 @@
 #include "build/build_config.h"
 
 #if BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
+#if !BUILDFLAG(IS_NATIVE_TARGET_BUILD)
```
andrewsavage1: I would think that the native target build would have enable_cobalt_hermetic_hacks=false (and similarly is_cobalt_hermetic_build=false), as it is not a hermetic build

starboard/elf_loader/BUILD.gn
```diff
@@ -108,10 +108,13 @@ if (current_toolchain == starboard_toolchain) {
       "//starboard/content/fonts:copy_font_data",
     ]
 
-    if (!sb_is_evergreen_compatible || enable_cobalt_hermetic_hacks) {
-      deps += [ "//starboard/crashpad_wrapper:wrapper_stub" ]
-    } else {
+    # TODO: b/446904317 - use the real wrapper for other 3P platforms once they
+    # can build it and its deps.
+    if (sb_is_evergreen_compatible && target_os == "linux" &&
```
andrewsavage1, line 113: Consider making a temporary variable should_build_crashpad so you know what you need to remove and can easily change it to true

third_party/crashpad/crashpad/build/crashpad_buildconfig.gni
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if (is_cobalt) {
+  import("//cobalt/build/configs/initialize_variables.gni")
```
andrewsavage1: This is imported in the top-level BUILDCONFIG.gn already so I don't understand why this is necessary?

base/process/launch_posix.cc
hlwarriner: The intent of `is_cobalt_hermetic_build` wasn't totally clear to me. I thought it described the *overall* build and roughly meant that the *cobalt layer* is building with the starboard interface. And because the setting of `use_custom_libcxx` to false at https://github.com/youtube/cobalt/blob/main/build/config/c%2B%2B/c%2B%2B.gni#L50 is conditioned on `is_cobalt_hermetic_build`, I decided to just set `is_cobalt_hermetic_build` to true in native target builds.

But since you clarified offline that "any target when is_cobalt_hermetic_build=true is being treated as if the Starboard interface is the only system functions it has access to," I agree that  `is_cobalt_hermetic_build` should be false for native target builds. I'll upload another commit to address this.

third_party/crashpad/crashpad/build/crashpad_buildconfig.gni
hlwarriner: This might just be a relic from some local changes I had earlier. I'll try to remove it.

cobalt/build/configs/BUILD.gn
```diff
@@ -24,6 +24,8 @@ config("cobalt_config") {
   if (current_toolchain == starboard_toolchain &&
       starboard_toolchain != default_toolchain) {
     configs += [ "//${starboard_path}/platform_configuration" ]
+  } else if (current_toolchain == native_target) {
```
andrewsavage1: Combine this with the above condition (`|| current_toolchain == native_target`)

cobalt/BUILD.gn
```diff
@@ -49,6 +49,15 @@ group("gn_all") {
   if (is_starboard && build_updater_sandbox) {
     deps += [ "//chrome/updater/cobalt:updater_sandbox" ]
   }
+
+  # TODO: b/446904317 - add these deps for all 3P platforms once build issues
+  # are resolved.
+  if (target_os == "linux" && target_cpu == "x64") {
+    deps += [
+      "//third_party/crashpad/crashpad/handler:crashpad_handler(//$starboard_path/toolchain:native_target)",
```
andrewsavage1: Can't we use the `native_target` variable instead of specifying the path here?

build/config/c++/c++.gni
```diff
@@ -45,8 +45,8 @@ use_custom_libcxx = use_custom_libcxx && !is_nacl
 
 # Cobalt comment: Override the use_custom_libcxx variable for partner toolchains to
 # avoid linking with chromium's C++ libraries.
-if (is_cobalt_hermetic_build && (current_toolchain == starboard_toolchain ||
-                                 current_toolchain == native_target)) {
+if (is_cobalt_hermetic_build && current_toolchain == starboard_toolchain ||
+    defined(native_target) && current_toolchain == native_target) {
```
andrewsavage1: When is native_target not defined?

third_party/boringssl/BUILD.gn
```diff
@@ -187,7 +187,7 @@ component("boringssl") {
                 rebase_path(version_script, root_build_dir) ]
   }
 
-  if (is_cobalt_hermetic_build) {
+  if (is_cobalt_hermetic_build && !is_native_target_build) {
```
andrewsavage1: I think `!is_native_target_build` is unnecessary

third_party/crashpad/crashpad/build/crashpad_buildconfig.gni
```diff
@@ -60,7 +60,12 @@ if (crashpad_is_in_chromium) {
   crashpad_is_mac = is_mac
   crashpad_is_ios = is_ios
   crashpad_is_win = is_win
-  crashpad_is_linux = is_linux && !is_cobalt_hermetic_build || is_chromeos
+  crashpad_is_linux = is_linux &&
+    # We don't actually use crashpad components built with the cobalt toolchain
+    # in hermetic builds, and this variable causes sources to be included that
+    # introduce compiler errors for this build configuration.
+    !(is_cobalt_hermetic_build && current_toolchain == cobalt_toolchain) ||
```
andrewsavage1, line 67: When is is_cobalt_hermetic_build true outside of the default (cobalt) toolchain?

third_party/crashpad/crashpad/client/BUILD.gn
```diff
@@ -72,7 +72,12 @@ crashpad_static_library("client") {
     sources += [ "crashpad_client_fuchsia.cc" ]
   }
 
-  if (enable_cobalt_hermetic_hacks) {
+  # TODO: b/406511608 - Cobalt: these sources are needed because Cobalt shell
+  # still depends on the crashpad client through its dependency on
+  # //components/crash/core/app, although the functionality of this component
+  # is stubbed out. We might consider removing the dependency entirely, which
+  # would obviate the need for these client stubs.
+  if (current_toolchain == default_toolchain && enable_cobalt_hermetic_hacks) {
```
andrewsavage1, line 80: The `current_toolchain == default_toolchain` check here is redundant

third_party/crashpad/crashpad/client/BUILD.gn
```diff
@@ -148,7 +157,7 @@ static_library("common") {
     ]
   }
 
-  if (enable_cobalt_hermetic_hacks) {
+   if (current_toolchain == default_toolchain && enable_cobalt_hermetic_hacks) {
```
andrewsavage1, line 160: Redundant (also remove leading space)

third_party/crashpad/crashpad/handler/BUILD.gn
```diff
@@ -45,10 +45,12 @@ static_library("handler") {
   }
 
   if (crashpad_is_linux) {
-    sources += [
-      "linux/cros_crash_report_exception_handler.cc",
-      "linux/cros_crash_report_exception_handler.h",
-    ]
+    if (!defined(is_native_target_build) || !is_native_target_build) {
```
andrewsavage1: Are we not able to compile this?

third_party/crashpad/crashpad/snapshot/BUILD.gn
```diff
@@ -65,6 +65,13 @@ crashpad_static_library("snapshot") {
     "unloaded_module_snapshot.h",
   ]
 
+  if (defined(is_native_target_build) && is_native_target_build) {
```
andrewsavage1: I assume this is because is_native_target_build isn't defined for chromium configs—I think `if (is_cobalt && is_native_target_build)` would be more clear

third_party/crashpad/crashpad/util/net/tls.gni
```diff
@@ -15,7 +15,11 @@
 import("../../build/crashpad_buildconfig.gni")
 
 declare_args() {
-  if (crashpad_is_fuchsia || crashpad_is_android) {
+  # Cobalt comment: we're still exploring, in b/446904324, whether we can use
+  # libcurl. For now will keep the c25 status quo and use raw sockets with
+  # BoringSSL as the backend.
+  if (crashpad_is_fuchsia || crashpad_is_android ||
+          (defined(is_native_target_build) && is_native_target_build)) {
```
andrewsavage1: Replace with is_cobalt && is_native_target_build

starboard/elf_loader/BUILD.gn
hlwarriner: Thanks for the suggestion. I'll keep this as is for now but will do that if this logic gets more complex (for example, if we use the real wrapper for raspi too but are not yet able to for rdk).

build/config/c++/c++.gni
hlwarriner: In chromium builds. I changed the condition to check `is_cobalt` instead of `defined(native_target)`.

cobalt/BUILD.gn
hlwarriner: Yeah, we can. Thanks.

cobalt/build/configs/BUILD.gn
hlwarriner: Done

third_party/boringssl/BUILD.gn
hlwarriner: It should be after the `Set is_cobalt_hermetic_build false in native build` commit, yeah.

third_party/crashpad/crashpad/util/net/tls.gni
hlwarriner: Done

third_party/crashpad/crashpad/snapshot/BUILD.gn
hlwarriner: Done

third_party/crashpad/crashpad/build/crashpad_buildconfig.gni
hlwarriner: When we're building with the `starboard_toolchain`.

third_party/crashpad/crashpad/client/BUILD.gn
hlwarriner: I don't think it is. `enable_cobalt_hermetic_hacks`, like `is_cobalt_hermetic_build`, is true in the `starboard_toolchain`.

See https://github.com/youtube/cobalt/pull/7280/files#r2376351959.

third_party/crashpad/crashpad/client/BUILD.gn
hlwarriner: See my response to https://github.com/youtube/cobalt/pull/7280/files#r2376357204

third_party/crashpad/crashpad/handler/BUILD.gn
hlwarriner: I just ported this customization forward from c25 without giving it much thought. These sources are responsible for writing reports to a "ChromeOS" crash reporter so it's unlikely we'd ever want to use them. If you'd prefer we build them just to avoid the customization in the code here, though, I'm ok with that. Based on the build flags in third_party/crashpad/crashpad/handler/handler_main.cc I don't think they'll actually be linked into the handler for us, so it wouldn't impact binary size. Let me know.

third_party/crashpad/crashpad/build/crashpad_buildconfig.gni
andrewsavage1: Yep you're right, I should fix this in the future but makes sense to do it this way rn

third_party/crashpad/crashpad/handler/BUILD.gn
andrewsavage1: Yeah if we can compile them let's do that to avoid these changes, but if there are any failures makes sense to exclude them

hlwarriner: It's possible it will be an issue for other platform toolchains, like for raspi, but they compile fine for x64. I'll add them back for now.

starboard/elf_loader/elf_loader_sandbox.cc
```diff
@@ -108,6 +121,8 @@ void SbEventHandle(const SbEvent* event) {
     const SbEventStartData* data = static_cast<SbEventStartData*>(event->data);
     const starboard::CommandLine command_line(
         data->argument_count, const_cast<const char**>(data->argument_values));
+    InstallCrashpadHandler(
```
y4vor, line 124: Shouldn't we install the crashpad handler early in the Starboard startup?

hlwarriner: I tried to explain the reasoning behind this change from c25 in the PR description:
> The Crashpad handler is installed from the elf_loader_sandbox and loader_app. This obviates the need for partners to do the installation in their ports and also removes a //starboard -> //base -> //starboard dependency cycle.

The ` //starboard -> //base -> //starboard` dependency cycle is something we should be able to solve. But I do think it's a significant improvement to move the installation out of partner-written code; we recently had a partner integration issue because the partner missed the function comment, and our integration docs, and failed to pass certificates to `InstallCrashpadHandler` (b/400398018#comment46).

I get that earlier is better for crashpad handler installation, but this is still very early: before we attempt to load the cobalt shared library. If you think the loader app is doing a significant amount of work between these two points (where we used to install crashpad handler vs. where we are now in this PR) and there's a significant chance of a crash there, then let's discuss. Otherwise, I'd like to do the installation in the loader app / elf_loader_sandbox.

y4vor (approved):

starboard/elf_loader/elf_loader_sandbox.cc
hlwarriner: Based on offline discussion we'll merge this as is and use b/447422326 to discuss further. Thanks.

PR #7285, opened 2025-09-23
[Rebase M138] Update h5vcc metrics to template code

sideb0ard:
Bug: 418842688

Upstream converted ScriptPromise and ScriptPromiseResolver to typed templates:
https://chromium-review.googlesource.com/c/chromium/src/+/5381805

TESTED:
$ autoninja -C out/linux-x64x11_devel obj/third_party/blink/renderer/bindings/modules/v8/v8/v8_h_5_vcc_metrics.o
Proxy started successfully.
ninja: Entering directory `out/linux-x64x11_devel'
ninja: no work to do.
RBE Stats: down 0 B, up 0 B,

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.h
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.h
@@ -37,7 +37,8 @@ namespace blink {
 class ExecutionContext;
 class LocalDOMWindow;
 class ScriptState;
-class ScriptPromiseResolver;
+
+using IDLVoid = IDLUndefined;
 
 class MODULES_EXPORT H5vccMetrics final
     : public EventTarget,
@@ -54,10 +55,10 @@ class MODULES_EXPORT H5vccMetrics final
 
   // Web-exposed interface:
   DEFINE_ATTRIBUTE_EVENT_LISTENER(metrics, kMetrics)
-  ScriptPromise enable(ScriptState*, ExceptionState&);
-  ScriptPromise disable(ScriptState*, ExceptionState&);
+  ScriptPromise<IDLUndefined> enable(ScriptState*, ExceptionState&);
+  ScriptPromise<IDLUndefined> disable(ScriptState*, ExceptionState&);
   bool isEnabled();
-  ScriptPromise setMetricEventInterval(ScriptState*, uint64_t, ExceptionState&);
+  ScriptPromise<IDLUndefined> setMetricEventInterval(ScriptState*, uint64_t, ExceptionState&);
 
   // EventTarget impl.
   ExecutionContext* GetExecutionContext() const override {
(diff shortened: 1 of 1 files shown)


madhurajayaraman:
Weird that this doesn't need the corresponding changes in the .cc files but I'm not going to look a gift horse in the mouth

third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.h
```diff
@@ -37,7 +37,8 @@ namespace blink {
 class ExecutionContext;
 class LocalDOMWindow;
 class ScriptState;
-class ScriptPromiseResolver;
+
+using IDLVoid = IDLUndefined;
```
andrewsavage1, line 41: This seems unnecessary?

madhurajayaraman: WIthout it, it tries to resolve to IDLVoid and then errors with something like this : 
gen/third_party/blink/renderer/bindings/modules/v8/v8_h_5_vcc_system.cc:284:59: error: use of undeclared identifier 'IDLVoid'
  284 | static_assert(bindings::IsReturnTypeCompatible<IDLPromise<IDLVoid>, std::remove_cvref_t<decltype(return_value)>>, "Return type from native call is incompatible to the type specified in IDL");
      |

madhurajayaraman: I can't figure out *why* it's doing the weird resolution, and why IDLUndefined is okay as a type only after it does a weird Undefined -> void -> Undefined chain

madhurajayaraman: Maybe add a TODO to investigate what's going on here.

madhurajayaraman:
https://github.com/youtube/cobalt/pull/7271 Has all of the changes along with the .cc changes -- can we get this in instead?

third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.h
andrewsavage1: I downloaded this change and was able to compile h_5_vcc_metrics.o fine without this statement, let's remove it

andrewsavage1: Ah, I see it's an issue with v8_h_5_vcc_metrics.o not h_5_vcc_metrics.o—I'll look

andrewsavage1: The issue is that our IDL files are using `Promise<void>` instead of `Promise<undefined>`. Chromium doesn't use `Promise<void>` anywhere afaics

PR #7184, opened 2025-09-16
Add deep links for 3p

andrewsavage1:
Fixed: 441559782

PR #7289, opened 2025-09-23
evergreen: Fix autoplay video with startup url

borongc:
Add `--autoplay-policy=no-user-gesture-required` to allow to autoplay the video with startup url, same as https://github.com/youtube/cobalt/pull/4415.

Issue: 378134542
Issue: 446945583
Issue: 446985442

### DIFF
--- a/cobalt/app/cobalt_switch_defaults_starboard.cc
+++ b/cobalt/app/cobalt_switch_defaults_starboard.cc
@@ -111,6 +111,9 @@ const base::CommandLine::SwitchMap GetCobaltParamSwitchDefaults() {
         // Enable precise memory info so we can make accurate client-side
         // measurements.
         {switches::kEnableBlinkFeatures, "PreciseMemoryInfo"},
+        // Enable autoplay video/audio, as Cobalt may launch directly into media
+        // playback before user interaction.
+        {switches::kAutoplayPolicy, "no-user-gesture-required"},
   });
   return cobalt_param_switch_defaults;
 }


cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -111,6 +111,8 @@ const base::CommandLine::SwitchMap GetCobaltParamSwitchDefaults() {
         // Enable precise memory info so we can make accurate client-side
         // measurements.
         {switches::kEnableBlinkFeatures, "PreciseMemoryInfo"},
+        // Enable autoplay video/audio.
```
yell0wd0g: micro-nit: I'd add also the "why" here, e.g.
`// Enable autoplay video/audio, otherwise loaded videos don't play 
automatically upon load.` Or some such.

yell0wd0g (approved):
RS LGTM w/ a micro-nit.

andrewsavage1:
I have this in https://github.com/youtube/cobalt/pull/7184/files#diff-8a68d4926d1bbaf7393dcc6b52a7c4d3e1d798a6737f25e0a4f70ec151fcc69e already—would you be fine just waiting for that?

cobalt/app/cobalt_switch_defaults_starboard.cc
andrewsavage1: If we're going to add a comment, I think it should be more telling of why we need it and say `We may launch directly into media playback before user interaction`

borongc: Done.

borongc:
> I have this in https://github.com/youtube/cobalt/pull/7184/files#diff-8a68d4926d1bbaf7393dcc6b52a7c4d3e1d798a6737f25e0a4f70ec151fcc69e already—would you be fine just waiting for that?

Could we land this PR first as this is required for passing some failing YTS tests on raspi?

Autoplay is not only for deeplink, but also all other YTS media tests.

PR #7291, opened 2025-09-24
Experimental/m138 lw exclude chrome

yunsik-jung:
This PR enables the use of exclude_chrome_build=true for lightweight Linux builds.
With this change, it is possible to build essential Chromium components without including the full Chrome browser dependencies.

It also ensures that chromedriver remains available for testing and automation, even in lightweight builds.

This is applied to Linux only as exclude_chrome_build currently supports Linux, and Android lightweight builds don't require GN-level exclusion.

### DIFF
--- a/build/config/chrome_exclude.gni
+++ b/build/config/chrome_exclude.gni
@@ -3,6 +3,7 @@
 # found in the LICENSE file.
 
 import("//build/config/cast.gni")
+import("//build/config/lightweight.gni")
 
 declare_args() {
   # Controls whether to exclude //chrome directory from the build graph.
@@ -18,7 +19,8 @@ declare_args() {
   # - Linux: Optional
   #
   # TODO: Consider adding is_cast_android support.
-  exclude_chrome_build = is_fuchsia || is_castos
+  exclude_chrome_build =
+      is_fuchsia || is_castos || (enable_lightweight && !is_android)
 }
 
 if (exclude_chrome_build && !(is_fuchsia || is_castos || is_linux)) {


.github/config/chromium_linux-lightweight.json
```diff
@@ -5,6 +5,7 @@
   ],
   "gn_args": [
     {
+      "exclude_chrome_build": true,
```
kaidokert: Did you want to put this here, rather than in `build/config/lightweight.gni` ? 

If you want this to apply only to Linux build, perhaps this would work in the args block there:

```
declare_args() {
   ...
  
  # Disable Chrome build
  exclude_chrome_build = enable_lightweight && !is_android 
}
```

This would be preferable so that we can keep all of the "lightweight" logic contained in GN with a single flag.  I'm not sure though - i have not tried, if it can't work like that this seems fine, too.

yunsik-jung: I was thinking about the same thing, but still not sure which would be better. 

In current draft patch, exclude_chrome_build is defined in `build/config/chrome_exclude.gni` not `build/config/lightweight.gni` .

Similarly, It would be nice to have some guidance on how to handle if we need to disable existing Chromium upstream arguments which are not defined for lightweight .

For example, for the upstream argument `enable_feature_A`,
Candidate 1: Add enable_lightweight condition to existing upstream code.
```
import ("build/config/lightweight.gni")
enable_feature_A = {existing upstream conditions} && !enable_lightweight
```

Candidate 2: Keep upstream code unchanged and just apply to `-lightweight.json`.


If considering `exclude_chrome_build` only, I can also define `exclude_chrome_build` in `build/config/lightweight.gni` not `build/config/chrome_exclude.gni` and set `exclude_chrome_build = enable_lightweight && !is_android ` as you suggested.

In this case, `exclude_chrome_build` is used for the lightweight feature and should not consider `is_fuchisa` and `is_castos` condition.
(F.Y.I, current draft version was implemented as:  `exclude_chrome_build = is_fuchsia || is_castos` 
and replaced `is_fuchsia` and `is_castos` related condition with `exclude_chrome_build` in `//BUILD.gn`.)

I think this option is also OK with me. 
Could you let me know which method you prefer?

Youngsoo: I think Kaido’s suggestion is similar to what we have in candidate 1, and I also prefer the approach in candidate 1 which is simpler.
If there are no objections, I suggest we proceed with candidate 1 for now.

kaidokert: Are we going ahead with Candidate 1 then ? I'm fine with that

yunsik-jung: Will go with candidate 1.

yunsik-jung: Done

PR #7294, opened 2025-09-24
FakeGraphicsContextProvider: Move display initialization to separate function

rakuco:
There are a lot of `#ifdef`s and different code paths that can be taken to determine whether `eglGetDisplay()`, `eglGetPlatformDisplay()` or `eglGetPlatformDisplayEXT()` will be called, and if/when we need to make ANGLE use Metal on tvOS it will get even more confusing.

Make the code easier to understand by:
* Moving it to a separate function to make `InitializeEGL()` smaller.
* Setting the type of the attributes being set separately from declaring the vector of attributes itself.
* Using a constexpr array rather than a vector for said list of attributes and initializing it separately from the main code.
* Switching from checking for `#if condition` rather than `#if !condition` in the case of the `EGL_VERSION_1_5` checks.

Fixed: 447076985


rakuco:

starboard/testing/fake_graphics_context_provider.cc
```diff
@@ -109,6 +109,59 @@ typedef SbEglDisplay(EGLAPIENTRYP PFNEGLGETPLATFORMDISPLAYEXTPROC)(
...
+// Calls the appropriate EGL function to get an SbEglDisplay depending on
+// platform and available EGL version.
+SbEglDisplay GetEglDisplay() {
+#if defined(EGL_VERSION_1_5)
+  // eglGetDisplay() takes an array of EGLAttrib's.
+  using EglGetDisplayAttributeType = SbEglAttrib;
+#else
+  // eglGetPlatformDisplayEXT() takes an array of EGLint's.
+  using EglGetDisplayAttributeType = EGLint;
+#endif
```
yell0wd0g: @sherryzy - do we need this distinction? If we decided to always go via
Starboard for EGL loading, then we should always use here `SbEglAttrib`,
right?

On that line, I see that `SbEglAttrib` is an `int64_t` in [starboard](https://github.com/youtube/cobalt/blob/6858cd52d33fe67357830afe0ae9979ff03b25e7/starboard/egl.h#L53)) 
whereas the reference [khronos eglplatform.h](https://registry.khronos.org/EGL/api/EGL/eglplatform.h) says it's
an `int32_t` (`typedef khronos_int32_t EGLint;`). 

Should we
a) Use here SbEglAttrib on all paths and
b) Fix in a subsequent CL starboard's `SbEglAttrib` to `int64_t`?

yell0wd0g (commented):
Looks like a clarifying cleanup. Have one q for @sherryzy

starboard/testing/fake_graphics_context_provider.cc
rakuco: There's a tricky distinction here though: `eglGetPlatformDisplayEXT()` takes a `const EGLInt*`, whereas `eglGetDisplay()` takes a `const EGLAttrib*`.

An `EGLInt` is a 32-bit integer, but an `EGLAttrib` is an `intptr_t`.

sherryzy: Yes, @rakuco is correct. We need to keep these two distinct because eglGetPlatformDisplayEXT() and eglGetDisplay() require different input types. We could handle type conversion, but keep different paths is just simpler.

yell0wd0g: Ah so then shouldn't we have two types, `EglGetDisplayAttributeType` and
`EglGetPlatformDisplayAttributeType` (or `EglGetPlatformDisplayExtAttributeType` ?
Calling them the same way is confusing. We could then rope the
```
  static constexpr EglGetDisplayAttributeType kEglGetDisplayAttributes[] = {
```
into the ifdef and just have

```
#if defined(EGL_VERSION_1_5)
  // eglGetDisplay() takes an array of EGLAttrib's.
  static constexpr SbEglAttrib kEglRequestedGetDisplayAttributes[] = { 
#else
  // eglGetPlatformDisplayEXT() takes an array of EGLint's.
  static constexpr EGLint kEglRequestedGetDisplayAttributes[] = {
#endif
      EGL_PLATFORM_ANGLE_DEVICE_TYPE_ANGLE,
      EGL_PLATFORM_ANGLE_DEVICE_TYPE_EGL_ANGLE, EGL_PLATFORM_ANGLE_TYPE_ANGLE,
      EGL_PLATFORM_ANGLE_TYPE_OPENGLES_ANGLE,
      EGL_NONE  // Terminate the attribute list.
  };
```
?
I don't think there's much added value in the `using` data type
naming since it's only used in one place and that's inmediately
there.
Also re. using `#if/else` for declarations like that, I think
it might be a personal preference, personally I don't see any
issue with mixing statements (r/lvalues) and preprocessor, but
WDYT?

rakuco: Done in commit [f072e36](https://github.com/youtube/cobalt/pull/7294/commits/f072e366a1218ebbf71f5c59301889dbf33625f2).

Both approaches look fine to me at the moment, but it will look more confusing if we start using Metal on tvOS, because the whole declaration will look like this:

```c++
#if defined(EGL_VERSION_1_5)
  static constexpr SbEglAttrib kEglRequestedGetDisplayAttributes[] = {
#else
  static constexpr EGLint kEglRequestedGetDisplayAttributes[] = {
#endif  // defined(EGL_VERSION_1_5)
#if BUILDFLAG(IS_APPLE)
    EGL_PLATFORM_ANGLE_TYPE_ANGLE,
    EGL_PLATFORM_ANGLE_TYPE_METAL_ANGLE,
#else
    EGL_PLATFORM_ANGLE_DEVICE_TYPE_ANGLE,
    EGL_PLATFORM_ANGLE_DEVICE_TYPE_EGL_ANGLE,
    EGL_PLATFORM_ANGLE_TYPE_ANGLE,
    EGL_PLATFORM_ANGLE_TYPE_OPENGLES_ANGLE,
#endif  // BUILDFLAG(IS_APPLE)
    EGL_NONE  // Terminate the attribute list.
  };
```

yell0wd0g: Hmm I see. My other idea was to just just fully separate the functions,
and duplicate the current atttributes, e.g.

```
#if BUILDFLAG(IS_ANDROID)
SbEglDisplay GetEglDisplay() {
  return EGL_CALL_SIMPLE(eglGetDisplay(EGL_DEFAULT_DISPLAY));
}
#else
#if defined(EGL_VERSION_1_5)
SbEglDisplay GetEglDisplay() {
  // Attributes...
  return EGL_CALL_SIMPLE(eglGetPlatformDisplay(
      EGL_PLATFORM_ANGLE_ANGLE, reinterpret_cast<void*>(EGL_DEFAULT_DISPLAY),
      kEglRequestedGetDisplayAttributes));
}
#else 
SbEglDisplay GetEglDisplay() {
  // Attributes...
  // Extract function pointer
  // Return calling that f pointer if it exists
  // Else return eglGetDisplay() 
}
#endif
#endif 
```
and move the 
```
  SB_DCHECK_EQ(EGL_SUCCESS, EGL_CALL_SIMPLE(eglGetError()));
  SB_CHECK_NE(EGL_NO_DISPLAY, display);
```
to the callsite (which is perfectly aware of the EGL being used
since it calls `eglInitialize()` almost right away.

The rationale for duplicating the attributes bit would be
to simplify having/maintaining separate lists for tvOS etc.

I might even go further and have give the `GetEglDisplay()`
a suffix with what's the intended platform, and then have
an overarching `GetEglDisplay()` to call the appropriate one.
Overkill in lines of code, but perhaps clearer for the occassional
reader?

Thoughts?

rakuco: Duplicating `GetEglDisplay()` like that would lead to a declaration like

```c++
  static constexpr Foo kEglRequestedGetDisplayAttributes[] = {
#if BUILDFLAG(IS_APPLE)
    EGL_PLATFORM_ANGLE_TYPE_ANGLE,
    EGL_PLATFORM_ANGLE_TYPE_METAL_ANGLE,
#else
    EGL_PLATFORM_ANGLE_DEVICE_TYPE_ANGLE,
    EGL_PLATFORM_ANGLE_DEVICE_TYPE_EGL_ANGLE,
    EGL_PLATFORM_ANGLE_TYPE_ANGLE,
    EGL_PLATFORM_ANGLE_TYPE_OPENGLES_ANGLE,
#endif  // BUILDFLAG(IS_APPLE)
    EGL_NONE  // Terminate the attribute list.
  };
```

twice (unless we can guarantee that Apple builds use one of the variants, which is another story because it's not clear to me what's currently setting `EGL_VERSION_1_5` in the first place). Another option would be moving the attributes array outside the functions and declaring it once in this case.

yell0wd0g: EGL 1.5 is pretty old, I'd be surprised if it's not supported by
Apple. For reference, the issues we got with not fully supporting
1.5. came from Raspi (old crappy raspbian, ~2012) and RDK (mistery
nobody has updated the driver?). I'd vote for leaving a 
```
#if defined(APPLE) && !defined(EGL_VERSION_1_5)
#error "Apple eGL should be at least 1.5"
#endif 
```
and relying on that assumption, that AppleTV supports EGL1.5. The
`EGL_VERSION_1_5` preprocessor should come from the system library
`egl.h` header which should be a verbatim copy of the Khronos one.

rakuco: > The
> `EGL_VERSION_1_5` preprocessor should come from the system library
> `egl.h` header which should be a verbatim copy of the Khronos one.

That brings me back to the "other story" part I mentioned, I think. AFAICS nothing's actually including `egl.h` in this file, neither on tvOS nor on linux-x11x64.

abhijeetk: > Thoughts?

A complete separation of platform paths might be cleaner, each platform gets its own entry point and attribute set, avoiding subtle mismatches and keeping the code easier to extend and debug.

rakuco: > That brings me back to the "other story" part I mentioned, I think. AFAICS nothing's actually including `egl.h` in this file, neither on tvOS nor on linux-x11x64.

OK, this was more intricate than I thought and there are multiple issues that we need to solve. I've filed b/447478802 with a longer explanation.

rakuco:
Closing in favor of #7347

PR #7297, opened 2025-09-24
[Rebase m138] set hls_demuxer=false to disable new media flags

madhurajayaraman:
Bug: 418842688


third_party/blink/renderer/modules/mediasource/media_source.cc
```diff
@@ -71,7 +71,7 @@ namespace blink {
 
 namespace {
 
-#if BUILDFLAG(ENABLE_MSE_MPEG2TS_STREAM_PARSER)
+#if BUILDFLAG(ENABLE_MSE_MPEG2TS_STREAM_PARSER) && !BUILDFLAG(USE_STARBOARD_MEDIA)
```
andrewsavage1: Let's check w the media team if we want ENABLE_MSE_MPEG2TS_STREAM_PARSER set to true or false—if we want it to be false, we wouldn't make this change

madhurajayaraman: @borongc

borongc: I don't think supporting mp2t is required.

I briefly looked at the code, it seems in m114, `enable_hls_demuxer = false`.
https://source.chromium.org/chromium/chromium/src/+/refs/tags/114.0.5735.358:media/media_options.gni;l=108

However, after m120+, `enable_hls_demuxer` is true with `proprietary_codecs`:
https://source.chromium.org/chromium/chromium/src/+/main:media/media_options.gni;l=124
resulting in `enable_mse_mpeg2ts_stream_parser=true`.

I think setting `enable_hls_demuxer=false` in [cobalt/build/configs/common.gn](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/build/configs/common.gn) should fix the issue.

cobalt/build/configs/common.gn
```diff
@@ -45,3 +45,6 @@ v8_enable_concurrent_marking = false
 
 # This flag is also needed to disable concurrent marking.
 v8_enable_atomic_object_field_writes = false
+
+# Disable hls_demuxer so that enable_mse_mpeg2ts_stream_parser=false
+enable_hls_demuxer=false
```
kaidokert: Nitpick: end of line end missing from end of file

PR #7303, opened 2025-09-24
[rebase m138] add flag for implementation to match usage

madhurajayaraman:
Bug: 418842688

We gate the usage of this function behind `#if !BUILDFLAG(IS_APPLE) && !BUILDFLAG(IS_STARBOARD)` so added gate for implementation to match.

### DIFF
--- a/components/performance_manager/graph/process_node_impl_describer.cc
+++ b/components/performance_manager/graph/process_node_impl_describer.cc
@@ -60,7 +60,7 @@ std::string HostedProcessTypesToString(
   return str;
 }
 
-#if !BUILDFLAG(IS_APPLE)
+#if !BUILDFLAG(IS_APPLE) && !BUILDFLAG(IS_STARBOARD)
 const char* GetProcessPriorityString(const base::Process& process) {
   switch (process.GetPriority()) {
     case base::Process::Priority::kBestEffort:


kaidokert (approved):
This is one of the patches that should be cleaned up down the line. It's just a enum->string conversion function, ideally the enum would be still available even though we aren't multi-process, and it shouldn't need stubbing out.

Also, `#if !BUILDFLAG(IS_MULTI_PROCESS_CAPABLE)` would be better here - but it would need upstreaming.

madhurajayaraman:
> This is one of the patches that should be cleaned up down the line. It's just a enum->string conversion function, ideally the enum would be still available even though we aren't multi-process, and it shouldn't need stubbing out.
> 
> Also, `#if !BUILDFLAG(IS_MULTI_PROCESS_CAPABLE)` would be better here - but it would need upstreaming.

Is there a TODO of some kind I should add here? Or should we just keep a list of PRs that need to be revisited in a bug somewhere?

andrewsavage1:
> > This is one of the patches that should be cleaned up down the line. It's just a enum->string conversion function, ideally the enum would be still available even though we aren't multi-process, and it shouldn't need stubbing out.
> > Also, `#if !BUILDFLAG(IS_MULTI_PROCESS_CAPABLE)` would be better here - but it would need upstreaming.
> 
> Is there a TODO of some kind I should add here? Or should we just keep a list of PRs that need to be revisited in a bug somewhere?

PRs that need to edit an original commit are listed in b/446191460

kaidokert:
> Is there a TODO of some kind I should add here? Or should we just keep a list of PRs that need to be revisited in a bug somewhere?

I don't know what's better for recordkeeping tbh. You could tag them all in a single bug, adding a TODO works but adds more commit noise, you could add a github label to PRs where things like these come up.

At the end, those things will be visible in the actual diff against baseline as well, but they won't be necessarily immediate obvious and organized - and you lose context of review comments.

PR #7304, opened 2025-09-24
[rebase m138] remove unused field from function call

madhurajayaraman:
Bug: 418842688
Bug: 189889720

Chromium change:
[CSP] Remove code for navigate-to | https://chromium-review.googlesource.com/c/chromium/src/+/5850982


kaidokert:
We even had an old bug open for this: b/189889720

PR #7307, opened 2025-09-24
[Rebase m138] Remove unreachable code

madhurajayaraman:
Bug: 418842688

Chromium change : https://chromium-review.googlesource.com/c/chromium/src/+/5782602

### DIFF
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -1048,7 +1048,6 @@ base::TimeDelta ChunkDemuxer::GetWriteHead(const std::string& id) const {
   if (iter == id_to_streams_map_.end() || iter->second.empty()) {
     // Handled just in case.
     NOTREACHED();
-    return base::TimeDelta();
   }
 
   return iter->second[0]->GetWriteHead();


media/filters/chunk_demuxer.cc
```diff
@@ -1048,7 +1048,6 @@ base::TimeDelta ChunkDemuxer::GetWriteHead(const std::string& id) const {
   if (iter == id_to_streams_map_.end() || iter->second.empty()) {
     // Handled just in case.
     NOTREACHED();
```
kjyoun, line 1050: can you replace this with `NOTREACHED_NORETURN`?


https://source.corp.google.com/h/github/youtube/cobalt/+/main:base/notreached.h
```
// TODO(crbug.com/851128): Migrate NOTREACHED() callers to NOTREACHED_NORETURN()
// which is [[noreturn]] and always FATAL. Once that's done, rename
// NOTREACHED_NORETURN() back to NOTREACHED() and remove the non-FATAL version.
```

madhurajayaraman: This is the error when I try to use NOTREACHED_NORETURN():
```
siso_rule: clang/cxx
stderr:
../../media/filters/chunk_demuxer.cc:1050:5: error: use of undeclared identifier 'NOTREACHED_NORETURN'
 1050 |     NOTREACHED_NORETURN();
      |     ^~~~~~~~~~~~~~~~~~~
1 error generated.
```

`NOTREACHED_NORETURN()` doesn't exist anymore, `NOTREACHED` implements no return now, so `NOTREACHED` is the same thing as what `NOTREACHED_NORETURN()` used to be.
Chromium main also has no references to this anymore: https://source.chromium.org/search?q=NOTREACHED_NORETURN&ss=chromium%2Fchromium%2Fsrc&ssfr=1

kjyoun: I think that we have to follow the change in this PR

https://github.com/youtube/cobalt/pull/6808

kjyoun: BTW, my concern on NOTREACHED() is that it's removed on release build

> NOTREACHED():  ... If run, it will cause a crash if `DCHECK_IS_ON` is enabled. Prefer to unconditionally CHECK() instead of conditionally hitting a NOTREACHED().

madhurajayaraman: Ah thank you! I will cherry pick that in instead of this change, didn't know that we had a change for it already. I'll abandon this PR.

The current rebase branch is a couple months behind cobalt (should be up to date soon) so there are definitely little things like this that we're missing

PR #7308, opened 2025-09-24
feat(starboard): Add prctl wrapper for VMA naming

sherryzy:
This change introduces a wrapper for the `prctl` system call to support naming Virtual Memory Areas (VMAs) using `PR_SET_VMA_ANON_NAME`.

When the underlying kernel does not support this feature, the wrapper provides a fallback mechanism that writes the VMA information to a temporary file (`/tmp/cobalt_vma_tags_<pid>.txt`). This allows for memory debugging and attribution even on systems without the latest kernel features.

A new NPLB test, `PosixPrctlTest.SetVmaAnonName`, has been added to verify both the direct `prctl` call and the fallback mechanism.

The `prctl` wrapper is now used in various parts of the codebase, including the partition allocator and persistent memory allocator, under the `IS_STARBOARD` build flag.

Bug: 443798603, 444471844

### DIFF
--- a/starboard/elf_loader/exported_symbols.cc
+++ b/starboard/elf_loader/exported_symbols.cc
@@ -59,6 +59,7 @@
 #include "starboard/shared/modular/starboard_layer_posix_mmap_abi_wrappers.h"
 #include "starboard/shared/modular/starboard_layer_posix_pipe2_abi_wrappers.h"
 #include "starboard/shared/modular/starboard_layer_posix_poll_abi_wrappers.h"
+#include "starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.h"
 #include "starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.h"
 #include "starboard/shared/modular/starboard_layer_posix_semaphore_abi_wrappers.h"
 #include "starboard/shared/modular/starboard_layer_posix_signal_abi_wrappers.h"
@@ -230,6 +231,7 @@ ExportedSymbols::ExportedSymbols() {
 
   // POSIX APIs
   REGISTER_SYMBOL(aligned_alloc);
+  REGISTER_SYMBOL(atexit);
   REGISTER_SYMBOL(calloc);
   REGISTER_SYMBOL(close);
   REGISTER_SYMBOL(fdatasync);
@@ -348,6 +350,7 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_WRAPPER(pathconf);
   REGISTER_WRAPPER(pipe2);
   REGISTER_WRAPPER(poll);
+  REGISTER_WRAPPER(prctl);
   REGISTER_WRAPPER(pthread_attr_init);
   REGISTER_WRAPPER(pthread_attr_destroy);
   REGISTER_WRAPPER(pthread_attr_getdetachstate);
--- a/starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
@@ -1,0 +1,91 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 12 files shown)


starboard/elf_loader/exported_symbols.cc
```diff
@@ -230,6 +231,7 @@ ExportedSymbols::ExportedSymbols() {
 
   // POSIX APIs
   REGISTER_SYMBOL(aligned_alloc);
+  REGISTER_SYMBOL(atexit);
```
kaidokert, line 234: Is this related to the rest of the pr ? I could we we may need `atexit` but it's not clear from here as to how those two are linked.

third_party/musl/include/sys/prctl.h
```diff
@@ -177,6 +177,14 @@ struct prctl_mm_map {
 #define PR_PAC_SET_ENABLED_KEYS 60
 #define PR_PAC_GET_ENABLED_KEYS 61
 
+#if !defined(PR_SET_VMA)
+#define PR_SET_VMA 0x53564d41
```
kaidokert, line 181: This seems like a magic number - what's the origin of it ? Does upstream `musl` have this defined ? 

EDIT: Actually i checked, it's not there. I'd recommend making a pull request to musl to add it: https://github.com/kraj/musl/blob/kraj/master/include/sys/prctl.h

starboard/elf_loader/exported_symbols.cc
sherryzy: It's related. I used this function for cleaning up this tmp file in starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc

third_party/musl/include/sys/prctl.h
sherryzy: I get this number from "linux/prctl.h", and it should be consistent to all Linux kernel systems.

sherryzy: @kaidokert I created this PR to update upstream code: https://github.com/kraj/musl/pull/9, though I'm not sure how is the reviewing process going, and I can not add reviewers.

sherryzy:
Friendly ping..

third_party/musl/include/sys/prctl.h
kaidokert: They replied - they need patches emailed to mailing list. Different contribution workflow

niranjanyardi:
We would like to get this in so that smaps works on evergreen-rdk platform : https://issuetracker.google.com/454095852 RDK - Memory leak during endurance tests on 26.eap

talked to @johnedocampo  - john already added multiple functions under prctl - so vma(being added in this PR) would be an addition to that . John  also mentioned some changes will be needed on this PR to make it work.

@sherryzy when you are back - it'll be good to co-ordinate with John and get this in.

starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
```diff
@@ -0,0 +1,91 @@
...
+                               unsigned long arg3,
+                               unsigned long arg4,
+                               unsigned long arg5) {
+  int result = prctl(option, arg2, arg3, arg4, arg5);
+  if (result == -1 && errno == EINVAL && option == PR_SET_VMA &&
+      arg2 == PR_SET_VMA_ANON_NAME) {
+    // Kernel does not support PR_SET_VMA_ANON_NAME. Fallback to writing to a
+    // file.
+    char file_path[256];
+    snprintf(file_path, sizeof(file_path), "/tmp/cobalt_vma_tags_%d.txt",
```
jellefoks, line 64: I find a few things at issue with this: 

- The base name of the file should be a const string so that it doesn't have to be repeated in the different places where it's needed.
- The path should not be hardcoded to /tmp because tempdir may not be /tmp on all devices,
- IMHO this logic should live above Starboard in the Cobalt layer wrapper.
- The file should also not be just stored in tmpdir because tmpdir system-wide readable for all processes, and we don't know if the umask is set to block others from reading the file, and this information should probably no tbe exposed to every process on the device. A good alternative is probably `mkstemps` (we can add the implementation from musl).

PR #7309, opened 2025-09-24
starboard: Remove not-used StarboardBridge::OnStop method 

kjyoun:
This change removes the `onStop` native method and the corresponding JNI function. whic are not used.

Bug: 150410605


johnxwork (approved):

starboard/android/shared/starboard_bridge.cc
```diff
@@ -73,11 +73,6 @@ jboolean JNI_StarboardBridge_InitJNI(
   return true;
 }
 
-void JNI_StarboardBridge_OnStop(JNIEnv* env) {
-  SbAudioSinkImpl::TearDown();
```
johnxwork, line 77: I actually think this function's usage is lost during iterations. 

In C25, this was part of the application shutdown: https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/android/shared/application_android.cc;drc=655b3e166675fcb74a12b2f52a37c4e79149fc44;l=202

Would you add onStop native call to StarboardBridge::onActivityStop()?

Really nice catch.

kjyoun: I think that a person who is familiar with the Android lifecycle would be a better person for this task.

I filed b/447434347

PR #7311, opened 2025-09-24
[Rebase m138] migrate callers to getNormalizedHeader overload

madhurajayaraman:
Bug: 418842688

Chromium change:
Migrate all callers to optional-based GetNormalizedHeader overload | https://chromium-review.googlesource.com/c/chromium/src/+/5939082

Change based on edits in file : https://chromium-review.googlesource.com/c/chromium/src/+/5939082/9/chrome/browser/enterprise/connectors/device_trust/navigation_throttle.cc

### DIFF
--- a/cobalt/browser/cobalt_secure_navigation_throttle.cc
+++ b/cobalt/browser/cobalt_secure_navigation_throttle.cc
@@ -94,8 +94,7 @@ CobaltSecureNavigationThrottle::EnforceCSPHeaders() {
         content::NavigationThrottle::CANCEL, net::ERR_BLOCKED_BY_CLIENT);
   }
 
-  if (std::optional<std::string> CSP_value = 
-        response_headers->GetNormalizedHeader("Content-Security-Policy")) {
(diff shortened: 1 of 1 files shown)


cobalt/browser/cobalt_secure_navigation_throttle.cc
```diff
@@ -95,8 +94,8 @@ CobaltSecureNavigationThrottle::EnforceCSPHeaders() {
         content::NavigationThrottle::CANCEL, net::ERR_BLOCKED_BY_CLIENT);
   }
 
-  if (response_headers->GetNormalizedHeader("Content-Security-Policy",
-                                            &CSP_value)) {
+  if (std::optional<std::string> CSP_value = 
+        response_headers->GetNormalizedHeader("Content-Security-Policy")) {
```
andrewsavage1: I assume this function before returned true on success and now returns a value for the std::optional on success?

It doesn't seem like we need a variable CSP_value anymore though, and can just check `if (response_headers->GetNormalizedHeader(...))` (I don't see CSP_value used below)

PR #7317, opened 2025-09-24
[Rebase m138] Remove profanity masking

madhurajayaraman:
Bug: 418842688

Chromium change: https://chromium-review.googlesource.com/c/chromium/src/+/5651681


kaidokert (approved):
uh, i guess profanity is okay again in 2025

PR #7324, opened 2025-09-25
Deprecate SbCPUFeaturesGet

y4vor:
The functionality is migrated to using getauxval() which we alrady support.

Issue: 437416109


yell0wd0g (approved):
RS LGTM since this seems like a pure removal.

andrewsavage1 (approved):

PR #7325, opened 2025-09-25
[Rebase m138] Get shell_permission_manager up to date

madhurajayaraman:
Bug: 418842688

Chromium changes:
https://chromium-review.googlesource.com/c/chromium/src/+/4756727
https://chromium-review.googlesource.com/c/chromium/src/+/5498921
https://chromium-review.googlesource.com/c/chromium/src/+/5896593
https://chromium-review.googlesource.com/c/chromium/src/+/6438921
https://chromium-review.googlesource.com/c/chromium/src/+/6387077
https://chromium-review.googlesource.com/c/chromium/src/+/4746465
https://chromium-review.googlesource.com/c/chromium/src/+/5545382

Does not include https://chromium-review.googlesource.com/c/chromium/src/+/5154368
Changes to IsAllowlistedPermissionType are spread over many CLs, not linking them here.

### DIFF
--- a/cobalt/shell/browser/shell_permission_manager.cc
+++ b/cobalt/shell/browser/shell_permission_manager.cc
@@ -47,11 +47,6 @@ bool IsAllowlistedPermissionType(PermissionType permission) {
 
     case PermissionType::IDLE_DETECTION:
 
-    // Storage Access API web platform tests require permission to be granted by
-    // default.
-    case PermissionType::STORAGE_ACCESS_GRANT:
-    case PermissionType::TOP_LEVEL_STORAGE_ACCESS:
-
     // WebNFC browser tests require permission to be granted by default.
     case PermissionType::NFC:
       return true;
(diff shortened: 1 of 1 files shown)


cobalt/shell/browser/shell_permission_manager.cc
```diff
@@ -67,17 +61,28 @@ bool IsAllowlistedPermissionType(PermissionType permission) {
...
     case PermissionType::WAKE_LOCK_SYSTEM:
+    case PermissionType::HAND_TRACKING:
     case PermissionType::VR:
     case PermissionType::AR:
+    case PermissionType::STORAGE_ACCESS_GRANT:
     case PermissionType::CAMERA_PAN_TILT_ZOOM:
     case PermissionType::WINDOW_MANAGEMENT:
     case PermissionType::LOCAL_FONTS:
     case PermissionType::DISPLAY_CAPTURE:
+    case PermissionType::TOP_LEVEL_STORAGE_ACCESS:
```
andrewsavage1, line 72: It's worth noting in the description that the changes you're making here don't relate to the CLs you linked and are spread out in a number of CLs (at least, that's what I saw)—I don't think you need to link each of those CLs because they're often just a single line of change, but it's worth mentioning that there are changes outside of the CLs you note

andrewsavage1 (approved):

cobalt/shell/browser/shell_permission_manager.cc
madhurajayaraman: Updated

PR #7329, opened 2025-09-25
android: Handle undocumented exceptions in getOutputFormat()

kjyoun:
The `getOutputFormat()` method in Android's `MediaCodec` API seems to throw various undocumented exceptions on specific devices due to OS fragmentation and OEM customizations. 

This commit broadens the exception handling in `MediaCodecBridge.java` to catch `RuntimeException`. This is a more robust approach that prevents a wider range of undocumented, device-specific exceptions from propagating to the native layer and causing a crash.

Bug: 445694177


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -713,7 +713,7 @@ private void getOutputFormat(GetOutputFormatResult outGetOutputFormatResult) {
     int status = MediaCodecStatus.OK;
     try {
       format = mMediaCodec.get().getOutputFormat();
-    } catch (IllegalStateException e) {
+    } catch (Exception e) {
```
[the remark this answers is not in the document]
kjyoun: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -713,7 +713,7 @@ private void getOutputFormat(GetOutputFormatResult outGetOutputFormatResult) {
     int status = MediaCodecStatus.OK;
     try {
       format = mMediaCodec.get().getOutputFormat();
-    } catch (IllegalStateException e) {
+    } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
kjyoun: I would keep RuntimeException

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -421,6 +421,11 @@ FrameSize MediaCodecBridge::GetOutputSize() {
   JNIEnv* env = AttachCurrentThread();
   Java_MediaCodecBridge_getOutputFormat(env, j_media_codec_bridge_,
                                         j_reused_get_output_format_result_);
+  if (Java_GetOutputFormatResult_status(
+          env, j_reused_get_output_format_result_) == MEDIA_CODEC_ERROR) {
+    SB_LOG(WARNING) << "GetOutputSize() failed, returning empty size.";
+    return {{0, 0}};
```
johnxwork: I'm not familiar with codec and not sure about the implication of returning zeros here. Would you also ask for a media member to double check?

kjyoun: sure will do.

BTW, this code returns the same value as the existing does.

Existing code "eventually" return empty size, since JNI methods below returns either 0 or -1, when output format result is null

[code](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java;l=160-202;drc=8a59f5f35487b2a60c73032e474aaea9c6576b88)

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -421,6 +421,11 @@ FrameSize MediaCodecBridge::GetOutputSize() {
   JNIEnv* env = AttachCurrentThread();
   Java_MediaCodecBridge_getOutputFormat(env, j_media_codec_bridge_,
                                         j_reused_get_output_format_result_);
+  if (Java_GetOutputFormatResult_status(
```
borongc: It seems like we didn't check if `j_reused_get_output_format_result_format` is nullptr or not.

Chromium does this:
1. Return null if it has errors: https://source.chromium.org/chromium/chromium/src/+/main:media/base/android/java/src/org/chromium/media/MediaCodecBridge.java;l=724
2. if the format is null, handle it: https://source.chromium.org/chromium/chromium/src/+/main:media/base/android/media_codec_bridge_impl.cc;l=531

kjyoun: I agree, the correct long-term fix is to return an error/null here and have the callsite handle it.

For now, I've added a TODO to track this.

Making that change is too large and out of scope for this PR, especially since this is a targeted fix which will land on 26.android release branch.

borongc: It seems we only call this to check if the format is valid, so using
```
if (Java_GetOutputFormatResult_format(env, j_reused_get_output_format_result_)) {
    SB_LOG(WARNING) << "GetOutputSize() failed, returning empty size.";
    return {};
}
```
should be sufficient?

And, I agree we should handle error case and report to H5 player as Chromium does.

kjyoun: I reverted this change to make this PR focus on fixing the original cause.

I added comments on `MediaCodecBridge.getOutputFormat` about why I changed it to `RuntimeException`

We can handle return value of `getOutputSize` in a separate PR.

borongc (approved):

PR #7332, opened 2025-09-25
[Rebase m138] Fixes for shell_platform_delegate.o

madhurajayaraman:
Bug: 418842688

Chromium changes:
chromium-review.googlesource.com/c/chromium/src/+/5729956
chromium-review.googlesource.com/c/chromium/src/+/5272353
https://chromium-review.googlesource.com/c/chromium/src/+/4918014
https://chromium-review.googlesource.com/c/chromium/src/+/5791853
https://chromium-review.googlesource.com/c/chromium/src/+/5561092
https://chromium-review.googlesource.com/c/chromium/src/+/5017280
https://chromium-review.googlesource.com/c/chromium/src/+/5148491
https://chromium-review.googlesource.com/c/chromium/src/+/5670973
https://chromium-review.googlesource.com/c/chromium/src/+/6090650

### DIFF
--- a/cobalt/shell/browser/shell.cc
+++ b/cobalt/shell/browser/shell.cc
@@ -647,7 +647,9 @@ bool Shell::IsBackForwardCacheSupported(WebContents& web_contents) {
   return true;
 }
 
-PreloadingEligibility Shell::IsPrerender2Supported(WebContents& web_contents) {
+PreloadingEligibility Shell::IsPrerender2Supported(
+  WebContents& web_contents,
+  PreloadingTriggerType trigger_type) {
   return PreloadingEligibility::kEligible;
 }
 
--- a/cobalt/shell/browser/shell.h
+++ b/cobalt/shell/browser/shell.h
@@ -172,7 +172,8 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
                       const blink::mojom::FileChooserParams& params) override;
   bool IsBackForwardCacheSupported(WebContents& web_contents) override;
   PreloadingEligibility IsPrerender2Supported(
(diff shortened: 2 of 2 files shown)


cobalt/shell/browser/shell.cc
```diff
@@ -401,8 +402,11 @@ gfx::NativeWindow Shell::window() {
 }
 #endif
 
-WebContents* Shell::OpenURLFromTab(WebContents* source,
-                                   const OpenURLParams& params) {
+WebContents* Shell::OpenURLFromTab(
+    WebContents* source,
+    const OpenURLParams& params,
+    base::OnceCallback<void(content::NavigationHandle&)>
+        navigation_handle_callback) {
```
andrewsavage1, line 409: Shouldn't we also change L452-454 as is done in https://chromium-review.googlesource.com/c/chromium/src/+/4918014 to use navigation_handle_callback?

andrewsavage1 (commented):
I don't think chromium-review.googlesource.com/c/chromium/src/+/5272353 is relevant?

madhurajayaraman:
> I don't think chromium-review.googlesource.com/c/chromium/src/+/5272353 is relevant?

Relevant for deleting the string_piece.h header in shell.h . It's so minimal but ultimately it is a break so included the PR.

cobalt/shell/browser/shell.cc
madhurajayaraman: You're totally right, updated.

madhurajayaraman: Added the fix to the right commit & rebased so hopefully it's straightforward to review!

PR #5735, opened 2025-05-12
Build starboard loader_app for evergreen raspi builds

niranjanyardi:
b/389972301

Add changes to support building starboard loader_app on raspi

PR #7346, opened 2025-09-26
system_egl: Set eglGetPlatformDisplay on all cases

rakuco:
Follow-up to #5735, which added a check for `EGL_VERSION_1_5` without an `#else` clause, leading to not all fields of `SbEglInterface` being set in the non-Android, non-EGL 1.5 case (i.e. `eglWaitSync` was never set because all fields past `eglGetPlatformDisplay` were off-by-one).

Bug: 389972301
Bug: 447478802


rakuco:

rakuco:
@sherryzy could you take a look at this one? Thanks!

rakuco:
@yell0wd0g could you merge this one?

PR #5950, opened 2025-06-02
Add EGL 1.4 fallback for eglGetPlatformDisplay in nplb tests

sherryzy:
Context:
Some Cobalt platforms, such as the Raspberry Pi 2, support EGL 1.4 but not EGL 1.5. The nplb test suite fails on these devices because it attempts to use `eglGetPlatformDisplay()`, a function that is only standardized in EGL 1.5.

This change introduces a compile-time workaround to ensure backward compatibility with EGL 1.4. It adds a preprocessor check for `EGL_VERSION_1_5`. If EGL 1.5 is not defined, the code falls back to using `eglGetPlatformDisplayEXT()`, which is available through the `EGL_EXT_platform_base` extension in EGL 1.4.

Fixed: 420988131

PR #5468, opened 2025-04-18
Fix nplb SbPlayerTest*

sherryzy:
Fixes tests failing due to a missing EGL display. This change explicitly sets the display attribute for Angle to ensure proper initialization when a default display cannot be automatically found.

b/402459390

PR #7347, opened 2025-09-26
FakeGraphicsContextProvider: Clean up display initialization code in InitializeEGL()

rakuco:
The logic added in #5468, #5950 and #6438 was not working as expected:

* This file does not include the EGL headers directly, so nothing ever sets
  `EGL_VERSION_1_5` and the EGL 1.4 code path was always being used.
* Use of `eglGetPlatformDisplay()` with ANGLE parameters only makes use when
  ANGLE is used, which at the moment is the case only for linux-x64x11.
  On platforms that do not use ANGLE but do provide
  `eglGetPlatformDisplayEXT()`, the call was failing because ANGLE
  attributes were being used.

The new logic is simpler and should work with Android, linux-x64x11, Raspi2,
RDK and other platforms:

* Remove all checks for `EGL_VERSION_1_5`.
* Only call `eglGetPlatformDisplay()` when ANGLE is in use, as all the
  platform being passed and its attributes are ANGLE-specific at the moment.
* Call `eglGetDisplay()` in all other cases (i.e. Android, Raspi2, RDK etc).
  This was already the case for Android; Raspi2 does not have
  `eglGetPlatformDisplayEXT()` so it was always doing this too; RDK has a
  custom `SbEglGetDisplay()` that tries to use `eglGetPlatformDisplayEXT()`
  when it is available (without the wrong ANGLE parameters that were being
  passed here).

Note that other platforms that do not use ANGLE but have EGL 1.5 could also
call `eglGetPlatformDisplay()`, but they need to pass the right attributes
and a valid platform value to the function.

Also note that at the moment raspi2 and RDK skip all NPLB tests using this
class.

---

This PR contains 2 commits but since they are probably going to be squashed, some explanation about the changes to `starboard/egl_and_gles` are in order:

* Clean up list of valid `gl_type` values in the documentation as well as in
  the validation (i.e. "glimp" and "none" were either not documented or not
  allowed).
* Define the `egl_and_gles_angle` group, there is no need to define it
  conditionally.
* Set the `STARBOARD_GL_TYPE_{ANGLE,SYSTEM_GLES2}` macros depending on
  `gl_type`'s value. This is useful when determining whether we can use
  ANGLE values (in FakeGraphicsContextProvider, for example).

---

Bug: 402459390
Bug: 420988131
Bug: 432259527
Fixed: 447076985
Fixed: 447478802

### DIFF
--- a/starboard/testing/fake_graphics_context_provider.cc
+++ b/starboard/testing/fake_graphics_context_provider.cc
@@ -177,8 +177,8 @@ void FakeGraphicsContextProvider::InitializeEGL() {
   // - Android only has eglGetDisplay()
   // - Raspberry Pi 2 has EGL 1.4 without platform extensions so only
   // eglGetDisplay() is available.
-  // - RDK has a custom SbEglGetDisplay() implementation that tries to use
-  // eglGetPlatformDisplayEXT() and falls back to eglGetDisplay() when
+  // - RDK has a custom SbEglInterface::eglGetDisplay implementation that tries
+  // to use eglGetPlatformDisplayEXT() and falls back to eglGetDisplay() when
   // necessary.
   //
   // Note that other platforms that do not use ANGLE but have EGL 1.5 could


rakuco:
@andrewsavage1 for the GN bits, @sherryzy for the EGL bits, @yell0wd0g for everything :-)

starboard/egl_and_gles/BUILD.gn
```diff
@@ -23,16 +23,25 @@
 # in the configuration.gni for the current platform.
 
 group("egl_and_gles") {
-  assert(gl_type != "none", "gl_type must be set to a valid value")
+  _valid_gl_type_values = [
+    "angle",
+    "system_gles2",
+  ]
+  assert(filter_include([ gl_type ], _valid_gl_type_values) != [],
```
andrewsavage1: Keeping it simple with an `assert(gl_type == "angle" || gl_type == "system_gles2")` would be clearer

rakuco: Done!

rakuco:
The linux-modular and evergreen-x64 failures were real. For some reason, in a linux-modular build the `//starboard/egl_and_gles:defines` config was not being propagated to `starboard_group` even though it was in `starboard`'s `all_dependent_configs`. The only reason I can think of is that these configs are not propagated when different toolchains are in use.

For now, I've cheated and just started setting `STARBOARD_GL_TYPE_*` in `//starboard/build/config:starboard` that a ton of targets already depend on.

andrewsavage1:
>  The only reason I can think of is that these configs are not propagated when different toolchains are in use.

That's correct, see `propagates_configs` in https://gn.googlesource.com/gn/+/master/docs/reference.md#functions-and-variables

> For now, I've cheated and just started setting `STARBOARD_GL_TYPE_*` in `//starboard/build/config:starboard` that a ton of targets already depend on.

Is it possible to make these defines a public_config as part of the `//starboard/egl_and_gles` target and have the correct targets depend on it, or are there too many transitive dependencies? public_configs provides a more clear config structure than all_dependent_configs so is generally preferred

rakuco:
> Is it possible to make these defines a public_config as part of the `//starboard/egl_and_gles` target and have the correct targets depend on it

These definitions should be independent from the main `egl_and_gles` target: on linux-x64x11, for example, depending on this target will make e.g. `nplb` include ANGLE's headers and link against ANGLE rather than deferring usage of an actual EGL/GLES library to the starboard platform implementation.

What could be done is make each target that needs to use those macros depend on the config target; I only went for adding them by default because I thought most people wouldn't know this is necessary in the first place.

rakuco:
> What could be done is make each target that needs to use those macros depend on the config target; I only went for adding them by default because I thought most people wouldn't know this is necessary in the first place.

I've ended up going for another approach: adding a buildflag header with the `gl_type` settings so that other files can include it and check for `BUILDFLAG(STARBOARD_GL_TYPE_{ANGLE,SYSTEM_GLES2})`. Using this requires depending on the new buildflag target, so the situation I was trying to avoid above won't happen.

starboard/egl_and_gles/BUILD.gn
```diff
@@ -22,17 +22,26 @@
...
+  assert(gl_type == "angle" || gl_type == "system_gles2")
   public_deps = [ ":egl_and_gles_$gl_type" ]
 }
 
-group("egl_and_gles_system_gles2") {
-  # Use the system-provided implementation of GLES2.
+buildflag_header("buildflags") {
+  header = "buildflags.h"
+
+  _enable_angle = gl_type == "angle"
```
andrewsavage1: No need to prepend an underscore here and below, it should already be scoped to this target

rakuco: Done!

rakuco:
@sherryzy @yell0wd0g could you take a look at this one?

starboard/testing/fake_graphics_context_provider.cc
```diff
@@ -190,40 +168,35 @@ void FakeGraphicsContextProvider::RunLoop() {
...
+  // ANGLE is known to implement EGL 1.5 (and therefore
+  // eglGetPlatformDisplay()), and when it is used eglGetPlatformDisplay()
+  // needs to be called instead of eglGetDisplay() anyway because the right
+  // attributes must be specified.
+  //
+  // When ANGLE is not used, eglGetDisplay() is always used:
+  // - Android only has eglGetDisplay()
+  // - Raspberry Pi 2 has EGL 1.4 without platform extensions so only
+  // eglGetDisplay() is available.
+  // - RDK has a custom SbEglGetDisplay() implementation that tries to use
```
yell0wd0g: nit: `SbEglGetDisplay()`? IIUC here we're inside Starboard
and `SbBla()` functions are only provided to clients of
Starboard.

rakuco: It's [a custom implementation of `SbEglInterface::eglGetDisplay` that happens to be called `SbEglGetDisplay()`](https://cobalt.googlesource.com/external/components/generic/cobalt/+/137d5f3214aea87d80d65b3ea3dbebd13856f38f/src/third_party/starboard/rdk/shared/system/system_egl.cc#138). Should I explain this in more detail?

rakuco: Done just in case.

yell0wd0g: Thanks! Beat me to it.

PR #7348, opened 2025-09-26
Add strip tool to all Evergreen toolchains

andrewsavage1:
Fixed: 447437870

### DIFF
--- a/build/toolchain/linux/BUILD.gn
+++ b/build/toolchain/linux/BUILD.gn
@@ -8,6 +8,11 @@ import("//build/config/sysroot.gni")
 import("//build/config/ui.gni")
 import("//build/toolchain/gcc_toolchain.gni")
 
+if (is_cobalt_hermetic_build && is_official_build &&
+    current_toolchain == default_toolchain) {
+  import("//starboard/build/toolchain/linux/strip.gni")
+}
+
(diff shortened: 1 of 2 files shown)


build/toolchain/linux/BUILD.gn
```diff
@@ -21,13 +26,21 @@ clang_toolchain("clang_arm") {
     current_cpu = "arm"
     current_os = "linux"
   }
+  if (is_cobalt_hermetic_build && is_official_build &&
+      current_toolchain == default_toolchain) {
```
hlwarriner, line 30: Why do we need the `current_toolchain == default_toolchain` condition here?

I noticed we don't when assigning `strip` for the "clang_x64" toolchain.

andrewsavage1: The clang_x64 toolchain under //cobalt is only a cobalt toolchain, while these arm toolchains can technically be used as host toolchains (or other secondary toolchains) in the Chromium config. So it would be fine to put `current_toolchain == default_toolchain` in the `//cobalt` `clang_x64` one, just redundant

hlwarriner: Sounds good - thanks.

y4vor (approved):

PR #7349, opened 2025-09-26
Add Cobalt configuration for ICU

haozheng-cobalt:
`make_data_all.sh` generates the icudtl.dat file from the ICU source text files in a temporary directory third_party/icu/data/out/tmp, then it calls `copy_data.sh` to move this temporary file to a permanent source-controlled location, which is `third_party/icu/cobalt/icudtl.dat` for Cobalt. Then `third_party/icu/BUILD.gn` copies over  `third_party/icu/cobalt/icudtl.dat` to out folder when Cobalt builds.

I do not find Cobalt Linux is building ICU rn, this PR is considering ATV only.
This PR creates the Cobalt's configuration for ICU reusing the existing Android configuration; `third_party/icu/filters/cobalt.json` is copied from `third_party/icu/filters/android.json`, so that we could customize Cobalt's configuration without tampering other platform in the next step.

Recommend run this shell commands rather than running `make_data_all.sh` directly, because the sh file runs for all platforms. https://paste.googleplex.com/4824386649587712

Test: this PR creates a same icudtl.dat file for Cobalt as before, doesn't change behavior. 
Bug: 446966279

### DIFF
--- a/.github/config/linux.json
+++ b/.github/config/linux.json
@@ -9,7 +9,6 @@
   "targets": [
     "cobalt:gn_all",
     "chromedriver",
-    "crashpad_handler",
     "dump_syms",
     "minidump_stackwalk"
   ],
--- a/.github/workflows/gemini-commit-message.yml
+++ b/.github/workflows/gemini-commit-message.yml
@@ -107,9 +107,18 @@ jobs:
       # This step now uses the direct response from the Gemini step.
       - name: 'Post Commit Message as Comment'
         uses: actions/github-script@v7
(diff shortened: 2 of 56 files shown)


johnxwork:
The code change looks simple and benign. But the comments are missing and commit message is unclear.
- In the PR description, can you clarify if you are stating the behavior for before- or after- this PR?
- Why do we need to add the two new output files to the source tree?
- Do we need to update these output artifacts for e.g. every icu text file update?

PR #7353, opened 2025-09-26
[Rebase M138] Avoid unused variable warning in GpuInit [upstreamable]

sideb0ard:
Bug: 418842688

Add conditional block around variable declaration to match usage.

### DIFF
--- a/gpu/ipc/service/gpu_init.cc
+++ b/gpu/ipc/service/gpu_init.cc
@@ -891,9 +891,7 @@ bool GpuInit::InitializeAndStartSandbox(base::CommandLine* command_line,
       std::move(supported_buffer_formats_for_gl_native_pixmap_import);
   [[maybe_unused]] auto* factory =
       ui::OzonePlatform::GetInstance()->GetSurfaceFactoryOzone();
-#if BUILDFLAG(IS_COBALT)
-  [[maybe_unused]] bool filter_set = false;
-#else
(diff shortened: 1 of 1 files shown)


gpu/ipc/service/gpu_init.cc
```diff
@@ -891,7 +891,11 @@ bool GpuInit::InitializeAndStartSandbox(base::CommandLine* command_line,
       std::move(supported_buffer_formats_for_gl_native_pixmap_import);
   [[maybe_unused]] auto* factory =
       ui::OzonePlatform::GetInstance()->GetSurfaceFactoryOzone();
+#if BUILDFLAG(IS_COBALT)
```
kaidokert: This looks .. not great just to avoid an unused var. Maybe we can do an override in GN in the local module ( e.g. under gpu/ipc/*gn* or gpu/ipc/service/* , or perhaps better to use 

```
#if !BUILDFLAG(ENABLE_VULKAN) && !BUILDFLAG(SKIA_USE_DAWN) 
```

rather here - the latter would be upstreamable ?

_globally_ disabling unused variables in all our compilations in a central GN file is also an option, but it's a risky/bad option.

kaidokert (approved):
Okay - now this change looks upstreamable as we don't mention Cobalt anywhere. Could you put a CL up for Chromium ?

Or, again - label this PR for sending upstream

sideb0ard:
> Okay - now this change looks upstreamable as we don't mention Cobalt anywhere. Could you put a CL up for Chromium ?
> 
> Or, again - label this PR for sending upstream

will do, thanks!

PR #7356, opened 2025-09-26
Turn off the plugin support

y4vor:
This is a cleanup to use the gn
variable instead of guarding the code.

Issue: 409755808

### DIFF
--- a/cobalt/build/configs/linux_common.gn
+++ b/cobalt/build/configs/linux_common.gn
@@ -110,3 +110,6 @@ use_dbus = false
 
 # Embed ICU data in the cobalt binary
 icu_use_data_file = false
+
+# Disable the Pepper API (PPAPI) plugin support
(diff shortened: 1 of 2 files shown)


ppapi/buildflags/buildflags.gni
```diff
@@ -16,7 +16,7 @@ declare_args() {
 
 declare_args() {
   # Enables Pepper API (PPAPI) plugin support.
-  enable_ppapi = enable_plugins && !is_fuchsia
+  enable_ppapi = enable_plugins && !is_fuchsia && !is_starboard
```
andrewsavage1: Can't we set this via linux_common.gn instead?

y4vor: Yeah, that should work too. I'll move it there.

andrewsavage1 (approved):

PR #7358, opened 2025-09-26
Prune crash reports as required by cobalt

hlwarriner:
This implements the changes to our report pruning approach described in go/crashpad-db-mgmt. CrashReportUploadThread from upstream is used to prune reports, but based on our own total database size (and report age) requirements and only after pending reports are processed.

As part of the implementation, go/crashpad-db-mgmt suggested moving ownership of the prune thread from the main crashpad handler thread to the upload thread. But it turned out to be simpler - and have a smaller customization footprint - to keep the prune thread as a ScopedStoppable instantiated by the main thread and to leverage an existing callback parameter for the upload thread to trigger report pruning after it processes and attempts to upload pending reports.

The threading model, combined with ScopedStoppable, should allow pruning to complete before the handler process exits.

Issue: 446889385
Issue: 406511608

### DIFF
--- a/starboard/crashpad_wrapper/wrapper.cc
+++ b/starboard/crashpad_wrapper/wrapper.cc
@@ -230,7 +230,15 @@ void InstallCrashpadHandler(const std::string& ca_certificates_path) {
   std::map<std::string, std::string> default_annotations = {
       {kCrashpadVersionKey, kCrashpadVersion},
       {kCrashpadProductKey, product_name}};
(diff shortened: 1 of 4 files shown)


hlwarriner:
@y4vor @yuying-y, Andrew already approved this but please let me know if you'd like to review and need more time. I'll plan to merge tomorrow morning if I don't hear anything.

yuying-y (approved):

PR #7360, opened 2025-09-26
BACKPORT: Don't set dirty bits for attribs that are out of range.

sherryzy:
This CL backports https://chromium-review.googlesource.com/c/angle/angle/+/5008034 to address an out-of-bounds access bug.

I did some minor modifications to resolve conflict. Some of the upstream code has been moving around, for example FrontendFeatures_autogen.h is in include/platform/autogen/FrontendFeatures_autogen.h in the upstream. And in Context.cpp, it calls caps->maxVertexAttributes and there are some dependency on other PR, I keep the minimum change, and fix it by directly call mState.mCaps.maxVertexAttributes.

Bug: 437918195

Original commit description:
PrivateState::setAllDirtyBits sets all bits in mDirtyCurrentValues. When the context has fewer max attibutes than MAX_VERTEX_ATTRIBS, this can cause out-of-bounds access to
PrivateState::mVertexAttribCurrentValues if the dirty bits are iterated over without range validation.

### DIFF
--- a/third_party/angle/include/platform/FrontendFeatures_autogen.h
+++ b/third_party/angle/include/platform/FrontendFeatures_autogen.h
@@ -116,6 +116,13 @@ struct FrontendFeatures : FeatureSetBase
     FeatureInfo disableProgramCaching = {"disableProgramCaching", FeatureCategory::FrontendFeatures,
                                          "Disables saving programs to the cache", &members,
                                          "http://anglebug.com/1423136"};
+
+    FeatureInfo forceMinimumMaxVertexAttributes = {
+        "forceMinimumMaxVertexAttributes",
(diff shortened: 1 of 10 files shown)


sherryzy:
There are some minor modifications to resolve conflicts. Updated the PR description to reflect this.

yell0wd0g:
> I did some minor modifications to resolve conflict. Some of the upstream code has been moving around, for example FrontendFeatures_autogen.h is in include/platform/autogen/FrontendFeatures_autogen.h in the upstream. And in Context.cpp, it calls caps->maxVertexAttributes and there are some dependency on other PR, I keep the minimum change, and fix it by directly call mState.mCaps.maxVertexAttributes.

Thanks! Still LGTM.

PR #7364, opened 2025-09-26
[Rebase M138] Update media::mojo:TypeConvertor

sideb0ard:
Bug: 418842688

DecoderBuffer Mojo transport was split into EOS and data types via a Union:
https://chromium-review.googlesource.com/c/chromium/src/+/5918375

### DIFF
--- a/media/mojo/common/media_type_converters.cc
+++ b/media/mojo/common/media_type_converters.cc
@@ -170,17 +170,16 @@ TypeConverter<scoped_refptr<media::DecoderBuffer>,
     return media::DecoderBuffer::CreateEOSBuffer();
   }
 
+  const auto& mojo_buffer = input->get_data();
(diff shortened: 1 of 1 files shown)


media/mojo/common/media_type_converters.cc
```diff
@@ -175,8 +175,9 @@ TypeConverter<scoped_refptr<media::DecoderBuffer>,
   // a new DecoderBuffer. Note that DecoderBuffer is released
   // here as its ref-count was increased manually to ensure
   // media thread won't release it before MojoRenderer has it.
+  const auto& mojo_data_buffer = input->get_data();
```
borongc: Should we move L183 before L173 to save this line (L178)?

sideb0ard: done, thanks!

PR #7367, opened 2025-09-26
Remove unnecessary media flags, to save binary size

niranjanyardi:
Issue: 437151722

For 3P hermetic builds disable unnecessary media flags to save 2MB of the compressed lz4 binary

1. enable_libaom = false 
2. enable_dav1d_decoder = false
3. media_use_libvpx = false 
4. media_use_openh264 = false

### DIFF
--- a/cobalt/build/configs/linux-x64x11-modular/args.gn
+++ b/cobalt/build/configs/linux-x64x11-modular/args.gn
@@ -21,3 +21,15 @@ enable_library_cdms = false
 # is respected with LTO.
 is_cfi = false
 use_thin_lto = false
+
+# Disables AV1 encoding/decoding using libaom.
+enable_libaom = false
+
+# Disable AV1 decoding using dav1d.
+enable_dav1d_decoder = false
+
+# Disables VP8/VP9 encoding/decoding using libvpx.
+media_use_libvpx = false
+
+# Disables H.264 encoding/decoding using OpenH264.
+media_use_openh264 = false
--- a/cobalt/build/configs/linux_common.gn
+++ b/cobalt/build/configs/linux_common.gn
@@ -110,15 +110,3 @@ use_dbus = false
 
 # Embed ICU data in the cobalt binary
 icu_use_data_file = false
-
-# Disables AV1 encoding/decoding using libaom.
-enable_libaom = false
-
-# Disable AV1 decoding using dav1d.
-enable_dav1d_decoder = false
-
-# Disables VP8/VP9 encoding/decoding using libvpx.
-media_use_libvpx = false
-
-# Disables H.264 encoding/decoding using OpenH264.
-media_use_openh264 = false


niranjanyardi:
Ran into this error:

https://paste.googleplex.com/5834710261563392

niranjanyardi:
last time i checked the hermetic checks were mostly passing - idk if the tests will run 

it will be good if the reviewers can atleast do a intial review - i mostly added these flags as Miguel suggested here: https://b.corp.google.com/issues/437151722#comment4

I'm not very familiar with the 3P media pipeline.

Further more fel free to make changes yourself - as this is an experimental branch - you can push to it

yell0wd0g (commented):
You should be able to remove a few more, and
also for all platforms, not just linux, see
#7053 (I didn't land it due to upstream unit
tests failing).

niranjanyardi:
> You should be able to remove a few more, and also for all platforms, not just linux, see #7053 (I didn't land it due to upstream unit tests failing).

I talked to @borongc and he mentioned that if any chromium tests are failing because of this change- this is because of some issue in upstream chromium 
I would like to decouple the binary size changes with the chromium media files analysis
1) That means i can disable the tests here and the media team can work on re-enabling them 
2) Otherwise I'll pretty much give up on this PR and just disable one flag at a time slowly 

I prefer 1) over 2) as we can get the binary size win quickly - it's also something i want to move on from . 1) does create more tech debt than 2) - but atleast that is tracked in the disabled tests - we can work on re-enabling them slowly

An example of a failure is:
[ RUN      ] SupportedTypesTest.IsSupportedVideoTypeBasics
../../media/base/supported_types_unittest.cc:44: Failure
Value of: IsSupportedVideoType( {VideoCodec::kVP9, VP9PROFILE_PROFILE0, kUnspecifiedLevel, kColorSpace})
  Actual: false
Expected: true
[  FAILED  ] SupportedTypesTest.IsSupportedVideoTypeBasics (0 ms)
[----------] 1 test from SupportedTypesTest (0 ms total)

jasonzhangxx:
These are chromium flags. By disabling them, I guess some third party libraries are excluded so that binary size is reduced. It's weird that disabling chromium flags causes chromium test failures. 

For the example of failure:
[ RUN ] SupportedTypesTest.IsSupportedVideoTypeBasics
../../media/base/supported_types_unittest.cc:44: Failure
Value of: IsSupportedVideoType( {VideoCodec::kVP9, VP9PROFILE_PROFILE0, kUnspecifiedLevel, kColorSpace})
Actual: false
Expected: true
[ FAILED ] SupportedTypesTest.IsSupportedVideoTypeBasics (0 ms)
[----------] 1 test from SupportedTypesTest (0 ms total)
By disabling media_use_libvpx, chromium may lose the capability to support vp9, so the expected result of `IsSupportedVideoType( {VideoCodec::kVP9, VP9PROFILE_PROFILE0, kUnspecifiedLevel, kColorSpace})` is likely to be false, though in the test it expects `true`. 

For 3P, cobalt linux starboard reference implementation needs libvpx, libaom, libdav1d and libopenh264.

PR #7371, opened 2025-09-29
Add build and migration instructions for EAP

andrewsavage1:
Bug: 436853146

### DIFF
--- a/starboard/doc/eap/25_to_26.md
+++ b/starboard/doc/eap/25_to_26.md
@@ -18,7 +18,7 @@ Below are the high level steps for what is needed when migrating a platform:
     Using the appropriate OS and CPU for your target platform and setting the starboard path to the directory containing your port (i.e. the starboard_path for the linux-x64x11 reference platform is `starboard/linux/x64x11`).
 * Must update Starboard functions (this will mostly just involve getting rid of deprecated functions, see [starboard/CHANGELOG.md](../../../starboard/CHANGELOG.md), but you may need to modify wrappers under `//starboard/shared/modular/` if NPLB test fail).
 * Some Chromium settings may cause issues. If you run into any of these issues, please contact the Cobalt team so we can make the changes upstream.
-* Validate with NPLB test suite and by running Cobalt.
+* Validate with NPLB test suite and by running Cobalt. Note that NPLB should be run via an Evergreen loader instead of directly as an executabe.
 * Validate with running YTS. Ideally aim for parity with C25 results
 
 ## Building and Running


starboard/doc/eap/25_to_26.md
```diff
@@ -0,0 +1,10 @@
+# Migrating a platform
+
+We use RDK as our reference platform, and performed the migration from Cobalt 25 to Cobalt 26 support for it as one of the first platforms. You can find the vast majority of the changes we needed the make in the following changelist: https://cobalt.googlesource.com/external/components/generic/cobalt/+/d7236a995f023bba1c74ad95602f460c2fa72297.
+
+Below are the high level steps for what is needed when migrating a platform:
+* Must update Starboard functions (this will mostly just involve getting rid of deprecated functions, see [starboard/CHANGELOG.md](../../../starboard/CHANGELOG.md), but you may need to modify wrappers under `//starboard/shared/modular/` if NPLB test fail).
+* Some Chromium settings may cause issues. Use GN args to change these when possible, but if you need to change upstream we can talk about submitting those changes in Cobalt
```
y4vor: I think we should ask partners to report any Chromium settings issues to us and make the changes only upstream.

starboard/doc/eap/open_source_install.patch
```diff
@@ -0,0 +1,23 @@
+diff --git a/build/install-build-deps.sh b/build/install-build-deps.sh
```
y4vor: Can't we checkin these changes in build/install-build-deps.sh? 
Why do we need a patch?

starboard/doc/eap/25_to_26.md
```diff
@@ -0,0 +1,10 @@
...
+* Add an args.gn file for your platform. You should likely inherit from linux-x64x11-modular/args.gn by importing it with `import("//cobalt/build/configs/linux-x64x11-modular/args.gn")`.
+* Validate with NPLB test suite and by running Cobalt.
```
y4vor: I think we should add instructions for how to build and run for their platform e.g.

# Build Evergreen for your platform
cobalt/build/gn.py -p evergreen-arm-softfp-<platform> -c devel --no-rbe
autoninja -C out/evergreen-arm-softfp-<platform>_devel/ cobalt_loader nplb_loader

# Package files 
TODO... 

# Run Cobalt on your platform
./loader_app

# Run NPLB
./elf_loader_sandbox --evergreen_content=. --evergreen_libary=libnplb.so
```

File: starboard/doc/eap/25_to_26.md (hunk 1 above)
Comment (lamdbui) [reply starboard/doc/eap/25_to_26.md]: We can follow-up with this with a survey as development goes underway on more 3P platforms.

File: starboard/doc/eap/25_to_26.md (hunk 3 above)
Comment (lamdbui) [reply starboard/doc/eap/25_to_26.md]: Is the only way to run `nplb` via the `elf_loader_sandbox` now? If so, let's make that clear that they should not be using the standalone monolithic `nplb` build for validation.

File: starboard/doc/eap/getting_started.md [hunk 4]
```diff
@@ -0,0 +1,53 @@
+# Open Source Setup
```
lamdbui, line 1: Can we also link to reference port documentation in perhaps another section?

We have RDK here and can also link to Raspi-2 when available:
- https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/site/docs/development/setup-rdk.md

lamdbui, line 1: Can we also link to reference port documentation in perhaps another section?

We have RDK here and can also link to Raspi-2 when available:
- https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/site/docs/development/setup-rdk.md

starboard/doc/eap/open_source_install.patch
andrewsavage1: It's machine-dependent which is why I was hesitant to check them in, but it's probably fine to do so

starboard/doc/eap/25_to_26.md
andrewsavage1: Right now we have the build setup so that you need to run nplb via an evergreen loader (and this is our preference because we want to validate that the loader actually works correctly as well), but it would be possible to change it so that we allow `nplb` to build monolithically if that's necessary for some reason. Right now, it shouldn't be possible to build `nplb` monolithically unless the partner actually changes GN build files, but I can still make a note that they should be using a loader for the tests.

starboard/doc/eap/getting_started.md
andrewsavage1: Are the RDK instructions for C25 something we want to link here?

lamdbui: I think we'll want it eventually for reference, but I realize those are the C25-based ones and we can add C26-based instructions later.

PR #7397, opened 2025-09-29
[Rebase M138] Update SystemTrustStoreStarboard

sideb0ard:
Bug: 418842688

Upstream changes:
* PR which adds use of BoringSSL
https://chromium-review.googlesource.com/c/chromium/src/+/5001690

* PR which adds GetPlatformTrustStore() 
Add new interface PlatformRootStore:
https://chromium-review.googlesource.com/c/chromium/src/+/5667088

* PR which adds GetChromeRootConstraints(..)
Add Chrome Root Store constraints, starting with SCT-based constraints.
https://chromium-review.googlesource.com/c/chromium/src/+/5278551

* PR which adds eutl_trust_store()
Implement 1-QWAC verification
https://chromium-review.googlesource.com/c/chromium/src/+/6332054

* PR which adds IsLocallyTrustedRoot(..)
Allow locally trusted root certificates to override CRS constraints
https://chromium-review.googlesource.com/c/chromium/src/+/5567490

### DIFF
--- a/net/cert/internal/system_trust_store_starboard.cc
+++ b/net/cert/internal/system_trust_store_starboard.cc
@@ -37,15 +37,20 @@ class SystemTrustStoreChromeOnly : public SystemTrustStore {
   }
 
   net::PlatformTrustStore* GetPlatformTrustStore() override { return nullptr; }
+
   bool IsLocallyTrustedRoot(
       const bssl::ParsedCertificate* trust_anchor) override {
     return false;
   }
+
   base::span<const ChromeRootCertConstraints> GetChromeRootConstraints(
       const bssl::ParsedCertificate* cert) const override {
-    return {};
+    return trust_store_chrome_->GetConstraintsForCert(cert);
   }
-  bssl::TrustStore* eutl_trust_store() override { return nullptr; }
+
+  bssl::TrustStore* eutl_trust_store() override {
+    return trust_store_chrome_->eutl_trust_store();
+  }
 
  private:
   std::unique_ptr<TrustStoreChrome> trust_store_chrome_;


net/cert/internal/system_trust_store_starboard.cc
```diff
@@ -1,3 +1,4 @@
+
```
andrewsavage1: remove

net/cert/internal/system_trust_store_starboard.cc
```diff
@@ -28,14 +29,26 @@ class SystemTrustStoreChromeOnly : public SystemTrustStore {
...
+  bool IsKnownRoot(const bssl::ParsedCertificate* trust_anchor) const override {
     return trust_store_chrome_->Contains(trust_anchor);
   }
-  int64_t chrome_root_store_version() override {
+  int64_t chrome_root_store_version() const override {
     return trust_store_chrome_->version();
   }
+
+  // Stubs for methods not supported in Starboard.
+  net::PlatformTrustStore* GetPlatformTrustStore() override { return nullptr; }
```
andrewsavage1, line 39: These don't seem to be related to the change you linked in the PR description

net/cert/internal/system_trust_store_starboard.cc
sideb0ard: done

net/cert/internal/system_trust_store_starboard.cc
sideb0ard: Updated the PR description with the additional upstream changes.

net/cert/internal/system_trust_store_starboard.cc
```diff
@@ -28,14 +28,25 @@ class SystemTrustStoreChromeOnly : public SystemTrustStore {
...
+  int64_t chrome_root_store_version() const override {
     return trust_store_chrome_->version();
   }
+
+  net::PlatformTrustStore* GetPlatformTrustStore() override { return nullptr; }
+  bool IsLocallyTrustedRoot(
+      const bssl::ParsedCertificate* trust_anchor) override {
+    return false;
+  }
+  base::span<const ChromeRootCertConstraints> GetChromeRootConstraints(
```
andrewsavage1, line 46: I don't think we should be implementing this here, other platform-specific trust stores do not override the default implementation

net/cert/internal/system_trust_store_starboard.cc
```diff
@@ -28,14 +28,25 @@ class SystemTrustStoreChromeOnly : public SystemTrustStore {
...
+  bool IsKnownRoot(const bssl::ParsedCertificate* trust_anchor) const override {
     return trust_store_chrome_->Contains(trust_anchor);
   }
-  int64_t chrome_root_store_version() override {
+  int64_t chrome_root_store_version() const override {
     return trust_store_chrome_->version();
   }
+
+  net::PlatformTrustStore* GetPlatformTrustStore() override { return nullptr; }
+  bool IsLocallyTrustedRoot(
```
andrewsavage1, line 41: This is implemented in net/cert/internal/system_trust_store.cc, are you sure we need to override that? It would deviate from other platforms

net/cert/internal/system_trust_store_starboard.cc
```diff
@@ -28,14 +28,25 @@ class SystemTrustStoreChromeOnly : public SystemTrustStore {
...
+  bssl::TrustStore* GetTrustStore() override { return trust_store_chrome_.get(); }
+  bool IsKnownRoot(const bssl::ParsedCertificate* trust_anchor) const override {
     return trust_store_chrome_->Contains(trust_anchor);
   }
-  int64_t chrome_root_store_version() override {
+  int64_t chrome_root_store_version() const override {
     return trust_store_chrome_->version();
   }
+
+  net::PlatformTrustStore* GetPlatformTrustStore() override { return nullptr; }
```
andrewsavage1, line 39: I don't think we should be overriding the default behavior in net/cert/internal/system_trust_store.cc?

net/cert/internal/system_trust_store_starboard.cc
```diff
@@ -28,14 +28,25 @@ class SystemTrustStoreChromeOnly : public SystemTrustStore {
...
+  bool IsLocallyTrustedRoot(
+      const bssl::ParsedCertificate* trust_anchor) override {
+    return false;
+  }
+  base::span<const ChromeRootCertConstraints> GetChromeRootConstraints(
+      const bssl::ParsedCertificate* cert) const override {
+    return {};
+  }
+  bssl::TrustStore* eutl_trust_store() override { return nullptr; }
```
andrewsavage1: same as other cases, I would expect to use the default

andrewsavage1 (changes requested):

PR #7402, opened 2025-09-29
Finch Config Expiration

jammel-yeboah:
This change introduces a mechanism to expire a Finch config if it hasn't been updated within a server-configurable number of days. This prevents devices that have been offline for extended periods from using a potentially stale or problematic configuration.

An expired config is treated as an empty config, causing the device to fall back to its default, compiled-in feature flags.

**Implementation details:**
- This entire feature is gated by a new feature flag, `kExperimentConfigExpiration`
- The expiration period is not hardcoded. The logic reads the variations_expiration_threshold_days integer from the `kFinchParameters` dictionary, which is populated by the existing H5vcc `setFinchParameters` API. If the server does not send this parameter, the expiration check is disabled by default. If the server does not send this parameter, the expiration check is disabled by default.
- The expiration check relies on the `kVariationsLastFetchTime` timestamp, adopted from the existing Chrome Variations infrastructure
- To gather data on config age before activating the feature, two UMA metrics are logged on every startup, regardless of the feature flag's state:
  - `Cobalt.Finch.ConfigAgeInDays`: A custom count of the config's age in days (1-90).
  - `Cobalt.Finch.ConfigState`: An enum indicating the config's status (`kValid`, `kExpired`, `kMissingTimestamp`).

Bug: 407734088

### DIFF
--- a/tools/metrics/histograms/metadata/cobalt/histograms.xml
+++ b/tools/metrics/histograms/metadata/cobalt/histograms.xml
@@ -56,8 +56,8 @@ Always run the pretty print utility on this file after editing:
 </histogram>
 
 <histogram name="Cobalt.Finch.ConfigAgeInDays" units="days"
-    expires_after="2025-01-01">
-  <owner>jammel@google.com</owner>
+    expires_after="never">
+  <owner>yijiazhang@google.com</owner>
   <summary>
     Records the age of the Finch configuration in days each time it is
     checked at startup. This data is used to determine an appropriate
@@ -66,8 +66,8 @@ Always run the pretty print utility on this file after editing:
 </histogram>
 
 <histogram name="Cobalt.Finch.ConfigState" enum="VariationsConfigState"
-    expires_after="2025-01-01">
-  <owner>jammel@google.com</owner>
+    expires_after="never">
+  <owner>yijiazhang@google.com</owner>
   <summary>
     Records the state of the Finch configuration at startup. This is used
     to monitor the frequency of expired or invalid configurations.


cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -80,6 +88,56 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
...
+  base::TimeDelta config_age = base::Time::Now() - fetch_time;
+
+  UMA_HISTOGRAM_CUSTOM_COUNTS("Cobalt.Finch.ConfigAgeInDays",
+                              config_age.InDays(), 1, 90, 50);
+
+  // Get the expiration threshold from the server config. Default to 30 days.
+  const int expiration_threshold_in_days =
+      experiment_prefs->GetDict(kFinchParameters)
+          .FindInt("variations_expiration_threshold_days")
+          .value_or(30);
```
AlecsisDuarte: nit: Could you define this value as a constant at the top of the file for visibility and easier to change if needed?

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -80,6 +88,56 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
...
+                              VariationsConfigState::kMissingTimestamp);
+    return false;
+  }
+
+  base::Time fetch_time =
+      experiment_prefs->GetTime(variations::prefs::kVariationsLastFetchTime);
+  base::TimeDelta config_age = base::Time::Now() - fetch_time;
+
+  UMA_HISTOGRAM_CUSTOM_COUNTS("Cobalt.Finch.ConfigAgeInDays",
+                              config_age.InDays(), 1, 90, 50);
```
AlecsisDuarte: super-nit: Should we define the custom count MACRO to something like "UMA_HISTOGRAM_DAYS_ELAPSED"? Mostly to add notations around the values range? No need if this is widely known on the cobalt side!

AlecsisDuarte (approved):

cobalt/browser/cobalt_content_browser_client.cc
jammel-yeboah: > super-nit: Should we define the custom count MACRO to something like "UMA_HISTOGRAM_DAYS_ELAPSED"? Mostly to add notations around the values range? No need if this is widely known on the cobalt side!

I noticed that other uses of the macro leave it as it is so we can keep it like that for consistency. Also, if we decide to change the value range to hours, weeks, etc, we wouldn't have to redefine the macro. Maybe a comment clarifying that the value range is in days should suffice

cobalt/browser/features.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
jammel-yeboah, line 1: Created this file to define global features for Cobalt because it doesn't seem like we have one but please lmk if I'm wrong and one exists

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -407,6 +486,16 @@ void CobaltContentBrowserClient::CreateFeatureListAndFieldTrials() {
       ->metrics_services_manager()
       ->InstantiateFieldTrialList();
 
+  const bool config_expired = HasConfigExpired();
+  if (IsVariationsConfigExpirationEnabled() && config_expired) {
+    // Config is expired, so we proceed without applying it.
+    // Set up an empty feature list and return early.
```
yjzhang111: This block can be moved to SetUpCobaltFeaturesAndParams(). One reason for it is HasConfigExpired() reads from disk and stores feature/feature params. If executed in SetUpCobaltFeaturesAndParams(), we can re-use it.
config_expired is essentially the same as applying an empty config. We can combine the two logic instead of instantiating an empty feature list here.

yjzhang111 (changes requested):
Have you done any manual testing? Could you briefly mention here what kind of tests you've done

cobalt/browser/features.cc
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cobalt/browser/features.h"
+
+namespace cobalt {
+namespace features {
+
+// Enables the variations config expiration check.
+BASE_FEATURE(kVariationsConfigExpiration,
```
yjzhang111: ExperimentConfig might be a better name to so it could be consistent with the references in the rest of the files (e.g. global_features.cc) in cobalt/browser

cobalt/browser/features.h
```diff
@@ -0,0 +1,29 @@
...
+#ifndef COBALT_BROWSER_FEATURES_H_
+#define COBALT_BROWSER_FEATURES_H_
+
+#include "base/feature_list.h"
+
+namespace cobalt {
+namespace features {
+
+// Enables the variations config expiration check.
+extern const base::Feature kVariationsConfigExpiration;
```
yjzhang111: Have you tried testing this feature? I think you might need BASE_DECLARE_FEATURE macro.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -81,6 +85,81 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
 constexpr base::FilePath::CharType kTrustTokenFilename[] =
     FILE_PATH_LITERAL("Trust Tokens");
 
+// An enum for the UMA histogram to track the state of the variations config.
+// These values are persisted to logs. Entries should not be renumbered and
+// numeric values should never be reused.
+enum class VariationsConfigState {
```
yjzhang111: I think it's better to move these changes to a file like experiment_config_manager.cc, including the following two expiration related functions. They don't have much to do with the lifecycle of the browser client.

jammel-yeboah:
> Have you done any manual testing? Could you briefly mention here what kind of tests you've done

**Test 1:**

- Goal: Ensure we record last fetch time and add it to the config
- How: Launch cobalt, send config via h5vcc API from browser console:
    
    `await window.h5vcc.experiments.setExperimentState({"features":{"FinchTestFeature":true},"featureParams":{"FinchTestFeatureParam":1},"experimentIds":[113,223]})`
    
- Result: We successfully record last fetch time (`variations_last_fetch_time`), as seen by reading from cache on disk:
  
       
image

      
    

**Test 2:**

- Goal: Ensure we don’t apply expired config when `kVariationsConfigExpiration` feature is ON
- How: Set `kVariationsConfigExpiration` to `ENABLED_BY_DEFAULT` in code, manually set `expiration_threshold_in_days` to `-1` so that every new config is effectively considered expired
- Result:
    - `IsVariationsConfigExpirationEnabled()` evaluates to true, we enter `HasConfigExpired`  function, we log that the config is expired, and we set an empty feature list
    
    
image (1)

    

**Test 3:**

- Goal: Ensure we still apply config even if it’s "expired", because `kVariationsConfigExpiration` feature is OFF
- How: Set `kVariationsConfigExpiration` to `DISABLED_BY_DEFAULT` in code, manually set `expiration_threshold_in_days` to -1 so that every config is considered expired
- Result: Config still applied
    - No log output of expired config getting ignored, we set the feature list and apply the config

cobalt/browser/features.h
jammel-yeboah: How is the BASE_DECLARE_FEATURE macro different/better in this case? I tested the current implementation (see other comment) and it seems to be working fine.

cobalt/browser/features.cc
jammel-yeboah: Agree, I'll update it

cobalt/browser/cobalt_content_browser_client.cc
jammel-yeboah: Makes sense, so essentially since  `SetUpCobaltFeaturesAndParams()` and `HasConfigExpired()` both read directly from the `PrefService`, we can combine that logic into one. I'll make the changes

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -30,13 +88,37 @@ ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
...
   if (num_crashes >= kCrashStreakEmptyConfigThreshold) {
-    return ExperimentConfigType::kEmptyConfig;
+    config_type = ExperimentConfigType::kEmptyConfig;
+  } else if (num_crashes >= kCrashStreakSafeConfigThreshold) {
+    config_type = ExperimentConfigType::kSafeConfig;
+  } else {
+    config_type = ExperimentConfigType::kRegularConfig;
+  }
+
+  // If the config is already empty, no need to check for expiration.
```
yjzhang111: Do you need this block? You can just return kEmptyConfig when checking for kCrashStreakEmptyConfigThreshold in the previous if block.

jammel-yeboah: True, I just wanted to make the logic clear and separated by setting the `config_type` in all three scenarios

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -30,13 +90,37 @@ ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
...
   }
-  if (num_crashes >= kCrashStreakSafeConfigThreshold) {
-    return ExperimentConfigType::kSafeConfig;
+
+  // Now, check if the expiration feature is enabled in the determined config.
+  const bool use_safe_config =
+      (config_type == ExperimentConfigType::kSafeConfig);
+  const base::Value::Dict& feature_map = experiment_config_->GetDict(
+      use_safe_config ? kSafeConfigFeatures : kExperimentConfigFeatures);
+  const bool expiration_enabled =
```
yjzhang111, line 112: Could you add a comment here explaining why we're accessing it directly and not using the standard IsEnabled function? Also, we probably don't want Cobalt behavior to change during an app lifecycle. Since it's reading the feature directly, there's a risk of that happening. Could you document it explicitly that it is only called before the featurelist is initialized?

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -30,13 +90,37 @@ ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
...
+      feature_map.FindBool(features::kExperimentConfigExpiration.name)
```
yjzhang111, line 113: Is there any benefit of defining this as a base::Feature instead of a string constant if we're only using the string name?

tools/metrics/histograms/metadata/cobalt/histograms.xml
```diff
@@ -55,6 +55,25 @@ Always run the pretty print utility on this file after editing:
   </summary>
 </histogram>
 
+<histogram name="Cobalt.Finch.ConfigAgeInDays" units="days"
+    expires_after="2025-01-01">
+  <owner>jammel@google.com</owner>
```
johnxwork: I expect this histogram to show up in next year, do we want Yijia to be the owner instead?

Also, is the expires_after date important in any way? It's expired already

jammel-yeboah: Yes I can transfer ownership to @yjzhang111 

Also, from my understanding the `expires_after` is important because that's when the UMA gets removed from the system. I'll fix the date to be something like 2027-01-01 instead

cobalt/browser/experiments/experiment_config_manager.cc
jammel-yeboah: This is just a specific case where we have to check the string directly on startup, but in the future, if we want to check that the feature is enabled after  `FeatureList` is initialized, we might want to use the `base::Feature` object like we do for all other features right? So we can probably keep it this way to make the code future-proof

tools/metrics/histograms/metadata/cobalt/histograms.xml
jammel-yeboah: Actually, just making the expiry `never` makes more sense. At least that's how other Cobalt histograms seem to be defined

PR #7403, opened 2025-09-29
[Rebase m138] Fixes for shell_content_browser_client

madhurajayaraman:
Bug: 418842688

Chromium changes are in commit messages.

### DIFF
--- a/cobalt/shell/browser/shell_content_browser_client.cc
+++ b/cobalt/shell/browser/shell_content_browser_client.cc
@@ -217,16 +217,6 @@ class ShellVariationsServiceClient
       PrefService* local_state) override {}
 };
 
-// Returns the full user agent string for the content shell.
-std::string GetShellFullUserAgent() {
-  std::string product = "Chrome/" CONTENT_SHELL_VERSION;
-  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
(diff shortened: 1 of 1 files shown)


cobalt/shell/browser/shell_speech_recognition_manager_delegate.cc
```diff
@@ -37,4 +37,9 @@ ShellSpeechRecognitionManagerDelegate::GetEventListener() {
   return nullptr;
 }
 
-}  // namespace content
+#if !BUILDFLAG(IS_FUCHSIA) && !BUILDFLAG(IS_ANDROID)
+void ShellSpeechRecognitionManagerDelegate::BindSpeechRecognitionContext(
+    mojo::PendingReceiver<media::mojom::SpeechRecognitionContext> receiver) {}
+#endif
+
+}  // namespace content
```
andrewsavage1, line 45: nit: add newline

andrewsavage1 (approved):
I think this is all fine, but it is admittedly hard to tell

madhurajayaraman:
> I think this is all fine, but it is admittedly hard to tell

I agree, and I'm the one who went through the Chromium history! I think for most of the shell folder, it'll make sense later to just have Hao/Adam/whoever has expertise in the area go through and look at the diff between us and 138. This is the minimum set of changes needed to make it build, and is not at all the actual set of changes we will need to be up to date.

PR #7404, opened 2025-09-30
android: Disabling esa commandLineArgs for gold builds

ckim1346:
As per [go/chrobalt-system-dev-dd](https://goto.google.com/chrobalt-system-dev-dd), devtools access should not be allowed in release builds. Adding a new versionInfo isReleaseBuild to check for release build. Couldn't use the existing isReleaseBuild() starboard function as it's not ready by the time createContent() is called. Checking in both createContent() and getArgs() disables esa commandLineArgs for release builds.

Bug: 446019286

### DIFF
--- a/components/version_info/android/BUILD.gn
+++ b/components/version_info/android/BUILD.gn
@@ -6,14 +6,10 @@ import("//build/config/android/channel.gni")
 import("//build/config/android/rules.gni")
 import("//build/util/process_version.gni")
 
-# This is a safe way to check for a variable that may not be defined in
-# all build configurations.
-_is_cobalt = false
 if (defined(is_cobalt)) {
-  _is_cobalt = is_cobalt
-}
-if (_is_cobalt) {
-  import("//cobalt/build/configs/cobalt.gni")
+  if (is_cobalt) {
+    import("//cobalt/build/configs/cobalt.gni")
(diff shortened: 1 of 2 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -103,7 +103,10 @@ protected void createContent(final Bundle savedInstanceState) {
     if (!CommandLine.isInitialized()) {
       CommandLine.init(null);
 
-      String[] commandLineArgs = getCommandLineParamsFromIntent(getIntent(), COMMAND_LINE_ARGS_KEY);
+      String[] commandLineArgs = null;
+      if (!VersionInfo.isReleaseBuild()) {
```
johnxwork, line 107: Are we able to just use `isReleaseBuild()` like below?

ckim1346: `isReleaseBuild()` isn't available from the StarboardBridge in `createContent()` yet. Calling it in `createContent()` causes the app to crash on start up

johnxwork (approved):

PR #7405, opened 2025-09-30
Remove os_definitions.gni

andrewsavage1:
This file can incorrectly define some variables that should only be defined by upstream GN code.

Also removes a few unused config targets.

Bug: 443797215


hlwarriner (approved):
LGTM as long as checks pass

PR #7408, opened 2025-09-30
[Rebase m138] Update shell_browser_context

madhurajayaraman:
Bug: 418842688

Chromium changes:
[tvOS] Set `switches::kContentShellUserDataDir` | https://chromium-review.googlesource.com/c/chromium/src/+/6480975
Code Health: Cleanup stale PersistentOriginTrials | https://chromium-review.googlesource.com/c/chromium/src/+/6082699
[FedCM] Move delegate implementation to content/browser | https://chromium-review.googlesource.com/c/chromium/src/+/5740871

### DIFF
--- a/cobalt/shell/browser/shell_browser_context.cc
+++ b/cobalt/shell/browser/shell_browser_context.cc
@@ -28,7 +28,7 @@
 #include "cobalt/shell/browser/shell_content_browser_client.h"
 #include "cobalt/shell/browser/shell_content_index_provider.h"
 #include "cobalt/shell/browser/shell_download_manager_delegate.h"
-#include "cobalt/shell/browser/shell_paths.h"
+#include "cobalt/shell/common/shell_paths.h"
 #include "cobalt/shell/browser/shell_permission_manager.h"
 #include "components/keyed_service/content/browser_context_dependency_manager.h"
 #include "components/keyed_service/core/simple_dependency_manager.h"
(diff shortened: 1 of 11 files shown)


cobalt/shell/browser/shell_browser_context.cc
```diff
@@ -83,24 +81,6 @@ void ShellBrowserContext::InitWhileIOAllowed() {
   if (cmd_line->HasSwitch(switches::kIgnoreCertificateErrors)) {
     ignore_certificate_errors_ = true;
   }
-  if (cmd_line->HasSwitch(switches::kContentShellDataPath)) {
```
andrewsavage1, line 86: We should follow https://chromium-review.googlesource.com/c/chromium/src/+/6480975 and add this code to shell_paths.cc

cobalt/shell/common/shell_paths.cc
andrewsavage1, line 1: Eventually we'll likely want to have this be as similar as we can to content/shell's shell_paths.cc, which would mean applying the other changes from https://chromium-review.googlesource.com/c/chromium/src/+/6480975 to it (including changing its parent directory)—maybe we should just do that now

andrewsavage1 (approved):

cobalt/shell/common/shell_paths.cc
madhurajayaraman: Updated!

PR #7410, opened 2025-09-30
[Rebase m138] Update shell_main_delegate

madhurajayaraman:
Bug: 418842688

Chromium changes:
Reland "[tracing] Cleanup up trace_log" | https://chromium-review.googlesource.com/c/chromium/src/+/6163823
Add reporting of statistics for AllocationTraceRecorder | https://chromium-review.googlesource.com/c/chromium/src/+/4566075
Reland "Allow child processes to inherit logging file handle on Windows" | https://chromium-review.googlesource.com/c/chromium/src/+/5373414
base/logging: Prevent test-only crash caused by std::string(nullptr) | https://chromium-review.googlesource.com/c/chromium/src/+/5416213

### DIFF
--- a/cobalt/shell/app/shell_main_delegate.cc
+++ b/cobalt/shell/app/shell_main_delegate.cc
@@ -116,8 +116,6 @@ void InitLogging(const base::CommandLine& command_line) {
 
   if (dest == LoggingDest::kFile) {
     settings.log_file_path = log_filename.value();
-  } else {
-    settings.log_file_path = nullptr;
   }
 
   if (dest == LoggingDest::kStderr) {


cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -82,26 +82,50 @@
...
+    }
   }
 
   logging::LoggingSettings settings;
-  settings.logging_dest = logging::LOG_TO_ALL;
-  settings.log_file_path = log_filename.value().c_str();
+
+  if (dest == LoggingDest::kFile) {
+    settings.log_file_path = log_filename.value();
+  }
```
andrewsavage1, line 119: Don't we need the else statement to set it to nullptr instead?

andrewsavage1: From https://chromium-review.googlesource.com/c/chromium/src/+/5373414

madhurajayaraman: yes and i am shocked i missed that seeing as i copy pasted the entire file??? good catch, ty, fixed now.

madhurajayaraman:
```
stderr:
../../cobalt/shell/app/shell_main_delegate.cc:120:30: error: null passed to a callee that requires a non-null argument [-Werror,-Wnonnull]
  120 |     settings.log_file_path = nullptr;
      |                              ^~~~~~~
1 error generated.
```

Weirdly I get a compile error when I have the nullptr assignment, which is extra odd because it's the same as Chromium? We probably have some build flag somewhere, any idea where?

PR #7421, opened 2025-10-01
[Rebase m138] fixes for cobalt_main_delegate

madhurajayaraman:
Bug: 418842688

Chromium changes:
Add reporting of statistics for AllocationTraceRecorder | https://chromium-review.googlesource.com/c/chromium/src/+/4566075
[tracing] Cleanup up trace_log | https://chromium-review.googlesource.com/c/chromium/src/+/6120057
[CodeHealth] Remove stale base::Feature EnableHangWatcherInZygoteChildren | https://chromium-review.googlesource.com/c/chromium/src/+/5987891
Enable HangWatcher crash reports for Canary/Dev without finch on Desktop platforms. | https://chromium-review.googlesource.com/c/chromium/src/+/5233337

### DIFF
--- a/cobalt/app/cobalt_main_delegate.cc
+++ b/cobalt/app/cobalt_main_delegate.cc
@@ -74,6 +74,10 @@ std::optional<int> CobaltMainDelegate::PostEarlyInitialization(
 
   InitializeHangWatcher();
 
+  const std::string process_type =
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
+          switches::kProcessType);
+
   // ShellMainDelegate has GWP-ASan as well as Profiling Client disabled.
   // Consequently, we provide no parameters for these two. The memory_system
   // includes the PoissonAllocationSampler dynamically only if the Profiling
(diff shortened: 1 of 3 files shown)


cobalt/app/cobalt_main_delegate.cc
```diff
@@ -147,7 +150,8 @@ void CobaltMainDelegate::InitializeHangWatcher() {
   } else {
     hang_watcher_process_type = base::HangWatcher::ProcessType::kUnknownProcess;
   }
+  const bool emit_crashes = false;
```
madhurajayaraman, line 153: https://chromium-review.googlesource.com/c/chromium/src/+/5233337
This is set to false since I wasn't sure what the right conditional would be here, relevant chromium change linked.

andrewsavage1: It looks like setting this to false maintains the same behavior as m114, but let's see if @yjzhang111 or @hlwarriner know

andrewsavage1: Let's not wait on this though, let's just log a follow-up bug and submit it

andrewsavage1 (approved):

cobalt/app/cobalt_main_delegate.cc
madhurajayaraman: Added it to the "notes" tab on the compile errors sheet of things to follow up on!

hlwarriner: > It looks like setting this to false maintains the same behavior as m114, but let's see if @yjzhang111 or @hlwarriner know

Yeah I agree with that.

@madhurajayaraman can you please just update the PR description? The actual change here doesn't match "Enable HangWatcher crash reports for Canary/Dev without finch on Desktop platforms."

andrewsavage1: Yeah, putting a line saying what we're doing here and referencing that we're going to follow up on it would be good (maybe we can make a go link for the tracking spreadsheet and mention that as well)

madhurajayaraman: @hlwarriner all of the change are copied from upstream, I've already made a note that we need to fix this change since the upstream had a lot of conditionals that I"m not sure we use. I'll assign a bug to you. b/448621078

PR #7422, opened 2025-10-01
Make kokoro pipeline work for evergreen-arm, evergreen-x64

niranjanyardi:
Issue: 445523026

This change adds Kokoro CI/CD pipeline configurations for Evergreen ARM
hardfp (Raspi and RDK variants) and Evergreen x64 platforms.
It introduces new package definitions to specify artifact content and
updates build scripts to correctly process these platforms.
Necessary third-party licenses for components used in these builds
are also registered.

The primary motivation is to enable automatic generation and storage
of build artifacts on GCS for these Evergreen targets, streamlining
their internal testing, deployment, and release workflows.

### DIFF
--- a/cobalt/build/evergreen-arm-hardfp-raspi/package.json
+++ b/cobalt/build/evergreen-arm-hardfp-raspi/package.json
@@ -7,6 +7,7 @@
                 "cobalt_shell.pak",
                 "loader_app",
                 "gen/build_info.json",
+                "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt"
             ],
             "dirs": [
@@ -18,6 +19,10 @@
                     "to_file": "build_info.json"
                 },
                 {
+                    "from_file": "gen/licenses_cobalt.spdx",
+                    "to_file": "licenses_cobalt.spdx"
+                },
+                {
                     "from_file": "gen/licenses_cobalt.txt",
                     "to_file": "licenses_cobalt.txt"
                 }
@@ -27,12 +32,17 @@
         {
             "files": [
                 "gen/build_info.json",
+                "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt"
             ],
             "rename_files": [
                 {
                     "from_file": "gen/build_info.json",
                     "to_file": "build_info.json"
+                },
+                {
+                    "from_file": "gen/licenses_cobalt.spdx",
+                    "to_file": "licenses_cobalt.spdx"
                 },
                 {
                     "from_file": "gen/licenses_cobalt.txt",
--- a/tools/licenses/licenses.py
+++ b/tools/licenses/licenses.py
@@ -395,7 +395,7 @@ SPECIAL_CASES = {
         "Name": "LLVM",
         "URL": "http://llvm.org",
         "License": "Apache 2.0 with LLVM exceptions",
-        "License File": "/third_party/llvm-project/compiler-rt/LICENSE.TXT",
+        "License File": "NOT_SHIPPED",
     },
     os.path.join('third_party', 'lz4_lib'): {
         "Name": "lz4",
--- a/cobalt/build/evergreen-arm-hardfp-rdk/package.json
+++ b/cobalt/build/evergreen-arm-hardfp-rdk/package.json
@@ -7,6 +7,7 @@
                 "cobalt_shell.pak",
                 "loader_app",
                 "gen/build_info.json",
+                "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt"
             ],
             "dirs": [
@@ -18,6 +19,10 @@
                     "to_file": "build_info.json"
                 },
                 {
+                    "from_file": "gen/licenses_cobalt.spdx",
+                    "to_file": "licenses_cobalt.spdx"
+                },
+                {
                     "from_file": "gen/licenses_cobalt.txt",
                     "to_file": "licenses_cobalt.txt"
                 }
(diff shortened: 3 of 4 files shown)


cobalt/build/evergreen-arm-hardfp-raspi/package.json
```diff
@@ -0,0 +1,45 @@
...
+            ],
+            "rename_files": [
+                {
+                    "from_file": "gen/build_info.json",
+                    "to_file": "build_info.json"
+                },
+                {
+                    "from_file": "gen/licenses_cobalt.txt",
+                    "to_file": "licenses_cobalt.txt"
+                }
```
niranjanyardi, line 50: i removed license_cobalt.spdx - do we need that for evergreen ? i thought licenses_cobalt.txt is enough

tools/licenses/licenses.py
```diff
@@ -383,6 +383,30 @@
...
+        "Name": "RDK",
+        "URL": "https://github.com/rdkcentral/rdk",
+        "License": "Apache 2.0",
+        "License File": "/starboard/contrib/rdk/LICENSE",
+    },
+    os.path.join('third_party', 'llvm-project'): {
+        "Name": "LLVM",
+        "URL": "http://llvm.org",
+        "License": "Apache 2.0 with LLVM exceptions",
+        "License File": "/third_party/llvm-project/compiler-rt/LICENSE.TXT",
```
niranjanyardi: there's actually another license file for libunwind: https://source.corp.google.com/h/github/youtube/cobalt/+/main:third_party/llvm-project/libunwind/LICENSE.TXT

it has the same license, is it okay to exclude it or should i add a duplicate entry for  os.path.join('third_party', 'llvm-project') - i guess i can add an entry for third_party', 'llvm-project', 'libunwind'

All these 4 exceptions were added because licenses.py complained - atleast the tool doesnt see llvm-project as 2 repos

niranjanyardi: @kaidokert  ptal

niranjanyardi:
i extracted the archive and ran it on my rdk - it worked

Here's a passing run: http://shortn/_zVW6QCzas8

cobalt/build/evergreen-arm-hardfp-raspi/package.json
briantting: we should support both, just to keep it consistent and available for future use. json > txt

arjungm (approved):

tools/licenses/licenses.py
kaidokert: As is, this doesn't look entirely correct. 

This is probably the right way to declare it:

```
os.path.join('third_party', 'llvm-project','compiler-rt`): {
```

And then libunwind would be 
```
os.path.join('third_party', 'llvm-project','libunwind`): {
```

We treat those as separate packages

niranjanyardi: Hmm, atleast the way we have added it is not great. Chromium expects the license file to be present in third_part/foo
```
if there's additional sub-repos with licenses they need to be added in a different set called ADDITIONAL_PATH
# A list of paths that contain license information but that would otherwise
# not be included.  Possible reasons include:
#   - Third party directories in //clank which are considered to be Google-owned
#   - Directories that are directly checked out from upstream, and thus
#     don't have a README.chromium
#   - Directories that contain example code, or build tooling.
#   - Nested third_party code inside other third_party libraries.
ADDITIONAL_PATHS
```

i wonder if this means we should add it like chromium - directly under third_party/compiler-rt:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/compiler-rt/
I'll create a bug : http://b/450339148 Consider adding libunwind, compiler-rt directly from chromium

niranjanyardi: Also the license tool currently expects something under third_party/llvm-project otherwise i get this error:
Failed to parse README.chromium: missing README.chromium or licenses.py SPECIAL_CASES entry in third_party/llvm-project

cobalt/build/evergreen-arm-hardfp-raspi/package.json
niranjanyardi: fixed, i think i had removed it because of an unrelated to spdx failure

tools/licenses/licenses.py
```diff
@@ -383,6 +385,30 @@
...
+        "Name": "RDK",
+        "URL": "https://github.com/rdkcentral/rdk",
+        "License": "Apache 2.0",
+        "License File": "/starboard/contrib/rdk/LICENSE",
+    },
+    os.path.join('third_party', 'llvm-project'): {
+        "Name": "LLVM",
+        "URL": "http://llvm.org",
+        "License": "Apache 2.0 with LLVM exceptions",
+        "License File": "NOT_SHIPPED",
```
niranjanyardi, line 398: Given (https://github.com/youtube/cobalt/pull/7422#discussion_r2403306859)  , i added "NOT_SHIPPED" here and added the 2 sub licenses under ADDTIONAL_PATHS - does that sound reasonable @kaidokert  ?

kaidokert: If it lets it get submitted, it seems okay. 

All of it seems a bit messy though, i would create a technical debt review / follow-up ticket and send it to @briantting to follow up and organize.

niranjanyardi: http://b/450339148 assigned to brian

niranjanyardi:
doing a test run here: http://shortn/_lbjV5y3Tjm

niranjanyardi:
> doing a test run here: http://shortn/_lbjV5y3Tjm

passed

niranjanyardi:
Note that i extracted the gcs artifacts locally and tested them on the RDK device-I could launch cobalt on RDK with the gcs archive files

PR #7424, opened 2025-10-01
[tvOS] Replace tvos-arm64 configuration with tvos-arm64-device

rakuco:
Follow-up to #7065 now that #7107 has landed:

* Use a suffix so that we now have both "-simulator" and "-device" configurations.
* Make `args.gn` include `tvos_common.gni` and behave like the tvos-arm64-simulator configuration.
* Add the missing bits to starboard/build and starboard/tvos that hook this configuration into the build.

Note that this is not usable yet: actually having a build that works requires more backports, so it is better to just wait for the work on M138 to start. These bits are independent from the Chromium milestone and can be landed now.

Co-authored-by: Julie Jeongeun Kim <jkim@igalia.com>

Bug: 442923940

### DIFF
--- a/cobalt/build/configs/tvos-arm64-device/args.gn
+++ b/cobalt/build/configs/tvos-arm64-device/args.gn
@@ -1,0 +1,4 @@
+import("//cobalt/build/configs/tvos_common.gn")
+
+target_cpu = "arm64"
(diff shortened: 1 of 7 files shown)


rakuco:

PR #7426, opened 2025-10-01
Add Evergreen softfp and arm64 builds

andrewsavage1:
Also consolidates the evergreen github jobs into a single file.

Bug: 448904392

### DIFF
--- a/.github/config/evergreen-arm-softfp.json
+++ b/.github/config/evergreen-arm-softfp.json
@@ -7,7 +7,6 @@
     "cobalt"
   ],
   "build_configs": [
-    "debug",
     "devel",
     "qa",
     "gold"


andrewsavage1:
I still need to add the workflows, but please take an initial look at the approach. We use a minimal "no_platform" platform to fake out the build

niranjanyardi:
> I still need to add the workflows, but please take an initial look at the approach. We use a minimal "no_platform" platform to fake out the build

1) Somehow we could build evergreen-arm-softp without any starboard platform in C25 - is it possible to copy that same mechanism here - like literally not building against any starboard file ? 

2) name seems a bit confusing: "no_platform" is still a starboard "platform" ..?

3) how would you prevent someone from adding more files/ adding more starboard content in this "no_platform" ?. like ideally evergreen-arm-softp is not "contaminated" with any starboard artifacts. Maybe adding a github owners group makes sense?

Dunno if @kaidokert has any more thoughts as we discussed this very briefly few days back

andrewsavage1:
> 1. Somehow we could build evergreen-arm-softp without any starboard platform in C25 - is it possible to copy that same mechanism here - like literally not building against any starboard file ?

The way we were doing in in c25 was quite different, because the loader platform build and evergreen builds were completely separate, making it easier when you don't have a loader platform like in this case and harder when you do have a loader platform. It's possible, but would require more involved changes.

> 2. name seems a bit confusing: "no_platform" is still a starboard "platform" ..?

Well, it's not a starboard platform. It has no Starboard implementations or configuration other than the minimum necessary to actually have the build work.

> 3. how would you prevent someone from adding more files/ adding more starboard content in this "no_platform" ?. like ideally evergreen-arm-softp is not "contaminated" with any starboard artifacts. Maybe adding a github owners group makes sense?

I don't think there's much point to this—we have hermetic builds for evergreen, which necessitate statically linked shared libraries as final build objects. I'm not sure what kind of "contamination" you're imagining.

niranjanyardi:
> > 1. Somehow we could build evergreen-arm-softp without any starboard platform in C25 - is it possible to copy that same mechanism here - like literally not building against any starboard file ?
> 
> The way we were doing in in c25 was quite different, because the loader platform build and evergreen builds were completely separate, making it easier when you don't have a loader platform like in this case and harder when you do have a loader platform. It's possible, but would require more involved changes.
> 
> > 2. name seems a bit confusing: "no_platform" is still a starboard "platform" ..?
> 
> Well, it's not a starboard platform. It has no Starboard implementations or configuration other than the minimum necessary to actually have the build work.
> 
> > 3. how would you prevent someone from adding more files/ adding more starboard content in this "no_platform" ?. like ideally evergreen-arm-softp is not "contaminated" with any starboard artifacts. Maybe adding a github owners group makes sense?
> 
> I don't think there's much point to this—we have hermetic builds for evergreen, which necessitate statically linked shared libraries as final build objects. I'm not sure what kind of "contamination" you're imagining.

Contamination - one thing i can think of is via defines - you define something in no_starboard headers which inadvertently gets pulled into chromium builds. configuration_public.h is one place something could be added.  it's a long shot

starboard/no_platform/configuration_public.h
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+// The Starboard configuration for a stub implementation designed to be
+// built on Desktop Linux. Other devices will have real Starboard
+// implementations.
+
+// Other source files should never include this header directly, but should
+// include the generic "starboard/configuration.h" instead.
+
+#ifndef STARBOARD_STUB_CONFIGURATION_PUBLIC_H_
```
y4vor, line 22: Are we doing mini_stub instead of fixing the stub platform?
Also what targets would use the headers here. Are we actually building things like elf_loader or nplb?

andrewsavage1: //starboard/common files include starboard/configuration.h which in turn includes this, though in my latest commit I make it as minimal as possible. We don't actually build elf_loader (though note we would build libnplb.so, as that's built with the cobalt/default toolchain), or other starboard level targets, so we should expect this file to have any purpose.

andrewsavage1:
> Contamination - one thing i can think of is via defines - you define something in no_starboard headers which inadvertently gets pulled into chromium builds. configuration_public.h is one place something could be added. it's a long shot

I left a comment in the file—I think using CODEOWNERS for this would be overkill

starboard/no_platform/configuration_public.h
andrewsavage1: As for no_platform vs using stub—we could use stub, but I'm trying to make it clear that this isn't supposed to build/run the starboard layer by having no Starboard implementation at all, not even stubs. We should probably separately make the stub platform build, but it seemed like a worse solution for this—wdyt?

y4vor: It is a bit strange. The no_platform is actually x64 architecture and we would use it to build the arm libraries. 
What happens if you remove configuration_pubilc.h or if you make it empty?

andrewsavage1: If it doesn't exist starboard/configuration.h won't be able to include it and won't have its required define of SB_IS_WCHAR_T_UTF32

PR #7429, opened 2025-10-02
android: Migrate VideoSurfaceTexture to JNI Generator

osagie98:
Migrates the VideoSurfaceTexture class to JNI Generator. A class VideoSurfaceTextureBridge is implemented to serve as the JNI interface between native code and VideoSurfaceTexture, and to ensure VideoSurfaceTexture methods are only called by a single class.

Bug: 417242469

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java
@@ -15,6 +15,7 @@
 package dev.cobalt.media;
 
 import android.graphics.SurfaceTexture;
+import android.view.Surface;
 import org.chromium.base.annotations.CalledByNative;
 import org.chromium.base.annotations.JNINamespace;
 import org.chromium.base.annotations.NativeMethods;
@@ -44,6 +45,21 @@ public class VideoSurfaceTexture extends SurfaceTexture {
   @CalledByNative
   void removeOnFrameAvailableListener() {
     super.setOnFrameAvailableListener(null);
+  }
+
+  @CalledByNative
+  static Surface createSurface(VideoSurfaceTexture surfaceTexture) {
+    return new Surface(surfaceTexture);
+  }
+
+  @CalledByNative
+  public void updateTexImage() {
+    super.updateTexImage();
+  }
+
+  @CalledByNative
+  public void getTransformMatrix(float[] mtx) {
+    super.getTransformMatrix(mtx);
   }
 
   @NativeMethods
--- a/starboard/android/shared/decode_target.cc
+++ b/starboard/android/shared/decode_target.cc
@@ -33,35 +33,6 @@ namespace starboard {
 namespace {
 
 using base::android::AttachCurrentThread;
-using base::android::ScopedJavaGlobalRef;
-using base::android::ScopedJavaLocalRef;
-
-ScopedJavaGlobalRef<jobject> CreateSurfaceTexture(JNIEnv* env,
-                                                  int gl_texture_id) {
-  ScopedJavaGlobalRef<jobject> surface_texture(
-      VideoSurfaceTextureBridge::CreateVideoSurfaceTexture(env, gl_texture_id));
-  SB_CHECK(surface_texture);
-
-  return surface_texture;
-}
-
-ScopedJavaGlobalRef<jobject> CreateSurfaceFromSurfaceTexture(
-    JNIEnv* env,
-    jobject surface_texture) {
-  ScopedJavaLocalRef<jclass> surface_class(
-      env, env->FindClass("android/view/Surface"));
-  SB_CHECK(surface_class);
-  jmethodID surface_constructor = env->GetMethodID(
-      surface_class.obj(), "<init>", "(Landroid/graphics/SurfaceTexture;)V");
-  SB_CHECK(surface_constructor);
-
-  ScopedJavaGlobalRef<jobject> surface(
-      env, env->NewObject(surface_class.obj(), surface_constructor,
-                          surface_texture));
-  SB_CHECK(surface);
-
-  return surface;
-}
 
 void RunOnContextRunner(void* context) {
   std::function<void()>* closure = static_cast<std::function<void()>*>(context);
@@ -109,11 +80,15 @@ void DecodeTarget::CreateOnContextRunner() {
 
   JNIEnv* env = AttachCurrentThread();
   // Wrap the GL texture in an Android SurfaceTexture object.
-  surface_texture_ = CreateSurfaceTexture(env, texture);
+  surface_texture_ =
+      VideoSurfaceTextureBridge::CreateVideoSurfaceTexture(env, texture);
+  SB_CHECK(surface_texture_);
 
   // We will also need an Android Surface object in order to obtain a
   // ANativeWindow object that we can pass into the AMediaCodec library.
-  surface_ = CreateSurfaceFromSurfaceTexture(env, surface_texture_.obj());
+  surface_ =
+      VideoSurfaceTextureBridge::CreateSurface(env, surface_texture_.obj());
+  SB_CHECK(surface_texture_);
 
   native_window_ = ANativeWindow_fromSurface(env, surface_.obj());
 
--- a/starboard/android/shared/video_decoder.cc
+++ b/starboard/android/shared/video_decoder.cc
@@ -28,6 +28,7 @@
 #include "build/build_config.h"
 #include "starboard/android/shared/media_common.h"
 #include "starboard/android/shared/video_render_algorithm.h"
+#include "starboard/android/shared/video_surface_texture_bridge.h"
 #include "starboard/common/check_op.h"
 #include "starboard/common/log.h"
 #include "starboard/common/media.h"
@@ -46,6 +47,7 @@ namespace starboard {
 namespace {
 
 using base::android::AttachCurrentThread;
+using base::android::JavaParamRef;
 using base::android::ScopedJavaLocalRef;
 using std::placeholders::_1;
 using std::placeholders::_2;
@@ -710,7 +712,8 @@ bool MediaCodecVideoDecoder::InitializeCodec(
       JNIEnv* env = AttachCurrentThread();
       ScopedJavaLocalRef<jobject> surface_texture(
           env, decode_target->surface_texture());
-      bridge_->SetOnFrameAvailableListener(env, surface_texture);
+      bridge_->SetOnFrameAvailableListener(
+          env, JavaParamRef<jobject>(env, surface_texture.obj()));
 
       std::lock_guard lock(decode_target_mutex_);
       decode_target_ = decode_target;
@@ -787,7 +790,8 @@ void MediaCodecVideoDecoder::TeardownCodec() {
       JNIEnv* env = AttachCurrentThread();
       ScopedJavaLocalRef<jobject> surface_texture(
           env, decode_target_->surface_texture());
-      bridge_->RemoveOnFrameAvailableListener(env, surface_texture);
+      bridge_->RemoveOnFrameAvailableListener(
+          env, JavaParamRef<jobject>(env, surface_texture.obj()));
 
       decode_target_to_release = decode_target_;
       decode_target_ = nullptr;
@@ -985,36 +989,11 @@ bool MediaCodecVideoDecoder::IsBufferDecodeOnly(
 
 namespace {
 
-jmethodID GetUpdateTexImageMethod(JNIEnv* env) {
-  jmethodID update_tex_image_method;
-
-  ScopedJavaLocalRef<jclass> surface_class(
-      env, env->FindClass("android/graphics/SurfaceTexture"));
-  SB_CHECK(surface_class);
-  update_tex_image_method =
-      env->GetMethodID(surface_class.obj(), "updateTexImage", "()V");
-  SB_CHECK(update_tex_image_method);
-
-  return update_tex_image_method;
-}
-
-jmethodID GetTransformMatrixMethod(JNIEnv* env) {
-  jmethodID get_transform_matrix_method;
-
-  ScopedJavaLocalRef<jclass> surface_class(
-      env, env->FindClass("android/graphics/SurfaceTexture"));
-  SB_CHECK(surface_class);
-  get_transform_matrix_method =
-      env->GetMethodID(surface_class.obj(), "getTransformMatrix", "([F)V");
-  SB_CHECK(get_transform_matrix_method);
-
-  return get_transform_matrix_method;
-}
-
 void updateTexImage(jobject surface_texture) {
   JNIEnv* env = AttachCurrentThread();
 
-  env->CallVoidMethod(surface_texture, GetUpdateTexImageMethod(env));
+  VideoSurfaceTextureBridge::UpdateTexImage(
+      env, JavaParamRef<jobject>(env, surface_texture));
 }
 
 void getTransformMatrix(jobject surface_texture, float* matrix4x4) {
@@ -1023,8 +1002,9 @@ void getTransformMatrix(jobject surface_texture, float* matrix4x4) {
   jfloatArray java_array = env->NewFloatArray(16);
   SB_CHECK(java_array);
 
-  env->CallVoidMethod(surface_texture, GetTransformMatrixMethod(env),
-                      java_array);
+  VideoSurfaceTextureBridge::GetTransformMatrix(
+      env, JavaParamRef<jobject>(env, surface_texture),
+      JavaParamRef<jfloatArray>(env, java_array));
 
   jfloat* array_values = env->GetFloatArrayElements(java_array, 0);
   memcpy(matrix4x4, array_values, sizeof(float) * 16);
--- a/starboard/android/shared/video_surface_texture_bridge.cc
+++ b/starboard/android/shared/video_surface_texture_bridge.cc
@@ -18,28 +18,54 @@
 
 namespace starboard {
 namespace {
+using base::android::JavaParamRef;
+using base::android::ScopedJavaGlobalRef;
 using base::android::ScopedJavaLocalRef;
 }  // namespace
 
 void VideoSurfaceTextureBridge::SetOnFrameAvailableListener(
     JNIEnv* env,
-    ScopedJavaLocalRef<jobject> surface_texture) const {
+    const JavaParamRef<jobject>& surface_texture) const {
   Java_VideoSurfaceTexture_setOnFrameAvailableListener(
       env, surface_texture, reinterpret_cast<jlong>(this));
 }
 
 void VideoSurfaceTextureBridge::RemoveOnFrameAvailableListener(
     JNIEnv* env,
-    ScopedJavaLocalRef<jobject> surface_texture) const {
+    const JavaParamRef<jobject>& surface_texture) const {
   Java_VideoSurfaceTexture_removeOnFrameAvailableListener(env, surface_texture);
 }
 
 // static
-ScopedJavaLocalRef<jobject>
+ScopedJavaGlobalRef<jobject>
 VideoSurfaceTextureBridge::CreateVideoSurfaceTexture(JNIEnv* env,
                                                      int gl_texture_id) {
   SB_CHECK(env);
-  return Java_VideoSurfaceTexture_Constructor(env, gl_texture_id);
+  return ScopedJavaGlobalRef<jobject>(
+      Java_VideoSurfaceTexture_Constructor(env, gl_texture_id));
+}
+
+// static
+ScopedJavaGlobalRef<jobject> VideoSurfaceTextureBridge::CreateSurface(
+    JNIEnv* env,
+    jobject surface_texture) {
+  return ScopedJavaGlobalRef<jobject>(Java_VideoSurfaceTexture_createSurface(
+      env, ScopedJavaLocalRef<jobject>(env, surface_texture)));
+}
+
+// static
+void VideoSurfaceTextureBridge::UpdateTexImage(
+    JNIEnv* env,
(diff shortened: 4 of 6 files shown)


starboard/android/shared/video_surface_texture_bridge.cc
```diff
@@ -0,0 +1,42 @@
...
+// limitations under the License.
+
+#include "starboard/android/shared/video_surface_texture_bridge.h"
+
+#include "cobalt/android/jni_headers/VideoSurfaceTexture_jni.h"
+
+namespace starboard {
+namespace {
+  using base::android::ScopedJavaLocalRef;
+}  // namespace
```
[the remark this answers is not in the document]
osagie98: The fully qualified names are used in the header file

starboard/android/shared/video_surface_texture_bridge.cc
```diff
@@ -0,0 +1,45 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/android/shared/video_surface_texture_bridge.h"
+
+#include "cobalt/android/jni_headers/VideoSurfaceTexture_jni.h"
+
+namespace starboard {
+namespace {
+using base::android::ScopedJavaLocalRef;
```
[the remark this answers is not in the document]
osagie98: It is, not useful

starboard/android/shared/video_decoder.cc
```diff
@@ -408,6 +407,10 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
     }
   }
 
+  if (output_mode_ == kSbPlayerOutputModeDecodeToTexture) {
+    bridge_.reset(new VideoSurfaceTextureBridge(this));
```
kjyoun: [go/totw/126](https://goto.google.com/totw/126)
> By default, use std::make_unique for dynamic allocation

```
bridge_ = std::make_unique<VideoSurfaceTextureBridge>(this);
```

starboard/android/shared/video_decoder.cc
```diff
@@ -408,6 +407,10 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
     }
   }
 
+  if (output_mode_ == kSbPlayerOutputModeDecodeToTexture) {
```
kjyoun: I think that you had better initialize through intialization list. then you can set `const` to bridge

````
video_decoder.cc
MediaCodecVideoDecoder::MediaCodecVideoDecoder(....)
    : video_codec_(video_stream_info.codec),
       ...
       bridge_(output_mode_ == kSbPlayerOutputModeDecodeToTexture ?
                           std::make_unique<VideoSurfaceTextureBridge>(this) :
                            nullptr) {



video_decoder.h

const std::unique_ptr<VideoSurfaceTextureBridge> bridge_;
````

[go/totw/labs/const#where](http://go/totw/labs/const#where)

> In short, you should use const whenever possible, as long as it makes sense to do so

File: cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java [hunk 5]
```diff
@@ -15,34 +15,41 @@
...
-  @UsedByNative
+  @CalledByNative
   void removeOnFrameAvailableListener() {
     super.setOnFrameAvailableListener(null);
   }
 
-  private native void nativeOnFrameAvailable(long nativeVideoDecoder);
+  @NativeMethods
+  interface Natives {
+    @NativeClassQualifiedName("VideoSurfaceTextureBridge")
```
kjyoun: you don't need `@NativeClassQualifiedName("VideoSurfaceTextureBridge")`, since parameter name already matches native class name (native"VideoSurfaceTextureBridge")

[@NativeClassQualifiedName("VideoSurfaceTextureBridge")](https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/jni_zero/README.md#usage)

starboard/android/shared/decode_target.cc
```diff
@@ -22,33 +22,60 @@
...
+using base::android::AttachCurrentThread;
+using base::android::ScopedJavaGlobalRef;
+using base::android::ScopedJavaLocalRef;
+
 jobject CreateSurfaceTexture(JNIEnv* env, int gl_texture_id) {
-  jobject local_surface_texture = JniNewObjectOrAbort(
-      env, "dev/cobalt/media/VideoSurfaceTexture", "(I)V", gl_texture_id);
+  jobject local_surface_texture =
+      VideoSurfaceTextureBridge::CreateVideoSurfaceTexture(env, gl_texture_id);
+  SB_CHECK(local_surface_texture);
```
kjyoun: How about using `ScopedJavaLocalRef`?

Then it's automatically destroyed. you dont' need to call `DeleteLocalRef`

```
jobject CreateSurfaceTexture(JNIEnv* env, int gl_texture_id) {
  ScopedJavaLocalRef<jobject> local_surface_texture(
      env, VideoSurfaceTextureBridge::CreateVideoSurfaceTexture(env, gl_texture_id));
  SB_CHECK(local_surface_texture.obj());

  jobject global_surface_texture = env->NewGlobalRef(local_surface_texture.obj());
  SB_CHECK(global_surface_texture); // It's also good to check the result of this call.

  return global_surface_texture;
}
````

File: starboard/android/shared/video_decoder.cc (hunk 4 above)
Comment (osagie98) [reply starboard/android/shared/video_decoder.cc]: Sure, though we don't need the bridge for punch out playback. Should I create the bridge anyway?

File: starboard/android/shared/video_decoder.cc (hunk 3 above)
Comment (osagie98) [reply starboard/android/shared/video_decoder.cc]: Noted

File: starboard/android/shared/video_decoder.cc (hunk 4 above)
Comment (osagie98) [reply starboard/android/shared/video_decoder.cc]: Updated it to be const and initialized in the initialization list

File: cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java (hunk 5 above)
Comment (osagie98) [reply cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java]: Done

File: starboard/android/shared/decode_target.cc (hunk 6 above)
Comment (osagie98) [reply starboard/android/shared/decode_target.cc]: Done

File: starboard/android/shared/video_decoder.cc [hunk 7]
```diff
@@ -378,7 +377,8 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
       is_video_frame_tracker_enabled_(IsFrameRenderedCallbackEnabled() ||
                                       tunnel_mode_audio_session_id != -1),
       has_new_texture_available_(false),
-      number_of_preroll_frames_(kInitialPrerollFrameCount) {
+      number_of_preroll_frames_(kInitialPrerollFrameCount),
+      bridge_(new VideoSurfaceTextureBridge(this)) {
```
kjyoun: With this code, we can create brige_ only when `output_mode_ == kSbPlayerOutputModeDecodeToTexture `

if not, bridge_ is set to null unique pointer.
and prefer `std:make_unique`

```
bridge_(output_mode_ == kSbPlayerOutputModeDecodeToTexture ?
                           std::make_unique<VideoSurfaceTextureBridge>(this) :
                            nullptr)
````

File: starboard/android/shared/video_decoder.cc (hunk 4 above)
Comment (kjyoun) [reply starboard/android/shared/video_decoder.cc]: I added a comment about this: https://github.com/youtube/cobalt/pull/7429#discussion_r2397351857

you can create bridge_ only when `output_mode_ == kSbPlayerOutputModeDecodeToTexture`

File: starboard/android/shared/video_decoder.cc (hunk 7 above)
Comment (osagie98) [reply starboard/android/shared/video_decoder.cc]: Done

File: starboard/android/shared/decode_target.cc [hunk 8]
```diff
@@ -22,33 +22,61 @@
...
+                         "(Landroid/graphics/SurfaceTexture;)V");
+    SB_CHECK(cache.surface_constructor);
+  });
+
+  ScopedJavaLocalRef<jobject> local_surface(
+      env, env->NewObject(cache.surface_class.obj(), cache.surface_constructor,
+                          surface_texture));
+  SB_CHECK(local_surface);
+
+  ScopedJavaGlobalRef<jobject> global_surface(env, local_surface.obj());
```
johnxwork: I know this is the same as behavior before, but it smells like a memory leak

starboard/android/shared/decode_target.cc
```diff
@@ -22,33 +22,61 @@
...
-
-  jobject global_surface = JniConvertLocalRefToGlobalRef(env, local_surface);
-
-  return global_surface;
+  struct SurfaceCache {
+    ScopedJavaGlobalRef<jclass> surface_class;
+    jmethodID surface_constructor;
+  };
+
+  static SurfaceCache cache;
```
johnxwork: A few questions:
1. Is this a new invention? I thought this PR is only to migrate JNI.
2. Why is this not a member variable and instead a static object? Do we expect multiple DecodeTargets and all of them to share the same cache?
3. Why do we need it now?

starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -0,0 +1,60 @@
...
+      delete;
+
+  void SetOnFrameAvailableListener(
+      JNIEnv* env,
+      base::android::ScopedJavaLocalRef<jobject> surface_texture) const;
+  void RemoveOnFrameAvailableListener(
+      JNIEnv* env,
+      base::android::ScopedJavaLocalRef<jobject> surface_texture) const;
+
+  static jobject CreateVideoSurfaceTexture(JNIEnv* env, int gl_texture_id);
```
johnxwork: Typically you'd want to deal with Local object in function returns instead of raw objects for ownership counting. What's the reason to use jobject here?

osagie98: This is to keep the PR as similar as possible to the current behavior which create a jobject and instantiates a local ref (line 36) in decode_target.cc. This change immediately assigns it to a `ScopedJavaLocalRef`, but I should be able to return a ref directly.

starboard/android/shared/decode_target.cc
osagie98: 1. Not new, but it's a way to get around continually looking up the relevant class and method with JNI genrator. Previously, on the right, we could directly call the Surface constructor. Now, we need to call `FindClass` for the Surface, then `GetMethodID` for the constructor before we can call it. Instead of doing these extra 2 function calls each time, I added this cache instead, which I also do in video_decoder.*. [Here](https://github.com/youtube/cobalt/blob/9cd2db17c5ae068805d2c20025ecb792ea15b657/starboard/android/shared/media_capabilities_cache.cc#L74) are some [other](https://github.com/youtube/cobalt/blob/9cd2db17c5ae068805d2c20025ecb792ea15b657/starboard/android/shared/audio_track_bridge.cc#L275) examples of this idea.
2. It can be a member variable, generally there will only be a single `DecodeTarget` per playback. Though the cache will never change once set - and it just stores information to look up the Surface class and constructor, which shouldn't change or go stale during the life of the app. I'll change it to a member if that's preferable.
3. We don't need it now, there's no functional changes here - just managing tech debt.

starboard/android/shared/decode_target.cc
osagie98: I'll take a deeper look. The local surface ref should release at the end of this function, where the global surface ref will do so during destruction, which should also clean up the surface itself.

starboard/android/shared/decode_target.cc
```diff
@@ -22,33 +22,48 @@
...
+  static std::once_flag once_flag;
+  std::call_once(once_flag, [env, cache]() {
+    cache->surface_class = ScopedJavaGlobalRef<jclass>(
+        env, env->FindClass("android/view/Surface"));
+    SB_CHECK(cache->surface_class);
+    cache->surface_constructor =
+        env->GetMethodID(cache->surface_class.obj(), "<init>",
+                         "(Landroid/graphics/SurfaceTexture;)V");
+    SB_CHECK(cache->surface_constructor);
+  });
```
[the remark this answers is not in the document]
osagie98: I restored the original static idea

starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -0,0 +1,64 @@
...
+      base::android::ScopedJavaLocalRef<jobject> surface_texture) const;
+  void RemoveOnFrameAvailableListener(
+      JNIEnv* env,
+      base::android::ScopedJavaLocalRef<jobject> surface_texture) const;
+
+  static ScopedJavaLocalRef<jobject> CreateVideoSurfaceTexture(
+      JNIEnv* env,
+      int gl_texture_id);
+
+  void OnFrameAvailable(JNIEnv* env) { host_->OnFrameAvailable(); }
```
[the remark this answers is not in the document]
osagie98: Done

kjyoun:
Whoops, I accidentally made a PR doing the same thing! 
Sorry for stepping on your toes.
https://github.com/youtube/cobalt/pull/7476

While looking into this, I found a potentially simpler approach for the JNI migration that avoids the need for `video_surface_texture_bridge.h`.

 The design uses a direct, bidirectional JNI communication model:

From Java, `VideoSurfaceTexture` calls the native `MediaVideoCodecDecoder`.

From native, `MediaVideoCodecDecoder` calls back to the Java `VideoSurfaceTexture`.

starboard/android/shared/decode_target.h
```diff
@@ -53,8 +56,8 @@ class DecodeTarget final : public SbDecodeTargetPrivate {
 
   // Java objects which wrap the texture.  We hold on to global references
   // to these objects.
-  jobject surface_texture_;
-  jobject surface_;
+  ScopedJavaGlobalRef<jobject> surface_texture_;
```
kjyoun: > Never declare namespace aliases or convenience using-declarations at namespace scope in header files, only in .cc files.

base::android::ScopedJavaGlobalRef<jobject> surface_texture_;

starboard/android/shared/decode_target.h
```diff
@@ -19,20 +19,23 @@
...
 #include <jni.h>
 
+#include "base/android/jni_android.h"
 #include "starboard/common/size.h"
 #include "starboard/decode_target.h"
 #include "starboard/shared/starboard/decode_target/decode_target_internal.h"
 
 namespace starboard {
 
+using base::android::ScopedJavaGlobalRef;
```
kjyoun: can you delete this using?

[go/totw/119](https://goto.google.com/totw/119)

> Never declare namespace aliases or convenience using-declarations at namespace scope in header files, only in .cc files.

starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <jni.h>
+
+#include "base/android/jni_android.h"
+#include "base/memory/raw_ptr.h"
+#include "starboard/common/log.h"
+
+namespace starboard {
+
+using base::android::ScopedJavaLocalRef;
```
kjyoun: > Never declare namespace aliases or convenience using-declarations at namespace scope in header files, only in .cc files.

[go/totw/119](https://goto.google.com/totw/119)

starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -0,0 +1,64 @@
...
+      base::android::ScopedJavaLocalRef<jobject> surface_texture) const;
+
+  static ScopedJavaLocalRef<jobject> CreateVideoSurfaceTexture(
+      JNIEnv* env,
+      int gl_texture_id);
+
+  void OnFrameAvailable(JNIEnv*) { host_->OnFrameAvailable(); }
+
+ private:
+  const raw_ptr<Host> host_;
```
kjyoun: raw_ref, since `host_` is guaranteed to be not-null

per gemini.

> In the Chromium codebase, both raw_ptr<T> and raw_ref<T> are non-owning smart pointers designed to enhance memory safety by mitigating use-after-free vulnerabilities. They are intended to replace raw C++ pointers (T*) and references (T&) for class and struct fields. The fundamental difference between them lies in their nullability.

starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -0,0 +1,64 @@
...
+    Host() = default;
+    ~Host() = default;
+
+    virtual void OnFrameAvailable() = 0;
+  };
+
+  VideoSurfaceTextureBridge(Host* host) : host_(host) { SB_CHECK(host_); }
+  ~VideoSurfaceTextureBridge() = default;
+
+  VideoSurfaceTextureBridge(const VideoSurfaceTextureBridge&) = delete;
```
kjyoun, line 41: nit: Add // Disallow copy and assign. 

https://source.corp.google.com/search?q=%22%2F%2F%20Disallow%20copy%20and%20assign.%22&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

starboard/android/shared/video_decoder.h
```diff
@@ -47,7 +50,8 @@ namespace starboard {
 class MediaCodecVideoDecoder : public VideoDecoder,
                                public MediaCodecDecoder::Host,
                                private JobQueue::JobOwner,
```
kjyoun, line 53: These ordering look strange.

can you make them sorted public to private?

```
public VideoDecoder,
public MediaCodecDecoder::Host,
ublic VideoSurfaceTextureBridge::Host,
private VideoSurfaceHolder,
private JobQueue::JobOwner,
```

starboard/android/shared/video_decoder.cc
```diff
@@ -981,19 +985,48 @@ bool MediaCodecVideoDecoder::IsBufferDecodeOnly(
...
 
+// Cache the SurfaceTexture methods so that we minimize JNI calls in
+// updateTexImage() and getTransformMatrix().
+struct SurfaceTextureJniCache {
+  ScopedJavaGlobalRef<jclass> surface_class;
+  jmethodID update_tex_image_method;
+  jmethodID get_transform_matrix_method;
+};
+
+SurfaceTextureJniCache cache;
```
kjyoun: how about g_cache?

since this is a global variable.

By convention, we added g_ prefix for global variable

https://source.corp.google.com/search?q=%5Cbg_%20case:yes%20lang:cc&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

starboard/android/shared/video_decoder.cc
```diff
@@ -981,19 +985,48 @@ bool MediaCodecVideoDecoder::IsBufferDecodeOnly(
...
+
+void EnsureCacheIsInitialized(JNIEnv* env) {
```
kjyoun: how about `InitializeCacheOnce` ?

starboard/android/shared/decode_target.cc
kjyoun: JNI handles caching under the hood, so that we don't need it manually

Here is the generated code

```
// This method id is cached
static std::atomic<jmethodID> g_dev_cobalt_coat_StarboardBridge_getAppStartTimestamp(nullptr);

static jlong Java_StarboardBridge_getAppStartTimestamp(JNIEnv* env, const
    base::android::JavaRef<jobject>& obj) {
  jclass clazz = dev_cobalt_coat_StarboardBridge_clazz(env);
  CHECK_CLAZZ(env, obj.obj(),
      dev_cobalt_coat_StarboardBridge_clazz(env), 0);

  jni_generator::JniJavaCallContextChecked call_context;
  // Here method id is saved to the cache at first run
  call_context.Init<
      base::android::MethodID::TYPE_INSTANCE>(
          env,
          clazz,
          "getAppStartTimestamp",
          "()J",
          &g_dev_cobalt_coat_StarboardBridge_getAppStartTimestamp);

  jlong ret =
      env->CallLongMethod(obj.obj(),
          call_context.base.method_id);
  return ret;
}
```

starboard/android/shared/video_decoder.cc
```diff
@@ -981,19 +985,48 @@ bool MediaCodecVideoDecoder::IsBufferDecodeOnly(
 
 namespace {
 
+// Cache the SurfaceTexture methods so that we minimize JNI calls in
```
kjyoun: we don't need this caching logic

JNI generator already handled it.

starboard/android/shared/video_surface_texture_bridge.h
osagie98: Good tip, thanks

starboard/android/shared/decode_target.cc
```diff
@@ -22,33 +22,45 @@
...
-      env, "android/view/Surface", "(Landroid/graphics/SurfaceTexture;)V",
-      surface_texture);
-
-  jobject global_surface = JniConvertLocalRefToGlobalRef(env, local_surface);
-
-  return global_surface;
+ScopedJavaGlobalRef<jobject> CreateSurfaceFromSurfaceTexture(
+    JNIEnv* env,
+    jobject surface_texture) {
+  ScopedJavaLocalRef<jclass> surface_class(
```
kjyoun: i don't think that we need to write legacy JNI code any more.

Here is the example that I made on top of this PR, to soley use JNI generator

https://github.com/youtube/cobalt/pull/7529/commits/d07c8afed7c7c763528d046603921e38f73c0b1c

kjyoun:
can you update the title of commit message to `android: Migrate VideoSurfaceTexture to JNI Generator`

per [go/cobalt-style#commit-message](http://go/cobalt-style#commit-message), prefix ends with ":"

```
When your commit relates to one of these topics, select the most relevant tag from either category and add it to the beginning of your subject line, followed by a colon.

Examples:

media: Add support for AV1 playback
starboard: Implement SbWindowGetPlatformHandle for Wayland
...

```

kjyoun:
thanks for the changes

PR #7366, opened 2025-09-26
android: Change symbol for top 4 java stack frames on exception

ckim1346:
Looking at the [6.23 primes](https://primes.corp.google.com/products/5063451/metrics/crashes/clusters/1000000001517437853?duration_seconds=2592000&source=8594998787&reportType=NATIVE&version_A=5.30.308&version_B=6.23.300&sortBy=4&sortOrder=2&measure=reports&trendsTab=overview&timeGranularity=daily) for [b/439066169](https://b.corp.google.com/439066169), the @ symbol seems to be translated to `<EMAIL_ADDRESS>`: 

`Abort message: [20459:20459:0925/122243.045405:FATAL:jni_android.cc(364)] JNI exception: <EMAIL_ADDRESS>:335&<EMAIL_ADDRESS>:162&<EMAIL_ADDRESS>:294&<EMAIL_ADDRESS>:8177`

Changing to output match[0] instead.

Local testing output (forcibly throwing an exception): 
```09-29 14:27:05.706  7465  7465 F cobalt  : [7465:7465:0929/142705.651175:FATAL:jni_android.cc(364)] JNI exception: .onSuccess(CobaltActivity.java:170)&.executeEnqueuedCallbacks(BrowserStartupControllerImpl.java:388)&.browserStartupComplete(BrowserStartupControllerImpl.java:65)&.next(MessageQueue.java:335)```

Bug: 430385989

PR #7440, opened 2025-10-02
starboard: JniEnvExt logs JNI exception on abort 

kjyoun:
Previously, when `JniEnvExt` aborted, it did not leave any log other than the crash stack. This change modifies JniEnvExt to log a fatal error, which will be included in Prime reports, providing more information about the crash.

```
Before
(No log)

After
[FATAL:jni_android.cc(400)] JNI exception: java.lang.RuntimeException: Hello @ .MethodC(VideoFrameReleaseTimeHelper.java:111)&.MethodB(VideoFrameReleaseTimeHelper.java:115)&.MethodA(VideoFrameReleaseTimeHelper.java:119)&.adjustReleaseTime(VideoFrameReleaseTimeHelper.java:135)
```

This change also enhances the log message to include the exception error message, in addition to the source location information.

Bug: 448896865


base/android/jni_android.cc
```diff
@@ -67,6 +67,40 @@ bool shouldAddCobaltPrefix() {
   }
   return g_add_cobalt_prefix;
 }
+
+std::string GetFirstLine(const std::string& stack_trace) {
+  std::stringstream ss(stack_trace);
+  std::string first_line;
+  std::getline(ss, first_line);
+  return first_line;
+}
```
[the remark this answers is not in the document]
kjyoun: Done

base/android/jni_android.cc
```diff
@@ -67,6 +67,40 @@ bool shouldAddCobaltPrefix() {
...
+
+std::string FindTopJavaMethodsAndFiles(const std::string& stack_trace, const size_t max_matches) {
+    std::regex pattern("\\.([^.(]+)\\(([^)]+\\.java:\\d+)\\)");
```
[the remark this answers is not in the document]
kjyoun: ack.

But will stick with the existing code. This method is executed only once. this code is not in hot path

base/android/jni_android.cc
```diff
@@ -67,7 +67,45 @@ bool shouldAddCobaltPrefix() {
...
+//     at dev.cobalt.media.VideoFrameReleaseTimeHelper.MethodC(VideoFrameReleaseTimeHelper.java:111)
+//     at dev.cobalt.media.VideoFrameReleaseTimeHelper.MethodB(VideoFrameReleaseTimeHelper.java:115)
+//     at dev.cobalt.media.VideoFrameReleaseTimeHelper.MethodA(VideoFrameReleaseTimeHelper.java:119)
+//     at dev.cobalt.media.VideoFrameReleaseTimeHelper.adjustReleaseTime(VideoFrameReleaseTimeHelper.java:135)
+std::string GetFirstLine(const std::string& stack_trace) {
+  return stack_trace.substr(0, stack_trace.find('\n'));
+}
+
+std::string FindTopJavaMethodsAndFiles(const std::string& stack_trace, const size_t max_matches) {
+    std::regex pattern("\\.([^.(]+)\\(([^)]+\\.java:\\d+)\\)");
```
andrewsavage1, line 83: Building this from string pieces would make it more readable, i.e. having separate method and file regexes

Also, can we use `\w` instead of `[^.(]` and `[^)]`?

andrewsavage1: I see now that it's existing code, so it's fine if you want to keep it the same

andrewsavage1 (approved):

andrewsavage1:
Please have Colin or Charley take a look before submitting

base/android/jni_android.cc
kjyoun: yes it comes from existing code.
I moved the method into anonymous namespace, since it's used only inside this file

I want to keep existing logic

base/android/jni_android.cc
```diff
@@ -350,7 +388,9 @@ void CheckException(JNIEnv* env) {
 #if BUILDFLAG(IS_COBALT)
       std::string exception_info = GetJavaExceptionInfo(env, java_throwable);
       base::android::SetJavaException(exception_info.c_str());
-      exception_token = FindTopJavaMethodsAndFiles(exception_info, 4);
+      exception_token =
+          GetFirstLine(exception_info) + " @ " +
```
ckim1346: The '@' symbol was being converted to a string `EMAIL_ADDRESS` in primes: #7366 so I'm a little wary of that. Don't know if that would happen in this case but something to consider. Otherwise LGTM

ckim1346 (approved):

base/android/jni_android.cc
kjyoun: Thanks for the tip. Will change to "at"

kjyoun: Done

base/android/jni_android.cc
```diff
@@ -67,7 +67,45 @@ bool shouldAddCobaltPrefix() {
...
+    std::ostringstream oss;
+    for (size_t i = 0; i < all_matches.size(); ++i) {
+        oss << all_matches[i];
+        if (i < all_matches.size() - 1) {
+            oss << "&";
+        }
+    }
+
+    return oss.str();
+}
```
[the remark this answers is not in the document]
kjyoun: They are already in anonymous namespace

PR #7444, opened 2025-10-02
feat: Integrate musl stdio

jellefoks:
This commit integrates the buffered IO libc functions from musl. Specifically, this is all functions using 'FILE *'.

Key changes:

- **Musl stdio Integration:**
  - Integrated a comprehensive set of musl `stdio` functions and their dependencies into the build.
  - Replaced musl's internal, futex-based file-locking mechanism with a per-file locking mechanism that uses a `pthread_once_t` to ensure thread-safe initialization of a `pthread_mutex_t` and `pthread_cond_t` for each `FILE` object. This resolves incompatibilities with the Starboard threading model.
  - Replaced `syscall()` invocations in musl's I/O functions with a libc wrapper layer that maps musl's internal syscalls to standard C library functions (e.g., `read`, `write`, `open`). This is more portable and avoids potential conflicts with the underlying system's syscall interface.
  - Adapted musl's `pthread_impl.h` to allow musl's internal code to use `__pthread_self()->tid` unchanged by providing a Starboard-compatible implementation.

- **Starboard:**
  - Added `readv` and `unlinkat` to Starboard.

- **Build Configuration:**
  - Updated `third_party/musl/BUILD.gn` to include all necessary musl source files and their dependencies, and to reflect the new file structure for the syscall wrappers.
  - Removed the integrated `stdio` functions from the API leak detector manifest (`evergreen/manifest`).
  - Removed stdio symbols from `starboard/elf_loader/exported_symbols.cc` as they are now provided by musl.

- **Testing:**
  - Added a new test, `posix_stdio_test.cc`, to verify the correctness of basic stdio operations (`fopen`, `fclose`) in modular builds.

- **Googletest Cleanup:**
  - Removed the `ENABLE_COBALT_HERMETIC_HACKS` macro and associated workarounds from the googletest library. With the integration of a proper stdio implementation, these hacks are no longer necessary. This simplifies the code and removes platform-specific workarounds.
  
Bug: 406081586

### DIFF
--- a/third_party/musl/src/starboard/misc/syscall.c
+++ b/third_party/musl/src/starboard/misc/syscall.c
@@ -64,7 +64,6 @@ long syscall(long number, ...) {
     case __NR_close: {
       int fildes = va_arg(ap, int);
       va_end(ap);
-      printf("Syscall function is called with number __NR_close\n");
       return libc_wrapper_SYS_close(fildes);
     }
 #endif
@@ -74,14 +73,12 @@ long syscall(long number, ...) {
       int op = va_arg(ap, int);
       void* arg = va_arg(ap, void*);
       va_end(ap);
-      printf("Syscall function is called with number __NR_gettid\n");
       return libc_wrapper_SYS_fcntl(fd, op, arg);
     }
 #endif
 #ifdef __NR_gettid
     case __NR_gettid:
       va_end(ap);
-      printf("Syscall function is called with number __NR_gettid\n");
       return libc_wrapper_SYS_gettid();
 #endif
 #ifdef __NR_lseek
@@ -90,7 +87,6 @@ long syscall(long number, ...) {
       off_t offset = va_arg(ap, off_t);
       int whence = va_arg(ap, int);
       va_end(ap);
-      printf("Syscall function is called with number __NR_lseek\n");
       return libc_wrapper_SYS_lseek(fildes, offset, whence);
     }
 #endif
@@ -99,7 +95,6 @@ long syscall(long number, ...) {
       const char *pathname = va_arg(ap, const char *);
       struct stat * statbuf = va_arg(ap, struct stat *);
       va_end(ap);
-      printf("Syscall function is called with number __NR_lstat\n");
       return libc_wrapper_SYS_lstat(pathname, statbuf);
     }
 #endif
@@ -109,7 +104,6 @@ long syscall(long number, ...) {
       int flags = va_arg(ap, int);
       mode_t mode = va_arg(ap, mode_t);
       va_end(ap);
-      printf("Syscall function is called with number __NR_open\n");
       return libc_wrapper_SYS_open(pathname, flags, mode);
     }
 #endif
@@ -119,7 +113,6 @@ long syscall(long number, ...) {
       void* buf = va_arg(ap, void*);
       size_t nbyte = va_arg(ap, size_t);
       va_end(ap);
-      printf("Syscall function is called with number __NR_read\n");
       return libc_wrapper_SYS_read(fildes, buf, nbyte);
     }
 #endif
@@ -129,7 +122,6 @@ long syscall(long number, ...) {
       const struct iovec* iov = va_arg(ap, const struct iovec*);
       int iovcnt = va_arg(ap, int);
       va_end(ap);
-      printf("Syscall function is called with number __NR_readv\n");
       return libc_wrapper_SYS_readv(fildes, iov, iovcnt);
     }
 #endif
@@ -137,7 +129,6 @@ long syscall(long number, ...) {
     case __NR_rmdir: {
       const char* pathname = va_arg(ap, const char*);
       va_end(ap);
-      printf("Syscall function is called with number __NR_rmdir\n");
       return libc_wrapper_SYS_rmdir(pathname);
     }
 #endif
@@ -146,7 +137,6 @@ long syscall(long number, ...) {
       const char *pathname = va_arg(ap, const char *);
       struct stat * statbuf = va_arg(ap, struct stat *);
       va_end(ap);
-      printf("Syscall function is called with number __NR_stat\n");
       return libc_wrapper_SYS_stat(pathname, statbuf);
     }
 #endif
@@ -154,7 +144,6 @@ long syscall(long number, ...) {
     case __NR_unlink: {
       const char* pathname = va_arg(ap, const char*);
       va_end(ap);
-      printf("Syscall function is called with number __NR_unlink\n");
       return libc_wrapper_SYS_unlink(pathname);
     }
 #endif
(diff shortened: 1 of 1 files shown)


third_party/googletest/src/googletest/src/gtest.cc
```diff
@@ -3337,16 +3327,11 @@ static void ColoredPrintf(GTestColor color, const char* fmt, ...) {
...
-// TODO: b/399507045 - Cobalt: Fix build error, remove hack
-#elif BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
-  posix::PrintF("\033[0;3%sm", GetAnsiColorCode(color));
-  posix::VPrintF(fmt, args);
-  posix::PrintF("\033[m");  // Resets the terminal to default.
 #else
   printf("\033[0;3%sm", GetAnsiColorCode(color));
   vprintf(fmt, args);
   printf("\033[m");  // Resets the terminal to default.
-#endif  // GTEST_OS_WINDOWS && !GTEST_OS_WINDOWS_MOBILE
```
andrewsavage1, line 3400: nit: keep the comment here (I assume it's from upstream)

third_party/musl/arch/generic/bits/ioctl_fix.h
andrewsavage1, line 1: how does this file get included?

jellefoks: This gets included by musl's `arch/generic/bits/ioctl.h` , which in turn gets included by `include/sys/ioctl.h`, which is the include for the `ioctl` symbol.

third_party/googletest/src/googletest/src/gtest.cc
jellefoks: good catch

jellefoks:
Back to draft. The locking approach in this version is incomplete.

jellefoks:
Shouldn't have been closed

starboard/CHANGELOG.md
```diff
@@ -109,6 +110,7 @@ The functionality is migrated to use `getauxval` instead.
 * `symlink`
 * `sysconf`
 * `uname`
+* `unlinkat`
```
y4vor, line 113: I don't see `unlinkat` as registered symbol.

starboard/tools/api_leak_detector/api_leak_detector.py
```diff
@@ -261,6 +262,7 @@
     'sysconf',
     'uname',
     'unlink',
+    'unlinkat',
```
y4vor: why add it to the api_leak_detector?

third_party/abseil-cpp/absl/base/internal/sysinfo.cc
```diff
@@ -26,7 +26,7 @@
 #include <unistd.h>
 #endif
 
-#ifdef __linux__
+#if defined(__linux__) || defined(STARBOARD)
```
y4vor, line 29: Why do we need the starboard check? 
Also it should be BUILDFLAG(IS_STARBOARD) if we really need it.

third_party/musl/arch/arm/bits/ioctl_fix.h
```diff
@@ -1,2 +1,4 @@
+#include "../../generic/bits/ioctl_fix.h"
```
y4vor, line 2: Gate the change with defined(STARBOARD)?

third_party/musl/src/internal/pthread_impl.h
```diff
@@ -13,6 +13,78 @@
...
+	} StarboardPthreadCondMutex;
+
+	static inline void __cond_mutex_pair_init(StarboardPthreadCondMutex* pair) {
+	pthread_mutexattr_t attr;
+	pthread_mutexattr_init(&attr);
+	pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+	pthread_mutex_init(&pair->mutex, &attr);
+	pthread_mutexattr_destroy(&attr);
+	pthread_cond_init(&pair->cond, NULL);
+}
```
y4vor: This looks like wrong indentation.

Also it is a bit strange to not check for errors. I think we should add error checking.

third_party/musl/src/starboard/misc/syscall.c
```diff
@@ -12,22 +12,176 @@
...
+  printf("Unexpected syscall %d (0x%x)\n", number, number);
+  SbLogRawDumpStack(0);
+#endif  // defined(ENABLE_DEBUGGE
+
+  switch (number) {
+#ifdef __NR_close
+    case __NR_close: {
+      int fildes = va_arg(ap, int);
+      va_end(ap);
+      printf("Syscall function is called with number __NR_close\n");
```
y4vor: #if defined(ENABLE_DEBUGGER) here and below?

starboard/CHANGELOG.md
jellefoks: Solved. I had accidentally removed^H^H^H^H^H^H^H it from the manifest(s) during a rebase.

starboard/tools/api_leak_detector/api_leak_detector.py
jellefoks: Solved. I had accidentally removed^H^H^H^H^H^H^H it from the manifest(s) during a rebase.

third_party/abseil-cpp/absl/base/internal/sysinfo.cc
jellefoks: We need this because somehow, we're not always `__linux__` when this is compiled and we're not getting the `#include <sys/syscall.h>` in the line below.

I followed the pattern used around musl, but I'll try `BUILDFLAG(IS_STARBOARD)` instead of `defined(STARBOARD)`. We'll see if CI agrees that it work...

jellefoks: Update: nope:

```
./../third_party/abseil-cpp/absl/base/internal/sysinfo.cc:29:27: error: function-like macro 'BUILDFLAG' is not defined
   29 | #if defined(__linux__) || BUILDFLAG(IS_STARBOARD)
      |                           ^
../../third_party/abseil-cpp/absl/base/internal/sysinfo.cc:401:37: error: use of undeclared identifier '__NR_gettid'
  401 |   return static_cast<pid_t>(syscall(SYS_gettid));
      |                                     ^~~~~~~~~~
../../third_party/abseil-cpp/absl/base/internal/sysinfo.cc:397:20: note: expanded from macro 'SYS_gettid'
  397 | #define SYS_gettid __NR_gettid
      |                    ^~~~~~~~~~~
2 errors generated.
```

third_party/musl/src/internal/pthread_impl.h
jellefoks: Solved. Fixed during gardening.

third_party/musl/arch/arm/bits/ioctl_fix.h
jellefoks: Done

third_party/musl/src/starboard/misc/syscall.c
jellefoks: Ohhh. unexpected debugging leftover. I can just remove them all, because it's already logged with a stack trace above. Technically, this should be dead code now, but leaving it in, in case there is still a library with unexpected incorrect include paths, skipping our Starboard syscall expansion.

PR #7445, opened 2025-10-02
[Rebase M138] Fix Shell Paths

sideb0ard:
Bug: 418842688

This is a partial revert  of "[Rebase m138] Update shell_browser_context (#7408)"

Although chrome commit - https://chromium-review.googlesource.com/c/chromium/src/+/6480975 moves
* content/shell/browser/shell_paths.* -> content/shell/common/shell_paths.* 
we do not need to move our copy 
* cobalt/shell/browser/shell_paths.cc -> cobalt/shell/common/shell_paths.cc 

This creates a later merge conflict e.g. here - https://github.com/youtube/cobalt/pull/5978

This partially reverts commit 8c8291d62019a25322624a6609b5264507c147c2.


andrewsavage1:
Repeating from the chat, I'd prefer not to do this as I think it's better for us to follow //content/shell structure in //cobalt/shell where we can (for now) and the conflicting commit looks like it's already in this branch in https://github.com/youtube/cobalt/commit/ceca6de99afc5589b171b8aac4bf59492fba50c2

PR #7449, opened 2025-10-03
[Rebase m138] Fixes for cobalt_unittests

madhurajayaraman:
Chromium changes:
5242238: Make MetricsServicesManagerClient the owner of SyntheticTrialRegistry. | https://chromium-review.googlesource.com/c/chromium/src/+/5242238
5034217: [RWS] Remove CanonicalCookie::IsSameParty method | https://chromium-review.googlesource.com/c/chromium/src/+/5034217
5741777: Add a new GetHeader overload, and deprecate old version | https://chromium-review.googlesource.com/c/chromium/src/+/5741777

### DIFF
--- a/cobalt/browser/metrics/cobalt_metrics_services_manager_client_test.cc
+++ b/cobalt/browser/metrics/cobalt_metrics_services_manager_client_test.cc
@@ -49,6 +49,9 @@ class CobaltMetricsServicesManagerClientTest : public ::testing::Test {
   void SetUp() override {
     ASSERT_TRUE(temp_dir_.CreateUniqueTempDir());
 
+    synthetic_trial_registry_ =
+        std::make_unique<variations::SyntheticTrialRegistry>();
+
     path_override_ = std::make_unique<base::ScopedPathOverride>(
         base::DIR_CACHE, temp_dir_.GetPath());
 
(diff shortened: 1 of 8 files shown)


starboard/shared/starboard/player/player_internal.h
```diff
@@ -30,7 +30,7 @@
 #include "starboard/window.h"
 
 #if SB_PLAYER_ENABLE_VIDEO_DUMPER
-#include SB_PLAYER_DMP_WRITER_INCLUDE_PATH
+#include "starboard/shared/starboard/player/video_dmp_writer.h"
```
andrewsavage1: I would assume we need to be able to configure this, could you get someone from media to take a look? (Maybe @sideb0ard  would know)

madhurajayaraman: https://github.com/youtube/cobalt/blob/feature/rebase_m138/starboard/linux/shared/configuration_public.h#L54
We do have this, but it is for some reason not grabbing that value. This error is making every test that references player fail (which is a lot of them) :(

andrewsavage1: Is that define getting hit? I.e. if you add an #error next to it does that error trigger?

madhurajayaraman: This turned out to be a weird local error, ran a `git clean -fdx` and rebuilt and it worked??

cobalt/browser/metrics/cobalt_metrics_services_manager_client_test.cc
```diff
@@ -73,61 +77,47 @@ class CobaltMetricsServicesManagerClientTest : public ::testing::Test {
 };
 
 TEST_F(CobaltMetricsServicesManagerClientTest, ConstructorInitializesState) {
-  EXPECT_THAT(manager_client_->GetEnabledStateProvider(), NotNull());
```
andrewsavage1, line 76: Removing all of the metrics checks without replacement seems strange to me—is there no replacement function? What is the upstream change that requires this?

andrewsavage1 (approved):

PR #7453, opened 2025-10-03
Cleanup most linker stub hack

y4vor:
- Keep the changes in the launch_posix.cc as removing the file requires too many customisations in other places.

- Cleanup the code in base/process/process_metrics_posix_starboard_stub.cc content/utility/utility_main.cc media/capture/video/linux/video_capture_device_factory_linux_starboard_stub.cc

- Enable the proxy_config_service_linux.cc and remove the stub.

Issue: 413130400


net/BUILD.gn
```diff
@@ -1203,8 +1203,6 @@ component("net") {
     sources += [
       "base/network_change_notifier_linux.cc",
       "base/network_change_notifier_linux.h",
-      "proxy_resolution/proxy_config_service_linux.cc",
-      "proxy_resolution/proxy_config_service_linux.h",
```
niranjanyardi, line 1207: this is a bit confusing:

is_linux && !is_cobalt_hermetic_build : this condition will apply to chromium builds as well - so you are unintentionally changing upstream code ?

better way is to leave upstream code as it and add a explicit condition which ensures its only for cobalt if(is_cobalt && !is_cobalt_hermetic_build) > remove these files

y4vor: Updated.

PR #7454, opened 2025-10-03
Keep the v8 cpu.cc change

y4vor:
Originally we though V8_OS_LINUX would have to be turned off. But we kept it. The current customizations is considered the correct way and is no longer a hack.

Issue: 409756628


andrewsavage1 (approved):
Fine by me to do this, but if we're going to have this approach we should likely get rid of V8_OS_STARBOARD (do we have a ticket for that already?)

y4vor:
> V8_OS_STARBOARD

Yes b/440078577

PR #7455, opened 2025-10-03
starboard: Make SbPlayerBridge handle SetBounds directly

kjyoun:
This PR removes `SbPlayerSetBoundsHelper` and moves its logic into `SbPlayerBridge` and `StarboardRenderer`. This simplifies the code and removes an unnecessary class.

#vibe-coded

Bug: 448196546

### DIFF
--- a/media/starboard/sbplayer_bridge.cc
+++ b/media/starboard/sbplayer_bridge.cc
@@ -377,9 +377,10 @@ void SbPlayerBridge::WriteBuffers(
 void SbPlayerBridge::SetBounds(const gfx::Rect& rect) {
   DCHECK(task_runner_->RunsTasksInCurrentSequence());
 
-  // Cobalt renderer renders from back to front, using a monotonically
-  // increasing sequence guarantees that all video layers are correctly ordered
-  // on z axis.
+  // On certain platforms (e.g., Linux), the Cobalt renderer draws layers from
+  // back to front. A monotonically increasing sequence is used to generate a
+  // unique z-index for each player, ensuring video layers are stacked
+  // correctly.
   static base::AtomicSequenceNumber s_z_index;
 
   set_bounds_z_index_ = s_z_index.GetNext();


media/starboard/sbplayer_bridge.cc
```diff
@@ -381,10 +374,15 @@ void SbPlayerBridge::WriteBuffers(
...
 
-void SbPlayerBridge::SetBounds(int z_index, const gfx::Rect& rect) {
+void SbPlayerBridge::SetBounds(const gfx::Rect& rect) {
   DCHECK(task_runner_->RunsTasksInCurrentSequence());
 
-  set_bounds_z_index_ = z_index;
+  // Cobalt renderer renders from back to front, using a monotonically
+  // increasing sequence guarantees that all video layers are correctly ordered
+  // on z axis.
+  static base::AtomicSequenceNumber s_z_index;
```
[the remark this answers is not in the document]
kjyoun: this variable is used only inside `SetBounds`

Keeping it inside `SetBounds` is more clear in its intention (other methods don't need to access this variable)

media/starboard/sbplayer_bridge.cc
```diff
@@ -381,10 +374,15 @@ void SbPlayerBridge::WriteBuffers(
                                            &video_stream_info_);
 }
 
-void SbPlayerBridge::SetBounds(int z_index, const gfx::Rect& rect) {
+void SbPlayerBridge::SetBounds(const gfx::Rect& rect) {
   DCHECK(task_runner_->RunsTasksInCurrentSequence());
 
-  set_bounds_z_index_ = z_index;
+  // Cobalt renderer renders from back to front, using a monotonically
```
borongc: This is actually for linux platform (not android), we should consider to make the commit like:

```
// Starboard on certain platform (e.g., Linux) renders from back to front, using a monotonically xxx
```

kjyoun: This was a comment copy-n-pasted from deleted media/starboard/sbplayer_set_bounds_helper.cc

Now  I update comment accordingly. PTAL

borongc (approved):

yell0wd0g (approved):
Changes LGTM! (Non Owner)

media/starboard/sbplayer_bridge.cc
```diff
@@ -597,8 +596,6 @@ void SbPlayerBridge::Suspend() {
 
   sbplayer_interface_->SetPlaybackRate(player_, 0.0);
 
-  set_bounds_helper_->SetPlayerBridge(NULL);
```
xiaomings, line 600: The logic here seems to be slightly altered before and after the change.  On the other hand, it's possible that Suspend() isn't used.

kjyoun: I think that we can remove this line without the change of behavior.

This line is to disable SetBounds during suspended[1].
but `SetBounds` already has a logic to disable itself when suspended.

```
[1]
void SbPlayerSetBoundsHelper::SetPlayerBridge(SbPlayerBridge* player_bridge) {
  CHECK(task_runner_->RunsTasksInCurrentSequence());
  player_bridge_ = player_bridge;
  if (player_bridge_ && rect_.has_value()) {
    player_bridge_->SetBounds(s_z_index.GetNext(), rect_.value());
  }
}

[2]
void SbPlayerBridge::SetBounds(int z_index, const gfx::Rect& rect) {
  DCHECK(task_runner_->RunsTasksInCurrentSequence());

  set_bounds_z_index_ = z_index;
  set_bounds_rect_ = rect;

  if (state_ == kSuspended) {
    return;
  }

  UpdateBounds();
}

```

xiaomings: I see.  Thanks for the comment.

PR #7459, opened 2025-10-06
[Rebase m138] Build fixes for gpu_unittests

madhurajayaraman:


andrewsavage1:
Was this the result of a bad cherry pick? Can you reference the original bad cherry pick in the PR description if so?

madhurajayaraman:
> Was this the result of a bad cherry pick? Can you reference the original bad cherry pick in the PR description if so?

I believe the issue is that in Chromium there used to be another if/else block around this, and we added our own extra `!hacks` conditional to it. So when they removed the if/else block, it actually cherry picked correctly but left us with a dangling if.

This is the chromium change that adds some of this code, https://chromium-review.googlesource.com/c/chromium/src/+/4598828
Here's our existing if/else : https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:gpu/ipc/service/gpu_channel_manager_unittest.cc
Here's the code that removes the if/else blocks: https://chromium-review.googlesource.com/c/chromium/src/+/5505454

andrewsavage1:
> > Was this the result of a bad cherry pick? Can you reference the original bad cherry pick in the PR description if so?
> 
> I believe the issue is that in Chromium there used to be another if/else block around this, and we added our own extra `!hacks` conditional to it. So when they removed the if/else block, it actually cherry picked correctly but left us with a dangling if.
> 
> This is the chromium change that adds some of this code, https://chromium-review.googlesource.com/c/chromium/src/+/4598828 Here's our existing if/else : https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:gpu/ipc/service/gpu_channel_manager_unittest.cc Here's the code that removes the if/else blocks: https://chromium-review.googlesource.com/c/chromium/src/+/5505454

Let's add this to the description so it's in the commit history

PR #7462, opened 2025-10-06
Set version number for 26.eap.1

kaidokert:
Bug: 404845662


kaidokert:
Can't really do this as this would break fast-forward merges.

andrewsavage1:
We should do this now, right?

PR #7463, opened 2025-10-06
Deprecate V8_OS_STARBOARD

y4vor:
Removing dead code gated by V8_OS_STARBOARD which is not defined.
We use V8_OS_LINUX.

Issue: 440078577


andrewsavage1 (commented):
Please make it clear in the description that this just deletes dead code as V8_OS_STARBOARD is not currently defined in the build.

PR #7466, opened 2025-10-07
[Rebase m138] fixes for blink_unittests

madhurajayaraman:


third_party/blink/renderer/modules/mediarecorder/audio_track_mojo_encoder_unittest.cc
```diff
@@ -159,6 +163,19 @@ class TestInterfaceFactory final : public media::mojom::InterfaceFactory {
...
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  void CreateStarboardRenderer(
+      mojo::PendingRemote<media::mojom::MediaLog> media_log_remote,
+      const media::StarboardRendererConfig& config,
+      mojo::PendingReceiver<media::mojom::Renderer> receiver,
+      mojo::PendingReceiver<media::mojom::StarboardRendererExtension>
+          renderer_extension_receiver,
+      mojo::PendingRemote<media::mojom::StarboardRendererClientExtension>
+          client_extension_remote){
+    NOTREACHED();
```
andrewsavage1, line 175: Someone from media should review this and see if we want to add a TODO or keep it this way

madhurajayaraman: @borongc can you take a look? should I add a todo to implement this function in these tests, or leave as is?

borongc: This should be fine as `CreateStarboardRenderer` mostly follows `CreateMediaFoundationRenderer`, using the same way here is sufficient.

andrewsavage1 (approved):

PR #7470, opened 2025-10-07
starboard: Make SbSystemRequestStop a no-op on Android

kjyoun:
`SbSystemRequestStop()` was implemented using a legacy JNI call, but this function is not used.
 
This change removes the unnecessary JNI implementation and replaces it with a warning log. This simplifies the native codebase by removing an unused JNI call.

Bug: 450024477

### DIFF
--- a/starboard/android/shared/system_request_stop.cc
+++ b/starboard/android/shared/system_request_stop.cc
@@ -14,10 +14,10 @@
 
 #include "starboard/system.h"
 
-#include "base/android/jni_android.h"
-#include "starboard/android/shared/starboard_bridge.h"
+#include "starboard/common/log.h"
 
 void SbSystemRequestStop(int error_level) {
-  JNIEnv* env = base::android::AttachCurrentThread();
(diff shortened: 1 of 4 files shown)


starboard/android/shared/system_request_stop.cc
```diff
@@ -15,14 +15,9 @@
...
 
 #include "base/android/jni_android.h"
-#include "starboard/android/shared/jni_env_ext.h"
-#include "starboard/android/shared/jni_state.h"
-
-using ::starboard::JniCallVoidMethodOrAbort;
-using ::starboard::JNIState;
+#include "starboard/android/shared/starboard_bridge.h"
 
 void SbSystemRequestStop(int error_level) {
```
johnxwork, line 19: I think this is unused on Android port now. It's kept for AOSP usage in the future

kjyoun: I made `SbSystemRequestStop` to NO-OP, instead of keeping legacy JNI method calls.
We can easily reimplement this method, when AOSP is ready.

johnxwork (approved):

PR #7479, opened 2025-10-08
Enable crashpad for arm, and cxx17, toolchains

hlwarriner:
This PR includes a few sets of changes to enable crash reporting on our evergreen-raspi reference platform:
* Customizations to build //base wtih cxx17, which partner toolchains can opt into with a toolchain arg
* Customizations to build zlib with arm toolchains that are on a version less than ARMv8
* Compiler flags added to the raspi port to disable warnings that are not issued by more modern compilers.

With these changes it should now be fairly straightforward for most partners to enable crashpad on their platforms.

Issue: 446904317
Issue: 406511608

### DIFF
--- a/starboard/raspi/2/toolchain/BUILD.gn
+++ b/starboard/raspi/2/toolchain/BUILD.gn
@@ -33,6 +33,9 @@ gcc_toolchain("starboard") {
     sysroot = "$raspi_home/busterroot"
     use_cxx17 = true
     build_base_with_cpp17 = true
+
+    # TODO: 451695911 - enable remote compilation, when we can.
+    use_remoteexec = false
   }
   propagates_configs = !enable_cobalt_hermetic_hacks
 }
@@ -54,5 +57,8 @@ gcc_toolchain("native_target") {
     sysroot = "$raspi_home/busterroot"
     use_cxx17 = true
     build_base_with_cpp17 = true
+
+    # TODO: 451695911 - enable remote compilation, when we can.
+    use_remoteexec = false
   }
 }


base/allocator/partition_allocator/partition_alloc.gni
```diff
@@ -47,7 +47,14 @@ declare_args() {
   #     and doesn't wish to incur the library size increase (crbug.com/674570).
   # 2.  On NaCl (through this declaration), where PartitionAlloc doesn't
   #     build at all.
-  use_partition_alloc = !is_nacl
+  #
+  # Cobalt comment: it would be a significant change to use PartitionAlloc as
+  # the default allocator in components built by partners. We're considering
+  # that change in b/447346129 but for now set this false in partner toolchains
+  # so that PartitionAlloc is not even linked into partner built binaries.
+  use_partition_alloc =
```
andrewsavage1, line 55: Do we still need the above two changes if we're not using partition alloc?

hlwarriner:
@y4vor @andrewsavage1 re the `./native_target/crashpad_handler: symbol lookup error: ./native_target/crashpad_handler: undefined symbol: _dl_stack_flags` runtime error I mentioned yesterday, I identified what I think is the issue: `-lpthread` is getting added to the end of the link command and introducing multiple undefined symbols in the executable (e.g., `_dl_stack_flags`, `_dl_init_static_tls`). I manually ran the link command after modifying it to remove `-lpthread` and this had the desired effect: these undefined symbols disappeared and `crashpad_handler` ran as expected on my raspberry pi.

I'll add another commit when I figure out where in the build code to make a change to address this. Let me know if you have any suggestions.

cobalt/build/configs/BUILD.gn
```diff
@@ -147,4 +147,10 @@ config("buildflag_defines") {
   if (is_native_target_build) {
     defines += [ "ENABLE_BUILDFLAG_IS_NATIVE_TARGET_BUILD" ]
   }
+
+  # Use build_base_with_cpp17 and BUILDFLAG(BUILD_BASE_WITH_CPP17) to guard
+  # customizations to //base in order to build the library using C++17.
+  if (build_base_with_cpp17) {
```
andrewsavage1, line 160: Instead of declaring it in base_configuration.gni, I think we should declare build_base_with_cpp17 in this file as it's the only place it's used, so there's no reason for it to be global

base/allocator/partition_allocator/partition_alloc.gni
hlwarriner: The changes to source code under `base/allocator/partition_allocator/`? We do because, to pass `gn check`, the upstream build code unconditionally compiles the PartitionAlloc code and then removes it at the link stage if `use_partition_alloc = false`: https://github.com/youtube/cobalt/blob/main/base/BUILD.gn#L1722.

We could alternatively try to change that behavior in the build code but I think the current approach is cleaner, especially since similar changes are already needed in `base/cxx20_is_constant_evaluated.h` and `base/numerics/safe_conversions_impl.h`.

cobalt/build/configs/BUILD.gn
hlwarriner: That makes sense to narrow the scope to that one file. Thanks.

base/allocator/partition_allocator/partition_alloc.gni
```diff
@@ -47,7 +47,15 @@ declare_args() {
...
+      !is_nacl && (!is_cobalt || ((current_toolchain != starboard_toolchain ||
```
andrewsavage1: nit: I think this would read a little cleaner if you extracted the not, i.e.
`!(is_cobalt && (current_toolchain == starboard_toolchain && starboard_toolchain != default_toolchain || current_toolchain == native_target))`
wdyt?

third_party/zlib/BUILD.gn
```diff
@@ -8,6 +8,15 @@ declare_args() {
   # Expose zlib's symbols, used by Node.js to provide zlib APIs for its native
   # modules.
   zlib_symbols_visible = false
+
+  if (is_cobalt) {
+    # TODO: b/446904317 - Cobalt: consider extracting this variable for use in
+    # other scopes.
+    is_platform_defined_toolchain =
```
andrewsavage1: nit: prepend an underscore to remove this from the global scope and remove it from the declare_args block as we don't want to allow overrides

hlwarriner: Done

base/allocator/partition_allocator/partition_alloc.gni
hlwarriner: I can convince myself that either is more readable. I don't feel strongly though and made the change you suggested.

hlwarriner:
> @y4vor @andrewsavage1 re the `./native_target/crashpad_handler: symbol lookup error: ./native_target/crashpad_handler: undefined symbol: _dl_stack_flags` runtime error I mentioned yesterday, I identified what I think is the issue: `-lpthread` is getting added to the end of the link command and introducing multiple undefined symbols in the executable (e.g., `_dl_stack_flags`, `_dl_init_static_tls`). I manually ran the link command after modifying it to remove `-lpthread` and this had the desired effect: these undefined symbols disappeared and `crashpad_handler` ran as expected on my raspberry pi.
> 
> I'll add another commit when I figure out where in the build code to make a change to address this. Let me know if you have any suggestions.

@andrewsavage1 identified https://github.com/youtube/cobalt/blob/main/build/config/BUILD.gn#L224 and I confirmed that this addition of "pthread" to `libs` is adding  `-lpthread` to the link line. We already have an exception there for the hermetic cobalt and starboard toolchains; I'll upload a commit to include the native target toolchain in that exception.

hlwarriner:
> > @y4vor @andrewsavage1 re the `./native_target/crashpad_handler: symbol lookup error: ./native_target/crashpad_handler: undefined symbol: _dl_stack_flags` runtime error I mentioned yesterday, I identified what I think is the issue: `-lpthread` is getting added to the end of the link command and introducing multiple undefined symbols in the executable (e.g., `_dl_stack_flags`, `_dl_init_static_tls`). I manually ran the link command after modifying it to remove `-lpthread` and this had the desired effect: these undefined symbols disappeared and `crashpad_handler` ran as expected on my raspberry pi.
> > I'll add another commit when I figure out where in the build code to make a change to address this. Let me know if you have any suggestions.
> 
> @andrewsavage1 identified https://github.com/youtube/cobalt/blob/main/build/config/BUILD.gn#L224 and I confirmed that this addition of "pthread" to `libs` is adding `-lpthread` to the link line. We already have an exception there for the hermetic cobalt and starboard toolchains; I'll upload a commit to include the native target toolchain in that exception.

Addressed in the latest commit.

build/config/BUILD.gn
```diff
@@ -218,7 +218,12 @@ config("default_libs") {
       "CoreText.framework",
       "Foundation.framework",
     ]
-  } else if (is_linux && !is_cobalt_hermetic_build || is_chromeos) {
+  } else if (is_cobalt && is_native_target_build) {
+    # This build configuration is only used for the crashpad handler executable.
+    # This binary needs dynamic linking but gets undefined symbols, which result
+    # in symbol lookup errors at runtime, when also linked with pthread.
+    libs = [ "dl" ]
```
andrewsavage1, line 225: Can we put this somewhere more targeted and/or that doesn't touch upstream code? If this is platform-dependent (because it built without this on linux, right?), maybe the platform_configuration/BUILD.gn?

hlwarriner: 1. We already have cobalt customizations in this config (see the condition directly below this one).
2. If we don't add this condition we would still need to modify the one directly below it so that it doesn't result in `true` for native target builds (unless we want to remove pthread from libs later, which I don't think is a good idea).

Given that, I think this change is fine.

hlwarriner: (Resolved offline)

PR #7483, opened 2025-10-08
media: Refine AdjustForDiscardedDurations

johnedocampo:
Currently, our implementation for
`DecodedAudio::AdjustForDiscardedDurations` assumes that `discarded_duration_from_front` and `discarded_duration_from_back` can never be negative. However, some upstream codes suggest that WebM files could have negative discard padding (check bug ID to see more on this). 

While inquiring on if we need to properly support negative discard padding, this PR makes some slight changes to DecodedAudio::AdjustForDiscardedDurations to now set `discarded_duration_from_front` and `discarded_duration_from_back` to 0 if they are negative. 

If it turns out that we don't need to worry about negative discard padding, we can just leave this as is. If we do need to support negative discard padding, we'll create a follow-up PR that properly handles this case.

Bug: 448163604

### DIFF
--- a/starboard/shared/starboard/player/decoded_audio_internal.cc
+++ b/starboard/shared/starboard/player/decoded_audio_internal.cc
@@ -151,8 +151,16 @@ void DecodedAudio::AdjustForDiscardedDurations(
     int sample_rate,
     int64_t discarded_duration_from_front,
     int64_t discarded_duration_from_back) {
-  SB_DCHECK_GE(discarded_duration_from_front, 0);
-  SB_DCHECK_GE(discarded_duration_from_back, 0);
+  if (discarded_duration_from_front < 0) {
+    SB_LOG(WARNING) << "discarded_duration_from_front is negative with value "
+                    << discarded_duration_from_front << ". Setting to 0.";
+    discarded_duration_from_front = 0;
+  }
+  if (discarded_duration_from_back < 0) {
+    SB_LOG(WARNING) << "discarded_duration_from_back is negative with value "
+                    << discarded_duration_from_back << ". Setting to 0.";
+    discarded_duration_from_back = 0;
+  }
   SB_DCHECK_EQ(storage_type(), kSbMediaAudioFrameStorageTypeInterleaved);
 
   if (discarded_duration_from_front == 0 && discarded_duration_from_back == 0) {


starboard/shared/starboard/player/decoded_audio_internal.cc
```diff
@@ -151,8 +151,10 @@ void DecodedAudio::AdjustForDiscardedDurations(
     int sample_rate,
     int64_t discarded_duration_from_front,
     int64_t discarded_duration_from_back) {
-  SB_DCHECK_GE(discarded_duration_from_front, 0);
-  SB_DCHECK_GE(discarded_duration_from_back, 0);
+  discarded_duration_from_front =
```
kjyoun: How about this?

```
#include <algorithm> // Required for std::max

// ...

discarded_duration_from_front = std::max(0, discarded_duration_from_front);
discarded_duration_from_back = std::max(0, discarded_duration_from_back);
````

Comment (kjyoun) [reply starboard/shared/starboard/player/decoded_audio_internal.cc]: or 

don't we need to log warning when discarded_duration_from_front or discarded_duration_from_back has negative values at least?

Comment (johnedocampo) [reply starboard/shared/starboard/player/decoded_audio_internal.cc]: I agree that we should add log warnings. Adjusted implementation to also output warnings. Thanks for the quick review KJ!

File: starboard/shared/starboard/player/decoded_audio_internal.cc [hunk 2]
```diff
@@ -151,8 +151,16 @@ void DecodedAudio::AdjustForDiscardedDurations(
...
+  if (discarded_duration_from_front < 0) {
+    SB_LOG(WARNING) << "discarded_duration_from_front is negative with value "
+                    << discarded_duration_from_front << ". Setting to 0.";
+    discarded_duration_from_front = 0;
+  }
+  if (discarded_duration_from_back < 0) {
+    SB_LOG(WARNING) << "discarded_duration_from_back is negative with value "
+                    << discarded_duration_from_back << ". Setting to 0.";
+    discarded_duration_from_back = 0;
+  }
```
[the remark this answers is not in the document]
johnedocampo: I think we can just keep the implementation as is.

PR #7484, opened 2025-10-09
starboard: Add starboard_platform_tests to CI on android

johnedocampo:
This PR adds the test suite "starboard_platform_tests" to CI, and also makes it a buildable test binary, so that the test suite can be ran locally. Changes have also been made to the tests that this suite runs to address compiler warnings.

A follow up bug will tag all failing tests from the suite to investigate on whether we should support them, remove them etc.

Bug: 450374396

### DIFF
--- a/cobalt/testing/filters/android-arm/starboard_platform_tests_filter.json
+++ b/cobalt/testing/filters/android-arm/starboard_platform_tests_filter.json
@@ -1,10 +1,12 @@
 {
+  "comment": "TODO: b/450568428 - Investigate these failing tests",
   "failing_tests": [
     "JniEnvExtTest.EmptyNewStringStandardUTF",
     "JniEnvExtTest.NewStringStandardUTF",
     "MimeUtilTest.ChecksSupportedFlacContainers",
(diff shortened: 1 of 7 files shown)


starboard/android/shared/BUILD.gn
```diff
@@ -388,10 +390,11 @@ static_library("starboard_base_symbolize") {
   check_includes = false
 }
 
-source_set("starboard_platform_tests") {
+test("starboard_platform_tests") {
```
kjyoun, line 393: can you add an instruction somewhere about how to run this test locally?

johnedocampo:
> Looks good to me.
> 
> My only request is, can you write instructions somewhere about how to run these tests locally? Otherwise, we have to wait ~15 min for CI to run the Android unit tests to see the results.

@kjyoun Will do. Looking to update go/cobalt-android-tv-setup#running-tests; once that passes through I'll merge this PR.

starboard/android/shared/BUILD.gn
johnedocampo: Done.

PR #7487, opened 2025-10-09
[Rebase m138] Fixes for nplb

madhurajayaraman:
Fix for unused variable error and
`error: unknown warning option '-Wno-enum-constexpr-conversion' [-Werror,-Wunknown-warning-option]`
which led to signedness test failing.

### DIFF
--- a/starboard/nplb/sabi/signedness_and_size_of_enum_test.cc
+++ b/starboard/nplb/sabi/signedness_and_size_of_enum_test.cc
@@ -21,7 +21,7 @@ typedef enum GenericEnumType {
   kOnlyTag,
 } GenericEnumType;
 
-// TODO: Fix this test to work with new compiler.
+// TODO b/452970985 : Fix this test to work with new compiler.
 // SB_COMPILE_ASSERT((static_cast<GenericEnumType>(-1) < 0) == SB_HAS_SIGNED_ENUM,
 //                   SB_HAS_SIGNED_ENUM_is_inconsistent_with_sign_of_enum);
 


starboard/nplb/sabi/signedness_and_size_of_enum_test.cc
```diff
@@ -14,16 +14,15 @@
...
+#include <type_traits>
+
 namespace nplb {
 namespace {
 
 typedef enum GenericEnumType {
   kOnlyTag,
 } GenericEnumType;
 
-SB_COMPILE_ASSERT((static_cast<GenericEnumType>(-1) < 0) == SB_HAS_SIGNED_ENUM,
```
madhurajayaraman, line 24: This is false for both "std::is_signed_v<GenericEnumType>" and "std::is_unsigned_v<GenericEnumType>" so I'm not sure what this test is looking for or testing --needs to be rewritten.

andrewsavage1:
Why is this change necessary—can you add it to the PR description?

starboard/nplb/posix_compliance/posix_mkdir_test.cc
```diff
@@ -85,7 +85,7 @@ TEST_F(PosixMkdirTest, SuccessfulCreation) {
   ASSERT_NE(dirp, nullptr);
   int entry_count = 0;
   errno = 0;
-  while (struct dirent* dp = readdir(dirp)) {
+  while (readdir(dirp) != nullptr) {
```
andrewsavage1: Can this just be `while(readdir(dirp)) {`?

starboard/nplb/BUILD.gn
```diff
@@ -32,6 +32,9 @@ config("wnos") {
     "-Wno-unused-local-typedef",
     "-Wno-unreachable-code",
     "-Wno-unreachable-code-break",
+
+# Do not warn about variable length arrays
```
andrewsavage1: Let's format this correctly

starboard/nplb/sabi/signedness_and_size_of_enum_test.cc
andrewsavage1: See if you can take some of Yavor's time to ask about this

y4vor: The test is enforcing that the enum should have a signed type implementation. I'm surprised it is failing. 
Did we change the compiler in m138?

andrewsavage1: I think the chromium compiler (the one we use for the cobalt/default toolchain) updates fairly often, so I would assume it's changed from m114, yes

starboard/nplb/sabi/signedness_and_size_of_enum_test.cc
```diff
@@ -21,8 +21,9 @@ typedef enum GenericEnumType {
...
-                  SB_HAS_SIGNED_ENUM_is_inconsistent_with_sign_of_enum);
+// TODO: Fix this test to work with new compiler.
```
andrewsavage1: Let's create a bug for this and put it as a blocker for b/452422975

PR #7488, opened 2025-10-09
[Rebase m138] fixes for media_unittests

madhurajayaraman:
Chromium change:
Delete MediaUrlDemuxer | https://chromium-review.googlesource.com/c/chromium/src/+/6444139

### DIFF
--- a/media/base/decoder_buffer.cc
+++ b/media/base/decoder_buffer.cc
@@ -60,7 +60,7 @@ class ExternalSharedMemoryAdapter : public DecoderBuffer::ExternalMemory {
 
 // --- Starboard-specific Constructor Implementations ---
 DecoderBuffer::DecoderBuffer(size_t size) : size_(size) {
-  if (size_ > 0) {
+  if (size_ >= 0) {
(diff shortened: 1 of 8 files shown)


media/base/decoder_buffer.cc
```diff
@@ -60,7 +60,7 @@ class ExternalSharedMemoryAdapter : public DecoderBuffer::ExternalMemory {
 
 // --- Starboard-specific Constructor Implementations ---
 DecoderBuffer::DecoderBuffer(size_t size) : size_(size) {
-  if (size_ > 0) {
+  if (size_ >= 0) {
```
andrewsavage1, line 63: Do the tests have issues when size_ is 0 and we don't run Initialize?

madhurajayaraman: Yeah, because we only initialize data_ if size >0, but we set end_of_stream_ = !data_. When data_ = 0, basically all of the tests fail since we set end_of_stream_ = false, and that's a DCHECK at the start of basically every decoder buffer function.

PR #7489, opened 2025-10-09
Fix linker error for cobalt browser tests

linxinan-yt:
Move the build rule for shell_test_switches to avoid linker error like below:

```
ld.lld: error: duplicate symbol: switches::IsRunWebTestsSwitchPresent()
>>> defined at shell_test_switches.cc:25 (../../cobalt/shell/common/shell_test_switches.cc:25)
>>>            obj/cobalt/shell/browsertests_sources/shell_test_switches.o:(switches::IsRunWebTestsSwitchPresent())
>>> defined at shell_test_switches.cc:25 (../../cobalt/shell/common/shell_test_switches.cc:25)
>>>            cobalt_shell_test_lib/shell_test_switches.o:(.text+0x0) in archive obj/cobalt/testing/browser_tests/libcobalt_shell_test_lib.a
```
Test: autoninja -C out/android-x86_debug cobalt_browsertests
Bug: 432313695

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -45,6 +45,9 @@ import dev.cobalt.shell.Shell;
 import dev.cobalt.shell.ShellManager;
 import dev.cobalt.util.DisplayUtil;
 import dev.cobalt.util.Log;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -96,6 +99,7 @@ public abstract class CobaltActivity extends Activity {
   // Tracks the status of the FLAG_KEEP_SCREEN_ON window flag.
   private Boolean isKeepScreenOnEnabled = false;
   private String diagnosticFinishReason = "Unknown";
+  private PlatformError mPlatformError;
 
   // Initially copied from ContentShellActiviy.java
   protected void createContent(final Bundle savedInstanceState) {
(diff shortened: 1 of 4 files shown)


cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -407,16 +407,23 @@ test("cobalt_browsertests") {
...
 
-static_library("cobalt_shell_test_lib") {
+source_set("shell_test_switches_lib") {
   testonly = true
   sources = [
-    # TODO(b/428999732): move test switches files to cobalt/testing.
     "//cobalt/shell/common/shell_test_switches.cc",
     "//cobalt/shell/common/shell_test_switches.h",
+  ]
+  public_deps = [ "//base:base" ]
```
Awallky: Can this just be a deps instead of public deps?

cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -407,16 +407,23 @@ test("cobalt_browsertests") {
   }
 }
 
-static_library("cobalt_shell_test_lib") {
+source_set("shell_test_switches_lib") {
```
Awallky: Why don't we just add this directly to cobalt/shell/BUILD and add the target to the test lib deps?

cobalt/testing/browser_tests/BUILD.gn
linxinan-yt: Done.

cobalt/testing/browser_tests/BUILD.gn
linxinan-yt: SGTM. Done.

linxinan-yt:
> I think your test and bug footers are expected to be "Bug" and "Test".

Nice catch. I was not aware of it. Bug and BUG are different in Cobalt...

linxinan-yt:
Sorry mistakenly closed PR. Reopen it.

PR #7490, opened 2025-10-09
android: Remove obsolete JNI implementation in Android Media Session client

kjyoun:
The Android implementation now uses the Chromium MediaSession, making the Starboard-based MediaSession extension and its JNI client obsolete. The existing JNI implementation is out-dated, broken, and no longer has a counterpart in the Java layer.

This change removes the unused JNI implementation from android_media_session_client.cc. The GetMediaSessionApi function now logs a fatal error and returns nullptr to prevent accidental use of this unsupported extension.

Bug: 377019873
Bug: 450024477


jasonzhangxx:
The media session extension is no longer used. It should be ok to remove android_media_session_client.cc. But I'm not sure if it would be re-introduced back in the future for Android 3P AOSP. @johnxwork and @andrewsavage1 to correct me.

andrewsavage1:
> The media session extension is no longer used. It should be ok to remove android_media_session_client.cc. But I'm not sure if it would be re-introduced back in the future for Android 3P AOSP. @johnxwork and @andrewsavage1 to correct me.

Our approach has been to ignore AOSP changes and worry about those later—they'll still be in history. This is fine.

PR #7493, opened 2025-10-09
Use UnsafeDanglingUntriaged() for SbPlayer callbacks

jasonzhangxx:
|player| pointer in the SbPlayerBridge callbacks can be dangling by design, so we should use UnsafeDanglingUntriaged() for it.

Bug: 450337617


borongc:
We should merge this to the branch `feature/rebase_m138`.

jasonzhangxx:
Close this PR as it's not needed on m114.

PR #6417, opened 2025-07-14
content/browser: avoid launching unnecessary services/managers

yell0wd0g:
**Note to reader**
I'm unsure whether this is too big of a cannon - ideally we'd disable
services one by one, with a GN flag, and also disable its associated
tests etc (maybe also removing the code from the build, since linker
might not be smart enough to remove it...). This CL/PR is here as 
reference until we figure out what to do.

This CL/PR cuts off a few classes/managers/services that are known not to
be used at the moment in Cobalt, namely:
- OopDataDecoder, 
- Subsystem:Audio,
- Subsystem:Midi, 
- GamepadService, 
- FidoHidDiscovery, 
- WebRTCInternals, 
- MediaStreamManager (this is for both audio and video capture in browser), 
- SpeechRecognitionManager,

This should be upstreamed, rather than peppering IS_COBALT in a //content
file, but perhaps the savings in execution resources might justify this
transgression.

Bug: 383140932

PR #7496, opened 2025-10-10
cobalt/content: Don't launch unnecessary services

yell0wd0g:
This CL disables some Chrome services via either GN flag or a new
command line feature "UserInteractionPeripherals". 

This CL also adds code to disable said feature on Cobalt builds.
This is needed to keep content_shell building as on our ToT, 
otherwise some of the disabled services (MediaStreamManager)
cause crashes when running Web Platform Tests on this CI (there's
a specific "bot" for it which builds content_shell with IS_COBALT).

For good measure, chrome:webrtc-internals is also removed entirely,
since nobody is serving it anyway.

With this there're 4 less threads on a linux monolithic build
(similar reduction is expected for any other platform, but not
confirmed): from 78 -> 74.

The concrete list of Services not launched is, concretely:
- OopDataDecoder (via GN flag),
- WebRTCInternals (via GN flag),
- Subsystem:Audio,
- Subsystem:Midi,
- GamepadService,
- FidoHidDiscovery,
- MediaStreamManager (this is for both audio and video capture in browser),
- SpeechRecognitionManager,

This PR supersedes #6417 .

bug: 383140932
Bug: 450888267

### DIFF
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -33,6 +33,7 @@ generate_jni("jni_headers") {
     "apk/app/src/main/java/dev/cobalt/media/MediaDrmBridge.java",
     "apk/app/src/main/java/dev/cobalt/media/VideoFrameReleaseTimeHelper.java",
     "apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java",
+    "apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java",
     "apk/app/src/main/java/dev/cobalt/util/DisplayUtil.java",
   ]
 
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
@@ -19,7 +19,6 @@ import static dev.cobalt.media.Log.TAG;
 import android.media.AudioAttributes;
 import android.media.AudioFormat;
 import android.media.AudioManager;
-import android.media.AudioTimestamp;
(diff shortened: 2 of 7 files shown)


cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -78,7 +79,11 @@ static constexpr auto kCobaltToggleSwitches = std::to_array<const char*>({
 const base::CommandLine::SwitchMap GetCobaltParamSwitchDefaults() {
   const base::CommandLine::SwitchMap cobalt_param_switch_defaults({
     // Disable Vulkan.
-    {switches::kDisableFeatures, "Vulkan"},
+    {switches::kDisableFeatures,
+     "Vulkan,"
+     // Don't launch services related to user-interaction via peripherals,
+     // e.g. video capture.
+     "UserInteractionPeripherals"},
```
johnxwork, line 90: The indentation looks bit wield

yell0wd0g:
> The content modifications don't have macro guards which I suppose indicates the intent to upstream?

I tried building them out in Cobalt - but that removed those services from
content_shell, which is used in CI for Web Tests. 

In an ideal world, we could run Web Tests with cobalt_shell, but we're not
there yet (there's no test runner - or some such). So we are forced to the
suboptimal flag solution. We'll revisit this when upstreaming.

cobalt/app/cobalt_switch_defaults_starboard.cc
yell0wd0g: I know! I copy-pasted the weird stuff given to me by the CI style presubmit bot 🤷

yell0wd0g:
Continuing in #7786 .

PR #7499, opened 2025-10-10
starboard: Harden thread safety checks and simplify ThreadChecker

kjyoun:
This change improves thread safety and robustness by enabling thread checks in release builds and simplifying the underlying ThreadChecker implementation.

Key changes:
- `ThreadChecker` is no longer a no-op in release builds. It has been simplified to always bind to its creation thread, removing complex atomic and lazy-initialization logic.
- All usages of `SB_DCHECK(thread_checker_.CalledOnValidThread())` have been converted to `SB_CHECK`. This elevates the check from a debug-only assertion to a fatal error in all build types, ensuring that any threading violations are caught immediately.
- `JobQueue` is updated to use the new, simpler ThreadChecker, removing its own redundant thread ID tracking.

Bug: 435425692

### DIFF
--- a/.github/actions/docker/action.yaml
+++ b/.github/actions/docker/action.yaml
@@ -76,22 +76,21 @@ runs:
         set -x
         DOCKER_COMPOSE_IMAGE="${GITHUB_DOCKER_REGISTRY_URL}/${{ github.repository }}/${DOCKER_SERVICE}"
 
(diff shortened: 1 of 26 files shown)


kaidokert (approved):
It doesn't seem wrong to me. 

I did check a bit of Chromium code, and it looks like most of Chromium still uses DCHECKS for CalledonValidThread() for some reason - perhaps historical.

Please have Jelle or Xiaoming also have a look before merging, just in case.

PR #7500, opened 2025-10-10
Revert "android: Preserve Viz components in ProGuard"

johnxwork:
Reverts youtube/cobalt#7428

Bug: 448659685


zhongqiliang:
Note, this could be still needed to prevent the crash after vis is added in the dependency.

johnxwork:
> Note, this could be still needed to prevent the crash after vis is added in the dependency.

Correct, just reverting to see if this affects memory.

PR #7503, opened 2025-10-10
RDK: Add build documentation

denisyuji:
Introduce a new markdown file with reviewed steps for building and deploying Cobalt for RDK. 

Issue: 443126811

### DIFF
--- a/starboard/doc/eap/getting_started.md
+++ b/starboard/doc/eap/getting_started.md
@@ -1,38 +1,73 @@
 # Open Source Setup
 
+## Setup workstation ##
+
+Install necessary packages.
+
 ```
-# Install necessary packages
 sudo apt update && sudo apt install -qqy --no-install-recommends \
   git curl python3-dev xz-utils lsb-release file
+```
 
-# Install depot_tools
+Clone `depot_tools` and add the directory to `PATH` for this session.
+
+```
 git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
-
-# Consider adding the following to your ~/.bashrc or equivalent as well:
 export PATH="/path/to/depot_tools:$PATH"
+```
 
-# Download the repository. Use whichever protocol you prefer.
+Consider adding the `depot_tools` directory to `PATH` in `.bashrc`, `.profile`, or equivalent, to make it permanent.
+
+## Get source code ##
+
+Download the repository. Use whichever protocol you prefer.
+
+```
 git clone --single-branch git@github.com:youtube/cobalt.git chromium/src
 git -C chromium/src remote add _gclient git@github.com:youtube/cobalt.git
+```
 
-# Configure gclient and download subrepos
+Configure gclient and download subrepos
+
+```
 cd chromium
 gclient config --name=src git@github.com:youtube/cobalt.git
 cd src
 gclient sync --no-history -r $(git rev-parse @)
+```
 
-# Install build dependencies.
+
+## Building Evergreen for x64 ##
+
+Install build dependencies:
+
+```
 ./build/install-build-deps.sh
 
-# Build Evergreen for linux
+```
+
+Build Evergreen for linux:
+
+```
 cobalt/build/gn.py -p evergreen-x64 -c devel --no-rbe
 autoninja -C out/evergreen-x64_devel/ cobalt_loader nplb_loader
+```
 
-# Run Cobalt
+## Running Cobalt ##
+
+After build complete, you can run Cobalt with:
+
+```
 out/evergreen-x64_devel/loader_app
-# Run NPLB
+```
+
+To run `nplb`:
+
+```
 out/evergreen-x64_devel/elf_loader_sandbox --evergreen_content=. --evergreen_libary=libnplb.so
 ```
+
+## Additional Tips: ##
 
 Before building, you can optionally set cc_wrapper to a build accelerator of your choice. For example, setting it to ccache:
 
--- a/starboard/doc/setup-rdk.md
+++ b/starboard/doc/setup-rdk.md
@@ -1,0 +1,264 @@
+# Building and Deploying Cobalt for RDK
+
+## Prerequisites
+
+### Get an RDK device running the correct image
+1. Get an RDK reference device
+
+    You need to have a YouTube reference device (Amlogic S905X4 (AH212)) to run the commands.
+
+    If you don't have one, you can purchase a reference device from
+[rdklogic](https://rdklogic.tv/collections/all-products/products/amlogic-s905x4-developer-box-rdk?variant=43178224484592).
+
+2. Get a detailed overview of the RDK reference port
+
+    Please get the device information from YouTube Partner portal: [AH212](https://developers.google.com/youtube/devices/living-room/compliance/ah212).
+
+    Note: If you do not have access to the page above, please see
+    [Accessing Living Room Partnership Resources.](https://developers.google.com/youtube/devices/living-room/access/accessing-lr-partnership-resources)
+
+    More information about the RDK reference box can be found on RDK Central wiki: [RDK-Google IPSTB profile stack on Amlogic reference board](https://wiki.rdkcentral.com/display/RDK/RDK-Google+IPSTB+profile+stack+on+Amlogic+reference+board).
+
+3. Ensure that you have flashed the latest system image. If you are a google partner, see the [Cobalt Development on RDK box doc](https://docs.google.com/document/d/1hTlGU15_BKnaa_yT_Z8Omolb7rdZAZj57wtADoKgAfg/edit?resourcekey=0-v-QJNWJumUuksd8JTu76eQ&tab=t.0).
+
+### SSH Setup
+
+1. Configure passwordless SSH to your target RDK device:
+
+```bash
+ssh-copy-id root@<TARGET_IP>
+ssh root@<TARGET_IP> echo "SSH working"
+```
+
+**Note:** Replace `<TARGET_IP>` with your device's IP address.
+
+## 1. Fetch Cobalt Sources
+
+Follow the standard Cobalt setup process from [the starboard documentation](/starboard/doc/eap/getting_started.md):
+
+1. **Install required packages and depot_tools** (see [Setup workstation](/starboard/doc/eap/getting_started.md#setup-workstation))
+2. **Clone Cobalt repository and configure gclient** (see [Get source code](/starboard/doc/eap/getting_started.md#get-source-code))
+3. **Download subrepos** using `gclient sync`
+
+**Important:**
+- You do **NOT** need to install build dependencies or build Evergreen for x64 as shown in the starboard documentation.
+- The starboard documentation demonstrates building for `evergreen-x64` (Linux desktop). For RDK boards, you will use the ARM platform: `evergreen-arm-hardfp-rdk`. Skip the `evergreen-x64` build and proceed to the RDK-specific build section below.
+
+**RDK-specific directory setup:**
+
+```bash
+COBALT_BUILD_DIR="$HOME/cobalt-build"
+mkdir -p "$COBALT_BUILD_DIR"
+cd "$COBALT_BUILD_DIR"
+# Continue with standard setup from starboard docs, then proceed to step 2
+```
+
+## 2. Build Docker Image
+
+### Docker Registry Authentication
+
+Before building the Docker image, authenticate with Google Cloud to access the marketplace container registry:
+
+```bash
+gcloud auth login
+gcloud auth configure-docker marketplace.gcr.io
+```
+
+This is required because the `$COBALT_BUILD_DIR/chromium/src/cobalt/docker/rdk/Dockerfile` pulls base images from Google's marketplace registry.
+
+### Building Cobalt
+
+The Cobalt repo includes a `docker-compose.yaml` that describes the build images.
+
+This section is split into two parts: what to do on the host (Outside Container) and what the container will execute (Inside Container).
+
+#### Outside Container
+
+On the host, ensure the key directories and variables are set. These will be bind-mounted into the container so build artifacts are available on the host after the build completes.
+
+```bash
+COBALT_BUILD_DIR="$HOME/cobalt-build"
+DEPOT_TOOLS_DIR="/path/to/depot_tools"
+```
+
+Then, build the `rdk` image and launch an interactive shell inside a temporary container:
+
+```bash
+# Build the container image and run an interactive shell in the rdk service container
+docker compose -f "$COBALT_BUILD_DIR/chromium/src/docker-compose.yaml" run --rm --build \
+  -v "$COBALT_BUILD_DIR:$COBALT_BUILD_DIR" \
+  -e COBALT_BUILD_DIR="$COBALT_BUILD_DIR" \
+  -v "$DEPOT_TOOLS_DIR:$DEPOT_TOOLS_DIR" \
+  -e DEPOT_TOOLS_DIR="$DEPOT_TOOLS_DIR" \
+  rdk bash
+```
+
+#### Inside Container
+
+After using the command above, the container will start with a interactive shell. Paste the following commands into the container shell to perform the build.
+
+Inside Container (paste these):
+
+```bash
+# Configure git for safe directory access
+git config --global --add safe.directory '*'
+
+# Add depot_tools to PATH for build tools (gn, ninja)
+export PATH=$DEPOT_TOOLS_DIR:$PATH
+
+# Navigate to Cobalt source directory
+cd $COBALT_BUILD_DIR/chromium/src
+
+# Generate build files for RDK platform
+cobalt/build/gn.py -p evergreen-arm-hardfp-rdk -c devel --no-rbe
+
+# Compile Cobalt binaries
+ninja -C out/evergreen-arm-hardfp-rdk_devel/ cobalt_loader nplb_loader
+```
+
+Build output will be placed in `$COBALT_BUILD_DIR/chromium/src/out/evergreen-arm-hardfp-rdk_devel/` on the host because of the bind mount.
+
+After building successfully, you can use `ctrl+d` to finish the docker execution.
+
+## 3. Deploying Cobalt on RDK
+
+### Configuration Variables
+
+Set these variables on your host machine before proceeding with deployment steps:
+
+```bash
+TARGET="root@<TARGET_IP>"
+COBALT_CONTENT_DIR="/data/out_cobalt"
+COBALT_BUILD_DIR="$HOME/cobalt-build"
+OUT="$COBALT_BUILD_DIR/chromium/src/out/evergreen-arm-hardfp-rdk_devel"
+```
+
+**Note:** Replace `<TARGET_IP>` with your RDK device's IP address.
+
+### Setup Target Device (One-time)
+
+Create the Cobalt environment setup script on the target device. This script sets up environment variables, kills WPE Framework (RDK UI menu), and ensures westeros is running.
+
+**Note:** This setup only needs to be done once on a fresh install of RDK.
+
+```bash
+ssh "$TARGET" "
+    mkdir -p $COBALT_CONTENT_DIR /data/cobalt/evergreen-arm-hardfp-rdk_devel /home/root
+    cat > /home/root/setup-cobalt-env.sh << 'SCRIPT_EOF'
+    #!/bin/sh
+    # Cobalt environment preparation for busybox
+    cd $COBALT_CONTENT_DIR
+    killall -9 elf_loader_sandbox loader_app westeros-init 2>/dev/null || true
+    sleep 1
+
+    # Stop WPE Framework if running
+    if systemctl is-active --quiet wpeframework; then
+        systemctl stop wpeframework >/dev/null 2>&1
+        sleep 2
+    fi
+
+    # Set up westeros environment
+    eval \$(systemctl show-environment | sed 's/^/export /') >/dev/null 2>&1
+    export WAYLAND_DISPLAY=wayland-0
+    export XDG_RUNTIME_DIR=/tmp
+
+    # Configure system parameters
+    echo 0 > /sys/module/drm/parameters/vblankoffdelay 2>/dev/null
+    echo 1 > /sys/module/aml_drm/parameters/video_axis_zoom 2>/dev/null
+
+    # Always kill and restart westeros compositor
+    killall -9 westeros westeros-init 2>/dev/null || true
+    sleep 1
+    /usr/bin/westeros-init >/dev/null 2>&1 &
+    sleep 3
+
+    echo 'Cobalt environment ready. WPE Framework stopped and westeros running.'
+    SCRIPT_EOF
+    chmod +x /home/root/setup-cobalt-env.sh"
+```
+
+**What the setup script does:**
+
+The `/home/root/setup-cobalt-env.sh` script is an auxiliary script used for development only. It sets up the Cobalt environment by:
+- Stopping WPE Framework (RDK UI) to free up the display
+- Starting westeros compositor for Cobalt's display functionality
+- Setting up environment variables and system parameters
+- Ensuring Cobalt has exclusive access to the display
+
+**Note:** WPE Framework is stopped as a development workaround to ensure Cobalt has exclusive display access. Future RDK patches will integrate Cobalt as a WPE plugin, eliminating the need for this workaround. The westeros compositor is required to provide Cobalt with keyboard input events and other RDK services.
+
+### Transfer Cobalt Files
+
+**Note:** This step can be repeated multiple times during development.
+
+Transfer Cobalt files to target using the variables defined in the Configuration Variables section above:
+
+**Method 1: Full deployment (tar) - Faster for transferring all the build artifacts**
+
+Use this method for initial deployment or when you want to ensure a clean state:
+
+```bash
+cd "$OUT"
+tar czf - -T cobalt_loader.runtime_deps | ssh "$TARGET" "cd $COBALT_CONTENT_DIR && tar xzf -"
+
+# For NPLB (Non-Platform Loader Binary) testing:
+# tar czf - -T nplb_loader.runtime_deps | ssh "$TARGET" "cd $COBALT_CONTENT_DIR && tar xzf -"
+```
+
+**Method 2: Incremental deployment (rsync) - Faster for updates**
+
+Use this method for faster incremental updates as it only transfers changed files:
+
+```bash
+cd "$OUT"
+rsync -avz --ignore-missing-args --files-from=cobalt_loader.runtime_deps ./ "$TARGET:$COBALT_CONTENT_DIR/"
+
+# For NPLB loader - only sync changed files:
+# rsync -avz --ignore-missing-args --files-from=nplb_loader.runtime_deps ./ "$TARGET:$COBALT_CONTENT_DIR/"
+```
+
+**Alternative: Using ADB for file transfer**
+
+If you have a USB type-C cable connected, you can use `adb` to deploy the runtime dependencies:
+
+```bash
+cd "$OUT"
+tar czf /tmp/cobalt_runtime_deps.tar.gz -T cobalt_loader.runtime_deps
+adb push /tmp/cobalt_runtime_deps.tar.gz $COBALT_CONTENT_DIR/
+adb shell "cd $COBALT_CONTENT_DIR && tar xzf cobalt_runtime_deps.tar.gz && rm cobalt_runtime_deps.tar.gz"
+rm /tmp/cobalt_runtime_deps.tar.gz
+```
+
+### Run on Target
+
+SSH into the target:
+
+```bash
+ssh "$TARGET"
+```
+
+Then, on the target device, run the commands interactively.
+
+**Important:** You must execute the `setup-cobalt-env.sh` script before running Cobalt each new `ssh` session.
+
+```bash
+source /home/root/setup-cobalt-env.sh; /data/out_cobalt/loader_app --content=/data/out_cobalt --url="https://www.youtube.com/tv"
+```
+
+## 4. Additional Information
+
+### Display Configuration
+
+The `/home/root/setup-cobalt-env.sh` script configures `WAYLAND_DISPLAY=wayland-0` and starts the westeros compositor for Cobalt display functionality. Execute this script before running Cobalt each time.
+
+### Runtime Dependencies
+
+Available build targets:
+- `cobalt_loader.runtime_deps` - Main Cobalt application
+- `nplb_loader.runtime_deps` - Non-Platform Loader Binary for testing
+
+### Important Notes
+
+- **Deployment Directory:** Uses `/data` for storage due to RDK storage restrictions. The `/data/` partition typically has more space and is designed for application deployments.
+- **Environment Setup:** Execute `/home/root/setup-cobalt-env.sh` before running Cobalt each time. This script handles all Cobalt environment setup including westeros compositor and WPE Framework management.
+- **Build Configuration:** This guide uses the `devel` configuration for faster iteration. For production builds, use `-c qa` or `-c gold`.


cobalt/docker/rdk/Dockerfile
```diff
@@ -12,9 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-ARG BASE_IMAGE_SHA=sha256:be15f84c44c4ee68b4f187128f0278df1b0f424c04fb5f08b098967d5b896388
-
-FROM marketplace.gcr.io/google/debian12@${BASE_IMAGE_SHA}
+FROM debian:12
```
lamdbui: @denisyuji , any reason why you had to switch this out? 

For all our Docker configs, we pin the image so that the environment is the same across all our ports (e.g. raspi-2, linux).

starboard/doc/setup-rdk.md
lamdbui, line 1: Let's rename to `setup-rdk.md` to match the format we used in C25. 
- https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/site/docs/development/setup-rdk.md

We can probably lift much of the common information as well from the [C25 version](https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/site/docs/development/setup-rdk.md) and see if we can keep it as consistent as we can:
- e.g. Prerequisites (has additional helpful platform info)

cobalt/docker/rdk/Dockerfile
```diff
@@ -43,36 +41,60 @@ RUN apt-get update && apt-get install -y \
...
+    libtool \
     && rm -rf /var/lib/apt/lists/*
 
-ARG GCLOUD=google-cloud-cli-linux-x86_64.tar.gz
-RUN cd /tmp \
-    && curl -L -O "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/${GCLOUD}" \
-    && tar xf "${GCLOUD}" -C /usr/local --strip-components=1 google-cloud-sdk/bin google-cloud-sdk/lib \
-    && rm "${GCLOUD}" \
-    && gcloud --version
+# Install Node.js via NVM
```
lamdbui: I'm not sure if this is necessary as I thought it was already setup in the image.

@niranjanyardi , any thoughts on this?

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,120 @@
...
+## 1. Build Docker Image
+
+The Dockerfile is located at `chromium/src/cobalt/docker/rdk/Dockerfile`.
+
+```bash
+cd chromium/src/cobalt/docker/rdk
+docker build -t rdk .
+```
+
+## 2. Fetch Cobalt Sources
```
lamdbui: I think we can probably cross-reference some of the case documentation here, to avoid some duplication and simplify.
- https://github.com/youtube/cobalt/blob/main/starboard/doc/eap/getting_started.md

cobalt/docker/rdk/Dockerfile
niranjanyardi: yea, i think it's already setup

niranjanyardi: @denisyuji there's already a github action which checks if cobalt builds for rdk: 

Image

so there's no need to modify the dockerfile AFAIK

cobalt/docker/rdk/Dockerfile
andrewsavage1: If we are going to change the dockerfile in this way, we should definitely have someone from the T&I team take a look and make sure this is sustainable

andrunko: From what I can see these dockerfile changes are not really required but more cosmetic? I would try to avoid changing the docker config unless really needed - reason for that is that the current config is based on the other configs we ship (uses same base image, etc) and already tested/working on both CI and locally, so I think any change here should be in sync with the other configs as well.

cobalt/docker/rdk/Dockerfile
andrunko: Same as mentioned previously, I would try to avoid changing the docker config unless really needed.

starboard/doc/setup-rdk.md
andrunko: Agree, I think it would be good to add info on setting up the device.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,120 @@
...
+
+**Build Output Location:** `$HOME/cobalt-build/chromium/src/out/evergreen-arm-hardfp-rdk_devel/`
+
+The bind mount ensures all build artifacts are directly accessible on your host machine after the build completes.
+
+## 4. Deploy to Target
+
+```bash
+TARGET="root@<TARGET_IP>"
+DEST="/opt/cobalt/evergreen-arm-hardfp-rdk_devel"
```
andrunko: We usually use `/data/` as the root directory when deploying to device due to storage restrictions, not sure if `/opt/` also has enough space (same partition? dont have the device to check now), but I would update it to `/data/` in any case as this is what we used previously on examples/tests/etc.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,120 @@
...
+# Cobalt environment variables
+WAYLAND_DISPLAY=wayland-0
+EOF
+"
+
+# Sync build artifacts
+rsync -az -e ssh "$OUT/app/" "$TARGET:$DEST/app/"
+rsync -az --include='*.pak' --include='*.pak.info' --exclude='*' -e ssh "$OUT/" "$TARGET:$DEST/"
+rsync -az -e ssh "$OUT/fonts/" "$TARGET:$DEST/fonts/"
+rsync -az --progress -e ssh "$OUT/elf_loader_sandbox" "$OUT/libcobalt.so" "$TARGET:$DEST/"
```
andrunko: Each GN target has a corresponding `.runtime_deps` file that gets generated during build and includes all files needed for that target during runtime.

I believe using it should be less error prone and make sure all required files are included. For example, to create a tarball for `loader_app` one would do:
```
tar czpf cobalt_loader.tgz -C out/evergreen-arm-hardfp-rdk_devel/ -T out/evergreen-arm-hardfp-rdk_devel/cobalt_loader.runtime_deps
```

Same applies to `nplb`, where one would use `nplb_loader.runtime_deps` instead.

andrunko: Also may be worth adding a note (or maybe even instructions) using `adb`, which is also mentioned in the C25 docs.

starboard/doc/rdk.md
```diff
@@ -0,0 +1,120 @@
...
+
+# Create destination directory
+ssh "$TARGET" "mkdir -p $DEST"
+
+# Configure environment variables
+ssh "$TARGET" "
+grep -q 'Cobalt environment' /etc/default/westeros-env || cat >> /etc/default/westeros-env << 'EOF'
+
+# Cobalt environment variables
+WAYLAND_DISPLAY=wayland-0
```
andrunko: How about the other env variables, are those not required? Currently I use the following when testing on device:
```
# Set environment
export WESTEROS_GL_USE_BEST_MODE=1
export WESTEROS_GL_MAX_MODE=3840x2160
export WESTEROS_GL_GRAPHICS_MAX_SIZE=1920x1080
export WESTEROS_GL_USE_AMLOGIC_AVSYNC=1
export WESTEROS_GL_REFRESH_PRIORITY=F,80
export WESTEROS_SINK_AMLOGIC_USE_DMABUF=1
export WESTEROS_SINK_USE_FREERUN=1
export WESTEROS_SINK_USE_ESSRMGR=1
export WESTEROS_GL_USE_REFRESH_LOCK=1
export WESTEROS_GL_USE_UEVENT_HOTPLUG=1
export RDKSHELL_KEYMAP_FILE=/etc/rdkshell_keymapping.json
export ESSOS_NO_EVENT_LOOP_THROTTLE=1
export AVPK_SKIP_HDMI_VALIDATION=1
echo 0 > /sys/module/drm/parameters/vblankoffdelay
echo 1 > /sys/module/aml_drm/parameters/video_axis_zoom
export LD_PRELOAD=/usr/lib/libwesteros_gl.so.0.0.0
export XDG_RUNTIME_DIR="/run"
export WAYLAND_DISPLAY=wayland-0
```

Not sure if all of them are required but please confirm.

We also need to set `COBALT_CONTENT_DIR` (e.g. `export COBALT_CONTENT_DIR=/data/test`), to where the runtime data was deployed otherwise cobalt will try to load the data from the default system prefix (e.g. `/usr/`).

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,120 @@
...
+    cd $DEST
+    killall -9 elf_loader_sandbox loader_app 2>/dev/null || true
+    systemctl stop wpeframework 2>/dev/null || true
+    sleep 1
+
+    # Source environment variables at runtime
+    set -a; . /etc/default/westeros-env; set +a
+    eval \$(systemctl show-environment | sed 's/^/export /')
+
+    ./elf_loader_sandbox --evergreen_content=. --evergreen_library=libcobalt.so --disable-hang-monitor
```
andrunko: Any reason why using `--disable-hang-monitor`? Unless there is a know issue which makes it required I would avoid mentioning it here - fwiw, for loader app, using `./loader_app` directly should also work, but using `elf_loader_sandbox` should be fine also.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,120 @@
...
+ssh "$TARGET" "
+    cd $DEST
+    killall -9 elf_loader_sandbox loader_app 2>/dev/null || true
+    systemctl stop wpeframework 2>/dev/null || true
+    sleep 1
+
+    # Source environment variables at runtime
+    set -a; . /etc/default/westeros-env; set +a
+    eval \$(systemctl show-environment | sed 's/^/export /')
+
```
andrunko, line 241: I guess we also need to start `westeros` here before running the app? At least for `loader_app` we should need it (`nplb` probably works without it but needs confirming).

```
/usr/bin/westeros-init &
```

starboard/doc/setup-rdk.md
denisyuji: Good catch. In my setup, Cobalt was being terminated by the hanging monitor. Since I fixed my setup, this flag is no longer necessary. I removed it.

starboard/doc/setup-rdk.md
denisyuji: Makes sense, renamed!

cobalt/docker/rdk/Dockerfile
denisyuji: That makes sense.Since I was carefully following the instruction on a fresh machine to validate on the user perspective, and I changed these lines to make easier for developers, since I was not being able to fetch this docker image from `marketplace.gcr.io` before doing:

```
gcloud auth login
gcloud auth configure-docker marketplace.gcr.io
```
I added this instructions, now `docker build` can fetch the images from `marketplace.gcr.io`.
I reverted the changes on the Dockerfile and adjusted the documentation.

cobalt/docker/rdk/Dockerfile
denisyuji: Make sense, this step was only required for the debian image from dockerhub. I removed the changes for the Dockerfile

starboard/doc/setup-rdk.md
denisyuji: Makes sense, it will be better for safety. I updated it to `/data/`

starboard/doc/setup-rdk.md
denisyuji: @lamdbui , that makes sense! I cross referenced the code fetching process. See below

starboard/doc/setup-rdk.md
denisyuji: That makes sense. I updated the doc with the instructions above. Thanks @andrunko

starboard/doc/rdk.md
denisyuji: Thanks, @andrunko,

The line `eval \$(systemctl show-environment | sed 's/^/export /')` will export most of the variables required, since:

```
root@AmlogicFirebolt:~# systemctl show-environment | sed 's/^/export /' 
export LANG=C
export PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
export AVPK_SKIP_HDMI_VALIDATION=1
export CTRLM_RESTART_DELAY=0
export DEBUGINIFILE=/etc/debug.ini
export ENABLE_FOG_OPT_OVERRIDE=1
export ESSOS_NO_EVENT_LOOP_THROTTLE=1
export GST_REGISTRY=/data/gstreamer/registry.arm.bin
export LD_PRELOAD=/usr/lib/libwesteros_gl.so.0.0.0:/usr/lib/libRCECHal.so
export NETSRVMGRINIFILE=/etc/netsrvmgr.conf
export RDKSHELL_KEYMAP_FILE=/etc/rdkshell_keymapping.json
export RMFCONFIGINIFILE=/etc/rmfconfig.ini
export WEBPANOTIFYINCFG=/etc/notify_webpa_cfg.json
export WESTEROS_GL_GRAPHICS_MAX_SIZE=1920x1080
export WESTEROS_GL_MAX_MODE=3840x2160
export WESTEROS_GL_MODE=2160p60hz
export WESTEROS_GL_REFRESH_PRIORITY=F,80
export WESTEROS_GL_USE_AMLOGIC_AVSYNC=1
export WESTEROS_GL_USE_BEST_MODE=1
export WESTEROS_GL_USE_GBM_MODIFIERS=1
export WESTEROS_GL_USE_REFRESH_LOCK=1
export WESTEROS_GL_USE_UEVENT_HOTPLUG=1
export WESTEROS_SINK_AMLOGIC_USE_DMABUF=1
export WESTEROS_SINK_USE_ESSRMGR=1
export WESTEROS_SINK_USE_FREERUN=1
export XDG_RUNTIME_DIR=/run
```

starboard/doc/setup-rdk.md
denisyuji: Thanks, @andrunko. I tested it by only killing Cobalt and keeping `westeros` running, so we don't need to restart it at every attempt. Does it makes sense?

andrunko: hmm, I think `westeros` is not running in this case but `wpeframework` instead? I haven't tried using the default wayland display as I'd always stop `wpeframework` and start `westeros` (that comes from instructions used on C25). I would probably keep that  for now and if the case update later on once the wpeframework plugin integration updates land.

If you update here please make sure to update the `Display Configuration` section and `WAYLAND_DISPLAY` setting acccordingly.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,160 @@
...
+## Runtime Dependencies
+
+**Available Targets:**
+- `cobalt_loader.runtime_deps` - Main Cobalt application
+- `nplb_loader.runtime_deps` - Non-Platform Loader Binary for testing
+
+## Notes
+
+- **Deployment Directory:** Uses `/data/cobalt/` instead of `/opt/cobalt/` due to RDK device storage partitioning. The `/data/` partition typically has more space and is designed for application deployments.
```
andrunko: I'm fine with this note, but not sure about the `instead of /opt/cobalt` given we never mentioned `/opt` before. Maybe rephrase to something like:
```
- **Deployment Directory:** Uses `/data`  for storage due to RDK storage restrictions. The `/data/` partition typically has more space and is designed for application deployments.
```

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,160 @@
...
+- **Environment Variables:** Cobalt requires specific environment variables for display functionality. The script now sets these directly for better reliability.
```
andrunko: Not clear what script this is referring to as we never had one before (or instructions)? I would either drop this note or rephrase it to something like:
```
- **Environment Variables:** Cobalt requires specific environment variables for display functionality. The script above should set these variables accordingly.
```

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,160 @@
...
+- **Westeros/Wayland Services:** These services are assumed to be always running. The script no longer attempts to stop/start them.
+- **Build Configuration:** This guide uses the `devel` configuration for faster iteration. For production builds, use `-c qa` or `-c gold`.
+- **Display Issues:** If Cobalt runs but doesn't display, verify the wayland display setting matches the active compositor.
```
andrunko: Please see previous comment on wpeframework/westerons and update here if needed.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Available Targets:**
+- `cobalt_loader.runtime_deps` - Main Cobalt application
+- `nplb_loader.runtime_deps` - Non-Platform Loader Binary for testing
+
+## Notes
+
+- **Deployment Directory:** Uses `/data/cobalt/` instead of `/opt/cobalt/` due to RDK device storage partitioning. The `/data/` partition typically has more space and is designed for application deployments.
+- **Environment Variables:** Cobalt requires specific environment variables for display functionality. The script now sets these directly for better reliability.
+- **Westeros/Wayland Services:** These services are assumed to be always running. The script no longer attempts to stop/start them.
```
andrunko: As above, not clear what script this refers to as we never had one (so "no longer" doesnt apply). Also if we go with the "stop wpeframework and start westeros"  route this should be dropped, please check.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,160 @@
+# Building and Deploying Cobalt for RDK
+
+## Prerequisites
+
+- Docker installed on your build machine
+- SSH access to your RDK target device
+- ~20GB free disk space for Cobalt sources and build artifacts
```
andrunko: Can we add some instructions on setting up the device as done in the C25 docs?

starboard/doc/setup-rdk.md
denisyuji: Ok! updated!

starboard/doc/setup-rdk.md
denisyuji: Yes, it was not clear that it is the script inside ssh command. updated

starboard/doc/setup-rdk.md
denisyuji: @andrunko , sorry, you are right, `westeros` is not running when the device boots. Instead, ResidentApp (the UI) uses direct direct DRM/KMS rendering (via ESSOS). However, RDK creates per default the Wayland socket (location: `/run/ResidentApp/wst-ResidentApp` ), and since ResidentApp is already managed by RDKShell it integrates seamslessly with RDK. I tested it multiple times on fresh image to certify Cobalt is displayed consistently with `WAYLAND_DISPLAY=ResidentApp/wst-ResidentApp`. I updated the `Display Configuration` as per your instructions. Thanks for pointing that.

starboard/doc/setup-rdk.md
denisyuji: Thx! Please see previous comment above

starboard/doc/setup-rdk.md
denisyuji: Thanks! I fixed it. We will not stop wpe framework, only start the application using the ResidentApp's wayland socket.

starboard/doc/setup-rdk.md
denisyuji: Makes sense! I copied the `Get an RDK device running the correct image` section from the C25 doc!

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+## 2. Fetch Cobalt Sources
+
+First follow the steps on , see For  
+
+1. **See the detailed source setup instructions** on the [Open Source Setup document](cobalt/starboard/doc/eap/getting_started.md). Follow the standard Cobalt until you configure gclient and download subrepos. **Note:** For the instructions below, you do **NOT** need to install build dependencies or build Evergreen for x64, as it is demonstrated on the doc.
+2. **Install required packages and depot_tools** (see starboard documentation)
+3. **Clone Cobalt repository and configure gclient** (see starboard documentation)  
+4. **Install build dependencies:** `./build/install-build-deps.sh` (see starboard documentation)
```
andrunko: Afaik you dont need to install build deps to build using docker as the docker image already does it, but in any case it won't hurt.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,174 @@
...
+# Alternative: For NPLB (Non-Platform Loader Binary) testing
+# tar czf - -T nplb_loader.runtime_deps | ssh "$TARGET" "cd $DEST && tar xzf -"
+```
+
+### Alternative: Using ADB for File Transfer
+
+If you have a USB type-C cable connected, you can use `adb` to update the library file:
+
+```sh
+adb push libloader_app.so /usr/lib/libloader_app.so
```
andrunko: Replacing the system `libloader_app.so` is only needed if you are replacing the system installed cobalt (C25). This may be confusing, I suggest to change this example to copy the cobalt_loader tarball instead (created from `cobalt_loader.runtime_deps`).

starboard/doc/setup-rdk.md
andrunko: Right, I haven't tried running with the default wayland display and don't have a device to verify atm but if it works I am Ok with it - note though I'm not sure if using the same display from wpeframework (but without the wpeframework plugin integration) and leave wpeframework running can cause some unrelated issue (e.g. app stopping due to some watchdog), so let's see what others say here but as I said, if it works I'm Ok with it.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,174 @@
...
+    ./elf_loader_sandbox --evergreen_content=. --evergreen_library=libcobalt.so
+
+    # To run NPLB:
+    # ./elf_loader_sandbox --evergreen_content=. --evergreen_library=libnplb.so
+"
+```
+
+## Display Configuration
+
+**Wayland Display Setup:** We use `WAYLAND_DISPLAY=ResidentApp/wst-ResidentApp` because it provides RDKShell coordination for proper Z-order management and follows the standard RDK integration pattern. The socket is located at `/run/ResidentApp/wst-ResidentApp`, though ResidentApp itself uses direct rendering via Essos rather than westeros compositing.
```
andrunko: I would not even mention westeros here if we dont use it as per the docs to avoid confusion.

andrunko (commented):
Some minor suggestions, otherwise LGTM.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,174 @@
...
+TARGET="root@<TARGET_IP>"
+DEST="/data/cobalt/evergreen-arm-hardfp-rdk_devel"
+OUT="$HOME/cobalt-build/chromium/src/out/evergreen-arm-hardfp-rdk_devel"
+
+# Create destination directory
+ssh "$TARGET" "mkdir -p $DEST"
+
+# Configure environment variables
+ssh "$TARGET" "
+grep -q 'Cobalt environment' /etc/default/westeros-env || cat >> /etc/default/westeros-env << 'EOF'
```
andrunko: One more thing, we should rename `westeros-env` here if we end up not using westeros.

andrunko: Looking at it again, looks like this file is no longer in use in the instructions? Please verify.

starboard/doc/setup-rdk.md
denisyuji: good catch! I removed it!

starboard/doc/setup-rdk.md
denisyuji: Thanks, @andrewsavage1 . I changed it to copy all the files

starboard/doc/setup-rdk.md
denisyuji: ok, I removed the mention about westeros

andrunko: Did you forget to push? I still see it mentioned here? Also please check the previous comment about "westeros-env".

denisyuji: @andrunko , sorry, I forgot to remove this `westeros-env` file.

starboard/doc/setup-rdk.md
denisyuji: I removed this `westeros-env` file.

andrunko (approved):
LGTM.

starboard/doc/setup-rdk.md
lamdbui: Thanks both. I'd say in the "Display Configuration" section, if we can make clear if it's required or not, it'd be helpful. Sounds like we need to export it based on the run instructions.

Given that we are generally working with the same base image, would it just make sense to have a script to auto-configure these things such as environment variables to simplify?

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+```bash
+cd chromium/src/cobalt/docker/rdk
+docker build -t rdk .
+```
+
+## 2. Fetch Cobalt Sources
+
+First follow the steps on , see For  
+
+1. **See the detailed source setup instructions** on the [Open Source Setup document](cobalt/starboard/doc/eap/getting_started.md). Follow the standard Cobalt until you configure gclient and download subrepos. **Note:** For the instructions below, you do **NOT** need to install build dependencies or build Evergreen for x64, as it is demonstrated on the doc.
```
jonastsai: This link is broken: [Open Source Setup document](cobalt/starboard/doc/eap/getting_started.md)

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+2. **Install required packages and depot_tools** (see starboard documentation)
+3. **Clone Cobalt repository and configure gclient** (see starboard documentation)
+
+**Important Platform Difference:** The starboard documentation on `cobalt/starboard/doc/eap/getting_started.md` demonstrates building for `evergreen-x64` (Linux desktop). For RDK boards, you will use the ARM platform instead: `evergreen-arm-hardfp-rdk`. Skip the starboard build for `evergreen-x64` and proceed to the RDK-specific build section below.
```
jonastsai: For `cobalt/starboard/doc/eap/getting_started.md`, please use a hybrid link like  [Open Source Setup document] above instead show the full documents

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+docker build -t rdk .
+```
+
+## 2. Fetch Cobalt Sources
+
+First follow the steps on , see For  
+
+1. **See the detailed source setup instructions** on the [Open Source Setup document](cobalt/starboard/doc/eap/getting_started.md). Follow the standard Cobalt until you configure gclient and download subrepos. **Note:** For the instructions below, you do **NOT** need to install build dependencies or build Evergreen for x64, as it is demonstrated on the doc.
+2. **Install required packages and depot_tools** (see starboard documentation)
+3. **Clone Cobalt repository and configure gclient** (see starboard documentation)
```
jonastsai: It is good to reuse the same document in starboard/doc/getting_started.md to get source code and setup the workstation. But the instruction here makes the user confused. Can you also update getting_started.md and split the first section into "Setup workstation", "Get source code" etc. and the document here can be simplify tell the user to finish "Setup workstation", "Get source code" steps in the getting_started.md file.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+```bash
+TARGET="root@<TARGET_IP>"
+DEST="/data/cobalt/evergreen-arm-hardfp-rdk_devel"
+COBALT_CONTENT_DIR="/data/cobalt-content"
+OUT="$HOME/cobalt-build/chromium/src/out/evergreen-arm-hardfp-rdk_devel"
+
+ssh "$TARGET" "
+    mkdir -p $COBALT_CONTENT_DIR $DEST
+    rm -rf /usr/share/content/data
+    ln -sf $COBALT_CONTENT_DIR /usr/share/content/data
```
jonastsai: Since you are running Cobalt as an executable via elf_loader_sandbox or loader_app, why will you remove /usr/share/content/data and link it to /usr/share/content/data? It will break the existing CB25.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+
+All the commands below runs on target and uses `ssh` to communicate with the target when needed. Ensure you followed the instructions from the section **SSH Setup** in the beginning of this doc.
+
+**Note:** The `elf_loader_sandbox` expects Cobalt files to be in `/usr/share/content/data/`, but this directory is on the root filesystem which may be full. We use a symbolic link to redirect it to the `/data` partition which has more space.
+
+Set the required environment variables, create destination directory on target and setup RDK content directory:
+
+```bash
+TARGET="root@<TARGET_IP>"
+DEST="/data/cobalt/evergreen-arm-hardfp-rdk_devel"
```
jonastsai: This folder is not used

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+COBALT_CONTENT_DIR="/data/cobalt-content"
```
jonastsai: The original folder in the document is `/data/out_cobalt`. Both folders are fine to me but please keep them consistent.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+## 3. Build Cobalt
+
+The Docker container bind-mounts `$HOME/cobalt-build` to access sources and write build artifacts:
+
+```bash
+COBALT_BUILD_DIR="$HOME/cobalt-build"
+
+docker run --rm -v $COBALT_BUILD_DIR:$COBALT_BUILD_DIR -t rdk bash -c "
+    git config --global --add safe.directory '*'
+    export PATH=$COBALT_BUILD_DIR/depot_tools:\$PATH
```
jonastsai: the `depot_tools` is not installed in `$COBALT_BUILD_DIR` if user follow the steps in `getting_started.md`

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,201 @@
...
+
+**RDK-specific directory setup:**
+```bash
+COBALT_BUILD_DIR="$HOME/cobalt-build"
+mkdir -p "$COBALT_BUILD_DIR"
+cd "$COBALT_BUILD_DIR"
+# Continue with standard setup from starboard docs, then return here for RDK-specific build
+```
+
+## 3. Build Cobalt
```
jonastsai: Can we have the steps to build Chrobalt without docker, too?

starboard/doc/setup-rdk.md
andrunko: Good catch, missed it.

starboard/doc/setup-rdk.md
denisyuji: Thanks @lamdbui . I've just add instructions to create a `.profile` script that will stop wpe framework, start westeros and source the env variables on every run. I tried to make Cobalt altogether with the AI, but it was not reliable, since it sometimes starts in wrong z-order (behind the RDK's Home UI). So, I will go back to the previous strategy to kill wpe framework and start westeros.

starboard/doc/setup-rdk.md
denisyuji: Thank you! Fixed it!

starboard/doc/setup-rdk.md
denisyuji: ack, @jonastsai , I am working to adapt the doc to show how to build it without docker

starboard/doc/setup-rdk.md
denisyuji: Thx! I removed it

starboard/doc/setup-rdk.md
denisyuji: Thx, @jonastsai . I fixed it

starboard/doc/setup-rdk.md
denisyuji: Ok, I am working modifying `starboard/doc/getting_started.md`

starboard/doc/setup-rdk.md
denisyuji: good catch! Changed it to `/data/out_cobalt`

starboard/doc/setup-rdk.md
denisyuji: that makes sense. I am removing this `rm` and symlink and using `--evergreen_content=.` instead

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,230 @@
...
+rm /tmp/cobalt_runtime_deps.tar.gz
+```
+
+## 6. Run on Target
+
+**Note:** The environment is automatically set up via `/home/root/.profile` on every interactive SSH session. See previous section on creation of this file.
+
+```bash
+ssh "$TARGET" "cd /data/out_cobalt && ./elf_loader_sandbox --evergreen_content=. --evergreen_library=libcobalt.so"
+```
```
niranjanyardi: we generall launch libcobalt.so with loader_app, elf_loader_sandbox is generally used for tests. Can you please update the instructions here to run with loader_app

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,255 @@
...
+
+This is required because the `$COBALT_BUILD_DIR/chromium/src/cobalt/docker/rdk/Dockerfile` pulls base images from Google's marketplace registry.
+
+### Build the RDK Docker Image
+
+Navigate to the Docker directory and build the image:
+
+```bash
+cd chromium/src/cobalt/docker/rdk
+docker build --progress=plain -t rdk -f "$COBALT_BUILD_DIR/chromium/src/cobalt/docker/rdk/Dockerfile" "$COBALT_BUILD_DIR/chromium/src/cobalt/docker/rdk"
```
andrunko: Should we use docker-compose here instead? I build it with "docker-compose run rdk" (we ship a docker-compose.yml at the root dir), any reason to use docker directly?

andrunko: iirc CI also uses docker compose here, see `.github/actions/docker/action.yaml` - on C25 docker-compose.yml would already mount some dirs, this was simplified on C26 but I think we can still use "docker compose" here to build at least (`docker compose build rdk`) and then if needed use docker directly to run the container by passing the extra mount points.

starboard/doc/setup-rdk.md
denisyuji: @niranjanyardi , thanks, it makes sense, I changed it to use `loader_app`.
Just FYI, while I was testing this change I faced issues building `cobalt_loader` in the `26.eap` branch (A Warning being treated as an error - This warning was not showing up in previous commits), which was leading the build to not produce `loader_app` . This took me some time investigating. I will treat this build warning/error of `26.eap` branch in a separated thread and just put the instruction for `loader_app` for now.

starboard/doc/setup-rdk.md
denisyuji: @jonastsai , I updated it, please, see [starboard/doc/eap/getting_started.md](https://github.com/denisyuji/cobalt/blob/main/starboard/doc/eap/getting_started.md) from this PR.

starboard/doc/setup-rdk.md
denisyuji: @jonastsai , good catch, I reverted it to bind mount

starboard/doc/setup-rdk.md
denisyuji: @jonastsai , since this will add some more lines of complexity to this doc, it would be ok for you if I add these lines on a separated PR?

starboard/doc/setup-rdk.md
denisyuji: @andrunko , the main rationale for this is to provide a clean instruction for the user to run the image and build in a single command:

```bash
docker run --rm \
    -v $COBALT_BUILD_DIR:$COBALT_BUILD_DIR \
    -v $DEPOT_TOOLS_DIR:$DEPOT_TOOLS_DIR \
    -t rdk bash -c "
    # Configure git for safe directory access
    git config --global --add safe.directory '*'
    
    # Add depot_tools to PATH for build tools (gn, ninja)
    export PATH=$DEPOT_TOOLS_DIR:\$PATH
    
    # Navigate to Cobalt source directory
    cd $COBALT_BUILD_DIR/chromium/src
    
    # Generate build files for RDK platform
    cobalt/build/gn.py -p evergreen-arm-hardfp-rdk -c devel --no-rbe
    
    # Compile Cobalt binaries
    ninja -C out/evergreen-arm-hardfp-rdk_devel/ cobalt_loader nplb_loader
"
```

That would prevent user to make confusion about what commands should run inside/outside container. 
Do you think we need to revert to docker-compose, than put some **warning** messages explaining that the commands should run inside the container?

andrunko: I would probably go with something like:
```
# Docker build instructions

## Outside container
- xyz
- ...
- <build/enter container>

## Inside container

- ...
```

I think having a single command to run it is probably easier once you get used to it but for instructions I think it is probably easier to follow step-by-step instructions.

But no hard requirement here, as long as the commands work I'm fine either way.

starboard/doc/eap/getting_started.md
andrewsavage1, line 1: IMO the changes to this file are just unnecessary, but if other people think they're helpful I'll defer

starboard/doc/eap/getting_started.md
```diff
@@ -1,39 +1,67 @@
 # Open Source Setup
 
+## Setup workstation ##
+
+Install necessary packages. In this example, we consider installing `depot_tools` in $HOME. We will add the `depot_tools` directory to `PATH` for this session (consider adding this to `.bashrc` or `.profile`):
```
andrewsavage1: > In this example, we consider installing `depot_tools` in $HOME. We will add the `depot_tools` directory to `PATH` for this session (consider adding this to `.bashrc` or `.profile`)

This is currently misleading. The below `git clone` command installs depot_tools to `$PWD/depot_tools`, which is not necessarily the home directory. Change this to `$HOME/depot_tools` to fix.

Additionally this statement could use better phrasing. Change it to i.e. "We install `depot_tools` to `$HOME` and add the `depot_tools` directory to `PATH` for this session. Consider also adding the export statement to your `~/.bashrc` or equivalent as well."

But personally, I think the way I had it before was better.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,255 @@
...
+
+**Method 1: Full deployment (tar) - Faster for transferring all the build artifacts**
+
+Use this method for initial deployment or when you want to ensure a clean state:
+
+```bash
+cd "$OUT"
+tar czf - -T cobalt_loader.runtime_deps | ssh "$TARGET" "cd $COBALT_CONTENT_DIR && tar xzf -"
+
+# For NPLB (Non-Platform Loader Binary) testing:
```
andrewsavage1, line 204: NPLB stands for "No Platform Left Behind"

denisyuji: Thanks @andrewsavage1 ,  I removed this acronym explanation

starboard/doc/eap/getting_started.md
denisyuji: Thanks @andrewsavage1. I reverted to your original version.
I only split the content in starboard/doc/eap/getting_started.md to make it clearer for users coming from starboard/doc/setup-rdk.md so they can easily see which steps they need to follow. Let me know if it sounds to you now.

starboard/doc/eap/getting_started.md
```diff
@@ -1,39 +1,72 @@
...
+
 ```
-# Install necessary packages
 sudo apt update && sudo apt install -qqy --no-install-recommends \
   git curl python3-dev xz-utils lsb-release file
+```
 
-# Install depot_tools
-git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+Clone `depot_tools` and add the directory to `PATH` for this session. Consider adding this to `.bashrc`, `.profile`, or equivalent, to make it permanent:
```
andrewsavage1: "This" is ambiguous here. Rephrase, i.e. with "Consider adding depot_tools to your PATH by adding the export call to `~/.bashrc`, `~/.profile`, or equivalent to make it permanent."

starboard/doc/eap/getting_started.md
andrewsavage1: Thanks for addressing, splitting it up into separate code blocks as you to LGTM

starboard/doc/eap/getting_started.md
niranjanyardi: +1, let's do it in another PR if needed. It'll be good to get these instructions landed and copied over to EAp where partners can take a look

starboard/doc/setup-rdk.md
niranjanyardi: I'm reading the existing doc and i think it's okay to point to the existing sections . we can always modify the instructions in starboard/doc/eap/getting_started.md in a follow up if needed .

starboard/doc/eap/getting_started.md
denisyuji: good catch, @andrewsavage1, I rephrased it!

starboard/doc/setup-rdk.md
denisyuji: @andrunko , I separated the image build/run commands and the commands inside the container. Also, I added instruction to use `docker compose`. Check if it is clear now

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,263 @@
...
+```
+
+Then, build the `rdk` image and launch an interactive shell inside a temporary container:
+
+```bash
+# Build (if needed) and run an interactive shell in the rdk service container
+docker compose -f "$COBALT_BUILD_DIR/chromium/src/docker-compose.yaml" run --rm --build \
+  -v "$COBALT_BUILD_DIR:$COBALT_BUILD_DIR" \
+  -v "$DEPOT_TOOLS_DIR:$DEPOT_TOOLS_DIR" \
+  -e DEPOT_TOOLS_DIR="$DEPOT_TOOLS_DIR" \
```
andrunko, line 92: Do we need to set COBALT_BUILD_DIR env variable as well here, looks like it is used inside the container below without getting set?

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,263 @@
...
+```bash
+COBALT_BUILD_DIR="$HOME/cobalt-build"
+DEPOT_TOOLS_DIR="/path/to/depot_tools"
+```
+
+Then, build the `rdk` image and launch an interactive shell inside a temporary container:
+
+```bash
+# Build (if needed) and run an interactive shell in the rdk service container
+docker compose -f "$COBALT_BUILD_DIR/chromium/src/docker-compose.yaml" run --rm --build \
```
andrunko, line 88: You can skip the `-f ...` if running from the same dir which has the `docker-compose.yaml` file, but it doesn't hurt to pass it anyway.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,263 @@
...
+```bash
+ssh "$TARGET"
+```
+
+Then, on the target device, run the commands interactively.
+
+**Important:** You must execute the `setup-cobalt-env.sh` script before running Cobalt each new `ssh` session.
+
+```bash
+source /home/root/setup-cobalt-env.sh; /data/out_cobalt/loader_app --content=/data/out_cobalt --url="https://www.youtube.com/tv"
```
andrunko, line 245: Have you confirmed `--content=` does the same as setting `COBALT_CONTENT_DIR` env variable? Asking cause I haven't tried with `--content=` before.

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,263 @@
...
+```bash
+ssh-copy-id root@<TARGET_IP>
+ssh root@<TARGET_IP> echo "SSH working"
+```
+
+**Note:** Replace `<TARGET_IP>` with your device's IP address.
+
+## 1. Fetch Cobalt Sources
+
+Follow the standard Cobalt setup process from [the starboard documentation](cobalt/starboard/doc/eap/getting_started.md):
```
jonastsai: This URL still doesn't work. Try `/starboard/doc/eap/getting_started.md`
and verify it on https://github.com/denisyuji/cobalt/blob/main/starboard/doc/setup-rdk.md

starboard/doc/setup-rdk.md
```diff
@@ -0,0 +1,263 @@
...
+
+1. **Install required packages and depot_tools** (see [Setup workstation](cobalt/starboard/doc/eap/getting_started.md#setup-workstation))
+2. **Clone Cobalt repository and configure gclient** (see [Get source code](cobalt/starboard/doc/eap/getting_started.md#get-source-code))
```
jonastsai: These two links are broken, too

starboard/doc/setup-rdk.md
denisyuji: I'll leave with `-f` so user can call the command from anywhere

starboard/doc/setup-rdk.md
denisyuji: Yes, I validated that they have the same effect:

```bash
root@AmlogicFirebolt:/data/out_cobalt# ./loader_app --content=/data/out_cobalt -h
[loader_app/11692:1104/163525.616032(UTC):INFO:loader_app.cc(242)] alternative_content=/data/out_cobalt
[loader_app/11692:1104/163525.624157(UTC):INFO:slot_management.cc(423)] content=/data/out_cobalt
[loader_app/11692:1104/163525.624196(UTC):INFO:elf_loader.cc(82)] evergreen_config: content_path=/data/out_cobalt
[loader_app/11692:1104/163525.830018(UTC):INFO:elf_loader.cc(82)] evergreen_config: content_path=/data/out_cobalt
[loader_app/11692:1104/163525.830054(UTC):INFO:elf_loader_impl.cc(61)] Loading /data/out_cobalt/app/cobalt/lib/libcobalt.lz4 using compression
[loader_app/11692:1104/163533.724744(UTC):FATAL:loader_app.cc(139)] Check failed: false. Failed to load library at '/data/out_cobalt/app/cobalt/lib/libcobalt.lz4'.
```

```bash
root@AmlogicFirebolt:/data/out_cobalt# ./loader_app --content=/data/out_cobalt_2/ -h
[loader_app/12264:1104/163617.478129(UTC):INFO:loader_app.cc(242)] alternative_content=/data/out_cobalt_2/
[loader_app/12264:1104/163617.487214(UTC):INFO:slot_management.cc(423)] content=/data/out_cobalt_2/
[loader_app/12264:1104/163617.487254(UTC):INFO:elf_loader.cc(82)] evergreen_config: content_path=/data/out_cobalt_2/
[loader_app/12264:1104/163617.719597(UTC):INFO:elf_loader.cc(82)] evergreen_config: content_path=/data/out_cobalt_2/
[loader_app/12264:1104/163617.719659(UTC):INFO:elf_loader_impl.cc(61)] Loading /data/out_cobalt_2//app/cobalt/lib/libcobalt.lz4 using compression
[loader_app/12264:1104/163624.112093(UTC):FATAL:loader_app.cc(139)] Check failed: false. Failed to load library at '/data/out_cobalt_2//app/cobalt/lib/libcobalt.lz4'.
```

```bash
root@AmlogicFirebolt:/data/out_cobalt# COBALT_CONTENT_DIR=/data/out_cobalt ./loader_app -h
[1104/163717.545836:WARNING:wrapper.cc(214)] ca_certificates_path is unused:/data/out_cobalt/app/cobalt/content/ssl/certs
[loader_app/12905:1104/163717.549485(UTC):INFO:slot_management.cc(410)] lib_path=/data/out_cobalt/app/cobalt/lib/libcobalt.lz4
[loader_app/12905:1104/163717.549529(UTC):INFO:slot_management.cc(423)] content=/data/out_cobalt/app/cobalt/content
[loader_app/12905:1104/163717.549563(UTC):INFO:elf_loader.cc(82)] evergreen_config: content_path=/data/out_cobalt/app/cobalt/content
[loader_app/12905:1104/163717.549599(UTC):INFO:elf_loader_impl.cc(61)] Loading /data/out_cobalt/app/cobalt/lib/libcobalt.lz4 using compression
[loader_app/12905:1104/163723.932743(UTC):ERROR:loader_app.cc(126)] No library found at compressed /data/out_cobalt/app/cobalt/content/app/cobalt/lib/libcobalt.lz4 or uncompressed /data/out_cobalt/app/cobalt/content/app/cobalt/lib/libcobalt.so path
```

```bash
root@AmlogicFirebolt:/data/out_cobalt# COBALT_CONTENT_DIR=/data/out_cobalt_2/ ./loader_app -h
[1104/163743.486065:WARNING:wrapper.cc(214)] ca_certificates_path is unused:/data/out_cobalt_2//app/cobalt/content/ssl/certs
[loader_app/13161:1104/163743.489751(UTC):INFO:slot_management.cc(410)] lib_path=/data/out_cobalt_2//app/cobalt/lib/libcobalt.lz4
[loader_app/13161:1104/163743.489787(UTC):INFO:slot_management.cc(423)] content=/data/out_cobalt_2//app/cobalt/content
[loader_app/13161:1104/163743.489820(UTC):INFO:elf_loader.cc(82)] evergreen_config: content_path=/data/out_cobalt_2//app/cobalt/content
[loader_app/13161:1104/163743.489860(UTC):INFO:elf_loader_impl.cc(61)] Loading /data/out_cobalt_2//app/cobalt/lib/libcobalt.lz4 using compression
[loader_app/13161:1104/163749.875085(UTC):ERROR:loader_app.cc(126)] No library found at compressed /data/out_cobalt_2//app/cobalt/content/app/cobalt/lib/libcobalt.lz4 or uncompressed /data/out_cobalt_2//app/cobalt/content/app/cobalt/lib/libcobalt.so path
```

starboard/doc/setup-rdk.md
denisyuji: Thanks @jonastsai ! I had to merge/validate the different instruction sources we have, and also address the feedback from all, so that took a bit, but this flow is now tested and working.
I updated the links and we are ready to move ahead with the merge.

starboard/doc/setup-rdk.md
denisyuji: fixed!

starboard/doc/setup-rdk.md
denisyuji: Oops! sorry! you are right. I fixed it

andrunko: Yes, that is why I asked, but the command above is *not* setting it atm.

jonastsai:
LGTM. The broken links are also fixed.

PR #7511, opened 2025-10-11
starboard: Remove unused SbWindowGetDiagonalSizeInInches

kjyoun:
The SbWindowGetDiagonalSizeInInches function is deprecated and no longer in use.

This was discovered during a JNI migration. The function was using a legacy JNI implementation, but it has no callers. Instead of migrating the unused code, it is being deleted.

Bug: 150410605
Bug: 450024477


andrewsavage1:
It is a bigger deal to remove Starboard functions than it is anything else—some may be unused currently due to Chrobalt not yet implementing everything, but we are discussing whether or not we can the h5vcc GetDiagonal web api from Cobalt or if it is still necessary, which will determine whether or not we keep this function (see b/376893909). For now, it does not look like we can remove this function.

kjyoun:
Understood.

Reverted this PR. Instead, will create another PR to migrate JNI code in this file.

PR #7516, opened 2025-10-13
cobalt/shell: Remove unnecessary enable_arc config

rakuco:
As of M138, `//build/config/compiler:enable_arc` is opt-out rather than opt-in, so there is no need to explicitly depend on it here.

Bug: 433180638


rakuco:

rakuco:
None of the CI failures are related, the CI bots seem to be having problems authenticating with the RBE backend

yell0wd0g (approved):
LGTM I guess? : )

yell0wd0g:
@sideb0ard WDYT?

PR #7525, opened 2025-10-13
cobalt/browser: Disable QUIC by default

borongc:
This allows to run experiments to see if this can improve the overall performance, especially on low-end-devices, as QUIC has more CPU usages on m114 than m138, which Cobalt is currently based on.

Issue: 448423139
Issue: 451631829

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -75,6 +75,9 @@ public final class CommandLineOverrideHelper {
         paramOverrides.add("--disable-accelerated-video-encode");
         // Rasterize Tiles directly to GPU memory.
         paramOverrides.add("--enable-zero-copy");
+        // Disable QUIC to save CPU budgets on m114.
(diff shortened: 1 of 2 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -75,6 +75,8 @@ public static List<String> getDefaultCommandLineOverridesList() {
         paramOverrides.add("--disable-accelerated-video-encode");
         // Rasterize Tiles directly to GPU memory.
         paramOverrides.add("--enable-zero-copy");
+        // Disable QUIC to save CPU budgets on m114.
```
johnxwork, line 78: Add a TODO for removing on m138 so the rebase work group don't CP this.

borongc: Done.

kaidokert:
Did we verify this works ? We had a synthetic test page that let us test throughput impact easily for this.

borongc:
> Did we verify this works ? We had a synthetic test page that let us test throughput impact easily for this.

I can test it, and how should I setup it?

PR #7527, opened 2025-10-13
cobalt: Tune CC GPU memory settings to reduce footprint

sherryzy:
This change tunes two GPU memory settings in the compositor to reduce its overall memory footprint, particularly on Android devices.

The primary change reduces the amount of memory allocated for offscreen content prepainting. On most Android devices, this was previously set to use up to 67% of the compositor's memory budget for prepainting textures outside the visible viewport. This is lowered to 25%.

Experimental results on an android-qa build on Sabrina show this change significantly reduces the memory usage of the compositor (the 'cc' component). When navigating quickly through video content, memory usage dropped from ~70-80 MB to a more stable ~40-50 MB. Further local testing showed that reducing this to 0% (disabling prepainting) could lower usage to 30-40 MB with a minimal impact on user experience. We are proceeding with 25% for now, but could consider further reductions in the future.

Additionally, the GPU memory allocation priority was changed from CUTOFF_ALLOW_NICE_TO_HAVE to CUTOFF_ALLOW_REQUIRED_ONLY. While this did not show a significant memory reduction in testing, it is a logical change for our use case, as scrolling is not a primary interaction on most target devices, making "nice-to-have" allocations for scrolling less critical.

It's worth noting that while the compositor's component memory decreased, the peak RSS numbers reported by memory-infra did not show a clear reduction. However, this metric is known to fluctuate. The impact of these changes on overall system memory should be verified in production.

Bug: 424591313

### DIFF
--- a/third_party/blink/renderer/platform/widget/compositing/layer_tree_settings.cc
+++ b/third_party/blink/renderer/platform/widget/compositing/layer_tree_settings.cc
@@ -91,7 +91,11 @@ cc::ManagedMemoryPolicy GetGpuMemoryPolicy(
   cc::ManagedMemoryPolicy actual = default_policy;
   actual.bytes_limit_when_visible = 0;
   actual.priority_cutoff_when_visible =
+#if BUILDFLAG(IS_COBALT)
+      gpu::MemoryAllocation::CUTOFF_ALLOW_REQUIRED_ONLY;
+#else
       gpu::MemoryAllocation::CUTOFF_ALLOW_NICE_TO_HAVE;
+#endif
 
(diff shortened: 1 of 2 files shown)


johnxwork:
If you have the apk, can you try it on the Bravia in office? If there's no obvious regression on rendering, this should be a good change.

sherryzy:
Tested in Bravia, everything seems working fine with this change. I'm being very aggressive changed it to 0 and still didn't notice any bad user experience.

sherryzy:
Applying the cherry-pick label prior to PR submission may result in hung CI test suites.

yell0wd0g (approved):
\o/

PR #7528, opened 2025-10-13
Flush cookies and local storage added to h5vcc.

ghost:
Adding h5vcc.storage.flush() similar to c25 in support of STFC60.

We have a client app that updates the cookies and local storage in app lifecycle handlers. It calls h5vcc.storage.flush() which in c25 persisted cookies and local storage to disk. This is to ensure after the app is suspended, the values are persisted. It's possible for the app to be killed before this.

In a follow-up, I'll implement flushing cookies and local storage when blur/conceal/freeze happens similar to c25

https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/browser/browser_module.cc#L1848

This will likely first be for android with an additional PR for the 3P and other starboard platforms. Chrobalt android has CobalyActivity.onPause() and CobaltActivity.onStop(). It doesn't have full support of the rest of the lifecycle.

The API changes slightly, but this shouldn't cause an issue since the argument is ignored in c25 and will be ignored in c26. For c26, the returned promise can be ignored or used for sequencing.

c25:
```
interface H5vccStorage {
  // We don't support synchronous flush, this is legacy.
  void flush(optional boolean sync = false);
};
```

c26:
```
interface H5vccStorage {
  Promise<void> flush();
};
```

Issue: 371433332

### DIFF
--- a/cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
+++ b/cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
@@ -1,0 +1,52 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cobalt/browser/h5vcc_storage/h5vcc_storage_impl.h"
+
+// #include "build/build_config.h"
+#include "content/public/browser/storage_partition.h"
+#include "services/network/public/mojom/cookie_manager.mojom.h"
+
+namespace h5vcc_storage {
+
+H5vccStorageImpl::H5vccStorageImpl(
+    content::RenderFrameHost& render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver)
+    : content::DocumentService<mojom::H5vccStorage>(render_frame_host,
+                                                    std::move(receiver)) {
+  DETACH_FROM_THREAD(thread_checker_);
+}
+
+H5vccStorageImpl::~H5vccStorageImpl() {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+}
+
+void H5vccStorageImpl::Create(
+    content::RenderFrameHost* render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver) {
+  new H5vccStorageImpl(*render_frame_host, std::move(receiver));
+}
+
+void H5vccStorageImpl::Flush(FlushCallback flush_callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  auto* storage_partition = render_frame_host().GetStoragePartition();
+  CHECK(storage_partition);
+  // Flushes localStorage.
+  storage_partition->Flush();
+  auto* cookie_manager = storage_partition->GetCookieManagerForBrowserProcess();
+  CHECK(cookie_manager);
+  cookie_manager->FlushCookieStore(std::move(flush_callback));
+}
+
+}  // namespace h5vcc_storage
--- a/cobalt/browser/h5vcc_storage/h5vcc_storage_impl.h
+++ b/cobalt/browser/h5vcc_storage/h5vcc_storage_impl.h
@@ -1,0 +1,54 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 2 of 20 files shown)


cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -167,26 +199,21 @@ void H5vccSystemImpl::GetUserOnExitStrategy(
...
+void H5vccSystemImpl::Exit(bool force_close, ExitCallback exit_callback) {
+  auto* storage_partition = render_frame_host().GetStoragePartition();
+  CHECK(storage_partition);
+  // Flushes localStorage.
+  storage_partition->Flush();
+  auto* cookie_manager = storage_partition->GetCookieManagerForBrowserProcess();
+  CHECK(cookie_manager);
+  cookie_manager->FlushCookieStore(base::BindOnce(
+      [](bool force_close, base::OnceClosure exit_callback) {
+        PerformExitStrategy(force_close);
```
johnxwork: This modifies the current behavior of suspending the app on `h5vcc.system.exit()` to killing the app?

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.h
```diff
@@ -0,0 +1,54 @@
...
+  H5vccStorageImpl& operator=(const H5vccStorageImpl&) = delete;
+
+  void Flush(FlushCallback) override;
+
+ private:
+  H5vccStorageImpl(content::RenderFrameHost& render_frame_host,
+                   mojo::PendingReceiver<mojom::H5vccStorage> receiver);
+  ~H5vccStorageImpl();
+
+  THREAD_CHECKER(thread_checker_);
```
yell0wd0g, line 49: nit: This will be compiled out in non-debug builds so we're trying
to just keep a refptr to the SequencedTaskRunner, and CHECK for
BelongsToCurrentThread() on each method (in this case only one).
QQ: is this object really constructed on a different TaskRunner?
(I say so because of the DETACH in ctor).

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void H5vccStorageImpl::Create(
+    content::RenderFrameHost* render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver) {
+  new H5vccStorageImpl(*render_frame_host, std::move(receiver));
+}
+
+void H5vccStorageImpl::Flush(FlushCallback flush_callback) {
+  auto* storage_partition = render_frame_host().GetStoragePartition();
```
yell0wd0g, line 43: nit: Add `DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);`?

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -94,6 +96,36 @@ std::string GetTrackingAuthorizationStatusShared() {
...
+void PerformExitStrategy(bool force_close) {
+#if BUILDFLAG(IS_STARBOARD)
+  if (force_close) {
+    SbSystemRequestStop(0);
+    return;
+  }
+  auto strategy = GetUserOnExitStrategyInternal();
+  switch (strategy) {
+    case h5vcc_system::mojom::UserOnExitStrategy::kClose:
+      SbSystemRequestStop(0);
```
yell0wd0g: nit: `SbSystemRequestStop(/*error_level=*/0);`
Same in l.102.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
ghost: It allows h5vcc.system.exit(true) to force stop the app.

h5vcc.system.exit() will perform the exit strategy.

ghost: Perhaps it's not needed. yts can perform device.stop() instead.

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.h
ghost: I'll need to figure out what that means 😄 

I copied from the other h5vcc files. We can change all of them in a followup.

ghost:
Breaking this up into multiple PRs makes sense. This will just be adding `h5vcc.storage.flush()`.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
ghost: Moved this out of the PR.

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
ghost: Done

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
ghost: Reconsidering adding this. h5vcc.storage.flush() is still needed for now as well as flushing on blur/conceal/freeze.

third_party/blink/renderer/modules/cobalt/h5vcc_storage/h_5_vcc_storage.idl
```diff
@@ -0,0 +1,25 @@
...
+// This interface mirrors a subset of the H5vccStorage Web IDL defined in Cobalt
+// LTS 25:
+//   https://github.com/youtube/cobalt/blob/25.lts.stable/cobalt/h5vcc/h5vcc_storage.idl
+[
+  Exposed=Window,
+  SecureContext
+]
+interface H5vccStorage : EventTarget {
+  [CallWith=ScriptState, RaisesException]
+  Promise<void> flush();
```
hlwarriner, line 24: Are the web app teams aware of this decision to bring `flush()` forward from c25?

I noticed this API is changed from sync to async, which we do prefer in Chrobalt. But since the return type is changed from `void` to `Promise<void>` I believe the calling mechanics would need to change *if* the web app needs to ensure the flush is complete before proceeding with some other actions. It would be good to check with the author of go/chrobalt-web-apis-compatibility to see any compatibility changes are required.

ghost: Thanks. This PR is getting closed. 😄

ghost:
Closing since if we use indexeddb instead, we can get the same assurances as updating localStorage/cookies and calling a custom flush().

PR #7442, opened 2025-10-02
starboard: Introduce Result<T> for robust error handling

kjyoun:
This commit introduces a modern, robust error-handling mechanism using two new headers: `expected.h` and `result.h`.

`expected.h` provides a lightweight `Expected<T, E>` class, modeled after C++23's `std::expected`. It serves as a vocabulary type for functions that can return either a success value or an error, replacing out-of-band error codes. This implementation is a placeholder until `std::expected` or `base::expected` is available.

`result.h` builds upon `Expected` to provide helpers tailored for this codebase:
- `Result<T>`: A type alias for `Expected<T, std::string>`.
- `NonNullResult<T>`: A wrapper that guarantees a successful result for pointer types is never null.
- `Success()` and `Failure()`: Factory functions for creating result objects, improving readability.

As part of this change, several components have been refactored to use the new `Result` types, making their error handling more explicit and type-safe:
- `PlayerComponents::Factory::CreateComponents`
- `MediaCodecVideoDecoder::InitializeCodec`
- `FilterBasedPlayerWorkerHandler` (formerly `HandlerResult`)

This removes the need for output parameters for error messages and clarifies the success and failure paths in the player creation logic.

[go/cobalt-introduce-result-type](http://go/cobalt-introduce-result-type)

Bug: 449188714

PR #7530, opened 2025-10-13
android: Implement ExoPlayer native classes

osagie98:
Introduce a new media player implementation for Android platforms
leveraging ExoPlayer. This change adds C++ bridge classes to
interface with Java ExoPlayer components and integrates them into the
Starboard player worker architecture.

This enables using ExoPlayer as an alternative backend for audio and
video stream playback, aiming to improve media capabilities and
performance on Android. The integration is controlled by the new
feature flag, kEnableExoPlayer.

Bug: 191693650


starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+using base::android::ToJavaByteArray;
+using starboard::EndedCB;
+using starboard::ErrorCB;
+using starboard::InputBuffer;
+using starboard::InputBuffers;
+using starboard::PrerolledCB;
+
+constexpr int kADTSHeaderSize = 7;
+constexpr int kNoOffset = 0;
+constexpr int kWaitForInitializedTimeout = 250'000;  // 250 ms
```
kjyoun: Can you add "Us" suffix to denote the unit is microseconds

`kWaitForInitializedTimeoutUs`

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+
+DECLARE_INSTANCE_COUNTER(ExoPlayerBridge)
+
+ExoPlayerAudioCodec SbAudioCodecToExoPlayerAudioCodec(SbMediaAudioCodec codec,
```
kjyoun: Can you make this method to return `std::optional<ExoPlayerAudioCodec>`?

Then we can return either value or error (std::nullopt) in the same channel (by return value)

starboard/android/shared/exoplayer/exoplayer_bridge.h
```diff
@@ -0,0 +1,145 @@
...
+
+  base::android::ScopedJavaGlobalRef<jobject> j_exoplayer_manager_;
+  base::android::ScopedJavaGlobalRef<jobject> j_exoplayer_bridge_;
+  base::android::ScopedJavaGlobalRef<jobject> j_audio_media_source_;
+  base::android::ScopedJavaGlobalRef<jobject> j_video_media_source_;
+  base::android::ScopedJavaGlobalRef<jobject> j_sample_data_;
+  base::android::ScopedJavaGlobalRef<jobject> j_output_surface_;
+
+  bool error_occurred_ = false;
+  starboard::AudioStreamInfo audio_stream_info_;
```
kjyoun: const  starboard::AudioStreamInfo audio_stream_info_;
const  starboard::VideoStreamInfo video_stream_info_;

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+  }
+}
+
+}  // namespace
+
+ExoPlayerBridge::ExoPlayerBridge(const AudioStreamInfo& audio_stream_info,
+                                 const VideoStreamInfo& video_stream_info)
+    : audio_stream_info_(audio_stream_info),
+      video_stream_info_(video_stream_info) {
+  int64_t start = CurrentMonotonicTime();
```
kjyoun:
```
  int64_t start_us = CurrentMonotonicTime();
  InitExoplayer();
  int64_t end_us = CurrentMonotonicTime();
  SB_LOG(INFO) << "ExoPlayer init took " << (end_us - start_us) / 1'000 << " msec.";
```

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+                                   const PrerolledCB& prerolled_cb,
+                                   const EndedCB& ended_cb) {
+  SB_CHECK(error_cb);
+  SB_CHECK(prerolled_cb);
+  SB_CHECK(ended_cb);
+  SB_CHECK(!error_cb_);
+  SB_CHECK(!prerolled_cb_);
+  SB_CHECK(!ended_cb_);
+
+  error_cb_ = error_cb;
```
kjyoun: actually I think here std::move is more efficient. we can avoid copy


error_cb_ = std::move(error_cb);

Then you have update signature of this method.

void ExoPlayerBridge::SetCallbacks(ErrorCB error_cb,
                                   PrerolledCB prerolled_cb,
                                   EndedCB ended_cb)

starboard/android/shared/exoplayer/exoplayer_bridge.h
```diff
@@ -0,0 +1,145 @@
...
+  void OnReady(JNIEnv*);
+  void OnError(JNIEnv* env, jstring error_message);
+  void OnBuffering(JNIEnv*);
+  void SetPlayingStatus(JNIEnv*, jboolean isPlaying);
+  void OnPlaybackEnded(JNIEnv*);
+
+  // VideoSurfaceHolder method
+  void OnSurfaceDestroyed() override {}
+
+  bool IsEndOfStreamWritten(SbMediaType type) {
```
kjyoun: `bool IsEndOfStreamWritten(SbMediaType type)  const`

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+  playback_ended_ = false;
+  audio_eos_written_ = false;
+  video_eos_written_ = false;
+  seeking_ = true;
+  underflow_ = false;
+}
+
+void ExoPlayerBridge::WriteSamples(const InputBuffers& input_buffers) {
+  SB_CHECK(!input_buffers.empty());
+  SB_CHECK(input_buffers.size() == 1)
```
kjyoun: please use SB_CHECK_EQ, which logs value, when CHECK failed

include "starboard/common/check_op.h"

```
  SB_CHECK_EQ(input_buffers.size(), 1)
      << "ExoPlayer can only write 1 sample at a time"'
```

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+  seek_time_ = seek_to_timestamp;
+  playback_ended_ = false;
+  audio_eos_written_ = false;
+  video_eos_written_ = false;
+  seeking_ = true;
+  underflow_ = false;
+}
+
+void ExoPlayerBridge::WriteSamples(const InputBuffers& input_buffers) {
+  SB_CHECK(!input_buffers.empty());
```
kjyoun: you don't need empty check. since you checks length of input_buffers below

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+  SB_CHECK(input_buffers.size() == 1)
+      << "ExoPlayer can only write 1 sample at a time, received "
+      << input_buffers.size() << " samples.";
+
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
```
kjyoun, line 180: this would be enough. you don't need more expensive unique_lock here
````
std::lock_guard lock(mutex_);
```

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+    is_key_frame = input_buffers.front()->video_sample_info().is_key_frame;
+  }
+
+  bool is_eos = false;
+
+  JNIEnv* env = AttachCurrentThread();
+  // Remove the ADTS header from AAC frames before writing to the player.
+  int offset = (type == kSbMediaTypeAudio &&
+                        audio_stream_info_.codec == kSbMediaAudioCodecAac
+                    ? kADTSHeaderSize
```
kjyoun: outer-most brackets are not necessary

````
 int offset = type == kSbMediaTypeAudio &&
                        audio_stream_info_.codec == kSbMediaAudioCodecAac
                         ? kADTSHeaderSize : kNoOffset;
```

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+                                       j_exoplayer_bridge_,
+                                       static_cast<float>(playback_rate));
+
+  std::unique_lock<std::mutex> lock(mutex_);
+  playback_rate_ = playback_rate;
+}
+
+int ExoPlayerBridge::GetDroppedFrames() {
+  jint dropped_frames = Java_ExoPlayerBridge_getDroppedFrames(
+      AttachCurrentThread(), j_exoplayer_bridge_);
```
kjyoun: ?

``` 
return  Java_ExoPlayerBridge_getDroppedFrames(
      AttachCurrentThread(), j_exoplayer_bridge_);
```

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+  return dropped_frames;
+}
+
+int64_t ExoPlayerBridge::GetCurrentMediaTime(MediaInfo& info) {
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
```
kjyoun: std::lock_guard lock(mutex_);

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+}
+
+bool ExoPlayerBridge::EnsurePlayerIsInitialized() {
+  bool wait_timeout;
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
+    wait_timeout =
+        initialized_cv_.wait_for(
+            lock, std::chrono::microseconds(kWaitForInitializedTimeout)) ==
+        std::cv_status::timeout;
```
kjyoun: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rconc-wait

Add condition variable and check it

```
void ExoPlayerBridge::OnInitialized(JNIEnv* env) {
 { 
   std::lock_guard lock(mutex_);
   intialized_ = true;
 }
 initialized_cv_.notify_one();
}


bool ExoPlayerBridge::EnsurePlayerIsInitialized() {
  bool wait_timeout;
  {
    std::unique_lock<std::mutex> lock(mutex_);
    initialized_cv_.wait_for(
            lock, std::chrono::microseconds(kWaitForInitializedTimeout). [this] {
                  return initialized_;
            });
```

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,248 @@
...
+}
+
+HandlerResult ExoPlayerWorkerHandler::Init(
+    SbPlayer player,
+    UpdateMediaInfoCB update_media_info_cb,
+    GetPlayerStateCB get_player_state_cb,
+    UpdatePlayerStateCB update_player_state_cb,
+    UpdatePlayerErrorCB update_player_error_cb) {
+  // This function should only be called once.
+  SB_CHECK(update_media_info_cb_ == NULL);
```
kjyoun: idiom for null check

```
SB_CHECK(!update_media_info_cb_ );
```

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,248 @@
...
+  // later report an error if seek fails.
+  return HandlerResult{true};
+}
+
+HandlerResult ExoPlayerWorkerHandler::WriteSamples(
+    const InputBuffers& input_buffers,
+    int* samples_written) {
+  SB_CHECK(!input_buffers.empty());
+  SB_CHECK(BelongsToCurrentThread());
+  SB_CHECK(samples_written != NULL);
```
kjyoun: SB_CHECK(!samples_written);

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,248 @@
...
+// limitations under the License.
+
+#include "starboard/android/shared/exoplayer/exoplayer_worker_handler.h"
+
+#include "starboard/common/player.h"
+
+namespace starboard {
+namespace {
+
+using ::starboard::GetPlayerStateName;
```
kjyoun: you don't need `using ::starboard::GetPlayerStateName;`

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,248 @@
...
+
+const int64_t kUpdateIntervalUsec = 200'000;  // 200ms
+}  // namespace
+
+ExoPlayerWorkerHandler::ExoPlayerWorkerHandler(
+    const SbPlayerCreationParam* creation_param)
+    : JobOwner(kDetached),
+      audio_stream_info_(creation_param->audio_stream_info),
+      video_stream_info_(creation_param->video_stream_info) {
+  update_job_ = std::bind(&ExoPlayerWorkerHandler::Update, this);
```
kjyoun: can you initialize member variable though initialization list and make the const?

````
  audio_stream_info_(creation_param->audio_stream_info),
      video_stream_info_(creation_param->video_stream_info),
     update_job_(std::bind(&ExoPlayerWorkerHandler::Update, this),


const .. update_job_
const std::unique_ptr<ExoPlayerBridge> bridge_;
```

starboard/android/shared/exoplayer/exoplayer_worker_handler.h
```diff
@@ -0,0 +1,93 @@
...
+#include "starboard/shared/starboard/media/media_util.h"
+#include "starboard/shared/starboard/player/filter/common.h"
+#include "starboard/shared/starboard/player/input_buffer_internal.h"
+#include "starboard/shared/starboard/player/job_queue.h"
+#include "starboard/shared/starboard/player/player_worker.h"
+
+namespace starboard {
+
+class ExoPlayerWorkerHandler : public starboard::PlayerWorker::Handler,
+                               private starboard::JobQueue::JobOwner {
```
kjyoun: you can omit starboard:: here, which make code more concise

```
public PlayerWorker::Handler,
private JobQueue::JobOwner 
```

starboard/android/shared/exoplayer/exoplayer_worker_handler.h
```diff
@@ -0,0 +1,93 @@
...
+ private:
+  starboard::PlayerWorker::Handler::HandlerResult Init(
+      SbPlayer player,
+      UpdateMediaInfoCB update_media_info_cb,
+      GetPlayerStateCB get_player_state_cb,
+      UpdatePlayerStateCB update_player_state_cb,
+      UpdatePlayerErrorCB update_player_error_cb) override;
+  starboard::PlayerWorker::Handler::HandlerResult Seek(int64_t seek_to_time,
+                                                       int ticket) override;
+  starboard::PlayerWorker::Handler::HandlerResult WriteSamples(
```
kjyoun: You can use just HandlerResult here and below

```
HandlerResult WriteSamples(...);
````

as we already did in starboard/shared/starboard/player/filter/filter_based_player_worker_handler.h

File: starboard/android/shared/exoplayer/exoplayer_worker_handler.h [hunk 20]
```diff
@@ -0,0 +1,93 @@
...
+  void Update();
+  void OnError(SbPlayerError error, const std::string& error_message);
+  void OnPrerolled();
+  void OnEnded();
+
+  SbDecodeTarget GetCurrentDecodeTarget() override {
+    return kSbDecodeTargetInvalid;
+  }
+
+  std::unique_ptr<ExoPlayerBridge> bridge_;
```
kjyoun: const ?

starboard/android/shared/exoplayer/exoplayer_worker_handler.h
osagie98: Done in the header file, for some reason it doesn't work in the cc file, though I've shortened it

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,424 @@
...
+    case kSbMediaAudioCodecFlac:
+      return EXOPLAYER_AUDIO_CODEC_FLAC;
+    case kSbMediaAudioCodecPcm:
+      return EXOPLAYER_AUDIO_CODEC_PCM;
+    case kSbMediaAudioCodecIamf:
+      return EXOPLAYER_AUDIO_CODEC_IAMF;
+    default:
+      SB_LOG(ERROR) << "ExoPlayerBridge encountered unknown audio codec "
+                    << codec;
+      return EXOPLAYER_AUDIO_CODEC_MAX;
```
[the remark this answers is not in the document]
osagie98: Removed

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,249 @@
...
+  } else {
+    if (bridge_->IsEndOfStreamWritten(kSbMediaTypeVideo)) {
+      SB_LOG(WARNING) << "Tried to write video EOS after EOS is enqueued";
+    } else {
+      SB_LOG(INFO) << "Video EOS enqueued";
+      bridge_->WriteEndOfStream(kSbMediaTypeVideo);
+    }
+  }
+
+  return PlayerWorker::Handler::HandlerResult{true};
```
kjyoun:
```
return Success();
```

you may meet build error, since HandleResult is updated in #7442 to `Result<void>`

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,249 @@
...
+    int* samples_written) {
+  SB_CHECK(!input_buffers.empty());
+  SB_CHECK(BelongsToCurrentThread());
+  SB_CHECK(samples_written);
+  SB_CHECK(bridge_);
+  for (const auto& input_buffer : input_buffers) {
+    SB_CHECK(input_buffer);
+  }
+
+  *samples_written = 0;
```
kjyoun, line 107: I don't think  you don't need this line, once you change line 114 to `samples_written = input_buffers.size();

starboard/android/shared/exoplayer/exoplayer_worker_handler.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace starboard {
+
+class ExoPlayerWorkerHandler : public PlayerWorker::Handler,
+                               private JobQueue::JobOwner {
+ public:
+  explicit ExoPlayerWorkerHandler(const SbPlayerCreationParam* creation_param);
+
+ private:
+  HandlerResult Init(SbPlayer player,
```
kjyoun: Can you replace `HandlerResult` with `Result<void>` ?

HandlerResult will be gone in #7560

with `Result<void>`, this PR would not be affected

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,249 @@
...
+    }
+  } else {
+    SB_CHECK_EQ(input_buffers.front()->sample_type(), kSbMediaTypeVideo);
+    if (bridge_->IsEndOfStreamWritten(kSbMediaTypeVideo)) {
+      SB_LOG(WARNING) << "Tried to write video sample after EOS is reached";
+    }
+  }
+
+  *samples_written += input_buffers.size();
+  bridge_->WriteSamples(input_buffers);
```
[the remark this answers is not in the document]
osagie98: Enabled batch writes

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,559 @@
...
+using base::android::ScopedJavaGlobalRef;
+using base::android::ScopedJavaLocalRef;
+using base::android::ToJavaByteArray;
+
+constexpr int kADTSHeaderSize = 7;
+constexpr int kNoOffset = 0;
+constexpr int kWaitForInitializedTimeoutUs = 250'000;  // 250 ms.
+const SbMediaMasteringMetadata kEmptyMasteringMetadata = {};
+
+const jint COLOR_RANGE_FULL = 1;
```
kjyoun: constexpr  here and below?

starboard/android/shared/exoplayer/exoplayer_bridge.h
```diff
@@ -0,0 +1,146 @@
...
+  mutable std::mutex mutex_;
+  // Signaled once player initialization is complete.
+  std::condition_variable initialized_cv_;
+  bool initialized_ = false;
+  bool audio_eos_written_ = false;
+  bool video_eos_written_ = false;
+  bool playback_ended_ = false;
+  double playback_rate_ = 1.0;
+  bool seeking_ = false;
+  bool underflow_ = false;
```
kjyoun: can you add comments to member variables that are guarded by mutex_ ?

e.g.)
double playback_rate_ = 1.0;  // Guarded by |mutex_|

starboard/android/shared/BUILD.gn
```diff
@@ -187,6 +187,10 @@ static_library("starboard_platform") {
     "drm_system.cc",
     "drm_system.h",
     "egl_swap_buffers.cc",
+    "exoplayer/exoplayer_bridge.cc",
```
kjyoun, line 190: can you add BUILD.gn to starboard/android/shared/exoplayer and set build configuration there?

Instead of importing source codes, we can put the build target under deps.

Then we isolate build configuration for Exoplayer cleanly.

e.g.)


[
starboard/android/shared/BUILD.gn](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/shared/starboard/media/BUILD.gn)

starboard/android/shared/exoplayer/exoplayer_bridge.h
```diff
@@ -0,0 +1,146 @@
...
+  void InitExoplayer();
+
+  base::android::ScopedJavaGlobalRef<jobject> j_exoplayer_manager_;
+  base::android::ScopedJavaGlobalRef<jobject> j_exoplayer_bridge_;
+  base::android::ScopedJavaGlobalRef<jobject> j_audio_media_source_;
+  base::android::ScopedJavaGlobalRef<jobject> j_video_media_source_;
+  base::android::ScopedJavaGlobalRef<jobject> j_output_surface_;
+
+  bool error_occurred_ = false;
+  const AudioStreamInfo audio_stream_info_;
```
kjyoun: add mutex comments?

looks like audio_stream_info_ and video_stream_info_ are guarded by mutex

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+  // This function should only be called once.
+  SB_CHECK(!update_media_info_cb_);
+
+  // All parameters have to be valid.
+  SB_CHECK(SbPlayerIsValid(player));
+  SB_CHECK(update_media_info_cb);
+  SB_CHECK(get_player_state_cb);
+  SB_CHECK(update_player_state_cb);
+
+  AttachToCurrentThread();
```
kjyoun: why this method is called, though we don't use env?

kjyoun (commented):
LGTM  % comments.

Can you add separate BUILD.gn for files under exoplayer?

and please add some explanation about this PR in description. you can refer the commit message that Gemini generates

starboard/extension/feature_config.h
```diff
@@ -133,6 +133,8 @@ STARBOARD_FEATURE(kUseStubVideoDecoder, "UseStubVideoDecoder", false)
 STARBOARD_FEATURE(kVideoDecoderDelayUsecOverride,
                   "VideoDecoderDelayUsecOverride",
                   false)
+// Enables the ExoPlayer-based SbPlayer for playback.
+STARBOARD_FEATURE(kEnableExoPlayer, "EnableExoPlayer", false)
```
jasonzhangxx, line 146: nit: the starboard feature is actually not used in this pr.

starboard/android/shared/exoplayer/exoplayer_bridge.h
osagie98: I made `error_occurred_` an atomic bool, though the rest shouldn't need to be protected. Possibly j_output_surface but it's only used by Java until destruction.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
osagie98: This is calling `JobQueue::JobOwner::AttachToCurrentThread()`.

starboard/android/shared/BUILD.gn
osagie98: I explored this, because I call a method from `StarboardBridge` in `starboard/android/shared`, the exoplayer module needs to be exposed to it, causing a dependency cycle. I can do this if I also isolate the `StarboardBridge` files.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  initialized_cv_.notify_one();
+}
+
+void ExoPlayerBridge::OnReady(JNIEnv* env) {
+  prerolled_cb_();
+}
+
+void ExoPlayerBridge::OnError(JNIEnv* env, jstring error_message) {
+  SB_CHECK(error_cb_);
+  error_cb_(kSbPlayerErrorDecode, ConvertJavaStringToUTF8(env, error_message));
```
jasonzhangxx: `error_occurred_` is not set to `true` here. Please check if it's intended. I recommend to add a error reporting  function and always use that function to report error.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+
+HandlerResult ExoPlayerWorkerHandler::WriteSamples(
+    const InputBuffers& input_buffers,
+    int* samples_written) {
+  SB_CHECK(!input_buffers.empty());
+  SB_CHECK(BelongsToCurrentThread());
+  SB_CHECK(samples_written);
+  SB_CHECK(bridge_);
+  for (const auto& input_buffer : input_buffers) {
+    SB_CHECK(input_buffer);
```
jasonzhangxx: `SB_CHECK` will evaluate the condition in gold build (DCHECK does not). Let's avoid none trivial SB_CHECKs or use SB_DCHECK here.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  *samples_written = 0;
+  if (input_buffers.front()->sample_type() == kSbMediaTypeAudio) {
+    if (bridge_->IsEndOfStreamWritten(kSbMediaTypeAudio)) {
+      SB_LOG(WARNING) << "Tried to write audio sample after EOS is reached";
```
jasonzhangxx: In `FilterBasedPlayerWorkerHandler::WriteSamples`, if EOS is already written, besides the warning log, the hanlder won't write the sample to the underlying renderer. But the code here will always call `bridge_->WriteSamples(input_buffers)`.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+  bridge_->SetVolume(volume_);
+}
+
+void ExoPlayerWorkerHandler::Update() {
+  SB_CHECK(BelongsToCurrentThread());
+
+  if (get_player_state_cb_() == kSbPlayerStatePresenting) {
+    int dropped_frames = 0;
+    if (video_stream_info_.codec != kSbMediaVideoCodecNone) {
+      dropped_frames = bridge_->GetDroppedFrames();
```
jasonzhangxx: If no video codec, can `bridge_->GetDroppedFrames()` works? If it works, we can have code like
```
if(bridge_.is_valid()) {
  dropped_frames = bridge_->GetDroppedFrames();
}
```

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  AttachToCurrentThread();
+
+  player_ = player;
+  update_media_info_cb_ = update_media_info_cb;
+  get_player_state_cb_ = get_player_state_cb;
+  update_player_state_cb_ = update_player_state_cb;
+  update_player_error_cb_ = update_player_error_cb;
+
+  bridge_->SetCallbacks(
```
jasonzhangxx: `ExoPlayerBridge::InitExoplayer` is called in ExoPlayerBridge ctor, which is in ExoPlayerWorkerHandler ctor. But it's possible that error occurs during initialization. In that case, we'd better to be able to report the error. I recommend to pass the callbacks below in ExoPlayerBridge ctor, and in ExoPlayerWorkerHandler::OnError, we do
```
if(update_player_error_cb_) {
  // use |update_player_error_cb_| to report error.
}
else {
  // store the error and message, and report them after  |update_player_error_cb_| is set.
}
```

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+}
+
+void ExoPlayerWorkerHandler::Stop() {
+  SB_CHECK(BelongsToCurrentThread());
+
+  SB_LOG(INFO) << "ExoPlayerWorkerHandler stopped.";
+
+  RemoveJobByToken(update_job_token_);
+
+  bridge_->Stop();
```
jasonzhangxx: Let's clear callbacks `update_player_state_cb_`, `update_player_state_cb_` and `update_player_state_cb_` to ensure they won't be triggered after `Stop()`. Also, related code in `OnError`, `OnPrerolled` and `OnEnded` need to be updated.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+    int bits_per_sample = audio_stream_info_.bits_per_sample;
+
+    ScopedJavaLocalRef<jbyteArray> configuration_data;
+    if (!audio_stream_info_.audio_specific_config.empty()) {
+      configuration_data =
+          ToJavaByteArray(env, audio_stream_info_.audio_specific_config.data(),
+                          audio_stream_info_.audio_specific_config.size());
+    }
+
+    bool is_passthrough = audio_stream_info_.codec == kSbMediaAudioCodecEac3 ||
```
jasonzhangxx: `is_passthrough` will be override in `SupportedAudioCodecToMimeType()`, so the init value here doesn't make sense.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+
+  if (get_player_state_cb_() == kSbPlayerStatePresenting) {
+    int dropped_frames = 0;
+    if (video_stream_info_.codec != kSbMediaVideoCodecNone) {
+      dropped_frames = bridge_->GetDroppedFrames();
+    }
+
+    ExoPlayerBridge::MediaInfo info;
+    auto media_time = bridge_->GetCurrentMediaTime(info);
+    update_media_info_cb_(media_time, dropped_frames, !info.is_underflow);
```
jasonzhangxx: Consider to add `bridge_->GetCurrentMediaTime()` and `bridge_->IsUnderflow()` instead of `auto media_time = bridge_->GetCurrentMediaTime(info);`. `is_playing`, `is_eos_played` and `playback_rate` are actually not used.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  }
+}
+
+}  // namespace
+
+ExoPlayerBridge::ExoPlayerBridge(const AudioStreamInfo& audio_stream_info,
+                                 const VideoStreamInfo& video_stream_info)
+    : audio_stream_info_(audio_stream_info),
+      video_stream_info_(video_stream_info) {
+  int64_t start_us = CurrentMonotonicTime();
```
jasonzhangxx: nit: don't forget to remove debug code.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  SB_CHECK(error_cb_);
+  error_cb_(kSbPlayerErrorDecode, ConvertJavaStringToUTF8(env, error_message));
+}
+
+void ExoPlayerBridge::OnBuffering(JNIEnv* env) {
+  std::lock_guard lock(mutex_);
+  underflow_ = true;
+}
+
+void ExoPlayerBridge::SetPlayingStatus(JNIEnv* env, jboolean is_playing) {
```
jasonzhangxx: The `SetPlayingStatus` is more like a callback of `OnPlayeringStatusChanged`.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  ScopedJavaLocalRef<jbooleanArray> j_key_frames(
+      env, env->NewBooleanArray(number_of_samples));
+  env->SetBooleanArrayRegion(j_key_frames.obj(), 0, number_of_samples,
+                             key_frames.data());
+
+  auto j_media_source =
+      type == kSbMediaTypeAudio ? j_audio_media_source_ : j_video_media_source_;
+  ScopedJavaLocalRef<jobject> j_combined_samples =
+      Java_ExoPlayerMediaSource_getSampleBuffer(env, j_media_source,
+                                                total_size);
```
jasonzhangxx: It's good to support batched writing, but we need to make sure it's supported. The size of combined sample can't be too large, otherwise it might have problems like 1. memory allocation failure, 2. performance concern, 3. blocking the thread for too long time.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  for (size_t i = 0; i < number_of_samples; ++i) {
+    auto& input_buffer = input_buffers[i];
+    const uint8_t* data_start = input_buffer->data() + offset;
+    int sample_size = sizes[i];
+
+    memcpy(current_write_ptr, data_start, sample_size);
+    current_write_ptr += sample_size;
+  }
+
+  Java_ExoPlayerMediaSource_writeSamples(
```
jasonzhangxx: Can we consider to pass media type (audio or video) down to Java_ExoPlayerMediaSource_writeSamples() and let ExoPlayerBridge java class to handle the inputs? So that, the ExoPlayerBridge C++ interface can be simplified and it doesn't have to hold pointers to audio source and video source.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  if (error_occurred_.load()) {
+    SB_LOG(ERROR) << "Tried to seek after an error occurred.";
+    return;
+  }
+
+  Java_ExoPlayerBridge_seek(AttachCurrentThread(), j_exoplayer_bridge_,
+                            seek_to_timestamp);
+
+  std::lock_guard lock(mutex_);
+  ended_ = false;
```
jasonzhangxx: `ended_` is only set to true in dtor. Do we need it?

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,560 @@
...
+  video_eos_written_ = false;
+  seeking_ = true;
+  underflow_ = false;
+}
+
+void ExoPlayerBridge::WriteSamples(const InputBuffers& input_buffers) {
+  {
+    std::lock_guard lock(mutex_);
+
+    if (ended_) {
```
jasonzhangxx: It doesn't make sense to check `ended_` here.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,250 @@
...
+  } else {
+    SB_CHECK_EQ(input_buffers.front()->sample_type(), kSbMediaTypeVideo);
+    if (bridge_->IsEndOfStreamWritten(kSbMediaTypeVideo)) {
+      SB_LOG(WARNING) << "Tried to write video sample after EOS is reached";
+    }
+  }
+
+  *samples_written += input_buffers.size();
+  bridge_->WriteSamples(input_buffers);
+  return Success();
```
jasonzhangxx, line 122: Success sample writing would make PlayerWorker keep pulling data. `FilterBasedPlayerWorkerHandler::WriteSamples` checks if underlying components can accept more inputs. For exoplayer, I think we need a mechanism to limit inputs.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
```diff
@@ -0,0 +1,343 @@
...
+      reinterpret_cast<const jbyte*>(input_buffer->data() + offset));
+  ScopedJavaLocalRef<jbyteArray> sample_data_local_ref(
+      env, static_cast<jbyteArray>(env->NewLocalRef(j_sample_data_.obj())));
+
+  Java_ExoPlayerBridge_writeSample(
+      env, j_exoplayer_bridge_, sample_data_local_ref, size,
+      input_buffer->timestamp(), is_key_frame, type);
+
+  return true;
+}
```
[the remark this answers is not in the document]
osagie98: I can do this, but we only expect a single sample at a time, and there's a DCHECK in any case. This would lessen the changed needed for batched writes in the future.

starboard/android/shared/exoplayer/exoplayer_bridge.h
```diff
@@ -0,0 +1,106 @@
...
+                  const SbMediaVideoStreamInfo& video_stream_info);
+  ~ExoPlayerBridge();
+
+  // VideoSurfaceHolder method.
+  void OnSurfaceDestroyed() override;
+
+  bool Init(ErrorCB error_cb, PrerolledCB prerolled_cb, EndedCB ended_cb);
+
+  bool Seek(int64_t timestamp);
+  bool WriteSamples(const InputBuffers& input_buffers, SbMediaType type);
```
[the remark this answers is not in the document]
osagie98: See above

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,243 @@
...
+      return Success();
+    }
+    if (!bridge_->WriteSamples(input_buffers, sample_type)) {
+      return Failure("Failed to write samples to the ExoPlayer.");
+    }
+    *samples_written = static_cast<int>(input_buffers.size());
+  }
+
+  return Success();
+}
```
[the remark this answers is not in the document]
osagie98: It should only contain one sample, which is DCHECKed in the bridge. I could DCHECK again here but it would be redundant. Also, when we implement batched writes this function shouldn't need to be updated.

starboard/extension/feature_config.h
osagie98: Added the remaining c++ files, it's used now in player_create.cc

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
osagie98: Changed to DCHECK. I changed a few other CHECKs in functions called often to DCHECK as well.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
osagie98: It'll work with no codec, it'll just return 0 as the underlying player will never report dropped frames. In any case, I removed the separate function for it.

starboard/android/shared/exoplayer/exoplayer_bridge.cc
osagie98: Removed batched writing

starboard/android/shared/exoplayer/exoplayer_bridge.cc
osagie98: I moved it so the Java ExoPlayerBridge determines the correct MediaSource to write to

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
osagie98: Added a method `CanAcceptMoreData`. It's not reflected here in C++, but in the Java code I set the max buffer duration to 30 seconds, and it'll return false for `CanAcceptMoreData` at 27 seconds or higher. These are arbitrarily chosen and I can change them to values that make some more sense, but I'm still working out that side.

osagie98:
Note that I removed batched writing support to keep it simple. The earlier implementation works but it requires more consideration for memory. However, I left `SbPlayerGetMaximumNumberOfSamplesPerWrite` untouched at 256, which is technically not possible with the current implementation.

starboard/android/shared/exoplayer/exoplayer_worker_handler.cc
```diff
@@ -0,0 +1,243 @@
...
+
+  update_job_token_ = Schedule(update_job_, kUpdateIntervalUsec);
+
+  return Success();
+}
+
+Result<void> ExoPlayerWorkerHandler::Seek(int64_t seek_to_time, int ticket) {
+  SB_CHECK(bridge_->is_valid());
+
+  audio_eos_written_ = false;
```
osagie98, line 86: Not sure if this is thread-safe, I can add a mutex here or make them atomic

starboard/android/shared/exoplayer/exoplayer_util.cc
```diff
@@ -0,0 +1,241 @@
...
+
+bool IsIdentity(const SbMediaColorMetadata& color_metadata) {
+  return color_metadata.primaries == kSbMediaPrimaryIdBt709 &&
+         color_metadata.transfer == kSbMediaTransferIdBt709 &&
+         color_metadata.matrix == kSbMediaMatrixIdBt709 &&
+         color_metadata.range == kSbMediaRangeIdLimited &&
+         Equal(color_metadata.mastering_metadata, kEmptyMasteringMetadata);
+}
+
+jint SbMediaPrimaryIdToColorStandard(SbMediaPrimaryId primary_id) {
```
jasonzhangxx, line 70: We can move all duplicate functions and variables to `media_common.h`.

starboard/android/shared/exoplayer/exoplayer_util.cc
```diff
@@ -0,0 +1,241 @@
...
+// metadata value and Android HDR metadata value is possible.  This value
+// implies that HDR playback should not be attempted.
+constexpr jint COLOR_VALUE_UNKNOWN = -1;
+
+bool Equal(const SbMediaMasteringMetadata& lhs,
+           const SbMediaMasteringMetadata& rhs) {
+  return memcmp(&lhs, &rhs, sizeof(SbMediaMasteringMetadata)) == 0;
+}
+
+bool IsIdentity(const SbMediaColorMetadata& color_metadata) {
```
jasonzhangxx, line 62: Looks like this function is to check if the color metadata is sdr?

osagie98:
Closing in favor of https://github.com/youtube/cobalt/pull/8529

PR #7531, opened 2025-10-14
Reduce Skia's GPU resource cache size to lower memory usage

sherryzy:
This change reduces the Skia GrContext's max resource cache size to 0 on low-end devices.

Skia uses this cache to keep GPU resources like textures and shaders in memory for reuse, which improves rendering performance. However, this can consume a significant amount of memory.

Testing on Android shows that when navigating quickly through video content, this cache can grow to 20 MB or more. By disabling it, the memory usage for Skia's cache is reduced to a stable ~5 MB.

Manual testing indicates that this change does not introduce significant user experience regressions, and pages continue to render correctly. This is a worthwhile trade-off for the memory savings on resource-constrained devices.

Bug: 424591313


johnxwork:
Would you be able to count the rough cache hit rate of the to-be-disabled rendering resource cache?

PR #7532, opened 2025-10-14
android: Remove legacy JNI helper files

kjyoun:
This change removes several legacy JNI helper files that are no longer needed. The functionality provided by these files has been replaced by more modern JNI practices from `base/android/`.

The following files are removed:
- starboard/android/shared/jni_env_ext.h
- starboard/android/shared/jni_env_ext.cc
- starboard/android/shared/jni_env_ext_test.cc
- starboard/android/shared/jni_state.h
- starboard/android/shared/jni_state.cc

Along with the file deletions, build dependencies were updated, and the corresponding `JniEnvExtTest` cases were removed from the Android test filters.

Bug: 450024477

### DIFF
--- a/cobalt/build/android/package.json
+++ b/cobalt/build/android/package.json
@@ -9,7 +9,6 @@
                 "icudtl.dat",
                 "lib.unstripped/libchrobalt.so",
                 "libchrobalt.so",
-                "libstarboard_jni_state.so",
                 "obj/cobalt/shell/android/cobalt_shell_java_resources.resources.zip",
                 "obj/components/metrics/metrics_java.javac.jar",
                 "obj/third_party/android_deps/chromium_play_services_availability_java.javac.jar"


starboard/android/shared/jni_env_ext.cc
```diff
@@ -1,289 +0,0 @@
...
-  SB_DCHECK_NE(JNIState::GetVM(), nullptr);
-
-  SB_DCHECK_EQ(JNIState::GetApplicationClassLoader(), nullptr);
-  JNIState::SetApplicationClassLoader(JniConvertLocalRefToGlobalRef(
-      env, JniCallObjectMethodOrAbort(
-               env, env->GetObjectClass(starboard_bridge), "getClassLoader",
-               "()Ljava/lang/ClassLoader;")));
-
-  SB_DCHECK_EQ(JNIState::GetStarboardBridge(), nullptr);
-  JNIState::SetStarboardBridge(env->NewGlobalRef(starboard_bridge));
```
johnedocampo, line 60: Are you able to verify that the app runs and with no issues with this deletion? I remember encountering this line for https://github.com/youtube/cobalt/pull/6981, and if I tried to comment out this line my app wouldn't run.

johnedocampo (approved):
Thanks for doing this! I was getting confused on how these codes worked when I was interacting with them, but it looks like we do not need them anymore. Changes LGTM. Have one question about how deleting these files affects runtime.

starboard/android/shared/jni_env_ext.cc
kjyoun: Yes, I verified that app works fine.

'JniState` is to keep state for JniEnvExt. It does not call any Android methods under the hood. https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/jni_state.cc;l=46;drc=2ab9ec976c84bd4d76d09801fb022398f88782c2

Since now we don't use JniEnvEXt any more, we can safely delete JniState and its call sites.

If you rebase your change to ToT and the delete this method call, app still work properly.

johnedocampo: Thanks for the clarification!

johnxwork (approved):
I'm actually really surprised we could remove JNIEnvExt so easily. This is quite enjoyable!

kjyoun:
Now all JNI method that  used JniEnvExt is now migrated to JNI generator,  though we have several  legacy JNI code, which do not use JniEnvExt

PR #7536, opened 2025-10-14
Refactor ShellPlatformDelegate and implement TestShellPlatformDelegate

haozheng-cobalt:
This PR decouples the base class from concrete delegate implementations and allows for cleaner test-specific overrides by introducing TestShellPlatformDelegate.

Bug: 446894603

### DIFF
--- a/cobalt/shell/browser/shell_platform_delegate.h
+++ b/cobalt/shell/browser/shell_platform_delegate.h
@@ -25,6 +25,12 @@
 #include "ui/gfx/geometry/size.h"
 #include "ui/gfx/native_widget_types.h"
 
+#if defined(USE_AURA) && defined(SHELL_USE_TOOLKIT_VIEWS)
+namespace views {
+class ViewsDelegate;
+}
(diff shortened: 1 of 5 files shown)


cobalt/shell/browser/shell_platform_delegate.h
```diff
@@ -141,6 +147,10 @@ class ShellPlatformDelegate {
 #endif
 
  protected:
+#if defined(USE_AURA) && defined(SHELL_USE_TOOLKIT_VIEWS)
```
Awallky, line 150: nit: you could combine this with the preprocessor define and change the lower oner to `#elif defined(USE_AURA) && !defined(SHELL_USE_TOOLKIT_VIEWS)` as they depend on the same defines with differing values.

haozheng-cobalt: I'm using the AND condition to better match the corresponding sources and dependencies conditions in the gn file.

PR #7537, opened 2025-10-14
Add custom InsertAnnotationForHandler API from c25

hlwarriner:
This enables Crashpad annotations to be provided by a client after the signal handler has been installed. This is needed to enable annotations to be set by the Cobalt layer, including on behalf of a web app.

These changes are mostly just ported forward c25, with a couple significant updates:
* Use of locks from //base/synchronization instead of Starboard
* The scope is much smaller since we are not currently supporting the RequestCrashDumpHandler SignalHandler in Chrobalt.

A separate PR will wire up the CrashHandler Starboard extension so that this API is used by the setString() web API for 3P.

Issue: 446908034

### DIFF
--- a/third_party/crashpad/crashpad/client/crashpad_client_linux.cc
+++ b/third_party/crashpad/crashpad/client/crashpad_client_linux.cc
@@ -51,6 +51,7 @@
 
 #if BUILDFLAG(IS_STARBOARD)
 #include "base/notreached.h"
+#include "base/synchronization/lock.h"
 #include "starboard/elf_loader/evergreen_info.h"
 #endif  // BUILDFLAG(IS_STARBOARD)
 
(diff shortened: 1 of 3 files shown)


third_party/crashpad/crashpad/client/crashpad_client_linux.cc
```diff
@@ -206,11 +242,15 @@ class SignalHandler {
...
 
+#if BUILDFLAG(IS_STARBOARD)
+  const EvergreenInfo& GetEvergreenInfo() { return evergreen_info_; }
+#endif  // BUILDFLAG(IS_STARBOARD)
+
   virtual void HandleCrashImpl() = 0;
 
 #if BUILDFLAG(IS_STARBOARD)
   virtual bool SendEvergreenInfoImpl() = 0;
-  const EvergreenInfo& GetEvergreenInfo() { return evergreen_info_; }
```
andrewsavage1, line 213: Why move this? Wouldn't it be better to group our functions together?

hlwarriner: In the spirit of Classes/Declaration Order from Google's C++ style guide, we should group the virtual method declarations together. In my opinion we should prioritize that over grouping our customizations together.

PR #7540, opened 2025-10-14
Fix linker error for cobalt browser tests

linxinan-yt:
Bug: b/432313695
Ttest: autoninja -C out/android-x86_debug cobalt_browsertests


linxinan-yt:
Close.

PR #7546, opened 2025-10-15
Add StarboardMain entry point function

santoshmahto7:
When loader_app is run as wpeframework plugin, the plugin expect StarboardMain as entry point to libloader_app library

### DIFF
--- a/starboard/loader_app/loader_app.cc
+++ b/starboard/loader_app/loader_app.cc
@@ -309,6 +309,13 @@ void SbEventHandle(const SbEvent* event) {
   SB_CHECK_EQ(pthread_mutex_unlock(&mutex), 0);
 }
 
+#if SB_IS(EVERGREEN_COMPATIBLE)
+// StarboardMain is called from wpeframework cobalt plugin
+extern "C" SB_EXPORT_PLATFORM int StarboardMain(int argc, char** argv) {
+  return SbRunStarboardMain(argc, argv, SbEventHandle);
+}
+#endif
+
 int main(int argc, char** argv) {
   return SbRunStarboardMain(argc, argv, SbEventHandle);
 }


starboard/loader_app/loader_app.cc
```diff
@@ -309,6 +309,13 @@ void SbEventHandle(const SbEvent* event) {
   SB_CHECK_EQ(pthread_mutex_unlock(&mutex), 0);
 }
 
+#if SB_IS(EVERGREEN_COMPATIBLE)
+// StarboardMain is called from wpeframework cobalt plugin
+extern "C" SB_EXPORT_PLATFORM int StarboardMain(int argc, char** argv) {
```
andrunko, line 314: Can we add this to the RDK SB sources instead? Maybe on `src/third_party/starboard/rdk/shared/main_rdk.cc`? Asking cause I dont remember having to add it for cobalt 25, unsure if it was removed on cobalt 26 though, but I believe it would be better if we can make it specific to RDK.

santoshmahto7: C25 does it differently,  creates a preprocessor directive to replace main  with StarboardMain at buildtime like this
```
      if (!sb_is_modular || sb_is_evergreen) {
        # Rewrite main() functions into StarboardMain. TODO: This is a
        # hack, it would be better to be more surgical, here.
        defines = [ "main=StarboardMain" ]
      }
```
https://github.com/youtube/cobalt/blob/6f52f5041219a06a4f62b51a91a354e9f0ded0a1/starboard/build/config/BUILD.gn#L152C1-L156C8

But this can be added in main_rdk.cc - need fixing(isn't built for now in evergreen_compatible mode) also main already there conflicts (double definition main) when built

lamdbui: It's preferred to do this in the `rdk` SB source, if possible to keep this clean as this is a generic implementation.

> StarboardMain is called from wpeframework cobalt plugin

We don't want anything referencing more platform-specific stuff like `wpeframework cobalt plugin` in here.

santoshmahto7: Make sense, I will move it to main_rdk.cc

santoshmahto7:
This PR is superseded with another Starboard PR
https://cobalt-review.googlesource.com/c/external/components/generic/cobalt/+/7130 
So I am closing it

PR #7548, opened 2025-10-15
Implement CrashAnnotator interface for 3P

hlwarriner:
This implements the `CrashLog.setString` web API on 3P platforms (confirmed with local testing).

This change involves introducing a new version of the CrashHandler Starboard Extension and wiring it up for 3P platforms. The new version enables a callback to be provided to the extension implementation so that it can delegate implementation of the `SetString`  function to a dependency. This allows us to avoid a build dependency cycle for 3P platforms: `//starboard/crashpad_wrapper` indirectly depends on `//starboard:starboard_group` via `//base`, so we want to avoid also having `//starboard:starboard_group` depend on `//starboard/crashpad_wrapper`. We wouldn't have had this problem in c25 because the Crashpad client library was built on top of `mini_chromium` instead of `//base`.

Issue: 446908034
Issue: 400709539

### DIFF
--- a/starboard/extension/extension_test.cc
+++ b/starboard/extension/extension_test.cc
@@ -233,11 +233,15 @@ TEST(ExtensionTest, CrashHandler) {
 
   EXPECT_STREQ(extension_api->name, kExtensionName);
   EXPECT_GE(extension_api->version, 1u);
-  EXPECT_LE(extension_api->version, 2u);
+  EXPECT_LE(extension_api->version, 3u);
   EXPECT_NE(extension_api->OverrideCrashpadAnnotations, nullptr);
 
   if (extension_api->version >= 2) {
     EXPECT_NE(extension_api->SetString, nullptr);
+  }
+
+  if (extension_api->version >= 3) {
+    EXPECT_NE(extension_api->RegisterSetStringCallback, nullptr);
   }
 
   const ExtensionApi* second_extension_api =


andrewsavage1:
> note that SB_IS(MODULAR) is true for Evergreen compatible platforms on the main branch, which is a change from c25

This was true in c25, see i.e. https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/build/config/BUILDCONFIG.gn;l=126?q=%22sb_is_modular%20%3D%20%22&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F25.lts.1%2B

hlwarriner:
> > note that SB_IS(MODULAR) is true for Evergreen compatible platforms on the main branch, which is a change from c25
> 
> This was true in c25, see i.e. https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/build/config/BUILDCONFIG.gn;l=126?q=%22sb_is_modular%20%3D%20%22&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F25.lts.1%2B

But in c25 `sb_is_evergreen` was true for Evergreen platforms and false for Evergreen-compatible platforms, which are the platforms that needed to implement the Starboard extension. This Evergreen vs. Evergreen-compatible platform distinction was done away with on the main branch, IIUC, since we now just use toolchains to build Evergreen components vs. partner-built components.

SB_IS(MODULAR) couldn't have ben true for Evergreen compatible platforms on c25; if it were, the `setString()` web API wouldn't have been implemented correctly (https://github.com/youtube/cobalt/blob/main/starboard/shared/starboard/crash_handler.cc#L30).

andrewsavage1:
> > > note that SB_IS(MODULAR) is true for Evergreen compatible platforms on the main branch, which is a change from c25
> > 
> > 
> > This was true in c25, see i.e. https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/build/config/BUILDCONFIG.gn;l=126?q=%22sb_is_modular%20%3D%20%22&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F25.lts.1%2B
> 
> But in c25 `sb_is_evergreen` was true for Evergreen platforms and false for Evergreen-compatible platforms, which are the platforms that needed to implement the Starboard extension. This Evergreen vs. Evergreen-compatible platform distinction was done away with on the main branch, IIUC, since we now just use toolchains to build Evergreen components vs. partner-built components.
> 
> SB_IS(MODULAR) couldn't have ben true for Evergreen compatible platforms on c25; if it were, the `setString()` web API wouldn't have been implemented correctly (https://github.com/youtube/cobalt/blob/main/starboard/shared/starboard/crash_handler.cc#L30).

Sorry, I misread your original message and missed "compatible"—yes you're right

cobalt/browser/crash_annotator/crash_annotator_impl.cc
```diff
@@ -34,10 +36,17 @@ void CrashAnnotatorImpl::Create(
...
                                    const std::string& value,
                                    SetStringCallback callback) {
-  // TODO(cobalt, b/383301493): actually implement this.
-  LOG(INFO) << "CrashAnnotatorImpl::SetString key=" << key
-            << " value=" << value;
-  std::move(callback).Run(false);
+  auto crash_handler_extension =
+      static_cast<const CobaltExtensionCrashHandlerApi*>(
+          SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+  if (crash_handler_extension && crash_handler_extension->version >= 2) {
```
andrewsavage1, line 42: Will this work for version 2 or only version 3?

starboard/shared/starboard/crash_handler.cc
```diff
@@ -22,23 +25,35 @@ namespace starboard {
...
+  SB_CHECK_EQ(pthread_mutex_unlock(&g_set_string_callback_mutex), 0);
 }
 
 const CobaltExtensionCrashHandlerApi kCrashHandlerApi = {
-    kCobaltExtensionCrashHandlerName,
-    2,
-    &OverrideCrashpadAnnotations,
-    &SetString,
+    kCobaltExtensionCrashHandlerName, 3,
+    &OverrideCrashpadAnnotations,     &SetString,
```
andrewsavage1, line 55: Is this how clang-format wants to format it?

hlwarriner: I was surprised by this too but yeah, that's what clang-format recommended when it ran on commit 3.

cobalt/browser/crash_annotator/crash_annotator_impl.cc
hlwarriner: Yeah it should work for versions 2 and 3.

From the perspective of the cobalt layer there's no difference between versions 2 and 3. The cobalt layer just needs the `SetString` function to be present and to do its job - how it does that is up to the platform that implements it.

hlwarriner:
I'm merging this change now, before the main branch is locked for the M138 merge on Monday. I'll address any remaining feedback in a separate PR once the main branch is unlocked.

PR #7552, opened 2025-10-15
Fix the tracing service

y4vor:
- Disable video input which starts the Video Capture Service which in turn locks up the utility thread and prevents any other utility service to run e.g Tracing.
- Enable the OS memory metrics in Tracing and just turn off the build id read which relies or dladdr.
- Fixed the uname wrapper for modular builds. It wasn't hooked up in the build.

Issue: 452366512

### DIFF
--- a/content/browser/renderer_host/media/media_devices_dispatcher_host.cc
+++ b/content/browser/renderer_host/media/media_devices_dispatcher_host.cc
@@ -149,8 +149,11 @@ void MediaDevicesDispatcherHost::EnumerateDevices(
   MediaDevicesManager::BoolDeviceTypes devices_to_enumerate;
   devices_to_enumerate[static_cast<size_t>(
       MediaDeviceType::MEDIA_AUDIO_INPUT)] = request_audio_input;
+// Starboard doesn't support video input.
+#if !BUILDFLAG(IS_STARBOARD)
   devices_to_enumerate[static_cast<size_t>(
       MediaDeviceType::MEDIA_VIDEO_INPUT)] = request_video_input;
+#endif
   devices_to_enumerate[static_cast<size_t>(
       MediaDeviceType::MEDIA_AUDIO_OUTPUT)] = request_audio_output;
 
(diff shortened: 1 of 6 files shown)


sherryzy (approved):
LGTM!

content/browser/renderer_host/media/media_devices_dispatcher_host.cc
```diff
@@ -149,8 +149,11 @@ void MediaDevicesDispatcherHost::EnumerateDevices(
   MediaDevicesManager::BoolDeviceTypes devices_to_enumerate;
   devices_to_enumerate[static_cast<size_t>(
       MediaDeviceType::MEDIA_AUDIO_INPUT)] = request_audio_input;
+// Starboard doesn't support video input.
+#if !BUILDFLAG(IS_STARBOARD)
```
yell0wd0g, line 153: Could you try once with the command flag [`--use-fake-device-for-media-stream`](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:media/base/media_switches.cc;l=138-141?q=fake-video-ca&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain) ?

That should enumerate a few known test formats and should unblock
all this enumeration. 

(We'd just have to add `switches::kUseFakeDeviceForMediaStream`
to [this line](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/app/cobalt_switch_defaults_starboard.cc;l=79)).

Only if you want to avoid instrumenting this code. 

All this MediaStreamManager/MediaDevicesManager/MediaDevices*Host etc,
I would like to remove it altogether via build time flag, but it's
very embedded and ideally we'd solve it upstream first.

y4vor: The flag works! However looking at the documentation of what it does it would actually turn off the microphone support too, which we don't want to do.
https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/base/media_switches.cc;l=138?q=kUseFakeDeviceForMediaStream&sq=r:github%2Fyoutube%2Fcobalt%20b:main.

I'll keep the current code that only turns off the Video Input enumeration.

PR #7555, opened 2025-10-15
Fixes for EG starboard and native toolchains

andrewsavage1:
Bug: 452357142

### DIFF
--- a/base/BUILD.gn
+++ b/base/BUILD.gn
@@ -121,6 +121,12 @@ if (is_fuchsia) {
 
 _build_date_header = "$target_gen_dir/generated_build_date.h"
 
+if (is_cobalt) {
+  _should_build_for_platform =
+      (is_cobalt_hermetic_build && current_toolchain == starboard_toolchain) ||
+      current_toolchain == native_target
+}
+
 if (is_win) {
   # This is in a separate config so the flags can be applied to dependents.
   # ldflags in GN aren't automatically inherited.
@@ -1062,7 +1068,7 @@ component("base") {
     "//build/config/compiler:compiler_buildflags",
     "//third_party/modp_b64",
   ]
-  if (!is_nacl) {
+  if (!is_nacl && (!is_cobalt || !_should_build_for_platform)) {
     # Used by metrics/crc32, except on NaCl builds.
     deps += [ "//third_party/zlib" ]
 
@@ -1106,7 +1112,7 @@ component("base") {
     "//third_party/abseil-cpp:absl",
   ]
 
-  if (!is_nacl) {
+  if (!is_nacl && (!is_cobalt || !_should_build_for_platform)) {
     sources += [
       "containers/span_rust.h",
       "strings/string_view_rust.h",
@@ -1785,6 +1791,13 @@ component("base") {
     if (use_allocator_shim) {
       public_deps += [ "allocator/partition_allocator:allocator_shim" ]
     }
+
+    if (is_cobalt && _should_build_for_platform) {
+      sources -= [
+        "logging/rust_log_integration.cc",
+        "logging/rust_log_integration.h",
+      ]
+    }
   }
 
   if (is_starboard) {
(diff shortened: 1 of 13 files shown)


base/BUILD.gn
```diff
@@ -121,6 +121,12 @@ if (is_fuchsia) {
 
 _build_date_header = "$target_gen_dir/generated_build_date.h"
 
+if (is_cobalt) {
+  _should_build_for_platform =
```
y4vor, line 125: Is this something we are introducing? I'm not sure what it represents.

base/BUILD.gn
```diff
@@ -1106,7 +1112,9 @@ component("base") {
     "//third_party/abseil-cpp:absl",
   ]
 
-  if (!is_nacl) {
+  if (!is_nacl && (!is_cobalt || ((!is_cobalt_hermetic_build ||
```
y4vor: Shouldn't this be using _should_build_for_platform?

base/containers/span_rust.h
```diff
@@ -11,7 +11,7 @@
 #include "build/build_config.h"
 #include "third_party/rust/cxx/v1/cxx.h"
 
-#if BUILDFLAG(IS_NACL)
+#if BUILDFLAG(IS_NACL) || BUILDFLAG(IS_COBALT) && !defined(SB_IS_DEFAULT_TC)
```
y4vor, line 14: What is the case where it is Cobalt but not using the cobalt toolchain?

base/numerics/byte_conversions.h
```diff
@@ -133,7 +157,7 @@ inline constexpr int64_t I64FromNativeEndian(
 // storage, and explicit big endian for network order.
 inline constexpr float FloatFromNativeEndian(
     std::span<const uint8_t, 4u> bytes) {
-  return std::bit_cast<float>(U32FromNativeEndian(bytes));
+  return BitCast<float>(U32FromNativeEndian(bytes));
```
y4vor, line 158: Same here and below.

base/numerics/byte_conversions.h
```diff
@@ -15,11 +14,36 @@
...
+#if !BUILDFLAG(IS_COBALT) || defined(SB_IS_DEFAULT_TC)
+#include <bit>
+#endif
+
 // Chromium only builds and runs on Little Endian machines.
 static_assert(ARCH_CPU_LITTLE_ENDIAN);
 
 namespace base {
 
+namespace {
```
y4vor, line 26: Guard with BUILDFLAG(IS_COBALT)

build/config/compiler/BUILD.gn
```diff
@@ -1081,6 +1081,9 @@ config("compiler") {
     # We use clang-rt sanitizer runtimes.
     "-Zexternal-clangrt",
   ]
+  if (current_toolchain == default_toolchain) {
```
y4vor: add is_cobalt?

build/config/compiler/BUILD.gn
```diff
@@ -2094,6 +2097,7 @@ config("default_warnings") {
 
     cflags_cc += [
       # Disables for C++ only
+      "-Wno-shift-count-overflow",
       "-Wno-invalid-offsetof",  # offsetof on non-standard-layout type
```
y4vor, line 2100: is_cobalt?

third_party/abseil-cpp/absl/time/time.cc
```diff
@@ -503,5 +503,42 @@ struct tm ToTM(absl::Time t, absl::TimeZone tz) {
   return tm;
 }
 
+// Not all Cobalt toolchains can compile these functions as constexpr as is
```
y4vor, line 507: Guard with BUILDFLAG(IS_COBALT) here and the rest of the files.

base/BUILD.gn
andrewsavage1: I'm trying to distinguish between the evergreen default/cobalt and host toolchains (which can use modern c++ and support rust) and the native and starboard toolchains (which cannot right now)

base/containers/span_rust.h
andrewsavage1: BUILDFLAG(IS_COBALT) is true across all toolchains for cobalt platforms

base/BUILD.gn
andrewsavage1: It means it is the hermetic build building for the native or starboard toolchains (which is what I mean by "for the platform)—I can rename in the future if necessary

build/config/compiler/BUILD.gn
andrewsavage1: removed

PR #7561, opened 2025-10-16
Changes to get a simplified CI to build for m138.

SortaCookie:
Stripped out tests, simplified build configs to only build devel for now and skipping rewrapper cipd fetch.

Bug: 452643486
Change-Id: Ia1538253f11d1e3fb8d8a6adff405b1b3e958bcc

### DIFF
--- a/.github/config/android-arm.json
+++ b/.github/config/android-arm.json
@@ -3,7 +3,7 @@
   "platforms": [
     "android-arm"
   ],
-  "test_on_device": true,
+  "test_on_device": false,
   "test_device_family": "android",
   "test_e2e": true,
   "test_root_target": "//cobalt/android:cobalt_apk",
--- a/.github/config/android-arm64.json
+++ b/.github/config/android-arm64.json
@@ -3,9 +3,9 @@
   "platforms": [
     "android-arm64"
   ],
-  "test_on_device": true,
+  "test_on_device": false,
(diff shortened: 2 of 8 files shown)


.github/actions/depot_tools/action.yaml
```diff
@@ -27,6 +27,16 @@ runs:
           echo "target_os=['android']" >> .gclient
           gclient validate
         shell: bash
+      - name: Inject siso env file [ hack ]
```
andrewsavage1: Siso is commented out in the DEPS: https://github.com/youtube/cobalt/blob/feature/rebase_m138/DEPS#L5723-L5732

Is this still necessary?

.github/config/android-arm.json
```diff
@@ -3,35 +3,8 @@
   "platforms": [
     "android-arm"
   ],
-  "test_on_device": true,
```
andrewsavage1, line 6: Instead of removing all of this could we simply set test_on_device and test_e2e both to false for now?

(relevant for all instances like this)

.github/actions/depot_tools/action.yaml
SortaCookie: I think it's still necessary because when I ran this without the step there were complaints about a missing config file and I had used feature/rebase_m138 from Monday when it was commented out.

PR #7571, opened 2025-10-16
Plug Evergreen API leaks

andrewsavage1:
Bug: 452357142

### DIFF
--- a/content/app/content_main_runner_impl.cc
+++ b/content/app/content_main_runner_impl.cc
@@ -1096,7 +1096,7 @@ int ContentMainRunnerImpl::Initialize(ContentMainParams params) {
   }
 #endif  // BUILDFLAG(USE_ZYGOTE)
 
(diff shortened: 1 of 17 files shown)


briantting (approved):

madhurajayaraman:
I think we generally want to keep cobalt specific #if else statements on separate lines to reduce future merge conflicts - the majority of rebase issues came from if/else blocks - if it's a pain to change we can leave as is and clean up later.

andrewsavage1:
> I think we generally want to keep cobalt specific #if else statements on separate lines to reduce future merge conflicts - the majority of rebase issues came from if/else blocks - if it's a pain to change we can leave as is and clean up later.

filed b/452634589, but we should probably think about if we actually want to do this, as if the original check changes, we may very well might want to change our check and _not_ have it resolve the conflict automatically

PR #7575, opened 2025-10-16
Fix the evergreen cobalt_loader build

andrewsavage1:
Bug: 452357142


andrewsavage1:
> If it's not too annoying can you separate out the starboard/cobalt specific conditionals onto a separate line? that's where most of the rebase merge conflicts come from.

Done where possible, though I'm still not sure whether we want to do this long term (see b/452634589)

PR #7578, opened 2025-10-16
android: Replace JNI prefix flag with GN build flag

zhongqiliang:
Replace the cobalt-jni-prefix command-line switch with the
is_cobalt_on_google3 GN build flag. The JNI prefix logic in
jni_android.cc now uses this compile-time flag to determine if the
cobalt/ prefix should be applied to JNI class names. The Android
packaging script is also updated to enable this flag for relevant builds.

This change addresses a bug where base::CommandLine::InitializedForCurrentProcess()
would return false at app startup. This occurred even before libcronet.so
was excluded, preventing the command-line switch from being honored.
This led to failures in loading org/chromium/base/JNIUtils. Using a GN
flag ensures the correct JNI behavior is established at build time,
resolving this class loading bug.

Bug: 452059316
Change-Id: I38a02bc54928e892d28190930df0b3c3eab6711

### DIFF
--- a/cobalt/build/android/package.py
+++ b/cobalt/build/android/package.py
@@ -36,7 +36,7 @@ PLATFORMS = ['android-arm', 'android-arm64', 'android-x86']
 CONFIGS = ['gold', 'qa']
 
 
-def patch_args(out_dir):
+def patch_args_for_google3(out_dir):
   """
   Ensures 'is_cobalt_on_google3 = true' is in the args.gn file.
   Will raise FileNotFoundError if args.gn is missing.
@@ -61,7 +61,7 @@ def build(platforms_to_package):
       subprocess.call(
           ['cobalt/build/gn.py', out_dir, '-p', platform, '-c', config])
 
-      patch_args(out_dir)
+      patch_args_for_google3(out_dir)
 
       subprocess.call(['autoninja', '-C', out_dir, 'cobalt:gn_all'])
 


base/android/jni_android.cc
```diff
@@ -61,11 +59,8 @@ std::string getRepackagedName(const char* signature) {
 }
 
 bool shouldAddCobaltPrefix() {
-  if (!g_checked_command_line && base::CommandLine::InitializedForCurrentProcess()) {
-    g_add_cobalt_prefix = base::CommandLine::ForCurrentProcess()->HasSwitch(switches::kCobaltJniPrefix);
-    g_checked_command_line = true;
-  }
-  return g_add_cobalt_prefix;
+  // The flag is set in cobalt/build/android/package.py
+  return BUILDFLAG(IS_COBALT_ON_GOOGLE3);
```
andrewsavage1: Let's use the macro as a macro—

```
#if BUILDFLAG(IS_COBALT_ON_GOOGLE3)
  return true;
#else
  return false;
#endif

File: cobalt/build/configs/cobalt_for_google.gni [hunk 2]
```diff
@@ -0,0 +1,3 @@
+declare_args() {
```
andrewsavage1, line 15: Add a copyright header

cobalt/build/android/package.py
```diff
@@ -30,15 +30,45 @@
...
+      if not any(line.strip().startswith('is_cobalt_on_google3') for line in f):
+        # The argument was not found, append it, ensuring a newline.
+        f.write('\nis_cobalt_on_google3 = true\n')
+  except FileNotFoundError:
+    print(f'ERROR: Could not find {args_gn_path} to modify.')
+    return False
+
+  # Run 'gn gen' again to apply the new argument
+  print(f'Re-running gn gen for {out_dir} to apply appended args...')
+  if subprocess.call(['gn', 'gen', out_dir]) != 0:
```
andrewsavage1: You don't need to run this explicitly, autoninja does it automatically

cobalt/build/android/package.py
```diff
@@ -30,15 +30,45 @@
...
+  'gn gen' to apply the change.
+  """
+  args_gn_path = os.path.join(out_dir, 'args.gn')
+
+  try:
+    with open(args_gn_path, 'r+', encoding='utf-8') as f:
+      if not any(line.strip().startswith('is_cobalt_on_google3') for line in f):
+        # The argument was not found, append it, ensuring a newline.
+        f.write('\nis_cobalt_on_google3 = true\n')
+  except FileNotFoundError:
```
andrewsavage1: What's the reason we want to catch errors and call sys.exit(1) instead of just using the error originally raised? If you do want to do it this way, instead use `os.path.exists` with a print and return instead of a try/except

cobalt/build/android/package.py
zhongqiliang: removed try catch

cobalt/build/android/package.py
```diff
@@ -30,28 +30,53 @@
 
 import argparse
 import subprocess
+import os
 
 PLATFORMS = ['android-arm', 'android-arm64', 'android-x86']
 CONFIGS = ['gold', 'qa']
 
 
+def patch_args(out_dir):
```
johnxwork: You are doing a very specific thing here, don't use such a generic name. Make this function name more obvious about what it does. People will ask what's the implication of "is_cobalt_on_google3".

johnxwork (approved):

PR #7580, opened 2025-10-16
Disable pgo profiles for now

andrewsavage1:
Will break android official builds, but currently gclient sync is broken because of this.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -164,8 +164,7 @@ vars = {
   'checkout_clang_coverage_tools': False,
 
   # Fetch the pgo profiles to optimize official builds.
-  # Cobalt: Set to True for android official builds.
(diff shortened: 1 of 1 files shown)


DEPS
```diff
@@ -164,8 +164,7 @@ vars = {
   'checkout_clang_coverage_tools': False,
 
   # Fetch the pgo profiles to optimize official builds.
-  # Cobalt: Set to True for android official builds.
-  'checkout_pgo_profiles': True,
+  'checkout_pgo_profiles': False,
```
oxve, line 167: Do we want to add a TODO? Or is that superfluous?

andrewsavage1: I'm not sure of the right solution here, whether it's to change this value or something else—so I'm going to leave it without one for now

PR #7582, opened 2025-10-16
[Rebase m138] fix viz tests

madhurajayaraman:


components/viz/common/quads/draw_quad_perftest.cc
```diff
@@ -71,8 +71,8 @@ class DrawQuadPerfTest : public testing::Test {
   }
 
   void CleanUpRenderPass() {
-    render_pass_.reset();
```
andrewsavage1, line 74: Why this change? Is there any equivalent upstream change? If we're going to make it, even though it's small, it's in upstream code so we should wrap it with BUILDFLAG(IS_COBALT)

madhurajayaraman: This fixes the dangling pointer error, I'm not sure why we're suddenly throwing it when the test file hasn't changed since 2015?? I'll add the gates

components/viz/common/quads/draw_quad_perftest.cc
```diff
@@ -71,8 +71,13 @@ class DrawQuadPerfTest : public testing::Test {
   }
 
   void CleanUpRenderPass() {
+#if BUILDFLAG(IS_COBALT)
```
andrewsavage1, line 76: Add a short note about why we do this (it's impossible to tell just by reading the code)

PR #7584, opened 2025-10-16
Refactor MediaCapabilitiesCache for Tests

johnedocampo:
The current implementation of `MediaCapabilitiesCache` makes it difficult to create isolated unit tests. This PR aims to refactor the `MediaCapabilitiesCache` implementation to allow for robust unit testing. This PR also adds some unittests for MediaCapabilitiesCache, verifying its behavior.

This is mainly achieved through dependency injection. Instead of `MediaCapabilitiesCache` directly calling other APIs (e.x. `MediaDrmBridge`, `AudioOutputManager`) to fill its cache, we instead inject provider methods that the cache will call. This enables us to inject our own mock methods for the cache, which helps create a well isolated testing environment.

The original logic of how the cache obtains its values is largely unchanged; the original code has been transported to provider classes that will be injected into `MediaCapabilitiesCache` in production. 

This PR is based off of https://github.com/youtube/cobalt/pull/7564, which provided the main idea of how this implementation should look like.

Bug: 433773029

### DIFF
--- a/starboard/android/shared/media_capabilities_cache.cc
+++ b/starboard/android/shared/media_capabilities_cache.cc
@@ -85,66 +85,114 @@ void ConvertStringToLowerCase(std::string* str) {
   }
 }
 
-bool GetIsWidevineSupported() {
-  return MediaDrmBridge::IsWidevineSupported(AttachCurrentThread());
-}
-
-bool GetIsCbcsSupported() {
-  return MediaDrmBridge::IsCbcsSupported(AttachCurrentThread());
-}
-
-std::set<SbMediaTransferId> GetSupportedHdrTypes() {
-  std::set<SbMediaTransferId> supported_transfer_ids;
-
-  JNIEnv* env = AttachCurrentThread();
-  ScopedJavaLocalRef<jintArray> j_supported_hdr_types =
-      StarboardBridge::GetInstance()->GetSupportedHdrTypes(env);
-
-  if (!j_supported_hdr_types) {
-    // Failed to get supported hdr types.
-    SB_LOG(ERROR) << "Failed to load supported hdr types.";
-    return std::set<SbMediaTransferId>();
+class MediaCapabilitiesProviderImpl : public MediaCapabilitiesProvider {
+  bool GetIsWidevineSupported() override {
+    return MediaDrmBridge::IsWidevineSupported(AttachCurrentThread());
   }
+  bool GetIsCbcsSchemeSupported() override {
+    return MediaDrmBridge::IsCbcsSupported(AttachCurrentThread());
+  }
+  std::set<SbMediaTransferId> GetSupportedHdrTypes() override {
+    std::set<SbMediaTransferId> supported_transfer_ids;
 
-  jsize length = env->GetArrayLength(j_supported_hdr_types.obj());
-  jint* numbers =
-      env->GetIntArrayElements(j_supported_hdr_types.obj(), nullptr);
-  for (int i = 0; i < length; i++) {
-    switch (numbers[i]) {
-      case HDR_TYPE_DOLBY_VISION:
+    JNIEnv* env = AttachCurrentThread();
+    ScopedJavaLocalRef<jintArray> j_supported_hdr_types =
+        StarboardBridge::GetInstance()->GetSupportedHdrTypes(env);
+
+    if (!j_supported_hdr_types) {
+      // Failed to get supported hdr types.
+      SB_LOG(ERROR) << "Failed to load supported hdr types.";
+      return std::set<SbMediaTransferId>();
+    }
+
+    jsize length = env->GetArrayLength(j_supported_hdr_types.obj());
+    jint* numbers =
+        env->GetIntArrayElements(j_supported_hdr_types.obj(), nullptr);
+    for (int i = 0; i < length; i++) {
+      switch (numbers[i]) {
+        case HDR_TYPE_DOLBY_VISION:
+          continue;
+        case HDR_TYPE_HDR10:
+          supported_transfer_ids.insert(kSbMediaTransferIdSmpteSt2084);
+          continue;
+        case HDR_TYPE_HLG:
+          supported_transfer_ids.insert(kSbMediaTransferIdAribStdB67);
+          continue;
+        case HDR_TYPE_HDR10_PLUS:
+          continue;
+      }
+    }
+    env->ReleaseIntArrayElements(j_supported_hdr_types.obj(), numbers, 0);
+
+    return supported_transfer_ids;
+  }
+  bool GetIsPassthroughSupported(SbMediaAudioCodec codec) override {
+    SbMediaAudioCodingType coding_type;
+    switch (codec) {
+      case kSbMediaAudioCodecAc3:
+        coding_type = kSbMediaAudioCodingTypeAc3;
+        break;
+      case kSbMediaAudioCodecEac3:
+        coding_type = kSbMediaAudioCodingTypeDolbyDigitalPlus;
+        break;
+      default:
+        return false;
+    }
+    int encoding = GetAudioFormatSampleType(coding_type);
+    JNIEnv* env = AttachCurrentThread();
+    return AudioOutputManager::GetInstance()->HasPassthroughSupportFor(
+        env, encoding);
+  }
+  bool GetAudioConfiguration(
+      int index,
+      SbMediaAudioConfiguration* configuration) override {
+    JNIEnv* env = AttachCurrentThread();
+    return AudioOutputManager::GetInstance()->GetAudioConfiguration(
+        env, index, configuration);
+  }
+  void GetCodecCapabilities(
+      std::map<std::string, AudioCodecCapabilities>& audio_codec_capabilities,
+      std::map<std::string, VideoCodecCapabilities>& video_codec_capabilities)
+      override {
+    JNIEnv* env = AttachCurrentThread();
+    ScopedJavaLocalRef<jobjectArray> j_codec_infos =
+        Java_MediaCodecUtil_getAllCodecCapabilityInfos(env);
+    jsize length = env->GetArrayLength(j_codec_infos.obj());
+
+    // Note: Codec infos are sorted by the framework such that the best
+    // decoders come first.
+    // This order is maintained in the cache.
+    for (int i = 0; i < length; i++) {
+      ScopedJavaLocalRef<jobject> j_codec_info(
+          env, env->GetObjectArrayElement(j_codec_infos.obj(), i));
+      SB_CHECK(j_codec_info);
+
+      ScopedJavaLocalRef<jstring> j_mime_type =
+          Java_CodecCapabilityInfo_getMimeType(env, j_codec_info);
+      std::string mime_type = ConvertJavaStringToUTF8(env, j_mime_type.obj());
+      // Convert the mime type to lower case.
+      ConvertStringToLowerCase(&mime_type);
+
+      ScopedJavaLocalRef<jobject> j_audio_capabilities =
+          Java_CodecCapabilityInfo_getAudioCapabilities(env, j_codec_info);
+      if (j_audio_capabilities) {
+        // Found an audio decoder.
+        audio_codec_capabilities[mime_type].push_back(
+            std::make_unique<AudioCodecCapability>(env, j_codec_info,
+                                                   j_audio_capabilities));
         continue;
-      case HDR_TYPE_HDR10:
-        supported_transfer_ids.insert(kSbMediaTransferIdSmpteSt2084);
-        continue;
-      case HDR_TYPE_HLG:
-        supported_transfer_ids.insert(kSbMediaTransferIdAribStdB67);
-        continue;
-      case HDR_TYPE_HDR10_PLUS:
-        continue;
+      }
+      ScopedJavaLocalRef<jobject> j_video_capabilities =
+          Java_CodecCapabilityInfo_getVideoCapabilities(env, j_codec_info);
+      if (j_video_capabilities) {
+        // Found a video decoder.
+        video_codec_capabilities[mime_type].push_back(
+            std::make_unique<VideoCodecCapability>(env, j_codec_info,
+                                                   j_video_capabilities));
+      }
     }
   }
-  env->ReleaseIntArrayElements(j_supported_hdr_types.obj(), numbers, 0);
-
-  return supported_transfer_ids;
-}
-
-bool GetIsPassthroughSupported(SbMediaAudioCodec codec) {
-  SbMediaAudioCodingType coding_type;
-  switch (codec) {
-    case kSbMediaAudioCodecAc3:
-      coding_type = kSbMediaAudioCodingTypeAc3;
-      break;
-    case kSbMediaAudioCodecEac3:
-      coding_type = kSbMediaAudioCodingTypeDolbyDigitalPlus;
-      break;
-    default:
-      return false;
-  }
-  int encoding = GetAudioFormatSampleType(coding_type);
-  JNIEnv* env = AttachCurrentThread();
-  return AudioOutputManager::GetInstance()->HasPassthroughSupportFor(env,
-                                                                     encoding);
-}
+};
 }  // namespace
 
 CodecCapability::CodecCapability(JNIEnv* env,
@@ -237,9 +285,27 @@ bool VideoCodecCapability::AreResolutionAndRateSupported(int frame_width,
 SB_ONCE_INITIALIZE_FUNCTION(MediaCapabilitiesCache,
                             MediaCapabilitiesCache::GetInstance)
 
+std::unique_ptr<MediaCapabilitiesCache> MediaCapabilitiesCache::CreateForTest(
+    std::unique_ptr<MediaCapabilitiesProvider> media_capabilities_provider) {
+  return std::unique_ptr<MediaCapabilitiesCache>(
+      new MediaCapabilitiesCache(std::move(media_capabilities_provider)));
+}
+
+MediaCapabilitiesCache::MediaCapabilitiesCache()
+    : MediaCapabilitiesCache(
+          std::make_unique<MediaCapabilitiesProviderImpl>()) {}
+
+MediaCapabilitiesCache::MediaCapabilitiesCache(
+    std::unique_ptr<MediaCapabilitiesProvider> media_capabilities_provider)
+    : media_capabilities_provider_(std::move(media_capabilities_provider)) {
+  // Enable mime and key system caches.
+  MimeSupportabilityCache::GetInstance()->SetCacheEnabled(true);
+  KeySystemSupportabilityCache::GetInstance()->SetCacheEnabled(true);
+}
+
 bool MediaCapabilitiesCache::IsWidevineSupported() {
   if (!is_enabled_) {
-    return GetIsWidevineSupported();
+    return media_capabilities_provider_->GetIsWidevineSupported();
   }
   std::lock_guard scoped_lock(mutex_);
   UpdateMediaCapabilities_Locked();
@@ -248,7 +314,7 @@ bool MediaCapabilitiesCache::IsWidevineSupported() {
 
 bool MediaCapabilitiesCache::IsCbcsSchemeSupported() {
   if (!is_enabled_) {
-    return GetIsCbcsSupported();
+    return media_capabilities_provider_->GetIsCbcsSchemeSupported();
   }
   std::lock_guard scoped_lock(mutex_);
   UpdateMediaCapabilities_Locked();
@@ -258,7 +324,8 @@ bool MediaCapabilitiesCache::IsCbcsSchemeSupported() {
 bool MediaCapabilitiesCache::IsHDRTransferCharacteristicsSupported(
     SbMediaTransferId transfer_id) {
   if (!is_enabled_) {
-    std::set<SbMediaTransferId> supported_transfer_ids = GetSupportedHdrTypes();
+    std::set<SbMediaTransferId> supported_transfer_ids =
+        media_capabilities_provider_->GetSupportedHdrTypes();
     return supported_transfer_ids.find(transfer_id) !=
            supported_transfer_ids.end();
   }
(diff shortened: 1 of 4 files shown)


starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -237,9 +319,54 @@ bool VideoCodecCapability::AreResolutionAndRateSupported(int frame_width,
...
                             MediaCapabilitiesCache::GetInstance)
 
+std::unique_ptr<MediaCapabilitiesCache> MediaCapabilitiesCache::CreateForTest(
+    std::unique_ptr<DrmCapabilitiesProvider> drm_capabilities_provider,
+    std::unique_ptr<HdrCapabilitiesProvider> hdr_capabilities_provider,
+    std::unique_ptr<PassthroughCapabilitiesProvider>
+        passthrough_capabilities_provider,
+    std::unique_ptr<AudioConfigurationProvider> audio_configuration_provider,
+    std::unique_ptr<CodecCapabilitiesProvider> codec_capabilities_provider) {
+  return std::unique_ptr<MediaCapabilitiesCache>(
```
kjyoun, line 290: How about this?

```
return std::make_unique<MediaCapabilitiesCache>(
    std::move(drm_capabilities_provider),
    std::move(hdr_capabilities_provider),
    std::move(passthrough_capabilities_provider),
    std::move(audio_configuration_provider),
    std::move(codec_capabilities_provider));
```

kjyoun (commented):
Can you add explanation about this PR in description?

e.g.) why/how you introduced dependencies injection,.

starboard/android/shared/media_capabilities_cache.cc
johnedocampo: Interestingly, std::make_unique doesn't work as it is trying to use a private ctor. I think its important that the normal ctors for MediaCapabilitiesCache are kept private, to push the idea that this class should only be created/used with `GetInstance.` I think the normal method is okay.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -237,9 +319,54 @@ bool VideoCodecCapability::AreResolutionAndRateSupported(int frame_width,
...
+MediaCapabilitiesCache::MediaCapabilitiesCache()
+    : MediaCapabilitiesCache(
+          std::make_unique<DrmCapabilitiesProviderImpl>(),
+          std::make_unique<HdrCapabilitiesProviderImpl>(),
+          std::make_unique<PassthroughCapabilitiesProviderImpl>(),
+          std::make_unique<AudioConfigurationProviderImpl>(),
+          std::make_unique<CodecCapabilitiesProviderImpl>()) {
+  // Enable mime and key system caches.
+  MimeSupportabilityCache::GetInstance()->SetCacheEnabled(true);
+  KeySystemSupportabilityCache::GetInstance()->SetCacheEnabled(true);
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -85,66 +85,148 @@ void ConvertStringToLowerCase(std::string* str) {
...
+      }
+    }
+    return video_codec_capabilities_map;
   }
-  int encoding = GetAudioFormatSampleType(coding_type);
-  JNIEnv* env = AttachCurrentThread();
-  return AudioOutputManager::GetInstance()->HasPassthroughSupportFor(env,
-                                                                     encoding);
-}
+};
```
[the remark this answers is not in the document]
johnedocampo: Done; CodecCapabilitiesProvider now only has one function call that will return both video and audio capabilities.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -289,9 +401,13 @@ bool MediaCapabilitiesCache::GetAudioConfiguration(
...
-    JNIEnv* env = AttachCurrentThread();
-    return AudioOutputManager::GetInstance()->GetAudioConfiguration(
-        env, index, configuration);
+    std::vector<SbMediaAudioConfiguration> configurations =
+        audio_configuration_provider_->GetAudioConfigurations();
+    if (static_cast<size_t>(index) < configurations.size()) {
+      *configuration = configurations[index];
+      return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -115,10 +115,57 @@ class VideoCodecCapability : public CodecCapability {
...
+ public:
+  virtual ~PassthroughCapabilitiesProvider() = default;
+  virtual bool GetIsPassthroughSupported(SbMediaAudioCodec codec) = 0;
+};
+
+class AudioConfigurationProvider {
+ public:
+  virtual ~AudioConfigurationProvider() = default;
+  virtual std::vector<SbMediaAudioConfiguration> GetAudioConfigurations() = 0;
+};
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -435,69 +545,17 @@ void MediaCapabilitiesCache::UpdateMediaCapabilities_Locked() {
...
+    is_cbcs_supported_ = drm_capabilities_provider_->IsCbcsSchemeSupported();
+    supported_transfer_ids_ =
+        hdr_capabilities_provider_->GetSupportedHdrTypes();
+
+    CodecCapabilities codecs =
+        codec_capabilities_provider_->GetCodecCapabilities();
+    audio_codec_capabilities_map_ = std::move(codecs.audio_capabilities);
+    video_codec_capabilities_map_ = std::move(codecs.video_capabilities);
+    audio_configurations_ =
+        audio_configuration_provider_->GetAudioConfigurations();
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -85,66 +85,135 @@ void ConvertStringToLowerCase(std::string* str) {
...
+    SbMediaAudioConfiguration configuration;
+    JNIEnv* env = AttachCurrentThread();
+    while (audio_configurations.size() < kMaxAudioConfigurations &&
+           AudioOutputManager::GetInstance()->GetAudioConfiguration(
+               env, audio_configurations.size(), &configuration)) {
+      audio_configurations.push_back(configuration);
+    }
+    return audio_configurations;
+  }
+};
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -0,0 +1,312 @@
...
+              (int index, SbMediaAudioConfiguration* configuration),
+              (override));
+};
+
+ACTION(ReturnEmptyCodecCapabilities) {
+  return CodecCapabilities{
+      std::map<std::string,
+               std::vector<std::unique_ptr<AudioCodecCapability>>>(),
+      std::map<std::string,
+               std::vector<std::unique_ptr<VideoCodecCapability>>>()};
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.cc
kjyoun: It makes sense

starboard/android/shared/media_capabilities_cache.h
kjyoun, line 204: deleted typo that I made

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -420,12 +523,6 @@ std::string MediaCapabilitiesCache::FindVideoDecoder(
...
 }
 
-MediaCapabilitiesCache::MediaCapabilitiesCache() {
-  // Enable mime and key system caches.
-  MimeSupportabilityCache::GetInstance()->SetCacheEnabled(true);
-  KeySystemSupportabilityCache::GetInstance()->SetCacheEnabled(true);
-}
-
 void MediaCapabilitiesCache::UpdateMediaCapabilities_Locked() {
   if (capabilities_is_dirty_.exchange(false)) {
```
kjyoun: suggestion.

1) How about changing this method to `UpdateMediaCapabilitie()` and lock mutex inside?
```
void MediaCapabilitiesCache::UpdateMediaCapabilities() {
  std::lock_guard lock(mutex_);
  ...
```

Looks like that all callsite locks mutex just to call this method


2)
How about this pattern to reduce nesting?

go/tott/733.md
```
void MediaCapabilitiesCache::UpdateMediaCapabilities_Locked() {
  if (!capabilities_is_dirty_.exchange(false)) 
     return false; // fast fail here
  }

  // Then normal flow
  audio_codec_capabilities_map_.clear();
  ... 
```

starboard/android/shared/media_capabilities_cache.h
kjyoun: Do we need unique_ptr here?

How about just std::vector<AudioCodecCapability>?
```
using AudioCodecCapabilities = std::vector<AudioCodecCapability>
```

Though this is out of the scope of this PR, 
I think that it's a good chance to clean this up. Now we are protected by unit tests.

kjyoun:
Thanks a lot for enabling unit tests for Android and refactoring code to enable more unit tests.
It will definitely help the health of our code base.

starboard/android/shared/media_capabilities_cache.cc
johnedocampo: > 1. How about changing this method to `UpdateMediaCapabilitie()` and lock mutex inside?
> 
> ```
> void MediaCapabilitiesCache::UpdateMediaCapabilities() {
>   std::lock_guard lock(mutex_);
>   ...
> ```
> 
> Looks like that all callsite locks mutex just to call this method

I think its okay just to leave this as is, as some callsites like `MediaCapabilitiesCache::IsPassthroughSupported` and `MediaCapabilitiesCache::GetAudioConfiguration` have additional logic after `UpdateMediaCapabilities_Locked` that read the internals of the cache. If we move the mutex away from those functions, there may be race conditions where while one of these functions is reading values, the cache gets reset.
 
> How about this pattern to reduce nesting?
> 
> go/tott/733.md
> 
> ```
> void MediaCapabilitiesCache::UpdateMediaCapabilities_Locked() {
>   if (!capabilities_is_dirty_.exchange(false)) 
>      return false; // fast fail here
>   }
> 
>   // Then normal flow
>   audio_codec_capabilities_map_.clear();
>   ... 
> ```

Done.

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -115,10 +115,59 @@ class VideoCodecCapability : public CodecCapability {
   const Range supported_frame_rates_;
 };
 
+class DrmCapabilitiesProvider {
```
jasonzhangxx: I recommend to combine the providers into one MediaCapabilitiesProvider interface, to shorten the code. We can also consider to make MediaCapabilitiesProvider nested in MediaCapabilitiesCache.

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -115,10 +115,59 @@ class VideoCodecCapability : public CodecCapability {
...
+  std::map<std::string, std::vector<std::unique_ptr<AudioCodecCapability>>>
+      audio_capabilities;
+  std::map<std::string, std::vector<std::unique_ptr<VideoCodecCapability>>>
+      video_capabilities;
+};
+
+class CodecCapabilitiesProvider {
+ public:
+  virtual ~CodecCapabilitiesProvider() = default;
+  virtual CodecCapabilities GetCodecCapabilities() = 0;
```
jasonzhangxx: We can pass two references of |audio_capabilities| and |video_capabilities|, to avoid introducing a new struct.

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -0,0 +1,291 @@
...
+  EXPECT_TRUE(cache_->IsWidevineSupported());
+}
+
+TEST_F(MediaCapabilitiesCacheTest, IsWidevineSupported_DisabledCache) {
+  cache_->SetCacheEnabled(false);
+  EXPECT_CALL(*mock_drm_provider_, IsWidevineSupported())
+      .Times(2)
+      .WillRepeatedly(Return(true));
+  EXPECT_TRUE(cache_->IsWidevineSupported());
+  EXPECT_TRUE(cache_->IsWidevineSupported());
```
jasonzhangxx, line 88: Can we make the test first return true and then return false for second call?

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -0,0 +1,291 @@
...
+
+  EXPECT_FALSE(cache_->GetAudioConfiguration(2, &config));
+}
+
+TEST_F(MediaCapabilitiesCacheTest, ClearCache) {
+  EXPECT_CALL(*mock_drm_provider_, IsWidevineSupported())
+      .Times(2)
+      .WillRepeatedly(Return(true));
+  EXPECT_TRUE(cache_->IsWidevineSupported());
+  EXPECT_TRUE(cache_->IsWidevineSupported());
```
jasonzhangxx, line 283: We should also test other functions for cache clear.

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -0,0 +1,291 @@
...
+  SbMediaAudioConfiguration config;
+  EXPECT_TRUE(cache_->GetAudioConfiguration(0, &config));
+  EXPECT_EQ(config.coding_type, kSbMediaAudioCodingTypeAc3);
+
+  EXPECT_TRUE(cache_->GetAudioConfiguration(1, &config));
+  EXPECT_EQ(config.coding_type, kSbMediaAudioCodingTypeDolbyDigitalPlus);
+
+  EXPECT_FALSE(cache_->GetAudioConfiguration(2, &config));
+}
+
```
jasonzhangxx, line 235: Will there be another pr for tests of `HasAudio/VideoDecoderFor` ands `FindAudio/VideoDecoder`?

starboard/android/shared/media_capabilities_cache_test.cc
johnedocampo: Done. Added additional ClearCache tests for other functions.

starboard/android/shared/media_capabilities_cache_test.cc
johnedocampo: Yes, we can add another PR to test these functions.

starboard/android/shared/media_capabilities_cache_test.cc
johnedocampo: Done. Other eligible tests that use the DisabledCache have also been given this adjustment.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -420,83 +487,44 @@ std::string MediaCapabilitiesCache::FindVideoDecoder(
...
-    video_codec_capabilities_map_.clear();
-    audio_configurations_.clear();
-    is_widevine_supported_ = GetIsWidevineSupported();
-    is_cbcs_supported_ = GetIsCbcsSupported();
-    supported_transfer_ids_ = GetSupportedHdrTypes();
-    LoadCodecInfos_Locked();
-    LoadAudioConfigurations_Locked();
+  if (!capabilities_is_dirty_.exchange(false)) {
+    return;
   }
```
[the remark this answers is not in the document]
johnedocampo: We've changed the implementation so that if the cache is not dirty, we return immediately. The code should be fine as is.

starboard/android/shared/media_capabilities_cache.h
johnedocampo: Done. This function now takes in two references to the capability containers.

starboard/android/shared/media_capabilities_cache.h
johnedocampo: Done. All providers have been merged into one provider.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -420,83 +487,44 @@ std::string MediaCapabilitiesCache::FindVideoDecoder(
...
+  passthrough_supportabilities_.clear();
+
+  audio_codec_capabilities_map_.clear();
+  video_codec_capabilities_map_.clear();
+  audio_configurations_.clear();
+  is_widevine_supported_ = media_capabilities_provider_->IsWidevineSupported();
+  is_cbcs_supported_ = media_capabilities_provider_->IsCbcsSchemeSupported();
+  supported_transfer_ids_ =
+      media_capabilities_provider_->GetSupportedHdrTypes();
+  LoadCodecInfos_Locked();
```
kjyoun: Why not just calling `  media_capabilities_provider_->GetCodecCapabilities(
      audio_codec_capabilities_map_, video_codec_capabilities_map_);` here?

The only things that `LoadCodecInfos_Locked` do are `SB_CHECK`, but which are not necessary if we call `GetCodecCapabilities` here, since the code is clearing `audio_codec_capabilities_map_` and `video_codec_capabilities_map_` in line 498 and 499

starboard/android/shared/media_capabilities_cache_test.cc
jasonzhangxx: As ClearCache() is supposed to clear all cached data in on call, can we combine all ClearCache* tests into one? That makes the test case closer to our use case.

And after clear cache, we can change the return value to ensure it doesn't use the value from previous call,  like what we did for cache disabled test cases.

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -420,83 +487,44 @@ std::string MediaCapabilitiesCache::FindVideoDecoder(
...
+  LoadAudioConfigurations_Locked();
```
kjyoun: do we need a separate `LoadAudioConfigurations_Locked` here?

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -169,14 +195,13 @@ class MediaCapabilitiesCache {
 
   std::mutex mutex_;
 
+  // Provider for abstracting data sources.
```
kjyoun: // Provider for abstracting data sources. This is non-null. ?

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -115,10 +115,35 @@ class VideoCodecCapability : public CodecCapability {
   const Range supported_frame_rates_;
 };
 
+typedef std::vector<std::unique_ptr<AudioCodecCapability>>
```
jasonzhangxx: Let's move the typedef into MediaCapabilitiesProvider and MediaCapabilitiesCache class.

starboard/android/shared/media_capabilities_cache.cc
johnedocampo: That makes sense. Done.

starboard/android/shared/media_capabilities_cache.h
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -115,10 +115,35 @@ class VideoCodecCapability : public CodecCapability {
...
+typedef std::vector<std::unique_ptr<VideoCodecCapability>>
+    VideoCodecCapabilities;
+
+class MediaCapabilitiesProvider {
+ public:
+  virtual ~MediaCapabilitiesProvider() = default;
+  virtual bool IsWidevineSupported() = 0;
+  virtual bool IsCbcsSchemeSupported() = 0;
+  virtual std::set<SbMediaTransferId> GetSupportedHdrTypes() = 0;
+  virtual bool GetIsPassthroughSupported(SbMediaAudioCodec codec) = 0;
```
kjyoun, line 124: IsPassthroughSupported ?

like IsWidevineSupported

johnedocampo: I think GetIsPassthroughSupported makes sense, as the function gets whether or not a specific `SbMediaAudioCodec` supports passthrough, and then stores that result in its own internal std::map (`passthrough_supportabilities_`). I think this pattern is more in line with `GetAudioConfiguration` and `GetCodecCapabilities`, and `GetSupportedHdrTypes`.

starboard/android/shared/media_capabilities_cache.h
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache_test.cc
johnedocampo: Done.

starboard/android/shared/media_capabilities_cache.h
johnedocampo: Following this logic, I actually think that `IsWidevineSupported` and `IsCbcsSupported` should be changed to `GetIsWidevineSupported` and `GetIsCbcsSupported` as these functions also get their respective values from an API call, and store that value. This also lines up with how these functions were named in the original code. WDYT?

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -0,0 +1,316 @@
...
+              GetAudioConfiguration,
+              (int index, SbMediaAudioConfiguration* configuration),
+              (override));
+  MOCK_METHOD(
+      void,
+      GetCodecCapabilities,
+      ((std::map<std::string, AudioCodecCapabilities>)&audio_codec_capabilities,
+       (std::map<std::string,
+                 VideoCodecCapabilities>)&video_codec_capabilities),
+      (override));
```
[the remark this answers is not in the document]
johnedocampo: This is not an issue.

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -158,25 +183,28 @@ class MediaCapabilitiesCache {
...
+  const std::unique_ptr<MediaCapabilitiesProvider> media_capabilities_provider_;
+
+  // Cached data.
   std::set<SbMediaTransferId> supported_transfer_ids_;
   std::map<SbMediaAudioCodec, bool> passthrough_supportabilities_;
 
   typedef std::vector<std::unique_ptr<AudioCodecCapability>>
       AudioCodecCapabilities;
   typedef std::vector<std::unique_ptr<VideoCodecCapability>>
       VideoCodecCapabilities;
```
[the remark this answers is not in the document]
johnedocampo: I think it makes sense for the typedefs to be defined twice, as this helps with readability. In the future, when other people read what these values are, it would be handy to have the typedefs right next to where these values are declared.

starboard/android/shared/media_capabilities_cache.h
kjyoun: `GetIsWidevineSupported` and `GetIsCbcsSupported` sound reasonable for consistency.

jasonzhangxx (approved):

PR #5538, opened 2025-04-23
[media] Reduce extra copying in MP4StreamParser

xiaomings:
Removes extra copying for codecs other than h264, hevc, dolby vision, and aac.  This is good enough for our use cases where the majority of large frames are in av1 and vp9 only.

This not only avoids redundant copying of all DecoderBuffers, but also reduces peak allocation on large key frames.

Note that the upstream has similar optimizations that apply to more codecs (e.g. aac) across multiple large PRs, e.g.:
  https://chromium-review.googlesource.com/c/chromium/src/+/5578211
  https://chromium-review.googlesource.com/c/chromium/src/+/5734508

This change works as good for our use cases without introducing the complexity of cherry-picking.

b/409347653

PR #7586, opened 2025-10-16
media: Fix check failures on AAC playback

borongc:
For AAC playbacks, chromium updates to use `heap_frame_buf` instead of `frame_buf`. As the AAC bitstream conversion is feed into `heap_frame_buf`, Cobalt's original implementation using `frame_buf` causing CHECK failures in `CHECK(frame_buf.empty() || heap_frame_buf.empty())` introduced by https://chromium-review.googlesource.com/c/chromium/src/+/5734508.

Issue: 452105968

### DIFF
--- a/media/formats/mp4/mp4_stream_parser.cc
+++ b/media/formats/mp4/mp4_stream_parser.cc
@@ -1085,9 +1085,6 @@ ParseResult MP4StreamParser::EnqueueSample(BufferQueueMap* buffers) {
 
   if (audio) {
     if (ESDescriptor::IsAAC(runs_->audio_description().esds.object_type)) {
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
-      frame_buf.assign(buf, buf + sample_size);
-#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 #if BUILDFLAG(USE_PROPRIETARY_CODECS)
       heap_frame_buf = PrepareAACBuffer(runs_->audio_description().esds.aac,
                                         {buf, buf + sample_size}, &subsamples);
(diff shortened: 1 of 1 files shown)


media/formats/mp4/mp4_stream_parser.cc
```diff
@@ -1133,9 +1130,15 @@ ParseResult MP4StreamParser::EnqueueSample(BufferQueueMap* buffers) {
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   if (frame_buf.empty()) {
-    stream_buf = StreamParserBuffer::CopyFrom(
-        base::span<const uint8_t>{buf, buf + sample_size}, is_keyframe,
-        buffer_type, runs_->track_id());
+    if (heap_frame_buf.empty()) {
```
kjyoun: How about refactoring this using the ternary operator?

This approach makes the data source selection (raw buffer vs. heap buffer) immediately visible, reduces duplication, and improves overall conciseness.

````
stream_buf = StreamParserBuffer::CopyFrom(
      // Selects the data span based on whether the heap buffer is empty. ?
      heap_frame_buf.empty()
          ? base::span<const uint8_t>{buf, buf + sample_size}
          : base::span<const uint8_t>{&heap_frame_buf[0], heap_frame_buf.size()},,
      is_keyframe, buffer_type, runs_->track_id());
```

borongc: Done.

media/formats/mp4/mp4_stream_parser.cc
```diff
@@ -1134,8 +1131,10 @@ ParseResult MP4StreamParser::EnqueueSample(BufferQueueMap* buffers) {
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   if (frame_buf.empty()) {
     stream_buf = StreamParserBuffer::CopyFrom(
-        base::span<const uint8_t>{buf, buf + sample_size}, is_keyframe,
-        buffer_type, runs_->track_id());
+        heap_frame_buf.empty()
+        ? base::span<const uint8_t>{buf, buf + sample_size}
+        : base::span<const uint8_t>{&heap_frame_buf[0], heap_frame_buf.size()},
+        is_keyframe, buffer_type, runs_->track_id());
```
jasonzhangxx: The chrobalt optimization was from #5538, which backported some changes in https://chromium-review.googlesource.com/c/chromium/src/+/5734508. After rebase, these changes are actually no longer needed. In new code, the copy happens at line 1157-1173, it copies data from either |frame_buf| or |heap_frame_buf|, as in some cases |frame_buf| is used.

borongc: Simplify the implementation.

PR #7588, opened 2025-10-17
This change is to fix the error in getting the socket address on IPv6 networks

Jimmy-ge:
This change is to fix the error in getting the socket address on IPv6 networks
Bug:401441408


lamdbui:
@jellefoks , I recall we discussed this was a reasonable change. Can you take a quick look?

jellefoks:
Thanks for submitting this. I used this fix in https://github.com/youtube/cobalt/pull/8053  after fixing the line formatting and adding tests that confirm that this fix works (the tests added fail without this fix).

lamdbui:
Thank you, @Jimmy-ge for your contribution!  Closing this as this was integrated into #8053 . Please try that when you get a chance.

PR #7590, opened 2025-10-17
content/compositor: Allow more Workers for raster/image decoding for Android

yell0wd0g:
For some reason Android (including AndroidTV) has a provision to
only allow 1 worker thread (which is used for e.g. decoding images,
uploading them to textures if needed, and other raster work). 
Kabuki likes to load carrousels of ~4 images at once, and having
only one such Worker forces sequential operations (see e.g. 
https://screenshot.googleplex.com/QZicmhnCUGbNhFP - concretely the
sequential 4 image decode + upload operations). 

This CL fixes that by not avoiding such provision, which puts us
in parity with other platforms. On Kirkwood (Google Streamer Pro)
this means 2 Workers (from the rather large pool) can be used and
I anecdotally saw a bit of a more responsive UI as navigating
along. I didn't try on a MiStick nor any other ultra-low-power
device though.

Bug: 424591313


yell0wd0g:
@sherryzy PTAL - I should try this on MiStick but I think it should
be safe since it's capped by the number of cores or 4, whichever is
smallest, divided by 2 🙃

sherryzy (approved):

yell0wd0g:
Tested this on Xiaomi Stick and nothing was out of line either to my
testing or in a tracing I took.

PR #7592, opened 2025-10-17
Refactor ShellContentBrowserClient into prod and test classes

haozheng-cobalt:
Move browser test-specific functionality from ShellContentBrowserClient into a new subclass, ShellContentBrowserTestClient.

This change isolates test-only code paths, such as Mojo interface binding for browser control, custom scheme handlers for tests, and field trial setup for testing, into a dedicated client. This reduces the complexity of the core ShellContentBrowserClient, removes all #if defined(RUN_BROWSER_TESTS) blocks, and clarifies the separation
between production and test-time browser client behavior.

Bug: 446892644

### DIFF
--- a/cobalt/shell/browser/shell_content_browser_client.h
+++ b/cobalt/shell/browser/shell_content_browser_client.h
@@ -38,6 +38,21 @@ namespace content {
 class ShellBrowserContext;
 class ShellBrowserMainParts;
 
+// In content browser tests we allow more than one ShellContentBrowserClient
+// to be created (actually, ContentBrowserTestContentBrowserClient). Any state
+// needed should be added here so that it's shared between the instances.
+struct SharedState {
+  SharedState();
+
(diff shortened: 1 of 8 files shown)


cobalt/shell/browser/shell_content_browser_client.h
```diff
@@ -38,6 +38,21 @@ namespace content {
 class ShellBrowserContext;
 class ShellBrowserMainParts;
 
+// In content browser tests we allow more than one ShellContentBrowserClient
+// to be created (actually, ContentBrowserTestContentBrowserClient). Any state
+// needed should be added here so that it's shared between the instances.
+struct SharedState {
```
johnxwork, line 44: This should be an implementation detail and not exposed in the header.

haozheng-cobalt: The structure is needed in the inheriting test class, so moving it to the header.

PR #7594, opened 2025-10-17
Copy base from m114 into //copied_base

andrewsavage1:
m114 //base is needed for C++17 compilation with the starboard and native toolchains for crashpad.

Bug: 452357142


madhurajayaraman (approved):
yikes

PR #7606, opened 2025-10-17
[Rebase M138] fix Android build for base_unittests

sideb0ard:
Bug: 418842688

Fix error:
In file included from ../../base/sys_byteorder.h:16:                                                                                                                                                                                         
../../base/numerics/byte_conversions.h:9:15: error: token is not a valid binary operator in a preprocessor subexpression                                                                                                                     
    9 | #if !BUILDFLAG(IS_COBALT) || defined(SB_IS_DEFAULT_TC)                                                                                                                                                                               
      |     ~~~~~~~~~~^                                                                                                                                                                                                                      
1 error generated.

### DIFF
--- a/base/numerics/byte_conversions.h
+++ b/base/numerics/byte_conversions.h
@@ -6,9 +6,6 @@
 #define BASE_NUMERICS_BYTE_CONVERSIONS_H_
 
 #include <array>
-#if !BUILDFLAG(IS_COBALT) || defined(SB_IS_DEFAULT_TC)
-#include <bit>
-#endif
 #include <cstdint>
 #include <cstring>
 #include <span>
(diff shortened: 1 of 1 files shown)


base/numerics/byte_conversions.h
```diff
@@ -5,6 +5,8 @@
 #ifndef BASE_NUMERICS_BYTE_CONVERSIONS_H_
 #define BASE_NUMERICS_BYTE_CONVERSIONS_H_
 
+#include "build/build_config.h"
+
 #include <array>
 #if !BUILDFLAG(IS_COBALT) || defined(SB_IS_DEFAULT_TC)
```
andrewsavage1: Let's actually move the conditional include below the existing build_config include, usually the conditional includes go after others

andrewsavage1 (approved):

base/numerics/byte_conversions.h
sideb0ard: cools, will do

PR #7621, opened 2025-10-18
poc: Fire memory pressure signal 1 min after video playback.

kjyoun:
This is to see how memory foorprint responds when we trigger low memory pressure signal

This PR is intentially wait 1 min before triggering the event, to avoid noise of many event happening at the beginning of the playbackk.

Bug: 1234

### DIFF
--- a/cc/resources/resource_pool.cc
+++ b/cc/resources/resource_pool.cc
@@ -17,6 +17,7 @@
 #include "base/containers/contains.h"
 #include "base/format_macros.h"
 #include "base/functional/bind.h"
+#include "base/logging.h"
 #include "base/ranges/algorithm.h"
 #include "base/strings/stringprintf.h"
 #include "base/task/single_thread_task_runner.h"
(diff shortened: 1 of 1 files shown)


starboard/android/shared/media_decoder.cc
```diff
@@ -316,7 +316,21 @@ void MediaCodecDecoder::DecoderThreadFunc() {
...
+    int64_t started_us = CurrentMonotonicTime();
+    bool fired = false;
+
     while (!destroying_.load()) {
+      if (!fired) {
+        using base::MemoryPressureListener;
+        // Fire memory pressure signal 60 sec after playback starts.
+        if (int64_t elapsed_us = CurrentMonotonicTime() - started_us;
+            elapsed_us > 60'000'000) {
+          SB_LOG(INFO) << " Fire memory pressure signal";
```
yell0wd0g, line 328: QQ: why not fire right away?

kjyoun: This is to run an  experiment to see whether firing memory pressure signal actually reduce memory footprint.

Hence, I want to fire a single while memory footprint is in steady state, immediately after playback starts, when many things happens simultaneously.

BTW, the result is no apparent change in memory footprint.

PR #7634, opened 2025-10-20
Work around arm64 secondary ABI build issues

kaidokert:
Disables secondary ABI as this is not used by Cobalt build configs.
It causes a broken include chain for `is_official_build:True`

Bug: 418842688

### DIFF
--- a/build/config/android/abi.gni
+++ b/build/config/android/abi.gni
@@ -30,7 +30,7 @@ declare_args() {
   # needed only for release builds or for webview testing. Slows down "gn gen"
   # and ninja parse time due to having to write rules for most native targets
   # a second time. Applicable only when target_cpu is 64-bit.
-  enable_android_secondary_abi = is_official_build
+  enable_android_secondary_abi = is_official_build && !is_cobalt
 }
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,3 +23,4 @@ Android WebView, Ash). Even if these products have multiple executables, the
 code should be in subdirectories of the product.
 
 If you found a bug, please file it at https://crbug.com/new.
+FOO
```
oxve: Add an L to make line 25 a little more Mr T

oxve (approved):

sideb0ard (approved):

build/config/android/abi.gni
```diff
@@ -30,7 +30,7 @@ declare_args() {
   # needed only for release builds or for webview testing. Slows down "gn gen"
   # and ninja parse time due to having to write rules for most native targets
   # a second time. Applicable only when target_cpu is 64-bit.
-  enable_android_secondary_abi = is_official_build
+  enable_android_secondary_abi = is_official_build && !is_cobalt
```
andrewsavage1, line 33: This should be set in common.gn instead

PR #7636, opened 2025-10-20
DO NOT SUBMIT - [Rebase M138] Fix allocator shim DEPS issue

sideb0ard:
Bug: 418842688

TEST - UNSURE IF THIS IS THE CORRECT SOLUTION

 Fix GN dependency errors for android-arm64 builds

  Problem:
  The android-arm64 build fails during gn gen with multiple "Can't include this header from here" errors related to partition_alloc headers. This issue does not occur on the android-arm build.

  Root Cause:
  The issue stems from V8's pointer compression (v8_enable_pointer_compression), which is enabled by default on 64-bit architectures like arm64. This feature creates a dependency on partition_alloc headers that is
   not present in the 32-bit arm build.

  The dependency chain was broken in two places for the arm64 configuration:
   1. The //base component's dependency on the allocator shim is conditional on the use_allocator_shim variable, which was defaulting to false.
   2. Even with that dependency enabled, the allocator_shim component itself had a private (deps) dependency on allocator_core, which prevented the necessary headers from being visible to upstream targets like //v8
      and //gin.

  Solution:
  This was resolved with two targeted changes:
   1. In cobalt/build/configs/android-arm64/args.gn, use_allocator_shim is explicitly set to true. This ensures //base correctly depends on the allocator shim.
   2. In base/allocator/partition_allocator/src/partition_alloc/BUILD.gn, the dependency on :allocator_core within the allocator_shim component was changed from deps to public_deps. This correctly propagates the
      header visibility to all dependent targets.

  These changes align the android-arm64 build's dependency structure with its feature set, resolving the GN errors.

### DIFF
--- a/cobalt/build/configs/android-arm64/args.gn
+++ b/cobalt/build/configs/android-arm64/args.gn
@@ -1,2 +1,4 @@
 import("//cobalt/build/configs/android_common.gn")
 import("//cobalt/build/configs/chromium_android-arm64/args.gn")
+
+use_allocator_shim = true
--- a/base/allocator/partition_allocator/src/partition_alloc/BUILD.gn
+++ b/base/allocator/partition_allocator/src/partition_alloc/BUILD.gn
@@ -951,7 +951,7 @@ if (is_clang_or_gcc) {
       sources += [ "shim/empty.cc" ]
     }
 
-    deps += [
+    public_deps = [
       ":allocator_base",
       ":allocator_core",
       ":buildflags",


cobalt/build/configs/android-arm64/args.gn
```diff
@@ -1,2 +1,4 @@
 import("//cobalt/build/configs/android_common.gn")
 import("//cobalt/build/configs/chromium_android-arm64/args.gn")
+
+use_allocator_shim = true
```
kaidokert, line 4: Would we have any side effects from this ? Here's the documentation
```
  # Causes all the allocations to be routed via allocator_shim.cc. Usually,
  # the allocator shim will, in turn, route them to PartitionAlloc, but
  # other allocators are also supported by the allocator shim.
  use_allocator_shim = use_allocator_shim_default && is_clang_or_gcc
```

PR #7650, opened 2025-10-21
Display network connection dialog for all cases of no/weak connectivity

ckim1346:
Changing the network connection check logic from using `DidFinishNavigation` to a network check using `https://www.google.com/generate_204` ([Chromium Network Portal Detection](https://www.chromium.org/chromium-os/chromiumos-design-docs/network-portal-detection/#:~:text=This%20determination%20of%20being%20in%20a%20captive,by%20attempting%20to%20retrieve%20the%20webpage%20http://clients3.google.com/generate_204.)). This way we don't have to wait (30+ seconds) for a navigation to finish before checking for connectivity in the weak connectivity case before raising the error dialog, thus avoiding a black screen. This now happens `onResume()` so both on startup and resume.

Bug: 394176365

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
@@ -78,7 +78,6 @@ public class PlatformError
             showDialogOnUiThread();
           }
         });
-    // return dialog;
   }
 
   public void setResponse(@Response int response) {


johnxwork (approved):

yell0wd0g:
If anybody is reading -- I just ran gemini code review to see whether it would catch
any bug(s) introduced by this CL and it seems like it found something! (I didn't add
this info in the gemini query below because I'm not sure it wouldn't confuse it).

andrewsavage1:
@ckim1346 could you CP this to 26.eap as well?

ckim1346:
> If anybody is reading -- I just ran gemini code review to see whether it would catch any bug(s) introduced by this CL and it seems like it found something! (I didn't add this info in the gemini query below because I'm not sure it wouldn't confuse it).

Funnily enough, these are all things that were addressed in follow-up PR's

ckim1346:
> @ckim1346 could you CP this to 26.eap as well?

Sure thing, although I'm still working on a few edits to the network dialog to make sure it's fully functional and working as intended. I'll cherry pick all of those changes over once things are finalized.

PR #7654, opened 2025-10-21
Use head commit in build action for correct git history

oxve:
By default the merge commit is checked out which masks the original last docker change.

Bug: 376324653

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -48,7 +48,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 1
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yaml
```diff
@@ -48,7 +48,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 1
+          fetch-depth: 0
```
isarkis, line 51: Do we really need to fetch all commits here?  This will break cobalt_sandbox as repository is considerably larger there.  See https://github.com/youtube/cobalt/pull/6819

oxve:
Recreated PR here: https://github.com/youtube/cobalt/pull/7768

PR #7655, opened 2025-10-21
[Rebase m138] Fix for wm_unittests on evergreen

madhurajayaraman:
Move IsScaleFactorSupported to resource_scale_factor.h | https://chromium-review.googlesource.com/c/chromium/src/+/4523179

### DIFF
--- a/ui/wm/core/transient_window_manager_unittest.cc
+++ b/ui/wm/core/transient_window_manager_unittest.cc
@@ -103,7 +103,7 @@ class TransientWindowManagerTest : public aura::test::AuraTestBase {
   }
 };
 
-#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if GTEST_HAS_DEATH_TEST
 // Tests that creating a transient tree with a cycle in it will crash on a
 // CHECK. See a crash that can happen if we allow cycles http://b/286947509.
 TEST_F(TransientWindowManagerTest, TransientCycle) {


ui/wm/core/transient_window_manager_unittest.cc
```diff
@@ -103,6 +103,7 @@ class TransientWindowManagerTest : public aura::test::AuraTestBase {
   }
 };
 
+#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
```
andrewsavage1: Use `#if GTEST_HAS_DEATH_TEST` directly (we can upstream that)

ui/wm/test/starboard/run_all_unittests.cc
```diff
@@ -47,7 +47,7 @@ class WMTestSuite : public base::TestSuite {
     ASSERT_TRUE(base::PathService::Get(ui::UI_TEST_PAK, &ui_test_pak_path));
     ui::ResourceBundle::InitSharedInstanceWithPakPath(ui_test_pak_path);
 
-    if (ui::ResourceBundle::IsScaleFactorSupported(ui::k200Percent)) {
+    if (ui::IsScaleFactorSupported(ui::k200Percent)) {
```
andrewsavage1, line 50: Ensure this setup is similar to the upstream wm run_all_unittests.cc (i.e. changes they've made there are similar to ones made here)

andrewsavage1 (approved):

ui/wm/test/starboard/run_all_unittests.cc
madhurajayaraman: Yeah that's the PR linked in the description, they moved IsScaleFactorSupported one level up so we have to as well

andrewsavage1: Yep makes sense, just wanted to check that that was all they changed (and same with the other run_all_unittests files in the future). Thanks!

PR #7662, opened 2025-10-21
[Rebase m138] Build net_unittests for evergreen

madhurajayaraman:


kaidokert (approved):

cobalt/build/testing/targets/evergreen-x64/test_targets.json
oxve, line 70: Also add "out/evergreen-x64_devel/net_unittests_loader.py" here

madhurajayaraman: It doesn't run, so I've left it in the list of failing targets on top. I'll add the loader when it stops crashing.

PR #7674, opened 2025-10-22
[Rebase m138] gpu_perftests and nplb for evergreen

madhurajayaraman:
Bug: 454142578


third_party/musl/src/exit/assert.c
```diff
@@ -1,5 +1,7 @@
 #include <stdio.h>
 #include <stdlib.h>
+typedef struct _IO_FILE {};
```
andrewsavage1: Revert these change—we'll use https://github.com/youtube/cobalt/pull/7684 instead

andrewsavage1 (approved):
lgtm

PR #7678, opened 2025-10-22
Change getFeature to async

yjzhang111:
Bug: 450080869

getFeature needs mojo call to access the browser process, therefore changing its definition to async. getFeatureParam can access all the params within the renderer process, so removing the mojo calls and implementations to keep it synchronous.


third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-experiments/resources/mock-h5vcc-experiments.js
```diff
@@ -55,15 +55,19 @@ class MockH5vccExperiments {
...
 
   getConfigHash() {
     return this.STUB_KEY_CONFIG_CONFIG_HASH;
   }
 
-  getFeature(feature_name) {
-    throw new Error('Sync methods not supported in MojoJS (b/406809316');
+  async getFeature(feature_name) {
+    return {
+      feature_value: this.stub_result_.get(feature_name)
```
hlwarriner: This is very subtle but in the JavaScript bindings, the `feature_value` response value from the Mojo interface becomes camel case. So this line needs to be changed to `featureValue: this.stub_result_.get(feature_name)`.

This was not easy to debug...maybe the Mojo framework can be improved to provide some error messaging when an unexpected response parameter is set. I can look into that.

hlwarriner: Because this feature value response value that is set here will be evaluated by the real `H5vccExperiments::OnGetFeature()` method, it needs to be of type `OverrideState` from the Mojo definition.

If there's a way for the web test to create a value of this type and call ` mockH5vccExperiments.stubResult()` with it, that would be great.

Alternatively, I would recommend something like this (I verified this works, locally):
* Also import `OverrideState` in the mock Mojo impl, when `H5vccExperiments` and `  H5vccExperimentsReceiver` are imported
* Add a dedicated stub method for this, e.g., `mockH5vccExperiments.stubGetFeatureResult()`, that converts a true boolean to `OverrideState.OVERRIDE_ENABLE_FEATURE` and a false boolean to `OverrideState.OVERRIDE_DISABLE_FEATURE`. This would be intuitive enough I think.
* The web test would for example call `mockH5vccExperiments.stubGetFeatureResult(true)` and then expect the web API to return "ENABLED" (from the corresponding Blink enum).

yjzhang111: Thanks for helping to debug this!

yjzhang111: Updated to import OverrideState in the test

third_party/blink/renderer/modules/cobalt/h5vcc_experiments/h_5_vcc_experiments.cc
```diff
@@ -219,6 +218,26 @@ void H5vccExperiments::OnResetExperimentState(
...
+    case h5vcc_experiments::mojom::blink::OverrideState::
+        OVERRIDE_ENABLE_FEATURE:
+      resolver->Resolve(V8OverrideState(V8OverrideState::Enum::kENABLED));
+      return;
+    case h5vcc_experiments::mojom::blink::OverrideState::
+        OVERRIDE_DISABLE_FEATURE:
+      resolver->Resolve(V8OverrideState(V8OverrideState::Enum::kDISABLED));
+      return;
+  }
+  NOTREACHED() << "Invalid feature OverrideState for feature";
```
andrewsavage1, line 238: Sholud this call `resolver->Reject` before calling NOTREACHED?

andrewsavage1 (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_experiments/h_5_vcc_experiments.cc
yjzhang111: Is it necessary since NOTREACHED is a fatal call?

andrewsavage1: I had thought NOTREACHED was not fatal in official builds, but rereading the code I believe you're right

PR #7679, opened 2025-10-22
[Rebase M138] - Fix Evergreen arm64 vp9 compilation issues

sideb0ard:
Bug: 418842688

Needs tidy up but testing this #vibecoded solution.

### DIFF
--- a/base/allocator/partition_allocator/src/partition_alloc/aarch64_support.h
+++ b/base/allocator/partition_allocator/src/partition_alloc/aarch64_support.h
@@ -14,9 +14,11 @@
 #define HAS_HW_CAPS
 #endif
 
-#if PA_BUILDFLAG(PA_ARCH_CPU_ARM64) && defined(HAS_HW_CAPS)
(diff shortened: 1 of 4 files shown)


starboard/evergreen/arm64/platform_configuration/BUILD.gn
```diff
@@ -14,7 +14,7 @@
 
 config("sabi_flags") {
   cflags = [
-    "-march=arm64v8a",
+    "-march=armv9-a+sve+sme+dotprod+i8mm",
```
andrewsavage1, line 17: This is a fix for arm64, please separate it into a different PR

base/allocator/partition_allocator/src/partition_alloc/tagging.cc
andrewsavage1, line 1: use `has_memory_tagging = false` instead

base/allocator/partition_allocator/src/partition_alloc/aarch64_support.h
andrewsavage1, line 13: Revert the other changes here and just ensure this define isn't hit

PR #7681, opened 2025-10-22
Turn on AutoDetect for network status

johnxwork:
We were creating the NetworkChangeMonitor on Android but not using the AutoDetect feature. This resulted in `navigator.onLine` always returning meaningless true.

Bug: 378718120


Awallky (commented):
IIUC, all of the removed code is useless right now, since it always returned true.
Question: Does the auto detect feature fix `navigator.onLine` or do we need a follow-up to fix this?

johnxwork:
> IIUC, all of the removed code is useless right now, since it always returned true. Question: Does the auto detect feature fix `navigator.onLine` or do we need a follow-up to fix this?

The online in activity class to turn on AutoDetect is all we need to fix the `onLine` property.

Awallky (approved):

starboard/android/shared/BUILD.gn
```diff
@@ -260,7 +260,6 @@ static_library("starboard_platform") {
     "system_has_capability.cc",
     "system_info_api.cc",
     "system_info_api.h",
-    "system_network_is_disconnected.cc",
```
zhongqiliang, line 263: should this file be deleted?

johnxwork: Removing, thanks

PR #7689, opened 2025-10-22
Disable plugins, pdf and printing

y4vor:
Use the gn variables to disable the
plugins, pdf and printing and reduce the
binary size with ~ 16KB on arm.

Issue: 421171908


niranjanyardi (approved):

andrewsavage1:
Once this PR passes checks I'd be fine with it

PR #7692, opened 2025-10-22
cobalt: Enable cache age limit to reduce memory usage

sherryzy:
Enables the `kLimitImageDecodeCacheAge` feature by default and reduces the `kCacheAgeLimitSeconds` from 10 to 5 seconds. This allows the GPU image decode cache to be purged more frequently based on age, in addition to its size limit.

This change significantly reduces memory consumption. Experimental results show the persistent cache memory usage is reduced from up to 25MB down to approximately 8MB during fast video navigation.

By observing the total PSS given by "dumpsys meminfo", I feel this change can hopefully reduce average PSS by 20 - 30 MB when testing a devel build on Sabrina.

Bug: 424591313

### DIFF
--- a/cc/tiles/gpu_image_decode_cache.cc
+++ b/cc/tiles/gpu_image_decode_cache.cc
@@ -99,12 +99,12 @@ constexpr base::FeatureParam<int> kCacheSizeLimitMb{&kLimitImageDecodeCacheSize,
 // kCacheAgeLimitSeconds parameter defined below.
 BASE_FEATURE(kLimitImageDecodeCacheAge,
              "LimitImageDecodeCacheAge",
-             base::FEATURE_DISABLED_BY_DEFAULT);
+             base::FEATURE_ENABLED_BY_DEFAULT);
 
 // The max number of seconds since an LRUCache entry has been used before we
 // will try to evict it from the cache.
 constexpr base::FeatureParam<int> kCacheAgeLimitSeconds{
-    &kLimitImageDecodeCacheAge, "seconds", 10};
+    &kLimitImageDecodeCacheAge, "seconds", 5};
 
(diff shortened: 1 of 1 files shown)


yell0wd0g (approved):
Niiice LGTM!

Would you have to merge this to 114 and 138?
(And then to 26.android and 27.android...?
So many branches).

sherryzy:
@yell0wd0g Yes, I guess so. And in M138 the code changes a little bit, and there is a feature features::EnablePurgeGpuImageDecodeCache which is still disabled by default. Will test this after the rebase work is finished.

johnxwork:
Do we want to use command line switches instead if that's more manageable?https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java;l=94

Awallky:
> Do we want to use command line switches instead if that's more manageable?

+1
See `LimitImageDecodeCacheSize` in that class for an example of how to change the flag param value. See also - [link](go/finch-command-line-flags#configuring-basefeatureparam-state).
If we add things to the CLI helper class, please update the expectations in the test class as well - `cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java`.
Test Target is `cobalt_coat_junit_tests` and the test command is `out/android-arm_qa/bin/run_cobalt_coat_junit_tests`.
I am hoping we can get JUnit/RoboElectric tests supported in our CI eventually.

sherryzy:
Because of the M138 rebase work, I'll split this PR to 2: 1) apply this change to Android.26 in https://github.com/youtube/cobalt/pull/7733; 2) apply similar change in M138 branch for both 3p and Android.27.

PR #7693, opened 2025-10-22
Add wrappers for missing rdk functions

andrewsavage1:
Bug: 450956860

### DIFF
--- a/third_party/musl/src/starboard/largefile.c
+++ b/third_party/musl/src/starboard/largefile.c
@@ -28,16 +28,16 @@ int fstat64(int fildes, struct stat* info) {
 
 int open64(const char *filename, int flags, ...)
 {
-	mode_t mode = 0;
+  mode_t mode = 0;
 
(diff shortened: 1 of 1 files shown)


starboard/shared/modular/cobalt_layer_posix_stat_abi_wrappers.cc
```diff
@@ -22,6 +22,12 @@ int fstat(int fildes, struct stat* info) {
   return __abi_wrap_fstat(fildes, info);
 }
 
+int __abi_wrap_fstatat(int fd, const char* path, struct stat* buf, int flag);
```
jellefoks: If we add `fstatat()` to Starboard, we should implement `stat`, `lstat`, and `fstat` in Cobalt's libc/musl with calls to `fstatat()` to reduce the Starboard API surface. From a quick look it appears use musl (f.e. third_party/musl/src/stat/fstat.c)

y4vor (approved):

PR #7695, opened 2025-10-22
Fix musl includes

jellefoks:
This fixes the musl includes to be in the correct order, and declared in a single location.

This fixes unintentional leaks to use files from /usr/include and the build sysroot in Cobalt code.

Fixed: 454425109

### DIFF
--- a/build/config/c++/BUILD.gn
+++ b/build/config/c++/BUILD.gn
@@ -32,6 +32,7 @@ config("runtime_library") {
     }
   }
 
+  # This include path is needed for |__config_site| in that folder.
   include_dirs += [ "//buildtools/third_party/libc++" ]
 
   # libc++ has two levels of additional checking:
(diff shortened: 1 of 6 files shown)


jellefoks:
I think this interesting build failure is due to the ongoing m138 rebase causing issues in the CI

```
2025-10-22T23:48:03.1075493Z In file included from ../../third_party/libvpx/source/libvpx/vp9/encoder/arm/neon/vp9_error_sve.c:17:
2025-10-22T23:48:03.1076899Z ../../third_party/libvpx/source/libvpx/vpx_dsp/arm/vpx_neon_sve_bridge.h:30:10: error: SVE vector type 'svuint64_t' (aka '__SVUint64_t') cannot be used in a target without sve
```

Especially since there is no `vpx_neon_sve_bridge.h`...

```
$ find third_party/libvpx -name vpx_neon_sve_bridge.h
(no output)
```

build/config/c++/BUILD.gn
```diff
@@ -32,6 +32,7 @@ config("runtime_library") {
     }
   }
 
+  # This include path is needed for |__config_site| in that folder.
```
y4vor, line 35: why do we need this comment? Isn't this just upstream code unrelated to musl?

third_party/musl/BUILD.gn
```diff
@@ -22,19 +22,42 @@ if (target_cpu == "x64") {
   musl_arch = "aarch64"
 }
 
-config("external_settings") {
-  include_dirs = [
-    "arch/$musl_arch",
-    "arch/generic",
+config("system_includes") {
+  libmusl_prefix = "//third_party/musl"
```
y4vor, line 26: do we need the variable? I would assume the following works:
rebase_path("//third_party/musl/src/starboard/include",
                              root_build_dir)

build/config/c++/BUILD.gn
jellefoks: This is the c++ build config, so unrelated to musl, but this folder has no include files except for the one mystery file  "__config_site" So it was surprising to me to see it was added as an include folder, and it took me an iteration of removing it and building to discover that the folder actually was used... Added this comment to reduce future confusion.

third_party/musl/BUILD.gn
jellefoks: Yes, it works, but it's used 4 times for isystem incantations below, so I find this more readable.

PR #7700, opened 2025-10-23
Implement h5vcc system api

niranjanyardi:
Issue: 445433945 , 395650827

Implement h5vcc system tracking authorization

This change implements the tracking authorization status and request
functionality for the h5vcc system API. It integrates with the Starboard
IFA extension to retrieve the current authorization status and to trigger
user prompts for authorization, leveraging IFA API version 2.

The implementation includes updating the Blink renderer module to manage
ongoing asynchronous requests, ensuring proper handling of Mojo connection
errors. Starboard Linux shared components also receive stub
implementations for the new IFA extension functions to facilitate platform
integration. This addresses long-standing TODOs related to tracking
authorization.

### DIFF
--- a/cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
+++ b/cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
@@ -139,10 +139,34 @@ H5vccSystemImpl::H5vccSystemImpl(
     : content::DocumentService<mojom::H5vccSystem>(render_frame_host,
                                                    std::move(receiver)) {
   DETACH_FROM_THREAD(thread_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  const StarboardExtensionIfaApi* ifa_api =
+      static_cast<const StarboardExtensionIfaApi*>(
+          SbSystemGetExtension(kStarboardExtensionIfaName));
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
+  if (is_ifa_version_supported) {
+    ifa_api->RegisterTrackingAuthorizationCallback(this, [](void* context) {
+      DCHECK(context) << "Callback called with NULL context";
+      if (context) {
+        static_cast<H5vccSystemImpl*>(context)
+            ->ReceiveTrackingAuthorizationComplete();
+      }
+    });
+  }
+#endif
 }
 
 H5vccSystemImpl::~H5vccSystemImpl() {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  const StarboardExtensionIfaApi* ifa_api =
+      static_cast<const StarboardExtensionIfaApi*>(
+          SbSystemGetExtension(kStarboardExtensionIfaName));
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
+  if (is_ifa_version_supported) {
+    ifa_api->UnregisterTrackingAuthorizationCallback();
+  }
+#endif
 }
 
 void H5vccSystemImpl::Create(
@@ -200,6 +224,11 @@ void H5vccSystemImpl::RequestTrackingAuthorization(
   ifa_api->RequestTrackingAuthorization();
   std::move(callback).Run();
 #endif
+}
+
+void H5vccSystemImpl::ReceiveTrackingAuthorizationComplete() {
+  // TODO - b/457866495: Implement after studying Cobalt 25 implementation
+  NOTIMPLEMENTED();
 }
 
 void H5vccSystemImpl::GetUserOnExitStrategy(
--- a/cobalt/browser/h5vcc_system/h5vcc_system_impl.h
+++ b/cobalt/browser/h5vcc_system/h5vcc_system_impl.h
@@ -58,6 +58,7 @@ class H5vccSystemImpl : public content::DocumentService<mojom::H5vccSystem> {
   H5vccSystemImpl(content::RenderFrameHost& render_frame_host,
                   mojo::PendingReceiver<mojom::H5vccSystem> receiver);
   ~H5vccSystemImpl();
+  void ReceiveTrackingAuthorizationComplete();
 
   THREAD_CHECKER(thread_checker_);
 };


cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -89,8 +89,16 @@ bool GetLimitAdTrackingShared() {
 }
 
 std::string GetTrackingAuthorizationStatusShared() {
-  // TODO - b/395650827: Connect to Starboard extension.
-  NOTIMPLEMENTED();
+  const starboard::extension::IfaApi* ifa_api =
+      static_cast<const starboard::extension::IfaApi*>(
+          SbSystemGetExtension(starboard::extension::kIfaApiName));
+  if (ifa_api && ifa_api->version >= 2 &&
+      ifa_api->GetTrackingAuthorizationStatus) {
```
yell0wd0g: Can we have early return? E.g.
```
  const bool is_ifa_good = ifa_api && 
      ifa_api->version >= 2 &&
      ifa_api->GetTrackingAuthorizationStatus;
  if (!is_ifa_good) {
    return "NOT_SUPPORTED";
  }
``` 
etc ?

yell0wd0g: Can we have early return? E.g.
```
  const bool is_ifa_good = ifa_api && 
      ifa_api->version >= 2 &&
      ifa_api->GetTrackingAuthorizationStatus;
  if (!is_ifa_good) {
    return "NOT_SUPPORTED";
  }
``` 
etc ?

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -89,8 +89,16 @@ bool GetLimitAdTrackingShared() {
...
+    char status[128];
```
yell0wd0g: Can we change/overload `GetTrackingAuthorizationStatus()` to use
std::string? Otherwise it's a better practice to not have C arrays,
e.g. in this case:
```
  std::string status;
  status.reserve(128);
  if (ifa_api->GetTrackingAuthorizationStatus(status.data(), status.capacity())) {
    return status;
  }
```
?
Or just `std::array<char, 128> status`.

yell0wd0g (approved):
I have two comments but otherwise looks good? (Sorry
I didn't notice it was put to draft again).

Also consider running `/generate-commit-message` to
auto populate the commit/PR/CL message,

starboard/linux/shared/ifa.cc
andrewsavage1: I imagine you want to update this to 2

third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc
```diff
@@ -34,6 +34,7 @@ ScriptPromise H5vccSystem::getAdvertisingId(ScriptState* script_state,
 
   EnsureReceiverIsBound();
 
+  ongoing_requests_.insert(resolver);
```
niranjanyardi, line 37: these changes will affect android as well

should i cp it to 26.android ? ( if they are approved)

third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc
```diff
@@ -123,7 +129,7 @@ ScriptPromise H5vccSystem::requestTrackingAuthorization(
 
 void H5vccSystem::OnRequestTrackingAuthorization(
     ScriptPromiseResolver* resolver) {
-  // TODO - b/395650827: Reject when this fails.
```
niranjanyardi, line 126: maybe i should keep this todo ?

ill need to change the function signature to accept something if it's supposed to know when the underlying callback fails

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -91,9 +92,21 @@ bool GetLimitAdTrackingShared() {
 }
 
 std::string GetTrackingAuthorizationStatusShared() {
-  // TODO - b/395650827: Connect to Starboard extension.
-  NOTIMPLEMENTED();
-  return "NOT_SUPPORTED";
+#if BUILDFLAG(IS_STARBOARD)
```
niranjanyardi, line 95: should i add BUILDFLAG(IS_APPLETV) ?

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -174,10 +211,24 @@ void H5vccSystemImpl::GetTrackingAuthorizationStatusSync(
...
+    return;
+  }
+  ifa_api->RequestTrackingAuthorization();
   std::move(callback).Run();
+#endif
+}
+
+void H5vccSystemImpl::ReceiveTrackingAuthorizationComplete() {
+  // TODO - b/457866495: Implement after studying Cobalt 25 implementation
+  NOTIMPLEMENTED();
```
niranjanyardi, line 231: didnt get too deep into this, can implement it in this PR if requested.

C25 implementation looked non-trivial

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
niranjanyardi: fixed

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
niranjanyardi: i used std::vector<char> , hopefully that's okay

starboard/linux/shared/ifa.cc
niranjanyardi: done

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
yell0wd0g: In general, yes, just leave in this file a
```
#if BUILDFLAG(IS_STARBOARD) 
  // Do stuff
#elif BUILDFLAG(IS_IOS_TVOS)
  NOTIMPLEMENTED(); // or #error 
#endif
```

This is because the tvOS implementation won't need to go
via Starboard, and is unclear to me currently whether 
Igalians will decide to implement the Mojo interface
directly in Obj-C (case in which these two files won't even 
be compiled).

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -91,9 +92,21 @@ bool GetLimitAdTrackingShared() {
...
-  return "NOT_SUPPORTED";
+#if BUILDFLAG(IS_STARBOARD)
+  const StarboardExtensionIfaApi* ifa_api =
+      static_cast<const StarboardExtensionIfaApi*>(
+          SbSystemGetExtension(kStarboardExtensionIfaName));
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
+  if (!is_ifa_version_supported) {
+    return "NOT_SUPPORTED";
+  }
+  std::vector<char> status_buffer(128);
```
yell0wd0g, line 103: std::vector<char> is strange - how about:
```
  std::string status;
  status.resize(128);
  if (ifa_api->GetTrackingAuthorizationStatus(status.data(),
                                              status.size())) {
    return status;
  }
```
?

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -126,10 +139,34 @@ H5vccSystemImpl::H5vccSystemImpl(
...
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
+  if (is_ifa_version_supported) {
+    ifa_api->RegisterTrackingAuthorizationCallback(this, [](void* context) {
+      DCHECK(context) << "Callback called with NULL context";
+      if (context) {
+        static_cast<H5vccSystemImpl*>(context)
+            ->ReceiveTrackingAuthorizationComplete();
+      }
+    });
+  }
```
yell0wd0g, line 155: nit: Prefer early return:
```
  if (!is_ifa_version_supported) {
    return;
  }
  // The rest.
```

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -126,10 +139,34 @@ H5vccSystemImpl::H5vccSystemImpl(
...
   DETACH_FROM_THREAD(thread_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  const StarboardExtensionIfaApi* ifa_api =
+      static_cast<const StarboardExtensionIfaApi*>(
+          SbSystemGetExtension(kStarboardExtensionIfaName));
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
+  if (is_ifa_version_supported) {
+    ifa_api->RegisterTrackingAuthorizationCallback(this, [](void* context) {
+      DCHECK(context) << "Callback called with NULL context";
+      if (context) {
```
yell0wd0g, line 150: In general, it's uncommon to find a DCHECK followed by 
handling the condition: DCHECK should always verify, so
why test it in l.150? Consider leaving a CHECK() and 
assuming it's not null in l.151.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -174,10 +211,24 @@ void H5vccSystemImpl::GetTrackingAuthorizationStatusSync(
...
 void H5vccSystemImpl::RequestTrackingAuthorization(
     RequestTrackingAuthorizationCallback callback) {
+#if BUILDFLAG(IS_STARBOARD)
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  // TODO - b/395650827: Connect to Starboard extension.
-  NOTIMPLEMENTED();
+  const StarboardExtensionIfaApi* ifa_api =
+      static_cast<const StarboardExtensionIfaApi*>(
+          SbSystemGetExtension(kStarboardExtensionIfaName));
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
```
yell0wd0g, line 219: QQ: Should the number `2`s in this file and ifa.cc be a constant?

third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.h
```diff
@@ -66,6 +66,11 @@ class MODULES_EXPORT H5vccSystem final
   void EnsureReceiverIsBound();
   HeapMojoRemote<h5vcc_system::mojom::blink::H5vccSystem> remote_h5vcc_system_;
 
+  // Holds promises associated with outstanding async remote_h5vcc_system_
```
yell0wd0g, line 69: nit: `|remote_h5vcc_system_|` or `` `remote_h5vcc_system_` `` : escape variable
names in comments.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -126,10 +139,34 @@ H5vccSystemImpl::H5vccSystemImpl(
     : content::DocumentService<mojom::H5vccSystem>(render_frame_host,
                                                    std::move(receiver)) {
   DETACH_FROM_THREAD(thread_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  const StarboardExtensionIfaApi* ifa_api =
+      static_cast<const StarboardExtensionIfaApi*>(
+          SbSystemGetExtension(kStarboardExtensionIfaName));
+  const bool is_ifa_version_supported = ifa_api && ifa_api->version >= 2;
+  if (is_ifa_version_supported) {
```
yell0wd0g, line 147: Question: Is this `ifa_api` changing during the browser lifetime?
If not, could we just keep it in a member variable, and assume that
the version is the appropriate one if it's not null?

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -174,10 +211,24 @@ void H5vccSystemImpl::GetTrackingAuthorizationStatusSync(
 
 void H5vccSystemImpl::RequestTrackingAuthorization(
     RequestTrackingAuthorizationCallback callback) {
+#if BUILDFLAG(IS_STARBOARD)
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  // TODO - b/395650827: Connect to Starboard extension.
-  NOTIMPLEMENTED();
+  const StarboardExtensionIfaApi* ifa_api =
```
hlwarriner, line 216: This is a higher level comment that probably shouldn't impact this PR, but...I'm not sure if we need to keep this callback-based implementation in chrobalt long-term.

If I understand the c25 code approach correctly, `H5vccSystem.requestTrackingAuthorization()` works like this:
* The web API returns a void promise, which is fulfilled if/when a user makes a choice in the prompt
* After the promise is fulfilled, the web app can query `H5vccSystem.trackingAuthorizationStatus` to get the actual value
* The web API implementation, in `cobalt/h5vcc/h5vcc_system.cc`, registers a callback with the platform, through the Starboard extension, that will be called if/when the user completes the prompt. This is how the web API impl knows if/when it should fulfill the promise.

It seems like it would be simpler / more direct if in chrobalt we changed the `requestTrackingAuthorization` web API so that it returns a promise of the actual authorization status value. In the Starboard extension,  `RequestTrackingAuthorization()` would just be a synchronous function that returns the result from the prompt, and the Mojo framework would automatically give us the asynchronous effect.

The approach we've taken migrating web APIs from c25 to chrobalt is to minimize interface changes, so I think we can table this question for a future discussion and keep the `H5vccSystem.requestTrackingAuthorization()` interface+implementation as similar to c25 as possible for now. But I did want to raise it.

Cc @yell0wd0g @andrewsavage1

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
niranjanyardi: qq: have i got the abstraction level correctly here ? I think i'm wrong 

Atleast in 25.lts.1+ seems like there 's a [separate vector tacking the promises ](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/h5vcc/h5vcc_system.h;l=76 ) resolved and unresolved so maybe all these 3 functions should be present in third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc ?

By 3 methods I mean - UnregisterTrackingAuthorizationCallback , RegisterTrackingAuthorizationCallback and
ReceiveTrackingAuthorizationComplete

niranjanyardi: @andrewsavage1 / @hlwarriner

hlwarriner: Yeah, I think that's right.

If we want to keep the IFA Starboard extension unchanged from c25, it looks to me like we'd want to:
*  Add these three methods to the `H5vccSystem` mojo interface
* The `blink::H5vccSystem` implementation would be very similar to the `cobalt::h5vcc::H5vccSystem` implementation from c25, but `blink::H5vccSystem` would call the three new methods in the Mojo interface instead of directly calling the corresponding functions from the Starboard extension
* The implementations of these three methods in `h5vcc_system::H5vccSystemImpl` would just delegate to the functions in the Starboard extension.

But related to https://github.com/youtube/cobalt/pull/7700/files#r2495677251, I think there are two possible ways to simplify this even if we don't change the web interface.

**Keeping the Starboard extension unchanged**

The `H5vccSystemImpl::RequestTrackingAuthorization` implementation just calls the extension's `RegisterTrackingAuthorizationCallback` with its `RequestTrackingAuthorizationCallback callback` param, and then calls the extension's `RequestTrackingAuthorization`. Basically, the extension, instead of  `H5vccSystemImpl::RequestTrackingAuthorization`, would end up running this callback.

**Changing the Starboard extension**

The Mojo signature changes from `RequestTrackingAuthorization() => ()` to `RequestTrackingAuthorization() => (string authorization_status)`. The `H5vccSystemImpl::RequestTrackingAuthorization` implementation just synchronously calls a function in the extension that prompts the user and returns the value. Or, it could call `RequestTrackingAuthorization` from the extension - which would no longer need to execute a callback - and then call `GetTrackingAuthorizationStatus` from the extension. In either case, once  `H5vccSystemImpl::RequestTrackingAuthorization` has synchronously gotten the value, it runs its callback param.

Let me know what you think and if you want to talk through this more.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
niranjanyardi: Hmm, ill avoid adding the #elif BUILDFLAG(IS_IOS_TVOS) - they'll ahve to do it in all the API's like cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc

it's good if it's done in all places or not done at all.. wdyt ?

I didnt know if APPLETV was using starboard or not , and i know that the starboard ifa extension is used by appletv atleast in C25. If they won't use starboard , they'll have to figure out how to make it work for all API's not just this system api/ ifa api's

niranjanyardi:
@yell0wd0g / @hlwarriner  I just talked to @kaidokert and i think I'll get a PR up with a completely different implementation. I'll mostly abandon this PR as the changes will be very different based on this discussion:
https://docs.google.com/document/d/1LGHzgcTosnfd-MrExkeCTMGU2YrFfpC2OW_Xp_Kef4A/edit?disco=AAABtnoidU4

niranjanyardi:
This PR is deprecated, please review https://github.com/youtube/cobalt/pull/7928 instead.

@yell0wd0g - I'll try to be more proactive next time in asking for exact feedback. I acknowledge the time you spent writing this feedback - thank you! and sorry for the late response .

niranjanyardi:
The 'Bug ID Check' is failing because the PR title or description is missing a bug ID. Please add a bug ID (e.g., 'b/123456789') to the PR title or description to fix this. I am unable to do this automatically due to a tool error.

PR #7706, opened 2025-10-23
Remove root from ICU

haozheng-cobalt:
This change removes the "root" locale data from the ICU data bundle (icudtl.dat) generated for Cobalt. It updates the cobalt.json filter definition to no longer explicitly include "root" for various data trees (e.g., locale, currency, date, region, unit, zone). The translit data is also now explicitly excluded. 
Note root is not included in C25's ICU data.

Bug: 446966279


y4vor (approved):

haozheng-cobalt:
Cannot remove root from translit rule for it caused crash in TransliteratorTest.LatinASCIICorrect unit test in base.

PR #7711, opened 2025-10-23
Temporarily disable pre-commits for Cobalt and Starboard

sideb0ard:
Bug: 454687024

During the rebase to m138, we seem to have inadvertently pulled in different java code formatting rules, resulting in code being reformatted to 4 spaces, rather than 2. Temporarily disable for now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -77,7 +77,7 @@ repos:
         language: python
         types: [java]
         args: [-i]
-        # (TODO:b/418842688) Re-enable. Temporarily disabled post m138 merge
+        # (TODO:b/454687024) Re-enable. Temporarily disabled post m138 merge
         # due to unintended update of java code formatting rules
         exclude: |
             (?x)^(


.pre-commit-config.yaml
```diff
@@ -7,8 +7,8 @@ default_language_version:
 
 files: |
     (?x)^(
-        (.*/)?cobalt/.*|
-        (.*/)?starboard/.*|
+        # (.*/)?cobalt/.*|
```
andrewsavage1: Let's just do this for google-java-format

andrewsavage1 (changes requested):

.pre-commit-config.yaml
sideb0ard: done

.pre-commit-config.yaml
```diff
@@ -71,12 +71,12 @@ repos:
         entry: python3 ./cobalt/precommit/python3_check.py
         language: python
         types: [python]
-    -   id: google-java-format
-        name: google-java-format
-        entry: python3 ./cobalt/precommit/google_java_format_wrapper.py
-        language: python
-        types: [java]
-        args: [-i]
+    # -   id: google-java-format
```
andrewsavage1: I wouldn't comment this out, instead use the `exclude:` field for cobalt and starboard directories (see an example below in gcheckstyle)

.pre-commit-config.yaml
```diff
@@ -77,6 +77,11 @@ repos:
         language: python
         types: [java]
         args: [-i]
+        exclude: |
```
andrewsavage1, line 82: Maybe add a todo comment here, ideally with a bug number

sideb0ard: done.

andrewsavage1: This is not a good TODO bug—it should be about android style standards and re-enabling this check, not an umbrella ticket

PR #7714, opened 2025-10-23
cobalt/v8: Modify V8 Params to Improve Cold Start

Awallky:
Currently, Cobalt cold starts when navigating to deeplinks is quite slow.

This PR attempts to improve the state of things by doing two things:

* Match some C25 V8 tuning parameters
* Adding some new V8 tuning parameters

This adds the following new fields after consulting with the V8 team:

* max_old_space_size - 1024 MB

These changes may help improve cold start init without increasing Cobalt memory consumption. Local tests showed some improvement over an untuned deeplink navigation.

Test: out/android-arm_qa/bin/run_cobalt_coat_junit_tests
Bug: 454628424


Awallky:
@yell0wd0g It's not clear to me what we should be doing in regards to back-patching to M114 and 26.android. 
What might make sense is to make it land on M114, 26.android, 26.epa, and then a separate effort to land on M138. 

I don't think we have an M114 branch though. We just have a pre-138 tag. I think we need to make an M114 branch so people can make commits to some branch that easily cherrypicks to 26.android though.

yell0wd0g:
> @yell0wd0g It's not clear to me what we should be doing in regards to back-patching to M114 and 26.android. What might make sense is to make it land on M114, 26.android, 26.epa, and then a separate effort to land on M138.
> 
> I don't think we have an M114 branch though. We just have a pre-138 tag. I think we need to make an M114 branch so people can make commits to some branch that easily cherrypicks to 26.android though.

I think for this PR specifically we need to 
a) patch against M114, or the closest thing which I think is 26.android directly and 
b) patch against main (M138), which in some cases would mean partially/rewriting the contents of the PR/CL.

This is a temporary situation ofc.

yell0wd0g:
Since this PR/CL is based on an M138-chrobalt, and mimics what is done
on 26.android, land them in tandem (to avoid diverging the two branches).

Awallky:
@yell0wd0g 
> Since this PR/CL is based on an M138-chrobalt, and mimics what is done
on 26.android, land them in tandem (to avoid diverging the two branches).

It looks like these flags caused crashing in the 26.android branch. I don't think we can land the changes there.

PR #7723, opened 2025-10-24
Use DecoderBufferAllocator based on flag

kjyoun:
Previously, `DecoderBufferAllocator` is used if USE_STARBOARD_MEDIA is defined.

This PR is to provide an option of whether to use  `DecoderBufferAllocator` or not, when  USE_STARBOARD_MEDIA is defined.

Still we need `DecoderBufferAllocator` instance to get memory budget, but we don't use its main feature of allocate

### DIFF
--- a/media/base/decoder_buffer.cc
+++ b/media/base/decoder_buffer.cc
@@ -14,6 +14,9 @@ namespace media {
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 namespace {
 DecoderBuffer::Allocator* s_allocator = nullptr;
+
+// TODO(kjyoun): Connect this to a command line switch.
+constexpr bool kUseStarboardDecoderBufferAllocator = true;
 }  // namespace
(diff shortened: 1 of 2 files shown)


media/base/decoder_buffer.cc
```diff
@@ -14,6 +14,9 @@ namespace media {
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 namespace {
 DecoderBuffer::Allocator* s_allocator = nullptr;
+
+// TODO(kjyoun): Connect this to a command line switch.
+constexpr bool kUseStarboardDecoderBufferAllocator = true;
```
xiaomings, line 19: Since Finch hasn't been ready yet, we'll need something like https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/h5vcc/h5vcc_settings.cc#L55 to run experiments tentatively.

PR #7725, opened 2025-10-24
Fix Finch Safe Mode

jammel-yeboah:
Bug: 452149713

Safe mode is a crucial mechanism for Cobalt to fall back to a "safe" or "default" config if it receives a faulty config that causes a crash.

If we crash 3 times, we fall back to the safe config. If we crash 4 times, we fall back to the default, or "empty" config.

This PR addresses a bug where Cobalt's safe mode was never activating because the underlying crash streak was never being incremented after a crash.

The root cause was a failure to signal the start of a new session to the metrics system. The CleanExitBeacon mechanism, which is responsible for detecting unclean shutdowns, must first be set to an "unclean" state at startup. This was not
happening. Consequently, if a crash occurred, the beacon from the previous clean shutdown was still present. On the next startup, the system would incorrectly read this old beacon, conclude the last session was clean, and therefore fail to increment the crash streak at all.

Main fixes overview:

1. Enabling crash streak incrementing
    - Bug: We were never marking the CleanExitBeacon as "unclean" at startup. Without this, a crash was indistinguishable from a clean shutdown from the beacon's perspective.
    - Fix: Added a call to `metrics_state_manager_->LogHasSessionShutdownCleanly(false, false)` in `CobaltMetricsServiceClient::Initialize()`. This marks the session as unclean at startup. Now, only a graceful shutdown that
    calls `LogHasSessionShutdownCleanly(true, ...)` in the destructor can mark it as clean.
2. Reading crash streak from the right place
    - Bug: With the crash streak now being correctly incremented in `metrics_local_state_`, there was another bug: `ExperimentConfigManager` was reading the streak from the wrong file (experiment_config_).
    - Fix: Updated `ExperimentConfigManager` to read from `metrics_local_state_.`
3. Caveat: Synchronizing `Variations` Beacon and `Metrics Config` files
    - Problem: The system uses two files: the Variations beacon at startup and the Metrics Config at runtime. A race condition existed where resetting the streak to 0 after a config fetch would update the Metrics Config file
    but not the Variations beacon. A subsequent crash would cause the stale, high value from the beacon to be used.
    - Fix: After resetting the streak to 0 in `metrics_local_state_`,  explicitly call `clean_exit_beacon()->WriteBeaconValue(true)`. This forces the beacon to re-read the now-zeroed streak from `metrics_local_state_` and immediately write it to the
    Variations file, ensuring both files are synchronized.
4. Caveat: Ensuring the crash streak reset is not lost on a crash
    - Problem: PrefService writes are batched and only guaranteed on a clean shutdown. If we reset the crash streak to 0 and then crashed, that reset would be lost.
    - Fix: Explicitly call `CommitPendingWrite()` on `metrics_local_state_` after setting the crash streak to 0. This forces an immediate write to disk, making the reset durable.
    
PR tested by YTS safe mode tests here: https://critique.corp.google.com/cl/825319049

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -70,8 +70,6 @@ public final class CommandLineOverrideHelper {
         paramOverrides.add("--disable-accelerated-video-encode");
         // Rasterize Tiles directly to GPU memory.
         paramOverrides.add("--enable-zero-copy");
-        // Disable mipmap generation to reduce memory usage.
-        paramOverrides.add("--disable-mipmap-generation");
 
         return paramOverrides;
     }
--- a/cobalt/app/cobalt_switch_defaults_starboard.cc
+++ b/cobalt/app/cobalt_switch_defaults_starboard.cc
@@ -69,8 +69,6 @@ static constexpr auto kCobaltToggleSwitches = std::to_array<const char*>({
     // Cobalt doesn't use Chrome's accelerated video decoding/encoding.
     switches::kDisableAcceleratedVideoDecode,
     switches::kDisableAcceleratedVideoEncode,
-    // Disable mipmap generation to reduce memory usage.
-    switches::kDisableMipmapGeneration,
 });
 
 // Map of switches with parameters and their defaults.
--- a/media/audio/ios/audio_session_manager_ios.mm
+++ b/media/audio/ios/audio_session_manager_ios.mm
@@ -29,13 +29,24 @@ AudioSessionManagerIOS::AudioSessionManagerIOS() {
   AVAudioSession* audio_session = [AVAudioSession sharedInstance];
 
   NSError* error = nil;
-  auto options = AVAudioSessionCategoryOptionAllowBluetooth |
-                 AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+  AVAudioSessionCategoryOptions options;
 #if !BUILDFLAG(IS_IOS_TVOS)
-                 AVAudioSessionCategoryOptionDefaultToSpeaker |
+  options = AVAudioSessionCategoryOptionAllowBluetooth |
+            AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+            AVAudioSessionCategoryOptionDefaultToSpeaker |
+            AVAudioSessionCategoryOptionMixWithOthers;
+#else
+  options = 0;
 #endif
-                 AVAudioSessionCategoryOptionMixWithOthers;
-  [audio_session setCategory:AVAudioSessionCategoryPlayAndRecord
+
+  AVAudioSessionCategory category;
+#if !BUILDFLAG(IS_IOS_TVOS)
+  category = AVAudioSessionCategoryPlayAndRecord;
+#else
+  category = AVAudioSessionCategoryPlayback;
+#endif
+
+  [audio_session setCategory:category
                         mode:AVAudioSessionModeDefault
                      options:options
                        error:&error];
@@ -85,16 +96,13 @@ AudioSessionManagerIOS::AudioSessionManagerIOS() {
   // Find the desired audio output device
   AVAudioSessionRouteDescription* currentRoute = [audio_session currentRoute];
   if ([currentRoute.outputs count] > 0) {
+#if !BUILDFLAG(IS_IOS_TVOS)
     AVAudioSessionPortDescription* output = currentRoute.outputs.firstObject;
     if ([output.portType isEqualToString:AVAudioSessionPortBuiltInReceiver] ||
         [output.portType isEqualToString:AVAudioSessionPortBuiltInSpeaker]) {
-      auto option =
-#if !BUILDFLAG(IS_IOS_TVOS)
-          AVAudioSessionPortOverrideSpeaker;
-#else
-          AVAudioSessionPortOverrideNone;
-#endif
-      [audio_session overrideOutputAudioPort:option error:&error];
+      error = nil;
+      [audio_session overrideOutputAudioPort:AVAudioSessionPortOverrideSpeaker
+                                       error:&error];
       if (error) {
         NSLog(@"Error overriding output audio port: %@",
               [error localizedDescription]);
@@ -102,6 +110,7 @@ AudioSessionManagerIOS::AudioSessionManagerIOS() {
         NSLog(@"Set default output device to Speaker");
       }
     } else {
+      error = nil;
       [audio_session overrideOutputAudioPort:AVAudioSessionPortOverrideNone
                                        error:&error];
       if (error) {
@@ -111,6 +120,12 @@ AudioSessionManagerIOS::AudioSessionManagerIOS() {
         NSLog(@"Using System chosen default audio output device");
       }
     }
+#else
+    // tvOS manages audio output routing at the system level and does not
+    // support port overrides. Hardware devices use HDMIOutput as the single
+    // available route, while the simulator uses BuiltInSpeaker.
+    NSLog(@"Using System chosen default audio output device");
+#endif
   }
 
   [audio_session setActive:YES error:nil];
--- a/media/filters/source_buffer_stream.cc
+++ b/media/filters/source_buffer_stream.cc
@@ -1872,14 +1872,6 @@ bool SourceBufferStream::UpdateVideoConfig(const VideoDecoderConfig& config,
     }
   }
 
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
-  // Dynamically increase |memory_limit_| when video resolution goes up.
-  memory_limit_ = std::max(
-      memory_limit_,
-      GetDemuxerStreamVideoMemoryLimit(DemuxerType::kChunkDemuxer,
-                                       &config));
-#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
-
   return true;
 }
 
--- a/services/webnn/BUILD.gn
+++ b/services/webnn/BUILD.gn
@@ -190,7 +190,7 @@ component("webnn_service") {
     deps += [ "//third_party/onnxruntime_headers" ]
   }
 
-  if (is_mac || webnn_use_tflite) {
+  if (is_apple || webnn_use_tflite) {
     sources += [
       "queueable_resource_state.h",
       "queueable_resource_state_base.cc",


cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -80,17 +82,26 @@ bool HasConfigExpired(PrefService* experiment_prefs) {
 
 }  // namespace
 
+ExperimentConfigManager::ExperimentConfigManager(
```
jammel-yeboah, line 83: removed the old inline constructor and created this new constructor to add the second `PrefService* metrics_local_state` parameter so that the class could access the crash streak

cobalt/browser/global_features.cc
```diff
@@ -158,6 +158,11 @@ void GlobalFeatures::InitializeActiveExperimentIds() {
   }
 }
 
+void GlobalFeatures::Shutdown() {
```
jammel-yeboah, line 158: Explicitly destroys the metrics manager. This is then hooked into the application's main lifecycle by calling it from [CobaltBrowserMainParts::PostMainMessageLoopRun()](https://github.com/youtube/cobalt/pull/7725/files#diff-bb37c95a0cf66c08afd410e6adc7e216d34b20fe76c4983c90a2f0b5199843aeR53)

cobalt/browser/metrics/cobalt_metrics_service_client.h
```diff
@@ -57,7 +57,7 @@ constexpr base::TimeDelta kMinIdleRefreshInterval = base::Seconds(30);
 // much.
 class CobaltMetricsServiceClient : public metrics::MetricsServiceClient {
  public:
-  ~CobaltMetricsServiceClient() override = default;
+  ~CobaltMetricsServiceClient() override;
```
jammel-yeboah: removed `default` so we can create a custom destructor which contains clean up code, ie `metrics_state_manager_->LogHasSessionShutdownCleanly(true, false)`

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -80,17 +82,26 @@ bool HasConfigExpired(PrefService* experiment_prefs) {
...
-  int num_crashes =
-      experiment_config_->GetInteger(variations::prefs::kVariationsCrashStreak);
+  int num_crashes = metrics_local_state_->GetInteger(
+      variations::prefs::kVariationsCrashStreak);
   static_assert(
       kCrashStreakEmptyConfigThreshold > kCrashStreakSafeConfigThreshold,
       "Threshold to use an empty experiment config should be larger "
       "than to use the safe one.");
-
+  std::cout << "num_crashes" << num_crashes << std::endl;
```
AlecsisDuarte: Are these debugging logs still needed? If so, could you make them a bit more descriptive? (e.g., `std::cout << "Number of crashes in streak = " << num_crashes << std::endl`). And same with other `std::cout` statements.

cobalt/browser/global_features.cc
AlecsisDuarte: I think it would be good idea to surface this as a comment to make it clear for future maintainers

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -80,17 +82,26 @@ bool HasConfigExpired(PrefService* experiment_prefs) {
...
+    : experiment_config_(experiment_config),
+      metrics_local_state_(metrics_local_state) {
+  DCHECK(experiment_config_);
+  DCHECK(metrics_local_state_);
+}
+
 ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
   // First, determine the config type based on the crash streak.
-  DCHECK(experiment_config_);
+  DCHECK(metrics_local_state_);
```
AlecsisDuarte, line 94: Curious, why are we using `DCHECK` here? Is it because this is a performance intensive? If not, you could probably use `CHECK` instead!

See go/totw/labs/avoid-dcheck

cobalt/browser/experiments/experiment_config_manager.h
```diff
@@ -30,14 +30,10 @@ enum class ExperimentConfigType {
 // This class manages the content of the experiment config stored on disk.
```
AlecsisDuarte, line 30: Nit: Probably worth updating this class' description since it is now doing extra work as tracking both experiment config and the state of the system (Crashes)

cobalt/browser/global_features.cc
```diff
@@ -42,8 +42,8 @@ GlobalFeatures::GlobalFeatures() {
   CreateMetricsServices();
   // InitializeActiveExperimentIds needs ExperimentConfigManager to determine
   // the experiment config type.
-  experiment_config_manager_ =
-      std::make_unique<ExperimentConfigManager>(experiment_config_.get());
+  experiment_config_manager_ = std::make_unique<ExperimentConfigManager>(
```
AlecsisDuarte, line 45: super-nit: Since this is only used for initializing the `activeExperimentIds` consider moving it to the `InitializeActiveExperimentIds` function to keep all relevant steps around initializing experiments all together.

cobalt/browser/experiments/experiment_config_manager.cc
jammel-yeboah: They are just for debugging -- I'll remove them!

cobalt/browser/experiments/experiment_config_manager.cc
jammel-yeboah: Mostly just following existing patterns (like `DCHECK(experiment_config_);`) but I'll go ahead and change that

AlecsisDuarte: SGTM! I'm also not that experienced on this repo so I'm okay with going with the existing approach .

cobalt/browser/experiments/experiment_config_manager.h
jammel-yeboah: So the class hasn't functionally changed. I discovered that it was checking the crash streak in the wrong file `experiment_config`, instead of `metrics_local_state` (see `ExperimentConfigManager` params). The class itself functionally remains the same 🙂

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "cobalt/browser/cobalt_content_browser_client.h"
 
+#include <iostream>
```
yjzhang111: Remove..?

cobalt/browser/global_features.cc
yjzhang111: I believe setExperimentState also calls this config manager. I think it's ok either way (since this is a member variable) as long as the GlobalFeatures ctor initializes it

cobalt/browser/global_features.cc
yjzhang111: Agreed. Maybe add a comment explaining why we need to explicitly destroy it.

cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
```diff
@@ -94,6 +94,7 @@ void H5vccExperimentsImpl::SetExperimentState(
   // CommitPendingWrite not called here to avoid excessive disk writes.
   // Features and featureParams won't be applied until the next Cobalt cold
   // start so the delay is acceptable.
+  global_features->metrics_local_state()->CommitPendingWrite();
```
yjzhang111, line 119: Is there a particular reason we need to call CommitPendingWrite here?

cobalt/browser/metrics/cobalt_metrics_services_manager_client.h
```diff
@@ -69,7 +69,11 @@ class CobaltMetricsServicesManagerClient
 
   ::metrics::MetricsStateManager* GetMetricsStateManager() override;
 
-  void ClearMetricsServiceClientRawPtrForTest() { metrics_service_client_ = nullptr; }
+  void ClearMetricsServiceClient() { metrics_service_client_ = nullptr; }
+
+  void ClearMetricsServiceClientRawPtrForTest() {
```
yjzhang111, line 74: What exactly is `ClearMetricsServiceClientRawPtrForTest` for? Are you planning to add tests in this PR?

jammel-yeboah: I was using this for debugging. Thanks for pointing it out, I'll remove it

cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
jammel-yeboah: (resolved in person)

TLDR: Without `CommitPendingWrite()`, we never write to disk if we shutdown too early.

For other reviewers: see [b/456583508](https://b.corp.google.com/issues/456583508)

yjzhang111:
Instead of explaining what safe mode is / how safe mode works, could you explain in the PR description about what the fix is about. e.g. experiment config didn't get updated when crashing so Cobalt never enters safe mode. This PR modifies experiment config manager to read from metrics local state instead which do get updated after a crash (on top of the beacon file).

cobalt/browser/global_features.cc
```diff
@@ -155,6 +155,17 @@ void GlobalFeatures::InitializeActiveConfigData() {
           : kExperimentConfigActiveConfigData);
 }
 
+void GlobalFeatures::Shutdown() {
+  // Explicitly shuts down the metrics services. This is to ensure the
+  // CobaltMetricsServiceClient destructor is called, which logs a clean
+  // shutdown. The specific shutdown order here is required to nullify
+  // a raw pointer and prevent a use-after-free crash
+  // that would otherwise occur on exit.
```
AlecsisDuarte: nit: Moving these function documentation on top of the function in the header makes it easier for a user of this API to understand what and when the shutdown is called instead of inside of it.

cobalt/browser/global_features.cc
```diff
@@ -155,6 +155,17 @@ void GlobalFeatures::InitializeActiveConfigData() {
           : kExperimentConfigActiveConfigData);
 }
 
+// Explicitly shuts down the metrics service. This is to ensure the
+// CobaltMetricsServiceClient destructor is called, which logs a clean
+// shutdown. The specific shutdown order here is required to nullify
+// a raw pointer and prevent a use-after-free crash
+// that would otherwise occur on exit.
```
AlecsisDuarte: I think these would be more helpful in the header for callers and IDEs making them available. See go/cstyle#function_declarations.

cobalt/browser/experiments/experiment_config_manager.cc
yjzhang111: CHECK might be a little too aggressive. We almost never want to intentionally crash production builds (with a few exceptions). DCHECK should be enough to catch problems in CI/early testing

cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
```diff
@@ -69,8 +73,26 @@ void H5vccExperimentsImpl::SetExperimentState(
...
   // successful config fetch do not prevent updating to a new update, and
   // therefore do not necessitate falling back to a safe config.
-  experiment_config_ptr->SetInteger(variations::prefs::kVariationsCrashStreak,
-                                    0);
+  global_features->metrics_local_state()->SetInteger(
+      variations::prefs::kVariationsCrashStreak, 0);
+  // At this point, have set the crash streak to 0 in the metrics_local_state
+  // file. Do the same for the Variations beacon file to keep them in sync by
+  // calling WriteBeaconValue(true), which leads to the call of
+  // WriteBeaconFile(). This function reads the 0 value we just set in
```
yjzhang111: Are you saying WriteBeaconValue reads from metrics_local_state? How so?

cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
```diff
@@ -69,8 +73,26 @@ void H5vccExperimentsImpl::SetExperimentState(
   // Note: It's important to clear the crash streak. Crashes that occur after a
   // successful config fetch do not prevent updating to a new update, and
   // therefore do not necessitate falling back to a safe config.
-  experiment_config_ptr->SetInteger(variations::prefs::kVariationsCrashStreak,
-                                    0);
+  global_features->metrics_local_state()->SetInteger(
+      variations::prefs::kVariationsCrashStreak, 0);
+  // At this point, have set the crash streak to 0 in the metrics_local_state
+  // file. Do the same for the Variations beacon file to keep them in sync by
+  // calling WriteBeaconValue(true), which leads to the call of
```
yjzhang111: nit: I think "which leads to the call of WriteBeaconFile()" is an implementation detail, you can leave it out.

cobalt/browser/global_features.cc
yjzhang111, line 1: Can you remove metrics::MetricsService::RegisterPrefs(registry); from GlobalFeatures::RegisterPrefs(PrefRegistrySimple* registry) now that we're not using metrics related stuff in experiment config? It's a one line change so you don't have to do it in a separate PR

cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
jammel-yeboah: Yes, but WriteBeaconValue itself doesn't directly; it calls WriteBeaconFile which does:

https://source.corp.google.com/h/github/youtube/cobalt/+/main:components/metrics/clean_exit_beacon.cc;l=434;bpv=1?q=CleanExitBeacon::WriteBeaconFile&sq=repo:github%2Fyoutube%2Fcobalt%20AND%20branch:main

cobalt/browser/global_features.cc
```diff
@@ -155,6 +155,12 @@ void GlobalFeatures::InitializeActiveConfigData() {
           : kExperimentConfigActiveConfigData);
 }
 
+void GlobalFeatures::Shutdown() {
+  metrics_services_manager_client_->ClearMetricsServiceClient();
```
yjzhang111: Instead of explicitly destroying metrics_services_manager_client_, can we call metrics_service()->LogCleanShutdown()? It does the same thing and we don't need to worry about managing the pointers

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -42,8 +42,15 @@ CobaltMetricsServiceClient::CobaltMetricsServiceClient(
...
 }
 
+CobaltMetricsServiceClient::~CobaltMetricsServiceClient() {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  metrics_state_manager_->LogHasSessionShutdownCleanly(true, false);
+}
+
 void CobaltMetricsServiceClient::Initialize() {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  metrics_state_manager_->LogHasSessionShutdownCleanly(false, false);
```
yjzhang111: Can we move this to where we setup field trials (right after InstantiateFieldTrialList) to prevent access before beacon file initialization?

cobalt/browser/global_features.cc
jammel-yeboah: `LogCleanShutdown()` only updates the clean exit beacon, it does not tear down the metrics service. Do you mind clarifying?

cobalt/browser/metrics/cobalt_metrics_service_client.cc
jammel-yeboah: Thanks for this, moving this line to `CobaltContentBrowserClient::CreateFeatureListAndFieldTrials` would guarantee the beacon file is initialized before we write to it. Made the change.

cobalt/browser/global_features.cc
yjzhang111: What do we need to explicitly tear down metrics service for? Please correct me if I'm wrong, I was under the impression that we just need to update the beacon file

jammel-yeboah: Oh I see... we tear down metrics service to ensure the `CobaltMetricsServiceClient` destructor is called (which contains the line to signal a clean shutdown to the beacon file).

Before adding this line ` metrics_services_manager_client_->ClearMetricsServiceClient();`, I was getting a dangling pointer error after the next line (which tears down metrics services) was executed:  `metrics_services_manager_.reset();.` 

From my understanding, the flow looks something like this:
- The `metrics_services_manager_` owns the `metrics_services_manager_client_`
- The `metrics_services_manager_client_` holds a raw pointer to the `CobaltMetricsServiceClient` instance.
- `metrics_services_manager_` was getting destroyed first, then the client ended up with a dangling pointer. When the client was eventually destroyed, accessing this dangling pointer was causing a crash.

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
```diff
@@ -212,6 +212,7 @@ class CobaltMetricsServiceClientTest : public ::testing::Test {
         &prefs_, enabled_state_provider_.get(), std::wstring(),
         temp_dir_.GetPath(), metrics::StartupVisibility::kForeground);
     ASSERT_THAT(metrics_state_manager_, NotNull());
+    metrics_state_manager_->clean_exit_beacon()->Initialize();
```
yjzhang111: Now that you moved things around a little bit, could you also try removing this and see if the tests still pass?

cobalt/browser/experiments/experiment_config_manager_unittest.cc
Awallky, line 35: IIUC this `pref_service` provides the finch experiment config, something we read once on startup, and then the `metrics_pref_service_` is something that manages the current state of metrics.

Can we add a comment to the declaration here to clarify that or change the names to match whatever is in the `ExperimentConfigManager` ctor? 

This would make things a bit easier to follow what their purpose is in these tests.

cobalt/browser/metrics/cobalt_metrics_services_manager_client_test.cc
```diff
@@ -66,6 +66,7 @@ class CobaltMetricsServicesManagerClientTest : public ::testing::Test {
 
     manager_client_ =
         std::make_unique<CobaltMetricsServicesManagerClient>(&prefs_);
+    manager_client_->GetMetricsStateManager()->InstantiateFieldTrialList();
```
Awallky: nit: to shorten the calls at various callsites for this, you could expose a public method of the same name that just calls  `manager_client_->GetMetricsStateManager()` underneath it.

yjzhang111: Agreed. Or just remove this now that you have LogHasSessionShutdownCleanly calls in the right places

Awallky (approved):
LGTM % comments

cobalt/browser/experiments/experiment_config_manager_unittest.cc
jammel-yeboah: Sounds good, I'll add that in

cobalt/browser/global_features.cc
yjzhang111: Discussed offline to remove the custom destructor.

PR #7729, opened 2025-10-24
media: Make DecoderBuffer use external allocator only when it's set.

kjyoun:
Previously, DecoderBufferAllocator is always used for memory allocation if USE_STARBOARD_MEDIA is defined.

With this PR, DecoderBuffer uses DecoderBufferAllocator only when it's set.

This PR alone does not change the behavior of `DecoderBuffer`'s allocation, since we always set `DecoderBufferAllocator`. #7747 will set `DecoderBufferAllocator` based on a run-time flag.

Bug: 454441375


media/base/decoder_buffer.h
```diff
@@ -202,29 +203,31 @@ class MEDIA_EXPORT DecoderBuffer
...
+      return allocator_data_->data;
+    }
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
     DCHECK(!end_of_stream());
     DCHECK(!read_only_mapping_.IsValid());
     DCHECK(!writable_mapping_.IsValid());
     DCHECK(!external_memory_);
     return data_.get();
-#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
   }
```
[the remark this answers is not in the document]
kjyoun: It's the logic of exising code. This PR is intended not to change the behavior existing code.

media/base/decoder_buffer.cc
```diff
@@ -147,11 +153,11 @@ void DecoderBuffer::Initialize(DemuxerStream::Type type) {
...
   int padding = s_allocator->GetBufferPadding();
-  allocated_size_ = size_ + padding;
-  data_ = static_cast<uint8_t*>(s_allocator->Allocate(type,
-                                                      allocated_size_,
-                                                      alignment));
-  memset(data_ + size_, 0, padding);
+  size_t allocated_size = size_ + padding;
+  allocator_data_.emplace(static_cast<uint8_t*>(s_allocator->Allocate(
+                            type, allocated_size, alignment)),
+                        allocated_size);
```
xiaomings: nit: alignment

media/base/decoder_buffer.cc
```diff
@@ -57,10 +56,10 @@ DecoderBuffer::DecoderBuffer(const uint8_t* data,
   Initialize();
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  memcpy(data_, data, size_);
-#else // BUILDFLAG(USE_STARBOARD_MEDIA)
+  memcpy(writable_data(), data, size_);
```
borongc, line 59: `writable_data()` is deprecating, we should avoid to use it.

borongc: m138 (main branch) has more refactoring in this part. This is fine to leave it in 26.android branch, but for main branch, we should avoid using it.

media/base/decoder_buffer.cc
```diff
@@ -18,7 +18,6 @@ DecoderBuffer::Allocator* s_allocator = nullptr;
 
 // static
 DecoderBuffer::Allocator* DecoderBuffer::Allocator::GetInstance() {
-  DCHECK(s_allocator);
```
borongc, line 21: Is there any reason we want to remove this DCHECK (or not convert it to CHECK)? This is independent from this PR.

media/base/decoder_buffer.cc
```diff
@@ -84,9 +83,12 @@ DecoderBuffer::DecoderBuffer(DemuxerStream::Type type,
     return;
   }
 
-  Initialize(type);
-
-  memcpy(data_, data, size_);
+  if (s_allocator) {
+    Initialize(type);
+  } else {
```
borongc, line 88: Now, without `s_allocator`, DecoderBuffer seems that it won't initialize neither `data_` or `allocator_data_`. Basically, it won't work if `s_allocator` is not set? If that's the case, should we log (or CHECK) it so it's easier for us to debug?

media/base/decoder_buffer.cc
kjyoun: Agree. when we have to cherry-pick the change to m138, I will change this part.
I hope that it happens.

For now, let's see whether this disabling DecoderBuffeAllocator can pass the experiment.
If it's doesn't pass, we don't have to cherry-pick  this to m138

I expect that I need to modify this change anyway since `DecoderBuffer` in m138 quite differs from the one in 26.android. (e.g. Constructure signature), if I cherry-picked to m138.

media/base/decoder_buffer.cc
kjyoun: With this change, DecoderBuffer can work without `s_allocator` .
If there is no `s_allocator`, it will use partition_alloc and then `GetInstance` returns nullptr.

media/base/decoder_buffer.cc
kjyoun: If s_allocator is set,  allocator_data_ is initialized and we use `DecoderBufferAllocator` as external allocator
If it is not set (== nullptr),  data_ is initialized and we use `partition_alloc` as default allocator (as Chromium does)

re: log, 
I did not add it, since existing `DecoderBuffer` doesn't have any log by itself and I interpreted "no log" as intention of this class.

Do you think that it's better with the log?

johnxwork (approved):

PR #7731, opened 2025-10-24
cobalt/v8: Modify V8 Params to Improve Cold Start

Awallky:
Currently, Cobalt cold starts when navigating to deeplinks is quite slow.

This PR attempts to improve the state of things by doing two things:

* Match some C25 V8 tuning parameters
* Adding some new V8 tuning parameters

This adds the following new fields after consulting with the V8 team:

* max_old_space_size - 1024 MB

These changes may help improve cold start init without increasing Cobalt memory consumption. Local tests showed some improvement over an untuned deeplink navigation. With some feedback from the V8 team, these parameters.
Note: The V8 team heavily discouraged use of `gc_interval` at all as it stresses CPU resources significantly.

Test: out/android-arm_qa/bin/run_cobalt_coat_junit_tests
Bug: 454628424


Awallky:
Yeah, this is causing crashes on 26.android. So, I don't think we should pursue this on this branch.

PR #7736, opened 2025-10-24
CP: Refactor MediaCapabilitiesCache for Tests (#7584)

sideb0ard:
Missed Cherry-pick of https://github.com/youtube/cobalt/pull/7584

The current implementation of `MediaCapabilitiesCache` makes it difficult to create isolated unit tests. This PR aims to refactor the `MediaCapabilitiesCache` implementation to allow for robust unit testing. This PR also adds some unittests for MediaCapabilitiesCache, verifying its behavior.

This is mainly achieved through dependency injection. Instead of `MediaCapabilitiesCache` directly calling other APIs (e.x. `MediaDrmBridge`, `AudioOutputManager`) to fill its cache, we instead inject provider methods that the cache will call. This enables us to inject our own mock methods for the cache, which helps create a well isolated testing environment.

The original logic of how the cache obtains its values is largely unchanged; the original code has been transported to provider classes that will be injected into `MediaCapabilitiesCache` in production.

This PR is based off of https://github.com/youtube/cobalt/pull/7564, which provided the main idea of how this implementation should look like.

Bug: 433773029

### DIFF
--- a/starboard/android/shared/media_capabilities_cache.cc
+++ b/starboard/android/shared/media_capabilities_cache.cc
@@ -23,6 +23,7 @@
 #include "base/android/jni_array.h"
 #include "base/android/jni_string.h"
 #include "base/android/scoped_java_ref.h"
+#include "cobalt/android/jni_headers/MediaCodecUtil_jni.h"
 #include "starboard/android/shared/audio_output_manager.h"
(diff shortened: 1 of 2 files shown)


johnedocampo (approved):
Thanks for the help Thor!

PR #7738, opened 2025-10-24
Cherry pick PR #7131: evergreen: more customizations to update_client

yuying-y:
This is a manual cherry-pick with no sustantial changes. Some legacy
functions and classes in m114 are ported over because the Cobalt
updater customizations depend on them. TODOs are filed and included in
the files. Follow-up PRs will resolve them, and this PR is kept as a
cherry-pick.

TODO(454962891): Investigate the upstream method to cancel a download
TODO(455004672): Investigate the legacy classes and functions from m114.

Original PR #7131 
These customizations support in-memory update and installation slot
manipulation. They mostly come from C25, with modifications to work with
the new codebase. Added a new CrxDownloaderFactoryCobalt class.

Issue: 431862767

### DIFF
--- a/components/update_client/component.cc
+++ b/components/update_client/component.cc
@@ -51,7 +51,53 @@
 #include "components/update_client/update_engine.h"
 #include "components/update_client/utils.h"
 
+#if BUILDFLAG(IS_STARBOARD)
+#include "components/update_client/crx_downloader_factory.h"
+#include "components/update_client/cobalt_slot_management.h"
+#endif
+
+// TODO(b/448186580): consider replacing all LOG with D(V)LOG.
+
 namespace update_client {
+
+// TODO(b/455004672): Investigate the legacy functions from m114. 
+// These functions are ported from m114 because Cobalt customizations
+// depend on them.
+#if defined(USE_M114_FUNCTIONS)
+namespace {
+
+using InstallOnBlockingTaskRunnerCompleteCallback = base::OnceCallback<
+    void(ErrorCategory error_category, int error_code, int extra_code1)>;
+
+void StartInstallOnBlockingTaskRunner(
+    scoped_refptr<base::SequencedTaskRunner> main_task_runner,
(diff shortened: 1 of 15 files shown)


components/update_client/component.cc
```diff
@@ -51,8 +51,327 @@
...
+#include "components/update_client/component_unpacker.h"
+#include "components/update_client/crx_downloader_factory.h"
+#include "components/update_client/cobalt_slot_management.h"
+#endif
+
+// TODO(b/448186580): consider replacing all LOG with D(V)LOG.
+
 namespace update_client {
 
+// TODO(b/455004672): Investigate the legacy functions from m114.
```
andrewsavage1, line 63: I'm fine with this temporarily, but think this is a relatively high-priority issue to follow up on as the functions that we should be calling in Chromium code that got refactored may also have functionality changes that we need.

Maybe you already know this, but it looks to me like a lot of the refactoring happened in and around https://chromium-review.googlesource.com/c/chromium/src/+/5797506, so that might be a good jumping off point for addressing

andrewsavage1 (commented):
Have you been able to test this locally?

yuying-y:
> Have you been able to test this locally?

This PR builds on evergreen-x64. The checks are running for the other platforms. I can't test the updater E2E yet, because of the big change with m138, tracked by b/455004672. For this PR, it's best to keep the change roughly consistent with the original PR #7131, because it's a cherry-pick PR and the original PR has been reviewed and merged.

components/update_client/component.cc
yuying-y: Yes, I'm afraid I'll have to get to this right after submitting this cherry-pick PR, in order to make the updater work again after the m138 rebase. Thanks for the pointer.

andrewsavage1: Okay, then makes sense to me to submit this if you're planning to follow up on it immediately after

PR #7748, opened 2025-10-27
media: Decouple decoder memory budget from DecoderBufferAllocator

kjyoun:
This PR changes `demuxer_memory_limit_starboard.cc` to get memory budget info directly from Starboard API, skipping `DecoderBufferAllocator`.

This is to decouple budget info from  `DecoderBufferAllocator`, so that Cobalt can get budget info, regardless of whether  `DecoderBufferAllocator` is used or not.

Bug: 454441375

### DIFF
--- a/media/starboard/decoder_buffer_memory_info.h
+++ b/media/starboard/decoder_buffer_memory_info.h
@@ -41,10 +41,10 @@ class StubDecoderBufferMemoryInfo : public DecoderBufferMemoryInfo {
 };
 
 // Returns the memory limit for audio decoder buffers in bytes.
-int GetDecoderAudioBufferLimitBytes();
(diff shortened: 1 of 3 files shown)


media/starboard/decoder_buffer_memory_info.h
```diff
@@ -38,6 +40,13 @@ class StubDecoderBufferMemoryInfo : public DecoderBufferMemoryInfo {
   size_t GetMaximumMemoryCapacity() const override { return 0; }
 };
 
+// Returns the memory limit for audio decoder buffers in bytes.
+int GetDecoderAudioBufferLimitBytes();
```
xiaomings: Maybe `GetAudioDecoderBufferLimitBytes()`, and the same for the one below?

PR #7661, opened 2025-10-21
Reapply "android: Replace JNI prefix flag with GN build flag" (#7631)

zhongqiliang:
This reverts commit 2dd294e.

Bug: 452059316
Change-Id: I5c62604c777d08c545b3ccd735e668a03f5dbdbc

PR #7754, opened 2025-10-27
Revert "Cherry pick PR #7661: Reapply "android: Replace JNI prefix flag with GN build flag" (#7631)"

johnxwork:
Reverts youtube/cobalt#7671

Reason 1: To check if this caused the startup memory increase we see on perfgate. This is one of two suspected PRs.
Reason 2: This PR seems to have caused b/455589818

Bug: 455589818


zhongqiliang:
Startup memory increase should be irrelevant.

I 'd prefer to fix the failing tests instead of revert since cl/824570121 is submitted.

PR #7755, opened 2025-10-27
Fix undefined zlib symbol for ARMv7 toolchains

hlwarriner:
Issue: 454950085


hlwarriner:
h/t @andrewsavage1 for identifying the root cause

PR #7756, opened 2025-10-27
Revert "Temporarily disable crashpad_handler build for RDK Kokoro"

andrewsavage1:
Reverts youtube/cobalt#7751

Bug: 454950085


hlwarriner:
With how we currently use `use_crashpad`, we don't have a way to build `crashpad_handler` without also using the real `crashpad_wrapper` library to install the signal handler at app startup. I'll test this on an RDK - hopefully tomorrow - just to make sure that doing so at least doesn't introduce a runtime crash.

PR #7762, opened 2025-10-28
Add docker service mapping for EG builds

oxve:
Bug: 448904392
Bug: 446227047


isarkis:
> Looks fine, but is there a way to see it working when still in a PR or does it have to be submitted for that?

There is way to see if it's working by running a presubmit in Kokoro via Gerrit.

PR #7767, opened 2025-10-28
cobalt/build: Disable building some TFLite bits and SLES

yell0wd0g:
This CL simply disables a few GN flags that reduce the binary size of Linux
builds by not building TFLite/WebNN related bits. Gold linux-x64x11 monolithic
before/after: 924MB vs 909MB.

Since I was in the area I also added a media override to not build SLES.

Either of those removed three symbols, so I in turn had to remove them from
the leak detector config file.

The line `enable_hls_demuxer=false` was tripping the style check so had to
fix it in this CL.

Bug: 455626752


yell0wd0g:
@y4vor some minor improvement, PTAL.

y4vor (approved):

starboard/tools/api_leak_detector/evergreen/manifest
andrewsavage1, line 1: If this is intentional, you'll likely need to update dev_manifest as well

yell0wd0g: I was planning on getting the builds green and then ask again
whether this reduction in leak was real or not. Thanks for 
the pointer!

yell0wd0g: Removed those 3 symbols from `evergreen/dev_manifest`. 
Let's see the colour of the CI!

PR #7768, opened 2025-10-28
Fix docker image caching

oxve:
Use sha256 checksum on Docker files as the image tag instead of commit commit hash. Also refactor to always rebuilding image on post submit builds.

Bug: 376324653


oxve:
If we are also able to move build id generation to the init step we can remove the deep checkout in the build step and replace it will a shallow one. With the cache though the time saved will probably be negligible.

.github/workflows/main.yaml
```diff
@@ -48,8 +48,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 1
-          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+          # Checkout commit directly instead of using merge commit to maintain correct history.
+          ref: ${{ (github.event_name == 'pull_request' && github.event.pull_request.head.sha) || github.sha }}
```
linxinan-chops: Do you know the local path of this checkout? ~/src? Will it impact the following steps?

.github/workflows/main.yaml
```diff
@@ -48,8 +48,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          fetch-depth: 1
-          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
```
isarkis: Do we really need to fetch all commits here? This will break cobalt_sandbox as repository is considerably larger there. See https://github.com/youtube/cobalt/pull/6819

isarkis (changes requested):

.github/workflows/main.yaml
oxve: I changed to use file sha checksums instead. It does not rely on git history.

.github/actions/docker/action.yaml
```diff
@@ -80,18 +81,15 @@ runs:
...
 
-        elif [[ "${IS_PR}" == "true" ]]; then
-          # No post-submit image was found. Check PR registry to see if there's an image there.
+        else
+          # No fitting post-submit image was found. Check PR registry to see if there's an image there.
           PROJECT_NAME=$(gcloud config get-value project)
           DOCKER_COMPOSE_IMAGE="${PR_DOCKER_REGISTRY_URL}/${PROJECT_NAME}/github/${DOCKER_SERVICE}"
 
           if ! docker manifest inspect "${DOCKER_COMPOSE_IMAGE}:${DOCKER_CONTENT_SHA}"; then
             echo "::warning::No pre-built docker image found for commit ${DOCKER_CONTENT_SHA}."
```
arjungm: nit: commit -> Docker content SHA256

isarkis (approved):

.github/actions/docker/action.yaml
oxve: Done

PR #7776, opened 2025-10-29
media,gpu: guard Android code depending on SDK > 24

yell0wd0g:
This CL/PR has three commits:

- Commit 1 works in //gpu/command_buffer/service: it avoids building
`ImageReaderGLOwner` when the SDK is <16, which is when many NDK methods
in that class were introduced. Current code should a) not be used on
Cobalt and b) have a fallback to `SurfaceTextureGLOwner`, which is
supported all right. //base function `EnableAndroidImageReader()` is
tweaked to always return false for Cobalt, for good measure.
 For future reference, much code was changed in this area when Chrome
deprecated SDKs before 26, see https://crrev.com/c/5177892 - I 
considered restating the code removed by that CL, but since this code
path (starting in //media/gpu/android/video_frame_factory_impl.*) 
should not be used in Cobalt anyway, the easiest path is to guard
the offending code from the build.

- Commit 2 works in //media/(audio|gpu) and avoids linking `aaudio`
when the SDK is < 26 since that's when said library was introduced.

- Commit 3 simply restates `default_min_sdk_version` to 24 for cobalt.

Tested via android-arm_qa local builds.

Bug: 451783049


yell0wd0g:
@sideb0ard  PTAL -- take a good look though, there were decisions re `ImageReaderGLOwner` that are not trivial : )

@johnxwork for RS/PTAL

sideb0ard (approved):
LGTM! It builds for me locally on arm and arm64, and code changes look clean.

PR #4244, opened 2024-10-09
[media] Check IAMF codec params for validity

osagie98:
1. Implements IamfMimeUtil to parse the "codecs" portion of an IAMF mime to check it for validity. Also adds tests for IamfMimeUtil.
2. Implements SplitString() in //starboard/common/string.h. Adds tests for SplitString().
3. Updates GetAudioCodecFromString() to use IamfMimeUtil. The Iamf test for GetAudioCodecFromString() now has a few more cases to check for validity.

   b/356704307

PR #4188, opened 2024-10-02
[media] Add codec util test for IAMF

osagie98:
Adds a test for GetAudioCodecFromString() to parse the codecs parameter of an IAMF MIME.

`GetAudioCodecFromString()` will more strictly parse for valid values of the IAMF codec param in a follow up https://github.com/youtube/cobalt/pull/4061.

b/356704307

PR #7777, opened 2025-10-29
Cherry-pick IAMF utility commits

osagie98:
This cherry-picks two Cobalt 25 IAMF commits:

- https://github.com/youtube/cobalt/pull/4188 adds a CodecUtilTest to check that IAMF mime types are parsed correctly.
- https://github.com/youtube/cobalt/pull/4244 add a class `IamfMimeUtil` to aid in parsing IAMF mime types.

Bug: 376312767


starboard/common/string.h
```diff
@@ -21,6 +21,7 @@
 
 #include <stdarg.h>
 #include <stdio.h>
+
```
borongc, line 24: nit: is this on purpose?

osagie98: It looks like the whitespace between C headers and C++ headers was removed at some point, so this cherry-pick restored it. Not sure why it was removed initially.

PR #7778, opened 2025-10-29
Package unstripped Evergreen libcobalt.so

briantting:
Package unstripped Evergreen libcobalt.so to Kokoro builds including setting up package.jsons for evergreen-arm-softfp and evergreen-arm64.

Bug: 445996864

### DIFF
--- a/cobalt/build/evergreen-arm-hardfp-raspi/package.json
+++ b/cobalt/build/evergreen-arm-hardfp-raspi/package.json
@@ -2,12 +2,13 @@
     "archive_datas": [
         {
             "files": [
-                "libcobalt.so",
                 "cobalt_shell.pak",
+                "libcobalt.so",
                 "loader_app",
                 "gen/build_info.json",
                 "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt",
+                "lib.unstripped/libcobalt.so",
                 "native_target/crashpad_handler"
             ],
             "dirs": [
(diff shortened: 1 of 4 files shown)


cobalt/build/evergreen-arm-hardfp-raspi/package.json
```diff
@@ -2,12 +2,13 @@
...
         {
             "files": [
-                "libcobalt.so",
                 "cobalt_shell.pak",
+                "libcobalt.so",
                 "loader_app",
                 "gen/build_info.json",
                 "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt",
+                "lib.unstripped/libcobalt.so",
```
andrewsavage1, line 11: This is only available for qa and gold builds, not devel or debug builds. Will that cause any issues?

cobalt/build/evergreen-arm-softfp/package.json
```diff
@@ -0,0 +1,61 @@
+{
+    "archive_datas": [
+        {
+            "files": [
+                "cobalt_shell.pak",
+                "libcobalt.so",
+                "loader_app",
```
andrewsavage1: This and the crashpad_handler would not be available for softfp or arm64 builds

cobalt/build/evergreen-arm-hardfp-raspi/package.json
briantting: It should be fine, Kokoro only builds qa and gold.

cobalt/build/evergreen-arm-softfp/package.json
briantting: ack

PR #7786, opened 2025-10-30
cobalt/content: Don't launch unnecessary services

yell0wd0g:
This CL disables some Content services that Cobalt doesn't
need nor use, via BUILDFLAG. This is a less ambitious version
of #7496 , but based on preprocessor carveouts.

With this there're 2 less threads on a linux monolithic build
(similar reduction is expected for any other platform, but not
confirmed): from 75 -> 73.

The list of Services not launched via `BrowserMainLoop` is, 
concretely:

- HidManager
- OopDataDecoder,
- Subsystem:Midi,
- GamepadService,
- FidoHidDiscovery,
- SpeechRecognitionManager,
- SavefileManager

And via content/utility/services:

- AuctionWorklet (this is for Ad Auctions)
- DataDecoder
- StorageService
- VideoCapture

bug: 383140932
Bug: 450888267


yell0wd0g:
@y4vor PTAL -- this is against main, a separate, reduced version can be landed vs
`26.eap` or `26.android` but this particular CL won't apply cleanly that way back.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -73,7 +73,8 @@ public static List<String> getDefaultCommandLineOverridesList() {
         paramOverrides.add("--disable-accelerated-video-encode");
         // Rasterize Tiles directly to GPU memory.
         paramOverrides.add("--enable-zero-copy");
-
+        // Don't enumerate or use real capture devices (audio, video).
+        paramOverrides.add("--use-fake-device-for-media-stream");
```
y4vor: Doesn't that break our microphone implementation?

cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -69,6 +71,8 @@ static constexpr auto kCobaltToggleSwitches = std::to_array<const char*>({
     // Cobalt doesn't use Chrome's accelerated video decoding/encoding.
     switches::kDisableAcceleratedVideoDecode,
     switches::kDisableAcceleratedVideoEncode,
+    // Don't enumerate or use real capture devices (audio, video).
+    switches::kUseFakeDeviceForMediaStream,
```
y4vor: We should keep the audio capture for the microphone.

content/utility/services.cc
```diff
@@ -415,13 +421,17 @@ void RegisterIOThreadServices(mojo::ServiceFactory& services) {
 }
 
 void RegisterMainThreadServices(mojo::ServiceFactory& services) {
+#if !BUILDFLAG(IS_COBALT)
   services.Add(RunAuctionWorkletService);
   services.Add(RunAudio);
```
y4vor, line 429: Probably we want to keep `RunAudio`.

yell0wd0g: Gotcha!

PR #7788, opened 2025-10-30
result_test.cc: Use EXPECT_DEATH_IF_SUPPORTED

rakuco:
EXPECT_DEATH() does not always exist (it does not on tvOS, for example). Use the `IF_SUPPORTED` variant to avoid weird build errors like

../../starboard/common/result_test.cc:128:12: error: expected '(' for function-style cast or type construction
  128 |         int* n = nullptr;
      |         ~~~^
../../starboard/common/result_test.cc:128:14: error: use of undeclared identifier 'n'
  128 |         int* n = nullptr;
      |              ^
../../starboard/common/result_test.cc:137:30: error: expected '(' for function-style cast or type construction
  137 |         std::unique_ptr<int> null_ptr;
      |         ~~~~~~~~~~~~~~~~~~~~ ^

Bug: 449188714

### DIFF
--- a/starboard/common/result_test.cc
+++ b/starboard/common/result_test.cc
@@ -114,15 +114,9 @@ TEST(NonNullResultTest, Failure) {
   EXPECT_NE(result.error().find("Something went wrong"), std::string::npos);
 }
 
-#if SB_IS(EVERGREEN)
-#define MAYBE_DeathTest DISABLED_DeathTest
-#else
-#define MAYBE_DeathTest DeathTest
-#endif
-
 class NonNullResultDeathTest : public ::testing::Test {};
 
-TEST_F(NonNullResultDeathTest, MAYBE_DeathTest_RawPointer) {
+TEST_F(NonNullResultDeathTest, DeathTest_RawPointer) {
   EXPECT_DEATH_IF_SUPPORTED(
       {
         int* null_ptr = nullptr;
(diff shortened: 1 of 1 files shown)


rakuco:

starboard/common/result_test.cc
kjyoun: With EXPECT_DEATH_IF_SUPPORTED, We may not need this #if/#endif block any more.

Can you remove this block?

starboard/common/result_test.cc
```diff
@@ -123,7 +123,7 @@ TEST(NonNullResultTest, Failure) {
 class NonNullResultDeathTest : public ::testing::Test {};
 
 TEST_F(NonNullResultDeathTest, MAYBE_DeathTest_RawPointer) {
```
kjyoun: Then this will be "DeathTest_RawPointer"

starboard/common/result_test.cc
rakuco: Done!

starboard/common/result_test.cc
rakuco: Done.

kjyoun (approved):
CI checks are green

LGTM

rakuco:
Not sure if a second LGTM, but cc @fayev-g just in case. Could someone enable auto-merge for this PR as well?

andrewsavage1 (approved):
It would be great to do this more places!

PR #7789, opened 2025-10-30
BACKPORT: MessagePumpIOSForIO: Fix handling of read+write one-shot events

rakuco:
This is a clean backport of https://crrev.com/c/6633541, which fixes a corner case issue in the I/O message pump used by tvOS.

Bug: 432824509

---

Original CL description:

MessagePumpIOSForIO had the same problem that
https://crrev.com/c/6043387 fixed in MessagePumpEpoll: when watching read and write events in one-shot mode, the code was still dispatching both events even though the contract is that only one will fire.

Apply the fix from the CL above by taking into account whether we have a persistent watcher or not when deciding whether to fire both events.

This fixes IOWatcherFdTest.ReadWriteUnifiedOneShot on tvOS.

Change-Id: I0e7adf7e16a04af45b947f15bfc7dd616bb5408e
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6633541
Cr-Commit-Position: refs/heads/main@{#1472485}

### DIFF
--- a/cobalt/build/configs/common.gn
+++ b/cobalt/build/configs/common.gn
@@ -50,7 +50,7 @@ v8_enable_atomic_object_field_writes = false
 v8_use_external_startup_data = false
 
 # Disable hls_demuxer so that enable_mse_mpeg2ts_stream_parser=false
-enable_hls_demuxer=false
+enable_hls_demuxer = false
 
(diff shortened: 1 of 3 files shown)


rakuco:

PR #7790, opened 2025-10-30
BACKPORT: [tvOS] Use launch_mac.cc with posix_spawn*() functions on the simulator

rakuco:
This is a clean backport of https://crrev.com/c/6687371, which is required for properly running tests on the tvOS simulator with multiple processes.

Bug: 432824509

---

Original CL description:

On tvOS (as of version 26), all posix_spawn*() functions in <spawn.h> except for posix_spawnp() are marked unavailable despite being available.

This makes running tests very tricky, as by default base::TestLauncher assumes it can spawn tests in another process, and running tests with --single-process-tests makes monitoring timeouts harder and breaks browser tests (which assume each test runs in a separate process).

Since the symbols do exist, work around this issue by declaring these functions with different visibility annotations before including <spawn.h> (the fact that it is included as a system header disables some warnings and makes it unnecessary to explicitly ignore -Wavailability here).

Using multiple processes is still disallowed in the App Store, so ensure this file is built only on simulator builds by explicitly #error'ing out if launch_mac.cc is used on a device build.

Change-Id: I416171ede66975ba1176ad77fc8c998a30b59e0f
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6687371
Cr-Commit-Position: refs/heads/main@{#1483338}

### DIFF
--- a/base/BUILD.gn
+++ b/base/BUILD.gn
@@ -2329,10 +2329,16 @@ component("base") {
         ]
       }
 
-      # We include launch_mac on simulator builds so unittests can fork. But,
-      # platforms like tvOS continue to use launch_ios since they do not support
-      # multi-processes.
-      if (target_environment == "simulator" && target_platform == "iphoneos") {
+      # We include launch_mac on simulator builds so unit and browser tests can
+      # fork.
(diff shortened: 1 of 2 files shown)


rakuco:

PR #7793, opened 2025-10-30
Restrict multi-packets per call QUIC optimizations to Starboard or Android

rakuco:
These optimizations depend on sendmmsg() and recvmmsg(), which are Linux-specific APIs, so restrict it to platforms that we know to be running the Linux kernel: `IS_STARBOARD` and `IS_ANDROID`.

On other platforms (e.g. tvOS), just rely on Socket::ReadMultiplePackets() returning `ERR_NOT_IMPLEMENTED` to let the code fall back to the upstream behavior.

Add a new build flag to indicate whether this optimization is enabled instead of adding checks for Starboard and Android without context in the tree.

Bug: 433539383

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,5 +1,5 @@
 # See the following for configuration options and the default values
-# https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github#default-configuration
+# https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github#config.yaml-schema
 
 have_fun: true # Haikus and other bits
 
@@ -7,6 +7,7 @@ code_review:
   disable: false
   comment_severity_threshold: MEDIUM # Default medium severity
   pull_request_opened:
-    code_review: false # Post a code review when PR opened
+    include_drafts: false # Disables agent functionality on draft pull requests.
+    code_review: true # Post a code review when PR opened
     summary: false # Summarize the change when PR opened 
     help: false # Post a help text
--- a/.github/actions/build/action.yaml
+++ b/.github/actions/build/action.yaml
@@ -10,18 +10,6 @@ inputs:
   test_targets_json_file:
     description: "The path to the file the test targets are written to."
     required: true
-  upload_on_host_test_artifacts:
-    description: "Indicates if on-host test artifacts should be uploaded."
-    required: true
-  upload_on_device_test_artifacts:
-    description: "Indicates if on-device test artifacts should be uploaded."
-    required: true
-  upload_web_test_artifacts:
-    description: "Indicates if web test artifacts should be uploaded."
-    required: true
-  test_artifacts_key:
-    description: "Artifact key used to store on-host test artifacts."
-    required: true
 runs:
   using: "composite"
   steps:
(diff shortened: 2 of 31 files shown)


rakuco:

ckim1346 (commented):
IIUC and the intent is just to exclude `IS_APPLE`, instead of adding `IS_ANDROID` and `IS_LINUX` could we just use `!BUILDFLAG(IS_APPLE)`?

rakuco:
We could, I just thought this would be more future-proof if/when e.g. Windows support is readded.

jellefoks:
I think this is ok. While the purest way may be to split the |recvmmsg| version into a `udp_socket_linux.cc` since `recvmmsg` is a Linux kernel API (and Android is Linux), there is precedent in calling it POSIX because quiche is also calls it POSIX (in its `quic_udp_socket_posix.inc`), so insisting on the difference may be more pedantic than practical.

That said, this PR is like still missing matching changes to `net/quic/quic_chromium_packet_reader.cc`.

rakuco:
> That said, this PR is like still missing matching changes to `net/quic/quic_chromium_packet_reader.cc`.

Hmm, do you mean for completeness/to reduce binary size? AFAICS, the current code works because `QuicChromiumPacketReader::StartReadingMultiplePackets()` will call `socket_->ReadMultiplePackets()`, get `ERR_NOT_IMPLEMENTED` and disable this code path at runtime.

rakuco:
@jellefoks @ckim1346 could you take another look? Is there anything you'd like me to address?

jellefoks:
> > That said, this PR is like still missing matching changes to `net/quic/quic_chromium_packet_reader.cc`.
> 
> Hmm, do you mean for completeness/to reduce binary size? AFAICS, the current code works because `QuicChromiumPacketReader::StartReadingMultiplePackets()` will call `socket_->ReadMultiplePackets()`, get `ERR_NOT_IMPLEMENTED` and disable this code path at runtime.

Ah, right, `ReadMultiplePackets` is virtual and net::Socket returns that.

net/socket/udp_client_socket.cc
```diff
@@ -184,7 +184,7 @@ void UDPClientSocket::ApplySocketTag(const SocketTag& tag) {
   socket_.ApplySocketTag(tag);
 }
 
-#if BUILDFLAG(IS_COBALT)
+#if BUILDFLAG(IS_COBALT) && (BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_LINUX))
```
jellefoks: This actually does not appear to be correct: I just checked the build flags for the modular build, for example `net/url_request/url_request.c` (*), and `IS_LINUX` is not defined, nor is `OS_LINUX` defined that could be used to set BUILDFLAG_IS_LINUX (in build/build_config.h).

*: `p="linux-x64x11-modular" ; c="devel" ; t="obj/net/net/url_request.o" ; f="out/${p}_${c}" ; rm -v ${f}/${t} ; autoninja --verbose -C ${f} $t`

rakuco: Do you know which flag needs to checked for in this case? Something like `SB_IS(MODULAR) || (BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_LINUX))`? Or just go with not-tvOS for now and expand it for other platforms like Windows in the future?

andrewsavage1: You should likely use IS_STARBOARD instead

rakuco: Do you mean just `IS_STARBOARD` or `IS_STARBOARD || IS_ANDROID || IS_LINUX`?

rakuco:
@andrewsavage1 @jellefoks I've reworded the PR description and tried a new approach after talking to Andrew: it looks like the `#ifdef` combination we need is `Starboard || Android`, meaning "this platform is running the Linux kernel".

I've added a new build flag based on this combination and adjusted the code accordingly.

net/socket/udp_socket_posix.cc
```diff
@@ -280,10 +280,13 @@ int UDPSocketPosix::GetLocalAddress(IPEndPoint* address) const {
...
 }
 
-#if BUILDFLAG(IS_COBALT)
+#if BUILDFLAG(ENABLE_MULTI_PACKETS_PER_CALL_QUIC_OPTIMIZATIONS)
 #ifdef UNSAFE_BUFFERS_BUILD
 #pragma allow_unsafe_buffers
 #endif
+#endif  // BUILDFLAG(ENABLE_MULTI_PACKETS_PER_CALL_QUIC_OPTIMIZATIONS)
+
+#if BUILDFLAG(ENABLE_MULTI_PACKETS_PER_CALL_QUIC_OPTIMIZATIONS)
```
andrewsavage1: Why do you want to separate this into its own block?

rakuco: No particular reason, I think I'd done that in one of the first commits and just maintained the change. I've now merged the blocks again.

fayev-g (approved):

PR #7794, opened 2025-10-30
Include starboard player test data in the test bundles

rakuco:
These test files are required for e.g. nplb to run; if they are not found we hit a CHECK early on and cannot even list the available tests.

To avoid requiring people to know that they need to depend on this new target, make it the only possible dependency by defining it for all platforms (on non-Apple ones it is just a group with `data_deps` set to "player_download_test_data") and making "player_download_test_data" an internal target that other files do not depend on explicitly.

Bug: 433180638

### DIFF
--- a/starboard/shared/starboard/player/BUILD.gn
+++ b/starboard/shared/starboard/player/BUILD.gn
@@ -92,7 +92,6 @@ if (is_apple) {
 } else {
   group("player_test_data") {
     testonly = true
-    data_deps =
-        [ "//starboard/shared/starboard/player:player_download_test_data" ]
+    data_deps = [ ":player_download_test_data" ]
   }
 }


rakuco:

starboard/shared/starboard/player/BUILD.gn
```diff
@@ -73,3 +73,18 @@ action("player_download_test_data") {
     rebase_path("$outdir/testdata", root_build_dir),
   ]
 }
+
+if (is_apple) {
```
andrewsavage1, line 79: This is necessary in the same places player_download_test_data is needed on other platforms, correct? If that's the case, let's make it easier for dependents by creating a group `player_test_data_resources` on non-apple platforms that simply has a `public_deps` of `:player_download_test_data` and change the dependencies to instead be on `player_test_data_resources` regardless of platform (and you can set `player_download_test_data` to have a visibility of just the `player_test_data_resources` target).

rakuco: Done!

starboard/nplb/BUILD.gn
```diff
@@ -329,6 +329,7 @@ test("nplb") {
     "//starboard/nplb/compiler_compliance:cpp14_supported",
     "//starboard/shared/starboard/drm:drm_test_helpers",
     "//starboard/shared/starboard/media:media_util",
+    "//starboard/shared/starboard/player:player_test_data_resources",
```
andrewsavage1: I think it should be a data_dep still, is that not possible for some reason?

starboard/shared/starboard/player/BUILD.gn
```diff
@@ -73,3 +75,23 @@ action("player_download_test_data") {
...
+    public_deps = [ ":player_download_test_data" ]
+
+    sources = get_target_outputs(":player_download_test_data")
+
+    outputs = [ "{{bundle_resources_dir}}/content/test/starboard/shared/starboard/player/testdata/{{source_file_part}}" ]
+  }
+} else {
+  group("player_test_data_resources") {
+    testonly = true
+    data_deps = [ "//starboard/shared/starboard/player:player_download_test_data" ]
```
andrewsavage1: I would make this a public_dep if possible and have dependents use data_deps instead

starboard/nplb/BUILD.gn
rakuco: Doesn't that break the tvOS case, which needs the dependency at build-time to put the files in the bundle?

rakuco:
@andrewsavage1 fyi, I've had to rename the `player_test_data_resources` target to `player_test_data` to fix `gn gen` on Android. Apparently `nplb`'s target type was causing us to hit https://issues.chromium.org/issues/40603958 because "*_resources" is listed [here](https://source.chromium.org/chromium/chromium/src/+/main:build/config/android/internal_rules.gni;l=64;drc=6ba8004a30479e13a7e0f1dc93521ea97e043f6a)

andrewsavage1:
> @andrewsavage1 fyi, I've had to rename the `player_test_data_resources` target to `player_test_data` to fix `gn gen` on Android. Apparently `nplb`'s target type was causing us to hit https://issues.chromium.org/issues/40603958 because "*_resources" is listed [here](https://source.chromium.org/chromium/chromium/src/+/main:build/config/android/internal_rules.gni;l=64;drc=6ba8004a30479e13a7e0f1dc93521ea97e043f6a)

Oh that's a nasty implicit bug—good work finding it, I wonder how we could avoid hitting it in the future

PR #7798, opened 2025-10-30
Add github actions for 26.eap

andrewsavage1:
Bug: 452945577

### DIFF
--- a/.github/workflows/android_26.eap.yaml
+++ b/.github/workflows/android_26.eap.yaml
@@ -1,4 +1,4 @@
-name: android
+name: android_26.eap
 
 on:
   pull_request:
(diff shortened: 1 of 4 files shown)


.github/workflows/android_26.eap.yaml
oxve, line 1: To avoid mixing up the workflows on https://github.com/youtube/cobalt/actions can you rename the workflow files?

Compare with `26.android`: https://github.com/youtube/cobalt/blob/26.android/.github/workflows/android_26.android.yaml

.github/workflows/android.yaml
oxve, line 1: And update these names.

oxve (approved):

PR #7806, opened 2025-10-30
cobalt/precommit: Adds a presubmit entry for Java Strings w/o/ Locale

yell0wd0g:
Vibe coded  : )

```
mcasas@ydog2 ~/p/cobalt/src (b_456490110__add_presubmit_for_java_string_without_locale) $ git diff
diff --git a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java
index a7270c4d38e..b8d634ad772 100644
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java
@@ -121,6 +121,7 @@ public abstract class CobaltService {
             + "})(window)";
 
     String jsCode = String.format(Locale.US, jsCodeTemplate, nativeService, base64Data);
+    String miguelao = String.format( "foo foo");
     this.cobaltActivity.evaluateJavaScript(jsCode);
   }
 }
mcasas@ydog2 ~/p/cobalt/src (b_456490110__add_presubmit_for_java_string_without_locale) $ git commit -m  "Should not work" -a
check for case conflicts.................................................Passed
...
// Many checks
...
gn-format............................................(no files to check)Skipped
Check Java String.format includes Locale.................................Failed
- hook id: check-java-strings-have-locale
- exit code: 1

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java:124: String.format call should include a java.util.Locale argument.
```

Bug: 456490110


yell0wd0g:
Someone from @youtube/cobalt-build PTAL ?

cobalt/precommit/check_java_strings_have_locale.py
```diff
@@ -0,0 +1,33 @@
...
+
+  for file_path in args.files:
+    try:
+      with open(file_path, 'r', encoding='utf-8') as f:
+        for line_num, line in enumerate(f, 1):
+          if pattern.search(line):
+            errors.append(f'{file_path}:{line_num}: String.format call '
+                          'should include a java.util.Locale argument.')
+    except Exception as e:
+      print(f"Error processing file {file_path}: {e}", file=sys.stderr)
```
andrewsavage1, line 26: You want to catch any possible error and succeed? Why?

yell0wd0g: I think the rationale in these filters is to only catch one thing,
in this case the string-w/o-locale; if the files fail to open for
some reason, the presubmit won't hold the commit. Makes sense? 
Otherwise LMK and I can append an `error` here as well.

oxve: I think this is reasonable.

oxve (approved):

cobalt/precommit/check_java_strings_have_locale.py
yell0wd0g: Gonna resolve this convo to submit this CL/PR. @andrewsavage1 if you
still have concerns we can address them in subsequent CL/PRs.

yell0wd0g:
@kaidokert PTAL

cobalt/precommit/check_java_strings_have_locale.py
```diff
@@ -0,0 +1,33 @@
...
+
+def main():
+  parser = argparse.ArgumentParser()
+  parser.add_argument('files', nargs='*')
+  args = parser.parse_args()
+
+  errors = []
+  # This regex looks for String.format( where the rest of the line does not
+  # contain "Locale". This is a heuristic for single-line calls.
+  pattern = re.compile(r'\bString\.format\s*\((?!.*Locale)')
```
zhongqiliang: We only need to add "Locale.US" when there is number in the parameter. It should search for String.format and %d.

yell0wd0g: Done, I think.

yell0wd0g:
@zhongqiliang PTAL

cobalt/precommit/check_java_strings_have_locale.py
zhongqiliang: Sorry, the important part is that "%d" needs to be there.
String.format("Some thing, %s", str); --> this does not need Locale.US.
String.format(Locale.US, "Some thing, %d", int_or_float); --> this needs Locale.US.

yell0wd0g: Gotcha, then the regex should be 

`String\.format\((?!.*Locale\.US)(?=.*".*?\%d.*").*?\);`

Some examples I ran:
```
String miguelao = String.format("foo %d foo", 123);              <------ BAD!

String miguelao = String.format(Locale.US, "foo %d foo", 123);   <-- GOOD

String miguelao = String.format(Locale.US, "foo %s foo", "bla"); <-- GOOD

String.format("Some thing, %s", str);                            <-- GOOD
```

yell0wd0g:
@zhongqiliang sorry I forgot about this CL, PTAL again at the latest changes.

kaidokert (changes requested):
I'd rather request we enable ErrorProne Java checking. We already have this in our tooling / infra - needs to be configured and enabled.

One off regexes are going to be brittle and fail - false sense of security

PR #5551, opened 2025-04-24
[android] Refine video render algorithm

jasonzhangxx:
1. Allow VideoRenderAlgorithm release first frame as soon as possible.
2. Besides the first frame, other frames will be released after any
   audio frame has been consumed.

Bug: 186660620

PR #7808, opened 2025-10-31
Fix all highest level [5] cpplint errors

fayev-g:
The cpplint tool (Google C++ style) was added in C25 and cherry-picked into main (m138) in https://github.com/youtube/cobalt/pull/7783. This PR fixes all highest level [5] errors that exist in the codebase. (In theory your PRs shouldn't be blocked by errors in files you don't touch.) These errors are mostly header guard, missing copyright, missing header inclusion, and syntax errors (missing terminating newline).

Bug: 456554108

### DIFF
--- a/cobalt/shell/app/resource.h
+++ b/cobalt/shell/app/resource.h
@@ -17,39 +17,35 @@
 // Used by test_shell.rc
 //
 
-#ifndef COBALT_SHELL_APP_RESOURCE_H_
-#define COBALT_SHELL_APP_RESOURCE_H_
 
-#define IDR_MAINFRAME 128
-#define IDM_EXIT 105
-#define IDM_CLOSE_WINDOW 106
-#define IDM_NEW_WINDOW 107
-#define IDM_SHOW_DEVELOPER_TOOLS 108
-#define IDC_CONTENTSHELL 109
-#define IDD_ALERT 130
-#define IDD_CONFIRM 131
-#define IDD_PROMPT 132
-#define IDC_NAV_BACK 1001
-#define IDC_NAV_FORWARD 1002
-#define IDC_NAV_RELOAD 1003
-#define IDC_NAV_STOP 1004
-#define IDC_PROMPTEDIT 1005
-#define IDC_DIALOGTEXT 1006
+#define IDR_MAINFRAME             128
+#define IDM_EXIT                  105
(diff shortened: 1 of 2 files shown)


fayev-g:
Will follow up with addressing [4] errors.

kjyoun (approved):

starboard/extension/feature_config.h
```diff
@@ -27,6 +27,9 @@
 // Doing it this way guarantees that features defined in the Chrobalt side and
 // starboard side share the same names, param types and initial default values.
 
+#ifndef STARBOARD_EXTENSION_FEATURE_CONFIG_H_
+#define STARBOARD_EXTENSION_FEATURE_CONFIG_H_
```
jasonzhangxx: Please ignore the lint error in this file by `// NOLINT(build/header_guard)`, like in https://github.com/youtube/cobalt/pull/5551. It's intended to not define header macro.

fayev-g: Done, thanks! Was wondering what was going on here.

cobalt/shell/app/resource.h
andrewsavage1, line 1: More of a general thought, but I'm unsure if we should be using our formatters/linters on upstream code in //cobalt/shell like this, as much of it is taken directly from upstream //content/shell code. Maybe we want to exclude it from the pre-commit checks and instead have the Chromium PRESUBMITs run on it, but I'm not sure

fayev-g:
> This seems like good cleanup, but making changes like this in the main branch when we are developing in multiple branches (main, 26.android, and 26.eap) can be annoying when cherry-picking isn't clean. Would you be able to check if #7783 and then this change apply cleanly to the 26.eap branch as well?

Discussed offline. I tried cherry-picking but there are a lot of files that don't exist in 26.android and 26.eap. We think it will be easier to just run cpplint on those branches separately. The lint errors that get fixed at [5] should generally be straightforward - like incorrect header guards and random semicolons in places.

cobalt/shell/app/resource.h
fayev-g: Thanks! I restored and excluded the one //cobalt/shell file that modified in this PR for now (resource.h). I think we'll have a broader discussion about how to prioritize linter recommendations between chromium and cobalt.

andrewsavage1 (approved):

PR #7811, opened 2025-10-31
evergreen: Prep the updater for Cobalt

yuying-y:
This change resolve remaining issues of the updater module to prepare it for integrating with the cobalt shell.

Key changes:
   - Resolved a circular dependency between cobalt/updater and cobalt/shell by moving updater-related switches to cobalt/updater/configurator.cc.
   - Remove circular dependency between //cobalt/updater and //components/update_client by moving relevant constants and functions to components/update_client/utils.*
   - Improved the destruction sequence of the UpdaterModule to prevent crashes on exit.
   - The hash verification functions in components/update_client/utils.cc are ported from 25.lts.1+. 

Issues: 431862767, 213388707

### DIFF
--- a/cobalt/updater/configurator.cc
+++ b/cobalt/updater/configurator.cc
@@ -18,8 +18,8 @@
 #include <set>
 #include <utility>
 #include "base/command_line.h"
+#include "base/time/time.h"
 #include "base/version.h"
-#include "cobalt/browser/switches.h"
 #include "cobalt/updater/network_fetcher.h"
 #include "cobalt/updater/patcher.h"
 #include "cobalt/updater/prefs.h"
@@ -46,6 +46,14 @@ const char kUpdaterJSONDefaultUrlQA[] =
     "https://omaha-qa.sandbox.google.com/service/update2/json";
 const char kUpdaterJSONDefaultUrl[] =
     "https://tools.google.com/service/update2/json";
+
+// Whether to request, download, and install uncompressed (rather than
+// compressed) Evergreen binaries.
+const char kUseUncompressedUpdates[] = "use_uncompressed_updates";
+
+// Uses the QA update server to test the changes to the configuration of the
+// PROD update server.
+const char kUseQAUpdateServer[] = "use_qa_update_server";
 
 std::string GetDeviceProperty(SbSystemPropertyId id) {
   char value[kSystemPropertyMaxLength];
@@ -86,7 +94,7 @@ Configurator::Configurator(
     SetChannel(persisted_channel);
   }
   if (base::CommandLine::ForCurrentProcess()->HasSwitch(
-          switches::kUseUncompressedUpdates)) {
+          kUseUncompressedUpdates)) {
     use_compressed_updates_.store(false);
   } else {
     use_compressed_updates_.store(true);
@@ -113,13 +121,13 @@ base::TimeDelta Configurator::UpdateDelay() const {
 }
 
 std::vector<GURL> Configurator::UpdateUrl() const {
-#if !defined(COBALT_BUILD_TYPE_GOLD)
+#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
+  base::AutoLock auto_lock(const_cast<base::Lock&>(update_server_url_lock_));
   if (allow_self_signed_packages_ && !update_server_url_.empty()) {
     return std::vector<GURL>{GURL(update_server_url_)};
   }
-#endif  // !defined(COBALT_BUILD_TYPE_GOLD)
-  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
-          switches::kUseQAUpdateServer)) {
+#endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(kUseQAUpdateServer)) {
     return std::vector<GURL>{GURL(kUpdaterJSONDefaultUrlQA)};
   } else {
     return std::vector<GURL>{GURL(kUpdaterJSONDefaultUrl)};
@@ -304,11 +312,11 @@ std::string Configurator::GetAppGuidHelper(const std::string& updater_channel,
             << "combination is undefined with the new Omaha configs.";
 
   // All undefined channel requests go to prod configs except for static
-  // channel requestsf for C24 and older.
+  // channel requests for C24 and older.
   // TODO(b/449024263): Replace regex matchers with substring_set_matcher or re2
   if (!std::regex_match(updater_channel, std::regex("2[0-4]lts\\d+")) &&
       sb_version >= 14 && sb_version <= 16) {
-    const auto it = kChannelAndSbVersionToOmahaIdMap.find(
+    it = kChannelAndSbVersionToOmahaIdMap.find(
         "prod" + std::to_string(sb_version));
     if (it != kChannelAndSbVersionToOmahaIdMap.end()) {
       return it->second;
@@ -319,6 +327,7 @@ std::string Configurator::GetAppGuidHelper(const std::string& updater_channel,
 }
 
 std::string Configurator::GetAppGuid() const {
+  base::AutoLock auto_lock(const_cast<base::Lock&>(updater_channel_lock_));
   const std::string version(COBALT_VERSION);
   return GetAppGuidHelper(updater_channel_, version, SB_API_VERSION);
 }
--- a/cobalt/updater/updater_module.cc
+++ b/cobalt/updater/updater_module.cc
@@ -14,10 +14,13 @@
 
 #include "cobalt/updater/updater_module.h"
 
+#include <string>
 #include <utility>
 #include <vector>
 #include "third_party/abseil-cpp/absl/types/optional.h"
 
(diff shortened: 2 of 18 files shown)


cobalt/updater/configurator.cc
```diff
@@ -114,12 +125,12 @@ base::TimeDelta Configurator::UpdateDelay() const {
 
 std::vector<GURL> Configurator::UpdateUrl() const {
 #if !defined(COBALT_BUILD_TYPE_GOLD)
```
andrewsavage1: Note this will never be defined, use `#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD)` instead

cobalt/updater/updater_module.cc
```diff
@@ -44,12 +50,22 @@
...
 namespace {
 
 using update_client::CobaltSlotManagement;
+using update_client::component_to_updater_status_map;
 using update_client::ComponentState;
 using update_client::UpdateCheckError;
+using update_client::updater_status_string_map;
+using update_client::UpdaterStatus;
+
+#if defined(COBALT_BUILD_TYPE_DEBUG)
```
andrewsavage1: Use BUILDFLAG(IS_DEBUG)

cobalt/updater/updater_module.cc
```diff
@@ -88,14 +104,55 @@ ComponentStateToCobaltExtensionUpdaterNotificationState(
...
+  DCHECK(task_may_run) << "Task that will never run posted with " << __func__
+                       << " from " << from_here.ToString();
+
+  if (task_may_run) {
+    // Wait for the task to complete before proceeding.
+    do {
+      if (task_finished.TimedWait(base::Milliseconds(kTimeWaitInterval))) {
+        break;
+      }
+#if !defined(COBALT_BUILD_TYPE_GOLD)
```
andrewsavage1: Use the buildflag

components/update_client/component_unpacker.cc
```diff
@@ -86,20 +86,34 @@ bool ComponentUnpacker::Verify() {
...
+#else
   VLOG(1) << "Verification successful: " << path_.value();
+#endif
   return true;
 }
 
 bool ComponentUnpacker::BeginUnzipping() {
   // Mind the reference to non-const type, passed as an argument below.
   base::FilePath& destination = is_delta_ ? unpack_diff_path_ : unpack_path_;
+  #if !BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: unindent

components/update_client/component.cc
```diff
@@ -1011,7 +1011,13 @@ void Component::StateNew::DoHandle() {
...
 
   auto& component = State::component();
-  if (component.crx_component()) {
+
+#if BUILDFLAG(IS_STARBOARD)
+  auto& config = component.update_context_->config;
+  config->SetPreviousUpdaterStatus(config->GetUpdaterStatus());
+#endif
+
+   if (component.crx_component()) {
```
andrewsavage1: Revert the indentation change here

components/update_client/component_unpacker.cc
```diff
@@ -196,10 +210,12 @@ void ComponentUnpacker::EndPatching(UnpackerError error, int extended_error) {
 }
 
 void ComponentUnpacker::EndUnpacking() {
+#if !BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 213: Should this be checking IN_MEMORY_UPDATES instead?

components/update_client/protocol_serializer_json.cc
```diff
@@ -70,11 +70,15 @@ std::string ProtocolSerializerJSON::Serialize(
   // OS version and platform information.
   base::Value::Dict os_node;
   os_node.Set("platform", request.os.platform);
-  os_node.Set("arch", request.os.arch);
-  if (!request.os.version.empty())
-    os_node.Set("version", request.os.version);
-  if (!request.os.service_pack.empty())
-    os_node.Set("sp", request.os.service_pack);
+  // TODO(b/456632915): Device info OS returns garbled characters.
```
andrewsavage1: It'd be good to similarly add a NOTIMPLEMENTED call

andrewsavage1: go/todo-format
Our TODOs should have the form `// TODO: b/123, Cobalt - Description`

components/update_client/update_checker.cc
```diff
@@ -25,6 +25,9 @@
 #include "base/task/sequenced_task_runner.h"
 #include "base/task/thread_pool.h"
 #include "build/build_config.h"
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: Put conditional includes below the rest of the includes

cobalt/shell/browser/shell.cc
```diff
@@ -57,6 +57,14 @@
 #include "third_party/blink/public/mojom/choosers/file_chooser.mojom-forward.h"
 #include "third_party/blink/public/mojom/window_features/window_features.mojom.h"
 
+#if BUILDFLAG(USE_EVERGREEN)
+#include "cobalt/updater/updater_module.h"
+#include "content/public/browser/browser_thread.h"
```
andrewsavage1: Why do you need all of these content/public includes? The only references you're adding is to two cobalt::updater methods

cobalt/shell/browser/shell.cc
andrewsavage1, line 1: Remove the changes to shell.cc and shell.h from this PR and make them in a follow up PR. This PR should focus on the changes necessary to build the updater and update_client instead, and the follow up should be explicitly about enabling the updater as you do in this file.

yuying-y: I don't see why it's necessary to separate the changes. In order to build the updater with Cobalt, the following is added to cobalt/shell/BUILD.gn
  if (use_evergreen) {
    deps = [ "//cobalt/updater" ]
  }
Then after this dependency is added, it doesn't make sense to not hook the updater with the shell. The changes in shell.* is small, and it makes sense to enable the updater in one go. Why do you think it's necessary to separate this PR?

andrewsavage1: This PR is currently doing two things: fixing the build issues in the updater code now that there is a dependency, and actually enabling the updater during startup. Because the build fixes to the updater code are fairly extensive, I want to separate them out so that anyone can easily see how the updater gets enabled at runtime during startup.

components/update_client/component_unpacker.cc
yuying-y: No, the unpack_path_ points to the installation slot, which we'll keep around.

andrewsavage1: Makes sense why we wouldn't want to delete the unpack path, but why wouldn't we want to delete the unpack diff path?

components/update_client/protocol_serializer_json.cc
yuying-y: Hmm, since this is in the middle of the function, not sure where to add it. Suggestions?

andrewsavage1: Not sure why that's an issue, NOTIMPLEMENTED doesn't return like NOTREACHED, it'll just log an error. You should be able to put it on line 75 right after the todo comment

cobalt/shell/browser/shell.cc
yuying-y: Cool. Moving them to a separate PR.

cobalt/shell/browser/shell.cc
yuying-y: I certainly need the storage_partition.h. For the others, I recall I hit compile errors without them. Will try again after this PR is merged and resolve in https://github.com/youtube/cobalt/pull/7854.

components/update_client/component_unpacker.cc
yuying-y: unpack_diff_path_ is the equivalent of unpack_path_ for delta update - an update containing just the diff compared to the previous version, not a complete package. It should be treated the same as unpack_path_.

andrewsavage1: Got it, thanks for clarifying!

cobalt/updater/configurator.cc
```diff
@@ -47,6 +47,17 @@ const char kUpdaterJSONDefaultUrlQA[] =
 const char kUpdaterJSONDefaultUrl[] =
     "https://tools.google.com/service/update2/json";
 
+// Moved the switches from cobalt/browser/switches.cc to avoid circular
```
andrewsavage1: I think we can remove this comment, imo it's not surprising for the variables to be here

cobalt/updater/updater_module.cc
```diff
@@ -88,14 +104,55 @@ ComponentStateToCobaltExtensionUpdaterNotificationState(
   }
 }
 
+// This function is ported from cobalt/base/task_runner_util.cc from 25.lts.1+
```
andrewsavage1: I think it's fine to remove this comment

components/update_client/protocol_serializer_json.cc
yuying-y: Cool, done.

PR #7812, opened 2025-10-31
Add cobalt-specific paths_mac implementation for tvOS

rakuco:
Originally done by @abhijeetk for M114, and later adjusted for M138.

Copied content/shell/app/paths_mac.h and paths_mac.mm to cobalt/shell/app/ to provide Mac-specific path utilities for the cobalt shell, replacing the dependency on content/shell/app/paths_mac.h.

The copyright header was updated following guidance from bug 451947281.

Bug: 451947281
Bug: 456692793


rakuco:

andrewsavage1:
Can you link the change that was made in m114?

rakuco:
> Can you link the change that was made in m114?

Sorry, I meant the change was originally written in our internal fork by Abhijeet, but it was never upstreamed to this repository.

cobalt/shell/BUILD.gn
```diff
@@ -368,6 +368,13 @@ static_library("cobalt_shell_app") {
     "app/shell_main_delegate.h",
   ]
   public_deps = [ ":cobalt_shell_app_deps" ]
+
+  if (is_apple) {
```
fayev-g, line 372: Just for posterity - is paths_mac for all apple platforms and badly named or is it for mac only? It looks like in the .mm it specifies `.Append("MacOS")` at line 80 also. Thanks in advance for clarifying

fayev-g (approved):

cobalt/shell/BUILD.gn
rakuco: > is paths_mac for all apple platforms and badly named or is it for mac only?

The former (just like the upstream version from content shell) :melting_face: I believe this was originally Mac-specific and later adopted by iOS (which only calls some of those functions).

abhijeetk: > Just for posterity - is paths_mac for all apple platforms and badly named or is it for mac only? 

@fayev-g I have created CL : https://chromium-review.googlesource.com/c/chromium/src/+/7117381 to rename file. Thanks for pointing it out.

fayev-g: Nice! Thanks for being proactive about making this clearer.

y4vor (approved):

rakuco:
Hmm, I thought @y4vor was part of the cobalt-embedded-owners team. Maybe @andrewsavage1 is?

PR #7814, opened 2025-10-31
[tvOS] cobalt shell: Adapt to https://crrev.com/c/5507654

rakuco:
This upstream change moved and renamed asset files and changed some build system rules. This needs to be done in Cobalt Shell as well.

Bug: 456692793


rakuco:

andrewsavage1 (approved):
Seems okay, I'll let Faye take a look too and submit

PR #7815, opened 2025-10-31
Trim the feature filters of ICU

haozheng-cobalt:
Updated the ICU feature filter configuration in third_party/icu/filters/cobalt.json to reduce the number of included locales for currency, collation, region, unit, and zone data trees in the feature list.

This change aligns Cobalt's ICU data filters with the ICU63 version from Chromium's source, and C25 filter rules.

This PR reduces Cobalt.apk size by 1.32 MB.

Test: built devel and ran CUJ locally.
Bug: 446966279


johnxwork:
Can you describe the impact of removing these locale/languages?

That might aid in the discussion of why we remove and why we don't remove more.

haozheng-cobalt:
> Can you describe the impact of removing these locale/languages?
> 
> That might aid in the discussion of why we remove and why we don't remove more.

The removal decision is based on the combination of ICU filers in C25: https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:third_party/icu/filters/cobalt.json, and YouTube UI text https://ytglobal.googleplex.com/config/languages.

The removal configs are all supposed to not be used in Youtube.

PR #7822, opened 2025-10-31
Remove testonly from non-test build targets

sideb0ard:
Bug: 375655377

Removes testonly and adds needed jni dependencies uncovered when testonly is removed.

### DIFF
--- a/cobalt/shell/BUILD.gn
+++ b/cobalt/shell/BUILD.gn
@@ -23,9 +23,6 @@ import("//tools/grit/repack.gni")
 # true.
 shell_use_toolkit_views = toolkit_views && !is_castos
 
-# Web test support only built for the linux-x64x11-no-starboard target.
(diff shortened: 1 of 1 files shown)


johnxwork:
Can we try cherry-picking this to 26android?

cobalt/android/BUILD.gn
```diff
@@ -206,6 +204,7 @@ shared_library_with_jni("libchrobalt") {
       "//media/midi",
     ]
   }
+  remove_uncalled_jni = true
```
zhongqiliang, line 207: Does gold build have this by default?
Is there a list of what JNI methods in c++ code that are not used?

sideb0ard: the unused ones are the ones in your bug report - https://buganizer.corp.google.com/issues/456268748#comment1

PR #7831, opened 2025-11-02
Add smaps capture scripts add end to end pipeline script

Awallky:
cobalt/tools: Introduce unified smaps analysis pipeline script

This commit adds a new script, run_analysis_pipeline.py, to streamline the smaps memory analysis workflow. This script integrates the existing read_smaps_batch.py, analyze_smaps_logs.py, and visualize_smaps_analysis.py tools into a single, convenient command.

Key features of the unified pipeline:

       - Simplified Workflow: Users can now generate a complete visual memory analysis report from raw smaps logs with a single command, eliminating the need to run
         multiple scripts sequentially.
       - Automated Processing: The script handles the creation of temporary directories for processed logs and their subsequent cleanup, reducing manual overhead.
       - End-to-End Testing: A new unit test, run_analysis_pipeline_test.py, has been added to verify the correct execution and integration of all components within the
          pipeline.
       - Documentation: The README.md has been updated to introduce the new pipeline script, its usage, and examples.

This enhancement significantly improves the usability and efficiency of the smaps memory analysis toolchain.

#vibe-coded
Bug: 456178181

### DIFF
--- a/cobalt/tools/performance/smaps/README.md
+++ b/cobalt/tools/performance/smaps/README.md
@@ -217,11 +217,13 @@ python3 run_analysis_pipeline.py <RAW_LOG_DIR> [OPTIONS]
     The path to the directory containing the raw smaps log files.
 *   `--output_image` (type: `str`, default: `smaps_analysis.png`)
     The path where the final output PNG image will be saved.
+*   `--platform` (type: `str`, choices: `android`, `linux`, default: `android`)
+    Specify the platform for platform-specific aggregations.
 
 #### Example
 
 ```bash
-python3 run_analysis_pipeline.py cobalt_smaps_logs --output_image my_analysis.png
+python3 run_analysis_pipeline.py cobalt_smaps_logs --output_image my_analysis.png --platform android
 ```
 
 ## Improving Aggregation Rules
--- a/cobalt/tools/performance/smaps/read_smaps.py
+++ b/cobalt/tools/performance/smaps/read_smaps.py
@@ -82,13 +82,7 @@ def read_smap(args):
     # Find the name of the memory region
     key = '<anon>'
     if len(head) > 5:
-      for i in reversed(range(5, len(head))):
-        if head[i].startswith('[') or head[i].startswith('/'):
-          key = head[i]
-          break
-        elif not head[i].isdigit():
-          key = head[i]
-          break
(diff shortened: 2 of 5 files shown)


Awallky:
Potential Improvements:

   1. Type Hinting: The code would benefit from the addition of Python type hints. This would improve readability and allow for static analysis to catch potential bugs.
   2. Centralized Logging: Consider replacing the print() statements used for status updates and errors with a centralized logging configuration. This would provide more granular control over log levels (e.g., DEBUG, INFO, ERROR) and output destinations.
   3. Configuration File for Aggregation Rules: The aggregation rules in read_smaps.py are hardcoded as a series of re.sub() calls. For easier maintenance and extensibility, these rules could be moved to an external configuration file (e.g., a YAML or JSON file). This would allow users to add new rules without modifying the Python code.
   4. Enhanced Output and Interactivity
   * Interactive Visualizations: Instead of a static PNG, the visualization script could generate an interactive HTML file using libraries like Plotly or Bokeh. This would allow users to hover over data points to see exact values, zoom into specific time ranges, and toggle the visibility of different memory regions, making exploration much easier.
   * Comprehensive HTML Reports: The final output could be a single HTML report that includes the summary text from the analysis (top consumers, top growers) alongside the interactive plots. This would create a single, shareable artifact for each analysis run.
5. Performance and Robustness
   * Parallel Processing: The read_smaps_batch.py script processes files serially. For a large number of log files, this could be parallelized using Python's multiprocessing module to significantly speed up the processing step on multi-core machines.
   * Formal Dependency Management: The project relies on pandas and matplotlib. Adding a requirements.txt file would formalize these dependencies, making it easier for new users to set up the environment correctly (pip install -r requirements.txt).
6. Advanced Analysis and Extensibility
   * Historical Comparison / Diffing: A new tool could be added to compare the JSON outputs of two different runs (e.g., before and after a code change). This would be powerful for regression testing, allowing you to see exactly which memory regions increased or decreased between the two versions.
   * Platform-Specific Logic Abstraction: As the tool evolves, the if args.platform == 'android': checks could be refactored into a more scalable, object-oriented design. You could have a base Platform class with subclasses like AndroidPlatform and LinuxPlatform that encapsulate the specific aggregation rules for each, making it cleaner to add support for new platforms in the future.

PR #7841, opened 2025-11-03
ci: Use zstd compression for test artifacts

oxve:
Replaces gzip with zstd compression for test artifact archives used in linux workflows. This algorithm provides a better compression ratio and is significantly faster, leading to a reduction in CI job times. Time saved (compress+extract+network) using zstd over gzip is >4 minutes on linux and >5 minutes on evergreen-x64.

The change is implemented conditionally, falling back to gzip for platforms where zstd is not available on the devices.

Bug: 432536319


.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -38,9 +38,11 @@ runs:
 
         FLAGS=""
         if [[ "${{ matrix.platform }}" == *android* ]]; then
-          FLAGS="--archive-per-target --use-android-deps-path --flatten-deps"
+          FLAGS="--compression gz --archive-per-target --use-android-deps-path --flatten-deps"
```
isarkis, line 42: nit: can we add a comment before this if statement explaining why zstd can't be used on these platforms?

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -38,9 +38,11 @@ runs:
...
         elif [[ "${{ matrix.platform }}" == *raspi* || "${{ matrix.platform }}" == *rdk* ]]; then
-          FLAGS="--archive-per-target --flatten-deps"
+          FLAGS="--compression gz --archive-per-target --flatten-deps"
+        else
+          FLAGS="--compression zstd --compression-level=9"
```
oxve, line 46: Running `zstd` with compression level 9 adds ~10-15 seconds to the compress/extract times and saves ~25% (30s-1.5m) on network transfer.

.github/actions/upload_test_artifacts/action.yaml
oxve: Added a comment explaining that android, raspi, and rdk don't have zstd installed

PR #7848, opened 2025-11-03
3p: Enable cache age limit to reduce memory usage

sherryzy:
Enables the kLimitImageDecodeCacheAge feature by default and reduces the kCacheAgeLimitSeconds from 10 to 5 seconds. This allows the GPU image decode cache to be purged more frequently based on age, in addition to its size limit.

Bug: 424591313


johnxwork:
Would we want `LimitImageDecodeCacheAge` on 26.android?

sherryzy:
@johnxwork I added this for 26.android first: https://github.com/youtube/cobalt/pull/7733.

PR #7849, opened 2025-11-04
Add missing dependencies needed for Kimono

sideb0ard:
Bug: 454931711

Under m114, we have a load of jar files under out/$TARGET/lib.java which are missing under m138.
CoAT runs fine, so these are all only needed for Kimono. 

As listed in b/454931711 these are missing and are used in our build scripts
lib.java/third_party/android_deps/utils/java.jar
lib.java/third_party/androidx/androidx_appcompat_appcompat_resources_java.jar
lib.java/third_party/androidx/androidx_core_core_java.jar

Using 'androidx_appcompat_appcompat_resources_java.jar' as an example, it seems this is now packaged as an AAR file - an android specific archive format. Gemini tells me we just need to include the missing depencies to our build target. 

Using the list of needed/missing java files in b/456268748 , I've now added the target for each of these as a dependency.

### DIFF
--- a/cobalt/shell/android/BUILD.gn
+++ b/cobalt/shell/android/BUILD.gn
@@ -65,7 +65,6 @@ android_resources("cobalt_shell_java_resources") {
 
 android_library("cobalt_shell_java") {
   testonly = true
-
   resources_package = "dev.cobalt.shell"
   deps = [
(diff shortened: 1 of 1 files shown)


johnxwork:
Do we know why we didn't need these explicit deps in m114?

PR #7859, opened 2025-11-04
cobalt/shell: Exclude web tests code from tvOS too

rakuco:
Extend the existing Android and Starboard checks to include tvOS as well.

Bug: 456692793

### DIFF
--- a/cobalt/shell/app/shell_main_delegate.cc
+++ b/cobalt/shell/app/shell_main_delegate.cc
@@ -58,7 +58,7 @@
 #define IPC_MESSAGE_MACROS_LOG_ENABLED
 #include "content/public/common/content_ipc_logging.h"
 #define IPC_LOG_TABLE_ADD_ENTRY(msg_id, logger) \
-    content::RegisterIPCLogger(msg_id, logger)
+  content::RegisterIPCLogger(msg_id, logger)
 #endif
 
 #if !BUILDFLAG(IS_ANDROID) && !BUILDFLAG(IS_IOS_TVOS) && \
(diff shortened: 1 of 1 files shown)


rakuco:

fayev-g (approved):

cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -58,10 +58,11 @@
 #define IPC_MESSAGE_MACROS_LOG_ENABLED
 #include "content/public/common/content_ipc_logging.h"
 #define IPC_LOG_TABLE_ADD_ENTRY(msg_id, logger) \
-    content::RegisterIPCLogger(msg_id, logger)
+  content::RegisterIPCLogger(msg_id, logger)
 #endif
 
-#if !BUILDFLAG(IS_ANDROID) && !BUILDFLAG(IS_STARBOARD)
+#if !BUILDFLAG(IS_ANDROID) && !BUILDFLAG(IS_IOS_TVOS) && \
+    !BUILDFLAG(IS_STARBOARD)
```
yell0wd0g, line 65: Looking at this condition my first thought was that this
would be tantamount to  `#if !BUILDFLAG(IS_COBALT)` but
the GN file says that there is a situation in which we
build with `IS_COBALT` but no starboard/android/tvos/
other future Cobalt platforms, i.e,. Cobalt on Linux 
desktop it's quite confusing, so could we just add a 
gn section to define a new `SUPPORT_WEB_TESTS` `BUILDFLAG`?
(Iterations on the name welcomed).

An example is `USE_VAAPI` under `//media`, [here](https://source.chromium.org/chromium/chromium/src/+/main:media/gpu/BUILD.gn;l=16-21;drc=459d9db973100448f4e2ac77e39623bfe930ee48).
We'd need to include a new file but I think it'll make
this code more legible down the line.

rakuco: This makes sense. I've filed bug 458876995 to track this independently.

PR #7805, opened 2025-10-30
cobalt: Add h5vcc-embedded custom scheme

borongc:
Introduce the h5vcc-embedded custom URL scheme to serve resources
directly from the Cobalt binary. A custom URLLoaderFactory is
implemented to resolve URLs using this scheme, retrieving content
from embedded resources.

This enables the loading of pre-packaged, trusted content, such as
splash screens, ensuring instant availability and removing external
network dependencies. Navigations using h5vcc-embedded URLs are
exempted from HTTPS enforcement and Content Security Policy (CSP)
checks, as this content is considered secure and internal to the
application.

This is required for https://github.com/youtube/cobalt/pull/7803.

Issue: 456482732
Issue: 441738743

PR #7860, opened 2025-11-04
[Review only]Cache manifest defined splash video

linxinan-yt:
Cache manifest defined splash video

1. A new SplashScreenManager class is introduced to handle all splash screen logic. 
2. CobaltActivity is updated to use the new SplashScreenManager and to load the splash screen from a data URL.
3. CobaltWebContentsObserver is updated to trigger caching splash video at DidUpdateWebManifestURL event.

Bug: 454630524

### DIFF
--- a/cobalt/android/apk/app/src/main/assets/splash.html
+++ b/cobalt/android/apk/app/src/main/assets/splash.html
@@ -1,11 +1,9 @@
 <!doctype html>
 <html lang="en">
   <body>
+    <h1>Splash</h1>
     <video id="primary-video" class="video" muted autoplay playsinline></video>
     <script>
-      // Dictionary mapping string descriptions to media file descriptions in the
-// form of [contentType, url, maxVideoCapabilities (for videos only),
-// licenseUrl]
 const MEDIA_FILES = {
   'splash_clear': {
     contentType: 'video/webm; codecs="vp9"',
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -256,7 +256,7 @@ public abstract class CobaltActivity extends Activity {
                 }, mSplashTimeoutMs);
           }
         });
-    if (mDisableNativeSplash || !hasSplashVideoCached()) {
+    if (mDisableNativeSplash || !getSplashVideoFile().exists()) {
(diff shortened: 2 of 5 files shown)


linxinan-yt:
@yell0wd0g  Quote your questions here:

```
@linxinan-chops IIUC there's two parts here, one is the refactor of splash related
code to SplashScreenManager and then the addition of the base64 caching. Could we
have two separate PRs? The former, SplashScreenManager, should be easy to review
since it'd be just a cleanup.

Another question is: I thought @borongc made the splash resources
accessible via a Chrome UI protocol ("h5vcc-embedded"), does this
PR/CL respect that?
```

For two PRs, sure will do.

Regarding, "Chrome UI protocol ("h5vcc-embedded")", I saw it in the bug, but not sure how to consume it yet, maybe @borongc may provide more info around it.

cobalt/android/apk/app/src/main/assets/splash.html
```diff
@@ -0,0 +1,227 @@
+<!doctype html>
```
borongc, line 1: Putting this file as an Android asset made this solution only Android-specific, we should consider 3P platform as well. Same for `SplashScreenManager.java`.

linxinan-yt: The thing is splash requires fast response once the app spin up, any chromium aka platform agnostic solution may take extra time. So I think platform specific implementation is a trade off. 

The splash.html is a temp and handy solution for now, I guess. In long term, we should have a native API to consume it, probably.

borongc: This PR (https://github.com/youtube/cobalt/pull/7805) allows you to access the file in memory using `h5vcc-embedded://{filename}`, just put `{filename}` under `cobalt/shell/embedded_resources/` and add it to BUILD.gn, then you can access it both on 3P and Android platform.

This also bypasses CSP and HTTPs enforcement, as it is only local, so it is safe and won't cause any security issues.

linxinan-yt: Synced with @borongc, 

1. better not write any files to file asset for security
2. use PWA solution for platform agnostic

For 2, I can use the installable manager in my old https://github.com/youtube/cobalt/pull/7642 to fetch and parse manifest. But two issues need to figure out:

1. how to notify the java native side if the chromium cache is available.
2. hot to detect if the video file has updated and need to fetch again.

PR #7863, opened 2025-11-04
[third_party] Import ICU as owned dependency

kaidokert:
Import ICU package as owned code, to allow modifications.

Bug: 446966279


kaidokert:
Note this PR needs to be merged, not squashed when it lands

y4vor (approved):

oxve (approved):

PR #7865, opened 2025-11-04
[26.android] Revert `GetHeapSize` to only `used_heap_size`

fayev-g:
Upstream Chromium calculates `used_js_heap_size` by including `external_memory`. This is not backwards compatible with previous versions of Cobalt where it was only `used_heap_size`. Revert this for now and check canary metrics. Follow up with introducing a new API that takes into account `external_memory`: b/457773763

Bug: 456183913


yell0wd0g (commented):
Just a note to please try to add a prefix with the high level 
folder, e.g. in this case "blink: Revert ...". The 
generate-commit-message Gemini hook can suggest it too. Otherwise
this LGTM! Any plan for future maintenance, or are we keeping this difference with Chromium for the foreseeable future?

PR #7885, opened 2025-11-05
android: Skip MediaCodec.stop() on Android 11

kjyoun:
On some Android 11 devices, calling `MediaCodec.stop()` can trigger a native race condition and crash in `libstagefright` if an error occurs during the stop phase.

Bug: 369372033


sideb0ard (approved):
code deletion is always nice!

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -324,7 +324,6 @@ MediaCodecBridge::~MediaCodecBridge() {
   }
 
   JNIEnv* env = AttachCurrentThread();
-  Java_MediaCodecBridge_stop(env, j_media_codec_bridge_);
```
borongc, line 327: Looks safe, but I am not sure if this will cause other issues.

kjyoun: Theoretically, this should be OK.

My plan is to submit this PR to main, and, when cherry-picking to 26.android. add version guard this logic to apply only to Android 11 and below

```
if (android.os.Build.VERSION.SDK_INT <= android.os.Build.VERSION_CODES.R) {
    // This code will run on Android 11 ("Red Velvet Cake") and all older devices
}
```

borongc: cc @jasonzhangxx for more opinions on this.

osagie98:
LGTM, though the MediaCodec API documentation explicitly calls [stop() before release()](https://developer.android.com/reference/android/media/MediaCodec.html). If release() already does the needed work it would be good to ask the Android Media team to update their documentation.

starboard/android/shared/media_codec_bridge.cc
xiaomings: Consider adding the version guard to main too, and cherry-pick to 26.android with it.  Note that this may no longer apply to this code location per my other comment.

Technically for such a change we should guard it behind a feature.  However, given the current launch cadence and percentage it's ok to use the launch as the experiment vehicle if @johnxwork is ok with it, and we should monitor this closely.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -680,18 +680,6 @@ private int flush() {
...
-  @CalledByNative
-  private void stop() {
-    synchronized (mNativeBridgeLock) {
-      mNativeMediaCodecBridge = 0;
-    }
-    try {
-      mMediaCodec.get().stop();
-    } catch (Exception e) {
-      Log.e(TAG, "Failed to stop MediaCodec", e);
-    }
```
xiaomings, line 692: Consider keeping this function and the whole call chain, and just checking for the API level and not calling stop() for old ATV version.  In this case we'd keep the above logic to reset `mNativeMediaCodecBridge`, and have a chance to add a comment here describing our intention.

starboard/android/shared/media_codec_bridge.cc
kjyoun: changed the logic to call stop() conditionally

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
kjyoun: Updated the logic to call stop() conditionally,

kjyoun:
> LGTM, though the MediaCodec API documentation explicitly calls [stop() before release()](https://developer.android.com/reference/android/media/MediaCodec.html). If release() already does the needed work it would be good to ask the Android Media team to update their documentation.

Thanks for the finding. I though that we don't need to call stop() at all, before release. It's not the case.
Though release() implicilitly calls stop(), it is strongly recommended to call stop()

kjyoun:
Updated PR. PTAL

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -597,6 +597,16 @@ public static void createVideoMediaCodecBridge(
   }
 
   @CalledByNative
+  public void destroy() {
+    // We are calling stop() only on Android 12 and afterwards, since Android 11
+    // has a race condition regarding error happening during stop().
+    // See b/369372033 for details.
+    if (android.os.Build.VERSION.SDK_INT > android.os.Build.VERSION_CODES.R) {
```
xiaomings: Would it work if we revert all other changes in this PR, and move this check and the above comment to stop() below?  In this way there are less code change and the logic for newer devices are kept as is.

kjyoun: Technically, it would work, but I believe this PR enforces correct behavior more reliably.

We should only skip stop() when we are calling release() immediately after. If the app intends to call only stop(), we must ensure stop() (or an equivalent) is actually called; we cannot simply skip it in that case.

By removing Stop() from the MediaCodecBridge C++ interface and introducing destroy() in Java, we prevent stop() from being called in isolation from native code. The app can now only call destroy(), which will handle stop() and release() appropriately based on the Android version.

If you look at the logs in b/369372033#comment28, you can see we were previously making redundant stop() calls during destruction(stop() -> stop() -> release()). This change also cleans that up.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -597,6 +597,17 @@ public static void createVideoMediaCodecBridge(
   }
 
   @CalledByNative
+  public void destroy() {
+    // We skip calling stop() on Android 11, as this version has a race condition
+    // if an error occurs during stop(). See b/369372033 for details.
+    if (android.os.Build.VERSION.SDK_INT == android.os.Build.VERSION_CODES.R) {
```
yell0wd0g: Wouldn't this rather be
`if (android.os.Build.VERSION.SDK_INT <= android.os.Build.VERSION_CODES.R) {` 
?

So it works on all versions R (SDK 30, Android 11) and previous ?
IIUC correctly the bug (and crash) manifests on multiple versions
pre-11. 

Ah never mind I just re-read [comment 7](https://buganizer.corp.google.com/issues/369372033#comment7). Code
is OK as-is.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
xiaomings: The change would be local and minimal by keeping the stop() function and move this check to it, and this PR doesn't reset `mNativeMediaCodecBridge` for old ATV releases.

Changing the stop() sequence could have subtle side effects that we may not want to deal with now.  We should remove the redundant stop() later, when we have time to ensure all code paths work properly.

kjyoun: OK, changed this to minimal change to skip stop().

I may create a separate PR to do refactoring that I tried to do in this PR.

xiaomings (approved):

PR #7890, opened 2025-11-05
cleanup: Move WeakPtrFactory as the last class member

borongc:
Moves base::WeakPtrFactory instances to be the last member in their
respective classes. This ensures the factory is destructed first,
invalidating all associated weak pointers before other class members
are destroyed. This prevents potential use-after-free issues during
object teardown.

Issue: 458068496

### DIFF
--- a/media/starboard/starboard_cdm.h
+++ b/media/starboard/starboard_cdm.h
@@ -107,29 +107,6 @@ class MEDIA_EXPORT StarboardCdm : public ContentDecryptionModule,
     std::string ToString() const;
   };
 
-  // Default task runner.
-  scoped_refptr<base::SequencedTaskRunner> const task_runner_;
-
-  const SbDrmSystem sb_drm_;
-
-  int next_ticket_ = 0;
-  TicketToSessionUpdateRequestMap ticket_to_session_update_request_map_;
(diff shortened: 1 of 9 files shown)


media/starboard/starboard_cdm.h
```diff
@@ -192,6 +169,27 @@ class MEDIA_EXPORT StarboardCdm : public ContentDecryptionModule,
...
+  TicketToSessionUpdateMap ticket_to_session_update_map_;
+  TicketToServerCertificateUpdatedMap ticket_to_server_certificate_updated_map_;
+
+  CdmPromiseAdapter promises_;
+
+  std::list<std::string> session_list_;
+
+  CallbackRegistry<EventCB::RunType> event_callbacks_;
+
+  base::WeakPtrFactory<StarboardCdm> weak_factory_{this};
```
kjyoun, line 195: How about adding comment like  below to inform future-developer not to add variables below?

// NOTE: Do not add member variables after weak_factory_
//  It should be the first one destroyed among all members.

borongc: Done.

Awallky (approved):
LGTM! 
Thanks for doing this

PR #7895, opened 2025-11-05
Add stdbuf in runfile to work around log lines appearing out of order

oxve:
There is a bug in gtest starboardization causing output appearing out of order in CI. This is due to output buffering affecting the streams (stdout/err) differently.

Bug: 458084372

### DIFF
--- a/starboard/build/util/generate_runfile.py
+++ b/starboard/build/util/generate_runfile.py
@@ -12,6 +12,7 @@ import subprocess
 import sys
 
 command = [
+    'stdbuf', '-i0', '-o0', '-e0',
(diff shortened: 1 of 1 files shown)


starboard/build/util/generate_runfile.py
```diff
@@ -12,6 +12,7 @@
 import sys
 
 command = [
+    'stdbuf', '-i0', '-o0', '-e0',
```
niranjanyardi, line 15: this will fail for rdk.. right ?

niranjanyardi: im assuming youll submit the PR when RDK gets updated

niranjanyardi (approved):

PR #7904, opened 2025-11-05
[DNS][POC] expose SbMediaGetVideoBufferBudget to h5vcc

johnedocampo:
Bug: 416105795

### DIFF
--- a/media/starboard/BUILD.gn
+++ b/media/starboard/BUILD.gn
@@ -93,6 +93,8 @@ source_set("starboard") {
       "starboard_memory_allocator.h",
       "starboard_renderer.cc",
       "starboard_renderer.h",
+      "starboard_settings.cc",
+      "starboard_settings.h",
       "starboard_utils.cc",
       "starboard_utils.h",
     ]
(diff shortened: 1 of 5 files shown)


kjyoun:
How about setting specific value for decoder_bufffe_memory_info instead of creating h5vcc settings instance?

as you mentioned in https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:media/starboard/decoder_buffer_memory_info.h

I want to avoid exposing whole h5vcc settings as global singleton instance. 
I think that we depend on singletones too much

---

```
std::optional<int> g_video_buffer_limit_override;

void SetVideoDecoderBufferLimitBytes(int limit) {
  g_video_buffer_limit_override = limit;
}

int GetVideoDecoderBufferLimitBytes(VideoCodec codec,
                                    const gfx::Size& resolution,
                                    int bits_per_pixel) {
  if (g_video_buffer_limit_override) {
    return *g_video_buffer_limit_override;
  }
  return SbMediaGetVideoBufferBudget(MediaVideoCodecToSbMediaVideoCodec(codec),
                                     resolution.width(), resolution.height(),
                                     bits_per_pixel);
}
```

PR #7908, opened 2025-11-05
build: Update required branch protection checks

oxve:
This commit updates the list of required GitHub branch protection checks
to reflect changes in the CI/CD pipeline. The script now explicitly
includes relevant checks, moving from an exclusion model to an inclusion
model. Special release branches (26.android and 26.eap) are manually
added to the list of protected branches.

Bug: 452945577

### DIFF
--- a/cobalt/tools/update_required_branch_checks.py
+++ b/cobalt/tools/update_required_branch_checks.py
@@ -21,7 +21,7 @@ Requires PyGithub to run:
 import argparse
 from typing import List
 
-from github import Github
+import github
 
 # Issue a Personal Access Token with 'repo' permission on
 # https://github.com/settings/tokens.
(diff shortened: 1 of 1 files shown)


cobalt/tools/update_required_branch_checks.py
```diff
@@ -69,11 +65,13 @@ def get_protected_branches() -> List[str]:
   branches = ['main']
   for i in range(MINIMUM_LTS_RELEASE, LATEST_LTS_RELEASE + 1)[:-1]:
     branches.append(f'{i}.lts.1+')
+  branches.append('26.android')
```
andrewsavage1, line 68: nit: Is there a reason you don't want to just include this in the declaration of the list?

andrewsavage1 (approved):

PR #7928, opened 2025-11-06
Implement h5vcc authorization tracking api's for evergreen and stub out for appletv

niranjanyardi:
Issue: 445433945
Issue: 395650827

All h5vcc system API's for evergreen related functionality are implemented.
This prepares the codebase for future specific implementation on tvOS (AppleTV) platforms by adding conditional compilation guards for H5vccSystem's tracking authorization APIs.

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc
@@ -122,12 +122,17 @@ ScriptPromise H5vccSystem::requestTrackingAuthorization(
 }
 
 void H5vccSystem::OnRequestTrackingAuthorization(
-    ScriptPromiseResolver* resolver) {
+    ScriptPromiseResolver* resolver,
+    bool is_tracking_authorization_supported) {
+  if (is_tracking_authorization_supported) {
 #if BUILDFLAG(IS_IOS_TVOS)
-  // TODO - b/458160672: Reject when this fails.
-  NOTIMPLEMENTED();
+    // TODO - b/458160672: Reject when this fails.
+    NOTIMPLEMENTED();
 #endif
-  resolver->Resolve();
+    resolver->Resolve();
+  } else {
+    resolver->Reject();
+  }
 }
 
 void H5vccSystem::exit() {
--- a/cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
+++ b/cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
@@ -180,8 +180,10 @@ void H5vccSystemImpl::RequestTrackingAuthorization(
 #if BUILDFLAG(IS_IOS_TVOS)
   // TODO - b/458160672: Add implementation for AppleTV
   NOTIMPLEMENTED();
+  std::move(callback).Run(true);
+#else
+  std::move(callback).Run(false);
 #endif
-  std::move(callback).Run();
 }
 
 void H5vccSystemImpl::GetUserOnExitStrategy(
--- a/cobalt/browser/h5vcc_system/public/mojom/h5vcc_system.mojom
+++ b/cobalt/browser/h5vcc_system/public/mojom/h5vcc_system.mojom
@@ -34,7 +34,7 @@ interface H5vccSystem {
   GetTrackingAuthorizationStatus() => (string tracking_authorization_status);
   [Sync]
   GetTrackingAuthorizationStatusSync() => (string tracking_authorization_status);
-  RequestTrackingAuthorization() => ();
+  RequestTrackingAuthorization() => (bool is_tracking_authorization_supported);
 
   // Returns the platform-specific user on exit strategy.
   [Sync]
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.h
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.h
@@ -62,7 +62,7 @@ class MODULES_EXPORT H5vccSystem final
   void OnGetAdvertisingId(ScriptPromiseResolver*, const String&);
   void OnGetLimitAdTracking(ScriptPromiseResolver*, bool);
   void OnGetTrackingAuthorizationStatus(ScriptPromiseResolver*, const String&);
-  void OnRequestTrackingAuthorization(ScriptPromiseResolver*);
+  void OnRequestTrackingAuthorization(ScriptPromiseResolver*, bool);
   void EnsureReceiverIsBound();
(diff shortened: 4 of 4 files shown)


niranjanyardi:
FYI @fayev-g - added some todo's for appletv

niranjanyardi:
Build failures look unrelated:
ERROR at //components/update_client/update_checker.cc:47:11: Can't include this header from here.
#include "starboard/system.h"
          ^-----------------

andrewsavage1:
Let's update the title for this pr

third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc
```diff
@@ -123,7 +123,10 @@ ScriptPromise H5vccSystem::requestTrackingAuthorization(
 
 void H5vccSystem::OnRequestTrackingAuthorization(
     ScriptPromiseResolver* resolver) {
-  // TODO - b/395650827: Reject when this fails.
+#if BUILDFLAG(IS_IOS_TVOS)
+  // TODO - b/458160672: Reject when this fails.
+  NOTIMPLEMENTED();
+#endif
   resolver->Resolve();
```
hlwarriner: In c25 we reject the promise when the platform doesn't support the `requestTrackingAuthorization` web API: https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/h5vcc/h5vcc_system.cc#L157.

Looking at the web app code, I think we want to keep this behavior.

One way to do this would be to
* Change the `RequestTrackingAuthorization` Mojo method to reply with a boolean
* `H5vccSystemImpl::RequestTrackingAuthorization` would do `std::move(callback).Run(true)` for TVOS and `std::move(callback).Run(false)` otherwise
* `H5vccSystem::OnRequestTrackingAuthorization` would resolve the promise if the value is true and reject it if false.

But maybe you can think of a better way to implement it.

Lmk what you think.

fayev-g: There's actually an edge-case tvOS bug related to this because we don't correctly reject the promise if the tvOS API doesn't support it also (came up in YTTV launch): https://github.com/youtube/cobalt/blob/main/starboard/shared/uikit/ifa.mm#L89, b/424267768#comment8

I think your suggestion of being able to let the h5vcc system know when to resolve or reject is helpful. I haven't fixed this in main yet because we're still in the process of moving over tvOS files and getting them hooked up.

I don't know how much of this is in scope of this particular PR, but I can follow up with the tvOS-specific implementation.

cc: @rakuco @jkim-julie @Gyuyoung @abhijeetk

hlwarriner: I think it is helpful to have confirmation that that web app needs to know whether the feature is/isn't supported. Thanks!

niranjanyardi: Thanks @hlwarriner . Your suggestion makes sense :
```
One way to do this would be to

Change the RequestTrackingAuthorization Mojo method to reply with a boolean
H5vccSystemImpl::RequestTrackingAuthorization would do std::move(callback).Run(true) for TVOS and std::move(callback).Run(false) otherwise
H5vccSystem::OnRequestTrackingAuthorization would resolve the promise if the value is true and reject it if false.
```

Ill implement this 

+1 faye - this is great- seems like we are improving the API for apple

niranjanyardi:
> Thanks, Niranjan. These changes look good to me.
> 
> I think the only thing missing is web test coverage for `requestTrackingAuthorization`. Since we don't have plans to run these web tests on TVOS, it would probably be fine for now to just add a simple test under `third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-system/` that verifies the promise returned by the web API is rejected. I appreciate that you've had to deal with changing requirements around the web interface so if you prefer to get this PR in as is and add the test coverage in a follow-up, I'm totally fine with that.

added it to my follow up todo's: https://b.corp.google.com/issues/445433945#comment11
Ill submit this for now asa i ahve all approvals and would be good to get something in after a while

third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.cc
```diff
@@ -122,9 +122,17 @@ ScriptPromise H5vccSystem::requestTrackingAuthorization(
...
+    ScriptPromiseResolver* resolver,
+    bool is_tracking_authorization_supported) {
+  if (is_tracking_authorization_supported) {
+#if BUILDFLAG(IS_IOS_TVOS)
+    // TODO - b/458160672: Reject when this fails.
+    NOTIMPLEMENTED();
+#endif
+    resolver->Resolve();
+  } else {
+    resolver->Reject();
```
yell0wd0g, line 134: Just a note that `Reject()` should be reserved for errors:
https://www.w3.org/2001/tag/doc/promises-guide#rejections-should-be-exceptional

So my question is whether a user would interpret `Reject()`
correctly (API not supported as opposed to "error in 
retrieving authorization"). It's possible that this is just 
a legacy behaviour and I'm not finding that info in the IDL 
file?  The legacy behaviour did not `Reject()` IIUC in 
https://github.com/youtube/cobalt/blob/25.lts.stable/cobalt/h5vcc/h5vcc_system.cc

hlwarriner: I asked Niranjan to do this because I saw that this **_was_** the legacy behavior in c25: https://github.com/youtube/cobalt/pull/7928#discussion_r2499576247.

Please let me know if I misread the code.

hlwarriner: > Just a note that Reject() should be reserved for errors:
https://www.w3.org/2001/tag/doc/promises-guide#rejections-should-be-exceptional

That's interesting, thanks for sharing. I could see the argument that this fits as a good case for rejection, though, as "When it will be impossible to complete the requested task" - since here, the web app has requested something that the platform does not support. Wdyt?

niranjanyardi: Maybe we can add a reason for rejection ?
resolver->Reject("Platform not supported");  - will that be better ?

rakuco: > resolver->Reject("Platform not supported");

Just a side note that the usual idiom here would be something like `resolver->RejectWithDOMException(DOMExceptionCode::kNotSupportedError, message)`.

PR #7836, opened 2025-11-03
media: Pass H5vcc settings from GlobalFeatures to StarboardRenderer

kjyoun:
This PR plumbs H5vcc settings from `GlobalFeatures` in the browser process to `StarboardRenderer` in the renderer process.
 
It introduces a new Mojo interface, `cobalt::mojom::H5vccSettings`, to allow the renderer to synchronously query all settings from the browser during initialization.
  
`CobaltContentRendererClient` now queries these settings during `RenderThreadStarted` and passes them to `StarboardRenderer` via `RendererFactoryTraits` and `StarboardRendererConfig`.
 
Bug: 455661813

PR #7933, opened 2025-11-06
media: Switch between DecoderBufferAllocator and partition_alloc at renderer process

kjyoun:
This PR refactors the logic for switching between `DecoderBufferAllocator` and the default system allocator (PartitionAlloc).

Key changes:
1.  **Safe Runtime Switching:**
    *   When disabled via H5vcc settings ("Media.DisableAllocator"), `DecoderBufferAllocator` gracefully releases its pre-allocated memory pool back to the system once all outstanding buffers are freed.
    *   This ensures thread-safe transitions and prevents memory leaks or invalid access if the allocator is disabled while buffers are still in use.

2.  **Renderer Process Switching:**
    *   The switching logic is now executed in the renderer process, instead of `StarboardRenderer` in the GPU process.
    *   IPC path will be removed in #8056 

Bug: 456552294

### DIFF
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -156,13 +156,11 @@ const T* GetSettingValue(
 
 void ProcessH5vccSettings(
     const std::map<std::string, media::H5vccSettingValue>& settings) {
-  bool enable_allocator = true;
   if (auto* val = GetSettingValue<int64_t>(
-          settings, kH5vccSettingsKeyMediaDisableAllocator);
-      val && *val == 1) {
-    enable_allocator = false;
+          settings, kH5vccSettingsKeyMediaDisableAllocator)) {
+    bool disable_allocator = *val != 0;
+    media::DecoderBuffer::EnableAllocator(!disable_allocator);
   }
-  media::DecoderBuffer::EnableAllocator(enable_allocator);
 
   for (const auto& [setting_name, setting_value] : settings) {
     AppendSettingToSwitch(setting_name, setting_value);


media/starboard/starboard_renderer.cc
```diff
@@ -116,29 +118,48 @@ void ConfigureDecoderBufferAllocator(bool use_external_allocator) {
...
+        DecoderBuffer::Allocator::Set(nullptr);
+        g_external_allocator->reset();
+      }
     } else {
-      LOG(INFO) << "Keeping current default DecoderBufferAllocator.";
+      LOG(INFO)
+          << __func__
+          << " > Already using default allocator(partition_alloc). Keeping it.";
     }
   }
```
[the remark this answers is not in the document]
kjyoun: I would keep current structure. it's more intuitive

media/starboard/starboard_renderer.cc
```diff
@@ -119,25 +121,45 @@ void ConfigureDecoderBufferAllocator(bool use_external_allocator) {
...
-      LOG(INFO) << "Destroying DecoderBufferAllocator instance. Using "
-                   "default allocator from now on.";
-      // NOTE: The use_external_allocator flag, controlled by the YouTube
-      // experimentation tooling, changes when a new Kabuki app is loaded.
-      // A change in this flag signifies a new app load, making it safe
-      // to destroy the DecoderBufferAllocator from the previous session.
-      g_external_allocator->reset();
-      DecoderBuffer::Allocator::Set(nullptr);
+      auto allocator = static_cast<DecoderBufferAllocator*>(instance);
+      if (allocator->GetAllocatedMemory() != 0) {
```
xiaomings: There could be a few issues here:
1. During the experiment, some of the samples believed to be using PartitionAlloc are actually using the DecoderBufferAllocator, which leads to skewed metrics.  I am not too concerned about this.
2. There is a race condition, where this function returns 0, but the DecoderBufferAllocator is used on another thread to allocate some memory.  This will lead to a crash.  We should avoid this even if this isn't encountered in production for our existing usage patterns.
3. The DecoderBufferAllocator lives in the same process as h5vcc (i.e. renderer process), and StarboardRenderer lives in gpu process.  We shouldn't configure DecoderBufferAllocator in StarboardRenderer.  @borongc FYI.

For this particular setting, we may try to require that it can only be set once before any playbacks.  This would be extra tricky given the splash screen, where we have to ensure that the experiment code kicks in before that.

kjyoun: This is temporary scaffolding logic. Once the experiment is completed, we will remove this logic entirely.
  
re: 1. Agreed.

re: 2. I agree a race condition is theoretically possible, but I believe the chance is extremely small. The experimentation setting remains consistent once the player library is loaded. The only scenario where the allocator might switch is between the splash screen (not using H5 player) and the actual video. There is a considerable gap during this switch (at least 1s+ on Kirkwood).
  
I have added a if (allocator->GetAllocatedMemory() != 0) guard to further minimize this risk, though it does not  prevent the race condition, ocated, we will log a warning and refuse to switch, preventing the crash scenario.
  
re: 3. Implementing the ideal logic (signaling across processes to confirm memory release) is too complex for this short-term experiment on Cobalt 26. Given the constraints and the temporary nature of this code, I believe the current implementation with the added safety guard is an acceptable trade-off. 

WDYT?

---
[1]  I did several measurement. Gap is always  at least 1 sec or more on Kirkwood device
Considering Cobalt app has lots of things to do,  this gap would be even larger on low-end device.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -81,6 +84,43 @@ void BindHostReceiverWithValuation(mojo::GenericPendingReceiver receiver) {
...
+
+void ProcessH5vccSettings(
+    const std::map<std::string, media::H5vccSettingValue>& settings) {
+  bool use_allocator = true;
+  if (auto* val = GetSettingValue<int64_t>(
+          settings, kH5vccSettingsKeyMediaDisableAllocator);
+      val && *val == 1) {
+    use_allocator = false;
+  }
+  media::DecoderBuffer::UseAllocator(use_allocator);
```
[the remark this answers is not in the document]
kjyoun: it's hard to read

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -227,4 +227,23 @@ void DecoderBufferAllocator::TryFlushAllocationLog_Locked() {
 }
 #endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
 
+void DecoderBufferAllocator::SetEnabled(bool enabled) {
+  auto to_string = [](bool value) { return value ? "enabled" : "disabled"; };
```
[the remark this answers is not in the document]
kjyoun: Done

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -200,17 +240,9 @@ void CobaltContentRendererClient::GetStarboardRendererFactoryTraits(
...
-      } else if (value->is_int_value()) {
-        renderer_factory_traits->h5vcc_settings.emplace(key,
-                                                        value->get_int_value());
-      } else {
-        NOTREACHED();
-      }
-    }
+    auto h5vcc_settings = ParseH5vccSettings(std::move(settings));
+    ProcessH5vccSettings(h5vcc_settings);
+    renderer_factory_traits->h5vcc_settings = std::move(h5vcc_settings);
```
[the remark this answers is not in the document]
kjyoun: Done

media/starboard/starboard_renderer.cc
kjyoun: Moved the switching logic to renderer and made DecoderBufferAllocator free allocated memory when it's done

PTAL

cobalt/renderer/cobalt_content_renderer_client.cc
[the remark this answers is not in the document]
kjyoun: Done

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -81,6 +84,43 @@ void BindHostReceiverWithValuation(mojo::GenericPendingReceiver receiver) {
   content::RenderThread::Get()->BindHostReceiver(std::move(receiver));
 }
 
+std::map<std::string, media::H5vccSettingValue> ParseH5vccSettings(
```
xiaomings, line 131: There is similar processing merged in [PR 7964](https://github.com/youtube/cobalt/pull/7964/files#diff-e71945ae96a2b2b3fcb37e21d26b20fea195bdce73048fbf2d9c3645c1d2029cR101), consider rebasing and merging.

media/base/decoder_buffer.h
```diff
@@ -81,9 +81,8 @@ class MEDIA_EXPORT DecoderBuffer
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   class Allocator {
    public:
-    static Allocator* GetInstance();
```
xiaomings, line 84: The majority of the changes in this PR is a revert of https://github.com/youtube/cobalt/pull/7836/.  It would be easier to review if we put these reverted changes (not a complete revert as there are code in PR 7836 that we want to keep) in a separate change, that can be merged first.

cobalt/renderer/cobalt_content_renderer_client.cc
kjyoun: Done

media/base/decoder_buffer.h
kjyoun: I revert the commit to remove IPC path.

https://github.com/youtube/cobalt/pull/7933/commits/ffd33be3eb23846ca04ac45d05b735240dcb0fd9

Will make it into a separate PR, once this PR is merged.

kjyoun: Create #8056 for removal

cobalt/renderer/cobalt_content_renderer_client.cc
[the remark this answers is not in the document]
xiaomings: It can be
```
if (auto* val = GetSettingValue<int64_t>(
          settings, kH5vccSettingsKeyMediaDisableAllocator)) {
    const bool use_allocator = !*val;
    media::DecoderBuffer::UseAllocator(use_allocator);
}
```

This has a few benefits:
1. It handles values like 2 correctly.  While the app shouldn't set the value to 2, unless we explicitly reject any values other than 0 or 1, we should treat non-zero values as true, instead of only treating 1 as true.
2. It avoids calling `media::DecoderBuffer::UseAllocator()` when the setting value isn't set.

Note that we can even name the setting "Media.EnableAllocator", so we don't have to use the temporary variable `use_allocator`, as a non-negative setting is much easier to understand.

media/base/decoder_buffer.h
```diff
@@ -93,10 +92,13 @@ class MEDIA_EXPORT DecoderBuffer
     virtual int GetBufferPadding() const = 0;
     virtual base::TimeDelta GetBufferGarbageCollectionDurationThreshold()
         const = 0;
+    virtual void SetEnabled(bool enabled) = 0;
 
    protected:
     ~Allocator() {}
   };
+
+  static void UseAllocator(bool enabled);
```
xiaomings: Consider naming it `EnableAllocator()`?

media/base/decoder_buffer.h
xiaomings: Please feel free to keep this as is, as the PR looks good overall.

The main reason I was asking is, I am not sure how I can review the change without the revert part.

As I have to treat every changed line as a new change and check if they work.  With a separate revert change, I can simply scan that and more focused on the real change in this PR.  On the other hand, please let me know if there is a way on github that I can exclude the revert commit during reviewing.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -227,4 +227,24 @@ void DecoderBufferAllocator::TryFlushAllocationLog_Locked() {
...
+
+  base::AutoLock scoped_lock(mutex_);
+  if (enabled_ == enabled) {
+    return;
+  }
+
+  LOG(INFO) << "DecoderBufferAllocator::SetEnabled: " << to_string(enabled_)
+            << " -> " << to_string(enabled);
+  enabled_ = enabled;
+  if (!enabled_ && strategy_ && strategy_->GetAllocated() == 0) {
```
xiaomings, line 245: We should also add extra handling into `DecoderBufferAllocator::Free()`, to reset `strategy_` when `enabled_` is false and `strategy_->GetAllocated()` is 0.

media/base/decoder_buffer.h
kjyoun: Agree that the part to delete IPC path makes this PR bloated so that code review becomes difficult

I think that it's better to keep two separate PRs. Otherwise, PR is too large

media/base/decoder_buffer.h
```diff
@@ -81,9 +81,8 @@ class MEDIA_EXPORT DecoderBuffer
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   class Allocator {
    public:
-    static Allocator* GetInstance();
     static void Set(Allocator* allocator);
-
+
```
borongc: nit: extra spaces

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -30,11 +31,16 @@ namespace cobalt {
 
 namespace {
 
+const char kH5vccSettingsKeyMediaDisableAllocator[] = "Media.DisableAllocator";
+const char kH5vccSettingsKeyMediaVideoBufferSizeClampMb[] =
```
borongc, line 35: If this is moved here, we should clean up `cobalt_content_browser_client.cc`.

media/base/decoder_buffer.h
kjyoun: Done

media/base/decoder_buffer.h
kjyoun: Done

cobalt/renderer/cobalt_content_renderer_client.cc
kjyoun: I converted these raw strings into constants so that we can see all the setting keys in one place.                                       
                                                                                                                                            
Even if `cobalt_content_browser_client.cc `is the only other file that needs them, I think it's better to keep these strings centralized  rather than spread across the file.                                                                                                       
                                                                                                                                            
Let me know if you have any suggestions.

borongc: Right. You can ignore this as I thought `kH5vccSettingsKeyMediaVideoBufferSizeClampMb` was in `cobalt_content_browser_client.cc`.

cobalt/renderer/cobalt_content_renderer_client.cc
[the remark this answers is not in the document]
kjyoun: Since `DecoderBufferAllocator` is a critical component that is enabled by default, I want to be very careful about disabling it. I prefer  to disable it only when the condition is explicitly met (integer 1) rather than interpreting non-zero values loosely. I think it is better to be safe here.                                                                                                                   
                                                                                                                                            
Regarding the name, while `EnableAllocator` is readable, `DisableAllocator` better reflects the behavior: the allocator is enabled by default until we explicitly disable it. This ensures that if there is a JavaScript error or the flag is missing, we default to using `DecoderBufferAllocator`.

Regarding the logic: `UseAllocator()` modifies a static variable. If we skip calling it, the allocator will retain its state from the previous session (e.g., if it was enabled, it remains enabled). In short, skipping the call does not inherently disable the allocator

media/starboard/decoder_buffer_allocator.cc
kjyoun: Yes, you are right.

I forgot to add there. Done

cobalt/renderer/cobalt_content_renderer_client.cc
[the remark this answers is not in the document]
xiaomings: 1. I am ok if the implementation rejects a value of not 0 or 1 (e.g. if the value is 2) by returning false to the set function.  However, if we allows the web app to successfully call `h5vcc.settings.set('Media.DisableAllocator', 2)`, the result should be the same as calling `h5vcc.settings.set('Media.DisableAllocator', 1)` instead of being the same as calling `h5vcc.settings.set('Media.DisableAllocator', 0)`.
2. We can keep the name as is, as I don't have a strong preference here.
3. We are changing the default behavior of whether DecoderBuffer allocator is enabled.  Currently, whether the DecoderBuffer allocator is enabled is inside decoder_buffer.cc, and is default to enabled.  Assume someone changes the code inside decoder_buffer.cc so it's default to false, they would have to change the code here too, otherwise it will be set to true by this code, even if the web app doesn't try to set the setting.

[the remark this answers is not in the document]
kjyoun: Done

xiaomings (approved):

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -227,4 +229,24 @@ void DecoderBufferAllocator::TryFlushAllocationLog_Locked() {
...
+    LOG(INFO) << "Freed " << strategy_->GetCapacity()
+              << " bytes of media buffer pool since allocator is disabled.";
+    strategy_.reset();
+  }
```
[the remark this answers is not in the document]
kjyoun: I prefer keeping current structure, which does not touch existing codes.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -129,6 +128,45 @@ bool AppendSettingToSwitch(
   return true;
 }
 
+std::map<std::string, media::H5vccSettingValue> ParseH5vccSettings(
+    cobalt::mojom::SettingsPtr settings) {
+  std::map<std::string, media::H5vccSettingValue> h5vcc_settings;
+  for (auto& [key, value] : settings->settings) {
+    if (value->is_string_value()) {
+      h5vcc_settings.emplace(key, std::move(value->get_string_value()));
```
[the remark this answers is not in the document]
kjyoun: We don't use `value->get_string_value()` any more, hence I would stick with `std::move`, which means we give up the ownership.

johnxwork (approved):

PR #7943, opened 2025-11-06
starboard: Use default copy constructor for CreationParameters

kjyoun:
Replaces the explicit copy constructor implementation for `PlayerComponents::Factory::CreationParameters` with the defaulted one. The explicit implementation was performing a member-wise copy, which is exactly what the defaulted copy constructor does.

This removes the need to manually update the copy constructor when adding new members to `CreationParameters`, reducing the risk of future bugs.

Bug: 414009070


starboard/shared/starboard/player/filter/player_components.h
```diff
@@ -69,7 +69,7 @@ class PlayerComponents {
                          SbDecodeTargetGraphicsContextProvider*
                              decode_target_graphics_context_provider,
                          SbDrmSystem drm_system = kSbDrmSystemInvalid);
-      CreationParameters(const CreationParameters& that);
+      CreationParameters(const CreationParameters& that) = default;
```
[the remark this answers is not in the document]
kjyoun: good suggestion, though it's out of the scope of this PR

starboard/shared/starboard/player/filter/player_components.cc
```diff
@@ -129,16 +129,7 @@ PlayerComponents::Factory::CreationParameters::CreationParameters(
...
-  this->audio_stream_info_ = that.audio_stream_info_;
-  this->video_stream_info_ = that.video_stream_info_;
-  this->player_ = that.player_;
-  this->output_mode_ = that.output_mode_;
-  this->max_video_input_size_ = that.max_video_input_size_;
-  this->decode_target_graphics_context_provider_ =
-      that.decode_target_graphics_context_provider_;
-  this->drm_system_ = that.drm_system_;
-}
+    const CreationParameters& that) = default;
```
jasonzhangxx: nit: we can put it in header file.

kjyoun: sounds good

kjyoun: Done

PR #7944, opened 2025-11-06
android: Switch to main shell after splash ends

borongc:
Synchronize the transition from the splash screen to the main application
shell. Previously, the main shell could be shown based purely on a
timeout, leading to potential race conditions where the splash screen
had not fully completed or torn down.

This change modifies CobaltActivity to check if the splash app shell has
been destroyed before showing the main shell. The splash.html now
explicitly calls window.close() when its video finishes, allowing the
splash shell to be properly destroyed. A slight delay is added to
the main shell load timer to account for this.

Also updates the MediaSource VP9 codec string in splash.html for
improved capability detection. The default splash timeout in the
AndroidManifest is also adjusted.

Issue: 441738743

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -246,7 +246,7 @@ public abstract class CobaltActivity extends Activity {
                     synchronized(lock) {
                       if (isMainFrameLoaded == false) {
                         // Main app loaded in App shell, switch to it.
-                        Log.i(TAG, "main shell is loaded");
+                        Log.i(TAG, "NativeSplash: main shell is loaded with splash screen is finished: " + mShellManager.getSplashAppShell().isDestroyed() + ".");
                         isMainFrameLoaded = true;
                         mShellManager.showAppShell();
                       }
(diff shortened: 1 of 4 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -274,10 +274,21 @@ public void onWebContentsLoaded() {
                   @Override
                   public void run() {
                     synchronized(lock) {
+                      // If main is loaded, this is no-op due to it is switched
```
johnxwork, line 277: nit: The indentation is getting a little crazy

cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2
```diff
@@ -58,7 +58,7 @@
                   android:exported="true">
             <meta-data android:name="cobalt.APP_URL" android:value="https://www.youtube.com/tv"/>
             <meta-data android:name="cobalt.APP_SPLASH_URL" android:value="h5vcc-embedded://splash.html"/>
-            <meta-data android:name="cobalt.APP_SPLASH_TIMEOUT_MS" android:value="1500"/>
+            <meta-data android:name="cobalt.APP_SPLASH_TIMEOUT_MS" android:value="1000"/>
```
johnxwork: The fallback timeout above is still 1500

johnxwork: I thought the splash screen is longer than 1 second, which means your new code to check if splash screen finished will always be false?

johnxwork (approved):

cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2
borongc: Changed all timeout to 1.5s.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: This is aligned with L278?

kaidokert (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
kaidokert: seems correct, if unwieldy

borongc:
Closed this as this is handled by https://github.com/youtube/cobalt/pull/8128.

PR #7950, opened 2025-11-06
Fix missing file in android package step

arjungm:
Bug: 456537711

Manual cherrypick of #7903


oxve (approved):

arjungm:
This is the fix for the currently broken Android release jobs on main. I'm not able to submit this PR due to branch protections and the checks aren't valid here since its for the Borgcron side of things. Could someone with branch push privileges submit this on my behalf?

PR #7961, opened 2025-11-07
Revert "cleanup: Disable mipmap generation by default (#7871)"

yell0wd0g:
This reverts commit c73c23223a948d635931eb1a39143683f43efad4 
because it broke some cobalt builds, see:

Bug: 458658891

Original commit message ------------------------------------- 

This change disables mipmap generation in the main GPU process. The feature is not currently used, so this aims to reduce potential memory usage.

Initial testing did not show a clear memory reduction.

Bug: 452075779


aee-google (approved):

sherryzy:
Oops, I shouldn't have landed all the changes in hurry assuming they were safe. Will take a look why those DCHECK() fail.

PR #7964, opened 2025-11-07
media: Expose decoder video budget clamp to h5vcc

johnedocampo:
This PR adds the ability for us to experiment with reducing the size of the video DecoderBuffer pool. This is done through a new `memory_limit_clamp_` value in `SourceBufferStream`, which will clamp the value of `memory_limit_` if `memory_limit_` exceeds `memory_limit_clamp_`. The `memory_limit_clamp_` value is then exposed to an h5vcc setting, allowing us to set custom values for this clamp.

To get this implemented two major things were done:

1. In `cobalt_content_renderer_client.cc`, we query for the h5vcc value during `GetStarboardRendererFactoryTraits`. If it exists, we enable a new media switch `kMSEVideoBufferSizeLimitClampMb`, associating this switch with the set value from h5vcc.

2. When configuring a new video config, we set its memory limit to the smaller value between `SbMediaGetVideoBufferBudget` or our set clamp value, which is retrieved from the media switch.

Bug: 416105795
Bug: 460292554

### DIFF
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -30,13 +30,11 @@ namespace cobalt {
 
 namespace {
 
-constexpr auto kVideoBufferSizeClampSetting = "Media.UseVideoBufferSizeClamp";
-
 // Map that stores all current bindings of H5vcc settings to media switches.
 // If a setting has a corresponding switch, we will enable the switch with the
 // corresponding value.
-const base::flat_map<std::string, const char*> kSettingToSwitchMap = {
-    {kVideoBufferSizeClampSetting, switches::kMSEVideoBufferSizeLimitClampMb},
+const base::flat_map<std::string, const char*> kH5vccSettingToSwitchMap = {
+    {"Media.VideoBufferSizeClampMb", switches::kMSEVideoBufferSizeLimitClampMb},
 };
 
 // TODO(b/376542844): Eliminate the usage of hardcoded MIME string once we
@@ -94,20 +92,41 @@ void BindHostReceiverWithValuation(mojo::GenericPendingReceiver receiver) {
   content::RenderThread::Get()->BindHostReceiver(std::move(receiver));
 }
 
+// TODO: b/460292554 - This code is a tentative solution, and will be replaced
+// once base::Feature is fully supported.
+//
 // Append the h5vcc setting to the corresponding media switch, if such mapping
-// exists.
-void AppendSettingToSwitch(
-    const std::string& switch_name,
+// exists. H5vcc settings are either pass their value to a media switch for code
+// in /media to use, or are given to Starboard Renderer for direct usage.
+bool AppendSettingToSwitch(
+    const std::string& setting_name,
     const cobalt::mojom::SettingValuePtr& setting_value) {
-  if (setting_value->is_int_value()) {
-    base::CommandLine::ForCurrentProcess()->AppendSwitchASCII(
-        switch_name, base::NumberToString(setting_value->get_int_value()));
-    LOG(INFO) << "Applied command line switch: " << switch_name << " = "
-              << setting_value->get_int_value();
-  } else {
-    LOG(WARNING) << "Attempted to apply switch " << switch_name
-                 << " but the setting value was not an integer.";
+  auto it = kH5vccSettingToSwitchMap.find(setting_name);
+  if (it == kH5vccSettingToSwitchMap.end()) {
+    return false;
   }
+  std::string switch_name = it->second;
+  std::string setting_str;
+  switch (setting_value->which()) {
+    case cobalt::mojom::SettingValue::Tag::kStringValue: {
+      setting_str = setting_value->get_string_value();
+      break;
+    }
+    case cobalt::mojom::SettingValue::Tag::kIntValue: {
+      setting_str = base::NumberToString(setting_value->get_int_value());
+      break;
+    }
+    default: {
+      LOG(WARNING) << "Attempted to apply switch " << switch_name
+                   << " but the setting value was not an integer or string.";
+      return false;
+    }
+  }
+  base::CommandLine::ForCurrentProcess()->AppendSwitchASCII(switch_name,
+                                                            setting_str);
+  LOG(INFO) << "Applied command line switch: " << switch_name << " = "
+            << setting_str;
+  return true;
 }
 
 }  // namespace
@@ -229,20 +248,17 @@ void CobaltContentRendererClient::GetStarboardRendererFactoryTraits(
 
   cobalt::mojom::SettingsPtr settings;
   if (h5vcc_settings_remote_->GetSettings(&settings) && settings) {
-    // H5vcc settings are either given to Starboard Renderer for direct usage,
-    // or pass their value to a media switch for code in //media to use.
     for (auto& [key, value] : settings->settings) {
-      auto it = kSettingToSwitchMap.find(key);
-      if (it != kSettingToSwitchMap.end()) {
-        AppendSettingToSwitch(it->second, value);
-      } else if (value->is_string_value()) {
-        renderer_factory_traits->h5vcc_settings.emplace(
-            key, std::move(value->get_string_value()));
-      } else if (value->is_int_value()) {
-        renderer_factory_traits->h5vcc_settings.emplace(key,
-                                                        value->get_int_value());
-      } else {
-        NOTREACHED();
+      if (!AppendSettingToSwitch(key, value)) {
+        if (value->is_string_value()) {
+          renderer_factory_traits->h5vcc_settings.emplace(
+              key, std::move(value->get_string_value()));
+        } else if (value->is_int_value()) {
+          renderer_factory_traits->h5vcc_settings.emplace(
+              key, value->get_int_value());
+        } else {
+          NOTREACHED();
+        }
       }
     }
   }
--- a/media/filters/chunk_demuxer.h
+++ b/media/filters/chunk_demuxer.h
@@ -161,10 +161,10 @@ class MEDIA_EXPORT ChunkDemuxerStream : public DemuxerStream {
   void SetStreamMemoryLimit(size_t memory_limit);
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  // Sets the memory limit and memory limit clamp, in bytes, on the SourceBufferStream. If 
-  // the memory limit is greater than the |memory_limit_clamp|, the memory limit will be 
-  //set to the clamp value.
-  void SetAndApplyStreamMemoryLimitClamp(size_t memory_limit_clamp);
+  // Sets the memory limit and memory limit clamp, in bytes, on the SourceBufferStream. If
+  // the memory limit is greater than the |memory_limit_clamp|, the memory limit will be
+  // set to the clamp value.
+  void SetStreamMemoryLimitClamp(size_t memory_limit_clamp);
 #endif // BUILDFLAG(USE_STARBOARD_MEDIA)
 
   void SetLiveness(StreamLiveness liveness);
--- a/media/filters/source_buffer_state.cc
+++ b/media/filters/source_buffer_state.cc
@@ -969,7 +969,7 @@ void SourceBufferState::SetStreamMemoryLimits() {
         << "Custom video per-track SourceBuffer size clamp limit(MiB)="
         << (video_buf_size_limit_clamp / 1024 / 1024);
     for (const auto& it : video_streams_)
-      it.second->SetAndApplyStreamMemoryLimitClamp(video_buf_size_limit_clamp);
+      it.second->SetStreamMemoryLimitClamp(video_buf_size_limit_clamp);
   }
 #endif // BUILDFLAG(USE_STARBOARD_MEDIA)
 }
--- a/media/filters/source_buffer_stream.cc
+++ b/media/filters/source_buffer_stream.cc
@@ -1875,6 +1875,8 @@ bool SourceBufferStream::UpdateVideoConfig(const VideoDecoderConfig& config,
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   // Dynamically increase |memory_limit_| when video resolution goes up.
+  // Note: m138+ introduces new ways to set the stream memory limit.
+  // TODO: b/460460519 - Readjust this code after rebasing to m138+.
   memory_limit_ = std::max(
       memory_limit_,
       GetDemuxerStreamVideoMemoryLimit(Demuxer::DemuxerTypes::kChunkDemuxer,
--- a/media/filters/source_buffer_stream.h
+++ b/media/filters/source_buffer_stream.h
@@ -184,7 +184,7 @@ class MEDIA_EXPORT SourceBufferStream {
   }
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  void set_and_apply_memory_limit_clamp(size_t memory_limit_clamp) {
+  void set_memory_limit_clamp(size_t memory_limit_clamp) {
     memory_limit_clamp_ = memory_limit_clamp;
     memory_limit_ = std::min(memory_limit_, memory_limit_clamp);
   }
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -440,9 +440,9 @@ void ChunkDemuxerStream::SetStreamMemoryLimit(size_t memory_limit) {
 }
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-void ChunkDemuxerStream::SetAndApplyStreamMemoryLimitClamp(size_t memory_limit_clamp) {
+void ChunkDemuxerStream::SetStreamMemoryLimitClamp(size_t memory_limit_clamp) {
   base::AutoLock auto_lock(lock_);
-  stream_->set_and_apply_memory_limit_clamp(memory_limit_clamp);
+  stream_->set_memory_limit_clamp(memory_limit_clamp);
 }
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 


media/filters/source_buffer_stream.h
```diff
@@ -504,6 +514,13 @@ class MEDIA_EXPORT SourceBufferStream {
   // constructor.
   size_t memory_limit_;
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // The hard limit of possible data in bytes the stream will keep in memory.
+  // This value is only used if the switch |kMSEVideoBufferSizeLimitClampMb| is 
+  // used. If this switch is not enabled, this value is not used.
+  size_t memory_limit_clamp_;
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/chunk_demuxer.h
```diff
@@ -160,6 +160,10 @@ class MEDIA_EXPORT ChunkDemuxerStream : public DemuxerStream {
   // Sets the memory limit, in bytes, on the SourceBufferStream.
   void SetStreamMemoryLimit(size_t memory_limit);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  void SetStreamMemoryLimitClamp(size_t memory_limit_clamp);
+#endif // USE_STARBOARD_MEDIA
```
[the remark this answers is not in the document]
johnedocampo: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +106,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+
+  const cobalt::mojom::SettingValuePtr& setting_value = it->second;
+  if (setting_value->is_string_value()) {
+    command_line->AppendSwitchASCII(switches::kMSEVideoBufferSizeLimitClampMb,
+                                    setting_value->get_string_value());
+  } else {
+    LOG(WARNING)
+        << "H5vcc setting Media.UseVideoBufferSizeClamp was found, but its "
+        << "associated value was not a string. Make sure the value set is"
+        << " a string.";
```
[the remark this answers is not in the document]
johnedocampo: Done. Pre-commit reformatted this message.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +106,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
   }
+
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  cobalt::mojom::SettingsPtr settings;
+  if (!h5vcc_settings_remote_->GetSettings(&settings) || !settings) {
+    return;
+  }
+
+  auto it = settings->settings.find("Media.UseVideoBufferSizeClamp");
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/source_buffer_stream.cc
```diff
@@ -1879,6 +1880,14 @@ bool SourceBufferStream::UpdateVideoConfig(const VideoDecoderConfig& config,
       memory_limit_,
       GetDemuxerStreamVideoMemoryLimit(Demuxer::DemuxerTypes::kChunkDemuxer,
                                        &config));
+
+  // If the |memory_clamp| experiment is enabled, we ensure that the |memory_limit_|
+  // is not above the clamp.
+  const base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+  if (command_line->HasSwitch(switches::kMSEVideoBufferSizeLimitClampMb)) {
+    memory_limit_ = std::min(memory_limit_, memory_limit_clamp_);
+  }
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/source_buffer_state.cc
```diff
@@ -960,6 +960,18 @@ void SourceBufferState::SetStreamMemoryLimits() {
...
       it.second->SetStreamMemoryLimit(video_buf_size_limit);
   }
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+ size_t video_buf_size_limit_clamp =
+      GetMSEBufferSizeLimitIfExists(switches::kMSEVideoBufferSizeLimitClampMb);
+  if (video_buf_size_limit_clamp) {
+    MEDIA_LOG(INFO, media_log_)
+        << "Custom video per-track SourceBuffer size clamp limit="
+        << video_buf_size_limit;
```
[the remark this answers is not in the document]

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +109,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+  const cobalt::mojom::SettingValuePtr& setting_value = it->second;
+  if (setting_value->is_string_value()) {
+    command_line->AppendSwitchASCII(switches::kMSEVideoBufferSizeLimitClampMb,
+                                    setting_value->get_string_value());
+  } else {
+    LOG(WARNING) << "H5vcc setting" << kVideoBufferClampSettingName
+                 << " was found, but its "
+                 << "associated value was not a string. Make sure the value "
+                    "set is a string.";
+  }
```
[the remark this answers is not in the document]
johnedocampo: RenderFrameCreated is only called once, so this should be okay. Fixed comment issue.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -81,6 +83,9 @@ void BindHostReceiverWithValuation(mojo::GenericPendingReceiver receiver) {
   content::RenderThread::Get()->BindHostReceiver(std::move(receiver));
 }
 
+const std::string kVideoBufferClampSettingName =
+    "Media.UseVideoBufferSizeClamp";
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/source_buffer_stream.cc
```diff
@@ -10,6 +10,7 @@
 #include <sstream>
 #include <string>
 
+#include "base/command_line.h"
```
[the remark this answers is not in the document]
johnedocampo: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +108,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
     content::RenderThread::Get()->BindHostReceiver(
         h5vcc_settings_remote_.BindNewPipeAndPassReceiver());
   }
+
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  cobalt::mojom::SettingsPtr settings;
+  if (!h5vcc_settings_remote_->GetSettings(&settings) || !settings) {
+    return;
+  }
```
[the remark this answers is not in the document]
johnedocampo: As this is called only once at app startup before any video is played, this should be okay.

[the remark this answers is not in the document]
xiaomings: Sorry that I am not familiar with the process model here, and it's probably not in the scope of this PR.  The h5vcc settings is sent to the browser process from the renderer process, then SourceBuffer is retrieving it back from the browser process.  If h5vcc and SourceBuffer both live in the renderer process, is it common for them to exchange information without the browser process?

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +108,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+
+  auto it = settings->settings.find(kVideoBufferClampSettingName);
+  if (it == settings->settings.end()) {
+    return;
+  }
+
+  const cobalt::mojom::SettingValuePtr& setting_value = it->second;
+  if (setting_value->is_string_value()) {
```
xiaomings: Consider using the int type.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +108,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+    command_line->AppendSwitchASCII(switches::kMSEVideoBufferSizeLimitClampMb,
+                                    setting_value->get_string_value());
```
xiaomings: This probably works but not sure if it's a common way to pass settings around.  Is there any precedence in Chromium?

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done. We now check for an int type and convert it to a string if it exists.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +109,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
     content::RenderThread::Get()->BindHostReceiver(
         h5vcc_settings_remote_.BindNewPipeAndPassReceiver());
   }
+
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  cobalt::mojom::SettingsPtr settings;
```
borongc: Consider to put L115-L122 as a helper function, like `CheckIfH5vccAttributeExists`.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -104,6 +109,29 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+    return;
+  }
+  auto it = settings->settings.find(kVideoBufferSizeClampSetting);
+  if (it == settings->settings.end()) {
+    return;
+  }
+
+  const cobalt::mojom::SettingValuePtr& setting_value = it->second;
+  if (setting_value->is_int_value()) {
+    command_line->AppendSwitchASCII(
```
kjyoun: Does this API for this case?

Per Gemini[1], to be effective, this method should be called before a certain momont.

You can test it by loading custom player that I used for testing
It would set "Media.DisableExternalAllocator", but modifying the code to interpret it as the signal that you want, you can test timing is correct.

https://b.corp.google.com/issues/456552294#comment7

---
[1]

In Chromium, you technically *can* call `AppendSwitchASCII` on a `base::CommandLine` object at any moment if you have mutable access to it, but for it to be **effective**, it usually must be called **before** a specific event, depending on what you are trying to influence.

It is highly time-sensitive regarding initialization phases.

### 1. For the Current Process
If you are trying to influence the behavior of the *currently running* process by modifying `base::CommandLine::ForCurrentProcess()`:

* **Must happen early:** Most subsystems (Networking, Rendering, V8, `base::FeatureList`) read their switches strictly once during their initialization phase.
* **The cutoff event:** Usually, the absolute latest you can effectively add global switches is during early startup delegates like `ContentMainDelegate::PreSandboxStartup` or `ContentBrowserClient::AppendExtraCommandLineSwitches`.
* **Too late:** If you append a switch *after* the relevant subsystem has initialized, the `CommandLine` object will contain your new switch, but the subsystem will already be running with its original configuration and will ignore your change.

### 2. For Launching Subprocesses
If you are preparing to launch a *new* process (e.g., a Renderer, GPU, or Utility process):

* **The cutoff event:** You must append the switches before the actual launch call is made.
* **Where to do it:** typically done by overriding specific delegates designed for this purpose, such as:
    * `ContentBrowserClient::AppendExtraCommandLineSwitches` (influences all child processes).
    * `BrowserChildProcessHostDelegate::OnProcessLaunchCommandLine` (for specific child processes).

### Summary
While the `CommandLine` object itself doesn't lock you out from appending later, standard Chromium architecture requires you to append switches **before initialization** of the target subsystem or **before launch** of the target process.

cobalt/renderer/cobalt_content_renderer_client.cc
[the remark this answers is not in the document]
borongc: It seems like we don't implement h5vcc_settings.idl as what we did on C25 (https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/25.lts.1+:cobalt/h5vcc/h5vcc_settings.idl), and it's a Chromium global feature to h5vcc_settings_impl in https://github.com/youtube/cobalt/pull/7836.

cc @kjyoun @at-ninja

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Thanks for the loader, I was trying to test this myself but couldn't find a way to set a setting on startup.

Running the loader, it seems that the **h5vcc settings are loaded after** the renderer frame is initialized.
Looks like we'll have to find a spot later in the pipeline to query for this value. Maybe in media_factory.cc?

cc: @borongc

johnedocampo: Moved the code from `RenderFrameCreated` to `GetStarboardRendererFactoryTraits`. 

While it isn't ideal that `GetStarboardRendererFactoryTraits` loads the switches as well, I feel it would be overkill/redundant to add a new function that calls the settings again and iterate through each one to apply certain switches. This new function would also just be called right before `GetStarboardRendererFactoryTraits` is called, which would be redundant. 

I tested this using your loader app, and I can confirm that this method now works.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: From looking around the chromium code base, there does not seem to be a precedent for using `AppendSwitchASCII` to pass settings around. The closest I found was in chrome_content_browser_client.cc (https://github.com/youtube/cobalt/blob/26.android/chrome/browser/chrome_content_browser_client.cc#L1311 and https://github.com/youtube/cobalt/blob/26.android/chrome/browser/chrome_content_browser_client.cc#L2616). For the most part, `AppendSwitchASCII` is used for setting up test environments.

media/base/media_switches.cc
```diff
@@ -190,6 +190,13 @@ const char kForceVideoOverlays[] = "force-video-overlays";
 const char kMSEAudioBufferSizeLimitMb[] = "mse-audio-buffer-size-limit-mb";
 const char kMSEVideoBufferSizeLimitMb[] = "mse-video-buffer-size-limit-mb";
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+// Allows explicitly specifiying MSE video buffer size maximum as megabytes.
+// Any video buffer size greater than this value will be clamped down to the
+// associated switch value.
+const char kMSEVideoBufferSizeLimitClampMb[] = "mse-video-buffer-size-limit-clamp-mb";
```
kjyoun, line 197: Do we really need this switch?

How about adding a `SetBufferLimit` method to `source_buffer_stream.h` and calling it directly? I know it's a bit hacky, but I'm not sure the current solution is that clean either.

johnedocampo: I think I prefer to use the switch method, as the code that uses it already follows precedence in Chromium. In `SourceBufferState::SetStreamMemoryLimits()`, Chromium uses the switches `kMSEVideoBufferSizeLimitMb` to override the memory limit if the switch is active.

With this switch, we can easily just add additional code to  `SourceBufferState::SetStreamMemoryLimits()` that checks for our own custom added switch. Chromium already has built code like [GetMSEBufferSizeLimitIfExists](https://github.com/youtube/cobalt/blob/26.android/media/filters/source_buffer_state.cc#L64) that we can leverage.

With the `SetBufferLimit` method, we'd most likely have to add a static method to `SourceBufferStream` that would be called in `cobalt/renderer/cobalt_content_renderer_client` which sets the clamp value. This may get a bit more confusing to maintain since the renderer_client must now depends about the SourceBufferStream (though there may be a better way to implement this). WDYT?

borongc: Both ways work, but StarboardRenderer cannot call  `SourceBufferStream` and `ChunkDemuxerStream` directly, we have to expose it (with more LOCs).

media/filters/source_buffer_state.cc
```diff
@@ -960,6 +960,18 @@ void SourceBufferState::SetStreamMemoryLimits() {
     for (const auto& it : video_streams_)
       it.second->SetStreamMemoryLimit(video_buf_size_limit);
   }
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+ size_t video_buf_size_limit_clamp =
+      GetMSEBufferSizeLimitIfExists(switches::kMSEVideoBufferSizeLimitClampMb);
+  if (video_buf_size_limit_clamp) {
+    MEDIA_LOG(INFO, media_log_)
+        << "Custom video per-track SourceBuffer size clamp limit="
```
kjyoun: limit(MiB)=  (video_buf_limit_clamp_bytes/ 1024/1024)  for readability ?

media/filters/source_buffer_stream.h
```diff
@@ -183,6 +183,13 @@ class MEDIA_EXPORT SourceBufferStream {
     memory_limit_ = memory_limit;
   }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  void set_memory_limit_clamp(size_t memory_limit_clamp) {
+    memory_limit_clamp_ = memory_limit_clamp;
+    memory_limit_ = std::min(memory_limit_, memory_limit_clamp);
```
kjyoun, line 189: I think we don't need this logic

```
memory_limit_ = std::min(memory_limit_, memory_limit_clamp);
```

method name is `set_memory_limit_clamp`

media/filters/source_buffer_stream.h
```diff
@@ -504,6 +511,13 @@ class MEDIA_EXPORT SourceBufferStream {
   // constructor.
   size_t memory_limit_;
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // The hard limit of possible data in bytes the stream will keep in memory.
+  // This value is only used if the switch |kMSEVideoBufferSizeLimitClampMb| is 
+  // used. If this switch is not enabled, this value is not used.
+  size_t memory_limit_clamp_;
```
kjyoun: ctor always initialize this value to max

then why not set max as default value here, then you don't need to set it ctor.

media/filters/source_buffer_state.cc
kjyoun: limit(MiB) =" << (clamp / 1024 / 1024)  for log readability?

media/filters/source_buffer_stream.h
johnedocampo: Done.

media/filters/source_buffer_state.cc

media/filters/source_buffer_stream.h
johnedocampo: This logic is here as we still need to clamp the memory here. If we remove this logic, the memory limit won't actually be clamped when the video starts.

I do agree though that the method does not define this reasoning well. Updated the call to `set_and_apply_memory_limit_clamp`, along with `ChunkDemuxerStream::SetStreamMemoryLimitClamp` -> `ChunkDemuxerStream::SetAndApplyStreamMemoryLimitClamp` for clarity.

borongc:
We should have @xiaomings for approving this PR.

johnedocampo:
> We should have @xiaomings for approving this PR.

I wanted to try to get this in before the cutoff, but I agree. I don't think we need this too urgently. I'll try to get approval from @xiaomings.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -26,6 +30,15 @@ namespace cobalt {
 
 namespace {
 
+constexpr auto kVideoBufferSizeClampSetting = "Media.UseVideoBufferSizeClamp";
```
xiaomings: Consider removing this constant, and use the string literal directly in the `flat_map` below.

Also nit: `constexpr char kH5vccSettingVideoBufferSizeClamp[] = "Media.VideoBufferSizeClamp";`

1. It's a bit too much to use auto here.
2. `VideoBufferSizeClamp` is a better name than `UseVideoBufferSizeClamp`, as the setting is to set the size instead of enable a preset size.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -26,6 +30,15 @@ namespace cobalt {
...
+
+// Map that stores all current bindings of H5vcc settings to media switches.
+// If a setting has a corresponding switch, we will enable the switch with the
+// corresponding value.
+const base::flat_map<std::string, const char*> kSettingToSwitchMap = {
```
xiaomings: kSettingToSwitchMap => kH5vccSettingToSwitchMap

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -81,6 +94,22 @@ void BindHostReceiverWithValuation(mojo::GenericPendingReceiver receiver) {
...
+void AppendSettingToSwitch(
+    const std::string& switch_name,
+    const cobalt::mojom::SettingValuePtr& setting_value) {
+  if (setting_value->is_int_value()) {
+    base::CommandLine::ForCurrentProcess()->AppendSwitchASCII(
+        switch_name, base::NumberToString(setting_value->get_int_value()));
+    LOG(INFO) << "Applied command line switch: " << switch_name << " = "
+              << setting_value->get_int_value();
+  } else {
+    LOG(WARNING) << "Attempted to apply switch " << switch_name
```
xiaomings: We are using a map above, so we are trying to make this work for adding more parameters in future.  In that case, we should handle the string values too.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -200,8 +229,13 @@ void CobaltContentRendererClient::GetStarboardRendererFactoryTraits(
 
   cobalt::mojom::SettingsPtr settings;
   if (h5vcc_settings_remote_->GetSettings(&settings) && settings) {
+    // H5vcc settings are either given to Starboard Renderer for direct usage,
+    // or pass their value to a media switch for code in //media to use.
     for (auto& [key, value] : settings->settings) {
-      if (value->is_string_value()) {
+      auto it = kSettingToSwitchMap.find(key);
+      if (it != kSettingToSwitchMap.end()) {
+        AppendSettingToSwitch(it->second, value);
```
xiaomings: If we make `AppendSettingToSwitch()` return a bool, we would be able to move the check of kSettingToSwitchMap and the comment above into `AppendSettingToSwitch()`, and change the call here to
`if (AppendSettingToSwitch(...)) { ... } else { ... `

media/filters/chunk_demuxer.h
```diff
@@ -160,6 +160,13 @@ class MEDIA_EXPORT ChunkDemuxerStream : public DemuxerStream {
   // Sets the memory limit, in bytes, on the SourceBufferStream.
   void SetStreamMemoryLimit(size_t memory_limit);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Sets the memory limit and memory limit clamp, in bytes, on the SourceBufferStream. If 
+  // the memory limit is greater than the |memory_limit_clamp|, the memory limit will be 
+  //set to the clamp value.
+  void SetAndApplyStreamMemoryLimitClamp(size_t memory_limit_clamp);
```
xiaomings: Shall we just name it `SetStreamMemoryLimitClamp()`?

media/filters/chunk_demuxer.h
```diff
@@ -160,6 +160,13 @@ class MEDIA_EXPORT ChunkDemuxerStream : public DemuxerStream {
   // Sets the memory limit, in bytes, on the SourceBufferStream.
   void SetStreamMemoryLimit(size_t memory_limit);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Sets the memory limit and memory limit clamp, in bytes, on the SourceBufferStream. If 
+  // the memory limit is greater than the |memory_limit_clamp|, the memory limit will be 
+  //set to the clamp value.
```
xiaomings: nit: missing space.  Also, does the cl formatter work on the branch?

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -200,8 +229,13 @@ void CobaltContentRendererClient::GetStarboardRendererFactoryTraits(
 
   cobalt::mojom::SettingsPtr settings;
   if (h5vcc_settings_remote_->GetSettings(&settings) && settings) {
+    // H5vcc settings are either given to Starboard Renderer for direct usage,
+    // or pass their value to a media switch for code in //media to use.
```
xiaomings: Let's add a TODO with bug, explicitly mention that this is tentative code, which will be replaced by base::Feature soon.

cobalt/renderer/cobalt_content_renderer_client.cc
xiaomings: In that case we are creating a precedence of this usage, which probably isn't the best way.  It's ok if this is tentative and will be replaced with base::Feature soon, let's add a comment explicitly mention this.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done. Filed b/460292554.

media/filters/chunk_demuxer.h
johnedocampo: Done. Switched names back to SetStreamMemoryLimitClamp(). Also changed the function name in `source_buffer_stream.h` from `set_and_apply_memory_limit_clamp` to `set_memory_limit_clamp`.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done. Constant has been renamed to VideoBufferSizeClamp and has been moved to the flat_map directly.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done.

media/filters/chunk_demuxer.h
johnedocampo: Fixed.

It seems that the cl formatter only checks for files under `/cobalt` or `/starboard` (https://github.com/youtube/cobalt/blob/26.android/.pre-commit-config.yaml#L8). IIUC, this is because our cl formatter, `pre-commit`, is not the same as chromium's.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done. Also simplified the logic in `GetStarboardRendererFactoryTraits` due to `AppendSettingToSwitch()` change.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done. Filed b/460292554.

borongc: Add this bug id to commit messages.

johnedocampo: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -26,6 +30,14 @@ namespace cobalt {
 
 namespace {
 
+// Map that stores all current bindings of H5vcc settings to media switches.
+// If a setting has a corresponding switch, we will enable the switch with the
+// corresponding value.
+const base::flat_map<std::string, const char*> kH5vccSettingToSwitchMap = {
+    {"Media.UseVideoBufferSizeClamp",
```
borongc: I think `Media.SetVideoBufferSizeClamp` may be better to reflect the meaning here.

media/filters/source_buffer_stream.cc
```diff
@@ -1879,6 +1879,7 @@ bool SourceBufferStream::UpdateVideoConfig(const VideoDecoderConfig& config,
       memory_limit_,
       GetDemuxerStreamVideoMemoryLimit(Demuxer::DemuxerTypes::kChunkDemuxer,
                                        &config));
+  memory_limit_ = std::min(memory_limit_, memory_limit_clamp_);
```
borongc, line 1884: note as m138 (main branch) has different codes to handle this, add a comment here like "this should be readjusted after rebase to m138+".

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Changed. In an [earlier comment](https://github.com/youtube/cobalt/pull/7964#discussion_r2519470530) from @xiaomings, this variable is now `Media.VideoBufferSizeClamp`.

media/filters/source_buffer_stream.cc
johnedocampo: Done. Filed b/460460519 tracking this.

media/filters/source_buffer_stream.h
```diff
@@ -183,6 +183,13 @@ class MEDIA_EXPORT SourceBufferStream {
     memory_limit_ = memory_limit;
   }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  void set_memory_limit_clamp(size_t memory_limit_clamp) {
+    memory_limit_clamp_ = memory_limit_clamp;
+    memory_limit_ = std::min(memory_limit_, memory_limit_clamp);
+  }
```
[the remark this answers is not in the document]
johnedocampo: As this is a short term solution and there is no current plans to change when set_memory_limit is called, this should be okay.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -26,6 +30,13 @@ namespace cobalt {
 
 namespace {
 
+// Map that stores all current bindings of H5vcc settings to media switches.
+// If a setting has a corresponding switch, we will enable the switch with the
+// corresponding value.
+const base::flat_map<std::string, const char*> kH5vccSettingToSwitchMap = {
+    {"Media.VideoBufferSizeClamp", switches::kMSEVideoBufferSizeLimitClampMb},
```
xiaomings: If the size will be set in MBytes, let's name it as VideoBufferSizeClamp**Mb**.

Also wanted to double check that the path for using the setting has been verified.

cobalt/renderer/cobalt_content_renderer_client.cc
xiaomings: We can keep the current implementation as is as it's tentative, but this kind of complexity is one of the reasons I want to avoid using such a technique.

cobalt/renderer/cobalt_content_renderer_client.cc
johnedocampo: Done.

By path, do you mean the code to activate the h5vcc setting? If so, not yet.

xiaomings: To clarify, we should try to somehow set the setting (e.g. via dev tools) and verify that the implementation works (e.g. there is log message indicating that the clamp is set, and you may also observe the significant reduction of buffered health or frequent rebuffering when you set the clamp to a very small value).  While we can verify this in production, it's lower cost to sanity check locally now.

johnedocampo: Thank you for the clarification. Yes, I have verified that this setting is working when set from devtools.

PR #7969, opened 2025-11-10
[tvOS] Remove posix_emu/ directory

rakuco:
This directory includes a `pthread_setname_np()` implementation with a signature that matches glibc's and musls's (by taking 2 arguments instead of 1), and it came from C25.

Although it works as expected with Starboard code (once plugged into the build system), it breaks when called from Chromium code. Chromium code already calls `pthread_setname_np()` with the right number of arguments, which leads to a wrong comparison that always fails in this code (pthread_self() being compared to a `const char*`) and a lot of misleading error messages about failing to set the thread name being printed on the console.

Since most usages of `pthread_setname_np()` have been fixed to work with Apple's version in PR #7052 and the remaining case is being updated in this commit, we can drop this workaround altogether.

(posix_emu also has a memrchr.c that is being removed here. It came from C25 and is not used anywhere at the moment)

Bug: 432507888
Bug: 456692793

### DIFF
--- a/cobalt/testing/browser_tests/frame_tree_browsertest.cc
+++ b/cobalt/testing/browser_tests/frame_tree_browsertest.cc
@@ -126,12 +126,7 @@ IN_PROC_BROWSER_TEST_F(FrameTreeBrowserTest, MAYBE_FrameTreeShape) {
 // TODO(ajwong): Talk with nasko and merge this functionality with
 // FrameTreeShape.
 // TODO(b/437413685): Investigate test failure.
-#if BUILDFLAG(IS_ANDROID)
-#define MAYBE_FrameTreeShape2 FrameTreeShape2
-#else
-#define MAYBE_FrameTreeShape2 DISABLED_FrameTreeShape2
-#endif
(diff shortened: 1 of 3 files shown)


rakuco:

fayev-g:
Nice, thanks for the clean up! Was also wondering about this clean up when I moved the files over. posix_emu was removed in Android between c25 and c26 also.

starboard/darwin/posix_emu/include/pthread.h
fayev-g, line 1: @rakuco Should we remove the references in starboard/darwin/tvos/shared/BUILD.gn and starboard/darwin/tvos/shared/platform_configuration/BUILD.gn also?

rakuco: I can do it if you prefer. My original plan was to just remove these entire files once the files in `starboard/tvos` are plugged into the build.

fayev-g: Ok! Just wanted to make sure the references won't cause additional issues. lgtm

PR #4889, opened 2025-02-11
[linux] Support SbPlayer via punching a video hole frame

borongc:
Enable OverlayProcessorOzone with OverlayStrategyUnderlayStarboard to support punch hole display on Linux.

The SbPlayer works with punch-out mode. This is the same approach used on android for viz: https://github.com/youtube/cobalt/pull/4287.

b/394685755
b/395902954

PR #7970, opened 2025-11-10
Move video geometry setter code out of OverlayStrategyUnderlayStarboard

rakuco:
`OverlayStrategyUnderlayStarboard` is not used by tvOS, as the code path in Viz for Apple platforms is different and none of the OverlayProcessorUsingStrategy classes are used.

As a first step towards making hole punching work for tvOS, move the video geometry setter out of `OverlayStrategyUnderlayStarboard` and into a separate file that can be used by both the existing class as well as the upcoming tvOS code in Viz.

Bug: 459366150
Bug: 456692793

### DIFF
--- a/cobalt/build/configs/starboard.gni
+++ b/cobalt/build/configs/starboard.gni
@@ -20,6 +20,11 @@ declare_args() {
 declare_args() {
   # TODO: b/371241293 - Remove this variable when the hermetic toolchain is setup
   use_custom_libc = false
+
+  # TODO: b/460538637 - Deprecate this variable
+  # Whether this is an Evergreen compatible platform. A compatible platform
+  # can run the elf_loader and launch the Evergreen build.
+  sb_is_evergreen_compatible = is_starboard
 }
 
 # TODO: b/371241293 - Remove this variable when the hermetic toolchain is setup
--- a/starboard/build/config/base_configuration.gni
+++ b/starboard/build/config/base_configuration.gni
@@ -42,10 +42,6 @@ declare_args() {
   # Directory path to static contents' data.
   sb_static_contents_output_data_dir = "$root_out_dir/content/data"
 
-  # Whether this is an Evergreen compatible platform. A compatible platform
-  # can run the elf_loader and launch the Evergreen build.
-  sb_is_evergreen_compatible = false
-
   # Whether to use the libunwind library on Evergreen compatible platform.
   sb_evergreen_compatible_use_libunwind = false
 
--- a/starboard/crashpad_wrapper/wrapper.cc
+++ b/starboard/crashpad_wrapper/wrapper.cc
@@ -14,9 +14,8 @@
 
 #include "starboard/crashpad_wrapper/wrapper.h"
 
-#include <sys/stat.h>
-
 #include <string.h>
+#include <sys/stat.h>
 
 #include <map>
 #include <vector>
@@ -210,9 +209,6 @@ void RecordStatus(CrashpadInstallationStatus status) {
 }  // namespace
 
 void InstallCrashpadHandler(const std::string& ca_certificates_path) {
-#if BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
-  LOG(WARNING) << "ca_certificates_path is unused:" << ca_certificates_path;
-#endif  // BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
   ::crashpad::CrashpadClient* client = GetCrashpadClient();
 
   const base::FilePath handler_path = GetPathToCrashpadHandlerBinary();
@@ -261,10 +257,7 @@ void InstallCrashpadHandler(const std::string& ca_certificates_path) {
 
   if (!client->StartHandlerAtCrash(handler_path, database_directory_path,
                                    default_metrics_dir, kUploadUrl,
-// TODO: b/446932156 - pass Cobalt's CA store path to the handler.
-#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
-                                   ca_certificates_path,
-#endif  // !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
+                                   base::FilePath(ca_certificates_path.c_str()),
                                    default_annotations, default_arguments)) {
     LOG(ERROR) << "Failed to install the signal handler";
     RecordStatus(
--- a/starboard/linux/x64x11/shared/platform_configuration/configuration.gni
+++ b/starboard/linux/x64x11/shared/platform_configuration/configuration.gni
@@ -26,5 +26,4 @@ if (current_toolchain == default_toolchain &&
   sabi_path = "//starboard/sabi/x64/sysv/sabi-v$sb_api_version.json"
 }
 
-sb_is_evergreen_compatible = true
 use_crashpad = true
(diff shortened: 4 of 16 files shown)


rakuco:

borongc:
Is `OverlayProcessorMac` used for tvOS? If so, you can refer to this PR https://github.com/youtube/cobalt/pull/4889.

If tvOS uses `CALayerOverlayProcessor`, consider to find `DrawQuad::Material::kVideoHole` and make it transparent there to verify if punch-hole can work on tvOS before moving video geometry setter.

It is not required that `OverlayStrategyUnderlayStarboard` need to be used on tvOS, and you can define a mojo::Remote on `CALayerOverlayProcessor` as well.

rakuco:
> Is `OverlayProcessorMac` used for tvOS? If so, you can refer to this PR #4889.

It is, but I don't understand how that PR helps in this case. Could you elaborate?

borongc:
> It is, but I don't understand how that PR helps in this case. Could you elaborate?

Updated my comment above.

rakuco:
> If tvOS uses `CALayerOverlayProcessor`, consider to find `DrawQuad::Material::kVideoHole` and make it transparent there to verify if punch-hole can work on tvOS before moving video geometry setter.
> 
> It is not required that `OverlayStrategyUnderlayStarboard` need to be used on tvOS, and you can define a mojo::Remote on `CALayerOverlayProcessor` as well.

Yes, the code that makes `CALayerOverlayProcessor` work with punch-out mode is written and it basically works like that. This PR just lays out some groundwork so that the same video geometry setter code doesn't need to be duplicated for the 3rd time (tvOS, Starboard and Chromecast).

rakuco:
@borongc does the above make sense? The alternative I see is to implement this code again for tvOS and add an ifdef to `cobalt/gpu/cobalt_content_gpu_client.cc` to pick the right version, unless I'm missing something?

borongc:
> @borongc does the above make sense? The alternative I see is to implement this code again for tvOS and add an ifdef to `cobalt/gpu/cobalt_content_gpu_client.cc` to pick the right version, unless I'm missing something?

I don't have a clear picture how this will be with tvOS, although we agree that we may need to move `VideoGeometrySetter` out of `OverlayStrategyUnderlayStarboard`.

It would be good if we have a WIP PR to make punch-hole video working on tvOS first, so I can have a better picture how we should do for `VideoGeometrySetter`. As this is only used for setting the bounds of SbPlayer, I think we should have punch-hole video working first, then back to this PR.

Let me know if you think this doesn't make sense to you.

rakuco:
Sure, I've uploaded the rest of the work to my branch tvos/punchout-mode-chrobalt: https://github.com/youtube/cobalt/compare/main...rakuco:cobalt:tvos/punchout-mode-chrobalt. This is a simplified version because the changes to `starboard/tvos/shared/application_darwin.mm` require other PRs, but the core of the work in https://github.com/rakuco/cobalt/commit/3b072c0dd8a50606c37f798b8584e7271525e92f is ready. PTAL.

components/viz/service/BUILD.gn
```diff
@@ -429,6 +429,8 @@ viz_component("service") {
       sources += [
         "display/starboard/overlay_strategy_underlay_starboard.cc",
         "display/starboard/overlay_strategy_underlay_starboard.h",
+        "display/starboard/video_geometry_setter.cc",
```
borongc, line 430: As OverlayStrategyUnderlayStarboard is not used on tvOS at all, I feel like here if Cobalt is built on tvOS, we should exclude it.

Also, looking at https://github.com/youtube/cobalt/compare/main...rakuco:cobalt:tvos/punchout-mode-chrobalt, it seems like it is hacky to get `PutQuadInSeparateOverlay()` to be called. Have you tried to use `--force-video-overlays` on content shell on tvOS (https://source.chromium.org/chromium/chromium/src/+/main:media/base/media_switches.cc;l=146) to check how does chromium use overlay video on tvOS?

rakuco: > As OverlayStrategyUnderlayStarboard is not used on tvOS at all, I feel like here if Cobalt is built on tvOS, we should exclude it.

Done.

> Also, looking at https://github.com/youtube/cobalt/compare/main...rakuco:cobalt:tvos/punchout-mode-chrobalt, it seems like it is hacky to get PutQuadInSeparateOverlay() called. Have you tried to use --force-video-overlays on content shell on tvOS (https://source.chromium.org/chromium/chromium/src/+/main:media/base/media_switches.cc;l=146) to check how does chromium use overlay video on tvOS?

Hmm, I'm not sure what I was supposed to noticed with `--force-video-overlays` enabled.

As far as I can see, the code path that puts everything in one layer via `PutQuadInSeparateOverlay()` is similar to what C25 ended up doing.

In upstream Chromium, video playback is coordinated via https://source.chromium.org/chromium/chromium/src/+/main:ui/accelerated_widget_mac/ca_renderer_layer_tree.mm, with a (mostly up-to-date) design doc here: https://docs.google.com/document/d/1DtSN9zzvCF44_FQPM7ie01UxGHagQ66zfF5L9HnigQY/edit?usp=sharing, and it does something similar to decode to texture with multiple hierarchies of `CALayers` that make it really hard to integrate with punchout mode.

borongc: For Chromium content shell, I used the following command first to understand how overlay video works on ATV/Linux.

```
# ATV
$ out/Default/bin/content_shell_apk launch --args="--single-process --force-video-overlays" website_url

# Linux
$ out/Default/bin/content_shell launch --args="--single-process --force-video-overlays" website_url
```

As I am not sure how tvOS works on this, I am not sure what's the right way for this (or someone with expertise should help review these PRs).

rakuco: > For Chromium content shell, I used the following command first to understand how overlay video works on ATV/Linux.

Hmm, was I supposed to see anything different? I've tested an upstream Linux build and I don't see any difference when launching content_shell with or without `--force-video-overlays` and https://crosvideo.appspot.com, for example.

> As I am not sure how tvOS works on this, I am not sure what's the right way for this (or someone with expertise should help review these PRs).

I thought you were this someone :-) Is there anyone else who could also look at this? Maybe @jasonzhangxx?

borongc: Note for Linux, overlay video requires HW support. So I think even you use the flag, it is fallback to texture mode.

Android has MediaCodec (usually with HW decoder), so use that flag will see overlay video mode.

Not sure how this would be on tvOS / iOS / MacOS.

sherryzy: I'm doing a simple code search for this --force-video-overlays, and according to the code comment "// Force media player using SurfaceView instead of SurfaceTexture on Android.", this can be Android specific, and ignored by TvOS.

borongc: Right, discussed with @sherryzy offline. Refactoring `VideoGeometrySetter` looks good to me.

However, for tvOS part, we need @yell0wd0g for approval.

sherryzy (approved):

yell0wd0g (approved):
RS LGTM!

yell0wd0g:
@y4vor or @andrewsavage1  could you PTAL as starboard-owners? Thanks!

components/viz/service/BUILD.gn
rakuco: Thank you both!

PR #7971, opened 2025-11-10
tvos_common.gn: Enable use_dawn and skia_use_dawn.

rakuco:
These are disabled in common.gn, but the Apple build needs both to be on (iosurface_image_backing.mm depends on dawn_shared_texture_cache.cc, for example).

Bug: 433180638
Bug: 456692793


rakuco:

yell0wd0g (commented):
Post-submit LGTM! On tvOS IIUC we should use Skia on Dawn, 
Dawn on Metal. And Angle for GL-like workloads like e.g. 
WebGL.

PR #7973, opened 2025-11-10
[tvOS] Replace kSbInt64Max with std::numeric_limits equivalent

rakuco:
This applies #4331 to the tvOS code imported from C25.

Bug: 366064934
Bug: 456692793


rakuco:

fayev-g (approved):

PR #7977, opened 2025-11-10
cobalt: Apply JNI package prefix for Google3

zhongqiliang:
Apply a "cobalt" package prefix to generated JNI Java class references
when building Cobalt for the Google3 monorepo. This modifies the JNI
generation system to prepend "cobalt" to "org.chromium" package names,
ensuring proper class resolution for internal build variants like Kimono.
This change resolves Bug: 459562843 by aligning JNI lookups with the
expected internal packaging and prevents runtime class lookup failures.

Bug: 459562843

Change-Id: I00733bd40f2e31a61b2cfaa17188a82a1825e365

### DIFF
--- a/content/browser/android/app_web_message_port.cc
+++ b/content/browser/android/app_web_message_port.cc
@@ -43,10 +43,14 @@ base::android::ScopedJavaLocalRef<jobjectArray> CreateJavaMessagePort(
   JNIEnv* env = base::android::AttachCurrentThread();
   return base::android::ToTypedJavaArrayOfObjects(
       env, base::span(j_descriptors),
-#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_COBALT_ON_GOOGLE3)
+#if BUILDFLAG(IS_COBALT)
+  #if BUILDFLAG(IS_COBALT_ON_GOOGLE3)
       cobalt_org_chromium_content_browser_AppWebMessagePort_clazz(env));
-#else
+  #else
       org_chromium_content_browser_AppWebMessagePort_clazz(env));
+  #endif
+#else
+    org_chromium_content_browser_AppWebMessagePort_clazz(env));
 #endif
 }
 


third_party/jni_zero/jni_zero.gni
```diff
@@ -268,6 +277,10 @@ template("generate_jni_registration") {
     if (defined(is_cronet_build) && is_cronet_build) {
       args += _cronet_renaming_extra_args
     }
+
+    if (defined(is_cobalt_on_google3) && is_cobalt_on_google3) {
```
andrewsavage1: When is is_cobalt_on_google3 not defined? It wouldn't be a bad idea to conditionally define with (only when `is_cobalt = true`) but right now I don't think you do

zhongqiliang: It should be fixed, take a look again.

third_party/jni_zero/jni_zero.gni
```diff
@@ -268,6 +279,10 @@ template("generate_jni_registration") {
...
```
andrewsavage1: It would be better to check `is_cobalt && is_cobalt_on_google3`

third_party/jni_zero/jni_zero.gni
```diff
@@ -48,6 +51,14 @@ _cronet_renaming_extra_args = [
   "org.jni_zero",
 ]
 
+_cobalt_renaming_extra_args = [
```
andrewsavage1: Only define this when `is_cobalt && is_cobalt_on_google3`

third_party/jni_zero/jni_zero.gni
zhongqiliang: done

third_party/jni_zero/jni_zero.gni
zhongqiliang: done

content/browser/android/app_web_message_port.cc
```diff
@@ -22,6 +22,7 @@
 #include "third_party/blink/public/common/messaging/transferable_message_mojom_traits.h"
 #include "third_party/blink/public/common/messaging/web_message_port.h"
 #include "third_party/blink/public/mojom/messaging/transferable_message.mojom.h"
+#include "third_party/jni_zero/cobalt_for_google3_buildflags.h"
```
andrewsavage1: It's probably better to guard this and other changes in this file additionally with `BUILDFLAG(IS_COBALT)` (and the GN change in content/browser/BUILD.gn with `is_cobalt`)

zhongqiliang: When build Cobalt.apk, we want org_chromium_content_browser_AppWebMessagePort_clazz(env));
When build libchrobalt.so for Kimono apk, we want cobalt_org_chromium_content_browser_AppWebMessagePort_clazz(env));

zhongqiliang: I think IS_COBALT_ON_GOOGLE3 is the right flag to use here.

zhongqiliang: oh, you mean additionally with.

andrewsavage1 (approved):

PR #7979, opened 2025-11-10
Rebase cobalt_browsertests to M138

haozheng-cobalt:
This commit updates the cobalt_browsertests to align with Chromium M138. It integrates API changes, refactors test support classes, and re-enables previously disabled test components. This ensures the browser tests remain compatible and provide accurate coverage for the updated Chromium codebase.

Test: this change does not impact non-test code, and `obalt/testing:cobalt_browsertests` builds successfully. More PRs to truly bring up browsertests for running.
Bug: 458665232

### DIFF
--- a/cobalt/testing/browser_tests/resource_load_observer.cc
+++ b/cobalt/testing/browser_tests/resource_load_observer.cc
@@ -14,11 +14,13 @@
 
 #include "cobalt/testing/browser_tests/resource_load_observer.h"
 
+#include <optional>
 #include <string>
 #include <vector>
 
 #include "base/path_service.h"
 #include "base/run_loop.h"
+#include "cobalt/testing/browser_tests/browser/test_shell.h"
 #include "content/public/browser/render_frame_host.h"
 #include "content/public/common/content_paths.h"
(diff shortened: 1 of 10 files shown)


cobalt/testing/browser_tests/resource_load_observer.cc
```diff
@@ -41,9 +43,12 @@ ResourceLoadObserver::ResourceLoadEntry&
 ResourceLoadObserver::ResourceLoadEntry::operator=(
     ResourceLoadObserver::ResourceLoadEntry&&) = default;
 
-ResourceLoadObserver::ResourceLoadObserver(Shell* shell)
+ResourceLoadObserver::ResourceLoadObserver(TestShell* shell)
```
Awallky, line 46: nit: I don't think there is a need to change the method signature here, since the TestShell *is* a Shell object as well. It leaves the delta smaller.

haozheng-cobalt: Thanks for bringing this up, yes I agree, I'll bulk address the reference of Shell in another PR.

PR #7980, opened 2025-11-10
Enable optimize_for_size for 3P

y4vor:
Turn on the optimize_for_size wich for the most
part just enables -Os instead of -O2. The flag is on for Android but for some reason was not turned of for 3P. This optimization was also used in Cobalt 25.

The result is 14MB binary size reduction and ~ 10MB memory reduction.

 libcobalt.so 85MB -> 71MB
 libcobalt.lz4 52MB -> 43MB
 RSS home, signed out 135MB -> 124MB

Issue: 421171908

### DIFF
--- a/cobalt/build/configs/linux_common.gn
+++ b/cobalt/build/configs/linux_common.gn
@@ -113,3 +113,6 @@ icu_use_data_file = false
 
 # Disable the Pepper API (PPAPI) plugin support
 enable_ppapi = false
+
+# Enable the compiler size optimizations e.g. -Os.
+optimize_for_size = true
--- a/mojo/core/node_channel.cc
+++ b/mojo/core/node_channel.cc
@@ -907,9 +907,11 @@ void NodeChannel::InitializeLocalCapabilities() {
     return;
   }
 
+#if !BUILDFLAG(IS_STARBOARD)
   if (core::Channel::SupportsChannelUpgrade()) {
     SetLocalCapabilities(kNodeCapabilitySupportsUpgrade);
   }
+#endif
 }
 
 }  // namespace core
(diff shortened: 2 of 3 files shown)


cobalt/build/configs/linux_common.gn
```diff
@@ -113,3 +113,6 @@ icu_use_data_file = false
 
 # Disable the Pepper API (PPAPI) plugin support
 enable_ppapi = false
+
+# Enable the compiler size optimizations e.g. -Os.
+optimize_for_size = true
```
niranjanyardi, line 118: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-Os

"Optimize for size. -Os enables all -O2 optimizations except those that often increase code size"

Did you see any hit to performance ?  If not, seems safe to submit

niranjanyardi (approved):

cobalt/build/configs/linux_common.gn
y4vor: I didn't notice any regressions. I'm trying to run some of the t2t, r2r routines to get something more precise.

yell0wd0g, line 118: I couldn't find any more instances of `optimize_for_size` in this
folder (`main` and `26.eap` branches) (I'm guessing AndroidTV adds
it in the internal build?) Could/should we add this gn variable to
`common.gni` instead for all platforms to benefit from it?

y4vor: https://source.corp.google.com/h/github/youtube/cobalt/+/26.eap:build/config/compiler/compiler.gni;l=42?q=optimize_for_size&sq=r:github%2Fyoutube%2Fcobalt%20b:26.eap&start=11

PR #7988, opened 2025-11-11
Add ssl certificate build artifacts needed for running cobalt

niranjanyardi:
Issue: 458753458

This change includes the 'ssl' directory within Evergreen application
packages across various platforms. Missing these certificates would prevent Cobalt from
properly communicating with secure endpoints, impacting its core
functionality and ability to run.

### DIFF
--- a/cobalt/build/evergreen-arm-hardfp-raspi/package.json
+++ b/cobalt/build/evergreen-arm-hardfp-raspi/package.json
@@ -13,7 +13,8 @@
             ],
             "dirs": [
                 "app",
-                "fonts"
+                "fonts",
+                "ssl"
             ],
             "rename_files": [
                 {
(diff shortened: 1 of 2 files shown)


niranjanyardi:
> This makes sense so leaving an LGTM to unblock, assuming you take up Gemini's commit description.

thanks for the LGTM! i discussed this some more with andrew and found an issue with the rdk evergreen setup - ill revise this PR.

generally if i dont respond immediately on the comment - im waiting on feedback from a reviewer regarding the PR on a different issue. I'm happy to address comments but just keep in mind that if i dont explicitly add you on the PR sometuimes - I may be looking for feedback from someone else and may not reply to your comments until the original issue is resolved.

niranjanyardi:
closing this after discussion with andrew: https://github.com/youtube/cobalt/pull/8026

PR #7990, opened 2025-11-11
[tvOS] Make OutputModeSupported() return false for decode to texture

rakuco:
Decode to texture is broken in the main branch for tvOS and the feature is also being redesigned on other platforms, so make the port explicitly indicate that this mode is not supported when SbPlayer asks for it.

Bug: 447334535


rakuco:

fayev-g (approved):

PR #7991, opened 2025-11-11
BACKPORT: [tvOS] Use spatial navigation by default, supported by gesture recognizers

rakuco:
This is an M139 backport that is part of the series of patches that are necessary for navigation and input handling for tvOS.

Bug: 432824509

---

Original CL description:

This CL translates some swipe gestures into key presses so that users move focus up/left/right/down or click elements by, respectively, swiping up/left/right/down or tapping the remote control.

The translation process is not perfect and not all possible swiping gestures in the Siri remote are turned into sensible spatial navigation options, but it does allow switching focus and navigating in general web content.

Change-Id: I80051025696c3d288d59c54386d81618be4ba66f
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6630858
Cr-Commit-Position: refs/heads/main@{#1471774}

### DIFF
--- a/cobalt/shell/app/shell_main_delegate.cc
+++ b/cobalt/shell/app/shell_main_delegate.cc
@@ -15,6 +15,8 @@
 #include "cobalt/shell/app/shell_main_delegate.h"
 
 #include <iostream>
+#include <memory>
+#include <string>
 #include <tuple>
 #include <utility>
 #include <variant>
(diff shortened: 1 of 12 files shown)


rakuco:

andrewsavage1 (approved):

PR #7992, opened 2025-11-11
[tvOS] Remove search_controller_focus_delegate.h

rakuco:
This is part of the non-media code that is not used in Chrobalt. It is quite self-contained, and thus a good first candidate for removal.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #7993, opened 2025-11-11
[tvOS] Use explicit types with the "weak self" idiom

rakuco:
Instead of using `typeof(self)`, use explicit types to fix errors like

```
../../starboard/tvos/shared/game_controller_input_device.mm:31:10: error: expected unqualified-id
   31 |   __weak typeof(self) weakself = self;
      |          ^
../../starboard/tvos/shared/game_controller_input_device.mm:34:52: error: use of undeclared identifier 'weakself'
   34 |         SBDGameControllerInputDevice* strongself = weakself;
      |                                                    ^~~~~~~~
```

Bug: 456692793


rakuco:

PR #7997, opened 2025-11-11
Cherry pick PR #8570: Allow render to fetch splash video on h5vcc schema and move hard coded webm to the embedded file

linxinan-yt:
See original #8570:

Register the H5VCC URL schema to the fetch API at render side.

With the change, native splash implementation could call H5VCC to get the cached contents via fetch API.

In the Cherry pick, I have to refactor a new target for "//cobalt/shell:url_constants", as in M114 cobalt_shell_lib is still under "testonly = true", which blocks us to use the constants in production targets.

Issue: 454630524

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -66,9 +66,6 @@ public final class CommandLineOverrideHelper {
         // causes rendering artifacts when
         // low-end-device-mode is enabled.
         paramOverrides.add("--disable-rgba-4444-textures");
-        // Force CobaltActivity to use AndroidOverlay instead
-        // of VideoSurfaceView.
-        paramOverrides.add("--CobaltUsingAndroidOverlay");
 
         return paramOverrides;
     }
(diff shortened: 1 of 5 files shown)


cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -28,14 +34,66 @@
...
-      : client_(std::move(client)), url_(request.url) {
+      : client_(std::move(client)),
+        url_(request.url),
+        task_runner_(base::SequencedTaskRunner::GetCurrentDefault()) {
     std::string key = url_.host();
+    // TODO(b/454630524): Implement splash video fetcher to populate the cache.
+    if (base::EndsWith(url_.ExtractFileName(), ".webm",
+                       base::CompareCase::SENSITIVE)) {
+      base::FilePath user_data_dir;
+      if (base::PathService::Get(SHELL_DIR_USER_DATA, &user_data_dir)) {
```
linxinan-yt: @kaidokert this could be a path like "/data/user/0/dev.cobalt.coat/cache/" on Android. Are there any concerns to write/read the video as a file?

linxinan-yt: Thread is out of date.

johnxwork:
The CP only seems to add h5vcc embedded url scheme. Are you sure the CP is done right? The description seems to mention much more.

linxinan-yt:
> The CP only seems to add h5vcc embedded url scheme. Are you sure the CP is done right? The description seems to mention much more.

Yes. This CP only enables the h5vcc schema for render. The main change will be in the follow-up CP.

kaidokert (approved):

PR #8000, opened 2025-11-11
Migrate runner's build directory to /runner-cache

linxinan-yt:
This is shadowing PR following the sandbox [PR 815](https://github.com/youtube/cobalt_sandbox/pull/815). 
This change allows runners to mount /runner-cache from the host, thus the tool chain files and dependent repo fetched by gclient could be used for the next runner.

After this change, below files will be cached crossing runners on the same node:

- checked in code on cobalt/src: to save the artifacts downloaded by gclient runhook, mainly the tool chain files.
- CIPD cache
- source code fetched by gclient sync
As a results, the main checkout will move to ~/cobalt/src(a soft link to /runner-cache/cobalt) from ~/src.

Bug: 407051026

### DIFF
--- a/.github/actions/depot_tools/action.yaml
+++ b/.github/actions/depot_tools/action.yaml
@@ -8,15 +8,22 @@ runs:
   using: "composite"
   steps:
       - name: Get depot tools
-        run: git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+        run: |
+          if [ -d "/runner-cache/depot_tools" ]; then
+            echo "depot_tools directory found. Resetting repository..."
+            git -C /runner-cache/depot_tools reset --hard origin/main
+          else
+            echo "depot_tools directory not found. Cloning repository..."
+            git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git /runner-cache/depot_tools
+          fi
         shell: bash
       - name: Bootstrap bundled python3
         run: |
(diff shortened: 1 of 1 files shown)


oxve (approved):

rakuco:
I think this is causing weird CI issues:
* https://github.com/youtube/cobalt/actions/runs/19461161761/job/55685817480?pr=8092
* https://github.com/youtube/cobalt/actions/runs/19461161781/job/55685808714?pr=8092
* https://github.com/youtube/cobalt/actions/runs/19458825769/job/55678357591
* https://github.com/youtube/cobalt/actions/runs/19458825779/job/55678340885

linxinan-yt:
Hi @rakuco 

Thanks for reporting the issue.  First off, does rerun fix the issues? I want to understand if they are persistent or not. Breaking down the 4 links you shared, 

- https://github.com/youtube/cobalt/actions/runs/19458825769/job/55678357591 is due to the git lock. Probably because the previous run exit abnormally. I will file follow up PR to clean them.
- The rest are build errors due to third_party deps. Since the node has been deleted due to auto scale. It is very hard to dig now. I will search similar build errors in MTV biz hour and see if I could dig into the cached files.

rakuco:
> First off, does rerun fix the issues? I want to understand if they are persistent or not.

It generally does fix the issue. Note that CI has been flaky lately, so sometimes it means rerunning will trigger new issues, but at least the failures are not always the same.

PR #8001, opened 2025-11-11
[media m138] fix side data issue

sideb0ard:
DecoderBufferSideData was added as a formalized data structure in:
https://chromium-review.googlesource.com/c/chromium/src/+/4757645 

Since https://chromium-review.googlesource.com/c/chromium/src/+/5918375 an empty DecoderBufferSideData is also used to in an EOS to tell the decoder that it should flush its current state and prepare for a new configuration.

This PR adds an additional check to ensure the alpha_data vector has contents before copying.

Tested on Boreal, can trigger the DCHECK without this change at the end of an ad, and passes with the change.

Bug: 456780532

### DIFF
--- a/media/starboard/sbplayer_bridge.cc
+++ b/media/starboard/sbplayer_bridge.cc
@@ -948,7 +948,7 @@ void SbPlayerBridge::WriteBuffersInternal(
     sample_info.buffer_size = buffer->size();
     sample_info.timestamp = buffer->timestamp().InMicroseconds();
 
-    if (buffer->side_data()) {
+    if (buffer->side_data() && !buffer->side_data()->alpha_data.empty()) {
       // We only support at most one side data currently.
       side_data->type = kMatroskaBlockAdditional;
       side_data->data = buffer->side_data()->alpha_data.data();


media/starboard/sbplayer_bridge.cc
```diff
@@ -948,7 +948,7 @@ void SbPlayerBridge::WriteBuffersInternal(
     sample_info.buffer_size = buffer->size();
     sample_info.timestamp = buffer->timestamp().InMicroseconds();
 
-    if (buffer->side_data()) {
+    if (buffer->has_side_data()) {
```
kjyoun: ?

```
if (buffer->has_side_data() && buffer->side_data()) 
```

kjyoun:
super nit: On description, Bug id is usually put at the bottom.


It's required to call has_side_data() before accessing side_data()
DecoderBufferSideData was added as a formalized data structure in:
In https://chromium-review.googlesource.com/c/chromium/src/+/4757645 side


Bug: 456780532

media/starboard/sbplayer_bridge.cc
sideb0ard: turns out has_side_data had a short lived life! It was removed in https://chromium-review.googlesource.com/c/chromium/src/+/6011506 - though its still referenced in https://source.chromium.org/chromium/chromium/src/+/main:media/base/decoder_buffer.h;l=235?q=decoder_buffer.h&ss=chromium%2Fchromium%2Fsrc

media/starboard/sbplayer_bridge.cc
```diff
@@ -948,7 +948,7 @@ void SbPlayerBridge::WriteBuffersInternal(
     sample_info.buffer_size = buffer->size();
     sample_info.timestamp = buffer->timestamp().InMicroseconds();
 
-    if (buffer->side_data()) {
+    if (buffer->side_data() && buffer->side_data()->alpha_data.size() > 0) {
```
kjyoun: [alpha_data](https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/mojo/common/validation_utils.cc;l=47;drc=524126904340bf2104b6615d6bb6352ba6d7d844)


`&& !buffer->side_data()->alpha_data.empty()` ?

sideb0ard: oh nice! will add, yes!

sideb0ard: done!

kjyoun (approved):

PR #8003, opened 2025-11-11
android: Independent timeout for network error dialog

ckim1346:
The network error dialog is seeing some edge cases where it is not displaying (ie. failed DNS resolution) and thus causing a black screen. This is due to the fact the the `urlConnection` timeouts only begin once DNS lookup is finished. This PR adds an independent timeoutHandler that will display the error dialog if the check is not successful within 10 seconds, regardless of a successful/non-successful/hanging DNS connection.

Bug: 394176365
Bug: 457471731

### DIFF
--- a/cobalt/app/cobalt_main_delegate.cc
+++ b/cobalt/app/cobalt_main_delegate.cc
@@ -30,8 +30,7 @@
 
 namespace cobalt {
 
-CobaltMainDelegate::CobaltMainDelegate(bool is_content_browsertests)
-    : content::ShellMainDelegate(is_content_browsertests) {}
+CobaltMainDelegate::CobaltMainDelegate() : content::ShellMainDelegate() {}
 
 CobaltMainDelegate::~CobaltMainDelegate() {}
 
@@ -157,6 +156,7 @@ void CobaltMainDelegate::InitializeHangWatcher() {
   }
   const bool emit_crashes = false;
 
-  base::HangWatcher::InitializeOnMainThread(hang_watcher_process_type, emit_crashes);
+  base::HangWatcher::InitializeOnMainThread(hang_watcher_process_type,
+                                            emit_crashes);
 }
 }  // namespace cobalt
--- a/cobalt/app/cobalt_main_delegate.h
+++ b/cobalt/app/cobalt_main_delegate.h
@@ -29,7 +29,7 @@ namespace cobalt {
 
 class CobaltMainDelegate : public content::ShellMainDelegate {
  public:
-  explicit CobaltMainDelegate(bool is_content_browsertests = false);
+  explicit CobaltMainDelegate();
 
   CobaltMainDelegate(const CobaltMainDelegate&) = delete;
   CobaltMainDelegate& operator=(const CobaltMainDelegate&) = delete;
--- a/cobalt/build/android/package.json
+++ b/cobalt/build/android/package.json
@@ -2,6 +2,7 @@
     "archive_datas": [
         {
             "files": [
+                "apks/Cobalt.apk",
                 "cobalt_shell.pak",
                 "gen/build_info.json",
                 "gen/licenses_cobalt.spdx",
@@ -14,6 +15,10 @@
                 "obj/third_party/android_deps/chromium_play_services_availability_java.javac.jar"
             ],
             "rename_files": [
+                {
+                  "from_file": "apks/Cobalt.apk",
+                  "to_file": "Cobalt.apk"
+                },
                 {
                     "from_file": "gen/build_info.json",
                     "to_file": "build_info.json"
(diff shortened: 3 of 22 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -637,6 +645,23 @@ private void createNewSurfaceView() {
   // Try generate_204 with a timeout of 5 seconds to check for connectivity and raise a network
   // error dialog on an unsuccessful network check
   protected void activeNetworkCheck() {
+    if (timeoutRunnable != null) {
+      timeoutHandler.removeCallbacks(timeoutRunnable);
+    }
+    timeoutRunnable =
+      () -> {
+        Log.w(TAG, "Charley: Active Network check timed out after 5 seconds.");
```
borongc: nit: log fix

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -637,6 +645,23 @@ private void createNewSurfaceView() {
...
+        if (mPlatformError == null || !mPlatformError.isShowing()) {
+          mPlatformError =
+              new PlatformError(
+                  getStarboardBridge().getActivityHolder(), PlatformError.CONNECTION_ERROR, 0);
+          mPlatformError.raise();
+        }
+        mShouldReloadOnResume = true;
+        timeoutRunnable = null;
+      };
+    timeoutHandler.postDelayed(timeoutRunnable, 10000);
```
borongc: consider to put `10000` as a final int.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -637,6 +645,23 @@ private void createNewSurfaceView() {
   // Try generate_204 with a timeout of 5 seconds to check for connectivity and raise a network
   // error dialog on an unsuccessful network check
   protected void activeNetworkCheck() {
+    if (timeoutRunnable != null) {
```
borongc, line 650: Put comment here so people can know why this is needed.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: If this is timeout, should we cancel the following new Thread for network probe? As this is likely hang in `urlConnection.connect();` and the connection is never closed.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: Is there a reliable way to cancel the thread once it's started? To my knowledge something like `Thread.interrupt()` wouldn't work on the `HttpURLConnection`.

borongc: `urlConnection.disconnect();` should be able to allow thread to be un-blocked, and thread will be terminated once it is not blocked.

That is, I think `timeoutRunnable` can `urlConnection.disconnect();`, so thread will be un-blocked and terminated.

And, we can raise a platform error in this case. If we don't handle this, it is possible that the system has many open network connections by Cobalt, but they are not closed properly.

johnxwork:
Gemini Recommendation: For network operations, especially one as frequent as a network connectivity check, you should never use new Thread(). Instead, use an ExecutorService (like a ScheduledThreadPoolExecutor or FixedThreadPoolExecutor) to manage a fixed pool of worker threads. This limits resource usage and allows for thread reuse.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
johnxwork: The connection does have a timeout though so in theory it shouldn't hang indefinitely?

I need to catch tonight's release. If this PR is merged in time, shall we address this in a follow-up PR?

PR #8005, opened 2025-11-11
Delete test for non-definite error behavior for madvise

madhurajayaraman:
Bug: 461567547

Removing a test for an ambiguous behavior - the spec says we "may" return EINVAL, but it's not mandatory. Tests should be for definite conditions and behaviors.

### DIFF
--- a/starboard/nplb/posix_compliance/madvise_test.cc
+++ b/starboard/nplb/posix_compliance/madvise_test.cc
@@ -37,11 +37,6 @@ class MadviseTest : public ::testing::Test {
   int* memory;
 };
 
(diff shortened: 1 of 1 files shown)


andrewsavage1:
The change here is fine, but please make the title and description more descriptive

yell0wd0g:
Just a drive-by that you can get gemini to generate a commit message for you thanks to @zhongqiliang by adding, if I'm not mistaken:

/generate-commit-message

see e.g. here: https://github.com/youtube/cobalt/pull/8059#issuecomment-3535430232

andrewsavage1 (approved):

PR #8007, opened 2025-11-12
Cherry pick PR #7725: Fix Finch Safe Mode

yjzhang111:
Refer to the original PR: https://github.com/youtube/cobalt/pull/7725, resolved merge conflicts and removed unused parts. Patching the cherry pick PR manually as the original author Jammel has left the team and the cherry pick (https://github.com/youtube/cobalt/pull/7966) fails CLA/google requirement.

Bug: 452149713

Safe mode is a crucial mechanism for Cobalt to fall back to a "safe" or "default" config if it receives a faulty config that causes a crash.

If we crash 3 times, we fall back to the safe config. If we crash 4 times, we fall back to the default, or "empty" config.

This PR addresses a bug where Cobalt's safe mode was never activating because the underlying crash streak was never being incremented after a crash.

The root cause was a failure to signal the start of a new session to the metrics system. The CleanExitBeacon mechanism, which is responsible for detecting unclean shutdowns, must first be set to an "unclean" state at startup. This was not
happening. Consequently, if a crash occurred, the beacon from the previous clean shutdown was still present. On the next startup, the system would incorrectly read this old beacon, conclude the last session was clean, and therefore fail to increment the crash streak at all.

Main fixes overview:

Enabling crash streak incrementing
Bug: We were never marking the CleanExitBeacon as "unclean" at startup. Without this, a crash was indistinguishable from a clean shutdown from the beacon's perspective.
Fix: Added a call to metrics_state_manager_->LogHasSessionShutdownCleanly(false, false) in CobaltMetricsServiceClient::Initialize(). This marks the session as unclean at startup. Now, only a graceful shutdown that
calls LogHasSessionShutdownCleanly(true, ...) in the destructor can mark it as clean.
Reading crash streak from the right place

Bug: With the crash streak now being correctly incremented in metrics_local_state_, there was another bug: ExperimentConfigManager was reading the streak from the wrong file (experiment_config_).
Fix: Updated ExperimentConfigManager to read from metrics_local_state_.
Caveat: Synchronizing Variations Beacon and Metrics Config files

Problem: The system uses two files: the Variations beacon at startup and the Metrics Config at runtime. A race condition existed where resetting the streak to 0 after a config fetch would update the Metrics Config file
but not the Variations beacon. A subsequent crash would cause the stale, high value from the beacon to be used.
Fix: After resetting the streak to 0 in metrics_local_state_, explicitly call clean_exit_beacon()->WriteBeaconValue(true). This forces the beacon to re-read the now-zeroed streak from metrics_local_state_ and immediately write it to the
Variations file, ensuring both files are synchronized.
Caveat: Ensuring the crash streak reset is not lost on a crash

Problem: PrefService writes are batched and only guaranteed on a clean shutdown. If we reset the crash streak to 0 and then crashed, that reset would be lost.
Fix: Explicitly call CommitPendingWrite() on metrics_local_state_ after setting the crash streak to 0. This forces an immediate write to disk, making the reset durable.
PR tested by YTS safe mode tests here: [critique.corp.google.com/cl/825319049](https://critique.corp.google.com/cl/825319049)

### DIFF
--- a/cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
+++ b/cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
@@ -113,8 +113,8 @@ void H5vccExperimentsImpl::SetExperimentState(
       cobalt::kLatestConfigHash,
       std::move(
           experiment_config.Find(cobalt::kLatestConfigHash)->GetString()));
-  // TODO (b/442825834): Remove CommitPendingWrite to decrease storage writes
-  // TODO (b/456583508): Without CommitPendingWrite, we should still write to
+  // TODO: b/442825834 - Remove CommitPendingWrite to decrease storage writes
+  // TODO: b/456583508 - Without CommitPendingWrite, we should still write to
   // storage if we shutdown early
   global_features->metrics_local_state()->CommitPendingWrite();
   experiment_config_ptr->CommitPendingWrite();
@@ -126,8 +126,8 @@ void H5vccExperimentsImpl::ResetExperimentState(
   PrefService* experiment_config =
       cobalt::GlobalFeatures::GetInstance()->experiment_config();
   experiment_config->ClearPref(cobalt::kExperimentConfig);
-  // TODO (b/442825834): Remove CommitPendingWrite to decrease storage writes
-  // TODO (b/456583508): Without CommitPendingWrite, we should still write to
+  // TODO: b/442825834 - Remove CommitPendingWrite to decrease storage writes
+  // TODO: b/456583508 - Without CommitPendingWrite, we should still write to
   // storage if we shutdown early
   experiment_config->CommitPendingWrite();
   std::move(callback).Run();


cobalt/browser/h5vcc_experiments/h5vcc_experiments_impl.cc
```diff
@@ -91,9 +113,11 @@ void H5vccExperimentsImpl::SetExperimentState(
       cobalt::kLatestConfigHash,
       std::move(
           experiment_config.Find(cobalt::kLatestConfigHash)->GetString()));
-  // CommitPendingWrite not called here to avoid excessive disk writes.
-  // Features and featureParams won't be applied until the next Cobalt cold
-  // start so the delay is acceptable.
+  // TODO (b/442825834): Remove CommitPendingWrite to decrease storage writes
```
AlecsisDuarte: super-nit: Based on the cobalt coding style (go/cobalt-style#todo-format) the TODOs should have the following format: `// TODO: b/BUG_ID -`

PR #7792, opened 2025-10-30
BACKPORT: [tvos] Move kContentShellUserDataDir to shell_main_delegate.cc

rakuco:
Backport https://crrev.com/c/6607747 from @jkim-julie that is part of the work to get Cobalt/Content Shell to work correctly on tvOS.

Bug: 432824509

---

Original CL description:

This CL moves appending kContentShellUserDataDir to the command line to shell_main_delegate.cc to keep it at //content/shell.

Change-Id: I8bd84dddf9212ff4798bc793bb5ed984f9083d54 Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6607747 (cherry picked from commit e3f126fc994ee87aae0a674ba0c1830765033b8c)

PR #8008, opened 2025-11-12
[tvOS] cobalt shell: Move kContentShellUserDataDir to shell_main_delegate.cc

rakuco:
This is the Cobalt Shell version of #7792, setting `kContentShellUserDataDir` so that the right directory is used given the platform's cache directory limitations.

Original change by @jkim-julie.

Bug: 456692793

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,6 +7,7 @@
 /starboard/shared/starboard/drm             @youtube/cobalt-media
 /starboard/shared/starboard/media           @youtube/cobalt-media
 /starboard/shared/starboard/player          @youtube/cobalt-media
+/starboard/tvos/shared/media                @youtube/cobalt-media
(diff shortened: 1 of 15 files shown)


rakuco:

cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -208,6 +208,17 @@ std::optional<int> ShellMainDelegate::BasicStartupComplete() {
   }
 #endif
 
+#if BUILDFLAG(IS_IOS_TVOS)
```
Awallky, line 211: I would love for these preprocessor defines to instead be built into their own platform-specific version of the various classes, as Chromium's [C++ style guide](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++-dos-and-donts.md#minimizing-preprocessor-conditionals) suggests minimizing them, but I think that is outside of the scope of this change.

PR #8009, opened 2025-11-12
cobalt: Support custom UA strings via --user-agent in non-official builds

rakuco:
This ports https://crrev.com/c/6169001 to `CobaltContentBrowserClient`, allowing one to specify a custom user agent string in the command-line when `OFFICIAL_BUILD` is not defined.

Bug: 442760345

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,6 +7,7 @@
 /starboard/shared/starboard/drm             @youtube/cobalt-media
 /starboard/shared/starboard/media           @youtube/cobalt-media
 /starboard/shared/starboard/player          @youtube/cobalt-media
+/starboard/tvos/shared/media                @youtube/cobalt-media
 /third_party/blink/renderer/platform/media/ @youtube/cobalt-media
 
 /.github/ # no owners
@@ -26,6 +27,8 @@
 /starboard/common                          @youtube/cobalt-starboard-owners
 /components/viz/service/display/starboard/ @youtube/cobalt-starboard-owners
 /ui/ozone/platform/starboard/              @youtube/cobalt-starboard-owners
+
+/starboard/tvos @youtube/cobalt-tvos
 
 /cobalt/build/gn.py @youtube/cobalt-build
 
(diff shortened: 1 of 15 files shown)


rakuco:

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -87,9 +89,32 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
 constexpr base::FilePath::CharType kTrustTokenFilename[] =
     FILE_PATH_LITERAL("Trust Tokens");
 
+#if !defined(OFFICIAL_BUILD)
```
Awallky, line 92: Why do we need to re-define the method here when it already exists inside of [user_agent_utils.h](https://source.corp.google.com/h/github/youtube/cobalt/+/main:components/embedder_support/user_agent_utils.h;drc=524126904340bf2104b6615d6bb6352ba6d7d844;l=42)?

Can we just use it directly and then check the return value?

cobalt/browser/cobalt_content_browser_client.cc
Awallky, line 123: nit: I'm not sure if we will run into issues with the metadata slightly differing from the UA string, but it looks like content_shell does the same thing.

cobalt/browser/cobalt_content_browser_client.cc
rakuco: Heh ironically I was the one who started exporting that function upstream... I guess the version in this PR was based on M114, where I had to manually add some extra code to get it working.

rakuco:
Well it turns out this was actually already done for M138 in Cobalt Shell: https://github.com/youtube/cobalt/blob/67599662fdb99931151970c9e2b859a8c1beb012/cobalt/shell/browser/shell_content_browser_client.cc#L531, I just hadn't noticed it before. Closing.

rakuco:
Hmm, it turns out the existing implementation doesn't work. I've sent #8092 to fix it.

PR #8010, opened 2025-11-12
Add CODEOWNERS entries for tvOS

rakuco:
As suggested by @yell0wd0g, make `//starboard/tvos` be owned by the cobalt-tvos team, but make the media-specific directory be under cobalt-media's purview.

Fixed: 459669677

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -7,6 +7,7 @@
 /starboard/shared/starboard/drm             @youtube/cobalt-media
 /starboard/shared/starboard/media           @youtube/cobalt-media
 /starboard/shared/starboard/player          @youtube/cobalt-media
+/starboard/tvos/shared/media                @youtube/cobalt-media
 /third_party/blink/renderer/platform/media/ @youtube/cobalt-media
 
 /.github/ # no owners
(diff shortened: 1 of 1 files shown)


rakuco:

.github/CODEOWNERS
```diff
@@ -27,6 +27,9 @@
 /components/viz/service/display/starboard/ @youtube/cobalt-starboard-owners
 /ui/ozone/platform/starboard/              @youtube/cobalt-starboard-owners
 
+/starboard/tvos              @youtube/cobalt-tvos
```
yell0wd0g: QQ: What about //starboard/darwin? Or will that disappear?

.github/CODEOWNERS
```diff
@@ -27,6 +27,9 @@
...
+/starboard/tvos/shared/media @youtube/cobalt-media
```
yell0wd0g: The style chosen back then was to add this near l.9 .

.github/CODEOWNERS
rakuco: That one will disappear, hopefully Really Soon (TM). It's a C25 artifact and most of the contents are outdated or will be migrated to `//starboard/tvos`.

.github/CODEOWNERS
rakuco: Done!

PR #8012, opened 2025-11-12
[tvOS] Stop synthesizing SBDApplicationPlayer's duration property

rakuco:
Commit daecd125ee19df ("Remove SbMediaTime from Starboard interface") stop setting `_duration` anywhere, and `-duration` does not reference it anywhere, leading to the following warning:

    ./../starboard/tvos/shared/media/application_player.mm:615:1: warning: ivar '_duration' which backs the property is not referenced in this property's accessor [-Wunused-property-ivar]

Bug: 459507205
Bug: 456692793


rakuco:

PR #8013, opened 2025-11-12
[tvOS] Fix smaller issues in code imported from C25

rakuco:
Fix a bunch of erros and warnings in the code that came from C25. There are many small changes that would be too annoying to send in separate PRs:

- Fix some unsigned-to-signed comparisons
- Fix some narrowing (float to int) issues
- Avoid shadowing variables in `game_controller_input_device.mm`
- Remove some trailing or wrong semicolons in declarations
- Remove unused constants and variables
- Remove a break after a return statement in a switch clause

Bug: 459507205
Bug: 456692793


rakuco:

fayev-g (approved):

PR #8014, opened 2025-11-12
Use Cobalt CA root certs for crash report uploads

hlwarriner:
Use Cobalt CA root certs for crash report uploads
    
This is primarily to make Crashpad more portable and hermetic: we don't
want to rely on the system certificates for report uploads. This PR
ports this feature forward from c25.
    
Issue: 446932156

### DIFF
--- a/third_party/crashpad/crashpad/util/net/http_transport.cc
+++ b/third_party/crashpad/crashpad/util/net/http_transport.cc
@@ -55,7 +55,7 @@ void HTTPTransport::SetTimeout(double timeout) {
 
 #if BUILDFLAG(IS_NATIVE_TARGET_BUILD)
 void HTTPTransport::SetRootCACertificatesDirectoryPath(
-    const std::string& path) {
+    const base::FilePath& path) {
   root_ca_certificates_directory_path_ = path;
 }
 #else  // BUILDFLAG(IS_NATIVE_TARGET_BUILD)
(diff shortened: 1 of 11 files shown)


third_party/crashpad/crashpad/util/net/http_transport.cc
```diff
@@ -52,8 +53,16 @@ void HTTPTransport::SetTimeout(double timeout) {
   timeout_ = timeout;
 }
 
+#if BUILDFLAG(IS_NATIVE_TARGET_BUILD)
```
andrewsavage1, line 56: For crashpad util code, what's the right way to approach which BUILDFLAGs to use? Handler will check IS_NATIVE_TARGET and client will check IS_STARBOARD, and I generally expect util to OR the two, but I understand that here for example the code is only used in the handler.

Maybe one way to think about it is if the client ever needed to use certs, would we want it to similarly be able to set a directory or a single path? I assume the former, so I would lean towards adding `|| BUILDFLAG(IS_STARBOARD)` here, but I could be convinced otherwise

hlwarriner: Yeah, it's a good point and I've spent some time thinking about it when making these types of customizations. The general approach I've settled on is to make the customizations as narrowly as they're currently needed. I feel there are some benefits to this approach: the customizations can sometimes be simpler because they don't need to be as general; more of our builds end up aligning more closely with upstream builds. It's also just a simple rule that is easy to apply in practice, which is nice.

Applying that rule in this example means we only would make the customization for native target builds. But I think we have an even stronger case to do that here. Given the Crashpad client's core responsibilities - https://chromium.googlesource.com/crashpad/crashpad/+/HEAD/doc/overview_design.md#the-crashpad-client - it seems unlikely that it would ever need to send or receive anything over the network. 

Lmk how that sounds / if you want to discuss more.

PR #8023, opened 2025-11-13
android: refactor activeNetworkCheck() to use ScheduledThreadPool

ckim1346:
This follow-up CL to #8003 addresses closing threads that might have been left hanging with the newly implemented timeout. It refactors `activeNetworkCheck()` to use `ScheduledThreadPool` instead of the alternative of trying to interrupt or cancel them manually, which may or may not work due to the implementation of `HttpURLConnection`. It also sets a few methods on the URL request to make it more thorough and closer to Chromium's version.

Bug: 394176365
Bug: 457471731
Bug: 460232580


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -644,44 +647,37 @@ private void createNewSurfaceView() {
   }
 
   // Try to generate_204 with a timeout of 5 seconds to check for connectivity and raise a network
-  // error dialog on an unsuccessful network check
+  // error dialog on an unsuccessful network check. This functions runs 2 threads: one for the
+  // generate_204 check and one for a separate Future timeout check in the case that a connection
+  // can't be established ie. a DNS resolution hang.
   protected void activeNetworkCheck() {
```
borongc, line 653: As this function gets more complex, we should consider move this to a single Java file.

Chromium puts this as [ConnectivityDetector.java](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/net/connectivitydetector/ConnectivityDetector.java), and we should consider to use the same flavor.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -65,15 +67,15 @@
...
 import org.chromium.ui.base.ActivityWindowAndroid;
 import org.chromium.ui.base.IntentRequestTracker;
-import org.chromium.net.NetworkChangeNotifier;
 
 /** Native activity that has the required JNI methods called by the Starboard implementation. */
 public abstract class CobaltActivity extends Activity {
   private static final String URL_ARG = "--url=";
   private static final String META_DATA_APP_URL = "cobalt.APP_URL";
-  private static final int NETWORK_CHECK_TIMEOUT_MS = 10000;
+  private static final int NETWORK_CHECK_TIMEOUT_MS = 5000;
```
johnxwork, line 78: what's the reason to shorten the timeout?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -644,44 +647,37 @@ private void createNewSurfaceView() {
...
+            Log.w(TAG, "Primary connectivity check failed, trying fallback.");
+            probeURL = performSingleProbe("http://connectivitycheck.gstatic.com/generate_204");
           }
 
-          if (timeoutRunnable != null) {
-            timeoutHandler.removeCallbacks(timeoutRunnable);
-            timeoutRunnable = null;
+          if (!probeURL) {
+            // Throw an exception to trigger the error dialog logic
+            throw new IOException("Both primary and fallback connectivity checks failed.");
```
johnxwork, line 677: What's the reason to use exception instead of state variables here?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: See the discussion in b/457471731#comment37

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: Many of the url functions can possibly throw an IOException in the case that connection fails, times out, etc.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: Will definitely consider this maybe in a future ticket

PR #8024, opened 2025-11-13
Add rollback detection for Finch config

yjzhang111:
Bug: 447453844

This change introduces a mechanism to prevent the application of Finch experiment configurations when a Cobalt version rollback is detected. The current Cobalt version is now stored alongside the experiment configuration. If the recorded version in the configuration is newer than the running Cobalt version, the experiment config will be ignored to prevent features from being accidentally enabled or misconfigured by an incompatible configuration. This ensures better stability and prevents unexpected behavior after a downgrade.


cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -126,6 +127,18 @@ ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
...
     return ExperimentConfigType::kEmptyConfig;
   }
+
+  // Check if a rollback happened. If so, apply the empty config.
+  std::string recorded_cobalt_version =
+      use_safe_config
+          ? experiment_config_->GetString(kSafeConfigMinVersion)
+          : experiment_config_->GetString(kExperimentConfigMinVersion);
+
+  if (!recorded_cobalt_version.empty() &&
```
AlecsisDuarte, line 139: Just to confirm I understand, this means we default to empty config type if the last recorded cobalt version is older (less) than current device's cobalt version, right? Meaning if there was no cobalt version recorded for whatever reason we proceed to return the same config type? In which cases could we not have a cobalt version?

yjzhang111: Good catch - it should be the other way around. Default to empty config when recorded cobalt version is newer. Updated.

I added a comment about not having a recorded cobalt version - in theory Finch would only be sent to Cobalt version that has all the Finch functionalities (including this one) from the server. But this pref is added later than the rest so technically it is possible for a client to have the config without the "min version" pref if they're using a cobalt version somewhere in between the initial Finch implementation and this PR, and they somehow got a Finch config from the server.

AlecsisDuarte (approved):

cobalt/browser/experiments/experiment_config_manager.cc
AlecsisDuarte: Great, thanks for the context!

cobalt/browser/experiments/experiment_config_manager_unittest.cc
```diff
@@ -20,6 +20,7 @@
 #include "base/values.h"
 #include "cobalt/browser/constants/cobalt_experiment_names.h"
 #include "cobalt/browser/features.h"
+#include "cobalt/version.h"
```
Awallky, line 23: While I think the unittests here are fairly thorough, I think this would really benefit from using `cobalt/testing/browser_tests`, which is a port of `content_browsertests`. I can't find any direct comparison to this test in `//chrome` or content browser tests, but I think we should at least file a bug to add a browser test.

Awallky: *as a follow-up

Awallky (approved):

PR #8025, opened 2025-11-13
Workaround to disable the updater for RDK

yuying-y:
Use ENABLE_COBALT_HERMETIC_HACKS to guard the updater initialization temporarily, because it works on Linux and Raspi-2, but causes a crash on RDK.

Bug: 431862767

### DIFF
--- a/cobalt/shell/browser/shell.cc
+++ b/cobalt/shell/browser/shell.cc
@@ -142,6 +142,9 @@ void Shell::FinishShellInitialization(Shell* shell) {
   if (raw_web_contents->GetPrimaryMainFrame()->IsRenderFrameLive()) {
     GetPlatform()->MainFrameCreated(shell);
   }
+
+// TODO(b/460219182): Remove this once the updater works on RDK.
+#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
 #if BUILDFLAG(USE_EVERGREEN)
   // Create the updater module
   auto* storage_partition =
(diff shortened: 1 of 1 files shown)


cobalt/shell/browser/shell.cc
```diff
@@ -142,17 +142,20 @@ void Shell::FinishShellInitialization(Shell* shell) {
...
-      raw_web_contents->GetPrimaryMainFrame()->GetStoragePartition();
-  if (storage_partition) {
-    LOG(INFO) << "Creating updater_module_ for Shell";
-    shell->updater_module_.reset(new cobalt::updater::UpdaterModule(
-        storage_partition->GetURLLoaderFactoryForBrowserProcess(),
-        cobalt::updater::kDefaultUpdateCheckDelay));
-  }
-#endif
+
+#if BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
```
y4vor: I think you can do !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) and keep the code uncommented.

niranjanyardi: you can uncomment below code and just say #if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)

commenting the code seems a bit odd. @andrewsavage1 may know of a better flag to use here

niranjanyardi: Also good to add a TODO here with the bug to fix it

niranjanyardi:
Also i generally use the following command for a descriptive commit message. Since this is pretty straightforward upto you
to use it

/generate-commit-message

niranjanyardi (approved):

y4vor (approved):

PR #8026, opened 2025-11-13
Set sb_is_evergreen_compatible=is_starboard

niranjanyardi:
Issue: 458753458

Move the default definition of sb_is_evergreen_compatible to the
Cobalt build configuration. This change centralizes how Evergreen
compatibility is determined for Starboard platforms.

The flag now defaults to 'is_starboard', aligning with the intent
that most Starboard platforms support Evergreen. This simplifies
platform-specific GN files by removing redundant explicit
assignments.

### DIFF
--- a/cobalt/build/configs/starboard.gni
+++ b/cobalt/build/configs/starboard.gni
@@ -21,6 +21,7 @@ declare_args() {
   # TODO: b/371241293 - Remove this variable when the hermetic toolchain is setup
   use_custom_libc = false
 
+  # TODO: b/460538637 - Deprecate this variable
   # Whether this is an Evergreen compatible platform. A compatible platform
   # can run the elf_loader and launch the Evergreen build.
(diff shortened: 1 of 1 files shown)


niranjanyardi:
blocked until https://cobalt-review.git.corp.google.com/c/external/components/generic/cobalt/+/7330 goes in

cobalt/build/configs/starboard.gni
```diff
@@ -20,6 +20,10 @@ declare_args() {
 declare_args() {
   # TODO: b/371241293 - Remove this variable when the hermetic toolchain is setup
   use_custom_libc = false
+
+  # Whether this is an Evergreen compatible platform. A compatible platform
```
andrewsavage1, line 25: Let's add a TODO to remove this variable entirely too

niranjanyardi: Done

PR #8028, opened 2025-11-13
Fixup: Address is_debug=false issue for exclude_chrome_build

yunsik-jung:
This patch resolves issues in the previous patch when building with is_debug=false and exclude_chrome_build=true for lightweight Linux builds.


Youngsoo (commented):
LGTM

PR #8029, opened 2025-11-13
[tvOS] Do not synthesize SBDApplicationPlayer.playbackRate

rakuco:
This fixes the following warning

    ../../starboard/tvos/shared/media/application_player.mm:519:1: warning: ivar '_playbackRate' which backs the property is not referenced in this property's accessor [-Wunused-property-ivar]

`_playbackRate` is not read anywhere, so we can just stop creating it altogether.

Bug: 459507205
Bug: 456692793


rakuco:

fayev-g (approved):
Thanks for cleaning up even these small things. It just makes things make sense :)

PR #6634, opened 2025-07-30
Remove pthread::ThreadSetPriority() in favor of SbThreadSetPriority()

rakuco:
`starboard::shared::pthread::ThreadSetPriority()` was added in #3174; when `SbThreadSetPriority()` was introduced in #3299 none of the callers of `starboard::shared::pthread::ThreadSetPriority()` were converted.

This PR performs this conversion and, in another commit, removes `starboard::shared::pthread::ThreadSetPriority()` since it serves the same purpose.

Bug: 302335657

PR #8030, opened 2025-11-13
[tvOS] Adapt to #6634 (removal of pthread::ThreadSetPriority())

rakuco:
Make the tvOS work after #6634 removed pthread::ThreadSetPriority() in favor of calls to SbThreadSetPriority():

- Update call sites
- Adapt the tvOS implementation to the new API

Bug: 459507205
Bug: 456692793


rakuco:

PR #8031, opened 2025-11-13
[tvOS] Flatten namespaces in starboard/tvos

rakuco:
Adapt to what PRs like #7283, #7224 and others have done, and get rid of the "shared::uikit" namespaces internally.

Bug: 456692793
Bug: 441955897


rakuco:

PR #8033, opened 2025-11-13
[tvOS] Port code away from starboard::{Mutex,ConditionVariable}

rakuco:
Use std::condition_variable and std::mutex since the starboard types have been removed in #6894.

Most of the porting was trivial, except for audio_queue_audio_sink_type.mm which uses a mutex and two condition variables for bi-directional communication between threads.

Bug: 390503926
Bug: 456692793
Bug: 441955897

### DIFF
--- a/starboard/tvos/shared/media/audio_queue_audio_sink_type.mm
+++ b/starboard/tvos/shared/media/audio_queue_audio_sink_type.mm
@@ -448,7 +448,7 @@ void TvosAudioSinkType::Destroy(SbAudioSink audio_sink) {
   audio_thread_condition_.notify_one();
   {
     std::unique_lock lock(audio_thread_mutex_);
-    destroy_condition_.wait(lock, [&]() { return sinks_to_destroy_.empty(); });
+    destroy_condition_.wait(lock, [&] { return sinks_to_destroy_.empty(); });
   }
   delete audio_sink;
 }


rakuco:

starboard/tvos/shared/media/audio_queue_audio_sink_type.mm
```diff
@@ -441,11 +441,14 @@ bool IsValid(SbAudioSink audio_sink) override {
...
-  while (!sinks_to_destroy_.empty()) {
-    destroy_condition_.Wait();
+  {
+    std::lock_guard lock(audio_thread_mutex_);
+    sinks_to_destroy_.push_back(static_cast<TvosAudioSink*>(audio_sink));
+  }
+  audio_thread_condition_.notify_one();
+  {
+    std::unique_lock lock(audio_thread_mutex_);
+    destroy_condition_.wait(lock, [&]() { return sinks_to_destroy_.empty(); });
```
kjyoun: super nit: you can omit "()" here

kjyoun (approved):

starboard/tvos/shared/media/audio_queue_audio_sink_type.mm
rakuco: Done!

fayev-g (approved):

PR #8034, opened 2025-11-13
Skeleton changes for h5vcc updater

madhurajayaraman:
Bug: 454440974

### DIFF
--- a/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
+++ b/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
@@ -45,7 +45,7 @@ void H5vccUpdaterImpl::Create(
 
 void H5vccUpdaterImpl::GetUpdateServerUrl(GetUpdateServerUrlCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  std::move(callback).Run("");
+  NOTREACHED();
 }
 
 }  // namespace h5vcc_updater
--- a/cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
+++ b/cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
@@ -19,4 +19,4 @@ module h5vcc_updater.mojom;
 interface H5vccUpdater {
   // Get the update server url for the device.
   GetUpdateServerUrl() => (string update_server_url);
-  };
+};
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
@@ -1,4 +1,4 @@
-// Copyright 2020 The Cobalt Authors. All Rights Reserved.
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.


third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/resources/mock-h5vcc-updater.js
```diff
@@ -0,0 +1,48 @@
...
+
+  reset() {
+    this.stub_result_ = new Map();
+  }
+
+  stubResult(key, value) {
+    this.stub_result_.set(key, value);
+  }
+
+  stubUseCompressedUpdates(limitAdTracking) {
```
andrewsavage1: Update the variable names here

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Custom API for getting and setting Evergreen update settings.
+
+interface H5vccUpdater {
+
+  [CallWith=ScriptState, RaisesException]
+  Promise<DOMString> getUpdaterChannel();
```
andrewsavage1: Did we check that it's okay to make this an async API?

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+module h5vcc_updater.mojom;
+
+// The browser process must provide an implementation of this interface so that
+// the renderer process can implement the H5vccUpdater Blink API.
+interface H5vccUpdater {
+  // Get the advertisingId for the device.
```
andrewsavage1: cruft comment

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cobalt/browser/h5vcc_updater/h5vcc_updater_impl.h"
+
+#include "base/functional/callback.h"
+#include "build/build_config.h"
+#include "cobalt/updater/updater_module.h"
+
+#if BUILDFLAG(IS_ANDROIDTV)
```
andrewsavage1: Remove androidtv code

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "cobalt/configuration/configuration.h"
+#include "starboard/common/system_property.h"
+#include "starboard/system.h"
+#endif
+
+namespace h5vcc_updater {
+
+namespace {
+
+std::string GetUpdaterChannelShared() {
```
andrewsavage1: It doesn't look like this is used—you could add a NOTIMPLEMENTED call here and use this instead of calling GetInstance()->GetUpdaterChannel() to stub out the behavior for now and just add the skeleton (I recommend that), or you could remove this and actually hook it up to the updater (this may be a longer process, especially as we may need to convert the updater to use base::Singleton if we're going through with the current approach)

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: +1, that sounds good for now, but we'll need to figure out what to do when we cherry pick these changes to the `main` branch.

It would be best to follow the approach from c25 and not even expose the interface, or build the implementation code, for non-evergreen platforms. If that's not feasible (I think it is) then we could provide a stub implementation for non-evergreen platforms.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
hlwarriner: @madhurajayaraman, I think you made the right decision here to try to make the API async but Andrew's right that we should check with the client teams to make sure it's ok. I'll ping you offline about that - and sorry about not bringing it up sooner. As long as this change doesn't introduce a crash I'd say let's go ahead and merge it as async for now, and we can revert that change later if needed.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
```diff
@@ -0,0 +1,23 @@
...
+  readonly attribute DOMString? updaterChannel;
```
hlwarriner: We hopefully won't need to keep a sync version of this API around. But if you want to include this for now until we get clarity after talking with the client teams, that's ok.

third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/h5vcc-updater-getUpdaterChannel.https.any.js
```diff
@@ -0,0 +1,13 @@
+// META: global=window
+// META: script=/resources/test-only-api.js
+// META: script=resources/automation.js
+
+h5vcc_updater_tests(async (t, mockH5vccUpdater) => {
+  assert_implements(window.h5vcc, "window.h5vcc not supported");
+  assert_implements(window.h5vcc.updater, "window.h5vcc.updater not supported");
+
+  const expected = '1';
```
hlwarriner, line 11: Please add a comment explaining that this value comes from the stub implementation in the browser.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
andrewsavage1: Shouldn't we just wrap most of these changes in sb_is_evergreen/SB_IS(EVERGREEN) now?

hlwarriner: I just wasn't sure if we could conditionally expose the interface in the IDL, like we did at https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/h5vcc/h5vcc_updater.idl#L17. Do you know if that's possible? I don't see `CONDITIONAL` in https://chromium.googlesource.com/chromium/src/+/HEAD/third_party/blink/renderer/bindings/IDLExtendedAttributes.md. I guess we could use a partial interface to conditionally add the updater interface to h5vcc?

For the actual build variables/flags, we should prefer `use_evergreen`/`BUILDFLAG(USE_EVERGREEN)`, right?

madhurajayaraman: I've followed the conventions we're using for the other h5vcc APIs in this branch - should updater be gated differently?

third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/resources/mock-h5vcc-updater.js
madhurajayaraman: oops can you guess where i modeled these from

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: For most other interfaces, we want to expose the APIs for all platforms but then we have different implementations for different platforms at the Mojo level.

Here, the interface should ideally only be exposed, at all, for Evergreen platforms. If we don't have a good way to do that - I'll check with the chrome team about that - then we'll just want some kind of stub implementation for non-Evergreen platforms. So yeah, this interface is different than most of the other ones in this way.

yell0wd0g: Drive-by: WebIDL in Chrome has a `RuntimeEnabled` conditional 
that allows pegging an API to a flag, that then can be turned 
on or off per platform ([docs](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/blink/renderer/platform/RuntimeEnabledFeatures.md?q=RuntimeEnabled&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain&start=1). But
ideally the same API would be present in all platforms (the idea
being that the Web platform is the same everywhere) and calls
to a non-implemented method would `throw` synchronously upon
call. WDYT?

andrewsavage1: Yeah I think we should avoid using a RuntimeEnabled flag, as that would build the updater idl/implementation regardless. It should be fine to throw for unsupported platforms

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Custom API for getting and setting Evergreen update settings.
+
+interface H5vccUpdater {
+
+  [CallWith=ScriptState, RaisesException]
+  Promise<DOMString> getUpdateServerUrl();
+
```
hlwarriner: nit: please remove the blank lines here.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
```diff
@@ -0,0 +1,23 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Custom API for getting and setting Evergreen update settings.
+
+interface H5vccUpdater {
+
+  [CallWith=ScriptState, RaisesException]
+  Promise<DOMString> getUpdateServerUrl();
```
hlwarriner, line 21: Please carry forward the web API comment from c25: https://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/h5vcc/h5vcc_updater.idl#L43.

Maybe just change "Sets" to "Gets" for now, until the setter is added. So
```
// Gets the URL the updater will use for updates. This should only be used for
// testing and should not be available in production.
```

third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/h5vcc-updater-getUpdaterChannel.https.any.js
```diff
@@ -0,0 +1,15 @@
...
+// META: script=resources/automation.js
+
+h5vcc_updater_tests(async (t, mockH5vccUpdater) => {
+  assert_implements(window.h5vcc, "window.h5vcc not supported");
+  assert_implements(window.h5vcc.updater, "window.h5vcc.updater not supported");
+
+  // currently stubbed out and not implemented, so test will fail.
+  // placeholder value here till the updater plumbing is done.
+  const expected = '1';
+  mockH5vccUpdater.stubUpdateServerUrl(expected);
```
hlwarriner: This doesn't match the stub method from the mock, i.e., `stubGetUpdateServerUrl`. I think either is fine but they'll need to match.

third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/h5vcc-updater-getUpdaterChannel.https.any.js
```diff
@@ -0,0 +1,15 @@
...
+  let actual = window.h5vcc.updater.updateServerUrl();
```
hlwarriner: This should be `window.h5vcc.updater.getUpdateServerUrl`.

third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/h5vcc-updater-getUpdaterChannel.https.any.js
hlwarriner, line 1: Please just update the file name to match the API that it now tests after the latest commits.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -0,0 +1,51 @@
...
+
+void H5vccUpdaterImpl::Create(
+    content::RenderFrameHost* render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccUpdater> receiver) {
+  new H5vccUpdaterImpl(*render_frame_host, std::move(receiver));
+}
+
+void H5vccUpdaterImpl::GetUpdateServerUrl(GetUpdateServerUrlCallback callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  std::move(callback).Run();
```
hlwarriner: Does this build? I think it needs a string arg. It could just be an empty string or something for now until it's actually implemented.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: That SGTM to throw for unsupported platforms. I'm ok with that being a fast-follow, since the impl is stubbed anyway for now and since this is on the 26.eap branch.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: It does build, it doesn't seem to throw anything, so I'll leave as is. in the follow up where I add in the actual updater functionality, it'll be edited anyway.

hlwarriner: The `evergreen-x64 / x64_qa` check has this compiler error for the `obj/cobalt/browser/h5vcc_updater/h5vcc_updater/h5vcc_updater_impl.o` target:

```
../../cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc:48:23: error: no matching member function for call to 'Run'
  std::move(callback).Run();
  ~~~~~~~~~~~~~~~~~~~~^~~
../../base/functional/callback.h:133:5: note: candidate function not viable: requires single argument 'args', but no arguments were provided
  R Run(Args... args) const& {
    ^
../../base/functional/callback.h:144:5: note: candidate function not viable: requires single argument 'args', but no arguments were provided
  R Run(Args... args) && {
    ^
1 error generated.
```

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+module h5vcc_updater.mojom;
+
+// The browser process must provide an implementation of this interface so that
+// the renderer process can implement the H5vccUpdater Blink API.
+interface H5vccUpdater {
+  // Get the update server url for the device.
+  GetUpdateServerUrl() => (string update_server_url);
+  };
```
andrewsavage1: nit: fix whitespace

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cobalt/browser/h5vcc_updater/h5vcc_updater_impl.h"
+
+#include "base/functional/callback.h"
+#include "build/build_config.h"
+
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 20: Probably unnecessary, as we should likely only build this file for starboard platforms, but we can address in a follow up

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
andrewsavage1: Call NOTIMPLEMENTED or NOTREACHED here for now to show it's a stub

Also, does this build as is? Don't you need a string for the callback?

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
```diff
@@ -0,0 +1,22 @@
+// Copyright 2020 The Cobalt Authors. All Rights Reserved.
```
andrewsavage1: As always, fix the copyright year :)

third_party/blink/web_tests/wpt_internal/cobalt/h5vcc-updater/h5vcc-updater-getUpdateServerUrl.https.any.js
```diff
@@ -0,0 +1,15 @@
...
+// META: script=/resources/test-only-api.js
+// META: script=resources/automation.js
+
+h5vcc_updater_tests(async (t, mockH5vccUpdater) => {
+  assert_implements(window.h5vcc, "window.h5vcc not supported");
+  assert_implements(window.h5vcc.updater, "window.h5vcc.updater not supported");
+
+  // currently stubbed out and not implemented, so test will fail.
+  // placeholder value here till the updater plumbing is done.
+  const expected = '1';
```
andrewsavage1, line 11: Put a comment here noting that this specifically comes from the stub mojo implementation

hlwarriner: The test is using this expected value to configure the mock impl (`mockH5vccUpdater.stubGetUpdateServerUrl(expected)`) so I think this reads ok without a comment.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: It should need a string, yeah. https://github.com/youtube/cobalt/pull/8034/files#r2546224890

PR #8036, opened 2025-11-13
Introduce Safe Config Fetch Time

yjzhang111:
Bug: 458397954

### DIFF
--- a/cobalt/browser/experiments/experiment_config_manager_unittest.cc
+++ b/cobalt/browser/experiments/experiment_config_manager_unittest.cc
@@ -254,7 +254,7 @@ TEST_F(ExperimentConfigManagerTest,
   // Enable the feature in the SAFE config.
   pref_service_->SetDict(kSafeConfigFeatures, std::move(feature_map));
 
-  pref_service_->SetTime(variations::prefs::kVariationsLastFetchTime,
+  pref_service_->SetTime(variations::prefs::kVariationsSafeSeedFetchTime,
                          base::Time::Now() - base::Days(5));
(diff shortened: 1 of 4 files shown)


cobalt/browser/experiments/experiment_config_manager_unittest.cc
```diff
@@ -400,4 +402,21 @@ TEST_F(ExperimentConfigManagerTest,
             ExperimentConfigType::kRegularConfig);
 }
 
+TEST_F(ExperimentConfigManagerTest, StoreSafeConfigSetsFetchTime) {
+  base::Time fetch_time = base::Time::Now() - base::Days(5);
```
Awallky, line 481: Gemini's suggestion isn't bad, but I'd recommend using a ClockUtil across these unittests to inject/mock a clock whose time we control instead of using the system clock - go/totw/111#injecting-a-utilclock.

Awallky: Oh, I see it looks like you might already be doing that. 
GH just hides some of it. nvm

Awallky:
@yjzhang111 What's the plan of action to make use of this pref? Are we planning on logging this in a histogram? Using it to alter SafeMode behavior?

yell0wd0g:
Please consider adding a descriptive commit message to help reviewers and git history.
You can now simply ask gemini to suggest a message by dropping 

/generate-commit-message

see e.g. https://github.com/youtube/cobalt/pull/8059#issuecomment-3535430232

PR #8039, opened 2025-11-14
android: Ensure SplashShell creation before AppShell

borongc:
The SplashShell is now explicitly created during activity
initialization if it does not already exist. This guarantees
the splash screen appears correctly, especially after scenarios
like switching Android TV accounts where the shell might not
be initialized by default. Renames getSplashAppShell to getSplashShell.

Issue: 459802298
Issue: 441738743

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -286,7 +286,7 @@ public abstract class CobaltActivity extends Activity {
                   public void run() {
                     synchronized(lock) {
                       if (isMainFrameLoaded == false) {
-                        Log.i(TAG, "NativeSplash:switch to main shell after timeout " + mSplashTimeoutMs + "ms");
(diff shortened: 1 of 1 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -230,6 +230,12 @@ private void finishInitialization(Bundle savedInstanceState) {
     //     it loads the native splash screen URL.
     //   - mShellManager.showAppShell() switches the visible shell from
     //     the active shell to the App shell.
+    if (mShellManager.getSplashShell() == null) {
```
zhongqiliang, line 233: can you refactor the code from here to line 301 into 2 separate functions.
one for native splash enabled, the other for disabled.

borongc: SplashShell is needed to be created first. If native splash screen is disabled, this will be close by calling `mShellManager.showAppShell();` (L272) anyway.

johnxwork (approved):

PR #8042, opened 2025-11-14
[tvOS] Remove SB_API_VERSION checks for old versions

rakuco:
Remove all blocks that are enabled only for SB_API_VERSION <= 16, which sometimes include entire files.

Bug: 459507205


rakuco:

PR #5765, opened 2025-05-14
Move eztime out of Starboard

jellefoks:
This moves eztime out of Starboard, removes it's usage and remaining ICU usage from Starboard.

eztime is temporarily added in cobalt/common. In a followup, it wil be deprecated when the matching standard POSIX time functions are implemented instead.

b/406083355

PR #8043, opened 2025-11-14
[tvOS] Implement a local copy of PosixTimeToWindowsTime()

rakuco:
starboard::PosixTimeToWindowsTime() was removed in #5765, but an implementation is still needed in order to set SbPlayerInfo::start_date.

Add a copy of the original function identical to the one that also exists in `//starboard/loader_app/drain_file.h`.

Bug: 437419202
Bug: 456692793

### DIFF
--- a/starboard/tvos/shared/media/player_get_info.mm
+++ b/starboard/tvos/shared/media/player_get_info.mm
@@ -14,14 +14,25 @@
 
 #import <Foundation/Foundation.h>
 
-#import <limits>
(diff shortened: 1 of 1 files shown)


rakuco:

fayev-g (approved):
I can see wanting to write a util file to use between starboard/loader_app/drain_file.h, but it's only a few lines of code and not likely to change. Maybe if it gets used again it would be worth it.

PR #8044, opened 2025-11-14
[tvOS] Adapt to #7442 (starboard: Introduce Result<T> for robust error handling)

rakuco:
Update function signatures in the tvOS implementation. While here, remove a reference to the `uikit` namespace that was removed in #8031.

Bug: 449188714
Bug: 456692793


rakuco:

fayev-g (approved):

PR #8045, opened 2025-11-14
[tvOS] Remove SBD_UNAVAILABLE_INITIALIZER_IMPL

rakuco:
The declarations of these functions are already marked with NS_UNAVAILABLE, so calling them is already forbiden at build-time and throwing an exception at runtime is redundant.

Furthermore, the macro completely confuses clang-format when used without a semicolon (and using it _with_ a semicolon results in a trailing semicolon warning).

Bug: 456692793


rakuco:

PR #8048, opened 2025-11-14
[tvOS] Remove ui_nav_*

rakuco:
These files are part of the hybrid nav implementation that is not going to be used anymore.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #8050, opened 2025-11-14
Reland "feat(cc): Disable resource pool reuse to reduce memory #7853"

sherryzy:
This relands PR# 7853, which was reverted due to breaking `PartialRasterTileManagerTest.PartialRasterSuccessfullyEnabled` and `PartialRasterTileManagerTest.PartialTileImageDecode` on Cobalt.

This reland re-introduces the test refactoring but disables the failing
tests specifically for Cobalt builds using
`#if BUILDFLAG(IS_COBALT) GTEST_SKIP()`. This allows the valuable
test improvements to be landed without affecting the Cobalt build.

Original PR description---------------------------------------

By marking resources with 'avoid_reuse', they are freed immediately
after use instead of being kept in the resource pool. This ensures tiles
are rastered on-demand.

This change significantly reduces the tile_memory usage, ~20M when
navigating fast through videos, with no obvious negative performance
impact.

Original PR: https://github.com/youtube/cobalt/pull/7853

Bug: 424591313

### DIFF
--- a/cc/tiles/tile_manager_unittest.cc
+++ b/cc/tiles/tile_manager_unittest.cc
@@ -2454,11 +2454,19 @@ void RunPartialTileDecodeCheck(std::unique_ptr<LayerTreeHostImpl> host_impl,
 // Ensures that the tile manager successfully reuses tiles when partial
 // raster is enabled.
 TEST_F(PartialRasterTileManagerTest, PartialRasterSuccessfullyEnabled) {
+#if BUILDFLAG(IS_COBALT)
+  GTEST_SKIP() << "Cobalt doesn't support partial tile raster.";
+#else
   RunPartialRasterCheck(TakeHostImpl(), true /* partial_raster_enabled */);
+#endif
 }
 
 TEST_F(PartialRasterTileManagerTest, PartialTileImageDecode) {
+#if BUILDFLAG(IS_COBALT)
+  GTEST_SKIP() << "Cobalt doesn't support partial tile raster.";
+#else
   RunPartialTileDecodeCheck(TakeHostImpl(), true /* partial_raster_enabled */);
+#endif
 }
 
 TEST_F(PartialRasterTileManagerTest, CompleteTileImageDecode) {
--- a/cc/resources/resource_pool.cc
+++ b/cc/resources/resource_pool.cc
@@ -190,6 +190,9 @@ ResourcePool::InUsePoolResource ResourcePool::AcquireResource(
   PoolResource* resource = ReuseResource(size, format, color_space);
   if (!resource)
     resource = CreateResource(size, format, color_space);
+#if BUILDFLAG(IS_COBALT)
+  resource->mark_avoid_reuse();
+#endif
   resource->set_debug_name(debug_name);
   return InUsePoolResource(resource, !!context_provider_);
 }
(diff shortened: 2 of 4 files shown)


cc/tiles/tile_manager_unittest.cc
```diff
@@ -2460,6 +2462,7 @@ TEST_F(PartialRasterTileManagerTest, PartialRasterSuccessfullyEnabled) {
 TEST_F(PartialRasterTileManagerTest, PartialTileImageDecode) {
   RunPartialTileDecodeCheck(TakeHostImpl(), true /* partial_raster_enabled */);
```
yell0wd0g, line 2468: As a common rule of thumb, I'd rather disable tests with `GTEST_SKIP()`
rather than commenting out. I.e. something like
```
TEST_F(PartialRasterTileManagerTest, PartialRasterSuccessfullyEnabled) {
#if BUILDFLAG(IS_COBALT)
  GTEST_SKIP() << "Cobalt doesn't support partial tile raster";
#endif 
  RunPartialRasterCheck(TakeHostImpl(), true /* partial_raster_enabled */);
}
```

This is because skipped tests are accounted for at the end of test
runs (i.e. there's four categories out of a GTest run: pass, fail,
skipped, and disabled). 

Skipping for a certain case also has the (minor) bonus of positive 
logic: if Bla - then skip, as opposed to if !Bla - then run.

In some cases there's so many tests that it might not be practical,
but in this PR it seems like we could try said skip.

yell0wd0g (approved):
LGTM with a suggestion to rather use SKIP. Thanks
for the followup and the writeup!

Re. commit description, Chromium has a way to keep the original
CL description, see e.g. https://chromium-review.googlesource.com/c/chromium/src/+/4808891 (the
original CL description is copied verbatim after a line,
and the first commit -patchset in gerrit- is the reverted
CL also verbatim). I'm not sure if there's a docs/
entry though, so take this as a nit-suggestion.

cc/tiles/tile_manager_unittest.cc
andrewsavage1: I would think it's better to disable tests in cases like this following Chromium's pattern of MAYBE_ prefixes, i.e. 

#if BUILDFLAG(IS_COBALT)
#define MAYBE_PartialRasterEnabled DISABLED_PartialRasterEnabled
#else
#define MAYBE_PartialRasterEnabled PartialRasterEnabled
#endif

yell0wd0g: Hmm yeah, I think it's a matter of subtle meaning: I would read
SKIP as "platform doesn't support it, so no sense trying", whereas
DISABLED I'd read as "it's disabled now but we'd like to enable
it" (DISABLED tests can be run via command line flag, contrary to
the SKIP ones, which are always skipped). So I guess the question
is whether partial raster is broken and we'd like to fix it, or
is it not supported at all?  (I thought we were in the second case).

sherryzy: Yeah, we are the second case. Those tests are skipped on purpose because we don't support resource reuses. This change currently only applies on 26.eap, in M138, I'll try to find a better approach to disable resource pool so that we don't need to disable all the tests accordingly. For now, I'll vote for GTEST_SKIP().

PR #8051, opened 2025-11-14
Revert "Cherry pick PR #7527: cobalt: Tune CC GPU memory settings to reduce footprint"

borongc:
Reverts youtube/cobalt#7577

It seems https://github.com/youtube/cobalt/pull/7577 improves UI latency with the tradeoff of increased crashes in blink. As a result, reverting this PR to verify if this is the root cause.

Issue: 424591313 
Issue: 437916038


sherryzy (approved):
Leave a LGTM in case the other revert doesn't fix the STFC regressions.

andrewsavage1:
What's the status of this?

borongc:
> What's the status of this?

Looks like this has no impact on STFC, but leave it for more days, so we can have more data from 6.33.

borongc:
It seems https://github.com/youtube/cobalt/pull/8070 doesn't lower the crash at at media::WebMStreamParser::AppendToParseBuffer(), we may merge this PR to 26.android.

PR #8054, opened 2025-11-14
Only wire up CrashHandler extension for Evergreen

hlwarriner:
This fixes a crash on monolithic and non-evergreen modular builds that was introduced by #7548. There are other reasonable ways to fix the crash but I think this makes sense, at least for now, since we currently do not support crashpad - or any real crash handler - for non-evergreen builds.

Issue: 460489024

### DIFF
--- a/starboard/nplb/BUILD.gn
+++ b/starboard/nplb/BUILD.gn
@@ -363,7 +363,6 @@ test("nplb") {
   if (sb_is_evergreen_compatible) {
     sources += [
       "nplb_evergreen_compat_tests/checks.h",
-      "nplb_evergreen_compat_tests/crashpad_config_test.cc",
       "nplb_evergreen_compat_tests/executable_memory_test.cc",
       "nplb_evergreen_compat_tests/fonts_test.cc",
       "nplb_evergreen_compat_tests/loader_app_metrics_test.cc",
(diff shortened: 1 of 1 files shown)


hlwarriner:
I'm not sure if we intend to support a non-evergreen modular build for RDK (we do not have one in CI). If so, we'd probably need a similar change in the RDK port. I'll look into that next week.

starboard/linux/shared/system_get_extensions.cc
```diff
@@ -14,12 +14,15 @@
 
 #include "starboard/system.h"
 
+#include "build/build_config.h"
 #include "starboard/common/string.h"
 #if SB_IS(EVERGREEN_COMPATIBLE)
 #include "starboard/elf_loader/evergreen_config.h"
 #endif
 #include "starboard/extension/configuration.h"
+#if BUILDFLAG(USE_EVERGREEN)
```
andrewsavage1: nit: I know you're just following the existing pattern, but it's better to put conditional includes below other includes (see go/cstyle#Names_and_Order_of_Includes under "Exception")

hlwarriner: Done

hlwarriner:
> We won't support non-evergreen RDK so no need to worry about that. I do wonder if there are any partners that try to build without Evergreen though—maybe we should put an error pragma in the crash_handler extension if `!SB_IS(EVERGREEN)`?

Cool, thanks.

Re the error pragma, there's nothing wrong with non-evergreen platforms using this shared extension implementation per se. But they do need to use it correctly, and non-evergreen platforms were not because they were not calling `RegisterSetStringCallback`, even with a stub. The nullptr check I suggested in b/460489024#comment4 will make the extension safer and I think we can skip the error pragma.

PR #8066, opened 2025-11-17
tvOS: Add cobalt_switch_defaults_tvos.cc

rakuco:
Just like ATV has its own set of command-line defaults and switches, tvOS needs something separate from the Starboard settings.

This commit movesove the platform-independent parts out of `cobalt_switch_defaults_starboard.cc` so that the newly-added `cobalt_switch_defaults_tvos.cc` can use them as well.

At the moment, the difference in switches between Starboard and tvOS is:
- tvOS does not have `switches::kEnableLowEndDeviceMode` and `blink::switches::kDisableRGBA4444Textures`
- tvOS does not disable Vulkan because it is not used on Apple platforms.
- Ozone-specific settings for ANGLE were removed for tvOS.

Patch by @jkim-julie.

Bug: 456692793

### DIFF
--- a/cobalt/app/cobalt_switch_defaults.h
+++ b/cobalt/app/cobalt_switch_defaults.h
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef COBALT_APP_COBALT_SWITCH_DEFAULTS_STARBOARD_H_
-#define COBALT_APP_COBALT_SWITCH_DEFAULTS_STARBOARD_H_
+#ifndef COBALT_APP_COBALT_SWITCH_DEFAULTS_H_
+#define COBALT_APP_COBALT_SWITCH_DEFAULTS_H_
 
 #include <map>
 #include <string>
@@ -53,4 +53,4 @@ class CommandLinePreprocessor {
 
 }  // namespace cobalt
 
-#endif  // COBALT_APP_COBALT_SWITCH_DEFAULTS_STARBOARD_H_
+#endif  // COBALT_APP_COBALT_SWITCH_DEFAULTS_H_
--- a/cobalt/app/cobalt_switch_defaults_starboard.cc
+++ b/cobalt/app/cobalt_switch_defaults_starboard.cc
@@ -78,36 +78,35 @@ CommandLinePreprocessor::GetCobaltToggleSwitches() {
 
 const base::CommandLine::SwitchMap&
 CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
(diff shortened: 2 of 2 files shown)


rakuco:

rakuco:
@yell0wd0g PTAL. The downside of this approach is that command-line settings are defined in 3 different places: the Java code for ATV, `cobalt_switch_defaults_starboard.cc` for Starboard and `cobalt_switch_defaults_tvos.cc` for tvOS.

A possible different approach is to rename `cobalt_switch_defaults_starboard.cc` to something more generic and have an `#ifdef` soup to set different values depending on the platform (I guess the benefit is that people would be less likely to forget to update one of the files).

rakuco:
(I've filed b/461065037 to track the weird formatting required in `cobalt_switch_defaults_starboard.cc`)

fayev-g:
> @yell0wd0g PTAL. The downside of this approach is that command-line settings are defined in 3 different places: the Java code for ATV, `cobalt_switch_defaults_starboard.cc` for Starboard and `cobalt_switch_defaults_tvos.cc` for tvOS.
> 
> A possible different approach is to rename `cobalt_switch_defaults_starboard.cc` to something more generic and have an `#ifdef` soup to set different values depending on the platform (I guess the benefit is that people would be less likely to forget to update one of the files).

Fwiw, separating out files per platform and avoiding #ifdef soup is actually Google-preferred style, and I also prefer it. Google C++ style [0] on avoiding preprocessor macros enforces readability where the code you see is the same as what the compiler sees. I've noticed Chromium diverges from this pattern quite a bit though.

[0]: https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros

andrewsavage1:
> > @yell0wd0g PTAL. The downside of this approach is that command-line settings are defined in 3 different places: the Java code for ATV, `cobalt_switch_defaults_starboard.cc` for Starboard and `cobalt_switch_defaults_tvos.cc` for tvOS.
> > A possible different approach is to rename `cobalt_switch_defaults_starboard.cc` to something more generic and have an `#ifdef` soup to set different values depending on the platform (I guess the benefit is that people would be less likely to forget to update one of the files).
> 
> Fwiw, separating out files per platform and avoiding #ifdef soup is actually Google-preferred style, and I also prefer it. Google C++ style [0](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros) on avoiding preprocessor macros enforces readability where the code you see is the same as what the compiler sees. I've noticed Chromium diverges from this pattern quite a bit though.

I'm fine with this approach, but I think it would be possible (and good) at some point to have a layer of shared flags and a separate layer of platform-specific flags, so for the majority of flags we can add them once instead of needing to add them multiple times across multiple platforms.

PR #8067, opened 2025-11-17
[tvOS] Add empty implementation of h5vcc functions

rakuco:
Stop `#error`ing out on tvOS in h5vcc function implementations on tvOS by adding some empty implementations, which at least allows the build to proceed.

Patch by @abhijeetk.

Bug: 447135715

### DIFF
--- a/cobalt/build/gn.py
+++ b/cobalt/build/gn.py
@@ -38,7 +38,9 @@ _BUILD_TYPES = {
     },
     'qa': {
         'symbol_level': 1,
-        'is_official_build': 'true'
+        'is_official_build': 'true',
+        # Enable stack traces, disabled by is_official_build
+        'exclude_unwind_tables': 'false'
     },
     'gold': {
         'symbol_level': 1,


rakuco:

cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_impl.cc
```diff
@@ -67,6 +67,10 @@ void H5vccAccessibilityImpl::IsTextToSpeechEnabledSync(
   bool enabled =
       CobaltTextToSpeechHelper::GetInstance()->IsTextToSpeechEnabled(env);
   std::move(callback).Run(enabled);
+#elif BUILDFLAG(IS_IOS_TVOS)
+  // TODO: Implement text-to-speech availability check for tvOS.
```
andrewsavage1: Add a bug number—feel free to use b/447135715 as an umbrella for all tvos h5vcc implementations

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -162,6 +168,10 @@ void H5vccSystemImpl::GetUserOnExitStrategy(
   std::move(callback).Run(GetUserOnExitStrategyInternal());
 #elif BUILDFLAG(IS_ANDROIDTV)
   std::move(callback).Run(h5vcc_system::mojom::UserOnExitStrategy::kMinimize);
+#elif BUILDFLAG(IS_IOS_TVOS)
+  // TODO: Determine appropriate user exit strategy for tvOS.
+  NOTIMPLEMENTED();
+  std::move(callback).Run(h5vcc_system::mojom::UserOnExitStrategy::kNoExit);
```
andrewsavage1: It should still build if you don't do this, right? Can you use std::ignored instead to be more explicit about this not being implemented?

andrewsavage1 (approved):

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
rakuco: > It should still build if you don't do this, right?

Yeah, it will build but crash at runtime with something like `DCHECK failed: !connected. H5vccSystem::GetUserOnExitStrategyCallback was destroyed without first either being run or its corresponding binding being closed. It is an error to drop response callbacks which still correspond to an open interface pipe.`

> Can you use std::ignored instead to be more explicit about this not being implemented?

Hmm, I couldn't find other occurrences of `std::ignore` used for this, did you mean something like `std::ignore = std::move(callback)`? That crashes at runtime, just like if I call `callback.Reset()`.

cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_impl.cc
rakuco: Done.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
andrewsavage1: Got it, is this the same exit strategy for tvos as c25?

rakuco: In C25 it was "suspend", which I think maps to `kMinimize` here. I've changed it to match the original version (there's a patch in the works actually implementing this behavior FWIW).

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -82,14 +85,17 @@ bool GetLimitAdTrackingShared() {
...
+  NOTIMPLEMENTED();
 #else
 #error "Unsupported platform."
 #endif
   return limit_ad_tracking;
 }
 
 std::string GetTrackingAuthorizationStatusShared() {
-  // TODO - b/395650827: Connect to Starboard extension.
+  // TODO(b/447135715): b/395650827: Connect to Starboard extension.
```
andrewsavage1: nit: use the style `TODO: b/123 - Description`. For multiple bugs, let's do `TODO: b/123, b/456 - Description`

rakuco: Done.

PR #8068, opened 2025-11-17
net: Fix memory leak in CobaltTrustedURLLoaderHeaderClient

kjyoun:
The `CobaltTrustedURLLoaderHeaderClient` was unconditionally appending new `CobaltTrustedHeaderClient` instances to a vector for every request, creating a memory leak.

This CL changes the ownership model to use `mojo::MakeSelfOwnedReceiver`, allowing the `CobaltTrustedHeaderClient` to manage its own lifetime tied to the Mojo connection.

Bug: 456171826


cobalt/browser/client_hint_headers/cobalt_trusted_header_client.h
```diff
@@ -24,8 +24,7 @@ namespace browser {
```
Awallky, line 24: Suggest something describing the purpose of this class like: 

```
// Responsible for handling trusted headers for individual network requests.                                                                                                                                 
// This class is used to add Cobalt-specific client hint headers before a request is sent. 
```

cobalt/browser/client_hint_headers/cobalt_trusted_header_client.cc
Awallky, line 40: suggest adding a comment like:

```
// Cobalt does not currently need to act on response headers, so this is a no-op.
```

cobalt/browser/client_hint_headers/cobalt_trusted_url_loader_header_client.h
Awallky, line 24: suggest adding comment like:

```
// This class acts as a factory for creating CobaltTrustedHeaderClient instances
// for each new URL loader. It is responsible for binding the Mojo receiver
// for the TrustedHeaderClient interface.
```

Awallky (approved):
LGTM % comments.
Great find!

cobalt/browser/client_hint_headers/cobalt_trusted_header_client.h
kjyoun: Done

cobalt/browser/client_hint_headers/cobalt_trusted_url_loader_header_client.h
kjyoun: Done

cobalt/browser/client_hint_headers/cobalt_trusted_header_client.cc
kjyoun: Done

zhongqiliang:
LGTM, if the request headers are still added.

kjyoun:
verified network request to have the wanted header

https://screenshot.googleplex.com/8UkdePWfeYARVT5

cobalt/browser/client_hint_headers/cobalt_trusted_url_loader_header_client.cc
```diff
@@ -36,9 +38,8 @@ void CobaltTrustedURLLoaderHeaderClient::OnLoaderForCorsPreflightCreated(
 
 void CobaltTrustedURLLoaderHeaderClient::CreateAndBindCobaltTrustedHeaderClient(
     mojo::PendingReceiver<network::mojom::TrustedHeaderClient> receiver) {
-  auto client =
-      std::make_unique<browser::CobaltTrustedHeaderClient>(std::move(receiver));
-  cobalt_header_clients_.push_back(std::move(client));
+  mojo::MakeSelfOwnedReceiver(std::make_unique<CobaltTrustedHeaderClient>(),
+                              std::move(receiver));
```
zhongqiliang, line 42: I thought mojo::PendingReceiver will be connected by the remote, and it will be closed and destroyed by the remote. Not true? @yell0wd0g 
This is the first time I see mojo::MakeSelfOwnedReceiver.

Awallky: I'm not sure if that's true or not, but even if it is, wouldn't the vector here still hold onto objects of that type indefinitely, even if they become null at some point?

kjyoun: Per [self_owned_receiver.h](https://source.corp.google.com/h/github/youtube/cobalt/+/main:mojo/public/cpp/bindings/self_owned_receiver.h;l=132-143;drc=524126904340bf2104b6615d6bb6352ba6d7d844), it seems to be designed to handle the use case that we want.

```
// Binds the lifetime of an interface implementation to the lifetime of the
// Receiver. When the Receiver is disconnected (typically by the remote end
// closing the entangled Remote), the implementation will be deleted.
//
// Any incoming method calls or disconnection notifications will be scheduled
// to run on |task_runner|. If |task_runner| is null, this defaults to the
// current SequencedTaskRunner.
//
// Note: self-owned receivers are unsafe to use when the interface
// implementation has lifetime dependencies outside of its control.
```

kjyoun: merge this PR so that it can be included tonight's nightly build and test
(then we have one more day to fix any issue, before Wednesday build cut)

kjyoun: Will create a separate PR for any comment

yell0wd0g: If a class `A` fully owns a `Receiver<X>` (e.g. [here](https://chromium.googlesource.com/chromium/src/+/main/docs/mojo_and_services.md#Implementing-the-Interface)) then 
at any point in time there's only one such alive; `A` can 
reuse it by `Unbind()`ing and `Bind()`ing it again, but
there's only one.

Aside, when the `Remote` connected to `X` is closed, `X` 
is not destroyed automatically (the `Remote` destruction 
causes a connection error on the `Receiver` side, which 
can be handled to manage the receiver's lifetime). 
`SelfOwnedReceiver` is a convenience for that operation
(see [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:mojo/public/cpp/bindings/self_owned_receiver.h;l=107-108;drc=21fb7d148f9e0948484b16d75a2467ef6c4b0787), [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:mojo/public/cpp/bindings/self_owned_receiver.h;l=113-120;drc=21fb7d148f9e0948484b16d75a2467ef6c4b0787) and [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:mojo/public/cpp/bindings/self_owned_receiver.h;l=84;drc=21fb7d148f9e0948484b16d75a2467ef6c4b0787)).

I think the misunderstanding was that destroying a 
`Remote<CobaltTrustedHeaderClient>` was enough to destroy
 the associated entry in `cobalt_header_clients_`, right?

kjyoun: Yes, exactly. This change makes CobaltTrustedHeaderClient a SelfOwnedReceiver, which is destroyed automatically when the remote is  disconnected.

PR #8070, opened 2025-11-17
Revert "Cherry pick PR #7590: content/compositor: Allow more Workers for raster/image decoding for Android"

borongc:
Reverts youtube/cobalt#7605

It seems https://github.com/youtube/cobalt/pull/7605 improves UI latency with the tradeoff of increased crashes in blink. As a result, reverting this PR to verify if this is the root cause.

Issue: 424591313
Issue: 437916038


yell0wd0g (approved):
LGTM - let's see if this was a bad (TM) CL!!

PR #8072, opened 2025-11-17
android: Fix CobaltA11yHelper SurfaceView attachment

borongc:
Previously, CobaltA11yHelper was initialized with the video-specific
SurfaceView. This led to accessibility features not correctly
interacting with the main content SurfaceView.

Bind CobaltA11yHelper to the ContentViewRenderView's SurfaceView
instead. This ensures accessibility features operate on the primary
display surface.

Issue: 431317298

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -161,6 +161,8 @@ public abstract class CobaltActivity extends Activity {
     // SurfaceView's 'hole' clipping during animations that are notified to the window.
     mWindowAndroid.setAnimationPlaceholderView(
         mShellManager.getContentViewRenderView().getSurfaceView());
+    mA11yHelper = new CobaltA11yHelper(this,
+        mShellManager.getContentViewRenderView().getSurfaceView());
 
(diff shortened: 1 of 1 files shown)


haozheng-cobalt (approved):
I've tested all a11y edge cases susceptible to focus issue and did not run into any issues!

Awallky:
@borongc Could we add some unit test to cover this in any reasonable way to prevent this from regressing? 
In addition, can we either directly add or file a bug to add a cobalt_browsertest to provide e2e coverage of this feature? 

I think this one is at risk to regress at slight changes and testing that runs in our CI would go a long way in preventing this.

borongc:
> @borongc Could we add some unit test to cover this in any reasonable way to prevent this from regressing? In addition, can we either directly add or file a bug to add a cobalt_browsertest to provide e2e coverage of this feature?
> 
> I think this one is at risk to regress at slight changes and testing that runs in our CI would go a long way in preventing this.

If you are asking about unit tests for CobaltA11yHelper, this should be a follow-up for @haozheng-cobalt.

Note this PR doesn't introduce any new feature.

borongc:
> I filed b/461783385 for @haozheng-cobalt to follow up with tests for this change.
> 
> @borongc, could you note this somewhere in either a test footer or in the PR description?

The PR is auto merged, but I can do this in the cherry pick PR.

PR #8090, opened 2025-11-18
BACKPORT: [tvOS] Add support for editing text input fields

rakuco:
Backport M139's https://crrev.com/c/6633696, another commit in a series to improve input handling (including the Siri remote) on tvOS.

Bug: 432824509

---

Original CL description:

The implementation is much simpler than the iOS one because there is no BrowserEngineKit to deal with.

Furthermore, tvOS has some peculiarities:

1. Contrary to iOS, simply adopting the UIKeyInput protocol does not cause the system keyboard to be displayed when RenderWidgetUIView becomes the first responder.
2. The system keyboard takes up the entire screen.

What we do to work around 1) is add a dummy UITextField instance that we selectively add to the view hierarchy when we need to open or close the on-screen keyboard.

For 2), we need to make WidgetBase::ShowVirtualKeyboardOnElementFocus() behave like on ChromeOS and not immediately show the system keyboard when a text element is focused. Instead, we only show it when it is focused and a tap occurs.

Change-Id: Id5507b6c8b8f247b485ba76f55f330c4a2519fa7
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6633696
Cr-Commit-Position: refs/heads/main@{#1473775}

### DIFF
--- a/content/browser/renderer_host/render_widget_host_view_ios.mm
+++ b/content/browser/renderer_host/render_widget_host_view_ios.mm
@@ -758,7 +758,6 @@ void RenderWidgetHostViewIOS::OnUpdateTextInputStateCalled(
     TextInputManager* text_input_manager,
     RenderWidgetHostViewBase* updated_view,
     bool did_update_state) {
-#if !BUILDFLAG(IS_IOS_TVOS)
   if (text_input_manager->GetActiveWidget()) {
     [ui_view_->view_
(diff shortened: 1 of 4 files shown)


rakuco:

rakuco:
@andrewsavage1 @yell0wd0g could either of you rubber-stamp this one?

PR #8092, opened 2025-11-18
cobalt: Make passing a UA string in the command-line work

rakuco:
Commit d9b2353f151 ("Fix linker issues for linux-x64x11 cobalt build (#7437)") implemented this behavior that is part of content shell as a side effect of trying to fix the build, but it does not work as expected in Cobalt because CobaltContentBrowserClient overrides GetUserAgent().

Move the call to embedder_support::GetUserAgentFromCommandLine to CobaltContentBrowserClient and restrict it to non-official builds while here.

Bug: 442760345


rakuco:

PR #7563, opened 2025-10-16
starboard: Replace starboard/types.h with standard headers

kjyoun:
This commit removes `starboard/types.h` and replaces its inclusions with standard C/C++ headers like <stdint.h>, <stddef.h>, and <stdbool.h>. Static asserts previously located in `starboard/types.h` have been moved to `starboard/nplb/sabi/size_test.cc`.

Bug: 366064934

PR #8093, opened 2025-11-18
[tvOS] Adapt to PR #7563 ("starboard: Replace starboard/types.h with standard headers")

rakuco:
Remove starboard/types.h in the headers that are actually used in the (for now internal) build and replace it with the appropriate system headers.

Bug: 366064934


rakuco:

fayev-g (approved):

PR #5899, opened 2025-05-28
android: Implement ExoPlayer based SbPlayer

osagie98:
Introduce a new media playback pipeline utilizing the Android ExoPlayer
library. This implementation provides an alternative SbPlayer handler
that forwards encoded media data and playback commands to the Java-side
ExoPlayer.

This change is foundational for integrating ExoPlayer, with initial
support for basic playback. Key areas such as tunnel mode seek/pause
and DRM playback are noted as future work. The new pipeline is enabled
via the kEnableExoPlayer feature flag.

Bug: 191693650

PR #8094, opened 2025-11-18
android: Integrate ExoPlayer Java components

osagie98:
android: Integrate ExoPlayer Java components

This change introduces the core Java classes for the ExoPlayer
rendering implementation on Android. These classes manage the ExoPlayer
instance, facilitate communication between native and Java code,
handle Digital Rights Management (DRM) sessions, and provide the
mechanism for streaming media samples to ExoPlayer.

This is a foundational step to leverage ExoPlayer for media playback,
aiming to improve media capabilities, performance, and compatibility
on Android devices.

Bug: 191693650


cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,200 @@
...
+        mediaPeriod = new ExoPlayerMediaPeriod(format, allocator);
+        return mediaPeriod;
+    }
+
+    @Override
+    public void releasePeriod(MediaPeriod mediaPeriod) {
+        // Ignore the passed-in MediaPeriod and call the ExoPlayerMediaPeriod directly. As there's
+        // only a single MediaPeriod, this will match the passed MediaPeriod.
+        this.mediaPeriod.destroySampleStream();
+    }
```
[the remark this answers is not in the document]
osagie98: Added a log if this happens, with the current implementation this shouldn't be possible.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+            return new TrackGroupArray(new TrackGroup(format));
+        }
+
+        @Override
+        public long selectTracks(ExoTrackSelection[] selections, boolean[] mayRetainStreamFlags,
+                SampleStream[] streams, boolean[] streamResetFlags, long positionUs) {
+            for (int i = 0; i < selections.length; ++i) {
+                if (selections[i] != null) {
+                    stream =
+                            new ExoPlayerSampleStream(allocator, selections[i].getSelectedFormat());
```
jasonzhangxx: Will `selectTracks()` be called more than once? And do we need to re-create a new stream every time?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,126 @@
...
+    ExoPlayerSampleStream(Allocator allocator, Format format) {
+        sampleQueue = SampleQueue.createWithoutDrm(allocator);
+        sampleQueue.format(format);
+    }
+
+    void discardBuffer(long positionUs, boolean toKeyframe) {
+        sampleQueue.discardTo(positionUs, toKeyframe, false);
+    }
+
+    void writeSample(byte[] samples, int size, long timestamp, boolean isKeyFrame) {
```
jasonzhangxx: `writeSample()` might come from starboard player thread. Will that be safe for ExoPlayerSampleStream?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        @Override
+        public long getNextLoadPositionUs() {
+            return getBufferedPositionUs();
+        }
+
+        @Override
+        public boolean continueLoading(@NonNull LoadingInfo loadingInfo) {
+            return stream == null || !stream.endOfStreamWritten();
```
jasonzhangxx: According to the description of MediaPeriod.continueLoading function, it says `@return True if progress was made, meaning that {@link #getNextLoadPositionUs()} will return a different value than prior to the call. False otherwise.`, please make sure if we should always return true before eos here.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,126 @@
...
+import dev.cobalt.util.Log;
+import java.io.IOException;
+
+/** Queues encoded media to be retrieved by the player renderers */
+@UnstableApi
+public class ExoPlayerSampleStream implements SampleStream {
+    private final SampleQueue sampleQueue;
+    private boolean endOfStream = false;
+    private final ParsableByteArray sampleData = new ParsableByteArray();
+    private static final long MAX_BUFFER_DURATION_US = 30 * 1000 * 1000; // 30 seconds.
```
jasonzhangxx: The written samples will be copied and stored in sampleQueue, which is not managed by DecoderBufferAllocator, and input buffers will be released. So, the actual memory used for encoded buffers might exceed DecoderBufferAllocator budget. That might increase the memory pressure.

30s is too large, maybe we can try 5s. And we should add a comment to mention that special memory handling with exoplayer (or we can consider to file a follow-up bug).

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,50 @@
...
+@JNINamespace("starboard")
+public class ExoPlayerManager {
+    private Context context;
+
+    public ExoPlayerManager(Context context) {
+        this.context = context;
+    }
+
+    @CalledByNative
+    public synchronized ExoPlayerBridge createExoPlayerBridge(long nativeExoPlayerBridge,
```
jasonzhangxx: Do we still need ExoPlayerManager? With the existing code, it seems that we can call `ExoPlayerBridge.createExoPlayerBridge()` and  `ExoPlayerBridge.destroy()` without the manager class.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,50 @@
...
+            ExoPlayerMediaSource audioSource, ExoPlayerMediaSource videoSource, Surface surface,
+            boolean enableTunnelMode) {
+        return ExoPlayerBridge.createExoPlayerBridge(nativeExoPlayerBridge, context, audioSource, videoSource,
+                surface, enableTunnelMode);
+    }
+
+    @CalledByNative
+    public synchronized void destroyExoPlayerBridge(ExoPlayerBridge exoPlayerBridge) {
```
jasonzhangxx: There used to be a list of exo players, which has been removed. So, do we still need `ExoPlayerManager.destroyExoPlayerBridge`? Can it just call ExoPlayerBridge.destroy() directly?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,50 @@
...
+import android.view.Surface;
+import dev.cobalt.util.Log;
+import org.jni_zero.CalledByNative;
+import org.jni_zero.JNINamespace;
+
+/** Creates and destroys ExoPlayer instances */
+@JNINamespace("starboard")
+public class ExoPlayerManager {
+    private Context context;
+
```
jasonzhangxx, line 92: nit: I recommend to add a comment to explain why we need an ExoPlayerManager on top. My understanding is, 1. Creating ExoPlayer needs the application Context which can stored in manager class, 2. We can add the ExoPlayer list back in the future if it's needed (maybe for encrypted contents?).

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+        @Override
+        public void discardBuffer(long positionUs, boolean toKeyframe) {
+            stream.discardBuffer(positionUs, toKeyframe);
+        }
+
+        @Override
+        public long readDiscontinuity() {
+            if (seekTimeUs != C.TIME_UNSET) {
+                long positionToReport = seekTimeUs;
+                seekTimeUs = C.TIME_UNSET;
```
jasonzhangxx: Why do we need to report discontinuity after seek? A comment of the explanation might be needed here to help reader understand it.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+    public MediaItem getMediaItem() {
+        return this.mediaItem;
+    }
+
+    @Override
+    public void maybeThrowSourceInfoRefreshError() throws IOException {}
+
+    @Override
+    public MediaPeriod createPeriod(MediaPeriodId id, Allocator allocator, long startPositionUs) {
+        mediaPeriod = new ExoPlayerMediaPeriod(format, allocator);
```
jasonzhangxx: My understanding is `createPeriod` will be only called once for `ExoPlayerMediaSource`. If that's true, we should report an error if it's called more than once, so that if that goes wrong, we'll get the error message in qoe logs. Or we can throw out an exception if we can.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+        return mediaPeriod;
+    }
+
+    @Override
+    public void releasePeriod(MediaPeriod mediaPeriod) {
+        if (mediaPeriod != this.mediaPeriod) {
+            Log.e(TAG, "Called MediaSource.releasePeriod on an unknown MediaPeriod.");
```
jasonzhangxx: Same as `createPeriod`, we should report an error.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+        }
+        // Ignore the passed-in MediaPeriod and call the ExoPlayerMediaPeriod directly. As there's
+        // only a single MediaPeriod, this will match the passed MediaPeriod.
+        this.mediaPeriod.destroySampleStream();
```
jasonzhangxx: I'm wondering why we explicitly call `mediaPeriod.destroySampleStream` but still keep `this.mediaPeriod` here. If we'll only have one period, we can just release `this.mediaPeriod` here.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,206 @@
...
+
+    public void writeEndOfStream() {
+        mediaPeriod.writeEndOfStream();
+    }
+
+    public boolean canAcceptMoreData() {
+        return mediaPeriod.canAcceptMoreData();
+    }
+
+    private static final class ExoPlayerMediaPeriod implements MediaPeriod {
```
jasonzhangxx: `ExoPlayerMediaPeriod` is complicated enough to be moved to a separate file.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    void writeSample(byte[] samples, int size, long timestamp, boolean isKeyFrame) {
+        sampleData.reset(samples, size);
+        int flags = 0;
+        if (isKeyFrame) {
+            flags |= C.BUFFER_FLAG_KEY_FRAME;
+        }
+
+        sampleQueue.sampleData(sampleData, size);
```
jasonzhangxx: Let's add a TODO here to mention that this code introduces an extra memory copy, which has the potential to be optimized.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerCodecUtil.java
```diff
@@ -0,0 +1,131 @@
...
+        MediaCodecBridge.ColorInfo info = new MediaCodecBridge.ColorInfo(colorRange, colorSpace,
+                colorTransfer, primaryRChromaticityX, primaryRChromaticityY, primaryGChromaticityX,
+                primaryGChromaticityY, primaryBChromaticityX, primaryBChromaticityY,
+                whitePointChromaticityX, whitePointChromaticityY, maxMasteringLuminance,
+                minMasteringLuminance, maxCll, maxFall, false);
+        return new ColorInfo.Builder()
+                .setColorRange(info.colorRange)
+                .setColorSpace(info.colorStandard)
+                .setColorTransfer(info.colorTransfer)
+                .setHdrStaticInfo(info.hdrStaticInfo.array())
```
jasonzhangxx: We should add an `byte[] getHdrStaticInfo()` function in MediaCodecUtil.java, and use that function here and in ColorInfo ctor. We shouldn't use MediaCodecBridge.ColorInfo here just to extract the hdr static info.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerCodecUtil.java
```diff
@@ -0,0 +1,131 @@
...
+
+        if (colorInfo != null) {
+            builder.setColorInfo(colorInfo);
+        }
+
+        return new ExoPlayerMediaSource(builder.build());
+    }
+
+    @CalledByNative
+    public static ColorInfo createHdrColorInfo(int colorRange, int colorSpace, int colorTransfer,
```
jasonzhangxx: nit: let's rename it to `createExoPlayerColorInfo`, to make it straightforward for exoplayer only.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerCodecUtil.java
```diff
@@ -0,0 +1,131 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import org.jni_zero.CalledByNative;
+
+/** Creates MediaSources for specified codec configurations */
+public class ExoPlayerCodecUtil {
+    @CalledByNative
+    public static ExoPlayerMediaSource createAudioMediaSource(
```
jasonzhangxx: We can move `createAudioMediaSource`, `createHdrColorInfo` and `createVideoMediaSource` to ExoPlayerManager.java, and move FilteringMediaCodecSelector to ExoPlayerBridge.java. Then we no longer need ExoPlayerCodecUtil.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+
+    private final class DroppedFramesListener implements AnalyticsListener {
+        @Override
+        public void onDroppedVideoFrames(
+                @NonNull EventTime eventTime, int droppedFrames, long elapsedMs) {
+            ExoPlayerBridgeJni.get().onDroppedVideoFrames(mNativeExoPlayerBridge, droppedFrames);
+        }
+    }
+
+    public static ExoPlayerBridge createExoPlayerBridge(long nativeExoPlayerBridge, Context context,
```
jasonzhangxx: As we have  another `ExoPlayerManager.createExoPlayerBridge`, this function looks redundant to me.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+        }
+
+        DefaultTrackSelector trackSelector = new DefaultTrackSelector(context);
+        trackSelector.setParameters(new DefaultTrackSelector.ParametersBuilder(context)
+                                            .setTunnelingEnabled(enableTunnelMode)
+                                            .build());
+
+        DefaultRenderersFactory renderersFactory =
+                new DefaultRenderersFactory(context).setMediaCodecSelector(
+                        new ExoPlayerCodecUtil.FilteringMediaCodecSelector());
```
jasonzhangxx: nit: do we need to create new `DefaultRenderersFactory` every time?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+
+        if (audioSource != null) {
```
jasonzhangxx: nit:  the check looks redundant to me.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+    private final long mNativeExoPlayerBridge;
+    private Handler exoplayerHandler;
+    private long lastPlaybackPosUsec = 0;
+    private volatile float playbackRate = 0.0f;
+    private final ExoPlayerListener playerListener;
+    private final DroppedFramesListener droppedFramesListener;
+
+    private static final long PLAYER_RELEASE_TIMEOUT_MS = 300;
+    private static final int MAX_BUFFER_DURATION_MS = 30000; // 30 seconds
+    private static final int MIN_BUFFER_DURATION_MS = 450;
```
jasonzhangxx: Let's add a comment for the variables above and how they're selected.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+        if (enableTunnelMode) {
+            Log.i(TAG, "Tunnel mode is enabled for this playback.");
+        }
+
+        DefaultTrackSelector trackSelector = new DefaultTrackSelector(context);
+        trackSelector.setParameters(new DefaultTrackSelector.ParametersBuilder(context)
+                                            .setTunnelingEnabled(enableTunnelMode)
+                                            .build());
+
+        DefaultRenderersFactory renderersFactory =
```
jasonzhangxx: nit: do we need to create new `DefaultRenderersFactory` every time?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+                         .build();
+
+        playerListener = new ExoPlayerListener();
+        droppedFramesListener = new DroppedFramesListener();
+        exoplayerHandler.post(() -> {
+            player.addListener(playerListener);
+            player.addAnalyticsListener(droppedFramesListener);
+            player.setMediaSource(playbackMediaSource);
+            player.setVideoSurface(surface);
+            player.prepare();
```
jasonzhangxx: Will any of the functions above get error? If there's any error, should it continue to next step and would it cause other problem?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+            updatePlaybackPos();
+        });
+    }
+
+    private synchronized void updatePlaybackPos() {
```
jasonzhangxx: If `getCurrentPosition` can be called on other thread, we should call it directly in `getCurrentPositionUsec`, so that the current position returned can be more accurate.

If `getCurrentPosition` has to be called on `exoplayerHandler` every 150ms, we should also store the system time when `lastPlaybackPosUsec` is updated, and then in `getCurrentPosition`, return `lastPlaybackPosUsec + wall_time_difference`.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+        if (!isAbleToProcessCommands()) {
+            return;
+        }
+
+        // getCurrentPosition() returns milliseconds, convert to microseconds
+        lastPlaybackPosUsec = player.getCurrentPosition() * 1000;
```
jasonzhangxx: Android doc of `getCurrentPosition` says "This method must only be called if COMMAND_GET_CURRENT_MEDIA_ITEM is available.". Could you help me understand how to check if that command is available and if we need to check it?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+        exoplayerHandler.postDelayed(this::updatePlaybackPos, 150);
+    }
+
+    public void destroy() {
```
jasonzhangxx: nit: to make the function name more consistent with other media components, I feel `release` can be a better name.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+        if (!isAbleToProcessCommands()) {
+            Log.e(TAG, "Unable to destroy player with NULL ExoPlayer.");
+            return;
+        }
+
+        exoplayerHandler.post(() -> {
```
jasonzhangxx: By calling  `ExoPlayerBridge.destroy()`, upon returning, native callbacks (like onError, onEnded, onDroppedVideoFrames) should no longer be invoked.

But `player.stop()`, `player.release()` below are done asynchronously, so you need extra guardian code in ExoPlayerBridge to ensure not to invoke callback after `ExoPlayerBridge.destroy()` is returned and before underlying exoplayer is completely released.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+            player.stop();
+            player.removeListener(playerListener);
+            player.removeAnalyticsListener(droppedFramesListener);
+            player.release();
+            player = null;
+            exoplayerHandler = null;
```
jasonzhangxx: Do we need to reset `exoplayerHandler` to null?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+                type == ExoPlayerRendererType.AUDIO ? audioMediaSource : videoMediaSource;
+        return mediaSource.canAcceptMoreData();
+    }
+
+    @CalledByNative
+    private synchronized long getCurrentPositionUsec() {
+        return lastPlaybackPosUsec;
+    }
+
+    private boolean isAbleToProcessCommands() {
```
jasonzhangxx: nit: `isValid` makes more sense to me.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,336 @@
...
+        return player != null && exoplayerHandler != null;
```
jasonzhangxx: If `exoplayerHandler` is always non null, we won't need to check it here.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I moved the videosource and surface checks to ExoPlayerManager.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I've injected it from the ExoPlayerManager

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: These functions are not documented to throw errors/exceptions, and they are unrelated to each other. However, prepare may throw an exception that should be caught by the player listener

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: Removed it

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: Good tip, I've added handling for the time difference. All player methods must be called on the on`exoPlayerHandler`, so I can't do it synchronously.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: There is a [function](https://developer.android.com/reference/kotlin/androidx/media3/common/Player#getAvailableCommands())  to check if commands are available. Here we can safely assume this command is always available as everything in Java is happening locally - according to Gemini it should only be unavailable at times if the Player were downloading media data itself from a server.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: As the listeners are removed before `player.release()`, I think any pending callbacks from them will be suppressed, so we wouldn't see any more after `release()`

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I don't think so, removed

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerCodecUtil.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerCodecUtil.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerCodecUtil.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
osagie98: Added a comment, your assumptions are correct

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
osagie98: The native code now calls release() directly

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: `destroySampleStream()` is a custom ExoPlayerMediaPeriod function that can't be called on the passed-in MediaPeriod, which is why I ignore it. I've added a check to ensure that the MediaPeriods match

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: No, it'll only be called once per playback during player preparation. After this returns, the player will signal `kSbPlayerStateInitialized`

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: I don't believe it's needed, it's used if the samplestream contains samples prior to the playback start position. It returns C.TIME_UNSET, which should have the same effect as returning the seek position.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: That deals with fetching media over the network, which the player isn't doing here. It should load at all times unless the Allocator buffer is full, I've added an additional check here for CanAcceptMoreData()

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: The `SampleQueue` is thread-safe. My local variable `endOfStream` is not, so I've marked it volatile.

cobalt/android/BUILD.gn
```diff
@@ -45,25 +47,11 @@ generate_jni("jni_headers") {
...
-  srcjar_deps = [ ":browser_jni_headers" ]
-  deps = [ "//third_party/jni_zero:jni_zero_java" ]
-}
-
 android_library("cobalt_main_java") {
   testonly = true
   resources_package = "dev.cobalt.coat"
   deps = [
-    ":cobalt_browser_java",
+    ":jni_headers",
```
jasonzhangxx: Please make sure that change is intended, and let @johnxwork @zhongqiliang to review this change.

cobalt/android/BUILD.gn
```diff
@@ -83,6 +71,11 @@ android_library("cobalt_main_java") {
     "//third_party/androidx:androidx_annotation_annotation_java",
     "//third_party/androidx:androidx_core_core_java",
     "//third_party/androidx:androidx_customview_customview_java",
+    "//third_party/androidx:androidx_media3_media3_common_java",
+    "//third_party/androidx:androidx_media3_media3_datasource_java",
+    "//third_party/androidx:androidx_media3_media3_decoder_java",
+    "//third_party/androidx:androidx_media3_media3_exoplayer_java",
+    "//third_party/androidx:androidx_media3_media3_extractor_java",
```
jasonzhangxx, line 102: Is extractor used? If not, let's not include it.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+
+    // Duration after which an error is raised if the player doesn't release its resources.
+    private static final long PLAYER_RELEASE_TIMEOUT_MS = 300;
+    private static final int MAX_BUFFER_DURATION_MS = 5000; // 5 seconds.
+    // Based on the requirement to begin playback if at least 500 milliseconds of media has been
+    // appended.
+    private static final int MIN_BUFFER_DURATION_MS = 450;
+
+    /** Filters software video codecs from ExoPlayer codec selection. */
+    private static final class FilteringMediaCodecSelector implements MediaCodecSelector {
```
jasonzhangxx: FilteringMediaCodecSelector has been moved to ExoPlayerManager.java.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+import org.jni_zero.NativeMethods;
+
+/** Facilitates communication between the native ExoPlayerBridge and the Java ExoPlayer */
+@JNINamespace("starboard")
+@UnstableApi
+public class ExoPlayerBridge {
+    private ExoPlayer player;
+    private ExoPlayerMediaSource audioMediaSource;
+    private ExoPlayerMediaSource videoMediaSource;
+    private final long mNativeExoPlayerBridge;
```
jasonzhangxx: Only `mNativeExoPlayerBridge` uses a different naming convention.  Let's keep the consistency.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+    private long seekTimeUsec = 0;
+    private final ExoPlayerListener playerListener;
+    private final DroppedFramesListener droppedFramesListener;
+
+    // Duration after which an error is raised if the player doesn't release its resources.
+    private static final long PLAYER_RELEASE_TIMEOUT_MS = 300;
+    private static final int MAX_BUFFER_DURATION_MS = 5000; // 5 seconds.
+    // Based on the requirement to begin playback if at least 500 milliseconds of media has been
+    // appended.
+    private static final int MIN_BUFFER_DURATION_MS = 450;
```
jasonzhangxx: The comment says it needs at least 500ms of data, but `MIN_BUFFER_DURATION_MS` is 450, which is confusing.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
jasonzhangxx: `player.removeListener(playerListener)` will be executed on `exoplayerHandler` thread, which could happen after 200ms (random value just for example). During that 200ms, `onPlaybackStateChanged` might be called and it will invoke  `ExoPlayerBridgeJni.get().onReady(mNativeExoPlayerBridge);`, at that time, will  `mNativeExoPlayerBridge` still be available?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+
+        this.playbackRate = playbackRate;
+
+        if (playbackRate == 0.0f) {
+            pause();
+        } else if (playbackRate > 0.0f) {
+            exoplayerHandler.post(() -> {
+                player.setPlaybackParameters(new PlaybackParameters(playbackRate, 1.0f));
+            });
+            play();
```
jasonzhangxx: If `pause()` is called first, calling `setPlaybackRate( 1.0f)` would start the playback.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    @CalledByNative
+    private synchronized long getCurrentPositionUsec() {
+        if (isProgressing) {
+            return lastPlaybackPosUsec
+                    + ((SystemClock.elapsedRealtime() - playbackPosLastUpdatedMsec) * 1000);
+        }
+
+        return Math.max(lastPlaybackPosUsec, seekTimeUsec);
```
jasonzhangxx: The code would introduce gaps when switching between playing <-> paused. For example, `lastPlaybackPosUsec` is 500ms, if `pause()` is called at 600ms, calling `getCurrentPositionUsec()` at 550ms would get right time of 550ms, but calling `getCurrentPositionUsec() `at 650ms would get 500ms. After resumed, `getCurrentPositionUsec()` would return a timestamp larger than 600ms.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    public void destroy() {
+        sampleQueue.release();
+    }
+
+    public boolean canAcceptMoreData() {
+        long bufferedDurationUs =
+                sampleQueue.getLargestQueuedTimestampUs() - sampleQueue.getFirstTimestampUs();
+        return bufferedDurationUs < MAX_BUFFER_DURATION_US - MEMORY_PRESSURE_THRESHOLD_US;
```
jasonzhangxx: Can you help me understand what `MEMORY_PRESSURE_THRESHOLD_US` is for?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I added a comment to clarify

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I removed calls to `pause()` and `play()` as that should be determined by native code.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,173 @@
...
+        @Override
+        public List<MediaCodecInfo> getDecoderInfos(
+                String mimeType, boolean requiresSecureDecoder, boolean requiresTunnelingDecoder) {
+            List<MediaCodecInfo> defaultDecoderInfos = null;
+            try {
+                defaultDecoderInfos =
+                        androidx.media3.exoplayer.mediacodec.MediaCodecUtil.getDecoderInfos(
+                                mimeType, requiresSecureDecoder, requiresTunnelingDecoder);
+            } catch (androidx.media3.exoplayer.mediacodec.MediaCodecUtil.DecoderQueryException e) {
+                Log.i(dev.cobalt.media.Log.TAG,
```
Awallky: Instead of inlining the TAGs here, can we do something similar to what we did in the other classes?
That is, use a static import and just use `TAG` inline?

Awallky: Actually, now that I think about it, I think it is more standard practice to create a unique TAG for each individual class. Can you make a local `TAG` unique to each class?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import org.jni_zero.CalledByNative;
+import org.jni_zero.JNINamespace;
+import org.jni_zero.NativeMethods;
+
+/** Facilitates communication between the native ExoPlayerBridge and the Java ExoPlayer */
+@JNINamespace("starboard")
+@UnstableApi
+public class ExoPlayerBridge {
```
Awallky, line 45: Can you please add Javadoc to all non-trivial public methods, especially those in ExoPlayerBridge.java that are called by native code (@CalledByNative)? This will improve maintainability, as it clarifies the contract between the Java and C++ layers.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,365 @@
...
+    private final Handler exoplayerHandler;
+    // The following variables are accessed on both the Handler and native threads
+    private volatile long lastPlaybackPosUsec = 0;
+    private volatile long playbackPosLastUpdatedMsec = 0;
+    private volatile float playbackRate = 0.0f;
+    private volatile boolean isProgressing = false;
+
+    private long seekTimeUsec = 0;
+    private final ExoPlayerListener playerListener;
+    private final DroppedFramesListener droppedFramesListener;
```
Awallky: Could you make sure these variables adhere to the AOSP naming conventions?
https://source.android.com/docs/setup/contribute/code-style#follow-field-naming-conventions
Can you also check if any of the other variables need updates for this?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaPeriod.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    @Override
+    public long getAdjustedSeekPositionUs(long positionUs, SeekParameters seekParameters) {
+        return positionUs;
+    }
+
+    @Override
+    public long getBufferedPositionUs() {
+        return stream != null ? stream.getBufferedPositionUs() : 0l;
```
Awallky: nit: Consider changing the `0l` default value to `0L` as it improves readability.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I see what you mean now, I've added a volatile boolean `isReleasing` to prevent further Jni calls once `release()` is called

cobalt/android/BUILD.gn
osagie98: Removed

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: I've added a comment, instead of waiting for the queued duration to exceed `MAX_BUFFER_DURATION_US`, the queue will refuse to accept more data if the duration is beyond `MAX_BUFFER_DURATION_US - MEMORY_PRESSURE_THRESHOLD_US`, to ensure the queued duration never exceeds the maximum

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I now update the playback position after pausing, so calls to get the position while paused should be accurate. In play() I call `updateplaybackpos()`, which should update the relevant variables before the player reaches the `isPlaying` state, as this should all be executed on the same Handler.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
osagie98: I can do that, though all media files share the same `TAG` `starboard_media`, maybe @jasonzhangxx or @xiaomings can weigh in?

cobalt/android/BUILD.gn
osagie98: Unintended, I reverted it.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
jasonzhangxx: I don't think we have unique TAG for each individual class. Currently, we have `dev.cobalt.util.Log` and `dev.cobalt.media.Log`, which have tags `starboard` and `starboard_media`. I don't think it's necessary to keep two tags, so we can combine them into one, but let's not change it in this PR.

@osagie98 I think the original comment asks to import `static dev.cobalt.media.Log.TAG` and then use `TAG` here.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,392 @@
...
+            if (mIsReleasing) {
+                return;
+            }
+
+            mIsProgressing = isPlaying;
+            ExoPlayerBridgeJni.get().onIsPlayingChanged(mNativeExoPlayerBridge, isPlaying);
+        }
+
+        @Override
+        public void onPlayerError(@NonNull PlaybackException error) {
```
jasonzhangxx: Let's check `mIsReleasing` in `onPlayerError` too, to keep the consistency.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,392 @@
...
+
+        // getCurrentPosition() returns milliseconds, convert to microseconds
+        mLastPlaybackPosUsec = mPlayer.getCurrentPosition() * 1000;
+        mPlaybackPosLastUpdatedMsec = SystemClock.elapsedRealtime();
+        mExoplayerHandler.postDelayed(this::updatePlaybackPos, 150);
+    }
+
+    @CalledByNative
+    public void release() {
+        mIsReleasing = true;
```
jasonzhangxx: `mIsReleasing` is actually `mIsReleased`.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -51,58 +50,62 @@ public class ExoPlayerBridge {
...
-                case Player.STATE_BUFFERING:
-                case Player.STATE_IDLE:
-                    return;
-                case Player.STATE_READY:
-                    ExoPlayerBridgeJni.get().onReady(mNativeExoPlayerBridge);
-                    break;
-                case Player.STATE_ENDED:
-                    ExoPlayerBridgeJni.get().onEnded(mNativeExoPlayerBridge);
-                    break;
+            if (mPlayer == null || mIsReleased) {
```
jasonzhangxx: Why does the check need `mPlayer == null`? Lots of other functions in the file have this check.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -216,54 +222,38 @@ public void release() {
...
         ExoPlayerMediaSource mediaSource =
-                type == ExoPlayerRendererType.AUDIO ? mAudioMediaSource : mVideoMediaSource;
-        if (mediaSource == null || mIsReleasing) {
+            sample.getType() == ExoPlayerRendererType.AUDIO ? mAudioMediaSource : mVideoMediaSource;
+        if (mediaSource == null || mIsReleased) {
             reportError(
-                    String.format("Tried to write %s sample while ExoPlayer is in an invalid state",
-                            type == ExoPlayerRendererType.AUDIO ? "audio" : "video"));
+                String.format("Tried to write %s sample while ExoPlayer is in an invalid state",
+                    sample.getType() == ExoPlayerRendererType.AUDIO ? "audio" : "video"));
```
jasonzhangxx: After reporting an error, it should return immediately and not continue to write sampels.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I messed up the find and replace, it now checks that the player isn't null

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoMediaDrmWrapper.java
```diff
@@ -0,0 +1,222 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * A wrapper around {@link FrameworkMediaDrm} that captures the session ID when a session is opened.
+ */
+@UnstableApi
+public final class ExoMediaDrmWrapper implements ExoMediaDrm {
```
jasonzhangxx: Is it still needed in this pr? Should it be moved to drm support pr?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerDrmBridge.java
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * Manages DRM sessions for ExoPlayer and acts as a bridge between ExoPlayer's MediaDrmCallback
+ * and the native SbDrmSystem.
+ *
+ * This class implements {@link MediaDrmCallback} to handle key and provisioning requests,
+ * forwarding them to the native Starboard implementation. It also owns and manages the lifecycle
+ * of the {@link DefaultDrmSessionManager} used by ExoPlayer.
+ */
+@JNINamespace("starboard")
+public class ExoPlayerDrmBridge {
```
jasonzhangxx: Is it still needed in this pr? Should it be moved to drm support pr?

osagie98: No, removed

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoMediaDrmWrapper.java
osagie98: Removed, seems I removed it from the main pr https://github.com/youtube/cobalt/pull/5899 then accidentally copied it back over to this one. This file is completely gone as the Drm support pr uses an updated implementation.

jasonzhangxx (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,214 @@
...
+     * Helper to check if a mime type represents a passthrough audio codec.
+     */
+    private static boolean isPassthroughCodec(String mime) {
+        return MimeTypes.AUDIO_AC3.equals(mime)
+                || MimeTypes.AUDIO_E_AC3.equals(mime)
+                || MimeTypes.AUDIO_E_AC3_JOC.equals(mime);
+    }
+
+    /**
+     * A {@link MediaCodecSelector} that disables internal software decoders for AC3, E-AC3,
```
xiaomings: Do "internal software decoders" refer to the software decoders included inside ExoPlayer or the platform MediaCodec decoders (which aren't always software)?

The comment also doesn't capture what the function actually does.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,214 @@
...
+                return Collections.emptyList();
+            }
+            List<MediaCodecInfo> defaultDecoderInfos = null;
+            try {
+                defaultDecoderInfos =
+                        MediaCodecUtil.getDecoderInfos(
+                                mimeType, requiresSecureDecoder, requiresTunnelingDecoder);
+            } catch (MediaCodecUtil.DecoderQueryException e) {
+                Log.i(TAG, String.format("MediaCodecUtil.getDecoderInfos() error %s", e));
+                return defaultDecoderInfos;
```
xiaomings: Is `defaultDecoderInfos` null here?  If it is, consider returning null to be explict.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
```diff
@@ -0,0 +1,214 @@
...
+            }
+            // Skip video decoder filtering for emulators.
+            if (IsEmulator.isEmulator()) {
+                Log.i(TAG, "Allowing all available decoders for emulator.");
+                return defaultDecoderInfos;
+            }
+
+            List<MediaCodecInfo> hardwareVideoDecoderInfos = new ArrayList<>();
```
xiaomings: nit: this should be moved into the if block.

If there are only audio and video decoder, we could also check "audio/" here and return first.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSample.java
```diff
@@ -0,0 +1,71 @@
...
+  private final int mFlags;
+
+  @CalledByNative
+  public ExoPlayerMediaSample(
+      ByteBuffer samples,
+      int size,
+      long timestamp,
+      boolean isKeyFrame,
+      int type) {
+    this.mSamples = samples;
```
xiaomings: Just wanted to double check why it's named as "samples" instead of "sample".

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSample.java
```diff
@@ -0,0 +1,71 @@
...
+    this.mSize = size;
+    this.mTimestampUsec = timestamp;
+    this.mType = type;
+
+    int flags = 0;
+    if (isKeyFrame) {
+      flags |= C.BUFFER_FLAG_KEY_FRAME;
+    }
+    this.mFlags = flags;
```
xiaomings: this.mFlags = isKeyFrame ? C.BUFFER_FLAG_KEY_FRAME : 0;

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
```diff
@@ -0,0 +1,128 @@
...
+ * A custom {@link BaseMediaSource} that receives encoded media data from the native Starboard
+ * layer and provides it to ExoPlayer.
+ *
+ * This source is designed for a single-period lifecycle, mapping to a single audio or video
+ * stream provided by the native application.
+ */
+public final class ExoPlayerMediaSource extends BaseMediaSource {
+    private final Format mFormat;
+    // Guarded by mLock
+    private ExoPlayerMediaPeriod mMediaPeriod;
```
xiaomings: I am not familiar with the Java convention.  In C++, usually we declare the variable right after mLock and add the comment after `mMediaPeriod` on the same line.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,211 @@
...
+            if (!mBuffer.hasRemaining()) {
+                return C.RESULT_END_OF_INPUT;
+            }
+            int readLength = Math.min(length, mBuffer.remaining());
+            mBuffer.get(buffer, offset, readLength);
+            return readLength;
+        }
+    }
+
+    private static final long MAX_BUFFER_DURATION_US = 5 * 1000 * 1000; // 5 seconds.
```
xiaomings: This can occupy a non-trivial amount of memory for video streams.  If we make a copy of the data inside the ExoPlayer integration, this may affect overall memory allocation.  It should be less of a concern if it only holds a reference to the DecoderBuffer but doesn't allocate extra memory.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,211 @@
...
+    public void writeSample(ExoPlayerMediaSample sample) {
+        synchronized (mLock) {
+            try {
+                ByteBufferDataReader dataReader = new ByteBufferDataReader(sample.getSamples());
+                int bytesWritten = 0;
+                while (bytesWritten < sample.getSize()) {
+                    int result =
+                            mSampleQueue.sampleData(dataReader, sample.getSize() - bytesWritten, true);
+                    if (result == C.RESULT_END_OF_INPUT) {
+                        throw new IOException("Unexpected end of input from ByteBuffer");
```
xiaomings: If this is the only place throwing `IOException` to be caught below, can we turn this into an if statement and return here directly?

Also, if this can't happen under our implementation, consider add a comment to make it explicit.  I am not sure if there is a way in Java code to CHECK, which is better.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
```diff
@@ -0,0 +1,211 @@
...
+            @NonNull FormatHolder formatHolder, @NonNull DecoderInputBuffer buffer, int readFlags) {
+        synchronized (mLock) {
+            if (!mSampleQueue.isReady(mEndOfStream)) {
+                return C.RESULT_NOTHING_READ;
+            }
+
+            try {
+                return mSampleQueue.read(formatHolder, buffer, readFlags, mEndOfStream);
+            } catch (DecoderInputBuffer.InsufficientCapacityException e) {
+                Log.i(TAG,
```
xiaomings: Shall we report an error here?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    @CalledByNative
+    private void seek(long seekToTimeUsec) {
+        if (mPlayer == null || mIsReleased) {
+            reportError("Cannot seek with NULL or released ExoPlayer.");
+            return;
+        }
+        mExoPlayerHandler.post(() -> {
+          mPlayer.seekTo(seekToTimeUsec / 1000);
```
xiaomings, line 294: It's 2 space indent here but 4 space indent for post in the function above.  Has it gone through the Java code formatter?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,381 @@
...
+     */
+    @CalledByNative
+    private synchronized long getCurrentPositionUsec() {
+        long currentPositionUsec = mLastPlaybackPosUsec;
+        if (mIsProgressing) {
+            currentPositionUsec += (long) ((SystemClock.elapsedRealtime() - mPlaybackPosLastUpdatedMsec)
+                            * mPlaybackRate * 1000);
+        }
+
+        return Math.max(currentPositionUsec, mSeekTimeUsec);
```
xiaomings: If this is necessary, would this work when seeking backwards?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,381 @@
...
+    }
+
+    @CalledByNative
+    private void setPlaybackRate(float playbackRate) {
+        if (mPlayer == null || mIsReleased) {
+            reportError("Cannot set playback rate with NULL or released ExoPlayer");
+            return;
+        }
+
+        this.mPlaybackRate = playbackRate;
```
xiaomings: nit: the explicit "this." isn't consistent with other code in this file.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,381 @@
...
+            mPlayer.setMediaSource(playbackMediaSource);
+            mPlayer.setVideoSurface(surface);
+            mPlayer.prepare();
+            mPlayer.play();
+        });
+    }
+
+    private synchronized void updatePositionAnchor() {
+        if (!mIsReleased && mPlayer != null) {
+            mLastPlaybackPosUsec = mPlayer.getCurrentPosition() * 1000;
```
xiaomings: If ExoPlayer's internal implementation is fast enough (they probably have similar adjustment internally), we may use it directly inside getCurrentPositionUsec() without all the extra code.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,381 @@
...
+
+    @CalledByNative
+    public void release() {
+        mIsReleased = true;
+        if (mPlayer == null) {
+            Log.w(TAG, "Attempted to destroy ExoPlayer after it has already been released.");
+            return;
+        }
+
+        mExoPlayerHandler.post(() -> {
```
xiaomings: If this is executed asynchronously, is there any side effect of the mPlayer after the release() returns?  Does the user of the class expect all side effects to be gone on release() return?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSource.java
osagie98: No particular convention for Java, but I've made it more readable.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerMediaSample.java
osagie98: Should be sample, fixed

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: I removed the catch, so any exception thrown will be caught by the listener in the ExoPlayerBridge, which reports an error

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: I added an assert, but kept the exception, since in production code we'd want to report this an an error

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerSampleStream.java
osagie98: The player creates its own copy once it enters the samplequeue. I've lowered the duration for video (2 or 3 seconds depending on the stream), but I'm also working on a design change to eliminate the samplequeue and additional copies

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
osagie98: These refer to platform mediacodec decoders. This function disables the on-board decoders for Dolby Digital so they can only be handled with passthrough. It also disables all software mediacodec video decoders, except for emulators that must use software decoders. I've updated the comment.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerManager.java
osagie98: It should only be audio and video, though I'm not 100% sure, so it's a catch all line to be safe for all non video mime types

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: The player is finished after release returns. The native ExoPlayerBridge won't call the player again, and doesn't expect any more callbacks from the player. Only during mPlayer.release() there may be internal thread synchronization issues, which are handled without an error, though it shows up in logs.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: `updatePositionAnchor` runs on the handler thread since all player methods like `mPlayer.getCurrentPosition()` must run on the handler thread, so I can't query the position synchronously in `getCurrentPositionUsec()`. I added the extra code to account for the lack of synchronization.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: I was using a different formatter, I've formatted all the files with `google-java-format` which eliminated a lot of whitespace.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: .

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
osagie98: This works when seeking backwards but it's likely not necessary. I've removed it but I'm testing to see if it causes any issues.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/ExoPlayerBridge.java
```diff
@@ -0,0 +1,426 @@
...
+    if (mPlayer == null || mIsReleased) {
+      reportError("Cannot seek with NULL or released ExoPlayer.");
+      return;
+    }
+    mExoPlayerHandler.post(
+        () -> {
+          mPlayer.seekTo(seekToTimeUsec / 1000);
+          updatePositionAnchor();
+        });
+    mSeekTimeUsec = seekToTimeUsec;
```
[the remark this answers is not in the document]
osagie98: This may be possible but I'd rather not block the native side

osagie98:
Closing in favor of https://github.com/youtube/cobalt/pull/5899

PR #7995, opened 2025-11-11
[tvOS] Adapt to removal of Starboard headers post-C25

rakuco:
* starboard/memory.h was removed in #4126
* starboard/string.h was removed in #4106
* starboard/directory.h was removed in #4106
* starboard/common/optional.h was removed in #5986

Bug: 456692793

PR #8095, opened 2025-11-18
tvOS: Clean up remaining usages of removed Starboard headers

Gyuyoung:
After #7995 removed the removed Starboard headers, there were still a few places that continued to include them. This change cleans up those remaining references.

Bug: 456692793


Gyuyoung:

PR #8112, opened 2025-11-19
android: Close all Cobalt services on H5vcc unload

zhongqiliang:
Introduce a mechanism to shut down all active Cobalt platform
services when the H5vccSystem object is destructed. This
change addresses a bug where platform-specific services were
not reliably closed during page unloads or reloads (e.g.,
Kabuki reload), leading to potential resource leaks and
unexpected behavior on Android. Ensure proper cleanup via
StarboardBridge on the Java side.
Bug: 454969656

Change-Id: I466a52be4575173a98bed5cb749654ae0153444e


andrewsavage1:
The commit in this change should be a `git cherry-pick -x` of the original change

zhongqiliang:
okey, use this one instead https://github.com/youtube/cobalt/pull/8113

PR #8114, opened 2025-11-19
Enable updater for evergreen platforms

santoshmahto7:
It was disabled before due to crash reported on RDK, but seems to work well in latest code as verified.
So the  commit to disable the updater is reverted here.

Issue: 460219182


lamdbui:
@yuying-y , can you take point in reviewing this one?

borongc:
I still have the crash after this PR with `devel` build on RDK, where my device is connected to the wired ethernet.

PR #8116, opened 2025-11-19
Add a simpler SBDStarboardApplication implementation to ApplicationDarwin

rakuco:
The tvOS part
-------------

Tests were previously being run with SbRunStarboardMain() and SBDUIKitApplicationMain(). This starts the whole starboard::Application main loop and causes the tests to never stop running on tvOS, as it is not a Starboard platform and we do not integrate with the Starboard main loop infrastructure at all.

ApplicationDarwin essentially needs to be made more similar to ApplicationAndroid, which also does not need to integrate with the Starboard event loop. Instead of dealing with PlaybackCapabilities and AudioSink in Initialize() and Teardown(), which are not called in our case, use the constructor and destructor instead. SetCommandLine() is now called manually in the constructor as well.

A new inner Objective-C++ interface owned by ApplicationDarwin is now responsible for initializing the main objects needed by the tvOS code (window manager, DRM manager, EGL adapter etc).

All these changes are integrated into the Starboard testing infrastructure by creating a ::testing::Environment subclass whose purpose is to create and destroy ApplicationDarwin instances between each test. It is called by test_main.cc.

This is enough to get a lot of NPLB tests to work fine without further integration with Starboard, the only downside is that the Starboard test infrastructure on tvOS now depends on `//base` to avoid having to reimplement all the iOS integration done in `base/test`.

With these changes, it is now possible to remove uikit_application.mm altogether.

Patch by @abhijeetk and myself.

The general changes
-------------------

Supporting this means touching some platform-independent bits as well: since `test_main.cc` now depends on tvOS's `test_support` when targeting the platform, it makes sense to move it to a separate source set to avoid duplicating the same checks and dependencies in different locations.

Bug: 456692793
Bug: 433180638

### DIFF
--- a/starboard/BUILD.gn
+++ b/starboard/BUILD.gn
@@ -235,13 +235,13 @@ if (current_toolchain == starboard_toolchain) {
   }
   target(_starboard_unittests_target_type, "starboard_unittests") {
     testonly = true
-    deps = [ ":starboard_test_main" ]
     defines = [ "STARBOARD_IMPLEMENTATION_TEST" ]
 
     write_runtime_deps =
         "$root_build_dir/starboard_unittests_wrapper.runtime_deps"
 
-    deps += [
+    deps = [
+      ":starboard_test_main",
       "//starboard:starboard_headers_only",
       "//starboard/common:common_test",
       "//testing/gtest",


rakuco:

starboard/tvos/shared/application_darwin.mm
```diff
@@ -18,23 +18,72 @@
...
+
+void EmptyHandleCallback(const SbEvent* event) {
+  SB_LOG(ERROR) << "Starboard event DISCARDED:" << event->type;
+}
+
 }  // namespace
 
-void ApplicationDarwin::Initialize() {
+struct ApplicationDarwin::ObjCStorage {
+  ObjCApplication* __strong objc_application;
```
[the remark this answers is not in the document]
rakuco: Wrong, see "Struct Data Members" in https://google.github.io/styleguide/cppguide.html#Variable_Names

starboard/tvos/shared/application_darwin.h
fayev-g, line 1: nit: Were were planning to drop both `uikit` && `darwin` naming conventions? Just wondering if we wanted this to just be application_tvos. (The android one is application_android.) Can do this in another PR also

fayev-g: Overall this PR is great and glad to see uikit_application.mm cleaned up

rakuco: We're definitely getting rid of the uikit bits. As for "darwin", I guess it also makes sense to drop it, although it hadn't crossed my mind (there are many other commits touching this file so I kind of got used to the name :-)

rakuco:
@andrewsavage1 could you take a look at the platform-independent part of this change? Is there someone else who should also review this PR?

andrewsavage1:
> @andrewsavage1 could you take a look at the platform-independent part of this change? Is there someone else who should also review this PR?

Let's have @oxve take a look

rakuco:
> Platform independent parts lgtm

@andrewsavage1 can you take another look and (hopefully) lgtm this one too?

starboard/BUILD.gn
```diff
@@ -224,13 +235,13 @@ if (current_toolchain == starboard_toolchain) {
   }
   target(_starboard_unittests_target_type, "starboard_unittests") {
     testonly = true
-    sources = [ "//starboard/common/test_main.cc" ]
+    deps = [ ":starboard_test_main" ]
```
andrewsavage1: nit: combine with the rest of the deps

rakuco: done

PR #8117, opened 2025-11-19
[tvOS] Flatten some leftover namespaces in application_darwin.h

rakuco:
Bug: 456692793


rakuco:

PR #8119, opened 2025-11-19
media: Add AllocateOnDemand experiment

johnedocampo:
This PR exposes the `allocate_on_demand` setting for DecoderBufferAllocator to H5VCC experiments. This allows us to explicitly enable `allocate_on_demand` at runtime.

To make this happen, the following changes have been made:

- |is_memory_pool_allocated_on_demand_| is no longer const, as this experiment enables the possibility of it being changed while `DecoderBufferAllocator` exists.
- Locking the mutexes for `Suspend()` and `Resume()` earlier in the function call, as it is possible that |is_memory_pool_allocated_on_demand_| is changing when these functions are called.
- `cobalt_content_renderer_client.cc` will now query for this specific setting, and will update the value as necessary.

Bug: 459199176

### DIFF
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -260,9 +260,12 @@ void DecoderBufferAllocator::SetAllocateOnDemand(bool enabled) {
 
   is_memory_pool_allocated_on_demand_ = enabled;
   // If we enable |is_memory_pool_allocated_on_demand_|, we should try to
-  // reset the the strategy.
+  // reset the strategy.
   if (is_memory_pool_allocated_on_demand_ && strategy_ &&
       strategy_->GetAllocated() == 0) {
+    LOG(INFO) << "Freed " << strategy_->GetCapacity()
+              << " bytes of media buffer pool since allocator now allocates on "
+                 "demand.";
     strategy_.reset();
   }
 }


media/starboard/decoder_buffer_allocator.cc
```diff
@@ -249,4 +248,24 @@ void DecoderBufferAllocator::SetEnabled(bool enabled) {
...
+  if (is_memory_pool_allocated_on_demand_) {
+    // If we enable |is_memory_pool_allocated_on_demand_|, we should try to
+    // reset the the strategy.
+    if (strategy_ && strategy_->GetAllocated() == 0) {
+      strategy_.reset();
+    }
+  } else {
+    EnsureStrategyIsCreated();
+  }
+}
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -249,4 +248,27 @@ void DecoderBufferAllocator::SetEnabled(bool enabled) {
...
+                 "DecoderBufferAllocator";
+    if (strategy_ && strategy_->GetAllocated() == 0) {
+      strategy_.reset();
+    }
+  } else {
+    LOG(INFO) << "Disabling is_memory_pool_allocated_on_demand_ for "
+                 "DecoderBufferAllocator";
+    EnsureStrategyIsCreated();
+  }
+}
```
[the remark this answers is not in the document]
johnedocampo: Done. Since both SetEnabled and SetAllocateOnDemand both use the function, refactored both functions to reduce duplication.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -249,4 +249,27 @@ void DecoderBufferAllocator::SetEnabled(bool enabled) {
...
+  if (is_memory_pool_allocated_on_demand_) {
+    // If we enable |is_memory_pool_allocated_on_demand_|, we should try to
+    // reset the the strategy.
+    if (strategy_ && strategy_->GetAllocated() == 0) {
+      strategy_.reset();
+    }
+  } else {
+    EnsureStrategyIsCreated();
+  }
+}
```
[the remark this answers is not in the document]
johnedocampo: From my understanding, this should not be an issue for the following reasons:

1. Free() will still be called even if allocate_on_demand is on. From my understanding, setting the flag on does not automatically free the memory.

2. SbMemoryFree is not called at all.

media/starboard/decoder_buffer_allocator.h
```diff
@@ -81,7 +82,7 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
   void TryFlushAllocationLog_Locked() EXCLUSIVE_LOCKS_REQUIRED(mutex_);
 #endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
 
-  const bool is_memory_pool_allocated_on_demand_;
+  bool is_memory_pool_allocated_on_demand_;
```
kjyoun: EXCLUSIVE_LOCKS_REQUIRED(mutex_);

kjyoun (commented):
nit: can you update title prefix("[media]") to "media: " ?

go/cobalt-style#commit-message

media/starboard/decoder_buffer_allocator.h
johnedocampo: Done. Line is now:

```cpp
bool is_memory_pool_allocated_on_demand_ GUARDED_BY(mutex_);
```

johnedocampo:
> nit: can you update title prefix("[media]") to "media: " ?
> 
> go/cobalt-style#commit-message

Done.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -38,6 +38,10 @@ namespace {
 // be different.
 const size_t kSmallAllocationThreshold = 512;
 
+const char* GetEnabledString(bool value) {
```
kjyoun: ToString isn't more concise?

Since this is used only inside this file, I think we can choose concise one.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -249,4 +249,27 @@ void DecoderBufferAllocator::SetEnabled(bool enabled) {
...
+  is_memory_pool_allocated_on_demand_ = enabled;
+
+  if (is_memory_pool_allocated_on_demand_) {
+    // If we enable |is_memory_pool_allocated_on_demand_|, we should try to
+    // reset the the strategy.
+    if (strategy_ && strategy_->GetAllocated() == 0) {
+      strategy_.reset();
+    }
+  } else {
+    EnsureStrategyIsCreated();
```
kjyoun: I think we don't need to call this? since other method (e.g. Allocate) calls this if necessary.

media/starboard/decoder_buffer_allocator.cc
johnedocampo: Done, changed back to ToString.

media/starboard/decoder_buffer_allocator.cc
johnedocampo: You're right, this call is not needed due to other methods ensuring it is created. Removed and simplified the if statement.

kjyoun (approved):

johnxwork (approved):

PR #8124, opened 2025-11-19
Implement ShellPlatformTestDataAura

haozheng-cobalt:
Extract Aura test client initialization from ShellPlatformDataAura into
a new derived class, ShellPlatformTestDataAura. This separates test-
specific Aura configurations, such as test focus and parenting clients,
from the core ShellPlatformDataAura implementation. The base class's
constructor is updated to optionally create the focus client, allowing
ShellPlatformTestDataAura to manage its own test-specific clients. This
improves modularity and provides a dedicated setup for browser tests.

Test: compiled and ran Linux build, this PR does not impact ATV as aura is only used for desktop platforms like Linux.
Bug: 446895973

### DIFF
--- a/cobalt/shell/browser/shell_platform_data_aura.cc
+++ b/cobalt/shell/browser/shell_platform_data_aura.cc
@@ -100,35 +100,44 @@ class FocusRules : public wm::BaseFocusRules {
 
 ShellPlatformDataAura::ShellPlatformDataAura(const gfx::Size& initial_size,
                                              bool create_focus_client) {
+  CHECK(aura::Env::GetInstance());
+
 #if BUILDFLAG(IS_OZONE)
+  // Setup global display::Screen singleton.
   screen_ = std::make_unique<aura::ScreenOzone>();
-#endif
+#endif  // BUILDFLAG(IS_OZONE)
 
-  host_ = aura::WindowTreeHost::Create(ui::PlatformWindowInitProperties{
-      gfx::Rect(gfx::Point(), initial_size), "Cobalt"});
+  ui::PlatformWindowInitProperties properties;
+  properties.bounds = gfx::Rect(initial_size);
+
+  host_ = aura::WindowTreeHost::Create(std::move(properties));
   host_->InitHost();
-
-  // TODO(https://crbug.com/1336055): this is needed for
-  // mouse_cursor_overlay_controller_browsertest.cc on cast_shell_linux as
-  // currently, when is_castos = true, the views toolkit isn't used.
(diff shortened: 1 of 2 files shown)


cobalt/shell/browser/shell_platform_data_aura.cc
```diff
@@ -105,11 +95,11 @@ class FocusRules : public wm::BaseFocusRules {
     return true;
   }
 };
-#endif  // !defined(RUN_BROWSER_TESTS)
 
 }  // namespace
 
-ShellPlatformDataAura::ShellPlatformDataAura(const gfx::Size& initial_size) {
+ShellPlatformDataAura::ShellPlatformDataAura(const gfx::Size& initial_size,
+                                             bool create_focus_client) {
```
Awallky, line 102: What is the purpose of this parameter? It seems mostly unused, and always set to true, unless I am missing something.

Awallky: It looks like you are trying to control the focus_client_ ctor with that bool, but never changing it. Did you forget to change it in the test class?

cobalt/testing/browser_tests/browser/shell_platform_test_data_aura.cc
```diff
@@ -0,0 +1,50 @@
...
+#include "ui/aura/test/test_focus_client.h"
+#include "ui/aura/test/test_window_parenting_client.h"
+#include "ui/wm/core/cursor_loader.h"
+#include "ui/wm/core/default_activation_client.h"
+
+namespace content {
+
+ShellPlatformTestDataAura::ShellPlatformTestDataAura(
+    const gfx::Size& initial_size)
+    : ShellPlatformDataAura(initial_size, false) {
```
haozheng-cobalt, line 29: @Awallky for the test class the create_focus_client param is set to false, then the prod focus client will not be generated and the test focus client will be generated instead.

cobalt/shell/browser/shell_platform_data_aura.cc
haozheng-cobalt: The param is used, please see the other comment where I at you.

cobalt/testing/browser_tests/browser/shell_platform_test_data_aura.cc
Awallky: Oh, interesting. You essentially create a wrapper ctor to invoke the multi-param. My search was looking for a top-level ctor, hence the miss. Thanks!

PR #8127, opened 2025-11-19
fix: Complete memory leak fix for URLLoaderHeaderClient

sherryzy:
This commit provides the second part of a two-part fix to resolve
memory leak related to `CobaltTrustedURLLoaderHeaderClient`. 
                                                                     
The initial fix in PR#8068 addressed a leak within `CobaltTrustedHeaderClient`
by making it self-owned.  However, that change did not address 
the lifecycle of the factory object, `CobaltTrustedURLLoaderHeaderClient`,
which was itself being leaked from a long-lived vector.

This commit addresses the remaining leak by:
1. Refactoring `CobaltTrustedURLLoaderHeaderClient` to be fully
self-owned via `mojo::MakeSelfOwnedReceiver`. 
2. Removing the `cobalt_header_clients_` vector from 
`CobaltContentBrowserClient`, which was indefinitely storing
 pointers to every created client instance.
3. Updating the corresponding unit tests to reflect the new 
self-owned lifecycle management.                                                

Together, these two commits fully resolve the memory leak by      
ensuring both the client and the factory that creates them are    
properly garbage collected.                                       

Bug: 454095852

### DIFF
--- a/cobalt/browser/client_hint_headers/cobalt_trusted_url_loader_header_client.cc
+++ b/cobalt/browser/client_hint_headers/cobalt_trusted_url_loader_header_client.cc
@@ -19,10 +19,8 @@
 namespace cobalt {
 namespace browser {
 
-CobaltTrustedURLLoaderHeaderClient::CobaltTrustedURLLoaderHeaderClient(
-    mojo::PendingReceiver<network::mojom::TrustedURLLoaderHeaderClient>
-        receiver)
-    : receiver_(this, std::move(receiver)) {}
+CobaltTrustedURLLoaderHeaderClient::CobaltTrustedURLLoaderHeaderClient() =
+    default;
 
 void CobaltTrustedURLLoaderHeaderClient::OnLoaderCreated(
     int32_t request_id,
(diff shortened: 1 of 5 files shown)


zhongqiliang (approved):

zhongqiliang:
you should use `git cherry-pick -x <SHA>` to cherry pick.

andrewsavage1:
This causes a memory leak for Starboard/linux but not android?

kjyoun:
Can't we submit this PR to main and cherry-pick int 26.eap, following the normal flow?
(if there is no urgency to submit this PR directly to 26.eap first)

sherryzy:
@kjyoun The 3p workstream is currently prioritizing development on 26.eap over origin/main. Our standard procedure is to implement changes on 26.eap and then cherry-pick them to origin/main. I haven't tested this on origin/main yet. And last time I checked origin/main, Cobalt crashes immediately.

sherryzy:
@yell0wd0g Gemini generated the title and description. It must copied everything from https://github.com/youtube/cobalt/pull/8068.
Sorry for the confusion! T_T

sherryzy:
@andrewsavage1 I hypothesize that this issue is also causing a memory leak on Android. The leak is reliably reproduced in my current testing environment using an HTML page that reloads frequently. This scenario (frequent reloads) may not have been a high-priority test case for the Android, which could explain why this memory leak has not been detected in previous testing.

PR #7688, opened 2025-10-22
Reland flush on pause

ghost:
Revert "Revert flush local storage https://github.com/youtube/cobalt/pull/7669 and https://github.com/youtube/cobalt/pull/7587" https://github.com/youtube/cobalt/pull/7526

Reverted due to m138 conflict.
See: https://github.com/youtube/cobalt/pull/7526

First commit is revert. Other commits are fixing the conflicts.

Issue: 418842688
Issue: 371433332

PR #7587, opened 2025-10-16
[Android] Flush cookies and local storage when activity is paused

ghost:
Similar to c25 which flushes cookies and localStorage on the cobalt freeze event, this PR adds flushing cookies and localStorage to the Android activity onPause() handler.

In addition to this, flush to cookies and localStorage is performed when h5vcc.system.exit() is called in support of YTS tests. In production the flush may be redundant but will act as a no-op if it is.

Issue: 371433332

PR #7526, opened 2025-10-13
[Rebase M138 Android] - Fix Android compilation

sideb0ard:
Bug: 418842688

Biggest change is due to:
jni_zero: Migrating all JNI annotations in /base to jni_zero copies
https://chromium-review.googlesource.com/c/chromium/src/+/4935910

I used the following sed command to replace:
'find cobalt/ starboard/ -type f -exec sed -i 's/org.chromium.base.annotations/org.jni_zero/g' {} +'

Also:
jni_zero: moving jni_int_wrapper.h to jni_zero
https://chromium-review.googlesource.com/c/chromium/src/+/4894452

Spanification of jni_array - no longer accepts a pointer and size, must be a span:
https://chromium-review.googlesource.com/c/chromium/src/+/5731813

merge after https://github.com/youtube/cobalt/pull/7522

PR #8130, opened 2025-11-20
Reland: flush on pause

kaidokert:
Picking up reland from #7688

---

Revert "Revert flush local storage https://github.com/youtube/cobalt/pull/7669 and https://github.com/youtube/cobalt/pull/7587" https://github.com/youtube/cobalt/pull/7526

Reverted due to m138 conflict.
See: https://github.com/youtube/cobalt/pull/7526

First commit is revert. Other commits are fixing the conflicts.

Issue: 418842688
Issue: 371433332
Bug: 451215656
Bug: 450936106
Bug: 456836622
Bug: 444451968
Bug: 448130250


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -327,7 +331,7 @@ protected ImeAdapterImpl getImeAdapterImpl() {
   @Override
   protected void onCreate(Bundle savedInstanceState) {
     // Record the application start timestamp.
-    mTimeInNanoseconds = System.nanoTime();
+    mTimeInNanoseconds = System.nano-time();
```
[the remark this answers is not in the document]
kaidokert: Fixed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -433,6 +446,7 @@ protected void onStop() {
       webContents.updateWebContentsVisibility(Visibility.HIDDEN);
       // document.onfreeze event
       webContents.onFreeze();
+      CobaltContentBrowserClient.flushCookiesAndLocalStorage();
```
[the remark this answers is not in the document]
kaidokert: Addressed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -70,6 +71,7 @@
 import org.chromium.net.NetworkChangeNotifier;
 import org.chromium.ui.base.ActivityWindowAndroid;
 import org.chromium.ui.base.IntentRequestTracker;
+import org.chromium.net.NetworkChangeNotifier;
```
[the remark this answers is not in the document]
kaidokert: Fixed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -454,6 +468,10 @@ protected void onResume() {
       rootView.requestFocus();
       Log.i(TAG, "Request focus on the root view on resume.");
     }
+    WebContents webContents = getActiveWebContents();
+    if (webContents != null) {
+      CobaltContentBrowserClient.dispatchFocus();
+    }
```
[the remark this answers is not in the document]
kaidokert: Fixed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -422,6 +426,15 @@ protected void onStart() {
...
   }
 
+  @Override
+  protected void onPause() {
+    WebContents webContents = getActiveWebContents();
+    if (webContents != null) {
+      CobaltActivityJni.get().flushCookiesAndLocalStorage();
+    }
+    super.onPause();
+  }
```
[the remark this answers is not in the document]
kaidokert: The JNI call issue is fixed, the other one is invalid with current Starboard design

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -187,29 +211,29 @@ void H5vccSystemImpl::GetUserOnExitStrategy(
...
+  auto* storage_partition = render_frame_host().GetStoragePartition();
+  CHECK(storage_partition);
+  // Flushes localStorage.
+  storage_partition->Flush();
+  auto* cookie_manager = storage_partition->GetCookieManagerForBrowserProcess();
+  CHECK(cookie_manager);
+  // Sequencing exit strategy after flushing delays performing exit strategy by
+  // 20ms when tested on a chromecast.
+  cookie_manager->FlushCookieStore(base::BindOnce(&PerformExitStrategy));
 }
```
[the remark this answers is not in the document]
kaidokert: That's currently still a relatively minimal duplication of a few lines.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -422,6 +425,15 @@ protected void onStart() {
     MemoryPressureMonitor.INSTANCE.enablePolling(false);
   }
 
+  @Override
+  protected void onPause() {
+    WebContents webContents = getActiveWebContents();
```
Awallky: nit: do we need to hold the variable if we are not using it? Can we just inline the function call?

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
```diff
@@ -187,29 +211,29 @@ void H5vccSystemImpl::GetUserOnExitStrategy(
...
-      SbSystemRequestConceal();
-      break;
-    case h5vcc_system::mojom::UserOnExitStrategy::kNoExit:
-      return;
-  }
-#elif BUILDFLAG(IS_ANDROIDTV)
-  JNIEnv* env = base::android::AttachCurrentThread();
-  StarboardBridge* starboard_bridge = StarboardBridge::GetInstance();
-  starboard_bridge->RequestSuspend(env);
-#elif BUILDFLAG(IS_IOS_TVOS)
```
andrewsavage1, line 206: Make sure to add the TVOS back into the copied function

Awallky (approved):
I am wondering if an onPause event in Android would lead to kSbEventTypeBlur or kSbEventTypeFreeze? If so, the calls to flush cookies becomes redundant.
But this is more of a re-land than a new change, so we can worry about optimizing things later after fixing cookie loss issues.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -345,6 +383,63 @@ void CobaltContentBrowserClient::WillCreateURLLoaderFactory(
...
+  content::RenderFrameHost* primary_main_frame =
+      web_contents->GetPrimaryMainFrame();
+  CHECK(primary_main_frame);
+
+  std::u16string javascript = u"window.dispatchEvent(new Event('" +
+                              base::UTF8ToUTF16(event_name) + u"'));";
+  base::ConcurrentCallbacks<base::Value> concurrent;
+  primary_main_frame->ForEachRenderFrameHost(
+      [&concurrent, &javascript](content::RenderFrameHost* rfh) {
+        rfh->ExecuteJavaScript(javascript, concurrent.CreateCallback());
```
andrewsavage1: Is there no way to directly call the handling function instead of injecting javascript?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
kaidokert: Not necessarily, but when you step through this code in a debugger - i find it slightly more convenient to see a null value in a temporary value.

Also if you need to quickly add remove logging to / DCHECK for this nullness, it's less churn.

Do we have a style guide recommendation for this type of things ?

Awallky: I could not find anything that says it is explicitly disallowed. I think [go/clarity-code-best-practice](https://goto.google.com/clarity-code-best-practice) is pretty bare and [go/java-style](https://goto.google.com/java-style) does not mention anything about the choice between inlining or not. 

I think the argument for debuggability is fair as well. I am okay with either choice.

cobalt/app/cobalt.cc
```diff
@@ -163,14 +164,34 @@ void SbEventHandle(const SbEvent* event) {
       g_exit_manager = nullptr;
       break;
     }
-    case kSbEventTypeBlur:
-    case kSbEventTypeFocus:
+    case kSbEventTypeBlur: {
```
johnxwork, line 167: I think CoAT does not dispatch Starboard events now. It uses Chromium's event handler directly.

johnxwork: This is actually attempting to fix the problem I just described in the meeting, blur/focus is missing when the app is still in the foreground but behind a popup: https://docs.google.com/document/d/1p-dPwyS3RR85UydeNPNzZvv8Jii-IxEV0cXahByRvlE/edit?tab=t.0#bookmark=id.r2wlthwl7d3g

cobalt/android/apk/app/src/main/java/dev/cobalt/browser/CobaltContentBrowserClient.java
```diff
@@ -0,0 +1,41 @@
...
+import org.jni_zero.NativeMethods;
+
+/** Exposes the native CobaltContentBrowserClient class functionality. */
+public class CobaltContentBrowserClient {
+  @JNINamespace("cobalt")
+  @NativeMethods
+  interface Natives {
+    void flushCookiesAndLocalStorage();
+    void dispatchBlur();
+    void dispatchFocus();
```
zhongqiliang, line 27: These APIs are not used in CobaltActivity.java, how would you hook this up?

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -345,6 +383,63 @@ void CobaltContentBrowserClient::WillCreateURLLoaderFactory(
...
+    std::move(callback).Run();
+    return;
+  }
+  auto* web_contents = web_contents_observer_->web_contents();
+  CHECK(web_contents);
+  content::RenderFrameHost* primary_main_frame =
+      web_contents->GetPrimaryMainFrame();
+  CHECK(primary_main_frame);
+
+  std::u16string javascript = u"window.dispatchEvent(new Event('" +
```
zhongqiliang: CobaltActivity has a evaluateJavaScript() method, you probably can directly run the js code there to skip the JNI layers.

cobalt/browser/cobalt_content_browser_client.cc
kaidokert: `RenderWidgetHost` has Blur/Focus() methods, i'm trying to work out if those can be used directly.

cobalt/browser/h5vcc_system/h5vcc_system_impl.cc
kaidokert: added back, thanks

cobalt/app/cobalt.cc
kaidokert: > I think CoAT does not dispatch Starboard events now. 

I understand. The original implementation that i'm trying to re-land addressed both non-Android and Android together here. Maybe i should separate them for clarity ? 

> It uses Chromium's event handler directly.

Please have another look - is the current Android implementation wrong here ? Is there another place in Cobalt shell code where pause/receive are processed that is a more direct way to do this, rather than custom JNI ?

cobalt/android/apk/app/src/main/java/dev/cobalt/browser/CobaltContentBrowserClient.java
kaidokert: Please see : `CobaltActtivity.java`

```
  @Override
  protected void onPause() {
    CobaltContentBrowserClient.dispatchBlur();
    super.onPause();
  }
```

cobalt/browser/cobalt_content_browser_client.cc
kaidokert: I actually don't think i ended up needing `evaluateJavaScript` at all, as i can call `Focus()` and `Blur()` directly. Please take another pass through

cobalt/browser/cobalt_content_browser_client.cc
andrewsavage1: Looks way better using those, thanks

cobalt/app/cobalt.cc
andrewsavage1: > I understand. The original implementation that i'm trying to re-land addressed both non-Android and Android together here. Maybe i should separate them for clarity ?

I had trouble finding these changes in the original PRs, and the changes aren't in the 26.eap branch, but I assume we'll want them there as well (that is, they don't seem specific to m138). I think it makes sense to separate out parts of this PR that are specific to m138 and parts that will need to be cherry picked to 26.* branches—it looks like there are a few other parts that would need to be cherry picked besides this?

johnxwork: > Please have another look - is the current Android implementation wrong here ? Is there another place in Cobalt shell code where pause/receive are processed that is a more direct way to do this, rather than custom JNI ?

Sorry my mistake, the handling here is correct. You have implementation for both ATV and 3P in this PR. And that should be the right thing to do.

PR #6214, opened 2025-06-27
Remove redundant functionalities in cobalt/shell/browser

haozheng-cobalt:
See tracker: go/cobalt-shell-optimization.

Things removed:
Permission Manager, Download Manager, Bluetooth Delegate, Federated Permission.

New size:
```
haozheng@haozheng:~/chromium/src$ ls -lh out/chrobalt-android-arm/obj/cobalt/shell
total 712K
drwxr-xr-x 2 haozheng primarygroup 4.0K Jul  2 02:08 content_shell_app
-rw-r--r-- 1 haozheng primarygroup    0 Jul  2 02:08 content_shell_app_deps.stamp
-rw-r--r-- 1 haozheng primarygroup    0 Jul  2 00:35 content_shell_app.inputdeps.stamp
-rw-r--r-- 1 haozheng primarygroup 219K Jul  2 02:07 content_shell_app.ninja
drwxr-xr-x 2 haozheng primarygroup 4.0K Jul  2 02:08 content_shell_lib
-rw-r--r-- 1 haozheng primarygroup    0 Jul  2 02:08 content_shell_lib_deps.stamp
-rw-r--r-- 1 haozheng primarygroup    0 Jul  2 00:35 content_shell_lib.inputdeps.stamp
-rw-r--r-- 1 haozheng primarygroup 225K Jul  2 02:07 content_shell_lib.ninja
-rw-r--r-- 1 haozheng primarygroup 5.6K Jul  2 02:08 libcontent_shell_app.a
-rw-r--r-- 1 haozheng primarygroup 247K Jul  2 02:08 libcontent_shell_lib.a
```

Test: connect to a BT speaker and play a playback locally.

Bug: 427478103

PR #8139, opened 2025-11-20
cobalt shell: Remove dead iOS/tvOS Bluetooth code

rakuco:
Most of the Bluetooth code was removed from Cobalt Shell in #6214; remove some leftovers that caused build failures on tvOS.

Bug: 456692793
Bug: 427478103


rakuco:

PR #8140, opened 2025-11-20
[tvOS] Remove attachUiNavigationView from application_window.*

rakuco:
Follow-up to #8048, this code is never called now that the `ui_nav_*` have been removed.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #8141, opened 2025-11-20
[tvOS] Remove on_screen_keyboard.{mm,h} and related code

rakuco:
These files were used to implement the on-screen keyboard extension, which is not implemented by any other port in main and therefore probably is not necessary anymore.

Doing so allows us to get rid of a substantial amount of code from application_window.mm, with more cleanups to follow later.

Bug: 459507205


rakuco:

PR #8142, opened 2025-11-20
[tvOS] Remove libevent/ directory

rakuco:
This came from C25 and is not used in Chrobalt.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #8144, opened 2025-11-20
Remove RunFileChooser() implementation in //cobalt

Gyuyoung:
Cobalt is not intended to display any native UI. Therefore, the implementations of RunFileChooser() along with its helpers and listeners are no longer needed. This PR removes all of them from //cobalt.

Bug: 461675419


Gyuyoung:

fayev-g:
lgtm since my understanding is that we don't plan to rebase cobalt/shell onto upstream content/shell or anything, and these modifications make sense. As Adam referenced, hopefully it's not too bad to remove usages in tests if they're there.

fayev-g (approved):

Gyuyoung:
> LGTM. Thanks for the cleaup!

Do I still need to build cobalt_browsertests for Android?

rakuco:
> Do I still need to build cobalt_browsertests for Android?

I think it's a good idea. You also need to fix the lint issues (even though they aren't caused by this PR as far as I can see).

Gyuyoung:
> Can you just ensure that cobalt_browsertests builds properly before landing? Tests will not pass right now, due to an ongoing bug, but it should still build.
> 
> something like `autoninja -C out/android-arm_qa cobalt_browsertests` should work.

When I tried to build the target for Android, I faced the following error,

```
../../cobalt/testing/browser_tests/browser/shell_content_browser_test_client.cc:29:10: fatal error: 'cobalt/shell/common/shell_controller.test-mojom.h' file not found
   29 | #include "cobalt/shell/common/shell_controller.test-mojom.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
RBE Stats: down 0 B, up 0 B, 1 non zero exit
```

I think this is not related to this PR. FYI, I could build `cobalt_apk` target locally.

cobalt/shell/browser/shell_file_select_helper.cc
```diff
@@ -1,228 +0,0 @@
-// Copyright 2025 The Cobalt Authors. All Rights Reserved.
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
```
kaidokert, line 5: Linter is now complaining about missing copyright headers. If you are fully removing the file, lets fully delete it. 

If fully deleting it is complex because of other include paths, to get the linter job green, you may want to leave only the copyright header with a comment clarifying this file is intentionally empty.

Gyuyoung: This PR passed the linter after removing the files. Thanks.

rakuco:
@fayev-g could you enable auto-merge for this PR?

PR #8156, opened 2025-11-20
Add directory to store cookies and local storage.

ghost:
Bug: 371433332
Bug: 453724467

### DIFF
--- a/base/base_paths.h
+++ b/base/base_paths.h
@@ -10,7 +10,7 @@
 
 #include "build/build_config.h"
 
-#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if BUILDFLAG(IS_STARBOARD)
 #include "base/base_paths_starboard.h"
 #else
 #if BUILDFLAG(IS_WIN)
@@ -26,7 +26,7 @@
 #if BUILDFLAG(IS_POSIX)
 #include "base/base_paths_posix.h"
 #endif
-#endif  // BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#endif  // BUILDFLAG(IS_STARBOARD)
 
 namespace base {
 
--- a/cobalt/shell/common/shell_paths.cc
+++ b/cobalt/shell/common/shell_paths.cc
@@ -28,7 +28,7 @@
 #include "base/nix/xdg_util.h"
 #endif
 
-#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if BUILDFLAG(IS_STARBOARD)
 #include "starboard/configuration_constants.h"  // nogncheck
 #include "starboard/system.h"                   // nogncheck
 #endif
@@ -38,15 +38,12 @@ namespace content {
 namespace {
 
 bool GetDefaultUserDataDirectory(base::FilePath* result) {
-#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if BUILDFLAG(IS_STARBOARD)
   std::vector<char> path(kSbFileMaxPath, 0);
   bool success =
       SbSystemGetPath(kSbSystemPathFilesDirectory, path.data(), path.size());
   CHECK(success) << "kSbSystemPathFilesDirectory not defined.";
   *result = base::FilePath(path.data());
-#elif BUILDFLAG(IS_LINUX)
-  CHECK(base::PathService::Get(base::DIR_CACHE, result));
-  *result = result->Append("cobalt");
 #elif BUILDFLAG(IS_APPLE)
   CHECK(base::PathService::Get(base::DIR_APP_DATA, result));
   *result = result->Append("Chromium Content Shell");
--- a/starboard/CHANGELOG.md
+++ b/starboard/CHANGELOG.md
@@ -1329,3 +1329,9 @@ options have been left in Cobalt, though renamed to `cobalt_configuration.gypi`.
 
 Please see [contrib/README.md](contrib/README.md) for description of
 expectations for contents in this directory.
+
+### Introduces new system property kSbSystemPathFilesDirectory.
+
+Path to directory for permanent storage. Both read and write access are
+required. The path is used primarily for DOM localStorage and HTTP cookies
+persistance.
--- a/base/path_service_unittest.cc
+++ b/base/path_service_unittest.cc
@@ -147,12 +147,12 @@ TEST_F(PathServiceTest, Get) {
                                                  : &ReturnsValidPath,
                  key);
   }
-#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if BUILDFLAG(IS_STARBOARD)
   // In the three Starboard custom directories, DIR_CACHE should always be
   // valid while DIR_SYSTEM_FONTS and DIR_SYSTEM_FONTS_CONFIGURATION
   // can be invalid on some platforms.
   EXPECT_PRED1(ReturnsValidPath, DIR_CACHE);
-#else  // BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#else  // BUILDFLAG(IS_STARBOARD)
 #if BUILDFLAG(IS_WIN)
   for (int key = PATH_WIN_START + 1; key < PATH_WIN_END; ++key) {
     if (key == DIR_SYSTEM_TEMP) {
@@ -179,7 +179,7 @@ TEST_F(PathServiceTest, Get) {
     EXPECT_PRED1(ReturnsValidPath, key);
   }
 #endif  // BUILDFLAG(IS_WIN)
-#endif  // BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#endif  // BUILDFLAG(IS_STARBOARD)
 }
 
 // Tests that CheckedGet returns the same path as Get.


cobalt/shell/common/shell_paths.cc
```diff
@@ -28,16 +28,29 @@
...
-      env.get(), base::nix::kXdgConfigHomeEnvVar, base::nix::kDotConfigDir));
-  *result = config_dir.Append("content_shell");
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+  std::vector<char> path(kSbFileMaxPath, 0);
+  bool success =
+      SbSystemGetPath(kSbSystemPathFilesDirectory, path.data(), path.size());
+  DCHECK(success);
+  if (!success) {
+    DLOG(ERROR) << "kSbSystemPathFilesDirectory not defined.";
+    return false;
```
andrewsavage1: What happens when this returns false? Can the app reasonably continue? If not, the DCHECK above should be a CHECK

starboard/android/shared/system_get_path.cc
andrewsavage1, line 1: Does android use SbSystemGetPath at all?

starboard/system.h
```diff
@@ -69,6 +69,10 @@ typedef enum SbSystemPathId {
   // The directory should be used only for storing the updates.
   // See starboard/doc/evergreen/cobalt_evergreen_overview.md
   kSbSystemPathStorageDirectory,
+
+  // Path to the directory for permanent files. Used for cookies and
+  // localStorage.
+  kSbSystemPathFilesDirectory,
```
andrewsavage1, line 75: We should add an nplb test for this

starboard/android/shared/system_get_path.cc
ghost: Android TV uses SbSystemGetPath in the migration code to delete cobalt cache files after the migration is complete.

In normal operation Android TV does not use SbSystemGetPath. We will need Android to implement SbSystemGetPath for AOSP.

starboard/android/shared/system_get_path.cc
```diff
@@ -63,6 +62,16 @@ bool SbSystemGetPath(SbSystemPathId path_id, char* out_path, int path_size) {
...
       break;
     }
+    case kSbSystemPathFilesDirectory: {
+      if (starboard::strlcpy(path, g_app_files_dir, kPathSize) >= kPathSize) {
+        return false;
+      }
+      if (starboard::strlcat(path, "/files", kPathSize) >= kPathSize) {
+        return false;
+      }
+      mkdir(path, 0700);
```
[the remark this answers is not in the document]
ghost: This is copy-pasted code. We should fix all the uses or refactor with tests.

cobalt/shell/common/shell_paths.cc
```diff
@@ -28,16 +28,29 @@
...
+  }
+  *result = base::FilePath(path.data());
+#elif BUILDFLAG(IS_LINUX)
+  CHECK(base::PathService::Get(base::DIR_CACHE, result));
+  *result = result->Append("cobalt");
```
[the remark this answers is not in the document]
ghost: This is for development only. Putting the cookies and local storage in the .cache/cobalt directory is fine.

starboard/system.h
y4vor, line 75: Please update the `starboard/CHANGELOG.md`

starboard/android/shared/system_get_path.cc
```diff
@@ -63,6 +62,16 @@ bool SbSystemGetPath(SbSystemPathId path_id, char* out_path, int path_size) {
       mkdir(path, 0700);
       break;
     }
+    case kSbSystemPathFilesDirectory: {
+      if (starboard::strlcpy(path, g_app_files_dir, kPathSize) >= kPathSize) {
+        return false;
+      }
+      if (starboard::strlcat(path, "/files", kPathSize) >= kPathSize) {
```
y4vor, line 69: Isn't the g_app_files_dir directory already pointing to the correct place? The way I read the code this wold produce something like `<app_package>/files/files`

cobalt/shell/common/shell_paths.cc
```diff
@@ -28,16 +28,25 @@
...
-  base::FilePath config_dir(base::nix::GetXDGDirectory(
-      env.get(), base::nix::kXdgConfigHomeEnvVar, base::nix::kDotConfigDir));
-  *result = config_dir.Append("content_shell");
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+  std::vector<char> path(kSbFileMaxPath, 0);
+  bool success =
+      SbSystemGetPath(kSbSystemPathFilesDirectory, path.data(), path.size());
+  CHECK(success) << "kSbSystemPathFilesDirectory not defined.";
+  *result = base::FilePath(path.data());
+#elif BUILDFLAG(IS_LINUX)
```
y4vor: I think we want to keep the `IS_LINUX` branch unchanged.
To make it work for both modular/evergreen and monolithic Cobalt builds you can use  `IS_STARBOARD` instead of `IS_COBALT_HERMETIC_BUILD`.

starboard/android/shared/system_get_path.cc
ghost: Are you suggesting the following?

An Android Activity passes in two directories.

`::starboard::g_app_files_dir` is from `activity.getFilesDir()`.
https://developer.android.com/reference/android/content/Context#getFilesDir()

`::starboard::g_app_cache_dir` is from `activity.getCacheDir()`.
https://developer.android.com/reference/android/content/Context#getCacheDir()

For `g_app_files_dir`:
- `kSbSystemPathStorageDirectory` return `<g_app_files_dir>/storage`
- **`kSbSystemPathFilesDirectory` returns `<g_app_cache_dir>` (do not append `/files`)**

For `g_app_cache_dir`:
- `kSbSystemPathCacheDirectory` returns `<g_app_cache_dir>/cache`
- `kSbSystemPathDebugOutputDirectory` returns `<g_app_cache_dir>/log`
- `kSbSystemPathTempDirectory` returns `<g_app_cache_dir>`

cobalt/shell/common/shell_paths.cc
andrewsavage1: I think we would just want to replace `#elif BUILDFLAG(IS_LINUX)` with `#elif BUILDFLAG(IS_STARBOARD)` and remove the linux macro, there isn't a non-starboard linux platform

starboard/android/shared/system_get_path.cc
y4vor: OK, keep it as is.  I realized that it is better to be explicit and not take over the whole location.

cobalt/shell/common/shell_paths.cc
ghost: Okay, I'll add IS_STARBOARD to base_paths.h as well. That should clear things up.

y4vor (approved):

starboard/CHANGELOG.md
```diff
@@ -1329,3 +1329,9 @@ options have been left in Cobalt, though renamed to `cobalt_configuration.gypi`.
 
 Please see [contrib/README.md](contrib/README.md) for description of
 expectations for contents in this directory.
+
+### Introduces new system property kSbSystemPathFilesDirectory.
```
jellefoks, line 1333: FWIW, on Chromium on Linux cookies typically can be found at `${HOME}/.config/google-chrome/Default/Cookies`, where `Default` is the chrome profile, and the `${HOME}/.config` is the freedesktop base directory for `XDG_CONFIG_HOME`:

https://specifications.freedesktop.org/basedir/latest/ -> 
"_$XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used._"

So it has a 1:1 mapping to a commonly used standard, so I think it's good to have it as a path. These `SbSystemPathId` enum values all have the prefix `SbSystemPath` and postfix `Directory`. So the unique identifier here is `Files`, which I think is quite an ambiguous term. 

Would you consider naming it more similar to the freedesktop name and perhaps naming it `kSbSystemPathConfigDirectory`?

y4vor: I don't think we should use the XDG_CONFIG_HOME pattern. The current change is modeled after the Android APIs Context.getFilesDir() and Context.getCacheDir() which also maps well to Starboard https://b.corp.google.com/issues/371433332#comment18.

ghost:
@andrewsavage1: can you review for cobalt-embedder-owners?

base/base_paths.h
```diff
@@ -10,7 +10,7 @@
 
 #include "build/build_config.h"
 
-#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD) || BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: This is redundant, replace this whole thing with just `#if BUILDFLAG(IS_STARBOARD)`

cobalt/shell/common/shell_paths.cc
```diff
@@ -28,16 +28,22 @@
 #include "base/nix/xdg_util.h"
 #endif
 
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD) || BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: Replace with `#if BUILDFLAG(IS_STARBOARD)`

cobalt/shell/common/shell_paths.cc
```diff
@@ -28,16 +28,22 @@
...
 
 namespace {
 
 bool GetDefaultUserDataDirectory(base::FilePath* result) {
-#if BUILDFLAG(IS_LINUX)
-  std::unique_ptr<base::Environment> env(base::Environment::Create());
-  base::FilePath config_dir(base::nix::GetXDGDirectory(
-      env.get(), base::nix::kXdgConfigHomeEnvVar, base::nix::kDotConfigDir));
-  *result = config_dir.Append("content_shell");
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD) || BUILDFLAG(IS_STARBOARD)
```
andrewsavage1: replace

ghost:
Looking at the failure, my change does need to update the unittest.

```
../../base/path_service_unittest.cc:178:18: error: use of undeclared identifier 'PATH_POSIX_START'
  178 |   for (int key = PATH_POSIX_START + 1; key < PATH_POSIX_END; ++key) {
      |                  ^~~~~~~~~~~~~~~~
../../base/path_service_unittest.cc:178:46: error: use of undeclared identifier 'PATH_POSIX_END'
  178 |   for (int key = PATH_POSIX_START + 1; key < PATH_POSIX_END; ++key) {
      |                                              ^~~~~~~~~~~~~~
2 errors generated.
```

PR #8161, opened 2025-11-21
Enable in-memory update

yuying-y:
Download the update package to a large string in memory, instead of a file, in order to save required storage.

The vast majority of the code is ported from the original PR #849 to enable this feature, with the following modifications:
1. NetworkFetcher::DownloadToString() utilizes SimpleURLLoader instead of URLFetcher, to be consistent with the upstream change.
2. Increases SimpleURLLoader::kMaxBoundedStringDownloadSize to 100MB for Cobalt, because as tested locally, the QA package size is 70MB. The size range of the official release packages is unknown yet, because the infra is WIP. This number is open to change.
3. Relatively minor change in zip::Unzip(), to work with the updated upstream code.

Bug: 444006168
Bug: 431862767

### DIFF
--- a/cobalt/updater/network_fetcher.cc
+++ b/cobalt/updater/network_fetcher.cc
@@ -155,8 +155,45 @@ void NetworkFetcher::DownloadToString(
     ProgressCallback progress_callback,
     DownloadToStringCompleteCallback download_to_string_complete_callback) {
   LOG(INFO) << "NetworkFetcher::DownloadToString, url = " << url;
-  // TODO(b/444006168): implement this to support in-memory updates
-  NOTIMPLEMENTED();
+  CHECK(dst != nullptr);
+  dst_str_ = dst;
+
+  CHECK(!simple_url_loader_);
+  auto resource_request = std::make_unique<network::ResourceRequest>();
+  resource_request->url = url;
+  resource_request->method = "GET";
+  resource_request->load_flags = net::LOAD_DISABLE_CACHE;
+  simple_url_loader_ = network::SimpleURLLoader::Create(
+      std::move(resource_request), traffic_annotation);
+  simple_url_loader_->SetRetryOptions(
+      kMaxRetriesOnNetworkChange,
+      network::SimpleURLLoader::RetryMode::RETRY_ON_NETWORK_CHANGE);
+  simple_url_loader_->SetOnResponseStartedCallback(base::BindOnce(
+      &NetworkFetcher::OnResponseStartedCallback, base::Unretained(this),
+      std::move(response_started_callback)));
+  simple_url_loader_->SetOnDownloadProgressCallback(base::BindRepeating(
+      &NetworkFetcher::OnProgressCallback, base::Unretained(this),
+      std::move(progress_callback)));
+  simple_url_loader_->DownloadToString(
+      url_loader_factory_.get(),
+      base::BindOnce(&NetworkFetcher::OnDownloadToStringComplete,
+                     base::Unretained(this),
+                     std::move(download_to_string_complete_callback)),
+      network::SimpleURLLoader::kMaxBoundedStringDownloadSize);
+}
+
+void NetworkFetcher::OnDownloadToStringComplete(
+    DownloadToStringCompleteCallback download_to_string_complete_callback,
+    std::unique_ptr<std::string> response_body) {
+  if (!response_body) {
+    LOG(ERROR) << "DownloadToString failed to get response from a string";
+    dst_str_->clear();
+  } else {
+    *dst_str_ = std::move(*response_body);
+  }
+  std::move(download_to_string_complete_callback)
+      .Run(dst_str_, simple_url_loader_->NetError(),
+           simple_url_loader_->GetContentSize());
 }
 
 #else
@@ -178,7 +215,6 @@ void NetworkFetcher::DownloadToFile(
   simple_url_loader_->SetRetryOptions(
       kMaxRetriesOnNetworkChange,
       network::SimpleURLLoader::RetryMode::RETRY_ON_NETWORK_CHANGE);
-  simple_url_loader_->SetAllowPartialResults(true);
   simple_url_loader_->SetOnResponseStartedCallback(base::BindOnce(
       &NetworkFetcher::OnResponseStartedCallback, base::Unretained(this),
       std::move(response_started_callback)));
--- a/services/network/public/cpp/simple_url_loader.h
+++ b/services/network/public/cpp/simple_url_loader.h
@@ -80,7 +80,12 @@ class COMPONENT_EXPORT(NETWORK_CPP) SimpleURLLoader {
   };
 
   // The maximum size DownloadToString will accept.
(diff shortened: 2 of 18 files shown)


hlwarriner:
Great to see this! I started going through the changes but want to take my time. I'll provide any feedback by Monday at the latest.

andrewsavage1:
Can you make sure this is tested on RDK before it's submitted?

hlwarriner:
> Download the update package to a large string in memory, instead of a file, in order to save required storage.
> 
> The vast majority of the code is ported from the original PR #849 to enable this feature, with the following modifications:
> 
> 1. NetworkFetcher::DownloadToString() utilizes SimpleURLLoader instead of URLFetcher, to be consistent with the upstream change.
> 2. Increases SimpleURLLoader::kMaxBoundedStringDownloadSize to 100MB for Cobalt, because as tested locally, the QA package size is 70MB. The size range of the official release packages is unknown yet, because the infra is WIP. This number is open to change.
> 3. Relatively minor change in zip::Unzip(), to work with the updated upstream code.
> 
> Bug: 444006168 Bug: 431862767

Just have one initial comment, about point (2). Given the increased size of the update package from c25 to c26, and overall increased memory use in c26, we should definitely take a look at Kaido's suggestion from https://github.com/youtube/cobalt/pull/849#discussion_r1261889885 to avoid performing updates during playback. That shouldn't block this current/PR feature, just raising it as something we should investigate soon. I'll see if we have a feature request tracking that and file one if not.

yuying-y:
> Can you make sure this is tested on RDK before it's submitted?

Okay, then it'll take longer to verify. The release infra is not ready to produce official update packages, so I have to 
1. create a custom build that turns off verification and connects to a local server; load the custom build on the RDK
2. create a custom crx package for RDK
3. run a local update server that serves the crx package
4. make sure the RDK can connect to the local server to download the package

Unexpected network restrictions and connection issues can slow it down.

andrewsavage1:
> > Can you make sure this is tested on RDK before it's submitted?
> 
> Okay, then it'll take longer to verify. The release infra is not ready to produce official update packages, so I have to
> 
> 1. create a custom build that turns off verification and connects to a local server; load the custom build on the RDK
> 2. create a custom crx package for RDK
> 3. run a local update server that serves the crx package
> 4. make sure the RDK can connect to the local server to download the package
> 
> Unexpected network restrictions and connection issues can slow it down.

In that case testing that it doesn't crash cobalt on rdk is good enough

services/network/public/cpp/simple_url_loader.h
```diff
@@ -80,7 +80,12 @@ class COMPONENT_EXPORT(NETWORK_CPP) SimpleURLLoader {
   };
 
   // The maximum size DownloadToString will accept.
-  static constexpr size_t kMaxBoundedStringDownloadSize = 5 * 1024 * 1024;
+#if BUILDFLAG(IS_STARBOARD)
+  // The existing packages are as big as 70MB, so setting the limit to 100MB for now.
+  static constexpr size_t kMaxBoundedStringDownloadSize = 100 * 1024 * 1024;
```
hlwarriner, line 85: A couple things.
1. It would be nice to know why the upstream authors chose 20MB as the limit (unfortunately it's not documented). As a follow-up we should check with them to see if they anticipate any issues from downloading larger content.
2. I don't have any great suggestions here but we would ideally have some kind of process for managing hardcoded limits like this. This one worries me since our automated e2e tests wouldn't discover there's an issue until we're preparing to release version x that has packages exceeding this value, at which point devices are in the field running version x-1 still have this current value and therefore would not accept updates to x. We can also discuss this as a follow-up.

cobalt/updater/network_fetcher.cc
```diff
@@ -155,8 +155,46 @@ void NetworkFetcher::DownloadToString(
...
+  auto resource_request = std::make_unique<network::ResourceRequest>();
+  resource_request->url = url;
+  resource_request->method = "GET";
+  resource_request->load_flags = net::LOAD_DISABLE_CACHE;
+  simple_url_loader_ = network::SimpleURLLoader::Create(
+      std::move(resource_request), traffic_annotation);
+  simple_url_loader_->SetRetryOptions(
+      kMaxRetriesOnNetworkChange,
+      network::SimpleURLLoader::RetryMode::RETRY_ON_NETWORK_CHANGE);
+  simple_url_loader_->SetAllowPartialResults(true);
```
hlwarriner: Is there a reason to override the default and allow partial results? I doubt we'd be able to make use of partial Crx content so it would probably be better to keep the default behavior. But let me know if this helps with error handling or something else.

services/network/public/cpp/simple_url_loader.h
hlwarriner: Re (2), another option would be to use the`DownloadToStringOfUnboundedSizeUntilCrashAndDie` API. I'd say for  now let's stick with `DownloadToString` though.

services/network/public/cpp/simple_url_loader.h
hlwarriner, line 1: This seems like the recommended interface for loading URLs so I'm definitely in favor of using it. But after this PR is merged we should do some profiling on reference devices to see make sure there aren't significant performance regressions vs. c25, aside from those resulting from the change in package size. For example, AFAICT this interface doesn't support preallocation of the response string based on the content size, and we had been using preallocation to reduce the memory spike during download (https://github.com/youtube/cobalt/pull/849#discussion_r1259747020).

hlwarriner: > we should do some profiling on reference devices

I'm happy to help with that or share the effort in some way, btw.

services/network/public/cpp/simple_url_loader.h
yuying-y: Thanks for the comment. Filed b/465213840 as follow-up.

cobalt/updater/network_fetcher.cc
yuying-y: Good question. I was following a similar practice in https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/26.eap:components/update_client/net/network_impl.cc. But thinking about it twice, I agree that we won't make use of a partially downloaded crx package. Removed it.

services/network/public/cpp/simple_url_loader.h
yuying-y: Makes sense. Sounds good. Also tracked in b/465213840.

y4vor (approved):

yuying-y:
> > > Can you make sure this is tested on RDK before it's submitted?
> > 
> > 
> > Okay, then it'll take longer to verify. The release infra is not ready to produce official update packages, so I have to
> > 
> > 1. create a custom build that turns off verification and connects to a local server; load the custom build on the RDK
> > 2. create a custom crx package for RDK
> > 3. run a local update server that serves the crx package
> > 4. make sure the RDK can connect to the local server to download the package
> > 
> > Unexpected network restrictions and connection issues can slow it down.
> 
> In that case testing that it doesn't crash cobalt on rdk is good enough

Verified 👍

PR #8162, opened 2025-11-21
[tvOS] Remove system_raise_platform_error.mm

rakuco:
Nothing is actually calling SbSystemRaisePlatformError() in the main branch, and this implementation imported from C25 is specific to the non-Chrobalt tvOS code with the way it blocks key presses and expects Starboard to manage the UI.

Furthermore, most of the code handling `_processUIPresses` was already removed in #8141.

Bug: 459507205


rakuco:

PR #8164, opened 2025-11-21
[tvOS] Remove SBDSearch* code

rakuco:
This code is also no longer necessary in Chrobalt now that the UI work is done in Cobalt Shell.

Bug: 459507205


rakuco:

PR #8165, opened 2025-11-21
[tvOS] Remove several unused files

rakuco:
These files implemented code that is not used in the Chrobalt tvOS port (some of these Starboard APIs are entirely gone).

Bug: 459507205


rakuco:

fayev-g (approved):

PR #8177, opened 2025-11-21
BACKPORT: [tvos] Handle events from a remote control

rakuco:
Backport https://crrev.com/c/6661878 from M140, a patch by @jkim-julie. This is part of the set of backports required to get input handling to work correctly.

Bug: 456692793

---

Original CL description:

This adds a pan gesture recognizer to handle continuous left and right events, and it’s activated only if the swipe gesture fails to recognize.

It also generates directional events on pressing Up/Right/Left/Right buttons.

Change-Id: Ib7e76aecc4e0df4db34dbdd9b32ef3cb45ba1809
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6661878
Cr-Commit-Position: refs/heads/main@{#1478341}


rakuco:

PR #8182, opened 2025-11-22
Updater implementation

madhurajayaraman:
Bug: 454440974

### DIFF
--- a/cobalt/browser/h5vcc_updater/BUILD.gn
+++ b/cobalt/browser/h5vcc_updater/BUILD.gn
@@ -31,7 +31,7 @@ source_set("h5vcc_updater") {
     deps += [ "//cobalt/configuration" ]
   }
 
-  if (sb_is_evergreen) {
+  if (use_evergreen) {
     deps += [ "//cobalt/updater:updater" ]
   }
 }
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_updater/BUILD.gn
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_updater/BUILD.gn
@@ -19,7 +19,7 @@ blink_modules_sources("h5vcc_updater") {
     "h_5_vcc_updater.cc",
     "h_5_vcc_updater.h",
   ]
-  if (sb_is_evergreen) {
+  if (use_evergreen) {
     deps = [ "//cobalt/browser/h5vcc_updater/public/mojom:mojom_blink" ]
   }
 }


cobalt/updater/updater_module.h
```diff
@@ -137,6 +151,9 @@ class UpdaterModule {
   void Finalize();
   void MarkSuccessfulImpl();
   void Update();
+
+  // Holds the single instance of UpdaterModule.
+  static std::unique_ptr<UpdaterModule> instance_;
```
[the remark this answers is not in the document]
yuying-y: Ditto. Can the initialization of `updater_module_` be removed from updater_module.cc? Also, is there a better solution than creating this leaked singleton?

cobalt/updater/updater_module.cc
```diff
@@ -199,7 +217,7 @@ UpdaterModule::UpdaterModule(
     base::TimeDelta update_check_delay)
     : url_loader_factory_(std::move(url_loader_factory)),
       update_check_delay_(update_check_delay) {
-  LOG(INFO) << "UpdaterModule::UpdaterModule";
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
```
yuying-y: Why is this added (and the one below)? Please keep the original logs.

cobalt/updater/updater_module.cc
```diff
@@ -153,6 +153,24 @@ namespace updater {
...
 const base::TimeDelta kDefaultUpdateCheckDelay = base::Seconds(30);
 
+// static
+base::NoDestructor<UpdaterModule>* UpdaterModule::updater_module_ = nullptr;
+
+void UpdaterModule::CreateInstance(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    base::TimeDelta update_check_delay) {
+  DCHECK(!updater_module_)
+      << "UpdaterModule::CreateInstance() called more than once.";
```
yuying-y: "UpdaterModule is already created. Use UpdaterModule::GetInstance() to get the instance."

cobalt/updater/updater_module.cc
```diff
@@ -153,6 +153,24 @@ namespace updater {
...
+  updater_module_ = new base::NoDestructor<UpdaterModule>(
+      std::move(url_loader_factory), update_check_delay);
+}
+
+UpdaterModule* UpdaterModule::GetInstance() {
+  DCHECK(updater_module_)
+      << "UpdaterModule::GetInstance() called before CreateInstance().";
```
yuying-y: "UpdaterModule is not created yet, and cannot be retrieved by UpdaterModule::GetInstance()."

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -45,7 +46,9 @@ void H5vccUpdaterImpl::Create(
 
 void H5vccUpdaterImpl::GetUpdateServerUrl(GetUpdateServerUrlCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  std::string url =
+      cobalt::updater::UpdaterModule::GetInstance()->GetUpdateServerUrl();
```
yuying-y, line 54: This particular API is fine. For the other setters for the sideloading feature, please use the macro ALLOW_EVERGREEN_SIDELOADING to guard the calls. https://b.corp.google.com/issues/454440974#comment3

cobalt/updater/updater_module.h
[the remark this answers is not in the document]
yuying-y: We talked about this on the standup meeting. Singleton is an acceptable solution ATM.

cobalt/updater/updater_module.cc
madhurajayaraman: We call this dcheck in the impl file - accidentally put it in here too. Reverted.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: I think all of this maybe needs to be gated behind is_evergreen flags - I can still use the macro for the followup APIs but i'll also look into gating for everythign.

cobalt/updater/updater_module.cc
```diff
@@ -215,7 +234,7 @@ UpdaterModule::UpdaterModule(
 }
 
 UpdaterModule::~UpdaterModule() {
-  LOG(INFO) << "UpdaterModule::~UpdaterModule";
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
```
yuying-y: Remove the dcheck and keep the original logs.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
yuying-y: ALLOW_EVERGREEN_SIDELOADING is different from is_evergreen. It's for the sideloading feature.

cobalt/updater/updater_module.cc
madhurajayaraman: Ah ty! i thought I'd gotten them all in the last pass.

cobalt/updater/updater_module.h
```diff
@@ -120,6 +127,13 @@ class UpdaterModule {
   void MarkSuccessful();
 
  private:
+  // Private constructor and destructor to enforce singleton pattern.
+  explicit UpdaterModule(scoped_refptr<network::SharedURLLoaderFactory>,
+                         base::TimeDelta update_check_delay);
+  ~UpdaterModule();
+
+  friend class base::NoDestructor<UpdaterModule>;
```
andrewsavage1, line 136: Put a TODO here (or somewhere) to see if we can avoid making the updater a singleton

andrewsavage1 (approved):

andrewsavage1:
Removed my approval bc I want to take a quick look after all the build failures are fixed

cobalt/updater/updater_module.h
```diff
@@ -137,6 +152,9 @@ class UpdaterModule {
   void Finalize();
   void MarkSuccessfulImpl();
   void Update();
+
+  // Holds the single instance of UpdaterModule.
+  static base::NoDestructor<UpdaterModule>* updater_module_;
```
yuying-y, line 157: Just thought of this: does it mean ~UpdaterModule() is never called when the app exits? If so, we need to be careful, because the destruction sequence in ~UpdaterModule() is carefully laid out to avoid a crash when the app exits while the updater is still running. Please run this scenario locally to confirm no crash happens.

yuying-y: To repro this scenario, run loader_app, monitor the cobalt logs, when the updater sends update check request, immediately exit the app by clicking the "X" of the window, then check the cobalt logs to see if a crash happens. I can explain offline if it's not clear.

cobalt/shell/BUILD.gn
```diff
@@ -55,6 +55,7 @@ group("cobalt_shell_lib_deps") {
     "//base/third_party/dynamic_annotations",
     "//cc/base",
     "//cobalt/shell/browser/migrate_storage_record",
+    "//cobalt/updater",
```
yuying-y: The following is already present in the definition of `cobalt_shell_lib`. Is the new addition necessary?
```
  if (use_evergreen) {
    deps = [ "//cobalt/updater" ]
  }
```

cobalt/updater/updater_module.h
yuying-y: Madu and I verified. It doesn't cause a crash on exit.

cobalt/shell/browser/shell.h
```diff
@@ -35,7 +35,7 @@
 #include "ui/gfx/geometry/size.h"
 #include "ui/gfx/native_widget_types.h"
 #if BUILDFLAG(USE_EVERGREEN)
-#include "cobalt/updater/updater_module.h"
+#include "cobalt/updater/updater_module.h"  //nogncheck
```
andrewsavage1: nit: Do we need this include anymore or should it be moved to the cc file?

andrewsavage1 (commented):
LGTM but builds still seem to be failing (due to some new dependencies)

cobalt/browser/h5vcc_updater/BUILD.gn
```diff
@@ -30,4 +30,8 @@ source_set("h5vcc_updater") {
   if (is_starboard) {
     deps += [ "//cobalt/configuration" ]
   }
+
+  if (sb_is_evergreen) {
```
yuying-y: Nit: This should be `use_evergreen` to be consistent with the build flag in cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/BUILD.gn
```diff
@@ -19,6 +19,7 @@ blink_modules_sources("h5vcc_updater") {
     "h_5_vcc_updater.cc",
     "h_5_vcc_updater.h",
   ]
-
-  deps = [ "//cobalt/browser/h5vcc_updater/public/mojom:mojom_blink" ]
+  if (sb_is_evergreen) {
```
yuying-y: Same here: should be `use_evergreen`.

cobalt/shell/browser/shell.cc
```diff
@@ -207,14 +207,14 @@ void Shell::FinishShellInitialization(Shell* shell) {
...
 #if BUILDFLAG(USE_EVERGREEN)
-  // Create the updater module
+  // Create the updater module singleton.
   auto* storage_partition =
       raw_web_contents->GetPrimaryMainFrame()->GetStoragePartition();
   if (storage_partition) {
-    LOG(INFO) << "Creating updater_module_ for Shell";
-    shell->updater_module_.reset(new cobalt::updater::UpdaterModule(
+    LOG(INFO) << "Creating UpdaterModule singleton for Shell.";
+    cobalt::updater::UpdaterModule::CreateInstance(
```
borongc, line 215: On ATV, shell could be recreated multiple times, so this would be an issue for AOSP.

That is, calling `Shell::CreateShell()` second time will break the assumption, and it will fail on https://source.corp.google.com/h/github/youtube/cobalt/+/26.eap:cobalt/updater/updater_module.cc;l=162.

borongc: For example, I have a PR adding unit tests to splash screen (https://github.com/youtube/cobalt/pull/8522), where it will create shell for each unit test.

This triggers the DCHECK due to `updater_module_` is a singleton for Cobalt, but it doesn't prevent from re-creating Shell() object itself.

Hence, a unit test `SplashScreenTest.EarlyMainContentLoad` failed every time in https://github.com/youtube/cobalt/actions/runs/20717531195/job/59480499360.

andrewsavage1: Yeah, if we're going to stick with a singleton we should ensure that this can handle multiple calls gracefully—but ideally we won't use a singleton.

yuying-y: #8520 moves the updater initialization to `CobaltContentBrowserClient`. Is it created multiple times like the shell?

borongc: `CobaltContentBrowserClient::OnWebContentsCreated` is called when each WebContents is created, so it is still called multiple times.

However, as Cobalt only holds a single CobaltContentBrowserClient, and it won't be destroyed as Shell(), so you can just add a flag to check if Updater singleton is initialized or not.

yuying-y: Resolved in https://github.com/youtube/cobalt/pull/8520/changes/274113036b1534be71ff66ce8ecf4b9f5699cd98

PR #8183, opened 2025-11-22
Fix: GrowableIOBuffer memory leak on SetCapacity(0)

sherryzy:
Addressed a memory leak in GrowableIOBuffer::SetCapacity when resizing to 0. Previously, unique_ptr::release() followed by realloc could lead to memory not being deallocated if realloc returned NULL. The fix explicitly calls reset() on the unique_ptr for capacity 0, ensuring proper memory release.

Bug: 454095852

### DIFF
--- a/net/base/io_buffer.cc
+++ b/net/base/io_buffer.cc
@@ -110,6 +110,18 @@ GrowableIOBuffer::GrowableIOBuffer() = default;
 
 void GrowableIOBuffer::SetCapacity(int capacity) {
   DCHECK_GE(capacity, 0);
+// Calling reallocate with size 0 and a non-null pointer causes memory leaks
+// on many platforms, since it may return nullptr while also not deallocating
+// the previously allocated memory.
+#if BUILDFLAG(IS_STARBOARD)
+  if (capacity == 0) {
+    real_data_.reset();
+    capacity_ = 0;
+    offset_ = 0;
+    data_ = nullptr;
+    return;
+  }
+#endif
   // this will get reset in `set_offset`.
   data_ = nullptr;
   // realloc will crash if it fails.


net/base/io_buffer.cc
```diff
@@ -110,6 +110,13 @@ GrowableIOBuffer::GrowableIOBuffer() = default;
 
 void GrowableIOBuffer::SetCapacity(int capacity) {
   DCHECK_GE(capacity, 0);
```
kaidokert, line 112: Trunk Chrome turned this into a CHECK, maybe we should, too ? 

https://source.chromium.org/chromium/chromium/src/+/main:net/base/io_buffer.cc;l=120

kaidokert: E.g just cherry-pick https://source.chromium.org/chromium/chromium/src/+/71ec175d18c7b7feb1294f6be1492cd326f71f86

sherryzy: @kaidokert  Yeah, I agree we can certainly proceed with the cherry-pick. However, I believe that cherry-pick is not directly related to the memory leak fix in this PR and should be handled separately. Since the bug manifests specifically when capacity = 0, neither a DCHECK nor a CHECK on negative capacity would prevent the leak.

One thing I'm uncertain is that shall we upstream this change to chrome. I assume yes, but I need to do more testing on chrome. And it's unwise to hold up this PR. We can proceed to work on the upstream change in parallel while we finalize the review and land this PR, does that sound good?

kaidokert: Nevermind about upstreaming for now - trunk Chrome code is quite different, having gone through span~ification effort. I don't think they actually have a leak there. 

I also didn't intend to hold up your PR at all - you had it in draft, that's why i left a quick comment.


However, please have a look at this: go/cobalt-cl/242168 - this is the second time we are fixing this issue. Not upstreaming has very real cost. We also have bugs like b/266242187 still open.

kaidokert:
@jellefoks does this look familiar :)

net/base/io_buffer.cc
kaidokert: I think upstream/Chrome doesn't have a leak here because of `realloc(0)` is undefined, whereas with PartitionAlloc it's safe. b/269801115

sherryzy: Yeah, this is exactly the same issue that Jelle fixed in b/269801115.

net/base/io_buffer.cc
```diff
@@ -110,6 +110,13 @@ GrowableIOBuffer::GrowableIOBuffer() = default;
...
+  if (capacity == 0) {
```
andrewsavage1, line 117: Let's add a BUILDFLAG to make it clear this is a cobalt change. It'd also probably be good to add a comment explaining why this change is made from upstream code.

PR #8185, opened 2025-11-24
media: Implement flow control for MediaDecoder

kjyoun:
Integrates `DecoderStateTracker` into the Android `MediaDecoder` to throttle input processing when the decoder has too many pending frames. This helps manage memory usage and prevents the decoder from falling
too far behind.

Key changes:
*   Instantiates `DecoderStateTracker` in `MediaDecoder` if enabled.
*   Blocks input processing in `DecoderThreadFunc` when `CanAcceptMore()` returns false.
*   Updates frame tracking via `SetFrameAdded` and `SetFrameDecoded`.

Bug: 455938352

### DIFF
--- a/starboard/shared/starboard/media/decoder_state_tracker.cc
+++ b/starboard/shared/starboard/media/decoder_state_tracker.cc
@@ -15,14 +15,11 @@
 #include "starboard/shared/starboard/media/decoder_state_tracker.h"
 
 #include <algorithm>
-#include <optional>
 #include <ostream>
-#include <sstream>
 #include <string_view>
 #include <utility>
 
 #include "build/build_config.h"
-#include "starboard/common/check_op.h"
 #include "starboard/common/log.h"
 #include "starboard/common/string.h"
 #include "starboard/common/time.h"
--- a/starboard/shared/starboard/media/decoder_state_tracker.h
+++ b/starboard/shared/starboard/media/decoder_state_tracker.h
@@ -16,15 +16,15 @@
 #define STARBOARD_SHARED_STARBOARD_MEDIA_DECODER_STATE_TRACKER_H_
 
 #include <atomic>
-#include <functional>
-#include <map>
+#include <iosfwd>
 #include <memory>
 #include <mutex>
 #include <string>
 #include <string_view>
+#include <utility>
 #include <vector>
 
-#include "starboard/common/log.h"
+#include "build/build_config.h"
 #include "starboard/shared/internal_only.h"
 #include "starboard/shared/starboard/player/job_thread.h"
 
--- a/starboard/shared/starboard/media/decoder_state_tracker_test.cc
+++ b/starboard/shared/starboard/media/decoder_state_tracker_test.cc
@@ -14,8 +14,6 @@
 
 #include "starboard/shared/starboard/media/decoder_state_tracker.h"
 
-#include <atomic>
-#include <optional>
 #include <sstream>
 
 #include "starboard/common/time.h"


starboard/shared/starboard/media/decoder_state_tracker_test.cc
```diff
@@ -0,0 +1,287 @@
...
+  EXPECT_TRUE(decoder_state_tracker_->CanAcceptMore());
+
+  // Fill up to new max (4).
+  decoder_state_tracker_->SetFrameAdded(2);
+  decoder_state_tracker_->SetFrameDecoded(2);
+  decoder_state_tracker_->SetFrameAdded(3);
+  decoder_state_tracker_->SetFrameAdded(4);
+  EXPECT_FALSE(decoder_state_tracker_->CanAcceptMore());
+
+  // Release to dip again. Should increase max to 6.
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/android/shared/media_decoder.cc
```diff
@@ -665,6 +680,10 @@ void MediaDecoder::OnMediaCodecOutputBufferAvailable(
     return;
   }
 
+  if (size > 0 && decoder_state_tracker_) {
+    decoder_state_tracker_->SetFrameDecoded(presentation_time_us);
```
jasonzhangxx: |decoder_state_tracker_| need to be disabled for tunnel mode. Under tunnel mode, there's no output frame from the decoder.

starboard/android/shared/media_decoder.cc
```diff
@@ -509,6 +516,14 @@ bool MediaDecoder::ProcessOneInputBuffer(
     memcpy(address, data, size);
   }
 
+  if (!input_buffer_already_written && decoder_state_tracker_) {
+    if (pending_input.type != PendingInput::kWriteEndOfStream) {
+      decoder_state_tracker_->SetFrameAdded(input_buffer->timestamp());
+    } else {
+      decoder_state_tracker_->SetEosFrameAdded();
+    }
+  }
```
jasonzhangxx: These code should be moved to after the input buffer is written into the underlying media codec successfully (the block below).

starboard/android/shared/player_components_factory.cc
```diff
@@ -68,6 +70,39 @@ constexpr bool kForceResetSurfaceUnderTunnelMode = true;
 constexpr int64_t kResetDelayUsecOverride = 0;
 constexpr int64_t kFlushDelayUsecOverride = 0;
 
+std::optional<int> ReadSystemPropertyPositiveInt(const char* key) {
+  char value[PROP_VALUE_MAX];
+  if (__system_property_get(key, value) == 0) {
```
jasonzhangxx: How will we change the value? Or it's just for local testing?

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,245 @@
...
+  std::lock_guard lock(mutex_);
+
+  if (disabled_) {
+    return;
+  }
+  // frames_in_flight_ can exceed max_frames_ initially to accommodate
+  // decoders that require a larger batch of input frames before producing the
+  // first decoded output.
+  if (frames_in_flight_.size() >
+      std::max(max_frames_, kMaxAllowedFramesWhenNoDecodedFrameYet)) {
```
jasonzhangxx: If the added frames exceed the max allowance, we should still add the record to |frames_in_flight_| to prevent writing more frames. If we return here directly, the actual written frames and the records in |frames_in_flight_| will be mismatched.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,245 @@
...
+          std::lock_guard lock(mutex_);
+          if (disabled_) {
+            return;
+          }
+          auto it = frames_in_flight_.upper_bound(presentation_time_us);
+          frames_in_flight_.erase(frames_in_flight_.begin(), it);
+
+          if (reached_max_ && frames_in_flight_.size() <= kFramesLowWatermark) {
+            int old_max = max_frames_;
+            max_frames_++;
```
jasonzhangxx: Can you help to understand why we want to increase |max_frames_| here?

IIUC, `frames_in_flight_.size()` includes decoding and decoded frames, so if's for performance concern, my understanding is it should check on decoded frames.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,245 @@
...
+std::string to_ms_string(std::optional<int64_t> us_opt) {
+  if (!us_opt) {
+    return "(nullopt)";
+  }
+  return std::to_string(*us_opt / 1'000);
+}
+
+}  // namespace
+
+DecoderStateTracker::DecoderStateTracker(int initial_max_frames,
```
jasonzhangxx: What "initial_max_frames" are we going to use? If we want to hardcode a limit for 4k or 8k videos, I think we can consider to limit the decoded frames only for 4k, 8k playbacks on devices with limit memory.

The android MediaCodec has it's own resource management, I think it has it's own memory usage limit. We used to have issue b/226677552. Some devices have very few available output buffers.

starboard/android/shared/player_components_factory.cc
kjyoun: This is for local testing. added comments

```
// For local testing with non-Gold build, set properties to enable decoder throttling.
// $ adb shell setprop setprop debug.cobalt.max_pending_inputs_size 10
// $ adb shell setprop debug.cobalt.initial_max_frames_in_decoder 6
```

starboard/android/shared/media_decoder.cc
kjyoun: Done

starboard/android/shared/media_decoder.cc
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Added comments explaining why w need this.
```
+            // If the number of frames in flight drops to the low-water mark
+            // after we have reached the max frames, it means the current max
+            // frames is too small to keep the decoder busy. We bump up the
+            // max frames to avoid underrun.
```

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: This logic is to detect a malfunction in `DecoderStateTracker`.                                                                                  
                                                                                                                                                  
The tracker should never track more frames than `std::max(max_frames_, kMaxAllowedFramesWhenNoDecodedFrameYet)`. If this condition is met, it   means a basic premise of the tracker's logic has been violated.      

  When such an incident occurs, `DecoderStateTracker` activates its kill switch, effectively disabling itself. From that point on, it stops  throttling the decoder entirely to avoid blocking playback.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: - MediaCodec has its own internal limits, but relying solely on them can cause excessive memory usage since decoded video frames are very large. `DecoderStateTracker` is designed to throttle the decoder pipeline to keep this memory footprint in check.
- However, we must be careful not to throttle too strictly, as that causes playback underruns and stuttering.
- To solve this, I introduced an adaptive (not hard-coded) limit. `DecoderStateTracker` starts with `initial_max_frames`, but automatically bumps up this limit if it detects the decoder can be starving (i.e., when the number of in-flight frames drops to the low-water mark).

starboard/android/shared/player_components_factory.cc
kjyoun: I removed this code for local experiment

jasonzhangxx:
Another thought is, there's an existing VideoFrameTracker, and we can consider to combine it with DecoderStateTracker.

kjyoun:
> Another thought is, there's an existing VideoFrameTracker, and we can consider to combine it with DecoderStateTracker.

I considered this, but I believe it's safer and cleaner to keep them separate due to their distinct responsibilities:
- DecoderStateTracker: For Flow Control
- VideoFrameTracker: For Statistics (i.e. dropped frames)

For now, it's better to keep them separate to make each class that has different roles to be compact for its own goal. Merging them into one would create a very complicated hard-to-debug class.

We can consider this option of merging them into one, when `DecoderStateTracker` passed the experiment and its implementation gets finalized.

starboard/android/shared/media_decoder.cc
```diff
@@ -206,7 +217,6 @@ void MediaDecoder::WriteInputBuffers(const InputBuffers& input_buffers) {
                    &MediaDecoder::DecoderThreadEntryPoint, this);
     SB_DCHECK_NE(decoder_thread_, 0);
   }
-
```
xiaomings, line 209: Let's keep the empty lines, and the same to other empty lines removed.  Empty lines (at the right amount) makes code more readable.

starboard/android/shared/player_components_factory.cc
```diff
@@ -50,6 +50,8 @@
 #include "starboard/shared/starboard/player/filter/video_renderer_internal_impl.h"
 #include "starboard/shared/starboard/player/filter/video_renderer_sink.h"
 
+#include <sys/system_properties.h>
```
xiaomings: Consider removing if no longer needed.

starboard/android/shared/video_decoder.cc
```diff
@@ -189,7 +189,8 @@ void ParseMaxResolution(const std::string& max_video_capabilities,
 
 class VideoFrameImpl : public VideoFrame {
  public:
-  typedef std::function<void()> VideoFrameReleaseCallback;
+  typedef std::function<void(int64_t pts, int64_t release_us)>
```
xiaomings: pts_us, release_at_us?

starboard/android/shared/video_decoder.h
```diff
@@ -57,7 +57,10 @@ class VideoDecoder
       VideoRendererSink;
 
   class Sink;
-
+  struct FlowControlOptions {
+    std::optional<int> max_pending_input_frames;
+    std::optional<int> initial_max_frames_in_decoder;
+  };
```
xiaomings, line 64: Consider adding empty line before and after.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+
+// Low-water mark for in-flight frames.
+// If the total number of in-flight frames goes lower than this value,
+// max frames will increase.
+constexpr int kFramesLowWatermark = 2;
+
+// Maximum number of endoding frames to accept when no frame is generated.
+// Some devices need a large number of frames when generating the 1st
+// decoded frame. See b/405467220#comment36 for details.
+constexpr int kMaxAllowedFramesWhenNoDecodedFrameYet = 10;
```
xiaomings, line 38: This is taken from playbacks not at the highest resolution on one device, it would be great if we can make a decision based on more devices.  Feel free to sync offline.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+
+std::string to_ms_string(std::optional<int64_t> us_opt) {
+  if (!us_opt) {
+    return "(nullopt)";
+  }
+  return std::to_string(*us_opt / 1'000);
```
xiaomings: nit: `return us_opt ? std::to_string(*us_opt / 1'000) : "(nullopt)";`, or
```
  if (us_opt) {
    return std::to_string(*us_opt / 1'000);
  }
  return "(nullopt)";
```

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+
+namespace starboard {
+namespace {
+
+// Low-water mark for in-flight frames.
+// If the total number of in-flight frames goes lower than this value,
+// max frames will increase.
+constexpr int kFramesLowWatermark = 2;
+
+// Maximum number of endoding frames to accept when no frame is generated.
```
xiaomings: endoding => encoded

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+void DecoderStateTracker::SetFrameAdded(int64_t presentation_time_us) {
+  std::lock_guard lock(mutex_);
+
+  if (disabled_) {
+    return;
+  }
+  // frames_in_flight_ can exceed max_frames_ initially to accommodate
+  // decoders that require a larger batch of input frames before producing the
+  // first decoded output.
+  if (frames_in_flight_.size() >
```
xiaomings: Is it possible for you to elaborate on the specific cases under which the condition of the if statement evaluates to true?

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+  std::lock_guard lock(mutex_);
+
+  if (disabled_) {
+    return;
+  }
+
+  auto it = frames_in_flight_.upper_bound(presentation_time_us);
+  for (auto i = frames_in_flight_.begin(); i != it; ++i) {
+    i->second = FrameStatus::kDecoded;
+  }
```
xiaomings, line 153: This is O(n).  Do we need the status of individual frame other than for testing or logging?

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+                                             int64_t release_us) {
+  {
+    std::lock_guard lock(mutex_);
+    if (disabled_) {
+      return;
+    }
+  }
+
+  int64_t delay_us = std::max(release_us - CurrentMonotonicTime(), int64_t{0});
+  job_thread_->Schedule(
```
xiaomings: If we save this information somehow, we can move the processing into CanAcceptMore(), and avoid introducing an extra thread.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+        }
+        frame_released_cb_();
+      },
+      delay_us);
+}
+
+void DecoderStateTracker::Reset() {
+  std::lock_guard lock(mutex_);
+  frames_in_flight_.clear();
+  disabled_ = false;
```
xiaomings: If it's disabled during the playback, we should probably keep it as disabled for the same reason as we keep `max_frames_`.  We may need to handle eos differently so it's not tracked by `disabled`.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,89 @@
...
+  DecoderStateTracker(int initial_max_frames, FrameReleaseCB frame_released_cb);
+  DecoderStateTracker(int initial_max_frames,
+                      FrameReleaseCB frame_released_cb,
+                      std::optional<int> log_interval_us);
+  ~DecoderStateTracker();
+
+  void SetFrameAdded(int64_t presentation_time_us);
+  void SetEosFrameAdded();
+  void SetFrameDecoded(int64_t presentation_time_us);
+  void SetFrameReleasedAt(int64_t presentation_time_us, int64_t release_us);
```
xiaomings: AddFrame, AddEosFrame, OnFrameDecoded, OnFrameReleasedAt?

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,89 @@
...
+  void Reset();
+
+  State GetCurrentStateForTest() const;
+  bool CanAcceptMore();
```
xiaomings: This can be const.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+      [this, presentation_time_us] {
+        {
+          std::lock_guard lock(mutex_);
+          if (disabled_) {
+            return;
+          }
+          auto it = frames_in_flight_.upper_bound(presentation_time_us);
+          frames_in_flight_.erase(frames_in_flight_.begin(), it);
+
+          if (reached_max_ && frames_in_flight_.size() <= kFramesLowWatermark) {
```
xiaomings: We're trying to make this algorithm adaptive here.  However, once we run this in an experiment, and if this algorithm actually has negative impact on some platforms, this change will reduce the impact and make the metrics look better.  However, do we want to actually make it more obvious if this algorithm doesn't work on some devices?

Just bring this up, and please feel free to keep this as is.

starboard/android/shared/video_decoder.cc
```diff
@@ -412,7 +419,7 @@ VideoDecoder::VideoDecoder(const VideoStreamInfo& video_stream_info,
 
   if (is_video_frame_tracker_enabled_) {
     video_frame_tracker_ =
-        std::make_unique<VideoFrameTracker>(kMaxPendingInputsSize * 2);
+        std::make_unique<VideoFrameTracker>(max_pending_inputs_size_ * 2);
```
jasonzhangxx, line 424: nit: just fyi, note that the size here for VideoFrameTracker initialization is the limit of how many video timestamps entries the tracker can have. So, it won't have much performance impact. The limit will be hit only when there're too many dropped frames.

starboard/android/shared/video_decoder.cc
```diff
@@ -933,7 +942,9 @@ void VideoDecoder::ProcessOutputBuffer(
   decoder_status_cb_(
       is_end_of_stream ? kBufferFull : kNeedMoreInput,
       new VideoFrameImpl(dequeue_output_result, media_codec_bridge,
-                         std::bind(&VideoDecoder::OnVideoFrameRelease, this)));
+                         [this](int64_t pts, int64_t release_us) {
+                           OnVideoFrameRelease(pts, release_us);
+                         }));
```
jasonzhangxx, line 953: Why here it uses lamda instead std::bind?

starboard/android/shared/video_decoder.cc
```diff
@@ -1257,11 +1268,16 @@ void VideoDecoder::OnTunnelModeCheckForNeedMoreInput() {
...
 }
 
-void VideoDecoder::OnVideoFrameRelease() {
+void VideoDecoder::OnVideoFrameRelease(int64_t pts, int64_t release_us) {
   if (output_format_) {
     --buffered_output_frames_;
     SB_DCHECK_GE(buffered_output_frames_, 0);
   }
+
+  if (media_decoder_ && media_decoder_->decoder_state_tracker()) {
```
jasonzhangxx, line 1283: nit: note that `OnVideoFrameRelease` is called from media decoder thread, but the code here looks ok.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
jasonzhangxx: nit: the year should be 2026 now.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+
+DecoderStateTracker::DecoderStateTracker(int initial_max_frames,
+                                         FrameReleaseCB frame_released_cb,
+                                         std::optional<int> log_interval_us)
+    : max_frames_(initial_max_frames),
+      frame_released_cb_(std::move(frame_released_cb)),
+      job_thread_(std::make_unique<shared::starboard::player::JobThread>(
+          "DecStateTrack")) {
+  SB_CHECK(frame_released_cb_);
+  if (log_interval_us) {
```
jasonzhangxx: Instead of an optional `log_interval_us`, I'd recommend to have a pre-define marco to enable the logs for debugging, so that the reader would immediately understand the purpose of the codes and the debug code won't be included in the production binary.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+  {
+    std::lock_guard lock(mutex_);
+    if (disabled_) {
+      SB_LOG(INFO) << "DecoderStateTracker state: DISABLED";
+    } else {
+      SB_LOG(INFO) << "DecoderStateTracker state: " << GetCurrentState_Locked();
+    }
+  }
+
+  job_thread_->Schedule(
```
jasonzhangxx: My understanding is when releasing `std::unique_ptr job_thread_`, it will set `job_thread_` to null and then delete the underlying object. In that case, `LogStateAndReschedule()` could be still running while `job_thread_` has been already set to null, and the code will crash when `job_thread_` is used here.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,89 @@
...
+  bool IsFull_Locked() const;
+  void EngageKillSwitch_Locked(std::string_view reason, int64_t pts);
+  void LogStateAndReschedule(int64_t log_interval_us);
+
+  const FrameReleaseCB frame_released_cb_;
+  const std::unique_ptr<shared::starboard::player::JobThread> job_thread_;
+
+  mutable std::mutex mutex_;
+  std::map<int64_t, FrameStatus> frames_in_flight_;  // Guarded by |mutex_|.
+  bool disabled_ = false;                            // Guarded by |mutex_|.
```
jasonzhangxx: When the tracker is disabled, we can send `initial_max_frames` of null to MediaDecoder, so that it won't be created, and in MediaDecoder, other code actually relies on checks of if `decoder_state_tracker_` is non null to decide if it should use the tracker. So, I believe we can remove the `disabled_` here.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+  // decoders that require a larger batch of input frames before producing the
+  // first decoded output.
+  if (frames_in_flight_.size() >
+      std::max(max_frames_, kMaxAllowedFramesWhenNoDecodedFrameYet)) {
+    EngageKillSwitch_Locked("Too many frames in flight: state=" +
+                                ToString(GetCurrentState_Locked()),
+                            presentation_time_us);
+    return;
+  }
+  if (frames_in_flight_.find(presentation_time_us) != frames_in_flight_.end()) {
```
jasonzhangxx: nit: that should not happen and  if that happens, we can raise a playback error.

starboard/shared/starboard/media/decoder_state_tracker.cc
jasonzhangxx: `frames_in_flight_` is ordered, and decoded frames will be produced in same order, so we should be able to optimize it to O(1).

starboard/shared/starboard/media/decoder_state_tracker.cc
jasonzhangxx: I still don't understand the algorithm here. 
-  My understanding is`reached_max_` need to be recalculated once a frame is removed from `frames_in_flight_`. The existing code doesn't recalculate it.
- `reached_max_` is set to true at line 101 when  `frames_in_flight_.size() >= max_frames_`, that could happen when video is not playing (like paused by user or during prerolling). Then here it checks `frames_in_flight_.size() <= kFramesLowWatermark`. If `reached_max_` is recalculated properly and `max_frames_` is greater than `kFramesLowWatermark`, the if condition at line 149 should never be true?

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,249 @@
...
+            // max frames to avoid underrun.
+            int old_max = max_frames_;
+            max_frames_++;
+            reached_max_ = false;
+            SB_LOG(WARNING)
+                << "Bump up max frames from " << old_max << " to "
+                << max_frames_ << ": state=" << GetCurrentState_Locked();
+          }
+        }
+        frame_released_cb_();
```
jasonzhangxx: I don't recommend to have a `frame_released_cb_` in tracker class. Let's find a better approach to decouple it.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,89 @@
...
+
+// Tracks the number of frames currently inside the decoder (input queue) and
+// the output queue to enable flow control.
+//
+// By limiting the number of frames in flight, we can prevent the decoder from
+// buffering an excessive amount of data, which reduces system memory usage.
+//
+// This class is thread-safe and can be used from multiple threads or
+// SequencedTaskRunners.
+class DecoderStateTracker {
```
jasonzhangxx, line 33: > I considered this, but I believe it's safer and cleaner to keep them separate due to their distinct responsibilities:
> 
> * DecoderStateTracker: For Flow Control
> * VideoFrameTracker: For Statistics (i.e. dropped frames)
> 
> For now, it's better to keep them separate to make each class that has different roles to be compact for its own goal. Merging them into one would create a very complicated hard-to-debug class.
> 
> We can consider this option of merging them into one, when `DecoderStateTracker` passed the experiment and its implementation gets finalized.

I see. The existing `DecoderStateTracker` actually also tracks some frame status, and `VideoFrameTracker` does that too. Can we consider a more general solution? Like:

- DecoderStateTracker: For Flow Control only (we may need a new name), which can get frame status from VideoFrameTracker
- VideoFrameTracker: For frame status tracking only (i.e. how many frames is received, how many frames is sent to MediaCodec for decoding, how many frames is decoded, how many frames is rendered (received frames but now rendered are considered dropped), etc)

The tunnel pipeline will need such a VideoFrameTracker for workflow control too. Different pipelines might have difference in workflow control, but I think video frames tracker can be reused.

starboard/shared/starboard/media/decoder_state_tracker.cc
jasonzhangxx: My concern is, if there's no performance problem, the extra thread make released frame being removed from `frames_in_flight_` with a small latency, which makes `GetCurrentState_Locked()` not that accurate. And frame status updating on two threads is prone to unexpected results,  though as released is the last state of a frame, it should be fine for now.

I think the consideration here is to move the workflow control algorithm to a background thread. In that case, we can remove released frame synchronously but leave the control algorithm still on the background thread.

starboard/android/shared/media_decoder.cc
kjyoun: Done

starboard/android/shared/video_decoder.cc
kjyoun: Done

starboard/android/shared/player_components_factory.cc
kjyoun: Done

starboard/android/shared/video_decoder.h
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: add `eos_added_`

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Thanks for the suggestion. I've decided to use On for everything (OnFrameAdded, OnEosAdded, nFrameDecoded, OnRelease). 

Since this is a 'Tracker,' it's essentially an observer of the decoder's lifecycle, so the On prefix feels more accurate than Set or Add.

starboard/android/shared/video_decoder.cc
kjyoun: Gemini
> Lambdas are generally preferred over std::bind in modern C++ and Chromium style. They are more readable, allow the compiler to optimize/inline the call more effectively, and provide better type safety compared to std::bind's use of placeholders.

[go/totw/108](https://goto.google.com/totw/108)
> Avoid std::bind. Use a lambda or std::bind_front instead.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Upon second thought, I've standardized on the 'On' prefix for all these methods. 

Since this class acts as a tracker/observer of the decoder's lifecycle, 'On' more accurately describes the event-driven nature of these notifications compared to 'Add' or 'Set'.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Done

starboard/android/shared/video_decoder.cc
kjyoun: Ack

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Yes, we need to track the 'decoded' status to handle the initial 'cold start' burst. Some decoders require more input frames than the  steady-state `max_frames_` before producing their first output. By tracking this state, IsFull_Locked() can be more permissive until the decoder  actually starts producing, which prevents starvation.                                                                                           

Regarding the O(N) loop: because frames_in_flight_ is sorted and the decoder is typically a FIFO pipeline, OnFrameDecoded effectively only processes one frame at a time in practice, making it O(1) for most calls. Given that N is very small (< 20), the current implementation is simple and robust.

I also added comment explaining why I chose std::map here.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: This is fail-safe logic designed to handle unexpected desynchronization.                                                                     
                                                                                                                                              
 In theory, this condition should never be met. However, it serves as a guard for scenarios where our tracking logic cannot accurately reflect the decoder's state—for example, if a single encoded frame were to produce multiple decoded frames with different timestamps, or if release  signals are lost. In such cases, this logic 'kicks in' to disable the tracker and prevent it from indefinitely blocking the input pipeline.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: I've updated the comment to clarify that the '10' is based on the 6-frame requirement observed in b/405467220#comment36, with an added margin to account for 8K and the potential variations across other devices you mentioned. Since we don't have data for more devices yet, I've chosen this conservative limit to prevent starvation while still keeping a safety ceiling.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Because a strict limit could significantly interrupt playback, I've taken a conservative approach by making the limits adaptive and introducing a kill-switch.      

I agree with your point that this adaptivity might partially mask the 'cost' or side effects of the logic on certain platforms. However, my  goal for this initial experiment is to determine if the overall approach provides a net benefit without regressing playback stability. If the experiment is successful, we can follow up with more targeted trials to fine-tune the constants and identify exactly where the logic might be  under-performing

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: You are correct that `reached_max_` is not recalculated on every frame removal—this is by design. 

The logic uses a latch to "remember" that the decoder was throttled because it hit the current `max_frames_` limit. If the pipeline then drains all the way down to 2 frames before we can effectively refill it, it indicates that the current limit is too shallow to cover the latency of the refilling process on this specific device. Incrementing `max_frames_` is a dynamic adjustment to prevent such deep drains (and potential underruns) in the future.

Regarding the **paused/preroll** scenario: even if the limit is reached while paused, if the pipeline drains to 2 upon resume, it still suggests that the allowed buffer size isn't sufficient to keep the decoder busy while we burst-fill it back up to capacity.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Thanks for pointing it out. I've refactored the destructor to safely handle the race condition and prevent potential crashes during shutdown.

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: As we chatted offline, I agree with the general idea of refactoring to share common logic. My main consideration is the timing. 

Since the `DecoderStateTracker `is currently experimental, we don't yet know if it will be permanently adopted. I'd prefer not to modify or complicate the stable code (`VideoFrameTracker`) until the effectiveness of this new flow control logic is proven. 

Once the experiment is successful and the implementation is finalized, we can definitely move forward with a unified refactoring.

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: `frame_released_cb_` is essential for signaling the MediaDecoder thread to wake up as soon as the pipeline has room. Without it, the decoder would have to rely on polling or timeouts, leading to performance degradation.                                                          

A key goal of this PR was to encapsulate the experimental flow control logic entirely within the `DecoderStateTracker`, thereby  minimizing invasive changes to the stable `VideoDecoder` and `MediaDecoder` classes. This makes it easier to enable/disable the feature for the experiment.                                                                                                                                     

If you have a specific pattern in mind for decoupling this signaling while maintaining this level of encapsulation, I’d be happy to explore it.

starboard/android/shared/video_decoder.cc
kjyoun: Ack

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: In general, `DecoderFrameTracker` needs a thread to run its own timer, tracking the nearest moment that frame is released.  System-level events (e.g.  is too late. 

re: Xiamong's comment
`release_us` timestamp provided by the platform is in the future. Since there is no system-level event triggered at the exact moment the hardware actually releases the buffer, we must maintain our own timer to signal `frame_released_cb_`.  

If we moved this logic to `CanAcceptMore()`, the decoder thread would have to poll that function, which would lead to either high CPU usage (if polling frequently) or increased playback latency (if polling slowly). And it needs a significant refactoring of MediaDecoder, which this PR tries to avoid.

re:  Jason's comment
The frame is kept in `frames_in_flight_` until the `delay_us` expires because the hardware still holds that buffer until the scheduled release time. If we remove it synchronously in `OnReleased`, `CanAcceptMore()` would return true prematurely, allowing the decoder to push more frames than 
  the intended max_frames limit before the hardware has actually freed up space

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: I want to be fail-safe to make sure that this experiment not make the regression

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,259 @@
...
+  }
+  return state.total_frames() >= max_frames_;
+}
+
+void DecoderStateTracker::EngageKillSwitch_Locked(std::string_view reason,
+                                                  int64_t pts) {
+  SB_LOG(ERROR) << "KILL SWITCH ENGAGED: " << reason << ", pts=" << pts;
+  disabled_ = true;
+  frames_in_flight_.clear();
+  frame_released_cb_();
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/android/shared/player_components_factory.cc
```diff
@@ -555,7 +554,8 @@ class PlayerComponentsFactory : public starboard::shared::starboard::player::
         force_reset_surface, kForceResetSurfaceUnderTunnelMode,
         force_big_endian_hdr_metadata, max_video_input_size,
         creation_parameters.surface_view(), enable_flush_during_seek,
-        reset_delay_usec, flush_delay_usec, error_message);
+        reset_delay_usec, flush_delay_usec, VideoDecoder::FlowControlOptions{},
```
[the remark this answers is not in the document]
kjyoun: Updated PR description

starboard/android/shared/video_decoder.cc
jasonzhangxx: I don't see the benefit here to use lambda. Lambda significantly makes the callback longer. @xiaomings to decide.

starboard/android/shared/media_decoder.cc
```diff
@@ -137,7 +140,15 @@ MediaDecoder::MediaDecoder(
       first_tunnel_frame_ready_cb_(first_tunnel_frame_ready_cb),
       tunnel_mode_enabled_(tunnel_mode_audio_session_id != -1),
       flush_delay_usec_(flush_delay_usec),
+      decoder_state_tracker_(initial_max_frames && !tunnel_mode_enabled_
+                                 ? std::make_unique<DecoderStateTracker>(
+                                       *initial_max_frames,
+                                       [this] { condition_variable_.Signal(); })
```
jasonzhangxx: Same as in another comment, I don't agree that the lambda here having better readability.

starboard/shared/starboard/media/decoder_state_tracker.cc
jasonzhangxx: I see, the schedule here is to introduce a delay on purpose. I believe the lambda make the code hard to understand.

Do we have to remove the frame record from `frames_in_flight_` strictly at when it's rendered? We don't know how android manage the underlying memory (ram or vram?), so the restriction is based on assumption. And the frame would be released at most 50ms prior to be rendered (when playback rate 1x), which means we could have up to 3 frames untracked, but we still be able to limit the total total outputs.

If we run the algorithm immediately, it could simplify the code and we'll not need `frame_released_cb_` at line 171 to trigger the `condition_variable_`.

starboard/shared/starboard/media/decoder_state_tracker.cc
jasonzhangxx: I see, thanks for the detailed explanation.

starboard/android/shared/media_decoder.cc
xiaomings: It's ok to use lambda.  The main issue here is that we are trying to add too many logic to the ctor initializer.

starboard/android/shared/video_decoder.cc
kjyoun: Otherwise, we have to use std::bind and place holder, which is not recommend practice.

[go/totw/108](https://goto.google.com/totw/108) suggests to use either lambda or `std::bind_front`, which is not available on c++17.

starboard/android/shared/media_decoder.cc
kjyoun: Though ctor initializer, we can make `decoder_state_tracker_` to "const"

I rewrite initialization logic using IIFE[1] to improve readability.
I think that it's more readable to add static method (then we have to modify header file and read 2 different place to understand Tracker initialization logic)


---
[1] IIFE: Immediately Invoked Function Expression

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Yes, this thread is to track the estimated release time of decoded frames.

On kirkwood device, `OnReleased` is called \~50 msec before actual rendering time. If we decrease `frames_in_flight_` immediately on the callback, we can underestimate the buffer size by 1~3 frames, depending of fps.

[Test result](https://b.corp.google.com/issues/455938352#comment6) showed the improvement of this PR is about \~2 frames, so that 1~3 frame difference is significant

Considering size of decoded frame is huge (> 10MB for 4K),   1\~3 frame difference can results massive difference in memory usage.

EWhile the exact memory savings vary by device (as seen in the differences between Kirkwood and Mi Stick in the test results) because  underlying buffer management is outside our control, this PR aims to do our best to prevent overloading the decoder buffer across different implementations.

starboard/shared/starboard/media/decoder_state_tracker.cc
xiaomings: Consider trying a few more devices locally, especially the low powered devices in our working area.  Feel free to work with @osagie98 to get more metrics from devices in our labs.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,100 @@
...
+    int total_frames() const { return decoding_frames + decoded_frames; }
+  };
+
+  DecoderStateTracker(int initial_max_frames, FrameReleaseCB frame_released_cb);
+  ~DecoderStateTracker();
+
+  void OnFrameAdded(int64_t presentation_us);
+  void OnEosAdded();
+  void OnFrameDecoded(int64_t presentation_us);
+  void OnReleased(int64_t presentation_us, int64_t release_us);
```
xiaomings: nit: OnFrameReleased.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,100 @@
...
+    int decoding_frames = 0;
+    int decoded_frames = 0;
+
+    int total_frames() const { return decoding_frames + decoded_frames; }
+  };
+
+  DecoderStateTracker(int initial_max_frames, FrameReleaseCB frame_released_cb);
+  ~DecoderStateTracker();
+
+  void OnFrameAdded(int64_t presentation_us);
```
xiaomings: nit: OnFrameQueued, and OnEosQueued?

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,100 @@
...
+  State GetCurrentState_Locked() const;
+  bool IsFull_Locked() const;
+  void EngageKillSwitch_Locked(std::string_view reason, int64_t pts);
+
+#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
+  void StartPeriodicalLogging(int64_t log_interval_us);
+  void LogStateAndReschedule(int64_t log_interval_us);
+#endif
+
+  const FrameReleaseCB frame_released_cb_;
```
xiaomings: nit: FrameReleasedCB.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,100 @@
...
+  // Non-resettable members end.
+
+  // NOTE: |job_thread_| must be the last variable declared so that it is
+  // destroyed first (following the reverse order of declaration).
+  std::unique_ptr<shared::starboard::player::JobThread>
+      job_thread_;  // Guarded by |mutex_|.
+};
+
+std::ostream& operator<<(std::ostream& os,
+                         const DecoderStateTracker::State& status);
```
xiaomings: nit: `status` => `state`

starboard/shared/starboard/media/decoder_state_tracker.cc
xiaomings: The backlog can be longer on some low end platforms, and the O(n) would also take longer, and std::map<> isn't cache friendly.

On the other hand, I am ok with keeping the implementation as is.

starboard/shared/starboard/media/decoder_state_tracker.cc
xiaomings: For a 60 fps video, this callback wake up the decoder thread around 60 times a seconds in addition to the existing loop and waking ups.  Is this necessary?

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,269 @@
...
+  // decoded frame. See b/405467220#comment36 for details.
+  if (state.decoded_frames == 0) {
+    return state.total_frames() >= kMaxAllowedFramesWhenNoDecodedFrameYet;
+  }
+  return state.total_frames() >= max_frames_;
+}
+
+void DecoderStateTracker::EngageKillSwitch_Locked(std::string_view reason,
+                                                  int64_t pts) {
+  SB_LOG(ERROR) << "KILL SWITCH ENGAGED: " << reason << ", pts=" << pts;
```
xiaomings, line 245: Please work with @osagie98 to run YTS with this enabled on all devices in the lab, and see if it causes any regressions or how frequently the algorithm will be disabled.

starboard/android/shared/player_components_factory.cc
```diff
@@ -546,6 +546,8 @@ class PlayerComponentsFactory : public starboard::shared::starboard::player::
                    << " of " << flush_delay_usec << "us during Flush().";
     }
 
+    // TODO: b/455938352 - Connect this options to h5vcc settings.
+    VideoDecoder::FlowControlOptions flow_control_options;
```
xiaomings, line 550: Is it possible for you to provide us with the values you used for testing, so I could get a concrete idea about the potential values?

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,100 @@
...
+  bool reached_max_ = false;                         // Guarded by |mutex_|.
+
+  // Non-resettable members start.
+  // These variables are preserved across calls to Reset().
+  bool disabled_ = false;  // Guarded by |mutex_|.
+  int max_frames_;         // Guarded by |mutex_|.
+  // Non-resettable members end.
+
+  // NOTE: |job_thread_| must be the last variable declared so that it is
+  // destroyed first (following the reverse order of declaration).
```
xiaomings: It's ok to keep this comment as is as it's still better to have this as the last member variable, but this is already handled in the dtor and the order no longer matters.

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: changed method names. I consulted Gemini,which recommneds to use the authoritative verb as method names, since StateTracker that manages the frame lifecycle, 


Old Name | Final "Accounting" Name
-- | --
OnFrameAdded | TrackNewFrame
OnEosAdded | MarkEosReached
OnFrameDecoded | MarkAsDecoded
OnReleased | MarkAsReleased

</body></html>

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Agreed. I’ve removed the comment as it could provide a false sense of security. Simply placing `job_thread_` last in the declaration order isn't enough; we still need explicit destructor logic to clean up pending tasks.

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.h
kjyoun: Renamed

starboard/android/shared/player_components_factory.cc
kjyoun: During testing,  I used these values
- max_pending_inputs_size 10 
- initial_max_frames_in_decoder 6

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: I tested various devices and will try a couple of more devices in our office.
Let me know, if there any device that you want to check

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: limit the size of tracker to 24 and changed to std::vector to address cache-friendliness

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Ack. will work with @osagie98 to test with YTS

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: It turns out that MediaDecoder thread polls every 1 msec. Then we don't need this thread.

As long as MediaDecoder polls frequently enough, we can rely its polling logic.

kjyoun:
- Removed JobThread scheduling, since MediaCodecDecoder main loop is polling fast enough

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,302 @@
...
+  }
+  PruneReleasedFrames_Locked();
+}
+
+void DecoderStateTracker::MarkFrameReleased(int64_t presentation_us,
+                                            int64_t release_us) {
+  std::lock_guard lock(mutex_);
+  if (disabled_) {
+    return;
+  }
```
xiaomings: When the new feature is disabled, the existing implementation is almost the same as the previous one, except that this function is called with an extra lock.  It probably doesn't matter, but we would further reduce the performance cost by turning |disabled_| to an atomic.

kjyoun:
checked how DecoderStateTracker affects main loop interval.

and confirmed that it doesn't affect polling interval.

https://diff.googleplex.com/#key=1a8vjWQrlz1L

starboard/shared/starboard/media/decoder_state_tracker.cc
kjyoun: Done

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,326 @@
...
+      return "frame size exceeeds limit: size=" +
+             std::to_string(frames_in_flight_.size()) +
+             ", limit=" + std::to_string(kMaxFramesToTrack);
+    }
+    // If frames_in_flight_ grows to an unexpected size, we kill the
+    // |DecoderStateTracker| since something has gone wrong.
+    if (frames_in_flight_.size() >
+        std::max(max_frames_, kMaxAllowedFramesWhenNoDecodedFrameYet)) {
+      return "Unexpected # of frames in flight: state=" +
+             ToString(GetCurrentState_Locked());
```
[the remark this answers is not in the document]
kjyoun: will keep current code, which is more concise.

starboard/shared/starboard/media/decoder_state_tracker.cc
```diff
@@ -0,0 +1,326 @@
...
+  auto it = std::lower_bound(
+      frames_in_flight_.begin(), frames_in_flight_.end(), presentation_us,
+      [](const auto& pair, int64_t pts) { return pair.first < pts; });
+
+  // This is fail-safe logic.
+  std::string kill_reason = [&]() -> std::string {
+    if (frames_in_flight_.size() >= kMaxFramesToTrack) {
+      return "frame size exceeeds limit: size=" +
+             std::to_string(frames_in_flight_.size()) +
+             ", limit=" + std::to_string(kMaxFramesToTrack);
```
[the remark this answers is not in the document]
kjyoun: will keep current code, which is more concise.

starboard/shared/starboard/media/decoder_state_tracker.h
```diff
@@ -0,0 +1,91 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef STARBOARD_SHARED_STARBOARD_MEDIA_DECODER_STATE_TRACKER_H_
+#define STARBOARD_SHARED_STARBOARD_MEDIA_DECODER_STATE_TRACKER_H_
+
+#include <atomic>
+#include <functional>
+#include <map>
```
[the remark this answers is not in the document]
kjyoun: Done

PR #8186, opened 2025-11-24
[tvOS] Remove a few more unused files imported from C25

rakuco:
* `system_request_stop.mm` is only used when the app exit stragegy is Close, which is not the case for tvOS.
* `localized_strings.*` are no longer used after #8162.

Bug: 459507205


rakuco:

PR #8193, opened 2025-11-25
Ensure EvergreenConfig is initialized before Crashpad

niranjanyardi:
This commit fixes a race condition where the Crashpad handler was being initialized before the EvergreenConfig singleton, leading to a null pointer dereference.

The `InstallCrashpadHandler` call in `loader_app.cc` is moved to after the library loading logic, guaranteeing that `EvergreenConfig::Create` is invoked before any component attempts to access the singleton.

Additionally, a defensive `SB_CHECK` statement has been added to `evergreen_config.cc` to provide a clear error message if the `EvergreenConfig` is accessed before it has been initialized.

Issue: 461621299

### DIFF
--- a/starboard/elf_loader/evergreen_config.cc
+++ b/starboard/elf_loader/evergreen_config.cc
@@ -35,6 +35,9 @@ void EvergreenConfig::Create(
 
 const EvergreenConfig* EvergreenConfig::GetInstance() {
   std::lock_guard lock(g_evergreen_config_mutex);
+  SB_CHECK(g_evergreen_config)
+      << "EvergreenConfig::GetInstance() called but g_evergreen_config is null "
+         "and not initialized.";
   return g_evergreen_config.get();
 }
 
--- a/starboard/loader_app/loader_app.cc
+++ b/starboard/loader_app/loader_app.cc
@@ -270,9 +270,6 @@ void SbEventHandle(const SbEvent* event) {
       }
     }
 
-    InstallCrashpadHandler(
-        command_line.GetSwitchValue(elf_loader::kEvergreenContent));
-
     if (is_evergreen_lite) {
       loader_app::RecordSlotSelectionStatus(SlotSelectionStatus::kEGLite);
       LoadLibraryAndInitialize(alternative_content, use_memory_mapped_file);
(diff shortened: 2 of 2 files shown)


andrewsavage1:
1. Does a similar change need to be made for elf_loader_sandbox.cc? https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/elf_loader/elf_loader_sandbox.cc;l=124-128?q=elf_loader_sandbox.cc&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain
2. Is this a race condition or simply a misordering? I thought this was all initialized in a single-threaded fashion
3. Did this happen on multiple platforms or only RDK? Either way, why did we only notice it on RDK?

hlwarriner:
We should fix this a different way. Moving installation of Crashpad to after the evergreen library is loaded means we wouldn't catch and report crashes that occur during library loading.

I'm *guessing* the problem is that when the certificates directory is located so that it can be passed to `crashpad::InstallCrashpadHandler()`, the RDK implementation of `SbSystemGetPath()` calls `elf_loader::EvergreenConfig::GetInstance()` and doesn't check whether the pointer to the instance is null (if the library hasn't yet been loaded, it will be).

See https://github.com/youtube/cobalt/blob/26.eap/starboard/linux/shared/system_get_path.cc#L101. This shared linux implementation, which is used by the x64 and raspi ports, checks whether the pointer is null.

edit: `src/third_party/starboard/rdk/shared/system/system_get_path.cc` actually does also have this check. Let's talk offline or in the bug, as it's not clear to me where the null pointer dereference is happening.

hlwarriner:
> We should fix this a different way. Moving installation of Crashpad to after the evergreen library is loaded means we wouldn't catch and report crashes that occur during library loading.
> 
> I'm _guessing_ the problem is that when the certificates directory is located so that it can be passed to `crashpad::InstallCrashpadHandler()`, the RDK implementation of `SbSystemGetPath()` calls `elf_loader::EvergreenConfig::GetInstance()` and doesn't check whether the pointer to the instance is null (if the library hasn't yet been loaded, it will be).
> 
> See https://github.com/youtube/cobalt/blob/26.eap/starboard/linux/shared/system_get_path.cc#L101. This shared linux implementation, which is used by the x64 and raspi ports, checks whether the pointer is null.
> 
> edit: `src/third_party/starboard/rdk/shared/system/system_get_path.cc` actually does also have this check. Let's talk offline or in the bug, as it's not clear to me where the null pointer dereference is happening.

I added a new comment at b/461621299#comment2. I suggest we discuss there @niranjanyardi @andrewsavage1.

niranjanyardi:
thanks @hlwarriner , i opned a new cl on gerrit to address this, added you there

PR #8197, opened 2025-11-25
[v8] BACKPORT: [sandbox] Do not attempt to use an alternative stack on tvOS

rakuco:
Backport the build system part of https://crrev.com/c/6642758, as the code that actually uses the new V8_OS_TVOS macro is not present in M138. This is part of the work to get interpreted WASM to work for tvOS.

Bug: 462369287

---

Original CL description:

sigaltstack() is a forbidden API on tvOS, and calling it causes a build error.

* Add V8_OS_TVOS and V8_TARGET_OS_TVOS in a way that mimics what we already do in Chromium: it is a subset/variant of the iOS build, so that these macros are set if and only if the corresponding IOS ones are set as well.
* Make EnsureAlternativeSignalStackIsAvailableForCurrentThread() do nothing for tvOS.

Change-Id: I0b22308544b2ae6fb0128ea1c0d41b817616525c
Reviewed-on: https://chromium-review.googlesource.com/c/v8/v8/+/6642758
Cr-Commit-Position: refs/heads/main@{#100841}

### DIFF
--- a/v8/include/v8config.h
+++ b/v8/include/v8config.h
@@ -85,6 +85,7 @@ path. Add it with -I<path> to the command line
 //  V8_OS_DARWIN        - Darwin (macOS, iOS)
 //  V8_OS_MACOS         - macOS
 //  V8_OS_IOS           - iOS
+//  V8_OS_TVOS          - tvOS (also sets V8_OS_IOS)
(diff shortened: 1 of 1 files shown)


rakuco:

PR #8202, opened 2025-11-25
cobalt: s/std::vector<char>/std::string/ for SbSystem*()

yell0wd0g:
This CL refactors `std::vector<char>` uses for `SbSystem*()` function
calls into `std::string` calls. Gemini-generated.

The bug is only for one such case, but I roped in all similar 
situations.

Bug: 452143961


yell0wd0g:
@yuying-y PTAL assuming Gemini code review points to major flaws.

yuying-y (approved):

PR #8212, opened 2025-11-26
[tvOS] Create a basic cobalt.app using ios_app_bundle

rakuco:
Patches by @jkim-julie and @abhijeetk. This makes the `cobalt` target on tvOS build a `cobalt.app` bundle that can be run on the simulator or deployed on the device.

This PR contains the build system scaffolding, as the newly-added main.mm does not do much yet.

Bug: 434004373

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -46,10 +46,10 @@ group("gn_all") {
   testonly = true
 
   # TODO(b/371589344): Fix android build configs.
-  deps = [
-    "//starboard:starboard_group",
-    "//starboard/nplb",
-  ]
+  deps = [ "//starboard:starboard_group" ]
+  if (!cobalt_is_release_build) {
+    deps += [ "//starboard/nplb" ]
+  }
   if (is_android) {
     # TODO(cobalt): reuse build rules below to define Android target.
     deps += [ "android:cobalt_apk" ]
--- a/starboard/android/shared/test_filters.py
+++ b/starboard/android/shared/test_filters.py
@@ -59,6 +59,7 @@ _FILTERED_TESTS = {
         # This test is failing because localhost is not defined for IPv6 in
         # /etc/hosts.
         'SbSocketAddressTypes/SbSocketResolveTest.Localhost/filter_ipv6_type_ipv6',
+        'PosixSocketHints/PosixSocketResolveTest.Localhost/family_inet6_*',
 
         # SbDirectory has problems with empty Asset dirs.
         'SbDirectoryCanOpenTest.SunnyDayStaticContent',
--- a/starboard/nplb/posix_compliance/posix_socket_helpers.cc
+++ b/starboard/nplb/posix_compliance/posix_socket_helpers.cc
@@ -17,7 +17,13 @@
 #include <sched.h>
 #include <sys/socket.h>
 
+#include <string>
+#include <tuple>
+#include <utility>
+
+#include "build/build_config.h"
 #include "starboard/common/log.h"
+#include "starboard/common/string.h"
 #include "starboard/common/time.h"
 #include "starboard/shared/posix/handle_eintr.h"
 #include "starboard/thread.h"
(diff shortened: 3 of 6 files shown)


rakuco:

cobalt/BUILD.gn
```diff
@@ -23,11 +23,18 @@ if (!toolchain_for_rust_host_build_tools) {
...
+}
+
 # This arg temporarily enables or disables the updater_sandbox target.
 # It's disabled on CI by default due to missing libcurl pkg. To enable it
 # locally, set it to true before building the target.
 declare_args() {
   build_updater_sandbox = false
+
+  # Cobalt app bundle identifier for tvOS.
+  cobalt_bundle_identifier = "foo.cobalt.cobalt"
```
yell0wd0g: Update this? E.g. `dev.cobalt.something` where something could be
`tvos` or similar. For reference, the developer version of Cobalt
on AndroidTV has a class path starting with `dev.cobalt.coat` (the
`coat` is from "Cobalt on AndroidTV").

rakuco: FWIW this came from C25 (e.g. https://github.com/youtube/cobalt/blob/5dbf892d2c673f88f7427c3b30b1b5948a7dbe81/starboard/darwin/tvos/arm64/launcher.py#L29), the difference is that it's adjustable as a GN argument.

Glad to change it to something else though. `dev.cobalt.tvos`?

cobalt/BUILD.gn
```diff
@@ -69,7 +78,40 @@ group("cobalt_browsertests") {
   deps = [ "testing:cobalt_browsertests" ]
 }
 
-if (!is_android) {
+if (is_ios && target_platform == "tvos") {
+  ios_app_bundle("cobalt") {
+    info_plist = "shell/app/ios-Info.plist"
```
fayev-g: This wasn't included in C25, right? I couldn't find an Info.plist set in C25 and I need to case it out for the production build because we can only have one and blaze creates it.

rakuco: Correct, this [originally came from Content Shell](https://github.com/youtube/cobalt/blob/main/content/shell/app/ios/ios-app.plist). In C25, the plist was [here](https://github.com/youtube/cobalt/blob/25.lts.1%2B/starboard/darwin/tvos/shared/BuildInfo.plist).

Do you mean this should be a GN argument as well?

rakuco: > Correct, this [originally came from Content Shell](https://github.com/youtube/cobalt/blob/main/content/shell/app/ios/ios-app.plist). In C25, the plist was [here](https://github.com/youtube/cobalt/blob/25.lts.1%2B/starboard/darwin/tvos/shared/BuildInfo.plist).

(and there are other commits we need to upstream that fork this ios-Info.plist and make it look more similar to the C25 version)

cobalt/BUILD.gn
rakuco: Done: the default value is now `dev.cobalt.tvos`.

cobalt/BUILD.gn
rakuco: I've made the value configurable via the `cobalt_info_plist` GN argument.

cobalt/BUILD.gn
```diff
@@ -69,7 +81,40 @@ group("cobalt_browsertests") {
   deps = [ "testing:cobalt_browsertests" ]
 }
 
-if (!is_android) {
+if (is_ios && target_platform == "tvos") {
```
andrewsavage1, line 86: Should we add a subplatform gn flag for is_ios_tvos upstream similar to the  BUILDFLAG?

cobalt/BUILD.gn
```diff
@@ -23,11 +23,21 @@ if (!toolchain_for_rust_host_build_tools) {
...
+}
+
 # This arg temporarily enables or disables the updater_sandbox target.
 # It's disabled on CI by default due to missing libcurl pkg. To enable it
 # locally, set it to true before building the target.
 declare_args() {
   build_updater_sandbox = false
+
+  # tvOS: Default app bundle identifier.
+  cobalt_bundle_identifier = "dev.cobalt.tvos"
```
andrewsavage1: Should we declare these for all platforms?

andrewsavage1 (approved):

cobalt/BUILD.gn
rakuco: This was the result of a veeery long discussion upstream when we started enabling tvOS in upstream Chromium and we ended up settling on using `target_platform` so that this can be eventually used for e.g. "watchos" and other variants that are also kind of subsets of iOS itself.

What can work in the Cobalt case is just assuming that `is_ios` implies `target_platform == "tvos"` since it's the only iOS-based platform we care about, it'd just be tricky to only allow this in Cobalt-specific code.

cobalt/BUILD.gn
rakuco: Not necessarily. I thought it wouldn't hurt, but I've pushed a commit restricting these arguments to tvOS.

PR #8213, opened 2025-11-26
[XB] Fix suspend hang with synchronous event dispatching

msolya:
GOOGAMCONS-219

Refactored OnSuspending to prevent MOAPPLICATION_HANG errors during app suspension:

  - Run suspend continuation on background thread using use_arbitrary() context to avoid blocking CoreApplication thread
  - Dispatch lifecycle events (Blur, Conceal, Freeze) synchronously to CoreApplication thread via WaitForComplete + RunAsync pattern
  - Consolidate two-step task chain into single continuation for cleaner flow
  - Add threading diagnostics with thread ID logging to track execution

This ensures proper event completion before releasing extended resources and prevents the application from hanging during suspend operations.


msolya:
PR was closed, because corresponding codebase was move to another repository

PR #8214, opened 2025-11-26
[tvOS] Remove outdated @available() checks

rakuco:
Remove checks for versions that are always supported by Chrobalt.

Bug: 459507205


rakuco:

PR #8216, opened 2025-11-26
cobalt: Remove //cobalt/renderer/rasterizer

yell0wd0g:
I think it's unused?

Bug: None


yell0wd0g:
Never mind! It's used.

PR #8221, opened 2025-11-27
tvOS: Support Identifier for Advertising

Gyuyoung:
This PR implements Identifier for Advertising support
by using the tvOS APIs directly instead of relying on
SbSystemGetProperty.

The following APIs are used:

  - requestTrackingAuthorizationWithCompletionHandler
  - ASIdentifierManager.advertisingIdentifier

In addition, this PR introduces the implementation under
//cobalt instead of //starboard, since the value can be
retrieved directly on tvOS without interacting with
//starboard.

Lastly, this CL removes the existing ifa.h/mm files and
related code.

TEST: https://yts.devicecertification.youtube/2021/manual/ifa.html

Bug: 453538923
Bug: 458160672


cobalt/shell/app/ios-Info.plist
```diff
@@ -52,5 +52,7 @@
 	<string>Allow content_shell access to microphone</string>
 	<key>NSCameraUsageDescription</key>
 	<string>Allow content_shell access to camera</string>
+	<key>NSUserTrackingUsageDescription</key>
```
Gyuyoung: This PR adds it here for now. But this part will be added to tvos-Info.plist, which we will add later.

cobalt/browser/h5vcc_system/h5vcc_system_impl_starboard.cc
```diff
@@ -49,48 +37,25 @@ h5vcc_system::mojom::UserOnExitStrategy GetUserOnExitStrategyInternal() {
...
-  StarboardBridge* starboard_bridge = StarboardBridge::GetInstance();
-  limit_ad_tracking = starboard_bridge->GetLimitAdTracking(env);
-#elif BUILDFLAG(IS_IOS_TVOS)
-  // TODO: b/447135715 - Implement ad tracking limit status for tvOS.
-  NOTIMPLEMENTED();
-#else
-#error "Unsupported platform."
-#endif
   return limit_ad_tracking;
 }
```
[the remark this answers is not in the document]
Gyuyoung: This PR doesn't modify the existing logic except for the tvOS part.

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
```diff
@@ -0,0 +1,120 @@
...
+void GetAdvertisingIdShared(
+    H5vccSystemImpl::GetAdvertisingIdCallback callback) {
+  NSString* advertising_id =
+      [[ASIdentifierManager sharedManager].advertisingIdentifier UUIDString];
+  std::move(callback).Run(base::SysNSStringToUTF8(advertising_id));
+}
+
+bool GetLimitAdTrackingShared() {
+  return [ASIdentifierManager sharedManager].isAdvertisingTrackingEnabled == NO;
+}
```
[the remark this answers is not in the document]
Gyuyoung: We're only targeting the latest version(above 14.0 version at least?). So I know we decided not to use the available() keyword. Applied using `ATTrackingManager.trackingAuthorizationStatus`.

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
```diff
@@ -0,0 +1,120 @@
...
+      return std::string("AUTHORIZED");
+    case ATTrackingManagerAuthorizationStatusDenied:
+      return std::string("DENIED");
+    case ATTrackingManagerAuthorizationStatusRestricted:
+      return std::string("RESTRICTED");
+    case ATTrackingManagerAuthorizationStatusNotDetermined:
+      return std::string("NOT_DETERMINED");
+  }
+  NOTREACHED();
+}
```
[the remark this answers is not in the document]
Gyuyoung: We don't support old version of tvOS. But, if we should support the old version, please let me know.

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
```diff
@@ -0,0 +1,120 @@
...
+      base::BindPostTask(base::SequencedTaskRunner::GetCurrentDefault(),
+                         base::BindOnce(
+                             [](RequestTrackingAuthorizationCallback callback,
+                                ATTrackingManagerAuthorizationStatus status) {
+                               std::move(callback).Run();
+                             },
+                             std::move(callback))));
+  [ATTrackingManager
+      requestTrackingAuthorizationWithCompletionHandler:completion];
+}
```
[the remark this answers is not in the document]
Gyuyoung: Commented out `status` in the completion handler's lambda. And we don't support the old versions of tvOS. So we don't use available() here.

Gyuyoung:

rakuco:
Since all commits will be squashed together, it might be easier to send 2 PRs: one splitting the existing code into multiple files and another one adding the tvOS implementation.

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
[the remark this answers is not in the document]
fayev-g: When we launch Chrobalt, we won't, no. I'll make sure the App Store listing also doesn't support the old version.

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
```diff
@@ -0,0 +1,121 @@
...
+    case ATTrackingManagerAuthorizationStatusAuthorized:
+      return std::string("AUTHORIZED");
+    case ATTrackingManagerAuthorizationStatusDenied:
+      return std::string("DENIED");
+    case ATTrackingManagerAuthorizationStatusRestricted:
+      return std::string("RESTRICTED");
+    case ATTrackingManagerAuthorizationStatusNotDetermined:
+      return std::string("NOT_DETERMINED");
+  }
+  NOTREACHED();
```
fayev-g, line 59: The previous implementation returned "UNKNOWN", which I assume wouldn't cause a runtime crash/error. I think this is ok to fail silently? Or is this meant to throw an error?

fayev-g:
> Since all commits will be squashed together, it might be easier to send 2 PRs: one splitting the existing code into multiple files and another one adding the tvOS implementation.

Yeah, we discussed in another PR I personally prefer separating out impls like this (consistent with Google C++ style), but we can send that splitting code to android team, as well. Lgtm for the tvos code with one comment.

Thanks, @Gyuyoung !

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
Gyuyoung: I think that the previous implementation returns "UNKNOWN" when the SDK version is lower than tvOS 14, since that status is only supported on tvOS 14 and above. However, we don’t support older versions of tvOS, so I didn’t take the "UNKNOWN" case into consideration.

FYI, according to the documentation[1], the status can only be one of four defined values. Therefore, it should never be anything outside those four cases. In situations like this, we typically add an assertion. 

[1] https://developer.apple.com/documentation/apptrackingtransparency/attrackingmanager/authorizationstatus?language=objc

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
```diff
@@ -0,0 +1,149 @@
...
+    RequestTrackingAuthorizationCallback callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  auto completion = base::CallbackToBlock(base::BindPostTask(
+      base::SequencedTaskRunner::GetCurrentDefault(),
+      base::BindOnce(
+          [](RequestTrackingAuthorizationCallback callback,
+             ATTrackingManagerAuthorizationStatus /*status*/) {
+            std::move(callback).Run();
+          },
+          std::move(callback))));
```
yell0wd0g: nit: I vaguely recall there's a neat glue template to bind `callback` 
and ignore `status`, which is `base::IgnoreArgs<const AccountId&>`.
These lines would be something like
```
  base::BindOnce(base::IgnoreArgs<ATTrackingManagerAuthorizationStatus>(std::move(callback)));
```
WDYT?

https://source.chromium.org/chromium/chromium/src/+/main:base/functional/callback_helpers.h;l=120-137;drc=622f4ceb71ee4de06b2d856f10cf415af4c1773e;bpv=1;bpt=1

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
```diff
@@ -0,0 +1,149 @@
...
+  [ATTrackingManager
+      requestTrackingAuthorizationWithCompletionHandler:completion];
+}
+
+void H5vccSystemImpl::GetUserOnExitStrategy(
+    GetUserOnExitStrategyCallback callback) {
+  std::move(callback).Run(h5vcc_system::mojom::UserOnExitStrategy::kMinimize);
+}
+
+void H5vccSystemImpl::Exit() {
```
yell0wd0g: The method of this body is the same for the other platforms, 
could we refactor it?

starboard/tvos/shared/system_get_extensions.cc
```diff
@@ -47,9 +46,6 @@ const void* SbSystemGetExtension(const char* name) {
   if (strcmp(name, kCobaltExtensionPlatformServiceName) == 0) {
     return starboard::shared::uikit::GetPlatformServiceApi();
   }
-  if (strcmp(name, kStarboardExtensionIfaName) == 0) {
-    return starboard::shared::uikit::GetIfaApi();
-  }
```
yell0wd0g, line 52: Instead of removing the code here and in the next file, could
we leave a guard like
```
  DCHECK_NE(strcmp(name, kStarboardExtensionIfaName), 0) <<
    "IFA not supported via Starboard";
```
or some such? For posterity. Only for debug builds/logs should
be fine.

Gyuyoung:
Add @johnxwork for the Android changes.

rakuco:
> Could we add some @ guard to make sure these files are
> only built on tvOS > 14

Is there a particular reason for that? I removed some of the existing checks in #8214, especially since we're targeting newer tvOS versions anyway (see https://github.com/youtube/cobalt/blob/main/build/config/ios/ios_sdk_overrides.gni).

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
rakuco: Hmm true, this patch was written before all these cookie-related changes from #8130. I think it makes sense to move this to `_base.cc` and have each platform file implement `PerformExitStrategy()` instead.

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
Gyuyoung: Thank you for letting me know that. It looks like a better way to ignore unused arguments. I think it may be changed like below to keep the current behavior. Applied.

```
      base::IgnoreArgs<ATTrackingManagerAuthorizationStatus>(base::BindOnce(
          [](RequestTrackingAuthorizationCallback callback) {
            std::move(callback).Run();
          },
          std::move(callback)))));
```

cobalt/browser/h5vcc_system/h5vcc_system_impl_tvos.mm
Gyuyoung: ok, I moved this to `_base.cc`, then implement `PerformExitStrategy()` in each platform file.

starboard/tvos/shared/system_get_extensions.cc
Gyuyoung: IMHO, it's good to show a message, then return null or false as follows,

```
SB_LOG(INFO).  << "IFA is not supported via Starboard.";
return NULL;
```

Gyuyoung (commented):
I updated the changes according to the comments except for the tvOS version check. And the gemini already reviewed this PR as soon as I uploaded.

Gyuyoung:
Hmm, it looks like the test failures on android CIs are not related to this PR. @fayev-g @yell0wd0g Do you have more comments on this PR?

rakuco:
> Hmm, it looks like the test failures on android CIs are not related to this PR.

Unfortunately you just have to keep trying. You can either amend and force-push or add/re-add the "runtest" label here.

rakuco:
@fayev-g can you enable auto-merge for this PR? I think that's the only thing missing to land this.

PR #8223, opened 2025-11-28
[tvOS] Make tvOS code in Starboard use ANGLE

rakuco:
Set `gl_type="angle"` just like the linux-x64x11 port already does, which causes the `//starboard/egl_and_gles` GN config properly add the ANGLE dependencies to the `starboard_platform` target.

This also requires teaching FakeGraphicsContextProvider to use ANGLE's Metal backend on Apple platforms, as the OpenGL ES backend is not available at least on tvOS.

Note that this makes the custom EGL/GLES code in the tree, which came from C25, even more broken. It is only used for decode to texture (which is broken for all ports), and will be removed in a follow-up PR.

Bug: 456692793
Bug: 447334535


rakuco:

starboard/testing/fake_graphics_context_provider.cc
```diff
@@ -70,6 +70,8 @@
 
 #define EGL_PLATFORM_ANGLE_TYPE_OPENGL_ANGLE 0x320D
 #define EGL_PLATFORM_ANGLE_TYPE_OPENGLES_ANGLE 0x320E
+
+#define EGL_PLATFORM_ANGLE_TYPE_METAL_ANGLE 0x3489
```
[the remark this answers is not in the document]
rakuco: Wrong, `<EGL/eglext.h>` is not included by this file.

fayev-g, line 74: jooc: Is it just known to copy this from eglext_angle? Is it possible to reference this from third_party? No need to change in this PR.

fayev-g (approved):

starboard/testing/fake_graphics_context_provider.cc
rakuco: There are places in Chromium that do the same, e.g. https://source.chromium.org/chromium/chromium/src/+/main:ui/gl/gl_context_egl.cc;l=32;drc=d37e0bf8bdce6e702ecd770ed2054dfca765b39e and https://source.chromium.org/chromium/chromium/src/+/main:ui/ozone/platform/wayland/gpu/wayland_gl_egl_utility.cc;l=9;drc=f983ceceaea537af44c483bff969470b995a2641

This file does not directly include any EGL/GL headers directly because the idea is to go through the indirections in `starboard/egl.h` and `starboard/gl.h` that end up leaving the implementation of these symbols to each port.

It is possible to include the EGL headers when using ANGLE assuming there's a target in ANGLE that just pulls in the headers, it would just make the behavior different from the non-ANGLE case where everything is resolved at runtime or redefined in `starboard/{egl,gl}.h`.

fayev-g: Interesting. I noticed that about Chromium, yeah. I've seen in other Google codebases, copies are usually wrapped in some if-then linter checker to provide some reference at the very least. Thanks for the explanation!

PR #7099, opened 2025-09-08
Starboard: Add starboard experiment to Tunnel Mode

johnedocampo:
`kForceTunnelMode` used to be a boolean in
`player_components_factory.cc` that could be changed manually; this PR makes it so that this value can now be changed on startup through the command line (implemented through a Starboard Feature).

Bug: 443614771

PR #6178, opened 2025-06-26
[media] Refactor Buffer State Change in StarboardRenderer

borongc:
This PR refactors buffer state change as `OnBufferingStateChange()`, which can be used to report buffering is empty when there is a audio/video underflow.

This PR also reorders the initialization variables in starboard_renderer.h to align with starboard_renderer.cc.

Issue: 307362589

PR #8240, opened 2025-12-02
feat(media): Add critical memory pressure signal before playback

sherryzy:
Dispatches a `MEMORY_PRESSURE_LEVEL_CRITICAL` signal at the beginning of `StarboardRenderer::StartPlayingFrom` to proactively free up memory before video playback, which is a memory-intensive operation.

This change was validated by adding a temporary 10-second delay after the signal was fired, allowing for memory measurement via `adb shell dumpsys meminfo`. Testing showed a memory reduction of approximately 10 MB immediately after the signal was dispatched. The primary contributor to this memory saving was the eviction of resources from Blink's `MemoryCache`. The experiment is done on 26.android branch with Sabrina device.

This signal triggers a one-time cache eviction in listeners; they do not store the memory pressure state, making it unnecessary to revert the pressure level back to moderate or none.

Bug: 454078657

### DIFF
--- a/media/base/media_switches.cc
+++ b/media/base/media_switches.cc
@@ -1704,6 +1704,12 @@ BASE_FEATURE(kLibaomUseChromeThreads,
              "LibaomUseChromeThreads",
              base::FEATURE_ENABLED_BY_DEFAULT);
 
+// When enabled, Cobalt will notify memory pressure listeners before
+// video playback.
+BASE_FEATURE(kNotifyMemoryPressureBeforePlayback,
+             "NotifyMemoryPressureBeforePlayback",
+             base::FEATURE_DISABLED_BY_DEFAULT);
+
 #if BUILDFLAG(IS_WIN)
 // Controls whether to use D3D12 video decoder instead of D3D11 when supported.
 BASE_FEATURE(kD3D12VideoDecoder,
--- a/media/starboard/starboard_renderer.cc
+++ b/media/starboard/starboard_renderer.cc
@@ -16,6 +16,7 @@
 
 #include "base/feature_list.h"
 #include "base/logging.h"
+#include "base/memory/memory_pressure_listener.h"
 #include "base/strings/stringprintf.h"
 #include "base/task/sequenced_task_runner.h"
 #include "base/trace_event/trace_event.h"
@@ -644,6 +645,15 @@ void StarboardRenderer::CreatePlayerBridge() {
     player_bridge_->SetVolume(volume_);
 
     state_ = STATE_FLUSHED;
+    if (base::FeatureList::IsEnabled(
+            media::kNotifyMemoryPressureBeforePlayback)) {
+      // Send a one-time critical memory pressure signal to ask
+      // other components to release memory.
+      base::MemoryPressureListener::NotifyMemoryPressure(
+          base::MemoryPressureListener::MEMORY_PRESSURE_LEVEL_CRITICAL);
+      LOG(INFO) << "Firing a criticial memory pressure signal to reduce memory "
+                   "burden.";
+    }
     std::move(init_cb_).Run(PipelineStatus(PIPELINE_OK));
     return;
   }
--- a/media/base/media_switches.h
+++ b/media/base/media_switches.h
@@ -548,6 +548,8 @@ MEDIA_EXPORT BASE_DECLARE_FEATURE(kRenderMutedAudio);
 // Enable experimental headless captions.
 MEDIA_EXPORT BASE_DECLARE_FEATURE(kHeadlessLiveCaption);
 
+MEDIA_EXPORT BASE_DECLARE_FEATURE(kNotifyMemoryPressureBeforePlayback);
+
 // Based on a |command_line| and the current platform, returns the effective
 // autoplay policy. In other words, it will take into account the default policy
 // if none is specified via the command line and options passed for testing.


media/starboard/starboard_renderer.cc
```diff
@@ -341,6 +344,9 @@ void StarboardRenderer::StartPlayingFrom(TimeDelta time) {
     return;
   }
 
+  base::MemoryPressureListener::NotifyMemoryPressure(
```
borongc: `StartPlayingFrom` is pretty late as media pipeline already allocate DecoderBuffer and DemuxerStream, and this is called when starting the playback including seeks.

We may consider to move this to `StarboardRenderer::CreatePlayerBridge()` before `std::move(init_cb_).Run(PipelineStatus(PIPELINE_OK));`. This is when Cobalt initializes StarboardRenderer, and begins to start playbacks.

media/starboard/starboard_renderer.cc
```diff
@@ -18,6 +18,9 @@
 #include "base/logging.h"
 #include "base/strings/stringprintf.h"
 #include "base/task/sequenced_task_runner.h"
+#if BUILDFLAG(IS_COBALT)
```
borongc: StarboardRenderer is only used in Cobalt. See https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/media_options.gni;l=377

media/starboard/starboard_renderer.cc
```diff
@@ -341,6 +344,9 @@ void StarboardRenderer::StartPlayingFrom(TimeDelta time) {
     return;
   }
 
+  base::MemoryPressureListener::NotifyMemoryPressure(
+      base::MemoryPressureListener::MEMORY_PRESSURE_LEVEL_CRITICAL);
```
jasonzhangxx: Garbage collection at the beginning of every playback will bring a concern of increased playback startup latency.

kjyoun:
re: Testing showed a memory reduction of approximately 10 MB immediately after the signal was dispatched. 

Can you share more details about the test?
e.g. 
- how did you measured memory saving. PSS ? RSS?
- measuring interval . every 1 sec, or 10 sec ?

Previously, I couldn't see the difference by firing memory pressure signal. https://b.corp.google.com/issues/454078657#comment2

Hence I want to know what the difference is
Hence I

sherryzy:
@kjyoun The main difference for testing is that I added a 10 second sleep right after this event is dispatched, this is to avoid RSS/PSS go up quickly because of the video playback. 

My test plan:
1. Navigating through videos quickly.
2. Stop navigation and do ```adb sysdump meminfo```
3. Play a video -> you can observe a spinner and the memory pressure signal is dispatched.
4. Do``` adb sysdump meminfo``` again. Note: the video is not played, and shows a black screen.

You can observer the RSS/PSS drop.

media/starboard/starboard_renderer.cc
sherryzy: @jasonzhangxx Do you know the exact impact on playback startup latency? Alternatively, we could explore preventing the media component from reacting to this event. I've observed that this event significantly reduces memory in other areas (e.g., resource_pool, memory_cache, etc.). Guidance on how to test this playback startup latency would be greatly appreciated to quantify the impact.

sherryzy:
@kjyoun  Here is some data for reference:

Before the signal is dispatched: TOTAL PSS:   290811            TOTAL RSS:   330532       TOTAL SWAP PSS:     4807
After: TOTAL PSS:   278652            TOTAL RSS:   308352       TOTAL SWAP PSS:    14691

Note: Please add ```base::PlatformThread::Sleep(base::Seconds(10));``` right after the line that memory pressure signal is sent.  After you see the event is dispatched from logging, please wait for 2~3 second to let the cache being cleared, and let sysdump data reflect that.

kjyoun:
I gave several test runs (5 w/ this PR + 4 w/o it) on Kirkwood device.
Test results showed the this PR reduced ~23 MB of PSS. ( measured 20 secs after signal is fired)

https://paste.googleplex.com/4886143021744128

kjyoun:
As @jasonzhangxx  mentioned,this PR could negatively impact QoE. Starting a new video is one of the busiest moments, so performing garbage collection then may increase CPU burden.

Could we put this behind an experiment flag? That would allow us to clearly understand the implications of this change

sherryzy:
@kjyoun Thanks for redo the experiment and help confirm the memory reduction. For experiment flags, I assume you mean Finch which I think it's not entirely ready to be use yet. Can we already measure the impact through experiment flag today?

media/starboard/starboard_renderer.cc
```diff
@@ -644,6 +645,8 @@ void StarboardRenderer::CreatePlayerBridge() {
     player_bridge_->SetVolume(volume_);
 
     state_ = STATE_FLUSHED;
+    base::MemoryPressureListener::NotifyMemoryPressure(
```
yell0wd0g: micro-nit: Could we leave a comment here to clarify the
meaning? E.g.
```
  // Send a one-time critical memory pressure signal to ask
  // other components to release memory.
```
or some such?

kjyoun: and can you add log as well?

SB_LOG(INFO) << "Firing a criticial memory pressure signal to reduce memory burden" ?

kjyoun:
re: experiment,
main branch uses starboard::feature, which will be connected to Finch soon
m26 branch uses h5vcc settings.

since this PR goes to main branch, this PR should use starboard:feature.
e.g.) https://github.com/youtube/cobalt/pull/7099

My suggestion is
- In this PR, make this feature under constant disabled
```
// TODO: connect this constant to experiment flag.
constexpr bool kFireMemoryPressureSignal = false;

if (kFireMemoryPressureSignal) {
  base::MemoryPressureListener::NotifyMemoryPressure(....);
}
```
- Submit this PR to main and cherry-pick to 26.android
- On main, connect the constant to starboard::Feature
- On 26.android, connect the constant to h5vcc settgins (I will help you for this, since I need to make some changes on current h5vcc impl)

borongc:
> * On main, connect the constant to starboard::Feature

BASE feature should be sufficient, as this doesn't touch Starboard.

johnedocampo:
> ```
> // TODO: connect this constant to experiment flag.
> constexpr bool kFireMemoryPressureSignal = false;
> 
> if (kFireMemoryPressureSignal) {
>   base::MemoryPressureListener::NotifyMemoryPressure(....);
> }
> ```
> 
> * On main, connect the constant to starboard::Feature

`starboard::features` are features meant to be used under the starboard layer. Since this experiment flag is in `/media`, I think it would be more appropriate to put this experiment in `media/base/media_switches.cc` on main. This experiment can be added with other Cobalt experiments like [PR 6178](https://github.com/youtube/cobalt/pull/6178).

sherryzy:
Thanks all for the suggestions. I have updated the PR, and gated this experiment in media/base/media_switches.cc per @johnedocampo 's recommendations.

media/base/media_switches.cc
```diff
@@ -1704,6 +1704,12 @@ BASE_FEATURE(kLibaomUseChromeThreads,
              "LibaomUseChromeThreads",
              base::FEATURE_ENABLED_BY_DEFAULT);
 
+// When enabled, Cobalt will notify memory pressure listeners before
+// video playback.
+BASE_FEATURE(kNotifyMemoryPressureBeforePlayback,
+             "NotifyMemoryPressureBeforePlayback",
+             base::FEATURE_ENABLED_BY_DEFAULT);
```
kjyoun: don't  we have to make this to `FEATURE_DISABLED_BY_DEFAULT`  and then enable this when it passes the experiment?

sherryzy: Yeah, I can change it to DISABLED_BY_DEFAULT. I'm not clear about how to do this experiment though. Can you explain how to enable this when passes the experiment?

kjyoun: I think that this would  be process.

1. Add new code under experiment flag.
2. Run experiment
3. Verify the result.
4. If good, ramp up to 100%
5. Remove experiment flag, so that new code is always used, and remove the oid code path which are no more used

kjyoun (approved):

PR #8241, opened 2025-12-02
Update pinned pre-commit clang-format to m138 version (21.1.0)

fayev-g:
Bug: 461065037

Chromium m138 clang uses version 21. We still pin a very old version, 17, from m114. The discrepancy causes format-related merge conflicts when backporting changes from upstream. Chromium main is on 22. This PR closes the gap on clang versions and is consistent with the current rebase version.

### DIFF
--- a/starboard/common/murmurhash2.cc
+++ b/starboard/common/murmurhash2.cc
@@ -14,8 +14,9 @@
 
 #include "starboard/common/murmurhash2.h"
 
-#include <memory>
 #include <string.h>
+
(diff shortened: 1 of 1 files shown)


briantting (approved):

fayev-g:
> Formatting changes only affect Starboard/Cobalt code, right? No changes to upstream code FWICT. LGTM if yes.

Yes! Rebased/upstream code is using this version of clang-format already

PR #8244, opened 2025-12-02
android: Dismiss network error dialog once connected to Internet

ckim1346:
This PR introduces a `NetworkChangeNotifier` to the `CobaltConnectivityDetector.java` that watches for changes in network connectivity, and in this case specifically, for reconnection to an internet source. If a device goes from being unconnected to connected, another `activeNetworkCheck()` is performed and dismisses the dialog if it exists if the internet connection is valid. As observed in b/463699732#comment13. This PR also introduces a `WebContentsObserver` to `CobaltActivity.java` to determine whether a hard reload is required on reconnection to the internet (like on startup), or if the dialog should just dismiss without a reload (like when a video is playing).

Bug: 464033064
Bug: 463699732


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltConnectivityDetector.java
```diff
@@ -38,12 +40,30 @@ public class CobaltConnectivityDetector {
   private final CobaltActivity activity;
   private PlatformError platformError;
   protected boolean mShouldReloadOnResume = false;
+  private boolean mHasSuccessfullyLoaded = false;
+  private boolean mHasVerifiedConnectivity = false;
```
johnxwork, line 43: Too many variables and too many conditions. We just need one reliable online callback to dismiss the dialog, right?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -215,6 +219,33 @@ public void onWebContentsReady() {
...
+
+                @Override
+                public void didFinishNavigationInPrimaryMainFrame(NavigationHandle navigationHandle) {
+                  // A navigation can successfully commit a cached page while the network is
+                  // offline. We must check that the network is actually online before we
+                  // consider a page load to be a "successful first load".
+                  if (navigationHandle.hasCommitted()
+                      && !navigationHandle.isErrorPage()
+                      && NetworkChangeNotifier.isOnline()
+                      && mCobaltConnectivityDetector.hasVerifiedConnectivity()) {
```
johnxwork, line 238: can we simplify this to just NetworkChangeNotifier.isOnline()?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -215,6 +219,33 @@ public void onWebContentsReady() {
...
+                    String scheme = navigationHandle.getUrl().getScheme();
+                    if ("http".equals(scheme) || "https".equals(scheme)) {
```
borongc: We have a custom scheme `h5vcc-embedded` in https://github.com/youtube/cobalt/pull/8308.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: In the current implementation, I think it could be simplified to not include `NetworkChangeNotifier.isOnline`. In the weak wifi scenario, `NetworkChangeNotifier.isOnline()` will return true even though we don't have a valid internet connection which is why it's necessary to validate whether the `mCobaltConnectivityDetector` has validated connectivity. But this is a convoluted way to check essentially if the app has successfully loaded past startup -- is there an existing way to check this?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltConnectivityDetector.java
ckim1346: This has to do with the request in b/464033064#comment5 -- we need a way to determine whether to reload the app (no connection on startup) or to just dismiss the dialog (startup has successfully finished). This used to be handled by `cobalt/browser/browser_module` [OnNetworkFailureSystemPlatformResponse()](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/browser/browser_module.cc;drc=06382972bc9b7344d511798b29f2ffac8f5d3fd1;l=1345) which calls [Navigate](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/browser/browser_module.cc;drc=06382972bc9b7344d511798b29f2ffac8f5d3fd1;l=454) and is a pretty complicated process too.   We can either try to replicate that or find an easier way to verify whether the app has started up successfully (see below).

johnxwork: Would modifying shell's url achieve the same behavior as the old Navigate? https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/shell/browser/shell.cc;drc=c3f310866477aad0a66a03aef38821c31c2aec2e;l=280

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: ack

johnxwork (approved):

PR #8247, opened 2025-12-02
[tvOS] Disable AV1 codec support in media pipeline

abhijeetk:
Historically, tvOS does not support AV1 codec. This change adds
preprocessor guards to exclude AV1 from supported video codecs on tvOS
platforms.

This change ensures the media capabilities API and codec queries
accurately reflect that AV1 is not available on tvOS devices.

Test: https://yts.devicecertification.youtube/2021/?tests=87-123&test_type=formatsupport-test#1764679567726

Bug: 465356807

### DIFF
--- a/cobalt/build/configs/tvos_common.gn
+++ b/cobalt/build/configs/tvos_common.gn
@@ -22,3 +22,6 @@ treat_warnings_as_errors = false
 # dawn_shared_texture_cache.cc, for example).
 use_dawn = true
 skia_use_dawn = true
+
+# Disable AV1 decoder for tvOS
+enable_dav1d_decoder = false
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -76,8 +76,13 @@ std::string GetMimeFromAudioType(const ::media::AudioType& type) {
       ::media::EME_CODEC_VP8 | ::media::EME_CODEC_OPUS |
       ::media::EME_CODEC_VORBIS | ::media::EME_CODEC_MPEG_H_AUDIO |
       ::media::EME_CODEC_FLAC | ::media::EME_CODEC_HEVC_PROFILE_MAIN |
-      ::media::EME_CODEC_HEVC_PROFILE_MAIN10 | ::media::EME_CODEC_AV1 |
-      ::media::EME_CODEC_AC3 | ::media::EME_CODEC_EAC3;
+      ::media::EME_CODEC_HEVC_PROFILE_MAIN10 | ::media::EME_CODEC_AC3 |
+      ::media::EME_CODEC_EAC3;
+
+#if BUILDFLAG(ENABLE_AV1_DECODER)
+  codecs |= ::media::EME_CODEC_AV1;
+#endif
+
   // TODO(b/375232937) Add IAMF
   return codecs;
 }
--- a/starboard/tvos/shared/media_is_video_supported.mm
+++ b/starboard/tvos/shared/media_is_video_supported.mm
@@ -38,6 +38,10 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
                            int64_t bitrate,
                            int fps,
                            bool decode_to_texture_required) {
+  if (video_codec == kSbMediaVideoCodecAv1) {
+    return false;
+  }
+
   bool experimental_allowed = false;
 
   if (mime_type) {


abhijeetk:

abhijeetk:
cc : @jasonzhangxx @fayev-g @andrewsavage1 @yell0wd0g

rakuco:
You should add the relevant folks as Reviewers in the PR as well.

abhijeetk:
> You should add the relevant folks as Reviewers in the PR as well.

Done. Thanks for the suggestion.

jasonzhangxx:
Platform specific code is preferred to be kept in platform specific file (like in starboard/tvos/shared/media_is_video_supported.mm).

abhijeetk:
> Platform specific code is preferred to be kept in platform specific file (like in starboard/tvos/shared/media_is_video_supported.mm).

**OPTION 1 :**
Hi, the current implementation relies on a hardcoded codec list to determine EME-supported codecs. Similar to the open-source Chromecast implementation[1], we can use the `ENABLE_AV1_DECODER` compile-time flag to disable AV1.

`ENABLE_AV1_DECODER` is false on tvOS because it is derived from `enable_dav1d_decoder`[2], and the tvOS configuration explicitly sets `enable_dav1d_decoder` to false[3].

chromecast implementation in cobalt repository[4], do not include AV1 in supported list of eme codecs. 

[1] https://source.chromium.org/chromium/chromium/src/+/main:chromecast/renderer/media/key_systems_cast.cc;l=74?q=GetCastEmeSupportedCodecs  
[2] https://github.com/youtube/cobalt/blob/main/media/media_options.gni#L145 
[3] https://github.com/youtube/cobalt/blob/main/cobalt/build/configs/common.gn#L56
[4] https://github.com/youtube/cobalt/blob/main/chromecast/renderer/media/key_systems_cast.cc#L39

**OPTION 2 :**
As you suggested, we can detect codec support at runtime using `SbMediaCanPlayMimeAndKeySystem`, which takes a mime_type as its first parameter. We can derive this mime_type from the codec name by extending the existing `GetMimeFromVideoType`[1] function.

Example:
```
SbMediaCanPlayMimeAndKeySystem(GetMimeFromVideoType(::media::VideoType(::media::VideoCodec::kAV1)), "");
```
This will internally use the logic in starboard/tvos/shared/media_is_*.mm.

Since passing an empty key system ("") to `SbMediaCanPlayMimeAndKeySystem()` only validates audio/video codec support, we would also need a separate Starboard API (e.g., `SbGetSupportedKeySystemNames`) to return the supported key systems (such as com.youtube.widevine.l3). These values are currently hardcoded in starboard/tvos/shared/media_is_supported.mm[2]

Potential usage:
```
for (const auto& key_system : SbGetSupportedKeySystemNames()) {
  ...
  SbMediaCanPlayMimeAndKeySystem(
      GetMimeFromVideoType(::media::VideoType(::media::VideoCodec::kAV1)),
      key_system);
  ...
}
```
This approach would allow us to derive the accurate set of EME-supported codecs.
For targets like the simulator where no key systems are supported, we can simply return an empty EME codec list.

The main downside is that `GetMimeFromVideoType` has limitations, as it makes simplifying assumptions about codec profile and level support already mentioned in comment[3]

[1] https://github.com/youtube/cobalt/blob/main/cobalt/renderer/cobalt_content_renderer_client.cc#L36
[2] https://github.com/youtube/cobalt/blob/main/starboard/tvos/shared/media_is_supported.mm#L21
[3] https://github.com/youtube/cobalt/blob/main/cobalt/renderer/cobalt_content_renderer_client.cc#L31

abhijeetk:
For Option 1, https://github.com/youtube/cobalt/pull/8247/files

For Option 2, Proposed changes are https://github.com/abhijeetk/cobalt/commits/disable-av1-support-using-platform-check/, it contains two commits :

1. 81cf7e8f3da88bb5bb31ec8be96847b7a8dee916 : Use SbMediaCanPlayMimeAndKeySystem to detect platform-level codec support dynamically (as suggested by Jason).
2. d0af2c2322b5ee7ea9ec3d8a024a1d96b3a9d8c1 : An API to dynamically retrieve supported DRM key systems, allowing us to use it with SbMediaCanPlayMimeAndKeySystem to evaluate both codec and DRM compatibility. This builds on and improves the approach in commit 81cf7e8.

starboard/shared/starboard/media/mime_util.cc
```diff
@@ -148,6 +149,11 @@ bool IsSupportedVideoCodec(const ParsedMimeInfo& mime_info) {
       return false;  // No associated container in YT.
     case kSbMediaVideoCodecVc1:
     case kSbMediaVideoCodecAv1:
+#if BUILDFLAG(IS_IOS_TVOS)
```
yell0wd0g: @jasonzhangxx should have the last word, but if there's no specific
platform file(s) (e.g. //starboard/tvos/media/mime_util.cc or similar)
then adding a preprocessor guard here looks like the path of least
resistance.

abhijeetk: Just a friendly reminder, if there are no further comments, can we go ahead and merge this? Happy to adjust if Jason has a different preference once he gets a chance to weigh in. Thanks!

jasonzhangxx: There's platform file starboard/tvos/shared/media_is_video_supported.mm, which is used to decide if the codec can be supported by the platform. Please don't reject av1 in mime_util with platform flag.

abhijeetk: @jasonzhangxx : As per your suggestion, We have added explicit check in platform specific file starboard/tvos/shared/media_is_video_supported.mm to reject AV1.

jasonzhangxx: Yes, thanks for the change. It looks good to me.

PR #8249, opened 2025-12-02
Enable EG leak detector for most archs

andrewsavage1:
Additionally plugs or allows certain platform-specific leaks to align the platforms.

Bug: 463376302
Bug: 476128896
Bug: 476129004

### DIFF
--- a/third_party/cpuinfo/src/src/api.c
+++ b/third_party/cpuinfo/src/src/api.c
@@ -351,7 +351,7 @@ uint32_t CPUINFO_ABI cpuinfo_get_current_uarch_index(void) {
 		cpuinfo_log_fatal("cpuinfo_get_%s called before cpuinfo is initialized", "current_uarch_index");
 	}
 #if defined(ENABLE_BUILDFLAG_ENABLE_COBALT_HERMETIC_HACKS)
-  // TODO: b/476128896 - See if we need to support cpuinfo or need an alternate implementation here.
+	// TODO: b/476128896 - See if we need to support cpuinfo or need an alternate implementation here.
 	return 0;
 #else
 #if CPUINFO_ARCH_ARM || CPUINFO_ARCH_ARM64 || CPUINFO_ARCH_RISCV32 || CPUINFO_ARCH_RISCV64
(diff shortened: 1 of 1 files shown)


niranjanyardi (approved):

third_party/cpuinfo/src/src/api.c
```diff
@@ -378,13 +382,18 @@ uint32_t CPUINFO_ABI cpuinfo_get_current_uarch_index(void) {
...
 }
 
 uint32_t CPUINFO_ABI cpuinfo_get_current_uarch_index_with_default(uint32_t default_uarch_index) {
 	if CPUINFO_UNLIKELY (!cpuinfo_is_initialized) {
 		cpuinfo_log_fatal(
 			"cpuinfo_get_%s called before cpuinfo is initialized", "current_uarch_index_with_default");
 	}
+#if defined(ENABLE_BUILDFLAG_ENABLE_COBALT_HERMETIC_HACKS)
+	// TODO: b/476128896 - See if we need to support cpuinfo or need an alternate implementation here.
+	return 0;
```
niranjanyardi, line 395: Does this fail to build ?

BAsed on code below seems okay to return 0. 
if (cpuinfo_linux_cpu_to_uarch_index_map == NULL) {
	 * in the same package. */
	return 0;
#endif

niranjanyardi: Also might be good to change the format of the comment : go/cobalt-style#todo-format just to specify iut's a cobalt bug in upstream code

// TODO: b/87654321 - Cobalt: fix foo

starboard/tools/api_leak_detector/api_leak_detector.py
```diff
@@ -271,6 +271,10 @@
     'vswprintf',
     'write',
     'writev',
+
+    # TODO: b/476129004 - __clock_gettime64 is weakly defined for 32-bit arm
+    # platforms.
+    '__clock_gettime64',
```
niranjanyardi, line 275: I'm okay with solving this in a later PR. but idk if yavor has another opinion

third_party/cpuinfo/src/src/api.c
andrewsavage1: I don't know that we have any restrictions on how many types of cores systems can have, so we need to investigate still

PR #8255, opened 2025-12-02
This is a fix for fixing all cpp header pre-commit issues.

SortaCookie:
All incorrectly formated include guards and redundant imports should be fixed in this fix.

Bug: 435503470

Change-Id: Id11a22eb674e65f7d19971749b4cdd1b68e337c4

### DIFF
--- a/starboard/common/murmurhash2.cc
+++ b/starboard/common/murmurhash2.cc
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "starboard/common/murmurhash2.h"
+
 #include <string.h>
 
-#include "starboard/common/murmurhash2.h"
+#include <memory>
 
 namespace starboard {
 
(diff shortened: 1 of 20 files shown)


starboard/common/murmurhash2.cc
```diff
@@ -12,9 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "starboard/common/murmurhash2.h"
+
 #include <string.h>
 
-#include "starboard/common/murmurhash2.h"
```
Awallky, line 17: I think we want the system includes to be contiguous right?

Awallky (approved):
LGTM % comment

starboard/common/murmurhash2.cc
SortaCookie: My clang-format seems to want these two headers separated. Probably because one is c style?

PR #8260, opened 2025-12-02
cobalt/tools: Support Linux for UMA histogram polling

Awallky:
The UMA histogram polling script has been extended to support Cobalt
running on Linux, in addition to Android. This involves refactoring the
platform-specific logic for checking Cobalt process status and
establishing DevTools Protocol connections. A new, comprehensive
integration test suite has replaced the old unit tests to cover both
Android and Linux functionalities.

Bug: 462462878

### DIFF
--- a/cobalt/tools/uma/README.md
+++ b/cobalt/tools/uma/README.md
@@ -4,6 +4,8 @@ This directory contains Python scripts for interacting with and analyzing Cobalt
 
 ---
 
+**Note:** These tools have been tested on Android and Linux, but the  Linux builds have only been tested on Linux monolithic builds, not RDK builds. (TODO: b/470381238)
+
 ## 1. `pull_uma_histogram_set_via_cdp.py`
 
 A script to connect to a running Cobalt instance, poll for UMA histograms at a regular interval, and save the raw data.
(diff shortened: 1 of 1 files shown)


cobalt/tools/uma/README.md
```diff
@@ -30,19 +30,59 @@ vpython3 pull_uma_histogram_set_via_cdp.py [OPTIONS]
...
+
+This means the running Cobalt instance is rejecting the connection from the script. To fix this, you must launch Cobalt with the correct flags to enable the remote debugging port and allow connections from remote origins. The method for this differs by platform.
+
+**For Linux:**
+
+Launch Cobalt directly from your shell, adding the `--remote-debugging-port` and `--remote-allow-origins` flags.
+
+```bash
+# Example for Linux
+path/to/cobalt --remote-debugging-port=9222 --remote-allow-origins=*
```
sherryzy, line 69: Does this work for all builds (monolithics/modular/evergreen)? It would be great if you can help test this out on RDK, and add instructions here if extra steps are needed.

Awallky: So far, I have only tested this on monolithic Linux builds and Android. I don't have an RDK device set up for testing. Can we defer that to a separate test effort?

sherryzy: Yeah, I agree that we can follow up this later, but it would be good to add this information on the instruction file and mentioned it's for monolithic builds. 
Also can you create a bug for testing this out on RDK? It would be helpful to have this tool work on devices because the data (especially the memory usage)  will be more representative there.

Awallky: Done. I filed b/470381238 and noted this in the README for the linux section.

PR #8266, opened 2025-12-02
Add and rename jar files to Android package

briantting:
This moves logic from the snapshot script to packaging.

Bug: 454931711

### DIFF
--- a/cobalt/build/android/package.json
+++ b/cobalt/build/android/package.json
@@ -11,8 +11,260 @@
                 "lib.unstripped/libchrobalt.so",
                 "libchrobalt.so",
                 "obj/cobalt/shell/android/cobalt_shell_java_resources.resources.zip",
+                "obj/base/activity_state_java.javac.jar",
+                "obj/base/android_info_java.javac.jar",
+                "obj/base/base_cached_flags_java.javac.jar",
+                "obj/base/base_java.javac.jar",
+                "obj/base/base_java_test_support.javac.jar",
+                "obj/base/base_java_url_utils_for_test.javac.jar",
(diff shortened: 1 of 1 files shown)


zhongqiliang (approved):

cobalt/build/android/package.json
```diff
@@ -11,8 +11,260 @@
                 "lib.unstripped/libchrobalt.so",
                 "libchrobalt.so",
                 "obj/cobalt/shell/android/cobalt_shell_java_resources.resources.zip",
+                "obj/base/activity_state_java.javac.jar",
```
oxve, line 14: This looks like a lot of maintenance to me. How often does this list change?

Can you create a bug to set up a discussion about alternate solutions?

cobalt/build/android/package.json
```diff
@@ -34,6 +286,1022 @@
                 {
                     "from_file": "lib.unstripped/libchrobalt.so",
                     "to_file": "symbol/libchrobalt.so"
+                },
```
oxve, line 289: Does Chromium support globbing or wildcards here?

cobalt/build/android/package.json
briantting: We were in discussions to massively cut down this list to a minimal set, but that's pending them getting the bandwidth to figure that out. Not sure how different it'll be then. This list shouldn't change often otherwise. And yup.

cobalt/build/android/package.json
briantting: Not at the moment, but we can add it if we decide to go that route.

PR #8267, opened 2025-12-03
BACKPORT: [MediaSource] Don't change state for OOM append failures

borongc:
This CL backports https://chromium-review.googlesource.com/c/chromium/src/+/7218068 to address native crash at media::WebMStreamParser::AppendToParseBuffer().

Issue: 437916038

---
Original commit description:

The code in several parsers treated a ByteQueue::Push() failure as having modified the byte queue. Even though no bytes are appended to the queue in this case, the next append thinks there was and that those bytes were unprocessed.

There's not an easy way to add a test for this outside of adding a new `force_push_failure_for_testing_` type method, which seems of limited value at the unit test level.

(cherry picked from commit a5902827481ad4f138de1f2827e74b6ceb3a4bab)

Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/7218068


yell0wd0g:
OMG yes please \o/

jasonzhangxx (approved):

PR #8268, opened 2025-12-03
ci: Add comment body to gemini api call

oxve:
Update the commit message generation workflow to pass any text
following the '/generate-commit-message' command to the Gemini API.
This allows users to provide additional instructions or context, such
as bug numbers or specific formatting requirements, when requesting
a commit message.

Bug: 436660675


SortaCookie (approved):

zhongqiliang:
Is this tested? How good is the result?

.github/workflows/gemini-commit-message.yml
```diff
@@ -84,17 +85,21 @@ jobs:
...
+          COMMENT=${COMMENT_BODY#* }
+
+          # Use jq to safely combine the instructions with the variables.
           json_payload=$(jq -n \
             --arg instr "$PROMPT_INSTRUCTIONS" \
             --arg url "$PR_URL" \
             --arg title "$PR_TITLE" \
             --arg body "$PR_BODY" \
             --arg diff "$PR_DIFF" \
+            --arg comment "$COMMENT" \
```
zhongqiliang: Could too much info distract gemini? 
Even if we want to add comment, should we exclude comment from bots e.g gemini-review?

zhongqiliang (approved):

.github/workflows/gemini-commit-message.yml
oxve: This only applies to the comment that started with `/generate-commit-message`. This allows the comment poster to add context to the prompt.

PR #8270, opened 2025-12-03
Fix rpath linker error for tvOS

abhijeetk:
Exclude tvOS from rpath_for_built_shared_libraries config in cobalt BUILD.gn since tvOS platforms don't support the $ORIGIN syntax in rpath.

Bug: 432642933

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -80,7 +80,10 @@ if (!is_android) {
 
     sources = [ "app/cobalt.cc" ]
 
-    configs += [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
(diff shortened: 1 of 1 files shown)


abhijeetk:
@rakuco : How about using `(!is_apple)` to exclude all changes related to rpath ? 

I think, we should have same criteria for exclusion here and in file https://github.com/youtube/cobalt/blob/main/starboard/build/config/BUILD.gn#L67.

rakuco:
This is fixed by #8212 by having a separate block for tvOS, I don't see the point in changing the Starboard configurations while that PR isn't merged.

abhijeetk:
Ah, I missed that there’s ongoing work in this area. I’ll go ahead and abandon the PR.

PR #8277, opened 2025-12-03
test: Fix race condition in ArtworkDownloaderDefaultTest

kjyoun:
Address a race condition within `ArtworkDownloaderDefaultTest`. The test was unreliable because it asserted mock interactions before asynchronous background download operations completed. Introduce Mockito's timeout mechanism to ensure the test waits for these operations, thereby making the test reliable.

Bug: 456171826

### DIFF
--- a/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
+++ b/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
@@ -94,7 +94,7 @@ public class ArtworkDownloaderDefaultTest {
             },
             mDownloader) {
           @Override
-          public Bitmap consumeBitmapAndCropTo16x9(Bitmap bitmap) {
+          public Bitmap cropTo16x9(Bitmap bitmap) {
             // Return a dummy bitmap to simulate success and trigger the callback,
             // even if the download (simulated to fail here) produces a null bitmap.
             return Bitmap.createBitmap(1, 1, Bitmap.Config.ARGB_8888);


cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -48,7 +49,7 @@ public void testDownloadArtwork_CallsLoader() {
       mDownloader.downloadArtwork(url, mMockLoader);
 
       // Even if download fails, bitmap will be null, but it should still be processed.
-      verify(mMockLoader).consumeBitmapAndCropTo16x9(any());
-      verify(mMockLoader).onDownloadFinished(any(Pair.class));
+      verify(mMockLoader, timeout(1000)).consumeBitmapAndCropTo16x9(any());
+      verify(mMockLoader, timeout(1000)).onDownloadFinished(any(Pair.class));
```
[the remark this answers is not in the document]
kjyoun: adding constant here makes code verbose. Here `timeout(1000)` is readable enough

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
Awallky: I actually think we can do a bit better here. We can control the timing of the callbacks more here. Make a test class that overrides the `onDownloadFinished` method and the `consumeBitmapAndCropTo16x9`. We can then block on the callback, via a CountdownLatch from test code, make the test class unlock the latch, and \we can proceed from there.

What we probably want to add is a junit timeout so the test fails if we fail to unblock in our tests.
See https://source.chromium.org/chromium/chromium/src/+/main:android_webview/javatests/src/org/chromium/android_webview/test/AwContentsClientShouldInterceptRequestTest.java;drc=18c7228a07c6a6cb021d388820ba4b0206436807;l=474 for an example of the usage in WebView. 

But essentially in the test:

```
create lock (latch) 
pass test callback to unit under test
post mDownloader.downloadArtwork(url, mMockLoader); to a background thread
block on the lock in the test code
onDownloadFinished/consumeBitmapAndCropTo16x9 eventually call the callback
make our assertions in test code
```

Happy to chat offline if it is easier. :)

kjyoun: Done. Thanks for the tip

Awallky: Just to confirm, this does pass on local test runs right? Since we do not actually run our junit tests in CI.

kjyoun: yes, it passes local unit test run

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -40,15 +44,28 @@ public void setUp() {
...
 
-      String url = "http://example.com/image.png";
-      mDownloader.downloadArtwork(url, mMockLoader);
+    final CountDownLatch latch = new CountDownLatch(1);
+    doAnswer(
+            invocation -> {
+              latch.countDown();
+              return null;
+            })
+        .when(mMockLoader)
```
Awallky: Is there any reason we are not using a real ArtworkLoader here? 
https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkLoader.java;drc=1000c1227372e8c330dc9ff2b3a604eb283088d1;l=52

I think this ctor is a prime opportunity to inject our own callback into the class, which can unlock our latch.

kjyoun: I made change.

WDYT?

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
zhongqiliang, line 43: This pattern for setup is weird. 

Would it be better to do something like this?
  public void setUp() {
        mDownloader = new ArtworkDownloaderDefault();
        mSpyDownloader = spy(mDownloader);
  }

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -14,41 +14,70 @@
...
 
-      String url = "http://example.com/image.png";
-      mDownloader.downloadArtwork(url, mMockLoader);
+    ArtworkLoader artworkLoader =
+        new ArtworkLoader(
+            new ArtworkLoader.Callback() {
+              @Override
+              public void onArtworkLoaded(Bitmap bitmap) {
+                // This callback is posted to the main looper by ArtworkLoader.
+                // We will manually run the looper to trigger it.
```
Awallky: I was thinking we can just call latch countdown here, since onDownloadFinished calls the callback at the end. 
We could test both of these in separate unit tests to verify both calls are invoked as expected.

Awallky: So, let's keep this and add another where we override the callback and call latch countdown?

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
kjyoun: Removed mMockLoader.

WDYT?

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
kjyoun: Done

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
zhongqiliang: Oh, I got it wrong. 
You were mocking ArtworkLoader, I thought you mocked ArtworkDownloader interface.

zhongqiliang:
It would be good to cover the failure case. https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkDownloaderDefault.java;l=45

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -14,41 +14,85 @@
...
+          public synchronized void onDownloadFinished(Pair<String, Bitmap> urlBitmapPair) {
+            super.onDownloadFinished(urlBitmapPair);
+            latch.countDown();
+          }
+        };
 
-      // Even if download fails, bitmap will be null, but it should still be processed.
-      verify(mMockLoader).consumeBitmapAndCropTo16x9(any());
-      verify(mMockLoader).onDownloadFinished(any(Pair.class));
+    artworkLoader.mRequestedArtworkUrl = "http://example.com/image.png";
```
Awallky: Suggest making this String and the timeout constants declared at the top:

```
private static final String TEST_URL = "http://example.com/image.png";
private static final int TIMEOUT_SECONDS = 1;
```

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -14,41 +14,85 @@
...
+
+    String url = "http://example.com/image.png";
+    mDownloader.downloadArtwork(url, artworkLoader);
+
+    // The download runs on the calling thread (in this test context), calls onDownloadFinished,
+    // which posts to the main looper. Run the looper to execute the callback.
+    ShadowLooper.runUiThreadTasks();
+
+    assertThat(latch.await(1, TimeUnit.SECONDS)).isTrue();
```
Awallky: This seems shared between the two tests. Suggest refactoring and consolidating the shared logic. Something like: 

```
/**
   * Helper to execute the download, flush the looper, and assert the latch.
   */
  private void executeAndVerify(ArtworkLoader loader) throws InterruptedException {
   // doStuff...
  }
```

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
Awallky: Since they are reused at various spots.

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -14,41 +14,85 @@
...
-      // aren't mocking the network layer fully), the downloader still reports back to the loader.
+  public void testDownloadArtwork_CallsOnDownloadFinished() throws InterruptedException {
+    final CountDownLatch latch = new CountDownLatch(1);
 
-      String url = "http://example.com/image.png";
-      mDownloader.downloadArtwork(url, mMockLoader);
+    ArtworkLoader artworkLoader =
+        new ArtworkLoader(
+            mock(ArtworkLoader.Callback.class),
+            mock(ArtworkDownloader.class)) {
```
Awallky: Do we need to mock the downloader class here? We have an instance for our tests.

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
kjyoun: Done

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
kjyoun: Done

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
kjyoun: Done

kjyoun:
> It would be good to cover the failure case. https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkDownloaderDefault.java;l=45

Updated test to cover this case. When exception happens, bitmap is not set

kjyoun:
Yes, they passed locally

kjyoun:
gemini review

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
```diff
@@ -14,41 +14,93 @@
...
+      throws InterruptedException {
+    loader.mRequestedArtworkUrl = TEST_URL;
+
+    mDownloader.downloadArtwork(TEST_URL, loader);
+
+    // The download runs on the calling thread (in this test context), calls onDownloadFinished,
+    // which posts to the main looper. Run the looper to execute the callback.
+    ShadowLooper.runUiThreadTasks();
+
+    assertThat(latch.await(TIMEOUT_SECONDS, TimeUnit.SECONDS)).isTrue();
```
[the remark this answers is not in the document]
kjyoun: ack.

PR #8279, opened 2025-12-03
starboard: Permit DRM ticket re-take for provisioning retries

kjyoun:
Calling SessionUpdateRequest::ReleaseTicket() multiple times during DRM provisioning retries led to a crash due to an SB_CHECK_NE assertion, but which can happen on some devices.

This PR  renames ReleaseTicket() to TakeTicket() and removes the assertion. TakeTicket() now returns kSbDrmTicketInvalid on subsequent calls for the same request, signaling StarboardCdm to treat it as a
spontaneous message for an existing session. 

This ensures robust handling of multi-stage DRM provisioning retries without crashing.

Bug: 79941850


starboard/android/shared/drm_system.cc
```diff
@@ -112,8 +112,20 @@ DrmSystem::SessionUpdateRequest::SessionUpdateRequest(
...
+  //
+  // Subsequent calls to TakeTicket() for the same request will return
+  // `kSbDrmTicketInvalid`. This is expected behavior and signals to the
+  // `StarboardCdm`'s `OnSessionUpdateRequestGenerated` callback that this is a
+  // "spontaneous" message for an existing session (i.e., a retry or follow-up
+  // provisioning request) rather than an initial request requiring promise
+  // resolution.
   int ticket = ticket_;
   ticket_ = kSbDrmTicketInvalid;
   return ticket;
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/android/shared/drm_system.cc
```diff
@@ -112,11 +113,19 @@ DrmSystem::SessionUpdateRequest::SessionUpdateRequest(
...
+  // device requires multiple rounds of provisioning (See
+  // http://b/419320804#comment1).
+  //
+  // When TakeTicket() is called for the first time, it returns a valid
+  // ticket and sets 'ticket_' to `kSbDrmTicketInvalid`. Subsequent calls to
+  // TakeTicket() for the same request will return `kSbDrmTicketInvalid`. This
+  // is expected behavior and means that this is a "spontaneous" message for an
+  // existing session (i.e., a retry or deferred license request) rather than an
+  // initial request. See StarboardCdm's logic for "spontaneous" message:
+  // https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/starboard/starboard_cdm.cc;l=336-359;drc=274beb9dd42dc6c7d7dd3ff9415a41e1393f0133
```
jasonzhangxx: By removing `SB_CHECK_NE`,  `update_request_callback_` might be invoked with `kSbDrmTicketInvalid`, I don't think that's expected behavior, and there's a comment at https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/drm_system.h;l=103-106.

If android MediaDrm might generate error multiple times, I think we can make it only report error for once, thus `update_request_callback_` will be invoked only once and it won't fail the SB_CHECK here..

kjyoun: That comment was written by me, and this PR is intended to update it to reflect the correct behavior.

`kSbDrmTicketInvalid` is effectively a misnomer in this context. It doesn't signify an error; rather, it indicates a "spontaneous" DRM request generated by the DRM service itself, rather than one initiated by a fresh JavaScript call.

I updated the code to reflect this intention more clearly.

Here is the flow on an unprovisioned device:

1.  JavaScript initiates a license request (e.g., ticket ID 1).
2.  The DRM service responds with a provisioning request, using ticket ID 1.
    *   `TakeTicket` returns 1.
3.  Once the device is provisioned, the DRM service "spontaneously" generates the deferred license request.
    *   `TakeTicket` returns `kSbDrmTicketInvalid` (aka `kSpontaneousDrmTicketId`).
    *   This signals that the request is spontaneous (internal retry) and not tied to a new JavaScript request. As noted in `StarboardCdm`, this is a supported path.

kjyoun:
> nit: let's also update the comment at drm_system.h.

Updated drm_system.h.  

```
    // Returns the ticket. On the first call, it returns a valid ticket and
    // resets its internal state to kSbDrmTicketInvalid. Subsequent calls will
    // return `kSbDrmTicketInvalid`, which means a spontaneous drm request.
    int TakeTicket();
```

found that SbDrmSessionUpdateRequestFunc already has a necessary comment.

```
// |ticket| will be the same ticket that was passed to
// SbDrmGenerateSessionUpdateRequest() or |kSbDrmTicketInvalid| if the update
// request was generated by the DRM system.
```

> And I'm not sure if invoking `update_request_callback_` with invalid ticket is necessary but I believe it won't crash the app.

Yes, we have to call it. Otherwise, this spontaneous drm request(i.e. deferred license request) is not sent to javascript layer.

PR #8283, opened 2025-12-04
cobalt: Parallelize web contents load at startup

borongc:
This change introduces a mechanism to parallelize the loading of web
contents during application startup. A second WebContents instance
is created at shell initialization to pre-load content in the
background.

The system monitors load progress of the initial WebContents. Once
its main frame is loaded, a delayed task is posted to switch to the
pre-loaded WebContents. This aims to reduce perceived startup time
by making essential content available sooner, improving overall
application responsiveness and user experience.

The unit tests were generated with assistance from `gemini-cli` and
refined for correctness.

#vibe-coded

Issue: 463395742


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -100,6 +100,8 @@ public abstract class CobaltActivity extends Activity {
   private CobaltConnectivityDetector mCobaltConnectivityDetector;
   private WebContentsObserver mWebContentsObserver;
 
+  private Boolean mIsSplashScreenStarted = false;
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +882,63 @@ gfx::Size Shell::GetShellDefaultSize() {
...
     g_platform->SetTitle(this, entry->GetTitle());
   }
 }
 
+void Shell::SwitchToMainWebContents() {
+  // It is safe to use |is_switch_to_main_frame_|
+  // instead of a lock due to it is on a single thread.
+  // This could be called multiple times.
+  if (is_switch_to_main_frame_ == false) {
+    LOG(INFO) << "NativeSplash: Switching to main frarme WebContents.";
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/shell/browser/shell.h
```diff
@@ -112,7 +118,12 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
 
   static bool ShouldHideToolbar();
 
-  WebContents* web_contents() const { return web_contents_.get(); }
+  WebContents* web_contents() const {
+    return !is_main_frame_started_
+               ? web_contents_.get()
+               : (is_main_frame_loaded_ ? web_contents_.get()
+                                        : splash_screen_web_contents_.get());
+  }
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/shell/browser/splash_screen_web_contents_observer.cc
```diff
@@ -0,0 +1,36 @@
...
+namespace content {
+
+SplashScreenWebContentsObserver::SplashScreenWebContentsObserver(
+    WebContents* web_contents)
+    : WebContentsObserver(web_contents) {}
+
+SplashScreenWebContentsObserver::~SplashScreenWebContentsObserver() = default;
+
+void SplashScreenWebContentsObserver::WebContentsDestroyed() {
+  LOG(ERROR) << "NativeSplash: " << __func__;
```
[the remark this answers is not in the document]
borongc: Removed.

cobalt/shell/browser/splash_screen_web_contents_observer.cc
```diff
@@ -0,0 +1,36 @@
...
+  Observe(nullptr);
+}
+
+void SplashScreenWebContentsObserver::LoadProgressChanged(double progress) {
+  LOG(ERROR) << "NativeSplash: " << __func__ << " progress " << progress;
```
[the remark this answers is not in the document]
borongc: Removed.

cobalt/shell/browser/splash_screen_web_contents_delegate.cc
```diff
@@ -0,0 +1,33 @@
...
+namespace content {
+
+SplashScreenWebContentsDelegate::SplashScreenWebContentsDelegate(
+    base::OnceClosure on_close)
+    : on_close_(std::move(on_close)) {}
+
+SplashScreenWebContentsDelegate::~SplashScreenWebContentsDelegate() = default;
+
+void SplashScreenWebContentsDelegate::CloseContents(WebContents* source) {
+  LOG(ERROR) << "NativeSplash: " << __func__;
```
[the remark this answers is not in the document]
borongc: Removed.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +882,63 @@ gfx::Size Shell::GetShellDefaultSize() {
   return default_shell_size;
 }
 
-#if BUILDFLAG(IS_ANDROID)
 void Shell::LoadProgressChanged(double progress) {
+#if BUILDFLAG(IS_ANDROID)
   g_platform->LoadProgressChanged(this, progress);
-}
 #endif
+  LOG(ERROR) << "NativeSplash: " << __func__ << " progress " << progress;
```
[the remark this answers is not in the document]
borongc: Removed.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java
```diff
@@ -35,14 +35,7 @@ public abstract class CobaltService {
...
   public static String jsCodeTemplate =
-        "((w) => {\n"
-            + "  let targetWindow = w;\n"
-            + "  const appIframe = document.getElementById('anchor');\n"
-            + "  if (appIframe?.contentWindow) {\n"
-            + "    targetWindow = appIframe.contentWindow;\n"
-            + "  }\n"
-            + "  targetWindow.H5vccPlatformService.callbackFromAndroid(%d, '%s');\n"
-            + "})(window)";
+        "window.H5vccPlatformService.callbackFromAndroid(%d, '%s');";
```
borongc, line 38: Verified with local Kimono build, platform services can work.

cobalt/shell/browser/shell.h
```diff
@@ -112,7 +118,14 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
 
   static bool ShouldHideToolbar();
 
-  WebContents* web_contents() const { return web_contents_.get(); }
+  WebContents* web_contents() const {
+    // If splash screen WebContents is started, and main frame
+    // hasn't been loaded, display splash screen WebContents.
+    if (is_main_frame_started_ && !is_main_frame_loaded_) {
```
zhongqiliang: Is this thread safe?

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -153,7 +153,13 @@ void ShellBrowserMainParts::InitializeBrowserContexts() {
 
 void ShellBrowserMainParts::InitializeMessageLoopContext() {
   Shell::CreateNewWindow(browser_context_.get(), GetStartupURL(), nullptr,
-                         gfx::Size());
+                         gfx::Size(),
+#if BUILDFLAG(IS_ANDROID)
+                         false /* create_splash_screen_web_contents */
+#else
+                         true /* create_splash_screen_web_contents */
+#endif  // BUILDFLAG(IS_ANDROID)
```
zhongqiliang, line 157: Is it possible to unify this solution on all platforms?

Meaning, is the current Linux implementation much better than this version?

borongc: Linux creates Shell very early, and as noted here: https://source.chromium.org/chromium/chromium/src/+/main:content/shell/browser/shell_browser_main_parts.cc;l=83

We could make Android use the 1st Shell, instead of 2nd Shell by `mShellManager.launchShell(shellUrl);`, but I don't know if there is any impact.

cobalt/shell/browser/shell.h
borongc: It is only called on a single thread, so it should be thread safe.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -211,35 +213,41 @@ private void finishInitialization(Bundle savedInstanceState) {
         new Shell.OnWebContentsReadyListener() {
```
johnxwork, line 261: Can we put this listener into its own class and give it a proper name?

cobalt/shell/browser/shell.cc
```diff
@@ -367,11 +406,30 @@ void Shell::RegisterInjectedJavaScript() {
   }
 }
 
+void Shell::LoadSplashScreenWebContents() {
+  if (splash_screen_web_contents_) {
+    // Display splash screen.
+    LOG(INFO) << "NativeSplash: Loading splash screen WebContents.";
+    is_main_frame_started_ = true;
```
johnxwork: Is there a better place to put this line? I don't see how the main frame is started in this function?

cobalt/shell/browser/shell.h
```diff
@@ -112,7 +118,14 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
 
   static bool ShouldHideToolbar();
 
-  WebContents* web_contents() const { return web_contents_.get(); }
+  WebContents* web_contents() const {
```
johnxwork: I think you want to draw a clear boundary between web_contents_ and splash_screen_web_contents_ instead of letting this function return one of them based on timing.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +882,62 @@ gfx::Size Shell::GetShellDefaultSize() {
...
+    is_main_frame_loaded_ = true;
+
+    if (is_splash_screen_ended_) {
+      LOG(INFO) << "NativeSplash: Main frame WebContents is loaded.";
+      SwitchToMainWebContents();
+    } else {
+      LOG(INFO) << "NativeSplash: Main frame WebContents is loaded, but wait "
+                   "for splash screen WebContents.";
+      if (web_contents_) {
+        web_contents_->WasHidden();
```
johnxwork: This would result in an additional blur event on startup?

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +882,62 @@ gfx::Size Shell::GetShellDefaultSize() {
...
+        content::GetUIThreadTaskRunner({})->PostDelayedTask(
```
johnxwork: Do you expect `LoadProgressChanged()` to be invoked from non-UI thread such as the network thread? If so do we need something like a ThreadChecker?

cobalt/shell/browser/shell.h
```diff
@@ -219,19 +236,25 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
...
+  void ClosingSplashScreenWebContents();
 
   std::unique_ptr<JavaScriptDialogManager> dialog_manager_;
 
   std::unique_ptr<WebContents> web_contents_;
+  std::unique_ptr<WebContents> splash_screen_web_contents_;
+  bool is_main_frame_started_ = false;
+  bool is_main_frame_loaded_ = false;
+  bool is_splash_screen_ended_ = false;
+  bool is_switch_to_main_frame_ = false;
```
johnxwork: `has_switched_to_main_frame_`?

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +882,62 @@ gfx::Size Shell::GetShellDefaultSize() {
...
+            FROM_HERE,
+            base::BindOnce(&Shell::SwitchToMainWebContents,
+                           weak_factory_.GetWeakPtr()),
+            base::Milliseconds(kSplashTimeoutMs));
```
johnxwork: Would it be simple to do `kSplashTimeoutMs - time_since_start` to speed up startup?

cobalt/shell/browser/shell.h
johnxwork: Instead of using these two state variables, is it logically more intuitive to check splash_screen_web_contents_delegate_::on_close_ to tell if we have made the switch?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -211,35 +213,41 @@ private void finishInitialization(Bundle savedInstanceState) {
...
-            mShellManager.getActiveShell().loadUrl(mStartupUrl);
-
-            // Initialize and register a WebContentsObserver.
-            mWebContentsObserver =
-              new org.chromium.content_public.browser.WebContentsObserver(getActiveWebContents()) {
-                @Override
-                public void didStartNavigationInPrimaryMainFrame(NavigationHandle navigationHandle) {
-                  if (!navigationHandle.isSameDocument()) {
-                    mCobaltConnectivityDetector.setAppHasSuccessfullyLoaded(false);
+            if (!mIsSplashScreenStarted) {
```
johnxwork: Why do we expect onWebContentsReady() to be called twice?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: Would this be a TODO for @ckim1346?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: Yes, on ATV, switching WebContents in a single Shell will cause this be called more than once.

So for this PR, we have App WebContents --> Splash WebContents --> App WebContents, where App WebContents is the 1st created to ensure it can be initialized with Java Bridge.

cobalt/shell/browser/shell.cc
borongc: This should be `is_splash_screen_started_`.

cobalt/shell/browser/shell.cc
borongc: Right, would this cause any issue?

cobalt/shell/browser/shell.cc
borongc: No, this is only called on browser thread. For safety, I added `base::BindPostTaskToCurrentDefault()` to ensure it is posted back on the default browser thread.

cobalt/shell/browser/shell.cc
borongc: We could, but when Splash Screen is ended, it will call `Shell::ClosingSplashScreenWebContents()` to switch to main frame. This is expected to be earlier than this timeout-switching.

cobalt/shell/browser/shell.h
borongc: This is the common function used in content/shell/browser/shell_platform_delegate_views.cc (Linux) and content/shell/browser/shell_platform_delegate_android.cc (Android). Keeping it can avoid more modifications in platform specific files to have two WebContents.

cobalt/shell/browser/shell.h
borongc: > is it logically more intuitive to check splash_screen_web_contents_delegate_::on_close_ to tell if we have made the switch?

Note this is only applied to cold start to play splash screen.

We haven't implemented the cases for main app reloads (see go/chrobalt-splash-cases, test case C). That is, checking `splash_screen_web_contents_delegate_::on_close_` won't work for those cases.

cobalt/shell/browser/shell.h
borongc: Added a dedicate `splash_screen_web_contents()`.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: Removed `mIsSplashScreenStarted` as Shell.java ensures both WebContents.

cobalt/shell/browser/shell.cc
borongc: Done.

cobalt/shell/browser/shell.h
borongc: Added a variable State to manage the status of splash screen.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +885,79 @@ gfx::Size Shell::GetShellDefaultSize() {
...
+                   "screen elapsed: "
+                << splash_screen_elapsed.InMilliseconds()
+                << "ms, remaining delay: " << remaining_delay.InMilliseconds()
+                << "ms.";
+      if (web_contents_) {
+        web_contents_->WasHidden();
+        content::GetUIThreadTaskRunner({})->PostDelayedTask(
+            FROM_HERE,
+            base::BindPostTaskToCurrentDefault(base::BindOnce(
+                &Shell::SwitchToMainWebContents, weak_factory_.GetWeakPtr())),
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +885,79 @@ gfx::Size Shell::GetShellDefaultSize() {
...
   if (entry) {
     g_platform->SetTitle(this, entry->GetTitle());
   }
 }
 
+void Shell::SwitchToMainWebContents() {
+  // It is safe to use |has_switched_to_main_frame_|
+  // instead of a lock due to it is on a single thread.
+  // This could be called multiple times.
+  if (has_switched_to_main_frame_ == false) {
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +885,79 @@ gfx::Size Shell::GetShellDefaultSize() {
...
 }
 
-#if BUILDFLAG(IS_ANDROID)
 void Shell::LoadProgressChanged(double progress) {
+#if BUILDFLAG(IS_ANDROID)
   g_platform->LoadProgressChanged(this, progress);
-}
 #endif
+  if (progress >= 1.0 && state_ >= STATE_SPLASH_SCREEN_STARTED &&
+      is_main_frame_loaded_ == false) {
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/shell/browser/shell.cc
borongc: Keep it as it was, and this is suggested by gemini, where LoadProgressChanged is executed on the UI thread and this task is also posted to the UI thread, the base::BindPostTaskToCurrentDefault wrapper is redundant.

However, I added `DCHECK_CURRENTLY_ON(content::BrowserThread::UI);` to ensure it is on Browser UI thread.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
```diff
@@ -209,20 +219,63 @@ public ShellViewAndroidDelegate getViewAndroidDelegate() {
...
+        mWebContents = webContents;
+        mSplashScreenWebContents = null;
+        mNavigationController = mWebContents.getNavigationController();
+        if (mIsActivityVisible) {
+            mWebContents.updateWebContentsVisibility(Visibility.VISIBLE);
+        }
+        mContentViewRenderView.setCurrentWebContents(mWebContents);
+    }
+
+    public void onActivityStateChange(boolean visible) {
```
johnxwork: nit: `onActivityVisible`?

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
```diff
@@ -209,20 +219,63 @@ public ShellViewAndroidDelegate getViewAndroidDelegate() {
      */
     @CalledByNative
     private void initFromNativeTabContents(WebContents webContents) {
+        if (webContents == null || mContentViewRenderView == null) return;
```
johnxwork, line 222: Does this only happen when activity is quickly tore down after startup? Or is this possible in case of some startup race condition?

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
```diff
@@ -209,20 +219,63 @@ public ShellViewAndroidDelegate getViewAndroidDelegate() {
...
             mWebContentsReadyListener.onWebContentsReady();
         }
     }
 
+    /**
+     * Load the native splash screen contents.
+     * @param webContents A {@link WebContents} object.
+     */
+    @CalledByNative
+    private void loadSplashScreenNativeTabContents(WebContents webContents) {
```
johnxwork: nit: `webContents` -> `splashWebContents`

cobalt/shell/browser/shell.cc
johnxwork: It would be quite dangerous. And also we would fail YTS device lifecycle test. I guess the best method is to keep main frame invisible/blurred at startup, and only update when switching.

I'm not sure what changes are needed. Do we want a condition on CobaltActivity.java:443?

cobalt/shell/browser/splash_screen_unittest.cc
```diff
@@ -0,0 +1,385 @@
...
+
+  void CallLoadProgressChanged(Shell* shell, double progress) {
+    shell->LoadProgressChanged(progress);
+  }
+
+  void CallLoadSplashScreenWebContents(Shell* shell) {
+    shell->LoadSplashScreenWebContents();
+  }
+
+  void CallClosingSplashScreenWebContents(Shell* shell) {
```
johnxwork: Is this a test only function? Shall we name it that way?

cobalt/shell/browser/shell.h
```diff
@@ -219,19 +244,26 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
...
 
   void RegisterInjectedJavaScript();
+  void SwitchToMainWebContents();
+  void ClosingSplashScreenWebContents();
 
   std::unique_ptr<JavaScriptDialogManager> dialog_manager_;
 
   std::unique_ptr<WebContents> web_contents_;
+  std::unique_ptr<WebContents> splash_screen_web_contents_;
+  State state_;
```
johnxwork: nit: rename to `splash_state_` or `splash_frame_state_`

cobalt/shell/browser/shell.h
```diff
@@ -219,19 +244,26 @@ class Shell : public WebContentsDelegate, public WebContentsObserver {
...
+  bool skip_for_testing_;
+  bool is_main_frame_loaded_ = false;
```
johnxwork, line 265: instead of `is_main_frame_loaded_` plus `has_switched_to_main_frame_`, use `State main_frame_state_`?

cobalt/shell/browser/shell_browser_main_parts.cc
johnxwork: It's likely a TODO for when we switch to use h5vcc platform service

cobalt/shell/browser/shell.cc
```diff
@@ -367,11 +416,31 @@ void Shell::RegisterInjectedJavaScript() {
   }
 }
 
+void Shell::LoadSplashScreenWebContents() {
+  if (splash_screen_web_contents_) {
```
johnxwork, line 436: Do you want to turn this if() to a CHECK()?

cobalt/shell/browser/shell.h
borongc: Done. Renamed to `splash_state_`.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
borongc: Done.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
borongc: Done.

cobalt/shell/browser/shell.h
borongc: `is_main_frame_loaded_` is not always earlier than `has_switched_to_main_frame_`. If splash screen reaches its time-out before main frame is loaded, it will switch to main frame anyway, so we cannot put these two as a state.

cobalt/shell/browser/shell.cc
borongc: This ensures `splash screen WebContents` is only used once, but we will revisit this after the design review for `loading splash screen on-demand from WebApp when it reloads`.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
borongc: Right, there are edge cases as we saw in the prototype implementation.

johnxwork: I think crash is much easier to debug than ANR. Shall we just crash instead?

cobalt/shell/browser/shell.cc
johnxwork: I mean, we don't expect `splash_screen_web_contents_` to be null, right? In a successful bootup.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
borongc: There are cases that shell is terminated early before loading WebContents, e.g., https://github.com/youtube/cobalt/pull/8329. This is not a visible issue for users, but will still result in a crash in bug report.

cobalt/shell/browser/shell.cc
borongc: Right, we don't expect `splash_screen_web_contents_` to be null, but we don't strictly define the sequence of the function as well (e.g., initialing `splash_screen_web_contents_` before calling `LoadSplashScreenWebContents`).

However, I do agree we can put a `CHECK` here, but I prefer to leave it as a no-op if `splash_screen_web_contents_` is null.

cobalt/shell/browser/shell.cc
borongc: The intention here is to prevent from playing video from main app with the splash screen at the same time, as we only allow a single SbPlayer. So, if main frame WebContents is ready (it could be a video link), we should ask it to suspend until splash screen is finished.

We could switch to main frame anyway if it is loaded, but main app requires to play the full splash screen before switching to the main app.

johnxwork: I wonder if we can just switch, where do you see `main app requires to play the full splash screen before switching to the main app`?

cobalt/shell/browser/shell.cc
johnxwork: What's the consequence if `splash_screen_web_contents_` is null? I just don't want black screen.

borongc: It won't load splash screen and stay in main frame WebContents. There is one unit test case `NoSplashScreen` for this case.

cobalt/shell/browser/shell.cc
borongc: Tag you in the doc (go/chrobalt-splash-cases).

cobalt/shell/browser/shell_browser_main_parts.cc
johnxwork: Could you add the TODO comment? We expect idl platform service to be ready by end of Jan.

cobalt/shell/browser/shell.cc
johnxwork: Thanks. The cases look legit. Could you verify with `yts test <device_id> "Device Lifecycle"` before submission?

I think we have some redundant `updateWebContentsVisibility()`.

cobalt/shell/browser/shell_browser_main_parts.cc
borongc: I assume you refer to add a new api `h5vcc.system.showSplashScreen()` to allow WebApp to request playing splash screen on-demand. The plan is to bind the h5vcc idl with a function to create splash screen web contents again.

I don't have any plan to unify this solution on all platforms, as this is inherited from Content Shell.

cobalt/shell/browser/shell.cc
borongc: Removed redundant `updateWebContentsVisibility()` in Shell.java.

borongc: For yts test, as the current trunk is still broken for platform services, I used https://github.com/youtube/cobalt/pull/8423 to test it in 26.android branch using local Kimono build.

The results are the same as 6.34: https://paste.googleplex.com/4604105861496832.

cobalt/shell/browser/shell.cc
```diff
@@ -336,11 +381,31 @@ void Shell::PrimaryMainDocumentElementAvailable() {
       web_contents());
 }
 
+void Shell::DidFinishNavigation(NavigationHandle* navigation_handle) {
```
zhongqiliang, line 384: This is empty method, seems no need to overwrite it.

cobalt/shell/browser/shell.cc
```diff
@@ -823,16 +917,103 @@ gfx::Size Shell::GetShellDefaultSize() {
...
+    remaining_delay = base::TimeDelta();
+  }
+
+  VLOG(1) << "NativeSplash: Main frame WebContents is loaded, splash "
+             "screen elapsed: "
+          << splash_screen_elapsed.InMilliseconds()
+          << "ms, remaining delay: " << remaining_delay.InMilliseconds()
+          << "ms.";
+  if (web_contents_) {
+    web_contents_->WasHidden();
```
zhongqiliang, line 973: Should this line be moved into SwitchToMainWebContents()?
Or should it be moved into the PostDelayedTask() block?

cobalt/shell/browser/shell.cc
borongc: Right. This is for debugging purposes, which help in prod. I will make a follow-up PR to turn this to `LOG(INFO)`.

cobalt/shell/browser/shell.cc
borongc: This is to ensure main app WebContents should be suspended if splash screen is still in the front.

cobalt/shell/browser/shell.cc
borongc: Changed the type of logging in https://github.com/youtube/cobalt/pull/8660.

PR #8289, opened 2025-12-04
BACKPORT: ios_app_bundle: Allow intents_target to work with testonly=true

rakuco:
Backport https://crrev.com/c/7207678 (from M145), a build system fix that is required for the App Intents code from C25 to be built and integrated into Cobalt Shell without the extra scripts and alternative build system infrastructure that was used in the past.

Bug: 461350746
Bug: 433180638
Bug: 432824509

---

Original CL description:

Forward `testonly` from invokers of the action so that test bundles can also work with the `generate_intents` and `intents_target` GN infrastructure.

Change-Id: If724967c659c96677f979e6cfee46287e74774a6
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/7207678
Cr-Commit-Position: refs/heads/main@{#1553377}

### DIFF
--- a/build/config/ios/rules.gni
+++ b/build/config/ios/rules.gni
@@ -362,6 +362,8 @@ template("ios_app_bundle") {
         get_label_info(invoker.intents_target, "name") + ".module_info.json"
 
     action(_metadata_extraction) {
+      forward_variables_from(invoker, [ "testonly" ])
+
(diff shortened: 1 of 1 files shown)


rakuco:

fayev-g (approved):

PR #8290, opened 2025-12-04
evergreen: symbolize printed stack traces

hlwarriner:
This PR ports forward customizations to google::Symbolize from 25.lts.1+ that are needed to symbolize addresses from Evergreen libraries.

Issue: 460132262

### DIFF
--- a/base/third_party/symbolize/BUILD.gn
+++ b/base/third_party/symbolize/BUILD.gn
@@ -43,8 +43,7 @@ static_library("symbolize") {
 
   public_configs = [ ":symbolize_config" ]
 
-  if (is_starboard &&
-      use_evergreen &&
+  if (is_starboard && use_evergreen &&
       current_toolchain == starboard_toolchain) {
     deps = [ "//starboard/elf_loader:evergreen_info" ]
   }


hlwarriner:
I verified locally that, combined with https://github.com/youtube/cobalt/pull/8272, printed stack traces are symbolized both for the cobalt shared library (when an unstripped one is loaded), and for unit test libraries, when they crash.

base/third_party/symbolize/symbolize.cc
```diff
@@ -502,6 +507,18 @@ static char *GetHex(const char *start, const char *end, uint64_t *hex) {
   return const_cast<char *>(p);
 }
 
+#if BUILDFLAG(IS_STARBOARD) && BUILDFLAG(USE_EVERGREEN)
```
hlwarriner: I'm still deciding whether it would be better to guard these customizations more narrowly since they should only be needed by the starboard toolchain in evergreen builds, and not by the cobalt toolchain in evergreen builds. To do that we'd probably need to add a new buildflag so we can check `use_evergreen && !sb_is_evergreen` in C++ code.

hlwarriner: Note that `google::symbolize()` is also called from the `base/debug/stack_trace_posix.cc` implementation of `ProcessBacktrace()`: https://github.com/youtube/cobalt/blob/main/base/debug/stack_trace_posix.cc#L238.

I'm not sure if that code path is working correctly currently - see b/398296821 - but in any case I actually do think the correct thing to do here is to use these Evergreen customizations even when `google::Symbolize()` is called from the evergreen library itself, as opposed to from `SbSystemSymbolize()` (below Starboard). In both cases the call is happening in the same process memory space.

So I'll plan to keep this customization guard as is. But lmk if you disagree, @y4vor or anyone else.

y4vor: hm, using it from both Cobalt and Starboard could be a problem. We'll probably end up with 2 separate global instances of the EvergrenInfo struct. We might need to make EvergreenInfo to be SB extension to be able to use it from Cobalt.

hlwarriner: That's a good point. In that case are you ok with what I proposed above, for now? That is, guard these customizations more narrowly so that they're only made for the Starboard toolchain in Evergreen builds.

hlwarriner: I uploaded another commit to confine the customizations to the starboard toolchain. PTAL.

y4vor (approved):

base/third_party/symbolize/symbolize.cc
```diff
@@ -67,6 +67,20 @@
 #include "symbolize.h"
 #include "demangle.h"
 
+#include "build/build_config.h"
+// TODO: b/398296821 - Cobalt: to support google::Symbolize() when called from
```
andrewsavage1, line 71: Are you going to leave this bug open after this PR is merged? It'd make more sense to me to create a new bug in the backlog and add the context from the thread mentioned below into the bug instead of referring to it below.

base/third_party/symbolize/BUILD.gn
```diff
@@ -42,4 +42,10 @@ static_library("symbolize") {
   ]
 
   public_configs = [ ":symbolize_config" ]
+
+  if (is_starboard &&
```
andrewsavage1: nit: the formatting here seems wrong, we should `gn format` it (just the added code, not necessarily the whole file)

hlwarriner: done, thanks

base/third_party/symbolize/symbolize.cc
hlwarriner: Yeah, the todo is actually pointing to related bug b/398296821 rather than b/460132262 (the one fixed by this PR). b/398296821 will remain open and sure, I'll add a link to this thread there.

PR #8291, opened 2025-12-04
Skip over the CheckTreeIsOpen check.

SortaCookie:
This check is not relavant to Cobalt since it appears to be checking the status of builds of android tablet form factors.

Bug: 435503470


Awallky (approved):
LGTM, assuming this no longer causes the issue locally.

PR #8295, opened 2025-12-04
Move an impl file's "relevant header" to the top and exclude from clang-format when the naming is not the same

fayev-g:
Bug: 466129151

Follow up from https://github.com/youtube/cobalt/pull/8241, where we update clang-format to the m138 version. Google C++ style guide and clang-format enforce that an impl file's "relevant header" is at the top, but the cobalt codebase contains impl names that don't exactly match the header name and so clang-format moves those headers in the includes order. After offline discussion, exclude those headers from clang-format for now. Gemini recommends that we consolidate impls or split headers to satisfy formatting and style rules.


fayev-g:
I can apply this to 26.android and 26.eap, as well.

starboard/shared/starboard/player/filter/testing/audio_decoder_benchmark.cc
```diff
@@ -12,14 +12,17 @@
...
 
+// clang-format off
+#include "starboard/shared/starboard/player/filter/audio_decoder_internal.h"
+// clang-format on
+
 #include <algorithm>
 #include <memory>
 
-#include "starboard/common/log.h"
+#include "starboard/common/log.hh"
```
johnedocampo: There's an extra h here.

johnedocampo (commented):
Overall, everything looks great from the media side. It looks like [starboard/shared/starboard/player/player_create.cc ](https://github.com/youtube/cobalt/blob/main/starboard/shared/starboard/player/player_create.cc) is missing from the PR. Also left one small comment on a typo.

Thanks for doing this!!

starboard/shared/starboard/player/filter/testing/audio_decoder_benchmark.cc
fayev-g: Thanks for the catch!

johnedocampo (approved):
Changes on the media side LGTM!

PR #8299, opened 2025-12-05
media: Connect notify-memory-pressure-before-playback to h5vcc settings

kjyoun:
This change allows the h5vcc setting `Media.NotifyMemoryPressureBeforePlayback` to control the critical memory pressure signal sent before playback starts in `StarboardRenderer`.

Bug: 466374906


media/base/media_switches.cc
```diff
@@ -191,6 +191,7 @@ const char kMSEAudioBufferSizeLimitMb[] = "mse-audio-buffer-size-limit-mb";
 const char kMSEVideoBufferSizeLimitMb[] = "mse-video-buffer-size-limit-mb";
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
+const char kCobaltNotifyMemoryPressureBeforePlaybackSwitch[] = "cobalt-notify-memory-pressure-before-playback";
```
johnedocampo: Can you add a comment to this switch explaining what it is doing?

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -36,13 +36,17 @@ const char kH5vccSettingsKeyMediaEnableAllocateOnDemand[] =
     "Media.EnableAllocateOnDemand";
 const char kH5vccSettingsKeyMediaVideoBufferSizeClampMb[] =
     "Media.VideoBufferSizeClampMb";
+const char kH5vccSettingsKeyMediaNotifyMemoryPressureBeforePlayback[] =
```
johnedocampo, line 37: nit: move this definition to after "Media.EnableAllocateOnDemand", to keep the values sorted.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -36,13 +36,17 @@ const char kH5vccSettingsKeyMediaEnableAllocateOnDemand[] =
...
+    "Media.NotifyMemoryPressureBeforePlayback";
 
 // Map that stores all current bindings of H5vcc settings to media switches.
 // If a setting has a corresponding switch, we will enable the switch with the
 // corresponding value.
 const base::flat_map<std::string, const char*> kH5vccSettingToSwitchMap = {
     {kH5vccSettingsKeyMediaVideoBufferSizeClampMb,
      switches::kMSEVideoBufferSizeLimitClampMb},
+    {kH5vccSettingsKeyMediaNotifyMemoryPressureBeforePlayback,
+     switches::kCobaltNotifyMemoryPressureBeforePlaybackSwitch},
```
johnedocampo: nit: this entry should come before `kH5vccSettingsKeyMediaVideoBufferSizeClampMb`, to keep the map sorted.

johnedocampo (commented):
Changes for the h5vcc settings connection looks good, left some small comments about formatting.

cobalt/renderer/cobalt_content_renderer_client.cc
kjyoun: Done

cobalt/renderer/cobalt_content_renderer_client.cc
kjyoun: Done

media/base/media_switches.cc
kjyoun: Done

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -34,13 +34,17 @@ namespace {
...
+    "Media.NotifyMemoryPressureBeforePlayback";
 const char kH5vccSettingsKeyMediaVideoBufferSizeClampMb[] =
     "Media.VideoBufferSizeClampMb";
 
 // Map that stores all current bindings of H5vcc settings to media switches.
 // If a setting has a corresponding switch, we will enable the switch with the
 // corresponding value.
 const base::flat_map<std::string, const char*> kH5vccSettingToSwitchMap = {
+    {kH5vccSettingsKeyMediaNotifyMemoryPressureBeforePlayback,
+     switches::kCobaltNotifyMemoryPressureBeforePlayback},
```
borongc, line 47: Should this be `media::kCobaltNotifyMemoryPressureBeforePlayback`? So we don't need to change StarboardRenderer?

media/starboard/starboard_renderer.cc
```diff
@@ -110,6 +112,21 @@ int GetDefaultAudioFramesPerBuffer(AudioCodec codec) {
       return 1;
   }
 }
+
+bool ReadCommandLineSwitchForMemoryPressureSignal() {
```
borongc, line 116: We don't need this. As long as the BASE feature is enabled by switch flag, using `base::FeatureList::IsEnabled()` should be sufficient.

kjyoun: IIUC,  we cannot use `base::FeatureList` here, since `base::FeatureList` is initialized early in the process startup while  H5vcc settings are updated during the runtime.

To avoid confusion, I removed base feature(`BASE_FEATURE(kCobaltNotifyMemoryPressureBeforePlayback)`), keeping only switch(`kCobaltNotifyMemoryPressureBeforePlayback`)

cobalt/renderer/cobalt_content_renderer_client.cc
kjyoun: We need to use command line switch vs base feature.

media/starboard/starboard_renderer.cc
```diff
@@ -584,8 +605,7 @@ void StarboardRenderer::CreatePlayerBridge() {
     player_bridge_->SetVolume(volume_);
 
     state_ = STATE_FLUSHED;
-    if (base::FeatureList::IsEnabled(
```
borongc, line 587: I think we would like to keep BASE feature `kCobaltNotifyMemoryPressureBeforePlayback` as it is.

`kH5vccSettingToSwitchMap` is only a tentative solution, which we should remove it after finch is ready.

As a result, here should be something like:
```
    if (base::FeatureList::IsEnabled(
            media::kCobaltNotifyMemoryPressureBeforePlayback) ||
            ReadCommandLineSwitchForMemoryPressureSignal()) {
```

kjyoun: Done

PR #8309, opened 2025-12-05
Add Evergreen info in the ua string

yuying-y:
Issue: 466113841

### DIFF
--- a/base/threading/thread_restrictions.h
+++ b/base/threading/thread_restrictions.h
@@ -18,6 +18,10 @@
 #include "third_party/abseil-cpp/absl/types/optional.h"
 #endif
 
+#if BUILDFLAG(IS_STARBOARD)
+#include "base/version.h"
+#endif
+
 // -----------------------------------------------------------------------------
 // Usage documentation
 // -----------------------------------------------------------------------------
@@ -217,6 +221,7 @@ class SystemReportComponent;
 namespace cobalt {
 namespace updater {
 class UpdaterModule;
+base::Version ReadEvergreenVersion(base::FilePath installation_dir);
 }  // namespace updater
 }  // namespace cobalt
 #endif  // BUILDFLAG(IS_STARBOARD)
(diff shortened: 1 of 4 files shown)


cobalt/browser/user_agent/BUILD.gn
```diff
@@ -35,4 +35,8 @@ static_library("user_agent") {
       "ANDROID_ABI=$android_app_abi",
     ]
   }
+
+  if (use_evergreen) {
+    deps = [ "//cobalt/updater:updater" ]
```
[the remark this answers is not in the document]
yuying-y: `=` is correct instead of `+=`, because `deps` is not defined in the target. Changing it to `+=` caused this error:
ERROR at //cobalt/browser/user_agent/BUILD.gn:40:5: Undefined identifier.
    deps += [ "//cobalt/updater:updater" ]
    ^---
See //cobalt/BUILD.gn:217:5: which caused the file to be included.
    "//cobalt/browser/user_agent",
    ^----------------------------

andrewsavage1: nit: put this above the conditional block and use the shorthand `//cobalt/updater`

base/threading/thread_restrictions.h
andrewsavage1, line 74: Why is it not possible for us to use base::MayBlock instead for this call?

yuying-y: Tried to use base::MayBlock(). It doesn't work. The crash still happens.

andrewsavage1: Thanks for checking—would you create a bug in the 3P backlog to rework these methods to use async task posting and a TODO for it here?

yuying-y: At this point when ReadEvergreenVersion() is called, the UserAgentPlatformInfo object is being initialized, which is requested by GetCobaltUserAgent()/GetCobaltUserAgentMetadata() in cobalt/browser/cobalt_content_browser_client.cc. The thread has to wait for the EG version to be read, in order to get a correct ua string. I'm skeptical if this can be done in an async/non-blocking manner, if that's what you mean. Please explain if I misunderstood.

andrewsavage1: The task still could be (and sounds like it needs to be) blocking, but I think we could switch it so that the callsites use a PostTask with a MayBlock if we restructure more thoroughly—that's just one idea though, and it's possible I'm wrong, so I think the bug can be more broadly about how to get rid of the changes to upstream code there and not necessarily saying we have to switch it to async task posting like I said above

yuying-y: Ok, filed b/468067703, and added the TODO to the function definition. As a reference, when handling a similar issue for the updater destruction in cobalt/updater/updater_module.cc, I tried the PostTask() with base::MayBlock() approach. It didn't work, and the crash persisted. Then I figured out the approach to explicitly allow blocking in base/threading/thread_restrictions.h, leveraged by many classes/functions that need to be blocking. I believe that's why this approach exists.

PR #8313, opened 2025-12-06
Fix: SIGSEGV during FeatureList access in browser tests

Awallky:
Problem:
Browser tests were crashing with a SIGSEGV in `base::FeatureList::IsFeatureOverriddenFromCommandLine` during early browser process startup. The stack trace indicated this occurred during `BrowserMainLoop::PreCreateThreads()`, which is called within `BrowserTestBase::SetUp()`. Specifically, `GpuDataManagerImpl::GetInstance()` was being called during this phase, and its initialization attempted to access `base::FeatureList::GetInstance()`. However, `BrowserTestBase::SetUp()` intentionally calls `base::FeatureList::ClearInstanceForTesting()` just before this point, causing `base::FeatureList::GetInstance()` to return `nullptr` and resulting in a null-pointer dereference.

Previous attempts to fix this by modifying `NavigationBrowserTest`'s constructor or `SetUpOnMainThread()` were ineffective because the crash occurred earlier in the global test setup, before these test-specific methods were invoked.

The critical comment in the codebase, "It's unsafe to append the gpu command line switches to the global CommandLine::ForCurrentProcess object after threads are created," highlighted that `GpuDataManagerImpl` needs to be initialized early, which conflicted with the `FeatureList`'s lifecycle in the test environment.

Solution:
Introduced a temporary `std::unique_ptr<base::test::ScopedFeatureList> scoped_feature_list` immediately after `base::FeatureList::ClearInstanceForTesting()` in `content/public/test/browser_test_base.cc`. This `ScopedFeatureList` is initialized with `InitWithEmptyFeatureAndFieldTrialLists()`.

Rationale:
This solution ensures that a valid (even if empty) `FeatureList` instance is present during the critical early browser startup phases within `BrowserTestBase::SetUp()`, particularly when `GpuDataManagerImpl` is initialized. This prevents the null-pointer dereference that was causing the crash. The `scoped_feature_list` then goes out of scope, allowing the test harness to proceed with its intended `FeatureList` setup (which happens later in the Android test runner) without interference. This approach is idiomatic for handling such timing-dependent issues in Chromium's test infrastructure, providing a temporary valid state when the global instance is intentionally cleared.

Bug: 432313695

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkDownloaderDefault.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkDownloaderDefault.java
@@ -58,7 +58,7 @@ public class ArtworkDownloaderDefault implements ArtworkDownloader {
       }
     }
 
-    bitmap = artworkLoader.consumeBitmapAndCropTo16x9(bitmap);
+    bitmap = artworkLoader.cropTo16x9(bitmap);
     Log.i(TAG, "Artwork downloaded %s.", url);
     artworkLoader.onDownloadFinished(Pair.create(url, bitmap));
   }
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkLoader.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ArtworkLoader.java
@@ -110,12 +110,13 @@ public class ArtworkLoader {
    * wider, it is returned as is.
    *
    * The input bitmap is recycled if a new cropped bitmap is created.
+   * NOTE: This method is accessed from google3(http://shortn/_SoIqBrEPhR).
    *
    * @param bitmap The source bitmap.
    * @return The 16:9 cropped bitmap, or the original bitmap.
    */
   @CheckResult
-  public Bitmap consumeBitmapAndCropTo16x9(Bitmap bitmap) {
+  public Bitmap cropTo16x9(Bitmap bitmap) {
     // Crop to 16:9 as needed
     if (bitmap == null) {
       return null;
--- a/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
+++ b/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ArtworkDownloaderDefaultTest.java
@@ -48,7 +48,7 @@ public class ArtworkDownloaderDefaultTest {
       mDownloader.downloadArtwork(url, mMockLoader);
 
       // Even if download fails, bitmap will be null, but it should still be processed.
-      verify(mMockLoader).consumeBitmapAndCropTo16x9(any());
+      verify(mMockLoader).cropTo16x9(any());
       verify(mMockLoader).onDownloadFinished(any(Pair.class));
   }
 }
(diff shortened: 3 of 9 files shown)


content/public/test/browser_test_base.cc
```diff
@@ -534,6 +535,13 @@ void BrowserTestBase::SetUp() {
   // FeatureList::SetInstance, which expects no instance to exist.
   base::FeatureList::ClearInstanceForTesting();
 
+  // Temporarily initialize a ScopedFeatureList to ensure it's valid during
+  // early browser process setup, which might access it before the true
+  // FeatureList is established by the test harness. This bridges the gap
+  // between ClearInstanceForTesting() and the actual FeatureList setup.
+  base::test::ScopedFeatureList scoped_feature_list;
+  scoped_feature_list.InitWithEmptyFeatureAndFieldTrialLists();
+
```
yell0wd0g, line 557: One potential problem I think I see is that the GpuDataManagerImpl
creation, which in turn calls `base::CommandLine::ForCurrentProcess()`
([here](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/gpu/gpu_data_manager_impl_private.cc;l=468-476;drc=4594e8aee5a5d939426f2e605c454132aeaea5dc) if I'm not mistaken), would then see an empty list of flags and features,
correct? And if yes then the lines
```
  if (command_line->HasSwitch(switches::kSingleProcess) ||
      command_line->HasSwitch(switches::kInProcessGPU)) {
    AppendGpuCommandLine(command_line, GPU_PROCESS_KIND_SANDBOXED);
  }
```
would not see that we're using an in-process GPU and work,
perhaps, differently.

Although from the CL description (thanks BTW - very thorough) I 
see this would not be an issue because in browser tests, the command
line is empty. Could we leave an assert/comment in code e.g.
using `CHECK_IS_TEST()` ? Something like
```
  const bool feature_list_present = base::FeatureList::GetInstance();
  base::test::ScopedFeatureList scoped_feature_list_for_testing;
  if (!feature_list_present) {
    // Browser tests clear and reset the global FeatureList (via 
    // ClearInstanceForTesting()); production code always has a 
    // FeatureList, even if it's empty (which it never is, since 
    // at the very least it has the process type, 'browser'). 
    CHECK_IS_TEST();
    // If in testing, initialize a bogus FeatureList for 
    // GpuDataManagerImpl to be access, otherwise it crashes :(
    scoped_feature_list_for_testing.InitWithEmptyFeatureAndFieldTrialLists()
  }
```

WDYT? IOW my concern is to make explicit that this is a test-only
path, otherwise we would start tweaking the initialization sequence.

See [CHECK_IS_TEST()](https://source.chromium.org/chromium/chromium/src/+/main:base/check_is_test.h;l=12-14;drc=ad9b7a1a5e5b0aa9d2c57599ca2a7241ae17543b) documentation, seems very applicable.

Awallky: Done. I think that is a reasonable modification. I'm not sure if this will mask things or just be an improvement, but this is a fix for now. I think I will land this to get infra folks unblocked and file a bug to look into why Cobalt browsertest behavior deviates from upstream content browsertests. This will at least kick the can down the road to answering the question "why does Cobalt's test behavior differ that much from upstream content_browsertests? Why are we hitting this failure path and not upstream tests?"

Awallky: I filed b/467102341 to track this

PR #8316, opened 2025-12-08
ci: Add deflake workflow to retry failing test jobs

oxve:
Introduces a new GitHub Actions workflow to automatically retry failed test jobs, attempting to deflake flaky tests. This workflow is triggered upon the completion of the main CI workflows (e.g., android, linux).

The max number of attempts is configured to 3.

Bug: 461673387

### DIFF
--- a/.github/workflows/deflake_tests.yml
+++ b/.github/workflows/deflake_tests.yml
@@ -1,0 +1,44 @@
+name: Deflake Tests
+on:
+  workflow_run:
+    workflows: [android, raspi-2-modular, evergreen, linux, tvos]
+    types:
+      - completed
+permissions:
+  contents: read
+  checks: read
+
+jobs:
+  rerun-failing-jobs:
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'failure'
+    permissions:
+      actions: write
+    steps:
+      - name: Rerun Failing Tests
+        env:
+          MAX_ATTEMPTS: 3
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          WORKFLOW_NAME: ${{ github.event.workflow_run.name }}
+          WORKFLOW: ${{ github.event.workflow_run.id }}
+        shell: bash
+        run: |
+          set +x
+          # This call fails until the workflow has run $MAX_ATTEMPTS times.
+          if gh run view --repo "${REPO}" "${WORKFLOW}" --attempt "${MAX_ATTEMPTS}"; then
+            echo "Done retrying ${WORKFLOW_NAME}. It is still failing after ${MAX_ATTEMPTS} attempts."
+            exit 0
+          fi
+
+          # Fetch failed jobs containing _tests.
+          failed_job_ids=$(gh run view --repo "${REPO}" "${WORKFLOW}" --json jobs --jq '.jobs[] | select(.conclusion=="failure") | select(.name | contains("_tests")) | .databaseId')
+
+          if [ -z "${failed_job_ids}" ]; then
+            echo "No failing '_tests' jobs found to rerun."
+            exit 0
+          fi
+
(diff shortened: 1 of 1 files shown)


andrewsavage1:
I know we've avoided this before due to worries it could overload CI and that it would be better to just disable and later address flakiness. Has something changed?

Will datadog still be able to identify flakey tests if we're deflaking them in this way?

.github/workflows/deflake_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+    types:
+      - completed
+permissions: {}
+
+jobs:
+  rerun-failing-jobs:
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'failure'
+    permissions:
+      actions: write
```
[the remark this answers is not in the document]
oxve: The workflow only runs on submitted code. I've looked at the remediations and it looks to me like it should be ok to have `actions: write` on the job level, as long as the default permission is set on the workflow level.

oxve:
> I know we've avoided this before due to worries it could overload CI and that it would be better to just disable and later address flakiness. Has something changed?
> 
> Will datadog still be able to identify flakey tests if we're deflaking them in this way?

The strategy is still to disable flaky tests as they appear. Tests are not that flaky that I anticipate this to cause that much additional load. To further alleviate it I plan to add code that filters tests so only failed tests are re-run.

All attempts will be reported to DataDog so this is actually better for flakiness detection. It marks a test flaky if it fails and succeeds on the same configuration and commit.

.github/workflows/deflake_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+        env:
+          MAX_ATTEMPTS: 3
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          WORKFLOW: ${{ github.event.workflow_run.id }}
+        shell: bash
+        run: |
+          set +x
+          if gh run view ${WORKFLOW} --attempt ${MAX_ATTEMPTS}; then
+            # The last attemp was successfully loaded. Our work is done.
```
andrewsavage1: r/attemp/attempt

.github/workflows/deflake_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+    runs-on: ubuntu-latest
+    if: github.event.workflow_run.conclusion == 'failure'
+    permissions:
+      actions: write
+      checks: read
+      contents: read
+    steps:
+      - name: Rerun Failing Tests
+        env:
+          MAX_ATTEMPTS: 3
```
andrewsavage1, line 20: Where do we check that we've reached attempt number 4 and should exit, even if attempt number 3 failed?

.github/workflows/deflake_tests.yml
```diff
@@ -0,0 +1,46 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          WORKFLOW: ${{ github.event.workflow_run.id }}
+        shell: bash
+        run: |
+          set +x
+          if gh run view ${WORKFLOW} --attempt ${MAX_ATTEMPTS}; then
```
andrewsavage1: Aren't we always viewing attempt number 3 here?

.github/workflows/deflake_tests.yml
oxve: We only attempt 3 times total. MAX_ATTEMPTS includes the original attempt.

.github/workflows/deflake_tests.yml
oxve: Updated and clarified.

.github/workflows/deflake_tests.yml
oxve: Yes, and once it exists we are done trying to deflake it.

andrewsavage1 (approved):

isarkis (approved):

PR #8318, opened 2025-12-08
[tvOS] Remove speech synthesis code

rakuco:
It is unclear if this feature should be enabled at all (ATV does not, for example). Remove it in the meantime to clean up the code base and make it easier to work on the code that does get used.

Note: there are references to this code in the `accessibility_*` files in `starboard/tvos/shared`, but these need to be reworked and possibly rewritten anyway.

Bug: 461443482
Bug: 459507205


rakuco:

PR #8333, opened 2025-12-09
tvOS: Remove the platform service extension implementation

rakuco:
This implementation is not compatible with Chrobalt and needs to be rewritten in `//cobalt` and made to provide the data currently used in Kabuki (which does not include maximum FPS according to the Android implementation).

Bug: 460683731
Bug: 459507205


rakuco:

fayev-g (approved):

PR #8338, opened 2025-12-09
cobalt/test: Remove symlinks to prevent CI failures

Awallky:
This assumes some tech debt, since it affects //content. However, the cost of engineering a copy action versus this is too high, as every day the browsertests are not running in CI, the greater chance that the browsertests are broken by any CL. This has already happened once. So, the plan of action is to land this, get browsertests running in CI, then address this tech debt as tracked in b/467370154.

Bug: 459831767


Awallky:
> It'd be good to state why the existence of these symlinks breaks our CI

I added some more context to the linked bug.

> since they're all under file_chooser/, couldn't we just remove that target/data dep instead?

I think this assumes that there is some build target for file_chooser that is included in content/test/data, but unfortunately, there are no build targets or BUILD.gn files inside of content/test/data. In order to do this, we would have to make some kind of top level build, break down the directory into smaller build targets, and exclude the file_chooser directory for Cobalt builds. That is likely a huge effort.

What is simpler, but still requires some engineering work, is to make a copy action that will copy everything under //content/test/data and filter out the file_chooser directory.

PR #8339, opened 2025-12-09
android: platformError webContents null check

ckim1346:
Adding a null check for `webContents` before attempting to reload the navigation controller to prevent a java crash when the retry button of the platform error dialog is pressed. 

Bug: 466112455

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
@@ -145,6 +145,9 @@ public class PlatformError
             if (webContents != null) {
               webContents.getNavigationController().reload(true);
             }
+            else {
+              Log.e(TAG, "WebContents is null and not available to reload the application.");
+            }
             cobaltActivity.getCobaltConnectivityDetector().activeNetworkCheck();
           }
           mDialog.dismiss();


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -140,9 +141,12 @@ public void onClick(DialogInterface dialogInterface, int whichButton) {
         case RETRY_BUTTON:
           mResponse = POSITIVE;
           if (cobaltActivity != null) {
-            cobaltActivity.getActiveWebContents().getNavigationController().reload(true);
+            WebContents webContents = cobaltActivity.getActiveWebContents();
+            if (webContents != null) {
+              webContents.getNavigationController().reload(true);
```
zhongqiliang, line 146: When reload did not happen, will activeNetworkCheck() be different for retry?

Should we add an error message on the dialog when webContents is null?

johnxwork (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: the check could be different if webContents finishes initializing by then/the user connects to the network. Added a log indicating webContents is null

kaidokert (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
zhongqiliang: Does cobaltActivity.getCobaltConnectivityDetector().activeNetworkCheck(); do things differently without webContents.getNavigationController().reload(true); ?
That is should "webContents.getNavigationController().reload(true);" be in the if (webContents != null) block or not?

PR #8342, opened 2025-12-09
Play splash video from Main app's cache

linxinan-yt:
Add cache reader to h5vcc loader to extract the splash video from local storage

With it, youtube main app(www.youtube.com[1]) could cache the splash video into local storage.  At the following start,  if the cache hits, h5vcc loader will return the local stored video to media source to display.

h5vcc loader exposes two URL parameters for the callers:
- fallback: designate the webm file if cache is empty. For now only allow splash_480.webm for the low spec devices.
- cache: specify the cache name, default is "default".
Check the h5vcc_scheme_url_loader_factory_browsertest.cc for the expected behaviors.

[1] main app domain is hardcoded for now.

Test:  cobalt_browsertests_runner --gtest_filter=H5vccSchemeURLLoaderFactoryBrowserTest*
Issue: 454630524


cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -89,6 +211,75 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
+        storage_partition->GetCacheStorageControl();
+    url::Origin origin = url::Origin::Create(GURL(kSplashDomain));
+    blink::StorageKey storage_key = blink::StorageKey::CreateFirstParty(origin);
+    storage::BucketLocator bucket_locator =
+        storage::BucketLocator::ForDefaultBucket(storage_key);
+
+    cache_storage_control->AddReceiver(
+        network::CrossOriginEmbedderPolicy(), mojo::NullRemote(),
+        bucket_locator, ::storage::mojom::CacheStorageOwner::kCacheAPI,
+        cache_storage_remote_.BindNewPipeAndPassReceiver());
```
borongc, line 238: Would this bind Mojo::Remote multiple times?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -14,16 +14,38 @@
...
 #include "services/network/public/mojom/url_loader.mojom.h"
 #include "services/network/public/mojom/url_response_head.mojom.h"
+#include "third_party/blink/public/mojom/blob/blob.mojom.h"
+#include "third_party/blink/public/mojom/cache_storage/cache_storage.mojom.h"
 #include "url/gurl.h"
+#include "url/origin.h"
+
+// TODO(b/454630524): Move below constants to command args.
+namespace {
+const std::string kSplashDomain = "https://www.youtube.com";
```
borongc: Constants can be in cobalt/shell/common/shell_switches.h or cobalt/shell/common/url_constants.h.

linxinan-chops: Discussed with Orpheus we may pass kSplashPath and kSplashCacheName via the URL parameter.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Yes. Any concerns for the binding?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Done.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -14,39 +14,153 @@
...
+      if (read_result == MOJO_RESULT_SHOULD_WAIT) {
+        watcher_->Arm();
+        return;
+      }
+      if (read_result != MOJO_RESULT_OK) {
+        OnDataAvailable(read_result, state);
+        return;
+      }
+      content_.insert(content_.end(), buffer, buffer + num_bytes);
+    }
```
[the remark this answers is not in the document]
linxinan-yt: `The consumer_handle_->ReadData() method expects its third argument to be a uint32_t&`... I don't think it is correct.

See definition of ReadData at `mojo/public/cpp/system/data_pipe.h`:

```
  MojoResult ReadData(MojoReadDataFlags flags,
                      base::span<uint8_t> buffer,
                      size_t& bytes_read) const  ...
```

size_t should be the correct input type and it did not hit compile error.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -57,24 +171,44 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
+      }
+    }
+    if (content_.empty()) {
+      LOG(WARNING) << "URL: " << url_.spec() << ", host: " << key
+                   << " not found.";
     }
-    std::string content(reinterpret_cast<const char*>(file_contents.data),
-                        file_contents.size);
-    SendResponse(content, mime_type);
+    SendResponse(content_, mime_type);
```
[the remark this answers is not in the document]
linxinan-yt: Done.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,77 @@
...
+          return;
+        }
+        const mediaSource = new MediaSource();
+        mediaSource.addEventListener('sourceopen', () => {
+          const sourceBuffer = mediaSource.addSourceBuffer('video/webm; codecs="vp9"');
+          // Get cache from the html itself's URL parameter.
+          // Set the fallback suffix depending on the device spec.
+          // TODO(454630524): Pass the cache and fallback from splash.html's URL
+          // parameters.
+          fetch('h5vcc-embedded://splash.webm?cache=default')
```
linxinan-yt: @borongc, I am not an expert of mediaSource here, but need this to verify the cache implementation. We probably need follow-up PRs to get the cache name from splash.html's URL parameter.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,77 @@
...
+            .then(response => {
+              if (!response.ok) {
+                throw new Error('Network response was not ok');
```
Awallky: Is there any way we could plumb this to Java/native so we could log this somewhere? Or even better, record a histogram?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
borongc: If this binds Mojo::Remote per fetch, we should ensure it is disconnected after fetch is finished. Otherwise, it's likely this will cause memory leakage like https://github.com/youtube/cobalt/pull/8068.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -0,0 +1,173 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+#include "url/gurl.h"
+
+namespace content {
+
+class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
+ public:
+  H5vccSchemeURLLoaderFactoryBrowserTest() {
+    scoped_feature_list_.InitFromCommandLine(
+        "DisableExclusiveLockingOnDipsDatabase", "");
```
Awallky: Why are we touching the DIPS database?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -0,0 +1,173 @@
...
+  }
+  ~H5vccSchemeURLLoaderFactoryBrowserTest() override = default;
+
+  std::string GetScript(const std::string& url) {
+    return base::StringPrintf(R"(
```
Awallky, line 62: Shouldn't we share the real splash screen HTML that production uses?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -0,0 +1,173 @@
...
+          return 'Exception: ' + e.toString();
+        }
+      })();
+    )",
+                                            cache_name.c_str());
+    EXPECT_EQ("Success", EvalJs(shell(), script));
+
+    // 2. Fetch via h5vcc-embedded scheme.
+    // The loader should find the cached content from the test domain.
+    GURL splash_url("h5vcc-embedded://splash.html");
```
borongc: Consider to use `kH5vccEmbeddedScheme` in cobalt/shell/common/url_constants.h.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -0,0 +1,173 @@
...
+#include "content/public/test/browser_test.h"
+#include "content/public/test/browser_test_utils.h"
+#include "net/test/embedded_test_server/embedded_test_server.h"
+#include "testing/gtest/include/gtest/gtest.h"
+#include "url/gurl.h"
+
+namespace content {
+
+class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
+ public:
```
Awallky, line 35: Could you add SetUp and TearDown methods similar to [this](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/serial/chrome_serial_browsertest.cc;l=192-210?q=lang:cc%20teardown%20f:browsertest%20f:chrome&ss=chromium)? We would want to clear the `global_splash_domain_test_` for each test and maybe `resource_map_test_`.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Nice catch. I will reset it. Plus, once the splash is played, the Webcontents will be removed, so we should be safe right?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Nice catch. I got this sometimes but later I found this is because the tests are using the same user_data_dir. This is fixed in the test runner script. So I don't need it. Will remove.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Nice catch.

linxinan-yt: Done.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Done.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: You are right, I should use the production JS in the HTML. Done.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Removed.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Add reset in the disconnect method.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: I am not sure, leave it to @borongc. Is it possible?

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,79 @@
...
+        mediaSource.addEventListener('sourceopen', () => {
+          const sourceBuffer = mediaSource.addSourceBuffer('video/webm; codecs="vp9"');
+          // Get cache from the html itself's URL parameter.
+          // TODO(463395742): Set the fallback suffix depending on the device spec.
+          const params = new URLSearchParams(window.location.search);
+          if (!params.has('cache')) {
+            params.append('cache', 'default');
+          }
+          fetch('h5vcc-embedded://splash.webm?' + params.toString())
+            .then(response => {
```
Awallky: nit: using async/await could improve readability.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,79 @@
...
...Aien+EkEdl3HZXlsFR4AAAAO8xtccbfpym3n6UYg6X5j+36cGILpQA8+p76sQG7cbnmF0GVfQNa3KBEdS6CYlAMKxLJD2t0CwcI8oXXTn9FUng4QnMhcdCX/zcfTmuRgFv7OYJVq+LaDjZmZJNrHJ+izgzgjmwB3iKwAEk7YBWwbPsiU6vntMLAxaXaxfzKv/qV+gSHB/LRnitcgJ/0YeKiMUrzG5UzcEXJDfF26Yq1LWJlSvI8JyoHfzEMkhrRtEI3Mp62AAKnmXGv+jbiKwrFzaJlTv8nIgLMCEN2zKsEqbjXl6KiS2fb4J6UVuYCojWsVEu3elr7q6P4ZRzJ7UYgDICZOHBXwCjQgyBAU0AhgBBDvFRMAYABWC1qsAAAAAABjGwY6mr/gAAAHsyA4hTUSf0yVJ9+6j9yPEvYZj3nU0wGmpP2iPyU9fb9qWhsWvZE2vEOSmaH0ViXO3YmYXkY3H54aw39aNBi/LsWK28ASVpaUyIgFP3SU+CD9t/NSzOzdbLAomA+LXumCS4JTzqfLUAYxIPxGIPk0az1XkIYHbg0YvauRAAAAFlMbV/pliozeFZVsd86vBVHxPSfY4kqlLGK77gDjw92okjL/xnewI5VtMOMx/UBFnCCCckIV1suMKd09DwAChLUI939tI7Vh5YDLJ7SyWhhyoBmA7iuu2h6UWNNM7g9qylHRoGLB4KmxF5Na0bkxwPUBTNi2BqZ4UMR2kSAW7a8qpZXFmYL4ZmmMEC2gjTnpwu+VjvcsajCeJnI5KqEwnZYGsTJQkOgpMSOQiaZf5/XcvV1Yz7+W4LfPKJYiBxLXF2jQxxrTIWY+Lh10pyaQHQlWbuwhm4KaNUShOomqadvsvVCzVV1NHDLRXOyo95h1sLgamxC+zDUhuxMapNMc9qzk1IQyITd5vtO+bGMTUbYE2X/vde0+fJsBYtHYS018PH3I45NxjxmSJHQObDqBX3L4yrJ6lTByAlVW4/KJaEFz6uVoEVKkiG7s8vqg6D3tOSzCP6mD/MQoMWXysr91V1EQKwYh4AMgJk4cFfAKDBobmBAW4AhgCBCpwATAGAAoBg+JvbeVT+1AAAAAAABXbLGlzgAAAABXbFVCOgAAAABXbLGlzgdsVUI6CbgRH7gd8cU7tro7uPs4EAt4r3gQHxggGo8IEDu5CzggPHt4r3gQHxgh4w8IEE';
```
Awallky, line 21: This probably better maintained if we export it to a standalone JS file instead of putting it all in a script tag.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: You are right. Let's follow G3's best practices.

Done.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: I agree, but this HTML should be a simple and small file.
Eventually we should migrate it to a native implementation, e.g. call the player from C++. 
Given its size and simplicity, I think it is ok to leave the JS here. @borongc WDYT?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -14,40 +14,156 @@
...
+        }
+        std::move(callback_).Run(std::move(content_));
+      }
+      return;
+    }
+
+    while (true) {
+      uint8_t buffer[kReadBufferSize];
+      size_t num_bytes = sizeof(buffer);
+      MojoResult read_result = consumer_handle_->ReadData(
```
borongc, line 128: m138 has different `ReadData()` from m114. Consider to make it `writable_span()` as what m138 does. E.g., https://chromium-review.googlesource.com/c/chromium/src/+/5619056.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -14,40 +14,156 @@
...
+      if (callback_) {
+        std::move(callback_).Run({});
+      }
+    }
+  }
+
+ private:
+  void OnDataAvailable(MojoResult result,
+                       const mojo::HandleSignalsState& state) {
+    constexpr uint32_t kReadBufferSize = 256;
```
borongc: Is there any reason this is 256? Why not larger?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -87,7 +225,89 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
+    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE, base::BindOnce(&H5vccSchemeURLLoader::DisconnectCacheStorage,
+                                  weak_factory_.GetWeakPtr()));
+    if (!result->is_response()) {
+      LOG(ERROR) << "Failed to match cache for splash video"
+                 << ", error: " << result->get_status();
+      SendResponse(content_, mime_type);
+      return;
+    }
+    LOG(INFO) << "Found splash video in cache.";
```
borongc: nit: consider to pass meaningful argument (like url) and log it here, so we can know which cache is found.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_unittest.cc
```diff
@@ -122,7 +122,8 @@ const char kTestHtmlContent[] = "<html><body><h1>Test</h1></body></html>";
 class H5vccSchemeURLLoaderFactoryTest : public testing::Test {
  public:
   H5vccSchemeURLLoaderFactoryTest()
-      : task_environment_(base::test::TaskEnvironment::MainThreadType::IO) {}
+      : task_environment_(base::test::TaskEnvironment::MainThreadType::IO),
+        factory_(nullptr) {}
```
borongc, line 126: Why this needs to be nullptr here?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.h
```diff
@@ -59,10 +63,16 @@ class H5vccSchemeURLLoaderFactory final
       override;
 
   // Testing seam to inject a resource map.
-  void SetResourceMapForTesting(const GeneratedResourceMap* resource_map_test);
+  static void SetResourceMapForTesting(
```
borongc, line 66: Make this to static is for browser test?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Thanks for pointing it. The splash screen is usually around 1MB size. 64KB should suffice while safe for low spec devices.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Thanks for digging it. Learnt a good lesson here.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.h
linxinan-yt: I think it makes sense to cover it in browser tests later. But Adam's comment makes sense, we should Setup/Teardown resource_map_test_ and global_splash_domain_test_ for each test case to avoid flakiness in the infra.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_unittest.cc
linxinan-yt: See reply in the comment for SetResourceMapForTesting.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Done.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_unittest.cc
linxinan-yt: Move the thread to the below one.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Added the cache name. Note domain is hard coded so it does not provide much value here.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
kaidokert: I think one could argue that there should be a `.gn` rule that builds the final version with embedded string. E.g something that invokes a python script that runs a Jinja template and substitutes an encoded version of the asset. 

At the same time it's one more thing that can go wrong - simple is good for now. I'd maybe make a follow-up ticket for considering a build rule for it.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,77 @@
...
+                mediaSource.endOfStream();
+                videoEl.play();
+                afterVideoStart && afterVideoStart();
+              }
+            });
+            sourceBuffer.appendBuffer(data);
+          } catch (error) {
+            console.error('Failed to fetch splash video:', error);
+          }
+        });
```
kaidokert: consider adding a `, { once: true });` safeguard here, to prevent multiple events causing a bad state.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,77 @@
...
+              throw new Error('Network response was not ok');
+            }
+            const data = await response.arrayBuffer();
+            sourceBuffer.addEventListener('updateend', () => {
+              if (!sourceBuffer.updating && mediaSource.readyState === 'open') {
+                mediaSource.endOfStream();
+                videoEl.play();
+                afterVideoStart && afterVideoStart();
+              }
+            });
```
kaidokert: consider adding `,{once: true}` here as well, to ensure `updateend` gets processed only once.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,77 @@
...
+          const sourceBuffer = mediaSource.addSourceBuffer('video/webm; codecs="vp9"');
+          // Get cache from the html itself's URL parameter.
+          // TODO(463395742): Set the fallback suffix depending on the device spec.
+          const params = new URLSearchParams(window.location.search);
+          if (!params.has('cache')) {
+            params.append('cache', 'default');
+          }
+          try {
+            const response = await fetch(
+                'h5vcc-embedded://splash.webm?' + params.toString());
```
borongc, line 74: We should check if 4k is supported:
```
MediaSource.isTypeSupported("video/webm; codecs=\"vp9\"; width=3840; height=2160;")
```
use default resolution. Otherwise, it should be 480p.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: Gotcha. Thanks.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: Done.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
borongc, line 74: Also, it only plays vp9 videos. We should open a ticket and add it as TODO for showing static image or css animation as a comment.

linxinan-yt: Done.

linxinan-yt: Track it in b/458074360. Added a TODO here.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: You mean we build the webm file into the html via some gn rule? It is possible, but if our ultimate goal is to get rid of html, why not just leave the webm files in the embedded folder?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -57,24 +174,46 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
       LoaderEmbeddedResources::GenerateMap(resource_map);
     }
 
-    if (resource_map.find(key) == resource_map.end()) {
-      LOG(WARNING) << "URL: " << url_.spec() << ", host: " << key
-                   << " not found.";
-      SendResponse("Resource not found", "text/plain", net::HTTP_NOT_FOUND);
-      return;
+    // Specify the built-in video if the cache is unavailable.
+    // Only for webm.
```
borongc, line 178: nit: add TODO for adding static image or css animation in the comment, as this need to be revisited later.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -57,24 +174,46 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
+    if (resource_map.find(resource_key) != resource_map.end()) {
+      FileContents file_contents = resource_map[resource_key];
+      content_ = std::string(reinterpret_cast<const char*>(file_contents.data),
+                             file_contents.size);
+    }
 
+    std::string mime_type = "application/octet-stream";
     if (base::EndsWith(key, ".html", base::CompareCase::SENSITIVE)) {
-      mime_type = "text/html";
+      if (content_.empty()) {
```
borongc: Consider to use a helper for L194--L197 & L210--L213.

cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -685,8 +683,8 @@ ShellContentBrowserClient::GetShellContentBrowserClientInstances() {
 void ShellContentBrowserClient::RegisterH5vccScheme(
     NonNetworkURLLoaderFactoryMap* factories) {
   if (!h5vcc_scheme_url_loader_factory_) {
-    LOG(WARNING) << "h5vcc_scheme_url_loader_factory_ is not initialized!";
-    return;
+    h5vcc_scheme_url_loader_factory_ =
+        std::make_unique<H5vccSchemeURLLoaderFactory>(browser_context());
```
borongc: Do we ensure we will have `browser_context()` here? This is moved from L259--L260 to here.

If we need to ensure `browser_context()` is not nullptr, we should consider add CHECK here.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -18,80 +17,79 @@
...
+        const mediaSource = new MediaSource();
+        mediaSource.addEventListener('sourceopen', async () => {
+          const sourceBuffer = mediaSource.addSourceBuffer('video/webm; codecs="vp9"');
+          const params = new URLSearchParams(window.location.search);
+          if (!params.has('cache')) {
+            params.append('cache', 'default');
+          }
+          // TODO(458074360): fallback to static image if the device does not support VP9.
+          if (!MediaSource.isTypeSupported("video/webm; codecs=\"vp9\"; width=3840; height=2160;")) {
+            params.append('fallback', 'splash_480.webm');
```
borongc, line 70: As this is not specified in the code, we should note it that the default filename is `splash.webm` as L71 below, appending `fallback` here will allow it to find other files from either built-in or cache files.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Done.

cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -230,7 +230,14 @@ test("cobalt_browsertests") {
   }
 
   if (is_starboard) {
-    sources += [ "cobalt_browser_tests_main.cc" ]
+    sources += [
+      "cobalt_browser_tests_main.cc",
+
+      # Media source and Fetch are not supported yet in android's browser tests.
```
borongc: This looks like a TODO, and we should have a bug id listed here.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: ummm, seems the logic has some duplication. Here I only need to change the mime_type and let the block line205 to handle the SendResponse.

cobalt/shell/browser/shell_content_browser_client.cc
linxinan-yt: I can add a CHECK, but I think browser_context() is guaranteed. 

See cobalt/shell/browser/shell_browser_main_parts.cc:

```
int ShellBrowserMainParts::PreMainMessageLoopRun() {
  InitializeBrowserContexts();  --> create new ShellBrowserContext
  Shell::Initialize(CreateShellPlatformDelegate());
  net::NetModule::SetResourceProvider(PlatformResourceProvider);
  ShellDevToolsManagerDelegate::StartHttpHandler(browser_context_.get());
  InitializeMessageLoopContext(); --> Shell::CreateNewWindow the splash screen
  return 0;
}
```

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: Make sense.

cobalt/testing/browser_tests/BUILD.gn
linxinan-yt: Done.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: Already impl-ed.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: Resolve it for to unblock the PR. We can continue the discussion in the tracking bug.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: Resolve for now to unblock the PR. We can continue the discussion in the tracking bug. AFAICT, the ultimate goal is to get rid of the html and play video directly. But we are not there yet.

PR #8347, opened 2025-12-09
Refactor Android manifests for browser tests

Awallky:
This patch refactors the Android manifest files for browser tests to reuse the production manifest with test-specific additions. This is achieved by using Jinja templates and inheritance.

The main changes are:
- `cobalt/shell/android/javatests/AndroidManifest.xml` now serves as a base for test manifests.
- `cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2` is modified to be includable and defines blocks for customization.
- `cobalt/testing/browser_tests/BUILD.gn` is updated to use the new Jinja template structure.
- The standalone `cobalt/testing/browser_tests/browsertests_apk/AndroidManifest.xml.jinja2` is deleted as it's no longer needed.

This change simplifies manifest management and reduces duplication.

Bug: 459831767

### DIFF
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -13,7 +13,7 @@ jinja_template("cobalt_manifest") {
     "manifest_package=dev.cobalt.coat",
     "build_type=$build_type",
   ]
-  if (is_debug) {
+  if (is_official_build) {
     variables += [ "debuggable=true" ]
   } else {
     variables += [ "debuggable=false" ]


cobalt/android/BUILD.gn
```diff
@@ -13,6 +13,11 @@ jinja_template("cobalt_manifest") {
     "manifest_package=dev.cobalt.coat",
     "build_type=$build_type",
   ]
+  if (is_debug) {
+    variables += [ "debuggable=true" ]
```
haozheng-cobalt, line 17: This is great improvement!  I just have one suggestion, we should use the `is_official_build` variable in stead of `is_debug` if we would like to  support browser tests in both `debug` and `qa`.

haozheng-cobalt (changes requested):

haozheng-cobalt (approved):
Good to go except the single comment.

cobalt/android/BUILD.gn
Awallky: Done.

PR #8359, opened 2025-12-10
Migrate evergreen snapshot packaging logic

briantting:
Brings packaging logic to Cobalt side. Increases flexability and decreases brittleness.

Bug: 445996864

### DIFF
--- a/cobalt/build/evergreen-arm-hardfp-raspi/package.json
+++ b/cobalt/build/evergreen-arm-hardfp-raspi/package.json
@@ -1,21 +1,40 @@
 {
     "archive_datas": [
         {
+            "tar_path": [
+                {
+                    "from_path": "lib.unstripped/libcobalt.so",
+                    "to_tar": "symbols/libcobalt.so.tar.gz"
+                }
+            ],
             "files": [
+                "app/cobalt/lib/libcobalt.lz4",
                 "cobalt_shell.pak",
-                "libcobalt.so",
-                "loader_app",
+                "fonts/fonts.xml",
                 "gen/build_info.json",
                 "gen/licenses_cobalt.spdx",
                 "gen/licenses_cobalt.txt",
-                "lib.unstripped/libcobalt.so",
-                "native_target/crashpad_handler"
-            ],
-            "dirs": [
-                "app",
-                "fonts"
+                "libcobalt.so",
+                "loader_app",
+                "native_target/crashpad_handler",
+                {
+                    "from_file": "cobalt_shell.pak",
+                    "to_file": "content/cobalt_shell.pak"
+                },
+                {
+                    "from_file": "libcobalt.so",
+                    "to_file": "lib/libcobalt.so"
+                }
             ],
             "rename_files": [
+                {
+                    "from_file": "app/cobalt/lib/libcobalt.lz4",
+                    "to_file": "lib/compressed/libcobalt.lz4"
+                },
+                {
+                    "from_file": "fonts/fonts.xml",
+                    "to_file": "content/fonts/fonts.xml"
+                },
                 {
                     "from_file": "gen/build_info.json",
                     "to_file": "build_info.json"
@@ -27,10 +46,14 @@
                 {
                     "from_file": "gen/licenses_cobalt.txt",
                     "to_file": "licenses_cobalt.txt"
-                },
+                }
+            ],
+            "dirs": [
+                "app",
+                "fonts",
                 {
-                    "from_file": "lib.unstripped/libcobalt.so",
-                    "to_file": "symbols/libcobalt.so"
+                    "from_dir": "app/cobalt/content/ssl",
+                    "to_dir": "content/ssl"
                 }
             ],
             "archive_type": "ARCHIVE_TYPE_ZIP"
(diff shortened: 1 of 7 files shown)


briantting:
@andrewsavage1 do we need the entire `app` and `font` directories?

cobalt/build/evergreen-arm-hardfp-raspi/package.json
```diff
@@ -29,8 +44,18 @@
...
                 },
                 {
-                    "from_file": "lib.unstripped/libcobalt.so",
-                    "to_file": "symbols/libcobalt.so"
+                    "from_file": "libcobalt.so",
+                    "to_file": "lib/libcobalt.so"
+                }
+            ],
+            "dirs": [
+                "app/cobalt/content/ssl",
```
andrewsavage1: It's probably a good idea to take the full `app/cobalt/content` dir here

cobalt/build/evergreen-arm-hardfp-raspi/package.json
```diff
@@ -1,21 +1,36 @@
...
-                "fonts"
-            ],
             "rename_files": [
+                {
+                    "from_file": "app/cobalt/lib/libcobalt.lz4",
+                    "to_file": "lib/compressed/libcobalt.lz4"
+                },
+                {
+                    "from_file": "cobalt_shell.pak",
+                    "to_file": "content/cobalt_shell.pak"
```
andrewsavage1: I'm worried about moving these files—the linux implementation to get the path to these files will assume they are adjacent to the loader_app binary in c26, not in a content subdirectory, so if anywhere down the line we need to run cobalt after this packaging I imagine it will have issues

andrewsavage1 (changes requested):

cobalt/build/evergreen-arm-hardfp-raspi/package.json
briantting: we previously avoided doing this specifically so that we don't copy over all of `app/cobalt/content/fonts` (we do just fonts.xml separately) and `app/cobalt/content/test`

if that's changed I don't mind copying over all of content, otherwise ssl is the only directory in content at the moment

cobalt/build/evergreen-arm-hardfp-raspi/package.json
briantting: just cobalt_shell.pak or also libcobalt.so? i can duplicate them instead of moving them to accommodate

cobalt/build/evergreen-arm-hardfp-raspi/package.json
andrewsavage1: If more content is added in the future to this directory (which is possible), this is going to break. Can we just exclude app/cobalt/content/test?

cobalt/build/evergreen-arm-hardfp-raspi/package.json
andrewsavage1: Both of them—why do we need to have them in a content subdir in the first place?

briantting: we're trying to offload all the logic from the snapshot.py google3 scripts. this way if there are changes Cobalt can control them all on the package side rather than wait for a multistep change. the snapshot and future rapid executor should just be as simple of a directory copy as possible.

cobalt/build/evergreen-arm-hardfp-raspi/package.json
briantting: it looks like doing it that way will be messy, there's also individual files like `app/cobalt/content/cobalt_shell.pak` that we don't want to copy / move. since we're building up this package directory, i think an additive strategy is fine vs an exclude strategy.

breakages are "fine" because now all the packaging is done Cobalt side so changes can be easily understood / made

andrewsavage1: The loader_app expects the same relative location of files (app, fonts, libcobalt.lz4) as they are in the out directory (so yes, the app and fonts dirs should be adjacent to the loader_app binary)

briantting: okay that should be handled now

cobalt/build/evergreen-arm-hardfp-raspi/package.json
andrewsavage1: It makes sense to control the packaging from the cobalt side, but I'm confused how that relates to putting these files in a content/ subdirectory? Are there some assumptions on the g3 side about that directory structure?

briantting: The g3 side used to expect exact files and directories and would manipulate them in a very brittle manner. Now the g3 side simply expects certain broader directories (content, lib, symbols) and copies them in their entirety without modification.

So before any change on the cobalt side would break snapshotting, now cobalt just change packaging however it wants knowing that whatever is in those certain folders will be snapshotted just like that.

Android was changed to reflect this new paradigm, this is updating it for Evergreen.

briantting:
https://paste.googleplex.com/5002210276671488

sample package layout

andrewsavage1 (approved):
Thanks Brian, filed b/468051586 for myself to follow up on

PR #8360, opened 2025-12-10
CI: Add YTS test support

isarkis:
This change enables the execution of YTS tests for Kirkwood, Boreal,
Sabrina, and Shields platforms. It refactors the existing E2E test
GitHub Action into a more generic internal tests action, unifying
support for both E2E and YTS test types.

Bug: 454941148

### DIFF
--- a/.github/actions/internal_tests/action.yaml
+++ b/.github/actions/internal_tests/action.yaml
@@ -1,18 +1,18 @@
-name: E2E Tests
-description: Runs e2e tests.
+name: Internal Tests
+description: Runs internal tests.
 inputs:
   test_dimensions:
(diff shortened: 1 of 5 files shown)


.github/actions/internal_tests/action.yaml
oxve, line 1: Nice refactoring. Do you think we can bring in the unit tests here too?

oxve (approved):

.github/actions/internal_tests/action.yaml
isarkis: I think I tried some time ago and ran into difficulties, I don't remember what the problem was.  We can try again in a separate PR.

PR #8365, opened 2025-12-11
Remove cobalt/shell/app/shell_main.cc

rakuco:
This file is used in Content Shell, but not in Cobalt, where the entry point is actually in `cobalt/app` instead.

Fixed: 467905838


rakuco:

PR #8366, opened 2025-12-11
tvOS: Remove most UI-related code from starboard/tvos/shared

rakuco:
This was essentially being kept for reference, as the `SbWindow*` APIs are not being used now that the UI is handled in Cobalt Shell.

As a consequence, `SBDStarboardApplication.windowManager` did not create or manipulate `SBDApplicationWindow` objects, so `application_window.*` and `application_view.*` were not being used at all.

Removing them also allows us to get rid of the input handling code that injected events into the Starboard event loop, which also is not used in main.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #8367, opened 2025-12-11
starboard: Cherry pick PR #8053: Fix DNS lookups for IPv6 addresses.

jellefoks:
Refer to the original PR: https://github.com/youtube/cobalt/pull/8053

The Starboard layer wrapper for getaddrinfo() made incorrect copies of
the sockaddr structure for IPv6 addresses. This issue prevented proper
resolution and use of IPv6 addresses for DNS lookups.

This change modifies the wrapper to ensure that the entire sockaddr
structure, including the address data, is correctly copied, thus
resolving the IPv6 address lookup failures. Additionally, extensive
tests have been added and updated to thoroughly verify both IPv4 and
IPv6 address resolution for known external hosts and localhost.

Bug: 460175900
Bug: 438858995
Bug: 409716671
Bug: 460594994
Bug: 439002238
Bug: 405988446

### DIFF
--- a/starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
@@ -17,12 +17,18 @@
 #include <stdlib.h>
 #include <string.h>
 
+#include <algorithm>
 #include <iterator>
 #include <memory>
 
 #include "starboard/common/log.h"
 
 namespace {
+
+// Ensure that |struct musl_sockaddr| is large enough to hold either
+// |struct sockaddr_in| or |struct sockaddr_in6| on this platform.
+static_assert(sizeof(struct musl_sockaddr) >= sizeof(struct sockaddr_in));
+static_assert(sizeof(struct musl_sockaddr) >= sizeof(struct sockaddr_in6));
 
 struct platform_iov_deleter {
   void operator()(struct iovec* ptr) const { free(ptr); }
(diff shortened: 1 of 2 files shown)


starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
```diff
@@ -329,18 +335,18 @@ SB_EXPORT int __abi_wrap_getaddrinfo(const char* node,
         free(musl_ai);
         return -1;
       }
-      musl_ai->ai_addrlen = ai_copy.ai_addrlen;
-      musl_ai->ai_addr =
-          (struct musl_sockaddr*)calloc(1, sizeof(struct musl_sockaddr));
+      musl_ai->ai_addrlen =
+          std::min(static_cast<uint32_t>(ai_copy.ai_addrlen),
```
yell0wd0g, line 339: Shouldn't this be `std::max` to make sure `ai_addrlen` can fit
either of those two lengths?

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
```diff
@@ -329,18 +335,18 @@ SB_EXPORT int __abi_wrap_getaddrinfo(const char* node,
...
+                   static_cast<uint32_t>(sizeof(struct musl_sockaddr)));
+      musl_ai->ai_addr = (struct musl_sockaddr*)calloc(1, musl_ai->ai_addrlen);
```
yell0wd0g, line 341: Seeing that Gemini hasn't complained I guess this idiom is fine,
but at first glance it looked like a leak? I was going to suggest
a comment as to why this is fine but seems pervasive in the 
file?

Also shouldn't we prefer to use `new` and/or `unique_ptr<>` ?

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
jellefoks: This is an `std::min()` because we can't set ai_addrlen longer than the `struct musl_sockaddr` or the user could go out of bounds reading the address. `struct musl_sockaddr` is guaranteed large enough to old either `struct sockaddr_in` or `struct sockaddr_in6` in our existing definition in `starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.h`.

That said, re-reading the type definitions, POSIX does not specify the type size of `sa_family_t`, which is the first member of`sockaddr_in` and `sockaddr_in6`, making them non-ABI, so this does need do be changed to use a (new) `musl_sockaddr_in` and `musl_sockaddr_in6` and use a per-member copy instead of the `memcpy` below.

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
jellefoks: `freeaddrinfo` is the function that contains the counterpart of this allocation (which is wrapped in our `starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc`)

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
yell0wd0g: Gotcha thanks for the explanation!

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
yell0wd0g: Makes sense thanks!

I guess the whole Starboard API is based on the client caller making
sure that the memory allocated by Starboard is eventually released
via a compensating call, right? So it's OK to separate the individual
allocate and release calls.

jellefoks: Yes, that's how `getaddrinfo` is specified, as a pair with `freeaddrinfo`. This is done to allow the details of this allocation to remain opaque to the caller.  The interface appears fragile in current coding standards because this is an old plain C interface with a variable length of data being returned.

To make the data structures that we return to the Cobalt caller ABI compatible, we have copy the data structures returned by the platform into the musl equivalents.

Our wrapper for `freeaddrinfo` is right below this function in this file, and it calls `free` where needed (the `addrinfo` entries, and their members `ai_addr` and `ai_cannonname`). Additionally, we call `freeaddrinfo` on the platform `addrinfo`  (in `native_res`) to free up the allocations made by the platform.

yell0wd0g (approved):
Whoops didn't see my +1 was needed. Doing it now

PR #8374, opened 2025-12-11
Add cobalt-insecure-private-range CSP directive

arjungm:
Re-introduce Content Security Policy source expression, 'cobalt-insecure-private-range'. This Cobalt-specific directive allows websocket connections to IP addresses that fall within standard private network ranges, as defined by RFC 1918 for IPv4 and ULA for IPv6. This allows Cobalt to be used in various internal E2E testing suites such as Vega or YTS.

Previously in Cobalt 25, we had implementations for `cobalt-insecure-local-network` and `cobalt-insecure-private-range`. This PR makes sure Cobalt 26+ still retains this feature.

Bug: 460275338
Bug: 389723405

### DIFF
--- a/third_party/blink/renderer/core/frame/csp/local_ip.cc
+++ b/third_party/blink/renderer/core/frame/csp/local_ip.cc
@@ -12,6 +12,9 @@
 
 namespace {
 
+// TODO(b/470105792): Refactor the implementation of the private range IP checks
+// to avoid string manipulation and bitwise operations.
+
 std::string SanitizeIPAddressStr(const std::string& ip_str) {
   std::string clean_ip = ip_str;
   size_t start = clean_ip.find('[');


third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -51,4 +66,56 @@ bool IsIPInLocalNetwork(const std::string& target_ip_str) {
...
+    // Check for the 2002::/16 prefix
+    // The first two bytes must be 0x20 and 0x02
+    if (ipv6_addr.s6_addr[0] != 0x20 || ipv6_addr.s6_addr[1] != 0x02) {
+      // RFC 4193: fc00::/7 (The first 7 bits must be 1111110)
+      // This covers both fc00::/8 and fd00::/8
+      // We check the first byte: (byte & 11111110) == 11111100
+      return (ipv6_addr.s6_addr[0] & 0xFE) == 0xFC;
+    }
+
+    // Treat ipv6_addr as ipv4.
```
johnxwork: Just curious: why is this needed?

Also FYI there's a similar function in C25: https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/network/local_network.cc;drc=181dea793fdbcac4ac072600e5d351e1e0849d00;l=69

johnxwork (approved):

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: IPv6 addresses with `2002::` prefix are 6to4 addresses, so they can be converted to their corresponding IPv4 address. In that case it'll check against RFC 1918 instead.

third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -51,4 +66,56 @@ bool IsIPInLocalNetwork(const std::string& target_ip_str) {
...
+    // Extract the 4 bytes following the prefix (indices 2, 3, 4, 5)
+    uint32_t ipv4_network_order;
+    std::memcpy(&ipv4_network_order, &ipv6_addr.s6_addr[2], sizeof(uint32_t));
+
+    // Convert from Network Byte Order to Host Byte Order (uint32_t)
+    return ntohl(ipv4_network_order);
```
arjungm: Oops I didn't actually catch this error here. I'll fix this and rebase and hopefully it still all works!

arjungm: Fixed now

yell0wd0g (commented):
@jellefoks can you PTAL? 

I thought this PR was going to be smaller but it's not and
it touches upstream code as well :/

third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -51,4 +90,46 @@ bool IsIPInLocalNetwork(const std::string& target_ip_str) {
     return is_local;
 }
 
+bool IsIPInPrivateRange(const std::string& raw_ip_str) {
```
yell0wd0g, line 96: This file has a lot of manual string manipulation - could we/did we
consider using [IPAddr](https://source.chromium.org/chromium/chromium/src/+/main:net/base/ip_address.h;l=163;drc=89424e3fec685efa12c7a6f598dbf3e9ebfbcf4d) or logic under e.g.
third_party/blink/renderer/platform/network/network_utils.cc ?

I understand this CL is a port of existing code -in that case
leaving it mostly untouched makes sense - can we add a TODO+bug
to see if this code can be refactored then? (Possibly using 
Gemini or some other AI tool).

yell0wd0g (approved):
Assuming this is a forward porting RS LGTM with
the comment below.

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: I'll refactor it now. I'm planning on cherrypicking this to main and 26android so I would save time on subsequent refactors.

Also that looks like it [does the trick](https://source.chromium.org/chromium/chromium/src/+/main:net/base/ip_address.cc;l=113-117;drc=89424e3fec685efa12c7a6f598dbf3e9ebfbcf4d;bpv=0;bpt=1) exactly, so nice find.

arjungm: Nevermind, I'll let this submit and we can find a way to refactor this in a follow-up.

Bugs: 470105792, 470104197

PR #8375, opened 2025-12-11
Skip Chromium precommit checks on files that are Cobalt specific.

SortaCookie:
Bug: 435503470


.pre-commit-config.yaml
```diff
@@ -43,6 +43,11 @@ repos:
         entry: ./third_party/depot_tools/vpython3 --vpython-spec=./third_party/depot_tools/.vpython3 ./cobalt/precommit/run_chromium_presubmits.py
         language: system
         types: [file]
+        exclude: |
+            (?x)^(
+                starboard/|
```
borongc: Should this be something like: 

```
(.*/)?cobalt/.*|
(.*/)?starboard/.*|
```

We have many customization codes under `//xxx/cobalt/` or `//xxx/starboard` that is not from Chromium.

borongc (approved):

PR #8384, opened 2025-12-12
Disable devtools for gold builds.

ghost:
Bug: 442625426

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -174,6 +174,15 @@ CobaltContentBrowserClient::CreateBrowserMainParts(
   return browser_main_parts;
 }
 
+std::unique_ptr<content::DevToolsManagerDelegate>
+CobaltContentBrowserClient::CreateDevToolsManagerDelegate() {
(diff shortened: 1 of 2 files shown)


andrewsavage1 (approved):
Seems good!

yell0wd0g (commented):
Can you replace the current succinct commit message with
the 🤖 Gemini Suggested Commit Message? 

A question: does this mean there will be no way to 
connect a remote DevTools to a Gold build, regardless
of the `--remote-allow-origins=*` flag? Or would this
only remove the DevTools shipped with the browser
client? It would be good to clarify in the commit
message the extent of the removal.

ghost:
@yell0wd0g My comment seems sufficient. Feel free to edit as you wish if you would prefer different wording. Based on what I read in the bug and associated docs, we want to keep the devtools in the compilation target, but disable the ability to use devtools for release builds. This change does that. If you have a better idea of what needs to happen, you are welcome to take over the bug.

ghost:
I'm not sure how the failed checks are related to my change. I'll rebase, retry, and see what happens.

andrewsavage1:
This should be cherry picked into 26.eap

PR #8392, opened 2025-12-12
android: Make QUIC disabling configurable

zhongqiliang:
Move the --disable-quic command line argument from a hardcoded
default in CommandLineOverrideHelper to a conditional check in
CobaltActivity. This allows downstream applications to control
QUIC enablement via feature flags through the getFeatureFlags()
extension point.

Bug: 467778593

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -45,6 +45,7 @@ import dev.cobalt.media.VideoSurfaceView;
 import dev.cobalt.shell.Shell;
 import dev.cobalt.shell.ShellManager;
 import dev.cobalt.util.DisplayUtil;
+import dev.cobalt.util.JavaSwitches;
 import dev.cobalt.util.Log;
 import dev.cobalt.util.UsedByNative;
 import java.util.ArrayList;
@@ -81,10 +82,6 @@ public abstract class CobaltActivity extends Activity {
   private static final String META_DATA_APP_SPLASH_TIMEOUT_MS = "cobalt.APP_SPLASH_TIMEOUT_MS";
   private static final String DISABLE_NATIVE_SPLASH = "disable-native-splash";
 
-
-  // Features
-  public static final String FEATURE_ENABLE_QUIC = "EnableQUIC";
-
   // This key differs in naming format for legacy reasons
   public static final String COMMAND_LINE_ARGS_KEY = "commandLineArgs";
 
@@ -93,7 +90,7 @@ public abstract class CobaltActivity extends Activity {
   // Maintain the list of JavaScript-exposed objects as a member variable
   // to prevent them from being garbage collected prematurely.
   private List<CobaltJavaScriptAndroidObject> javaScriptAndroidObjectList = new ArrayList<>();
-  private Map<String, String> featureFlags = new HashMap<>();
+  private Map<String, String> javaSwitches = new HashMap<>();
 
   @SuppressWarnings("unused")
   private CobaltA11yHelper a11yHelper;
@@ -132,9 +129,9 @@ public abstract class CobaltActivity extends Activity {
         commandLineArgs = getCommandLineParamsFromIntent(getIntent(), COMMAND_LINE_ARGS_KEY);
       }
 
-      Map<String, String> featureFlags = getFeatureFlags();
+      Map<String, String> javaSwitches = getJavaSwitches();
       List<String> extraCommandLineArgs = new ArrayList<>();
-      if (!featureFlags.containsKey(FEATURE_ENABLE_QUIC)) {
+      if (!javaSwitches.containsKey(JavaSwitches.ENABLE_QUIC)) {
         extraCommandLineArgs.add("--disable-quic");
       }
 
@@ -671,9 +668,11 @@ public abstract class CobaltActivity extends Activity {
         .orElse(null);
   }
 
-  // Overrided by Kimono
-  protected Map<String, String> getFeatureFlags() {
-    return this.featureFlags;
+  /** 
+   * Overridden by Kimono to provide specific Java switch configurations.  
+   */
+  protected Map<String, String> getJavaSwitches() {
+    return this.javaSwitches;
   }
 
   /**
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -117,6 +117,7 @@ android_library("cobalt_main_java") {
     "apk/app/src/main/java/dev/cobalt/util/DisplayUtil.java",
     "apk/app/src/main/java/dev/cobalt/util/Holder.java",
     "apk/app/src/main/java/dev/cobalt/util/IsEmulator.java",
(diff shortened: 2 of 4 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -66,9 +66,6 @@ public static List<String> getDefaultCommandLineOverridesList() {
         // causes rendering artifacts when
         // low-end-device-mode is enabled.
         paramOverrides.add("--disable-rgba-4444-textures");
-        // Disable QUIC to save CPU budgets on m114.
-        // Remove below if Cobalt rebase to m138+.
-        paramOverrides.add("--disable-quic");
```
Awallky, line 71: Can you also update the test class for this? (Assuming the test class tracks this CLI change)

zhongqiliang: https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java;l=1?q=CommandLineOverrideHelperTest.java&sq=git:github%2Fyoutube%2Fcobalt@26.android
There is no existing test for the flag, which is good. 

This change is for an temporary experiment, I think add a unit test for using the flag to enable the flag is overkilling.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -87,6 +93,7 @@ public abstract class CobaltActivity extends Activity {
   // Maintain the list of JavaScript-exposed objects as a member variable
   // to prevent them from being garbage collected prematurely.
   private List<CobaltJavaScriptAndroidObject> javaScriptAndroidObjectList = new ArrayList<>();
+  private Map<String, String> featureFlags = new HashMap<>();
```
Awallky: We should use the Chromium Java naming convention for private member variables -
go/cobalt-style#description
https://source.android.com/docs/setup/contribute/code-style#follow-field-naming-conventions

Awallky: I think @johnedocampo is adding feature flag support in https://github.com/youtube/cobalt/pull/8377. Can we add a TODO here to make sure that those changes are integrated here?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
Awallky: Sounds good. I will resolve this then.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -652,6 +671,11 @@ private String parseArg(String[] args, String argName) {
         .orElse(null);
   }
 
+  // Overrided by Kimono
+  protected Map<String, String> getFeatureFlags() {
+    return this.featureFlags;
```
johnxwork: Can we just pass command line argument from Kimono?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -124,9 +131,21 @@ protected void createContent(final Bundle savedInstanceState) {
       if (!VersionInfo.isReleaseBuild()) {
         commandLineArgs = getCommandLineParamsFromIntent(getIntent(), COMMAND_LINE_ARGS_KEY);
       }
+
+      Map<String, String> featureFlags = getFeatureFlags();
+      List<String> extraCommandLineArgs = new ArrayList<>();
```
johnxwork, line 133: Why can't we append to `commandLineArgs`?

zhongqiliang: Java String[] does not support append operation. The array has a fixed length, append will need to change length, that would need to create a new string array.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: My first version is to modify the list returned by getArgs().
I think this version is better. The getArgs() method is a mess today, I do not want to make it worse.

This new getFeatureFlags() API is the dedicated place for Kimono to pass gcl controlled flag values to Cobalt.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: The goal of this featureFlags is to have Kimono read a flag value from gcl config (which is controlled by YT servers), and pass it into Cobalt to control startup features. 

https://github.com/youtube/cobalt/pull/8377 is different. It is for Cobalt to define feature in Java, it is within Cobalt.

Feel free to reopen this if you have other questions.

zhongqiliang: https://github.com/youtube/cobalt/pull/8393 apply the pattern, not this one as other fields on 26.android are not on it.

Awallky: Ah, I see. This makes CP simpler. Since it is fixed in a separate PR, that should be fine.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: For test, we can override the gcl flag, no need to use command line arg from Kimono.

johnxwork: Isn't this PR making it more messy? We are adding an unnecessary layer of translation with the lookup of `EnableQUIC`. This new "feature" actually just controls the addition of `--disable_quic` command line param. You could just add `--disable_quic` from `ChrobaltMainActivity.java` when the experiment flag is off.

`getFeatureFlags()` will be easily misinterpreted as Chromium Finch features.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -652,6 +671,11 @@ private String parseArg(String[] args, String argName) {
         .orElse(null);
   }
 
+  // Overrided by Kimono
+  protected Map<String, String> getFeatureFlags() {
```
johnxwork: As discussed offline, we need to change the name to avoid confusion with Finch features. A few suggestions for you to pick from:
- getJavaSwitches
- getKimonoExperiments
- getJavaExperiments

PR #8393, opened 2025-12-12
android: Add feature flag map to CobaltActivity

zhongqiliang:
Same as https://github.com/youtube/cobalt/pull/8392,
Provide the API as 26.android for Kimono to use so that we do not need to branch the file on google3.

Bug: 467778593


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -515,6 +518,11 @@ private static boolean hasArg(List<String> args, String argName) {
     return false;
   }
 
+  // Overrided by Kimono
+  protected Map<String, String> getFeatureFlags() {
```
johnxwork: Posted in the other PR which seems to contain the same change:
- Can we just pass a command line argument from Kimono?

zhongqiliang: I replied on https://github.com/youtube/cobalt/pull/8392

johnxwork (approved):

PR #8395, opened 2025-12-12
media: Use uintptr_t for sample buffer IDs

kjyoun:
This change converts the sample buffer pointer to uintptr_t immediately in the DeallocateSampleCB callback. This prevents the pointer from being flagged by the dangling pointer detector when passed to base::BindOnce, as it is now treated as an integer ID rather than a pointer. The decoding_buffers_ map key is also updated to uintptr_t to match.

It is safe to use the numerical value of the `sample_buffer` (even if its underlying memory has been deallocated) because it is used purely as a lookup ID in `decoding_buffers_` and not for dereferencing its content. This approach explicitly avoids the need for `base::UnsafeDanglingUntriaged`, which aligns with best practices.

Per [unretained_dangling_ptr_guide.md](https://chromium.googlesource.com/chromium/src/+/136.0.7103.59/docs/unretained_dangling_ptr_guide.md):
> Avoid if at all possible.

Bug: 1234

### DIFF
--- a/media/starboard/sbplayer_bridge.cc
+++ b/media/starboard/sbplayer_bridge.cc
@@ -84,6 +84,10 @@ void SetDiscardPadding(
       discard_padding.second.InMicroseconds();
 }
 
+uintptr_t GetMemoryId(const void* addr) {
+  return reinterpret_cast<uintptr_t>(addr);
+}
+
 }  // namespace
 
 #if COBALT_MEDIA_ENABLE_STARTUP_LATENCY_TRACKING
@@ -138,13 +142,13 @@ void SbPlayerBridge::CallbackHelper::OnPlayerError(void* player,
 }
 
 void SbPlayerBridge::CallbackHelper::OnDeallocateSample(
-    const void* sample_buffer) {
+    uintptr_t sample_buffer_id) {
   if (!player_bridge_) {
     return;
   }
   CHECK(player_bridge_->task_runner_->RunsTasksInCurrentSequence());
 
-  player_bridge_->OnDeallocateSample(sample_buffer);
+  player_bridge_->OnDeallocateSample(sample_buffer_id);
 }
 
 void SbPlayerBridge::CallbackHelper::ResetPlayer() {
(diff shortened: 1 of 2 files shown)


media/starboard/sbplayer_bridge.cc
```diff
@@ -910,8 +914,9 @@ void SbPlayerBridge::WriteBuffersInternal(
...
     }
 
-    if (auto [iter, inserted] = decoding_buffers_.try_emplace(
-            buffer->data(), buffer, /*usage_count=*/1, sample_type);
+    if (auto [iter, inserted] =
+            decoding_buffers_.try_emplace(GetMemoryId(buffer->data()), buffer,
+                                          /*usage_count=*/1, sample_type);
         !inserted) {
       ++iter->second.usage_count;
     }
```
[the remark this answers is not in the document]
kjyoun: this is per default formatter

media/starboard/sbplayer_bridge.h
```diff
@@ -216,7 +216,7 @@ class SbPlayerBridge {
     int usage_count;
     SbMediaType type;
   };
-  using DecodingBuffers = absl::flat_hash_map<const void*, DecodingBuffer>;
+  using DecodingBuffers = absl::flat_hash_map<uintptr_t, DecodingBuffer>;
```
xiaomings, line 219: Would `raw_ptr<const void, DisableDanglingPtrDetection>` or something similar work here?

kjyoun: Using `DisableDanglingPtrDetection` implies it is still a pointer and we are just suppressing the safety check.   
(then we need to explain why it disable it)

By converting to uintptr_t, we make it self-evident that we are treating the address strictly as an opaque ID for lookup, not as a memory  address to be accessed. This makes the concept of a "dangling pointer" semantically irrelevant.

xiaomings: If either way works, it's more of a philosophical decision that probably both are right depends on how they are interpreted.

I prefer to use `DisableDanglingPtrDetection` because:
1. The pointer was sent over Starboard C interface, where C++ smart pointer techniques cannot be used here.  It's a valid use case for pointers, and using `DisableDanglingPtrDetection` is probably a more universal solution.
2. `sample_buffer` isn't only an id, it's a pointer guaranteed to be valid inside the callback by our `SbPlayerBridge` implementation.  We should inform the runtime checking system this situation explicitly, instead of casting to `uintptr_t` to avoid the check implicitly.
3. It's technically still a pointer, even when we cast it to uintptr_t.

kjyoun: I investigated a little deeper why the callback is called with dangled pointer from the beginning

See my analysis: https://b.corp.google.com/issues/452431699#comment10

I created a new PR to address this problem: https://github.com/youtube/cobalt/pull/8402
closing this PR

PR #8408, opened 2025-12-15
Add h5vcc.storage.clearCrashpadDatabase() support for modular builds …

ghost:
…to support testing.

Bug: 466092749

### DIFF
--- a/cobalt/browser/h5vcc_storage/DEPS
+++ b/cobalt/browser/h5vcc_storage/DEPS
@@ -1,0 +1,5 @@
+include_rules = [
+  "+content/public/browser",
+  "+mojo/public/cpp/bindings",
+  "+starboard",
+]
--- a/cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
+++ b/cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
@@ -1,0 +1,95 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "cobalt/browser/h5vcc_storage/h5vcc_storage_impl.h"
+
+#include "base/files/file_path.h"
+#include "base/files/file_util.h"
+#include "base/logging.h"
+#include "base/task/sequenced_task_runner.h"
+#include "base/task/single_thread_task_runner.h"
+#include "base/task/task_traits.h"
+#include "base/task/thread_pool.h"
+#include "base/threading/thread_restrictions.h"
+#include "build/build_config.h"
+#if BUILDFLAG(USE_EVERGREEN)
+#include "starboard/configuration_constants.h"  // nogncheck
+#endif
+
+namespace h5vcc_storage {
+
+namespace {
+
+#if BUILDFLAG(USE_EVERGREEN)
+const char kCrashpadDBName[] = "crashpad_database";
+#endif
+
+}  // namespace
+
+H5vccStorageImpl::H5vccStorageImpl(
+    content::RenderFrameHost& render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver)
+    : content::DocumentService<mojom::H5vccStorage>(render_frame_host,
+                                                    std::move(receiver)) {
+  DETACH_FROM_THREAD(thread_checker_);
+}
+
+H5vccStorageImpl::~H5vccStorageImpl() {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+}
+
+void H5vccStorageImpl::Create(
+    content::RenderFrameHost* render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver) {
+  new H5vccStorageImpl(*render_frame_host, std::move(receiver));
+}
+
+void H5vccStorageImpl::ClearCrashpadDatabase(
+    ClearCrashpadDatabaseCallback callback) {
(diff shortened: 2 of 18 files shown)


cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
```diff
@@ -0,0 +1,84 @@
...
+void H5vccStorageImpl::Create(
+    content::RenderFrameHost* render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver) {
+  new H5vccStorageImpl(*render_frame_host, std::move(receiver));
+}
+
+void H5vccStorageImpl::ClearCrashpadDatabase(
+    ClearCrashpadDatabaseCallback callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD) && !defined(COBALT_IS_RELEASE_BUILD)
```
andrewsavage1: This is different than the implementation in c25 (see https://lbshell-internal-review.git.corp.google.com/c/cobalt_src/+/297840), as there it will always clear the database. I wonder if we'll ever want this functionality in gold builds or non-modular builds— @hlwarriner do you know?

ghost: I based the implementation on the description in the bug.

https://b.corp.google.com/issues/466092749#comment1

Do we want to reject the promise if we don't clear the database (non-evergreen or other reason)?

ghost: Since this is what was asked for in the bug, can we approve this and create a follow-up to address issues like if it should work in release or modular or all?

andrewsavage1: Let's reject the promise for non-evergreen platforms like you suggest. I'm less sure about gold builds and whether they will need this feature available, but we can address that later like you suggest.

andrewsavage1:
Have you tested this end to end?

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
```diff
@@ -0,0 +1,93 @@
...
+      FROM_HERE,
+      base::BindOnce(
+          [](scoped_refptr<base::SingleThreadTaskRunner> task_runner,
+             ClearCrashpadDatabaseCallback callback) {
+            std::vector<char> cache_dir(kSbFileMaxPath + 1, 0);
+            if (SbSystemGetPath(kSbSystemPathCacheDirectory, cache_dir.data(),
+                                kSbFileMaxPath)) {
+              base::FilePath crashpad_db_dir =
+                  base::FilePath(cache_dir.data()).Append(kCrashpadDBName);
+              base::DeletePathRecursively(crashpad_db_dir);
```
hlwarriner: nit: thanks for adding error checking; please add similar error checking for `base::DeletePathRecursively`, which returns false if not successful.

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
hlwarriner: We do want to support this API in gold builds. Partners generally expect to be able to run YTS on both qa and gold builds.

third_party/blink/renderer/modules/cobalt/h5vcc_storage/h_5_vcc_storage.idl
```diff
@@ -0,0 +1,26 @@
...
+// LTS 25:
+// https://github.com/youtube/cobalt/blob/25.lts.stable/cobalt/h5vcc/h5vcc_storage.idl
+[
+  Exposed=Window,
+  SecureContext
+]
+interface H5vccStorage {
+  // Only intended to be used by tests.
+  [CallWith=ScriptState, RaisesException]
+  Promise<undefined> clearCrashpadDatabase();
```
hlwarriner, line 25: Thanks for making the change to async. Did you check with our client teams to make sure the change is ok?

hlwarriner (commented):
Thanks for taking this! I left a few comments.

third_party/blink/renderer/modules/cobalt/h5vcc_storage/h_5_vcc_storage.idl
ghost: I followed the bug description. Who should I reach out to to discuss the change in behavior?

hlwarriner: Comment 1 of b/466092749 does cover this. But it's ok, you can land this PR and I'll talk with the client teams about the async change.

cobalt/browser/h5vcc_storage/h5vcc_storage_impl.cc
```diff
@@ -0,0 +1,95 @@
...
+void H5vccStorageImpl::Create(
+    content::RenderFrameHost* render_frame_host,
+    mojo::PendingReceiver<mojom::H5vccStorage> receiver) {
+  new H5vccStorageImpl(*render_frame_host, std::move(receiver));
+}
+
+void H5vccStorageImpl::ClearCrashpadDatabase(
+    ClearCrashpadDatabaseCallback callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
```
hlwarriner: I'd prefer that we use `BUILDFLAG(USE_EVERGREEN)` to guard this implementation, since in chrobalt we - at least for now - only use crashpad in evergreen builds and don't use it for non-evergreen modular builds. Please make that change unless there's a good reason not to. Sorry that I missed that in my first review.

ghost: No problem. I'll make the change and test that. I appreciate the quick review and actionable comments 👍

cobalt/browser/h5vcc_storage/DEPS
```diff
@@ -0,0 +1,5 @@
+include_rules = [
```
andrewsavage1, line 1: I guess this is to appease the chromium presubmit check that we added?

andrewsavage1 (approved):

cobalt/browser/h5vcc_storage/DEPS
ghost: Exactly!

PR #8415, opened 2025-12-15
cobalt: Fix missing DidStopLoading

borongc:
`DidStopLoading` is to set initial focus via ContentsWebObserver, and it was added in https://github.com/youtube/cobalt/pull/6845 and missed during 
https://github.com/youtube/cobalt/pull/8352.

Issue: 433965181
Issue: 466120031
Issue: 469163663


yell0wd0g (approved):
LGTM in case it's needed.

Gemini suggestion doesn't look bad tho.

PR #8419, opened 2025-12-16
feat: Set mapped_memory_reclaim_limit for 3p devices

sherryzy:
Previously, the display compositor's mapped memory reclaim limit was
not explicitly set on Starboard, leading to uncontrolled memory usage
that could exceed 100 MB.

This change introduces a dynamic calculation of the limit based on the
full screen texture size. This ensures memory is aggressively
reclaimed, preventing excessive accumulation of mapped memory, similar
to the approach used for Android builds.

Bug: 463684608

### DIFF
--- a/gpu/command_buffer/client/shared_memory_limits.h
+++ b/gpu/command_buffer/client/shared_memory_limits.h
@@ -75,7 +75,7 @@ struct SharedMemoryLimits {
     return limits;
   }
 
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_STARBOARD)
   static SharedMemoryLimits ForDisplayCompositor(const gfx::Size& screen_size) {
     DCHECK(!screen_size.IsEmpty());
 
--- a/content/browser/compositor/viz_process_transport_factory.cc
+++ b/content/browser/compositor/viz_process_transport_factory.cc
@@ -49,6 +49,10 @@
 #include "services/viz/public/cpp/gpu/context_provider_command_buffer.h"
 #include "ui/base/ui_base_features.h"
 
+#if BUILDFLAG(IS_STARBOARD)
+#include "ui/display/screen.h"
+#endif
+
 #if BUILDFLAG(IS_WIN)
 #include "ui/gfx/win/rendering_window_manager.h"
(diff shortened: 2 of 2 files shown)


gpu/command_buffer/client/shared_memory_limits.h
```diff
@@ -97,6 +97,16 @@ struct SharedMemoryLimits {
...
 
+    return limits;
+  }
+#elif BUILDFLAG(IS_STARBOARD)
+  static SharedMemoryLimits ForDisplayCompositor(const gfx::Size& screen_size) {
+    DCHECK(!screen_size.IsEmpty());
+    SharedMemoryLimits limits;
+    constexpr uint32_t kBytesPerPixel = 4;
+    const uint32_t full_screen_texture_size_in_bytes =
+        screen_size.width() * screen_size.height() * kBytesPerPixel;
```
yell0wd0g: These lines look similar to l.80-85 LHS - could we just merge this
code with the IS_ANDROID one, or perhaps even just use the IS_ANDROID..?

sherryzy: I tested to directly use the Android path and it seems work fine. But in the comment above, some settings are Android specific, do you know if those apply to 3P(line 87 - 95)?

yell0wd0g: I dug a bit in the codebase: These limits are used for the TransferBuffers
that are interchanged between CommandBuffer clients (e.g. the Browser
process for UI stuff, or a Renderer client such as Blink Compositor, 
WebGL or WebGPU) and the GPU process. Putting limits on the TransferBuffer
makes sense, specially since only commands/paint ops should be transferred
directly (all images should go a separate way as SharedImages, or perhaps
still GpuMemoryBuffers back in 114 - but when I asked Gemini they said
that a WebGL texture would be _inside_ the TransferBuffer - I doubt that).
So I would say using Android limits for our Starboard platform should 
be safe; IOW Starboard is more like a mobile device than a desktop
platform. Makes sense?

sherryzy: Yes, totally makes sense. And I just did more testing using the Android settings for Starboard, and everything looks fine. Thanks for looking into the code!

PR #8436, opened 2025-12-17
Add requestHistogram function to h5vcc metrics.

yjzhang111:
Introduce a new JavaScript API, h5vcc.metrics.requestHistograms(), to synchronously fetch and return histogram data as a JSON string. This provides a workaround for existing onMetrics functionality and follows the Chromium DevTools Protocol implementation for exposing histogram information. This allows direct polling of metric data.

Bug: 467384307

### DIFF
--- a/cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.cc
+++ b/cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.cc
@@ -15,13 +15,11 @@
 #include "cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.h"
 
 #include "base/base64url.h"
-#include "base/json/json_writer.h"
 #include "base/logging.h"
 #include "base/metrics/histogram.h"
 #include "base/metrics/statistics_recorder.h"
 #include "base/threading/thread_checker.h"
 #include "base/time/time.h"
-#include "base/values.h"
 #include "cobalt/browser/global_features.h"
 #include "cobalt/browser/metrics/cobalt_metrics_services_manager_client.h"
 #include "components/metrics/metrics_log.h"
@@ -76,8 +74,7 @@ void H5vccMetricsImpl::SetMetricEventInterval(
   std::move(callback).Run();
 }
 
-void H5vccMetricsImpl::RequestHistograms(bool monitor_mode,
-                                         RequestHistogramsCallback callback) {
+void H5vccMetricsImpl::RequestHistograms(RequestHistogramsCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   // Synchronously fetch subprocess histograms that live in shared memory.
   // This is the only mechanism available to embedders like Cobalt, as the
@@ -106,10 +103,6 @@ void H5vccMetricsImpl::RequestHistograms(bool monitor_mode,
                           metrics::MetricsLog::LogType::ONGOING_LOG,
                           service_client);
 
-  if (!monitor_mode) {
-    last_histogram_samples_.clear();
-  }
-
   for (base::HistogramBase* const histogram :
        base::StatisticsRecorder::GetHistograms()) {
     auto samples = histogram->SnapshotSamples();
@@ -117,17 +110,14 @@ void H5vccMetricsImpl::RequestHistograms(bool monitor_mode,
       continue;
     }
 
-    if (monitor_mode) {
-      auto it = last_histogram_samples_.find(
-          std::string(histogram->histogram_name()));
-      if (it != last_histogram_samples_.end()) {
-        samples->Subtract(*it->second);
-        it->second = histogram->SnapshotSamples();
-      } else {
-        last_histogram_samples_.emplace(
-            std::string(histogram->histogram_name()),
-            histogram->SnapshotSamples());
-      }
+    auto it =
+        last_histogram_samples_.find(std::string(histogram->histogram_name()));
+    if (it != last_histogram_samples_.end()) {
+      samples->Subtract(*it->second);
+      it->second = histogram->SnapshotSamples();
+    } else {
+      last_histogram_samples_.emplace(std::string(histogram->histogram_name()),
+                                      histogram->SnapshotSamples());
     }
     if (samples->TotalCount() > 0) {
       log.RecordHistogramDelta(histogram->histogram_name(), *samples);
--- a/cobalt/browser/h5vcc_metrics/public/mojom/h5vcc_metrics.mojom
+++ b/cobalt/browser/h5vcc_metrics/public/mojom/h5vcc_metrics.mojom
@@ -29,5 +29,5 @@ interface H5vccMetrics {
 
   [Sync] Enable(bool enable) => ();
   SetMetricEventInterval(uint64 interval_seconds) => ();
-  RequestHistograms(bool monitor_mode) => (string histograms_json);
+  RequestHistograms() => (string histograms_json);
 };
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl
@@ -63,5 +63,5 @@ interface H5vccMetrics : EventTarget {
   // being sent as part of the onmetrics Event. Remove this API when Cobalt
   // and client web apps have fully migrated to said Event.
   [CallWith=ScriptState, RaisesException]
-  Promise<DOMString> requestHistograms(boolean monitor_mode);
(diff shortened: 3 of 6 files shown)


cobalt/browser/h5vcc_metrics/public/mojom/h5vcc_metrics.mojom
```diff
@@ -29,4 +29,5 @@ interface H5vccMetrics {
 
   [Sync] Enable(bool enable) => ();
   SetMetricEventInterval(uint64 interval_seconds) => ();
+  [Sync] RequestHistograms() => (string histograms_json);
```
yell0wd0g: I know this is a draft, but I'd recommend to make this Async
since anyway the IDL corresponding version returns a Promise.

third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl
```diff
@@ -56,4 +56,6 @@ interface H5vccMetrics : EventTarget {
   // again with a different value."
   // This API is now Promise-based.
   [CallWith=ScriptState, RaisesException] Promise<undefined> setMetricEventInterval(unsigned long long intervalSeconds);
+
+  [CallWith=ScriptState, RaisesException] Promise<DOMString> requestHistograms();
```
yell0wd0g: Can we leave a comment here saying e.g. 
```
// Retrieves the current status of all collected histograms. 
// TODO(b/XYZ): This is a polling version of the corresponding
// data being sent as part of the onmetrics Event. Remove this
// when Cobalt and client WebApps have fully migrated to said
// Event.
```

A question we should answer here or elsewhere is how large can 
this DOMString get? IIRC the `onmetrics` only provides new stuff
but if we imitate the Chrome UI interface, we'll get all the 
histograms all the time. For example on my Chrome Mac Beta, the
histograms XML is 14MB. I'd say the JSON would be comparable -
could this interfere with the normal Web page activity?

Another question is whether it's OK for JS to send many of these
requests simultaneously / can JS send a `requestHistograms()`
without waiting for a previous one to finish. I'd say yes and
the IDL and the code seems to point in that direction.

cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.cc
```diff
@@ -66,4 +76,51 @@ void H5vccMetricsImpl::SetMetricEventInterval(
...
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  // Synchronously fetch subprocess histograms that live in shared memory.
+  // This is the only mechanism available to embedders like Cobalt, as the
+  // fully async HistogramSynchronizer is a content internal implementation.
+  base::StatisticsRecorder::ImportProvidedHistogramsSync();
+
+  auto* manager_client =
+      cobalt::GlobalFeatures::GetInstance()->metrics_services_manager_client();
+  auto* service_client = manager_client->metrics_service_client();
+  auto* state_manager = manager_client->GetMetricsStateManager();
```
yell0wd0g, line 87: QQ: can |state_manager| be nullptr? 

For context, |service_client| is retrieved with a camel case method
name, which by style guide tells that it's just a member variable
retrieval. But |state_manager| is retrieved via a full fledged method
-- which implies something more complex, that can possibly fail. Just
bringing that up.

cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.cc
```diff
@@ -66,4 +76,51 @@ void H5vccMetricsImpl::SetMetricEventInterval(
...
+
+  ::metrics::ChromeUserMetricsExtension uma_proto;
+  // We create a temporary MetricsLog object to facilitate serializing
+  // histograms. This class requires a session ID in its constructor, but the
+  // value is not used for the purposes of this function. The session ID is
+  // stored in the system profile, but this function only extracts the
+  // histogram data, and the rest of the log (including the system profile) is
+  // discarded. Therefore, a placeholder value of 1 is sufficient.
+  metrics::MetricsLog log(state_manager->client_id(), 1,
```
yell0wd0g: nit: Either:

```
metrics::MetricsLog log(state_manager->client_id(), /*session_id=*/1,
...
```

Or
```
constexpr int kDontCareSessionId = 1;
metrics::MetricsLog log(state_manager->client_id(), kDontCareSessionId,
...
```

yell0wd0g (changes requested):
Please re-generate the commit message and title
(`/generate-commit-message`) and use it as source
for the PR title and description. Then run Gemini
code reviews on the CL (it has found at least two
major bugs in recent past !). Thanks!

cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.cc
yjzhang111: No it creates a state_manager internally when it's a nullptr.

third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl
yjzhang111: Not exactly sure how big the increase would be but we'll probably need to do an experiment on Kabuki side to slowly ramp this up so we can probably infer the interference with web app from the experiment data...?

Yes I think it should be OK. Can also test this when making Kabuki changes.

cobalt/browser/h5vcc_metrics/h5vcc_metrics_impl.cc
yell0wd0g: Gotcha then consider leaving a `CHECK(state_manager);` to signal
that it can't fail.

third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl
yell0wd0g: I think it'll get very large very soon - and then stabilize since there
would not be (many) new histograms populated past a certain moment. But
this is a problem that JS would encounter anyway, instructing the UA to
run a large operation.

To reduce the potential size I see that the UI implementation also has a
monitoring mode in [//content/browser/metrics/histograms_internals_ui.cc](https://source.chromium.org/chromium/chromium/src/+/main:content/browser/metrics/histograms_internals_ui.cc;l=136-151;drc=38954254daa6a6682eba968b3465bd4c55a79d08).
Maybe we could implement it as a follow-up here? It would need two methods
or a more complex arrangement, to start monitoring / get values / stop,
so it can be done after this PR.

yell0wd0g (approved):
LGTM modulo:
a) Taken up Gemini's commit title and description suggestion and
b) Adding the information about where the inspiration for
RequestHistograms came from to a function comment in the code 
(e.g. `// Function inspired by Chrome's WebUI's HistogramsMessageHandler::HandleRequestHistograms()`).

What's the plan for unit tests (blink_unittest) and/or 
WebTests for this?

third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl
yjzhang111: I took a quick stab at implementing a incremental UMA upload, could you take another look? 

When I tested it through devtools, it seems like the encoded UMA data are at most ~60kb. I wonder where this discrepancy is from - it could be that we're only converting the UMA events and stripping a lot of the other protos like system profile...?

yell0wd0g: Possible - my number came from downloading chrome:histograms as text html :)
so perhaps in binary form is more reduced.

If we wanted an incremental UMA version, I'd add a separate 2-method.
Otherwise the API gets complicated quickly as questions mount on the
ergonomics, e.g. what happens if `requestHistograms()` is called in
sequence with `monitor_mode` true and then false? Maybe there're simple
explanations, but all this would need to be spelled in the IDL file and
there should be WebTests (or at least unit tests) to be added (this
CL has no testing at all FWICT). So I'd suggest taking that off and
landing the simple version first (which in turn will need cherry picking
etc).

yjzhang111: I removed the option to upload entire UMA snapshot bc I didn't find any Cobalt custom logic to process UMA proto into YT GEL logs and we likely inherited it from Chrome. Keeping the uploading delta only logic just in case GEL doesn't process the log and simply sums everything up.

Resolving this comment for now but please feel free to reopen if you think this isn't the right approach

PR #8438, opened 2025-12-17
cobalt/test: navigation_browsertest to use GetGlobalFrameToken API

Awallky:
This replaces `blink::LocalFrameToken` with the more robust `GetGlobalFrameToken()` API for identifying frames across processes.
It is one of several changes to navigation_browsertests as part of the effort to enable browser tests on the main branch for Cobalt.
**Note:** This is branched on top of https://github.com/youtube/cobalt/pull/8437 as part of a chain. Review that first!

Bug: 459831767


Awallky:
After the rebase, the delta looks clean and the std::optional changes are absent from the previous commit.

PR #8440, opened 2025-12-17
tvOS: Make media_can_play_mime_and_key_system.mm build

rakuco:
Patch by @Gyuyoung.

Adjust the file to the namespace flattening PRs that have landed in the past few months so that it can be plugged into the build system later.

Bug: 456692793


rakuco:

fayev-g (approved):

PR #8444, opened 2025-12-17
cobalt/test: Add race condition tests for navigation commits

Awallky:
Introduces new test fixtures (`UndoCommitNavigationBrowserTest`, `NavigationQueueingBrowserTest`) to specifically test race conditions that can occur when a new navigation begins while a previous one is in a pending commit state.

This is one of several changes to navigation_browsertests as part of the effort to enable browser tests on the main branch for Cobalt.
Note: This is branched on top of #8443 as part of a chain.

Bug: 459831767

PR #8445, opened 2025-12-17
cobalt/test:  Add tests for credentialless iframes and sandbox warnings

Awallky:
Adds new browser tests to verify the behavior of security-related iframe attributes:
  - Ensures cookies are correctly handled for iframes with the `credentialless` attribute.
  - Verifies that a console warning is emitted for iframes with an escapable sandbox policy (`allow-scripts` and `allow-same-origin`).

This is one of several changes to navigation_browsertests as part of the effort to enable browser tests on the main branch for Cobalt.
Note: This is branched on top of #8444 as part of a chain.

Bug: 459831767

### DIFF
--- a/cobalt/testing/browser_tests/navigation_browsertest.cc
+++ b/cobalt/testing/browser_tests/navigation_browsertest.cc
@@ -6389,9 +6389,79 @@ class NavigationBrowserTestCredentiallessIframe : public NavigationBrowserTest {
   }
 };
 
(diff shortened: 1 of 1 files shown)


yell0wd0g (approved):
Took a look at 37bb397ebc4871dbaa8f9f3a4463d1f1da64f9e0 and
it looks good -- I'm assuming the test logic is copied from
elsewhere?

PR #8447, opened 2025-12-17
cobalt: Fix sdallocx and OPENSSL_memory symbol leaks

jellefoks:
Neuter BoringSSL weak symbols for Cobalt to resolve leaks in the shared library.

This change disables the weak symbol logic in BoringSSL's `mem.c` by adding `!BUILDFLAG(IS_COBALT)` to the macro condition. This forces `sdallocx`, `OPENSSL_memory_alloc`, `OPENSSL_memory_free`, and `OPENSSL_memory_get_size` to resolve to `NULL` at compile time.

Since these symbols are now resolved to `NULL` (and defined as static function pointers), the compiler can optimize away the conditional blocks that use them, completely removing the symbol references from the final Cobalt binary. This prevents them from leaking as undefined weak symbols in the shared library.

As a result, these symbols have also been removed from the API leak detector manifests.

Bug: 459893921
Bug: 470083196

### DIFF
--- a/starboard/tools/api_leak_detector/evergreen/dev_manifest
+++ b/starboard/tools/api_leak_detector/evergreen/dev_manifest
@@ -2,10 +2,6 @@
 
 # This file was auto-generated using api_leak_detector.py.
 
-OPENSSL_memory_alloc
-OPENSSL_memory_free
-OPENSSL_memory_get_size
-sdallocx
 TLS init function for partition_alloc::internal::g_thread_cache
 TLS init function for v8::internal::trap_handler::g_thread_in_wasm_code
 TLS init function for v8::internal::wasm::CodeSpaceWriteScope::scope_depth_
--- a/starboard/tools/api_leak_detector/evergreen/manifest
+++ b/starboard/tools/api_leak_detector/evergreen/manifest
@@ -2,9 +2,5 @@
 
 # This file was auto-generated using api_leak_detector.py.
 
-OPENSSL_memory_alloc
-OPENSSL_memory_free
-OPENSSL_memory_get_size
-sdallocx
 TLS init function for partition_alloc::internal::g_thread_cache
 TLS init function for v8::internal::trap_handler::g_thread_in_wasm_code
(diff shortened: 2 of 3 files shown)


cobalt/common/libc/stdlib/sdallocx.cc
```diff
@@ -0,0 +1,29 @@
...
+
+extern "C" {
+
+// sdallocx is a sized |free| function. By passing the size (which we happen to
+// always know in BoringSSL), the malloc implementation can save work. We cannot
+// depend on |sdallocx| being available, however, so it's a weak symbol.
+//
+// This is defined here to satisfy the weak symbol in BoringSSL and prevent it
+// from being exported as an undefined weak symbol.
+void sdallocx(void* ptr, size_t size, int flags) {
```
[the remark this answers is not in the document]
jellefoks: Done.

cobalt/common/libc/stdlib/sdallocx.cc
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <stdlib.h>
+
+extern "C" {
+
+// sdallocx is a sized |free| function. By passing the size (which we happen to
```
andrewsavage1: This comment is a bit confusing to me because I read it as if we will use the `size` variable in the sdallocx implementation (but of course we don't, and I think it's fine not to)

Also, could you add a line saying the flags are irrelevant to sdallocx so we ignore them?

jellefoks: This is the exact comment by the function in boringssl, but of course, I'll add a commit that phrases it better for the casual reader.

jellefoks:
I changed the approach to eliminate the weak symbols closer to the source. Also added the OPENSLL_memory ones.

cobalt/common/libc/stdlib/sdallocx.cc
jellefoks: Changed the approach.

PR #8451, opened 2025-12-17
cobalt/test: Add Remaining missing tests - Part 4

Awallky:
Ports over a number of missing browser tests from the content_browsertests target and modifies various API surfaces from the M114-M138 rebase. This is the second in a multi-part effort.

This is one of several changes to navigation_browsertests as part of the effort to enable browser tests on the main branch for Cobalt.
Note: This is branched on top of https://github.com/youtube/cobalt/pull/8450 as part of a chain.

Bug: 459831767


Awallky:
The only real difference here would be the use of `cobalt/shell` instead of `content/shell`.

PR #8458, opened 2025-12-18
cobalt: Enable av1 codec

borongc:
Remove the explicit disabling of the dav1d AV1 decoder from the common
build configuration. This change enables the AV1 video codec, allowing
Cobalt to play media streams that utilize this format, improving content
playback capabilities.

Issue: 463540505

### DIFF
--- a/cobalt/build/configs/common.gn
+++ b/cobalt/build/configs/common.gn
@@ -53,7 +53,6 @@ v8_use_external_startup_data = false
 enable_hls_demuxer = false
 
 # Many of these come from media_options.gni.
-enable_dav1d_decoder = false
 enable_hevc_parser_and_hw_decoder = false
 enable_libaom = false
 enable_platform_dolby_vision = false


cobalt/build/configs/common.gn
```diff
@@ -53,7 +53,6 @@ v8_use_external_startup_data = false
 enable_hls_demuxer = false
 
 # Many of these come from media_options.gni.
-enable_dav1d_decoder = false
```
yell0wd0g, line 56: Two things: Just removing it here would rely on it set elsewhere,
could we leave instead either an explicit set to `true` with a comment
on why SW AV1 decoding is needed, or an assertion:

`assert(enable_dav1d_decoder, "Dav1d must be enabled for AV1 decoding e.g. for Splash screen");`


Other question is: with this change, `enable_dav1d_decoder` will be set
to true by `//media`, but the SW AV1 decoding bit would be driven within
SbPlayer implementation, i.e. under `//starboard`, so I guess we should
add the `enable_dav1d_decoder = true` or the assertion to 
`starboard.gni`?

niranjanyardi (approved):

cobalt/build/configs/common.gn
borongc: > Just removing it here would rely on it set to try elsewhere,
> could we leave instead either an explicit set to true with a comment
> on why SW AV1 decoding

We have YTS on Github now, and it should be able to catch this regression.

> with this change, enable_dav1d_decoder will be set
> to true by //media, but the SW AV1 decoding bit would be driven within
> SbPlayer implementation, i.e. under //starboard

Leave this as a TODO to decouple parse av1 from mp4_parser and build dav1d_decoder into Cobalt.

PR #8465, opened 2025-12-18
tvOS: Remove uikit.h and vector.h

rakuco:
* `uikit.h` has been unused since #8116
* `vector.h` has been unused since #8366

Bug: 459507205


rakuco:

PR #8466, opened 2025-12-18
Remove libdav1d from WebRTC

y4vor:
build: Disable libdav1d in WebRTC

This change disables the inclusion of the dav1d AV1 decoder library
within the WebRTC internal decoder factory. Doing so significantly
reduces the overall binary size by approximately 533KB, optimizing the
application for smaller footprints.

Bug: 421171908

### DIFF
--- a/cobalt/build/configs/linux_common.gn
+++ b/cobalt/build/configs/linux_common.gn
@@ -120,5 +120,5 @@ optimize_for_size = true
 # Disable cast remote RPC as we don't use it.
 enable_media_remoting_rpc = false
 
-# Enable v8 snapshot compression to reduce binary size (~167K) for a small startup cost (<10ms).
+# Enable v8 snapshot compression to reduce binary size (~167KB) for a small startup cost (<10ms).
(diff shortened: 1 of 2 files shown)


cobalt/build/configs/linux_common.gn
```diff
@@ -120,5 +120,8 @@ optimize_for_size = true
 # Disable cast remote RPC as we don't use it.
 enable_media_remoting_rpc = false
 
-# Enable v8 snapshot compression to reduce binary size (~167K) for a small startup cost (<10ms).
+# Enable v8 snapshot compression to reduce binary size (~167KB) for a small startup cost (<10ms).
 v8_enable_snapshot_compression = true
+
+# Disable dav1d in webrtc to reduce binar size (~533KB)
+rtc_include_dav1d_in_internal_decoder_factory = false
```
yell0wd0g: Should we do this for all platforms/ main branch?

y4vor: We could

andrewsavage1 (approved):

PR #8468, opened 2025-12-18
Nolibvpx

y4vor:
Cobalt: Disable libvpx for Linux builds

This change disables the libvpx library, including VP8 and VP9 codec
support within Blink and WebRTC for Cobalt Linux builds. This helps to reduce
the overall binary size of the Cobalt application by removing unused
video codec implementations. 

The binary size for arm gold builds is reduced with ~700KB.

Bug: 421171908

### DIFF
--- a/third_party/webrtc/media/engine/internal_encoder_factory_starboard_stub.cc
+++ b/third_party/webrtc/media/engine/internal_encoder_factory_starboard_stub.cc
@@ -1,0 +1,43 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "media/engine/internal_encoder_factory.h"
+
+#include <memory>
+#include <string>
+#include <vector>
+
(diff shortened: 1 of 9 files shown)


third_party/webrtc/media/engine/internal_encoder_factory_starboard_stub.cc
```diff
@@ -0,0 +1,43 @@
...
+#include <string>
+#include <vector>
+
+#include "absl/strings/match.h"
+#include "api/video_codecs/video_encoder_factory.h"
+#include "api/video_codecs/video_encoder_factory_template.h"
+
+namespace webrtc {
+
+std::vector<SdpVideoFormat> InternalEncoderFactory::GetSupportedFormats()
```
andrewsavage1, line 27: Should we add NOTIMPLEMENTEDs to these functions?

y4vor: I think it is fine for the rest of WebRTC it to call it. It just returns that no codecs are supported. I don't want it to crash when it is called.

y4vor:
> This LGTM, but are you against changing is_starboard to i.e. `use_libvpx = `is_starboard`? It seems like a pretty simple change and would make upstreaming easier

It is not that simple. The upstream change might look quite different.
There are already rtc_build_libvpx and rtc_libvpx_build_vp9, but they don't do much anymore and there is nothing for vp8. I think they used to have those flags, but the code was cleaned up to not use them anymore. Which would be interesting for us as we would actually want to re-introduce them.

Also he Blink part is separate and would need a different flag.

My thinking is that for upstream it would be a different task and would take some design discussion with the reviewers.

third_party/webrtc/media/engine/internal_encoder_factory_starboard_stub.cc
andrewsavage1: Note that NOTIMPLEMENTED will just log a warning, not crash like NOTREACHED, but it seems fine to omit if this is a valid implementation that we want to keep as is

PR #8472, opened 2025-12-18
android: Always enable debug and profile for CoAT Android app

kjyoun:
CoAT builds are intended for testing and debugging, not production
releases. This change hardcodes the debuggable and profileable flags
to always be enabled within the CoAT Android manifest.

This simplifies development workflows by ensuring debug and profiling
tools can always attach without needing to track the specific build
type in use.

Bug: 462494527


Awallky (approved):
If the manifest is not used by Kimono, then we can probably just do whatever with our Chrobalt manifest, since it is really just a test app and serves as a proxy for Kimono.

PR #8481, opened 2025-12-19
Test rdk files path

ghost:


ghost:
This served it's purpose. The changes have made it into the external RDK library.

PR #8489, opened 2025-12-20
Revert "h5vcc_accessibility: Split port-specific code into separate files"

haozheng-cobalt:
Reverts youtube/cobalt#8368

This is a very good improvement to segregate the logic per platform. 
However we are doing a refactor to replace the observer pattern with another instance H5vccAccessibilityManager to specifically handle ontexttospeechchange API. This way H5vccAccessibilityImpl deals only with the texttospeech API, and the PlatformTextToSpeechHelper class in this PR does not need to include the Client interface for the high level H5vccAccessibilityImpl to implement the NotifyTextToSpeechChange. And PlatformTextToSpeechHelper would only need to implement IsTextToSpeechEnabled per platform. 

Then I think this PR would be an overkill to create this many platform files just to support a single function, and we already provide platforms files under starboard directory. It seems unnecessary. 

Open to discuss further, I'll revert this for now in order to merge the more urgent PR to support ontexttospeechchange on 3P.

Bug: 447135715


andrewsavage1:
FYI @rakuco

PR #8503, opened 2025-12-23
android: Patch @NullMarked to Cobalt shell

borongc:
Introduce NullMarked and related nullability annotations to the Cobalt
Android shell classes. This change aligns the null-safety practices
with Chromium upstream's Content shell, improving static analysis
and preventing potential null pointer exceptions. It involves adding
`Nullable` to fields and parameters, and utilizing assertNonNull and
assumeNonNull for explicit null checks.

Issue: 471076830
Issue: 418842688

Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/6472891

### DIFF
--- a/cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
+++ b/cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
@@ -14,6 +14,9 @@
 
 package dev.cobalt.shell;
 
+import static org.chromium.build.NullUtil.assertNonNull;
+import static org.chromium.build.NullUtil.assumeNonNull;
+
 import android.app.Activity;
 import android.content.Context;
 import android.graphics.Rect;
@@ -25,6 +28,10 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
 import org.chromium.base.Callback;
+import org.chromium.base.ResettersForTesting;
+import org.chromium.build.annotations.Initializer;
+import org.chromium.build.annotations.NullMarked;
+import org.chromium.build.annotations.Nullable;
 import org.chromium.components.embedder_support.view.ContentViewRenderView;
 import org.chromium.content_public.browser.ActionModeCallbackHelper;
 import org.chromium.content_public.browser.LoadUrlParams;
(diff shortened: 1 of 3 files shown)


cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
```diff
@@ -174,7 +183,7 @@ public void loadUrl(String url) {
      * @return The sanitized URL.
      */
     public static String sanitizeUrl(String url) {
-        if (url == null) return null;
+        assert url != null;
```
Awallky: This gets stripped away in release builds by R8. If we remove the if check but keep the assert, we lose some of the check functionality. If we want this to throw an exception in all builds, we should throw an exception such as a `IllegalArgumentException`. If you want to keep the assert there for debug builds, you should wrap it in a block:

```
if (BuildConfig.ENABLE_ASSERTS) {
    // Any code here will be stripped in release builds by R8.
    ...
}
```

and then in the else block, throw an exception so release builds also get an exception thrown during runtime. For simplicity, I would recommend just throwing an exception if we expect this to always be non-null.

Source - https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/java/java.md#asserts

borongc: Right, nice catch. Reverted it back to the original code, as Cobalt Shell is in prod (not just a debug build for testing).

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -26,6 +26,7 @@
 import android.net.Uri;
 import android.os.Bundle;
 import android.os.SystemClock;
+import android.text.TextUtils;
```
Awallky, line 29: Can you remove this since we do not change anything in this class?

Awallky: I guess this will go away if we land PR#8530 first.

borongc: Right, this will be fixed with https://github.com/youtube/cobalt/pull/8530.

PR #3972, opened 2024-08-13
Prioritize IPv4 addresses over IPv6

kaidokert:
Changes the default RFC3484 address ordering to prefer IPv4 for current versions of Cobalt. This restores the behavior in prior Cobalt versions.

Test-on-device: True

b/359376180

PR #8508, opened 2025-12-23
Prioritize IPv4 addresses over IPv6 (#3972)

johnxwork:
Changes the default RFC3484 address ordering to prefer IPv4 for current versions of Cobalt. This restores the behavior in prior Cobalt versions.

Bug: 359376180


net/dns/address_sorter_posix.cc
```diff
@@ -142,7 +142,13 @@ const AddressSorterPosix::PolicyEntry kDefaultPrecedenceTable[] = {
     // ::/0 -- any
     {{}, 0, 40},
     // ::ffff:0:0/96 -- IPv4 mapped
+#if BUILDFLAG(IS_COBALT)
+    // Cobalt currently prioritizes IPv4 addresses higher, as
+    // suggested in section 10.3 of RFC3484
+    {{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF}, 96, 100},
+#else
     {{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF}, 96, 35},
+#endif
```
yell0wd0g, line 151: To reduce the delta, could we do e.g.

```
#if BUILDFLAG(IS_COBALT)
// Comment...
constexpr kMagicValueIPPriority = 100;
#else 
constexpr kMagicValueIPPriority = 35;
#endif 

// ...

// If constexpr is not enough, then use a #define .
    {{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF}, 96, kMagicValueIPPriority},
```

yell0wd0g (approved):
Looking good - is this ready to land, or just experimental?

net/dns/address_sorter_posix.cc
jellefoks: An issue with that approach is that we would then have a modification to upstream code that is not guarded by a "#if BUILDFLAG(IS_COBALT)" block, which we've typically tried to use as much as possible, to make it clear what the Cobalt modification is when resolving merge conflicts during rebases.

net/dns/address_sorter_posix.cc
```diff
@@ -142,7 +142,13 @@ const AddressSorterPosix::PolicyEntry kDefaultPrecedenceTable[] = {
     // ::/0 -- any
     {{}, 0, 40},
     // ::ffff:0:0/96 -- IPv4 mapped
+#if BUILDFLAG(IS_COBALT)
+    // Cobalt currently prioritizes IPv4 addresses higher, as
+    // suggested in section 10.3 of RFC3484
+    {{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xFF, 0xFF}, 96, 100},
```
jellefoks, line 148: A nit, but technically a device could be put on a network that returns invalid DNS results where this matters: I don't think we should set the precedence higher than the ipv6 loopback, which is set to 50. So a value of 45 is better and will also prefer ipv4 over ipv6 for non-loopback addresses.

Also, I didn't read the code that walks this array, but should this row maybe be moved such that it preserves the decreasing precedence order? It's probaby just a minor performance reduction if we don't.

johnxwork: I did some research and 45 vs 100 are both good answers. Gemini actually prefers 100 because we'd be able to avoid ipv6 compatibility issues completely. If a user's device does not handle ipv6 very well, there's possibility that ipv6 loopback address does not work there either.

Anyway, I don't think it's a big issue. We don't use loopback address except for yts tests AFAIK anyway.

PR #8509, opened 2025-12-23
cobalt/tools: Update instructions for android memory monitor

tasunnn:
Since vpython3 currently does not have support for several needed packages such as matplotlib, this change updates setup instructions for memory tools to use python3 instead of vpython3 as a workaround.

Bug: 472313590


Awallky:
For the description, we *do* have vpython support, but it does not have the required package for this tool.
It was a mistake on my end for claiming to use vpython instead of regular python.

Could you update the PR description? We can say something similar to what you wrote, but say "Because vpython does not currently support matplotlib, ...etc"

PR #8510, opened 2025-12-23
cobalt: Cache H5vccAccessibility text-to-speech state

yell0wd0g:
This PR introduces a cache for the result of H5vccAccessibility.textToSpeech().
This avoids repeated synchronous Mojo round trips to the browser process
when querying the text-to-speech enabled state. The cached value is
updated when the browser notifies the renderer of a state change. This
improves performance and responsiveness by reducing IPC overhead.

Also adds a unit test for this caching behavior.

Bug: 458102489


yell0wd0g:
@haozheng-cobalt this is what I meant.

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
```diff
@@ -62,9 +65,19 @@ class MODULES_EXPORT H5vccAccessibility final
...
  private:
+  FRIEND_TEST_ALL_PREFIXES(H5vccAccessibilityTest,
+                           IsTextToSpeechEnabledSyncWithCachedValue);
+
   void EnsureRemoteIsBound();
   void EnsureReceiverIsBound();
 
+  // TODO(b/458102489): Remove this value caching when Kabuki uses the Event
+  // exclusively.
+  std::optional<bool> last_text_to_speech_enabled_;
```
haozheng-cobalt: Even though changing the setting of Talkback is not a frequent use case, I'd still suggest to wrap this cache variable with std::atomic to handle race condition of read and write.

haozheng-cobalt (changes requested):
I tried to test this change locally, but the main branch is too broken to test A11y atm, I'll need to cherry pick the implementations to C26 first to test.

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
yell0wd0g: Ah sorry I just saw that - was busy trying to make the unittests work :)

IIUC there shouldn't be a race because updates to and reads from
`last_text_to_speech_enabled_` happen on a single TaskRunner. It
would be good to make that assumption explicit in this CL via
a ThreadChecker for example. Would you like me to do it in this
CL? (I'd say yes).

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
```diff
@@ -15,6 +15,9 @@
 #ifndef THIRD_PARTY_BLINK_RENDERER_MODULES_COBALT_H5VCC_ACCESSIBILITY_H_5_VCC_ACCESSIBILITY_H_
 #define THIRD_PARTY_BLINK_RENDERER_MODULES_COBALT_H5VCC_ACCESSIBILITY_H_5_VCC_ACCESSIBILITY_H_
 
+#include <optional>
+
+#include "base/gtest_prod_util.h"
```
yell0wd0g: Whoops this shouldn't be here.

yell0wd0g:
@haozheng-cobalt could you review (modulo the Gemini comments and my own broken include, I'll fix them before landing).

@y4vor FYI - this builds on your idea of caching the value to avoid the round trip and adds a (partial) unit test as bonus.

haozheng-cobalt:
> I tried to test this change locally, but the main branch is too broken to test A11y atm, I'll need to cherry pick the implementations to C26 first to test.

Verified this PR in C26. The improvement looks good to me. With this change browser avoids unnecessarily querying the platform TTS settings when Kabuki calls textToSpeech API. And thanks for adding the follow up TODO ticket.

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
```diff
@@ -31,7 +34,7 @@ namespace blink {
 class LocalDOMWindow;
 class ScriptState;
 
-class MODULES_EXPORT H5vccAccessibility final
+class MODULES_EXPORT H5vccAccessibility
     : public EventTarget,
       public ExecutionContextLifecycleObserver,
       public h5vcc_accessibility::mojom::blink::H5vccAccessibilityClient {
```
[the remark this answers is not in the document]
yell0wd0g: I like the `final` keyword but IIRC it prevented me from subclassing
something somewhere in the unit test? Maybe I can restate it, let me
try.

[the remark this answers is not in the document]
yell0wd0g: "final" is restored. Thanks Gemini!

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
```diff
@@ -65,10 +67,14 @@ class MODULES_EXPORT H5vccAccessibility final
   void EnsureRemoteIsBound();
   void EnsureReceiverIsBound();
 
-  // Proxy to the browser process’s H5vccAccessibilityBrowser implementation.
+  // TODO(b/458102489): Remove this value caching when Kabuki uses the Event
+  // exclusively.
+  std::optional<bool> last_text_to_speech_enabled_;
```
hlwarriner: Optional nit: `last_text_to_speech_enabled_` sounds fine but consider renaming it to `cached_text_to_speech_enabled_`. I think `cached_` more clearly captures the intent.

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
hlwarriner: Yes, please :) That's a good idea to make the assumption explicit.

And thanks for bringing this up, Hao.

yell0wd0g: Gotcha! I used WTF::IsMainThread() since ThreadChecker is
disabled in non-debug builds (like our QA or Gold). H5vcc
accessibility is not available on workers now (per IDL, IIRC).
I still need to build and run this locally to make double
sure that `H5vccAccessibility::NotifyTextToSpeechChange`
is indeed called on the main thread (and not on an IO thread).

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
yell0wd0g: Done!

third_party/blink/renderer/modules/cobalt/h5vcc_accessibility/h_5_vcc_accessibility.h
yell0wd0g: Built and run locally, and asked Gemini. All seems to
confirm that all methods on this file would run on
IsMainThread(), so we should be good here.

For completion, if we wanted to expose this IDL implementation
to Workers, this assumption won't hold, but then again
we'd need to modify the IDL when it now says `Exposed: Window`
to `Exposed: Window, Worker` so it's not trivial.

haozheng-cobalt:
This is similar idea as time zone change notification in Chromium's service: https://source.corp.google.com/h/github/youtube/cobalt/+/main:services/device/time_zone_monitor/time_zone_monitor.cc;l=43

PR #8514, opened 2025-12-29
EG: Fix GN platform_configuration for arm64/softfp

andrewsavage1:
Fixed: 472288676

### DIFF
--- a/starboard/no_platform/platform_configuration/BUILD.gn
+++ b/starboard/no_platform/platform_configuration/BUILD.gn
@@ -12,24 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import("//build/config/arm.gni")
-
 config("platform_configuration") {
-  if (current_toolchain == default_toolchain) {
-    if (current_cpu == "arm64") {
-      # TODO: b/473814802 - Use the correct configuration here once musl is fixed.
-      configs = [ ":no_platform_configuration" ]
-    } else if (current_cpu == "arm" && arm_float_abi == "softfp") {
-      configs = [ "//starboard/evergreen/arm/softfp/platform_configuration" ]
-    } else {
-      assert(false, "Unsupported architecture.")
-    }
-  } else {
-    configs = [ ":no_platform_configuration" ]
-  }
-}
-
-config("no_platform_configuration") {
   configs = [
     "//starboard/build/config/sabi",
     "//starboard/build/config:starboard",


starboard/no_platform/platform_configuration/BUILD.gn
```diff
@@ -12,7 +12,23 @@
...
+    if (current_cpu == "arm64") {
+      configs = [ "//starboard/evergreen/arm64/platform_configuration" ]
+    } else if (current_cpu == "arm" && arm_float_abi == "softfp") {
+      configs = [ "//starboard/evergreen/arm/softfp/platform_configuration" ]
+    } else {
+      assert(false, "Unsupported architecture.")
+    }
+  } else {
+    configs = [ ":no_platform_configuration" ]
+  }
```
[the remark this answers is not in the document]
andrewsavage1: While a nice idea, this would overall be worse. The variables in configuration.gni are added to the global scope, so helper variables like this should be avoided whenever possible. Additionally, the suggested code just doesn't work as variables prepended with an underscore do not get exported to GN files that import it at all.

starboard/no_platform/platform_configuration/configuration.gni
```diff
@@ -12,6 +12,20 @@
...
+    import(
+        "//starboard/evergreen/arm/softfp/platform_configuration/configuration.gni")
+  } else {
+    assert(false, "Unsupported architecture.")
+  }
+} else {
+  import("//starboard/build/config/base_configuration.gni")
+
+  sabi_path = "//starboard/sabi/x64/sysv/sabi-v$sb_api_version.json"
+}
```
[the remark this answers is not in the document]
andrewsavage1: See response above

starboard/no_platform/platform_configuration/BUILD.gn
```diff
@@ -12,7 +12,23 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//build/config/arm.gni")
+
 config("platform_configuration") {
+  if (current_toolchain == default_toolchain) {
+    if (current_cpu == "arm64") {
+      configs = [ "//starboard/evergreen/arm64/platform_configuration" ]
+    } else if (current_cpu == "arm" && arm_float_abi == "softfp") {
```
niranjanyardi: maybe outside this scope, but do you want to proactively add hardfp here ? dunno how many partners use hardfp 

else if (current_cpu == "arm" && arm_float_abi == "hardfp")

andrewsavage1: We only need the platforms that we don't have loader platforms for here

starboard/no_platform/platform_configuration/configuration.gni
```diff
@@ -12,6 +12,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import("//starboard/build/config/base_configuration.gni")
+import("//build/config/arm.gni")
```
y4vor, line 15: The  no_platform is arm based ;-)

andrewsavage1: Actually you're right, I moved the configuration picking to a shared location

PR #8515, opened 2025-12-29
Fix: Introduce backpressure to curb mapped memory growth

sherryzy:
When rapidly producing graphics commands (e.g., scrolling quickly through a video list), client-side mapped memory could grow uncontrollably. This occurred because new memory was allocated for GPU commands faster than the GPU could process them, preventing the timely release of resources.

This change introduces a backpressure mechanism in the `MappedMemoryManager`. A new configurable limit, `max_mapped_memory_in_use_bytes`, is added to cap the total in-use mapped memory.

When an allocation would exceed this limit, the manager now blocks and waits for the GPU to finish processing pending commands by calling `Finish()`. This synchronization forces the release of GPU resources, allowing the client to reclaim memory and preventing unbounded growth. A deadlock detection mechanism is also included to prevent the client from waiting indefinitely.

With this change, memory usage is stabilized under heavy load. For example, during fast video navigation, memory usage now holds steady around 750MB, whereas it previously grew beyond 1GB.

Bug: 463684608

### DIFF
--- a/gpu/command_buffer/client/mapped_memory.cc
+++ b/gpu/command_buffer/client/mapped_memory.cc
@@ -13,6 +13,7 @@
 #include "base/atomic_sequence_num.h"
 #include "base/check.h"
 #include "base/memory/ptr_util.h"
+#include "base/logging.h"
 #include "base/notreached.h"
 #include "base/numerics/checked_math.h"
 #include "base/numerics/safe_conversions.h"
@@ -49,7 +50,24 @@ MappedMemoryManager::MappedMemoryManager(CommandBufferHelper* helper,
       max_free_bytes_(unused_memory_reclaim_limit),
       max_allocated_bytes_(SharedMemoryLimits::kNoLimit),
       tracing_id_(g_next_mapped_memory_manager_tracing_id.GetNext()) {
+#if BUILDFLAG(IS_STARBOARD)
+  allocated_bytes_cleanup_threshold_ = SharedMemoryLimits::kNoLimit;
+#endif
 }
+
+#if BUILDFLAG(IS_STARBOARD)
+MappedMemoryManager::MappedMemoryManager(
+    CommandBufferHelper* helper,
+    size_t unused_memory_reclaim_limit,
+    size_t allocated_bytes_cleanup_threshold)
+    : MappedMemoryManager(helper, unused_memory_reclaim_limit) {
+  allocated_bytes_cleanup_threshold_ = allocated_bytes_cleanup_threshold;
+  if (allocated_bytes_cleanup_threshold_ != SharedMemoryLimits::kNoLimit &&
+      max_allocated_bytes_ != SharedMemoryLimits::kNoLimit) {
+    CHECK_GE(max_allocated_bytes_, allocated_bytes_cleanup_threshold_);
+  }
+}
+#endif
 
 MappedMemoryManager::~MappedMemoryManager() {
   helper_->OrderingBarrier();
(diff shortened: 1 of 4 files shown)


gpu/command_buffer/client/mapped_memory.h
```diff
@@ -129,7 +129,12 @@ class GPU_EXPORT MappedMemoryManager {
   // |unused_memory_reclaim_limit|: When exceeded this causes pending memory
   // to be reclaimed before allocating more memory.
   MappedMemoryManager(CommandBufferHelper* helper,
-                      size_t unused_memory_reclaim_limit);
+                      size_t unused_memory_reclaim_limit
+#if BUILDFLAG(IS_STARBOARD)
+                      ,
+                      size_t max_in_use_bytes
```
andrewsavage1: Add a comment explaining what this does here—which I currently don't understand. It seems to make max_allocated_bytes_ obsolete, is that true?

sherryzy: See the code comment in gpu/command_buffer/client/shared_memory_limits.h for explaining this. I will also add a comment here.

sherryzy: Added code comment below.

andrewsavage1: I saw that comment, but I still don't understand how it's used differently/in relation to max_allocated_bytes_

sherryzy: max_in_use_bytes_ serves as a soft limit to trigger backpressure, whereas max_allocated_bytes_ acts as a hard cap on the memory pool. While the hard limit effectively restricts the MappedMemoryManager footprint, reaching it causes visible thumbnail flickering—a significant regression in user experience. Transitioning to a soft-limit approach prevents this flickering while still managing memory growth. Maybe I should name it as soft_allocated_bytes_limit_for_backpressure_ ?

andrewsavage1: What about something like allocated_bytes_cleanup_threshold_? Unless backpressure is a term widely used in graphics

sherryzy: Done.

gpu/command_buffer/client/mapped_memory.cc
```diff
@@ -65,6 +76,41 @@ void* MappedMemoryManager::Alloc(unsigned int size,
...
+    while (allocated_memory_ >= allocated_bytes_cleanup_threshold_) {
+      // Before blocking, see if space has become available in existing chunks.
+      for (auto& chunk : chunks_) {
+        chunk->FreeUnused();
+        if (chunk->GetLargestFreeSizeWithoutWaiting() >= size) {
+          // We are over the total limit, but we found a free spot.
+          // This is OK, as we are re-using memory, not growing.
+          goto found_space;
+        }
+      }
```
yell0wd0g, line 106: Seeing a `goto` is strange, I'd rather put this new added code in a function and use return. Ah but I see `MemoryChunkVector` is a private typedef.

How about changing this a bit so it's more idiomatic:

```
  if (std::find_if(chunks_.begin(), chunks_.end(), [](const auto& chunk) {
        chunk.FreeUnused();
        return chunk->GetLargestFreeSizeWithoutWaiting() == size;
      }) {
    break;
  }
```

Oh but then I see that `GetLargestFreeSizeWithoutWaiting()` is likely iterating
over the container again, right? Could it make sense to run `FreeUnused()` on
every chunk first and then calling `std::find_if()` with `GetLargestFreeSizeWithoutWaiting()` ?

gpu/command_buffer/client/mapped_memory.cc
```diff
@@ -65,6 +76,41 @@ void* MappedMemoryManager::Alloc(unsigned int size,
...
+
+      // If we have already waited and made no progress, we are deadlocked.
+      if (memory_before_wait == allocated_memory_ && memory_before_wait != 0) {
```
yell0wd0g: I was a bit surprised by the `memory_before_wait != 0` here but now I understand is a guard to skip this step the first iteration of the while loop. What about

```
while (allocated_memory_ >= allocated_bytes_cleanup_threshold_) {
  // Here the |chunks_| check logic

  const auto allocated_memory_before_ = allocated_memory_;
  {
    TRACE_EVENT0("gpu", "MappedMemoryManager::Alloc::BackpressureWait");
    // This call flushes the CommandBuffer and is blocking.
    helper_->Finish();  
    FreeUnused();  // Updates |allocated_memory_|.
  }
  if (allocated_memory_before_ == allocated_memory_) {
    LOG(ERROR) << "MappedMemoryManager deadlock: waiting for GPU did not "
                  "free any memory chunks.";
    break;
  }
  
}
```

gpu/command_buffer/client/mapped_memory.cc
```diff
@@ -42,13 +43,23 @@ MemoryChunk::MemoryChunk(int32_t shm_id,
...
       max_free_bytes_(unused_memory_reclaim_limit),
       max_allocated_bytes_(SharedMemoryLimits::kNoLimit),
-      tracing_id_(g_next_mapped_memory_manager_tracing_id.GetNext()) {
+      tracing_id_(g_next_mapped_memory_manager_tracing_id.GetNext())
+#if BUILDFLAG(IS_STARBOARD)
+      ,
+      allocated_bytes_cleanup_threshold_(allocated_bytes_cleanup_threshold)
+#endif
+{
 }
```
yell0wd0g, line 56: To avoid modifying callsites (the unittests, mostly), add a new constructor with the extra `allocated_bytes_cleanup_threshold` parameter, and make it call the original 2 parameter ctor. 

In the body of that ctor add a 
```
  CHECK_GE(max_free_bytes_, allocated_bytes_cleanup_threshold);
```
to verify that the backpressure threshold is smaller than the maximum
allocated size.

yell0wd0g (approved):
Some comments. Idea looks good. Can this be upstreamed?

gpu/command_buffer/client/mapped_memory.cc
yell0wd0g: Or call `GetFreeSize()` on each `chunk`, instead of `GetLargestFreeSizeWithoutWaiting()`.

gpu/command_buffer/client/shared_memory_limits.h
```diff
@@ -45,6 +45,13 @@ struct SharedMemoryLimits {
   uint32_t mapped_memory_reclaim_limit = kNoLimit;
   uint32_t mapped_memory_chunk_size = 2 * 1024 * 1024;
   uint32_t max_mapped_memory_for_texture_upload = 0;
+#if BUILDFLAG(IS_STARBOARD)
+  // The threshold for in-flight mapped memory that triggers backpressure.
+  // When total allocated mapped memory exceeds this, the client will wait
+  // for the GPU to finish work before allocating more. This is a tunable
+  // trade-off between memory usage and performance.
+  uint32_t max_mapped_memory_in_use_bytes = 24 * 1024 * 1024;
```
andrewsavage1: I think we should name these other related variables consistently with the MappedMemoryManager member variable we're adding

sherryzy: This file contains more settings than only the ones from mapped_memory. I can rename it to a more matching name though.

sherryzy:
Yeah, I think this is something we can try to upstream, I'll send this to chromium for review.

PR #8516, opened 2025-12-30
Clean up use of android.util.Log in //cobalt.

SortaCookie:
Move usage of android.util.Log to org.chromium.base.Log. The new Log class proxies the original Log methods with additional functionality tacked on.

Bug: 472345388

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaDrmBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaDrmBridge.java
@@ -32,7 +32,6 @@ import android.os.Build;
 import android.util.Base64;
 import androidx.annotation.RequiresApi;
 import dev.cobalt.coat.CobaltHttpHelper;
-import dev.cobalt.util.Log;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
@@ -40,6 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
 import java.util.UUID;
+import org.chromium.base.Log;
 import org.jni_zero.CalledByNative;
 import org.jni_zero.JNINamespace;
 import org.jni_zero.NativeMethods;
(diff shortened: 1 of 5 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaDrmBridge.java
```diff
@@ -124,11 +124,11 @@ public static OperationResult operationFailed(String errorMessage) {
     }
 
     public static OperationResult operationFailed(String errorMessage, Throwable e) {
-      return operationFailed(String.format("%s StackTrace: %s", errorMessage, android.util.Log.getStackTraceString(e)));
+      return operationFailed(String.format("%s StackTrace: %s", errorMessage, org.chromium.base.Log.getStackTraceString(e)));
```
Awallky: Instead of the fully qualified path here, can you add the import  to the top and then just use `Log.getStackTraceString` here and elsewhere?

cobalt/android/apk/app/src/main/java/dev/cobalt/util/Log.java
```diff
@@ -41,19 +41,19 @@ private Log() {}
   private static void initLogging() {
     try {
       sLogV =
-          android.util.Log.class.getDeclaredMethod(
+          org.chromium.base.Log.class.getDeclaredMethod(
```
Awallky, line 44: Similar comment to `MediaDrmBridge`, instead of the fully qualified path here, can you add the import to the top and then just use `Log.class.getDeclaredMethod` here and elsewhere?

cobalt/android/apk/app/src/main/java/dev/cobalt/util/Log.java
```diff
@@ -94,21 +94,21 @@ private static int logWithMethod(
   }
 
   public static int v(String tag, String messageTemplate, Object... args) {
-    if (android.util.Log.isLoggable(TAG, android.util.Log.VERBOSE)) {
+    if (org.chromium.base.Log.isLoggable(TAG, org.chromium.base.Log.VERBOSE)) {
```
Awallky, line 97: Similar finding here as others above.

Awallky (approved):
LGTM % comments.
Thanks so much for doing this!

cobalt/android/apk/app/src/main/java/dev/cobalt/util/Log.java
SortaCookie: In this case not possible to do easily since we're inside a class called Log.

PR #8520, opened 2026-01-03
evergreen: Move updater initialization to browser

yuying-y:
Relocate Evergreen updater module initialization and success marking
from the shell component to the browser client component.

The updater module is now initialized upon WebContents creation within
CobaltContentBrowserClient and marked successful when the primary main
document element becomes available in CobaltWebContentsObserver. This
ensures closer integration with browser lifecycle events and more
accurate timing for marking a successful installation.

Issue: 458770751
Issue: 462442606

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -323,10 +323,10 @@ void CobaltContentBrowserClient::OnWebContentsCreated(
   VLOG(1) << "NativeSplash: Observing main frame WebContents.";
   web_contents_observer_.reset(new CobaltWebContentsObserver(web_contents));
 #if BUILDFLAG(USE_EVERGREEN)
-  // Create the updater module singleton.
+  // Create the updater module singleton if not already created.
   auto* storage_partition =
       web_contents->GetPrimaryMainFrame()->GetStoragePartition();
-  if (storage_partition) {
+  if (storage_partition and !updater::UpdaterModule::GetInstance()) {
     LOG(INFO) << "Creating UpdaterModule singleton.";
     updater::UpdaterModule::CreateInstance(
         storage_partition->GetURLLoaderFactoryForBrowserProcess(),


y4vor (approved):

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -22,4 +26,16 @@ CobaltWebContentsObserver::CobaltWebContentsObserver(
 
 CobaltWebContentsObserver::~CobaltWebContentsObserver() = default;
 
+#if BUILDFLAG(USE_EVERGREEN)
+void CobaltWebContentsObserver::PrimaryMainDocumentElementAvailable() {
```
borongc, line 30: Note we may consider to move this to Shell, as Shell is the WebContentObserver to WebApp.

We should deprecate CobaltWebContentsObserver, and unify it with Shell. Right now, the only functionality of CobaltWebContentsObserver is for managing cookies.

yuying-y: Filed b/473905515. Please fill in details in the bug about the reasoning. The move can be done when the deprecation happens.

borongc: Please check b/466120031.

borongc: @haozheng-cobalt and @Awallky have a long term goal to unify `//cobalt/shell`, and lot of customizations were landed before we copy `//content/shell` to `//cobalt/shell`. To avoid adding more works to them, we should avoid adding more functions to CobaltWebContentsObserver.

The only functionality of CobaltWebContentsObserver is for managing cookies and dispatching lifecycle events.

yuying-y: Well, in b/431862767#comment12, you suggested moving the initialization of the updater from Shell to CobaltContentBrowserClient, because as you said "in the long term we should get rid of shell". Now you're suggesting to move the implementation from CobaltWebContentsObserver back to Shell in b/466120031. Is Shell will be kept in the long run or not? Can you elaborate your reasoning in the related bugs? Also, this suggestion is not blocking this PR. Follow-up discussion and potential modification can happen in a follow-up PR.

borongc: >  you suggested moving the initialization of the updater from Shell to CobaltContentBrowserClient
Shell is WebContentsObserver, and should not own updater in terms of lifecycle management.

Shell is Content Shell's way to manage WebContentsObserver across different platforms, but for Chromium, the browser-specific features (e.g., updater in this case) should be in Content{Browser/Renderer/Gpu/Utility}Client.

> Now you're suggesting to move the implementation from CobaltWebContentsObserver back to Shell in b/466120031.

If you look at the design, Shell is a WebContentsObserver and WebContentsDelegate to the single tab's WebContents, and we have a duplicate one (CobaltWebContentsObserver), which was created at the time when we didn't modify `//content/shell`, because we won't be able to upstream these changes. As a result, we copied `//content/shell` to `//cobalt/shell`.

However, if you look at `//cobalt`, they are lot of duplicate codes (like `//cobalt/shell/browser/shell_content_browser_client.cc` and `//cobalt/browser/cobalt_content_browser_client.cc`, where they should be unified in the long term, and this is exactly what @haozheng-cobalt and @Awallky work on).

So, in short, if you want to customize browser-specific features, it should be content client. If it's WebContent-related features, it should be WebContentsObserver.

In your case, as you want to mark  updater is successful when document is ready, this is in Shell (not encourage to put it in CobaltWebContentObserver as it's duplicate code, and it is not always CobaltWebContentObserver will be WebApp, since it observes the LAST WebContents created, it could be other WebContents if we create other ones).

Therefore, there are two reasons for this change: (1) updater should be own by content client, and (2) put `PrimaryMainDocumentElementAvailable` to Shell as it's the WebContentsObserver to the single tab's WebContents. For example, storage migration is implemented [in the same place](https://source.corp.google.com/h/github/youtube/cobalt/+/2c180a559975a191c695ce7b29768189b42f599b:cobalt/shell/browser/shell.cc;l=379) as well.

It sounds like we are moving parts of functions in-and-out from Shell, but this is due to the current Cobalt Shell design, and it still has lot of TODO works.

Sorry about the confusion, and apologize if this is annoying you. Feel free to leave it as I don't think this is super high-priority. This is a design issue.

yuying-y: Thanks for the explanation. How do we view updater as a browser-specific feature or a WebContent-related feature? IMO, the updater should be owned and manipulated by one module, including initialization and marking an update as successful. But since you mentioned `CobaltWebContentsObserver::PrimaryMainDocumentElementAvailable` is triggered for every `WebContents`, I assume the splash screen's creation also triggers this? If so, it's worth considering associating the marking successful operation with the actual app page. But I'm still confused about your previous comment in b/431862767#comment12 about getting rid of the customized Cobalt shell. If so, why are we still investing in it? In summary, I strongly push for a solution to have the updater related logic managed by one place, either the content_browser_client or the shell. Open to more discussions for a solution.

PR #8524, opened 2026-01-05
[tvos] Add com.widevine.alpha key system support

abhijeetk:
Add support for the standard Widevine key system identifier
"com.widevine.alpha" to tvOS's MediaIsSupported function, aligning with
Android implementations. Android supports both "com.widevine" and
"com.widevine.alpha" via IsWidevineL1() in [1].

This change brings tvOS to parity by recognizing com.widevine.alpha
alongside the existing com.youtube.widevine.l3 and
com.youtube.widevine.forcehdcp key systems.

All YTS test cases expect "com.widevine.alpha" to be supported for EME
Conformance Tests.[2]

[1] https://github.com/youtube/cobalt/blob/main/starboard/android/shared/media_common.h#L29
[2] https://yts.devicecertification.youtube/2021/?test_type=encryptedmedia-test#1767626362127

Bug: 471365322

### DIFF
--- a/starboard/tvos/shared/media_is_supported.mm
+++ b/starboard/tvos/shared/media_is_supported.mm
@@ -14,6 +14,7 @@
 
 #include <string_view>
 
+#include "base/containers/contains.h"
 #include "starboard/common/string.h"
 #include "starboard/media.h"
 #include "starboard/tvos/shared/media/drm_system_platform.h"
@@ -28,19 +29,17 @@ bool MediaIsSupported(SbMediaVideoCodec video_codec,
     return false;
   }
 
-  constexpr std::string_view kWidevineL3SystemName{"com.youtube.widevine.l3"};
-  constexpr std::string_view kWidevineForceHdcpSystemName{
-      "com.youtube.widevine.forcehdcp"};
-  constexpr std::string_view kWidevineAlphaSystemName{"com.widevine.alpha"};
-  std::string_view key_system_sv{key_system};
+  constexpr std::string_view kSupportedWidevineSystems[] = {
+      "com.youtube.widevine.l3",
+      "com.youtube.widevine.forcehdcp",
+      "com.widevine.alpha",
+  };
 
-  if (key_system_sv == kWidevineL3SystemName ||
-      key_system_sv == kWidevineForceHdcpSystemName ||
-      key_system_sv == kWidevineAlphaSystemName) {
+  if (base::Contains(kSupportedWidevineSystems, key_system)) {
     return true;
   }
 
-  if (key_system_sv == ::starboard::DrmSystemPlatform::GetKeySystemName()) {
+  if (key_system == ::starboard::DrmSystemPlatform::GetKeySystemName()) {
     // We don't use AVPlayer for encrypted vp9.
     return video_codec != kSbMediaVideoCodecVp9;
   }


abhijeetk:

starboard/tvos/shared/media_is_supported.mm
```diff
@@ -29,13 +28,20 @@ bool MediaIsSupported(SbMediaVideoCodec video_codec,
...
+
+  for (const auto& system : kSupportedWidevineSystems) {
+    if (key_system_sv == system) {
+      return true;
+    }
   }
 
-  if (std::string_view(key_system) ==
-      ::starboard::DrmSystemPlatform::GetKeySystemName()) {
+  if (key_system_sv == ::starboard::DrmSystemPlatform::GetKeySystemName()) {
```
rakuco: I think this can be simplified by doing away with `key_system_sv` and using `base::Contains()`:

```c++
constexpr std::string_view kSupportedWidevineSystems[] = /* ... */;

if (base::Contains(kSupportedWidevineSystems, key_system)) {
  return true;
}
if (key_system == ::starboard::DrmSystemPlatform::GetKeySystemName()) {
  // ...
}
```

abhijeetk: Agree this is cleaner. I notice there's already some precedent for using
base:: utilities in starboard/tvos (av_sample_buffer_video_renderer.h
uses base::memory::weak_ptr.h), but I want to confirm, is it acceptable
for starboard code to depend on base:: libraries? Want to make sure I'm
following the right dependency patterns here.

If base:: is okay to use, I'll update to use base::Contains() which is 
definitely cleaner.

rakuco: My understanding is that it is OK to use it. Android TV and tvOS are treated differently from Linux and other platforms that use more of Starboard, and there's more leeway for using `//base`.

abhijeetk: Done.

jasonzhangxx:
tvOS doesn't support widevine l1, it can only support widevine l3. `forcehdcp` is a customized protocol which is based on widevine l3. This change will break encrypted pipeline on tvOS.

abhijeetk:
I added `com.widevine.alpha` support to enable YTS (YouTube Test Suite) EME tests to pass, which expect this standard key system to be available. However, I now understand the potential impact on YT's production pipeline. 

Are there any tests to validate existing Widevine L3 playback support?

  > tvOS doesn't support widevine l1, it can only support widevine
   l3.

  Thank you for confirming this. After analyzing the code, I can verify that tvOS always returns L3 regardless of which key
  system JavaScript selects. This is enforced at multiple levels.

  > forcehdcp is a customized protocol which is based on widevine 
  l3. This change will break encrypted pipeline on tvOS.

I understand the concern. The current HDCP enforcement logic ([drm_create_system.mm](https://github.com/youtube/cobalt/blob/main/starboard/tvos/shared/media/drm_create_system.mm#L53-L55)) only triggers for key systems containing "forcehdcp":

  ```cpp
  if (strstr(key_system, "forcehdcp")) {
    wvoec_mock::OEMCrypto_AlwaysEnforceOutputProtection();
  }
```
 When JavaScript applications detect `com.widevine.alpha` support, they might select it over `com.youtube.widevine.forcehdcp`, bypassing the HDCP enforcement YouTube expects.

Just an observation, I noticed that `com.youtube.widevine.l3` is also in the supported key systems list but doesn't trigger HDCP enforcement (no "forcehdcp" substring). Should this be addressed as well?

jasonzhangxx:
> Are there any tests to validate existing Widevine L3 playback support?

I don't think we have specific tests against it.

> When JavaScript applications detect `com.widevine.alpha` support, they might select it over `com.youtube.widevine.forcehdcp`, bypassing the HDCP enforcement YouTube expects.

YT web application should not. Or is it a behavior in Chromium? We might need double check on that.

> Just an observation, I noticed that `com.youtube.widevine.l3` is also in the supported key systems list but doesn't trigger HDCP enforcement (no "forcehdcp" substring). Should this be addressed as well?

Yes, that's intended. `widevine.l3` doesn't require HDCP.

jasonzhangxx (approved):

abhijeetk:
> > When JavaScript applications detect `com.widevine.alpha` support, they might select it over `com.youtube.widevine.forcehdcp`, bypassing the HDCP enforcement YouTube expects.
> 
> YT web application should not. Or is it a behavior in Chromium? We might need double check on that.

The key system selection is controlled by the web application(YT/YTS), not the browser/Chromium.

fayev-g (approved):

PR #8525, opened 2026-01-05
Skeleton implementation for remaining updater APIs

madhurajayaraman:
cobalt: Implement skeleton for H5vccUpdater APIs

This change introduces the IDL, Mojo interface, Blink API, and
browser-side skeleton implementations for several updater-related
APIs. Browser side changes to follow.

Bug: 454440974

idl file in 25.lts.1+ for reference:
https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/h5vcc/h5vcc_updater.idl
list of apis to be implemented in 26.eap:
go/h5vcc-interface-changes

### DIFF
--- a/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
+++ b/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
@@ -49,42 +49,42 @@ void H5vccUpdaterImpl::Create(
 
 void H5vccUpdaterImpl::GetUpdaterChannel(GetUpdaterChannelCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::SetUpdaterChannel(const std::string& channel,
                                          SetUpdaterChannelCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::GetUpdateStatus(GetUpdateStatusCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::ResetInstallations(ResetInstallationsCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::GetInstallationIndex(
     GetInstallationIndexCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::GetAllowSelfSignedPackages(
     GetAllowSelfSignedPackagesCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::SetAllowSelfSignedPackages(
     bool allow_self_signed_packages,
     SetAllowSelfSignedPackagesCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::GetUpdateServerUrl(GetUpdateServerUrlCallback callback) {
@@ -94,33 +94,33 @@ void H5vccUpdaterImpl::GetUpdateServerUrl(GetUpdateServerUrlCallback callback) {
       cobalt::updater::UpdaterModule::GetInstance()->GetUpdateServerUrl();
   std::move(callback).Run(url);
 #else
-  NOTREACHED();
+  NOTIMPLEMENTED();
 #endif
 }
 
 void H5vccUpdaterImpl::SetUpdateServerUrl(const std::string& update_server_url,
                                           SetUpdateServerUrlCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::GetRequireNetworkEncryption(
     GetRequireNetworkEncryptionCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::SetRequireNetworkEncryption(
     bool require_network_encryption,
     SetRequireNetworkEncryptionCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 void H5vccUpdaterImpl::GetLibrarySha256(unsigned short index,
                                         GetLibrarySha256Callback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  NOTREACHED();
+  NOTIMPLEMENTED();
 }
 
 }  // namespace h5vcc_updater


third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
```diff
@@ -15,8 +15,37 @@
 // Custom API for getting and setting Evergreen update settings.
 
 interface H5vccUpdater {
+  [CallWith=ScriptState, RaisesException]
+  Promise<DOMString> getUpdaterChannel();
+  [CallWith=ScriptState, RaisesException]
+  Promise<void> setUpdaterChannel(DOMString channel);
```
hlwarriner: Please change the `Promise<void>` return types to `Promise<undefined>` to align with other web interfaces. `Promise<undefined>` is preferred because of subtle differences between `undefined` and `void` in JavaScript.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
hlwarriner, line 1: We should probably bring `getLibrarySha256` forward from c25, too. I'll start a comment thread in one of our docs and we can make a decision there.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.h
```diff
@@ -50,12 +50,44 @@ class MODULES_EXPORT H5vccUpdater final
   void ContextDestroyed() override;
 
   // Web-exposed interface:
-  ScriptPromise getUpdateServerUrl(ScriptState*, ExceptionState&);
+  ScriptPromise getUpdaterChannel(ScriptState* script_state,
```
hlwarriner: For all of these methods, we should use the templated argument in the return type for better type safety.

For example, `ScriptPromise<IDLString> getUpdaterChannel` instead of `ScriptPromise getUpdaterChannel`.

See b/329702363 for more context.

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
```diff
@@ -18,5 +18,15 @@ module h5vcc_updater.mojom;
 // the renderer process can implement the H5vccUpdater Blink API.
 interface H5vccUpdater {
   // Get the update server url for the device.
```
hlwarriner: nit: remove comment since it's been separated from its method. I think it's ok to just omit the method comments here, since the methods correlate 1:1 with the web APIs.

madhurajayaraman: Sounds good, I'd kept them from 25 in case they were still necessary. Will omit.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.h
madhurajayaraman: https://source.corp.google.com/h/github/youtube/cobalt/+/26.eap:third_party/blink/renderer/modules/cobalt/h5vcc_system/h_5_vcc_system.h
https://source.corp.google.com/h/github/youtube/cobalt/+/26.eap:third_party/blink/renderer/modules/cobalt/h5vcc_experiments/h_5_vcc_experiments.h

In the other files we don't currently do this - I actually can't find a single instance of us using this structure in cobalt. I'll put up a PR once I'm done modifying this to fix the other h5vcc APIs to match.

madhurajayaraman: https://github.com/youtube/cobalt/pull/8535 for the other h5vcc apis, this pr has typing standards for the updater apis.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl
madhurajayaraman: Converting to draft while I bring this in.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.h
madhurajayaraman: https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/blink/renderer/bindings/core/v8/script_promise.h
https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/26.eap:third_party/blink/renderer/bindings/core/v8/script_promise.h

In main, we have the version of script_promise (from Chromium) that is a template class. We don't use this version in 26.eap so I don't believe we can actually use the templated types in this branch. This is something we'll have to bring into main.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
```diff
@@ -31,7 +31,133 @@ H5vccUpdater::H5vccUpdater(LocalDOMWindow& window)
...
+}
+
+ScriptPromise<IDLUndefined> H5vccUpdater::setAllowSelfSignedPackages(
+    ScriptState* script_state,
+    bool allow_self_signed_packages,
+    ExceptionState& exception_state) {
+  auto* resolver = MakeGarbageCollected<ScriptPromiseResolver>(
+      script_state, exception_state.GetContext());
+
+#if BUILDFLAG(USE_EVERGREEN)
```
yuying-y: Reminder: these APIs to allow sideloading should be guarded by macros !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING. See https://b.corp.google.com/issues/454440974#comment3 for context.

madhurajayaraman: Thanks! Gated these functions.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.h
hlwarriner: Ah, I didn't realize there was a dependency on the m138 rebase for the typed version. Thanks for clarifying.

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
```diff
@@ -14,9 +14,17 @@
...
+  GetUpdateStatus() => (string status);
+  ResetInstallations();
+  GetInstallationIndex() => (uint16 index);
+  GetAllowSelfSignedPackages() => (bool allow_self_signed_packages);
+  SetAllowSelfSignedPackages(bool allow_self_signed_packages);
   GetUpdateServerUrl() => (string update_server_url);
+  SetUpdateServerUrl(string update_server_url);
+  GetRequireNetworkEncryption() => (bool require_network_encryption);
+  SetRequireNetworkEncryption(bool require_network_encryption);
+  GetLibrarySha256() => (string sha256);
```
hlwarriner: The `getLibrarySha256` web API will need to pass its `index` param through to the Mojo method.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
```diff
@@ -31,6 +31,129 @@ H5vccUpdater::H5vccUpdater(LocalDOMWindow& window)
...
+  EnsureReceiverIsBound();
+
+  remote_h5vcc_updater_->GetUpdaterChannel(
+      WTF::BindOnce(&H5vccUpdater::OnGetUpdaterChannel, WrapPersistent(this),
+                    WrapPersistent(resolver)));
+#else
+  resolver->Reject();
+#endif
+  return resolver->Promise();
+}
```
yell0wd0g, line 49: Drive-by: All of these methods' bodies are nearly copy-pastes.
Could we have either a templatized or at least a macro solution?

If the repetition pattern was already existing before this CL
then I'd suggest filing a TODO bug and refactoring in bulk.

madhurajayaraman: This is the skeleton implementation, I haven't added the mojo implementation yet. I'll make a note of it in the bug to look at refactoring this file, but for now I'd like to keep it as is to make edits easier if needed when I add tests, and the mojo implementation. I agree that this is repetitive, and is something I'll look at once we have this fully functional.
The pattern is repeated in a couple of the other h5vcc APIs, but definitely looks the worst here due to both the number of asynchronous methods, and the buildflags.

yell0wd0g: Understood. Please file a bug and leave a TODO, and also make the CL/PR
description conform to the usual commit patterns (e.g. the title
doesn't have a prefix of the touched area, it could have "blink:").

Re.
> This is the skeleton implementation
it is common to land the blink-side and Mojo changes together, as this CL
does, and the browser- (or service-) side changes after, but I wouldn't 
expect the mojom or blink-side code to change after this CL.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -47,13 +47,80 @@ void H5vccUpdaterImpl::Create(
   new H5vccUpdaterImpl(*render_frame_host, std::move(receiver));
 }
 
+void H5vccUpdaterImpl::GetUpdaterChannel(GetUpdaterChannelCallback callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  NOTREACHED();
```
andrewsavage1: This doesn't cause any crash when trying to load Kabuki does it? I would've expected a NOTIMPLEMENTED to not actually cause a crash, but maybe Kabuki's not using any of these

madhurajayaraman: It was fine when I ran it, but now I'm nervous so I'll rerun just to confirm!

PR #8526, opened 2026-01-05
Fix string copies in the wrappers

y4vor:
starboard: Fix string copies in ABI wrappers
    
Replace memcpy with starboard::strlcpy in several POSIX ABI wrappers.
This addresses two issues: ASAN detected reading uninitialized memory
and destination strings were not guaranteed to be null-terminated when
trimmed or copied at exact buffer boundaries. Using strlcpy ensures
buffers are always null-terminated and correctly handled.
    
Bug: 473625551

### DIFF
--- a/starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
@@ -16,6 +16,7 @@
 #include <string.h>
 #include <algorithm>
 #include "starboard/common/log.h"
+#include "starboard/common/string.h"
 
 int __abi_wrap_readdir_r(musl_dir* dirp,
                          struct musl_dirent* musl_entry,
@@ -113,10 +114,13 @@ struct musl_dirent* __abi_wrap_readdir(musl_dir* dirp) {
   dirp->musl_dir_entry->d_reclen = result_platform->d_reclen;
   dirp->musl_dir_entry->d_type = result_platform->d_type;
 
-  memset(dirp->musl_dir_entry->d_name, 0, sizeof(dirp->musl_dir_entry->d_name));
-  constexpr auto minlen = std::min(sizeof(dirp->musl_dir_entry->d_name),
-                                   sizeof(result_platform->d_name));
-  memcpy(dirp->musl_dir_entry->d_name, result_platform->d_name, minlen);
+  if (starboard::strlcpy(dirp->musl_dir_entry->d_name, result_platform->d_name,
+                         sizeof(dirp->musl_dir_entry->d_name)) >=
+      sizeof(dirp->musl_dir_entry->d_name)) {
+    SB_LOG(WARNING) << "Truncated d_name in readdir wrapper."
+                    << " src_size=" << sizeof(result_platform->d_name)
+                    << " dst_size=" << sizeof(dirp->musl_dir_entry->d_name);
+  }
 
   return dirp->musl_dir_entry;
 }
--- a/starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
@@ -20,21 +20,17 @@
 #include <cstring>
 
 #include "starboard/common/log.h"
+#include "starboard/common/string.h"
 
 void CopyUtsField(const char* src, size_t src_sz, char* dest, size_t dest_sz) {
   if (src == nullptr || dest == nullptr) {
     return;
   }
 
-  if (dest_sz < src_sz) {
-    SB_LOG(WARNING) << "Wrapper utsname field size " << dest_sz
-                    << " is smaller than platform utsname field size "
-                    << src_sz;
+  if (starboard::strlcpy(dest, src, dest_sz) >= dest_sz) {
+    SB_LOG(WARNING) << "Truncated utsname field."
+                    << " src_sz=" << src_sz << " dest_sz=" << dest_sz;
   }
-
-  memset(dest, 0, dest_sz);
-  auto copy_size = std::min(src_sz, dest_sz);
-  memcpy(dest, src, copy_size);
 }
 
 SB_EXPORT int __abi_wrap_uname(struct musl_utsname* musl_uts) {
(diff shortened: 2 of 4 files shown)


starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
```diff
@@ -33,8 +34,7 @@ void CopyUtsField(const char* src, size_t src_sz, char* dest, size_t dest_sz) {
   }
 
   memset(dest, 0, dest_sz);
```
hlwarriner: Since `starboard::strlcpy` is guaranteed to NUL terminate the destination string, consider removing the `memset` call. If you're not convinced it's safe to do that then feel free to keep it.

(Please see a similar call sequence in `starboard_layer_posix_directory_abi_wrappers.cc`, too.)

starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
```diff
@@ -33,8 +34,7 @@ void CopyUtsField(const char* src, size_t src_sz, char* dest, size_t dest_sz) {
...
-  auto copy_size = std::min(src_sz, dest_sz);
-  memcpy(dest, src, copy_size);
+  starboard::strlcpy(dest, src, dest_sz);
```
hlwarriner: We should probably check the return value of all the `starboard::strlcpy` calls to check for truncation. For example here, if `dest_sz == src_sz`, the string may get truncated by a byte to make space for the NUL terminator.

andrewsavage1 (approved):
Should we add nplb tests for ensuring this behavior or directly use ASAN instead? If we clean up/categorize the ASAN issues we should be able to explicitly report an error in CI when this kind of bug occurs, at least in nplb tests.

starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
y4vor: The strlcpy should take care of adding a null terminator on trucation.

y4vor:
> Should we add nplb tests for ensuring this behavior or directly use ASAN instead? If we clean up/categorize the ASAN issues we should be able to explicitly report an error in CI when this kind of bug occurs, at least in nplb tests.

I think we should rely on ASAN. It is about memory corruption and touching outside the boundaries.

starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
hlwarriner: Yeah. My point is that we should check the return value to identify if truncation occurred so we can log an error/warning or something, instead of ignoring it. Truncation probably indicates a problem, even though the destination is guaranteed to be NUL terminated.

y4vor: The warning is covered on line 30 already. I think in your example of `dest_sz == src_sz` and having a truncation it would mean the input string was valid and not null terminated. Do you think we should still handle it?

starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
```diff
@@ -113,10 +114,8 @@ struct musl_dirent* __abi_wrap_readdir(musl_dir* dirp) {
   dirp->musl_dir_entry->d_reclen = result_platform->d_reclen;
   dirp->musl_dir_entry->d_type = result_platform->d_type;
 
-  memset(dirp->musl_dir_entry->d_name, 0, sizeof(dirp->musl_dir_entry->d_name));
-  constexpr auto minlen = std::min(sizeof(dirp->musl_dir_entry->d_name),
-                                   sizeof(result_platform->d_name));
-  memcpy(dirp->musl_dir_entry->d_name, result_platform->d_name, minlen);
+  starboard::strlcpy(dirp->musl_dir_entry->d_name, result_platform->d_name,
```
madhurajayaraman: Do we have tests that cover checking this entry? Or tests for srtlcpy generally?

madhurajayaraman: Oh just saw Andrew's comment about tests - lgtm!

madhurajayaraman (approved):

starboard/shared/modular/starboard_layer_posix_uname_abi_wrappers.cc
hlwarriner: Imagine `src` is a string containing one character, like "h". If `CopyUtsField()` is called with `src_sz=1` and `dest_sz=1`, then the warning on line 30 won't complain and `starboard::strlcpy` will end up writing only a NUL terminator to `dest`. To me, this truncation is interesting and probably worth logging as a warning/error.

There's an example in https://pubs.opengroup.org/onlinepubs/9799919799/functions/strlcpy.html of checking for truncation in this way.

If you don't feel this is worth the effort then I'm ok with skipping it, especially since `CopyUtsField` is a private function. Up to you. And let me know if you think I'm getting something wrong about the behavior.

y4vor: I'll update the code. The example you have is does use invalid string as it isn't null terminated but I think the more edge cases are covered the better.

hlwarriner: Thanks - LGTM.

In my example I meant that the string is a valid string with a NUL terminator and just one actual character, so that the actual size of the char array is 2. So CopyUtsField() *should* be called with src_sz=2, but I was thinking of the case where the caller makes a mistake and instead calls it with src_sz=1. But anyway, thanks thanks for adding the check to cover the additional edge case.

yell0wd0g:
Drive-by @y4vor : I wonder if Gemini/Gemini-CLI would have flagged this kind of potential vulnerabilities during code reviews? (I'm assuming it wouldn't since IIRC we did have Gemini code reviews back when this code was landed..?). 

If the answer is "no" - maybe now that we've identified at least one such example and associated fix(es), would it make sense to "unleash" Gemini-CLI in the `//starboard` code base to look for similar vulnerabilities?

PR #8532, opened 2026-01-06
tvOS: Add //starboard/shared/posix/system_get_error_string.cc to the build

rakuco:
PR #8442 started calling `SbSystemGetErrorString()` in shared code, so an implementation is needed for tvOS as well.

Bug: 469772944
Bug: 456692793


rakuco:

fayev-g (approved):

PR #8533, opened 2026-01-06
[starboard][widevine] Fix tvOS compilation errors in Widevine porting (1/n)

abhijeetk:
Resolve three compilation issues preventing Widevine DRM from building
on tvOS. This includes marking Registry::mutex_ as mutable to allow
locking in const methods, adding a required include for SB_DCHECK_NE,
and correcting the scope of a forward declaration. Verified on Linux to
ensure cross-platform compatibility.

These changes ensure Widevine DRM support compiles correctly on tvOS
without affecting other platforms.

Bug: 471365322

### DIFF
--- a/starboard/shared/widevine/drm_system_widevine.cc
+++ b/starboard/shared/widevine/drm_system_widevine.cc
@@ -82,7 +82,7 @@ class Registry {
   }
 
  private:
-  std::mutex mutex_;
+  mutable std::mutex mutex_;
   // Use std::vector<> as usually there is only one active instance of widevine
   // drm system.
   std::vector<SbDrmSystem> drm_systems_;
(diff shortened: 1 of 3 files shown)


abhijeetk:

starboard/shared/widevine/drm_system_widevine.cc
```diff
@@ -75,7 +75,7 @@ class Registry {
     drm_systems_.erase(iter);
   }
 
-  bool Contains(SbDrmSystem drm_system) const {
+  bool Contains(SbDrmSystem drm_system) {
```
kjyoun: I think that the better approach here is to change `mutex_` mutable.

https://stackoverflow.com/questions/3239905/c-mutex-and-const-correctness

```
mutable std::mutex mutex_;
```

abhijeetk: Thanks for review. Done.

abhijeetk:
> Can you address my comment about mutable/const?
> 
> Otherwise, LGTM

Done. Used `mutable` for member variable `mutex_` as suggested.

PR #8535, opened 2026-01-06
Make ScriptPromise typed for WebAPIs

madhurajayaraman:
Bug: 473830615


madhurajayaraman:
[source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/blink/renderer/bindings/core/v8/script_promise.h](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/blink/renderer/bindings/core/v8/script_promise.h)
[source.corp.google.com/h/lbshell-internal/cobalt_src/+/26.eap:third_party/blink/renderer/bindings/core/v8/script_promise.h](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/26.eap:third_party/blink/renderer/bindings/core/v8/script_promise.h)

In main, we have the version of script_promise (from Chromium) that is a template class. We don't use this version in 26.eap so I don't believe we can actually use the templated types in this branch. This is something we'll have to bring into main.

PR #8540, opened 2026-01-07
Enable Crashpad for hang detection and upload on Android TV

alunev:
This commit integrates Crashpad to detect and report thread hangs in Cobalt.
   When a hang is detected by `base::HangWatcher`, Crashpad now generates
   a minidump and uploads it to the Google Crash server.

   Feature will remain disabled until the internal design review is completed.

   Key changes include:

   - Initializes Crashpad in `CobaltMainDelegate::PreSandboxStartup`.
   - Implements `CobaltCrashReporterClient` to provide Cobalt-specific information to Crashpad, including product/version from `COBALT_VERSION`.
   - Configures the crash dump location within the app's cache directory.
   - Packages `libchrome_crashpad_handler.so` in the Cobalt Android APK.
   - Introduces `base::HangWatcher::Delegate` to allow Cobalt to dynamically control hang reporting via the "enableCobaltHangReporting" feature flag (set via H5vccSettings).
   - Prevents Crashpad from handling regular crash signals (SIGSEGV, etc.) in the main process, reserving those for Primes.
   - Enables BoringSSL for HTTPS transport on Android within Crashpad.
   - Includes various build system adjustments

### DIFF
--- a/base/threading/hang_watcher.cc
+++ b/base/threading/hang_watcher.cc
@@ -55,6 +55,7 @@ std::atomic<LoggingLevel> g_io_thread_log_level{LoggingLevel::kNone};
 std::atomic<LoggingLevel> g_main_thread_log_level{LoggingLevel::kNone};
 #if BUILDFLAG(IS_COBALT)
 std::atomic<LoggingLevel> g_browser_process_renderer_thread_log_level{LoggingLevel::kNone};
+static inline HangWatcher::Delegate* g_hang_watcher_delegate = nullptr;
 #endif
 
 // Indicates whether HangWatcher::Run() should return after the next monitoring.
@@ -183,6 +184,13 @@ bool ThreadTypeLoggingLevelGreaterOrEqual(HangWatcher::ThreadType thread_type,
 }
 
 }  // namespace
+
+#if BUILDFLAG(IS_COBALT)
+void HangWatcher::SetDelegate(Delegate* delegate) {
+  DCHECK(!g_instance) << "SetDelegate must be called before Start()";
+  g_hang_watcher_delegate = delegate;
+}
+#endif
 
 // Determines if the HangWatcher is activated. When false the HangWatcher
 // thread never started.
@@ -976,10 +984,18 @@ void HangWatcher::DoDumpWithoutCrashing(
   base::TimeTicks latest_expired_deadline =
       watch_state_snapshot.GetHighestDeadline();
 
-  if (on_hang_closure_for_testing_)
+  if (on_hang_closure_for_testing_) {
     on_hang_closure_for_testing_.Run();
-  else
+  } else {
+#if BUILDFLAG(IS_COBALT)
+    if (g_hang_watcher_delegate &&
+        g_hang_watcher_delegate->IsHangReportingEnabled()) {
+      RecordHang();
+    }
+#else
     RecordHang();
+#endif  // BUILDFLAG(IS_COBALT)
+  }
 
   // Update after running the actual capture.
   deadline_ignore_threshold_ = latest_expired_deadline;
@@ -1042,8 +1058,9 @@ void HangWatcher::UnregisterThread() {
 
   // Thread should be registered to get unregistered.
   DCHECK(it != watch_states_.end());
-
-  watch_states_.erase(it);
+  if (it != watch_states_.end()) {
+    watch_states_.erase(it);
+  }
 }
 
 namespace internal {
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -215,6 +215,8 @@ shared_library("libchrobalt") {
     ]
   }
   sources = [
+    "//cobalt/app/cobalt_crash_reporter_client.cc",
+    "//cobalt/app/cobalt_crash_reporter_client.h",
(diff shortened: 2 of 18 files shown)


base/threading/hang_watcher.cc
```diff
@@ -976,10 +984,17 @@ void HangWatcher::DoDumpWithoutCrashing(
   base::TimeTicks latest_expired_deadline =
       watch_state_snapshot.GetHighestDeadline();
 
-  if (on_hang_closure_for_testing_)
+  if (on_hang_closure_for_testing_) {
     on_hang_closure_for_testing_.Run();
-  else
+  } else {
+#if BUILDFLAG(IS_COBALT)
+    if (!g_hang_watcher_delegate || g_hang_watcher_delegate->IsHangReportingEnabled()) {
```
hlwarriner: Since this is a new feature I think we should flip the behavior when the embedder has not set `g_hang_watcher_delegate` so that, in this case, hang recording is not enabled. That is:

```
if (g_hang_watcher_delegate && g_hang_watcher_delegate->IsHangReportingEnabled()) {
    RecordHang();
}
```

This seems safer, and would also be consistent with the decision in the delegate to keep hang reporting disabled if no feature flag is found.

cobalt/app/cobalt_main_delegate.cc
```diff
@@ -129,6 +143,24 @@ absl::variant<int, content::MainFunctionParams> CobaltMainDelegate::RunProcess(
...
+      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
+          switches::kProcessType);
+  CobaltCrashReporterClient::Create();
+  crash_reporter::InitializeCrashpad(process_type.empty(), process_type);
+  crash_reporter::SetUploadConsent(true);
+
+  return content::ShellMainDelegate::PreSandboxStartup();
+}
+#else
+// Provide a default implementation for other platforms.
```
hlwarriner: Optional: instead, consider using the buildflag to conditionally override `PreSandboxStartup`. That seems a bit cleaner to me.

third_party/crashpad/crashpad/util/net/tls.gni
```diff
@@ -27,5 +27,6 @@ declare_args() {
   # was removed from the Fuchsia SDK. Re-enable it when we have a way to acquire
   # a BoringSSL lib again.
   crashpad_use_boringssl_for_http_transport_socket =
-      crashpad_is_in_fuchsia || (crashpad_is_linux && crashpad_is_in_chromium)
+      crashpad_is_in_fuchsia || (crashpad_is_linux && crashpad_is_in_chromium) ||
+      crashpad_is_android
```
hlwarriner, line 31: This comment is really more about lines 18/19, but `http_transport_socket.cc` is not well supported and there are likely edge cases that it does not handle very well. Let's talk more offline, but please see b/446904324 for more context.

third_party/crashpad/crashpad/util/net/http_transport_socket.cc
```diff
@@ -128,7 +128,13 @@ class SSLStream : public Stream {
         return false;
       }
     } else {
-#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS)
+#if BUILDFLAG(IS_ANDROID)
+      if (SSL_CTX_load_verify_locations(
+              ctx_.get(), nullptr, "/system/etc/security/cacerts") <= 0) {
```
hlwarriner, line 133: It may be ok to just use system certs but for improved reliability and security can consider instead using the browser's embedded certificates. This is the approach we've taken for Crashpad in Evergreen, where we pass the location of Cobalt's certificate store to the Crashpad handler process.

Cobalt on ATV now uses Chromium's embedded certs instead of Cobalt's - b/403620663 - but we could consider using these instead of system certs. Let's discuss offline and make a decision before making any changes here, though.

third_party/crashpad/crashpad/util/net/tls.gni
alunev: thanks, synced offlne, looks good to proceed for now and I've added issue for tracking this

third_party/crashpad/crashpad/util/net/http_transport_socket.cc
alunev: thanks, synced offlne, looks good to proceed for now and I've added issue for tracking this

johnxwork (approved):

cobalt/android/BUILD.gn
```diff
@@ -236,13 +238,20 @@ android_apk("cobalt_apk") {
...
     "//content/public/android:content_java",
     "//media/capture/video/android:capture_java",
     "//net/android:net_java",
     "//third_party/mesa_headers",
     "//ui/android:ui_java",
   ]
+
+  # This is the packaging strategy used by Chrome on Android L through M. In a
+  # production implementation we should consider the N through P or Q+
+  # approaches described in docs/android_native_libraries.md.
```
hlwarriner, line 252: Please file a feature request to track this. The more modern approaches would enable us to share code between binaries to reduce size.

alunev: thanks, created

PR #8549, opened 2026-01-07
net: Refactor and test private IP range checks

arjungm:
Replace custom string manipulation and bitwise logic for IP address
parsing and private range validation with utilities from the base and
net libraries. This change leverages net::IPAddress,
net::ParseURLHostnameToAddress, and net::IPAddressMatchesPrefix to
provide a more robust and streamlined implementation.

Bug: 470105792

### DIFF
--- a/third_party/blink/renderer/core/frame/csp/local_ip.cc
+++ b/third_party/blink/renderer/core/frame/csp/local_ip.cc
@@ -32,7 +32,7 @@ std::optional<net::IPAddress> ExtractIPv4From6to4(const net::IPAddress& addr) {
   return net::IPAddress(bytes[2], bytes[3], bytes[4], bytes[5]);
 }
 
-}
+}  // namespace
 
 // Queries host IP address and matches against netmask of target ip to determine.
 bool IsIPInLocalNetwork(const std::string& target_ip_str) {
--- a/third_party/blink/renderer/core/frame/csp/local_ip_test.cc
+++ b/third_party/blink/renderer/core/frame/csp/local_ip_test.cc
@@ -1,5 +1,6 @@
 #include "third_party/blink/renderer/core/frame/csp/local_ip.h"
 
+#include "base/strings/string_util.h"
 #include "net/base/ip_address.h"
 #include "testing/gtest/include/gtest/gtest.h"
 
@@ -13,9 +14,7 @@ struct TestCase {
 class IsIPInPrivateRangeTest : public testing::TestWithParam<TestCase> {};
 
 TEST_P(IsIPInPrivateRangeTest, CheckRanges) {
-  const TestCase& test_case = GetParam();
-  EXPECT_EQ(test_case.expected_result, IsIPInPrivateRange(test_case.ip_str))
-      << "Failed for IP: " << test_case.ip_str;
+  EXPECT_EQ(GetParam().expected_result, IsIPInPrivateRange(GetParam().ip_str));
 }
 
 const TestCase kTestCases[] = {
@@ -52,6 +51,16 @@ const TestCase kTestCases[] = {
 
 INSTANTIATE_TEST_SUITE_P(All,
                          IsIPInPrivateRangeTest,
-                         testing::ValuesIn(kTestCases));
+                         testing::ValuesIn(kTestCases),
+                         [](const testing::TestParamInfo<IsIPInPrivateRangeTest::ParamType>& info) {
+                           std::string ip = info.param.ip_str;
+                           bool is_private = info.param.expected_result;
+                           base::ReplaceChars(ip, "[]", "", &ip);
+                           base::ReplaceChars(ip, "[]:.", "_", &ip);
+                           if (ip == "") {
+                             ip = "empty";
+                           }
+                           return ip + "_" + (is_private ? "IsPrivate" : "IsNotPrivate");
+                         });
 
 }  // namespace


third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -94,44 +56,47 @@ bool IsIPInLocalNetwork(const std::string& target_ip_str) {
...
+    return IPAddressMatchesPrefix(address, k10, 8) ||
+           IPAddressMatchesPrefix(address, k172, 12) ||
+           IPAddressMatchesPrefix(address, k192, 16);
   }
 
-  struct in6_addr ipv6_addr;
-  if (inet_pton(AF_INET6, ip_str.c_str(), &ipv6_addr) == 1) {
-    // Check for the 2002::/16 prefix (6to4 address).
-    // The first two bytes must be 0x20 and 0x02
-    if (ipv6_addr.s6_addr[0] == 0x20 && ipv6_addr.s6_addr[1] == 0x02) {
```
jellefoks, line 113: Why did you drop the 6to4 addresses?

arjungm:
@jellefoks I had added tests previously in an earlier PR (https://github.com/youtube/cobalt/pull/8473) in these files to exercise some of this code:

* content/renderer/content_security_policy_util_unittest.cc
* services/network/public/cpp/content_security_policy/content_security_policy_unittest.cc
* third_party/blink/renderer/core/frame/csp/conversion_util_test.cc
* third_party/blink/renderer/core/frame/csp/source_list_directive_test.cc
* third_party/blink/renderer/platform/network/http_parsers_test.cc

However we do not have any existing test file for `local_ip.cc` itself and I didn't look into adding it. Do you think I should add that, or are the above tests sufficient?

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: It wasn't dropped, just moved up a few lines to make use of `ConvertIPv4MappedIPv6ToIPv4()` instead to converting it by hand. Both IPv4 and IPv6 should be covered by this function still.

jellefoks: ConvertIPv4MappedIPv6ToIPv4 is for mapped IPv6 which is a different IP range from 6to4.

arjungm: oh gotcha, fixed it then

jellefoks:
> Do we have a test for this?

If there are tests already, then can you expand them to test the behavior in the ranges affected by this code block? For example, apparently it didn't catch that the the 6to4 range got dropped.

third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -92,6 +92,10 @@ bool IsIPInPrivateRange(const std::string& raw_ip_str) {
     const net::IPAddress k172(172, 16, 0, 0);
     const net::IPAddress k192(192, 168, 0, 0);
 
+
+    bool res = IPAddressMatchesPrefix(address, k10, 8);
+    LOG(ERROR) << "ARJUN: ipv4 " << res;
```
jellefoks: Leftover debug print?

third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -58,7 +58,7 @@ bool IsIPInPrivateRange(const std::string& raw_ip_str) {
   net::IPAddress address;
 
   // Parse the string into an IPAddress object
-  if (net::ParseURLHostnameToAddress(raw_ip_str, &address)) {
+  if (!net::ParseURLHostnameToAddress(raw_ip_str, &address)) {
```
jellefoks, line 82: Aha, looks like the test found an issue in the code.

jellefoks (changes requested):

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: yup it was a good call!

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: oh dear, the shame of it all

third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -94,44 +55,49 @@ bool IsIPInLocalNetwork(const std::string& target_ip_str) {
...
-    // RFC 4193: fc00::/7 (The first 7 bits must be 1111110)
-    // This covers both fc00::/8 and fd00::/8
-    // We check the first byte: (byte & 11111110) == 11111100
-    return (ipv6_addr.s6_addr[0] & 0xFE) == 0xFC;
+  // For IPv4, the private address range is defined by RFC 1918
+  // available at https://tools.ietf.org/html/rfc1918#section-3.
+  if (address.IsIPv4()) {
+    const net::IPAddress k10(10, 0, 0, 0);
+    const net::IPAddress k172(172, 16, 0, 0);
+    const net::IPAddress k192(192, 168, 0, 0);
```
yell0wd0g: nit: Prefer constexpr since that's supported by [this ctor] (https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/openscreen/src/platform/base/ip_address.h;l=52;drc=e92b5f17ef26b70c8498b59b65974f7652855e60).

Actually this might apply to other const declarations around 
here: try constexpr.

third_party/blink/renderer/core/frame/csp/local_ip.cc
```diff
@@ -94,44 +55,49 @@ bool IsIPInLocalNetwork(const std::string& target_ip_str) {
...
-                      (static_cast<uint32_t>(ipv6_addr.s6_addr[4]) << 8)  |
-                      (static_cast<uint32_t>(ipv6_addr.s6_addr[5]));
-      return IsIPInPrivateRangeIPv4(addr);
+  //  Note: we don't use IsReserved() since it includes loopback, link-local,
+  //  multicast, and other special-use blocks.
+
+  // Handle 6to4 and IPv4-mapped addresses.
+  if (address.IsIPv6()) {
+    const auto& bytes = address.bytes();
+    if (bytes[0] == 0x20 && bytes[1] == 0x02) {
```
yell0wd0g: The comment in the LHS made sense to understand the magic numbers.
Could we recover it and give these two a name, e.g.
```
  // Check for the 2002::/16 prefix (6to4 address).
  constexpr uint8_t kBlaBlaUpperByte = 0x20;
  constexpr uint8_t kBlaBlaLowerByte = 0x02;
```
?

Apply giving names to stand along numbers throughout.

Would this potentially be an issue depending on byte
endianness? Is there no other code in Chromium or in
IPAddress supporting this kind of comparison? (Maybe
via an operator> ?)

third_party/blink/renderer/core/frame/csp/local_ip_test.cc
```diff
@@ -0,0 +1,57 @@
...
+  const char* ip_str;
+  bool expected_result;
+};
+
+class IsIPInPrivateRangeTest : public testing::TestWithParam<TestCase> {};
+
+TEST_P(IsIPInPrivateRangeTest, CheckRanges) {
+  const TestCase& test_case = GetParam();
+  EXPECT_EQ(test_case.expected_result, IsIPInPrivateRange(test_case.ip_str))
+      << "Failed for IP: " << test_case.ip_str;
```
yell0wd0g: Instead of adding this to the error, you can add a name generator,
see e.g. the documentation example:
```
INSTANTIATE_TEST_SUITE_P(
    MyInstantiation, MyTestSuite,
    testing::Values(...),
    [](const testing::TestParamInfo<MyTestSuite::ParamType>& info) {
      // Can use info.param here to generate the test suffix
      std::string name = ...
      return name;
    });
```
e.g. 
```
INSTANTIATE_TEST_SUITE_P(
    // ...
    testing::Values(kTestCases),
    [](const testing::TestParamInfo<MyTestSuite::ParamType>& info) {
      auto [ip, is_internal] = info.param;
      std::string name = 
      return ip + "_" + (is_internal ? "Internal" : :"NotInternal";
    });
```

(the generated test case name has some restrictions, not sure
whether the generated IP name would not collide).

In any case with that name generator each test case, passing or 
failing, will have a descriptive name. Then you can just have here
```
 TEST_P(IsIPInPrivateRangeTest, CheckRanges) {
   EXPECT_EQ(test_case.expected_result, IsIPInPrivateRange(GetParam().ip_str));
 }
```

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: Okay I did some more digging and it is codified here: https://github.com/youtube/cobalt/blob/main/services/network/public/cpp/ip_address_space_util.cc#L414-L425

But notably there is one difference between this implementation and the one I've put here (which replicates our legacy implementation from Cobalt 25). It treats Link-Local and `0.0.0.0/8` as Private:

* [IPv4 Link-Local](https://github.com/youtube/cobalt/blob/main/services/network/public/cpp/ip_address_space_util.cc#L258)
* [IPv6 Link-Local](https://github.com/youtube/cobalt/blob/main/services/network/public/cpp/ip_address_space_util.cc#L249)
* [0.0.0.0/8 rule](https://github.com/youtube/cobalt/blob/main/services/network/public/cpp/ip_address_space_util.cc#L271)

I swapped out the implementation and it passes the unit tests except on `{"[2002:0a00:0001::]", true}` which I'll investigate now.

third_party/blink/renderer/core/frame/csp/local_ip_test.cc
arjungm: took a bit of work, but managed to get better looking test case names :+1:

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: managed to rewrite it to use that method, but it still assumes the above IP types are Private, and potentially has to warn that it's unwrapping a 6to4 address to parse it as private.

third_party/blink/renderer/core/frame/csp/local_ip.cc
arjungm: ack

PR #8561, opened 2026-01-07
Fix syscall implementation for arm64 builds

andrewsavage1:
Fixed: 473814802

### DIFF
--- a/starboard/CHANGELOG.md
+++ b/starboard/CHANGELOG.md
@@ -28,6 +28,7 @@ The functionality is migrated to use `getauxval` instead.
 * `fchmod`
 * `fchown`
 * `fdatasync`
+* `fstatat`
 * `gai_strerror`
 * `getauxval`
 * `getcwd`
@@ -40,13 +41,13 @@ The functionality is migrated to use `getauxval` instead.
 * `isatty`
 * `kill`
 * `link`
-* `lstat`
 * `madvise`
 * `malloc_usable_size`
 * `mincore`
 * `mkdtemp`
 * `mkostemp`
 * `mkstemp`
+* `openat`
 * `pathconf`
 * `pause`
 * `pipe`
@@ -112,6 +113,11 @@ The functionality is migrated to use `getauxval` instead.
 * `unlinkat`
 * `utimensat`
 * `writev`
+
+### Removed the following POSIX Symbols:
+The following were removed in favor of more modern POSIX APIs added above.
+* `open`
+* `stat`
 
 ### From `starboard/time_zone.h`
  * Removed `SbTimeZoneGetCurrent`. The time offset is now derived


starboard/shared/modular/cobalt_layer_posix_fcntl_abi_wrappers.cc
```diff
@@ -52,4 +52,19 @@ int fcntl(int fildes, int cmd, ...) {
...
 }
 
+int __abi_wrap_openat(int fildes, const char* path, int oflag, ...);
+
+int openat(int fildes, const char* path, int oflag, ...) {
+  if ((oflag & O_CREAT) || (oflag & O_TMPFILE) == O_TMPFILE) {
+    int result;
+    va_list ap;
+    va_start(ap, oflag);
+    result = __abi_wrap_openat(fildes, path, oflag, va_arg(ap, mode_t));
```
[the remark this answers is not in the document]
andrewsavage1: This is in the cobalt layer so already is a musl tpye

starboard/shared/modular/starboard_layer_posix_fcntl_abi_wrappers.cc
```diff
@@ -297,3 +298,16 @@ SB_EXPORT int __abi_wrap_fcntl(int fildes, int cmd, ...) {
   va_end(ap);
   return result;
 }
+
+int __abi_wrap_openat(int fildes, const char* path, int oflag, ...) {
+  fildes = (fildes == MUSL_AT_FDCWD) ? AT_FDCWD : fildes;
+  if ((oflag & O_CREAT) || (oflag & O_TMPFILE) == O_TMPFILE) {
+    va_list ap;
+    va_start(ap, oflag);
+    musl_mode_t mode = va_arg(ap, mode_t);
```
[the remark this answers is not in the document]
andrewsavage1: Fixed

third_party/musl/src/fcntl/open.c
```diff
@@ -12,10 +12,5 @@ int open(const char *filename, int flags, ...)
 		mode = va_arg(ap, mode_t);
 		va_end(ap);
 	}
-
-	int fd = __sys_open_cp(filename, flags, mode);
```
andrewsavage1, line 16: @jellefoks should we be using these types of syscalls as well? It does seem to pass nplb tests if I use the upstream code instead

jellefoks, line 16: I think this is solveable in a way that doesn't need a modification to the upstream code here. Let's at least file a followup ticket for doing that.

jellefoks (approved):

third_party/musl/src/fcntl/open.c
jellefoks: Right, see third_party/musl/src/internal/syscall.h line 432, this is supported: It redirects to libc_wrapper_SYS_open and applies the necessary errno/return value conversion.

And third_party/musl/src/starboard/include/syscall_arch.h line 84, also `__syscall(SYS_fcntl)` is supported.

andrewsavage1: Great, I switched it to use the upstream code

y4vor (approved):
Please update the starboard/CHANGELOG.md too.

jellefoks (approved):

PR #8565, opened 2026-01-08
media: Pass SbWindow to StarboardRenderer

borongc:
The StarboardRenderer requires a valid SbWindow handle to initialize
the underlying SbPlayer for video rendering. Previously, the renderer
was initialized with an invalid window handle.

This change introduces a Mojo interface to enable the browser process
to pass the SbWindow handle to the renderer process. The
CobaltContentBrowserClient registers a callback to capture the
SbWindow created by PlatformWindowStarboard. The
CobaltContentRendererClient then requests this handle via Mojo and
provides it to the StarboardRenderer during its initialization.
This ensures the SbPlayer can be created with the correct platform
window, improving video rendering capabilities.

Issue: 473814226

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -41,6 +41,7 @@
 #include "cobalt/browser/user_agent/user_agent_platform_info.h"
 #include "cobalt/common/features/starboard_features_initialization.h"
 #include "cobalt/media/service/mojom/video_geometry_setter.mojom.h"
+#include "cobalt/media/service/platform_window_provider_service.h"
 #include "cobalt/media/service/video_geometry_setter_service.h"
 #include "cobalt/shell/browser/shell.h"
 #include "cobalt/shell/common/shell_paths.h"
@@ -92,6 +93,17 @@ constexpr base::FilePath::CharType kTransportSecurityPersisterFilename[] =
     FILE_PATH_LITERAL("TransportSecurity");
 constexpr base::FilePath::CharType kTrustTokenFilename[] =
     FILE_PATH_LITERAL("Trust Tokens");
+
+void BindPlatformWindowProviderService(
+    uint64_t window_handle,
+    mojo::PendingReceiver<cobalt::media::mojom::PlatformWindowProvider>
+        receiver) {
+  mojo::MakeSelfOwnedReceiver(
+      std::make_unique<cobalt::media::PlatformWindowProviderService>(
+          base::BindRepeating([](uint64_t handle) { return handle; },
+                              window_handle)),
+      std::move(receiver));
+}
 
 }  // namespace
 
@@ -153,9 +165,19 @@ CobaltContentBrowserClient::CobaltContentBrowserClient()
               nullptr,
               base::OnTaskRunnerDeleter(nullptr))) {
   DETACH_FROM_THREAD(thread_checker_);
+#if BUILDFLAG(IS_STARBOARD)
+  // TODO: b/476434249 - Revisit if Cobalt supports multiple tabs/windows.
+  ui::PlatformWindowStarboard::SetWindowCreatedCallback(
+      base::BindRepeating(&CobaltContentBrowserClient::OnSbWindowCreated,
+                          weak_factory_.GetWeakPtr()));
+#endif  // BUILDFLAG(IS_STARBOARD)
 }
 
-CobaltContentBrowserClient::~CobaltContentBrowserClient() = default;
+CobaltContentBrowserClient::~CobaltContentBrowserClient() {
+#if BUILDFLAG(IS_STARBOARD)
+  ui::PlatformWindowStarboard::SetWindowCreatedCallback(base::NullCallback());
+#endif  // BUILDFLAG(IS_STARBOARD)
+}
 
 // static
 CobaltContentBrowserClient* CobaltContentBrowserClient::Get() {
@@ -416,6 +438,27 @@ void CobaltContentBrowserClient::DispatchFocus() {
       web_contents->GetRenderViewHost()->GetWidget()->Focus();
     }
   }
+}
+
+void CobaltContentBrowserClient::AddPendingWindowReceiver(
+    mojo::PendingReceiver<cobalt::media::mojom::PlatformWindowProvider>
+        receiver) {
+  if (cached_sb_window_) {
+    BindPlatformWindowProviderService(cached_sb_window_, std::move(receiver));
+  } else {
+    pending_window_receivers_.push_back(std::move(receiver));
+  }
+}
+
+void CobaltContentBrowserClient::OnSbWindowCreated(SbWindow window) {
+  // TODO: b/476434249 - Revisit if Cobalt supports multiple tabs/windows. This
+  // assumes only single PlatformWindowStarboard() in Cobalt.
+  CHECK(!cached_sb_window_);
+  cached_sb_window_ = reinterpret_cast<uint64_t>(window);
+  for (auto& receiver : pending_window_receivers_) {
+    BindPlatformWindowProviderService(cached_sb_window_, std::move(receiver));
+  }
+  pending_window_receivers_.clear();
 }
 
 void CobaltContentBrowserClient::FlushCookiesAndLocalStorage(
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -8,6 +8,7 @@
 
 #include "base/task/bind_post_task.h"
 #include "base/time/time.h"
+#include "cobalt/media/service/mojom/platform_window_provider.mojom.h"
 #include "cobalt/renderer/cobalt_render_frame_observer.h"
 #include "cobalt/shell/common/url_constants.h"
 #include "components/cdm/renderer/widevine_key_system_info.h"
@@ -22,6 +23,9 @@
 #include "mojo/public/cpp/bindings/generic_pending_receiver.h"
 #include "starboard/media.h"
 #include "starboard/player.h"
+#include "third_party/blink/public/common/thread_safe_browser_interface_broker_proxy.h"
+#include "third_party/blink/public/platform/browser_interface_broker_proxy.h"
+#include "third_party/blink/public/platform/platform.h"
 #include "third_party/blink/public/platform/web_string.h"
 #include "third_party/blink/public/web/web_security_policy.h"
 #include "third_party/blink/public/web/web_view.h"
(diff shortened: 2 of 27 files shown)


cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -418,6 +424,32 @@ void CobaltContentBrowserClient::DispatchFocus() {
...
+  cached_sb_window_ = reinterpret_cast<uint64_t>(window);
+  for (auto& receiver : pending_window_receivers_) {
+    mojo::MakeSelfOwnedReceiver(
+        std::make_unique<cobalt::media::PlatformWindowProviderService>(
+            base::BindRepeating([](uint64_t handle) { return handle; },
+                                cached_sb_window_)),
+        std::move(receiver));
+  }
+  pending_window_receivers_.clear();
+}
```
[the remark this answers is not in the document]
borongc: Done.

ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -25,6 +26,19 @@
...
+}
+
+// static
+void PlatformWindowStarboard::SetWindowCreatedCallback(
+    WindowCreatedCallback cb) {
+  if (g_created_callback) {
+    delete g_created_callback;
+  }
+  g_created_callback = new WindowCreatedCallback(std::move(cb));
+}
```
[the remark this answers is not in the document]
borongc: Done.

andrewsavage1:
I can take a look, but is there someone on the media team you can add to review this as well?

media/base/starboard/renderer_factory_traits.h
```diff
@@ -34,6 +35,8 @@ struct MEDIA_EXPORT RendererFactoryTraits {
   base::TimeDelta audio_write_duration_remote = kNoTimestamp;
   std::string max_video_capabilities;
   gfx::Size viewport_size;
+  GetSbWindowHandleCallback get_sb_window_handle_callback =
+      base::NullCallback();
```
kjyoun: No need to initialize to NullCallback

per Gemini
> the default constructor for base::RepeatingCallback (and base::OnceCallback) automatically creates a "null" callback. Therefore, explicitly assigning base::NullCallback() is redundant.

media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -55,6 +56,7 @@ StarboardRendererClient::StarboardRendererClient(
       pending_renderer_extension_(std::move(pending_renderer_extension)),
       pending_client_extension_receiver_(std::move(client_extension_receiver)),
       client_extension_receiver_(this),
+      get_sb_window_handle_callback_(get_sb_window_handle_callback),
```
kjyoun, line 59: ?
```
std::move(get_sb_window_handle_callback) 
````

File: media/mojo/services/starboard/starboard_renderer_wrapper.h [hunk 5]
```diff
@@ -80,6 +80,7 @@ class StarboardRendererWrapper : public Renderer,
   void OnGpuChannelTokenReady(
       mojom::CommandBufferIdPtr command_buffer_id) override;
   void GetCurrentVideoFrame(GetCurrentVideoFrameCallback callback) override;
+  void OnSbWindowHandleReady(const uint64_t sb_window_handle) override;
 #if BUILDFLAG(IS_ANDROID)
```
kjyoun, line 84: const here is not necessary

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -107,6 +111,23 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+
+  if (!window_provider_.is_bound()) {
+    render_frame->GetBrowserInterfaceBroker().GetInterface(
+        window_provider_.BindNewPipeAndPassReceiver());
+
+    window_provider_->GetSbWindow(base::BindOnce(
+        [](std::atomic<uint64_t>* out_handle, uint64_t handle) {
+          *out_handle = handle;
+          if (handle) {
+            LOG(INFO) << "Renderer received SbWindow handle: "
```
kjyoun: nit: how about adopting this fast-fail pattern?

```
if (!handle) {
  LOG(ERROR) << "Renderer received invalid SbWindow handle.";
  return;
}

LOG(INFO) << "Renderer received SbWindow handle: "
                      << reinterpret_cast<void*>(handle);
*out_handle = handle;
````

File: media/mojo/clients/starboard/starboard_renderer_client_factory.cc [hunk 7]
```diff
@@ -52,6 +52,7 @@ StarboardRendererClientFactory::StarboardRendererClientFactory(
               : traits->audio_write_duration_remote),
       max_video_capabilities_(traits->max_video_capabilities),
       viewport_size_(traits->viewport_size),
+      get_sb_window_handle_callback_(traits->get_sb_window_handle_callback),
```
kjyoun, line 55: ?

std::move(traits->get_sb_window_handle_callback)

media/mojo/clients/starboard/starboard_renderer_client.cc
borongc: `get_sb_window_handle_callback` is needed for all StarboardRendererClients, so we cannot move it, as it will be null for the following StarboardRendererClients.

media/mojo/clients/starboard/starboard_renderer_client_factory.cc
borongc: Same as above.

media/base/starboard/renderer_factory_traits.h
borongc: Done.

media/mojo/services/starboard/starboard_renderer_wrapper.h
borongc: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
borongc: Done.

andrewsavage1:
Why is it that we only want to do most of this when BUILDFLAG(IS_STARBOARD) instead of also doing it for android, which also uses SbPlayer?

andrewsavage1:
Also, our default implementations of SbPlayerCreate omit a variable for the SbWindow parameter (see i.e. [here](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/shared/starboard/player/player_create.cc;l=41?q=SbPlayerCreate&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain)). I assume that needs to be changed as well?

andrewsavage1:
Also, our default implementations of SbPlayerCreate omit a variable for the SbWindow parameter (see i.e. [here](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/shared/starboard/player/player_create.cc;l=41?q=SbPlayerCreate&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain)). I assume that needs to be changed as well?

borongc:
> Why is it that we only want to do most of this when BUILDFLAG(IS_STARBOARD) instead of also doing it for android, which also uses SbPlayer?

ATV doesn't use SbWindow, and it seems our partner's implementation does use it and change the SbPlayerCreate as well. See b/474119280#comment2.

ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -37,6 +51,10 @@ PlatformWindowStarboard::PlatformWindowStarboard(
   sb_window_ = SbWindowCreate(&options);
   CHECK(SbWindowIsValid(sb_window_));
 
+  if (g_created_callback) {
+    (*g_created_callback).Run(sb_window_);
```
jasonzhangxx, line 55: If we're confident that we'll only create SbWindow once, let's add a CHECK() somewhere to ensure `g_created_callback` can be triggered once with valid SbWindow, also with a comment.

ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -25,6 +28,17 @@
...
 
+namespace {
+std::unique_ptr<PlatformWindowStarboard::WindowCreatedCallback>
+    g_created_callback;
+}
+
+// static
+void PlatformWindowStarboard::SetWindowCreatedCallback(
+    WindowCreatedCallback cb) {
+  g_created_callback = std::make_unique<WindowCreatedCallback>(std::move(cb));
```
jasonzhangxx, line 39: Can we add a CHECK() to ensure under `IS_STARBOARD` PlatformWindowStarboard::SetWindowCreatedCallback would be called only once with non null callback?

ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -37,6 +51,10 @@ PlatformWindowStarboard::PlatformWindowStarboard(
   sb_window_ = SbWindowCreate(&options);
   CHECK(SbWindowIsValid(sb_window_));
 
+  if (g_created_callback) {
```
jasonzhangxx, line 54: Also, can we add a CHECK() to ensure under `IS_STARBOARD` when creating the system window, `g_created_callback` is not null? That is to ensure the orders of ContentBrowserClient and platform windows creations are as expected.

jasonzhangxx (approved):

media/starboard/starboard_renderer.cc
```diff
@@ -493,6 +501,17 @@ void StarboardRenderer::OnVideoGeometryChange(const gfx::Rect& output_rect) {
...
 
+void StarboardRenderer::OnSbWindowHandleReady(const uint64_t sb_window_handle) {
+  CHECK(task_runner_->RunsTasksInCurrentSequence());
+  if (sb_window_handle != 0) {
+    sb_window_ = reinterpret_cast<SbWindow>(sb_window_handle);
+    if (!SbWindowIsValid(sb_window_)) {
+      LOG(WARNING) << "SbWindow is not valid.";
+    }
+  }
+  CreatePlayerBridge();
```
andrewsavage1, line 520: We would want to create the player bridge regardless of if the window is valid?

borongc: Right. C25 does the same thing. I think it's partner's job to report error to us if `SbWindow` is not valid during SbPlayerCreate().

ui/ozone/platform/starboard/platform_window_starboard.cc
borongc: Assume this will only be called once for now, but we need to revisit it if we can support multiple windows.

ui/ozone/platform/starboard/platform_window_starboard.cc
borongc: Done. Note `platform_window_starboard` is already under `IS_STARBOARD`.

ui/ozone/platform/starboard/platform_window_starboard.cc
borongc: Changed to `base::BindOnce(&CobaltContentBrowserClient::OnSbWindowCreated, weak_factory_.GetWeakPtr())`, so it will fail to CHECK(g_created_callback) when Cobalt creates the 2nd PlatformWindowStarboard.

ui/ozone/platform/starboard/platform_window_starboard.cc
borongc: Opened b/476434249 to revisit this if Cobalt supports multiple tabs on Linux platforms.

borongc: Turning this to CHECK fails many unit tests, as not all unit tests will have CobaltContentBrowserClient. Alternatively, adding CHECK in CobaltContentBrowserClient to ensure `CobaltContentBrowserClient::OnSbWindowCreated` is only called once.

PR #8568, opened 2026-01-08
Fix deadlock in fflush() for stdout/stderr/stdin

jellefoks:
Fix deadlock in V8 by synchronizing __wake with __futexwait.

This fixes a race condition in Starboard's mutex/condvar-based futex emulation where a signal could be broadcast after a waiter checked the condition but before it entered pthread_cond_wait. By acquiring and releasing the mutex in __wake, we ensure the waiter has either not started checking (will see new value) or has fully entered the wait state (will receive signal), preventing a missed wakeup that caused threads to hang indefinitely (e.g. during V8 garbage collection when fflush is called).

Bug: 473609573

### DIFF
--- a/third_party/musl/src/internal/pthread_impl.h
+++ b/third_party/musl/src/internal/pthread_impl.h
@@ -61,7 +61,11 @@ static inline void __wake(volatile void *addr, int cnt, int priv)
 	// before we signal, by acquiring the mutex. We broadcast while holding
 	// the mutex to ensure no waiter can be in the "check-then-wait" window.
 	pthread_mutex_lock(&lock->mutex);
-	pthread_cond_broadcast(&lock->cond);
+	if (cnt == 1) {
+		pthread_cond_signal(&lock->cond);
+	} else {
+		pthread_cond_broadcast(&lock->cond);
+	}
 	pthread_mutex_unlock(&lock->mutex);
 }
 


v8/src/compiler/node.cc
```diff
@@ -434,7 +434,6 @@ void Node::RemoveUse(Use* use) {
 #if DEBUG
 void Node::Verify() {
   // Check basic validity of input data structures.
-  fflush(stdout);
```
sherryzy, line 437: Shall we gate this change with a build flag? If it's only a work around, maybe add a comment or a TODO to get this code back when the root cause is identified and resolved?

jellefoks: Done

sherryzy (approved):
LGTM. Thanks for fixing the issue so fast!

v8/src/compiler/node.cc
y4vor: I think we need BUILD_FLAG(IS_STARBOARD) or BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)

aee-google (approved):
nit: remove commented out code

jellefoks:
> nit: remove commented out code

Oh, good point, this is wrong. It should just be compiled out with an ifdef.

v8/src/compiler/node.cc
kaidokert: Yes, this should be build-time gated. Makes updates / upstream conflict resolution a lot easier

third_party/musl/src/stdio/fflush.c
```diff
@@ -7,6 +7,14 @@ weak_alias(dummy, __stderr_used);
 
 int fflush(FILE *f)
 {
+#if defined(STARBOARD)
+	if (!f ||
```
jellefoks: Note: This only calls `__ensure_std_files_initialized` when it needs to lock them, because `__ensure_std_files_initialized` uses pthread_once, which is likely more expensive than this expression.

v8/src/compiler/node.cc
jellefoks: Resolved with the root cause fix.

sherryzy (changes requested):
I patched this change locally and tested a bit, the renderer unresponsive issue is still there. However, the issue is gone when disabling ```fflush(stdout)```. Can you take another look and verify if anything are missing?

kaidokert (approved):
lgtm assuming it addresses the problem - @sherryzy to confirm. Changes in musl are certainly better than in v8

sherryzy:
I verified the renderer unresponsive issue is resolved with this fix. Just some minor comments:
1. We should still gate this change with a BUILDFLAG(IS_STARBOARD). Even better shall we consider send this change to upstream?
2. GEMINI's suggestion sounds like some valid improvement. Do we want to apply that change?

third_party/musl/src/internal/pthread_impl.h
```diff
@@ -57,11 +57,12 @@ static inline void __cond_mutex_pair_destroy(StarboardPthreadCondMutex* pair) {
...
 	StarboardPthreadCondMutex* lock = (StarboardPthreadCondMutex*)(addr);
-	// Signal threads that are waiting on the condition variable. It is safe
-	// to call pthread_cond_broadcast without holding the associated mutex.
-	// Waiting threads will be woken up, and they will then attempt to
-	// re-acquire the mutex within their call to pthread_cond_wait.
+	// Ensure the waiter has entered the wait state (released the mutex)
+	// before we signal, by acquiring the mutex. We broadcast while holding
+	// the mutex to ensure no waiter can be in the "check-then-wait" window.
+	pthread_mutex_lock(&lock->mutex);
 	pthread_cond_broadcast(&lock->cond);
```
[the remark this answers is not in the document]
jellefoks: I like this suggestion. Done.

jellefoks:
> I verified the renderer unresponsive issue is resolved with this fix. Just some minor comments:
> 
> 1. We should still gate this change with a BUILDFLAG(IS_STARBOARD). Even better shall we consider send this change to upstream?
> 2. GEMINI's suggestion sounds like some valid improvement. Do we want to apply that change?

I took the suggestion, and the code block already is gated with a '#if defined(STARBOARD)' (earlier in the file).

sherryzy (approved):
Oh, I missed that this file is the internal file for starboard. LGTM.

PR #8590, opened 2026-01-09
[DRAFT] Fix compilation error when built with clang on a Windows host.

v-kryachko:
Visual C only understands cflag "/std:c++17". Remove it when built with clang.
/gemini review.


v-kryachko:
Wrong branch.

PR #8614, opened 2026-01-12
tvOS: Add an implementation for h5vcc_accessibility

rakuco:
(Note that for this to work at runtime #9219, #9221 and a commit that
enables the accessibility extension on tvOS are required, but this code
builds fine without them)

* `H5vccAccessibilityImpl::IsTextToSpeechEnabledSync()` shares its
  implementation with the Starboard version: it just queries the Starboard
  accessibility extension, and the tvOS implementation of
  `GetTextToSpeechSettings` is essentially a wrapper around the
  `UIAccessibilityIsVoiceOverRunning()` call.

* Monitoring VoiceOver status changes is done by installing an observer that
  monitors the `UIAccessibilityVoiceOverStatusDidChangeNotification`
  notification and calls
  H5vccAccessibilityManager::OnTextToSpeechStateChanged() accordingly.

This implementation does not require integration with Chromium's
accessibility infrastructure, as Kabuki simulates screen reader support by
using the Speech Synthesizer Web API.

Bug: 447135715
Bug: 478127016


cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_impl.cc
```diff
@@ -29,6 +29,12 @@
 #include "starboard/extension/accessibility.h"
 #endif
 
+#if BUILDFLAG(IS_IOS_TVOS)
+#include "cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_ax_mode_observer.h"
+#include "ui/accessibility/ax_mode.h"
+#include "ui/accessibility/platform/ax_platform.h"
+#endif  // BUILDFLAG(IS_IOS_TVOS)
```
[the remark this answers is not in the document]
rakuco: This is incorrect.

cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_ax_mode_observer.cc
```diff
@@ -0,0 +1,30 @@
+// Copyright 2026 The Chromium Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_ax_mode_observer.h"
+
+#include "cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_manager.h"
+#include "ui/accessibility/platform/assistive_tech.h"
+#include "ui/accessibility/platform/ax_platform.h"
```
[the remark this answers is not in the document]
rakuco: incorrect

rakuco:

rakuco:
@haozheng-cobalt PTAL, this is another approach to implementing this API that does not need #8368

cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_ax_mode_observer.cc
```diff
@@ -0,0 +1,30 @@
+// Copyright 2026 The Chromium Authors
```
haozheng-cobalt: To follow Cobalt's code structure standards, the tvOS-specific implementation should be placed under `starboard/tvos/shared/`. And the filename appears generic despite being only for tvOS.

You could probably reference how Android implements the accessibility functions at `starboard/android/shared/text_to_speech_helper.cc`.

haozheng-cobalt: As a follow-up: you mentioned that this change is intended to be platform-agnostic, which runs contrary to how I would expect us to implement platform-specific functionality. I would envision an implementation that directly hooks into `native tvOS UIKit` functions instead.

Could you also clarify how you tested this change locally? Does VoiceOver work as expected, and how did you verify it on the device?"

rakuco: > To follow Cobalt's code structure standards, the tvOS-specific implementation should be placed under `starboard/tvos/shared/`. You could probably reference how Android implements the accessibility functions at `starboard/android/shared/text_to_speech_helper.cc`.

I thought that'd be the case only when there was a dependency on files in `//starboard`, as tvOS-specific code landed in `//cobalt` for IFA (#8221) and Siri integration (#8424), for example. Are these guidelines documented somewhere?

> And the filename appears generic despite being only for tvOS.

Since there was nothing tvOS-specific there, I thought it'd make sense to keep it generic since it could in theory be used by other ports in the future too. Should I rename them to something like `h5vcc_accessibility_ax_mode_observer_tvos.{cc,h}`?

> As a follow-up: you mentioned that this change is intended to be platform-agnostic, which runs contrary to how I would expect us to implement platform-specific functionality. I would envision an implementation that directly hooks into `native tvOS UIKit` functions instead.

It's platform-agnostic in the sense that it's integrating with the existing Chromium infrastructure that already abstracts the platform-specific details from consumers. The "native UIKit" part is the VoiceOver stuff added in #8611. That code is responsible for monitoring VoiceOver notifications and relaying those changes to Chromium via `content::ScopedAccessibilityMode`. What we do here is just consume the notifications that Chromium sends later as a response to those messages.

> Could you also clarify how you tested this change locally? Does VoiceOver work as expected, and how did you verify it on the device?"

I've tested it with https://rakuco.github.io/h5vcc-scraps/h5vcc_a11y.html by verifying that enabling and disabling VoiceOver triggers the notifications and fires the "ontexttospeechchange" event.

haozheng-cobalt: I’m hesitant to pull in the Chromium accessibility module solely to support queries and VoiceOver change notifications. It might be overkill and go against our ongoing goal of sliming down the Cobalt binary size.

In tvOS C25, we implemented the Starboard extension and Starboard event to support the voice over functionalities. ([link](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/shared/uikit/accessibility_extension.cc;l=24?q=kStarboardExtensionAccessibilityName&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F25.lts.1%2B&start=1)). This is the same approach we are sticking to for 3P Chrobalt today,([kSbEventTypeAccessibilityTextToSpeechSettingsChanged](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/event.h;l=234?q=kSbEventTypeAccessibilityTextToSpeechSettingsChanged&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) and [kStarboardExtensionAccessibilityName](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/extension/accessibility.h;l=30?q=kStarboardExtensionAccessibilityName&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain)),

I don’t believe it is necessary to include the Chromium assistive component as a middle layer just to support two simple functionalities: querying a boolean state and registering an event listener for that state. I suggest we wire these up directly with native UIKit calls or implement the corresponding Starboard APIs and merge the tvOS path into the 3P code path.

rakuco: The Chromium a11y module is in use regardless of this PR, as it's part of what makes VoiceOver actually read the contents of the page so e.g. reverting #8611 breaks screen reader support and I don't think it's possible to remove the dependency altogether to reduce binary size in this case.

With that said, I've also tried just adding calling `UIAccessibilityIsVoiceOverRunning()` directly as well as subscribing to VoiceOver status changes instead of getting them from Chromium, and the resulting `cobalt` binary inside the bundle had the exact same size as the one generated with the code in this PR (in a gold build).

I could add a commit with this approach here if you prefer, it just won't help with the binary size or getting rid of the a11y infrastructure in Chromium.

rakuco: > I’m hesitant to pull in the Chromium accessibility module solely to support queries and VoiceOver change notifications.

This comment makes *a lot* more sense once one figures out that Kabuki is actually broken with Chromium's a11y/screen reader implementation, does not use it and emulates screen reader support by using the Speech Synthesizer Web API :-)

I have a reworked version of this patch that doesn't use the rest of Chromium and `//ui` that I will send once I land some required changes first.

rakuco: Done. @haozheng-cobalt please take another look. The commit message has been rewritten to reflect the new approach.

rakuco:
@haozheng-cobalt friendly ping. This PR has been rewritten and integrates with the speech synthesis code that landed in other PRs. It's been tested on the device with VoiceOver on and off.

yell0wd0g (approved):
LGTM to unblock since Hao is OOO for a while and the code seems good, limited to tvOS and understandable. Hao PTAL and if there're suggestions or concerns we can follow afterwards.

rakuco:
@yell0wd0g @fayev-g could either of you enable auto-merge for this PR?

PR #8615, opened 2026-01-12
Mojo implementation for updater APIs 

madhurajayaraman:
evergreen: Implement updater H5vcc APIs

This change implements the H5vcc updater APIs in H5vccUpdaterImpl.
Previously, these APIs were stubbed out with NOTIMPLEMENTED() calls.
Now, for Evergreen builds, these methods delegate to the UpdaterModule,
providing concrete functionality for managing updater channels,
installation status, and other related update features. This enables
the browser to expose full updater capabilities via the H5vcc API.

Bug: 454440974

### DIFF
--- a/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
+++ b/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
@@ -148,7 +148,7 @@ void H5vccUpdaterImpl::GetAllowSelfSignedPackages(
   }
   std::move(callback).Run(updater_module->GetAllowSelfSignedPackages());
 #else
-  std::move(callback).Run("");
+  std::move(callback).Run(false);
   NOTIMPLEMENTED();
 #endif
 }
@@ -211,7 +211,7 @@ void H5vccUpdaterImpl::GetRequireNetworkEncryption(
   }
   std::move(callback).Run(updater_module->GetRequireNetworkEncryption());
 #else
-  std::move(callback).Run("");
+  std::move(callback).Run(false);
   NOTIMPLEMENTED();
 #endif
 }


madhurajayaraman:
I need to implement GetLibrarySha256 in the updater, which will be in a follow up PR.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +49,83 @@ void H5vccUpdaterImpl::Create(
...
+#if BUILDFLAG(USE_EVERGREEN)
+  std::string channel =
+      cobalt::updater::UpdaterModule::GetInstance()->GetUpdaterChannel();
+  std::move(callback).Run(channel);
+#else
   NOTIMPLEMENTED();
+#endif
 }
 
 void H5vccUpdaterImpl::SetUpdaterChannel(const std::string& channel,
```
yuying-y, line 66: Can you refer to the implementation of C25? https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/25.lts.1+:cobalt/h5vcc/h5vcc_updater.cc;l=44

For example, this SetUpdaterChannel() has significant logic to check the channel and do different things. It doesn't simply call updater->SetUpdaterChannel(channel)

Check the other functions as well.

yuying-y (commented):
For all functions, can you do null check after calling cobalt::updater::UpdaterModule::GetInstance()?

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +49,83 @@ void H5vccUpdaterImpl::Create(
 
 void H5vccUpdaterImpl::GetUpdaterChannel(GetUpdaterChannelCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(USE_EVERGREEN)
+  std::string channel =
+      cobalt::updater::UpdaterModule::GetInstance()->GetUpdaterChannel();
+  std::move(callback).Run(channel);
```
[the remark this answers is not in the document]
hlwarriner: I prefer what gemini suggested but consider it optional / up to you.

hlwarriner (commented):
I agree with Ying's comments about the nullptr checks and capturing any logic from the corresponding layer in c25 but I have no further feedback / everything else looks good.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: Yes, done.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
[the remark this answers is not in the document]
madhurajayaraman: Added!

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +54,115 @@ void H5vccUpdaterImpl::Create(
...
+#if BUILDFLAG(USE_EVERGREEN)
+  auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
+  if (updater_module) {
+    // Force an update if SetUpdaterChannel is called, even if
+    // the channel doesn't change, as long as it's not
+    // "prod"/"experiment"/"control" channel
+    if (updater_module->GetUpdaterChannel().compare(channel) != 0) {
+      updater_module->SetUpdaterChannel(channel);
+    } else if (channel == "prod" || channel == "experiment" ||
+               channel == "control") {
```
yuying-y: can you also add "rollback" to the list? Sorry it was missed in c25.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -108,26 +183,51 @@ void H5vccUpdaterImpl::GetUpdateServerUrl(GetUpdateServerUrlCallback callback) {
...
+#else
   NOTIMPLEMENTED();
+#endif
 }
 
 void H5vccUpdaterImpl::SetRequireNetworkEncryption(
     bool require_network_encryption,
     SetRequireNetworkEncryptionCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
```
yuying-y, line 223: Should BUILDFLAG(USE_EVERGREEN) also be checked? At other places as well.

yuying-y (approved):
Lgtm with two minor comments

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: Will discuss in meeting today, but the current proposed implementation plan is to move the binding to be conditional on `USE_EVERGREEN`, as otherwise web tests don't work. Wouldn't `ALLOW_EVERGREEN_SIDELOADING` already be conditional on `USE_EVERGREEN`? Either way, we check for `USE_EVERGREEN` separately - that change is in https://github.com/youtube/cobalt/pull/8637 as I realized the binding issue when writing the web tests.

yuying-y: Sounds good. For now, `ALLOW_EVERGREEN_SIDELOADING` is set false. When the feature is approved, this macro will be removed, instead of guarded by `USE_EVERGREEN`.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +50,116 @@ void H5vccUpdaterImpl::Create(
...
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(USE_EVERGREEN)
+  auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
+  if (!updater_module) {
+    std::move(callback).Run("");
+    return;
+  }
+  std::move(callback).Run(updater_module->GetUpdaterChannel());
+#else
   NOTIMPLEMENTED();
```
yell0wd0g, line 62: In general, Mojo callbacks need to be `Run()` or they would CHECK-crash.
So here we should have 
```
  std::move(callback).Run("");
  NOTIMPLEMENTED();
```

Same elsewhere in the file.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +50,116 @@ void H5vccUpdaterImpl::Create(
...
 
 void H5vccUpdaterImpl::SetUpdaterChannel(const std::string& channel,
                                          SetUpdaterChannelCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(USE_EVERGREEN)
+  auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
+  if (updater_module) {
+    // Force an update if SetUpdaterChannel is called, even if
+    // the channel doesn't change, as long as it's not
+    // "prod"/"experiment"/"control" channel
```
yell0wd0g: nit: In comments, mention functions/methods with (), and
escape variable names, i.e.
```
  // Force an update if SetUpdaterChannel() is called, even if
  // the |channel| doesn't change...
```

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +50,116 @@ void H5vccUpdaterImpl::Create(
...
   NOTIMPLEMENTED();
+#endif
 }
 
 void H5vccUpdaterImpl::GetInstallationIndex(
     GetInstallationIndexCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if BUILDFLAG(USE_EVERGREEN)
+  const uint16_t kInvalidInstallationIndex = 1000;
+  auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
```
yell0wd0g, line 126: If this is a singleton, how can it fail? If it can't fail,
then instead of handling it as nullptr, leave a `CHECK()`.
Sme comment throughout.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +50,116 @@ void H5vccUpdaterImpl::Create(
...
+  if (!updater_module) {
+    std::move(callback).Run(kInvalidInstallationIndex);
+    return;
+  }
+  int index = updater_module->GetInstallationIndex();
```
yell0wd0g, line 131: nit: const? Apply throughout.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +50,116 @@ void H5vccUpdaterImpl::Create(
...
+  std::move(callback).Run(index == -1 ? kInvalidInstallationIndex
+                                      : static_cast<uint16_t>(index));
```
yell0wd0g: nit: Prefer `base::checked_cast<uint6_t>(index)`

yell0wd0g (changes requested):
Looking good with a few nits, will LGTM after they
are addressed. Is there a unittest for this?

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: The updater module instance isn't available until `UpdaterModule::CreateInstance()` is called - currently done in `CobaltContentBrowserClient::OnWebContentsCreated()`.

I think it's reasonable to program this defensively and return `kInvalidInstallationIndex` here, as is currently done. For example, the web API could have been called prematurely.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: Thanks for catching this - agreed, let's fix it.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
hlwarriner: From the Google C++ style guide, "Using const on local variables is neither encouraged nor discouraged." Unless a Cobalt or Chromium style guide overrides that, I think it should be up to author discretion.

hlwarriner (approved):
LGTM aside from Miguel's comment about running the Mojo callbacks before `NOTIMPLEMENTED`.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: This becomes obsolete once we make the binding changes, I can add this in for now but seems redundant since it's going to almost immediately be deleted.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: Ying had asked me to implement the defensive programming. I initially didn't have this check implemented. I agree with Holden that it's better to have this check defensively to prevent premature calling.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
yell0wd0g: I don't fully understand -- what I meant is that if
`callback` reaches the end of the function and has not
been `std::move()`d, it will reach a `(D)CHECK`, because
all Mojo callbacks must be `Run()`. Just to clarify.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
yell0wd0g: If `GetInstance()` can return nullptr, then it makes sense to 
handle that case, yeah.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -49,42 +50,123 @@ void H5vccUpdaterImpl::Create(
...
+      return;
+    }
+    updater_module->CompareAndSwapForcedUpdate(0, 1);
+    updater_module->RunUpdateCheck();
+  }
+  std::move(callback).Run();
+#else
+  std::move(callback).Run();
   NOTIMPLEMENTED();
+#endif
```
yell0wd0g, line 89: Instead of having multiple `std::move(callback).Run()`, you could
use `base::ScopedClosureRunner`, which will make sure `callback`
is `Run()` on every return path (unless it's run before). For
example in this method the body would be:
```
  base::ScopedClosureRunner scoped_callback(std::move(callback));
#if BUILDFLAG(USE_EVERGREEN)
  // The same, but without std::move(callback).Run(); references.
  // Instead of the std::move(callback).Run(blabla), you'd do
  // std::move(scoped_callback).Run(blabla).
```

All in all, shorter!

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: Yes that's fine, I've added it so that this PR is clean when it goes in.
There's a follow up change that deletes the code (we get rid of the if/else block) but I definitely shouldn't introduce something broken even if I'm about to delete it :)

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
madhurajayaraman: Agreed, but if you look at https://github.com/youtube/cobalt/pull/8774 we get rid of the if/else blocks in favor of moving the gates to the mojo binding, and having a disconnect handler for the else case so we end up in a much cleaner state shortly! I added the second `std::move(callback).Run()` call to ensure we weren't introducing broken code even if the intention is to almost immediately clean it up. If we still think `base::ScopedClosureRunner` makes sense, I can introduce it - but can also revisit at the end of all of this work when we look at templatizing the web APIs (they're all basically the exact same code in different trenchcoats)

madhurajayaraman: No wait you're so right, updating.

madhurajayaraman: I cannot figure out why, but it's failing locally when I use `base::ScopedClosureRunner` ; theoretically it shouldn't but I don't want to spend more time debugging this so will submit as is.

PR #8616, opened 2026-01-12
Play splash video from Main app's cache

linxinan-yt:
PR extracts the splash video from the local storage cache, in which the main app's service worker could decide the splash video they want to show to users.

Issue: 454630524


linxinan-yt:
Close this and move the thread to #8342

PR #8622, opened 2026-01-13
Revert "cobalt/build: Add cobalt browsertests to monolithic Linux builds."

Awallky:
Reverts youtube/cobalt#8619

The tests are still running for some reason, causing CI failures - https://github.com/youtube/cobalt/actions/runs/20939228360/job/60172726340?pr=8578.

Bug: 433354983


cobalt/build/testing/targets/linux-x64x11/test_targets.json
```diff
@@ -14,7 +14,6 @@
     "base:base_unittests",
     "cc:cc_perftests",
     "cc:cc_unittests",
-    "cobalt:cobalt_browsertests",
```
linxinan-yt, line 17: I've thought it is only building the target, but seems it also tries to run the tests.
Currently there are some navigation tests are failing on Linux. What about android?

yell0wd0g (approved):
RS LGTM since this is a clean revert. (Consider adding the Bug: line in the description to get a ping in the bug.)

cobalt/build/testing/targets/linux-x64x11/test_targets.json
Awallky: Android works fine. I have seen runs where it skips the browser tests due to the test filters - https://github.com/youtube/cobalt/actions/runs/20940487416/job/60178844136?pr=8622

I'm not sure why there is a difference on Linux, since the filters look the same. I noticed that there is a _x64 build where these builds were failing on Linux - https://github.com/youtube/cobalt/actions/runs/20939228360/job/60172726340?pr=8578.

PR #8623, opened 2026-01-13
Fix cobalt_browsertests not execute test code

linxinan-yt:
This CL fixes an issue where cobalt_browsertests were not executing their test bodies on Starboard.

**Root Cause**: 
On Starboard (similar to Android and iOS), ShellMainDelegate::RunProcess initializes the BrowserMainRunner but returns 0 early, skipping the call to BrowserMainRunner::Run().
   * The test body (ui_task) is normally executed within BrowserMainLoop::InterceptMainMessageLoopRun.
   * On standard platforms, InterceptMainMessageLoopRun is called during Run().
   * Since Run() is skipped on Starboard, the test body was never executed.

  **Fix**:
  Updated BrowserMainLoop::CreateStartupTasks to include BUILDFLAG(IS_STARBOARD).
   * This explicitly schedules InterceptMainMessageLoopRun as a startup task.
   * This ensures the ui_task is executed synchronously during BrowserMainRunner::Initialize(), matching the behavior required by ShellMainDelegate for these platforms.

  **Additional Changes**:
   * Added assertions to ContentMainRunnerImplBrowserTest.StartupSequence in content_main_runner_impl_browsertest.cc to verify that FeatureList, ThreadPoolInstance, and ContentClient are correctly initialized after startup.

Issue: 433354983

### DIFF
--- a/content/browser/browser_main_loop.cc
+++ b/content/browser/browser_main_loop.cc
@@ -895,7 +895,10 @@ void BrowserMainLoop::CreateStartupTasks() {
 // is entered and startup tasks are run asynchronously from it.
 // InterceptMainMessageLoopRun() thus needs to be forced instead of happening
 // from MainMessageLoopRun().
-#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_IOS) || BUILDFLAG(IS_STARBOARD)
+// Make the Starboard same behavior like Android but only in non-release
+// builds for cobalt_browsertests.
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_IOS) || \
+    (!defined(OFFICIAL_BUILD) && BUILDFLAG(IS_STARBOARD))
   StartupTask intercept_main_message_loop_run = base::BindOnce(
       [](BrowserMainLoop* self) {
         // Lambda to ignore the return value and always keep a clean exit code


cobalt/testing/browser_tests/content_main_runner_impl_browsertest.cc
```diff
@@ -297,6 +297,9 @@ class ContentMainRunnerImplBrowserTest : public ContentBrowserTest {
 
 IN_PROC_BROWSER_TEST_F(ContentMainRunnerImplBrowserTest, StartupSequence) {
   // All of the work is done in SetUp().
+  EXPECT_TRUE(base::FeatureList::GetInstance());
+  EXPECT_TRUE(base::ThreadPoolInstance::Get());
+  EXPECT_TRUE(GetContentClientForTesting());
```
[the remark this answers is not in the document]
linxinan-yt: There is no EXPECT_NOT_NULL in chromium code base.

content/browser/browser_main_loop.cc
```diff
@@ -895,7 +895,7 @@ void BrowserMainLoop::CreateStartupTasks() {
 // is entered and startup tasks are run asynchronously from it.
 // InterceptMainMessageLoopRun() thus needs to be forced instead of happening
 // from MainMessageLoopRun().
-#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_IOS)
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_IOS) || BUILDFLAG(IS_STARBOARD)
```
linxinan-yt: I think this should work for starboard too. But if this impact the production, we can also wrap it with the BUILDFLAG(IS_BROWSERTEST).

haozheng-cobalt: Please do not use BUILDFLAG(IS_BROWSERTEST), as we plan to remove this guard in the future, you could use `#if !defined(OFFICIAL_BUILD)` for debug and debug.

linxinan-yt: Nice idea. Ty!

PR #8627, opened 2026-01-13
media: Enable protected media permission for tvOS

abhijeetk:
This enables protected media identifier permission for tvOS.
Without this, tvOS renderer processes crash with BAD_MESSAGE
when requesting Widevine DRM permissions. PermissionServiceImpl
cannot convert PermissionName to PermissionType, causing a
ReceivedBadMessage() and renderer termination.

Bug: 475154230
Bug: 471365322

### DIFF
--- a/third_party/blink/common/BUILD.gn
+++ b/third_party/blink/common/BUILD.gn
@@ -2,6 +2,7 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+import("//build/config/apple/mobile_config.gni")
 import("//build/config/features.gni")
 import("//build/config/rust.gni")
 import("//build/nocompile.gni")
@@ -343,7 +344,8 @@ source_set("common") {
     deps += [ "//ui/gfx/" ]
   }
 
-  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia || is_starboard) {
+  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia || is_starboard ||
+      (is_ios && target_platform == "tvos")) {
     defines = [ "ENABLE_PROTECTED_MEDIA_IDENTIFIER_PERMISSION" ]
   }
 }


abhijeetk:
cc: @Gyuyoung @jkim-julie @rakuco

third_party/blink/common/BUILD.gn
```diff
@@ -343,7 +343,8 @@ source_set("common") {
     deps += [ "//ui/gfx/" ]
   }
 
-  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia || is_starboard) {
+  if (is_chromeos || is_android || is_win || is_castos || is_fuchsia || is_starboard ||
+      (is_ios && target_platform == "tvos")) {
```
rakuco, line 348: For completeness, I'd explicitly import `//build/config/apple/mobile_config.gni` for `target_platform` to avoid relying on it being set by other files.

abhijeetk: Done! Yes, it make sense. Thanks for the review.

abhijeetk:
Hi all, quick reminder on this PR when you get a chance. Thanks!

fayev-g (approved):

third_party/blink/common/BUILD.gn
```diff
@@ -343,7 +344,8 @@ source_set("common") {
...
```
y4vor, line 348: We didn't add `is_tvos`?

rakuco: No, it existed for a brief period of time when we were enabling tvOS support upstream, but the OWNERS there ultimately opted for using the `target_platform` variable instead (https://chromium-review.googlesource.com/c/chromium/src/+/6306066)

fayev-g: I've been ok with this convention because it makes explicit that tvos support is a subset of ios support. (Some things can be applied to is_ios generally.) We can theoretically have a higher level wrapper - I'd just prefer consistency either way.

abhijeetk: Agreed with the reviewers. Using `is_ios && target_platform == "tvos"` is consistent with the established convention and keeps tvOS correctly scoped under iOS.

If there are no further concerns, can we go ahead and land this PR?

fayev-g: Thanks for confirming Kubo and Abhijeet!

PR #8630, opened 2026-01-13
Add test Finch features.

yjzhang111:
Introduce a dedicated feature and parameter for internal end-to-end testing of the Finch (field trial and variations) system. This enables verification that Finch configurations, including feature enablement and parameter distribution, are correctly applied and reported within Cobalt builds.

UMA histograms are added to track the feature's status and the parameter's value. This provides observability to confirm the integrity of the Finch pipeline without impacting user-facing functionality or existing features.

Bug: 456555897


cobalt/browser/features.cc
```diff
@@ -22,5 +22,13 @@ BASE_FEATURE(kExperimentConfigExpiration,
              "ExperimentConfigExpiration",
              base::FEATURE_DISABLED_BY_DEFAULT);
 
+// Test feature, does not enable any features.
+BASE_FEATURE(kTestFinchFeature,
+             "TestFinchFeature",
+             base::FEATURE_DISABLED_BY_DEFAULT);
+
+const base::FeatureParam<std::string> kTestFinchFeatureParam{
+    &kTestFinchFeature, "TestFinchFeatureParam", ""};
```
AlecsisDuarte, line 31: nit: Could you add a parameter name for the empty string? Just to make it clearer on the reader why the value is empty. 

See go/c-readability-advice#comment-argument-literals-with-their-parameter-names.

AlecsisDuarte (approved):
LGTM! Thanks

cobalt/browser/features.cc
yjzhang111: Seems like it's a common practice in chromium to leave default value blank
https://source.chromium.org/search?q=base::featureparam%3Cstd::string%3E%20f:.cc

PR #8642, opened 2026-01-14
evergreen: Fix DCHECK crash at UpdaterModule initialization

yuying-y:
Replace DCHECKs in UpdaterModule::CreateInstance and GetInstance
with null checks. This allows multiple calls to CreateInstance to
be safely ignored and GetInstance to return nullptr if not yet
initialized. Add a null check in H5vccUpdaterImpl to handle cases
where the UpdaterModule is not available, preventing a crash.

Bug: 475635275

### DIFF
--- a/cobalt/updater/updater_module.cc
+++ b/cobalt/updater/updater_module.cc
@@ -170,9 +170,11 @@ void UpdaterModule::CreateInstance(
 }
 
 UpdaterModule* UpdaterModule::GetInstance() {
-  LOG_IF(WARNING, !updater_module_)
-      << "UpdaterModule is not created yet, and cannot be "
-         "retrieved by UpdaterModule::GetInstance().";
+  if (!updater_module_) {
+    LOG(WARNING) << "UpdaterModule is not created yet, and cannot be "
+                    "retrieved by UpdaterModule::GetInstance().";
+    return nullptr;
+  }
   return updater_module_->get();
 }
 


yuying-y:
> ## Code Review
> This pull request aims to fix a `DCHECK` crash that occurs when `UpdaterModule::GetInstance()` is called before the module is created. The approach taken is to replace the `DCHECK` with a warning log and allow `GetInstance()` to be called at any time.
> 
> While the changes in `h5vcc_updater_impl.cc` correctly handle a potential `nullptr` return from `GetInstance()`, the implementation of `GetInstance()` itself in `updater_module.cc` is flawed. It logs a warning but then proceeds to dereference a null pointer, which will cause a crash in all builds, not just debug builds.
> 
> I've left a critical comment with a suggested fix to ensure `GetInstance()` safely returns `nullptr` when the module is not initialized, which will resolve the crash and fulfill the intent of this PR.

The original approach doesn't cause a crash by calling updater_module_->get(), because updater_module_ is a smart pointer. But the Gemini suggested approach is more clear, so it's adopted.

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
```diff
@@ -90,9 +90,16 @@ void H5vccUpdaterImpl::SetAllowSelfSignedPackages(
...
 #if BUILDFLAG(USE_EVERGREEN)
-  std::string url =
-      cobalt::updater::UpdaterModule::GetInstance()->GetUpdateServerUrl();
-  std::move(callback).Run(url);
+  auto updater_module = cobalt::updater::UpdaterModule::GetInstance();
+  if (updater_module) {
+    std::string url = updater_module->GetUpdateServerUrl();
+    std::move(callback).Run(url);
+  } else {
+    LOG(ERROR)
```
y4vor, line 98: Is there a race condition or are we creating it too late in the startup sequence?
Also would this lead to problems?

yuying-y: No, there's no identified race condition. This is only a safety check. The behavior is the same as C25 https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/25.lts.1+:cobalt/h5vcc/h5vcc_updater.cc;l=121

y4vor (approved):

PR #8657, opened 2026-01-14
Add an "Echo" platform service on linux for tests

hlwarriner:
This is a simple service that implements the contract and features provided by the PlatformService Starboard extension. It is intended to support e2e testsing of the H5vccPlatformService web interface.

Issue: 371419798

### DIFF
--- a/starboard/linux/shared/echo_service.cc
+++ b/starboard/linux/shared/echo_service.cc
@@ -1,4 +1,4 @@
-// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -42,15 +42,15 @@ struct ThreadArgs {
 // through Send(). It responds synchronously by returning the data from Send()
 // and asynchronously using its receive_callback.
 typedef struct EchoServiceImpl : public PlatformServiceImpl {
-  pthread_mutex_t m;
-  pthread_cond_t cv;
+  pthread_mutex_t shutdown_mutex;
+  pthread_cond_t shutdown_cond;
   bool should_shutdown;
   std::vector<pthread_t> workers;
 
   EchoServiceImpl(void* context, ReceiveMessageCallback receive_callback)
       : PlatformServiceImpl(context, receive_callback), should_shutdown(false) {
-    pthread_mutex_init(&m, NULL);
-    pthread_cond_init(&cv, NULL);
+    pthread_mutex_init(&shutdown_mutex, NULL);
+    pthread_cond_init(&shutdown_cond, NULL);
   }
 
   // Default constructor.
@@ -58,8 +58,8 @@ typedef struct EchoServiceImpl : public PlatformServiceImpl {
 
   ~EchoServiceImpl() {
     CleanUpWorkers();
-    pthread_mutex_destroy(&m);
-    pthread_cond_destroy(&cv);
+    pthread_mutex_destroy(&shutdown_mutex);
+    pthread_cond_destroy(&shutdown_cond);
   }
 
   static void* AsyncResponseTrampoline(void* args) {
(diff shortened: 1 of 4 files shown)


hlwarriner:
Cross referencing https://github.com/youtube/cobalt/pull/8685, which is now up for review.

andrewsavage1:
Most of the gemini comments seem reasonable, could we address those?

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,199 @@
...
+           uint64_t length,
+           uint64_t* output_length,
+           bool* invalid_state) {
+  SB_DCHECK(service);
+  SB_DCHECK(data);
+  SB_DCHECK(output_length);
+
+  char message[length + 1];
+  std::memcpy(message, data, length);
+  message[length] = '\0';
```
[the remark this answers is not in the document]
hlwarriner: I was following the pattern used by other linux service implementations but I think this makes sense. I added a comment to note the assumption that this service will only be sent data representing strings.

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,199 @@
...
+  // Rather than manage a thread pool we just create a new worker thread for
+  // each Send() request. This should be fine since this service is only used
+  // for testing.
+  static_cast<EchoServiceImpl*>(service)->StartAsyncResponseThread(async_response);
+
+  // Return the sync response.
+  *output_length = sync_response.length();
+  auto ptr = malloc(*output_length);
+  sync_response.copy(reinterpret_cast<char*>(ptr), sync_response.length());
+  return ptr;
```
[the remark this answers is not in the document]
hlwarriner: done

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,199 @@
...
+
+struct ThreadArgs {
+  PlatformServiceImpl* platform_service_instance;
+  std::string response;
+};
+
+// This platform service is for testing only and simply echoes data sent to it
+// through Send(). It responds synchronously by returning the data from Send()
+// and asynchronously using its receive_callback.
+typedef struct EchoServiceImpl : public PlatformServiceImpl {
```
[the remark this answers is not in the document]
hlwarriner: I'd prefer to remain consistent with the base struct and other linux implementations, like `PreAppRecommendationsPlatformServiceImpl`.

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,199 @@
...
+    pthread_mutex_unlock(&m);
+
+    for (pthread_t worker : workers) {
+      pthread_join(worker, nullptr);
+    }
+
+    workers.clear();
+  }
+
+} EchoServiceImpl;
```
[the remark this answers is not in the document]
hlwarriner: N/A - see https://github.com/youtube/cobalt/pull/8657/files#r2695976348.

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,199 @@
...
+};
+
+// This platform service is for testing only and simply echoes data sent to it
+// through Send(). It responds synchronously by returning the data from Send()
+// and asynchronously using its receive_callback.
+typedef struct EchoServiceImpl : public PlatformServiceImpl {
+  pthread_mutex_t m;
+  pthread_cond_t cv;
+  bool should_shutdown;
+  std::vector<pthread_t> workers;
```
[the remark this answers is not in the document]
hlwarriner: I'll keep the POSIX thread types. I think it's fine since this is a minimal implementation used only for testing. Engineers working Starboard are also experienced using these primitives, for example in `starboard/nplb/posix_compliance`.

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,199 @@
...
+    pthread_mutex_init(&m, NULL);
+    pthread_cond_init(&cv, NULL);
+  }
+
+  // Default constructor.
+  EchoServiceImpl() = default;
+
+  ~EchoServiceImpl() {
+    CleanUpWorkers();
+  }
```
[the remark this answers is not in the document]
hlwarriner: done

hlwarriner:
> Most of the gemini comments seem reasonable, could we address those?

Addressed all the comments. PTAL.

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,207 @@
...
+struct ThreadArgs {
+  PlatformServiceImpl* platform_service_instance;
+  std::string response;
+};
+
+// This platform service is for testing only and simply echoes data sent to it
+// through Send(). It responds synchronously by returning the data from Send()
+// and asynchronously using its receive_callback.
+typedef struct EchoServiceImpl : public PlatformServiceImpl {
+  pthread_mutex_t m;
```
andrewsavage1: Are these (m and cv) typical abbreviations? I think the code would be easier to read if they were spelled out

starboard/linux/shared/platform_service.cc
```diff
@@ -40,7 +41,8 @@ struct cmp_str {
 
 std::map<const char*, const void* (*)(), cmp_str> platform_service_registry = {
     {kSoftMicPlatformServiceName, GetSoftMicPlatformServiceApi},
-    {kPreappRecommendationServiceName, GetPreappRecommendationServiceApi}};
+    {kPreappRecommendationServiceName, GetPreappRecommendationServiceApi},
+    {kEchoServiceName, GetEchoServiceApi}};
```
andrewsavage1: Should we consider disabling this service in release builds? Or do you think it's unnecessary?

starboard/linux/shared/echo_service.cc
```diff
@@ -0,0 +1,207 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
andrewsavage1: nit: copyright years are wrong

andrewsavage1 (approved):

starboard/linux/shared/echo_service.cc
hlwarriner: Thanks for catching that.

starboard/linux/shared/platform_service.cc
hlwarriner: That's probably a good idea, mostly from a binary size perspective.

starboard/linux/shared/echo_service.cc
hlwarriner: In my experience they're fairly typical, yeah. The more important thing IMO is to name the mutex / cv variables after the resources / signals they control. But I'd say that's less important when a class/module only has one of each, since there's not much for the reader to keep track of. Anyway, I went ahead and renamed these variables as `shutdown_mutex` and `shutdown_cond` - hope that  helps.

andrewsavage1: Yes I think that's better, thanks!

PR #8660, opened 2026-01-14
cobalt: Force URL reload after storage migration

borongc:
After storage migration, the browser previously relied on
RenderFrameHost::Reload(). This could lead to an unreliable refresh
of content or navigation to the correct URL.

This change ensures a robust restart by explicitly loading the last
committed URL. Using NavigationController::LoadURLWithParams with
specific transition types guarantees the application reloads
correctly. This prevents issues with stale data or incomplete
reloads, ensuring proper state and content display. Web contents
are also marked as shown for correct rendering post-migration.

Issue: 474447216
Issue: 463395742

### DIFF
--- a/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
+++ b/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
@@ -136,14 +136,23 @@ Task LogElapsedTimeTask() {
 }
 #endif  // !defined(COBALT_IS_RELEASE_BUILD)
 
-Task ReloadTask(content::WeakDocumentPtr weak_document_ptr) {
+Task ReloadTask(content::WeakDocumentPtr weak_document_ptr, const GURL& url) {
   return base::BindOnce(
-      [](content::WeakDocumentPtr weak_document_ptr,
+      [](content::WeakDocumentPtr weak_document_ptr, const GURL& url,
          base::OnceClosure callback) {
-        RenderFrameHost(weak_document_ptr)->Reload();
+        auto* rfh = RenderFrameHost(weak_document_ptr);
+        content::WebContents* web_contents =
+            content::WebContents::FromRenderFrameHost(rfh);
+        if (web_contents) {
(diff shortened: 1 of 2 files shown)


cobalt/shell/browser/shell.cc
```diff
@@ -995,6 +995,7 @@ void Shell::SwitchToMainWebContents() {
     has_switched_to_main_frame_ = true;
     if (web_contents_) {
       GetPlatform()->UpdateContents(this);
+      web_contents_->WasShown();
```
johnxwork, line 998: I think you left out its corresponding `WasHidden()` intentionally right? Could you explain why we would not need that anymore?

johnxwork (approved):

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -136,14 +136,23 @@ Task LogElapsedTimeTask() {
...
+        auto* rfh = RenderFrameHost(weak_document_ptr);
+        content::WebContents* web_contents =
+            content::WebContents::FromRenderFrameHost(rfh);
+        if (web_contents) {
+          LOG(INFO) << "MigrationManager: Reloading URL: " << url.spec();
+          content::NavigationController::LoadURLParams params(url);
+          params.transition_type = ui::PageTransitionFromInt(
+              ui::PAGE_TRANSITION_TYPED | ui::PAGE_TRANSITION_FROM_ADDRESS_BAR);
+          web_contents->GetController().LoadURLWithParams(params);
+        }
```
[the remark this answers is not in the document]
borongc: URL should be valid in this case.

cobalt/shell/browser/shell.cc
borongc: It's still there. I just added `WasShown()` to pair it, but the visitibility should be updated in `GetPlatform()->UpdateContents(this)` from java side already.

PR #8674, opened 2026-01-15
Merge rdk repo into 26.eap

niranjanyardi:
This PR merges the RDK Starboard implementation directly into the Cobalt source tree.

Working with a separate RDK repository has caused frequent CI breakages and developer confusion. By integrating the RDK repository directly:

We improve CI stability.
We simplify the developer workflow.
We enable more consistent upstreaming of changes.
Additionally, this PR configures .pre-commit-config.yaml to exclude the starboard/contrib/rdk directory from pre-commit hooks, as it is treated as a third-party repository.

Bug: 474137620

### DIFF
--- a/starboard/.gitignore
+++ b/starboard/.gitignore
@@ -6,4 +6,3 @@ xcuserdata
 keyboxes
 keyboxes-dev
 android/apk/.gradle/
-contrib/rdk/


starboard/contrib/rdk/plugin/CobaltImplementation.cpp
```diff
@@ -0,0 +1,812 @@
...
+        Core::SystemInfo::SetEnvironment(_T("COBALT_CONTENT_DIR"), kDefaultContentDir);
+      }
+
+      {
+        string envVal, gstDebug = "gstplayer:4,2";
+        if (Core::SystemInfo::GetEnvironment(_T("GST_DEBUG"), envVal) && !envVal.empty())
+          gstDebug = "," + envVal;
+        if (config.GstDebug.IsSet() == true)
+          gstDebug += "," + config.GstDebug.Value();
+        Core::SystemInfo::SetEnvironment(_T("GST_DEBUG"), gstDebug);
```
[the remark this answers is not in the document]
niranjanyardi: 3p code, ignore

niranjanyardi:
@denisyuji / @santoshmahto7  we are merging the rdk repo into github for faster development (easily running on-device RDK tests ) . we 'll have to cherry-pick the changes manually to gerrit at a someaht regular cadence. let's discuss that in our next meeting

DEPS
```diff
@@ -3983,11 +3983,12 @@ deps = {
     'condition': 'checkout_mac',
   },
 
-  # Dependencies for RDK (starboard/contrib/rdk)
-  'src/starboard/contrib/rdk': {
-      'url': Var('rdk_starboard_git') + '/external/components/generic/cobalt' + '@' + '27.lts.youtube',
-      'condition': 'checkout_linux',
-  },
+# Cobalt: imported
```
andrewsavage1: This should just be deleted

niranjanyardi: https://g3doc.corp.google.com/company/teams/cobalt/team/infrastructure/code_repository/add_new_repo.md#push-the-repository-to-the-cobalt-repository-as-a-subtree 

atleast instructions says to leave it commented . should we change that @briantting . this file has some other repos with same pattern

andrewsavage1: That's because those instructions are for upstream repos, not a repo we own

niranjanyardi: this is an upstream repo- RDK owns it https://github.com/rdkcentral/larboard . We currently have our own branch for development but long term we are merging our changes into their branches and planning on using that .  Even the current changes are based of their 25.lts.xxxx dev branch . Like i created 27.lts.youtube/ 25.lts.youtube based of their dev branches

andrewsavage1: Further clarification—the instructions are for repos that come from Chromium upstream, because they are the ones that put them in the DEPS file. We don't want to just remove them from the DEPS file because we want to be clear that that's not what Chromium does. OTOH we are not beholden to any particular method of importing the rdk repo, so should not follow the same pattern as we do for Chromium imported repos.

briantting: I think the confusion here comes from the fact that we would normally comment out in DEPS a repo we're adding, but here we were also the ones that added this repo to DEPS in the first place. So we would remove it rather than comment it out since this does not exist in the upstream Chromium DEPS.

niranjanyardi: gotcha, will delete it

oxve (approved):

PR #8685, opened 2026-01-15
h5vcc: add e2e impl of platform services for 3P

hlwarriner:
This change ports forward the H5vccPlatformService web interface from c25. A requirement from our web clients is that we keep the interface unchanged from c25. There is on exception: because of an issue with the IDL compiler, the type of the data param in the ReceiveCallback is relaxed.

The requirement to use Mojo between the renderer and browser necessitates a fairly different implementation vs c25. Reviewers may want to look at the Mojo interfaces first to understand the implementation.

The changes are all guarded with !is_androidtv buildflags. Android TV will migrate from JavaBridge to web IDL in b/403638702.

The web interface and implementation have been manually tested e2e, with the help of #8657. Web tests are complicated by the synchronous Mojo methods and will be added separately.

This was partially #vibe-coded.

Issue: 371419798

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
@@ -234,7 +234,17 @@ void H5vccPlatformService::ContextDestroyed() {
 }
 
 void H5vccPlatformService::OnDataReceived(const WTF::Vector<uint8_t>& data) {
-  if (!receive_callback_ || !GetExecutionContext()) {
+  if (!receive_callback_) {
+    DLOG(WARNING) << "H5vccPlatformService::OnDataReceived: dropping the "
+                  << "message for " << service_name_ << " because the JS "
+                  << "callback is not set. The service was probably closed.";
+    return;
+  }
+
+  if (!GetExecutionContext()) {
+    DLOG(WARNING) << "H5vccPlatformService::OnDataReceived: dropping the "
+                  << "message for " << service_name_ << " because the "
+                  << "ExecutionContext is no longer valid.";
     return;
   }
 
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.idl
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.idl
@@ -12,6 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// This interface mirrors the H5vccPlatformService interface from 25.lts.stable:
+// https://github.com/youtube/cobalt/blob/25.lts.stable/cobalt/h5vcc/h5vcc_platform_service.idl.
+// To adhere to web and Chromium best practices we may consider changing these
+// APIs to be asynchronous (return promises). At the time of writing it was
+// deemed too expensive to migrate existing web clients to async APIs.
 [
   Exposed=Window,
   SecureContext,


hlwarriner:
CC @johnxwork

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
```diff
@@ -0,0 +1,92 @@
...
+  delete this;
+}
+
+void H5vccPlatformServiceManagerImpl::Has(const std::string& service_name,
+                                          HasCallback callback) {
+  auto platform_service_extension =
+      static_cast<const CobaltExtensionPlatformServiceApi*>(
+          SbSystemGetExtension(kCobaltExtensionPlatformServiceName));
+  if (!platform_service_extension) {
+    LOG(WARNING) << "The extension is not implemented on this platform";
```
andrewsavage1: Let's add the service_name to the log to make it clearer

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
```diff
@@ -0,0 +1,251 @@
...
+  DLOG(ERROR) << "PlatformService connection error for " << service_name_;
+  close();
+}
+
+DOMArrayBuffer* H5vccPlatformService::send(DOMArrayBuffer* data,
+                                           ExceptionState& exception_state) {
+  if (!service_opened_ || !platform_service_remote_.is_bound()) {
+    exception_state.ThrowDOMException(DOMExceptionCode::kInvalidStateError,
+                                      "Service is diconnected or not open.");
+    return nullptr;
```
zhongqiliang, line 191: This behavior is changed from C25 here https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/h5vcc/h5vcc_platform_service.cc;l=104?q=h5vcc_pla&sq=git:github%2Fyoutube%2Fcobalt@25.lts.1%2B. C25 return 0 buffer.

return DOMArrayBuffer::Create(nullptr, static_cast<size_t>(0)); will be the same as C25

But maybe nullptr is fine as it already throw the javascript exception.

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
```diff
@@ -0,0 +1,251 @@
...
+  WTF::Vector<uint8_t> input_data = ToWTFVector(data);
+  WTF::Vector<uint8_t> response_data;
+
+  bool mojo_result = platform_service_remote_->Send(input_data, &response_data);
+
+  if (!mojo_result) {
+    exception_state.ThrowDOMException(DOMExceptionCode::kInvalidStateError,
+                                      "Mojo call failed, connection lost.");
+    close();
+    return nullptr;
```
zhongqiliang, line 216: same

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
```diff
@@ -0,0 +1,190 @@
...
+    LOG(WARNING) << "PlatformService is not implemented on this platform.";
+    std::move(callback).Run({});  // Return empty vector on error
+    return;
+  }
+
+  uint64_t output_length = 0;
+  std::vector<uint8_t> mutable_data = data;
+  void* data_ptr = mutable_data.empty() ? nullptr : mutable_data.data();
+  uint64_t data_length = mutable_data.size();
+  bool invalid_state = 0;
```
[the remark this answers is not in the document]
zhongqiliang: +1

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
```diff
@@ -0,0 +1,190 @@
...
+  void* response_ptr = api->Send(platform_service_, data_ptr, data_length,
+                                 &output_length, &invalid_state);
+
+  if (invalid_state) {
+    LOG(ERROR) << "Send failed: Starboard service in invalid state for "
+               << service_name_;
+    if (response_ptr) {
+      free(response_ptr);
+    }
+    std::move(callback).Run({});  // Return empty vector on error
```
zhongqiliang: It did not run callback on C25. https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/h5vcc/h5vcc_platform_service.cc;l=123-129?q=h5vcc_pla&sq=git:github%2Fyoutube%2Fcobalt@25.lts.1%2B

cobalt/browser/h5vcc_platform_service/public/mojom/h5vcc_platform_service.mojom
```diff
@@ -0,0 +1,49 @@
...
+
+// Interface implemented by the browser-side H5vccPlatformService instance
+// to handle messages from the renderer process.
+interface PlatformService {
+  // Sends data to the service and optionally receives an immediate response.
+  //
+  // TODO: b/371419798 - consider using mojo_base.mojom.BigBuffer for more
+  // efficient transport, as the data and response data could be large.
+  [Sync]
+  Send(array<uint8> data) => (array<uint8> response_data);
```
zhongqiliang: invalid_state needs to be exposed on this API, so that the renderer can throw a javascript exception when the JSON bytes Kabuki sent do not align with the protocol.

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
```diff
@@ -0,0 +1,79 @@
...
+  void Send(const std::vector<uint8_t>& data, SendCallback callback) override;
+
+ private:
+  PlatformServiceImpl(
+      content::RenderFrameHost& render_frame_host,
+      const std::string& service_name,
+      mojo::PendingRemote<mojom::PlatformServiceObserver> observer,
+      mojo::PendingReceiver<mojom::PlatformService> receiver);
+  bool OpenStarboardService();
+  void OnDataReceivedFromStarboard(const std::vector<uint8_t>& data);
```
zhongqiliang, line 65: Android does not have Starboard, can we rename this method?

cobalt/browser/h5vcc_platform_service/public/mojom/BUILD.gn
```diff
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 # Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
andrewsavage1: nit: the copyright year is wrong on most of these files

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
andrewsavage1: Agreed, OpenService/OnDataReceived seem fine

andrewsavage1: Actually, scratch that—wouldn't Android also be using Starboard for its platform services implementation?

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
```diff
@@ -0,0 +1,92 @@
...
+    OpenCallback callback) {
+  if (!render_frame_host_.IsActive()) {
+    LOG(ERROR)
+        << "H5vccPlatformServiceManagerImpl::Open: RenderFrameHost not active";
+    return;
+  }
+
+  PlatformServiceImpl::Create(render_frame_host_, service_name,
+                              std::move(observer), std::move(receiver));
+  std::move(callback).Run();
```
zhongqiliang, line 90: C25 does not run callback immediately when the service is opened. https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/android/shared/platform_service.cc;l=59-83?q=CobaltExtensionPlatformServicePrivate&sq=git:github%2Fyoutube%2Fcobalt@25.lts.1%2B

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
```diff
@@ -0,0 +1,190 @@
...
+  bool invalid_state = 0;
+
+  void* response_ptr = api->Send(platform_service_, data_ptr, data_length,
+                                 &output_length, &invalid_state);
+
+  if (invalid_state) {
+    LOG(ERROR) << "Send failed: Starboard service in invalid state for "
+               << service_name_;
+    if (response_ptr) {
+      free(response_ptr);
```
andrewsavage1: nit: checking if the response_ptr is non-null here and below is unnecessary

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
```diff
@@ -0,0 +1,251 @@
...
+  ScriptState* script_state = ToScriptStateForMainWorld(GetExecutionContext());
+  if (!script_state || !script_state->ContextIsValid()) {
+    return;
+  }
+  ScriptState::Scope scope(script_state);  // For RAII
+
+  DOMUint8Array* dom_array = DOMUint8Array::Create(base::span(data));
+  v8::Local<v8::Value> v8_data =
+      ToV8Traits<DOMUint8Array>::ToV8(script_state, dom_array);
+  ScriptValue script_data(script_state->GetIsolate(), v8_data);
```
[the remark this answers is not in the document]
hlwarriner: done

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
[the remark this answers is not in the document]
hlwarriner: done

cobalt/browser/h5vcc_platform_service/public/mojom/BUILD.gn
hlwarriner: thanks - updated

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
hlwarriner: This error case is service-agnostic since it means the platform service extension itself is not implemented on the platform. That said, `H5vccPlatformServiceManagerImpl::Has` is of course invoked in order to check about a particular platform service. Do you still want to add the service name to the log?

andrewsavage1: Ah, no then let's just be more specific and say "The platform service extension is not implemented on this platform."

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
hlwarriner: Won't Android TV likely have its own implementation of the `mojom::PlatformService` interface, likely in Java?

If so, I'd like to keep the `OpenStarboardService`/`OnDataReceivedFromStarboard` names to make it clear which layer these methods are dealing with.

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
hlwarriner: sounds good - done

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
zhongqiliang: There is a thing call starboard bridge, which was migrated from C25, there is no starboard layer under the bridge.

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
hlwarriner: I believe that in both c25 and this current implementation in main the return value is irrelevant: it's not actually returned to the JavaScript caller since the JS exception is thrown and must be caught. And according to Gemini, returning the `nullptr` is preferred since it doesn't require a wasted memory allocation for the unused buffer.

I'll do more research / testing to confirm this and then add an implementation comment to the code to clarify.

cobalt/browser/h5vcc_platform_service/public/mojom/h5vcc_platform_service.mojom
hlwarriner: I changed `response_data` to be an optional (nullable) value. A response with no value indicates to the renderer that data could not be sent to the platform service.

PTAL.

See https://chromium.googlesource.com/chromium/src/mojo/+/HEAD/public/tools/bindings/README.md#primitive-types for information about the optional type.

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
hlwarriner: I confirmed that with either `return nullptr` or `return DOMArrayBuffer::Create(nullptr, static_cast<size_t>(0))`, the return value is `undefined` in JavaScript after the exception is caught.

Also, even if the return value were observable, I would have hoped web clients weren't relying on a 0 buffer since this behavior isn't part of the API contract in the IDL.

I kept the return value as `nullptr` and added comment to clarify.

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
hlwarriner: Thanks - this is a great catch. In the latest commit we now use `std::nullopt` in the callback here to signal to the renderer that there was an issue so that it can then raise the expected exception.

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
andrewsavage1: There is a Starboard layer on ATV, i.e. implementations for media-related functions (there are also a handful of other things that are more scattered). It doesn't use Starboard for system calls, but it still has to use Starboard for behavior specific to Cobalt, which I would say this falls under

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
hlwarriner: Just to be clear, this callback is not the `ReceiveCallback` from JS. This `OpenCallback` from the Mojo interface method is run, after `PlatformServiceImpl::Create` and `PlatformServiceImpl::OpenStarboardService` have attempted to synchronously open the underlying platform service and wire things up, to unblock the renderer thread and allow `H5vccPlatformService::open()` to return a `H5vccPlatformService` JS object to the web client.

Please lmk if you still see an issue or have any questions.

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
hlwarriner: Oh yeah, I see  hat now in https://pubs.opengroup.org/onlinepubs/009604499/functions/free.html. Thanks.

hlwarriner:
> It's worth asking @fayev-g to look to make sure this wouldn't cause any issues for tvos (I assume they are not using JS injection for platform services the same way ATV is, but want to check)

Based on b/460683731, I believe tvOS will be fine to use the IDL (ported forward, unchanged from c25) and Mojo framework introduced by this PR. @fayev-g, should we add kubo@ as a reviewer to confirm?

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
```diff
@@ -0,0 +1,267 @@
...
+  observer_receiver_.reset();
+  receive_callback_ = nullptr;
+}
+
+void H5vccPlatformService::ContextDestroyed() {
+  close();
+}
+
+void H5vccPlatformService::OnDataReceived(const WTF::Vector<uint8_t>& data) {
+  if (!receive_callback_ || !GetExecutionContext()) {
```
zhongqiliang: should we add a W level log here? 
on platform service async mode, if Kabuki close the platform service before the data come back, that data will be lost.

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
zhongqiliang: got it!

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
zhongqiliang: we can come back to this after ATV implementation is finished, and see how it goes.

zhongqiliang (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
hlwarriner: In reality these checks are probably excessively defensive: in both of these cases, the receiving end of the Mojo pipe in the renderer should be gone and `H5vccPlatformService::OnDataReceived` should not run. I still think it's good to have them, though, in case the events happen in an unexpected order, and I added warnings logs as you suggested.

cobalt/browser/h5vcc_platform_service/platform_service_impl.h
hlwarriner: Sounds good

PR #8690, opened 2026-01-15
starboard: Refactor WidevineTimer to use JobThread

kjyoun:
This change simplifies the threading model within WidevineTimer.
It replaces manual pthread management and custom synchronization with
the JobThread abstraction. This streamlines thread lifecycle management
and reduces boilerplate for asynchronous operations.

Bug: 468356861

### DIFF
--- a/starboard/shared/widevine/widevine_timer.cc
+++ b/starboard/shared/widevine/widevine_timer.cc
@@ -23,12 +23,16 @@
 namespace starboard {
 
 WidevineTimer::~WidevineTimer() {
+  std::lock_guard lock(mutex_);
+  if (job_thread_) {
+    // Flush any pending tasks before deleting |active_clients_|. This ensures
+    // that background tasks (like those scheduled by cancel()) that capture
+    // iterators into |active_clients_| finish before the data is destroyed.
+    // TODO: b/477902972 - Call stop method instead, once it's added.
+    job_thread_->ScheduleAndWait([] {});
+  }
   for (auto iter : active_clients_) {
     delete iter.second;
-  }
-  if (job_thread_) {
-    // Flush any pending tasks before job_thread_ is destroyed.
-    job_thread_->ScheduleAndWait([] {});
   }
 }
 
--- a/starboard/shared/widevine/widevine_timer.h
+++ b/starboard/shared/widevine/widevine_timer.h
@@ -41,8 +41,9 @@ class WidevineTimer : public ::widevine::Cdm::ITimer {
 
  private:
   std::mutex mutex_;
-  std::unique_ptr<JobThread> job_thread_;
-  std::map<IClient*, JobQueue::JobOwner*> active_clients_;
+  std::unique_ptr<JobThread> job_thread_;  // Guarded by |mutex_|.
+  std::map<IClient*, JobQueue::JobOwner*>
+      active_clients_;  // Guarded by |mutex_|.
 };
 
 }  // namespace starboard


starboard/shared/widevine/widevine_timer.cc
```diff
@@ -63,24 +31,18 @@ void WidevineTimer::setTimeout(int64_t delay_in_milliseconds,
...
+  SB_CHECK(job_thread_);
 
   auto iter = active_clients_.find(client);
   if (iter == active_clients_.end()) {
-    iter = active_clients_.emplace(client, new JobQueue::JobOwner(job_queue_))
-               .first;
+    iter =
+        active_clients_
+            .emplace(client, new JobQueue::JobOwner(job_thread_->job_queue()))
+            .first;
```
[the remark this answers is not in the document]
kjyoun: it's done by formatter

starboard/shared/widevine/widevine_timer.cc
```diff
@@ -95,53 +58,19 @@ void WidevineTimer::cancel(IClient* client) {
...
   }
 
-  SB_DCHECK(job_queue_);
+  SB_CHECK(job_thread_);
 
-  WaitEvent wait_event(mutex_);
-  job_queue_->Schedule([&]() { CancelAllJobsOnClient(client, &wait_event); });
-  wait_event.Wait(lock);
+  job_thread_->job_queue()->ScheduleAndWait(
+      [this, client] { CancelAllJobsOnClient(client); });
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/shared/widevine/widevine_timer.cc
```diff
@@ -95,53 +59,28 @@ void WidevineTimer::cancel(IClient* client) {
...
-}
-
-void WidevineTimer::CancelAllJobsOnClient(IClient* client,
-                                          WaitEvent* wait_event) {
-  SB_DCHECK(wait_event);
-  SB_DCHECK(job_queue_->BelongsToCurrentThread());
+void WidevineTimer::CancelAllJobsOnClient(IClient* client) {
+  SB_CHECK(job_thread_->BelongsToCurrentThread());
 
   {
```
rakuco: just a style nit: this nested scope is not needed anymore.

starboard/shared/widevine/widevine_timer.cc
```diff
@@ -95,53 +59,28 @@ void WidevineTimer::cancel(IClient* client) {
...
 
-  WaitEvent wait_event(mutex_);
-  job_queue_->Schedule([&]() { CancelAllJobsOnClient(client, &wait_event); });
-  wait_event.Wait(lock);
+  std::condition_variable cv;
+  bool done = false;
+  job_thread_->job_queue()->Schedule([&] {
+    CancelAllJobsOnClient(client);
+    {
+      std::lock_guard<std::mutex> lock(mutex_);
```
rakuco: Is it really necessary to guard `done` with a mutex? I don't see how it could be changed by multiple threads.

starboard/shared/widevine/widevine_timer.cc
kjyoun: Done

starboard/shared/widevine/widevine_timer.cc
kjyoun: Yes, 

Per gemini
> Actually, the mutex is necessary here to prevent a 'lost wakeup.' If we don't guard done with the same mutex used by the condition_variable, there's a risk that the main thread checks the flag, the worker thread updates it and notifies, and then the main thread goes to sleep anyway—missing the signal entirely. It also prevents a data race on the done variable itself.

rakuco: I think this is a theoretical issue at this point since `CancelAllJobsOnClient()` blocks on the mutex which inevitably makes this block be hit only once `cv` below is in a wait state, but I can see how that's just a side effect and not very safe in the long term.

With that said, can't this be achieved by calling `ScheduleAndWait()`?

kjyoun: Done. Refactored code to use `ScheduleAndWait`

starboard/shared/widevine/widevine_timer.cc
```diff
@@ -18,139 +18,63 @@
...
-};
-
-}  // namespace
-
 WidevineTimer::~WidevineTimer() {
   for (auto iter : active_clients_) {
     delete iter.second;
   }
+  if (job_thread_) {
+    // Flush any pending tasks before job_thread_ is destroyed.
```
rakuco: Is this to protect against the callback in the `ScheduleAndWait()` call in `cancel()` being invoked after `active_clients_` has been destroyed if you just rely on `job_thread_` being destroyed instead of flushing?

Also related, I wonder if the access to `active_clients_` above needs to be protected with a mutex?

kjyoun: Good point

Protected dtor with mutex

PR #8698, opened 2026-01-15
cobalt: Fix initial deep link for multiple webContents

zhongqiliang:
The initial deep link URL retrieval is now performed within the
initialDeepLink() getter method. Previously, it was fetched during
H5vccRuntime construction.

This change addresses an issue with multiple WebContents instances.
When a splash screen WebContent was present, it would consume the
initial deep link, preventing the main application WebContent (Kabuki)
from receiving it. Fetching on demand ensures each WebContent gets its
correct initial deep link.

Bug: 475657034

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_runtime/h_5_vcc_runtime.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_runtime/h_5_vcc_runtime.cc
@@ -24,11 +24,7 @@ namespace blink {
 H5vccRuntime::H5vccRuntime(LocalDOMWindow& window)
     : ExecutionContextLifecycleObserver(window.GetExecutionContext()),
       remote_h5vcc_runtime_(window.GetExecutionContext()),
-      deep_link_receiver_(this, window.GetExecutionContext()) {
-  EnsureRemoteIsBound();
-  remote_h5vcc_runtime_->GetAndClearInitialDeepLinkSync(&initial_deep_link_);
-  LOG(INFO) << "H5vccRuntime initial DeepLink: " << initial_deep_link_;
-}
+      deep_link_receiver_(this, window.GetExecutionContext()) {}
 
 void H5vccRuntime::ContextDestroyed() {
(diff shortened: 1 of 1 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_runtime/h_5_vcc_runtime.cc
```diff
@@ -24,18 +24,19 @@ namespace blink {
...
-}
+      deep_link_receiver_(this, window.GetExecutionContext()) {}
 
 void H5vccRuntime::ContextDestroyed() {
   remote_h5vcc_runtime_.reset();
   deep_link_receiver_.reset();
 }
 
 String H5vccRuntime::initialDeepLink() {
+  if (initial_deep_link_.empty()) {
```
xiaomings: If some code retrieves the deep link before this function is called, the initial deep link will be retrieved and cleared, and the link retrieved here is no longer the initial deep link.

zhongqiliang: The API design it that way, if the ondeeplink callback event is registered before, and the deeplink is returned through it, then call initialDeepLink will get nothing.

zhongqiliang: @jellefoks ^

third_party/blink/renderer/modules/cobalt/h5vcc_runtime/h_5_vcc_runtime.cc
```diff
@@ -24,18 +24,19 @@ namespace blink {
...
+    EnsureRemoteIsBound();
+    remote_h5vcc_runtime_->GetAndClearInitialDeepLinkSync(&initial_deep_link_);
+    LOG(INFO) << "H5vccRuntime initial DeepLink: " << initial_deep_link_;
+  }
```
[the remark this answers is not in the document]
zhongqiliang: accepted

PR #8704, opened 2026-01-16
Fix MUSL_REDIR_TIME64 macro definition mismatch

sherryzy:
This change ensuring that both ARM and X86 architectures MUSL_REDIR_TIME64 correctly defined
 as 0, aligning with their respective musl headers.

Bug: 476133776

### DIFF
--- a/starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.h
+++ b/starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.h
@@ -25,9 +25,7 @@
 extern "C" {
 #endif
 
-#if SB_IS(ARCH_ARM)
-#define MUSL_REDIR_TIME64 1
-#endif
+#define MUSL_REDIR_TIME64 0
 
 #define MUSL_PTHREAD_CREATE_JOINABLE 0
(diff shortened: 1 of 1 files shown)


starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.h
```diff
@@ -25,7 +25,9 @@
 extern "C" {
 #endif
 
-#if SB_IS(ARCH_ARM)
+#if SB_IS(ARCH_ARM) || SB_IS(ARCH_X86)
+#define MUSL_REDIR_TIME64 0
+#else
 #define MUSL_REDIR_TIME64 1
```
jellefoks: I think this should always be 0 for all platforms there this file is used.

sherryzy: While that holds true for the current implementation, several files under third_party/musl/arch/ still define REDIR_TIME64 = 1. Can we safely assume REDIR_TIME64 will always be 0 for Starboard? If so, I can remove this check, though it feels a bit fragile to rely on that assumption implicitly.

jellefoks (approved):

PR #8707, opened 2026-01-16
android: prevent unnecessary webContents reload from connectivity detector

ckim1346:
This PR adds a flag to `CobaltConnectivityDetector.java` to prevent unnecessary reloads of the webContents. The reload should only occur if we have encounter successful network check AND we have previously encountered a network failure.

Bug: 474529170


johnxwork:
We have been adding variables and checks to the detector in the past months. I think you want to introduce a state machine to make the flow simpler:
- The 204 ping has a few states: not stated, pending, success, failure. In the current design, success dismisses the dialog and failure shows the dialog. But for example, if a subsequent check request comes, you may want to keep the status as pending before the 204 completes. It's neither success nor failure.
- You want to minimize the number of variables that records particular state. The number of edge grows exponentially as we add more variables.

ckim1346:
> We have been adding variables and checks to the detector in the past months. I think you want to introduce a state machine to make the flow simpler:
> 
> * The 204 ping has a few states: not stated, pending, success, failure. In the current design, success dismisses the dialog and failure shows the dialog. But for example, if a subsequent check request comes, you may want to keep the status as pending before the 204 completes. It's neither success nor failure.
> * You want to minimize the number of variables that records particular state. The number of edge grows exponentially as we add more variables.

Agreed -- this is definitely a more temporary fix. Will file a bug to address the followup

PR #8708, opened 2026-01-16
Clean up crashpad hacks

andrewsavage1:


andrewsavage1:
Probably not the proper solution in some of these cases

PR #8709, opened 2026-01-16
posix: Clean up perror stdio hack

andrewsavage1:
Remove the conditional compilation that introduced a printf-based
implementation for perror when ENABLE_COBALT_HERMETIC_HACKS was enabled.
This change reverts perror to its standard musl implementation, as the
temporary hack method is no longer required due to improvements in
Cobalt's I/O infrastructure.

Fixed: 425993205

### DIFF
--- a/third_party/musl/src/stdio/perror.c
+++ b/third_party/musl/src/stdio/perror.c
@@ -2,22 +2,10 @@
 #include <string.h>
 #include <errno.h>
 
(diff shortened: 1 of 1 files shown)


jellefoks (approved):
Nice!

johnedocampo (approved):
Thanks for fix!

PR #8710, opened 2026-01-16
EG: Clean up dlsym fallback

andrewsavage1:
Remove the temporary dlsym fallback mechanism and its associated build
flag from the ELF loader's symbol lookup. This change improves symbol
resolution strictness and avoids misleading error logs for valid weak
symbols not found in the explicit map.

Fixed: 421944504

### DIFF
--- a/starboard/elf_loader/exported_symbols.cc
+++ b/starboard/elf_loader/exported_symbols.cc
@@ -14,14 +14,7 @@
 
 #include "starboard/elf_loader/exported_symbols.h"
 
-#include "build/build_config.h"
-
 #include <dirent.h>
-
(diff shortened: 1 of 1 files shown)


jellefoks (approved):

andrewsavage1:
Hm, right now unit tests have leaks—specifically `nplb_loader base_unittests_loader sql_unittests_loader boringssl_crypto_tests_loader perfetto_unittests_loader` leak at least the following:
```
alarm
fork
getopt
umask
vswprintf
```
Maybe I should only do this for devel/debug builds, but we should probably ensure at least 
nplb conforms to the expected exported symbols.

PR #8712, opened 2026-01-16
android: Enable AndroidOverlay for SbPlayer punch-out mode

borongc:
Enable the CobaltUsingAndroidOverlay command-line flag and feature
on Android. This forces CobaltActivity to utilize AndroidOverlay
instead of VideoSurfaceView for SbPlayer's punch-out mode.

This change ensures proper hardware accelerated video rendering and
resolves related playback issues. The explicit SetOverlayMode method
in Shell is no longer necessary with this new mechanism.

Issue: 460831614
Issue: 476418731
Issue: 476495060


borongc:
Note I didn't update `CommandLineOverrideHelperTest`. Put this as a future TODO if we can run it in CI.

PR #8717, opened 2026-01-16
Switch back to python test runner

linxinan-yt:
Issue: 433354983

### DIFF
--- a/cobalt/testing/browser_tests/run_browser_tests.py
+++ b/cobalt/testing/browser_tests/run_browser_tests.py
@@ -9,6 +9,7 @@ and iterate them.
 """
 
 import argparse
+import logging
 import os
 import subprocess
 import sys
@@ -45,9 +46,10 @@ class CobaltTestRunner:
       self._temp_dir_obj = tempfile.TemporaryDirectory(
           prefix="cobalt_user_data_")
       self.user_data_dir = self._temp_dir_obj.name
-      print(f"Using temporary user data directory: {self.user_data_dir}")
+      logging.info("Using temporary user data directory: %s",
+                   self.user_data_dir)
     else:
-      print(f"Using user data directory: {self.user_data_dir}")
+      logging.info("Using user data directory: %s", self.user_data_dir)
 
   def __enter__(self):
     """Enter the runtime context."""
@@ -62,14 +64,12 @@ class CobaltTestRunner:
     if self._temp_dir_obj:
       try:
         self._temp_dir_obj.cleanup()
-        print("Cleaned up temporary user data directory: "
-              f"{self.user_data_dir}")
+        logging.info("Cleaned up temporary user data directory: %s",
+                     self.user_data_dir)
       # pylint: disable=broad-exception-caught
       except Exception as e:
-        print(
-            "Error cleaning up temporary directory "
-            f"{self.user_data_dir}: {e}",
-            file=sys.stderr)
+        logging.error("Error cleaning up temporary directory %s: %s",
+                      self.user_data_dir, e)
       self._temp_dir_obj = None
     self._xml_root = None
     self._xml_tree = None
@@ -98,7 +98,7 @@ class CobaltTestRunner:
     if os.path.isfile(potential_binary_cwd):
       return potential_binary_cwd
 
-    print(f"Error: Binary not found at {binary_path}", file=sys.stderr)
+    logging.error("Binary not found at %s", binary_path)
     sys.exit(1)
 
   def _setup_sharding(self):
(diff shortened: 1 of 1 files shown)


cobalt/testing/browser_tests/run_browser_tests.py
```diff
@@ -7,137 +7,217 @@
...
-
-    # Sort by base name (Suite.TestBase), then by pre_count DESCENDING
-    return (f"{suite}.{case}", -pre_count)
+        test_name = f"{current_suite}{line}"
+        if "DISABLED_" not in test_name:
+          tests.append(test_name)
+
+  # Deterministic Sort
+  def get_sort_key(name):
+    return name
```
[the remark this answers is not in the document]
linxinan-yt: Done.

cobalt/testing/browser_tests/run_browser_tests.py
```diff
@@ -7,137 +7,217 @@
...
+              message="Test crashed or failed to output XML",
+              type="Crash")
+
+        # Update Counts
+        main_root.set("tests", str(passed_count + failed_count))
+        main_root.set("failures", str(failed_count))
+        main_tree.write(xml_output_file, encoding="UTF-8", xml_declaration=True)
+      # pylint: disable=broad-exception-caught
+      except Exception as e:
+        print(f"Error merging XML: {e}")
```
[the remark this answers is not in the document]
linxinan-yt: This happens between test cases, so not a big concern. The linux host have enough IO capacity to handle this.

linxinan-yt:
> Instead of one giant main function, can you break these up into smaller, more focused chunks and add a few unit tests for them? It will make it easier to maintain should we ever need to change it.

I had a version on that, but it quite similar to what you have done in #9260. I can upload it here for you to take a look at.

cobalt/testing/browser_tests/run_browser_tests.py
```diff
@@ -81,63 +171,256 @@ def get_sort_key(test_name):
...
-      ]
-
-      # Run the test and let it print to stdout/stderr
-      retcode = subprocess.run(cmd, check=False).returncode
-
-      if retcode != 0:
-        print(f"FAILED: {test} (Exit code: {retcode})")
-        failed_tests.append(test)
-      else:
+      root = ET.Element(
```
Awallky: In _merge_test_xml (Line 311) and _update_xml_summary (Line 321), the script performs a full read and write of the main XML output file for every single test case.
If the test suite grows to hundreds or thousands of tests, the overhead of parsing and re-writing the consolidated XML file repeatedly will become a significant bottleneck in CI.
Instead, we should accumulate test suite elements in memory during the run() loop and perform a single write at the end of the script, or use a more efficient append-only strategy.

cobalt/testing/browser_tests/run_browser_tests.py
```diff
@@ -7,71 +7,161 @@
...
+        self.total_shards = 1
+    if self.shard_index == 0 and "GTEST_SHARD_INDEX" in os.environ:
+      try:
+        self.shard_index = int(os.environ["GTEST_SHARD_INDEX"])
+      except ValueError:
+        print(
+            "Warning: Non-integer GTEST_SHARD_INDEX env var.", file=sys.stderr)
+        self.shard_index = 0
+
+    if not 0 <= self.shard_index < self.total_shards:
```
Awallky, line 124: If we misconfigure a shard (e.g., requests index 10 of 10), the script silently falls back to running shard 0. This results in shard 0's tests being run twice and shard 10's intended tests not being run at all.
Should this be a fatal error (sys.exit(1))? Could these silent fallbacks in sharding logic lead to hard-to-detect gaps.

linxinan-yt: Good point.

cobalt/testing/browser_tests/run_browser_tests.py
```diff
@@ -7,71 +7,161 @@
...
+  def list_tests(self) -> str:
+    """Lists tests from the binary using --gtest_list_tests.
+
+        Returns:
+            The standard output from the test listing command.
+
+        Raises:
+            SystemExit: If the list command fails.
+        """
+    print(f"Listing tests from {self.binary}...")
```
Awallky: We should probably replace all of these printf statements with logging.info/warning/error statements instead.

cobalt/testing/browser_tests/run_browser_tests.py
linxinan-yt: Make sense. Add a in memory xml root to store the temp results.

linxinan-chops (commented):
> We should probably replace all of these printf statements with logging.info/warning/error statements instead. 

Agree. Done.

cobalt/testing/browser_tests/run_browser_tests.py
linxinan-yt: Make sense. Done.

PR #8725, opened 2026-01-17
cobalt: Improve type casts in migration_manager.cc

yell0wd0g:
This PR/CL addresses a minor potential vulnerability (found by Gemini)
when `record->Read()` could return a `-1` on error.
 
It then then simplifies code by making |bytes| an std::vector<char> to 
reduce the number of type casts, and introduces `record_size` for that
same reason.

No new functionality intended.

Bug: 464009651

### DIFF
--- a/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
+++ b/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
@@ -55,7 +55,7 @@ std::string GetApplicationKey(const GURL& url) {
 }
 
 std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
-  constexpr std::string kRecordHeader = "SAV1";
+  constexpr std::string_view kRecordHeader = "SAV1";
   constexpr size_t kRecordHeaderSize = kRecordHeader.size();
 
   GURL initial_url(


yell0wd0g:
@Awallky PTAL

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -78,29 +79,30 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
     }
   }
 
-  if (record->GetSize() < static_cast<int64_t>(kRecordHeaderSize)) {
+  const auto record_size = record->GetSize();
```
Awallky, line 82: What's the benefit of not inlining this?

yell0wd0g: It's a bit shorter, and simplifies the `static_assert` in l.91:
I can do `decltype(record_size)`, versus if I had to get the
return type of `GetSize()` I'd have to use something like
```
      std::invoke_result_t<decltype(&StorageRecord::GetSize), StorageRecord>
```
which is hard to read : )

It might be faster if the compiler doesn't recognize that 
`record->GetSize()` should always return the same value?
`GetSize()` derives to `SbStorageGetSize()` (IIRC) and that
is implemented in a different .so for Evergreen builds...

yell0wd0g: Gonna resolve this conversation to land this but LMK
if this bothers you!

yell0wd0g:
After discussion with @zhongqiliang - this might be a good thing to bring to `26.android`.

PR #8732, opened 2026-01-20
tvOS: Fix hangs and timeouts when running NPLB player tests

rakuco:
Different tests in e.g. MaximumPlayerConfigurationExplorerTests, SbPlayerTest and MultiplePlayerTest were either hanging forever or timing out when run on tvOS.

The underlying cause in all cases is the same: either the tests themselves or one or more PlayerWorker instances created in the main thread were blocked joining other threads or waiting on a condition variable. At the same time, multiple parts of AVSBVideoRenderer (running on a different thread) were calling dispatch_sync()/dispatch_async() to run UI-related code in the main thread. The main thread was blocked, which resulted in a deadlock.

Fixing this means adopting a pattern of running the blocking code outside the main thread (fortunately all the places in question are thread-safe) while spinning the main loop until it is explicitly stopped from the background thread once the blocking code finishes running. The internals of how this is done are in a separate function, RunInBackgroundThreadAndWait(), that is invoked from some tests as well as tvOS-specific code.

More specifically:
* Spinning the UIKit main loop requires said loop to be started. This is done by invoking `UIApplicationMain()` and having it run `InitAndRunAllTests()`. Do this by pulling in some Chromium code from `//base/test` that takes care of setting up a UIKit application and running the tests at the right moment.
* Remove a SB_CHECK() call from attachPlayerView() since this code, when called from a test, has no corresponding player container view, and the subsequent no-op call is fine.
* Also invoke `RunInBackgroundThreadAndWait()` from the tvOS implementation of SbPlayerDestroy(), since there should be no more pending UI calls from AVSBVideoRenderer once the player is destroyed.

Fixed: 465634083
Bug: 450473125

### DIFF
--- a/starboard/tvos/shared/run_in_background_thread_and_wait.cc
+++ b/starboard/tvos/shared/run_in_background_thread_and_wait.cc
@@ -20,6 +20,14 @@
 #include "starboard/common/check_op.h"
 
 void RunInBackgroundThreadAndWait(std::function<void()>&& function) {
+  // This function is supposed to be used in very specific circumstances: test
+  // code that spawns one or more threads that need to make UI function calls
+  // on the main thread via dispatch_sync() or dispatch_async().
+  //
+  // In this context, it is not possible _not_ to have a run loop here: calling
+  // dispatch_sync() or using semaphores with GCD both have the same problem of
+  // inevitably blocking the main thread without running the main event loop,
+  // which leads to a deadlock.
   SB_CHECK_EQ(CFRunLoopGetMain(), CFRunLoopGetCurrent());
   dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
     function();


starboard/nplb/player_create_test.cc
```diff
@@ -114,8 +119,19 @@ class SbPlayerTest : public ::testing::Test {
...
       }
+
+#if BUILDFLAG(IS_IOS_TVOS)
+      // `ptr_lock` is needed because `lock` itself is not declared with __block
+      // and is therefore passed as const to the block below.
+      std::unique_lock<std::mutex>* const ptr_lock = &lock;
+      RunInBackgroundThreadAndWait(^{
+        condition_variable_.wait_for(*ptr_lock,
+                                     std::chrono::microseconds(wait_end - now));
+      });
```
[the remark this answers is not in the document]
rakuco: done

rakuco:

starboard/common/test_main.cc
```diff
@@ -47,6 +49,19 @@ int main(int argc, char** argv) {
 #else
 // If the OS is not Starboard use the regular main e.g. ATV.
 int main(int argc, char** argv) {
-  return InitAndRunAllTests(argc, argv);
+#if BUILDFLAG(IS_IOS_TVOS)
```
kjyoun: We already have `#if BUILDFLAG(IS_IOS_TVOS)` block inside `InitAndRunAllTests` at line 31

Can you merge this block int the existing one?

starboard/nplb/multiple_player_test.cc
```diff
@@ -43,6 +48,20 @@ class PlayerThread : public AbstractTestThread {
 
   void Run() override { functor_(); }
 
+  // A wrapper for Join() that, on tvOS, invokes it on a background GCD queue.
+  // This must be done to avoid a deadlock when the main thread is blocked on
+  // Join() and the worker thread is blocked attempting to invoke code in the
+  // main thread (in AVSBVideoRenderer).
+  void WaitForFinish() {
+#if BUILDFLAG(IS_IOS_TVOS)
+    RunInBackgroundThreadAndWait(^{
```
kjyoun: I don't think that it's a good practice to put objective-C syntaxax (i.e. ^) in common codes, even though it is under ifdef block


can you make RunInBackgroundThreadAndWait c++ compliant, (e.g. using std::function)?

starboard/common/test_main.cc
rakuco: Done!

starboard/nplb/multiple_player_test.cc
rakuco: Done!

starboard/tvos/shared/run_in_background_thread_and_wait.cc
```diff
@@ -19,10 +19,10 @@
 
 #include "starboard/common/check_op.h"
 
-void RunInBackgroundThreadAndWait(void (^block)(void)) {
+void RunInBackgroundThreadAndWait(const std::function<void()>& function) {
```
kjyoun: suggestion: pass-by-value and move?

Semantically, ownership of the function should be transferred to `RunInBackgroundThreadAndWait`. Taking it by value and moving it into the block makes this transfer explicit .

e.g.)
```
void RunInBackgroundThreadAndWait(std::function<void()> function) { // Pass by value
  SB_CHECK_EQ(CFRunLoopGetMain(), CFRunLoopGetCurrent());

  // Move the function into the block
  dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), [function = std::move(function)]() mutable {
    function();
    CFRunLoopStop(CFRunLoopGetMain());
  });

  CFRunLoopRun();
}
```

rakuco: How about `void RunInBackgroundThreadAndWait(std::function<void()>&& function)` instead?

kjyoun: Sounds good

rakuco: done!

rakuco:
@fayev-g could you take a look at this one?

starboard/tvos/shared/run_in_background_thread_and_wait.cc
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+#include "starboard/tvos/shared/run_in_background_thread_and_wait.h"
+
+#include <CoreFoundation/CoreFoundation.h>
+#include <dispatch/dispatch.h>
+
+#include "starboard/common/check_op.h"
+
+void RunInBackgroundThreadAndWait(std::function<void()>&& function) {
```
fayev-g, line 22: Totally understand you are trying to get around a problem with Join() here, but I believe using CFRunLoop this way is not common or recommended. If we cannot use dispatch_sync or dispatch_semaphore, can you add a comment above explaining why? And is it possible to move these files into a test-only target to make it clear this is a test-only work-around? Thanks!

From Gemini:

_Why this approach is problematic_
- _Nested Run Loops_: By calling CFRunLoopRun(), you are blocking the main execution path. If the OS sends a message to your app that isn't handled by the specific mode the run loop is in, your app will appear frozen.
- _The "Stop" Race Condition_: CFRunLoopStop(CFRunLoopGetMain()) only stops the current iteration of the run loop. If the background thread finishes before the main thread even enters CFRunLoopRun(), the stop signal might be lost or behave unpredictably.
- _Re-entrancy Hazards_: Since the main thread is technically "spinning" in a loop, it might still process other sources/timers. This can lead to code running in an order you didn't intend.

------------------------

_The Recommended Way_: dispatch_sync
If you are already on the main thread and want to run something on a background queue and wait for it, the standard tool is dispatch_sync. It is highly optimized and much cleaner.

```
void RunInBackgroundThreadAndWait(std::function<void()> function) {
    // Ensure we are on the main thread as per your original check
    assert(pthread_main_np() != 0); 
    
    dispatch_sync(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
        function();
    });
}
```

_The "Better" Way_: dispatch_group or dispatch_semaphore
If you need more control—or if you are worried about "Thread Starvation" (where dispatch_sync might behave unexpectedly under heavy load)—you can use a semaphore.

```
void RunInBackgroundThreadAndWait(std::function<void()> function) {
    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);

    dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
        function();
        dispatch_semaphore_signal(semaphore);
    });

    // Wait indefinitely on the main thread until the background task signals
    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
}
```
------------------------

_Important Architectural Note_
Synchronously waiting on the main thread for a background task is generally considered an anti-pattern in iOS/macOS development. It defeats the purpose of concurrency.

If function() takes 2 seconds, your UI will be completely unresponsive for those 2 seconds. The better approach is usually to make RunInBackgroundThread asynchronous and use a completion handler or block to update the UI when finished.

Warning: Never use dispatch_sync to target the current queue you are already on, as this will cause an immediate deadlock.

rakuco: > Totally understand you are trying to get around a problem with Join() here, but I believe using CFRunLoop this way is not common or recommended. If we cannot use dispatch_sync or dispatch_semaphore, can you add a comment above explaining why? And is it possible to move these files into a test-only target to make it clear this is a test-only work-around? Thanks!

Using CFRunLoop like this is not common, but manually spinning the event loop is the only way that works here as far as I can see. I've added a comment explaining why, but I can't move this to a test-only target because the tvOS `SbPlayerDestroy()` implementation also calls this function.

> From Gemini:

This is reasonable advice for the common case, it just fails to take into consideration the specificity of this problem and how the NPLB tests and AVSBRender code are structured (which I also discuss in b/450473125). The basic problem here is that we have a UIApplication that, when loaded, spawns a function that synchronously runs GTest-based unit tests but we need to wait for certain UI activity to happen, which only happens if the event loop is running and processing events. In Chromium code, we'd basically be doing something like

``` c++
base::RunLoop run_loop;
RunSomeCodeAsynchronously(run_loop.QuitClosure());
run_loop.Run();
```

We can't use `base::RunLoop` here because we're in Starboard and it'd require adding more `#ifdef`s and conditional GN deps because the tests also need a `base::test::TaskEnvironment`. This is a simplified version of the same idea.

> _Why this approach is problematic_
>
>     * _Nested Run Loops_: By calling CFRunLoopRun(), you are blocking the main execution path. If the OS sends a message to your app that isn't handled by the specific mode the run loop is in, your app will appear frozen.

True, but not really a concern in this controlled and trimmed down unit test environment. It would be possible to copy some of the `base::RunLoop` code that deals with nesting levels but it just does not seem to be necessary here.

>     * _The "Stop" Race Condition_: CFRunLoopStop(CFRunLoopGetMain()) only stops the current iteration of the run loop. If the background thread finishes before the main thread even enters CFRunLoopRun(), the stop signal might be lost or behave unpredictably.

The way this code is used, the background thread can't finish before the main thread enters `CFRunLoopRun()` because it is blocked waiting for GCD tasks to be processed, which happens when `CFRunLoopRun()` executes.

>     * _Re-entrancy Hazards_: Since the main thread is technically "spinning" in a loop, it might still process other sources/timers. This can lead to code running in an order you didn't intend.

True, but not a realistic issue in these NPLB tests AFAICS.

> _The Recommended Way_: dispatch_sync If you are already on the main thread and want to run something on a background queue and wait for it, the standard tool is dispatch_sync. It is highly optimized and much cleaner.

This does not work. It's still synchronously waiting for a `Join()` and the deadlock continues to occur.

> _The "Better" Way_: dispatch_group or dispatch_semaphore If you need more control—or if you are worried about "Thread Starvation" (where dispatch_sync might behave unexpectedly under heavy load)—you can use a semaphore.

This also does not work. `dispatch_semaphore_wait()` will also block the main thread without spinning the event loop, so we're basically in the same situation as if `Join()` had been invoked directly.

> _Important Architectural Note_ Synchronously waiting on the main thread for a background task is generally considered an anti-pattern in iOS/macOS development. It defeats the purpose of concurrency.
>
> If function() takes 2 seconds, your UI will be completely unresponsive for those 2 seconds. The better approach is usually to make RunInBackgroundThread asynchronous and use a completion handler or block to update the UI when finished.

This is true on a general level, and the ideal situation would be to refactor some of the tvOS media code so that it fits better into the more regular SbPlayer architecture so that this problem would not exist in the first place. However, this is what we need at the moment.

fayev-g: Thanks for adding the comment!

fayev-g (approved):

rakuco:
Hmm, I thought approval from cobalt-tvos was enough but apparently someone in cobalt-starboard-owners also needs to LGTM this PR :/ @fayev-g do you know who's part of this team?

PR #8743, opened 2026-01-20
Fix potential deadlock in __lockfile.

jellefoks:
If the thread ID has the MAYBE_WAITERS bit set (0x40000000), the recursion check fails because tid is not masked. This causes a thread to wait on a lock it already holds, leading to deadlock. Masking tid resolves this.

Bug: 473609573


jellefoks:
Closing, Linux kernels set PID_MAX_LIMIT to 4M, and have an implementation reason to limit it to below 1G, so this should never trigger on Linux.

PR #8749, opened 2026-01-21
media: Remove external allocator toggle

kjyoun:
Experiments showed that disabling the custom Starboard media allocator provides no benefit. This change removes the logic that allowed disabling the allocator via H5VCC settings, defaulting to always using the custom allocator when one is set.

RASTA result: https://screenshot.googleplex.com/5ESKgwuZjZPqncQ

Bug: 456552294

### DIFF
--- a/media/base/decoder_buffer.cc
+++ b/media/base/decoder_buffer.cc
@@ -29,13 +29,6 @@ void DecoderBuffer::Allocator::Set(Allocator* allocator) {
 }
 
 // static
-void DecoderBuffer::EnableAllocator(bool enabled) {
-  CHECK(s_allocator);
-  s_allocator->SetEnabled(enabled);
-  s_use_allocator = enabled;
-}
-
-// static
 void DecoderBuffer::EnableAllocateOnDemand(bool enabled) {
   CHECK(s_allocator);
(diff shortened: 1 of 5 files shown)


media/base/decoder_buffer.cc
```diff
@@ -28,13 +28,6 @@ void DecoderBuffer::Allocator::Set(Allocator* allocator) {
   s_use_allocator = true;
```
xiaomings, line 28: @kjyoun @osagie98 This change is supposed to be a partial revert of https://github.com/youtube/cobalt/pull/7933.  It doesn't seem to be a clean revert (e.g. `s_use_allocator` is left without being used) and is created as a new change.  We should consider revert experimental changes in future, instead of removing experimental code as new changes.

xiaomings: Another example is that `static Allocator* GetInstance();` was removed in the previous PR, and not added back.  Now functions for DecoderBuffer::Allocator global settings are forced into DecoderBuffer as static functions.

kjyoun: re: s_use_allocator = true,  made https://github.com/youtube/cobalt/pull/8838

re: static Allocator* GetInstance(), will create a separate PR

PR #8753, opened 2026-01-21
tvos: Fix SbMediaCanPlayMimeAndKeySystem.MinimumSupport

Gyuyoung:
Exclude AV1 codec tests from the SbMediaCanPlayMimeAndKeySystem.MinimumSupport
test on tvOS. tvOS is known not to support the AV1 codec, which causes
failures in this test if AV1-specific media capabilities are queried.
This ensures the test passes correctly on tvOS.

Bug: 473657163

### DIFF
--- a/starboard/darwin/tvos/shared/test_filters.py
+++ b/starboard/darwin/tvos/shared/test_filters.py
@@ -25,9 +25,6 @@ _FILTERED_TESTS = {
         # TODO(b/274023703): Enable partial audio tests once it's supported.
         'SbPlayerWriteSampleTests/SbPlayerWriteSampleTest.PartialAudio*',
         'SbPlayerWriteSampleTests/SbPlayerWriteSampleTest.DiscardAllAudio*',
-
-        # Enable once verified on the platform.
-        'SbMediaCanPlayMimeAndKeySystem.MinimumSupport',
     ],
     'base_unittests': [
         'StackTraceTest.OutputToStream',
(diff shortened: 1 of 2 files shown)


Gyuyoung:
CC @rakuco @abhijeetk @jkim-julie

rakuco:
After #8247, can't we check for `BUILDFLAG(ENABLE_AV1_DECODER)` instead? And instead of multiple `#ifdef`s I guess you could do

```c++
#if BUILDFLAG(...)
params_fhd.push_back("...");
params_4k.push_back("...");
params_8k.push_back("...");
#endif
```

Gyuyoung:
> After #8247, can't we check for `BUILDFLAG(ENABLE_AV1_DECODER)` instead? And instead of multiple `#ifdef`s I guess you could do
> 
> ```c++
> #if BUILDFLAG(...)
> params_fhd.push_back("...");
> params_4k.push_back("...");
> params_8k.push_back("...");
> #endif
> ```

yes, it looks better. Updated.

rakuco:
> Updated.

The code still looks the same but with a different `#ifdef` check. Unfortunately the check for AV1 won't work, I didn't realize it was set in `//media`, which I don't think can be a dependency here in `//starboard`.

I'd go with my original suggestion but checking for `IS_IOS_TVOS` along with an explanation in a comment.

Gyuyoung:
> > Updated.
> 
> The code still looks the same but with a different `#ifdef` check. Unfortunately the check for AV1 won't work, I didn't realize it was set in `//media`, which I don't think can be a dependency here in `//starboard`.
> 
> I'd go with my original suggestion but checking for `IS_IOS_TVOS` along with an explanation in a comment.

ok, I applied them to the new patch. Thanks.

PR #8755, opened 2026-01-21
[tvOS] Introduce `using_mac_runner` to depot_tools/action and use runner.temp

fayev-g:
Bug: 477593749

runner-cache is not available on default Github Mac runners and so we special case to use runner.temp for now behind `using_mac_runner`. Clean this up when self-hosted Mac runners are available.

### DIFF
--- a/.github/actions/depot_tools/action.yaml
+++ b/.github/actions/depot_tools/action.yaml
@@ -70,7 +70,7 @@ runs:
         gclient config \
           --name=src \
           --custom-var=download_remoteexec_cfg=True \
-          --custom-var='rbe_instance="${{ inputs.rbe_instance }}"'
+          --custom-var='rbe_instance="${{ inputs.rbe_instance }}"' \
           https://github.com/${{ github.repository }}
       shell: bash
 


.github/actions/depot_tools/action.yaml
```diff
@@ -79,10 +96,19 @@ runs:
...
         shell: bash
+
+      - name: Run gclient runhooks for Mac
+        if: inputs.run_sync == 'true' && inputs.using_mac_runner == 'true'
+        env:
+          CIPD_CACHE_DIR: "${{ runner.temp }}/cobalt/.cipd_cache"
+        run: |
+          cd cobalt
+          gclient runhooks --verbose
+        shell: bash
```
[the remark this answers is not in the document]
fayev-g: Done

.github/actions/depot_tools/action.yaml
```diff
@@ -10,79 +10,97 @@ inputs:
   rbe_instance:
     description: The RBE instance to use.
     default: projects/cobalt-actions-prod/instances/default_instance
+  # TODO: b/477593749 - runner-cache is not available on default Github Mac
+  # runners and so we special case to use runner.temp for now. Clean this up
+  # when self-hosted Mac runners are available.
+  using_mac_runner:
```
oxve, line 16: I'm not sure I like having a mac build carveout as a general principle. I think I would prefer to use feature gating instead.

The new param is really `use_cache` and the existing `use_cache` should be `clear_cache`.

.github/actions/depot_tools/action.yaml
oxve, line 1: Thanks for fixing the indent!

.github/actions/depot_tools/action.yaml
fayev-g: Filed b/478259367 to follow up on cleaning up the naming here!

PR #8767, opened 2026-01-22
cobalt: Rename H5vcc metrics histogram output

yjzhang111:
The RequestHistograms Mojom method and its corresponding C++ implementation in Blink used "histograms_json" for its return value. This name was misleading as the actual content is base64 encoded protocol buffer data, not JSON. Rename to "histograms_base64" to accurately reflect the content type and improve code clarity.

Bug: 467384307

### DIFF
--- a/cobalt/browser/h5vcc_metrics/public/mojom/h5vcc_metrics.mojom
+++ b/cobalt/browser/h5vcc_metrics/public/mojom/h5vcc_metrics.mojom
@@ -29,5 +29,5 @@ interface H5vccMetrics {
 
   [Sync] Enable(bool enable) => ();
   SetMetricEventInterval(uint64 interval_seconds) => ();
-  RequestHistograms() => (string histograms_json);
+  RequestHistograms() => (string histograms_base64);
 };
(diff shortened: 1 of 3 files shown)


cobalt/browser/h5vcc_metrics/public/mojom/h5vcc_metrics.mojom
```diff
@@ -29,5 +29,5 @@ interface H5vccMetrics {
 
   [Sync] Enable(bool enable) => ();
```
yell0wd0g, line 30: QQ and unrelated to this PR: Does this method need to be
"sync"? If so then maybe leave a comment as to why.

yjzhang111: Good point. The IDL definition is async and there's no reason for the mojo implementation to be sync. Filed b/478954561 to clean up
https://github.com/youtube/cobalt/blob/main/third_party/blink/renderer/modules/cobalt/h5vcc_metrics/h_5_vcc_metrics.idl#L35

yell0wd0g: Stating the obvious: no need to address that here! Follow up is OK.

PR #8768, opened 2026-01-22
Add file chooser test resources for content unit tests

Gyuyoung:
It appears that the cobalt port was missing the file chooser test data required by content unit tests. This change adds the missing resources so that content unit tests can be built on the tvOS port.

Bug: 473657163


Gyuyoung:

rakuco:
Hmm, it doesn't look like `content_unittests` is being run in CI at all so it's probably not relevant for Cobalt. Furthermore, isn't all the file chooser code disabled in Cobalt anyway?

Gyuyoung:
> Hmm, it doesn't look like `content_unittests` is being run in CI at all so it's probably not relevant for Cobalt. Furthermore, isn't all the file chooser code disabled in Cobalt anyway?

If so, I misunderstood. Cobalt only runs tests cases related to Cobalt. Then, we don't need to support content_unittests, and it's enough to run them in Chromium CI.

rakuco:
I originally thought these tests were run in CI too, but looking at what ran for this PR it doesn't look like `content_{unittests,browsertests}` are run.

Gyuyoung:
> I originally thought these tests were run in CI too, but looking at what ran for this PR it doesn't look like `content_{unittests,browsertests}` are run.

This PR was not to run content_unittests, instead this was to build `content unit tests` because content unit tests target included the file chooser test data files. The PR description might cause you confused. I updated it.

> If so, I misunderstood. Cobalt only runs tests cases related to Cobalt. Then, we don't need to support content_unittests, and it's enough to run them in Chromium CI.

BTW, it looks like android CI bot[1] didn't run `content_unittests`, but it ran other unit tests like `blink_heap_unittest`, `media_unittests`, `mojo_unittests`. Is there any reason we don't need to run `content_unittests` on CI?

[1] https://github.com/youtube/cobalt/actions/runs/21248524124/job/61146949002?pr=8753

rakuco:
> This PR was not to run content_unittests, instead this was to build `content unit tests` because content unit tests target included the file chooser test data files. The PR description might cause you confused. I updated it.

What I meant was that if it's not being run anywhere, it's not a problem that the target doesn't build either.

> Is there any reason we don't need to run `content_unittests` on CI?

My guess is that they test a lot of functionality that is not used by Cobalt. @fayev-g do you know who could answer this?

Awallky:
So, cobalt's test targets are defined at `//cobalt/build/testing/targets/`. We do test some non-cobalt targets, but that is because there are changes outside of `//cobalt` that we wanted to make sure we didn't break anything in upstream Chromium. 
We are only allowed to break Cobalt. :)

Gyuyoung:
> So, cobalt's test targets are defined at `//cobalt/build/testing/targets/`. We do test some non-cobalt targets, but that is because there are changes outside of `//cobalt` that we wanted to make sure we didn't break anything in upstream Chromium. We are only allowed to break Cobalt. :)

Thank you for the explanation. My understanding is that Cobalt runs additional tests only when there are modifications outside of //cobalt. For example, if there are changes in `//media`, Cobalt runs media_unittests on CI. Is that correct?

If so, I noticed that there are dozens of changes [1][2] in `//content` guarded by IS_STARBOARD_FOO or IS_COBALT_FOO. Given this, would it make sense to run `content_unittests` as well?

[1] https://github.com/search?q=repo%3Ayoutube%2Fcobalt%20%20IS_STARBOARD%20path%3Acontent&type=code
[2] https://github.com/search?q=repo%3Ayoutube%2Fcobalt++IS_COBALT+path%3Acontent&type=code

Gyuyoung:
It looks like running content_unittests is not a priority task for now. Let's consider this later again. Close this PR again.

PR #8771, opened 2026-01-22
media: Simplify StarboardRenderer initialization logic

borongc:
The StarboardRenderer previously used playing_start_from_time_ and
player_bridge_initialized_ to defer playback initiation if
StartPlayingFrom was called before the underlying SbPlayer was
fully initialized.

This change modifies the CreatePlayerBridge and OnPlayerStatus
flow to ensure the renderer's initialization callback is only
invoked after the SbPlayer reports kSbPlayerStateInitialized.
This guarantees that any subsequent call to StartPlayingFrom
will always find the SbPlayer in a ready state, eliminating the
need for internal state variables to manage deferred playback.

This simplifies the StarboardRenderer's logic and removes redundant
state.

Issue: 477930552

### DIFF
--- a/media/starboard/starboard_renderer.cc
+++ b/media/starboard/starboard_renderer.cc
@@ -297,13 +297,6 @@ void StarboardRenderer::Flush(base::OnceClosure flush_cb) {
 
   LOG(INFO) << "Flushing StarboardRenderer.";
 
-  // It's possible that Flush() is called immediately after StartPlayingFrom(),
-  // before the underlying SbPlayer is initialized.  Reset
-  // `playing_start_from_time_` here as StartPlayingFrom() checks for
-  // re-entrant.  This also avoids the stale `playing_start_from_time_` to be
-  // used.
-  playing_start_from_time_.reset();
-
   // Prepares the |player_bridge_| for Seek(), the |player_bridge_| won't
   // request more data from us before Seek() is called.
   player_bridge_->PrepareForSeek();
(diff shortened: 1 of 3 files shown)


media/starboard/starboard_renderer.cc
```diff
@@ -956,18 +945,12 @@ void StarboardRenderer::OnPlayerStatus(SbPlayerState state) {
 
   switch (state) {
     case kSbPlayerStateInitialized:
-      DCHECK(!player_bridge_initialized_);
-      player_bridge_initialized_ = true;
-
-      if (playing_start_from_time_) {
-        StartPlayingFrom(std::move(playing_start_from_time_).value());
-      }
+      DCHECK(init_cb_);
```
kjyoun: Can you change this to CHECK?

On prod  code, next line will crash the app, if init_cb_ is nullptr.
with CHECK, we can get clear reason about the crash.
and this would not incur performance burden.

go/cpplabs/avoid-dcheck.md

media/starboard/starboard_renderer.cc
```diff
@@ -677,7 +662,11 @@ void StarboardRenderer::CreatePlayerBridge() {
     player_bridge_->SetVolume(volume_);
 
     state_ = STATE_FLUSHED;
-    std::move(init_cb_).Run(PipelineStatus(PIPELINE_OK));
+
+    // Return `PipelineStatus(PIPELINE_OK)` if SbPlayer is in the state of
+    // `kSbPlayerStateInitialized` in OnPlayerStatus(). This ensures
+    // StartPlayingFrom() is called after SbPlayer is in
+    // `kSbPlayerStateInitialized`.
```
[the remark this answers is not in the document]
borongc: Done.

media/starboard/starboard_renderer.cc
borongc: Done.

PR #8774, opened 2026-01-22
Conditionally bind H5vccUpdater service

madhurajayaraman:
Cobalt: Conditionally bind H5vccUpdater service

Relocate the conditional binding of the H5vccUpdater Mojo interface
from the Blink renderer module to the Cobalt browser process.
This change ensures the Mojo service is only provided when the
USE_EVERGREEN build flag is enabled, centralizing control over
the API's availability. The client-side module now always attempts
to bind and gracefully handles Mojo connection errors by rejecting
pending promises if the service is unavailable.

Bug: 454440974

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
@@ -320,11 +320,15 @@ void H5vccUpdater::EnsureReceiverIsBound() {
 void H5vccUpdater::OnConnectionError() {
   remote_h5vcc_updater_.reset();
   HeapHashSet<Member<ScriptPromiseResolver>> h5vcc_updater_promises;
-  // Script may execute during a call to Resolve(). Swap these sets to prevent
+  // Script may execute during a call to Reject(). Swap these sets to prevent
   // concurrent modification.
   ongoing_requests_.swap(h5vcc_updater_promises);
   for (auto& resolver : h5vcc_updater_promises) {
+#if BUILDFLAG(USE_EVERGREEN)
     resolver->Reject("Mojo connection error.");
+#else
+    resolver->Reject("API not supported for this platform.");
+#endif
   }
   ongoing_requests_.clear();
 }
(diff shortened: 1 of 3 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
```diff
@@ -322,16 +313,27 @@ void H5vccUpdater::EnsureReceiverIsBound() {
...
+void H5vccUpdater::OnConnectionError() {
+  remote_h5vcc_updater_.reset();
+  HeapHashSet<Member<ScriptPromiseResolver>> h5vcc_updater_promises;
+  // Script may execute during a call to Resolve(). Swap these sets to prevent
+  // concurrent modification.
+  ongoing_requests_.swap(h5vcc_updater_promises);
+  for (auto& resolver : h5vcc_updater_promises) {
+    resolver->Reject("Mojo connection error.");
+  }
+  ongoing_requests_.clear();
```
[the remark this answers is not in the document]
madhurajayaraman: Leaving this as is for consistency with other h5vcc APIs.

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
```diff
@@ -36,15 +31,13 @@ ScriptPromise H5vccUpdater::getUpdaterChannel(ScriptState* script_state,
   auto* resolver = MakeGarbageCollected<ScriptPromiseResolver>(
       script_state, exception_state.GetContext());
 
-#if BUILDFLAG(USE_EVERGREEN)
   EnsureReceiverIsBound();
 
+  ongoing_requests_.insert(resolver);
   remote_h5vcc_updater_->GetUpdaterChannel(
```
yell0wd0g, line 37: What happens if we call this Web API (or any other in this 
file) on a  platform that doesn't define `USE_EVERGREEN`? 

Maybe `EnsureReceiverIsBound()` would not work (there'd
be no "service" side), but since that returns nothing,
the code would continue and use `remote_h5vcc_updater_`
and then...?

madhurajayaraman: The behavior is documented in a comment on top of the .idl file.
The renderer will attempt to connect to the Mojo interface, it will fail because no binder is registered. At that point, the disconnect handler catches this error and rejects all pending promises with "API not supported on this platform."

madhurajayaraman: I've also tested it with devtools to ensure that the behavior does happen, and this image has the behavior when I call one of the APIs on a linux build:
Screenshot 2026-01-22 at 1 48 03 PM

yell0wd0g: Oooh `OnConnectionError()` is called and the promise is rejected, gotcha,
perfect then. Thanks for the clarification!

PR #8776, opened 2026-01-22
android: simplify network error dialog and add dismiss button

ckim1346:
The network error dialog has been seeing persistent dialogs in some cases as well as other issues. This PR significantly simplifies the dialog's and `CobaltConnectivityDetector`'s functionality to only perform a network check on cold startup and only reload the web contents when the "Try again" button is pressed manually by the user. Additionally, this PR introduces a 'dismiss' button that only dismisses the dialog without any additional functionality.

Bug:477896593

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltConnectivityDetector.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltConnectivityDetector.java
@@ -28,7 +28,8 @@ import org.chromium.content_public.browser.WebContents;
 
 public class CobaltConnectivityDetector {
   private static final String TAG = "CobaltConnectivityDetector";
-  private static final int NETWORK_CHECK_TIMEOUT_MS = 5000; // 5-second overall timeout.
+  private static final int NETWORK_CHECK_TIMEOUT_MS = 5000;
+  private static final int NETWORK_CHECK_OVERALL_TIMEOUT_MS = 10000;
   private static final String DEFAULT_PROBE_URL = "https://www.google.com/generate_204";
   private static final String FALLBACK_PROBE_URL =
       "http://connectivitycheck.gstatic.com/generate_204";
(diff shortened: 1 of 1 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltConnectivityDetector.java
```diff
@@ -139,8 +127,8 @@ private boolean performSingleProbe(String urlString) {
       URL url = new URL(urlString);
       urlConnection = (HttpURLConnection) url.openConnection();
       // Shorter timeouts for connect/read, the overall timeout is the main guard.
-      urlConnection.setConnectTimeout(4000);
-      urlConnection.setReadTimeout(4000);
+      urlConnection.setConnectTimeout(NETWORK_CHECK_TIMEOUT_MS);
+      urlConnection.setReadTimeout(NETWORK_CHECK_TIMEOUT_MS);
```
johnxwork, line 128: Since you use 5s here, do you want the overall timeout for DNS to be 10s?

ckim1346: I believe when @ jainpoonam was testing, 10s felt too long for the user to be stuck on a black screen

ckim1346: Could update the overall to be 7.5 instead?

johnxwork: But do you have enough time to probe the two urls then?
Also, we moved the network check to `oncreate()` which happens much faster than `onresume`.

johnxwork: Gemini complaint the same

PR #8784, opened 2026-01-23
build: Explicitly set chrome_pgo_phase to 0 on tvOS

rakuco:
This is also done for all other platforms. On gold and qa builds targeting a device (which set `is_official_build` to true), this value will be 2 by default and break because Cobalt's git hash and changes to V8 make at least mksnapshot crash because the downloaded PGO profiles do not match the build.

Bug: 376141256
Bug: 456692793


rakuco:

rakuco:
@fayev-g could you either merge or enable auto-merge for this PR? Thanks!

PR #8795, opened 2026-01-23
Integrate kSbEventDateTimeConfigurationChanged with Chromium's device…

haozheng-cobalt:
…::TimeZoneMonitor service

### DIFF
--- a/ui/ozone/platform/starboard/platform_window_starboard.cc
+++ b/ui/ozone/platform/starboard/platform_window_starboard.cc
@@ -106,6 +106,8 @@ void PlatformWindowStarboard::ProcessFocusEvent(bool is_focused) {
   }
 }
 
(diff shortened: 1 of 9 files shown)


ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -106,6 +106,8 @@ void PlatformWindowStarboard::ProcessFocusEvent(bool is_focused) {
   }
 }
 
+void PlatformWindowStarboard::ProcessDateTimeConfigurationChangedEvent() {}
```
sherryzy, line 109: This timezone configuration doesn't seem related to window management. Ozone is primarily designed for the UI/Graphics stack, whereas this event belongs in the system stack. We should investigate a more appropriate bridge for this event—perhaps something closer to a system monitor or a dedicated configuration service rather than PlatformWindow.

PR #8803, opened 2026-01-26
Conditionally add binding in mojo for sideloading APIs

madhurajayaraman:
Evergreen: Isolate updater sideloading configuration

Move updater configuration setters to a dedicated sideloading Mojo
interface. This separates sensitive update settings, such as allowing
self-signed packages, update server URL, and network encryption
requirements, into an interface only bound in non-release Evergreen
builds with sideloading enabled. This prevents unauthorized modification
of update settings in production environments.

Bug: 454440974

### DIFF
--- a/cobalt/browser/cobalt_browser_interface_binders.cc
+++ b/cobalt/browser/cobalt_browser_interface_binders.cc
@@ -36,6 +36,8 @@
 
 #if BUILDFLAG(USE_EVERGREEN)
 #include "cobalt/browser/h5vcc_updater/h5vcc_updater_impl.h"
+// TODO(b/458483469): Remove the ALLOW_EVERGREEN_SIDELOADING check after
+// security review.
 #if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
 #include "cobalt/browser/h5vcc_updater/h5vcc_updater_sideloading_impl.h"
 #endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
@@ -105,6 +107,8 @@ void PopulateCobaltFrameBinders(
 #if BUILDFLAG(USE_EVERGREEN)
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterImpl::Create));
+// TODO(b/458483469): Remove the ALLOW_EVERGREEN_SIDELOADING check after
+// security review.
 #if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdaterSideloading>(
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterSideloadingImpl::Create));
--- a/cobalt/browser/h5vcc_updater/BUILD.gn
+++ b/cobalt/browser/h5vcc_updater/BUILD.gn
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# TODO(b/458483469): Conditionally add sideloading files to sources if
+# 'cobalt_is_release_build' and sideloading are enabled.
 source_set("h5vcc_updater") {
   if (use_evergreen) {
     sources = [
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
@@ -344,6 +344,8 @@ void H5vccUpdater::OnSideloadingConnectionError() {
   // concurrent modification.
   ongoing_sideloading_requests_.swap(h5vcc_updater_promises);
   for (auto& resolver : h5vcc_updater_promises) {
+// TODO(b/458483469): Remove the ALLOW_EVERGREEN_SIDELOADING check after
+// security review.
 #if BUILDFLAG(USE_EVERGREEN) && !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && \
     ALLOW_EVERGREEN_SIDELOADING
     resolver->Reject("Mojo connection error.");
--- a/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
+++ b/cobalt/browser/h5vcc_updater/h5vcc_updater_impl.cc
@@ -52,11 +52,11 @@ void H5vccUpdaterImpl::Create(
 void H5vccUpdaterImpl::GetUpdaterChannel(GetUpdaterChannelCallback callback) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
-  if (!updater_module) {
-    std::move(callback).Run("");
+  if (updater_module) {
+    std::move(callback).Run(updater_module->GetUpdaterChannel());
     return;
   }
-  std::move(callback).Run(updater_module->GetUpdaterChannel());
+  std::move(callback).Run("");
 }
 
 void H5vccUpdaterImpl::SetUpdaterChannel(const std::string& channel,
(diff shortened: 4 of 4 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
```diff
@@ -333,11 +321,48 @@ void H5vccUpdater::OnConnectionError() {
...
+#if BUILDFLAG(USE_EVERGREEN) && !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && \
+    ALLOW_EVERGREEN_SIDELOADING
+    resolver->Reject("Mojo connection error.");
+#else
+    resolver->Reject(
+        "API not supported for this build configuration. Enabled for "
+        "sideloading only.");
+#endif
+  }
+  ongoing_sideloading_requests_.clear();
```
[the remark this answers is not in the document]
madhurajayaraman: This is consistent with other web apis. leaving as is.

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
```diff
@@ -21,10 +21,13 @@ interface H5vccUpdater {
   ResetInstallations() => ();
   GetInstallationIndex() => (uint16 index);
   GetAllowSelfSignedPackages() => (bool allow_self_signed_packages);
```
hlwarriner: I think the getters should be moved to the new Sideloading interface, too.

cobalt/browser/cobalt_browser_interface_binders.cc
```diff
@@ -97,6 +97,10 @@ void PopulateCobaltFrameBinders(
 #if BUILDFLAG(USE_EVERGREEN)
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterImpl::Create));
+#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
+  binder_map->Add<h5vcc_updater::mojom::H5vccUpdaterSideloading>(
+      base::BindRepeating(&h5vcc_updater::H5vccUpdaterSideloadingImpl::Create));
+#endif
```
hlwarriner: Please add closing `#endif` comments on this line and the next line for readability since we now have nested conditions here.

```
#endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
#endif  // BUILDFLAG(USE_EVERGREEN)
```

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.h
```diff
@@ -71,6 +63,35 @@ class H5vccUpdaterImpl : public content::DocumentService<mojom::H5vccUpdater> {
   THREAD_CHECKER(thread_checker_);
 };
 
+class H5vccUpdaterSideloadingImpl
```
hlwarriner: Optional: consider extracting `H5vccUpdaterSideloadingImpl` into its own `.h`/`/cc` files. There's no official readability guidance around this AFAIK but I feel it would improve readability/organization. Also, separating them into their own files would allow the compiler to ignore `H5vccUpdaterSideloadingImpl` in non-sideloading builds.

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
madhurajayaraman: Do we know that we will only ever call the getters on sideloading builds? I assume that in a sideloading build we could set `allow_self_signed_packages=true` and then on any build we could get the value to determine the usage?

cobalt/browser/h5vcc_updater/h5vcc_updater_impl.h
madhurajayaraman: moved into their own files for readability

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
hlwarriner: @yuying-y can you please answer Madhura's question and share your recommendation?

hlwarriner: And by the way, it's a good question and fair point. In my opinion it would be better to treat all the sideloading setters+getters as one cohesive interface and rely on the rejected promise with "API not supported..." as the signal to the web client that even the getters are not supported on the platform/build, meaning sideloading is not supported on the platform/build. But @yuying-y should probably make the call since she designed the feature.

cobalt/browser/h5vcc_updater/BUILD.gn
```diff
@@ -16,6 +16,8 @@ source_set("h5vcc_updater") {
   sources = [
     "h5vcc_updater_impl.cc",
     "h5vcc_updater_impl.h",
+    "h5vcc_updater_sideloading_impl.cc",
+    "h5vcc_updater_sideloading_impl.h",
```
yell0wd0g: Looking at `cobalt_browser_interface_binders.cc`, 
`H5vccUpdaterSideloadingImpl` is only bound if 
```
  !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
```

can we just not build these two files if those 
conditions don't verify? There's usually a GN
variable version of buildflags etc.

This in turn would simplify the
`h5vcc_updater_sideloading_impl.cc` pattern of 

```
#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
  auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
  if (updater_module) {
    updater_module->SetAllowSelfSignedPackages(allow_self_signed_packages);
  }
  std::move(callback).Run();
#else
  std::move(callback).Run();
  NOTIMPLEMENTED();
#endif
```

since you could remove the guards (maybe leave an
`static_assert` at the top of the file to make sure
nobody builds it when
`!BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING`
doesn't verify).

madhurajayaraman: So we actually shouldn't really ever hit that code anyway - it's there for ease of reading more than anything else. The mojo binding is conditional on `!BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING`, so we shouldn't ever even get to a point where we call the function in `h5vcc_updater_sideloading_impl.cc`. Now that I'm typing this out I realize it's dead code and I should just throw a comment on top of each of the updater files to specify that the necessary conditionals are in the mojo binding layer, and not in any of the .cc or .h files.

tl;dr : yes, we can, and yes, we should.

cobalt/browser/h5vcc_updater/public/mojom/h5vcc_updater.mojom
yuying-y: Let's keep the getters and setters together, all moved to the sideloading interface.

cobalt/browser/cobalt_browser_interface_binders.cc
```diff
@@ -36,6 +36,7 @@
 
 #if BUILDFLAG(USE_EVERGREEN)
 #include "cobalt/browser/h5vcc_updater/h5vcc_updater_impl.h"
+#include "cobalt/browser/h5vcc_updater/h5vcc_updater_sideloading_impl.h"
```
yuying-y, line 42: This header can also be guarded by 

#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING

cobalt/browser/h5vcc_updater/h5vcc_updater_sideloading_impl.cc
```diff
@@ -0,0 +1,120 @@
...
+    updater_module->SetAllowSelfSignedPackages(allow_self_signed_packages);
+  }
+  std::move(callback).Run();
+}
+
+void H5vccUpdaterSideloadingImpl::GetAllowSelfSignedPackages(
+    GetAllowSelfSignedPackagesCallback callback) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  auto* updater_module = cobalt::updater::UpdaterModule::GetInstance();
+  if (!updater_module) {
```
hlwarriner, line 68: readability nit: unless there's a reason to do this that I'm missing, please choose a consistent conditional pattern and stick with it for similar methods in this class. For example, `GetAllowSelfSignedPackages()` checks `!updater_module` and early exits if that's the case, while `GetUpdateServerUrl()` checks `updater_module` in an `if-else`.

cobalt/browser/h5vcc_updater/BUILD.gn
```diff
@@ -13,10 +13,14 @@
 # limitations under the License.
 
 source_set("h5vcc_updater") {
-  sources = [
-    "h5vcc_updater_impl.cc",
-    "h5vcc_updater_impl.h",
-  ]
+  if (use_evergreen) {
+    sources = [
```
hlwarriner, line 19: We should probably only build the sideloading sources if the build allows evergreen sideloading and is not a release build.

I know we have a `cobalt_is_release_build` build variable. I don't see an `evergreen_allow_sideloading` build variable, in the branch or this PR, which is surprising to me. If you'd prefer to add that in a follow-up, then please add a todo comment here to conditionally build the sources once we have the build variable.

PR #8817, opened 2026-01-27
android: Make low-end-device-mode configurable

sherryzy:
Move the --enable-low-end-device-mode and --disable-rgba-4444-textures
command line arguments from a hardcoded default in
CommandLineOverrideHelper to a conditional check in CobaltActivity.
This allows downstream applications to control low-end-device-mode
enablement via feature flags through the getJavaSwitches() extension
point, while preserving the enabled-by-default behavior.

Bug: 477955599

### DIFF
--- a/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java
+++ b/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java
@@ -36,7 +36,11 @@ public class CommandLineOverrideHelperTest {
     @Test
     public void testDefaultCommandLineOverridesList() {
         List<String> overrides = CommandLineOverrideHelper.getDefaultCommandLineOverridesList();
-        assertThat(overrides.contains("--enable-low-end-device-mode")).isTrue();
+        assertThat(overrides.contains("--single-process")).isTrue();
+        assertThat(overrides.contains("--force-video-overlays")).isTrue();
+        assertThat(overrides.contains("--autoplay-policy=no-user-gesture-required")).isTrue();
+        assertThat(overrides.contains("--user-level-memory-pressure-signal-params")).isTrue();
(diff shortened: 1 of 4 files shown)


yell0wd0g:
I support this flag removal -- but how are we going to make sure there are no
unintended consequences? E.g. cc::ManagedMemoryPolicy [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:third_party/blink/renderer/platform/widget/compositing/layer_tree_settings.cc;l=184-189;drc=c5a8410e2008af297ae439cbf8985550466ea590) would start
taking the non-low-end mode, and I suspect there would be a number of similar
code paths around.

sherryzy:
@yell0wd0g I’m holding off on a direct merge to gate this change behind an experiment flag and assess its impact on topline metrics. We need a follow-up to evaluate whether other optimization paths should remain active to keep the memory improvements associated with this flag. See b/478989058 for the follow-up task.

sherryzy:
Updated this PR to gate the change by experiment flag.

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CommandLineOverrideHelperTest.java
```diff
@@ -36,7 +36,11 @@ public void setUp() {
     @Test
     public void testDefaultCommandLineOverridesList() {
         List<String> overrides = CommandLineOverrideHelper.getDefaultCommandLineOverridesList();
-        assertThat(overrides.contains("--enable-low-end-device-mode")).isTrue();
+        assertThat(overrides.contains("--single-process")).isTrue();
```
Awallky, line 39: I assume these are fixes to the existing test? 
If so, thanks for re-adding!

sherryzy: Yeah, I saw there are a few default commands are not added in the test, just added the missing ones.

PR #8821, opened 2026-01-27
Run splash screen browser tests on Android TV

linxinan-yt:
Enable LoadSplashVideo* browser tests on Android

1. Use a plain html to replace the splash.html to bypass the MediaSource since it is not yet supported by Android browser tests.
2. This change does not impact the test coverage on Linux, as LoadSplash* tests already covers MediaSource in splash.html. So the plain html will also run on Linux's browser tests.

Issue: 478334656

### DIFF
--- a/cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
+++ b/cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
@@ -93,8 +93,8 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
     H5vccSchemeURLLoaderFactory::SetSplashDomainForTesting(
         embedded_test_server()->base_url().spec());
 
-    // TODO(478334656): Before MediaSource is supported on Android's
-    // browser tests, use a simple HTML resource for testing.
+    // Use a plain html to simplify the test. This verifies the cache,
+    // hence bypass the MediaSource in the splash.html.
     auto test_resource_map = std::make_unique<GeneratedResourceMap>();
     LoaderEmbeddedResources::GenerateMap(*test_resource_map);
     const char kSafeSplashHtml[] = "<html><body>Safe</body></html>";


cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -93,6 +93,19 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
...
+    // production APK.
+    auto test_resource_map = std::make_unique<GeneratedResourceMap>();
+    LoaderEmbeddedResources::GenerateMap(*test_resource_map);
+    const char kSafeSplashHtml[] = "<html><body>Safe</body></html>";
+    FileContents safe_contents = {
+        reinterpret_cast<const unsigned char*>(kSafeSplashHtml),
+        sizeof(kSafeSplashHtml) - 1};
+    (*test_resource_map)["splash.html"] = safe_contents;
+    H5vccSchemeURLLoaderFactory::SetResourceMapForTesting(
+        test_resource_map.get());
```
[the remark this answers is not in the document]
Awallky: I think we can ignore this, since we clean this  up in the tearDown and setUp methods.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -93,6 +93,19 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
     H5vccSchemeURLLoaderFactory::SetSplashDomainForTesting(
         embedded_test_server()->base_url().spec());
 
+    // Inject a safe splash.html to avoid renderer crashes in the test
```
Awallky: Do we want to include this only in Android builds? Or is this also necessary for Linux?

linxinan-yt: This is used by LoadSplashVideoFrom* tests, which can run on both Linux and Android.

Awallky: Right, but my question is, does this work on Linux and only fails on Android or are we using a safe splash because it works on neither? 
If both, then could you clarify that in the PR message? If Android only, then we should add conditionals depending on the platform to only use safe splash on Android and use the real splash on Linux.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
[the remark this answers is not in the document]
linxinan-yt: Yes, this is what TearDown doing. I don't want to add them to the class member because the mock resouceMap is only for these two tests not all test in this suite. Also if mediaSource is supported, we even don't need this fake resouceMap.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Ah I see your question. 

1. These are tests for testing the storage cache API. They don't rely on the MediaSource.
2. I use a dummy page to replace splash.html, so that both Linux and Android can run this test.
3. This does not impact the test coverage on Linux, as the LoadSplash* have covered the MediaSource + Fetch in the splash.html.

I will add this to the PR descritpion.

PR #8831, opened 2026-01-27
android: Refactor StartupGuard disarming

zhongqiliang:
Move StartupGuard from dev.cobalt.util to dev.cobalt.shell to better
align it with shell-related components.

Simplify the initial disarming logic in CobaltActivity for non-YouTube
startup URLs. Introduce a StartupGuardNavigationObserver which
automatically disarms the guard once the primary main frame has
successfully committed a navigation. This change ensures the guard is
disarmed reliably after initial content has loaded.

Bug: 474529170

### DIFF
--- a/cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
+++ b/cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
@@ -6,7 +6,7 @@ import org.chromium.content_public.browser.WebContentsObserver;
 
 /**
  * An observer that monitors web content navigation to disarm the StartupGuard
- * once the primary page has successfully loaded.
+ * once the primary page has successfully loaded or begun navigating.
  */
 public class StartupGuardNavigationObserver extends WebContentsObserver {
     /**
@@ -19,10 +19,17 @@ public class StartupGuardNavigationObserver extends WebContentsObserver {
     }
 
     @Override
+    public void didRedirectNavigation(NavigationHandle navigation) {
+        // If the page starts to navigate, the app is functioning.
+        // Network-specific errors (e.g., DNS failure, offline) are handled by
+        // separate error-handling logic, so we disarm the guard here.
+        StartupGuard.getInstance().disarm();
+        destroy();
+    }
+
+    @Override
     public void didFinishNavigationInPrimaryMainFrame(NavigationHandle navigation) {
-        if (navigation.hasCommitted()) {
-            StartupGuard.getInstance().disarm();
-            destroy();
-        }
+        StartupGuard.getInstance().disarm();
+        destroy();
     }
 }


cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
```diff
@@ -0,0 +1,27 @@
...
+     * @param webContents The WebContents to observe.
+     */
+    public StartupGuardNavigationObserver(WebContents webContents) {
+        super(webContents);
+    }
+
+    @Override
+    public void didFinishNavigationInPrimaryMainFrame(NavigationHandle navigation) {
+        if (navigation.hasCommitted()) {
+            StartupGuard.getInstance().disarm();
```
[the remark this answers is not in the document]
zhongqiliang: thanks!

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -24,6 +22,8 @@ private static class LazyHolder {
         private static final StartupGuard INSTANCE = new StartupGuard();
     }
 
+    private static final String TAG = "cobalt";
```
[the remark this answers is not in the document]
zhongqiliang: disagree

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
```diff
@@ -0,0 +1,28 @@
...
+     * Constructs a new observer for the given WebContents.
+     *
+     * @param webContents The WebContents to observe.
+     */
+    public StartupGuardNavigationObserver(WebContents webContents) {
+        super(webContents);
+    }
+
+    @Override
+    public void didFinishNavigationInPrimaryMainFrame(NavigationHandle navigation) {
```
johnxwork, line 31: disarming the guard on this callback might defeat the purpose of StartupGuard. According to the API, this is called even when the navigation is aborted, in which case I think grey/white screen is expected.

zhongqiliang: disarm is called with navigation.hasCommitted() check, if it run into 404 or cancel, it won't disarm.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
```diff
@@ -264,6 +264,8 @@ private void updateNativeTabContents(WebContents webContents) {
             mWebContents.onShow();
         }
         mContentViewRenderView.setCurrentWebContents(mWebContents);
+
+        new StartupGuardNavigationObserver(mWebContents);
```
borongc, line 268: ATV has two shells (b/448910187), we should check if somehow app is slow, and CobaltActivity.java calls `mShellManager.launchShell()` with WebApp very late (which could be the case for low-end devices), would StartupGuard crash the app?

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
borongc: For storage migration, WebContents will be stop() for the task in [cobalt/shell/browser/migrate_storage_record/migration_manager.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/shell/browser/migrate_storage_record/migration_manager.cc;l=254).

We should verify if this will trigger this case.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
borongc: Also, right now Cobalt assumes `updateNativeTabContents()` is only called once during app startup.

However, WebApp team is under discussion that playing splash screen on-demand, which breaks the assumption (although it's safe to put StartupGuardNavigationObserver() here for now).

zhongqiliang: StartupGuardNavigationObserver cancel the StartupGuard crash when the page navigate away from Kabuki (in the case of a kabuki loader, hidesplashscreen is not called).

"if somehow app is slow, and CobaltActivity.java calls mShellManager.launchShell() with WebApp very late", it will crash if Kabuki failed to call hidesplashscreen within 1 min; it won't crash if Kabuki run normally to call hidesplashscreen within 1 min. this change won't affect that part.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
zhongqiliang: When will the migration code run? Does it run on the background or foreground? What will user see after WebContents stop() is called.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
borongc: Note that in https://github.com/youtube/cobalt/pull/8283, we don't have `hidesplashscreen` from WebApp.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
borongc: It will be a gray screen (or black screen). It is only triggered when you use C25 (with account sign-in), then update it to C26.

zhongqiliang: Discussed offline.

cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
zhongqiliang: Discussed offline

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuardNavigationObserver.java
```diff
@@ -0,0 +1,32 @@
...
+     * Constructs a new observer for the given WebContents.
+     *
+     * @param webContents The WebContents to observe.
+     */
+    public StartupGuardNavigationObserver(WebContents webContents) {
+        super(webContents);
+    }
+
+    @Override
+    public void didRedirectNavigation(NavigationHandle navigation) {
```
zhongqiliang, line 22: Note: kaido suggested to catch network errors with other code. 
If the page starts to navigate, it means the app is functioning, and if there is network error, we use other code to handle it.

PR #8833, opened 2026-01-27
cobalt: Refactor histogram fetching process for unit tests

yjzhang111:
Move histogram fetching and delta calculation logic from H5vccMetricsImpl to a new HistogramFetcher class. This refactoring ensures that histogram data snapshot logic can be tested by unit tests.

Bug: 467384307

### DIFF
--- a/cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
+++ b/cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
@@ -1,4 +1,4 @@
-// Copyright 2025 The Cobalt Authors. All Rights Reserved.
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,11 +17,11 @@
 #include "base/base64url.h"
 #include "base/logging.h"
 #include "base/metrics/histogram.h"
+#include "base/metrics/sample_vector.h"
 #include "base/metrics/statistics_recorder.h"
-#include "cobalt/browser/global_features.h"
-#include "cobalt/browser/metrics/cobalt_metrics_services_manager_client.h"
 #include "components/metrics/metrics_log.h"
-#include "components/metrics/metrics_service.h"
+#include "components/metrics/metrics_service_client.h"
+#include "components/metrics/metrics_state_manager.h"
 #include "components/metrics_services_manager/metrics_services_manager.h"
 #include "third_party/metrics_proto/chrome_user_metrics_extension.pb.h"
 #include "third_party/metrics_proto/cobalt_uma_event.pb.h"
@@ -34,11 +34,12 @@ HistogramFetcher::~HistogramFetcher() = default;
 std::string HistogramFetcher::FetchHistograms(
     metrics::MetricsStateManager* state_manager,
     metrics::MetricsServiceClient* service_client) {
-  // Synchronously fetch subprocess histograms that live in shared memory.
-  base::StatisticsRecorder::ImportProvidedHistogramsSync();
-
+  CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   CHECK(state_manager);
   CHECK(service_client);
+
+  // Synchronously fetch subprocess histograms that live in shared memory.
+  base::StatisticsRecorder::ImportProvidedHistogramsSync();
 
   if (state_manager->client_id().empty()) {
     return std::string();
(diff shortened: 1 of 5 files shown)


cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
```diff
@@ -0,0 +1,89 @@
...
+
+std::string HistogramFetcher::FetchHistograms(
+    metrics::MetricsStateManager* state_manager,
+    metrics::MetricsServiceClient* service_client) {
+  // Synchronously fetch subprocess histograms that live in shared memory.
+  base::StatisticsRecorder::ImportProvidedHistogramsSync();
+
+  CHECK(state_manager);
+  CHECK(service_client);
+
```
yell0wd0g, line 40: Should l.38 `base::StatisticsRecorder::ImportProvidedHistogramsSync();`
be moved here? To avoid importing if it's unnecessary.

cobalt/browser/h5vcc_metrics/histogram_fetcher.h
```diff
@@ -0,0 +1,51 @@
...
+  // Fetches all histograms, calculates the delta since the last call, and
+  // returns the result as a base64-encoded CobaltUMAEvent proto.
+  std::string FetchHistograms(metrics::MetricsStateManager* state_manager,
+                              metrics::MetricsServiceClient* service_client);
+
+ private:
+  // Stores the last snapshot of histogram samples.
+  std::map<std::string, std::unique_ptr<base::HistogramSamples>>
+      last_histogram_samples_;
+};
```
yell0wd0g, line 50: Add a `ThreadChecker` ?

cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
```diff
@@ -0,0 +1,89 @@
...
+  for (base::HistogramBase* const histogram :
+       base::StatisticsRecorder::GetHistograms()) {
+    auto samples = histogram->SnapshotSamples();
+    if (samples->TotalCount() == 0) {
+      continue;
+    }
+
+    auto it =
+        last_histogram_samples_.find(std::string(histogram->histogram_name()));
```
yell0wd0g: Having to make an `std::string` out of a `char*` every time is 
inefficient (there'll be a copy), and it makes me nervous that 
maybe under some circumstances it might be `nullptr` (a string
ctor with nullptr is undefined and in general bad). Looking at
`base::HistogramBase` I noticed that there's `name_hash()`, 
which is an `uint64_t`. If `std::string(histogram->histogram_name()`
is only used for indexing, could we make `last_histogram_samples_`
key a `uint64_t` and use `name_hash()` here? Less conversions.

cobalt/browser/h5vcc_metrics/histogram_fetcher.h
```diff
@@ -0,0 +1,51 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
yell0wd0g: nit: 2026 in the new files.

cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
```diff
@@ -0,0 +1,89 @@
...
+    if (it != last_histogram_samples_.end()) {
+      samples->Subtract(*it->second);
+    }
+    last_histogram_samples_[std::string(histogram->histogram_name())] =
+        histogram->SnapshotSamples();
+
+    if (samples->TotalCount() > 0) {
+      log.RecordHistogramDelta(histogram->histogram_name(), *samples);
+    }
```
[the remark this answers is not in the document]
yjzhang111: There's no Clone method or copy constructor for histogram snapshots. I tried a few things but all became quite convoluted. I moved the two snapshot calls back to back so that the window for race condition is minimal.

cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
```diff
@@ -0,0 +1,100 @@
...
+
+  // Synchronously fetch subprocess histograms that live in shared memory.
+  base::StatisticsRecorder::ImportProvidedHistogramsSync();
+
+  if (state_manager->client_id().empty()) {
+    return std::string();
+  }
+
+  ::metrics::ChromeUserMetricsExtension uma_proto;
+  metrics::MetricsLog log(state_manager->client_id(), /*session_id=*/1,
```
Awallky, line 49: Hard-coding the session ID seems not quite right. Wouldn't it be useful to track sessions here? Or do we expect that to be done via the main app instead?

cobalt/browser/h5vcc_metrics/histogram_fetcher_test.cc
```diff
@@ -0,0 +1,103 @@
...
+  TestingPrefServiceSimple prefs_;
+  metrics::TestMetricsServiceClient service_client_;
+  std::unique_ptr<cobalt::CobaltEnabledStateProvider> enabled_state_provider_;
+  std::unique_ptr<metrics::MetricsStateManager> state_manager_;
+};
+
+TEST_F(HistogramFetcherTest, FetchHistograms_CalculatesDeltaCorrectly) {
+  HistogramFetcher fetcher;
+
+  // stablish baseline and check for the histogram snapshot.
```
Awallky: nit: typo in establish

Awallky (approved):
LGTM
Could we add some browsertest in a follow-up?

cobalt/browser/h5vcc_metrics/histogram_fetcher.cc
yjzhang111: When we convert these histograms into Cobalt UMA event proto, we're only copying the histogram_event(). All other information are essentially stripped. I assume the other information like device info would be automatically attached in GEL.

PR #8839, opened 2026-01-28
cobalt: Fix rollback version logic

yjzhang111:
Update the experiment config manager to correctly identify cobalt version rollbacks. The previous logic used a simple string comparison for versions, which is incorrect for Cobalt's semantic version format (major.purpose.minor).

Implement a custom version comparison function that parses major and minor version numbers, ignoring the 'purpose' string. This ensures that experiment configurations are correctly invalidated when the client rolls back to an older major or minor version, preventing potential issues with stale experiment settings. Handles malformed version strings gracefully by not treating them as rollbacks.

Bug: 447453844

### DIFF
--- a/tools/metrics/histograms/metadata/cobalt/histograms.xml
+++ b/tools/metrics/histograms/metadata/cobalt/histograms.xml
@@ -74,6 +74,26 @@ Always run the pretty print utility on this file after editing:
   </summary>
 </histogram>
 
+<histogram name="Cobalt.Finch.RollbackDetected" enum="Boolean"
+    expires_after="never">
+  <owner>yijiazhang@google.com</owner>
+  <summary>
+    Records when a rollback is detected and prevented. This happens when the
+    cobalt version recorded for a previous Finch config is greater than the
+    current version of the Cobalt binary.
+  </summary>
+</histogram>
+
+<histogram name="Cobalt.Finch.VersionComparisonIsValid" enum="Boolean"
+    expires_after="never">
+  <owner>yijiazhang@google.com</owner>
+  <summary>
+    Records whether the version strings in the Finch config are valid and can be
+    parsed correctly. A false value indicates a malformed version string was
+    received.
+  </summary>
+</histogram>
+
 </histograms>
 
 </histogram-configuration>


cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -79,6 +83,38 @@ bool HasConfigExpired(PrefService* experiment_prefs) {
   return false;
 }
 
+// Custom version comparison for Cobalt. Returns true if version1 > version2.
+// Compares major and minor version numbers, ignoring the purpose string.
+// Format is assumed to be <major>.<purpose>.<minor>.
+bool IsVersionGreaterThan(const std::string& version1,
+                          const std::string& version2) {
+  auto parse_version = [](const std::string& v_str, int* major,
```
AlecsisDuarte, line 91: Couldn't we leverage base/version.h? I guess not because of the possibility of purpose being a string and not a number? If so, could you introduce a test that captures that it ignores the purpose being not a number?

AlecsisDuarte (approved):

cobalt/browser/experiments/experiment_config_manager.cc
yjzhang111: Yea base/version.h assumes integer versions like major.minor.patch. (I'm sure you know this already but) The middle part of Cobalt's browser version makes it tricky to reuse any existing utils.

Good catch. Tests added.

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -79,6 +83,39 @@ bool HasConfigExpired(PrefService* experiment_prefs) {
   return false;
 }
 
+// Custom version comparison for Cobalt. Returns true if version1 > version2.
+// Compares major and minor version numbers, ignoring the purpose string.
+// Format is assumed to be <major>.<purpose>.<minor>.
+bool ExperimentConfigManager::IsVersionGreaterThan(
```
Awallky: The way I see it, this should return 3 states: true, false, and invalid. If we get an invalid input, we should flag it. I think this will let us know if our versions are malformed. I would also suggest adding a histogram to track the invalid vs valid states.

yjzhang111: Thanks for reviewing! Updated.

PR #8840, opened 2026-01-28
tvos: Fix AnnexB to AVCC conversion for AUD NALUs

abhijeetk:
Previously, H.264 AnnexB keyframe processing for VideoToolbox only
skipped SPS/PPS NALUs. Optional NALUs, such as Access Unit
Delimiters (AUD), were not accounted for. This caused format
conversion failures when AUD NALUs were present in the stream.

This change ensures that both required (SPS/PPS) and optional (AUD)
NALUs are correctly stripped from the frame data prior to conversion.
The AvcParameterSets parser's combined_size_with_optionals_in_bytes_
field is now utilized for accurate frame preparation.

Bug: 471365322

### DIFF
--- a/starboard/tvos/shared/media/avc_av_video_sample_buffer_builder.mm
+++ b/starboard/tvos/shared/media/avc_av_video_sample_buffer_builder.mm
@@ -72,8 +72,14 @@ void AvcAVVideoSampleBufferBuilder::WriteInputBuffer(
       }
     }
     SB_DCHECK(parameter_sets.format() == AvcParameterSets::kAnnexB);
-    source_data += parameter_sets.combined_size_in_bytes();
-    data_size -= parameter_sets.combined_size_in_bytes();
+    size_t bytes_to_skip =
+        parameter_sets.combined_size_in_bytes_with_optionals();
+    if (bytes_to_skip > data_size) {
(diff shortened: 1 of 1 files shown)


starboard/tvos/shared/media/avc_av_video_sample_buffer_builder.mm
```diff
@@ -63,17 +63,20 @@ VideoConfig new_config(input_buffer->video_stream_info(),
     if (!new_config.is_valid()) {
       ReportError("Failed to parse video config.");
       return;
-    }
-    const auto& parameter_sets = new_config.avc_parameter_sets();
-    if (!video_config_ || video_config_.value() != new_config) {
-      video_config_ = new_config;
-      if (!RefreshAVCFormatDescription(parameter_sets)) {
-        return;
+    } else {
```
yell0wd0g: micro-nit: In Chromium in general, there is no `else` after a `return`.
Not sure about //starboard conventions though.

abhijeetk: Done.

abhijeetk:
cc: @rakuco @jkim-julie @Gyuyoung

PR #8845, opened 2026-01-28
cobalt/metrics: Implement Memory.Total.PrivateMemoryFootprint metric

Awallky:
Implement the collection of the private memory footprint metric.
This change leverages the Chromium memory_instrumentation service
to periodically request global memory dumps and record the total
private memory footprint as a UMA histogram. This provides crucial
data for monitoring memory usage in Cobalt. A new command-line
switch is added to allow configuration of the collection interval.

Note: Local testing was also done via `python3 cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py  --no-manage-cobalt --output-file=uma_output.txt`.

Test: `out/android-x86_devel/bin/run_cobalt_unittests --gtest_filter="*CobaltMetricsServiceClientTest*"`
Test: `out/android-x86_devel/bin/run_cobalt_browsertests --gtest_filter="*CobaltMetricsBrowserTest*"`

Bug: 479261728


cobalt/browser/metrics/cobalt_metrics_service_client.h
```diff
@@ -108,27 +112,17 @@ class CobaltMetricsServiceClient : public metrics::MetricsServiceClient {
...
-  // Virtual to be overridden in tests.
-  virtual void OnApplicationNotIdleInternal();
+  base::RepeatingTimer idle_refresh_timer_;
 
-  // Periodically tells UMA the app is not idle so that metrics payloads
-  // continue to be uploaded.
-  // TODO(cobalt, b/417477183): Consider removing this when user actions work in
-  // Kabuki.
-  void StartIdleRefreshTimer();
+  base::WeakPtrFactory<CobaltMetricsServiceClient> weak_ptr_factory_{this};
```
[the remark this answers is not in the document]
Awallky: Resolving these since I have made significant changes. Will request re-review from gemini.

cobalt/browser/global_features.cc
```diff
@@ -118,6 +121,7 @@ void GlobalFeatures::CreateMetricsServices() {
 
 void GlobalFeatures::CreateMetricsLocalState() {
   DCHECK(!metrics_local_state_);
+  base::ScopedAllowBlockingForTesting allow_blocking;
```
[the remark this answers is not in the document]
Awallky: This is basically a dup of another issue above. I'll resolve this to simplify the set of findings.

sherryzy:
I'm mid-review on this PR, but I've hit my time limit for today. I'll wrap up the rest tomorrow morning. Note: I noticed a pre-commit error—could you take a look at that in the meantime?

Awallky:
Note: I was also able to observe the histogram firing using this local UMA inspection tool:

`python3 cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py --no-manage-cobalt --output-file=uma_output.txt`

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -254,4 +287,31 @@ void CobaltMetricsServiceClient::SetMetricsListener(
...
+    // but the ProcessDump class doesn't expose raw_dump.
+    // Actually, looking at GlobalMemoryDump::ProcessDump, os_dump() returns *raw_dump_->os_dump.
+    // We should probably check if it's there before calling it if we could.
+    // But since it returns a reference, it's a bit tricky if it's null.
+    // Wait, let's look at the ProcessDump header again.
+    // It says: const mojom::OSMemDump& os_dump() const { return *raw_dump_->os_dump; }
+    // This is dangerous if os_dump is null.
+    
+    // In our case, RequestPrivateMemoryFootprint should ensure it's there for successful dumps.
+    total_private_footprint_kb += process_dump.os_dump().private_footprint_kb;
```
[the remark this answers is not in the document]
Awallky: There is no method of that name

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
```diff
@@ -195,11 +199,17 @@ def _get_histograms_from_file(file_path: str) -> list:
     ]
 
 
-def _get_chrome_guiding_metrics() -> list:
+def _get_chrome_guiding_metrics_for_memory() -> list:
   metrics = []
-  metrics.append('Memory.PartitionAlloc.MemoryReclaim')
```
sherryzy, line 200: What happens with these three removed metrics? We don't need them anymore?

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
```diff
@@ -195,11 +199,17 @@ def _get_histograms_from_file(file_path: str) -> list:
...
-  metrics.append('Memory.PartitionAlloc.PartitionRoot.ExtrasSize')
-  metrics.append('Memory.PartitionAlloc.PeriodicPurge')
+  # Desktop and Android
+  metrics.append('Memory.Browser.PrivateMemoryFootprint')
+  # Desktop only
```
sherryzy: What's the "Desktop" here? Do you mean linux? I would prefer to use linux if that's the case. Cobalt mostly works on the living room linux devices, not really "desktop".

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -254,4 +287,32 @@ void CobaltMetricsServiceClient::SetMetricsListener(
...
+void CobaltMetricsServiceClient::RecordMemoryMetrics(
+    memory_instrumentation::GlobalMemoryDump* global_dump) {
+  uint64_t total_private_footprint_kb = 0;
+  for (const auto& process_dump : global_dump->process_dumps()) {
+    // os_dump() can be null if the OS dump failed for this process.
+    // However, it returns a reference, so we must check via raw_dump if
+    // possible, but the ProcessDump class doesn't expose raw_dump. Actually,
+    // looking at GlobalMemoryDump::ProcessDump, os_dump() returns
+    // *raw_dump_->os_dump. We should probably check if it's there before
+    // calling it if we could. But since it returns a reference, it's a bit
```
sherryzy: These few line of comments doesn't look like needed. Remove?

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: I removed them mostly because I was following the metrics defined here - http://shortn/_fDDFtWPoOn.

They were some of the few metrics I saw firing on my machine at the time, hence why they made it in. We can keep them if you'd prefer, but just note that these are just default histograms to look up when no input file is given with a list of histograms of interest.

This tool is versatile enough that you can just specify whatever ones you want via an input text file.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: This is just pulled straight from the docs in http://shortn/_fDDFtWPoOn. But I agree with the suggestion to declare these differently for Cobalt as they likely will be specific to Cobalt and not re-use however other embedders record these metrics.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
Awallky: Oof. Good catch. Removed the comment as it is unnecessary.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: But since we have yet to figure out which ones will run on various platforms, having not implemented them yet, I went ahead and just removed the comments entirely for now. 
It will not harm anything anyways.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
```diff
@@ -195,11 +199,13 @@ def _get_histograms_from_file(file_path: str) -> list:
...
 
 
-def _get_chrome_guiding_metrics() -> list:
+def _get_chrome_guiding_metrics_for_memory() -> list:
   metrics = []
-  metrics.append('Memory.PartitionAlloc.MemoryReclaim')
-  metrics.append('Memory.PartitionAlloc.PartitionRoot.ExtrasSize')
-  metrics.append('Memory.PartitionAlloc.PeriodicPurge')
+  metrics.append('Memory.Browser.PrivateMemoryFootprint')
+  metrics.append('Memory.Gpu.PrivateMemoryFootprint')
```
kaidokert, line 205: @sherryzy does this actually work and give us a reliable GPU memory footprint ?

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -46,6 +48,9 @@ int CobaltBrowserMainParts::PreCreateThreads() {
 }
 
 int CobaltBrowserMainParts::PreMainMessageLoopRun() {
+  if (!memory_instrumentation::MemoryInstrumentation::GetInstance()) {
+    content::InitializeBrowserMemoryInstrumentationClient();
```
kaidokert: Are you sure this belongs here ? I see that it's normally done in `content/app/content_main_runner_impl.cc` within `ContentMainRunnerImpl::RunBrowser`. Do we need it earlier ?

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
yell0wd0g: It's supposed to be a very good quality signal: it's used in the 
guardails for Chrome experiments ([link](https://g3doc.corp.google.com/company/teams/chrome/guardrail_metrics/memory_usage.md?cl=head)).

Awallky: @kaidokert Only the metric I have added will work. We will need follow-ups for the others. 
Some are used only on a subset of platforms. Some are for Android, some for iOS, etc.

cobalt/browser/cobalt_browser_main_parts.cc
Awallky: I think when I added this, I thought it was not firing, since it takes it a minute or so to fire up. After it initializes, it *will* fire periodically. I should have made a test CLI switch to shorten the logging period in the prototype at least. 
It did seem to fire after removing it though.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
yjzhang111, line 272: Do we need to collect memory metrics before every uma upload? Not sure if I understood Chrome's impl correctly but it seems like they only do it upon browser start..?
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/metrics/chrome_browser_main_extra_parts_metrics.cc;l=1047;drc=72eb090be2c9a401d7f03a5b5f91c9e216e8bba3;bpv=1;bpt=1

Awallky: Great question! The answer is - no we do not.  I refactored things to run periodically on a BG task runner. Please take another look.

cobalt/browser/cobalt_browser_main_parts.cc
Awallky: It turns out that just waiting a bit longer was all that was needed. 
I thought it was not working since it was not firing, but I forgot about the delay in UMA upload period. 
I added a CLI switch to speed up how frequently this histogram fires and decoupled it from the UMA upload functionality. It now runs on its own task runner in a BG thread.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
Awallky: To answer your second question of "Chrome only collects this at startup".  Definitely not. They *initialize* it once at startup.

You will notice that [these two methods](https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/metrics/chrome_browser_main_extra_parts_metrics.cc;l=153-168;drc=1d1fcea6ebbc1c2c3d04ab5538c74f742c66444f) I have highlighted call each other recursively, at a delay. We now mimic this exact functionality after the latest change.

Awallky:
> lgtm. is this tested through the local uma collection through cdp script?

@yjzhang111  Yes, I confirmed it fired with the UMA polling script locally, and that it fired through the browsertests.
It does take a bit to fire, based on the memory polling frequency, but it will fire eventually. 
This is now orthogonal to the UMA upload cycles.

PR #6686, opened 2025-08-04
starboard/nplb: FakeGraphicsContextProvider doesn't need an SbWindow

yell0wd0g:
This CL/PR removes the SbWindow creation logic from FakeGraphicsContextProvider,
and also the provisions in NPLB build files to support SbWindow creation on
Android TV builds.

FakeGCP should not rely on SbWindow creation IIUC, unless -guess- it's to 
transitively initialize the windowing system in ways needed for graphics 
to operate and that'd look like relying on an implementation detail which
should be removed anyway.

I followed the few instances I could find of the created window being used
and they ended up in nothing, albeit partners might be using it.

This PR/CL is a followup to #6494.

Bug: 297808555

PR #8852, opened 2026-01-28
Revert "starboard/nplb: FakeGraphicsContextProvider doesn't need an S…

andrewsavage1:
…bWindow (#6686)"

This reverts commit c3a63553ca18451450906de05d3dcd0c28dc9832.

Bug: 459955810


borongc:
Would we revert this in main branch as well, or just 26.eap for now?

andrewsavage1:
> Would we revert this in main branch as well, or just 26.eap for now?

Yeah, we'll revert in both branches

borongc (approved):

PR #8860, opened 2026-01-29
cobalt: Use SharedRemote in CobaltContentRendererClient for thread safety

yell0wd0g:
Thsi CL makes CobaltContentRendererClient::window_provider_ a SharedRemote<> so 
that it can be destroyed on a different TaskRunner from where it was created.
This hits a DCHECK on exit (see bug). Gemini suggested this straightforward 
change.

Bug: 479296443


yell0wd0g:
Note that @borongc beat me to fix this on `main` #8880

PR #8861, opened 2026-01-29
android: refactoring network dialog check to use WebContentsObserver

ckim1346:
Removing the 204 ping and `CobaltConnectivityDetector.java` class in favor of using `WebContentsObserver` as the source of truth. Overriding functions in `cobalt_web_contents_observer` to monitor navigation results as well as implementing a timeout to catch weak wifi/DNS resolution errors. On a slow success, the network dialog will still pop up at the timeout, but the user can dismiss the dialog with the new dismiss button.

The dialog itself now has a new 'dismiss' button that only dismisses the dialog without any other functionality. The web contents are only reloaded on 'Try Again'. An `onKeyListener` was added for the back/escape button to only suspend the app when the back button is pressed, without dismissing the dialog or reloading the web contents. 

Bug: 479538263
Bug: 478284467


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -160,8 +163,19 @@ public void onClick(DialogInterface dialogInterface, int whichButton) {
   public void onDismiss(DialogInterface dialogInterface) {
     mDialog = null;
       CobaltActivity cobaltActivity = (CobaltActivity) mActivityHolder.get();
-      if (cobaltActivity != null && mResponse == CANCELLED) {
-        cobaltActivity.getStarboardBridge().requestSuspend();
+      if (cobaltActivity != null) {
+        if (mResponse != NEGATIVE) {
+          if (mResponse == CANCELLED) {
+            cobaltActivity.getStarboardBridge().requestSuspend();
```
johnxwork: Could you remind me why did we suspend?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -160,8 +163,19 @@ public void onClick(DialogInterface dialogInterface, int whichButton) {
   public void onDismiss(DialogInterface dialogInterface) {
     mDialog = null;
       CobaltActivity cobaltActivity = (CobaltActivity) mActivityHolder.get();
-      if (cobaltActivity != null && mResponse == CANCELLED) {
-        cobaltActivity.getStarboardBridge().requestSuspend();
+      if (cobaltActivity != null) {
+        if (mResponse != NEGATIVE) {
```
johnxwork: I think you just want
if (mResponse == POSITIVE) {
} else if (mResponse == CANCELLED) {
}

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+}
+
+void CobaltWebContentsObserver::DidFinishNavigation(
+    content::NavigationHandle* navigation_handle) {
+  if (!navigation_handle->IsInPrimaryMainFrame()) {
+    return;
+  }
+
+  // Only stop the navigation timer if the navigation finishes for
+  // a different document
```
johnxwork: Why is that? If Kabuki Browse does a redirect to Kabuki Watch, we would not stop the timer?

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+#include "content/public/browser/navigation_handle.h"
+#include "net/base/net_errors.h"
+#include "starboard/android/shared/starboard_bridge.h"
+#endif  // BUILDFLAG(IS_ANDROIDTV)
+
 namespace cobalt {
 
+#if BUILDFLAG(IS_ANDROIDTV)
+namespace {
+const int kNavigationTimeoutSeconds = 20;
```
johnxwork: I took a look at P95 startup latency and it's currently at ~14s excluding Cobalt cold start time. Can we set this to 30 seconds instead? Already talked to Kaido who's OK

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,71 @@
...
+#include "net/base/net_errors.h"
+#include "starboard/android/shared/starboard_bridge.h"
+#endif  // BUILDFLAG(IS_ANDROIDTV)
+
 namespace cobalt {
 
+#if BUILDFLAG(IS_ANDROIDTV)
+namespace {
+const int kNavigationTimeoutSeconds = 15;
+const int kJniErrorTypeConnectionError = 0;
```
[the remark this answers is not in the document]
johnxwork: The error is always kJniErrorTypeConnectionError? What's the point?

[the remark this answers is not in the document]
ckim1346: The `PlatformError.java` class takes in an error type that decides what kind of error to show. In its original and current implementation, 0 signifies a network error specifically; other error types are not handled/defined.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: If a user presses back while the dialog is active, the app is supposed to suspend. C25 behavior does the same b/438634236

cobalt/browser/cobalt_web_contents_observer.cc
ckim1346: As far as I could tell, the cobalt web contents observer only picked up on cobalt-level navigations. www.youtube.com/tv and variations (splash screen and sign-in flow). I might be wrong but I also observed that in DevTools: in C26, the URL only changes between those three while in C25 and older the URL does change when redirecting to Kabuki Watch, browse, etc. That's why the implementation of pendingNavigate() was difficult to implement in 26 from what I could tell

cobalt/browser/cobalt_web_contents_observer.cc
kaidokert: Yes - 30 is safer for sure.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: To have the suspend work correctly, we currently need a webContents reload on both POSITIVE and CANCELLED (but not NEGATIVE which is for the dismiss button). I could do something like

if (mResponse == POSITIVE || mResponse == CANCELLED) {
  reload
  if (CANCELLED): requestSuspend

if that's easier to read?

cobalt/browser/cobalt_web_contents_observer.cc
ckim1346: Done

cobalt/browser/cobalt_web_contents_observer.cc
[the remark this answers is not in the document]
kaidokert: This is not a critique on this PR specifically, but SbSystemPlatformErrorType defines `kSbSystemPlatformErrorTypeConnectionError=0`. It's .. not a great API design to be honest. Maybe we should make a ticket to rethink this for Starboard 18 separately.

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,71 @@
...
+#include "content/public/browser/navigation_handle.h"
+#include "net/base/net_errors.h"
+#include "starboard/android/shared/starboard_bridge.h"
+#endif  // BUILDFLAG(IS_ANDROIDTV)
+
 namespace cobalt {
 
+#if BUILDFLAG(IS_ANDROIDTV)
+namespace {
+const int kNavigationTimeoutSeconds = 15;
```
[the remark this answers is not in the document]
ckim1346: Isn't this why we use a namespace? Wouldn't adding `static` be redundant?

cobalt/browser/cobalt_web_contents_observer.cc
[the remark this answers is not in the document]
ckim1346: Filed a ticket: b/479657914

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
johnxwork: do you want to just implement these logic in button press block above and leave `onDimiss()` to pressing the back button on remote?

ckim1346: it was mostly to consolidate the reload logic as try again, open network settings, and the back button all reload. I can make it a separate function then to make it cleaner and more readable

cobalt/browser/cobalt_web_contents_observer.cc
johnxwork: I trust your observation. But why only stop the timer for navigation to different origins? I think this is only necessary if same-origin navigation does not trigger DidStartNavigation() but does trigger DidFinishNavigation(). Is that the case?

ckim1346: My reasoning was because there's only one timer for the observer. If a same page navigation starts, it could succeed easily without being indicative of the network condition. So if it's the same origin, I want the original navigation to handle the timeout

Ex.:
- navigation to b/1234 starts and timer starts
- b/1234 loads partially but is taking a long time
- user clicks on b/1234#comment1 which starts a new navigation, resetting the timer and returns successfully, which stops the timer
- b/1234 finishes but doesn't navigate successfully. but because b/1234#comment1 reset the timer and finished successfully, we don't see that the underlying b/1234 had a network issue

It's like if the splash page or sign in flow finish successfully but the underlying youtube.com/tv doesn't, which would lead to a black screen underneath.

yell0wd0g: The fact that reading the code doesn't quite explain how it's 
supposed to work or if it's hiding a complicated case in which
it might not work as expected/intended is a case for writing 
tests: unit tests and, ideally, content_browsertests. It's OK
if those are not running in CI entirely yet, the author will
be trusted that they pass locally :)

yell0wd0g: Even more convinced that we need example cases in the form of
cobalt_browsertests since I'm reading the WCO class and there's
a bazillion gotchas, e.g. `PrimaryPageChanged(Page& page)`
```
  // ................................................................. Prefer
  // listening to this method to listening to DidFinishNavigation and checking
  // NavigationHandle::IsInPrimaryMainFrame && !NavigationHandle::IsSameDocument
  // && NavigationHandle::HasCommitted (unless your code has to listen to
  // DidFinishNavigation for some other reason, in which case listening only to
  // DidFinishNavigation is recommended).
```
and then in `DidFinishNavigation`:
```
  // Listening to PrimaryPageChanged should be preferred to listening to
  // DidFinishNavigation and checking IsInPrimaryMainFrame, !IsSameDocument, and
  // HasCommitted.
```


Aside from it I'm reading `DidStartNavigation()`:
```
  // - There is no guarantee that DidFinishNavigation will be called
  // for any particular navigation before DidStartNavigation is called on the
  // next.
```

Would it be then ok for our purposes here that this happens (I don't
know, I'm just asking):
- t=0 start navigation -> we start the |timeout_timer_|
- t=1 another start navigation -> do nothing re |timeout_timer_|.
- t=2 finish navigation that started in t==0 -> we stop |timeout_timer_|.
- navigation that started in t==1 never finishes, but we have no
|timeout_timer_| running, so the user never sees the dialog.

IOW does this class cares only that _some_ navigation ever finishes,
or that _all_ navigations finish in less than `kNavigationTimeoutSeconds`?

ckim1346: Agreed, and I'm currently in the process of writing unit tests for a follow-up PR. 

As for `DidFinishNavigation` vs `PrimaryPageChanged`: it's true that we're checking `IsInPrimaryMainFrame` and `IsSameDocument` but the argument is that for the network check, we can't assume `HasCommitted`. `PrimaryPageChanged` and `HasCommitted` only catches successful page changes and wouldn't catch network errors that are thrown before a navigation commits ie. `net::ERR_CONNECTION_TIMED_OUT` and `net::ERR_NAME_NOT_RESOLVED`.

That's a fair point; I was under the assumption that www.youtube.com/tv would always precede same document pages like the splash screen (www.youtube.com/tv/splash) or the sign in page (www.youtube.com/tv?[]) but after some local testing, it's possible that splash goes first when coming back from the oauth (accounts.google.com/o/oauth2 + accounts.google.com/signin/ytlrtec) sign in flow. Perhaps the most catch-all solution is to have every navigation have its own timer? `RaisePlatformError()` prevents the raising of a dialog if one already exists.

yell0wd0g: Re. the `PrimaryPageChanged` that makes sense -- would you mind adding all that 
explanation as a comment for later developers? Thanks!

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -159,10 +164,13 @@ public void onClick(DialogInterface dialogInterface, int whichButton) {
   @Override
   public void onDismiss(DialogInterface dialogInterface) {
     mDialog = null;
+    if (mResponse == CANCELLED) {
       CobaltActivity cobaltActivity = (CobaltActivity) mActivityHolder.get();
-      if (cobaltActivity != null && mResponse == CANCELLED) {
+      if (cobaltActivity != null) {
         cobaltActivity.getStarboardBridge().requestSuspend();
+        reloadWebContents(cobaltActivity);
```
johnxwork: This reload was not there in the C25 implementation? https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java;drc=c788fb906621bc5fcace3d0a073b2ea7603dcff9;l=169

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+              << " not in primary mainframe, returning";
+    return;
+  }
+  LOG(INFO) << "DidStartNavigation: navigation to " << handle->GetURL()
+            << " in primary mainframe";
+
+  // Start a navigation timer with a timeout callback to raise a
+  // network error dialog
+  timeout_timer_.Stop();
+  timeout_timer_.Start(
```
johnxwork, line 56: sorry for the back and forth on this: I just took another look on the definition of same document navigation and we can safely ignore those: like don't anything if it's same document.

ckim1346: no worries. if we ignore same document navigations and say splash screen goes first after the oauth flow and returns `net::OK`, I believe this could lead to a black screen. 

/tv/splash (starts timer) --> `net::OK` (ends timer)
/tv?2945719 (same document, no timer) --> `net::ERR` --> black screen

Since I believe through investigation yesterday, we saw that they have the same `location.href` value

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Yeah, C25 has the dialog reappear when the user resumes the app: during resume `NavigatePendingURL` is called, starting a new navigation to the previous URL, re-triggering the dialog if the user doesn't have internet.

In our current implementation, back button dismisses the dialog and suspends but when we resume, it's a black screen since we don't re-navigate. I'll remove the dismiss and web contents reload for the network settings button; how should we handle the back button?

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+}
+
+void CobaltWebContentsObserver::DidFinishNavigation(
+    content::NavigationHandle* navigation_handle) {
+  if (!navigation_handle->IsInPrimaryMainFrame()) {
+    return;
+  }
+
+  timeout_timer_.Stop();
+  int net_error_code = navigation_handle->GetNetErrorCode();
```
yell0wd0g: Shouldn't this variable be a `net::Error`? 

https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:content/public/browser/navigation_handle.h;l=306?q=navigationhandle%20GetNetErrorCode&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain


To avoid these kind of forced-yet-silent casts, use auto. And
I think this could be const:
> const auto net_error_code = navigation_handle->GetNetErrorCode();

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+  if (net_error_code != net::OK && net_error_code != net::ERR_ABORTED) {
+    LOG(INFO) << "DidFinishNavigation: Raising platform error with code: "
+              << net_error_code;
```
yell0wd0g: This would just print "123" or some other number (I don't see
`operator<<` defined for `net::Error` and in any case in the 
current code, it's a forced-int). Consider adding here e.g. `<< net::ErrorToString(net_error_code)` to make this statement
readable.

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+void CobaltWebContentsObserver::RaisePlatformError() {
+  JNIEnv* env = base::android::AttachCurrentThread();
+  starboard::StarboardBridge* starboard_bridge =
+      starboard::StarboardBridge::GetInstance();
+
+  // Don't raise a new platform error if one is already showing
+  if (starboard_bridge->IsPlatformErrorShowing(env)) {
+    return;
+  }
+  starboard_bridge->RaisePlatformError(env, kJniErrorTypeConnectionError, 0);
```
yell0wd0g, line 90: nit: Could we consider instead

```
  if (!starboard_bridge->IsPlatformErrorShowing(env)) {
    starboard_bridge->RaisePlatformError(env, kJniErrorTypeConnectionError, 0);
  }
```

?
It's one line less, albeit uses negative logic which is always worse.

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -14,12 +14,77 @@
...
+    LOG(INFO) << "DidFinishNavigation: Raising platform error with code: "
+              << net_error_code;
+    RaisePlatformError();
+  }
+}
+
+void CobaltWebContentsObserver::RaisePlatformError() {
+  JNIEnv* env = base::android::AttachCurrentThread();
+  starboard::StarboardBridge* starboard_bridge =
+      starboard::StarboardBridge::GetInstance();
```
yell0wd0g: nit: In cases like this, you can use `auto` as the type since
the type is clear from the getter and would make the statement
more legible.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
johnxwork: I'd expect the back button to be handled as "dismiss". WDYT?

ckim1346: ie moving away from C25? b/438634236. I think both make sense although back = dismiss could lead to unintentional white/black screens so suspend might be the safer option.

cobalt/browser/cobalt_web_contents_observer.cc
ckim1346: Done

cobalt/browser/cobalt_web_contents_observer.cc
ckim1346: Done

cobalt/browser/cobalt_web_contents_observer.cc
ckim1346: ack but I do think current is more readable

cobalt/browser/cobalt_web_contents_observer.cc
ckim1346: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: added a key event listener for the back button to only suspend without dismissing the dialog or reloading web contents

PR #8865, opened 2026-01-29
Fix  MediaKeys.GetMetrics  failures in RDK

santoshmahto7:
Bug: 474508805

**Issue[RDK]:**  
mediakeys.getmetrics fails() as
- C26 : FAIL
- c25 newly built (25.lts.1+)  FAIL
- c25 with default RDK image:  PASS

This PR fixes the above YTS failure


lamdbui:
@niranjanyardi , can you help pull in the right folks on the Cobalt side to take a look?

third_party/blink/renderer/platform/media/cdm_result_promise.h
```diff
@@ -131,7 +131,7 @@ inline void CdmResultPromise<std::string>::resolve(const std::string& result) {
       key_system_uma_prefix_ + kTimeToResolveUmaPrefix + uma_name_,
       base::TimeTicks::Now() - creation_time_);
 
-  web_cdm_result_.CompleteWithString(WebString::FromUTF8(result));
+  web_cdm_result_.CompleteWithString(WebString::FromUTF8(result.data()));
```
sideb0ard: Hmm, this seems unsafe to me, changing a string to const char* for parsing unknown data. If the returned data has any null characters, we would be truncating the data, potentially breaking a CDM session.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
```diff
@@ -761,6 +761,16 @@ int DrmSystemOcdm::Decrypt(const std::string& id,
 }
 
 const void* DrmSystemOcdm::GetMetrics(int* size) {
+  // Taken from RDK patch:0007-GetMetrics-YT24-changes.patch
```
niranjanyardi: Is it possible to link the upstream code/change , now that it's on opensource github? https://github.com/rdkcentral/larboard

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
```diff
@@ -761,6 +761,16 @@ int DrmSystemOcdm::Decrypt(const std::string& id,
...
+  // This is used since below call to g_ocdmGetMetricSystemData i.e opencdm_get_metric_system_data
+  // throws NOT_IMPLEMENTED error.
+  // FIXME: Ensure valid cdm API is used here to get_metrics.
+  OpenCDMError result =  opencdm_get_metrics(ocdm_system_, metrics_data_);
+  if (ERROR_NONE == result) {
+    *size = metrics_data_.size();
+    return (void*)metrics_data_.data();
```
kjyoun: style nit: please use c++ style type cast

https://google.github.io/styleguide/cppguide.html#Casting

niranjanyardi:
> **Issue[RDK]:** mediakeys.getmetrics fails() as
> 
> * C26 : FAIL
> * c25 newly built (25.lts.1+)  FAIL
> * c25 with default RDK image:  PASS
> 
> This PR fixes the YTS issue, but the YTS tests are very minimal and don't verify that the metrics data are as expected.
> 
> To ensure all is good with the GetMetrics call, the above PR needs more attention from respective module experts since the issues are addressed at multiple levels (RDK, Blink, Media) in PR as:
> 
> * **RDK:**
>   ```
>   starboard(drm_system_ocdm.cc::GetMetrics) --> opencdm_get_metric_system_data(open_cdm.h)--> MetricData-->goes to OCDM plugins)
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   
>   * The CDM API call used is `opencdm_get_metric_system_data`, which doesn't seem to be implemented insidethe  CDM binary and returns
>     **NOT_IMPLEMENTED** error.
>   * In RDK(C25) starboard has been patched(0007-GetMetrics-YT24-changes.patch)[1] within yocto recipe to use another
>     API(`opencdm_get_metrics`), which returns metrics data. So the default C25 in the RDK image works.
>   
>   I applied the same patch in starboard(C26), and `drm_system_ocdm.cc::GetMetrics` seems to work and return string blobs, but it had an issue passing this data to blink layers because it was non-UTF8.
> * **BLINK:**
>   
>   * In blink, we convert data to WebString using  `WebString::FromUTF8(result)` to pass to JS/blink, which causes the data size to be **zero** since the string contains non-UTF8.
>   * In C25, this is handled differently; we use `std::vector<uint8_t>` instead of a string to pass, so no issue there.
> * **Consideration:**
>   
>   * In RDK, multiple yocto patches have been applied (both client and server side) to support the `getMetrics` API, and I am not sure which CDM API is the correct way to get metrics data?
>     ```
>     ./lib32-wpeframework-clientlibraries/4.4-r0/0001-RDK-add-GetMetrics-API-in-IPC-module-1-1-R4.patch
>     ./lib32-libloader-app/25.lts.stable+gitAUTOINC+e935b5ca17_8fe50f920f-r0/0007-GetMetrics-YT24-changes.patch
>     ./lib32-rdkservices/3.0+gitAUTOINC+b3996e014a-r1/0001-Add-a-new-metrics-punch-through-on-the-OCDM-framework-rdkservice.patch
>     ./lib32-rdkservices/3.0+gitAUTOINC+b3996e014a-r1/git/patches/0001-RDK-implement-GetMetric-API-at-OCDM-server-side-1-1.patch
>     ./lib32-wpeframework-interfaces/4.4-r0/git/patches/0001-RDK-add-new-API-GetMetrics-1-1-R4.patch
>          0007-GetMetrics-YT24-changes.patch
>     ```
>   * Should we also pass metrics data as std::vector<uint8_t> rather than WebString inside blink as done in C25 ?

jsut a suggestion , instead of adding the comments all here it might be better to add them near the appropriate code and then tag the code-owners/ me for reviewing the comments.

niranjanyardi:
> @niranjanyardi , can you help pull in the right folks on the Cobalt side to take a look?

done, added code-owners

third_party/blink/renderer/platform/media/cdm_result_promise.h
santoshmahto7: yeah, this change isn't required now. In this initial PR, metrics data was  NOT encoded as base64, so the data remain as a non-UTF8 blob, which was causing an issue. So the change here.
But as said, it is not required now.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
santoshmahto7: I searched on `github.com/rdkcentral,` but it seems the GitHub repo doesn't have  Amlogic-specific Yocto layer. The actual path for this patch in Amlogic repo was `./meta-rdk-aml/recipes-extended/cobalt/files/0007-GetMetrics-YT24-changes.patch`

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
santoshmahto7: yes

santoshmahto7:
@niranjanyardi  I found one issue with the current patch  and fixed it in the recent update, now some of the comments are outdated, so I will remove it from the description

third_party/blink/renderer/platform/media/cdm_result_promise.h
lamdbui: Looks like this change isn't in the latest patch any longer, so can resolve.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
lamdbui: Looks like this isn't in the latest patch and we can resolve this.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
lamdbui: Looks like this isn't in the latest patch and we can resolve this.

kaidokert (commented):
Overall looks good / correct to me. @sideb0ard could you take a look from Media side ?

media/mojo/clients/mojo_cdm.cc
```diff
@@ -435,8 +435,12 @@ void MojoCdm::OnMetricsReceived(
     uint32_t promise_id,
     const absl::optional<std::string>& metrics_string) {
   DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  cdm_promise_adapter_.ResolvePromise(promise_id,
-      metrics_string.value_or(std::string()));
+  if (metrics_string) {
+    cdm_promise_adapter_.ResolvePromise(promise_id, metrics_string.value_or(std::string()));
+  } else {
+    cdm_promise_adapter_.RejectPromise(promise_id, CdmPromise::Exception::NOT_SUPPORTED_ERROR, 0,
+                  "GeMetrics() is not supported.");
```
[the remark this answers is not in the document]
kaidokert: Lets fix

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
```diff
@@ -552,6 +556,14 @@ DrmSystemOcdm::DrmSystemOcdm(
     } else {
       SB_LOG(INFO) << "No opencdm_get_metric_system_data.";
     }
+
+    g_ocdmGetMetrics =
+        (OcdmGetMetrics)dlsym(RTLD_DEFAULT, "opencdm_get_metrics");
```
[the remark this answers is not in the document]
kaidokert: Valid comment - lets use reinterpret_cast

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
```diff
@@ -761,13 +773,14 @@ int DrmSystemOcdm::Decrypt(const std::string& id,
...
 const void* DrmSystemOcdm::GetMetrics(int* size) {
-  if ( !g_ocdmGetMetricSystemData )
-    return nullptr;
 
   SB_CHECK(ocdm_system_ != nullptr);
 
   const int kMaxRetry = 5;
   for (int i = 0; i < kMaxRetry; ++i) {
+    if ( !g_ocdmGetMetricSystemData )
+      break;
```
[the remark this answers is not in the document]
kaidokert: @gemini-code-assist i think this was moved intentionally here, as the global variable may get invalidated in any of the retry loops ?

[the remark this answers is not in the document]
santoshmahto7: geimini  has an incorrect interpretation since the loop will break if `g_ocdmGetMetricSystemData` is nullptr, so there is no question of checking in every iteration.
Thus statement from gemini is incorrect:
 "checking its validity on every iteration of the loop is"

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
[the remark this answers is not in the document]
santoshmahto7: ack

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
```diff
@@ -792,6 +805,24 @@ const void* DrmSystemOcdm::GetMetrics(int* size) {
...
+  // Use `opencdm_get_metrics` when `opencdm_get_metric_system_data` fails to
+  // get mertics data. This is observed in RDK/Amlogic where
+  // `opencdm_get_metric_system_data` returns with error code
+  // `INTERFACE_NOT_IMPLEMENTED`.
+  if (g_ocdmGetMetrics && metrics_.size() == 0) {
+    std::string buffer;
+    auto rc = g_ocdmGetMetrics(ocdm_system_, buffer);
+    if (rc == ERROR_NONE) {
+      uint32_t buffer_length = buffer.size();
+      uint16_t out_length = (((buffer_length * 8) / 6) + 4) * sizeof(TCHAR);
```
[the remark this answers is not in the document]
santoshmahto7: Base64Encode expects arguments as uint16_t  so let's use uint16_t for length

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/drm/drm_system_ocdm.cc
```diff
@@ -792,6 +805,24 @@ const void* DrmSystemOcdm::GetMetrics(int* size) {
...
+      metrics_.resize(out_length, '\0');
+      out_length = WPEFramework::Core::URL::Base64Encode(
+          reinterpret_cast<uint8_t*>(buffer.data()), buffer_length,
```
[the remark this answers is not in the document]
santoshmahto7: Done

media/mojo/clients/mojo_cdm.cc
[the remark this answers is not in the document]
santoshmahto7: Ah.. Done

sideb0ard (approved):
yeah, looks good from Media side, change to media/mojo/clients/mojo_cdm.cc is much simplified. thanks!

PR #8873, opened 2026-01-29
Move the signal handlers in SbRunStarboardMain

y4vor:
Move the signal handlers from StarboardMain to SbRunStarboardMain. This way the loader_app gets signal handlers properly registered. The current location in seem to be only for the plugin.

Issue: 478966141

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/main_rdk.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/main_rdk.cc
@@ -29,87 +29,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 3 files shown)


niranjanyardi:
Great job finding and fixing this!

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/main_rdk.cc
```diff
@@ -29,87 +29,10 @@
 // See the License for the specific language governing permissions and
```
niranjanyardi, line 29: there are 3 RDK platforms which share this code . If anything , yavor's change is an improvement which fixes all platforms IMO.

wdyt @jonastsai ?

PR #8875, opened 2026-01-29
cobalt: Use SharedRemote in CobaltContentRendererClient for thread safety

yell0wd0g:
Convert CobaltContentRendererClient::window_provider_ to
mojo::SharedRemote. This ensures thread-safe destruction,
preventing a DCHECK on exit that occurred when the remote
was destroyed on a different TaskRunner than it was created.

Same as #8860 but on main.

Verified locally:
> p="linux-x64x11-modular" ; c="devel" ; t="cobalt_loader" ; autoninja -C out/${p}_${c} $t && gdb -ex="set confirm off" -ex=r -ex=bt --args ./out/${p}_${c}/$t

hits a DCHECK on exit on ToT, the exit is clean with the patch.

Bug: 479296443

### DIFF
--- a/.github/actions/depot_tools/action.yaml
+++ b/.github/actions/depot_tools/action.yaml
@@ -54,7 +54,12 @@ runs:
     - name: Clone Depot Tools
       env:
         DEPOT_TOOLS_BASE_PATH: ${{ steps.paths.outputs.base_path }}/depot_tools
-      run: git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git "${DEPOT_TOOLS_BASE_PATH}"
+      run: |
+        if [ -d "${DEPOT_TOOLS_BASE_PATH}" ]; then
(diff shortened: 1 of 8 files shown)


yell0wd0g:
Thanks @jellefoks for the repro steps! Made this a breeze. PTAL!

yell0wd0g:
@y4vor PTAL

yell0wd0g:
Ehmm sorry apologies - this fix doesn't apply to `main` currently since
on this branch the `windo_provider` is not kept as member variable, is 
used on the spot (see code [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/renderer/cobalt_content_renderer_client.cc;l=125-133;drc=bb8109d73ceb123f69bbf906e51550210db5906c)). We still need the fix
but only on `26.eap` (#8860 ). Let me close this PR to avoid confusion.

PR #8876, opened 2026-01-29
Add static image fallback for native splash screen

tasunnn:
Implement static image fallback for the native splash screen for devices that don't support VP9 video playback. This ensures a visual splash screen is always presented even on low spec devices. Similar to splash video, the image can also be cached locally after first start, which could then be loaded by h5vcc loader for display in cases of cache hits. 

Test: cobalt_browsertests_runner --gtest_filter=H5vccSchemeURLLoaderFactoryBrowserTest*
Bug: 458074360

### DIFF
--- a/cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
+++ b/cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
@@ -237,7 +237,8 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
       mime_type_ = kMimeTypeVideoWebM;
     }
 
-    bool is_cacheable_type = (mime_type_ == kMimeTypeImagePng || mime_type_ == kMimeTypeVideoWebM);
+    bool is_cacheable_type =
+        (mime_type_ == kMimeTypeImagePng || mime_type_ == kMimeTypeVideoWebM);
     if (is_cacheable_type && browser_context_) {
       ReadSplashCache(key);
       return;


cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -194,6 +242,20 @@ IN_PROC_BROWSER_TEST_F(H5vccSchemeURLLoaderFactoryBrowserTest,
   EXPECT_EQ("Dimensions: 853x480", EvalJs(shell(), CheckVideoDimension(false)));
 }
 
+#if BUILDFLAG(IS_ANDROID)
+#define MAYBE_LoadSplashHtml DISABLED_LoadSplashHtml
+#else
+#define MAYBE_LoadSplashHtml LoadSplashHtml
```
linxinan-yt: You need a new name for this test case.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -43,25 +57,51 @@
...
+              throw new Error('Network response was not ok');
+            }
+            const imageBlob = await response.blob();
+            const imageUrl = URL.createObjectURL(imageBlob); 
+            placeholderEl.style.backgroundImage = `url(${imageUrl})`; 
+            placeholderEl.style.opacity = '100%';
+            // timeout? 
+            setTimeout(() => {
+              afterImage && afterImage();
+            }, 1500);
```
linxinan-yt: If we don't set timeout, will the splash screen get stopped and switched to main app?

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -43,25 +57,51 @@
...
-        playSplashAnimation(/*afterVideoStart=*/ showPlaceholderImage);
+        playSplashAnimation(/*afterVideoStart=*/ showPlaceholderImage, onSplashEnd);
       }
 
       /**
        * Plays the splash animation video in the given video element.
        * @param afterVideoStart Callback function when the video starts playing.
        */
-      function playSplashAnimation(afterVideoStart) {
+      async function playSplashAnimation(afterVideoStart, afterImage) {
```
linxinan-yt: why async?

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -43,25 +57,51 @@
...
+        const params = new URLSearchParams(window.location.search);
+        if (!params.has('cache')) {
+          params.append('cache', 'default');
+        }
+
+        if (!MediaSource.isTypeSupported("video/webm; codecs=\"vp9\";")) {
```
linxinan-yt: Seems the logic is getting complicated, let's move this to another function.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -88,10 +126,27 @@
...
+       * @param afterVideoStart Callback function when the video starts playing.
+       * @param afterImageLoad Callback function for static image.
+       */
+      function playSplashAnimation(afterVideoStart, afterImageLoad) {
+        const params = new URLSearchParams(window.location.search);
+        if (!params.has('cache')) {
+          params.append('cache', 'default');
+        }
+
+        if (MediaSource.isTypeSupported("video/webm; codecs=\"vp9\";")) {
```
linxinan-yt: Should it be !MediaSource.isTypeSupported(...)?

We handle static image if the device does not support VP9.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -43,35 +57,59 @@
...
         const mediaSource = new MediaSource();
         mediaSource.addEventListener('sourceopen', async () => {
           const sourceBuffer = mediaSource.addSourceBuffer('video/webm; codecs="vp9"');
-          const params = new URLSearchParams(window.location.search);
-          if (!params.has('cache')) {
-            params.append('cache', 'default');
-          }
           // The default splash video is splash.webm. If it does not exist in
           // the given cache, "fallback" parameter specifies which built-in video
           // to use.
```
linxinan-yt: This applies to the static image. So better move to line 138.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
tasunnn: Right. I modified the flag when testing and forgot to change it back, let me fix that

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -57,10 +57,58 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
...
+          const placeholder = document.getElementById('placeholder');
+          if (!placeholder) {
+            return 'No placeholder element found';
+          }
+
+          // check if image is displayed
+          let style = window.getComputedStyle(placeholder);
+          let retries = 0;
+          while (style.backgroundImage === 'none' && retries < 15) {
+            await new Promise(r => setTimeout(r, 100));
```
linxinan-yt: Set waiting time is not preferred in browser test because it relies on the spec of the test bot, aka the test could be flaky. You call this after `NavigateToURL(shell(), splash_url)`, so the placeholder should be rendered on the canvas. What error did you see if removing this timeout?

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -43,35 +57,59 @@
...
+          if (!response.ok) {
+            throw new Error('Network response was not ok');
+          }
+          return response.blob();
+        })
+        .then(imageBlob => {
+          const imageUrl = URL.createObjectURL(imageBlob);
+          placeholderEl.style.backgroundImage = `url(${imageUrl})`;
+          placeholderEl.style.opacity = '100%';
+          afterImageLoad && afterImageLoad();
```
borongc: How long this static image will be displayed?

tasunnn: I didn't have a timeout here for image, but it eventually finish displaying and app webcontent switches, do we want to set a timeout?

borongc: We should align with the length of the video (i.e., 1s here). This makes Cobalt handles this WebContents using the same logic regardless if it plays video or display a static image.

If we don't set the timeout here, it will rely on Cobalt's timeout (i.e., 1.5 second by default) to switch the contents, and if we set some large value, it will be different between playing video and displaying the image. For video right now, it will switch to main app frame if video is ended regardless the timeout is 1.5s.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
tasunnn: I got a "Fail to fetch" error if not setting wait time, it's probably taking little longer to set the url and render the image, would it still be fine to wait or there's any other workaround?

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
tasunnn: Ok I see, I will add a 1 second timeout for image display

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -230,6 +231,12 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
     // For html file, return from embedded resources.
     if (base::EndsWith(key, ".html", base::CompareCase::SENSITIVE)) {
       mime_type_ = kMimeTypeTextHtml;
+    } else if (base::EndsWith(key, ".png", base::CompareCase::SENSITIVE)) {
+      mime_type = kMimeTypeImagePng;
```
linxinan-yt: should be the class member.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Let's try to customize the `afterImageLoad` and avoid setting timeout:

```
 std::string CheckImageDimension() {
  return R"(
    (async () => {
      // 1. Mock the support to force the Image Fallback path
      MediaSource.isTypeSupported = () => false;

      return await new Promise((resolve) => {
        // This is the 'afterImageLoad' callback
        const onFinish = () => {
          const placeholder = document.getElementById('placeholder');
          if (!placeholder) return resolve('No placeholder element found');

          const style = window.getComputedStyle(placeholder);
          const bg = style.backgroundImage;
          const opacity = style.opacity;

          if (bg === 'none' || bg === '') return resolve('No image is displayed');
          if (parseFloat(opacity) < 1) return resolve('Image does not have 100% opacity');

          // Extract the blob URL from url("blob:...")
          const url = bg.replace(/^url\(["']?/, '').replace(/["']?\)$/, '');
          
          const image = new Image();
          image.onload = () => resolve('Dimensions: ' + image.naturalWidth + 'x' + image.naturalHeight);
          image.onerror = () => resolve('Image load error');
          image.src = url;
        };

        // 2. Trigger the animation. EvalJs will wait for resolve() to be called.
       // The splash should be an image with expected dimensions.
        playSplashAnimation(null, onFinish);
      });
    })();
  )";
}
```

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -58,10 +58,55 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
...
+          }
+
+          // use checkImage as callback to avoid setting timeout
+          playSplashAnimation(null, checkImage);
+        });
+      })();
+    )";
+  }
+
   std::string CheckVideoDimension(bool is_type_supported) {
```
linxinan-yt: Since now we have image and video, "is_type_supported" is not accurate. Would you change it to e.g. is_4k_supported?

linxinan-chops (approved):
LGTM % comment

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
tasunnn: sure

Awallky:
You can remove most of the reviewers here, since this only affects `//cobalt/shell`. That can be confined to ATV and/or cobalt embedders as the reviewer set.

cobalt/shell/BUILD.gn
```diff
@@ -461,6 +461,10 @@ action("embed_loader_resources_as_header_files") {
     # For low spec TV, 853x480. Copied from
     # google3/googledata/html/external_content/gstatic/ytlr/splash_screen/splash_480p.webm.
     "embedded_resources/loader_embedded_resources/splash_480.webm",
+
+    # For devices that do not support VP9. Copied from
+    # google3/video/youtube/web/living_room/kabuki/splash/chrobalt/splash_video_placeholder.png.
```
Awallky, line 467: I would not leave Google code paths here. If you must, can you link to the file using a shortn link?

andrewsavage1: I'm not sure it matters that much, there are already a lot of google3 path references in chromium

Awallky, line 467: Also, since this is from Google, I think you need to follow [this procedure](http://shortn/_9qZmqls7EY) in order to be able to port it to this Repo.

tasunnn: found the image link for external use, maybe we can replace with this? google3/googledata/html/external_content/gstatic/ytlr/splash_screen/splash_video_placeholder.png

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -225,6 +270,20 @@ IN_PROC_BROWSER_TEST_F(H5vccSchemeURLLoaderFactoryBrowserTest,
   EXPECT_EQ("Dimensions: 853x480", EvalJs(shell(), CheckVideoDimension(false)));
 }
 
+#if BUILDFLAG(IS_ANDROID)
```
Awallky, line 273: Could you add a bug to enable this on ATV?

cobalt/shell/BUILD.gn
Awallky: Out of an abundance of caution, I think we should at least consult with the link in a [related comment](https://github.com/youtube/cobalt/pull/8876/changes#r2766246147) about best practices for this type of change.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
tasunnn: Xinan already filed a bug above

Awallky: Ah, I see.
Marking this resolved.

cobalt/shell/BUILD.gn
tasunnn: maybe @borongc can chime in since he committed the video link

linxinan-yt: I am not sure since the procedure is for code, but the files we use here are binary media files. 

@borongc WDYT?

borongc: The file path was left by @linxinan-chops. I think it is good to leave it where it from, so I don't have any comment for this.

borongc: This is just a file (where it's from), not a patch from g3.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -57,6 +57,7 @@ namespace {
 const char kMimeTypeApplicationOctetStream[] = "application/octet-stream";
 const char kMimeTypeTextHtml[] = "text/html";
 const char kMimeTypeTextPlain[] = "text/plain";
+const char kMimeTypeImagePng[] = "image/png";
```
borongc: nit: move it to L58 for alphabet order.

borongc:
Maybe do a sanity check: use `enable_libvpx = false` in `//cobalt/build/configs/common.gn`, and use Linux build to verify if this fallback actually works.

cobalt/shell/BUILD.gn
tasunnn: after discussing with @linxinan-chops, we think it's better to leave the link as it is for now and have future work to store internal files for cobalt in some other way without directly exposing google code path. I've created a bug and will comment in file as TODO

linxinan-yt: We are preparing an internal folder for release build. I am thinking maybe eventually we could use that folder to store Youtube specific assets while keeping Cobalt an open source project like Chrome v.s. Chromium. We may continue the discussion in the tracking ticket b/481826565.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -230,6 +231,12 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
     // For html file, return from embedded resources.
     if (base::EndsWith(key, ".html", base::CompareCase::SENSITIVE)) {
       mime_type_ = kMimeTypeTextHtml;
+    } else if (base::EndsWith(key, ".png", base::CompareCase::SENSITIVE)) {
+      mime_type_ = kMimeTypeImagePng;
```
borongc, line 235: As `.png` and `.webm` are the similar cases, except `mime_type_`. Consider to use a helper to make the code more readable.

cobalt/shell/BUILD.gn
Awallky: I do not want to block submission. My stance is we should follow this policy http://shortn/_NlYfnRNo8N, including 

```
Data other than source code, including precompiled binaries and libraries, WSDL, XML, XSD, and anything else that you might not consider "code" falls under this restriction.
...when we refer to "code," assume we are also talking about whatever kind of data you were thinking about.
```

As mentioned above, there are already changes that list file paths and copy internal files, some literally in this same file. All of that being said, since it sounds like we filed a bug to address, I am okay with proceeding.

borongc:
This needs to CP to 26.android.

PR #8880, opened 2026-01-29
media: Fix CobaltContentRendererClient shutdown crash due to Mojo threading violation

borongc:
Previously, CobaltContentRendererClient could crash during shutdown.
This was caused by a Mojo threading violation when handling the
asynchronous response for GetSbWindow. The Mojo remote for
PlatformWindowProvider was a member variable, and its destruction
interfered with pending callbacks or was accessed from an incorrect
thread during shutdown.

To resolve this, make the PlatformWindowProvider remote local to the
RenderFrameCreated method. Ensure the GetSbWindow callback is posted
back to the client's thread using BindPostTaskToCurrentDefault, and
that the remote's lifetime is properly managed until the callback
executes. This prevents threading issues and ensures the client's
state is accessed safely.

Issue: 478302742

### DIFF
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -120,23 +120,36 @@ void CobaltContentRendererClient::RenderFrameCreated(
     LOG(WARNING) << "RenderFrameCreated is called with no webview.";
   }
 
-  if (!window_provider_.is_bound()) {
+  if (sb_window_handle_.load() == 0 && !window_handle_requested_) {
+    window_handle_requested_ = true;
+    mojo::Remote<media::mojom::PlatformWindowProvider> window_provider;
     render_frame->GetBrowserInterfaceBroker().GetInterface(
-        window_provider_.BindNewPipeAndPassReceiver());
+        window_provider.BindNewPipeAndPassReceiver());
 
-    window_provider_->GetSbWindow(base::BindOnce(
-        [](std::atomic<uint64_t>* out_handle, uint64_t handle) {
-          if (!handle) {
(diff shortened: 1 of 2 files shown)


cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -120,25 +120,38 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
+            },
+            weak_factory_.GetWeakPtr(), std::move(window_provider))));
   }
 }
 
+void CobaltContentRendererClient::OnGetSbWindow(uint64_t handle) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  if (!handle) {
+    LOG(ERROR) << "Renderer received invalid SbWindow handle.";
+    return;
```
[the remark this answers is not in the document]
borongc: `handle` would be valid if it is called.

y4vor (approved):
Thank you!

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -120,25 +120,38 @@ void CobaltContentRendererClient::RenderFrameCreated(
...
-            LOG(ERROR) << "Renderer received invalid SbWindow handle.";
-            return;
-          }
-
-          LOG(INFO) << "Renderer received SbWindow handle: "
-                    << reinterpret_cast<void*>(handle);
-          *out_handle = handle;
-        },
-        &sb_window_handle_));
+        window_provider.BindNewPipeAndPassReceiver());
```
Awallky, line 127: nit: Is this indentation off?

borongc: It's by `pre-commit` formatting.

PR #8888, opened 2026-01-30
Re-merge Enable Crashpad for hang detection and upload on Android TV + aab startup fix

alunev:
Enable Crashpad for hang detection and upload on Android TV
   This commit integrates Crashpad to detect and report thread hangs in Cobalt.
   When a hang is detected by `base::HangWatcher`, Crashpad now generates
   a minidump and uploads it to the Google Crash server.

   Key changes include:

   - Initializes Crashpad in `CobaltMainDelegate::PreSandboxStartup`.
   - Implements `CobaltCrashReporterClient` to provide Cobalt-specific
     information to Crashpad, including product/version from `COBALT_VERSION`.
   - Configures the crash dump location within the app's cache directory.
   - Packages `libchrome_crashpad_handler.so` in the Cobalt Android APK.
   - Introduces `base::HangWatcher::Delegate` to allow Cobalt to dynamically
     control hang reporting via the "enableCobaltHangReporting" feature flag
     (set via H5vccSettings).
   - Prevents Crashpad from handling regular crash signals (SIGSEGV, etc.)
     in the main process, reserving those for Primes.
   - Enables BoringSSL for HTTPS transport on Android within Crashpad.
   - Includes various build system adjustments
   - There's also a fix for ClassNotFoundException connected to crashpad when packaged as aab bundle: added explicit dependencies for //components/crash/android:java directly to the cobalt_apk target. This forces the Android App Bundle builder to include these required Java classes in the base APK.

### DIFF
--- a/cobalt/app/cobalt_crash_reporter_client.cc
+++ b/cobalt/app/cobalt_crash_reporter_client.cc
@@ -26,7 +26,11 @@
 #include "cobalt/version.h"
 
 namespace {
+#if BUILDFLAG(IS_ANDROIDTV)
 constexpr char kProductName[] = "Cobalt_ATV";
+#else
+constexpr char kProductName[] = "Cobalt";
+#endif
 constexpr char kCrashReportUrl[] = "https://clients2.google.com/cr/report";
 }  // namespace
 
--- a/cobalt/app/cobalt_main_delegate.cc
+++ b/cobalt/app/cobalt_main_delegate.cc
@@ -87,9 +87,8 @@ absl::optional<int> CobaltMainDelegate::PostEarlyInitialization(
   }
 
 #if BUILDFLAG(IS_ANDROIDTV)
-  // Set the HangWatcher delegate before initializing.
-  static cobalt::browser::CobaltHangWatcherDelegate g_hang_delegate;
-  base::HangWatcher::SetDelegate(&g_hang_delegate);
+  // This delegate is for reading the flag value.
+  cobalt::browser::CobaltHangWatcherDelegate::Initialize();
 #endif
 
   InitializeHangWatcher();
--- a/cobalt/browser/hang_watcher_delegate_impl.cc
+++ b/cobalt/browser/hang_watcher_delegate_impl.cc
@@ -17,11 +17,16 @@
 #include <cstdint>
 #include <optional>
 
-#include "base/logging.h"
 #include "cobalt/browser/global_features.h"
 
 namespace cobalt {
 namespace browser {
+
+// static
+void CobaltHangWatcherDelegate::Initialize() {
+  static base::NoDestructor<CobaltHangWatcherDelegate> instance;
+  base::HangWatcher::SetDelegate(instance.get());
+}
 
 bool CobaltHangWatcherDelegate::IsHangReportingEnabled() {
   std::optional<cobalt::GlobalFeatures::SettingValue> setting =
--- a/cobalt/browser/hang_watcher_delegate_impl.h
+++ b/cobalt/browser/hang_watcher_delegate_impl.h
@@ -25,6 +25,8 @@ class CobaltHangWatcherDelegate : public base::HangWatcher::Delegate {
   CobaltHangWatcherDelegate() = default;
   ~CobaltHangWatcherDelegate() override = default;
 
+  static void Initialize();
+
   bool IsHangReportingEnabled() override;
 };
 


cobalt/android/BUILD.gn
```diff
@@ -237,6 +237,8 @@ android_apk("cobalt_apk") {
     ":cobalt_apk_java",
     "//cobalt/shell/android:cobalt_shell_assets",
     "//cobalt/shell/android:cobalt_shell_java",
+    "//components/crash/android:handler_java",
+    "//components/crash/android:java",
```
hlwarriner, line 239: Just confirming that we need  `//components/crash/android:java` in particular, and not just `//components/crash/android:handler_java`? I ask because `//components/crash/android:java` has some sources that I believe we aren't using for hang reporting (e.g., `//components/minidump_uploader:minidump_uploader_java`).

hlwarriner (approved):
Left one comment but otherwise lgtm.

cobalt/android/BUILD.gn
alunev: hi, seems like it's the other way around: we need "//components/crash/android:java" but not "//components/crash/android:handler_java" since Cobalt is running native handler currently, not java one. Removed the redundant one.

base/threading/hang_watcher.cc
```diff
@@ -1042,8 +1058,9 @@ void HangWatcher::UnregisterThread() {
 
   // Thread should be registered to get unregistered.
   DCHECK(it != watch_states_.end());
-
-  watch_states_.erase(it);
+  if (it != watch_states_.end()) {
+    watch_states_.erase(it);
+  }
```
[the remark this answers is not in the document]
alunev: ok, let's fail fast with CHECK()

components/crash/core/app/crashpad_android.cc
```diff
@@ -410,8 +411,7 @@ void BuildHandlerArgs(CrashReporterClient* crash_reporter_client,
   crash_reporter_client->GetCrashDumpLocation(database_path);
   crash_reporter_client->GetCrashMetricsLocation(metrics_path);
 
-  // TODO(jperaza): Set URL for Android when Crashpad takes over report upload.
-  *url = std::string();
+  *url = kCrashReportUrl;
```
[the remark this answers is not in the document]
alunev: ok, let's do the override

cobalt/app/cobalt_crash_reporter_client.cc
```diff
@@ -0,0 +1,84 @@
...
+#include "base/android/path_utils.h"  // Needed for GetCacheDirectory
+#endif                                // BUILDFLAG(IS_ANDROID)
+#include "base/check.h"
+#include "base/files/file_path.h"
+#include "base/logging.h"
+#include "base/path_service.h"
+#include "cobalt/version.h"
+
+namespace {
+constexpr char kProductName[] = "Cobalt_ATV";
```
johnxwork, line 30: If you want this file to serve 3P later as well, you'd need to guard ATV-specific code with BUILDFLAG(IS_ANDROID), or you could make this file "atv_crash_reporter_client.cc". Either should be fine.

Or you could also change kProductName to be just "Cobalt"?

cobalt/app/cobalt_main_delegate.cc
```diff
@@ -78,6 +86,12 @@ absl::optional<int> CobaltMainDelegate::PostEarlyInitialization(
     content::InitializeMojoCore();
   }
 
+#if BUILDFLAG(IS_ANDROIDTV)
+  // Set the HangWatcher delegate before initializing.
+  static cobalt::browser::CobaltHangWatcherDelegate g_hang_delegate;
```
johnxwork: Init a static variable in the middle of a function looks a bit strange. I think the code is likely safe since `PostEarlyInitialization()` should only be called once right now.

But would you look into a paradigm that's more bullet proof? Like we would never call `SetDelegate` twice.

I'm also fine with this if you don't have a better one as it's really simple.

johnxwork (approved):

cobalt/app/cobalt_crash_reporter_client.cc
alunev: thanks, I've guarded(as in other places in this file)

For naming, I think the idea we've discussed is to differ ATV reports from others, and also to follow the convention: `Cobalt_Evergreen`.

cobalt/app/cobalt_main_delegate.cc
alunev: thanks' I've refactored a bit to look more like a standard singleton and to follow chromium conventions.

PR #8890, opened 2026-01-30
Correct content dir set by --evergreen_content

yuying-y:
Issue: 461621299

### DIFF
--- a/cobalt/build/modular_executable.gni
+++ b/cobalt/build/modular_executable.gni
@@ -114,8 +114,8 @@ template("evergreen_final_target") {
     args = [
       "--output",
       rebase_path(outputs[0]),
-      "--library_target",
-      test_lib_name,
+      "--library",
+      lib_name,
     ]
   }
 
--- a/starboard/build/util/generate_runfile.py
+++ b/starboard/build/util/generate_runfile.py
@@ -13,7 +13,7 @@ import sys
 
 command = [
     os.path.join(os.path.dirname(__file__), 'elf_loader_sandbox'),
-    '--evergreen_content=app/{library_target}/content', '--evergreen_library=lib{library_target}.so'
+    '--evergreen_content=.', '--evergreen_library={library}.so'
 ] + sys.argv[1:]
 try:
     result = subprocess.run(command, check=False)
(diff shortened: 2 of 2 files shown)


yuying-y:
> The library path you change this to doesn't exist—what is the content path that is missing from the out directory that's in the app//content subdirectory?

I see, the `lz4` lib exists but `so` doesn't, under `app/<target>/lib/`. I was following the reference 
```sh
.../elf_loader_sandbox --evergreen_library=app/nplb/lib/libnplb.so
                       --evergreen_content=app/nplb/content
```
in
[the Evergreen overview doc](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:starboard/doc/evergreen/cobalt_evergreen_overview.md;l=276-277). I'll update the docs accordingly.

For the content dir, the previous relevant issue is b/461621299. My understanding of the discussion was: the correct content path, if set by `--evergreen_content=`, should be `app/cobalt/content`, as it contains the correct cert dir `app/cobalt/content/ssl/certs`. I just ran
```
out/evergreen-x64_qa/elf_loader_sandbox --evergreen_content=app/cobalt/content --evergreen_library=libcobalt.so
```
It works!

andrewsavage1:
There should also be an `<out>/ssl/certs` directory not in the `app/` subdirectory—maybe there's a missing dependency?

yuying-y:
> There should also be an `<out>/ssl/certs` directory not in the `app/` subdirectory—maybe there's a missing dependency?

According to Niranjan and Holden, utilizing `<out>/ssl/certs` doesn't work. You can read the previous bug or talk to them about the details. My understanding is if a user wants to use `elf_loader_sandbox` to launch, they have to point to `app/` for the correct content dir. The best thing to do now is to update the documentations and make them consistent.

cobalt/build/modular_executable.gni
```diff
@@ -114,8 +114,8 @@ template("evergreen_final_target") {
     args = [
       "--output",
       rebase_path(outputs[0]),
-      "--library",
-      lib_name,
+      "--library_target",
+      test_lib_name,
```
andrewsavage1: You actually want original_target_name here (see how the path is constructed below)

andrewsavage1: Actually, you probably want to pass both lib_name and original_target_name and not try to reconstruct the names manually in generate_runfile.py

yuying-y: I figured it out. The previous approach failed because `test_lib_name` was not created correctly to represent the target name w/o the "lib" prefix, e.g. zlib_unittests->z_unittests. I modified the method to correctly identify names starting with "lib". Now "no_lib_target_name" represents the correctly stripped target name, and is the only param needed to passed in.

andrewsavage1: While this is right for the targets we currently run, it is wrong for targets that set output_prefix_override, as those would not have the `lib` prefix for the output shared library. We should pass both lib_name and original_target_name to generate_runfile.py.

yuying-y: I decided to skip the change in the script, because of the complication of how the path is manipulated and used by the tests, e.g. a test failed with: `The binary attempted to open the directory /home/pi/test/app/nplb/starboard/content/test/starboard/shared/starboard/player/testdata. However, the loader was configured with a different content path: /home/pi/test/app/nplb/content`. The script works fine with the current setting.

The purpose of the PR is to clarify the params with runing a target manually with `elf_loader_sandbox`. Updating the documentation is sufficient.

andrewsavage1 (approved):

PR #8905, opened 2026-01-31
cobalt: Introduce CobaltThreadChecker

yell0wd0g:
Introduce `CobaltThreadChecker`, replacing `base::ThreadChecker`
macros with a Cobalt-specific, always-enabled equivalents.

Chromium's `base::ThreadChecker` provides thread validation only when
`DCHECK_IS_ON()`. Cobalt requires more robust thread safety checks
across all build configurations, including release builds, to prevent
subtle runtime threading issues.

This CL also clarifies `CobaltContentRendererClient` by using 
`CHECK(content::RenderThread::IsMainThread())` for methods called on
the Renderer thread, and using the new `CobaltThreadChecker` for the 
constructor and destructor which run on the browser main thread.

Performance-wise this shouldn't introduce a big penalty since the
classes touched here are more scaffolding of an embedder, rather than
actions run per e.g. navigation, net etc.

Bug: 481078653

### DIFF
--- a/cobalt/app/tvos/main.mm
+++ b/cobalt/app/tvos/main.mm
@@ -27,6 +27,8 @@
 #include "cobalt/shell/browser/shell.h"
 #include "content/public/app/content_main.h"
 #include "content/public/app/content_main_runner.h"
+#include "content/public/browser/render_view_host.h"
+#include "content/public/browser/render_widget_host.h"
 #include "starboard/common/command_line.h"
 #include "starboard/tvos/shared/application_darwin.h"
 
@@ -37,6 +39,11 @@ static const char** g_argv = nullptr;
 @end
 
 @implementation CobaltAppSceneDelegate
+
+- (content::WebContents*)getWebContents {
+  CHECK_EQ(1u, content::Shell::windows().size());
+  return content::Shell::windows()[0]->web_contents();
+}
 
 - (void)scene:(UIScene*)scene
     willConnectToSession:(UISceneSession*)session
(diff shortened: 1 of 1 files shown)


yell0wd0g:
@andrewsavage1 PTAL

cobalt/common/cobalt_thread_checker.h
```diff
@@ -0,0 +1,44 @@
+// Copyright 2026 The Cobalt Authors
```
andrewsavage1: This looks quite different from our usual copyright header, are you sure it's right?

andrewsavage1:
The reference bug seems to be incorrect

Also, is this preferable to enabling base's thread check all the time? Why?

yell0wd0g:
> The reference bug seems to be incorrect

Ah yes, the bug is made up. I'll make one.

> 
> Also, is this preferable to enabling base's thread check all the time? Why?

Well, there is a cost to checking the threading on every call
everywhere. And, Chromium CI has browser tests running with
debug builds, so there's that coverage we get "for free". This
is specific to the //cobalt folder (I hope it'll be used more
pervasively to clarify threading).

andrewsavage1:
> > Also, is this preferable to enabling base's thread check all the time? Why?
> 
> Well, there is a cost to checking the threading on every call everywhere. And, Chromium CI has browser tests running with debug builds, so there's that coverage we get "for free". This is specific to the //cobalt folder (I hope it'll be used more pervasively to clarify threading).

Got it, so the cost isn't that high for the ones we're explicitly adding but would be high if we enabled all the existing ones—makes sense.

cobalt/common/cobalt_thread_checker.h
yell0wd0g: Whoops this is a Chromium header style, not a Cobalt one. Wrong copy-paste.
Let me fix it.

yell0wd0g:
> > > Also, is this preferable to enabling base's thread check all the time? Why?
> > 
> > 
> > Well, there is a cost to checking the threading on every call everywhere. And, Chromium CI has browser tests running with debug builds, so there's that coverage we get "for free". This is specific to the //cobalt folder (I hope it'll be used more pervasively to clarify threading).
> 
> Got it, so the cost isn't that high for the ones we're explicitly adding but would be high if we enabled all the existing ones—makes sense.

Added a line to this effect in `CobaltThreadChecker`.

PR #8916, opened 2026-02-02
media: Refactor MediaCodecBridge::release 

kjyoun:
Consolidates MediaCodec stopping and releasing into a single public
release() method. This ensures that the Android 11 workaround (skipping
stop() to avoid a framework bug during destruction) is only applied
when the codec is actually being released.

Issue: 480983077


kjyoun:
> this looks like a nice simplification and consolidation. The gemini comment regarding use-after-free sounds valid though.

Oops, I missed Gemin's critical comment.  will address it

kjyoun:
> I am having trouble creating individual review comments. Click here to see my feedback.
> ### starboard/android/shared/media_decoder.cc (193-194)
> security-high high
> 
> The removal of the explicit call to `Stop()` (which should now be replaced by an explicit release or destruction of the bridge) introduces a potential race condition and Use-After-Free (UAF) vulnerability. This call was responsible for setting `mNativeMediaCodecBridge = 0` in Java, which stops callbacks from the Android `MediaCodec` before the `MediaCodecDecoder` and its members are destroyed. Without this early signal, callbacks might be forwarded to the native side while the decoder is in a partially destroyed state, or even after it has been fully destroyed if the bridge is held by a shared pointer. It is recommended to explicitly destroy the bridge at the beginning of the destructor (e.g., by calling `media_codec_bridge_.reset()` or setting it to `nullptr`) to ensure callbacks are halted while all members are still alive.

Since `media_codec_bridge_` is the last member declared in `MediaCodecDecoder`, it is guaranteed to be destroyed first (among all members) immediately after the destructor body completes. Consequently, there is no functional difference between calling `media_codec_bridge_.reset()` at the bottom of the destructor and letting the compiler handle it automatically.

Moving the bridge destruction to the **beginning** of the destructor would change the existing call order (e.g., halting callbacks before the decoder thread is joined and the codec is flushed). I would prefer not to change this established behavior in the current PR.
 
Furthermore, potential races are already mitigated by the `destroying_` atomic flag set at the start of the destructor and the Java-side lock (`mNativeBridgeLock`) used during release, which ensures any active callbacks finish before the native pointer is cleared."

PR #8918, opened 2026-02-02
browser: Tie Mojo impl to RFH w/ DocumentUserData

hlwarriner:
This avoids a use-after-free vulnerability from a RenderFrameHost pointer held by H5vccPlatformServiceManagerImpl, as pointed out by Gemini in a review of #8785.

content::DocumentUserData seems preferable to content::DocumentService since there only needs to be a single ServiceManagerImpl per RenderFrameHost (see content/public/browser/document_service.h).

Issue: 478934144

### DIFF
--- a/cobalt/browser/cobalt_browser_interface_binders.cc
+++ b/cobalt/browser/cobalt_browser_interface_binders.cc
@@ -102,9 +102,8 @@ void PopulateCobaltFrameBinders(
 // TODO: b/403638702 - add a binding for a Java Mojo impl for 1P ATV.
 #if !BUILDFLAG(IS_ANDROIDTV)
   binder_map->Add<h5vcc_platform_service::mojom::H5vccPlatformServiceManager>(
-      base::BindRepeating(
-          &h5vcc_platform_service::
-              H5vccPlatformServiceManagerImpl::GetOrCreate));
+      base::BindRepeating(&h5vcc_platform_service::
+                              H5vccPlatformServiceManagerImpl::GetOrCreate));
 #endif
 }
 


cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
```diff
@@ -25,37 +26,48 @@
...
-    : render_frame_host_(render_frame_host),
-      receiver_(this, std::move(receiver)) {
-  receiver_.set_disconnect_handler(base::BindOnce(
-      &H5vccPlatformServiceManagerImpl::OnDisconnect, base::Unretained(this)));
+    content::RenderFrameHost* render_frame_host)
+    : content::DocumentUserData<H5vccPlatformServiceManagerImpl>(
+          render_frame_host) {
+  LOG(INFO) << "H5vccPlatformServiceManagerImpl created for RFH ID: "
+            << this->render_frame_host().GetGlobalId();
+  receivers_.set_disconnect_handler(base::BindRepeating(
```
andrewsavage1, line 65: Is there a point to having a disconnect handler if it seems not to do anything?

hlwarriner: It's just used to add debug logging, since disconnection is a potentially interesting event. I added a comment to clarify.

cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "cobalt/browser/h5vcc_platform_service/h5vcc_platform_service_manager_impl.h"
 
+#include <string>
```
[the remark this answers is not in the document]
hlwarriner: Google's C++ style guide recommends the exact opposite: "foo.cc should include bar.h if it uses a symbol from it even if foo.h includes bar.h."

hlwarriner:
CC @zhongqiliang

PR #8920, opened 2026-02-02
cobalt/metrics: Add Memory.*.Resident metrics to Cobalt

Awallky:
Note: This is based on [PR#8845](https://github.com/youtube/cobalt/pull/8845).

Note: Not all platforms support these metrics. For single-process mode, GPU or renderer process may not show up until cobalt moves to a multi-process architecture.

This change adds collection and recording of Resident Set Size (RSS) metrics 
for the browser, renderer, and GPU processes, as well as a total aggregate.
These are recorded as UMA histograms:
  - Memory.Total.Resident
  - Memory.Browser.Resident
  - Memory.Renderer.Resident
  - Memory.Gpu.Resident

Also updated cobalt_unittests, cobalt_browsertests, and
pull_uma_histogram_set_via_cdp.py to include these new metrics.

Test: out/android-x86_devel/bin/run_cobalt_unittests --gtest_filter="*CobaltMetricsServiceClientTest*"
Test: out/android-x86_devel/bin/run_cobalt_browsertests --gtest_filter="*CobaltMetricsBrowserTest*"

Bug: 480146289


cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
```diff
@@ -306,20 +306,44 @@ TEST_F(CobaltMetricsServiceClientTest, RecordMemoryMetricsRecordsHistogram) {
...
 
   CobaltMetricsServiceClient::RecordMemoryMetrics(global_dump.get());
 
-  histogram_tester.ExpectUniqueSample("Memory.Total.PrivateMemoryFootprint", 10,
+  histogram_tester.ExpectUniqueSample("Memory.Total.PrivateMemoryFootprint", 17,
                                       1);
+  histogram_tester.ExpectUniqueSample("Memory.Total.Resident", 34, 1);
+  histogram_tester.ExpectUniqueSample("Memory.Browser.Resident", 20, 1);
+  histogram_tester.ExpectUniqueSample("Memory.Renderer.Resident", 10, 1);
+  histogram_tester.ExpectUniqueSample("Memory.Gpu.Resident", 4, 1);
```
sherryzy: Can we confirm if Total Resident RSS is the sum of the Browser, Renderer, and GPU processes? Based on the test cases, that seems to be the current breakdown. Since we only have access to Total and Browser data right now, would it be accurate to use the delta (Total - Browser) as an estimate for graphics memory usage?

cobalt/browser/metrics/cobalt_metrics_browsertest.cc
```diff
@@ -68,10 +67,14 @@ IN_PROC_BROWSER_TEST_F(CobaltMetricsBrowserTest,
       histogram_tester.GetAllSamples("Memory.Total.PrivateMemoryFootprint")
           .size(),
       1u);
+  EXPECT_GE(histogram_tester.GetAllSamples("Memory.Total.Resident").size(), 1u);
+  EXPECT_GE(histogram_tester.GetAllSamples("Memory.Browser.Resident").size(),
+            1u);
+  // Renderer and Gpu processes may not be present depending on the platform and
```
linxinan-yt: Since cobalt is single process architecture in the foreseeable future, what about check if GPU and render are not sampled?

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
```diff
@@ -200,13 +200,22 @@ def _get_histograms_from_file(file_path: str) -> list:
...
+      'Memory.Total.PrivateMemoryFootprint',
+      'Memory.Browser.MemoryFootprint',
+  ]
+
+
+def _get_cobalt_resident_memory_metrics() -> list:
+  return [
+      'Memory.Total.Resident',
+      'Memory.Browser.Resident',
+      'Memory.Renderer.Resident',
```
linxinan-yt: Not blocker but curious if we are not using them for single process cobalt, why we have to report them now?

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -377,15 +377,53 @@ void CobaltMetricsServiceClient::ScheduleRecordForTesting(
...
+  uint64_t gpu_resident_kb = 0;
+
   for (const auto& process_dump : global_dump->process_dumps()) {
     total_private_footprint_kb += process_dump.os_dump().private_footprint_kb;
+
+    uint32_t resident_kb = process_dump.os_dump().resident_set_kb;
+    total_resident_kb += resident_kb;
+
+    switch (process_dump.process_type()) {
+      case memory_instrumentation::mojom::ProcessType::BROWSER:
```
sherryzy: Since Cobalt is single-process, do we always only have memory data for browser type which equals to the total RSS?
To avoid reporting misleading '0' values for sub-processes, we should consolidate the view to show only Total RSS and omit the specific process breakdowns for now. WDYT?

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
```diff
@@ -200,13 +200,22 @@ def _get_histograms_from_file(file_path: str) -> list:
...
-  metrics.append('Memory.Browser.PrivateMemoryFootprint')
-  metrics.append('Memory.Gpu.PrivateMemoryFootprint')
-  metrics.append('Memory.Renderer.PrivateMemoryFootprint')
-  metrics.append('Memory.Total.PrivateMemoryFootprint')
-  metrics.append('Memory.Browser.MemoryFootprint')
-  return metrics
+  return [
+      'Memory.Browser.PrivateMemoryFootprint',
+      'Memory.Gpu.PrivateMemoryFootprint',
+      'Memory.Renderer.PrivateMemoryFootprint',
```
sherryzy: For the privateMemoryFootprint, do we have data for GPU/renderer thread?

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: This is just the default set that runs when no one provides an input file. This is easily overridden. Users can  override them with other metrics by providing an input file.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: No, I think we would need multi-process mode for them to fire.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
Awallky: Yes, that is the case. 
I suppose we could omit the process-specific metrics and leave a TODO to re-add them if Cobalt ever runs in multi-process mode and reduce our code complexity and what is essentially dead code.

Note: As the PR currently stands, histograms only fire if they are > 0 MB.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: On second thought, since they do not fire in our current process architecture, maybe we remove these to reduce code complexity.

cobalt/browser/metrics/cobalt_metrics_browsertest.cc
Awallky: I removed them since they will not fire on single process architecture.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
Awallky: Removed.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: Removed.

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
Awallky: Removed.

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
Awallky: With the resident metrics, the total memory is whatever is seen in `/proc/PID/smaps`. One limitation we noticed when using smaps to check memory regressions was that smaps does *not* track Android GPU memory consumption. It only tracks the amount of shared memory buffers between the GPU process, and browser and renderer processes.  It is possible that when the multi-process mode is active, that the GPU process does something to track GPU memory consumption, but I am not sure.

I have removed the process-specific metrics now.

sherryzy: Where exactly does smaps fail to track Android GPU memory? Since Android uses a unified memory architecture, I expect GPU allocations to be reflected in the process's total RSS. Does smaps report these mappings differently and does the total RSS in smaps typically align with the 'Total' reported by adb shell dumpsys meminfo?

Awallky: You can see this detail in Richard's guide to memory, but yeah Graphics memory, is not completely visible to the kernel. Gfx Dev is the mmapped file visible to smaps. However, that does not compose all of graphics memory. There is also EGL memtrack and GL memtrack. 
See this [link](http://shortn/_XcPvYTOHpH) for more details.

Awallky: We saw this when investigating memory leaks last year, where smaps showed ~100MB less RSS than adb dumpsys meminfo.

PR #8952, opened 2026-02-04
Fix gclient config cache-dir

fayev-g:
Bug: 481463722

Should use the depot tools base_path set at the top of the script. The base_path is different when using the cache or not.


oxve:
You can ignore Gemini's review comment. I have a change in flight that will make it obsolete.

PR #8954, opened 2026-02-04
android: Add BRP configurations

jasonzhangxx:
This change introduces Java-side switches to configure the
PartitionAllocBackupRefPtr (BRP) memory safety feature. These switches
allow Cobalt Android applications to explicitly enable, disable, or
configure BRP with a memory reclaimer via Java properties. This provides
greater control for memory management and security on Android platforms.

Bug: 481385937

### DIFF
--- a/cobalt/android/apk/app/src/test/java/dev/cobalt/util/JavaSwitchesTest.java
+++ b/cobalt/android/apk/app/src/test/java/dev/cobalt/util/JavaSwitchesTest.java
@@ -54,6 +54,8 @@ public class JavaSwitchesTest {
     javaSwitches.put(JavaSwitches.DISABLE_SPLASH_SCREEN, "true");
     javaSwitches.put(JavaSwitches.FORCE_IMAGE_SPLASH_SCREEN, "true");
     javaSwitches.put(JavaSwitches.NUM_RASTER_THREADS, "4");
+    javaSwitches.put(JavaSwitches.DISABLE_BRP, "true");
+    javaSwitches.put(JavaSwitches.ENABLE_BRP_RECLAIMER, "true");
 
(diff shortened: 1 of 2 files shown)


jasonzhangxx:
> Could we add a test case to cobalt/android/apk/app/src/test/java/dev/cobalt/util/JavaSwitchesTest.java for this?
> 
> Additionally, I’d like to clarify the usage of these two flags: What is their current default state in Cobalt, and what is the proposed plan for experimentation?

Tests added. The default brp state is "enabled". We'll need experiment for both "disabled" and "enabled-with-memory-reclaimer" separately. "disabled" is preferred.

cobalt/android/apk/app/src/test/java/dev/cobalt/util/JavaSwitchesTest.java
```diff
@@ -63,6 +64,7 @@ public void getExtraCommandLineArgs_AllSwitches() {
     assertThat(args).contains("--js-flags=--initial-old-space-size=128");
     assertThat(args).contains("--js-flags=--max-old-space-size=256");
     assertThat(args).contains("--js-flags=--max-semi-space-size=16");
+    assertThat(args).contains("--disable-features=PartitionAllocBackupRefPtr");
     assertThat(args).hasSize(6);
```
sherryzy: Have you tried to run the test? It should fail on this check as it should be size 7. Note: this test currently is not running on CI, please manually build the test and verify it passes.

jasonzhangxx: Now the test can pass.

sherryzy (approved):

PR #8959, opened 2026-02-04
fix: add fuzztest deps/include config for base_unittests

hariprasad-hp:
**Problem**

base_unittests fails to compile when building json_reader_unittest.cc because FuzzTest headers use relative includes like:

   #include "./fuzztest/domain.h"

The build command for base_unittests did not include the FuzzTest source root (third_party/fuzztest/src) in its header search paths, so the compiler could not resolve ./fuzztest/... includes.

**Root Cause**

test("base_unittests") did not depend on //third_party/fuzztest:fuzztest, so GN did not propagate FuzzTest’s public include dirs into the unit test target.

**Fix**

Add //third_party/fuzztest:fuzztest to test("base_unittests") deps so FuzzTest include paths are applied when compiling base unit tests.

Export :fuzztest_public_config from fuzztest_internal to ensure dependents of fuzztest_internal also receive the required include dirs.

**Result**

Build proceeds past the error:

    fatal error: './fuzztest/domain.h' file not found

Files changed

    base/BUILD.gn

    third_party/fuzztest/BUILD.gn

https://buganizer.corp.google.com/issues/460879170#comment4

### DIFF
--- a/base/BUILD.gn
+++ b/base/BUILD.gn
@@ -3885,9 +3885,9 @@ test("base_unittests") {
     "//testing/gmock",
     "//testing/gtest",
     "//third_party/abseil-cpp:absl",
+    "//third_party/fuzztest:fuzztest",
     "//third_party/icu",
     "//third_party/modp_b64",
-    "//third_party/fuzztest:fuzztest",
 
     # TODO(crbug.com/40158212): Move partition_alloc tests into its repository.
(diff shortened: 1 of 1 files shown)


andrewsavage1:
We should address this with a different approach, see https://github.com/youtube/cobalt/pull/8967

PR #8963, opened 2026-02-04
Add a fixed adb binary to our browser tests

SortaCookie:
This was tested and verified by manually building and deploying to the MH lab.

Bug: 483488213

### DIFF
--- a/cobalt/testing/browser_tests/tools/Dockerfile
+++ b/cobalt/testing/browser_tests/tools/Dockerfile
@@ -4,7 +4,7 @@ FROM ubuntu@sha256:c7eb020043d8fc2ae0793fb35a37bff1cf33f156d4d4b12ccc7f3ef8706c3
 # Install base utilities, Android test dependencies (adb, netcat),
 # and Linux test dependencies (GUI/X11 libraries).
 RUN apt-get update && apt-get install -y \
-    adb \
+    unzip \
     netcat-openbsd \
     python3 \
     ca-certificates \
@@ -22,6 +22,7 @@ RUN apt-get update && apt-get install -y \
     libxrandr2 \
     libgbm1 \
     socat \
+    iproute2 \
     usbutils \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


cobalt/testing/browser_tests/tools/Dockerfile
```diff
@@ -22,9 +23,17 @@ RUN apt-get update && apt-get install -y \
     libxrandr2 \
     libgbm1 \
     socat \
+    iproute2 \
     usbutils \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
 
+# Download and install the Android SDK Platform-Tools (which includes adb).
+RUN wget https://dl.google.com/android/repository/platform-tools_r33.0.1-linux.zip -O /tmp/platform-tools.zip && \
```
Awallky: Instead of needing to download this everytime, what if we just commit this version to cobalt/build/tools or sommething like that along with a README explaining its necessity? Then we can just pull it into the tarball using the collect_test_artifacts script and keep the PATH update here in the Dockerfile.

cobalt/testing/browser_tests/tools/collect_test_artifacts.py
```diff
@@ -120,9 +120,28 @@ def main():
...
+      help='Output filename (default: cobalt_browsertests_artifacts.tar.gz)')
+  parser.add_argument(
+      '--output_dir', help='Output directory where the tarball will be placed.')
+  parser.add_argument(
+      '--compression',
+      choices=['xz', 'gz', 'zstd'],
+      default='gz',
+      help='The compression algorithm to use.')
+  parser.add_argument(
+      '--adb_platform_tools',
```
Awallky, line 132: No AI here.
I would like a test for the use of this flag, but I think it is a bit challenging to simulate a zip file in a test and I am not sure how much this code path could cause an issue.

Awallky (approved):
LGTM.

Please wait to land this until the dependent change is approved.

Awallky:
Also, please add appropriate commit message.
Could you also share how you tested and verified this change?

PR #8971, opened 2026-02-04
Add cherry-pick step to sync chromium branches.

briantting:
When the sync chromium branches action applies a new refresh diff commit to m138 that commit is now also cherry picked to main.

Bug: 481788398

### DIFF
--- a/.github/workflows/sync_chromium_branches.yaml
+++ b/.github/workflows/sync_chromium_branches.yaml
@@ -22,50 +22,66 @@ jobs:
           {milestone: m138, branch_num: 7204},
           {milestone: m144, branch_num: 7559},
         ]
-    outputs:
-      output1: ${{ steps.diff_step.outputs.diff_present }}
     steps:
-      # Attempt to checkout our chromium reference, if missing we will create a new branch
+      # Attempt to checkout our Chromium reference, if missing, create a new branch
       - uses: actions/checkout@v4
         id: checkout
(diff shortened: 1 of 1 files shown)


briantting:
> I think with something like this. I would like to see a run on the sandbox to make sure we don't break something.

Merged it just now with cobalt_sandbox, we can wait until the next rebase update to see what happens.

edit: going to force revert the sandbox's chromium/m138 latest commit so we can see the sync happen again

briantting:
https://github.com/youtube/cobalt_sandbox/actions/runs/21734334233/job/62695940965

https://github.com/youtube/cobalt_sandbox/commit/05ac1177880f3ae9807eacc7d760bb136f1653b1

works in sandbox!

kaidokert:
First review attempt didnt trigger for some reason

.github/workflows/sync_chromium_branches.yaml
```diff
@@ -23,49 +23,68 @@ jobs:
           {milestone: m144, branch_num: 7559},
         ]
     outputs:
-      output1: ${{ steps.diff_step.outputs.diff_present }}
+      output1: ${{ steps.create_diff.outputs.diff_present }}
```
SortaCookie: This output1 probably should be renamed.

briantting: removing outputs completely

PR #8975, opened 2026-02-05
tvOS: Remove unused //starboard/tvos/shared/get_home_directory.cc

Gyuyoung:
This file is used by Linux and RDK systems. Thus, the tvOS port
does not need to keep this file.
    
Bug: 456692793


Gyuyoung:

kjyoun:
> The change itself looks fine, I just wonder if this file is actually needed at all given it hasn't broken the build so far.

+1

If we don't actually use this file, why not remove it? Then we won't need to keep maintaining it like this any more.

Gyuyoung:
> > The change itself looks fine, I just wonder if this file is actually needed at all given it hasn't broken the build so far.
> 
> +1
> 
> If we don't actually use this file, why not remove it? Then we won't need to keep maintaining it like this any more.

yes, it looks like the file has only used by linux and rdk. Let's remove the file in tvOS.

rakuco (approved):
lgtm!

PR #8981, opened 2026-02-05
Cherry pick PR #8876: Add static image fallback for native splash screen

tasunnn:
original PR: #8876

Implement static image fallback for the native splash screen for devices that don't support VP9 video playback. This ensures a visual splash screen is always presented even on low spec devices. Similar to splash video, the image can also be cached locally after first start, which could then be loaded by h5vcc loader for display in cases of cache hits. 

Ran linux browsertest and changes passed locally. 

Bug: 458074360


linxinan-yt:
Please update your commit message. I think Linux br tests are not working on 26.android right now. You may simply state that you have tested the PR locally etc.

linxinan-chops (approved):
LGTM

PR #8990, opened 2026-02-06
android: Enable music splash screen

borongc:
Introduce a mechanism to display a specialized music splash screen
during application startup. This change adds music-themed splash
assets and modifies the Android activity and Cobalt shell logic
to select these assets based on a "topic" deep link parameter.

This allows for a tailored splash screen experience, improving
brand consistency and user experience for the YouTube Music
application during the loading phase.

Issue: 482144273


cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
```diff
@@ -142,12 +141,14 @@
         if (!params.has('cache')) {
           params.append('cache', 'default');
         }
+        const topicName = params.get('cache');
+        const splashTopic = (topicName === 'music') ? 'music_' : '';
```
linxinan-yt, line 147: This seems only serve music.

What about defining a set?

```
const supportTopics = new Set(['music']);
const splashTopic = supportTopics.has(topicName) ? topicName + '_' : '';
```

And later if we need to support more topic, we can simply append them to the support_topic set.

borongc: Right, this is only for music. For other types of splash screen, it is the default one.

That is, music splash screen is only the case at cold start with music deeplink.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -211,13 +211,21 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
-    // not support VP9.
+    // For html file, return from embedded resources.
+    if (base::EndsWith(key, ".html", base::CompareCase::SENSITIVE)) {
+      mime_type_ = kMimeTypeTextHtml;
+    } else if (base::EndsWith(key, ".png", base::CompareCase::SENSITIVE)) {
+      mime_type_ = kMimeTypeImagePng;
+    } else if (base::EndsWith(key, ".webm", base::CompareCase::SENSITIVE)) {
+      mime_type_ = kMimeTypeVideoWebM;
+    }
+    bool is_cacheable_type =
```
linxinan-yt, line 222: This seems not from this PR. Please rebase.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
linxinan-yt: If that is the case, e.g. only music, I would suggest leaving a comment here to clarify the context.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
borongc: It's already rebased. I moved the code due to reusing the flag `is_cacheable_type` in L228.

cobalt/shell/embedded_resources/loader_embedded_resources/splash.html
borongc: Done.

cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
```diff
@@ -108,19 +108,20 @@ public WindowAndroid getWindow() {
      */
     public void launchShell(String url) {
         // Calls the overloaded method with a null listener.
-        launchShell(url, null);
+        launchShell(url, "", null);
```
Awallky: Can we add parameter comments to the non-obvious ones?
https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/java/java.md#parameter-comments

borongc: Done.

borongc:
> I do not want to block this from landing before the release cutoff. Can we follow-up this change with some addition to cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc?

Thanks. Just want to clarify that adding music splash screen doesn't create new cases for h5vcc_scheme_url_loader_factory_browsertest.cc.

PR #8991, opened 2026-02-06
evergreen: Use flat_map for the Omaha ID map

yuying-y:
The kChannelAndSbVersionToOmahaIdMap is a read-only data structure.
Switching from std::unordered_map to base::fixed_flat_map optimizes
performance by reducing lookup times and improves memory usage due
to its compact, fixed-size nature. It also uses StringPiece to avoid
unnecessary string copies.

Issue: 474154953

### DIFF
--- a/cobalt/updater/util.h
+++ b/cobalt/updater/util.h
@@ -32,7 +32,7 @@ namespace cobalt {
 namespace updater {
 
 // Map of Omaha config IDs with channel and starboard version as indices.
-inline constexpr size_t kOmahaIdMapSize = 14;
+inline constexpr size_t kOmahaIdMapSize = 15;
 extern const base::
     fixed_flat_map<base::StringPiece, base::StringPiece, kOmahaIdMapSize>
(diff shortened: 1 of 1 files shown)


cobalt/updater/util.cc
```diff
@@ -50,8 +52,8 @@ const char kUncompressedLibraryPath[] = "lib/libcobalt.so";
 
 }  // namespace
 
-const std::unordered_map<std::string, std::string>
-    kChannelAndSbVersionToOmahaIdMap = {
+const auto kChannelAndSbVersionToOmahaIdMap = base::
+    MakeFixedFlatMap<base::StringPiece, base::StringPiece, kOmahaIdMapSize>({
```
hlwarriner, line 56: I think you just need to remove the size template parameter, `kOmahaIdMapSize`.

yuying-y: Turns out it's because kOmahaIdMapSize doesn't match the map size.

PR #8992, opened 2026-02-06
ci: Add YTS playback tests to presubmit

Libzu:
Integrate YTS playback tests into the GitHub Actions CI workflow.
This enables automatic execution of dedicated YTS playback tests
when the "yts_playback_test" label is applied to a pull request.
This ensures critical playback functionality is validated
before merging changes.

Bug: 481804189

### DIFF
--- a/.github/config/android-arm.json
+++ b/.github/config/android-arm.json
@@ -58,6 +58,12 @@
       "test_attempts": "6"
     }
   ],
+  "yts_playback_test_targets": [
+    {
+      "target": "testing/maneki/cobalt/kimono_tests:ATV_KIMONO_YTS_PLAYBACK_TESTS_ARM_DEVICES",
+      "test_attempts": "3"
+    }
+  ],
   "test_dimensions": {
     "device_type": "arm_devices",
     "device_pool": "maneki"
--- a/.github/config/android-arm.json
+++ b/.github/config/android-arm.json
@@ -8,6 +8,7 @@
   "test_device_family": "android",
   "test_e2e": true,
   "test_yts": true,
+  "test_yts_playback": true,
   "test_root_target": "//cobalt/android:cobalt_apk",
   "test_attempts": "3",
   "e2e_test_targets": [
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -538,9 +538,11 @@ jobs:
         config: [devel]
     steps:
       - name: Restore CI Essentials
-        uses: actions/download-artifact@v4
+        uses: actions/checkout@v4
         with:
-          name: ci-essentials-${{ inputs.platform }}
+          fetch-depth: 1
+          filter: blob:none
+          sparse-checkout: ${{ env.CI_ESSENTIALS }}
       - name: Run YTS Tests
         uses: ./.github/actions/internal_tests
         with:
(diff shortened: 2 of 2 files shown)


.github/config/android-arm.json
```diff
@@ -58,6 +58,12 @@
       "test_attempts": "6"
     }
   ],
+  "yts_playback_test_targets": [
+    {
+      "target": "testing/maneki/cobalt/kimono_tests:ATV_KIMONO_YTS_PLAYBACK_TESTS_ARM_DEVICES",
```
isarkis, line 64: Does this target already exist, I couldn't find it?

DhruvRaipure24 (approved):
lgtm

.github/config/android-arm.json
DhruvRaipure24: Yes, submitted a CL that creates this target.

.github/config/android-arm.json
isarkis, line 1: Add  `"test_yts_playback": true`

Libzu: done

Awallky (changes requested):
Can you rebase onto origin/main? I don't think all of these commits are intended for this change. It will make the diffs, and the required number of reviewers, likely smaller.

.github/workflows/main.yaml
```diff
@@ -502,6 +518,35 @@ jobs:
           test_type: 'yts_test'
           test_targets: '${{ needs.initialize.outputs.yts_test_targets }}'
 
+  yts-playback-test:
+    needs: [initialize, build]
+    # Run YTS Playback tests via presumit.
+    # Also, run YTS tests on push and schedule if not explicitly disabled via repo vars.
```
oxve: > if not explicitly disabled via repo vars

You are not checking the vars. They are used to disable ODT tests on sandbox so I think it should be added.

.github/workflows/main.yaml
```diff
@@ -502,6 +518,35 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        platform: ${{ fromJson(needs.initialize.outputs.platforms) }}
+        include: ${{ fromJson(needs.initialize.outputs.includes) }}
+        config: [devel]
+    steps:
+      - name: Restore CI Essentials
+        uses: actions/download-artifact@v4
+        with:
+          name: ci-essentials-${{ inputs.platform }}
```
oxve: This section is outdated. Check the other job for what should go here.

.github/workflows/main.yaml
```diff
@@ -502,6 +518,35 @@ jobs:
           test_type: 'yts_test'
           test_targets: '${{ needs.initialize.outputs.yts_test_targets }}'
 
+  yts-playback-test:
```
oxve, line 521: Why is this a new job and not just added to the list of yts_test_targets above?

The code looks like an exact copy except for the target list.

.github/workflows/main.yaml
Libzu: fixed, thanks

.github/workflows/main.yaml
Libzu: because the yts_playback tests are non blocking

.github/workflows/main.yaml
Libzu: this check is triggered by label, I am updating the comment

oxve (approved):

PR #8998, opened 2026-02-06
android: Refactor StartupGuard disable logic

zhongqiliang:
Refactor the handling of the StartupGuard's disabling logic within
CobaltActivity. The conditional check for DISABLE_STARTUP_GUARD is
moved from the onStart method to the onCreate method.

This change prevents the StartupGuard from being unnecessarily
scheduled in onCreate only to be disarmed immediately after in onStart
when it is meant to be disabled via Java switches. It streamlines the
initialization flow and ensures the StartupGuard is only scheduled if
it is intended to be active, improving clarity and efficiency.

Bug: 467729316


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -378,7 +378,11 @@ protected void onCreate(Bundle savedInstanceState) {
 
     super.onCreate(savedInstanceState);
 
-    StartupGuard.getInstance().scheduleCrash(HANG_APP_CRASH_TIMEOUT_SECONDS);
+    if (getJavaSwitches().containsKey(JavaSwitches.DISABLE_STARTUP_GUARD)) {
```
kaidokert, line 381: Should this and super.onCreate  be before setVolumeControlStream ?

zhongqiliang: I think it should be. But this is not for StartupGuard. I will find who wrote the code, and then confirm if we can move, if yes, in a follow up PR.

PR #8999, opened 2026-02-06
Enable sccache for TVOS targets and refactor workflow to use main.yaml.

SortaCookie:
This should turn on sccache for TVOS targets to improve compile times. Additionally, refactoring the tvos.yaml file to use main.yaml which is inline with the other workflows we have.

Bug: 432823450


SortaCookie:
The TvOS failure because the secret is not passed in from forks.

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,59 @@ inputs:
   test_targets_json_file:
     description: "The path to the file the test targets are written to."
     required: true
+  sccache_gcs_key_mac:
```
oxve: `sccache_gcs_key_tvos` for consistency. Or just `sccache_gcs_key`. We're almost certainly not going to be using sccache for any other platforms.

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,59 @@ inputs:
...
+    required: false
 runs:
   using: "composite"
   steps:
       - name: Clean out dir from previous runner
         run: |
           rm -rf cobalt/src/out
         shell: bash
+      - name: Install sccache
+        if: ${{ contains(matrix.platform, 'tvos') }}
```
oxve: Avoid gating features by OS as much as possible, especially inside composite actions.

In here, I think you should gate these steps (or new action, really) on the presence of the sccache key instead.

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,59 @@ inputs:
...
+            GCLOUD_ARCHIVE="google-cloud-cli-darwin-x86_64.tar.gz"
+          fi
+          mkdir -p "$GITHUB_WORKSPACE/.local"
+          cd ${{ runner.temp }}
+          curl -L -O "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/${GCLOUD_ARCHIVE}"
+          tar xf "${GCLOUD_ARCHIVE}" -C "$GITHUB_WORKSPACE/.local" --strip-components=1 google-cloud-sdk/bin google-cloud-sdk/lib
+          rm "${GCLOUD_ARCHIVE}"
+          export PATH="$GITHUB_WORKSPACE/.local/bin:$PATH"
+          gcloud --version
+        shell: bash
```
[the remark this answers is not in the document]
oxve: Investigate this comment. I too think this is a bit odd.

Is there a reason the sccache key is different from the key already present on our runners?

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,59 @@ inputs:
...
+    description: "The keyfile needed to access our sccache."
+    required: false
 runs:
   using: "composite"
   steps:
       - name: Clean out dir from previous runner
         run: |
           rm -rf cobalt/src/out
         shell: bash
+      - name: Install sccache
```
oxve, line 23: Move this whole section to a new sccache action.

.github/workflows/main.yaml
```diff
@@ -562,11 +605,13 @@ jobs:
       TEST_ARTIFACTS_KEY: ${{ matrix.platform }}_${{ matrix.name }}_test_artifacts
     steps:
       - name: Restore CI Essentials
+        if: ${{ needs.initialize.outputs.use_cache == 'true' }}
```
oxve: Remove this check. CI essentials is not a runner cache feature

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,59 @@ inputs:
...
+          if [[ "$(uname -m)" == "arm64" ]]; then
+            GCLOUD_ARCHIVE="google-cloud-cli-darwin-arm.tar.gz"
+          else
+            GCLOUD_ARCHIVE="google-cloud-cli-darwin-x86_64.tar.gz"
+          fi
+          mkdir -p "$GITHUB_WORKSPACE/.local"
+          cd ${{ runner.temp }}
+          curl -L -O "https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/${GCLOUD_ARCHIVE}"
+          tar xf "${GCLOUD_ARCHIVE}" -C "$GITHUB_WORKSPACE/.local" --strip-components=1 google-cloud-sdk/bin google-cloud-sdk/lib
+          rm "${GCLOUD_ARCHIVE}"
```
oxve: nit: not necessary

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,59 @@ inputs:
...
+      - name: Install sccache
+        if: ${{ contains(matrix.platform, 'tvos') }}
+        shell: bash
+        run: brew install sccache
+      - name: Configure sccache
+        if: ${{ contains(matrix.platform, 'tvos') }}
+        shell: bash
+        run: |
+          set -x
+          echo "${{ inputs.sccache_gcs_key_mac }}" | base64 --decode > "${HOME}/gcloud.json"
```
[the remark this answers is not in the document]
oxve: Do you really need to base64 encode the key? Bash should be able to handle a json string.

.github/workflows/main.yaml
```diff
@@ -218,6 +220,13 @@ jobs:
           # Create a zero-indexed list of shards for use by matrix, e.g. [0,1,2,3,4,5].
           gtest_shards="[$(seq -s, 0 1 $((${num_gtest_shards} - 1)))]"
           echo "gtest_shards=${gtest_shards}" >> $GITHUB_OUTPUT
+      - name: Set number Docker Requirement
```
oxve: `Set number Docker Requirement`?

.github/workflows/main.yaml
```diff
@@ -247,10 +256,12 @@ jobs:
       test_dimensions: ${{ steps.set-test-dimensions.outputs.test_dimensions }}
       test_attempts: ${{ steps.set-test-attempts.outputs.test_attempts }}
       num_gtest_shards: ${{ steps.set-gtest-shards.outputs.num_gtest_shards }}
+      use_cache: ${{ steps.set-use-cache.outputs.use_cache }}
 
   # Builds, tags, and pushes Cobalt docker build images to ghr.
   docker-build-image:
     needs: [initialize]
+    if: needs.initialize.outputs.use_cache == 'true'
```
oxve: Docker builds should not be gated by the use_cache flag. Use presence of `needs.initialize.outputs.docker_service` in the `build-docker-image` step below instead.

That way this job will pass without building a docker image.

This applies to the two steps below too (afaict they should get skipped already).

.github/config/tvos-arm64-simulator.json
```diff
@@ -1,4 +1,9 @@
 {
+  "docker_service": "",
```
oxve: nit: Don't use an empty field here. Instead add a fallback (`//`) to empty string in initialize. There are other existing examples of this you can look at.

.github/workflows/main.yaml
```diff
@@ -310,8 +321,35 @@ jobs:
     outputs:
       docker_tag: ${{ steps.build-docker-image.outputs.docker_tag }}
 
-  build:
+  build-ready:
```
oxve: What is the purpose of this new job? My guess is that you are trying to workaround the requiredness of the docker build step for the build step below? If that is the case my suggested fix above should take care of it.

.github/actions/sccache/action.yaml
```diff
@@ -0,0 +1,24 @@
...
+      echo "SCCACHE=1" >> $GITHUB_ENV
+      echo "SCCACHE_GCS_BUCKET=cobalt-kokoro-sccache-mac" >> $GITHUB_ENV
+      echo "SCCACHE_GCS_RW_MODE=READ_WRITE" >> $GITHUB_ENV
+      echo "SCCACHE_IDLE_TIMEOUT=0" >> $GITHUB_ENV
+    shell: bash
+  - name: Set Up Cloud SDK
+    if: ${{ contains(matrix.platform, 'tvos') }}
+    uses: isarkis/setup-gcloud@40dce7857b354839efac498d3632050f568090b6 # v1.1.1
+    with:
+      project_id: yt-cobalt
```
oxve: nit: missing newline

Did our newline checker miss this?

.github/config/tvos-arm64-simulator.json
```diff
@@ -14,4 +18,4 @@
       "platform": "tvos-arm64-simulator"
     }
   ]
-}
+}
```
oxve: nit: missing newline

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,27 @@ inputs:
...
+    required: false
 runs:
   using: "composite"
   steps:
       - name: Clean out dir from previous runner
         run: |
           rm -rf cobalt/src/out
         shell: bash
+      - name: Install sccache
+        if: inputs.sccache_gcs_key != ''
```
oxve: Change this to use new variable

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,27 @@ inputs:
   test_targets_json_file:
     description: "The path to the file the test targets are written to."
     required: true
```
oxve, line 12: Add a new param here: `enable_sccache`, default false

.github/actions/sccache/action.yaml
```diff
@@ -0,0 +1,28 @@
...
+    description: "The keyfile needed to access our sccache."
+    required: true
+runs:
+  using: "composite"
+  steps:
+  - name: Install sccache
+    shell: bash
+    run: brew install sccache
+  - name: Configure sccache
+    shell: bash
```
oxve, line 14: `if: inputs.sccache_gcs_key != ''`

.github/actions/sccache/action.yaml
```diff
@@ -0,0 +1,28 @@
...
+      echo "$SCCACHE_GCS_KEY" > "${HOME}/gcloud.json"
+      echo "SCCACHE_GCS_KEY_PATH=${HOME}/gcloud.json" >> $GITHUB_ENV
+      echo "SCCACHE=1" >> $GITHUB_ENV
+      echo "SCCACHE_GCS_BUCKET=cobalt-kokoro-sccache-mac" >> $GITHUB_ENV
+      echo "SCCACHE_GCS_RW_MODE=READ_WRITE" >> $GITHUB_ENV
+      echo "SCCACHE_IDLE_TIMEOUT=0" >> $GITHUB_ENV
+  - name: Set Up Cloud SDK
+    uses: isarkis/setup-gcloud@40dce7857b354839efac498d3632050f568090b6 # v1.1.1
+    with:
+      project_id: yt-cobalt
```
oxve: Is this required? Hopefully this works without a key

.github/actions/build/action.yaml
```diff
@@ -10,19 +10,27 @@ inputs:
...
+        uses: ./.github/actions/sccache
+        with:
+          sccache_gcs_key: ${{ inputs.sccache_gcs_key }}
       - name: GN gen
         run: |
           cd cobalt/src
           gn_flags=()
           if [[ "${{ matrix.platform }}" == *"tvos"* ]]; then
-            gn_flags+=(--no-rbe --no-check)
+            gn_flags+=(--no-rbe --no-check --use-sccache)
```
oxve, line 33: nit: move  `--use-sccache` under new if guarded by `enable_sccache`

Should be put in an env: block var for security reasons

.github/actions/build/action.yaml
```diff
@@ -118,6 +126,9 @@ runs:
           cd cobalt/src
           TARGETS=$(echo "${TARGETS_JSON}" | jq -cr '. | join(" ")')
           time autoninja -C out/${{ matrix.platform }}_${{ matrix.config }} ${TARGETS}
+          if [[ "${{ matrix.platform }}" == *"tvos"* ]]; then
```
oxve: Use `enable_sccache` instead

Remember env: block var

PR #9005, opened 2026-02-06
Remove browsertests from causing CI failures.

Awallky:
After landing PR#8982, CI was broken due to failing browsertest builds. This change removes them entirely, since they will not be supported in 26.android.

Bug: 481767228


borongc:
Thanks, @Awallky and @haozheng-cobalt.

PR #9009, opened 2026-02-07
starboard: Use factories for Android AudioTrack classes

kjyoun:
Introduce static Create() factory methods for AudioTrackBridge and AudioTrackAudioSink, guaranteeing that instantiated objects are always in a fully valid state upon creation.

This change enforces object invariants at the point of creation, eliminates the need for redundant is_valid() checks, and prevents the existence of partially initialized objects. The PassKey pattern is used to restrict constructor access to the factory methods.

Bug: 479994435

### DIFF
--- a/starboard/android/shared/audio_sink_min_required_frames_tester.cc
+++ b/starboard/android/shared/audio_sink_min_required_frames_tester.cc
@@ -126,19 +126,17 @@ void MinRequiredFramesTester::TesterThreadFunc() {
       is_test_complete_ = false;
     }
 
-    audio_sink_ =
-        AudioTrackAudioSink::Create(
-            /*type=*/nullptr, task.number_of_channels, task.sample_rate,
-            task.sample_type, frame_buffers, max_required_frames_,
-            min_required_frames_ * task.number_of_channels *
-                GetSampleSize(task.sample_type),
-            &MinRequiredFramesTester::UpdateSourceStatusFunc,
-            &MinRequiredFramesTester::ConsumeFramesFunc,
-            &MinRequiredFramesTester::ErrorFunc,
-            /*start_media_time=*/0,
-            /*tunnel_mode_audio_session_id=*/-1,
-            /*is_web_audio=*/false, /*context=*/this)
-            .release();
+    audio_sink_ = AudioTrackAudioSink::Create(
+        /*type=*/nullptr, task.number_of_channels, task.sample_rate,
+        task.sample_type, frame_buffers, max_required_frames_,
+        min_required_frames_ * task.number_of_channels *
+            GetSampleSize(task.sample_type),
+        &MinRequiredFramesTester::UpdateSourceStatusFunc,
+        &MinRequiredFramesTester::ConsumeFramesFunc,
+        &MinRequiredFramesTester::ErrorFunc,
+        /*start_media_time=*/0,
+        /*tunnel_mode_audio_session_id=*/-1,
+        /*is_web_audio=*/false, /*context=*/this);
     if (!audio_sink_) {
       SB_LOG(ERROR) << "Failed to create audio sink.";
       // Mark as complete so we can continue with the next task.
@@ -163,8 +161,7 @@ void MinRequiredFramesTester::TesterThreadFunc() {
     // |min_required_frames_| is shared between two threads. Release audio sink
     // to end audio sink thread before access |min_required_frames_| on this
     // thread.
-    delete audio_sink_;
-    audio_sink_ = nullptr;
+    audio_sink_.reset();
 
     if (wait_timeout) {
       SB_LOG(ERROR) << "Audio sink min required frames tester timeout.";
--- a/starboard/android/shared/audio_sink_min_required_frames_tester.cc
+++ b/starboard/android/shared/audio_sink_min_required_frames_tester.cc
@@ -126,14 +126,28 @@ void MinRequiredFramesTester::TesterThreadFunc() {
       is_test_complete_ = false;
     }
 
-    audio_sink_ = new AudioTrackAudioSink(
-        NULL, task.number_of_channels, task.sample_rate, task.sample_type,
-        frame_buffers, max_required_frames_,
-        min_required_frames_ * task.number_of_channels *
-            GetSampleSize(task.sample_type),
-        &MinRequiredFramesTester::UpdateSourceStatusFunc,
-        &MinRequiredFramesTester::ConsumeFramesFunc,
-        &MinRequiredFramesTester::ErrorFunc, 0, -1, false, this);
+    audio_sink_ =
+        AudioTrackAudioSink::Create(
+            /*type=*/nullptr, task.number_of_channels, task.sample_rate,
+            task.sample_type, frame_buffers, max_required_frames_,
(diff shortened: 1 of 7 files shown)


starboard/android/shared/audio_track_bridge.h
```diff
@@ -36,19 +37,22 @@ class AudioTrackBridge {
...
+  static std::unique_ptr<AudioTrackBridge> Create(
+      SbMediaAudioCodingType coding_type,
+      std::optional<SbMediaAudioSampleType> sample_type,
+      int channels,
+      int sampling_frequency_hz,
+      int preferred_buffer_size_in_bytes,
+      int tunnel_mode_audio_session_id,
+      bool is_web_audio);
+
+  AudioTrackBridge(
```
sideb0ard, line 49: should we make the constructor private here to ensure all uses of AudioTrackBridge come through the static Create(..) method?

kjyoun: With PassKey pattern, it's effective private, since only methods in AudioTrackBridge can call this method

I am following PassKey Pattern in chromium.
https://chromium.googlesource.com/chromium/src/+/HEAD/docs/patterns/passkey.md

e.g.) https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/base/audio_buffer.h?q=%5CbPassKey%5Cb&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:media%2Fbase%2F

xiaomings: Is it possible for you to elaborate the advantage of PassKey here if compared to use private?  Are we going to selectively allow other classes to access this ctor?

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -80,31 +81,52 @@ AudioTrackBridge::AudioTrackBridge(
...
         env,
         env->NewByteArray(channels * GetBytesPerSample(sample_type.value()) *
                           kMaxFramesPerRequest));
   } else {
     SB_NOTREACHED();
   }
 
-  SB_DCHECK(j_audio_data_) << "Failed to allocate |j_audio_data_|";
+  if (j_audio_data.is_null()) {
+    SB_LOG(ERROR) << "Failed to allocate |j_audio_data_|";
```
xiaomings: nit: j_audio_data

starboard/android/shared/audio_sink_min_required_frames_tester.cc
```diff
@@ -126,14 +126,28 @@ void MinRequiredFramesTester::TesterThreadFunc() {
...
+            task.sample_type, frame_buffers, max_required_frames_,
+            min_required_frames_ * task.number_of_channels *
+                GetSampleSize(task.sample_type),
+            &MinRequiredFramesTester::UpdateSourceStatusFunc,
+            &MinRequiredFramesTester::ConsumeFramesFunc,
+            &MinRequiredFramesTester::ErrorFunc,
+            /*start_media_time=*/0,
+            /*tunnel_mode_audio_session_id=*/-1,
+            /*is_web_audio=*/false, /*context=*/this)
+            .release();
```
xiaomings: Can we make |audio_sink_| a `std::unique_ptr<AudioTrackBridge>`?

starboard/android/shared/audio_track_bridge.h
sideb0ard: I'm not sure that is completely correct - from the PassKey header file it states that 
// The primary use case is restricting the construction of an object in
// situations where the constructor **needs** to be public
(https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/types/pass_key.h;l=10)

but yeah, it works, i'm fine with it!

sideb0ard (approved):

starboard/android/shared/audio_track_bridge.h
kjyoun: The main reason for using "PassKey" instead of a `private` constructor is to allow the use of **`std::make_unique`** within our factory method.

 If the constructor is `private`, the callsite (even within a static `Create()` method) is forced to use a bare `new` and wrap it manually: `return std::unique_ptr<Foo>(new Foo());`

However, the [Chromium guidelines](https://www.google.com/search?q=https://chromium.googlesource.com/chromium/src/%2B/main/styleguide/c%2B%2B/c%2B%2B-dos-and-donts.md%23use-std_make_unique-and-base_make_intrusive_ref_counted-instead-of-bare-new) strongly discourage bare `new`.

By using a `PassKey`, we keep the constructor technically `public` (so `std::make_unique` can see it) but effectively private (since only `AudioTrackBridge` can produce the key). This is a common technique adopted in Chromium codebase.

starboard/android/shared/audio_track_bridge.cc
kjyoun: Done

starboard/android/shared/audio_sink_min_required_frames_tester.cc
kjyoun: Good point. addressed it

starboard/android/shared/audio_track_bridge.h
xiaomings: That makes sense.  Thank you for the clarification.

starboard/android/shared/audio_sink_min_required_frames_tester.cc
```diff
@@ -126,14 +126,26 @@ void MinRequiredFramesTester::TesterThreadFunc() {
...
+        /*tunnel_mode_audio_session_id=*/-1,
+        /*is_web_audio=*/false, /*context=*/this);
+    if (!audio_sink_) {
+      SB_LOG(ERROR) << "Failed to create audio sink.";
+      // Mark as complete so we can continue with the next task.
+      {
+        std::lock_guard lock(mutex_);
+        is_test_complete_ = true;
+      }
+      test_complete_cv_.notify_one();
```
xiaomings, line 147: @jasonzhangxx is it possible for you to take a look as the logic is changed.

kjyoun: FYI, here is what Gemini says

---

Here is the breakdown of the improvement in your new logic:

1.  **Old Code Behavior:**
    *   `new AudioTrackAudioSink(...)` was called. Since it's a constructor, it always returns a pointer.
    *   If the internal initialization failed, the `AudioTrackOutThread` might never start or might terminate early without calling `ConsumeFramesFunc` (which is where `is_test_complete_` is usually set and signaled).
    *   The `wait_for(..., 5s)` call would be forced to wait for the entire 5 seconds before moving to the next task.

2.  **New Code Behavior:**
    *   `AudioTrackAudioSink::Create(...)` is called. It returns `nullptr` if the `AudioTrackBridge` cannot be created.
    *   The newly added `if (!audio_sink_)` block **manually signals completion**.
    *   This allows the `wait_for` call to return **immediately**, avoiding a 5-second stall for every task that fails creation.

**Conclusion:**
The logic is **correct and highly beneficial**. It specifically handles the case where the sink fails to initialize by ensuring the tester doesn't sit idle for 5 seconds waiting for a signal from a thread that was never successfully started.

PR #9011, opened 2026-02-07
Add --enable-low-end-device-mode-no-simulated-memory switch

sherryzy:
This switch allows enabling low-end device mode without triggering the memory simulation (which fakes a smaller amount of RAM). This is useful for testing or specific configurations where low-end optimizations are desired but the actual physical memory should be reported accurately.

- Added kEnableLowEndDeviceModeNoSimulatedMemory to base_switches.
- Updated SysInfo::AmountOfPhysicalMemory() and AmountOfAvailablePhysicalMemory() to skip simulation if this switch is present.
- Added ENABLE_LOW_END_DEVICE_MODE_NO_SIMULATED_MEMORY to JavaSwitches.java for Android.
- Added unit tests in JavaSwitchesTest.java.

Bug: 480174114

### DIFF
--- a/base/system/sys_info.cc
+++ b/base/system/sys_info.cc
@@ -39,7 +39,12 @@ int SysInfo::NumberOfEfficientProcessors() {
 // static
 uint64_t SysInfo::AmountOfPhysicalMemory() {
   if (base::CommandLine::ForCurrentProcess()->HasSwitch(
-          switches::kEnableLowEndDeviceMode)) {
+          switches::kEnableLowEndDeviceMode)
+  #if BUILDFLAG(IS_COBALT)
+   && !base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kEnableLowEndDeviceModeNoSimulatedMemory)
+  #endif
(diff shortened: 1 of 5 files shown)


base/system/sys_info.cc
```diff
@@ -55,7 +60,12 @@ uint64_t SysInfo::AmountOfPhysicalMemory() {
 // static
 uint64_t SysInfo::AmountOfAvailablePhysicalMemory() {
   if (base::CommandLine::ForCurrentProcess()->HasSwitch(
-          switches::kEnableLowEndDeviceMode)) {
+          switches::kEnableLowEndDeviceMode)
+#if BUILDFLAG(IS_COBALT)
+      && !base::CommandLine::ForCurrentProcess()->HasSwitch(
```
johnxwork, line 65: Why don't you just name it kDisableSimulatedMemory?

sherryzy: Because this switch is specific for the case where enable-low-end-device-mode is on and no simulated memory for this mode?

PR #9013, opened 2026-02-09
implement dynamic crash annotations using Crashpad

alunev:
This change establishes a pathway for capturing application context and client metadata during native hangs and crashes on Android.

Key components:
- CobaltCrashAnnotations: A thread-safe C++ singleton that manages dynamic Crashpad annotations using StringAnnotation.
- Increased buffer size to 1024 bytes to accommodate large JSON metadata strings passed from the application layer.
- JNI Bridge: Provides a mechanism for Java-side crash annotators on Android to feed annotations into the native singleton.

Verified on target device: metadata and background state annotations now correctly appear in captured minidumps after a triggered hang.

### DIFF
--- a/cobalt/browser/cobalt_crash_annotations.cc
+++ b/cobalt/browser/cobalt_crash_annotations.cc
@@ -14,6 +14,8 @@
 
 #include "cobalt/browser/cobalt_crash_annotations.h"
 
+#include "base/logging.h"
+
 namespace cobalt {
 namespace browser {
 
@@ -44,6 +46,8 @@ void CobaltCrashAnnotations::SetAnnotation(const std::string& name,
   }
 
   if (annotations_.size() >= kMaxAnnotations) {
+    LOG(WARNING) << "Maximum number of crash annotations reached ("
+                 << kMaxAnnotations << "). Ignoring: " << name;
     return;
   }
 


cobalt/browser/crash_annotator/crash_annotator_impl.cc
```diff
@@ -34,10 +37,9 @@ void CrashAnnotatorImpl::Create(
 void CrashAnnotatorImpl::SetString(const std::string& key,
                                    const std::string& value,
                                    SetStringCallback callback) {
-  // TODO(cobalt, b/383301493): actually implement this.
-  LOG(INFO) << "CrashAnnotatorImpl::SetString key=" << key
-            << " value=" << value;
-  std::move(callback).Run(false);
+  cobalt::browser::CobaltCrashAnnotations::GetInstance()->SetAnnotation(key,
```
hlwarriner: Do we actually need this change, here?

This C++ implementation of `mojom::CrashAnnotator` is meant for `!is_androidtv` platforms - see the preprocessor conditional in `cobalt/browser/cobalt_browser_interface_binders.cc` - and it looks to me like this code path still is not used by `is_androidtv` with this change.

Note that this method is implemented (not stubbed) on the `main` and `26.eap` branches. On the main branch at least, we probably want to keep the implementation that goes through a Starboard extension since not all Cobalt platforms use Crashpad.

alunev: thanks, makes sense, reverted this.

cobalt/android/apk/app/src/main/java/dev/cobalt/browser/crashannotator/CrashAnnotatorImplFirstParty.java
```diff
@@ -37,6 +38,7 @@ public void setString(String key,
                           String value,
                           SetString_Response callback) {
         CrashContext.INSTANCE.setCrashContext(key, value);
+        CrashAnnotatorImplFirstPartyJni.get().setAnnotation(key, value);
```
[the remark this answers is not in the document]
alunev: this is handled on the application side, before setting to the annotation. App has more insight into nature of the data sent.

cobalt/browser/cobalt_crash_annotations.cc
```diff
@@ -0,0 +1,58 @@
...
+void CobaltCrashAnnotations::SetAnnotation(const std::string& name,
+                                           const std::string& value) {
+  base::AutoLock lock(lock_);
+  auto it = annotations_.find(name);
+  if (it != annotations_.end()) {
+    it->second->annotation->Set(value);
+    return;
+  }
+
+  if (annotations_.size() >= kMaxAnnotations) {
```
johnxwork, line 48: Are there mechanism to evacuate annotations?

What are the annotations you expect to put in here?

johnxwork (approved):

cobalt/browser/cobalt_crash_annotations.h
```diff
@@ -0,0 +1,71 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
hlwarriner: nit: please update the copyright year in all new files.

cobalt/browser/cobalt_crash_annotations.h
```diff
@@ -0,0 +1,71 @@
...
+class CobaltCrashAnnotations {
+ public:
+  static CobaltCrashAnnotations* GetInstance();
+
+  // The maximum number of dynamic annotations allowed.
+  static constexpr size_t kMaxAnnotations = 32;
+
+  // The maximum size for a crash annotation value in bytes.
+  // Hard limit in crashpad is 20KB. We use 1k here to compromise with memory.
+  static constexpr size_t kMaxAnnotationSize = 1024;
```
hlwarriner: This seems reasonable, but did you check the size of the annotations that are currently sent from the web app?

cobalt/browser/crash_annotator/BUILD.gn
```diff
@@ -20,6 +20,7 @@ source_set("crash_annotator") {
 
   deps = [
     "//base",
+    "//cobalt/browser:cobalt_crash_annotations",
```
hlwarriner: Do you plan to cherry pick this change to `main`? If so, we should probably guard this dependency with `is_androidtv`.

cobalt/browser/cobalt_crash_annotations.h
```diff
@@ -0,0 +1,71 @@
...
+
+namespace cobalt {
+namespace browser {
+
+// A singleton class to manage dynamic Crashpad annotations (Crash Keys) in
+// Cobalt. This allows capturing context like application state or background
+// state that changes during the process lifetime.
+class CobaltCrashAnnotations {
+ public:
+  static CobaltCrashAnnotations* GetInstance();
```
hlwarriner, line 39: I'm concerned about this comment from https://github.com/youtube/cobalt/blob/26.android/third_party/crashpad/crashpad/client/annotation.h#L48:
```
Annotations should be declared with static storage duration.
```

It seems like the preference in Chromium/Crashpad is to use string literals for the annotation key names, but we obviously can't do that here - unless Cobalt were to have an agreement with the web app about the set of potential annotations it would send.

Anyway, I guess you're implementing the static duration here with the `static` pointer and `base::NoDestructor<CobaltCrashAnnotations>`?

alunev: Yes, to be precise, we use singleton via static base::NoDestructor + std::map + AnnotationSlot wrapper to provide that stability. Tricky but the point is that we want to support dynamic annotations to avoid coupling with the app.

cobalt/browser/cobalt_crash_annotations.h
alunev: yes, I've checked typical size is ~600 bytes. strings are UTF-8. I think I'll bump it to 2k to have some headroom.

cobalt/browser/cobalt_crash_annotations.h
alunev: done

cobalt/browser/cobalt_crash_annotations.cc
alunev: These annotations are intended for high-level application metadata (e.g., playback state, background status) to provide context during hang analysis. Added ClearAnnotation(), ClearAllAnnotations() to provide way to clear them.

cobalt/browser/crash_annotator/BUILD.gn
alunev: thanks, I've looked into this and seems like it's mergeable as is: ATV-specific JNI/Java path is already isolated via `#if BUILDFLAG(IS_ANDROIDTV)` guards in cobalt_browser_main_parts.cc and cobalt_browser_interface_binders.cc.
so that JNI/Java path is only reachable on ATV, while standard Android uses a C++ stub.

alunev: as discussed, that dependency is just redundant, removed, thanks.

cobalt/browser/cobalt_crash_annotations.cc
```diff
@@ -0,0 +1,82 @@
...
+  // crash reports uploaded to Google servers.
+  base::AutoLock lock(lock_);
+  auto it = annotations_.find(name);
+  if (it != annotations_.end()) {
+    it->second->annotation->Set(value);
+    return;
+  }
+
+  if (annotations_.size() >= kMaxAnnotations) {
+    return;
```
zhongqiliang, line 51: add a log here

alunev: thx, done.

PR #9021, opened 2026-02-09
doc: Fix evergreen content/lib paths

andrewsavage1:
Reverts changes to the older SB 16 posix doc.

Bug: 461621299

### DIFF
--- a/starboard/CHANGELOG.md
+++ b/starboard/CHANGELOG.md
@@ -197,7 +197,7 @@ Now the data field of this SbEvent type is a boolean indicating if text-to-speec
 ## Version 16
 A key update in Starboard version 16 is the adoption of POSIX APIs.
 For a full overview of Starboard POSIX migrations, please refer to
-[migration guide](/starboard/doc/starboard_16_posix.md).
+[migration guide](/starboard/doc/starboard_17_posix.md).
 
 #### Added new configuration constant `kHasPartialAudioFramesSupport`
 Set this to true if your platform supports partial audio frames.
--- a/starboard/doc/starboard_17_posix.md
+++ b/starboard/doc/starboard_17_posix.md
@@ -1,4 +1,4 @@
-# Starboard 16 POSIX APIs
+# Starboard 17 POSIX APIs
 
 
 ## Background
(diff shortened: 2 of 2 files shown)


starboard/doc/evergreen/cobalt_evergreen_overview.md
```diff
@@ -273,7 +273,7 @@ directory tree,
 we would use the following command to run NPLB:
 
 ```sh
-.../elf_loader_sandbox --evergreen_library=libnplb.so
+.../elf_loader_sandbox --evergreen_library=app/nplb/content/libnplb.so
```
yuying-y, line 276: Why are we making the documentation inconsistent, e.g. here using the `app/nplb/content/` prefix for both paths, while in starboard/doc/eap/25_to_26.md, emitting the prefix.

yuying-y: Also, the path is wrong. There is no libnplb.so under `app/nplb/lib/`, only the libnplb.lz4

yuying-y: Copying over what's discussed offline: let's keep the docs consistent, e.g. use `evergreen_library=libxxx.so` and
`evergreen_content=.`

starboard/doc/starboard_16_posix.md
```diff
@@ -33,7 +33,7 @@ ninja -C out/evergreen-arm-softfp_devel/ nplb_install
 3. Run the `nplb` test
 
 ```
-.../elf_loader_sandbox --evergreen_library=libnplb.so
+.../elf_loader_sandbox --evergreen_library=app/nplb/lib/libnplb.so
```
yuying-y: same here

starboard/doc/evergreen/cobalt_evergreen_overview.md
andrewsavage1: In the lines before this one, we are given a directory structure:

> For example, if we wanted to run the NPLB set of tests and had the following
directory tree,
> ```none
> .../elf_loader_sandbox
> .../content/app/nplb/lib/libnplb.so
> .../content/app/nplb/content
> ```
> we would use the following command to run NPLB:

And these commands follow that directory structure. Different partners can have different content paths for `kSbSystemPathContentDirectory`

starboard/doc/starboard_16_posix.md
andrewsavage1: I reverted the changes from the last PR here because this is a starboard 16 migration doc and is outdated for the main/26.eap branches. I'll update this to a Starboard 17 migration doc if that's easy

PR #9029, opened 2026-02-10
tvos: Fix VP9 header include path

abhijeetk:
The include path for the VP9 hardware AV video sample buffer
builder header was incorrectly using the 'internal-starboard'
prefix. This change corrects the path to 'starboard' and ensures
the header is properly included in the build system when the
internal build flag is enabled. This resolves build failures
when the preprocessor cannot locate the header file.

Bug: 452554667
Bug: 460479616

### DIFF
--- a/starboard/tvos/shared/BUILD.gn
+++ b/starboard/tvos/shared/BUILD.gn
@@ -151,6 +151,16 @@ static_library("starboard_platform") {
     "uikit_media_session_client.mm",
   ]
 
+  if (is_internal_build) {
+    # TODO(b/460479616): Update the source paths to the correct internal
+    # submodule location once the directory structure is finalized. These files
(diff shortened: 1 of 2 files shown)


abhijeetk:
cc: @jkim-julie @Gyuyoung

abhijeetk:
Hi Kubo, these changes are guarded by the `is_internal_build` flag. Since the public repository doesn't enable this flag, it won't impact the CI.

rakuco:
> Hi Kubo, these changes are guarded by the `is_internal_build` flag. Since the public repository doesn't enable this flag, it won't impact the CI.

I understand, my point is that if we don't know what the actual path will look like once bug 460479616 is solved, aren't you potentially replacing one broken reference with another broken reference, in which case it makes sense to keep the changes internal while this is sorted out?

PR #9030, opened 2026-02-10
nplb test: Change stack allocation size from 4Mb to 1MB

santoshmahto7:
nplb test `PosixThreadAttrTest.StackAddrAndSizeAttr` was failing due to large stack size allocation, possibly limitation on platform implementation

Bug: 479846471
Change-Id: If502181c9aef3ce889a0309c68f83847a9699058


starboard/nplb/posix_compliance/posix_thread_attr_test.cc
```diff
@@ -20,7 +20,7 @@
 namespace nplb {
 namespace {
 
-constexpr int kStackSize = 4 * 1024 * 1024;
+constexpr int kStackSize = 1 * 1024 * 1024;
```
niranjanyardi, line 23: is there any documentation on this ? if so, please add it

PR #9038, opened 2026-02-10
android: Enable 25% rollout for StartupGuard

zhongqiliang:
Limit the execution of the StartupGuard crash scheduling logic to
approximately 25% of app launches. This introduces a random check
to reduce the frequency of StartupGuard activations.

The purpose is to collect production data and to mitigate crash
volume if it hangs before Kimono get the ColdConfig updated, 
which would otherwise lead to an undesired crash.

Bug: 467729316

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -87,6 +87,10 @@ public abstract class CobaltActivity extends Activity {
   // How many seconds before the app exits if it fails to land YouTube home page.
   private static final int HANG_APP_CRASH_TIMEOUT_SECONDS = 60;
 
+  // The probability (between 0.0 and 1.0) that the StartupGuard's hang-detection
+  // logic will be activated for a given session.
+  private static final double STARTUP_GUARD_PROBABILITY = 0.25;
+
   // Maintain the list of JavaScript-exposed objects as a member variable
   // to prevent them from being garbage collected prematurely.
   private List<CobaltJavaScriptAndroidObject> javaScriptAndroidObjectList = new ArrayList<>();
@@ -388,8 +392,8 @@ public abstract class CobaltActivity extends Activity {
 
     super.onCreate(savedInstanceState);
 
-    // Use a random check to run the StartupGuard logic only 25% of the time
-    if (Math.random() < 0.25) {
+    // Use a random check to run the StartupGuard logic only a certain percentage of the time.
+    if (Math.random() < STARTUP_GUARD_PROBABILITY) {
       if (getJavaSwitches().containsKey(JavaSwitches.DISABLE_STARTUP_GUARD)) {
         Log.i(TAG, "StartupGuard is disabled by Java switch.");
       } else {


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -388,10 +388,15 @@ protected void onCreate(Bundle savedInstanceState) {
 
     super.onCreate(savedInstanceState);
 
-    if (getJavaSwitches().containsKey(JavaSwitches.DISABLE_STARTUP_GUARD)) {
-      Log.i(TAG, "StartupGuard is disabled by Java switch.");
+    // Use a random check to run the StartupGuard logic only 25% of the time
+    if (Math.random() < 0.25) {
```
Awallky: Can you inline this check with the Java switch to prevent nested block? 
You can hold onto the random number before the if block.

Awallky: Actually, gemini's suggestion is not bad either.

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -117,7 +117,7 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
           "--num-raster-threads="
               + javaSwitches.get(JavaSwitches.NUM_RASTER_THREADS).replaceAll("[^0-9]", ""));
     }
-    
+
```
Awallky, line 120: If you can, can you remove the delta to this file? If it is just formatting tool, then don't worry about it.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -388,10 +388,15 @@ protected void onCreate(Bundle savedInstanceState) {
...
+      if (getJavaSwitches().containsKey(JavaSwitches.DISABLE_STARTUP_GUARD)) {
+        Log.i(TAG, "StartupGuard is disabled by Java switch.");
+      } else {
+        StartupGuard.getInstance().scheduleCrash(HANG_APP_CRASH_TIMEOUT_SECONDS);
+      }
     } else {
-      StartupGuard.getInstance().scheduleCrash(HANG_APP_CRASH_TIMEOUT_SECONDS);
+      Log.i(TAG, "StartupGuard skipped by random 25% rollout check.");
     }
```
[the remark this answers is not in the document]
kaidokert: @gemini-code-assist Invalid feedback - the current logic is correct. We want the randomness to be taken into account first, in case there's a side effect or a problem caused by `getJavaSwitches` itself. None of this related code should run at all if the dice roll doesn't pass.

The feedback about rollout percentage -> named constant is reasonable here.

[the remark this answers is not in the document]
kaidokert: @gemini-code-assist i also disagree on the second point on trying to "optimize" a **single** call to Math.random - it makes code harder to follow for no particular gains. This is a piece of code that only runs ONCE during startup.

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
zhongqiliang: This is changed by the formatting tool automatically.

PR #9049, opened 2026-02-11
cobalt/build: Push browser test Docker images to registry

Awallky:
Enable the browser test runners in CI by building and pushing a specialized Docker image when cobalt_browsertests is a target.

This update:
1. Authenticates with the us-west1 Artifact Registry.
2. Colocates the browser test Dockerfile with its dependencies in the artifacts directory.
3. Builds and pushes the image to: us-west1-docker.pkg.dev/ytlr-image-hosting/integration-test-images/avvall-service-test
4. Uses a unique tag based on the run ID and platform for traceability.

This ensures that the necessary environment is available for downstream on-device or on-host browser test execution.

Test: `vpython3 cobalt/tools/test_on_device_tests_gateway_client.py`
Test: `vpython3 cobalt/build/test_archive_test_artifacts.py`
Test: `vpython3 cobalt/testing/browser_tests/tools/test_collect_test_artifacts.py`
Test: `vpython3 cobalt/testing/browser_tests/test_run_browser_tests.py`
Test: `vpython3 cobalt/tools/test_on_device_tests_gateway_client.py`

Bug: 483488213
Bug: 473909665
Bug: 473909293

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -670,6 +670,7 @@ jobs:
           test_attempts: '${{ needs.initialize.outputs.test_attempts }}'
           docker_tag: ${{ steps.build-docker-image.outputs.docker_tag }}
           test_targets: '${{ needs.initialize.outputs.browser_test_targets }}'
+          config: ${{ matrix.config }}
 
   browser-test-on-host:
     needs: [initialize, build, docker-unittest-image]
--- a/.github/actions/browser_tests/action.yaml
+++ b/.github/actions/browser_tests/action.yaml
@@ -22,6 +22,9 @@ inputs:
   test_targets:
     description: "JSON string representing the test targets."
     required: true
+  config:
+    description: "The build configuration (e.g. devel, qa)."
+    required: true
 
 runs:
   using: "composite"
@@ -68,9 +71,24 @@ runs:
         DOCKER_TAG: ${{ inputs.docker_tag }}
         GCS_RESULTS_PATH: ${{ inputs.gcs_results_path }}
         PLATFORM: ${{ matrix.platform }}
+        CONFIG: ${{ matrix.config }}
         TEST_TARGETS_JSON: ${{ inputs.test_targets }}
       run: |
         set -x
+
+        # Determine the gtest-filter from the filter file if it exists.
+        # We only handle the first target for filter application in ODT for now.
+        test_target=$(echo "${TEST_TARGETS_JSON}" | jq -r '.[0].target' | awk -F: '{print $NF}')
+        TEST_FILTER_JSON="${GITHUB_WORKSPACE}/cobalt/testing/filters/${PLATFORM}/${test_target}_filter.json"
+        GTEST_FILTER="*"
+        if [ -f "${TEST_FILTER_JSON}" ]; then
+            FAILING_TESTS=$(jq -r '.failing_tests | join(":")' "${TEST_FILTER_JSON}")
+            if [ -n "${FAILING_TESTS}" ]; then
+                GTEST_FILTER="-${FAILING_TESTS}"
+            fi
+        fi
+
+        CONTAINER_ARGS="--json-results-file=/app/results/summary.json|--logcat-output-file=/app/results/logcat.txt|--environment=local|--test-arguments=\\\"-e org.chromium.native_test.NativeTest.CommandLineFlags '--gtest_output=xml:/app/results/${test_target}_testoutput.xml --gtest_filter=${GTEST_FILTER}'\\\""
 
         python3 -u cobalt/tools/on_device_tests_gateway_client.py \
           --token "${GITHUB_TOKEN}" \
@@ -79,6 +97,10 @@ runs:
           --device_family "${DEVICE_FAMILY}" \
           --targets "${TEST_TARGETS_JSON}" \
           --filter_json_dir "${GITHUB_WORKSPACE}/cobalt/testing/filters/${PLATFORM}" \
+          --container_location "${DOCKER_TAG}" \
+          --container_env_variable "ANDROID_SERIAL=%deviceid%,TARGET_PLATFORM=${PLATFORM}_${CONFIG}" \
+          --container_args "${CONTAINER_ARGS}" \
+          --container_args_delimiter "|" \
           --label chrobalt_on_device_browser_test \
           --label github_${GITHUB_PR_NUMBER:-postsubmit} \
           --label builder-${{ matrix.platform }} \
@@ -97,7 +119,6 @@ runs:
           --label branch:${GITHUB_REF_NAME} \
           --dimensions "${TEST_DIMENSIONS}" \
           ${TEST_ATTEMPTS:+"--test_attempts" "$TEST_ATTEMPTS"} \
-          ${DOCKER_TAG:+"--docker_tag" "$DOCKER_TAG"} \
           --gcs_archive_path "${GCS_ARTIFACTS_PATH}" \
           --gcs_result_path "${GCS_RESULTS_PATH}" || {
             echo "Finished running tests..."
--- a/cobalt/testing/browser_tests/tools/run_tests.template.py
+++ b/cobalt/testing/browser_tests/tools/run_tests.template.py
@@ -56,6 +56,15 @@ def main():
       description="Portable test runner for Cobalt browser tests.",
       add_help=False)
   parser.add_argument("--init-command", help="Command to run before tests.")
+
+  # These arguments are often passed by the container environment.
+  # We sink them here so they don't cause errors or conflicts with the runner.
+  parser.add_argument("--test-filter", help="Test filter (sink).")
+  parser.add_argument("--test-arguments", help="Additional test args (sink).")
+  parser.add_argument("--json-results-file", help="JSON results file (sink).")
+  parser.add_argument("--logcat-output-file", help="Logcat output file (sink).")
+  parser.add_argument("--environment", help="Execution environment (sink).")
+
   parser.add_argument(
       "target", nargs="?", help="Target platform to run tests for.")
   parser.add_argument(
--- a/cobalt/tools/on_device_tests_gateway_client.py
+++ b/cobalt/tools/on_device_tests_gateway_client.py
@@ -261,6 +261,7 @@ def _process_test_requests(args: argparse.Namespace) -> List[Dict[str, Any]]:
         test_type = 'e2e_test'
         target_name = test_target.split(':')[-1]
         dir_on_device = _DIR_ON_DEV_MAP.get(args.device_family, '')
+
         # Browser tests running in the gateway use the specialized runner.
         # We need to tell it where to output the XML so the gateway can find it.
         if dir_on_device:
@@ -278,6 +279,18 @@ def _process_test_requests(args: argparse.Namespace) -> List[Dict[str, Any]]:
           params.append(f'docker_tag={args.docker_tag}')
         if args.gcs_result_path:
           params.append(f'gcs_result_path={args.gcs_result_path}')
+
+        # Overrides for browser test container configuration.
+        if args.container_location:
+          params.append(f'container_location={args.container_location}')
+        if args.container_env_variable:
+          params.append(f'container_env_variable={args.container_env_variable}')
+        if args.container_args:
+          params.append(f'container_args={args.container_args}')
+        if args.container_args_delimiter:
+          params.append(
+              f'container_args_delimiter={args.container_args_delimiter}')
+
       else:
         params = [f'yt_binary_name={_E2E_DEFAULT_YT_BINARY_NAME}']
         if args.device_family in _GCS_ARCHIVE_DEVICE_FAMILIES:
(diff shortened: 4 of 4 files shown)


.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -93,3 +93,36 @@ runs:
...
           "gs://${project_name}-test-artifacts/${WORKFLOW}/${GITHUB_RUN_NUMBER}/${{matrix.platform}}/"
       shell: bash
+    - name: Login to us-west1 GAR
+      if: inputs.upload_on_device_test_artifacts == 'true' || inputs.upload_on_host_test_artifacts == 'true'
+      run: |
+        set -eux
+        # Login to GAR using the service account token from metadata server.
+        TOKEN_ENDPOINT="http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+        ACCESS_TOKEN=$(curl -H 'Metadata-Flavor: Google' ${TOKEN_ENDPOINT} | jq -r .access_token)
+        printf ${ACCESS_TOKEN} | docker login -u oauth2accesstoken --password-stdin https://us-west1-docker.pkg.dev
```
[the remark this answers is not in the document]
Awallky: Done.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -93,3 +93,36 @@ runs:
...
+      shell: bash
+    - name: Build and Push Browser Test Docker Image
+      if: inputs.upload_on_device_test_artifacts == 'true' || inputs.upload_on_host_test_artifacts == 'true'
+      env:
+        TEST_TARGETS_JSON_FILE: ${{ inputs.test_targets_json_file }}
+      run: |
+        set -eux
+        # Check if cobalt_browsertests is in the targets.
+        if grep -q "cobalt_browsertests" "${TEST_TARGETS_JSON_FILE}"; then
```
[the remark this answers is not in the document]
Awallky: Done.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -93,3 +93,36 @@ runs:
...
+          echo "Found cobalt_browsertests in targets. Building Docker image..."
+
+          # Dockerfile expects cobalt_browsertests_artifacts.tar.gz in the context.
+          # archive_test_artifacts.py creates it in the artifacts directory.
+          cp "${GITHUB_WORKSPACE}/cobalt/src/cobalt/testing/browser_tests/tools/Dockerfile" "${GITHUB_WORKSPACE}/artifacts/"
+          cd "${GITHUB_WORKSPACE}/artifacts/"
+
+          TAG="${{ github.run_id }}-${{ matrix.platform }}-${{ matrix.config }}"
+          IMAGE="us-west1-docker.pkg.dev/ytlr-image-hosting/integration-test-images/avvall-service-test:${TAG}"
```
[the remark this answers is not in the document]
Awallky: Done.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -93,3 +97,35 @@ runs:
...
+        if jq -e '.. | strings | select(contains("cobalt_browsertests"))' "${TEST_TARGETS_JSON_FILE}" > /dev/null; then
+          echo "Found cobalt_browsertests in targets. Building Docker image..."
+
+          # Dockerfile expects cobalt_browsertests_artifacts.tar.gz in the context.
+          # archive_test_artifacts.py creates it in the artifacts directory.
+          cp "${GITHUB_WORKSPACE}/cobalt/src/cobalt/testing/browser_tests/tools/Dockerfile" "${GITHUB_WORKSPACE}/artifacts/"
+          cd "${GITHUB_WORKSPACE}/artifacts/"
+
+          TAG="${{ github.run_id }}-${{ matrix.platform }}-${{ matrix.config }}"
+          IMAGE="us-west1-docker.pkg.dev/ytlr-image-hosting/integration-test-images/${{ inputs.docker_image_name }}:${TAG}"
```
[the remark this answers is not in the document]
Awallky: Done.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -93,3 +97,35 @@ runs:
...
+        curl -s -H 'Metadata-Flavor: Google' ${TOKEN_ENDPOINT} | jq -r .access_token | docker login -u oauth2accesstoken --password-stdin https://us-west1-docker.pkg.dev
+      shell: bash
+    - name: Build and Push Browser Test Docker Image
+      if: inputs.upload_on_device_test_artifacts == 'true' || inputs.upload_on_host_test_artifacts == 'true'
+      env:
+        TEST_TARGETS_JSON_FILE: ${{ inputs.test_targets_json_file }}
+      run: |
+        set -eux
+        # Check if cobalt_browsertests is in the targets using jq for robust parsing.
+        if jq -e '.. | strings | select(contains("cobalt_browsertests"))' "${TEST_TARGETS_JSON_FILE}" > /dev/null; then
```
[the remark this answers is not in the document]
Awallky: Done.

.github/actions/internal_tests/action.yaml
```diff
@@ -84,6 +87,7 @@ runs:
           --dimensions '${{ inputs.test_dimensions }}' \
           --artifact_name '${{ inputs.artifact_name }}' \
           --device_family '${{ inputs.test_device_family }}' \
+          ${{ inputs.docker_tag && format('--docker_tag {0}', inputs.docker_tag) }} \
```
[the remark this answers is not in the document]
Awallky: Fixed.

.github/workflows/main.yaml
```diff
@@ -391,9 +392,77 @@ jobs:
         uses: ./.github/actions/api_leak_detector
         if: inputs.run_api_leak_detector
 
-  on-device-test:
+  docker-browsertests-image:
     needs: [initialize, build]
     if: |
+      !contains(fromJSON('["android-x86", "chromium_android-arm", "chromium_android-arm64", "chromium_android-x86", "evergreen-arm-softfp", "evergreen-arm64", "chromium_linux"]'), inputs.platform)
```
oxve: Can you make this a field in the platform config json and load it in initialize?

.github/actions/internal_tests/action.yaml
```diff
@@ -16,6 +16,9 @@ inputs:
   test_device_family:
     description: "The test device family (e.g., raspi, rdk)."
     default: ""
+  docker_tag:
```
SortaCookie: Would it be worth making a new container test action? @oxve any thoughts?

.github/workflows/main.yaml
```diff
@@ -391,9 +392,77 @@ jobs:
...
+            echo "cobalt_browsertests not found in targets. Skipping Docker build."
+          fi
+        shell: bash
+
+  on-device-test:
+    needs: [initialize, build, docker-browsertests-image]
+    if: |
+      always() &&
+      needs.initialize.result == 'success' &&
+      needs.build.result == 'success' &&
```
oxve: Why changing this?

.github/workflows/main.yaml
```diff
@@ -391,9 +392,77 @@ jobs:
...
+        with:
+          fetch-depth: 1
+          filter: blob:none
+          sparse-checkout: ${{ env.CI_ESSENTIALS }}
+      - name: Set up Cloud SDK
+        uses: isarkis/setup-gcloud@40dce7857b354839efac498d3632050f568090b6 # v1.1.1
+      - name: Parse Test Target JSON
+        id: test-target-json
+        uses: ./.github/actions/test_targets_json
+      - name: Build and Push Browser Test Docker Image
```
oxve: The re-usable docker action should be used here. All the docker code below is there already. It may need some modifications to get the GAR path how you want it but it shouldn't be too bad.

.github/workflows/main.yaml
SortaCookie: Better to add a bit (use_browser_test) to the platform json config

.github/workflows/main.yaml
```diff
@@ -468,12 +541,16 @@ jobs:
           test_targets: '${{ needs.initialize.outputs.e2e_test_targets }}'
           test_dimensions: '${{ needs.initialize.outputs.test_dimensions }}'
           test_device_family: '${{ needs.initialize.outputs.test_device_family}}'
+          docker_tag: ${{ github.run_id }}-${{ matrix.platform }}-${{ matrix.config }}
 
   yts-test:
-    needs: [initialize, build]
+    needs: [initialize, build, docker-browsertests-image]
```
SortaCookie: If we split out the container type we don't have to permate all the container stuff to the other tests. Then we dont have the dependency on the docker image.

.github/workflows/main.yaml
```diff
@@ -391,9 +392,77 @@ jobs:
...
+          REGISTRY_PROJECT: ytlr-image-hosting
+          REGISTRY_REPOSITORY: integration-test-images
+          PLATFORM: ${{ matrix.platform }}
+          CONFIG: ${{ matrix.config }}
+          WORKFLOW: ${{ github.workflow }}
+        run: |
+          set -eux
+          # Check if cobalt_browsertests is in the targets using jq for robust parsing.
+          if echo "${TEST_TARGETS_JSON}" | jq -e '.[] | select(endswith(":cobalt_browsertests") or . == "cobalt_browsertests")' > /dev/null; then
+            echo "Found cobalt_browsertests in targets. Building Docker image..."
```
SortaCookie: You can pre parse this in initialize and the use and if check to see if there's something to build.

.github/workflows/main.yaml
```diff
@@ -391,9 +392,77 @@ jobs:
...
+
+            docker build -f Dockerfile -t "${IMAGE}" .
+            docker push "${IMAGE}"
+          else
+            echo "cobalt_browsertests not found in targets. Skipping Docker build."
+          fi
+        shell: bash
+
+  on-device-test:
+    needs: [initialize, build, docker-browsertests-image]
```
oxve: I don't think you need to add the need here. Even if you technically need it to complete first you're not using any of its outputs.

I don't see the order being an issue. The build step should be much slower.

Then you don't need this additional if clauses.

Is this an early optimization or have you seen the docker build fail a lot?

.github/actions/internal_tests/action.yaml
oxve: I think browser tests might be a beast different enough that it should have its own test step, similar to how yts and e2e use the same composable action.

When adding the config flag we should remove the target from the test target json and instead add a new test job that runs builds the docker image and runs browser tests.

I think it was optimistic of me to add it to the unit test targets, even if they use the same run mechanism.

.github/workflows/main.yaml
```diff
@@ -391,9 +392,77 @@ jobs:
         uses: ./.github/actions/api_leak_detector
         if: inputs.run_api_leak_detector
 
-  on-device-test:
+  docker-browsertests-image:
```
oxve: Make this step both build the image and run browser tests, see discussion at the top.

Libzu (approved):

.github/workflows/main.yaml
Awallky: I refactored things into a browser_tests action

.github/workflows/main.yaml
Awallky: Done.

.github/workflows/main.yaml
Awallky: It was just an early optimization. I refactored things now and reverted this change.

.github/workflows/main.yaml
Awallky: I think I refactored things as you specified into a browser_tests action.

.github/workflows/main.yaml
Awallky: I think I refactored things as you intended into a browser_tests action.

.github/actions/internal_tests/action.yaml
Awallky: Done.

.github/actions/docker_browsertests/action.yaml
```diff
@@ -0,0 +1,96 @@
...
+        # Set floating to either the PR number (for PRs) or the branch name (for post-submit).
+        FLOATING_TAG="${{ inputs.is_pr == 'true' && github.event.pull_request.number || github.ref_name }}"
+        FLOATING_TAG="${FLOATING_TAG//\//__}"
+        FLOATING_TAG="${FLOATING_TAG%.1[+,-]}"
+        echo "DOCKER_FLOATING_TAG=${FLOATING_TAG}" >> $GITHUB_ENV
+
+        REGISTRY_REGION="us-west1"
+        REGISTRY_PROJECT="ytlr-image-hosting"
+        REGISTRY_REPOSITORY="integration-test-images"
+        DOCKER_IMAGE_NAME="avvall-service-test"
```
SortaCookie: Lets just change this value to a better name now.

.github/workflows/main.yaml
```diff
@@ -270,17 +279,18 @@ jobs:
 
   # Builds, tags, and pushes Cobalt unit test image to ghr.
   docker-unittest-image:
-    if: needs.initialize.outputs.test_on_host == 'true'
     needs: [initialize]
     runs-on: [self-hosted, chrobalt-linux-runner]
     steps:
       - name: Restore CI Essentials
+        if: needs.initialize.outputs.test_on_host == 'true'
```
SortaCookie: Why are we moving the check? Is it to ensure docker-unittest-image always runs?

.github/workflows/main.yaml
```diff
@@ -293,17 +303,18 @@ jobs:
 
   # Builds, tags, and pushes Cobalt web test image to ghr.
   docker-webtest-image:
-    if: needs.initialize.outputs.web_tests == 'true'
     needs: [initialize]
     runs-on: [self-hosted, chrobalt-linux-runner]
     steps:
       - name: Restore CI Essentials
+        if: needs.initialize.outputs.web_tests == 'true'
```
SortaCookie: Same there

Awallky: This no longer applies. I reverted this after refactoring the shared deps between the unittest-image and browsertests.

.github/workflows/main.yaml
Awallky: This no longer applies. I reverted this after refactoring the shared deps between the unittest-image and browsertests.

.github/actions/docker_browsertests/action.yaml
Awallky: Done

PR #9052, opened 2026-02-11
Add Browser Tests and YTS WPT Tests to Presubmit

Libzu:
Integrate Cobalt Browser tests and YTS WPT tests into the GitHub CI workflow.

Bug: 473909877

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -604,8 +604,8 @@ jobs:
       needs.initialize.outputs.test_browser == 'true' &&
         (
           (github.event_name == 'pull_request' && github.event.pull_request.draft == false ) ||
-          ((inputs.nightly == 'true' || github.event_name == 'schedule') && vars.RUN_ODT_TESTS_ON_NIGHTLY != 'False') ||
-          (github.event_name == 'push' && vars.RUN_ODT_TESTS_ON_POSTSUBMIT != 'False')
+          (inputs.nightly == 'true' || github.event_name == 'schedule') ||
+          (github.event_name == 'push')
         )
     runs-on: [self-hosted, odt-runner]
     name: ${{ matrix.name }}_browser_tests
@@ -634,9 +634,8 @@ jobs:
     if: |
       needs.initialize.outputs.test_yts_wpt == 'true' &&
         (
-          (github.event_name == 'pull_request' && github.event.pull_request.draft == false ) ||
-          ((inputs.nightly == 'true' || github.event_name == 'schedule') && vars.RUN_ODT_TESTS_ON_NIGHTLY != 'False') ||
-          (github.event_name == 'push' && vars.RUN_ODT_TESTS_ON_POSTSUBMIT != 'False')
+          (inputs.nightly == 'true' || github.event_name == 'schedule') ||
+          (github.event_name == 'push')
         )
     runs-on: [self-hosted, odt-runner]
     name: ${{ matrix.name }}_yts_wpt_tests


Awallky:
Don't we want this on arm64 targets as well? 
What about linux-x64x11?

.github/config/android-arm.json
```diff
@@ -58,6 +61,24 @@
       "test_attempts": "6"
     }
   ],
+  "yts_playback_test_targets": [
+    {
+      "target": "testing/maneki/cobalt/kimono_tests:ATV_KIMONO_YTS_PLAYBACK_TESTS_ARM_DEVICES",
```
isarkis, line 90: This will make YTS Playback test to run on random ARM devices (e.g. sabrina, boreal, kirkwood, etc.).   Have we checked with @osagie98 to see if it that would be ok vs having a separate workflow for Boreal, Kirkwood, Sabrina, etc ?

.github/workflows/main.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
       GITHUB_EVENT_NUMBER: ${{ github.event.number }}
     if: |
       github.event.action != 'labeled' ||
-      (github.event.pull_request.merged == false && github.event.label.name == 'runtest')
+      (github.event.pull_request.merged == false && (github.event.label.name == 'runtest' || github.event.label.name == 'yts_playback_test'))
```
SortaCookie, line 57: nit: maybe a label for integration-tests or something because the browser and wpt tests are not playback.

.github/workflows/main.yaml
```diff
@@ -502,6 +550,98 @@ jobs:
...
+          fetch-depth: 1
+          filter: blob:none
+          sparse-checkout: ${{ env.CI_ESSENTIALS }}
+      - name: Run YTS Playback Tests
+        uses: ./.github/actions/internal_tests
+        with:
+          test_type: 'yts_test'
+          test_targets: '${{ needs.initialize.outputs.yts_playback_test_targets }}'
+
+  browser-test:
```
SortaCookie, line 601: It might be worthwhile making a github action for container-tests in general there's a list of params that exist that could be easily configurable with gha inputs. See http://shortn/_aeSJB6SD4e

cobalt/tools/on_device_tests_gateway_client.py
```diff
@@ -244,24 +245,32 @@ def _process_test_requests(args: argparse.Namespace) -> List[Dict[str, Any]]:
...
       if test_attempts:
         test_args.extend([f'test_attempts={test_attempts}'])
       elif args.test_attempts:
         test_args.extend([f'test_attempts={args.test_attempts}'])
       test_cmd_args = []
-      params = [f'yt_binary_name={_E2E_DEFAULT_YT_BINARY_NAME}']
       files = []
-      if args.device_family in _GCS_ARCHIVE_DEVICE_FAMILIES:
-        params.append(f'gcs_cobalt_archive=gs://{args.cobalt_path}.zip')
+      if test_type in ('browser_test', 'yts_wpt_test'):
```
SortaCookie, line 257: This is fine for now. But we're going to need to add a number of params to be set. See: http://shortn/_aeSJB6SD4e

.github/config/android-arm.json
DhruvRaipure24: It will not run on random devices. It will run on all the devices mentioned in the BUILD rule, ie. boreal, bravia, sabrina and kirkwood. Austin is aware of this change.

.github/workflows/main.yaml
Libzu: I understand that we are adding yts_playback_test label here to trigger yts playback tests on demand, but browser/container tests are run as part of runtest, so it would not be necessary to create additional tags for it, right @isarkis?

cobalt/tools/on_device_tests_gateway_client.py
Libzu: sg

.github/workflows/main.yaml
isarkis: AFAIK no additional labels are needed, yts_playback_test label is not related to WPT and Browser tests.

.github/workflows/main.yaml
Libzu: will do in next pr

PR #9055, opened 2026-02-11
Add PR step to sync Chromium branches.

briantting:
Rather than attempt to force push the cherry picks of m138 refresh commits, create pull requests instead.

Bug: 481788398

### DIFF
--- a/.github/workflows/sync_chromium_branches.yaml
+++ b/.github/workflows/sync_chromium_branches.yaml
@@ -7,9 +7,10 @@ on:
 permissions: read-all
 jobs:
   sync:
-    runs-on: [self-hosted, chrobalt-linux-runner]
+    runs-on: ubuntu-latest
     permissions:
       contents: write
+      pull-requests: write
     strategy:
       fail-fast: false
       matrix:
@@ -30,12 +31,14 @@ jobs:
         with:
           ref: chromium/${{ matrix.branch.milestone }}
           fetch-depth: 1
+          filter: blob:none
       - name: Run fallback actions/checkout@v4
         uses: actions/checkout@v4
         if: steps.checkout.outcome == 'failure'
         with:
           ref: main
           fetch-depth: 1
+          filter: blob:none
       - name: Set up Git
         run: |
           git config --global user.name "GitHub Release Automation"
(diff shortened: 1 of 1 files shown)


briantting:
https://github.com/youtube/cobalt_sandbox/actions/runs/21927572936/job/63323780141
https://github.com/youtube/cobalt_sandbox/pull/831

tested on sandbox

.github/workflows/sync_chromium_branches.yaml
```diff
@@ -10,6 +10,7 @@ jobs:
     runs-on: [self-hosted, chrobalt-linux-runner]
```
oxve: Does this need to use our self-hosted runners? gh cli is pre-installed on the GH hosted ones.

.github/workflows/sync_chromium_branches.yaml
```diff
@@ -72,16 +73,36 @@ jobs:
             echo "main_cherry_pick=false" >> "$GITHUB_OUTPUT"
           fi
       - uses: actions/checkout@v4
-        id: checkout_main
+        id: checkout_cherry-pick-main-sync
```
oxve: nit: mixing underscores and dashes

That said, based on my comment below you don't need this id at all.

.github/workflows/sync_chromium_branches.yaml
```diff
@@ -72,16 +73,36 @@ jobs:
...
         if: steps.push_diff.outputs.main_cherry_pick == 'true' && contains(matrix.branch.milestone, 'm138')
         with:
-          ref: main
-          fetch-depth: 1
-      - name: Cherry Pick and Push upstream diff to main
-        if: steps.checkout_main.outcome == 'success'
+          ref: cherry-pick-main-sync
+          fetch-depth: 16
+      - name: Install GitHub CLI
+        if: steps.checkout_cherry-pick-main-sync.outcome == 'success'
```
oxve: This is the default unless you specify `continue-on-error: true` on the step above.

.github/workflows/sync_chromium_branches.yaml
```diff
@@ -72,16 +73,36 @@ jobs:
...
+            && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | sudo tee /etc/apt/sources.list.d/github-cli.list > /dev/null
+          sudo apt update
+          sudo apt install gh -y
+      - name: Cherry Pick upstream diff to main
+        if: steps.checkout_cherry-pick-main-sync.outcome == 'success'
         run: |
+          git pull --rebase origin main
           if ! git cherry-pick "${{ steps.push_diff.outputs.main_cherry_pick_hash }}"; then
             git status | sed -n "s/^[[:space:]]*deleted by us:[[:space:]]*//p" | xargs -d "\n" git rm
             git cherry-pick --continue --no-edit
```
oxve, line 91: I think this should also go a `git add .` to automatically add all conflicting files before continuing the cherry pick.

That's what we did for the cherry pick workflow.

.github/workflows/sync_chromium_branches.yaml
briantting: What do the self-hosted runners do that GH hosted ones can't? I didn't change this, but the asks here are pretty simple, mostly just git commands.

.github/workflows/sync_chromium_branches.yaml
briantting: If there isn't this if statement wouldn't this step run for all branches? Currently it only runs for m138 because the above step has the corresponding if check.

.github/workflows/sync_chromium_branches.yaml
briantting: just as an insurance that we have all the files conflicted? sgtm

.github/workflows/sync_chromium_branches.yaml
oxve: Add `filter: blob:none` to speed up checkup and avoid downloading unnecessary blobs. Here and the next step.
See https://github.com/youtube/cobalt/blob/main/.github/workflows/main.yaml#L64

.github/workflows/sync_chromium_branches.yaml
briantting: tested GH runners work

linxinan-chops (approved):
LGTM

PR #9061, opened 2026-02-11
Add runtime switch to disable storage migration

haozheng-cobalt:
Introduce the --disable-storage-migration command line switch.
This switch provides a runtime mechanism to prevent the automatic
storage migration process from executing. It allows operators to
disable migration, which can be useful for debugging, testing, or
in specific deployment scenarios where migration may not be desired
or could cause issues.

Bug: 483474656

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
@@ -24,6 +24,7 @@ import java.util.Map;
 public class JavaSwitches {
   public static final String ENABLE_QUIC = "EnableQUIC";
   public static final String DISABLE_STARTUP_GUARD = "DisableStartupGuard";
+  public static final String DISABLE_STORAGE_MIGRATION = "DisableStorageMigration";
   public static final String DISABLE_LOW_END_DEVICE_MODE = "DisableLowEndDeviceMode";
   public static final String ENABLE_LOW_END_DEVICE_MODE_NO_SIMULATED_MEMORY =
       "EnableLowEndDeviceModeNoSimulatedMemory";
@@ -107,6 +108,9 @@ public class JavaSwitches {
 
     if (javaSwitches.containsKey(JavaSwitches.DISABLE_SPLASH_SCREEN)) {
       extraCommandLineArgs.add("--disable-splash-screen");
+    }
+    if (javaSwitches.containsKey(JavaSwitches.DISABLE_STORAGE_MIGRATION)) {
+      extraCommandLineArgs.add("--disable-storage-migration");
     }
     if (javaSwitches.containsKey(JavaSwitches.FORCE_IMAGE_SPLASH_SCREEN)) {
       extraCommandLineArgs.add("--force-image-splash-screen");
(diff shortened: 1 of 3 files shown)


cobalt/shell/browser/migrate_storage_record/migrate-storage-experiment.md
oxve, line 1: Looks like you added some extra files to the PR

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -108,6 +109,9 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
     if (javaSwitches.containsKey(JavaSwitches.DISABLE_SPLASH_SCREEN)) {
       extraCommandLineArgs.add("--disable-splash-screen");
     }
+    if (javaSwitches.containsKey(JavaSwitches.DISABLE_STORAGE_MIGRATION)) {
+      extraCommandLineArgs.add("--disable-storage-migration");
+    }
```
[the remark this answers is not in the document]
haozheng-cobalt: I'd stick the existing pattern for now.

cobalt/shell/browser/migrate_storage_record/migrate-storage-experiment.md
haozheng-cobalt: yep haha, removed them.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -241,6 +241,12 @@ void MigrationManager::DoMigrationTasksOnce(
     return;
   }
 
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
```
zhongqiliang, line 244: move this above if (migration_attempted_.test_and_set()) {

haozheng-cobalt: I think it might be better to check this atomic flag first, so that we ensure that the switch check (and everything following it) only executes exactly once per application session.

zhongqiliang: The migration_attempted_.test_and_set() check ensures this method only executes once per session. Your new disable flag should act as a master override, taking precedence over this logic.

haozheng-cobalt: yes, but if the switch flag was moved before the migration_attempted_ check, the logging in the switch flag would be printed many times when the storage migration is disabled. right?

zhongqiliang: I am not sure I know which logging you mean. can you remove the logging? or does it matter?

haozheng-cobalt: I meant the logging at line 246.

zhongqiliang: The log isn't strictly necessary here. From a structural standpoint, it would be cleaner to move that logic further up or even outside the method entirely. 
I’m happy to approve this as-is since it’s functional, but it’s worth considering for better readability.

PR #9064, opened 2026-02-11
build: Fix race condition in Blink modules

oxve:
build: Fix race condition in Blink modules

Fix a build race condition where Blink bindings were being compiled before
the required Mojom headers from Cobalt's platform service were generated.

The failure was observed in Cobalt Evergreen nightly builds (e.g.,
arm-hardfp-rdk). Adding the explicit dependency on the Mojom target to the
make_modules_generated group ensures correct ordering in the build graph.

Bug: 483148600

### DIFF
--- a/third_party/blink/renderer/modules/BUILD.gn
+++ b/third_party/blink/renderer/modules/BUILD.gn
@@ -342,7 +342,7 @@ group("make_modules_generated") {
   ]
 
   if (is_cobalt) {
-    public_deps += [ "//cobalt/browser/h5vcc_platform_service/public/mojom:mojom_blink" ]
+    public_deps += [ "//cobalt/browser/h5vcc_platform_service/public/mojom:mojom_blink_headers" ]
   }
 }
 


third_party/blink/renderer/modules/BUILD.gn
```diff
@@ -340,6 +340,10 @@ group("make_modules_generated") {
     "//ui/accessibility:ax_enums_mojo_blink_headers",
     "//ui/display/mojom:mojom_blink_headers",
   ]
+
+  if (is_cobalt) {
+    public_deps += [ "//cobalt/browser/h5vcc_platform_service/public/mojom:mojom_blink" ]
```
hlwarriner: This looks ok but I think `//cobalt/browser/h5vcc_platform_service/public/mojom:mojom_blink_headers` would be better since only the headers are needed. Can you please try that?

oxve: It didn't fail so that's good. We'll have to monitor to make sure the flake is gone.

PR #9082, opened 2026-02-12
Only allow caller to access specified files from cache

linxinan-yt:
Fulfill the review comment form earlier PRs to restrict the access to storage cache

This PR aims to improve the security regarding cache. Only access to the file listed in the resource map are allowed.
But note, this is not a fix for any known bugs, as Cobalt does not accept user input URLs.

Issue: 454630524

### DIFF
--- a/cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
+++ b/cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
@@ -18,6 +18,7 @@
 #include <memory>
 
 #include "base/base64.h"
+#include "base/containers/contains.h"
 #include "base/containers/span.h"
 #include "base/strings/string_util.h"
 #include "base/strings/stringprintf.h"
@@ -235,7 +236,7 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
       resource_key = std::move(fallback);
     }
 
-    if (ExistsInResourceMap(resource_map, resource_key)) {
+    if (base::Contains(resource_map, resource_key)) {
       FileContents file_contents = resource_map[resource_key];
       content_ = std::string(reinterpret_cast<const char*>(file_contents.data),
                              file_contents.size);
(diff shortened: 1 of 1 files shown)


cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -238,7 +238,8 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
       LOG(WARNING) << "Resource not found: " << resource_key;
     }
 
-    if (is_cacheable_type && browser_context_) {
+    if (is_cacheable_type && browser_context_ &&
+        resource_map.find(key) != resource_map.end()) {
       ReadSplashCache(key);
       return;
     }
```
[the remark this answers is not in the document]
linxinan-yt: Solved.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -238,7 +238,8 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
       LOG(WARNING) << "Resource not found: " << resource_key;
     }
 
-    if (is_cacheable_type && browser_context_) {
+    if (is_cacheable_type && browser_context_ &&
+        resource_map.find(key) != resource_map.end()) {
```
borongc: nit: maybe put a bool var for L233, and use that bool for both if statements.

linxinan-yt: I created a helper function for this. But the tihing is that we have to check both key and resource key, so we cannot make one variable here.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
borongc, line 208: nit: as a result of these two variables, we should add a comment here why we need these two variables, `key` and `resource_key` due to fallback files.

linxinan-yt: Ack. Done.

PR #9089, opened 2026-02-13
android: Fix ClientLogInfo and AdvertisingId leaks

jasonzhangxx:
Ensure that executor services in ClientLogInfo and AdvertisingId are
properly shut down when their respective components are closed or
when the application is being destroyed. This prevents thread leaks
and ensures all resources are released, improving application stability
and preventing potential ANRs on shutdown.

Bug: 484183546

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
@@ -207,6 +207,7 @@ public class StarboardBridge {
       Log.i(TAG, "Activity destroyed after shutdown; killing app.");
       StarboardBridgeJni.get().closeNativeStarboard(nativeApp);
       closeAllServices();
+      advertisingId.shutdown();
       activity.finishAndRemoveTask();
     } else {
       Log.i(TAG, "Activity destroyed without shutdown; app suspended in background.");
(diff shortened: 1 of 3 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -207,6 +207,7 @@ protected void onActivityDestroy(Activity activity) {
       Log.i(TAG, "Activity destroyed after shutdown; killing app.");
       StarboardBridgeJni.get().closeNativeStarboard(nativeApp);
       closeAllServices();
+      advertisingId.shutdown();
```
zhongqiliang, line 210: It is probably better to create a thread pool in the constructor of StarboardBridge, and then pass it to AdvertisingId to use. 
You can shutdown the thread pool here.

jasonzhangxx: I actually didn't see AdvertisingId being created multiple times. @zhongqiliang do you think we need to shutdown the thread pool in advertisingId? I'm ok with removing the changes of advertisingId in this pr.

zhongqiliang (approved):

johnxwork (approved):

PR #9091, opened 2026-02-13
tvOS: Set firmware version in User Agent string

rakuco:
Pass the OS version as firmware version just like C25 did, otherwise this field would be set to "Unknown".

Bug: 442760345


rakuco:

PR #9103, opened 2026-02-15
starboard: Avoid unnecessary GetPrerollFrameCount() once seek is completed

kjyoun:
Add a check for the seeking state in `OnDecoderStatus()` prior to invoking `GetPrerollFrameCount()`.

This optimization prevents calling `GetPrerollFrameCount()` for every frame during normal playback, reducing overhead when the player is not actively seeking.

Bug: 1234

### DIFF
--- a/starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
+++ b/starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
@@ -283,7 +283,12 @@ void VideoRendererImpl::OnDecoderStatus(
       }
     }
 
-    if (number_of_frames_.load() >=
+    // Check seeking_ first to avoid an unnecessary call to
+    // GetPrerollFrameCount() during steady-state playback. The final
+    // exchange(false) acts as an atomic latch to ensure prerolled_cb_ is
+    // scheduled exactly once.
+    if (seeking_.load() &&
+        number_of_frames_.load() >=
             static_cast<int32_t>(decoder_->GetPrerollFrameCount()) &&
         seeking_.exchange(false)) {
 #if SB_PLAYER_FILTER_ENABLE_STATE_CHECK


starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -283,7 +283,10 @@ void VideoRendererImpl::OnDecoderStatus(
       }
     }
 
-    if (number_of_frames_.load() >=
+    // Check seeking_ first to avoid an unnecessary call to
+    // GetPrerollFrameCount() when not in a seeking state.
+    if (seeking_.load() &&
+        number_of_frames_.load() >=
             static_cast<int32_t>(decoder_->GetPrerollFrameCount()) &&
         seeking_.exchange(false)) {
```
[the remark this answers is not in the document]
kjyoun: This makes more complex for a very marginal safety gain.

This race condition (and the use of exchange) pre-exists this PR; my goal here is strictly to optimize the hot path by short-circuiting the virtual function call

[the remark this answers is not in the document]
kjyoun: Triple-nesting does not look good.

xiaomings:
Consider adding a valid bug number to the commit message, it's currently set to 1234.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -283,7 +283,12 @@ void VideoRendererImpl::OnDecoderStatus(
       }
     }
 
-    if (number_of_frames_.load() >=
+    // Check seeking_ first to avoid an unnecessary call to
+    // GetPrerollFrameCount() during steady-state playback. The final
+    // exchange(false) acts as an atomic latch to ensure prerolled_cb_ is
+    // scheduled exactly once.
+    if (seeking_.load() &&
```
xiaomings, line 290: Do we see any performance improvements with this change?

kjyoun: This is more about helping understand code.

The previous behavior that the player keep calling preroll count method, even after preroll stage is completed was confusing to me. When I met this behavior during debugging, I thought something went wrong.  After spending some time reading the code, I realized it wasn't a bug, 

I believe stopping these calls once preroll is finished makes the state machine much easier to follow and prevents future confusion.

xiaomings: Consider adding a comment with TODO for now, so anyone who's going to read the code again before this is addressed would understand it immediately, and we can get back to this once Chrobalt ATV is launched.

kjyoun: Filed b/485267396

Closing this PR

PR #9116, opened 2026-02-17
media: Parse IAMF config in VideoDmpReader

osagie98:
media: Parse IAMF config in VideoDmpReader

VideoDmpReader now extracts IAMF primary and additional profiles from the
stream's first access unit. This is achieved by parsing the IAMF
Sequence Header OBU.

This change enables the generation of an accurate and specific MIME type
string for IAMF audio, moving from a generic "iamf" to a detailed
"iamf.XXX.YYY.Opus" format. A precise MIME type is essential for
correct media pipeline configuration and playback compatibility.

Bug: 485254424


starboard/shared/starboard/player/video_dmp_reader.cc
```diff
@@ -157,7 +183,16 @@ std::string VideoDmpReader::audio_mime_type() const {
       ss << "audio/wav; codecs=\"1\";";
       break;
     case kSbMediaAudioCodecIamf:
-      ss << "audio/mp4; codecs=\"iamf\";";
+      if (!dmp_info_.iamf_primary_profile.has_value()) {
+        return "";
```
kjyoun: what does empty mimetype mean?

can you add comment here?

starboard/shared/starboard/player/video_dmp_reader.cc
```diff
@@ -305,6 +386,11 @@ void VideoDmpReader::Parse() {
   while (ParseOneRecord()) {
   }
 
+  if (dmp_info_.audio_codec == kSbMediaAudioCodecIamf &&
+      !dmp_info_.iamf_primary_profile.has_value()) {
+    ParseIamfConfigOBU();
```
kjyoun: this PR detects mimetype by parsing actual data, instead of depending mime type string?

Then can you remove IamfMimeUtil ?

starboard/shared/starboard/media/iamf_util.cc

starboard/shared/starboard/player/video_dmp_reader.cc
kjyoun, line 1: can you add unit tests for this file? 

Though it is out of the scope of this PR, I think that this is a good chance to add one.

We have video_dmp_reader and writer.  Then we can pair them in tests(
(For reader test, generate test data using writer. For writer, verify the written data using reader)
Gemini will do most of chores.

starboard/shared/starboard/player/video_dmp_reader.cc
```diff
@@ -71,6 +74,29 @@ SbPlayerSampleInfo ConvertToPlayerSampleInfo(
   return sample_info;
 }
 
+// Helper function to read a LEB128 value.
+bool ReadLeb128(const uint8_t** data, const uint8_t* end, uint32_t* value) {
+  SB_CHECK(data && *data);
```
kjyoun: nit: this will provide more helpful info. With current code, we don't know which is null, when crash happens.

```
 SB_CHECK(data);
SB_CHECK(*data);
```

starboard/shared/starboard/player/video_dmp_reader.cc
```diff
@@ -71,6 +74,29 @@ SbPlayerSampleInfo ConvertToPlayerSampleInfo(
...
+  SB_CHECK(end);
+  SB_CHECK(value);
+  *value = 0;
```
kjyoun: I would not update return value parameter, when the method fails.


Per Gemini

---

This function currently has side effects on failure. If the input data is malformed or ends prematurely, *value is left in a partially modified state. It is safer to use a local accumulator and only update the output parameter once the full LEB128 sequence is validated.

This follows the "strong exception guarantee" pattern—either the function succeeds completely, or it doesn't change the state at all.

```
// 1. Create a local variable to hold the progress
uint32_t decoded_value = 0; 

for (size_t i = 0; i < 5; ++i) {
    if (*data >= end) {
        return false;
    }
    uint8_t byte = *(*data)++;
    if (i == 4 && (byte & 0x7f) > 0x0f) {
        return false;
    }
    
    // 2. Update the local variable instead of the pointer
    decoded_value |= (uint32_t)(byte & 0x7f) << (i * 7);
    
    if (!(byte & 0x80)) {
        // 3. Only update the external value on guaranteed success
        *value = decoded_value;
        return true;
    }
}
return false;
```

starboard/shared/starboard/player/video_dmp_reader.cc
```diff
@@ -71,6 +74,25 @@ SbPlayerSampleInfo ConvertToPlayerSampleInfo(
...
+      return false;
+    }
+    uint8_t byte = *(*data)++;
+    *value |= (byte & 0x7f) << (i * 7);
+    if (!(byte & 0x80)) {
+      return true;
+    }
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
kjyoun: Good ?

starboard/shared/starboard/player/video_dmp_reader.cc
osagie98: Replaced with a `CHECK`, the file would be invalid if the profile info isn't parsed.

starboard/shared/starboard/player/video_dmp_reader.cc
osagie98: `IamfMimeUtil` validates and extracts information from an already formed mime type provided by the player. `VideoDmpReader` doesn't have this information so it has to create the correct mime type string by parsing the bitstream. I moved this new functionality to `IamfMimeUtil`.

starboard/shared/starboard/player/video_dmp_reader.cc
osagie98: Done

starboard/shared/starboard/player/video_dmp_reader.cc
osagie98: Done

starboard/shared/starboard/player/video_dmp_reader.cc
osagie98: I wrote some VideoDmpReader tests but the framework may be broken and will take time to debug. I'll create a bug to write full test for both the reader and writer.

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -122,4 +153,48 @@ IamfMimeUtil::IamfMimeUtil(const std::string& mime_type) {
   additional_profile_ = additional_profile;
 }
 
+// static.
+bool IamfMimeUtil::ParseIamfSequenceHeaderObu(const uint8_t* data,
```
kjyoun: you can make this method accept `const std::vector<uint8_t>&` directly.

Then callsite can be simpler.

```
  SB_CHECK(IamfMimeUtil::ParseIamfSequenceHeaderObu(
        audio_access_units_[0].data().data(),
        audio_access_units_[0].data().size(),
        &(*dmp_info_.iamf_primary_profile),
        &(*dmp_info_.iamf_additional_profile)));

-->

  SB_CHECK(IamfMimeUtil::ParseIamfSequenceHeaderObu(
        audio_access_units_[0].data(),
        &(*dmp_info_.iamf_primary_profile),
        &(*dmp_info_.iamf_additional_profile)));
```

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -122,4 +153,48 @@ IamfMimeUtil::IamfMimeUtil(const std::string& mime_type) {
...
+                                              size_t size,
+                                              uint8_t* primary_profile,
+                                              uint8_t* additional_profile) {
```
kjyoun: Returning by struct is the best practice

https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-out


In this case, we have to return struct or failure
Then you can use [starboard/common/result.h](https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/common/result.h?q=%20Result&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F)


Then implementation will be like.

```
struct IamfProfile {
  const uint8_t primary;
  const uint8_t additional;
};

Result<IamfProfile> IamfMimeUtil::ParseIamfSequenceHeaderObu(const std::vector<uint8_t>& data) {

 if (obu_type != kIamfSequenceHeaderObu) {
        return Failure ("Tried to read OBU: " + std::to_string(static_cast<int>(obu_type))
                   + " instead of IA Sequence Header OBU type "
                   + std::to_string(static_cast<int>(kIamfSequenceHeaderObu));
  }

// OnSuccess
 return IamfProfile{
   /*primary=*/*data++,
   /*additional=*/*data;
 }
```


And callsite will be like

````
    dmp_info_.iamf_primary_profile = 0;
    dmp_info_.iamf_additional_profile = 0;
    SB_CHECK(IamfMimeUtil::ParseIamfSequenceHeaderObu(
        audio_access_units_[0].data().data(),
        audio_access_units_[0].data().size(),
        &(*dmp_info_.iamf_primary_profile),
        &(*dmp_info_.iamf_additional_profile)));


--->

  auto result = IamfMimeUtil::ParseIamfSequenceHeaderObu(audio_access_units_[0].data());
  SB_CHECK(result) << result.error();
  dmp_info_.iamf_primary_profile = result->primary;
  dmp_info_.iamf_additional_profile = result->additional;
```

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -44,6 +47,34 @@ bool StringToProfile(const std::string& input, uint32_t* profile) {
   *profile = converted_val;
   return true;
 }
+
+// Helper function to read a LEB128 value.
+bool ReadLeb128(const uint8_t** data, const uint8_t* end, uint32_t* value) {
```
kjyoun: Can you update this signature ?

1) Return values only through return value, not output parameter?

you can use std::optional. Then on nullopt means failure, while valid value will be set on success

Then code will be like this
```
std::optional<uint32_t> ReadLeb128(const uint8_t** data, const uint8_t* end) {
  if (fail) {
     return std::nullopt;
  }

  // on success
  return value;
}
```

2) Use data struct encapsulating data array, instead of C-style data pair

https://share.google/aimode/dDmeisd2jkg6M5D3k

With `std::basic_string_view<uint8_t>`, code would be more cleaner

```
using ByteView = std::basic_string_view<uint8_t>;

std::optional<uint32_t> ReadLeb128(ByteView& view) {
  for (size_t i = 0; i < 5; ++i) {
    if (view.empty()) {
      return std::nullopt;
    }

    uint8_t byte = view.front();
    view.remove_prefix(1);

    // R52: A 32-bit value can't use more than 4 bits from the 5th LEB128 byte.
    if (i == 4 && (byte & 0x7f) > 0x0f) {
      return std::nullopt;
    }

    ....
  }

}

Comment (osagie98) [reply starboard/shared/starboard/media/iamf_util.cc]: Good suggestion, added with some changes suggested by Gemini

File: starboard/shared/starboard/media/iamf_util.cc (hunk 6 above)
Comment (osagie98) [reply starboard/shared/starboard/media/iamf_util.cc]: Done

File: starboard/shared/starboard/media/iamf_util.cc (hunk 7 above)
Comment (osagie98) [reply starboard/shared/starboard/media/iamf_util.cc]: Done

File: starboard/shared/starboard/media/iamf_util.cc [hunk 9]
```diff
@@ -49,31 +51,28 @@ bool StringToProfile(const std::string& input, uint32_t* profile) {
 }
 
 // Helper function to read a LEB128 value.
-bool ReadLeb128(const uint8_t** data, const uint8_t* end, uint32_t* value) {
-  SB_CHECK(data);
-  SB_CHECK(*data);
-  SB_CHECK(end);
-  SB_CHECK(value);
-
+std::optional<uint32_t> ReadLeb128(std::string_view& view) {
```
kjyoun: why & ? 

you should not need this.

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -154,47 +153,50 @@ IamfMimeUtil::IamfMimeUtil(const std::string& mime_type) {
...
-  if (!ReadLeb128(&data, end, &obu_size)) {
-    SB_LOG(ERROR) << "Failed to parse OBU size.";
-    return false;
+  std::string_view view(reinterpret_cast<const char*>(read_head),
+                        end - read_head);
+  const size_t size_before_leb128 = view.size();
+  std::optional<uint32_t> obu_size = ReadLeb128(view);
+
+  if (!obu_size.has_value()) {
+    return Failure(FormatString("Failed to parse OBU size."));
```
kjyoun: You don't need FormasString here

This would be enough

return Failure("Failed to parse OBU size.");

starboard/shared/starboard/media/iamf_util.cc
kjyoun: I see we need string_vew&.

Then I would suggest rename this method to something other than mere "Read", which suggests it read-only, while this method updating param string_view

ParseLeb128 ?

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -154,47 +153,50 @@ IamfMimeUtil::IamfMimeUtil(const std::string& mime_type) {
...
 
 // static.
-bool IamfMimeUtil::ParseIamfSequenceHeaderObu(const uint8_t* data,
-                                              size_t size,
-                                              uint8_t* primary_profile,
-                                              uint8_t* additional_profile) {
-  const uint8_t* end = data + size;
+Result<IamfMimeUtil::IamfProfileInfo> IamfMimeUtil::ParseIamfSequenceHeaderObu(
+    const std::vector<uint8_t>& data) {
+  const uint8_t* read_head = data.data();
```
kjyoun: data type that you need is `std::basic_string_view<uint8_t>`

You may want to create alias `using ByteView = std::basic_string_view<uint8_t>`

```

// Initialize a view of the entire vector
  ByteView view(data.data(), data.size());

  if (view.empty()) {
    return Failure("Empty OBU data.");
  }

  // Read first byte
  const uint8_t header_byte = view.front();
  uint8_t obu_type = (header_byte >> 3) & 0x1f;
  
  if (obu_type != kIamfSequenceHeaderObu) {
     //
  }

  // Then advance view
  view.remove_prefix(1); // Advance past header
```

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -154,47 +153,50 @@ IamfMimeUtil::IamfMimeUtil(const std::string& mime_type) {
...
+  const uint8_t* end = read_head + data.size();
 
-  const uint8_t header_byte = *data++;
+  const uint8_t header_byte = *read_head++;
```
kjyoun: The one of primary goals of using this "view" struct is to avoid pointer arithmetic.

You shouldn't need this.

kjyoun (commented):
Sorry for the multiple requests, but I believe we can make this logic significantly cleaner and more robust.

This time I wrote a sample code snippet to show the pattern I'm looking for:
https://paste.googleplex.com/5521076055572480

- Return-by-value, instead of modifying parameters or using out-params.
- Eliminate manual pointer arithmetic: Leverage std::basic_string_view<uint8_t> (or a ByteView alias) and its built-in methods like .remove_prefix() and .substr() to handle the buffer bounds safely

starboard/shared/starboard/media/iamf_util.cc
osagie98: Replaced with a BufferReader class

starboard/shared/starboard/media/iamf_util.cc
osagie98: Done

starboard/shared/starboard/media/iamf_util.cc
osagie98: Replaced with a BufferReader class

starboard/shared/starboard/player/video_dmp_reader.cc
```diff
@@ -157,7 +159,14 @@ std::string VideoDmpReader::audio_mime_type() const {
...
     case kSbMediaAudioCodecIamf:
-      ss << "audio/mp4; codecs=\"iamf\";";
+      SB_CHECK(dmp_info_.iamf_primary_profile.has_value());
+      // Only Opus IAMF substreams are currently supported.
+      ss << "audio/mp4; codecs=\"iamf.";
+      ss << std::setw(3) << std::setfill('0') << std::hex
+         << static_cast<int>(*dmp_info_.iamf_primary_profile) << "."
+         << std::setw(3) << std::setfill('0') << std::hex
+         << static_cast<int>(dmp_info_.iamf_additional_profile.value_or(0))
+         << ".Opus\";";
```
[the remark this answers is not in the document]
osagie98: Not useful, the spec says 3

starboard/shared/starboard/media/BUILD.gn
```diff
@@ -38,6 +38,7 @@ static_library("media_util") {
     "//starboard/shared/starboard/media/vp9_util.h",
   ]
 
+  deps = [ "//third_party/abseil-cpp:absl" ]
```
osagie98: Not sure if all of our platforms build with c++ 20 yet, otherwise I would use `<span>` instead.

starboard/shared/starboard/media/iamf_util.cc
```diff
@@ -16,13 +16,73 @@
 
 #include <cctype>
 #include <cstdlib>
+#include <optional>
 #include <string>
 #include <vector>
 
 #include "starboard/common/string.h"
+#include "third_party/abseil-cpp/absl/types/span.h"
```
kjyoun: I don't think we can use the absl library on Starboard; I cannot find any precedents.

https://source.corp.google.com/search?q=%22third_party%2Fabseil-cpp%2Fabsl%2Ftypes%2Fspan.h%22%20-f:third_party%20-f:chromeos&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

osagie98: If that's the case, I'll use string_view. However, `std::basic_string_view<uint8_t>` causes a build error:
```
../../third_party/libc++/src/include/string_view:310:42: error: implicit instantiation of undefined template 'std::char_traits<const unsigned char>'
  310 |   static_assert(is_same<_CharT, typename traits_type::char_type>::value,
      |                                          ^
../../starboard/shared/starboard/media/iamf_util.cc:83:41: note: in instantiation of template class 'std::basic_string_view<const unsigned char>' requested here
   83 |   std::basic_string_view<const uint8_t> view_;
      |                                         ^
../../third_party/libc++/src/include/__string/char_traits.h:46:8: note: template is declared here
   46 | struct char_traits;
```
Gemini says that `std::string_view` cannot support unsigned types. I will go back to C style buffer reads for now, since I've abstracted the operations it's easy to work with. We should update to `std::span` once it's supported

kjyoun: Hold a sec. Definitely, we can do it w/o C-style buffer read.

Can't we work together tomorrow, when we are in the office?

osagie98: Sure. Another option is to use `string::view` with signed types and cast to unsigned.

PR #9126, opened 2026-02-18
Include user agent in the update check

yuying-y:
Issue: 431862767


yuying-y:
> Piping looks good, is the user agent actually used yet or will we add that later?

It's used when we need to target specific devices by specific properties within the ua on the update server.

PR #9134, opened 2026-02-18
cobalt: Cache GetExperimentConfigType result

yjzhang111:
Cache the result of GetExperimentConfigType to avoid duplicate UMA enumerations. This ensures that UMA metrics related to experiment config type are only logged once, preventing redundant calculations and maintaining data integrity.

Bug: 481473168


Awallky (approved):
LGTM
I think some of the gemini comments seem reasonable, though refactoring a few duplicate lines of test setup seems unnecessary. PTAL at those findings before landing.

yjzhang111:
> LGTM I think some of the gemini comments seem reasonable, though refactoring a few duplicate lines of test setup seems unnecessary. PTAL at those findings before landing.

I looked into that and ik chromium guideline is use std whenever possible (?), do you know if that's the case for Cobalt as well? Do we prefer base/absl?

cobalt/browser/experiments/experiment_config_manager_unittest.cc
```diff
@@ -616,20 +641,44 @@ TEST_F(ExperimentConfigManagerTest,
...
   EXPECT_EQ(experiment_config_manager_->GetExperimentConfigType(),
             ExperimentConfigType::kRegularConfig);
+}
+
+TEST_F(ExperimentConfigManagerTest,
+       GetExperimentConfigTypeUsesServerConfiguredThresholds_Safe) {
+  base::Value::Dict finch_params;
+  finch_params.Set(kCrashStreakSafeConfigThreshold, 5);
+  finch_params.Set(kCrashStreakEmptyConfigThreshold, 10);
+  pref_service_->SetDict(kFinchParameters, std::move(finch_params));
```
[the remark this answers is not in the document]
yjzhang111: Not worth saving 3 lines for tests.

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -127,6 +127,10 @@ ExperimentConfigManager::ExperimentConfigManager(
 }
 
 ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
```
vincentmo: Can this be called from multiple threads in practice? If so, there are potential race conditions w/ accessing member variables.

cobalt/browser/experiments/experiment_config_manager.h
```diff
@@ -67,6 +67,9 @@ class ExperimentConfigManager {
   // PrefService for metrics local state.
   const raw_ptr<PrefService> metrics_local_state_;
 
+  bool has_cached_config_type_ = false;
```
vincentmo: Use an optional for cached_config_type_ instead, then you don't need 2 separate variables. That or add an UNKNOWN to the enum that this defaults to.

PR #9133, opened 2026-02-18
media: Implement incremental parse look-ahead for WebM and MP4

xiaomings:
Introduce incremental parse look-ahead optimization for WebM and MP4
stream parsers. This feature adds a global toggle and an outer retry
loop to authorize additional data chunks when no parsing progress is
made but more data is available.

This improves efficiency when processing large media elements that
exceed initial inspection limits, reducing redundant
return-and-re-entry cycles. It also fixes a potential pointer stalling
bug in WebM parsing.

This change is guarded by BUILDFLAG(USE_STARBOARD_MEDIA).

Bug: 484617192

PR #9146, opened 2026-02-18
media: Connect InPlaceReuseAllocatorBase to H5VCC

johnedocampo:
Expose the new in-place reuse allocator base strategy for decoder buffers
through an H5VCC setting. This allows applications to dynamically enable
the optimized buffer allocation strategy, potentially improving memory
utilization and performance by handling metadata in-place.

We introduce a new variable, `in_place_reuse_allocator_base_strategy_state_` 
for DecoderBufferAllocator to keep track on the current status of the strategy.
We also ensure that we can safely switch to this strategy if another strategy is
also in place. 

Bug: 470186535

### DIFF
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -157,6 +157,8 @@ void DecoderBufferAllocator::Free(DemuxerStream::Type type,
   bool should_reset_strategy =
       media_buffer_pool_strategy_state_ ==
           MediaBufferPoolStrategyState::kPendingEnabling ||
+      in_place_reuse_allocator_base_strategy_state_ ==
+          InPlaceReuseAllocatorBaseStrategyState::kPendingEnabling ||
       is_memory_pool_allocated_on_demand_;
 
   if (should_reset_strategy && strategy_->GetAllocated() == 0) {
@@ -264,6 +266,31 @@ void DecoderBufferAllocator::EnableMediaBufferPoolStrategy() {
   media_buffer_pool_strategy_state_ = MediaBufferPoolStrategyState::kEnabled;
 }
 
+void DecoderBufferAllocator::EnableInPlaceReuseAllocatorBase() {
+  base::AutoLock scoped_lock(mutex_);
+
+  if (in_place_reuse_allocator_base_strategy_state_ !=
+      InPlaceReuseAllocatorBaseStrategyState::kDisabled) {
+    return;
+  }
+
+  // There is another strategy being used, we have to wait until all
+  // allocations are freed before switching to InPlaceReuseAllocatorBase based
+  // strategy.
+  if (strategy_ && strategy_->GetAllocated() > 0) {
+    in_place_reuse_allocator_base_strategy_state_ =
+        InPlaceReuseAllocatorBaseStrategyState::kPendingEnabling;
+    return;
+  }
+
+  if (strategy_) {
+    strategy_.reset();
+  }
+
+  in_place_reuse_allocator_base_strategy_state_ =
+      InPlaceReuseAllocatorBaseStrategyState::kEnabled;
+}
+
 void DecoderBufferAllocator::EnsureStrategyIsCreated() {
   mutex_.AssertAcquired();
   if (strategy_) {
@@ -273,6 +300,12 @@ void DecoderBufferAllocator::EnsureStrategyIsCreated() {
   if (media_buffer_pool_strategy_state_ ==
       MediaBufferPoolStrategyState::kPendingEnabling) {
     media_buffer_pool_strategy_state_ = MediaBufferPoolStrategyState::kEnabled;
+  }
+
+  if (in_place_reuse_allocator_base_strategy_state_ ==
+      InPlaceReuseAllocatorBaseStrategyState::kPendingEnabling) {
+    in_place_reuse_allocator_base_strategy_state_ =
+        InPlaceReuseAllocatorBaseStrategyState::kEnabled;
   }
 
   if (media_buffer_pool_strategy_state_ ==
@@ -290,7 +323,9 @@ void DecoderBufferAllocator::EnsureStrategyIsCreated() {
   }
 
   if (base::FeatureList::IsEnabled(
-          kCobaltDecoderBufferAllocatorWithInPlaceMetadata)) {
+          kCobaltDecoderBufferAllocatorWithInPlaceMetadata) ||
+      in_place_reuse_allocator_base_strategy_state_ ==
+          InPlaceReuseAllocatorBaseStrategyState::kEnabled) {
     strategy_.reset(new BidirectionalFitDecoderBufferAllocatorStrategy<
                     starboard::common::InPlaceReuseAllocatorBase>(
         initial_capacity_, allocation_unit_));
--- a/media/starboard/decoder_buffer_allocator.h
+++ b/media/starboard/decoder_buffer_allocator.h
@@ -77,9 +77,16 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
 
   void SetAllocateOnDemand(bool enabled) override;
   void EnableMediaBufferPoolStrategy() override;
+  void EnableInPlaceReuseAllocatorBase() override;
 
  private:
   enum class MediaBufferPoolStrategyState {
+    kDisabled,
+    kEnabled,
+    kPendingEnabling,
+  };
+
+  enum class InPlaceReuseAllocatorBaseStrategyState {
     kDisabled,
     kEnabled,
     kPendingEnabling,
@@ -99,6 +106,9 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
   std::unique_ptr<Strategy> strategy_ GUARDED_BY(mutex_);
   MediaBufferPoolStrategyState media_buffer_pool_strategy_state_
       GUARDED_BY(mutex_) = MediaBufferPoolStrategyState::kDisabled;
+  InPlaceReuseAllocatorBaseStrategyState
+      in_place_reuse_allocator_base_strategy_state_ GUARDED_BY(mutex_) =
+          InPlaceReuseAllocatorBaseStrategyState::kDisabled;
 
 #if !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
   // The following variables are used for comprehensive logging of allocation
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
@@ -15,6 +15,7 @@
 #include "third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.h"
 
 #include "base/functional/callback.h"
+#include "base/no_destructor.h"
 #include "cobalt/browser/h5vcc_settings/public/mojom/h5vcc_settings.mojom-blink.h"
 #include "media/base/decoder_buffer.h"
 #include "media/base/stream_parser.h"
@@ -29,6 +30,7 @@
 #include "third_party/blink/renderer/platform/heap/garbage_collected.h"
 #include "third_party/blink/renderer/platform/heap/persistent.h"
 #include "third_party/blink/renderer/platform/wtf/functional.h"
+#include "third_party/blink/renderer/platform/wtf/hash_map.h"
 #include "third_party/blink/renderer/platform/wtf/text/wtf_string.h"
 
 namespace blink {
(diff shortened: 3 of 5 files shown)


media/starboard/decoder_buffer_allocator.cc
```diff
@@ -264,6 +266,28 @@ void DecoderBufferAllocator::EnableMediaBufferPoolStrategy() {
   media_buffer_pool_strategy_state_ = MediaBufferPoolStrategyState::kEnabled;
 }
 
+void DecoderBufferAllocator::EnableAllocateWithInPlaceMetaData() {
+  base::AutoLock scoped_lock(mutex_);
+
+  if (in_place_reuse_allocator_base_strategy_state_ !=
+      AllocateWithInPlaceMetaDataStrategyState::kDisabled) {
+    return;
+  }
```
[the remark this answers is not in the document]
johnedocampo: The current set-up will ensure that this will not happen.

media/starboard/decoder_buffer_allocator.h
```diff
@@ -99,6 +106,9 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
   std::unique_ptr<Strategy> strategy_ GUARDED_BY(mutex_);
   MediaBufferPoolStrategyState media_buffer_pool_strategy_state_
       GUARDED_BY(mutex_) = MediaBufferPoolStrategyState::kDisabled;
+  InPlaceReuseAllocatorBaseStrategyState
+      in_place_reuse_allocator_base_strategy_state_ GUARDED_BY(mutex_) =
+          InPlaceReuseAllocatorBaseStrategyState::kDisabled;
```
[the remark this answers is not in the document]
johnedocampo: This adjustment was automatically made by the pre-commit checks. We will keep this as-is.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -264,6 +266,31 @@ void DecoderBufferAllocator::EnableMediaBufferPoolStrategy() {
...
+
+  if (strategy_) {
+    strategy_.reset();
+  }
+
+  in_place_reuse_allocator_base_strategy_state_ =
+      InPlaceReuseAllocatorBaseStrategyState::kEnabled;
+}
```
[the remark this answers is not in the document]
johnedocampo: I think we can keep the change as-is here, as we don't expect both to be enabled at the same time.

xiaomings (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -74,6 +74,22 @@ ScriptPromise ProcessDecoderBufferSettings(ScriptState* script_state,
...
+      return promise;
+    }
+
+    LOG(WARNING) << name << " cannot be disabled.";
+    resolver->Reject(V8ThrowException::CreateTypeError(
+        script_state->GetIsolate(), name + " cannot be disabled."));
+
+    return promise;
+  }
+
```
kjyoun, line 107: This code block is repeatiton of the block above.

Can you refactor to share codes?

This is what gemini suggest.
````
using EnableFunc = std::function<void()>;

static const std::map<String, EnableFunc> kFlagMap{
    {"DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy",
     []() { ::media::DecoderBuffer::EnableMediaBufferPoolStrategy(); }},
    {"DecoderBuffer.EnableInPlaceReuseAllocatorBase",
     []() { ::media::DecoderBuffer::EnableInPlaceReuseAllocatorBase(); }},
};

auto it = kFlagMap.find(name);
if (it != kFlagMap.end()) {
  if (value->GetAsLong() != 0) {
    LOG(INFO) << "Enabling " << name << ".";
    it->second(); // Executes the std::function (lambda)
    resolver->Resolve();
  } else {
    LOG(WARNING) << name << " cannot be disabled.";
    resolver->Reject(V8ThrowException::CreateTypeError(
        script_state->GetIsolate(), name + " cannot be disabled."));
  }
  return promise;
}
```

johnedocampo: Done; refactored the code to reduce repetition. Added a static settings map that stores all the settings and simplified the parsing function to use this map.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -37,6 +39,19 @@ constexpr char kMediaAppendFirstSegmentSynchronously[] =
     "Media.AppendFirstSegmentSynchronously";
 constexpr char kDecoderBufferSettingPrefix[] = "DecoderBuffer.";
 
+using EnableFunction = void (*)();
+using SettingsMap = WTF::HashMap<WTF::String, EnableFunction>;
+
+const SettingsMap& GetDecoderBufferSettings() {
+  static const base::NoDestructor<SettingsMap> settings({
+      {"DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy",
```
kjyoun: can you define these hard-coded string as constant as above?

Then we can put all feature flags in the single places like [code](https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/renderer/cobalt_content_renderer_client.cc;l=39-52;drc=c1321336bc0591744bee0785c056179160bccc91)


When addding constexpr char, please add `static_assert` to enforce flags starts with given prefix.

```
constexpr char kDecoderBufferSettingPrefix[] = "DecoderBuffer.";

constexpr char kEnableMediaBufferPoolAllocatorStrategy[] = 
    "DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy";

constexpr char kEnableInPlaceReuseAllocatorBase[] = 
    "DecoderBuffer.EnableInPlaceReuseAllocatorBase";


static_assert(
    std::string_view(kEnableMediaBufferPoolAllocatorStrategy)
            .substr(0, std::string_view(kDecoderBufferSettingPrefix).size()) == 
        kDecoderBufferSettingPrefix);

static_assert(
    std::string_view(kEnableInPlaceReuseAllocatorBase)
            .substr(0, std::string_view(kDecoderBufferSettingPrefix).size()) == 
        kDecoderBufferSettingPrefix);
````

File: third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc [hunk 6]
```diff
@@ -37,6 +39,19 @@ constexpr char kMediaAppendFirstSegmentSynchronously[] =
     "Media.AppendFirstSegmentSynchronously";
 constexpr char kDecoderBufferSettingPrefix[] = "DecoderBuffer.";
```
kjyoun, line 43: How about using `std::string_view`, instead of const char[] ?

https://share.google/aimode/cWmty5Cio0TW00prq

The change would not need complex change.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
johnedocampo: Done.

Do the static_asserts help ensure that these settings always belong to the DecoderBuffer? I want to better understand the idea behind the static_asserts here.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
johnedocampo: I tried using `std::string_view` instead of `const char[]`, but the main issue I found was that `std::string_view` does not have an implicit way to be compared with `WTF::Strings`, which are the strings that are used in the H5VCC functions. To make it work, we'd have to wrap any `std::string_view` values in `WTF::String` ctors, like

```
WTF::String my_string = "DecoderBuffer.Foo";
std::string_view prefix = "DecoderBuffer.";

if (my_string.StartsWith(WTF::String(prefix.data(), prefix.length()))) {
...
}
```


Doing this in my opinion clutters the code a lot, reducing readability. With `const char[]` strings, no string type conversion is necessary to perform any comparisons with `WTF::Strings`. I think its okay to keep them as-is. WDYT?

I also tried seeing if we could use `const WTF::Strings` instead, but this change would come at cost to runtime performance rather than compile time, so I decided against it.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -121,7 +149,7 @@ ScriptPromise H5vccSettings::set(ScriptState* script_state,
       resolver->Reject(V8ThrowException::CreateTypeError(
           script_state->GetIsolate(),
           String("The value for '") + kMediaAppendFirstSegmentSynchronously +
-                 "' must be a number."));
+              "' must be a number."));
```
johnedocampo, line 155: note: this change was automatically made by pre-commit.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
kjyoun: yes, static_assert checks the given string starts with "Decoder." prefix, which it should.

static_assert is executed at compiler (== not runtime cost). If static_assert fails, compile will fail

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
kjyoun: ack

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -36,6 +38,33 @@ namespace {
 constexpr char kMediaAppendFirstSegmentSynchronously[] =
     "Media.AppendFirstSegmentSynchronously";
 constexpr char kDecoderBufferSettingPrefix[] = "DecoderBuffer.";
+constexpr char kEnableMediaBufferPoolAllocatorStrategy[] =
+    "DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy";
+constexpr char kEnableInPlaceReuseAllocatorBase[] =
+    "DecoderBuffer.EnableInPlaceReuseAllocatorBase";
+
+static_assert(
```
kjyoun, line 47: nit: 

I think that this static_assert is placed better just below kEnableMediaBufferPoolAllocatorStrategy

constexpr char kEnableMediaBufferPoolAllocatorStrategy[]
static_assert(kEnableMediaBufferPoolAllocatorStrategy...)

constexpr char kEnableInPlaceReuseAllocatorBase[]
static_assert(kEnableInPlaceReuseAllocatorBase...)

kjyoun (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
johnedocampo: Done.

kjyoun:
https://github.com/youtube/cobalt/pull/9133 is merged

Can you update  [third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc](https://github.com/youtube/cobalt/pull/9146/files#diff-7e7ad584945906b0791187a517764efe0f1f31d5f822f5c1dd8f7334e8b23a0c) to make newly merged code to use your refactored code?

johnedocampo:
> #9133 is merged
> 
> Can you update [third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc](https://github.com/youtube/cobalt/pull/9146/files#diff-7e7ad584945906b0791187a517764efe0f1f31d5f822f5c1dd8f7334e8b23a0c) to make newly merged code to use your refactored code?

The code in #9133 adds code for a "Media" setting, not "DecoderBuffer". My current changes only apply to settings whose name starts with "DecoderBuffer". Refactoring the logic for the `Media` settings seems to be getting out of scope for the original intent of this PR. Maybe we can create another PR to refactor this as well?

kjyoun:
> > #9133 is merged
> > Can you update [third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc](https://github.com/youtube/cobalt/pull/9146/files#diff-7e7ad584945906b0791187a517764efe0f1f31d5f822f5c1dd8f7334e8b23a0c) to make newly merged code to use your refactored code?
> 
> The code in #9133 adds code for a "Media" setting, not "DecoderBuffer". My current changes only apply to settings whose name starts with "DecoderBuffer". Refactoring the logic for the `Media` settings seems to be getting out of scope for the original intent of this PR. Maybe we can create another PR to refactor this as well?

Sure, we can make a separate PR

PR #9163, opened 2026-02-19
Apply v8 fix on execution/isolate-inl.h

alexanderbobrovnik:
The original fix can be found on
https://chromium-review.googlesource.com/c/v8/v8/+/3244420

b/485739172


amurovanyi (approved):

niranjanyardi (approved):
given this is an upstream fix, looks good

PR #9164, opened 2026-02-19
Apply v8 fix on builtins/promise-any.tq

alexanderbobrovnik:
The original fix can be found on
https://chromium-review.googlesource.com/c/v8/v8/+/3990627

b/485739172

### DIFF
--- a/third_party/v8/src/builtins/promise-any.tq
+++ b/third_party/v8/src/builtins/promise-any.tq
@@ -119,7 +119,19 @@ PromiseAnyRejectElementClosure(
           kPromiseAnyRejectElementRemainingSlot);
 
   // 9. Set errors[index] to x.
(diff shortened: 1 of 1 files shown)


amurovanyi (approved):

third_party/v8/src/builtins/promise-any.tq
```diff
@@ -119,7 +119,19 @@ PromiseAnyRejectElementClosure(
           kPromiseAnyRejectElementRemainingSlot);
 
   // 9. Set errors[index] to x.
-  const newCapacity = IntPtrMax(SmiUntag(remainingElementsCount), index + 1);
```
niranjanyardi, line 122: if there'sno upstream fix , jsut gaurd it with a build flag you use for building PS5 

#if defined(PS5) or something like that - so that it doesn't apply to all platforms

niranjanyardi (approved):

PR #9168, opened 2026-02-19
android: Enable AudioTrack playback params in tunnel mode

jasonzhangxx:
This commit enables dynamic playback rate control for audio tracks
on Android when operating in tunnel mode. It integrates the Android
AudioTrack.setPlaybackParams() API to delegate playback speed
adjustments to the platform.

Previously, only playback rates of 0.0 (pause) and 1.0 (normal) were
directly supported by the audio sink. Other rates were handled by
software time stretching or explicitly not implemented at the
platform level. By utilizing setPlaybackParams() for tunnel mode
audio, Cobalt can leverage hardware acceleration or platform-optimized
handling for variable speed playback, potentially improving
performance and reducing CPU overhead. The change also ensures the
audio buffer size is sufficiently increased to accommodate faster
playback speeds.

Bug: 311422213

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
@@ -20,6 +20,7 @@ import android.media.AudioAttributes;
 import android.media.AudioFormat;
 import android.media.AudioManager;
 import android.media.AudioTrack;
+import android.media.PlaybackParams;
 import android.os.Build;
 import androidx.annotation.GuardedBy;
 import androidx.annotation.RequiresApi;
@@ -227,11 +228,34 @@ public class AudioTrackBridge {
     mAvSyncPacketBytesRemaining = 0;
   }
 
+
+  @CalledByNative
+  public boolean setPlaybackRate(float playbackRate) {
+    if (mAudioTrack == null) {
+      Log.e(TAG, "Unable to setPlaybackRate with NULL audio track.");
+      return false;
+    }
+    if (!mTunnelModeEnabled) {
+      Log.i(TAG, "Skip SetPlaybackRate for non tunnel mode tracks.");
+      return true;
+    }
+
+    try {
+      PlaybackParams params = mAudioTrack.getPlaybackParams();
+      params.setSpeed(playbackRate);
+      mAudioTrack.setPlaybackParams(params);
+    } catch (IllegalArgumentException | IllegalStateException e) {
+      Log.e(TAG, String.format("Unable to setPlaybackRate, error: %s", e.toString()));
+      return false;
+    }
+    return true;
+  }
+
   @CalledByNative
   public int setVolume(float gain) {
     if (mAudioTrack == null) {
       Log.e(TAG, "Unable to setVolume with NULL audio track.");
-      return 0;
+      return AudioTrack.ERROR_INVALID_OPERATION;
     }
     return mAudioTrack.setVolume(gain);
   }
--- a/starboard/android/shared/audio_track_audio_sink_type.cc
+++ b/starboard/android/shared/audio_track_audio_sink_type.cc
@@ -188,13 +188,19 @@ AudioTrackAudioSink::~AudioTrackAudioSink() {
 
 void AudioTrackAudioSink::SetPlaybackRate(double playback_rate) {
   SB_DCHECK_GE(playback_rate, 0.0);
-  if (playback_rate != 0.0 && playback_rate != 1.0) {
-    SB_NOTIMPLEMENTED() << "TODO: Only playback rates of 0.0 and 1.0 are "
-                           "currently supported.";
-    playback_rate = (playback_rate > 0.0) ? 1.0 : 0.0;
+  SB_DLOG(INFO) << "Set playback rate to " << playback_rate;
+
+  {
+    std::lock_guard lock(mutex_);
+    playback_rate_ = playback_rate;
   }
-  std::lock_guard lock(mutex_);
-  playback_rate_ = playback_rate;
+
+  // AudioTrack doesn't support playback speed of 0.
+  if (playback_rate > 0.0) {
+    // AudioTrackBridge.setPlaybackRate() currently is only enabled for tunnel
+    // mode. It will be no-op for non tunnel player.
+    bridge_.SetPlaybackRate(playback_rate);
+  }
 }
 
 // TODO: Break down the function into manageable pieces.
@@ -439,6 +445,21 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
     int sampling_frequency_hz) {
   SB_CHECK(audio_track_audio_sink_type_);
   JNIEnv* env = AttachCurrentThread();
+
+  const bool force_tunnel_mode =
+      features::FeatureList::IsEnabled(features::kForceTunnelMode);
+  if (force_tunnel_mode) {
+    // AudioTrack.setPlaybackParams() needs extra buffer to support playback
+    // speed greater than 1.0x.
+    const double kMaxPlaybackSpeed = 2.0;
+    return std::max<int>(
+        AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
+            env, sample_type, channels, sampling_frequency_hz),
+        audio_track_audio_sink_type_->GetMinBufferSizeInFramesInternal(
+            channels, sample_type, sampling_frequency_hz) *
+            kMaxPlaybackSpeed);
+  }
+
   return std::max(
       AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
           env, sample_type, channels, sampling_frequency_hz),
--- a/starboard/android/shared/audio_track_bridge.cc
+++ b/starboard/android/shared/audio_track_bridge.cc
@@ -19,6 +19,7 @@
 #include "starboard/android/shared/audio_output_manager.h"
 #include "starboard/android/shared/media_common.h"
 #include "starboard/audio_sink.h"
+#include "starboard/common/check_op.h"
 #include "starboard/common/log.h"
 #include "starboard/shared/starboard/media/media_util.h"
 
@@ -239,6 +240,21 @@ int AudioTrackBridge::WriteSample(const uint8_t* samples,
     return bytes_written;
   }
   return bytes_written;
+}
+
+void AudioTrackBridge::SetPlaybackRate(
+    double playback_rate,
+    JNIEnv* env /*= AttachCurrentThread()*/) {
+  SB_DCHECK(env);
+  SB_DCHECK(is_valid());
+  // AudioTrack doesn't support playback speed of 0.
+  SB_DCHECK_GT(playback_rate, 0.0);
+
+  jboolean status = Java_AudioTrackBridge_setPlaybackRate(
+      env, j_audio_track_bridge_, static_cast<float>(playback_rate));
+  if (!status) {
+    SB_LOG(ERROR) << "Failed to set playback rate to " << playback_rate;
+  }
 }
 
 void AudioTrackBridge::SetVolume(double volume,
--- a/starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
+++ b/starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
@@ -214,10 +214,13 @@ void AudioRendererPcm::SetPlaybackRate(double playback_rate) {
 
   playback_rate_ = playback_rate;
 
-  audio_renderer_sink_->SetPlaybackRate(playback_rate_ > 0.0 ? 1.0 : 0.0);
+  double adjusted_playback_rate = playback_rate_ > 0.0 ? 1.0 : 0.0;
+  if (audio_renderer_sink_->AllowDirectPlaybackRateSetting()) {
+    adjusted_playback_rate = playback_rate_;
+  }
+
+  audio_renderer_sink_->SetPlaybackRate(adjusted_playback_rate);
   if (audio_renderer_sink_->HasStarted()) {
-    // TODO: Remove SetPlaybackRate() support from audio sink as it only need to
-    // support play/pause.
     if (playback_rate_ > 0.0) {
       if (process_audio_data_job_token_.is_valid()) {
         RemoveJobByToken(process_audio_data_job_token_);
@@ -334,14 +337,12 @@ int64_t AudioRendererPcm::GetCurrentMediaTime(bool* is_playing,
     frames_played =
         audio_frame_tracker_.GetFutureFramesPlayedAdjustedToPlaybackRate(
             elapsed_frames, playback_rate);
-#if BUILDFLAG(IS_ANDROID)
     if (audio_renderer_sink_->AllowOverflowAudioSamples()) {
       // A simple workaround to handle silence frames for tunnel mode player.
       // |playback_rate| is ignored as tunnel mode doesn't support
       // vsp.
       frames_played += audio_frame_tracker_.GetOverflowedFrames();
     }
-#endif  // BUILDFLAG(IS_ANDROID)
     media_time =
         seeking_to_time_ + frames_played * 1'000'000LL / samples_per_second;
     if (media_time < last_media_time_) {
@@ -490,7 +491,6 @@ void AudioRendererPcm::UpdateVariablesOnSinkThread_Locked(
       frames_consumed_set_at_ = system_time_on_consume_frames;
     }
 
-#if BUILDFLAG(IS_ANDROID)
     if (audio_renderer_sink_->AllowOverflowAudioSamples()) {
       auto silence_frames_consumed =
           frames_consumed_on_sink_thread_ - non_silence_frames_consumed;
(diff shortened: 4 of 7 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -227,6 +228,31 @@ public void release() {
...
   }
 
+
+  @CalledByNative
+  public int setPlaybackRate(float playbackRate) {
+    if (mAudioTrack == null) {
+      Log.e(TAG, "Unable to setPlaybackRate with NULL audio track.");
+      return 0;
+    }
+    if(!mTunnelModeEnabled) {
```
xiaomings: nit: add space between if and "(".

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -227,6 +228,31 @@ public void release() {
...
+      Log.e(TAG, "Don't setPlaybackRate with non tunnel track.");
+      return 0;
+    }
+    try {
+      PlaybackParams params = mAudioTrack.getPlaybackParams();
+      params.setSpeed(playbackRate);
+      mAudioTrack.setPlaybackParams(params);
+    } catch (IllegalArgumentException e) {
```
xiaomings: `(IllegalArgumentException | IllegalStateException e)`?

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -214,10 +214,15 @@ void AudioRendererPcm::SetPlaybackRate(double playback_rate) {
 
   playback_rate_ = playback_rate;
 
-  audio_renderer_sink_->SetPlaybackRate(playback_rate_ > 0.0 ? 1.0 : 0.0);
+  double adjusted_playback_rate = playback_rate_ > 0.0 ? 1.0 : 0.0;
+#if BUILDFLAG(IS_ANDROID)
+  if (audio_renderer_sink_->AllowDirectPlaybackRateSetting()) {
+    adjusted_playback_rate = playback_rate_;
+  }
```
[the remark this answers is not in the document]
jasonzhangxx: It looks the same as existing code.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -241,6 +241,21 @@ int AudioTrackBridge::WriteSample(const uint8_t* samples,
...
+    JNIEnv* env /*= AttachCurrentThread()*/) {
+  SB_DCHECK(env);
+  SB_DCHECK(is_valid());
+  // AudioTrack doesn't support playback speed of 0.
+  SB_DCHECK(playback_rate > 0.0);
+
+  jint status = Java_AudioTrackBridge_setPlaybackRate(
+      env, j_audio_track_bridge_, static_cast<float>(playback_rate));
+  if (status != 0) {
+    SB_LOG(ERROR) << "Failed to set playback rate to " << playback_rate;
```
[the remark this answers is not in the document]
jasonzhangxx: The return value in AudioTrackBridge.java has been updated.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -227,6 +228,28 @@ public void release() {
     mAvSyncPacketBytesRemaining = 0;
   }
 
+
+  @CalledByNative
+  public int setPlaybackRate(float playbackRate) {
```
kjyoun: why not bool ?

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -241,6 +241,21 @@ int AudioTrackBridge::WriteSample(const uint8_t* samples,
   return bytes_written;
 }
 
+void AudioTrackBridge::SetPlaybackRate(
+    double playback_rate,
+    JNIEnv* env /*= AttachCurrentThread()*/) {
+  SB_DCHECK(env);
```
kjyoun, line 248: SB_CHECK ?

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -241,6 +241,21 @@ int AudioTrackBridge::WriteSample(const uint8_t* samples,
...
+  SB_DCHECK(is_valid());
+  // AudioTrack doesn't support playback speed of 0.
+  SB_DCHECK(playback_rate > 0.0);
```
kjyoun: SB_CHECK_GT(playback_rate, 0.0);


in sb/common/check_op.h

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -214,10 +214,15 @@ void AudioRendererPcm::SetPlaybackRate(double playback_rate) {
 
   playback_rate_ = playback_rate;
 
-  audio_renderer_sink_->SetPlaybackRate(playback_rate_ > 0.0 ? 1.0 : 0.0);
+  double adjusted_playback_rate = playback_rate_ > 0.0 ? 1.0 : 0.0;
+#if BUILDFLAG(IS_ANDROID)
```
kjyoun: Isn't there another way to achieve this without using w/o using `#if BUILDFLAG(IS_ANDROID)` ?

It does not look good that we keep adding platform-specif build flags to common codes.

I think that we have to have generic way to add this feature. Especially, this code is on main branch.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
jasonzhangxx: It's to keep the consistency of the `setVolume()` function below.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
jasonzhangxx: We actually don't have to include `#if BUILDFLAG(IS_ANDROID)` in `audio_renderer_internal_pcm.cc`. The code is guarded by `AllowDirectPlaybackRateSetting()`. The extra protection is to make it clear that the code is used only on Android and ensure it won't be used anywhere else, as it's still an experiment. And after experiment is done, we can easily find the related code, to either keep them in prod code or remove them.

starboard/android/shared/audio_track_bridge.cc
jasonzhangxx: The CHECK here is actually redundant, I don't want to include it in prod, but it's still good to have DCHECK here to let reader know it cannot be null.

starboard/android/shared/audio_track_bridge.cc
jasonzhangxx: Done

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
kjyoun: Then please remove `#if BUILDFLAG` and add comments instead.
Comment can achieve what you are trying to achieve with `#if`
```
This logic is only used in Android. See b/XXX for more details
```
Actually, I don't think that you need to add even comments here. 

Using #if BUILDFLAG as a 'marker' in shared code is a overkill—especially since the logic is already safely guarded by the runtime check. I filed b/486910174 to remove other `#if BUILD(ANDROID)`. Apparently, those platform-specific build tag is recent additions.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
kjyoun: Could you  update `setVolume()` use bool, instead?

`setVolume` is the only method in this class returning int, in place of bool
I think that this is a good chance to clean up the code.
Otherwise, we will have 2 methods returning int, in place the bool is good.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -227,6 +228,28 @@ public void release() {
...
+      Log.e(TAG, "SetPlaybackRate is only supported for tunnel mode tracks.");
+      return 0;
+    }
+    try {
+      PlaybackParams params = mAudioTrack.getPlaybackParams();
+      params.setSpeed(playbackRate);
+      mAudioTrack.setPlaybackParams(params);
+    } catch (IllegalArgumentException | IllegalStateException e) {
+      Log.e(TAG, String.format("Unable to setPlaybackRate, error: %s", e.toString()));
+      return 1;
```
kjyoun: why 1 ?

don't you need to return false here?

```
  try {
...
      mAudioTrack.setPlaybackParams(params);
    } catch (IllegalArgumentException | IllegalStateException e) {
...
      return false;
 }
 return true;


```

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -439,6 +445,22 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
...
+  const bool force_tunnel_mode =
+      features::FeatureList::IsEnabled(features::kForceTunnelMode);
+  if (force_tunnel_mode) {
+    // AudioTrack.setPlaybackParams() needs extra buffer to support playback
+    // speed greater than 1.0x.
+    const double kMaxPlaybackSpeed = 2.0;
+    return std::max(
+        AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
+            env, sample_type, channels, sampling_frequency_hz),
+        static_cast<int>(
```
kjyoun: do you need `static_int<int>` here?

https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/audio_track_audio_sink_type.h;l=89;drc=4384f0a435d4b15f5669755f547421943d0ffaee

kjyoun: no need to static_cast ?

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -241,6 +242,21 @@ int AudioTrackBridge::WriteSample(const uint8_t* samples,
   return bytes_written;
 }
 
+void AudioTrackBridge::SetPlaybackRate(
+    double playback_rate,
+    JNIEnv* env /*= AttachCurrentThread()*/) {
+  SB_DCHECK(env);
+  SB_DCHECK(is_valid());
```
kjyoun, line 249: nit: SB_CHECK ?

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -723,16 +728,24 @@ bool AudioRendererPcm::AppendAudioToFrameBuffer(bool* is_frame_buffer_full) {
 
   int offset_to_append = total_frames_sent_to_sink_ % max_cached_frames_;
 
+  double adjusted_playback_rate = playback_rate_;
+#if BUILDFLAG(IS_ANDROID)
```
kjyoun: Can you remove this build tag?

starboard/android/shared/audio_track_bridge.cc
jasonzhangxx: The check is redundant, and all other functions use SB_DCHECK(). Let's keep it as is.

starboard/android/shared/audio_track_bridge.cc
jasonzhangxx: After think it again, I changed it back to use SB_DCHECK_GT() to keep the consistency with other DChecks. The check is redundant and if it accidentally happens and is rejected, the AudioTrack will throw out an exception to help us catch the problem in QA build.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
jasonzhangxx: Ok, done. The build flag checks have been removed in this file.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
jasonzhangxx: Done.

starboard/android/shared/audio_track_audio_sink_type.cc
jasonzhangxx: `GetMinBufferSizeInFramesInternal()` returns int, but `kMaxPlaybackSpeed` is double. I changed the code to explicit use `std::max<int>`, so the readability should be improved a little bit.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
jasonzhangxx: Ok, I changed it to return boolean. But I didn't change `setVolume` as the `AudioTrack.setVolume()` returns an int.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
jasonzhangxx: It has been removed, it now returns boolean.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -227,11 +228,34 @@ public void release() {
     mAvSyncPacketBytesRemaining = 0;
   }
 
+
+  @CalledByNative
+  public boolean setPlaybackRate(float playbackRate) {
+    if (mAudioTrack == null) {
+      Log.e(TAG, "Unable to setPlaybackRate with NULL audio track.");
+      return true;
```
kjyoun: This log. returning `true` while logging error,  is confusing to me.

I think we need to choose either one  among belows

```
// return true, while logging info
Log.i(TAG, "setPlaybackRate ignored: No active audio track.");
return true;
```

or 

```
// return false, while logging error
Log.e(TAG, "Unable to setPlaybackRate with NULL audio track.");
return false;
```

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -227,11 +228,34 @@ public void release() {
...
+    }
+    if (!mTunnelModeEnabled) {
+      Log.e(TAG, "SetPlaybackRate is only supported for tunnel mode tracks.");
+      return true;
```
kjyoun, line 240: ditto.  returning-true-while-logging-error is confusing

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -439,6 +445,21 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
...
   JNIEnv* env = AttachCurrentThread();
+
+  const bool force_tunnel_mode =
+      features::FeatureList::IsEnabled(features::kForceTunnelMode);
+  if (force_tunnel_mode) {
+    // AudioTrack.setPlaybackParams() needs extra buffer to support playback
+    // speed greater than 1.0x.
+    const double kMaxPlaybackSpeed = 2.0;
+    return std::max<int>(
+        AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
```
kjyoun, line 456: Style is weird

?


```
return std::max<int>(
    AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
        env, sample_type, channels, sampling_frequency_hz),
    audio_track_audio_sink_type_->GetMinBufferSizeInFramesInternal(
        channels, sample_type, sampling_frequency_hz) * kMaxPlaybackSpeed);
````

File: cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java (hunk 13 above)
Comment (jasonzhangxx) [reply cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java]: It was to match the pattern of `setVolume()`.  I updated the error logs and return values of `setPlaybackRate()` and `setVolume()`.

File: starboard/android/shared/audio_track_audio_sink_type.cc (hunk 15 above)
Comment (jasonzhangxx) [reply starboard/android/shared/audio_track_audio_sink_type.cc:456]: Done.

File: cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java (hunk 14 above)
Comment (jasonzhangxx) [reply cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java:240]: Done.

File: starboard/android/shared/audio_track_audio_sink_type.cc [hunk 16]
```diff
@@ -439,6 +445,21 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
...
   SB_CHECK(audio_track_audio_sink_type_);
   JNIEnv* env = AttachCurrentThread();
+
+  const bool force_tunnel_mode =
+      features::FeatureList::IsEnabled(features::kForceTunnelMode);
+  if (force_tunnel_mode) {
+    // AudioTrack.setPlaybackParams() needs extra buffer to support playback
+    // speed greater than 1.0x.
+    const double kMaxPlaybackSpeed = 2.0;
+    return std::max<int>(
```
kjyoun, line 455: nit: I think that you don't need <int> here, since both arguments are int type

jasonzhangxx: It's needed, the playback speed is double, so after multiplying it, the result is double. And compiler would throw out an error.

PR #9169, opened 2026-02-19
evergreen: Enable music splash screen

borongc:
Introduce the kLink command-line switch to accept a deep link URL.
Parse the provided deep link to extract a "topic" query parameter.
This topic is then passed to the splash screen creation logic to allow
dynamic configuration of the splash screen, specifically enabling the
music splash screen on Evergreen devices when triggered via a deep link.
This change aligns with similar functionality implemented
(https://github.com/youtube/cobalt/pull/8990) for ATV.

Issue: 484037780

### DIFF
--- a/cobalt/shell/browser/shell_browser_main_parts.cc
+++ b/cobalt/shell/browser/shell_browser_main_parts.cc
@@ -47,6 +47,7 @@
 #include "device/bluetooth/bluetooth_adapter_factory.h"
 #include "net/base/filename_util.h"
 #include "net/base/net_module.h"
+#include "net/base/url_util.h"
 #include "net/grit/net_resources.h"
 #include "ui/base/buildflags.h"
 #include "ui/base/resource/resource_bundle.h"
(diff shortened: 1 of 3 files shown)


cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -148,12 +149,24 @@ void ShellBrowserMainParts::InitializeBrowserContexts() {
...
+#if !BUILDFLAG(IS_ANDROID)
+  // Handle deeplink from non-ATV devices.
+  std::string topic;
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+  if (command_line->HasSwitch(switches::kLink)) {
+    GURL link_url(command_line->GetSwitchValueASCII(switches::kLink));
+    if (link_url.is_valid()) {
+      net::GetValueForKeyInQuery(link_url, "topic", &topic);
+    }
+  }
```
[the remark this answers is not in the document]
borongc: Right, but the current one can keep the code more readable.

cobalt/shell/common/shell_switches.cc
```diff
@@ -44,6 +44,9 @@ const char kContentShellHideToolbar[] = "content-shell-hide-toolbar";
 // comma-separated list of origins.
 const char kIsolatedContextOrigins[] = "isolated-context-origins";
 
+// The URL passed in via deep link.
+const char kLink[] = "link";
```
andrewsavage1, line 48: This currently overlaps with kLinkSwitch in starboard/shared/starboard/application.cc

borongc: If we want to use only one definition in Cobalt, we may need to move `kLinkSwitch` to starboard headers, so we can include it in Cobalt layer.

What's the best way to do this?

andrewsavage1: If there are other examples of having duplicate flag values like this then go ahead with it. Or if there are no examples of reusing a flag in the cobalt and starboard layer, this also seems fine. Otherwise I would follow some existing example of reusing the flag between the two layers.

borongc: Checked with other team members, it seems we don't have existing examples.

andrewsavage1 (approved):

PR #9170, opened 2026-02-19
android: increase MediaDecoder poll interval for TM

jasonzhangxx:
Increase the poll interval for the MediaCodecDecoder thread when tunnel mode is enabled. When in tunnel mode, the decoder's operations are less time-critical regarding polling, making it safe to wait longer. This change moves the wait from 1ms to 10ms to potentially reduce CPU cycles. We can revisit the value after the experiment of b/478320438 is done.

Bug: 485905052
Bug: 478320438

### DIFF
--- a/starboard/android/shared/media_decoder.cc
+++ b/starboard/android/shared/media_decoder.cc
@@ -442,7 +442,12 @@ void MediaCodecDecoder::DecoderThreadFunc() {
           // because the complex conditions are already checked by the
           // surrounding loop, which will re-evaluate the state when this wait
           // returns.
-          condition_variable_.wait_for(lock, std::chrono::milliseconds(1));
+          // Tunnel mode doesn't need Tick() function. It only waits for new
+          // inputs and outputs, so it's safe to wait for longer time.
+          const auto poll_interval = tunnel_mode_enabled_
+                                         ? std::chrono::milliseconds(10)
+                                         : std::chrono::milliseconds(1);
+          condition_variable_.wait_for(lock, poll_interval);
         }
       }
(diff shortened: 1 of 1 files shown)


starboard/android/shared/media_decoder.cc
```diff
@@ -438,11 +438,17 @@ void MediaCodecDecoder::DecoderThreadFunc() {
...
+            // Wait for a signal or a timeout. We don't use a predicate here
+            // because the complex conditions are already checked by the
+            // surrounding loop, which will re-evaluate the state when this wait
+            // returns.
+            condition_variable_.wait_for(lock, std::chrono::milliseconds(1));
+          } else {
+            // Tunnel mode doesn't need Tick() function. It only waits for new
+            // inputs and outputs, so it's safe to wait for longer time.
+            condition_variable_.wait_for(lock, std::chrono::milliseconds(10));
+          }
```
kjyoun: nit: How about this?

````
 // Tunnel mode doesn't need Tick() function. It only waits for new
 // inputs and outputs, so it's safe to wait for longer time
const auto poll_interval = tunnel_mode_enabled_ ?
       std::chrono::milliseconds(10) : 
       std::chrono::milliseconds(1);
condition_variable_.wait_for(lock, poll_interval);
```

xiaomings:
> LGTM except style comment, though we may need to hear @xiaomings opinion on this
> 
> He said that he has a plan to refactor this loop

Thanks for checking.  Please feel free to make any adjustments on the class, the refactor I am working on should work side by side with any other tuning of the class.

PR #9181, opened 2026-02-19
android: Load splash screen before main app URL to fix race condition

jellefoks:
On cold starts with high system load, the main app's URL load can finish before the splash screen initializes its internal WebContents. This triggers Cobalt's logic to skip or immediately hide the splash screen.

By swapping the initialization order, we ensure the splash screen state is moved out of UNINITIALIZED before the main app navigation can possibly finish, ensuring the splash animation is correctly displayed even during laggy startups.

Bug: 485979436


jellefoks:
Not sure if the current behavior is WAI, but there you go, if not, this may be the fix.

PR #9184, opened 2026-02-20
android: Fix platform service memory management

zhongqiliang:
Remove an unnecessary 'delete name' call in the destructor, this is double deletion that cause the app to crash.
Use malloc() for Send() to match the free() call in cobalt/browser/h5vcc_platform_service/platform_service_impl.cc.

Bug: 485655598

### DIFF
--- a/starboard/android/shared/platform_service.cc
+++ b/starboard/android/shared/platform_service.cc
@@ -29,13 +29,10 @@
 typedef struct CobaltExtensionPlatformServicePrivate {
   void* context;
   ReceiveMessageCallback receive_callback;
-  const char* name;
+  std::string name;
   jobject cobalt_service;
 
   ~CobaltExtensionPlatformServicePrivate() {
-    if (name) {
-      delete name;
-    }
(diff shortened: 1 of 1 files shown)


starboard/android/shared/platform_service.cc
```diff
@@ -33,9 +33,6 @@ typedef struct CobaltExtensionPlatformServicePrivate {
   jobject cobalt_service;
 
   ~CobaltExtensionPlatformServicePrivate() {
-    if (name) {
-      delete name;
-    }
```
hlwarriner, line 38: This works, but you can also consider changing the `name` field in `CobaltExtensionPlatformServicePrivate` to be a `std::string` and copying the data from the `name` argument when the struct is initialized. Then we would get automatic memory management.

See https://github.com/youtube/cobalt/blob/main/starboard/linux/shared/platform_service.h#L77.

zhongqiliang: done

PR #9190, opened 2026-02-20
Skip StorageMigration if legacy storage deletion is failed

haozheng-cobalt:
Issue: Users are repeatedly signed out of Cobalt-based apps (like YouTube) on certain devices after every reboot.

  Root Cause: A logic error in the storage migration process. The MigrationManager reads legacy (Cobalt 25) storage
  data but fails to verify if the file is successfully deleted afterwards. On devices where deletion fails (e.g.,
  due to file system locks), the legacy file persists. This causes the migration to run on every boot, repeatedly
  overwriting the user's current, valid session (in Modern/Cobalt 26 storage) with an expired session snapshot from
  the legacy file, which the server then rejects.
  
An ill-formatted investigation md file for reviewers to have an understanding of Cobalt's StorageMigration and relevant context can be found here: https://paste.googleplex.com/5273707011506176

Bug: 483485402
486267127

### DIFF
--- a/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
+++ b/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
@@ -85,7 +85,11 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
   auto bytes = std::vector<uint8_t>(record->GetSize());
   const int read_result =
       record->Read(reinterpret_cast<char*>(bytes.data()), bytes.size());
-  record->Delete();
+  if (!record->Delete()) {
+    LOG(ERROR) << "Failed to delete legacy storage record. Aborting migration "
+                  "to prevent overwrite loop.";
+    return nullptr;
+  }
   if (read_result != bytes.size()) {
     return nullptr;
   }


kaidokert (approved):
The fix itself looks like a good incremental step - and is certainly part of the bigger problem. Lets merge

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang, line 106: We should add a log here. This means it reads the old data, but failed to create the new one.

johnxwork:
A side-effect could be that: for devices that fail to delete storage. Storage migration would do nothing for them and sign them out.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
kaidokert: In general this needs a lot better instrumentation. However - i think Hao has a good, isolated two-line fix of a clear bug, and i would go ahead and get this submitted, with instrumentation as a follow-up

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -85,7 +85,11 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
   auto bytes = std::vector<uint8_t>(record->GetSize());
   const int read_result =
       record->Read(reinterpret_cast<char*>(bytes.data()), bytes.size());
-  record->Delete();
+  if (!record->Delete()) {
```
zhongqiliang, line 88: It hit this code eventually.
https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:starboard/shared/starboard/file_storage/storage_delete_record.cc;l=34?q=SbStorageDeleteRecord&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F26.android
And this is probably the reason why it fails to delete. https://screenshot.googleplex.com/3m6tAfwPhsRbinM

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang, line 57: I wonder, do we want to edit the recordheader on read success, instead of check if delete of the record is successful. if edit is successful, we can continue the migration, that way it won't lose the user data, and next time, the migration will be stopped.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
kaidokert: Hmm. This might actually mean that Delete returns false even though the file was actually committed for deletion on disk - and would get deleted on the next reboot.

From that perspective this change makes it more risky even ?

haozheng-cobalt: @kaidokert just to answer the second point - From that perspective this change makes it more risky even ?

I don't think it will, because the storage migration ideally only runs once -  when users update from C25 to C26. If this returns false when file is actually deleted, it only means that users will not carry sign in information to C26, and have to sign in to C26 when app is updated to C26, later on, users will be using the modern C26 storage. 
In other word, it's little harm to skip C25 storage migration.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
haozheng-cobalt: Good point, will consider in follow-up improvement. I'll merge the fix first.

haozheng-cobalt:
> A side-effect could be that: for devices that fail to delete storage. Storage migration would do nothing for them and sign them out.

Please see the discussion here: https://github.com/youtube/cobalt/pull/9190#discussion_r2835536688

PR #9191, opened 2026-02-21
cobalt: Unify H5vccSettings processing

kjyoun:
This PR is to avoid repeated code by introducing helper methods

Introduce a ProcessLongAs template helper to consolidate
redundant V8 type checking and promise rejection logic for boolean
settings. This streamlines the validation and conversion of
V8UnionLongOrString values, simplifying H5vccSettings::set and
ProcessDecoderBufferSettings. It also standardizes promise handling.

Issue: 480134029

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
@@ -14,8 +14,7 @@
 
 #include "third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.h"
 
-#include "base/functional/callback.h"
-#include "base/no_destructor.h"
+#include "base/types/expected.h"
 #include "cobalt/browser/h5vcc_settings/public/mojom/h5vcc_settings.mojom-blink.h"
 #include "media/base/decoder_buffer.h"
 #include "media/base/stream_parser.h"
@@ -32,7 +31,6 @@
 #include "third_party/blink/renderer/platform/heap/garbage_collected.h"
 #include "third_party/blink/renderer/platform/heap/persistent.h"
 #include "third_party/blink/renderer/platform/wtf/functional.h"
-#include "third_party/blink/renderer/platform/wtf/hash_map.h"
 #include "third_party/blink/renderer/platform/wtf/text/wtf_string.h"
 
 namespace blink {
@@ -70,7 +68,16 @@ ScriptPromise ProcessSettingAs(const ScriptContext& context,
         "The value for '" + name + "' must be a number."));
     return resolver->Promise();
   }
-  callback(resolver, static_cast<T>(value->GetAsLong()));
+
+  base::expected<void, String> result =
+      callback(static_cast<T>(value->GetAsLong()));
+  if (result.has_value()) {
+    resolver->Resolve();
+  } else {
+    resolver->Reject(V8ThrowException::CreateTypeError(
+        context.script_state->GetIsolate(), result.error()));
+  }
+
   return resolver->Promise();
 }
 
@@ -81,75 +88,16 @@ ScriptPromise ProcessEnableOnlySetting(const ScriptContext& context,
                                        ActionCallback action) {
   return ProcessSettingAs<bool>(
       context, value, name,
-      [&context, name, action = std::move(action)](
-          ScriptPromiseResolver* resolver, bool enable) {
+      [name, action = std::move(action)](
+          bool enable) -> base::expected<void, String> {
         if (!enable) {
           LOG(WARNING) << name << " cannot be disabled.";
-          resolver->Reject(V8ThrowException::CreateTypeError(
-              context.script_state->GetIsolate(),
-              name + " cannot be disabled."));
-          return;
+          return base::unexpected(name + " cannot be disabled.");
         }
 
         LOG(INFO) << "Enabling " << name << ".";
         action();
-        resolver->Resolve();
-      });
-}
-
-constexpr char kEnableMediaBufferPoolAllocatorStrategy[] =
-    "DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy";
-static_assert(
-    std::string_view(kEnableMediaBufferPoolAllocatorStrategy)
-        .substr(0, std::string_view(kDecoderBufferSettingPrefix).size()) ==
-    kDecoderBufferSettingPrefix);
-
-using EnableFunction = void (*)();
-using SettingsMap = WTF::HashMap<WTF::String, EnableFunction>;
-
-const SettingsMap& GetDecoderBufferSettings() {
-  static const base::NoDestructor<SettingsMap> settings({
-      {kEnableDecommitableAllocatorStrategy,
-       &::media::DecoderBufferAllocator::EnableDecommitableAllocatorStrategy},
-      {kEnableInPlaceReuseAllocatorBase,
-       &::media::DecoderBufferAllocator::EnableInPlaceReuseAllocatorBase},
-      {kEnableMediaBufferPoolAllocatorStrategy,
-       &::media::DecoderBufferAllocator::EnableMediaBufferPoolStrategy},
-  });
-  return *settings;
-}
-
-// Ideally this function should be moved to decoder_buffer.h.  It's kept here as
-// H5vccSettings will soon be deprecated and it's easier to remove from here.
-ScriptPromise ProcessDecoderBufferSettings(const ScriptContext& context,
-                                           const WTF::String& name,
-                                           const V8UnionLongOrString* value) {
-  DCHECK(name.StartsWith(kDecoderBufferSettingPrefix));
-
-  return ProcessSettingAs<bool>(
-      context, value, name, [&](ScriptPromiseResolver* resolver, bool enable) {
-        const auto& settings = GetDecoderBufferSettings();
-        auto it = settings.find(name);
-
-        if (it == settings.end()) {
-          LOG(WARNING) << name << " isn't a supported setting.";
-          // An unknown setting leads to TypeError.
-          resolver->Reject(V8ThrowException::CreateTypeError(
-              context.script_state->GetIsolate(),
-              name + " isn't a supported setting."));
-          return;
-        }
-
-        if (enable) {
-          LOG(INFO) << "Enabling " << name << ".";
-          it->value();
-          resolver->Resolve();
-        } else {
-          LOG(WARNING) << name << " cannot be disabled.";
-          resolver->Reject(V8ThrowException::CreateTypeError(
-              context.script_state->GetIsolate(),
-              name + " cannot be disabled."));
-        }
+        return base::ok();
       });
 }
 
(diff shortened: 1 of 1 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -127,62 +141,50 @@ ScriptPromise H5vccSettings::set(ScriptState* script_state,
...
+            LOG(INFO) << "Enable incremental parse look ahead.";
+            ::media::StreamParser::SetEnableIncrementalParseLookAhead(true);
+            resolver->Resolve();
+          } else {
+            LOG(WARNING) << kMediaIncrementalParseLookAhead
+                         << " cannot be disabled.";
+            resolver->Reject(V8ThrowException::CreateTypeError(
+                script_state->GetIsolate(),
+                kMediaIncrementalParseLookAhead +
+                    String(" cannot be disabled.")));
```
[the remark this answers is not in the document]
kjyoun: Done

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -127,62 +141,50 @@ ScriptPromise H5vccSettings::set(ScriptState* script_state,
                                  const WTF::String& name,
                                  const V8UnionLongOrString* value,
                                  ExceptionState& exception_state) {
+  ScriptContext context{script_state, exception_state.GetContext()};
```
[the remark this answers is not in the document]
kjyoun: ack

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -71,46 +95,36 @@ const SettingsMap& GetDecoderBufferSettings() {
 
 // Ideally this function should be moved to decoder_buffer.h.  It's kept here as
 // H5vccSettings will soon be deprecated and it's easier to remove from here.
-ScriptPromise ProcessDecoderBufferSettings(ScriptState* script_state,
+ScriptPromise ProcessDecoderBufferSettings(const ScriptContext& context,
                                            const WTF::String& name,
-                                           const V8UnionLongOrString* value,
-                                           ExceptionState& exception_state) {
+                                           const V8UnionLongOrString* value) {
```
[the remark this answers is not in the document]
kjyoun: ack

johnxwork:
If current use cases are all T==bool, shall we remove the template code and just add a function for boolean value?

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -71,46 +95,37 @@ const SettingsMap& GetDecoderBufferSettings() {
 
 // Ideally this function should be moved to decoder_buffer.h.  It's kept here as
 // H5vccSettings will soon be deprecated and it's easier to remove from here.
-ScriptPromise ProcessDecoderBufferSettings(ScriptState* script_state,
+ScriptPromise ProcessDecoderBufferSettings(const ScriptContext& context,
```
xiaomings: The current implementation uses runtime created map and nested functions to process the settings, which can be quite hard to understand.

Consider using a macro like:
MAYBE_PROCESS_BOOL_SETTING(context..., name, function);

And process the settings line by line in `H5vccSettings::set()`.  It could be easier to understand and maintain.  The macro could internally use a helper function as an implementation detail.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -129,85 +120,52 @@ ScriptPromise H5vccSettings::set(ScriptState* script_state,
...
-          script_state->GetIsolate(), String("The value for '") +
-                                          kMediaIncrementalParseLookAhead +
-                                          "' must be a number."));
-    }
-    return promise;
+  if (name_view == kMediaIncrementalParseLookAhead) {
+    return ProcessEnableOnlySetting(context, [] {
+      ::media::StreamParser::SetEnableIncrementalParseLookAhead(true);
+    });
   }
```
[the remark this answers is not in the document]
kjyoun: This is intended.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -219,15 +177,15 @@ ScriptPromise H5vccSettings::set(ScriptState* script_state,
     NOTREACHED();
     resolver->Reject(V8ThrowException::CreateTypeError(
         script_state->GetIsolate(), "Unsupported type."));
-    return promise;
+    return resolver->Promise();
```
xiaomings: Per http://shortn/_spYWHU9HSq:
```
// Note that an empty ScriptPromise will be returned after resolve or
// reject is called.
```

is this intended?

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -129,85 +120,52 @@ ScriptPromise H5vccSettings::set(ScriptState* script_state,
...
-    return ProcessDecoderBufferSettings(script_state, name, value,
-                                        exception_state);
+  SettingContext context{script_state, exception_state.GetContext(), value,
+                         name};
+
+  // Use StringView to enable template-based comparison with string literals.
+  // This allows the compiler to use the literal's size at compile-time (via
+  // sizeof), avoiding runtime length calculations (strlen) during equality
+  // checks.
+  WTF::StringView name_view(name);
```
xiaomings: For code that only used a few times to set experiment flags, it's probably better to keep it easy to read than optimizing for performance.

kjyoun: Done

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
kjyoun: Thanks for pointing this out.

Fixed.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
kjyoun: Removed map and refactored the code

kjyoun:
> If current use cases are all T==bool, shall we remove the template code and just add a function for boolean value?

Now we have int type settings

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
xiaomings: Just wanted to double check, would it be possible for us to use a one line macro like:

`MAYBE_PROCESS_BOOL_SETTING(context..., name, function);`

or even

`MAYBE_PROCESS_BOOL_SETTING(name, function);`

to add a new setting?

So `H5vccSettings::set()` will be:

```
MAYBE_PROCESS_BOOL_SETTING("DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy",
                           ::media::DecoderBuffer::EnableMediaBufferPoolStrategy);
MAYBE_PROCESS_BOOL_SETTING("DecoderBuffer.EnableInPlaceReuseAllocatorBase",
                           ::media::DecoderBuffer::EnableInPlaceReuseAllocatorBase);
...
```

kjyoun: I want to avoid using macros. 

I think that current code is concise enough.

WDYT?


```
  if (name == kDecoderBufferEnableMediaBufferPoolAllocatorStrategy) {
    return ProcessEnableOnlySetting(context, [] {
      ::media::DecoderBuffer::EnableMediaBufferPoolStrategy();
    });
  }
  if (name == kDecoderBufferEnableInPlaceReuseAllocatorBase) {
    return ProcessEnableOnlySetting(context, [] {
      ::media::DecoderBuffer::EnableInPlaceReuseAllocatorBase();
    });
  }
```

xiaomings: The above looks good, if we use string literals instead of constants.  Assuming we are adding a new setting "DecoderBuffer.EnableSomething", it would be great to just add a block like:
```
    if (name == "DecoderBuffer.EnableSomething") {
    return ProcessEnableOnlySetting(context, [] {
      ::media::DecoderBuffer::EnableMediaBufferPoolStrategy();
    });
  }
```

It's a bit tedious if we have to add a constant, then a static_assert to ensure it starts with "DecoderBuffer", then the block above (with the constant name).

kjyoun: Updated code.

Now looks like this

```

  if (name == "DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy") {
    return ProcessSettingAsEnableOnly(context, [] {
      ::media::DecoderBufferAllocator::EnableMediaBufferPoolStrategy();
    });
  }
  if (name == "DecoderBuffer.EnableInPlaceReuseAllocatorBase") {
    return ProcessSettingAsEnableOnly(context, [] {
      ::media::DecoderBufferAllocator::EnableInPlaceReuseAllocatorBase();
    });
  }
  // "DecoderBuffer." settings must be handled before this catch-all block.
  if (name.StartsWith("DecoderBuffer.")) {
    return Reject(context, name + " isn't a supported setting.");
  }

  if (name == "Media.AppendFirstSegmentSynchronously") {
    return ProcessSettingAs<bool>(context, [this](bool enable) -> Result {
      append_first_segment_synchronously_ = enable;
      return base::ok();
    });
  }
  if (name == "Media.ExperimentalMaxPendingBytesPerParse") {
    return ProcessSettingAsPositiveInt(context, [](int value) {
      ::media::SourceBufferState::SetMaxPendingBytesPerParseOverride(value);
    });
  }
  if (name == "Media.IncrementalParseLookAhead") {
    return ProcessSettingAsEnableOnly(context, [] {
      ::media::StreamParser::SetEnableIncrementalParseLookAhead(true);
    });
  }
```

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -31,98 +31,86 @@
...
+        return base::ok();
+      });
+}
+
+template <typename ActionCallback>
+ScriptPromise ProcessSettingAsPositiveInt(const SettingContext& context,
+                                          ActionCallback action) {
+  return ProcessSettingAs<int>(
+      context,
+      [name = context.name, action = std::move(action)](int value) -> Result {
```
xiaomings, line 106: Just wanted to double check, regarding `action = std::move(action)`, can we capture `action` directly without using an explicitly variable with the same name?

kjyoun: Technically, we could use &action here since the action is executed synchronously. 

But I prefer moving it using `std::move`, since it is more clearly showing that now the ownership of callback is moved into `ProcessSettingAs` and we don't care whether it handles callback sync or async.
(though it makes code more verbose)

xiaomings (approved):

johnxwork (approved):

PR #9196, opened 2026-02-23
tvOS: Revert most accessibility changes from #8611

rakuco:
It turns out Kabuki implements screen reader support by using the Web Speech Synthesis API rather than relying on a real AX tree and Chromium's accessibility infrastructure. In fact, enabling this while also implementing a tvOS backend for Speech Synthesis causes things to be pronounced twice (in addition to the fact that the Chromium AX tree implementation is buggy and very basic).

Bug: 447135715


rakuco:

haozheng-cobalt:
Yes, for Android TV, we rely on the AX tree  mainly to aid navigation, not for speech synthesis. Are you seeing navigation issues on Apple TV atm?

rakuco:
> Are you seeing navigation issues on Apple TV atm?

Yes, the iOS AX tree backend in Chromium is very primitive and limited at the moment. Furthermore, the way that Kabuki does a11y is also buggy and not spec-compliant (see bug 479112426 and bug 481964737, for example).

PR #9203, opened 2026-02-23
cobalt: Disable resource load throttling

zhongqiliang:
Cobalt operates as a single-tab application, which eliminates the
need for resource load throttling. In multi-tab browser environments,
throttling can manage resources or prioritize active tabs, but in
Cobalt, it provides no benefit and may hinder performance.

This change configures ResourceLoadScheduler::IsClientDelayable to
always return false for Cobalt builds. This ensures that all resource
requests are treated as non-delayable and processed without artificial
delays.

Bug: 484394054

### DIFF
--- a/third_party/blink/renderer/platform/loader/BUILD.gn
+++ b/third_party/blink/renderer/platform/loader/BUILD.gn
@@ -220,7 +220,8 @@ source_set("unit_tests") {
     "fetch/raw_resource_test.cc",
     "fetch/resource_fetcher_properties_test.cc",
     "fetch/resource_fetcher_test.cc",
-    "fetch/resource_load_scheduler_test.cc",
+    # resource_load_schedule is disabled, see b/484394054
+    # "fetch/resource_load_scheduler_test.cc",
     "fetch/resource_loader_defer_loading_test.cc",
     "fetch/resource_loader_test.cc",
     "fetch/resource_request_test.cc",


johnxwork (changes requested):
Actually, looking more at the implementation of `SchedulingLifecycleState`, I'm a bit worried that the root cause might be deeper? Why would a frame be considered "throttled" or "background"?

ckim1346:
> Actually, looking more at the implementation of `SchedulingLifecycleState`, I'm a bit worried that the root cause might be deeper? Why would a frame be considered "throttled" or "background"?

From the logs in the bugreport, it looked like the throttling was occurring while the app was actually in the background but the reason the app got backgrounded was somewhat unclear -- the android system started backgrounding several apps (youtube, hulu, etc. possibly due to memory pressure?) before Kimono was resumed and the background resource load throttling log appeared.

zhongqiliang:
> Actually, looking more at the implementation of `SchedulingLifecycleState`, I'm a bit worried that the root cause might be deeper? Why would a frame be considered "throttled" or "background"?

It’s possible the app is hitting a "hidden" state due to the native splash screen. However, even in the background, Kabuki requires unthrottled JS execution. Since this is a regression from C25, I’d suggest we land this change now to unblock the app and track the investigation into the state transitions as a separate task.

PR #8485, opened 2025-12-19
android: Enable Perfetto system tracing by default for CoAT

kjyoun:
CoAT (Cobalt on Android TV) builds are used primarily for development and testing. Enabling Perfetto system tracing by default simplifies performance profiling across all build types (debug and release).

Bug: 470076078

PR #9204, opened 2026-02-23
cobalt: Enable Perfetto system tracing on Android TV

xiaomings:
Inject the EnablePerfettoSystemTracing feature flag into the command line
during BasicStartupComplete for non-release Android TV builds. This allows
system tracing to be enabled early and safely in the application
lifecycle without impacting production environments.

Bug: 428025465

### DIFF
--- a/cobalt/app/cobalt_main_delegate.cc
+++ b/cobalt/app/cobalt_main_delegate.cc
@@ -14,8 +14,8 @@
 
 #include "cobalt/app/cobalt_main_delegate.h"
 
-#include "base/command_line.h"
 #include "base/base_switches.h"
+#include "base/command_line.h"
 #include "base/process/current_process.h"
 #include "base/threading/hang_watcher.h"
 #include "base/trace_event/trace_log.h"
--- a/cobalt/app/cobalt_main_delegate.cc
+++ b/cobalt/app/cobalt_main_delegate.cc
@@ -14,6 +14,8 @@
 
 #include "cobalt/app/cobalt_main_delegate.h"
 
+#include "base/command_line.h"
+#include "base/base_switches.h"
 #include "base/process/current_process.h"
 #include "base/threading/hang_watcher.h"
 #include "base/trace_event/trace_log.h"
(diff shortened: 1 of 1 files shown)


cobalt/app/cobalt_main_delegate.cc
```diff
@@ -54,6 +56,19 @@ absl::optional<int> CobaltMainDelegate::BasicStartupComplete() {
   cl->AppendSwitch(switches::kEnableAggressiveDOMStorageFlushing);
   cl->AppendSwitch(switches::kDisableGpuShaderDiskCache);
   cl->AppendSwitch("cobalt-custom-should-disable-http-caching");
+
+#if BUILDFLAG(IS_ANDROIDTV) && !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
```
johnxwork, line 60: COBALT_IS_RELEASE_BUILD is only true in gold build. That means Perfetto tracing will be enabled in Dogfood build and qa build. Is the performance overhead small enough to avoid user journey differences?

johnxwork (approved):

kjyoun:
Does this work ?

What I found is that perfetto is not working[1] on 26.android, while it works fine in main

---
[1] Perfetto trace is itself is working, but media marker is not visible.

kjyoun:
FYI, the reason that I did not CP https://github.com/youtube/cobalt/pull/8485/ to 26.android is that perfetto still did not work properly even with it.

cobalt/app/cobalt_main_delegate.cc
xiaomings: Thanks for bringing this up.  In that case I will just keep sending it via command line and close this PR.

xiaomings:
> Does this work ?
> 
> What I found is that perfetto is not working[1] on 26.android, while it works fine in main
> 
> [1] Perfetto trace is itself is working, but media marker is not visible.

I found it works on 26.android with `--enable-features=EnablePerfettoSystemTracing`.  I am going to close this PR as we can always manually use the command line parameter.

@sideb0ard FYI

kjyoun:
@xiaomings 

Can you see media markers?
e.g.) https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:media/starboard/sbplayer_bridge.cc;l=752;drc=229336364e0810d97b08b8b9df29dbdac6fcc1f2

I couldn't see them
Here is what I saw: https://screenshot.googleplex.com/6MNzt9aYNHSTteX

PR #9217, opened 2026-02-24
BACKPORT: iOS Blink: Stop changing AVAudioSession buffer/sample rate/output channel defaults

rakuco:
This fixes an issue with speech synthesis that was causing the accessibility utterances from Kabuki to be cut or garbled in the beginning when no speech is being produced at the same time.

Bug: 447135715

---

Original CL description:

The main goal here is to fix issues with speech synthesis (both with AVSpeechSynthesizer and announcements made via
UIAccessibilityPostNotification()): when no audio is being played, a lot of times the beginning of the speech will either be mangled or cut off completely when the preferred sample rate is changed from the default 44.1kHz to 48kHz.

While here, stop changing the preferred number of output channels and the buffer size by default too. All those values come from WebRTC, but do not seem to be required or really apply to general purpose web content.

The behavior of switching the number of input channels to 1 (mono) remains, as the consequences might cause issues (e.g. capturing/processing 2x data, sending 2x data over the network).

Change-Id: Iab3fad152c6ff855ed90498ba7bbbcae6d44c332
Reviewed-on: https://chromium-review.googlesource.com/c/chromium/src/+/7595224
Cr-Commit-Position: refs/heads/main@{#1589229}

### DIFF
--- a/media/audio/ios/audio_session_manager_ios.mm
+++ b/media/audio/ios/audio_session_manager_ios.mm
@@ -13,11 +13,9 @@
 
 namespace media {
 
-// Below constant values are taken from :
+// The constant below is taken from:
 // https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/sdk/objc/components/audio/RTCAudioSessionConfiguration.m
 const int kRTCAudioSessionPreferredNumberOfChannels = 1;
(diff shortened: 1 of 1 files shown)


rakuco:

PR #9218, opened 2026-02-24
Fix: Reset is_concealed_ in ThreadedImageDecoderProxy::Resume()

msolya:
Add Reveal() method to reset is_concealed_ flag in
ThreadedImageDecoderProxy when GPU becomes available again.

ThreadedImageDecoderProxy::Conceal() sets is_concealed_ = true which
causes Finish() to silently skip posting the decode-complete task.
This flag was never reset, so after a Conceal->Freeze->Unfreeze->Reveal
cycle, existing image loaders would permanently fail to complete.

The fix adds Reveal() through the full chain:
WebModule::Reveal -> LoaderFactory -> Loader -> ThreadedImageDecoderProxy
which resets is_concealed_ = false when GPU is restored, before
UpdateResourceProvider triggers Suspend+Resume on active loaders.

### DIFF
--- a/cobalt/loader/image/threaded_image_decoder_proxy.cc
+++ b/cobalt/loader/image/threaded_image_decoder_proxy.cc
@@ -142,6 +142,11 @@ void ThreadedImageDecoderProxy::Conceal() {
   is_concealed_ = true;
 }
 
+void ThreadedImageDecoderProxy::Reveal() {
+  base::AutoLock auto_lock(conceal_lock_);
+  is_concealed_ = false;
+}
+
 }  // namespace image
 }  // namespace loader
 }  // namespace cobalt
--- a/cobalt/browser/web_module.cc
+++ b/cobalt/browser/web_module.cc
@@ -1190,6 +1190,7 @@ void WebModule::Impl::Reveal(render_tree::ResourceProvider* resource_provider,
   window_->document()->PurgeCachedResources();
   PurgeResourceCaches(should_retain_remote_typeface_cache_on_freeze_);
 
+  loader_factory_->Reveal();
   loader_factory_->UpdateResourceProvider(resource_provider_);
   layout_manager_->Resume();
 
--- a/cobalt/loader/decoder.h
+++ b/cobalt/loader/decoder.h
@@ -68,6 +68,7 @@ class Decoder {
   virtual bool Suspend() = 0;
 
   virtual void Conceal() {}
+  virtual void Reveal() {}
 
   // Resumes the decode of this resource, starting over from the zero state.
   virtual void Resume(render_tree::ResourceProvider* resource_provider) = 0;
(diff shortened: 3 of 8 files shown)


kaidokert:
@andrewsavage1 taggig you for helping review quick - you worked on this very adjacent issue during C25 launch.

cobalt/loader/image/threaded_image_decoder_proxy.cc
```diff
@@ -131,6 +131,11 @@ bool ThreadedImageDecoderProxy::Suspend() {
 
 void ThreadedImageDecoderProxy::Resume(
     render_tree::ResourceProvider* resource_provider) {
+  {
+    base::AutoLock auto_lock(conceal_lock_);
+    is_concealed_ = false;
+  }
```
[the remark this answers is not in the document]
kaidokert: @gemini-code-assist yes but that would be a much bigger change, well outside of the scope of this PR. This PR is intended to mitigate the existing bug without redesigning the whole, very stable, codebase

jellefoks:
> It's a bit strange, because this doesn't match the modern conceal->freeze then unfreeze->reveal and still uses in part suspend/resume mixed with the new variants, but I think this is a reasonable change. cc @jellefoks

This is confusing because the image decoder is affected by both the loader lifecycle, which can be suspended and resumed in which case the decoder also has to suspend and resume activity, and separately by the application lifecycle which can be concealed and revealed in which case access to the GPU can be revoked (no GPU when concealed means no texture buffer to store the decoded pixtures in).

As a result, I don't think this PR is correct because loaders are resumed on Unfreeze, at which point the application is still concealed.

msolya:
> > It's a bit strange, because this doesn't match the modern conceal->freeze then unfreeze->reveal and still uses in part suspend/resume mixed with the new variants, but I think this is a reasonable change. cc @jellefoks
> 
> This is confusing because the image decoder is affected by both the loader lifecycle, which can be suspended and resumed in which case the decoder also has to suspend and resume activity, and separately by the application lifecycle which can be concealed and revealed in which case access to the GPU can be revoked (no GPU when concealed means no texture buffer to store the decoded pixtures in).
> 
> As a result, I don't think this PR is correct because loaders are resumed on Unfreeze, at which point the application is still concealed.

@jellefoks  Thank you for the explanation you're right. Resetting `is_concealed_` in `Resume()` is incorrect because `Resume()` is called on `Unfreeze`, when the application is still in the `Concealed` state and the GPU is not yet available. `Finish()` could then attempt to create a texture without a valid GPU context. 
I've updated the PR to add a proper `Reveal()` method through the full chain: `WebModule::Reveal - LoaderFactory - Loader -ThreadedImageDecoderProxy`. Now `is_concealed_` is reset to false only in `Reveal()`, which is called when the GPU is actually restored before `UpdateResourceProvider()` triggers `Suspend+Resume` on active loaders.

cobalt/loader/image/threaded_image_decoder_proxy.cc
```diff
@@ -142,6 +142,13 @@ void ThreadedImageDecoderProxy::Conceal() {
   is_concealed_ = true;
 }
 
+void ThreadedImageDecoderProxy::Reveal() {
+  LOG(INFO) << "ThreadedImageDecoderProxy::Reveal: Resetting is_concealed_ "
```
jellefoks: This looks like accidentally left debug logging.

msolya: @jellefoks thanks, sorry about that I’ve removed it now. Could you please re-review?

PR #9224, opened 2026-02-24
media: Cleanup Allocate Experiment

johnedocampo:
Bug: 487021459

### DIFF
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -86,10 +86,11 @@ DecoderBufferAllocator::~DecoderBufferAllocator() {
 }
 
 void DecoderBufferAllocator::Suspend() {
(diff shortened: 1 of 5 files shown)


media/starboard/decoder_buffer_allocator.cc
```diff
@@ -86,11 +86,12 @@ DecoderBufferAllocator::~DecoderBufferAllocator() {
 }
 
 void DecoderBufferAllocator::Suspend() {
-  base::AutoLock scoped_lock(mutex_);
   if (is_memory_pool_allocated_on_demand_) {
     return;
   }
 
+  base::AutoLock scoped_lock(mutex_);
```
johnedocampo, line 93: During this experiment, the lock had to be moved up as `is_memory_pool_allocated_on_demand_` could be changed at lifetime. With the experiment being removed, we can restore the locks to their original position.

PR #9225, opened 2026-02-24
Enable read-only for sccache for non-repo builds.

SortaCookie:
This should provide some sort of build acceleration for PRs that don't have access to read_write.

Bug: 432823450
Change-Id: I2af70dcfa2e9609bf5dba10d13aa07047cda2d94


.github/actions/sccache/action.yaml
oxve, line 1: Remove commented out lines

PR #9227, opened 2026-02-24
android: Make V8 optimize-for-size configurable

sherryzy:
The V8 --optimize-for-size flag is now configurable via a new Java switch
on Android. Previously, this flag was either hardcoded in some Cobalt
paths or applied based on Chromium's low-end device detection.

This change introduces JavaSwitches.DISABLE_V8_OPTIMIZE_FOR_SIZE to
allow disabling the flag, which is enabled by default. This provides
more granular control over V8 memory optimization settings, enabling
targeted testing and tuning for Cobalt on Android devices. The original
Chromium-side low-end device check for this flag is now bypassed for
Cobalt builds.

Bug: 477317552

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -67,9 +67,6 @@ public final class CommandLineOverrideHelper {
     public static StringJoiner getDefaultJsFlagOverridesList() {
         StringJoiner paramOverrides = new StringJoiner(",");
 
-        // Trades a little V8 performance for significant memory savings.
-        paramOverrides.add("--optimize-for-size");
-
         return paramOverrides;
     }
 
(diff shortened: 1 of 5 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -64,15 +64,6 @@ public static List<String> getDefaultCommandLineOverridesList() {
         return paramOverrides;
     }
 
-    public static StringJoiner getDefaultJsFlagOverridesList() {
```
borongc, line 67: keep the helper function, just remove L70-71.

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -38,6 +38,9 @@ public class JavaSwitches {
   /** V8 flag to enable write protection for code memory. Value type: Boolean (presence means true) */
   public static final String V8_WRITE_PROTECT_CODE_MEMORY = "V8WriteProtectCodeMemory";
 
+  /** V8 flag to disable optimize for size. Value type: Boolean (presence means true) */
+  public static final String DISABLE_V8_OPTIMIZE_FOR_SIZE = "DisableV8OptimizeForSize";
```
borongc: probably nit: should this be `V8_DISABLE_OPTIMIZE_FOR_SIZE`?

content/renderer/render_process_impl.cc
```diff
@@ -131,10 +131,15 @@ RenderProcessImpl::RenderProcessImpl()
   }
 #endif  // BUILDFLAG(DCHECK_IS_CONFIGURABLE)
 
+  // In Cobalt, the --optimize-for-size flag is managed through Cobalt's own
+  // configuration system (e.g., JavaSwitches on Android), allowing for
+  // more granular control and experimentation.
+#if !BUILDFLAG(IS_COBALT)
```
borongc, line 137: If this is the case, the experiments may have multiple arms for both flags (low-end-device-mode and optimize-for-size).

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -112,6 +115,10 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
               + javaSwitches.get(JavaSwitches.V8_MAX_SEMI_SPACE_SIZE).replaceAll("[^0-9]", ""));
     }
 
+    if (!javaSwitches.containsKey(JavaSwitches.V8_DISABLE_OPTIMIZE_FOR_SIZE)) {
+      extraCommandLineArgs.add("--js-flags=--optimize-for-size");
```
johnxwork, line 119: This will conflict with the other `--js-flags=...` specified through CommandLineOverrideHelper. V8 will only respect the last `--js-flags` arg

sherryzy: I'm unsure I understand. In this javaswitches, it will add a few --js-flags flags, and in that CommandLineOverrideHelper, it will process these command and merge them into one --js-flags and pass them all to V8? What other --js-flags will have conflict with this one?

johnxwork: My mistake. There will only be one `--js-flags`

johnxwork (approved):

PR #9231, opened 2026-02-24
android: use l1 drm system when secure pipeline is required

jasonzhangxx:
Use widevine l1 instead of l3, which is necessary to create secure pipeline, so that secured MediaCodec can be created properly.

Bug: 485905052

### DIFF
--- a/starboard/android/shared/video_decoder.cc
+++ b/starboard/android/shared/video_decoder.cc
@@ -418,8 +418,10 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
   if (force_secure_pipeline_under_tunnel_mode) {
     SB_DCHECK_NE(tunnel_mode_audio_session_id_, -1);
     SB_DCHECK(!drm_system_);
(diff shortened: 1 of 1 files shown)


starboard/android/shared/video_decoder.cc
```diff
@@ -419,7 +419,7 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
     SB_DCHECK_NE(tunnel_mode_audio_session_id_, -1);
     SB_DCHECK(!drm_system_);
     drm_system_to_enforce_tunnel_mode_ = std::make_unique<DrmSystem>(
```
kjyoun, line 423: Isn't it to better to add comment about why we we should select L1 for tunnel mode
It would help why we choose L1.

```
// To create secure pipeline for tunnel mode, we should use L1("com.widevine.alpha")
```

jasonzhangxx: Done

PR #9236, opened 2026-02-24
android: correct the calculation of audio track buffer size

jasonzhangxx:
Bug: 311422213

### DIFF
--- a/starboard/android/shared/audio_track_audio_sink_type.cc
+++ b/starboard/android/shared/audio_track_audio_sink_type.cc
@@ -477,20 +477,6 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
   SB_CHECK(audio_track_audio_sink_type_);
   JNIEnv* env = AttachCurrentThread();
 
-  const bool force_tunnel_mode =
-      features::FeatureList::IsEnabled(features::kForceTunnelMode);
-  if (force_tunnel_mode) {
-    // AudioTrack.setPlaybackParams() needs extra buffer to support playback
-    // speed greater than 1.0x.
-    const double kMaxPlaybackSpeed = 2.0;
-    return std::max<int>(
-        AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
-            env, sample_type, channels, sampling_frequency_hz),
-        audio_track_audio_sink_type_->GetMinBufferSizeInFramesInternal(
-            channels, sample_type, sampling_frequency_hz) *
-            kMaxPlaybackSpeed);
-  }
-
   return std::max(
       AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
           env, sample_type, channels, sampling_frequency_hz),
--- a/starboard/android/shared/player_components_factory.cc
+++ b/starboard/android/shared/player_components_factory.cc
@@ -417,12 +417,38 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
 
     // AudioRenderer prefers to use kSbMediaAudioSampleTypeFloat32 and only uses
     // kSbMediaAudioSampleTypeInt16Deprecated when float32 is not supported.
-    int min_frames_required = SbAudioSinkGetMinBufferSizeInFrames(
-        creation_parameters.audio_stream_info().number_of_channels,
+    const auto sample_type =
         SbAudioSinkIsAudioSampleTypeSupported(kSbMediaAudioSampleTypeFloat32)
             ? kSbMediaAudioSampleTypeFloat32
-            : kSbMediaAudioSampleTypeInt16Deprecated,
+            : kSbMediaAudioSampleTypeInt16Deprecated;
+
+    int min_frames_required = SbAudioSinkGetMinBufferSizeInFrames(
+        creation_parameters.audio_stream_info().number_of_channels, sample_type,
         creation_parameters.audio_stream_info().samples_per_second);
+
+    // To avoid redundant IsTunnelModeSupported() checks, we simply only check
+    // if tunnel mode is enabled here.
+    if (creation_parameters.audio_codec() != kSbMediaAudioCodecNone &&
+        creation_parameters.video_codec() != kSbMediaVideoCodecNone) {
+      const bool force_tunnel_mode =
+          FeatureList::IsEnabled(features::kForceTunnelMode);
+      MimeType video_mime_type(creation_parameters.video_mime());
+      if (force_tunnel_mode ||
+          video_mime_type.GetParamBoolValue("tunnelmode", false)) {
+        // AudioTrack.setPlaybackParams() might need extra buffer to support
+        // playback speed greater than 1.0x.
+        const double kMaxPlaybackSpeed = 2.0;
+        JNIEnv* env = AttachCurrentThread();
+        min_frames_required = std::max<int>(
+            min_frames_required,
+            AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
+                env, sample_type,
+                creation_parameters.audio_stream_info().number_of_channels,
+                creation_parameters.audio_stream_info().samples_per_second) *
+                kMaxPlaybackSpeed);
+      }
+    }
+
     // On Android 5.0, the size of audio renderer sink buffer need to be two
     // times larger than AudioTrack minBufferSize. Otherwise, AudioTrack may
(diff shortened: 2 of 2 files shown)


starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -454,10 +454,10 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
     const double kMaxPlaybackSpeed = 2.0;
     return std::max<int>(
```
kjyoun: I think you don't need `<int>` here

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -454,10 +454,10 @@ int AudioTrackAudioSinkType::GetMinBufferSizeInFrames(
...
         AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
-            env, sample_type, channels, sampling_frequency_hz),
+            env, sample_type, channels, sampling_frequency_hz) *
+            kMaxPlaybackSpeed,
```
kjyoun: formatting is weird
?

```
return std::max(
    AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames(
        env, sample_type, channels, sampling_frequency_hz) * kMaxPlaybackSpeed,
    audio_track_audio_sink_type_->GetMinBufferSizeInFramesInternal(
        channels, sample_type, sampling_frequency_hz)
);
```

jasonzhangxx: The pre-commit clang-format make it like this. I tried it again.

starboard/android/shared/audio_track_audio_sink_type.cc
jasonzhangxx: Yes, ` AudioOutputManager::GetInstance()->GetMinBufferSizeInFrames( env, sample_type, channels, sampling_frequency_hz) * kMaxPlaybackSpeed` is double, `audio_track_audio_sink_type_->GetMinBufferSizeInFramesInternal(channels, sample_type, sampling_frequency_hz)` is int, so std::max cannot know if it should use <int> or <double>.

starboard/android/shared/player_components_factory.cc
```diff
@@ -410,12 +411,29 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
...
         SbAudioSinkIsAudioSampleTypeSupported(kSbMediaAudioSampleTypeFloat32)
             ? kSbMediaAudioSampleTypeFloat32
-            : kSbMediaAudioSampleTypeInt16Deprecated,
+            : kSbMediaAudioSampleTypeInt16Deprecated;
+
+    int min_frames_required = SbAudioSinkGetMinBufferSizeInFrames(
+        creation_parameters.audio_stream_info().number_of_channels, sample_type,
         creation_parameters.audio_stream_info().samples_per_second);
+
+    if (is_tunnel_mode_used_) {
```
kjyoun: With `is_tunnel_mode_used_` as a member variable,  you are creating race condition between `CreateComponents` and `GetAudioRendererParams` :  `CreateComponents`  should be called before ``GetAudioRendererParams` 

Can you remove this member variable?

Gemini suggest  you can get `is_tunnel_mode enabled` from creation_param , which is one of input argument in this method.


----
```
bool IsTunnelModeUsed(const CreationParameters& creation_parameters) {
  // 1. Basic Requirement: Both audio and video must be present.
  if (creation_parameters.audio_codec() == kSbMediaAudioCodecNone ||
      creation_parameters.video_codec() == kSbMediaVideoCodecNone) {
    return false;
  }

  // 2. Intent Check: Parse the MIME type to see if "tunnelmode" is requested.
  MimeType video_mime_type(creation_parameters.video_mime());
  const bool force_tunnel_mode = FeatureList::IsEnabled(features::kForceTunnelMode);
  
  // Use MimeType::is_valid() to ensure we don't process garbage.
  bool tunnel_mode_requested = force_tunnel_mode;
  if (video_mime_type.is_valid()) {
    tunnel_mode_requested |= video_mime_type.GetParamBoolValue("tunnelmode", false);
  }

  if (!tunnel_mode_requested) {
    return false;
  }

  // 3. Support Check: Check hardware/platform capabilities.
  // We need a dummy variable for the secure pipeline pointer required by the signature.
  bool dummy_secure_pipeline = false;
  if (!IsTunnelModeSupported(creation_parameters, &dummy_secure_pipeline)) {
    return false;
  }

  // 4. Final Gate: Attempt to generate a Session ID.
  // In your original code, if this returns -1, is_tunnel_mode_used_ remains false.
  int session_id = GenerateAudioSessionId(creation_parameters);
  
  return session_id != -1;
}
```

jasonzhangxx: The `PlayerComponents::Factory::CreateComponents` is at https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:starboard/shared/starboard/player/filter/player_components.cc;l=158, which creates sub components first then create audio/video renderer. But I agree it's not a good way to get that. I'll refine the change.

jasonzhangxx: I updated the code. It now uses a compromised solution to only check if tunnel mode is enabled and no redundant `IsTunnelModeSupported()` checks, though it still need to parse video mime twice.

starboard/android/shared/player_components_factory.cc
```diff
@@ -417,12 +410,38 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
...
+
+    // To avoid redundant IsTunnelModeSupported() checks, we simply only check
+    // if tunnel mode is enabled here.
+    if (creation_parameters.audio_codec() != kSbMediaAudioCodecNone &&
+        creation_parameters.video_codec() != kSbMediaVideoCodecNone) {
+      const bool force_tunnel_mode =
+          FeatureList::IsEnabled(features::kForceTunnelMode);
+      MimeType video_mime_type(creation_parameters.video_mime());
+      if (force_tunnel_mode ||
+          video_mime_type.GetParamBoolValue("tunnelmode", false)) {
```
kjyoun, line 437: "tunnelmode" string keep being repeated in this file?

can you define a constant for it and use it instead?

starboard/android/shared/player_components_factory.cc
```diff
@@ -417,12 +410,38 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
...
         creation_parameters.audio_stream_info().samples_per_second);
+
+    // To avoid redundant IsTunnelModeSupported() checks, we simply only check
+    // if tunnel mode is enabled here.
+    if (creation_parameters.audio_codec() != kSbMediaAudioCodecNone &&
+        creation_parameters.video_codec() != kSbMediaVideoCodecNone) {
+      const bool force_tunnel_mode =
+          FeatureList::IsEnabled(features::kForceTunnelMode);
+      MimeType video_mime_type(creation_parameters.video_mime());
+      if (force_tunnel_mode ||
```
kjyoun, line 436: This pattern is used above at line 270-274

```
    const bool force_tunnel_mode =
          FeatureList::IsEnabled(features::kForceTunnelMode);
      enable_tunnel_mode =
          force_tunnel_mode ||
          video_mime_type.GetParamBoolValue("tunnelmode", false);
```

Can you make one helper method and use it both case?


`````

bool EnableTunnelMode(mime_type) {
  if (FeatureList::IsEnabled(features::kForceTunnelMode)) {
     return true;
  }
  return FeatureList::IsEnabled(features::kForceTunnelMode);
}

The here becomes more readable

if (creation_parameters.audio_codec() != kSbMediaAudioCodecNone &&
    creation_parameters.video_codec() != kSbMediaVideoCodecNone &&
    EnableTunnelMode(MimeType(creation_parameters.video_mime())) {
...

}
````

jasonzhangxx: There's already another `IsTunnelModeSupported()` function. I don't want to add another function with similar name. Let's keep it as is.

starboard/android/shared/player_components_factory.cc
jasonzhangxx: We used to have raw string for mime attribute, though other mime attributes have been removed. It's a temporary solution before we have Finch but still want to use mime string control the behavior. They'll eventually be removed. Let's keep it as is.

PR #9245, opened 2026-02-25
tvOS: Remove starboard/darwin

rakuco:
This directory came from C25 and is not used by anything. Most of its contents are either outdated or have more up-to-date counterparts in starboard/tvos with the exception of launcher.py and tvos_launcher.py, which have been moved to starboard/tvos in case they are useful in the future.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #9246, opened 2026-02-25
tvOS: Remove vpx_loop_filter_manager.*

rakuco:
These files were needed by vpx_video_decoder.*, which were removed in #8336 because they were only used in decode to texture mode.

Bug: 459507205
Bug: 447334535


rakuco:

PR #9248, opened 2026-02-25
tvos: Update VP9 header include path

abhijeetk:
Update the include and source paths for VP9 hardware accelerated video
sample buffer builder files within the tvOS Starboard component. This
change synchronizes the paths with recent internal submodule directory
structure updates. It resolves a compilation error in the internal
build pipeline by pointing to the correct locations.

This syncs changes from PR #9029 with the updated VP9 header include
paths introduced in PR #8867. 

Bug: 460479616
Bug: 487326366

### DIFF
--- a/starboard/tvos/shared/BUILD.gn
+++ b/starboard/tvos/shared/BUILD.gn
@@ -156,12 +156,9 @@ static_library("starboard_platform") {
   ]
 
   if (is_internal_build) {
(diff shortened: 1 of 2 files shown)


abhijeetk:

rakuco (commented):
informal lgtm from my side

as a side note, we should probably drop the `uikit/` part from the internal files too

abhijeetk:
@fayev-g Just a friendly reminder !

PR #9250, opened 2026-02-25
media: Refactor DecoderBufferAllocator strategies

xiaomings:
This change refactors how DecoderBufferAllocator strategies are managed.
It removes explicit Enable methods and a StrategyType enum from
media/base/decoder_buffer.h. Instead, strategy instantiation logic is
moved into a StrategyCreateCB lambda factory.

DecoderBufferAllocator now accepts this lambda to update its strategy.
This allows concrete strategy implementations to be injected from
Starboard-specific contexts, such as h5vcc_settings.cc, without
polluting the core media/base interface. It also exposes
DecoderBuffer::Allocator::Get() for global access to the allocator.
This improves modularity and testability of buffer allocation logic.

Bug: 454441375

### DIFF
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -17,6 +17,7 @@
 #include <algorithm>
 
 #include "base/feature_list.h"
+#include "base/functional/bind.h"
 #include "base/logging.h"
 #include "build/build_config.h"
 #include "media/base/media_switches.h"
@@ -35,6 +36,14 @@
 namespace media {
 
 namespace {
+
+using DefaultReuseAllocatorStrategy =
+    BidirectionalFitDecoderBufferAllocatorStrategy<
+        starboard::common::ReuseAllocatorBase>;
+using InPlaceReuseAllocatorStrategy =
+    BidirectionalFitDecoderBufferAllocatorStrategy<
+        starboard::common::InPlaceReuseAllocatorBase>;
+using starboard::common::experimental::MediaBufferPool;
 
 // Used to determine if the memory allocated is large. The underlying logic can
 // be different.
@@ -67,12 +76,6 @@ DecoderBufferAllocator::DecoderBufferAllocator(
   }
 
   base::AutoLock scoped_lock(mutex_);
-
-#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
-  // Uncomment the following line to default enable MediaBufferPoolStrategy.
-  // media_buffer_pool_strategy_state_ = MediaBufferPoolStrategyState::kEnabled;
-#endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
-
   EnsureStrategyIsCreated();
 }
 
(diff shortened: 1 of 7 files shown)


johnedocampo (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -19,6 +19,7 @@
 #include "cobalt/browser/h5vcc_settings/public/mojom/h5vcc_settings.mojom-blink.h"
 #include "media/base/decoder_buffer.h"
 #include "media/filters/source_buffer_state.h"
+#include "media/starboard/decoder_buffer_allocator.h"
```
borongc, line 22: Is this needed?

xiaomings: It's needed for the calls like `::media::DecoderBufferAllocator::EnableInPlaceReuseAllocatorBase()` below.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -221,6 +226,23 @@ size_t DecoderBufferAllocator::GetMaximumMemoryCapacity() const {
   return 0;
 }
 
+void DecoderBufferAllocator::UpdateAllocatorStrategy(
+    StrategyCreateCB create_cb) {
+  DCHECK(!create_cb.is_null());
+
+  base::AutoLock scoped_lock(mutex_);
+  experimental_strategy_create_cb_ = std::move(create_cb);
```
borongc, line 234: probably minor: this callback is moved, and it won't be null as it is RepeatingCallback after it is moved to `experimental_strategy_create_cb_`.

If this is intended to be called per playback, and only once, should we consider to make it OnceCallback?

xiaomings: `experimental_strategy_create_cb_` could be called multiple times to work with allocated on demand and suspend/resume.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -297,44 +305,31 @@ void DecoderBufferAllocator::EnsureStrategyIsCreated() {
...
+  is_strategy_switch_pending_ = false;
 
-  if (media_buffer_pool_strategy_state_ ==
-      MediaBufferPoolStrategyState::kEnabled) {
-    auto pool = starboard::common::experimental::MediaBufferPool::Acquire();
-    if (pool) {
-      strategy_.reset(new MediaBufferPoolDecoderBufferAllocatorStrategy(
-          pool, initial_capacity_, allocation_unit_));
-      LOG(INFO) << "DecoderBufferAllocator is using MediaBufferPool.";
+  if (!experimental_strategy_create_cb_.is_null()) {
```
borongc, line 310: If this is called multiple times, would CHECK(!experimental_strategy_create_cb_.is_null()) make more sense? I will leave this to you.

borongc (approved):

media/starboard/decoder_buffer_allocator.cc
xiaomings: It could be null if there is no experiments running.

johnxwork (approved):

PR #9256, opened 2026-02-25
cobalt: Guard storage migration with integrity cookie

zhongqiliang:
Introduce a mechanism to prevent storage migration tasks from
re-executing on a profile that has already been processed by
Cobalt 26+. Before running the migration logic, check for the
presence of the "yt-dev.storage-integrity" cookie. If this
cookie exists, it indicates that a previous Cobalt 26+ instance
has already performed the migration. Skipping further migration
in such cases prevents unintended removal of existing cookies
and localStorage data.

Bug: 487739842

### DIFF
--- a/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
+++ b/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
@@ -45,17 +45,6 @@ namespace migrate_storage_record {
 
 namespace {
 
-constexpr char kStorageIntegrityCookieName[] = "yt-dev.storage-integrity";
-
-bool HasIntegrityCookie(const std::vector<net::CanonicalCookie>& cookies) {
-  for (const auto& cookie : cookies) {
-    if (cookie.Name() == kStorageIntegrityCookieName) {
-      return true;
-    }
-  }
-  return false;
-}
-
 std::string GetApplicationKey(const GURL& url) {
   std::string encoded_url;
   base::Base64UrlEncode(url_matcher::util::Normalize(url).spec(),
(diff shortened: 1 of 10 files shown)


haozheng-cobalt:
the cookie `yt-dev.storage-integrity` is set when the app starts and [StorageIntegrityMonitor](https://source.corp.google.com/piper///depot/google3/video/youtube/web/living_room/core/lib/storage_integrity_monitor/storage_integrity_monitor.ts;l=103?q=INTEGRITY_COOKIE_KEY) is instantiated. 

Could it be possible that this cookie is supposed to exist before Cobalt's storage migration manager runs? and the issue we were running into is that Cobalt storage migration manager falsely purges the cookie? If that is the case, checking for this cookie is supposed to return true. Would only be able to confirm to test this locally, but I had difficulties testing the storage migration locally.

zhongqiliang:
> and the issue we were running into is that Cobalt storage migration manager falsely purges the cookie?

We had the bug, it is fixed by [this](https://github.com/youtube/cobalt/pull/8951)

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -45,6 +45,17 @@ namespace migrate_storage_record {
 
 namespace {
 
+constexpr char kStorageIntegrityCookieName[] = "yt-dev.storage-integrity";
```
johnxwork: I think checking a hard-coded cookie is extremely risky.

johnxwork:
Would you also update the description/title and remove local logging? Thanks!

PR #9258, opened 2026-02-25
cobalt/metrics: Expose vm metrics to UMA data stream

tasunnn:
cobalt: Expose VM metrics to UMA

This change introduces the collection and reporting of virtual memory
metrics, namely PrivateFootprintSwap and VmSize, to UMA.

The goal is to provide deeper insights into Cobalt's memory footprint,
especially regarding virtual memory usage and swap activity. This
additional data will aid in identifying and optimizing memory-related
performance issues and assist in general memory management analysis.

The metrics are collected from the OS memory dumps and propagated
through the memory instrumentation service before being emitted as
UMA histograms. Browser tests are updated to verify their emission.

Bug: 484063214


cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
```diff
@@ -336,6 +340,13 @@ void CobaltMemoryMetricsEmitter::CollateResults() {
   base::UmaHistogramMemoryLargeMB(
       "Memory.Total.SharedMemoryFootprint",
       static_cast<int>(shared_footprint_total_kb / kKiB));
+  // VM specific metrics
+  base::UmaHistogramMemoryLargeMB(
+      "Memory.Total.PrivateFootprintSwap",
+      static_cast<int>(private_footprint_swap_total_kb / kKiB));
+  base::UmaHistogramMemoryLargeMB(
```
johnxwork, line 353: VmSize is actually also expose through web API and collected by Kabuki. We can use it to study the correlation between UMA and Kabuki metrics.

tasunnn: Yeah I think this metric here is obtained from the same API, we also add them to UMA so they can still be available even if kabuki doesn't directly call the apis

Awallky (approved):
These changes look consistent with similar changes.
The only thing I would recommend to ensure this works as intended is that you locally test the MetricsBrowserTest with this command, since it is disabled in CI - 

`vpython3 cobalt/testing/browser_tests/run_browser_tests.py out/${PLAT}_devel/cobalt_browsertests *CobaltMetricsBrowserTest*`

PR #9262, opened 2026-02-26
net/: Use both QUIC and TCP for initial URL requests

ckim1346:
This PR reintroduces the `use_quic_for_unknown_origins` flag to try both QUIC and TCP for initial/unknown URL requests. If QUIC returns first, it will cancel the TCP request and otherwise fall back on the TCP ensuring we try quick but also have a reliable fallback. Additionally, we allow non-production builds to use any port whereas production builds can only use privileged ports.

Please refer to #2981 and #3514. 

Bug: 467778593


services/network/network_context.cc
```diff
@@ -2914,6 +2914,10 @@ URLRequestContextOwner NetworkContext::MakeURLRequestContext(
       *base::CommandLine::ForCurrentProcess(), is_quic_force_disabled,
       &session_params, quic_context->params());
 
+#if BUILDFLAG(IS_ANDROIDTV)
```
jellefoks: We should use `BUILDFLAG(IS_COBALT)` here so that it's also used for 3p.

net/http/http_stream_factory_job_controller.cc
```diff
@@ -1272,6 +1272,22 @@ HttpStreamFactory::JobController::GetAlternativeServiceInfoInternal(
           url::SchemeHostPort(original_url),
           request_info.network_anonymization_key);
   if (alternative_service_info_vector.empty()) {
+#if BUILDFLAG(IS_ANDROIDTV)
```
jellefoks: We should use `BUILDFLAG(IS_COBALT)` here so that it's also used for 3p.

net/http/http_network_session.h
```diff
@@ -303,6 +309,11 @@ class NET_EXPORT HttpNetworkSession : public base::PowerSuspendObserver {
   // Disable QUIC for new streams.
   void DisableQuic();
 
+#if BUILDFLAG(IS_ANDROIDTV)
```
jellefoks: We should use `BUILDFLAG(IS_COBALT)` here so that it's also used for 3p.

net/http/http_network_session.h
```diff
@@ -149,6 +149,12 @@ struct NET_EXPORT HttpNetworkSessionParams {
   // Enables QUIC support.
   bool enable_quic = true;
 
+#if BUILDFLAG(IS_ANDROIDTV)
```
jellefoks: We should use `BUILDFLAG(IS_COBALT)` here so that it's also used for 3p.

net/http/http_network_session.cc
```diff
@@ -393,6 +393,12 @@ void HttpNetworkSession::DisableQuic() {
   params_.enable_quic = false;
 }
 
+#if BUILDFLAG(IS_ANDROIDTV)
```
jellefoks: We should use `BUILDFLAG(IS_COBALT)` here so that it's also used for 3p.

net/http/http_stream_factory_job_controller.cc
```diff
@@ -1272,6 +1272,22 @@ HttpStreamFactory::JobController::GetAlternativeServiceInfoInternal(
...
+    if (session_->IsQuicEnabled() && session_->UseQuicForUnknownOrigin()) {
```
jellefoks, line 1280: I think the corresponding comment in C25 is useful and should be kept here too:

```
  // This block of code suggests QUIC connection for initial requests to a
  // new host. This method is proven to provide performance benefit while still
  // enabling Cobalt network module to fall back on TCP connection when QUIC
  // fails or is too slow.
```

net/http/http_stream_factory_job_controller.cc
```diff
@@ -1272,6 +1272,22 @@ HttpStreamFactory::JobController::GetAlternativeServiceInfoInternal(
...
+      url::SchemeHostPort origin(original_url);
+#if defined(COBALT_BUILD_TYPE_GOLD)
```
johnxwork, line 1282: When would port not be 443?

I also think it's dangerous to have qa and gold differ on this behavior. Prod could break if there's some issue with the prod-specific code.

ckim1346: #3514 I would assume for testing/debugging purposes -- @kaidokert do you remember the exact reasoning?

jellefoks: I think this is for mirroring the enforcing of https for gold builds.

jellefoks: I think we should at least enforce it to be a privileged port (<1024).

ckim1346: Updated, PTAL

jellefoks: You know, even though this is much the same as C25, reading this, it's kind of hard to follow because of the if block. Since the passthrough is to just return an empty object, it may be more readable with an `if (origin.port() >= kUnrestrictedPort) { return AlternativeServiceInfo(); }` in a single `#if defined(COBALT_BUILD_TYPE_GOLD)` block, and a comment that we're not defaulting to use QUIC for unprivileged ports.

jellefoks (approved):
Awesome!

PR #9268, opened 2026-02-26
starboard: const-ify |data| in CobaltExtensionPlatformServiceApi's Send()

rakuco:
Make the contract more explicit and indicate that the data received by Send() is read-only and any changes to it should be done by copying its contents.

Doing so also makes it easier and cleaner to use base::span in the browser side of the H5vccPlatformService implementation in a follow-up PR.

Bug: 487712437

### DIFF
--- a/cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
+++ b/cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
@@ -140,12 +140,9 @@ void PlatformServiceImpl::Send(const std::vector<uint8_t>& data,
   }
 
   uint64_t output_length = 0;
-  std::vector<uint8_t> mutable_data = data;
-  void* data_ptr = mutable_data.empty() ? nullptr : mutable_data.data();
-  uint64_t data_length = mutable_data.size();
   bool invalid_state = false;
 
-  void* response_ptr = api->Send(platform_service_, data_ptr, data_length,
+  void* response_ptr = api->Send(platform_service_, data.data(), data.size(),
                                  &output_length, &invalid_state);
 
   if (invalid_state) {
--- a/starboard/android/shared/platform_service.cc
+++ b/starboard/android/shared/platform_service.cc
@@ -90,7 +90,7 @@ void Close(CobaltExtensionPlatformService service) {
 }
 
 void* Send(CobaltExtensionPlatformService service,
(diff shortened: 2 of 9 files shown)


rakuco:

rakuco:
@hlwarriner not sure if this is actually allowed since it changes the function's signature?

hlwarriner:
> @hlwarriner not sure if this is actually allowed since it changes the function's signature?

I think there are two separate issues.

1. If we do want to change the signature, we should introduce a new version of the extension that adds the new function (with the `const void*` signature) and marks the old one as deprecated. Please see https://developers.google.com/youtube/cobalt/docs/codelabs/starboard_extensions/codelab#extension_versioning for more context.
2. I'm not sure we want to change the signature. I get your point that it seems like the data sent to the platform service should be `const`, but the Starboard extension wasn't defined that way and it may have been for good reason. For example, we may have wanted to give platform services the opportunity to modify the incoming buffer in-place. And because partners own their extension implementations, the Cobalt team isn't able to view them. @y4vor, I think you were involved in the original design + implementation; do you have an opinion on this?

rakuco:
@hlwarriner @y4vor feel free to revert this one if you think it makes more sense. It does change the API for partners so I'm still cautious about whether this is the way to go or not.

y4vor:
> > @hlwarriner not sure if this is actually allowed since it changes the function's signature?
> 
> I think there are two separate issues.
> 
> 1. If we do want to change the signature, we should introduce a new version of the extension that adds the new function (with the `const void*` signature) and marks the old one as deprecated. Please see https://developers.google.com/youtube/cobalt/docs/codelabs/starboard_extensions/codelab#extension_versioning for more context.
> 2. I'm not sure we want to change the signature. I get your point that it seems like the data sent to the platform service should be `const`, but the Starboard extension wasn't defined that way and it may have been for good reason. For example, we may have wanted to give platform services the opportunity to modify the incoming buffer in-place. And because partners own their extension implementations, the Cobalt team isn't able to view them. @y4vor, I think you were involved in the original design + implementation; do you have an opinion on this?

Technically we should bump the version of the extension as this is changing the interface. However for the current 27.lts release it would be fine as it isn't backwards compatible with previous releases. Going forward e.g. 28.lts we would need to bump the version for changes like this one.

PR #9270, opened 2026-02-26
cobalt: Build extra crashpad targets in gn_all only when use_evergreen=true

rakuco:
These targets do not exist on platforms like tvOS, and in any case they only seem to be used in 3p builds (see e.g.
`cobalt/build/evergreen-x64/package.json`).

Bug: 477518757

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -69,7 +69,7 @@ group("gn_all") {
   if (is_starboard && build_updater_sandbox) {
     deps += [ "//chrome/updater/cobalt:updater_sandbox" ]
   }
-  if (use_crashpad) {
+  if (use_evergreen && use_crashpad) {
     deps += [
       "//third_party/crashpad/crashpad/handler:crashpad_handler($native_target)",
       "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",


rakuco:

rakuco:
@andrewsavage1 @hlwarriner PTAL. I'm not sure if `use_evergeen` is the right variable in this case, but I think my understanding of why this is needed is correct.

rakuco:
Thanks for reviewing this. Can you merge the PR as well?

> That said, I'm not sure if it can be built for iOS. It's possible it can be built on a host machine, to which the Crashpad database can also be extracted, so that the database can be queried. But anyway, it sounds fine to me to disable the build for now.

FWIW `crashpad_database_util` can be built with `host_toolchain`. I can send a follow-up PR that either does that if you prefer, or even something that sets `native_target` to `host_toolchain` when `build_with_separate_cobalt_toolchain` is false (I'm just not very familiar with 3p and dunno if this would end up causing other kinds of failures).

hlwarriner:
> Thanks for reviewing this. Can you merge the PR as well?
>
Sure, will do.

> > That said, I'm not sure if it can be built for iOS. It's possible it can be built on a host machine, to which the Crashpad database can also be extracted, so that the database can be queried. But anyway, it sounds fine to me to disable the build for now.
> 
> FWIW `crashpad_database_util` can be built with `host_toolchain`. I can send a follow-up PR that either does that if you prefer, or even something that sets `native_target` to `host_toolchain` when `build_with_separate_cobalt_toolchain` is false (I'm just not very familiar with 3p and dunno if this would end up causing other kinds of failures).

Sounds good. I would probably prefer the first option.

PR #9275, opened 2026-02-26
ci: Add Gemini conflict resolution to cherry-pick

oxve:
This change integrates the Gemini API into the cherry-pick workflow to
automatically resolve merge conflicts. When a cherry-pick fails, the
gemini-cli action is invoked to attempt to resolve any conflicts.
This reduces manual intervention for routine cherry-picks.

Bug: 488071229

### DIFF
--- a/.github/workflows/label-cherry-pick.yaml
+++ b/.github/workflows/label-cherry-pick.yaml
@@ -6,6 +6,9 @@ on:
       - labeled
       - closed
 
+permissions:
+  contents: read
+
 jobs:
   prepare_branch_list:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


linxinan-yt (approved):
LGTM

.github/workflows/label-cherry-pick.yaml
```diff
@@ -23,45 +26,47 @@ jobs:
...
         for branch in "${branches[@]}"; do
-          if [[ $branch == $BASE_REF ]]; then
+          if [[ ${branch} == ${BASE_REF} ]]; then
             continue
           fi
-          if [[ ${labels[@]} =~ "cp-$branch" ]]; then
-            filtered_branches+=("$branch")
-          fi
+
+          for label in "${labels_array[@]}"; do
```
andrewsavage1, line 45: It looks this PR also fixes a case where not all labels are accounted for when a PR is closed if I'm reading this correctly—let's add that to the description

.github/workflows/label-cherry-pick.yaml
```diff
@@ -78,33 +83,72 @@ jobs:
       - name: Cherry pick merge commit
         id: cherry-pick
         continue-on-error: true
+        env:
+          GEMINI_API_KEY: ${{ secrets.GEMINI_ACTIONS_API_KEY }}
```
andrewsavage1, line 87: How often will we have to update this? What will be the failure if it expires? Does it have any other restrictions we'd have to think about like quota?

PR #9286, opened 2026-02-27
media: Implement madvise for idle buffer memory

xiaomings:
Introduce a mechanism to release physical memory associated with
decoder buffers when their allocation pools become idle. This change
adds a Decommit method to the common allocator interface, which is
implemented by StarboardMemoryAllocator using madvise(MADV_DONTNEED).

The ReuseAllocatorBase and InPlaceReuseAllocatorBase now invoke this
Decommit method when all internally managed blocks are freed or
during batched free operations. This effectively signals to the
operating system that the physical pages backing these buffers are
no longer actively needed, allowing them to be reclaimed.

A new DecoderBufferAllocator strategy, enabled via H5VCC settings,
configures the media pipeline to utilize this decommit-capable
allocator, reducing Cobalt's resident set size and improving overall
memory efficiency when media playback is not active.

Bug: 454441375

### DIFF
--- a/media/starboard/bidirectional_fit_decoder_buffer_allocator_strategy.h
+++ b/media/starboard/bidirectional_fit_decoder_buffer_allocator_strategy.h
@@ -27,11 +27,14 @@ class BidirectionalFitDecoderBufferAllocatorStrategy
  public:
   BidirectionalFitDecoderBufferAllocatorStrategy(
       std::size_t initial_capacity,
-      std::size_t allocation_increment)
-      : birectional_fit_allocator_(&fallback_allocator_,
+      std::size_t allocation_increment,
+      bool enable_decommit_on_idle)
+      : fallback_allocator_(enable_decommit_on_idle),
+        birectional_fit_allocator_(&fallback_allocator_,
                                    initial_capacity,
                                    kSmallAllocationThreshold,
-                                   allocation_increment) {}
+                                   allocation_increment,
+                                   enable_decommit_on_idle) {}
 
   void* Allocate(DemuxerStream::Type type,
                  size_t size,
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -90,8 +90,7 @@ DecoderBufferAllocator::~DecoderBufferAllocator() {
 
 // static
 DecoderBufferAllocator* DecoderBufferAllocator::Get() {
-  return static_cast<DecoderBufferAllocator*>(
-      DecoderBuffer::Allocator::Get());
+  return static_cast<DecoderBufferAllocator*>(DecoderBuffer::Allocator::Get());
 }
 
 void DecoderBufferAllocator::Suspend() {
@@ -163,8 +162,8 @@ void DecoderBufferAllocator::Free(DemuxerStream::Type type,
   }
 #endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
 
-  bool should_reset_strategy = is_strategy_switch_pending_ ||
-                               is_memory_pool_allocated_on_demand_;
+  bool should_reset_strategy =
+      is_strategy_switch_pending_ || is_memory_pool_allocated_on_demand_;
 
   if (should_reset_strategy && strategy_->GetAllocated() == 0) {
     // `strategy_->PrintAllocations()` will be called inside the dtor when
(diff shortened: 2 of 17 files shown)


media/starboard/starboard_memory_allocator.h
```diff
@@ -16,21 +16,39 @@
...
+                 << (enable_decommit_ ? "enabled" : "disabled") << ".";
+  }
+
   void* Allocate(std::size_t size) override { return Allocate(size, 1); }
 
   void* Allocate(std::size_t size, std::size_t alignment) override {
+    if (enable_decommit_) {
+      alignment = std::max(alignment, page_size_);
+      // Align up the size.
+      size = (size + page_size_ - 1) & ~(page_size_ - 1);
```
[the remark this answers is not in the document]
xiaomings: No longer applies.

starboard/common/in_place_reuse_allocator_base.cc
```diff
@@ -215,6 +215,15 @@ void InPlaceReuseAllocatorBase::Free(void* memory) {
...
       }
+
+      if (enable_decommit_on_idle_) {
+        SB_LOG(INFO) << "Batched free triggered idle state, decommitting "
+                     << fallback_allocations_.size() << " fallback allocations.";
+        for (const auto& fallback_allocation : fallback_allocations_) {
+          fallback_allocator_->Decommit(fallback_allocation.address,
+                                        fallback_allocation.size);
+        }
+      }
```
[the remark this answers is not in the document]
xiaomings: I'll keep this as is for now.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -265,6 +264,32 @@ void DecoderBufferAllocator::SetAllocateOnDemand(bool enabled) {
   }
 }
 
+// static
+void DecoderBufferAllocator::EnableDecommitableAllocatorStrategy() {
+  auto* allocator = Get();
+  CHECK(allocator);
+  allocator->UpdateAllocatorStrategy(base::BindRepeating(
+      [](int initial_capacity, int allocation_unit)
+          -> std::unique_ptr<DecoderBufferAllocator::Strategy> {
```
kjyoun, line 273: nit: I think that you don't need "-> std::unique_ptr<DecoderBufferAllocator::Strategy>" here, since this lambda has only one return value, so that the compiler can deduce its return type

media/starboard/starboard_memory_allocator.h
```diff
@@ -16,31 +16,74 @@
...
+    }
+    void* p = nullptr;
+    posix_memalign(&p, std::max(alignment, sizeof(void*)), *size);
+    return p;
   }
   void Free(void* memory) override { free(memory); }
+
+  void Decommit(void* memory, std::size_t size) override {
+    uintptr_t start = reinterpret_cast<uintptr_t>(memory);
+#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
```
kjyoun, line 72: why not CHECK always?

media/starboard/starboard_memory_allocator_test.cc
```diff
@@ -0,0 +1,59 @@
+// Copyright 2024 The Cobalt Authors. All Rights Reserved.
```
kjyoun: nit: 2026

starboard/common/bidirectional_fit_reuse_allocator_test.cc
```diff
@@ -26,21 +26,95 @@
...
+
 namespace starboard {
 namespace common {
 
 struct AlignedMemoryDeleter {
   void operator()(uint8_t* p) { free(p); }
 };
 
-static constexpr int kBufferSize = 1 * 1024 * 1024;
+static constexpr int kBufferSize = 16 * 1024 * 1024;
```
kjyoun: can you add the comment about why 16MB ?

media/starboard/decoder_buffer_allocator.cc
xiaomings: Removing this triggers compile errors, probably due to the return type isn't exactly the same as `std::unique_ptr<DecoderBufferAllocator::Strategy>`.

media/starboard/starboard_memory_allocator.h
xiaomings: I tend to add DCHECK()s whenever appropriate without taking too much of memory/performance into account, and I would avoid adding obvious CHECK()s around the code base.  While the cost is small, CHECK()s still have impact on both binary size and performance, which I would be more careful to add.

Totw often contain good suggestions but aren't official guidelines.

media/starboard/starboard_memory_allocator_test.cc
xiaomings: Done

starboard/common/bidirectional_fit_reuse_allocator_test.cc
xiaomings: The value is increased as the original one is no longer enough for test case `DecommitOnBatchedFree`.  I reduce it to 8 MB.

I can add a comment to it, on the other hand, quoting the test case `DecommitOnBatchedFree` here increases future maintenance effort and it's test code.  So please feel free to let me know if you prefer so.

media/starboard/bidirectional_fit_decoder_buffer_allocator_strategy.h
```diff
@@ -27,11 +27,14 @@ class BidirectionalFitDecoderBufferAllocatorStrategy
  public:
   BidirectionalFitDecoderBufferAllocatorStrategy(
       std::size_t initial_capacity,
-      std::size_t allocation_increment)
-      : birectional_fit_allocator_(&fallback_allocator_,
+      std::size_t allocation_increment,
+      bool enable_decommit_on_idle)
+      : fallback_allocator_(enable_decommit_on_idle),
+        birectional_fit_allocator_(&fallback_allocator_,
```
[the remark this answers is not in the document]
xiaomings: This should be addressed in a separate PR.

media/starboard/starboard_memory_allocator.h
```diff
@@ -16,21 +16,42 @@
...
+
   void* Allocate(std::size_t size) override { return Allocate(size, 1); }
 
   void* Allocate(std::size_t size, std::size_t alignment) override {
+    if (enable_decommit_) {
+      alignment = std::max(alignment, page_size_);
+      size = AlignUp(size, page_size_);
+    }
     void* p = nullptr;
     posix_memalign(&p, std::max(alignment, sizeof(void*)), size);
```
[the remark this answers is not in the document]
xiaomings: The function is expected to return nullptr on allocation failure.

media/starboard/starboard_memory_allocator.h
```diff
@@ -16,21 +16,42 @@
...
+    SB_LOG(INFO) << "StarboardMemoryAllocator: decommit is "
+                 << (enable_decommit_ ? "enabled" : "disabled") << ".";
+  }
+
   void* Allocate(std::size_t size) override { return Allocate(size, 1); }
 
   void* Allocate(std::size_t size, std::size_t alignment) override {
+    if (enable_decommit_) {
+      alignment = std::max(alignment, page_size_);
+      size = AlignUp(size, page_size_);
```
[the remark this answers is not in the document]
xiaomings: The code isn't supposed to be used for sizes close to limit.  Improving AlignUp() is not in the scope of this PR.

kjyoun (approved):

johnxwork (approved):

PR #9287, opened 2026-02-27
Enable Siso on builds

briantting:
Enables Siso as a drop-in replacement for Ninja. Sets up Siso backend
config via gclient sync and Siso toggles via GN. Deprecates
reclient_cfgs.

Force SISO to conditionally include musl_all_headers

When the use_custom_libc gn arg is true.

Bug: 454374226


build/config/siso/configure_siso.py
```diff
@@ -16,6 +16,11 @@
 
 _BACKEND_STAR = os.path.join(THIS_DIR, "backend_config", "backend.star")
 
+_COBALT_STAR = os.path.join(THIS_DIR, "backend_config", "cobalt.star")
+_KNOWN_COBALT_PROJECTS = (
+    'cobalt-actions-prod',
+)
+
 _GOOGLE_STAR = os.path.join(THIS_DIR, "backend_config", "google.star")
 _KNOWN_GOOGLE_PROJECTS = (
```
oxve, line 25: Can you add it to this list instead? Seems like it would be equivalent.

briantting: The problem is that we don't have a "large" worker type in our rbe cluster only "default"

briantting:
Having to include musl files in a SISO includes filegroup may be masking musl BUILD.gn hermeticity issues and be a hidden dependency.

build/config/siso/cobalt.star
```diff
@@ -0,0 +1,423 @@
...
+
+def __filegroups(ctx):
+    args = gn.args(ctx)
+    if args.get("use_custom_libc") == "false":
+        return {}
+
+    fg = {
+        "third_party/musl:musl_all_headers": {
+            "type": "glob",
+            "includes": [
```
kaidokert, line 21: Ugh. this looks gnarly. Is there no cleaner way to do this ? I'm expecting breakages to musl will be very surprising to people, as nobody will know to look here.

build/config/siso/backend_config/cobalt.star
```diff
@@ -0,0 +1,22 @@
+# -*- bazel-starlark -*-
+# Copyright 2024 The Chromium Authors
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+"""Siso backend config for Cobalt."""
+
+load("@builtin//struct.star", "module")
+
+def __platform_properties(ctx):
+    container_image = "docker://gcr.io/chops-public-images-prod/rbe/siso-chromium/linux@sha256:ef35d347f4a4a2d32b76fd908e66e96f59bf8ba7379fd5626548244c45343b2b"
```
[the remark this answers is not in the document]
kaidokert: Concur, this would be cleaner

Also would be helpful to have a comment above the version/container explaining how to obtain it or where does that value come from.

cobalt/reclient_cfgs/chromium-browser-clang/rewrapper_linux.cfg
```diff
@@ -1,8 +0,0 @@
-platform=container-image=docker://gcr.io/chops-public-images-prod/rbe/siso-chromium/linux@sha256:26de99218a1a8b527d4840490bcbf1690ee0b55c84316300b60776e6b3a03fe1,label:action_default=1
```
kaidokert, line 1: Should we delete them in this PR or just in case give people some time to switch over, and come back a couple weeks later ? I'm not sure if it helps anything

build/config/siso/cobalt.star
briantting: i can use * to help denote files, ** doesn't work unfortunately. This filelist is just a 1 for 1 copy from (https://source.corp.google.com/h/github/youtube/cobalt/+/main:third_party/musl/BUILD.gn;l=25)

briantting:
> Overall lgtm, but the musl-specific include file looks really bad. Quite a few mods to upstream too - ideally there would be a way to avoid them.

All the mods to build/config/siso/configure_siso.py can be reverted once it's on m144. The 3 lines to main.star are unavoidable for now, but that is a relatively limited upstream mod.

build/config/siso/backend_config/cobalt.star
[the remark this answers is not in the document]
briantting: Added a more detailed description.

cobalt/reclient_cfgs/chromium-browser-clang/rewrapper_linux.cfg
briantting: Since this is a drop in replacement, we should be able to switch over right away.

I also don't think, given current priorities, main's workflow needs to worry about the changes (compared to 26.eap and 26.android where devs probably would prefer no workflow changes at the moment).

build/config/siso/cobalt.star
briantting: added more comments for future reference

PR #9288, opened 2026-02-27
h5vcc: Use base::span for data transport in H5vccPlatformService

rakuco:
Use a ReadOnlyBuffer mojom type to represent data rather than an array, as this makes the C++ bindings use base::span instead of std::vector.

Doing so allows for less duplication and fewer manual conversions, as DOMArrayBuffer can deal with spans natively, and some simplification on the browser side as well. While here, const-ify and use `std::unique_ptr` with `base::FreeDeleter` on `response_ptr` to avoid manual memory management.

Tested with the softmic extension on Linux.

Fixed: 487712437

### DIFF
--- a/cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
+++ b/cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
@@ -21,6 +21,7 @@
 #include "base/functional/bind.h"
 #include "base/logging.h"
 #include "base/memory/free_deleter.h"
+#include "base/numerics/safe_conversions.h"
 #include "base/task/bind_post_task.h"
 #include "content/public/browser/browser_task_traits.h"
 #include "content/public/browser/browser_thread.h"
@@ -156,8 +157,8 @@ void PlatformServiceImpl::Send(base::span<const uint8_t> data,
     return;
   }
 
-  std::move(callback).Run(
-      base::span<const uint8_t>(response_ptr.get(), output_length));
+  std::move(callback).Run(base::span<const uint8_t>(
+      response_ptr.get(), base::checked_cast<size_t>(output_length)));
 }
 
 void PlatformServiceImpl::OnDataReceivedFromStarboard(


cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
```diff
@@ -140,31 +142,29 @@ void PlatformServiceImpl::Send(const std::vector<uint8_t>& data,
...
-  if (response_ptr && output_length > 0) {
-    const uint8_t* response_bytes = static_cast<const uint8_t*>(response_ptr);
-    response_data =
-        std::vector<uint8_t>(response_bytes, response_bytes + output_length);
-  }
-  free(response_ptr);
-
-  std::move(callback).Run(response_data);
+  std::move(callback).Run(
+      base::span<const uint8_t>(response_ptr.get(), output_length));
```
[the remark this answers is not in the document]
rakuco: done

rakuco:

hlwarriner:
I was waiting for resolution on #9268 before reviewing this current PR. I see that #9268 was merged this morning but I may revert it. I'll follow up here soon.

rakuco:
PR rebased after #9268 since it looks like it will stick.

cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
```diff
@@ -142,26 +145,20 @@ void PlatformServiceImpl::Send(const std::vector<uint8_t>& data,
...
-  std::vector<uint8_t> response_data;
-  if (response_ptr && output_length > 0) {
-    const uint8_t* response_bytes = static_cast<const uint8_t*>(response_ptr);
-    response_data =
-        std::vector<uint8_t>(response_bytes, response_bytes + output_length);
-  }
-  free(response_ptr);
-
-  std::move(callback).Run(response_data);
+  std::move(callback).Run(base::span<const uint8_t>(
```
hlwarriner, line 160: I noticed a small bug with our existing H5vccPlatformService implementation.

Per https://github.com/youtube/cobalt/blob/main/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.idl#L41, the `send()` web API should return `null`, instead of an empty `ArrayBuffer`, if the platform service declines to synchronously return data. But with our current implementation, an empty `ArrayBuffer` will be returned in this case.

From what I can tell, this current PR doesn't change the bug or make it any more difficult to fix. So I'll plan to just file a separate bug and fix it independently from this PR.

Please lmk if you agree with that assessment.

hlwarriner: @zhongqiliang , wdyt?

zhongqiliang: All of the existing platform services do not have the use case to return empty response. It will be good to take care of the edge case though.

hlwarriner: Sounds good, thanks. I filed b/493975470 to track this.

hlwarriner: @zhongqiliang, please do take a look at this overall PR as well.

PR #9299, opened 2026-02-27
Add StorageManager histograms

haozheng-cobalt:
Introduce UMA histograms to monitor the Cobalt storage migration process.
These metrics track the outcome of reading legacy Starboard storage,
the success of injecting data into Chromium storage, and the counts of
cookies and local storage items migrated. Also record the total
duration of the migration. This helps assess the reliability and
performance of the storage migration in production environments.

Bug: 484055661

### DIFF
--- a/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
+++ b/cobalt/shell/browser/migrate_storage_record/migration_manager.cc
@@ -22,6 +22,7 @@
 #include "base/containers/adapters.h"
 #include "base/files/file_path.h"
 #include "base/files/file_util.h"
+#include "base/metrics/histogram_functions.h"
 #include "base/path_service.h"
 #include "base/strings/utf_string_conversions.h"
 #include "base/task/thread_pool.h"
@@ -45,6 +46,40 @@ namespace migrate_storage_record {
 
 namespace {
 
+// These values are persisted to logs. Entries should not be renumbered and
+// numeric values should never be reused.
+// Tracks the outcome of attempting to read the legacy Starboard storage record.
+enum class StorageReadResult {
+  kSuccess = 0,
+  kPartitionKeyNotDefault = 1,
+  kRecordInvalid = 2,
+  kSizeTooSmall = 3,
+  kDeleteFailed = 4,
+  kReadMismatch = 5,
+  kHeaderMismatch = 6,
+  kParseError = 7,
+  kMaxValue = kParseError,
+};
+
+// Tracks the outcome of individual data injection operations into the new
+// Chromium-based storage partitions.
+enum class InjectionResult {
+  kSuccess = 0,
+  kError = 1,
+  kMaxValue = kError,
+};
+
+constexpr char kReadResultHistogram[] = "Cobalt.StorageMigration.ReadResult";
+constexpr char kCookieInjectionResultHistogram[] =
+    "Cobalt.StorageMigration.CookieInjectionResult";
+constexpr char kLocalStorageInjectionResultHistogram[] =
+    "Cobalt.StorageMigration.LocalStorageInjectionResult";
+constexpr char kCookiesCountHistogram[] =
+    "Cobalt.StorageMigration.CookiesCount";
+constexpr char kLocalStorageCountHistogram[] =
+    "Cobalt.StorageMigration.LocalStorageCount";
+constexpr char kDurationHistogram[] = "Cobalt.StorageMigration.Duration";
+
 std::string GetApplicationKey(const GURL& url) {
   std::string encoded_url;
   base::Base64UrlEncode(url_matcher::util::Normalize(url).spec(),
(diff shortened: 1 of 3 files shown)


cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -67,18 +88,24 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
     record->Delete();
     bool fallback = partition_key == GetApplicationKey(GURL(kDefaultURL));
     if (!fallback) {
+      base::UmaHistogramEnumeration("Cobalt.StorageMigration.ReadResult",
```
Awallky: Could probably move this string to a const char* or const std::string and reuse it across the callsites.
Also, for new histograms, be sure to follow http://shortn/_8QPjyFX1pJ for best practices when adding histograms to Cobalt.

haozheng-cobalt: Good tips, thanks Adam!

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -67,18 +100,24 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
     record->Delete();
     bool fallback = partition_key == GetApplicationKey(GURL(kDefaultURL));
     if (!fallback) {
+      base::UmaHistogramEnumeration(kReadResultHistogram,
+                                    StorageReadResult::kRecordInvalid);
```
Awallky: Should we differentiate between the record and fallback record here?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -45,6 +46,38 @@ namespace migrate_storage_record {
...
+// Tracks the outcome of individual data injection operations into the new
+// Chromium-based storage partitions.
+enum class InjectionResult {
+  kSuccess = 0,
+  kCookieSetError = 1,
+  kLocalStorageSetError = 2,
+  kMaxValue = kLocalStorageSetError,
+};
+
+constexpr char kReadResultHistogram[] = "Cobalt.StorageMigration.ReadResult";
```
Awallky, line 72: You'll want to add these to [tools/metrics/histograms/metadata/cobalt/histograms.xml](https://github.com/youtube/cobalt/tree/main/tools/metrics/histograms/metadata/cobalt/histograms.xml) in order to get them in the histograms dashboard. Otherwise, we will have to query tables directly to get the same information.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
haozheng-cobalt: yes, that will be better, amended.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -57,55 +92,75 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
...
-      std::make_unique<starboard::StorageRecord>(partition_key.c_str());
-  if (!record->IsValid()) {
-    record->Delete();
-    bool fallback = partition_key == GetApplicationKey(GURL(kDefaultURL));
-    if (!fallback) {
-      return nullptr;
+  StorageReadResult result = StorageReadResult::kSuccess;
+  std::unique_ptr<cobalt::storage::Storage> storage;
+
+  [&]() {
```
johnxwork, line 98: What's the reason to change this to a lambda function?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -57,55 +92,75 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
...
-  }
+    storage = std::make_unique<cobalt::storage::Storage>();
+    if (!storage->ParseFromArray(
+            reinterpret_cast<const char*>(bytes.data() + kRecordHeaderSize),
+            bytes.size() - kRecordHeaderSize)) {
+      result = StorageReadResult::kParseError;
+      storage.reset();
+      return;
+    }
+  }();
```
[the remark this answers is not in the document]
johnxwork: yeah please do this...

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
haozheng-cobalt: This is to break the nested conditions, to assign value for the StorageReadResult result, as the `base::UmaHistogramEnumeration(kReadResultHistogram, result);` can only be set once according to the histogram programming convention.

haozheng-cobalt: Before this change, the function `ReadStorage()` returns at any of the if conditions. Now when we add histogram logging, we can no longer simply return at any if condition, because base::UmaHistogramEnumeration(kReadResultHistogram, result); can only be set once in the code, this is the rule of histogram. So we need to remember the value of  StorageReadResult, and set base::UmaHistogramEnumeration(kReadResultHistogram, result); after all the if checks.

This IIFE lambda is basically used as a switch condition with minimal change to the existing logics in that function.

PR #9313, opened 2026-03-02
tvOS: Enable Platform Service extension

jkim-julie:
This enables PlatformService cobalt extension and adds ‘dev.cobalt.coat.clientloginfo’ which has the same behaviour with C25.

Bug: 487879114

### DIFF
--- a/cobalt/shell/browser/shell_platform_delegate_ios.mm
+++ b/cobalt/shell/browser/shell_platform_delegate_ios.mm
@@ -22,6 +22,7 @@
 #include "base/trace_event/trace_config.h"
 #include "cobalt/shell/app/resource.h"
 #include "cobalt/shell/browser/shell.h"
+#include "content/public/browser/tvos/simple_begin_frame_observer.h"
 #include "content/public/browser/visibility.h"
 #include "content/public/browser/web_contents.h"
 #include "services/tracing/public/cpp/perfetto/perfetto_config.h"
@@ -138,7 +139,6 @@ const char kAllTracingCategories[] = "*";
                           blue:53.0 / 255.0
                          alpha:1.0];
 }
-
 #if BUILDFLAG(IS_IOS_TVOS)
 // Intercept UIPressTypeMenu event and do not forward it to the
 // superclass's pressesBegan method, as it will cause the application to
@@ -175,6 +175,7 @@ const char kAllTracingCategories[] = "*";
     [super pressesEnded:nonMenuPresses withEvent:event];
   }
 }
+
 #endif  // BUILDFLAG(IS_IOS_TVOS)
 
 - (void)viewDidLoad {
@@ -707,6 +708,23 @@ bool ShellPlatformDelegate::DestroyShell(Shell* shell) {
 
   [shell_data.window resignKeyWindow];
   return true;  // The performClose() will do the destruction of Shell.
+}
+
+void ShellPlatformDelegate::SetVisible(WebContents* web_contents,
+                                       bool visible) {
+  if (!visible) {
+    return;
+  }
+
+  auto* observer = SimpleBeginFrameObserver::FromWebContents(web_contents);
+  if (!observer) {
+    return;
+  }
+
+  auto on_begin_frame_callback = base::BindRepeating([](double refresh_rate) {
+    [SBDGetApplication() updateLastDisplayRefreshRate:refresh_rate];
+  });
+  observer->StartObservingBeginFrame(on_begin_frame_callback);
 }
 
 void ShellPlatformDelegate::ToggleFullscreenModeForTab(
--- a/starboard/tvos/shared/application_darwin.mm
+++ b/starboard/tvos/shared/application_darwin.mm
@@ -49,18 +49,34 @@
   // Used for checking that certain methods are invoked from the UI thread (so
   // that UIKit calls can be made directly, for example).
   starboard::ThreadChecker _uiThreadChecker;
+
+  // The maximum number of frames per second a screen can render.
+  NSInteger _maximumFramesPerSecond;
+
+  // The last display refresh rate.
+  double _lastDisplayRefreshRate;
 }
 
 @synthesize drmManager = _drmManager;
 @synthesize playerManager = _playerManager;
+@synthesize maximumFramesPerSecond = _maximumFramesPerSecond;
 
 - (instancetype)init {
   self = [super init];
   if (self) {
     _drmManager = [[SBDDrmManager alloc] init];
     _playerManager = [[SBDPlayerManager alloc] init];
+    _maximumFramesPerSecond = [[UIScreen mainScreen] maximumFramesPerSecond];
   }
   return self;
+}
+
+- (void)updateLastDisplayRefreshRate:(double)refreshRate {
+  _lastDisplayRefreshRate = refreshRate;
+}
+
+- (double)displayRefreshRate {
+  return _lastDisplayRefreshRate;
 }
 
 - (void)setPlayerContainerView:(UIView*)view {
--- a/starboard/tvos/shared/platform_service.mm
+++ b/starboard/tvos/shared/platform_service.mm
@@ -1,0 +1,117 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/tvos/shared/platform_service.h"
+
+#import <UIKit/UIKit.h>
+
+#include "starboard/common/log.h"
+#include "starboard/common/string.h"
+#include "starboard/extension/platform_service.h"
+#import "starboard/tvos/shared/starboard_application.h"
+
+typedef struct CobaltExtensionPlatformServicePrivate {
+  void* Send(void* /*data*/,
+             uint64_t /*length*/,
+             uint64_t* output_length,
+             bool* invalid_state) {
+    if (!output_length || !invalid_state) {
+      return nullptr;
+    }
+
+    *invalid_state = false;
+    const std::string string = starboard::FormatString(
+        "MaximumFramesPerSecond:%ld;DisplayRefreshRate:%f;",
+        SBDGetApplication().maximumFramesPerSecond,
+        SBDGetApplication().displayRefreshRate);
+
+    // Other extensions do not include a NUL-terminating byte. This is done here
+    // for compatibility with C25.
+    *output_length = string.size() + 1;
+    auto* output = malloc(*output_length);
+    if (!output) {
+      *invalid_state = true;
+      return nullptr;
+    }
+
+    memcpy(output, string.c_str(), *output_length);
+    return output;  // caller must free()
+  }
+
+} CobaltExtensionPlatformServicePrivate;
+
+namespace starboard {
+
+namespace {
+
+constexpr std::string_view kClientLogInfoServiceName =
+    "dev.cobalt.coat.clientloginfo";
+
+bool Has(const char* name) {
+  if (!name) {
+    return false;
+  }
+
+  return name == kClientLogInfoServiceName;
+}
+
+CobaltExtensionPlatformService Open(
+    void* /*context*/,
+    const char* name,
+    ReceiveMessageCallback /*receive_callback*/) {
+  if (!Has(name)) {
+    SB_LOG(ERROR) << "Can't open Service " << name;
+    return kCobaltExtensionPlatformServiceInvalid;
+  }
+
+  CobaltExtensionPlatformService service =
+      new CobaltExtensionPlatformServicePrivate();
+  return service;
+}
+
+void Close(CobaltExtensionPlatformService service) {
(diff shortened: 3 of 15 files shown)


starboard/tvos/shared/application_darwin.mm
```diff
@@ -60,9 +66,26 @@ - (instancetype)init {
     _drmManager = [[SBDDrmManager alloc] init];
     _playerManager = [[SBDPlayerManager alloc] init];
   }
+  // Create a display link and add to current loop.
+  _displayLink = [CADisplayLink displayLinkWithTarget:self
+                                             selector:@selector(display:)];
+  [_displayLink addToRunLoop:[NSRunLoop currentRunLoop]
+                     forMode:NSRunLoopCommonModes];
```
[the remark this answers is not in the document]
jkim-julie: Done.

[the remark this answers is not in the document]
jkim-julie: Done.

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,124 @@
...
+             bool* invalid_state) {
+    if (!output_length || !invalid_state) {
+      return NULL;
+    }
+
+    *invalid_state = false;
+    std::string string = starboard::FormatString(
+        "MaximumFramesPerSecond:%ld;DisplayRefreshRate:%f;",
+        UIScreen.mainScreen.maximumFramesPerSecond,
+        SBDGetApplication().displayRefreshRate);
```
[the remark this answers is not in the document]
jkim-julie: Done.

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,124 @@
...
+#import <UIKit/UIKit.h>
+
+#include "starboard/common/log.h"
+#include "starboard/common/string.h"
+#include "starboard/extension/platform_service.h"
+#import "starboard/tvos/shared/starboard_application.h"
+
+typedef struct CobaltExtensionPlatformServicePrivate {
+  void* Send(void* data,
+             uint64_t length,
```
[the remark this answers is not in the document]
jkim-julie: Done.

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,124 @@
...
+
+  delete static_cast<CobaltExtensionPlatformServicePrivate*>(service);
+}
+
+void* Send(CobaltExtensionPlatformService service,
+           void* data,
+           uint64_t length,
+           uint64_t* output_length,
+           bool* invalid_state) {
+  SB_DCHECK(data);
```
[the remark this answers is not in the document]
jkim-julie: Done.

jkim-julie:

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+bool Has(const char* name) {
+  if (!name) {
+    return false;
+  }
+
+  if (strcmp(name, kClientLogInfoServiceName) == 0) {
+    return true;
+  }
+
+  return false;
```
rakuco: Let's take the opportunity and use more C++ here:

```c++
constexpr std::string_view kClientLogInfoServiceName = "...";

bool Has(const char* name) {
  if (!name) {
    return false;
  }
  return name == kClientLogInfoServiceName;
}
```

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+#include "starboard/tvos/shared/platform_service.h"
+
+#import <UIKit/UIKit.h>
+
+#include "starboard/common/log.h"
+#include "starboard/common/string.h"
+#include "starboard/extension/platform_service.h"
+#import "starboard/tvos/shared/starboard_application.h"
+
+typedef struct CobaltExtensionPlatformServicePrivate {
```
rakuco, line 24: I don't think this needs to exist at all, `context` and `receive_callback` are not used anywhere, so `Send()`'s implementation could be merged directly into `Send()` below.

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+  void* Send(void* /*data*/,
+             uint64_t /*length*/,
+             uint64_t* output_length,
+             bool* invalid_state) {
+    if (!output_length || !invalid_state) {
+      return NULL;
```
rakuco: s/NULL/nullptr/

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    *invalid_state = false;
+    std::string string = starboard::FormatString(
+        "MaximumFramesPerSecond:%ld;DisplayRefreshRate:%f;",
+        SBDGetApplication().maximumFramesPerSecond,
+        SBDGetApplication().displayRefreshRate);
+
+    *output_length = string.size() + 1;
+    char* output = (char*)malloc(*output_length);
```
rakuco:
```suggestion
    auto* output = malloc(*output_length);
```

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+    if (!output) {
+      *invalid_state = true;
+      return NULL;
```
rakuco:
```suggestion
      return nullptr;
```

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    *invalid_state = false;
+    std::string string = starboard::FormatString(
+        "MaximumFramesPerSecond:%ld;DisplayRefreshRate:%f;",
+        SBDGetApplication().maximumFramesPerSecond,
+        SBDGetApplication().displayRefreshRate);
+
+    *output_length = string.size() + 1;
```
rakuco, line 41:
```suggestion
    // Other extensions do not include a NUL-terminating byte. This is done here for compatibility with C25.
    *output_length = string.size() + 1;
```

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,121 @@
...
+  void* Send(void* /*data*/,
+             uint64_t /*length*/,
+             uint64_t* output_length,
+             bool* invalid_state) {
+    if (!output_length || !invalid_state) {
+      return NULL;
+    }
+
+    *invalid_state = false;
+    std::string string = starboard::FormatString(
```
rakuco:
```suggestion
    const std::string string = starboard::FormatString(
```

starboard/tvos/shared/platform_service.mm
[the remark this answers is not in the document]
rakuco: This comment applies to C25 and other platforms, but on Chrobalt+tvOS I'm pretty sure this is running in the UI thread.

What you could do anyway is use GCD to ensure this runs in the UI thread. Something like

```c++
NSInteger maximum_frames_per_second;
dispatch_sync(dispatch_get_main_queue(), ^{
  maximum_frames_per_second = [[UIScreen mainScreen] maximumFramesPerSecond];
});
const std::string string = ...;
```

cobalt/shell/browser/shell_platform_delegate_ios.mm
```diff
@@ -183,6 +187,29 @@ - (void)pressesEnded:(NSSet<UIPress*>*)presses
     [super pressesEnded:nonMenuPresses withEvent:event];
   }
 }
+
+- (void)update:(CADisplayLink*)sender {
```
rakuco: I haven't tried it out myself, but I think it should be possible to plug into https://source.chromium.org/chromium/chromium/src/+/main:components/viz/common/frame_sinks/external_begin_frame_source_ios.mm which already sets up a CADisplayLink, performs the `interval` calculation and dispatches it.

If you subclass [`SimpleBeginFrameObserver`](https://source.chromium.org/chromium/chromium/src/+/main:ui/compositor/host_begin_frame_observer.h;l=29;drc=56c66e417c83e2096a4e4e8a5c4ab7bbd525c9f3), you should be able to retrieve `interval` as `frame_interval`. If this works, you can just send this result to `SBDGetApplication()` and perform the division when the value is requested, as I don't think Kabuki will need this on every frame update.

starboard/tvos/shared/application_darwin.mm
```diff
@@ -49,20 +49,36 @@ @implementation ObjCApplication {
   // Used for checking that certain methods are invoked from the UI thread (so
   // that UIKit calls can be made directly, for example).
   starboard::ThreadChecker _uiThreadChecker;
+
+  // The maximum number of frames per second a screen can render.
+  NSInteger _maximumFramesPerSecond;
+
+  // The last display refresh rate.
+  std::atomic<double> _lastDisplayRefreshRate;
```
rakuco: I don't think this needs to be an atomic variable, especially if you use `PostTask()` from cobalt shell to make the writes sequential.

starboard/tvos/shared/platform_service.mm
jkim-julie: Since CobaltExtensionPlatformServicePrivate is the type used globally, https://github.com/youtube/cobalt/blob/main/starboard/extension/platform_service.h#L26, and `Open()` should return it. So, I think we still need it.

Instead, I cleaned `context` and `receive_callback` not used on tvOS from the structure.

starboard/tvos/shared/platform_service.mm
jkim-julie: Done

starboard/tvos/shared/platform_service.mm
jkim-julie: Done

starboard/tvos/shared/platform_service.mm
jkim-julie: Done.

starboard/tvos/shared/platform_service.mm
jkim-julie: Done.

starboard/tvos/shared/platform_service.mm
jkim-julie: Done

starboard/tvos/shared/application_darwin.mm
jkim-julie: Done

cobalt/shell/browser/shell_platform_delegate_ios.mm
jkim-julie: Indeed.
In the updated patch, I used SimpleBeginFrameObserver.
In order to utilize it,
- the observer is created after ui::Compositor is created at RenderWidgetHostViewTVOS::ShowWithVisibility.
- In ShellPlatformDelegate::SetVisible, the observing is started with a callback that sets the refresh rate to the SBDGetApplication.
- I created the observer files at //content/public to use it at //content/browser and //cobalt/shell
PTAL.

starboard/tvos/shared/platform_service.mm
```diff
@@ -0,0 +1,117 @@
...
+    auto* output = malloc(*output_length);
+    if (!output) {
+      *invalid_state = true;
+      return nullptr;
+    }
+
+    memcpy(output, string.c_str(), *output_length);
+    return output;  // caller must free()
+  }
+
```
rakuco, line 51: style nit: extra empty line here

starboard/tvos/shared/application_darwin.mm
```diff
@@ -49,20 +49,36 @@ @implementation ObjCApplication {
...
 @synthesize drmManager = _drmManager;
 @synthesize playerManager = _playerManager;
+@synthesize maximumFramesPerSecond = _maximumFramesPerSecond;
 
 - (instancetype)init {
   self = [super init];
   if (self) {
     _drmManager = [[SBDDrmManager alloc] init];
     _playerManager = [[SBDPlayerManager alloc] init];
+    _maximumFramesPerSecond = [[UIScreen mainScreen] maximumFramesPerSecond];
```
rakuco, line 69: Doesn't [this](https://github.com/youtube/cobalt/pull/9313#discussion_r2871796066) work to get rid of this property?

Another option that's probably better:
* Try calling `ui::Display::set_display_frequency()` in [`ScreenIos::ScreenChanged()`](https://github.com/youtube/cobalt/blob/bfc9423141fb146190ad9d1b7f22c4c3ddae7e38/ui/display/ios/screen_ios.mm#L80), [like the Mac version does](https://source.chromium.org/chromium/chromium/src/+/main:ui/display/mac/screen_mac.mm;l=185;drc=c04092ac310fd5203d2c607443617dd44f978cea).
* Remove `_maximumFramesPerSecond` and related code
* Call `ui::Screen::Get()->GetPrimaryDisplay().display_frequency()` in the platform service code.

starboard/tvos/shared/starboard_application.h
```diff
@@ -58,6 +58,16 @@ id<SBDStarboardApplication> SBDGetApplication(void);
  */
 @property(nonatomic, readonly) SBDPlayerManager* playerManager;
 
+/**
```
rakuco, line 61: Nit: these Doxygen-style comments were used in C25. In Chrobalt, it makes more sense to stick to Chromium's style.

starboard/tvos/shared/starboard_application.h
```diff
@@ -78,6 +88,8 @@ id<SBDStarboardApplication> SBDGetApplication(void);
 // suspending the application.
 - (void)registerMenuPressEnded:(UIPress*)press
                   pressesEvent:(UIPressesEvent*)pressesEvent;
+
+- (void)updateLastDisplayRefreshRate:(double)lastDisplayRefreshRate;
```
rakuco, line 92: Please add some documentation to this method.

jkim-julie:
Thanks a lot for the all comments!
Since it turned out we don't need this extension according to @fayev-g's comment on the Chat, I close this MR.

PR #8706, opened 2026-01-16
Prepend starboard OS name in user-agent string

santoshmahto7:
The current OS name in user-agent string is fetched using
chromium code which just return "Linux". This misses the real
OS name for starboard platform and result in YTS test failure.
So OS name in user-agent string should include the starboard platform
OS name.
 
Bug: 475670748

PR #9314, opened 2026-03-02
Prepend starboard OS name in user-agent string (#8706)

santoshmahto7:
The current OS name in user-agent string is fetched using chromium code which just return "Linux". This misses the real OS name for starboard platform and result in YTS test failure. So OS name in user-agent string should include the starboard platform OS name.

Bug: 475670748

(cherry picked from commit 0d02587505ef402fbe97471fd8fea1b5e53bd991) Change-Id: Idef1203f6d54668549c7154cdeba9ef3633f5d0a


santoshmahto7:
@niranjanyardi This passes now, could you land this

niranjanyardi (approved):

PR #9316, opened 2026-03-02
cobalt: Build crashpad_database_util on all platforms

rakuco:
Follow-up to #9270: the crashpad_database_util tool can be built on non-Evergreen platforms as well, and it is useful to inspect Crashpad databases.

Expand the checks in the gn_all target so that it is built on all platforms with the appropriate toolchain while restricting the building of the crashpad_handler binary to Evergreen platforms.

Bug: 477518757

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -70,20 +70,20 @@ group("gn_all") {
     deps += [ "//chrome/updater/cobalt:updater_sandbox" ]
   }
   if (use_crashpad) {
+    # On Starboard platforms, build both the crashpad_handler binary and the
+    # crashpad_database_util tool. On other platforms (e.g. ATV, tvOS) it is
+    # still useful to build the crashpad_database_util binary to inspect crash
+    # databases on the host.
     if (is_starboard) {
       deps += [
         "//third_party/crashpad/crashpad/handler:crashpad_handler($native_target)",
         "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",
       ]
+    } else if (is_ios) {
+      deps += [ "//third_party/crashpad/crashpad/tools:crashpad_database_util($host_toolchain)" ]
     } else {
-      # On other platforms (e.g. ATV, tvOS) it is still useful to build the
-      # crashpad_database_util binary to inspect crash databases on the host.
-      if (is_ios) {
-        deps += [ "//third_party/crashpad/crashpad/tools:crashpad_database_util($host_toolchain)" ]
-      } else {
-        deps +=
-            [ "//third_party/crashpad/crashpad/tools:crashpad_database_util" ]
-      }
+      # This applies to e.g. ATV, where the default toolchain is the host one.
+      deps += [ "//third_party/crashpad/crashpad/tools:crashpad_database_util" ]
     }
   }
 }


rakuco:

cobalt/BUILD.gn
```diff
@@ -69,11 +69,22 @@ group("gn_all") {
...
-      "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",
-    ]
+  if (use_crashpad) {
+    if (native_target != "") {
+      deps += [
+        "//third_party/crashpad/crashpad/handler:crashpad_handler($native_target)",
+        "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",
+      ]
+    } else {
+      # On other platforms (e.g. ATV, tvOS) it is still useful to build the
```
hlwarriner: Can you please move this comment to above line 82 and modify it to be something like:
```
# On other platforms (e.g. tvOS) it is still useful to build the
# crashpad_database_util binary to inspect crash database on the host
# after it's been pulled from the target device.
```

1. Clarify that the crashpad db is still on the target device for iOS, and must be copied from it.
2. For platforms that build `crashpad_database_util` with the default toolchain, I think the idea would be that it can somehow be pushed to the target device and run on it. I think this is actually possible for ATV.

CC @alunev - it may be useful to build this target for ATV whenever the HangWatcher changes are cherry picked to `main`. I think the `use_crashpad` build variable would prevent that from happening, but we can  make a change here at a later time.

cobalt/BUILD.gn
```diff
@@ -69,11 +69,22 @@ group("gn_all") {
   if (is_starboard && build_updater_sandbox) {
     deps += [ "//chrome/updater/cobalt:updater_sandbox" ]
   }
-  if (use_evergreen && use_crashpad) {
-    deps += [
-      "//third_party/crashpad/crashpad/handler:crashpad_handler($native_target)",
-      "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",
-    ]
+  if (use_crashpad) {
+    if (native_target != "") {
```
andrewsavage1: Would this check (`native_target != ""`) be equivalent to `is_starboard` (or is there another variable we could use in its place)? I can understand the logic of choosing the right toolchain to build the crashpad_database_util target with, but conditionally depending on the crashpad_handler if `native_target != ""` is a case that's harder for me to wrap my head around

rakuco: > Would this check (`native_target != ""`) be equivalent to `is_starboard` (or is there another variable we could use in its place)?

Good question. The logic that sets `native_target` is located [here](https://github.com/youtube/cobalt/blob/7c06df173d9082ca2e403e574303bdec5645ee28/cobalt/build/configs/initialize_variables.gni#L72), and it depends on both `starboard_toolchain` and `build_with_separate_cobalt_toolchain`. I'm not very familiar with all these different variables myself, so I'm wondering the same thing :-)

andrewsavage1: It should be okay to use `is_starboard` as all Starboard platforms require a native_target toolchain. I don't think we would have cases of non-starboard platforms with a native_target toolchain.

rakuco: done!

cobalt/BUILD.gn
```diff
@@ -69,11 +69,22 @@ group("gn_all") {
...
+  if (use_crashpad) {
+    if (is_starboard) {
+      deps += [
+        "//third_party/crashpad/crashpad/handler:crashpad_handler($native_target)",
+        "//third_party/crashpad/crashpad/tools:crashpad_database_util($native_target)",
+      ]
+    } else {
+      # On other platforms (e.g. ATV, tvOS) it is still useful to build the
+      # crashpad_database_util binary to inspect crash databases on the host.
+      if (is_ios) {
```
andrewsavage1: Consider flattening this conditional i.e. `if (is_starboard) ... else if (is_ios) ... else`

rakuco: good point, done!

PR #9327, opened 2026-03-02
starboard: Consolidate experimental feature into a struct

kjyoun:
Consolidate various experimental player features into a single
PlayerComponents::ExperimentalFeatures struct. This refactoring groups
parameters such as flush_decoder_during_reset, reset_audio_decoder,
and video decoder input/preroll counts.

This change improves code organization by reducing the number of
individual parameters in constructors and centralizing the management
of experimental player settings.

Bug: 480134029

### DIFF
--- a/starboard/shared/starboard/player/filter/player_components.cc
+++ b/starboard/shared/starboard/player/filter/player_components.cc
@@ -83,7 +83,9 @@ int AlignUp(int value, int alignment) {
 PlayerComponents::Factory::CreationParameters::CreationParameters(
     const media::AudioStreamInfo& audio_stream_info,
     SbDrmSystem drm_system)
-    : audio_stream_info_(audio_stream_info), drm_system_(drm_system) {
+    : audio_stream_info_(audio_stream_info),
+      experimental_features_(ExperimentalFeatures{}),
+      drm_system_(drm_system) {
   SB_DCHECK_NE(audio_stream_info_.codec, kSbMediaAudioCodecNone);
 }
 
@@ -92,15 +94,7 @@ PlayerComponents::Factory::CreationParameters::CreationParameters(
     SbPlayer player,
     SbPlayerOutputMode output_mode,
     int max_video_input_size,
-    bool flush_decoder_during_reset,
-    bool reset_audio_decoder,
-    std::optional<int> video_initial_max_frames_in_decoder,
-    std::optional<int> video_max_pending_input_frames,
-    std::optional<int> video_decoder_initial_preroll_count,
-    std::optional<int> video_decoder_poll_interval_ms,
-    std::optional<int> video_renderer_min_input_buffers,
-    std::optional<int> video_renderer_min_decoded_frames,
-    std::optional<int> media_codec_reset_delay_ms,
+    const ExperimentalFeatures& experimental_features,
     void* surface_view,
     SbDecodeTargetGraphicsContextProvider*
         decode_target_graphics_context_provider,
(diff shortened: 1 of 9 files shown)


starboard/shared/starboard/player/filter/testing/video_decoder_test_fixture.cc
```diff
@@ -81,22 +81,18 @@ void VideoDecoderTestFixture::Initialize() {
   ASSERT_TRUE(PlayerComponents::Factory::OutputModeSupported(
       output_mode, dmp_reader_.video_codec(), kSbDrmSystemInvalid));
   int max_video_input_size = 0;
-  bool flush_decoder_during_reset = false;
-  bool reset_audio_decoder = false;
+  PlayerComponents::ExperimentalFeatures experimental_features;
+  experimental_features.flush_decoder_during_reset = true;
```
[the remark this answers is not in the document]
kjyoun: This is intended change. 

since default value is false, setting false does not change anything.

Hence, it changed the test logic to set true, to see the value is actually changed.

starboard/shared/starboard/player/filter/player_components.cc
```diff
@@ -83,7 +83,9 @@ int AlignUp(int value, int alignment) {
 PlayerComponents::Factory::CreationParameters::CreationParameters(
     const media::AudioStreamInfo& audio_stream_info,
     SbDrmSystem drm_system)
-    : audio_stream_info_(audio_stream_info), drm_system_(drm_system) {
+    : audio_stream_info_(audio_stream_info),
+      experimental_features_(ExperimentalFeatures{}),
```
borongc, line 87: nit: we should put this after `drm_system` to make constructor initialization members in-order.

starboard/shared/starboard/player/filter/player_components.cc
```diff
@@ -160,28 +138,6 @@ PlayerComponents::Factory::CreationParameters::CreationParameters(
             video_stream_info_.codec != kSbMediaVideoCodecNone);
 }
 
-PlayerComponents::Factory::CreationParameters::CreationParameters(
```
borongc, line 171: Is this intended?

jasonzhangxx, line 171: What was the copy constructor used for?

starboard/shared/starboard/player/filter/player_components.h
```diff
@@ -54,6 +54,16 @@ class PlayerComponents {
   typedef ::starboard::shared::starboard::player::filter::VideoRenderer
       VideoRenderer;
 
+  struct ExperimentalFeatures {
```
jasonzhangxx, line 57: By adding this, we'll have `VideoDecoder::ExperimentalFeatures` and `PlayerComponents::ExperimentalFeatures`. Can we reuse one of them to simplify the codebase?

borongc:
nit: consider to put either "starboard:" or "android:" to the title so we can understand what this PR changes.

starboard/shared/starboard/player/filter/player_components.cc
kjyoun: Yes, this is intended.

I replace this one with the default copy constructor.
Then we don't need to update copy ctor, when we adds a new member.

kjyoun: We are still using copy ctor, but I replaced manual copy ctor with the default copy ctor.

The previous manual version was just performing member-wise assignment, which is what default copy ctor does. With default copy ctor, we don't need to update it, whenever a new member is added

starboard/shared/starboard/player/filter/player_components.h
kjyoun: Ack.

I will handle it in the separate PR.
Previously,  handing `VideoDecoder::ExperimentalFeatures` needed a big design change during the code, which eventually led to makiing regresssion.

Hence I want to make this PR just focusing on consolidating experiment features that PlayerComponent is using.

starboard/shared/starboard/player/filter/player_components.cc
kjyoun: I checked the header, and experimental_features_ is actually declared before drm_system_ in the class definition. 

https://screenshot.googleplex.com/BhgtyE69kGdSS46

Let me know if I missed something

kjyoun:
> nit: consider to put either "starboard:" or "android:" to the title so we can understand what this PR changes.

Done

jasonzhangxx (approved):

johnxwork (approved):

PR #5092, opened 2025-03-17
Load EGL and GLES through Angle

sherryzy:
In GLOzoneEGLStarboard::LoadGLES2Bindings(), we bypass Angle and load egl and gles library direclty. This should be loaded through Angle.

b/371272304

PR #9332, opened 2026-03-03
starboard: Integrate ANGLE into Ozone EGL backend

denisyuji:
Switch the Ozone Starboard implementation from direct EGL/GLES calls
to use ANGLE for function resolution.

Introduce SbGetNativeEGLDisplayType() to expose the platform's native
EGL display type. This allows ANGLE's native backend to correctly share
the platform display connection, resolving a GPU thread segfault
experienced on RDK/Wayland. This also removes custom manual lookups for
EGL/GLES function pointers, simplifying the graphics initialization.

Bug: 486053854

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -955,11 +955,13 @@ jobs:
       - name: Failed Tests
         if: needs.test-results.result == 'failure' || needs.browser-test-results.result == 'failure'
         run: |
-          # Print test result summary.
+          # Print test result summary when XML files are available.
           shopt -s globstar nullglob
-          python3 cobalt/tools/junit_mini_parser.py results/**/*.xml
-
-          if [[ "${{ needs.test-results.result }}" == "failure" || "${{ needs.browser-test-results.result }}" == "failure" ]]; then
+          xml_files=(results/**/*.xml)
+          if [ ${#xml_files[@]} -gt 0 ]; then
+            python3 cobalt/tools/junit_mini_parser.py results/**/*.xml
             exit 1
+          else
+            echo "No JUnit XML files to parse (artifacts may not have been uploaded)."
           fi
         shell: bash
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_egl.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_egl.cc
@@ -231,3 +231,13 @@ const SbEglInterface g_sb_egl_interface = {
 const SbEglInterface* SbGetEglInterface() {
   return &g_sb_egl_interface;
 }
+
+void* SbGetNativeEGLDisplayType() {
+  NativeDisplayType display_type = 0;
+  EssCtx* ctx =
+      third_party::starboard::rdk::shared::Application::Get()->GetEssCtx();
+  if (ctx && EssContextGetEGLDisplayType(ctx, &display_type)) {
+    return reinterpret_cast<void*>(display_type);
+  }
+  return nullptr;
+}
--- a/ui/ozone/platform/starboard/gl_ozone_egl_starboard.cc
+++ b/ui/ozone/platform/starboard/gl_ozone_egl_starboard.cc
@@ -14,221 +14,10 @@
 
 #include "ui/ozone/platform/starboard/gl_ozone_egl_starboard.h"
 
-#include <memory>
-
 #include "starboard/egl.h"
-#include "starboard/gles.h"
+#include "ui/ozone/common/egl_util.h"
 
 namespace ui {
-
-// Macro to simplify the lookup of function pointers from a given interface.
-// It checks if 'name' matches '#func_name' and returns the corresponding
-// function pointer from 'interface_ptr'.
-#define GET_INTERFACE_PROC_ADDRESS_ENTRY(func_name, interface_ptr) \
-  if (strcmp(name, #func_name) == 0) {                             \
-    return reinterpret_cast<gl::GLFunctionPointerType>(            \
-        interface_ptr->func_name);                                 \
-  }
-
-// Helper function to look up EGL function pointers from the SbEglInterface.
-gl::GLFunctionPointerType GetEglInterfaceProcAddress(
-    const char* name,
-    const SbEglInterface* egl) {
-  GET_INTERFACE_PROC_ADDRESS_ENTRY(eglChooseConfig, egl);
-  GET_INTERFACE_PROC_ADDRESS_ENTRY(eglCopyBuffers, egl);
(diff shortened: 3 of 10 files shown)


ui/ozone/platform/starboard/gl_ozone_egl_starboard.cc
```diff
@@ -252,58 +41,17 @@ scoped_refptr<gl::GLSurface> GLOzoneEGLStarboard::CreateOffscreenGLSurface(
...
-  gl::SetGLGetProcAddressProc(gl_proc);
-  return true;
-}
-
-void GLOzoneEGLStarboard::CreateDisplayTypeIfNeeded() {
-  if (!have_display_type_) {
-    display_type_ = reinterpret_cast<void*>(SB_EGL_DEFAULT_DISPLAY);
-    have_display_type_ = true;
-  }
+  return LoadDefaultEGLGLES2Bindings(implementation);
```
sherryzy: I don't think we can directly call this LoadDefaultEGLGLES2Bindings(). I had a PR previous and tried to do the same thing which got reverted, see: https://github.com/youtube/cobalt/pull/5092#discussion_r2001910158

@y4vor should be the best one to comment on this. But I think in general the idea is we can't load libraries such as libEGL.so, libGLESv2.so directly.

denisyuji: Thank you for your review, @sherryzy . This is a good point. I am following the same logic as the other Ozone implementations:

- X11 ([ui/ozone/platform/x11/x11_surface_factory.cc](https://github.com/youtube/cobalt/blob/24a9747eb6c2b368fb6aefd5b90f8199a6b70a94/ui/ozone/platform/x11/x11_surface_factory.cc#L158))
- DRM ([ui/ozone/platform/drm/gpu/gbm_surface_factory.cc](https://github.com/youtube/cobalt/blob/24a9747eb6c2b368fb6aefd5b90f8199a6b70a94/ui/ozone/platform/drm/gpu/gbm_surface_factory.cc#L198))
- Wayland ([ui/ozone/platform/wayland/gpu/wayland_surface_factory.cc](https://github.com/youtube/cobalt/blob/main/ui/ozone/platform/wayland/gpu/wayland_surface_factory.cc#L186))

They all delegate to `LoadDefaultEGLGLES2Bindings`, which will use ANGLE's resolver `EGL_GetProcAddress` ( if [BUILDFLAG(USE_STATIC_ANGLE) == true](https://github.com/youtube/cobalt/blob/24a9747eb6c2b368fb6aefd5b90f8199a6b70a94/ui/ozone/common/egl_util.cc#L146)). Then, the ANGLE's `EGL_GetProcAddress` resolver will use the starboard's `eglCreateContext` native driver.

If we use starboard's `eglGetProcAddress` resolver directly here, then it would bypass ANGLE, and in this PR the objective is to integrate ANGLE into the Ozone EGL backend.

@y4vor , is my assumption correct?

sherryzy: Yeah, In M114, the BUILDFLAG(USE_STATIC_ANGLE) path didn't exist; the system simply read the graphics library directly. I’ll dig into the M138 source later today to verify this new logic. In the meantime, @y4vor, would you mind taking a look as well?

sherryzy: > Then, the ANGLE's EGL_GetProcAddress resolver will use the starboard's eglCreateContext native driver.

I’m not sure that’s entirely accurate. This new approach bypasses the Starboard driver in favor of a direct ANGLE implementation, which is a cleaner and more robust architecture. We’ve seen issues previously where older devices, like the Raspberry Pi, struggled with outdated EGL/GLES versions. By using use_static_angle, ANGLE is compiled into libGLESv2_static and libEGL_static and linked directly into Cobalt. While this might lead to a slight increase in binary size, the improved stability and correctness make it a worthwhile trade-off.

.github/workflows/main.yaml
```diff
@@ -955,11 +955,13 @@ jobs:
       - name: Failed Tests
         if: needs.test-results.result == 'failure' || needs.browser-test-results.result == 'failure'
         run: |
-          # Print test result summary.
+          # Print test result summary when XML files are available.
```
sherryzy, line 958: I don't fully understand this part. @andrewsavage1 Can you take a look at the change of CI pipeline?

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_egl.cc
```diff
@@ -231,3 +231,13 @@ const SbEglInterface g_sb_egl_interface = {
 const SbEglInterface* SbGetEglInterface() {
   return &g_sb_egl_interface;
 }
+
+void* SbGetNativeEGLDisplayType() {
```
sherryzy, line 235: Shall we add a nplb test for this in starboard/nplb/egl_test.cc?

sherryzy:
@y4vor Would be great if can review the starboard related change.

ui/ozone/platform/starboard/gl_ozone_egl_starboard.cc
y4vor: You need to go through the SbGetEglInterface. Also is the extension that is missing critical? Please consider alternatives as disabling the lookup or providing the functionality in different way.

denisyuji: > You need to go through the SbGetEglInterface. Also is the extension that is missing critical? Please consider alternatives as disabling the lookup or providing the functionality in different way.

@y4vor , the extension (and ANGLE) are not critical at all for Starboard / RDK at this moment. We are implementing it just to pass the DCHECKs on `ui/gl/gl_context_egl.cc` (in the `main` branch, `devel` config, they currently crashes Cobalt on initialization if ANGLE extensions are not present)
Do you think [#9266](https://github.com/youtube/cobalt/pull/9266/changes) could be a better solution? It skips these DCHECKs for Cobalt. This workaround is proved to work too.

sherryzy: Discussed offline, inside ANGLE code, it still calls dlsym to load graphic libraries. And it should be go through starboard interface instead. We probably need to go through the original approach for relaxing the DCHECK() to avoid crashing.

niranjanyardi:
@denisyuji  I'm closing this PR as we have decided to go with the other approach. Feel free to reopen this if you have more questions

https://github.com/youtube/cobalt/pull/9332#discussion_r2892929507

Some more poitns from discussion:

 
The dependencies are:
Cobalt -> Angle -> Starboard -> system GL
Starboard -> system GL

If an extension is not supported on a platform - there's no way around it atm

PR #9349, opened 2026-03-03
starboard: Implement network API

andrewsavage1:
Changes out stub network connectivity implementation for real one on Starboard platforms.

Bug: 448158426

### DIFF
--- a/content/browser/browser_main_loop.h
+++ b/content/browser/browser_main_loop.h
@@ -184,6 +184,9 @@ class CONTENT_EXPORT BrowserMainLoop {
   }
 
 #if BUILDFLAG(IS_CHROMEOS) || BUILDFLAG(IS_STARBOARD)
+  // Cobalt: Expose on Starboard for passive updates. Note this will not be
+  // null for the preload/background state.
+
   // Only expose this on ChromeOS since it's only needed there. On Android this
   // be null if this process started in reduced mode.
   net::NetworkChangeNotifier* network_change_notifier() const {
--- a/content/public/browser/network_service_instance.h
+++ b/content/public/browser/network_service_instance.h
@@ -41,6 +41,7 @@ namespace content {
 CONTENT_EXPORT network::mojom::NetworkService* GetNetworkService();
 
 // Only on ChromeOS since it's only used there.
+// Cobalt: Used on Starboard as well.
 #if BUILDFLAG(IS_CHROMEOS) || BUILDFLAG(IS_STARBOARD)
 // Returns the global NetworkChangeNotifier instance.
 CONTENT_EXPORT net::NetworkChangeNotifier* GetNetworkChangeNotifier();


net/BUILD.gn
```diff
@@ -1657,16 +1657,16 @@ component("net") {
...
-  }
-  if (is_cobalt && is_linux && current_toolchain == host_toolchain) {
-    sources -= [
-      #  # TODO: b/413491685 - Cobalt: Fix errors building host tools.
-      "base/network_change_notifier_passive.cc",
+      # "base/network_change_notifier_starboard.h",
+      # "base/network_change_notifier_passive_starboard_stub.cc",
     ]
   }
+  # if (is_cobalt && is_linux && current_toolchain == host_toolchain) {
```
jellefoks: Commented out?

net/BUILD.gn
```diff
@@ -1657,16 +1657,16 @@ component("net") {
...
-      "base/network_change_notifier_starboard.h",
-      "base/network_change_notifier_passive_starboard_stub.cc",
-    ]
-  }
-  if (is_cobalt && is_linux && current_toolchain == host_toolchain) {
-    sources -= [
-      #  # TODO: b/413491685 - Cobalt: Fix errors building host tools.
-      "base/network_change_notifier_passive.cc",
+      # "base/network_change_notifier_starboard.h",
+      # "base/network_change_notifier_passive_starboard_stub.cc",
```
jellefoks: Remove commented out lines?

net/base/network_change_notifier_passive.cc
```diff
@@ -116,6 +116,12 @@ NetworkChangeNotifierPassive::NetworkChangeCalculatorParamsPassive() {
   params.connection_type_online_delay_ = base::Milliseconds(500);
 #elif BUILDFLAG(IS_ANDROID)
   params = NetworkChangeNotifierAndroid::NetworkChangeCalculatorParamsAndroid();
+#elif BUILDFLAG(IS_STARBOARD)
+  // TODO: b/123, Cobalt - Adjust these values with experimentation. Right now they just follow linux.
+  params.ip_address_offline_delay_ = base::Milliseconds(2000);
```
jellefoks, line 122: Do we have an h5vcc.settings equivalent that we could hookup to these? 

Optional followup: Maybe we can add histograms for the connection disruption durations, so we can learn the distribution to know how much we need to debounce? 

(edit for `s/denounce/debounce/`)

andrewsavage1: > Do we have an h5vcc.settings equivalent that we could hookup to these?

We don't

> Optional followup: Maybe we can add histograms for the connection disruption durations, so we can learn the distribution to know how much we need to denounce?

I can file a bug to follow up on this

net/BUILD.gn
andrewsavage1: Sorry, meant to open as a draft

net/base/network_change_notifier.h
```diff
@@ -331,7 +331,7 @@ class NET_EXPORT NetworkChangeNotifier {
         observer_list_;
   };
 
-#if BUILDFLAG(IS_CHROMEOS_LACROS)
+#if BUILDFLAG(IS_CHROMEOS_LACROS) || BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 334: Test if we want to change the initial connection type once I'm able to test this out

andrewsavage1: Confirmed that if we don't have this we start by being offline. In the future it may be better to manually set the connection type from the lifecycle manager instead.

andrewsavage1:
Was able to test on linux using 

```
# Setup
sudo modprobe dummy
sudo ip link add dummy0 type dummy
sudo ip link set dummy0 up

# Action
sudo ip addr add 192.168.255.255/24 dev dummy0
sudo ip addr del 192.168.255.255/24 dev dummy0
```

net/BUILD.gn
```diff
@@ -1657,14 +1657,6 @@ component("net") {
     sources += [
       "base/network_interfaces_starboard.cc",
       "base/platform_mime_util_starboard.cc",
-      "base/network_change_notifier_starboard.h",
-      "base/network_change_notifier_passive_starboard_stub.cc",
-    ]
-  }
-  if (is_cobalt && is_linux && current_toolchain == host_toolchain) {
-    sources -= [
-      #  # TODO: b/413491685 - Cobalt: Fix errors building host tools.
```
jellefoks, line 1666: Nit: double hash: "#  # "

y4vor (approved):

net/BUILD.gn
andrewsavage1: I'm not sure what you're suggesting, this is removed code

PR #9350, opened 2026-03-03
Add custom mapping for kSbKeyLaunchThisApplication

y4vor:
- Adding custom Web KeyboardEvent.key mapping for the kSbKeyLaunchThisApplication which is used for the
YouTube key on the remote.

Issue: 489424097

### DIFF
--- a/ui/events/keycodes/dom/dom_key_data.inc
+++ b/ui/events/keycodes/dom/dom_key_data.inc
@@ -536,4 +536,8 @@ DOM_KEY_MAP_DECLARATION {
   DOM_KEY_MAP("VideoModeNext",        VIDEO_MODE_NEXT,          0x0D4C),
   DOM_KEY_MAP("Wink",                 WINK,                     0x0D4D),
   DOM_KEY_MAP("ZoomToggle",           ZOOM_TOGGLE,              0x0D4E),
+#if BUILDFLAG(IS_COBALT)
+  // YouTube button delivered through kSbKeyLaunchThisApplication on Starboard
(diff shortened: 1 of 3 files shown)


ui/events/keycodes/dom/dom_key_data.inc
```diff
@@ -536,4 +536,8 @@ DOM_KEY_MAP_DECLARATION {
   DOM_KEY_MAP("VideoModeNext",        VIDEO_MODE_NEXT,          0x0D4C),
   DOM_KEY_MAP("Wink",                 WINK,                     0x0D4D),
   DOM_KEY_MAP("ZoomToggle",           ZOOM_TOGGLE,              0x0D4E),
+#if BUILDFLAG(IS_COBALT)
```
andrewsavage1, line 539: Using a BUILDFLAG here is a bit strange, as it's not totally clear what the mechanism to include #defines into this file would be and seems like it could break easily. I don't feel that strongly about it, but maybe if would be better to check `#if defined(ENABLE_BUILDFLAG_IS_COBALT)` directly here instead?

y4vor: I think it works fine. The file is a bit strange but gets included in the middle of regular Chromium code.

andrewsavage1: My point is that there's no guarantee about where the file gets included, and this assumes the BUILDFLAG macro is defined (i.e. `build/build_config.h` is `#include`d before this is), which isn't a very safe assumption

PR #9358, opened 2026-03-04
cobalt/metrics: Add UMA recording CPU usage

tasunnn:
cobalt: Add CPU.Total.Usage UMA

Introduce a CobaltCpuMetricsEmitter class responsible for
periodically fetching and emitting cobalt's total CPU usage
as a UMA histogram. This emitter is integrated into the
CobaltMetricsServiceClient, ensuring CPU usage metrics are
regularly collected alongside existing memory metrics.

This change provides valuable telemetry on system resource
utilization, aiding in performance monitoring and analysis.

Bug: 487654198

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -354,6 +354,10 @@ test("cobalt_unittests") {
     "//ui/events/devices",
   ]
 
+  if (is_starboard) {
+    deps += [ "//ui/ozone/platform/starboard" ]
+  }
+
   if (is_android) {
     deps += [
       "//base:base_java_test_support",
--- a/cobalt/app/app_lifecycle_delegate.cc
+++ b/cobalt/app/app_lifecycle_delegate.cc
@@ -33,11 +33,11 @@
 #include "cobalt/shell/browser/shell.h"
 #include "content/public/app/content_main.h"
 #include "content/public/app/content_main_runner.h"
+#include "content/public/browser/network_service_instance.h"
+#include "net/base/network_change_notifier_passive.h"
 
 #if BUILDFLAG(IS_STARBOARD)
 #include "cobalt/app/cobalt_switch_defaults.h"
-#include "content/public/browser/network_service_instance.h"
-#include "net/base/network_change_notifier_passive.h"
 #include "services/device/time_zone_monitor/time_zone_monitor_starboard.h"
 #endif
 
@@ -85,19 +85,31 @@ class DefaultAppLifecycleRunner : public cobalt::AppLifecycleRunner {
 
   void ShutDown() override {
     content::Shell::Shutdown();
+
+    if (content_main_delegate_) {
+      content_main_delegate_->Shutdown();
+    }
+
+    // Must be called after content_main_delegate_->Shutdown() to prevent
+    // unregistering the main thread from the SequenceManager which
+    // happens with the destruction of the BrowserTaskExecutor. If the order
+    // is reversed the SequenceManager complains (fails DCHECK) that the
+    // ContentMainRunnerImpl::Shutdown() is happening on the wrong thread as
+    // it no longer recognizes the main thread as a valid TaskEnvironment.
     if (main_runner_) {
       main_runner_->Shutdown();
     }
-    if (content_main_delegate_) {
-      content_main_delegate_->Shutdown();
-      content_main_delegate_.reset();
-    }
+
+    // Destroy only after main_runner_/ContentMainRunnerImpl is shutdown
+    // as the delegate is used internally.
+    content_main_delegate_.reset();
+
     // We intentionally do not null main_runner_ here to enforce the one-shot
     // rule: once stopped, the process cannot be re-started.
     exit_manager_.reset();
   }
 
-  bool IsRunning() const override { return content_main_delegate_ != nullptr; }
+  bool IsRunning() const override { return main_runner_ != nullptr; }
 
  private:
   std::unique_ptr<base::AtExitManager> exit_manager_;
@@ -133,35 +145,51 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
       OnStop(event);
       break;
     case kSbEventTypeBlur:
-    case kSbEventTypeFocus:
-#if BUILDFLAG(IS_STARBOARD)
-    {
-      auto* client = cobalt::CobaltContentBrowserClient::Get();
-      if (client) {
-        if (event->type == kSbEventTypeBlur) {
+      content::Shell::OnBlur();
+      {
+        auto* client = cobalt::CobaltContentBrowserClient::Get();
+        if (client) {
           client->DispatchBlur();
-        } else {
+        }
+      }
+#if BUILDFLAG(IS_STARBOARD)
+      if (platform_event_source_) {
+        platform_event_source_->HandleFocusEvent(event);
+      }
+#endif
+      break;
+    case kSbEventTypeFocus:
+      content::Shell::OnFocus();
+      {
+        auto* client = cobalt::CobaltContentBrowserClient::Get();
+        if (client) {
           client->DispatchFocus();
         }
       }
-    }
+#if BUILDFLAG(IS_STARBOARD)
       if (platform_event_source_) {
         platform_event_source_->HandleFocusEvent(event);
       }
 #endif
       break;
     case kSbEventTypeConceal:
+      content::Shell::OnConceal();
       break;
     case kSbEventTypeReveal:
       content::Shell::OnReveal();
       break;
-    case kSbEventTypeFreeze: {
-      auto* client = cobalt::CobaltContentBrowserClient::Get();
-      if (client) {
-        client->FlushCookiesAndLocalStorage();
+    case kSbEventTypeFreeze:
+      content::Shell::OnFreeze();
+      {
+        auto* client = cobalt::CobaltContentBrowserClient::Get();
+        if (client) {
+          client->FlushCookiesAndLocalStorage();
+        }
       }
-    } break;
+      break;
     case kSbEventTypeUnfreeze:
+      content::Shell::OnUnfreeze();
+      break;
     case kSbEventTypeInput:
 #if BUILDFLAG(IS_STARBOARD)
       if (platform_event_source_) {
@@ -187,7 +215,7 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
       // Chromium internally calls Reclaim/ReclaimNormal at regular interval
       // to claim free memory. Using ReclaimAll is more aggressive.
       // TODO: b/454095852 - Remove this when
-      // https://chromium-review.googlesource .com/c/chromium/src/+/7127962
+      // https://chromium-review.googlesource.com/c/chromium/src/+/7127962
       // lands on main
       ::partition_alloc::MemoryReclaimer::Instance()->ReclaimAll();
 
@@ -214,9 +242,8 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
 #endif
       break;
     case kSbEventTypeOsNetworkDisconnected:
-    case kSbEventTypeOsNetworkConnected:
+    case kSbEventTypeOsNetworkConnected: {
 #if BUILDFLAG(IS_STARBOARD)
-    {
       auto* notifier = content::GetNetworkChangeNotifier();
       if (notifier) {
         auto* passive_notifier =
@@ -232,9 +259,9 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
         passive_notifier->OnConnectionChanged(type);
         passive_notifier->OnConnectionSubtypeChanged(type, subtype);
       }
-    }
 #endif
-    break;
+      break;
+    }
     case kSbEventDateTimeConfigurationChanged:
 #if BUILDFLAG(IS_STARBOARD)
       device::NotifyTimeZoneChangeStarboard();
@@ -271,6 +298,7 @@ void AppLifecycleDelegate::OnStop(const SbEvent* event) {
   if (!runner_->IsRunning()) {
     return;
   }
+  content::Shell::OnStop();
   runner_->ShutDown();
 
 #if BUILDFLAG(IS_STARBOARD)
--- a/cobalt/app/app_lifecycle_delegate_unittest.cc
+++ b/cobalt/app/app_lifecycle_delegate_unittest.cc
@@ -17,6 +17,7 @@
 #include "base/functional/bind.h"
 #include "cobalt/shell/browser/shell.h"
 #include "cobalt/shell/browser/shell_test_support.h"
+#include "content/public/browser/render_widget_host_view.h"
 #include "content/test/test_web_contents.h"
 #include "testing/gmock/include/gmock/gmock.h"
 #include "testing/gtest/include/gtest/gtest.h"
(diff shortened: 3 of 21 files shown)


cobalt/browser/metrics/cobalt_metrics_service_client.h
```diff
@@ -140,6 +141,8 @@ class CobaltMetricsServiceClient : public metrics::MetricsServiceClient {
   virtual scoped_refptr<CobaltMemoryMetricsEmitter>
   CreateMemoryMetricsEmitter();
 
+  virtual scoped_refptr<CobaltCpuMetricsEmitter> CreateCpuMetricsEmitter();
```
yell0wd0g, line 149: Why is this virtual? I only see it implemented once, in 
the .cc file (maybe it's intended to be used for tests
later on)?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,39 @@
...
+
+void CobaltCpuMetricsEmitter::FetchAndEmitCpuMetrics() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  // Total CPU utilization in percentage of all cores over an interval
+  auto cpu_usage = process_metrics_->GetPlatformIndependentCPUUsage();
+  if (cpu_usage.has_value()) {
+    base::UmaHistogramCounts1000("CPU.Total.Usage", 
+                                static_cast<int>(cpu_usage.value() + 0.5));
+  }
```
yell0wd0g, line 35: What if GetPlatformIndependentCPUUsage() returns an error?

Maybe in that case log it as 0? Basically reserving one of the
histogram values to signify "invalid". I think for CPU, a value
of 0 would be intuitive enough, WDYT?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,39 @@
...
+          base::GetCurrentProcessHandle())) {
+  DETACH_FROM_SEQUENCE(sequence_checker_);
+}
+
+CobaltCpuMetricsEmitter::~CobaltCpuMetricsEmitter() = default;
+
+void CobaltCpuMetricsEmitter::FetchAndEmitCpuMetrics() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  // Total CPU utilization in percentage of all cores over an interval
```
yell0wd0g: nit: Period at the end of comment.

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,39 @@
...
+  auto cpu_usage = process_metrics_->GetPlatformIndependentCPUUsage();
```
yell0wd0g: nit: const?

cobalt/browser/metrics/cobalt_metrics_service_client.h
tasunnn: yes I haven't worked on tests for this PR since there's only one metric now but plan to include tests similar to the browsertests for memory metrics

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,39 @@
...
+  if (cpu_usage.has_value()) {
+    base::UmaHistogramCounts1000("CPU.Total.Usage", 
+                                static_cast<int>(cpu_usage.value() + 0.5));
```
yell0wd0g: I think the (value + 0.5) is to round up to the nearest integer?
If so, prefer `std::ceil()` so that intent is clear. 

What's the expected maximum value of `cpu_usage`? Is it 100 (like
a percent) or could it be e.g. 400 if there're four cores? (Seems
to be the latter as per [code comment](https://source.chromium.org/chromium/chromium/src/+/main:base/process/process_metrics.h;l=155-164;drc=d310efd6b1671b40cfab6ccd995a7cdc5c4009fc)). (`UmaHistogramCounts1000` 
seems to saturate at 1000, so inherently you're assuming a 
maximum of 10 cores). 

Logging directly this number has the problem of confusing 
comparisons among SoCs with different core counts, i.e. one
SoC with 2 cores could be fully utilized, hence logging 200,
while an SoC with four cores and the same load would also
log 200, while the CPU "load" would be more 50% to me. The
documentation recommends dividing `cpu_usage` by 
`SysInfo::NumberOfProcessors()` to have a single percentage,
which can then be logged via `UmaHistogramPercentage()`.
WDYT?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
tasunnn: Yes that makes sense, right now it just directly skip logging if returns on error, having a 0 can give clearer info about the state

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
tasunnn: yeah right now the max value would be (number of cores * 100%) so I used UmaHistogramCounts1000 for a looser range. 

I think your point makes sense, I was thinking to get the total usage for all cores but it does seem less intuitive when comparing across devices with different cores, I will divide it by number of processors to make it more clear.

tasunnn: Not sure if we want to use `std::ceil()` since it will always round up, which might be less accurate. I just found this [ClampRound ](https://source.chromium.org/chromium/chromium/src/+/main:base/numerics/safe_conversions.h;l=381;bpv=1;bpt=1?q=base::ClampRound&ss=chromium%2Fchromium%2Fsrc)in chromium that does rounding to nearest integer, maybe we can use this?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,41 @@
...
+  DETACH_FROM_SEQUENCE(sequence_checker_);
+}
+
+CobaltCpuMetricsEmitter::~CobaltCpuMetricsEmitter() = default;
+
+void CobaltCpuMetricsEmitter::FetchAndEmitCpuMetrics() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+
+  // Total CPU utilization in percentage of all cores in between every call.
+  // This will be zero on first call or when GetPlatformIndependentCPUUsage returns on error.
```
yell0wd0g: nit: This line and some of the ones below seem longer than
80 characters? Pro-tip: Running `git cl format` would fix
those style deviations.

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,41 @@
...
+  const double cpu_usage = process_metrics_->GetPlatformIndependentCPUUsage().value_or(0.0);
```
yell0wd0g: nit: Give constants a name:
```
  const expr double kInvalidCPUUsageValue = 0.0f;
  const double cpu_usage = ... value_or(kInvalidCPUUsageValue);
```

making part of the explanation in comment unnecessary (comments
are often not read, or become outdated).

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,41 @@
...
+  // Divide by the number of processors to log average per core CPU usage.
+  base::UmaHistogramPercentage("CPU.Total.UsageInPercentage",
+                               base::ClampRound<int>(cpu_usage / base::SysInfo::NumberOfProcessors()));
```
yell0wd0g: `base::SysInfo::NumberOfProcessors()` seems to always return non-zero, 
but there's a call to `SbSystemGetNumberOfProcessors()` of which I'm 
not sure. Two options: One is to fetch that value and leave a 
`CHECK()` so it crashes if this value is zero (which I expect will
happen for some Starboard implementor somewhere, the idea would be
to catch it early), or handle the case appropriately (e.g. via
`value_or` etc. IOW don't trust APIs you don't control : )

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "base/memory/ref_counted.h"
+#include "base/process/process_metrics.h"
+#include "base/sequence_checker.h"
+
+namespace cobalt {
+
+// This class fetches CPU metrics for the current process and its threads,
+// and emits UMA metrics. It maintains state to calculate deltas between
+// emissions.
```
yell0wd0g: I like this comment! Can you also add info on threading and
lifetime? E.g. can it be called from one TaskRunner only, or
is it thread safe? (If the former, add a CobaltThreadChecker
plz). Re the latter: this class is ref-counted (and 
ref-counted thread safe, at that), does it need to be? From
a cursory view of the CL, I don't think so, and in general
we should try not to use this ownership idiom if avoidable,
because it complicates code reading and reasoning.

tasunnn: done!

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
tasunnn: I don't think it currently make any calls to `SbSystemGetNumberOfProcessors()`, do you mean we need to check the return value for `base::SysInfo::NumberOfProcessors()` to make sure it's safe?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include <memory>
+
+#include "base/process/process_metrics.h"
+#include "cobalt/common/cobalt_thread_checker.h"
+
+namespace cobalt {
+
+// This class fetches CPU metrics for the current process and its threads,
+// and emits UMA metrics. It maintains state to calculate deltas between
```
yell0wd0g: nit: Re `and emits UMA metrics` -> rather than emitting, what
this code does is to record an entry in a histogram. UMA refers
to the aggregation of histograms (from Chrome) server side.
IOW the correct wording should be sth like "records a 
histogram entry" or some such.

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
```diff
@@ -0,0 +1,47 @@
...
+// emissions. This class is not thread-safe and must be called from the same
```
yell0wd0g: Does this class "maintains state to calculate deltas between emissions" ?
This might be from before, now I see independent record calls.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
yell0wd0g: nit: This should be a class now : )
https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes
> Use a struct only for passive objects that carry data; everything else is a class.

The change would change the privacy of methods/members:
for structs, they are by default public but private for
classes. Feel free to do it in a follow-up CL/PR (Gemini
should do this in less than 5').

yell0wd0g: This method name is no longer current: now it records
the memory metric explicitly in l.91 and then records the CPU
metrics (once) in l.94. Very different stuff. Similarly
`RecordMemoryMetricsAfterDelay()` in l.59 and the variables
such as kMemoryMetricsInterval etc.

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -83,9 +86,16 @@ struct CobaltMetricsServiceClient::State
...
 
-    scoped_refptr<CobaltMemoryMetricsEmitter> emitter =
+    scoped_refptr<CobaltMemoryMetricsEmitter> memory_emitter =
         parent_->CreateMemoryMetricsEmitter();
-    emitter->FetchAndEmitProcessMemoryMetrics();
+    memory_emitter->FetchAndEmitProcessMemoryMetrics();
+
+    if (!cpu_emitter_) {
+      cpu_emitter_ = parent_->CreateCpuMetricsEmitter();
+    }
```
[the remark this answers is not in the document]
yell0wd0g: Except `cpu_emitter_` is a std::unique_ptr<>, not a scoped_refptr...

I have a different perspective which is that `memory_emitter`
is a local variable (maybe `parent_` caches it?) but 
`cpu_emitter_` is a member variable here, and there's no
explanation of the difference.

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
```diff
@@ -0,0 +1,47 @@
...
+// TaskRunner it was created on.
+class CobaltCpuMetricsEmitter {
+ public:
+  CobaltCpuMetricsEmitter();
+
+  CobaltCpuMetricsEmitter(const CobaltCpuMetricsEmitter&) = delete;
+  CobaltCpuMetricsEmitter& operator=(const CobaltCpuMetricsEmitter&) = delete;
+
+  void FetchAndEmitCpuMetrics();
+  virtual ~CobaltCpuMetricsEmitter();
```
yell0wd0g, line 44: Add here `//TODO(b/...): Add unit tests.` to justify this class
not being final and having a virtual dtor.

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
tasunnn: I probably need to correct the wording here, the class itself is not maintaining any state per say. But the way how the class is initiated will ensure that the cumulative cpu usage tracked by `GetPlatformIndependentCPUUsage` does not get reset for every record and will be updated to ensure it's consistent with current state.

yell0wd0g: Oh so `GetPlatformIndependentCPUUsage()` has some state internally and
we must hold on to it here as a member variable for it to work 
properly?

If yes I'm thinking this `State` class in incorrectly named and it should
be called `MemoryPoller` or some such, and should not be reused for CPU 
tracking as is (because we'd be bundling two separate things, increasing
cohesion and reducing coherency). How about adding a new one?

tasunnn: yeah the current implementation lets `GetPlatformIndependentCPUUsage()` to handle the state tracking but if we don't want that, we could also maintain the state directly inside this class and pass that in as a parameter when calculating cpu usage https://source.chromium.org/chromium/chromium/src/+/main:base/process/process_metrics.h;l=168?q=GetPlatformIndependentCPUUsage&ss=chromium%2Fchromium%2Fsrc

tasunnn: for your second point, do you mean that we should separate CPU stuff from this state class `CobaltMetricsServiceClient::State` since that seems to be handling only memory-related states right now

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
```diff
@@ -0,0 +1,45 @@
...
+
+  // Total CPU utilization in percentage of all cores in between every call.
+  constexpr double kInvalidCPUUsageValue = 0.0;
+  const double cpu_usage =
+      process_metrics->GetPlatformIndependentCPUUsage().value_or(
+          kInvalidCPUUsageValue);
+  // Divide by the number of processors to log average per core CPU usage.
+  base::UmaHistogramPercentage(
+      "CPU.Total.UsageInPercentage",
+      base::ClampRound<int>(cpu_usage / base::SysInfo::NumberOfProcessors()));
```
Awallky: question: Should we protect against divide by zero errors here?

Awallky (approved):
LGTM

Can we follow-up with a browser test for this new metric?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.cc
tasunnn: yeah this is a good point, I will add checks

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -40,59 +41,88 @@
...
+};
 
- private:
-  friend class base::RefCountedThreadSafe<State>;
-  ~State() = default;
+class CobaltMetricsServiceClient::CpuPollingState : public MetricsPollingState {
+ public:
+  using MetricsPollingState::MetricsPollingState;
+
+  std::unique_ptr<base::ProcessMetrics> process_metrics_;
```
yell0wd0g, line 113: Can we move this guy to CobaltCpuMetricsEmitter, and 
initialize it in its ctor? That would make it more
obvious that it's needed to keep a state (currently
the class comment say so but there's no member vars
which I found a bit surprising).

That way, this CpuPollingState class would be in 
charge of repeatedly polling, and CobaltCpuMetricsEmitter
in charge of accessing and retrieving those metrics,
making the division of labour more clear IMHO.

tasunnn: This is a good point, I originally had it inside CobaltCpuMetricsEmitter but found issues synchronizing with CobaltMemoryMetricsEmitter when polling data since the cpu emitter would be persisted to manage its own state, which handles things/timing differently than memory emitter. So I moved process_metrics to CpuPollingState and I think it would make more sense here for the state object to manage the state while emitter to only provide metrics, similar to what's done in CobaltMemoryMetricsEmitter. I could modify the comments to reflect this change, WDYT?

yell0wd0g (approved):
Discussed offline that this is good as-is and
there could be follow ups to clarify ownership
and class diagram tracked in the bug+TODO. LGTM!

PR #9361, opened 2026-03-04
Reland "cobalt/test: Enable dedicated browser tests action for CI - Linux"

Awallky:
Reverts #9357

Fix validate-test-result CI failure on irrelevant platforms
    
    Modified junit_mini_parser.py to return 1 on test failures and exit silently with 0 on empty arguments.
    Updated main.yaml to:
    - Make validate-test-result and its artifact downloads conditional on whether tests were actually run for the platform.
    - Use junit_mini_parser.py exit code for step status.
    - Correctly handle 'skipped' and empty results from preceding test jobs.

Test: `vpython3 cobalt/build/test_archive_test_artifacts.py`
Test: `vpython3 cobalt/testing/browser_tests/tools/test_collect_test_artifacts.py`
Test: `vpython3 cobalt/testing/browser_tests/test_run_browser_tests.py`

Bug: 483488213
Bug: 473909665
Bug: 489793579

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -796,12 +796,14 @@ jobs:
           test_results_key: Blink Web Test Results ${{ github.workflow }}
 
   test-results:
-    needs: [initialize, build, on-device-test, on-host-test]
-    if: always() &&
-        (
-          (needs.on-device-test.result != 'skipped' && needs.on-device-test.result != '') ||
-          (needs.on-host-test.result != 'skipped' && needs.on-host-test.result != '')
-        )
+    needs: [initialize, build, on-device-test, on-host-test, browser-test-on-host]
+    if: |
+      always() &&
+      (
+        needs.on-device-test.result != 'skipped' ||
+        needs.on-host-test.result != 'skipped' ||
+        needs.browser-test-on-host.result != 'skipped'
+      )
     permissions: {}
     runs-on: ubuntu-latest
     name: ${{ matrix.test_target }}
@@ -968,7 +970,7 @@ jobs:
           pattern: ${{ matrix.platform }}_${{ matrix.name }}_browsertest_results
           path: results/
       - name: Failed Tests
-        if: always()
+        if: always() && (needs.test-results.result != 'skipped' || needs.browser-test-results.result != 'skipped')
         run: |
           # Print test result summary.
           shopt -s globstar nullglob


cobalt/testing/browser_tests/tools/run_tests.template.py
```diff
@@ -137,16 +139,24 @@ def main():
...
+      logging.info(
+          "Executing Linux test runner for '%s' using xvfb.py and "
+          "run_browser_tests.py", target_name)
+      xvfb_py = os.path.join(src_dir, "testing/xvfb.py")
+      run_browser_tests_py = os.path.join(
+          src_dir, "cobalt/testing/browser_tests/run_browser_tests.py")
+      cmd = [
+          vpython_path, xvfb_py, sys.executable, run_browser_tests_py,
+          test_runner
+      ] + runner_args
```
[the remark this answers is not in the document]
Awallky: Fixed.

.github/actions/linux_browser_tests/action.yaml
```diff
@@ -0,0 +1,68 @@
...
+        path: artifacts
+
+    - name: Run Browser Tests (Linux)
+      shell: bash
+      env:
+        XVFB_SERVER_ARGS: "-screen 0 1920x1080x24i +render +extension GLX -noreset"
+        PLATFORM: ${{ inputs.platform }}
+        CONFIG: ${{ inputs.config }}
+      run: |
+        set -eux
```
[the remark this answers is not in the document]
Awallky: Fixed.

linxinan-yt:
QQ, do the differences between commit 1 &  2 represent the change between first attempt and reland?

Awallky:
@linxinan-chops 

> QQ, do the differences between commit 1 & 2 represent the change between first attempt and reland?

Yes, exactly.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -39,7 +39,8 @@ runs:
         set -eux
         cd cobalt/src/
         mkdir -p "${GITHUB_WORKSPACE}/artifacts/"
-        for target in $(cat "${TEST_TARGETS_JSON_FILE}" | jq -r '.executables | .[]'); do
+        mapfile -t targets < <(jq -r '.executables | .[]' "${TEST_TARGETS_JSON_FILE}")
```
linxinan-yt, line 42: space?

Awallky: The `<(` serves as a process substitution.
The syntax for process substitution in Bash is strictly defined:

   1. Space between the redirection operator (<) and the process substitution (<(): This space is MANDATORY.
       - ... < <(command) is CORRECT.
       - ... <<(command) is INCORRECT (the shell interprets << as a "here-document" operator, which will break the script).

   2. Space between the second < and the opening parenthesis (: This space is FORBIDDEN.
       - ... < <(command) is CORRECT.
       - ... < < (command) is INCORRECT (the shell will fail to recognize this as process substitution and throw a syntax error).

  Summary:
  In the line `mapfile -t targets < <(jq ...)`, the single space between the first < and the second < is required for the script to function. No other spaces should be added within the `<(` operator itself.

.github/workflows/main.yaml
```diff
@@ -235,6 +235,13 @@ jobs:
              browser_test_targets_json=$(echo "${browser_test_targets}" | jq -c 'map(.target)')
           fi
           echo "browser_test_targets_json=${browser_test_targets_json}" >> $GITHUB_OUTPUT
+
+          # Centralize "should run browser tests" logic.
```
linxinan-yt, line 239: Would you explain in the comment in what cases that test_browser is true but targets json is empty?

cobalt/tools/validate_ci_config.py
```diff
@@ -0,0 +1,110 @@
...
+                       'android-arm must have test_browser: false')
+
+      ci_state = self._simulate_ci_logic(config)
+      self.assertFalse(
+          ci_state['browser-test-results'],
+          'Browser test results job must be skipped for android-arm')
+
+
+if __name__ == '__main__':
+  unittest.main()
```
linxinan-yt, line 110: Is this a unit test file or you want to add it in the pre-commit to check PR before they get uploaded?

.github/actions/upload_test_artifacts/action.yaml
linxinan-yt: Thanks for the explanation.  Good to learn.

.github/workflows/main.yaml
Awallky: Done.

cobalt/tools/validate_ci_config.py
Awallky: We will need to follow-up on this, since we ignore Chromium-like presubmits in cobalt and starboard.

PR #9379, opened 2026-03-06
tvOS: Run gn.py without --no-check in CI

rakuco:
`gn gen --check` is supposed to be working on tvOS builds, so stop passing `--no-check` to `gn.py` to ensure nothing breaks in the future.

Bug: 432823450


rakuco:

PR #9380, opened 2026-03-06
tvos: Use internal libvpx for internal builds

abhijeetk:
When is_internal_build is set, replace the standard libvpx dependency
with the tvOS-specific libvpx_tvos_arm64 internal library.

Without this, internal tvOS builds fail with undeclared identifier
errors for decoder_acquire_hw_image, decoder_release_hw_image,
and decoder_get_number_of_allocated_frame_buffers, all of which
are defined only in the internal libvpx library.

Bug: 487326366

### DIFF
--- a/starboard/tvos/shared/BUILD.gn
+++ b/starboard/tvos/shared/BUILD.gn
@@ -198,6 +198,11 @@ static_library("starboard_platform") {
     "//third_party/opus",
   ]
 
+  if (is_internal_build) {
+    deps -= [ "//third_party/libvpx" ]
+    deps += [ "//third_party/internal/libvpx_tvos:libvpx_tvos_arm64" ]
+  }
+
   frameworks = [
     "AVFoundation.framework",
     "AVKit.framework",


abhijeetk:
cc: @Gyuyoung @jkim-julie

starboard/tvos/shared/BUILD.gn
```diff
@@ -198,6 +198,11 @@ static_library("starboard_platform") {
     "//third_party/opus",
   ]
 
+  if (is_internal_build) {
+    deps -= [ "//third_party/libvpx" ]
```
rakuco, line 202: I'd rather do

```gn
if (is_internal_build) {
  deps += [ ... ]
} else {
  deps += [ "third_party/libvpx" ]
}
```

and remove the unconditional dependency from the `deps` block above.

abhijeetk: @rakuco  Thanks for taking the time to review this! After discussing the details in b/469309439, it looks like PR #9434 is the better path forward for our goals. WDYT ?

rakuco: Works for me too. If we can stick to the open-source libvpx, even better!

PR #9381, opened 2026-03-06
build: Stop disabling use_crashpad on tvOS

rakuco:
This is no longer necessary after #9270 and #9316.

Bug: 477518757


rakuco:

PR #9382, opened 2026-03-06
tvOS: Enable Crashpad integration in Cobalt Shell

rakuco:
This essentially means triggering the processing of any intermediate crash
dumps and later uploading them to the Crashpad server once the app's UI
finishes loading.

The code is based on the `ios/` implementation, especially
`ios/chrome/browser/crash_report/model/crash_helper.mm` (in particular the
way that the steps above are triggered only once and in a separate thread
when a scene finishes activating).

Integration with Cobalt's Crash Log Web API is done in #9408.

Bug: 477518757


cobalt/shell/app/shell_crash_reporter_client.cc
```diff
@@ -30,6 +31,24 @@
...
+  CHECK(base::PathService::Get(base::DIR_CACHE, &default_dir));
+  default_dir = default_dir.Append("Crashpad");
+#endif  // BUILDFLAG(IS_IOS)
+  if (!base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kCrashDumpsDir)) {
+    return default_dir;
+  }
+  return base::CommandLine::ForCurrentProcess()->GetSwitchValuePath(
+      switches::kCrashDumpsDir);
+}
```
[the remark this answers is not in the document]
rakuco: This comes from upstream, I'd rather keep both versions equal.

rakuco:

cobalt/shell/app/shell_crash_reporter_client.cc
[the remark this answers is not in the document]
andrewsavage1: Makes sense to keep it the same, but man that's strange logic in the upstream code imo

andrewsavage1 (approved):
Looks like this just follows upstream patterns so lgtm

cobalt/shell/app/shell_crash_reporter_client.cc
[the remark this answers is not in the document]
rakuco: Full disclosure, I wrote the upstream version, but the code I changed was also confusing :-)

I'll add a note to myself to make the Chromium code more readable and then backport it once the change lands.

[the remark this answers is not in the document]
andrewsavage1: oops! I should keep my comments to myself 😆

hlwarriner:
I was on vacation but am back today and have begun looking at this. I'll reply tomorrow with any feedback.

cobalt/shell/app/shell_crash_reporter_client.cc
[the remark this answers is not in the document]
rakuco: That's the push I needed :-) The good news is that since this PR hasn't been merged yet I was able to clean up the code upstream and port the change to this patch.

cobalt/app/tvos/main.mm
```diff
@@ -62,6 +64,11 @@ - (void)scene:(UIScene*)scene
 }
 
 - (void)sceneWillEnterForeground:(UIScene*)scene {
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          switches::kEnableCrashReporter)) {
+    ::crash_reporter::ProcessIntermediateDumps();
```
hlwarriner: From the function comment at https://source.chromium.org/chromium/chromium/src/+/main:components/crash/core/app/crashpad.h;l=256, we also need to call `StartProcessingPendingReports()` in order to trigger uploads of the generated minidumps. Is that called elsewhere? Or do you plan to enable uploads in a subsequent PR?

hlwarriner: I noticed that Chrome on iOS runs this function as a task in the thread pool, with `base::MayBlock()` and `base::TaskPriority::BEST_EFFORT`: https://source.chromium.org/chromium/chromium/src/+/main:ios/chrome/browser/crash_report/model/crash_helper.mm;l=191.

Given Chromium's recommendations around tasks/threading and the fact that this function involves file I/O, we should probably do the same thing. Can you please make that change or explain why you think we should execute the function directly on the current thread?

I'm also curious if you've reviewed `ios/chrome/browser/crash_report/model/` in general? There may be other logic there that we would find useful / want to reuse.

cobalt/shell/app/shell_crash_reporter_client.cc
```diff
@@ -40,13 +58,10 @@ base::FilePath ShellCrashReporterClient::GetReporterLogFilename() {
 #endif
 
 bool ShellCrashReporterClient::GetCrashDumpLocation(base::FilePath* crash_dir) {
-  if (!base::CommandLine::ForCurrentProcess()->HasSwitch(
```
hlwarriner: Thanks to #9233 we now have `cobalt/app/cobalt_crash_reporter_client.cc`. I feel it would be better to confine Cobalt's logic to customize its implementation of `crash_reporter::CrashReporterClient` in that single location. Would it be feasible for you to move these iOS customizations there?

Sorry about the inconvenience: it looks like #9233 landed while you were working on this.

hlwarriner:
CC @yell0wd0g

cobalt/app/tvos/main.mm
rakuco: I originally intended to deal with uploads separately, but since it just takes one call (together with the changes in #9506), I've added the changes to this PR.

cobalt/shell/app/shell_crash_reporter_client.cc
rakuco: Done in #9506, basically. There's another patch I have locally that needs that one to land that changes some tvOS code paths in `cobalt/app/cobalt_crash_reporter_client.cc` and `cobalt/app/cobalt_main_delegate.cc` that I will submit once #9506 lands.

cobalt/app/tvos/main.mm
rakuco: I hadn't paid much attention to the `ios/` implementation before because it was quite convoluted compared to what we needed here. It was a good suggestion though, and I've rewritten the code here to handle processing intermediate dumps _and_ enabling reporting in a separate thread called from `sceneDidBecomeActive`.

rakuco:
@hlwarriner I think I've addressed all your comments, please take another look.

The integration with `CobaltCrashReporterClient` is being done in #9506 (and a follow-up I need to send later), so this PR now only deals with activating the iOS/tvOS-specific parts from the right place.

cobalt/app/tvos/main.mm
hlwarriner: Nice! I know there is work to be done around product name, but what is the current state with this PR? That is, what is the product name for uploaded reports and how is it generated?

rakuco: This PR depends on #9506 to initialize Crashpad (I'll switch it back to draft to indicate it shouldn't be merged before that one), but other than that it basically works.

The product name is retrieved from [`crashpad_ios.mm`](https://github.com/youtube/cobalt/blob/ad583428b6cfb358203f31352a8c7feb81c8f8af/components/crash/core/app/crashpad_ios.mm#L35), which in practice means the product name comes from `CFBundleName` (currently set to `cobalt` by GN automatically, see b/493484176).

rakuco:
@fayev-g can you approve/merge this one as cobalt-tvos member?

PR #9390, opened 2026-03-06
Set --max-old-space-size=512 as default V8 flag.

sherryzy:
Based on experimentation, increasing max-old-space-size to 512 MiB has shown positive metrics impact. This change sets 512 as the default value while maintaining the ability to override it via the V8_MAX_OLD_SPACE_SIZE switch for further testing.

Bug: 477955599


johnxwork:
Enabled auto-merge. Just double check that you have already disabled other conflicting settings such as optimize-for-size and low-end-mode? I believe I've seen those PRs but just make sure this will take effect.

sherryzy:
The setting in this PR will override the value from low-memory and optimize-for-size. The change the other parameter --max-semi requires resolving the conflicts from optimize-for-size.

PR #9409, opened 2026-03-09
tvos: Enable VP9 SW decode unconditionally

abhijeetk:
Remove SB_IS_ARCH_ARM/ARM64 guards that restricted VP9 software decoding
to ARM devices only. VP9 SW decode is now available on all tvOS targets
unconditionally.

Bug: 488278125


abhijeetk:
cc : @Gyuyoung @jkim-julie

PR #9418, opened 2026-03-09
cobalt: Rework storage migration for early execution

zhongqiliang:
Rework the storage migration pipeline to ensure legacy storage (cookies
and LocalStorage) is correctly transferred to Chromium's storage
services before the main application launches.

go/chrobalt-pre-initialization-storage-migration-pipeline

Bug: 489453068

### DIFF
--- a/cobalt/browser/cobalt_browser_main_parts.cc
+++ b/cobalt/browser/cobalt_browser_main_parts.cc
@@ -20,6 +20,7 @@
 #include "base/command_line.h"
 #include "base/path_service.h"
 #include "base/run_loop.h"
+#include "base/sequence_checker.h"
 #include "base/trace_event/memory_dump_manager.h"
 #include "cobalt/browser/features.h"
 #include "cobalt/browser/global_features.h"
@@ -154,6 +155,7 @@ void CobaltBrowserMainParts::StartStorageMigration() {
 }
 
 void CobaltBrowserMainParts::OnMigrationComplete() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
   CHECK(sequence_checker_.CalledOnValidSequence());
   LOG(INFO) << "Migration complete. Proceeding with deferred launchShell.";
   migration_finished_ = true;
@@ -164,6 +166,7 @@ void CobaltBrowserMainParts::OnMigrationComplete() {
 
 void CobaltBrowserMainParts::PostOrRunIfStorageMigrationFinished(
     base::OnceClosure task) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
   CHECK(sequence_checker_.CalledOnValidSequence());
   if (!migration_finished_) {
     LOG(INFO) << "Deferring launchShell until migration finishes.";


cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,121 @@ constexpr char kLocalStorageCountHistogram[] =
...
+              if (base::PathExists(old_cache_subpath)) {
+                base::DeletePathRecursively(old_cache_subpath);
+              }
+            }
+          }));
+}
+
+// Creates a task that triggers the deletion of old cache directories and writes
+// the migration sentinel file, then immediately advances the migration pipeline.
+Task DeleteOldCacheDirectoryTask() {
```
haozheng-cobalt: This will not delete the sentinel file right? might be better to put the cache files names in macros for file paths rather than hardcode them in the functions.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -444,6 +494,75 @@ MigrationManager::ToCanonicalCookies(const cobalt::storage::Storage& storage) {
...
+            FROM_HERE,
+            base::BindOnce(
+                [](scoped_refptr<SharedClosureState> state) {
+                  if (state->cb) {
+                    LOG(ERROR) << "Cookie injection timed out after 2 "
+                                  "seconds. Proceeding anyway.";
+                    state->Run();
+                  }
+                },
+                shared_state),
```
haozheng-cobalt: My major concern of the PR is here, a 2-second fallback timeout for Mojo IPC calls during application initialization is a massive latency hit. Will 200ms - 500ms to be more reasonable?

cobalt/shell/browser/shell_browser_main_parts.h
```diff
@@ -56,6 +56,10 @@ class ShellBrowserMainParts : public BrowserMainParts {
   void PostMainMessageLoopRun() override;
   void PostDestroyThreads() override;
 
+  // Called to defer tasks (e.g. creating WebContents) until migration is
+  // finished.
+  virtual void PostOrRunIfMigrationFinished(base::OnceClosure task);
```
haozheng-cobalt: Maybe rename the function to `PostOrRunIfStorageMigrationFinished` to be clearer?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,121 @@ constexpr char kLocalStorageCountHistogram[] =
...
+Task DeleteOldCacheDirectoryTask() {
+  return base::BindOnce([](base::OnceClosure callback) {
+    DeleteOldCacheDirectoryAsync();
+    WriteMigrationSentinelAsync();
+    std::move(callback).Run();
+  });
+}
+
+// Emits UMA metrics for the total duration of the migration process.
+Task RecordMetricsTask(std::unique_ptr<base::ElapsedTimer> elapsed_timer) {
```
haozheng-cobalt: Would be better to change this function name to reflect that this function monitors time elapse, as we monitor multiple metrics in this class now.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,121 @@ constexpr char kLocalStorageCountHistogram[] =
...
+  }
+  return base::FilePath(path.data());
+}
+
+// Retrieves the path to the sentinel file used to fast-path skip the migration.
+// If this file exists, it means the migration was already completed on a prior
+// app launch.
+base::FilePath GetMigrationSentinelPath() {
+  base::FilePath current_cache_path;
+  if (!base::PathService::Get(base::DIR_CACHE, &current_cache_path)) {
```
jellefoks, line 127: `SbStorageOpenRecord` from `starboard/shared/starboard/file_storage/storage_open_record.cc` uses `GetStorageFilePath` from `starboard/shared/starboard/file_storage/storage_internal.h`  to which calls `GetHomeDirectory` which on Android  (`starboard/android/shared/get_home_directory.cc`)  returns `g_app_files_dir` which  is initialized by the `files_dir` parameter to `SbFileAndroidInitialize` (from `ApplicationAndroid`), from `j_files_dir` in `starboard_bridge.cc`, which gets it from `mAppContext.getFilesDir().getAbsolutePath()`.

Now, `g_app_files_dir` is returned from `kSbSystemPathStorageDirectory` and `kSbSystemPathFilesDirectory`, albeit those add '/storage' and '/files' to the `g_app_files_dir` path, and is *NOT* returned from `kSbSystemPathCacheDirectory` (see `starboard/android/shared/system_get_path.cc`, and `copied_base/base/base_paths_starboard.cc`)...

So, this way, if the device/user clears the cache but not the storage, then the migration will happen *again*, possibly restoring a different user if the user had switched after the last migration.

Ideally, we'd store the sentinel in the same folder as the storage record, so if we can call it from here, `GetHomeDirectory` instead of using DIR_CACHE.

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,47 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+  int result = ShellBrowserMainParts::PreMainMessageLoopRun();
+
+  StartStorageMigration();
+
+  return result;
+}
+
+void CobaltBrowserMainParts::StartStorageMigration() {
+  DLOG(INFO) << "CobaltBrowserMainParts::StartStorageMigration started.";
+  content::StoragePartition* partition =
```
haozheng-cobalt, line 148: This StoragePartition is checked multiple times in the storage migration manager, maybe worth passing it as param to the functions.

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,47 @@ int CobaltBrowserMainParts::PreCreateThreads() {
 
 int CobaltBrowserMainParts::PreMainMessageLoopRun() {
   StartMetricsRecording();
-  return ShellBrowserMainParts::PreMainMessageLoopRun();
+  int result = ShellBrowserMainParts::PreMainMessageLoopRun();
+
+  StartStorageMigration();
```
haozheng-cobalt, line 141: Should we gate all changes in the file cobalt/browser/cobalt_browser_main_parts.cc and cobalt/browser/cobalt_browser_main_parts.h in `#if BUILDFLAG(IS_ANDROIDTV)`?

johnxwork:
Can we get corresponding unittest added to migration_manager_unittest.cc?

cobalt/browser/cobalt_browser_main_parts.cc
haozheng-cobalt: Might be ok not to gate because this change will only be in 26.android, but ideally we might better add gate.

zhongqiliang: This code is common to evergreen, and it should also run this.

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: content::StoragePartition* partition is passed as a param already. Can you clarify your ask?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: No, it is different root directory. See it at the end of the design doc.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
jellefoks: I think the solution here is to use DIR_HOME and update `GetHomeDir()` in `base/files/file_util_posix.cc` to return the `kSbSystemPathFilesDirectory`, but to also then join `..` to go to the same folder where `SbStorageOpenRecord` stores.

Or, simpler, use DIR_HOME and just `setenv('HOME', g_app_files_dir);` where `g_app_files_dir` is initialized.

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,47 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+  DLOG(INFO) << "Migration complete. Proceeding with deferred tasks.";
+  migration_finished_ = true;
+  for (auto& task : pending_tasks_) {
+    std::move(task).Run();
+  }
+  pending_tasks_.clear();
+}
+
+void CobaltBrowserMainParts::PostOrRunIfMigrationFinished(
+    base::OnceClosure task) {
```
johnxwork, line 168: We need thread checker to make sure the `pending_tasks_` are executed on the same thread they are posted from.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
jellefoks: We spoke offline, and the sentinel is created right after the SbStorage record file is unlinked (deleted), so this is not going to be an issue, the sentinel existing or not existing is harmless.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: Done

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: synced offline, and reached an agreement to keep it as it is.

cobalt/shell/browser/shell_browser_main_parts.h
zhongqiliang: done

cobalt/browser/cobalt_browser_main_parts.cc
haozheng-cobalt: The call to check the storage migration completion and postpone window creation is placed inside `cobalt/shell/android/shell_manager.cc`, Are we missing the corresponding logics on other platforms?

zhongqiliang: yes, evergreen needs to hook this up in a separate PR.

cobalt/browser/cobalt_browser_main_parts.cc
haozheng-cobalt: You're right, sorry I looked at the wrong PR, I was cross referencing yours and mine, the partition has been passed in in this PR.

cobalt/browser/cobalt_browser_main_parts.cc
haozheng-cobalt: We should create a TODO ticket for other platforms. 
udpate: I created one b/493679298

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: This is a great point!!! Thanks.

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -82,8 +84,15 @@ void InitializeBrowserMemoryInstrumentationClient() {
...
 }  // namespace
 
-CobaltBrowserMainParts::CobaltBrowserMainParts(const std::string& deep_link)
-    : ShellBrowserMainParts(deep_link) {}
+void CobaltBrowserMainParts::InitializeMessageLoopContext() {
+  // On Android, we completely defer WebContents creation until the Java layer
+  // invokes ShellManager.launchShell() which reaches C++ via JNI.
+  // On Linux, we might need the base behavior.
+  // go/chrobalt-pre-initialization-storage-migration-pipeline
+#if !BUILDFLAG(IS_ANDROID)
```
vincentmo, line 102: Just checking, but should this be checking IS_ANDROIDTV instead? If not, what's the difference between the 2?

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,49 @@ int CobaltBrowserMainParts::PreCreateThreads() {
 
 int CobaltBrowserMainParts::PreMainMessageLoopRun() {
   StartMetricsRecording();
-  return ShellBrowserMainParts::PreMainMessageLoopRun();
+  int result = ShellBrowserMainParts::PreMainMessageLoopRun();
```
vincentmo, line 133: If PreMainMessageLoopRun fails, should StartStorageMigration still run? If so, why?

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,49 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+  content::StoragePartition* partition =
+      browser_context()->GetDefaultStoragePartition();
+
+  if (partition) {
+    cobalt::migrate_storage_record::MigrationManager::RunMigration(
+        partition, base::BindOnce(&CobaltBrowserMainParts::OnMigrationComplete,
+                                  base::Unretained(this)));
+  } else {
+    LOG(ERROR) << "BrowserContext not available for storage migration.";
+    OnMigrationComplete();
```
vincentmo: When would storage partition be null? In this scenario, will the web app be loaded w/o migration taking place?

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,49 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+  LOG(INFO) << "CobaltBrowserMainParts::StartStorageMigration started.";
+  content::StoragePartition* partition =
+      browser_context()->GetDefaultStoragePartition();
+
+  if (partition) {
+    cobalt::migrate_storage_record::MigrationManager::RunMigration(
+        partition, base::BindOnce(&CobaltBrowserMainParts::OnMigrationComplete,
+                                  base::Unretained(this)));
+  } else {
+    LOG(ERROR) << "BrowserContext not available for storage migration.";
```
vincentmo: Will we add telemetry logging so we know how often this happens?

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,7 +112,49 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+    std::move(task).Run();
+  }
+  pending_tasks_.clear();
+}
+
+void CobaltBrowserMainParts::PostOrRunIfStorageMigrationFinished(
+    base::OnceClosure task) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!migration_finished_) {
+    LOG(INFO) << "Deferring launchShell until migration finishes.";
```
vincentmo, line 172: Is the assumption that only 1 task (launchShell) is run (based on the logging)? If so, why is a vector used then?

cobalt/browser/cobalt_browser_main_parts.h
```diff
@@ -71,6 +73,20 @@ class CobaltBrowserMainParts : public content::ShellBrowserMainParts {
...
+  // C25 Storage Migration
+  void StartStorageMigration();
+  void OnMigrationComplete();
+  void PostOrRunIfStorageMigrationFinished(base::OnceClosure task) override;
+
+ protected:
+  void InitializeMessageLoopContext() override;
+
+ private:
+  bool migration_finished_ = false;
```
vincentmo: Are there any annotations/comments about whether this needs to be accessed w/ synchronization or thread sequence checks?

cobalt/shell/android/shell_manager.cc
```diff
@@ -65,14 +67,31 @@ static void JNI_ShellManager_Init(JNIEnv* env,
...
+  std::string topic = base::android::ConvertJavaStringToUTF8(env, jtopic);
+
+  // We wrap the WebContents creation into a task because we must ensure
+  // that any asynchronous storage migration is completely finished before
+  // the browser accesses the default storage partition. If WebContents
+  // is created too early, the JS environment will start with empty data.
+  auto create_window_task = base::BindOnce(
+      [](GURL url, std::string topic) {
+        ShellBrowserContext* browserContext =
+            ShellContentBrowserClient::Get()->browser_context();
```
vincentmo, line 98: Can either ShellContentBrowserClient::Get() or browserContext be null here that this should check for?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,123 @@ constexpr char kLocalStorageCountHistogram[] =
...
+// Writes the sentinel file `migration_completed.txt` in the background so that
+// future app launches can skip the migration process.
+void WriteMigrationSentinelAsync() {
+  base::ThreadPool::CreateSequencedTaskRunner(
+      {base::MayBlock(), base::TaskPriority::BEST_EFFORT,
+       base::TaskShutdownBehavior::CONTINUE_ON_SHUTDOWN})
+      ->PostTask(FROM_HERE, base::BindOnce([]() {
+                   base::FilePath sentinel_path = GetMigrationSentinelPath();
+                   if (!sentinel_path.empty()) {
+                     base::WriteFile(sentinel_path, "1");
```
vincentmo: Let's log when WriteFile fails

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,123 @@ constexpr char kLocalStorageCountHistogram[] =
...
+  return base::FilePath(path.data());
+}
+
+// Retrieves the path to the sentinel file used to fast-path skip the migration.
+// If this file exists, it means the migration was already completed on a prior
+// app launch.
+base::FilePath GetMigrationSentinelPath() {
+  base::FilePath current_cache_path;
+  if (!base::PathService::Get(base::DIR_CACHE, &current_cache_path)) {
+    return base::FilePath();
```
vincentmo, line 129: Should an error be logged here?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,123 @@ constexpr char kLocalStorageCountHistogram[] =
...
+       base::TaskShutdownBehavior::CONTINUE_ON_SHUTDOWN})
+      ->PostTask(
+          FROM_HERE, base::BindOnce([]() {
+            base::FilePath old_cache_path = GetOldCachePath();
+            if (old_cache_path.empty() || !base::PathExists(old_cache_path)) {
+              return;
+            }
+
+            base::FilePath current_cache_path;
+            base::PathService::Get(base::DIR_CACHE, &current_cache_path);
```
vincentmo: Check Get() status? If it fails then the following condition would pass and the old cache would get deleted.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,123 @@ constexpr char kLocalStorageCountHistogram[] =
...
+            if (!old_cache_path.IsParent(current_cache_path) &&
+                old_cache_path != current_cache_path) {
+              base::DeletePathRecursively(old_cache_path);
```
vincentmo: This can also fail, let's check status and log.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +119,123 @@ constexpr char kLocalStorageCountHistogram[] =
...
+                "service_worker_settings.json",
+                "settings.json",
+                "splash",
+                "splash_screen",
+                "uncompiled_js",
+            };
+            for (const auto& subpath : old_cache_subpaths) {
+              base::FilePath old_cache_subpath = old_cache_path.Append(subpath);
+              if (base::PathExists(old_cache_subpath)) {
+                base::DeletePathRecursively(old_cache_subpath);
```
vincentmo: Same, let's check status and log

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -293,139 +406,79 @@ Task MigrationManager::GroupTasks(std::vector<Task> tasks) {
...
+  // Build the pipeline of tasks to execute asynchronously.
   std::vector<Task> tasks;
-  tasks.push_back(CookieTask(weak_document_ptr, ToCanonicalCookies(*storage)));
-  const url::Origin& origin = render_frame_host->GetLastCommittedOrigin();
-  tasks.push_back(LocalStorageTask(weak_document_ptr,
+  tasks.push_back(CookieTask(partition, ToCanonicalCookies(*storage)));
+  tasks.push_back(LocalStorageTask(partition, origin,
                                    ToLocalStorageItems(origin, *storage)));
-  tasks.push_back(RecordMetricsTask(std::move(elapsed_timer)));
+  tasks.push_back(RecordMigrationDurationTask(std::move(elapsed_timer)));
```
vincentmo, line 536: Can we also add UMA that tracks migration outcome: fast path (sentinal found), skipped (no legacy data found), migration success, migration failed?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -293,139 +406,79 @@ Task MigrationManager::GroupTasks(std::vector<Task> tasks) {
...
+  // Fast Path: If the sentinel file exists, migration is already done.
+  base::ElapsedTimer sentinel_timer;
+  base::FilePath sentinel_path = GetMigrationSentinelPath();
+  bool sentinel_exists =
+      !sentinel_path.empty() && base::PathExists(sentinel_path);
+  if (sentinel_exists) {
+    LOG(INFO) << "Migration sentinel file found. Skipping migration.";
+    LOG(INFO) << "RunMigration fast path took "
+              << elapsed_timer->Elapsed().InMilliseconds() << " ms.";
+    std::move(done_callback).Run();
```
vincentmo, line 492: Should we maintain a separate fast path duration UMA so we can track how the disk read affects start up?

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -131,6 +290,9 @@ std::unique_ptr<cobalt::storage::Storage> ReadStorage() {
     auto bytes = std::vector<uint8_t>(record->GetSize());
     const int read_result =
         record->Read(reinterpret_cast<char*>(bytes.data()), bytes.size());
+
+    LOG(INFO) << "Read " << read_result << " bytes from legacy storage.";
+
     if (!record->Delete()) {
```
vincentmo: Does deleting the record before migration is complete introduce a scenario of a permanent data loss? Why isn't delete done only after migraiton is done?

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: The code here is correct. 

For Android, we do not want to use the default behavior of ShellBrowserMainParts::InitializeMessageLoopContext(); because we hijacked it to wait for migration to be done.

I did not test on linux, so keep it unchanged.

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: it should not! updated the code

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: It will not be null. I removed the if else check.

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: this is removed.

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: It is not necessary, removed the vector.
Thought about make this generic.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
haozheng-cobalt: @vincentmo  I believe permanent data loss is preferable to repeated migration attempts. We should only attempt storage migration ONCE, on a best-effort basis. If the migration fails, let it fail, the user will need to log in to their account manually after the app upgrades. We should NOT attempt to migrate the storage data again, as this could unexpectedly overwrite the new login data.

cobalt/browser/cobalt_browser_main_parts.h
zhongqiliang: made it GUARDED by sequence_checker_

cobalt/shell/android/shell_manager.cc
zhongqiliang: It won't be null here. browserContext() is already valid in the migration task.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: yes, added

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: sure

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: done

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: added logs

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: done

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: this is refactored.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: keep it as it for now, there is no data to indicate this is too long yet.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: done

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: done

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -116,7 +128,51 @@ int CobaltBrowserMainParts::PreMainMessageLoopRun() {
...
+    std::move(pending_task_).Run();
+  }
+}
+
+void CobaltBrowserMainParts::PostOrRunIfStorageMigrationFinished(
+    base::OnceClosure task) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!migration_finished_) {
+    LOG(INFO) << "Deferring launchShell until migration finishes.";
+    DCHECK(!pending_task_) << "Only one storage migration task is supported.";
```
johnxwork: I don't think DCHECK() is hit even in qa build. We can turn this into a CHECK()

cobalt/shell/android/shell_manager.cc
```diff
@@ -66,13 +70,46 @@ static void JNI_ShellManager_Init(JNIEnv* env,
...
-  Shell::CreateNewWindow(browserContext, url, nullptr, gfx::Size(),
-                         switches::ShouldCreateSplashScreen(), deeplink_url);
+
+  // We wrap the WebContents creation into a task because we must ensure
+  // that any asynchronous storage migration is completely finished before
+  // the browser accesses the default storage partition. If WebContents
+  // is created too early, the JS environment will start with empty data.
+  auto create_window_task = base::BindOnce(
+      [](GURL url, std::string deeplink_url) {
+        std::string status_param = cobalt::migrate_storage_record::
```
johnxwork: nit: const std::string

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -116,7 +128,51 @@ int CobaltBrowserMainParts::PreMainMessageLoopRun() {
...
+      browser_context()->GetDefaultStoragePartition();
+
+  DCHECK(partition);
+  cobalt::migrate_storage_record::MigrationManager::RunMigration(
+      partition, base::BindOnce(&CobaltBrowserMainParts::OnMigrationComplete,
+                                base::Unretained(this)));
+}
+
+void CobaltBrowserMainParts::OnMigrationComplete() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
```
johnxwork: DCHECK_CALLED_ON_VALID_SEQUENCE is not compiled in our builds since it's gated by DCHECK.

We can just use SequenceCheckerImpl directly.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -80,6 +102,57 @@ constexpr char kLocalStorageCountHistogram[] =
...
+void WriteMigrationSentinelAsync(scoped_refptr<MigrationState> state) {
+  base::ThreadPool::CreateSequencedTaskRunner(
+      {base::MayBlock(), base::TaskPriority::BEST_EFFORT,
+       base::TaskShutdownBehavior::CONTINUE_ON_SHUTDOWN})
+      ->PostTask(
+          FROM_HERE,
+          base::BindOnce(
+              [](scoped_refptr<MigrationState> state) {
+                base::FilePath sentinel_path = GetMigrationSentinelPath();
+                if (!sentinel_path.empty()) {
```
johnxwork: Shall we log to UMA when the sentinel path is empty? This will lead to repeated storage migration in the future.

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
```diff
@@ -293,139 +449,91 @@ Task MigrationManager::GroupTasks(std::vector<Task> tasks) {
...
-  return GroupTasks(std::move(tasks));
+  tasks.push_back(CookieTask(partition, ToCanonicalCookies(*storage), state));
+  tasks.push_back(LocalStorageTask(
+      partition, origin, ToLocalStorageItems(origin, *storage), state));
+  tasks.push_back(RecordMigrationDurationTask(std::move(elapsed_timer)));
+  tasks.push_back(CleanupLegacyFilesTask(state));
+
+  // Execute the chained tasks. The done_callback will be invoked when all
+  // steps, including async background work, have completed.
+  std::move(GroupTasks(std::move(tasks))).Run(std::move(done_callback));
```
johnxwork: We are not running asynchronously here. Why not just run the functions sequentially directly? Do you know what's the purpose to put them into tasks and execute like this?

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: done

cobalt/shell/android/shell_manager.cc
zhongqiliang: done

cobalt/browser/cobalt_browser_main_parts.cc
zhongqiliang: done

cobalt/shell/browser/migrate_storage_record/migration_manager.cc
zhongqiliang: done

PR #9423, opened 2026-03-09
cobalt: Implement app lifecycle and window management

jellefoks:
Introduce the core infrastructure for handling application lifecycle events and enhance Starboard window management robustness.

With this commit, conceal, reveal, focus, and blur functionality is fully operational with respect to lazy Starboard window creation/destruction and the corresponding Web API focus and visibility states.

Key Changes:
- Implement static entry points in content::Shell for lifecycle events (Blur, Focus, Conceal, Reveal, Freeze, Unfreeze, Stop).
- Add virtual handlers to ShellPlatformDelegate to propagate these events.
- Implement explicit focus/blur propagation to WebContents in ShellPlatformDelegate (OnFocus calls WebContents::Focus, OnBlur calls RenderWidgetHost::Blur).
- Update AppLifecycleDelegate to bridge global Starboard events to the newly added Shell handlers.
- Enhance PlatformWindowStarboard to robustly manage window creation, destruction, and visibility transitions, preventing redundant events.
- Map SIGWINCH to the Blur event in the POSIX signal handler to support integrated lifecycle testing.
- Introduce test_lifecycle.sh and enhance test_preload.sh to verify Visibility and Focus API states using the Chrome DevTools Protocol (CDP).
- Fix cyclic dependencies in Starboard Ozone by removing redundant ui/aura and ui/compositor inclusions.
- Ensure proper widget activation and visibility management in DesktopWindowTreeHostPlatform for Starboard.
- Enhance AppLifecycleDelegateTest to verify WebContents focus state transitions.

All changes are verified with unit tests and automated integration scripts on linux-x64x11-modular.

Bug: 448156280

### DIFF
--- a/starboard/shared/x11/application_x11.cc
+++ b/starboard/shared/x11/application_x11.cc
@@ -897,12 +897,10 @@ Application::Event* ApplicationX11::WaitForSystemEventWithTimeout(
 }
 
 void ApplicationX11::WakeSystemEventWait() {
-  if (!windows_.empty()) {
-    XSendAtom((*windows_.begin())->window, wake_up_atom_);
-  }
-  if (dev_input_) {
-    dev_input_->WakeSystemEventWait();
-  }
+  SB_CHECK(!windows_.empty());
+  XSendAtom((*windows_.begin())->window, wake_up_atom_);
+  SB_CHECK(dev_input_);
+  dev_input_->WakeSystemEventWait();
 }
 
 bool ApplicationX11::EnsureX() {
--- a/ui/ozone/platform/starboard/platform_window_starboard.cc
+++ b/ui/ozone/platform/starboard/platform_window_starboard.cc
@@ -77,18 +77,7 @@ PlatformWindowStarboard::PlatformWindowStarboard(
         ->AddPlatformEventObserverStarboard(this);
   }
 
-  SbWindowOptions options{};
-  SbWindowSetDefaultOptions(&options);
-  options.size.width = bounds.width();
-  options.size.height = bounds.height();
-  sb_window_ = SbWindowCreate(&options);
-  CHECK(SbWindowIsValid(sb_window_));
-
-  (*g_created_callback).Run(sb_window_);
-
-  widget_available_ = true;
-  delegate->OnAcceleratedWidgetAvailable(
-      reinterpret_cast<intptr_t>(SbWindowGetPlatformHandle(sb_window_)));
+  Show(false /* inactive */);
 }
 
 PlatformWindowStarboard::~PlatformWindowStarboard() {
@@ -163,34 +152,33 @@ gfx::Rect PlatformWindowStarboard::GetBoundsInDIP() const {
 }
 
 void PlatformWindowStarboard::Show(bool inactive) {
-  if (!SbWindowIsValid(sb_window_)) {
-    SbWindowOptions options{};
-    SbWindowSetDefaultOptions(&options);
-    options.size.width = bounds_.width();
-    options.size.height = bounds_.height();
-    sb_window_ = SbWindowCreate(&options);
-    CHECK(SbWindowIsValid(sb_window_));
-
-    (*g_created_callback).Run(sb_window_);
-  }
-
-  if (SbWindowIsValid(sb_window_) && !widget_available_) {
-    widget_available_ = true;
-    delegate_->OnAcceleratedWidgetAvailable(
-        reinterpret_cast<intptr_t>(SbWindowGetPlatformHandle(sb_window_)));
-  }
-}
-
-void PlatformWindowStarboard::Hide() {
-  if (!SbWindowIsValid(sb_window_)) {
+  if (SbWindowIsValid(sb_window_)) {
     return;
   }
 
+  SbWindowOptions options{};
+  SbWindowSetDefaultOptions(&options);
+  options.size.width = bounds_.width();
+  options.size.height = bounds_.height();
+  sb_window_ = SbWindowCreate(&options);
+  CHECK(SbWindowIsValid(sb_window_));
+
+  (*g_created_callback).Run(sb_window_);
+
+  CHECK(!widget_available_);
+  widget_available_ = true;
+  delegate_->OnAcceleratedWidgetAvailable(
+      reinterpret_cast<intptr_t>(SbWindowGetPlatformHandle(sb_window_)));
+}
+
+void PlatformWindowStarboard::Hide() {
+  CHECK(widget_available_);
   if (widget_available_) {
     widget_available_ = false;
     delegate_->OnAcceleratedWidgetDestroyed();
   }
 
+  CHECK(SbWindowIsValid(sb_window_));
   (*g_destroyed_callback).Run(sb_window_);
 
   SbWindowDestroy(sb_window_);


starboard/shared/signal/suspend_signals.cc
```diff
@@ -65,6 +65,13 @@ void Conceal(int signal_id) {
   SignalMask(kAllSignals, SIG_UNBLOCK);
 }
 
+void Blur(int signal_id) {
+  SignalMask(kAllSignals, SIG_BLOCK);
+  LogSignalCaught(signal_id);
+  Application::Get()->Blur(NULL, NULL);
```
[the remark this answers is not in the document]
jellefoks: Thanks for the suggested cleanup. Done.

cobalt/app/app_lifecycle_delegate.cc
```diff
@@ -144,7 +144,20 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
...
+      content::Shell::OnBlur();
+#if BUILDFLAG(IS_STARBOARD)
+      if (platform_event_source_) {
+        platform_event_source_->HandleFocusEvent(event);
+      }
+#else
+      // On Android, focus is handled natively via Activity and View lifecycle
+      // callbacks (e.g. CobaltActivity.onResume) which propagate directly to
+      // Chromium's WindowAndroid.
+      NOTIMPLEMENTED();
```
andrewsavage1: Should we call NOTIMPLEMENTED if it sounds like it is implemented elsewhere?

cobalt/shell/browser/shell_platform_delegate.cc
```diff
@@ -25,16 +27,64 @@ bool ShellPlatformDelegate::IsVisible() const {
   return is_visible_;
 }
 
+__attribute__((weak)) void ShellPlatformDelegate::Initialize(
```
andrewsavage1: Why do we need the weak attribute? Can we remove it?

cobalt/shell/browser/shell_test_support.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+// Copyright 2024 The Cobalt Authors. All Rights Reserved.
```
andrewsavage1: Change back to 2026

starboard/shared/x11/application_x11.cc
```diff
@@ -897,10 +897,12 @@ Application::Event* ApplicationX11::WaitForSystemEventWithTimeout(
 }
 
 void ApplicationX11::WakeSystemEventWait() {
-  SB_DCHECK(!windows_.empty());
-  XSendAtom((*windows_.begin())->window, wake_up_atom_);
-  SB_DCHECK(dev_input_);
-  dev_input_->WakeSystemEventWait();
+  if (!windows_.empty()) {
```
andrewsavage1: Why does this function need to change? Will we have cases in Chrobalt (as opposed to Cobalt 25 and earlier) where these checks would fail?

jellefoks:
Whoop! this is ready for review.

cobalt/app/app_lifecycle_delegate.cc
```diff
@@ -144,7 +144,20 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
...
     case kSbEventTypeBlur:
+      content::Shell::OnBlur();
+#if BUILDFLAG(IS_STARBOARD)
+      if (platform_event_source_) {
+        platform_event_source_->HandleFocusEvent(event);
+      }
+#else
+      // On Android, focus is handled natively via Activity and View lifecycle
+      // callbacks (e.g. CobaltActivity.onResume) which propagate directly to
+      // Chromium's WindowAndroid.
```
yell0wd0g: nit: This `!IS_STARBOARD` branch would apply to tvOS as well, not 
just Android.

But looking at BUILD.gn, the associated unittest file seems to
only be built for starboard platforms ([code](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/26.eap:cobalt/BUILD.gn;l=201?q=app_lifecycle_delegate.cc%20&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2F26.eap:cobalt%2F)) and
a quick glance at an android-arm-qa build like [here](https://github.com/youtube/cobalt/actions/runs/22967686521/job/66675953118?pr=9423) doesn't list it, 
so this buildflag should rather be:
```
#if !BUILDFLAG(IS_STARBOARD)
#error This file is only intended for Starboard platforms.
// Other platforms route events differently, e.g. on AndroidTV
// focus is handled natively... (same text you have).
#endif
```

cobalt/shell/browser/shell_test_support.h
yell0wd0g: Is this change intended? Usually file copyright years never change.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -161,12 +161,16 @@ CobaltContentBrowserClient::CobaltContentBrowserClient(
...
+  ui::PlatformWindowStarboard::SetWindowDestroyedCallback(
+      base::BindRepeating(&CobaltContentBrowserClient::OnSbWindowDestroyed,
+                          weak_factory_.GetWeakPtr()));
 #endif  // BUILDFLAG(IS_STARBOARD)
 }
 
 CobaltContentBrowserClient::~CobaltContentBrowserClient() {
 #if BUILDFLAG(IS_STARBOARD)
   ui::PlatformWindowStarboard::SetWindowCreatedCallback(base::NullCallback());
+  ui::PlatformWindowStarboard::SetWindowDestroyedCallback(base::NullCallback());
```
yell0wd0g: Just double checking: a NullCallback will crash (DCHECK) if
called (unless is tested, which is rarely done). Is this 
intended here? If the intention is to allow for such 
callbacks but do nothing, then `base::DoNothing()` is the 
appropriate callback.

cobalt/shell/browser/shell.cc
```diff
@@ -322,6 +365,12 @@ void Shell::Shutdown() {
     return;
   }
 
+  static bool is_shutting_down = false;
```
yell0wd0g: This is to protect the critical section in l.374-395
from reentrancy, right? If so then I'd rather use
concurrency specific patterns like a homebrew:

```
  std::atomic_flag is_shutting_down = ATOMIC_FLAG_INIT;;
  if (is_shutting_down.test_and_set()) {
    return;
  }
  // ...
  is_shutting_down.clear();
```

Or a `base::Lock` ([code](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:base/synchronization/lock.h;l=30;drc=21fb7d148f9e0948484b16d75a2467ef6c4b0787)).

cobalt/shell/browser/shell_platform_delegate.cc
```diff
@@ -25,16 +27,64 @@ bool ShellPlatformDelegate::IsVisible() const {
...
+    const gfx::Size& default_window_size,
+    bool is_visible) {
+  is_visible_ = is_visible;
+}
+
+void ShellPlatformDelegate::OnBlur() {
+  for (auto* shell : Shell::windows()) {
```
yell0wd0g, line 32: Question: Under what condition would Cobalt have multiple 
`windows()`? If none can be conjured then consider leaving 
a `CHECK_EQ(Shell::windows().size(), 1u)` and removing the
`for` loop. Same comment in the other methods below.

Note that `this` has a `web_contents()` and a
`splash_screen_web_contents()` IIUC.

cobalt/shell/browser/shell_platform_delegate.cc
```diff
@@ -25,16 +27,64 @@ bool ShellPlatformDelegate::IsVisible() const {
...
+}
+
+void ShellPlatformDelegate::OnFocus() {
+  for (auto* shell : Shell::windows()) {
+    shell->web_contents()->Focus();
+  }
+}
+
+void ShellPlatformDelegate::OnConceal() {
+  if (!IsVisible()) {
```
yell0wd0g: Question: is it possible to receive an `OnConceal()` while
not-visible? Same question in l.60. If there is a specific
use case consider leaving a comment, otherwise maybe a 
`CHECK(IsVisible());`?

cobalt/shell/browser/shell_platform_delegate.cc
```diff
@@ -25,16 +27,64 @@ bool ShellPlatformDelegate::IsVisible() const {
   return is_visible_;
 }
 
+__attribute__((weak)) void ShellPlatformDelegate::Initialize(
+    const gfx::Size& default_window_size,
```
yell0wd0g: micro-nit: In Chromium it's common to escape unused params, i.e.
` const gfx::Size& /*default_window_size*/,` 🤷

cobalt/shell/browser/shell_platform_delegate_aura.cc
```diff
@@ -111,6 +107,13 @@ void ShellPlatformDelegate::RevealShell(Shell* shell) {
 
   platform_->aura->host()->Show();
 }
+
+void ShellPlatformDelegate::ConcealShell(Shell* shell) {
+  if (platform_->aura) {
```
yell0wd0g: Is this possible? Maybe if we start the App "minimized"? A 
comment on this would be super helpful for future developers,
e.g.
```
  // This could happen if we start minimized or not visible,
  // e.g. in pre-app configuration
  if (...
```

cobalt/app/app_lifecycle_delegate.cc
jellefoks: Done

cobalt/shell/browser/shell_test_support.h
jellefoks: Oops. Done.

cobalt/browser/cobalt_content_browser_client.cc
jellefoks: They can also be null of SetWindowFooCallback is never called, so the code in platform_window_starboard.cc  referencing the callback already guards against it being null:

```
  if (g_created_callback) {
    (*g_created_callback).Run(sb_window_);
  }
```

But, I'll change the static initializers to `base::DoNothing()`, because that's cleaner. Needs #include "base/functional/callback_helpers.h"

cobalt/shell/browser/shell.cc
jellefoks: This is actually superfluous and was from an earlier attempt to solve a bug that was elsewhere. The g_platform check above covers this. Note: since this is always called from the same thread, it doesn't have to be atomic.

cobalt/shell/browser/shell_platform_delegate.cc
jellefoks: Done.

cobalt/shell/browser/shell_platform_delegate.cc
jellefoks: The fact that `Shell` in `cobalt/shell/browser/shell.h` has support for multiple windows with its `static std::vector<Shell*> windows` means that we should carry that through here. 

We could remove support for multiple windows altogether, but doing that would span outside of the scope of this PR. Filed as b/491969758. I do not know if there currently is an edge case or test setup where there is an additional window.

cobalt/shell/browser/shell_platform_delegate.cc
jellefoks: This mirrored the `if(visible_)` check in OnReveal added for a `LifecycleTest::RedundantReveal` test, but that is the only time it can actually occur in production.

So, good point, removed that test and changed to a CHECK() here and in the other state change functions. A followup PR is already prepared that ensures it will never happen redundantly.

cobalt/shell/browser/shell_platform_delegate_aura.cc
jellefoks: No, is not possible. Was a leftover defensive change. Removed

cobalt/app/app_lifecycle_delegate.cc
jellefoks: This doesn't build on `android-arm64 / arm64_devel` and `android-arm / arm_devel`...

```
2026-03-12T07:39:28.6071814Z #error This file is only intended for Starboard platforms.
2026-03-12T07:39:28.6072446Z  ^
2026-03-12T07:39:28.6073107Z ../../cobalt/app/app_lifecycle_delegate.cc:155:11: error: use of undeclared identifier 'platform_event_source_'
2026-03-12T07:39:28.6073944Z       if (platform_event_source_) {
2026-03-12T07:39:28.6074336Z           ^
2026-03-12T07:39:28.6075066Z ../../cobalt/app/app_lifecycle_delegate.cc:156:9: error: use of undeclared identifier 'platform_event_source_'
2026-03-12T07:39:28.6075955Z         platform_event_source_->HandleFocusEvent(event);
2026-03-12T07:39:28.6076445Z         ^
2026-03-12T07:39:28.6076725Z 3 errors generated.
```

ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -136,20 +150,52 @@ gfx::Rect PlatformWindowStarboard::GetBoundsInDIP() const {
...
+    widget_available_ = true;
+    delegate_->OnAcceleratedWidgetAvailable(
+        reinterpret_cast<intptr_t>(SbWindowGetPlatformHandle(sb_window_)));
+  }
 }
 
 void PlatformWindowStarboard::Hide() {
-  NOTIMPLEMENTED_LOG_ONCE();
+  VLOG(1) << "PlatformWindowStarboard::Hide";
+  if (!SbWindowIsValid(sb_window_)) {
```
andrewsavage1: When would this case be triggered?

ui/ozone/platform/starboard/platform_window_starboard.h
```diff
@@ -16,6 +16,7 @@
 #define UI_OZONE_PLATFORM_STARBOARD_PLATFORM_WINDOW_STARBOARD_H_
 
 #include "base/functional/callback.h"
+#include "base/memory/raw_ptr.h"
```
andrewsavage1, line 19: What is this needed for?

ui/views/widget/desktop_aura/desktop_window_tree_host_platform.cc
```diff
@@ -881,6 +881,11 @@ void DesktopWindowTreeHostPlatform::OnWillDestroyAcceleratedWidget() {
   desktop_native_widget_aura_->OnHostWillClose();
 }
 
+void DesktopWindowTreeHostPlatform::OnAcceleratedWidgetDestroyed() {
```
andrewsavage1, line 885: Should be guarded by buildflags

ui/views/widget/desktop_aura/desktop_window_tree_host_platform.h
```diff
@@ -152,6 +152,7 @@ class VIEWS_EXPORT DesktopWindowTreeHostPlatform
   void OnCloseRequest() override;
   void OnAcceleratedWidgetAvailable(gfx::AcceleratedWidget widget) override;
   void OnWillDestroyAcceleratedWidget() override;
+  void OnAcceleratedWidgetDestroyed() override;
```
andrewsavage1, line 156: Should be guarded by buildflags

cobalt/shell/browser/shell_platform_delegate.cc
jellefoks: Yeah, this is not needed anymore this was the 'base class' implementation added before the platform specializations were added to the non-derived classes in `shell_platform_delegate_(aura|views|android|ios).(cc|mm)`. Same for `RevealShell` and `ConcealShell` below.

cobalt/app/app_lifecycle_delegate.cc
yell0wd0g: Hmm so it is being built. While looking at who pulls from
it I noticed this is 26.eap, which doesn't make sense for
AndroidTV... Should we leave the current change (with `#error`)
and remove the file from android builds?

cobalt/app/app_lifecycle_delegate.cc
jellefoks: Addressed  in the 3rd commit of this PR.

cobalt/app/app_lifecycle_delegate.cc
jellefoks: I'll address it with an 'if (is_starboard)' build gate in a PR update.

ui/ozone/platform/starboard/platform_window_starboard.cc
jellefoks: Only if the platform sends duplicate conceal events, which I had a test for here but I'll prevent from propagating in #9424 (nearly ready, just needs rebasing on this PR). 

Added `CHECK()` instead, and for `widget_available_` here and in `Show()`.

ui/ozone/platform/starboard/platform_window_starboard.h
jellefoks: Style rule "include what you use". It's an addition for a missing  for the raw_ptr<> on line 103 below.

https://google.github.io/styleguide/cppguide.html#Include_What_You_Use

ui/views/widget/desktop_aura/desktop_window_tree_host_platform.cc
jellefoks: Done

ui/views/widget/desktop_aura/desktop_window_tree_host_platform.h
jellefoks: Done

starboard/shared/x11/application_x11.cc
jellefoks: I had encountered a crash here dereferencing the window. But auditing this, that should only be possible if DestroyWindow can be called from a diffent thread, which if that happens should be addressed differently. I changed this back but added SB_CHECK() to ensure it asserts instead of undefined behavior while dereferencing.

cobalt/BUILD.gn
```diff
@@ -109,14 +109,14 @@ source_set("common") {
...
-    "app/app_lifecycle_delegate.cc",
-    "app/app_lifecycle_delegate.h",
     "app/cobalt_main_delegate.cc",
     "app/cobalt_main_delegate.h",
   ]
 
   if (is_starboard) {
     sources += [
+      "app/app_lifecycle_delegate.cc",
+      "app/app_lifecycle_delegate.h",
```
yell0wd0g, line 119: \o/

ui/ozone/platform/starboard/platform_window_starboard.cc
jellefoks: Actually, in `Show()` the check is needed due to a convoluted initialization process around the various abstractions and the splash screen that I have yet to unravel.

cobalt/app/app_lifecycle_delegate.cc
```diff
@@ -146,7 +144,20 @@ void AppLifecycleDelegate::HandleEvent(const SbEvent* event) {
       OnStop(event);
       break;
     case kSbEventTypeBlur:
+      content::Shell::OnBlur();
+#if !BUILDFLAG(IS_STARBOARD)
+#error This file is only intended for Starboard platforms.
```
andrewsavage1, line 149: It's a bit of a strange place to put this error message, I would expect it at the top of the file—I don't think it's important enough to change now, but let's do it when we're cherry picking to main

andrewsavage1 (approved):

andrewsavage1:
Look great!

PR #9424, opened 2026-03-09
Cobalt Lifecycle: Formalize linear state machine

jellefoks:
Cobalt Lifecycle: Formalize linear state machine

With this change, platforms that may send events at unexpected times compared to other events, Cobalt will now behave in a predictable and tested way.

This replaces `app_lifecycle_delegate` with `app_event_delegate` and `app_event_runner`, which has better naming but more importantly it has robust handling of event sequencing. The sequence handling is kept in `app_event_delegate`, while `app_event_runner` contains the calls for the actions to be taken on incoming system events.

In addition to allowing much better handling of input event sequence edge cases, this split also allows thorough testing. 

In detail, this implements a formal, linear state machine for managing the application
lifecycle in ShellPlatformDelegate, ensuring predictable and traceable
state transitions.

Key Changes:
- Introduce the ApplicationState enum (Started, Blurred, Concealed,
  Frozen, Stopped) to track the high-level application lifecycle.
- Implement TransitionToState(state) logic that enforces a linear
  traversal of states, ensuring all intermediate side effects are
  triggered exactly once.
- Refactor lifecycle logic into granular "Do*" side-effect methods
  (DoBlur, DoConceal, DoFreeze, etc.) for better maintainability and
  testability.
- Enhance integration tests (test_lifecycle.sh and test_preload.sh) to
  verify internal state machine log sequences (e.g., ensuring Blur
  always precedes Conceal).
- Update MockShellPlatformDelegate to support base class delegation
  in unit tests, enabling verification of state machine transitions
  while mocking individual side effects.
- Ensure proper state initialization in platform-specific delegates
  (Android, Aura, Views, iOS).
- Update IntegratedBlurFocus unit test to verify WebContents focus
  state transitions during Blur/Focus events.

This formalization provides a robust framework for handling complex
application state transitions consistently across all Cobalt platforms.

Bug: 448156280

### DIFF
--- a/cobalt/app/app_event_delegate.cc
+++ b/cobalt/app/app_event_delegate.cc
@@ -47,7 +47,9 @@ AppEventDelegate::ApplicationState SbEventToTargetApplicationState(
     case kSbEventTypeStop:
       return AppEventDelegate::ApplicationState::kStopped;
     default:
-      NOTREACHED();
+      // This part of the code should only be reached if an unknown SbEventType
+      // is received.
+      NOTREACHED() << "Unexpected SbEventType: " << type;
       return AppEventDelegate::ApplicationState::kInitial;
   }
 }
@@ -98,6 +100,9 @@ void AppEventDelegate::HandleEvent(const SbEvent* event) {
         application_state_ = SbEventToTargetApplicationState(event->type);
         return;
       default:
+        // Robustly handle events received before the application has started or
+        // preloaded. This ensures that any early events are preceded by a
+        // valid initial state transition.
         LOG(WARNING) << "Received event " << event->type
                      << " before start or preload. "
                      << "An implicit preload event was inserted.";
@@ -115,6 +120,9 @@ void AppEventDelegate::HandleEvent(const SbEvent* event) {
     case kSbEventTypePreload:
     case kSbEventTypeStart: {
       // Redundant start/preload events are ignored, unless they contain a link.
+      // If they contain a link, they are treated as deep links. This avoids
+      // undefined behavior by providing defined transitions that maintain the
+      // lifecycle contract.
       const SbEventStartData* data =
           static_cast<const SbEventStartData*>(event->data);
       if (data && data->link && data->link[0] != '\0') {
@@ -170,6 +178,13 @@ void AppEventDelegate::HandleEvent(const SbEvent* event) {
 }
 
 void AppEventDelegate::TransitionToLifeCycleState(ApplicationState state) {
+  // TransitionToLifeCycleState ensures that the application moves from its
+  // current state to the target |state| by traversing all intermediate states
+  // in strict linear order. Each state transition triggers its corresponding
+  // side effects via the runner. This logic guarantees that no lifecycle events
+  // are skipped and that side effects are executed in a consistent, predictable
+  // sequence, regardless of whether the system events were received out of
+  // order, duplicated, or missing.
   CHECK_GT(state, ApplicationState::kInitial);
   CHECK_LE(state, ApplicationState::kStopped);
   // Ensure all intermediate state changes are triggered.
--- a/cobalt/app/app_event_delegate.h
+++ b/cobalt/app/app_event_delegate.h
@@ -34,13 +34,18 @@ namespace cobalt {
 // lifecycle actions and side effects is delegated to the AppEventRunner.
 class AppEventDelegate {
  public:
+  // ApplicationState defines the lifecycle states of the application.
+  // The order of these states is critical: TransitionToLifeCycleState relies on
+  // the invariant that states later in the enum are "further along" in the
+  // lifecycle towards being stopped. Any updates to this enum must preserve
+  // this monotonic ordering to ensure correct state machine transitions.
   enum class ApplicationState {
-    kInitial,
-    kStarted,
-    kBlurred,
-    kConcealed,
-    kFrozen,
-    kStopped
+    kInitial = 0,
+    kStarted = 1,
+    kBlurred = 2,
+    kConcealed = 3,
+    kFrozen = 4,
+    kStopped = 5
   };
 
   explicit AppEventDelegate(std::unique_ptr<AppEventRunner> runner = nullptr);


jellefoks:
Rename 'lifecyclle' to 'event', because all SbEventHandle calls go though this, including input events, deep links, etc.

All the logic from app_lifecycle_delegate now lives in app_event_runner, and the delegate, now named app_event_delegate only handles ensuring correct ordering of the events. It ensures that for lifecycle transitions, the intermediate states are never skipped. It will now correctly handle any out of sequence or duplicated event. Many tests are added for this.

third_party/musl/src/stdio/__toread.c
jellefoks, line 1: This change was added because it was causing a '__stdio_exit_needed'. A similar change already exists in main. I'm not sure why this PR triggered it.

cobalt/app/app_lifecycle_delegate.cc
jellefoks, line 1: The logic of this file is  moved to app_event_runner, and the delegate now only ensures correct sequencing of the EventHandle calls.

cobalt/app/app_event_delegate.h
```diff
@@ -0,0 +1,77 @@
...
+namespace cobalt {
+
+// AppEventDelegate translates Starboard system events into a sequence of
+// application lifecycle transitions. It maintains the canonical application
+// state and ensures that events are handled in a strict linear order,
+// synthesizing intermediate transitions when necessary. The actual execution of
+// lifecycle actions and side effects is delegated to the AppEventRunner.
+class AppEventDelegate {
+ public:
+  enum class ApplicationState {
```
jellefoks, line 42: This enum should allow to re-introduce reporting of the application state in the crash reports.

cobalt/shell/browser/lifecycle_unittest.cc
```diff
@@ -22,11 +22,11 @@ using testing::_;
 
 namespace content {
 
-class LifecycleTest : public CobaltShellTestBase {
+class LifecycleTest : public ShellTestBase {
```
jellefoks, line 25: This was unnecessarily renamed in an earlier PR, this renames it back.

cobalt/shell/browser/lifecycle_unittest.cc
```diff
@@ -40,104 +40,112 @@ class LifecycleTest : public CobaltShellTestBase {
...
-      platform_->CreatePlatformWindow(shell, gfx::Size(1920, 1080));
+      platform_->CreatePlatformWindow(shell_, gfx::Size(1920, 1080));
     }
 
-    Shell::FinishShellInitialization(shell);
+    Shell::FinishShellInitialization(shell_);
+  }
 
-    return shell;
+  void TearDown() override {
```
jellefoks, line 54: This logic was repeated every test, putting it here once makes each test much more readable.

cobalt/shell/browser/lifecycle_unittest.cc
```diff
@@ -22,11 +22,11 @@ using testing::_;
 
 namespace content {
 
-class LifecycleTest : public CobaltShellTestBase {
+class LifecycleTest : public ShellTestBase {
  public:
   LifecycleTest() = default;
 
-  Shell* CreateTestShell(bool is_visible) {
```
jellefoks, line 29: The return value is now stored as a member variable, and used in the TearDown().

cobalt/app/app_event_delegate.h
```diff
@@ -0,0 +1,77 @@
...
+  void TransitionToLifeCycleState(ApplicationState state);
+
+  std::unique_ptr<AppEventRunner> runner_;
+  ApplicationState application_state_ = ApplicationState::kInitial;
+
+#if BUILDFLAG(IS_STARBOARD)
+  // Ozone-specific bridge that converts Starboard events to Chromium events.
+  // Non-Starboard platforms (like Android) handle these events natively.
+  std::unique_ptr<ui::PlatformEventSourceStarboard> platform_event_source_;
+#endif
```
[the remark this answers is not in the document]
andrewsavage1: Used in cobalt/app/app_event_runner.cc

cobalt/app/app_event_delegate.cc
```diff
@@ -0,0 +1,244 @@
...
+    case kSbEventTypeReveal:
+      return AppEventDelegate::ApplicationState::kBlurred;
+    case kSbEventTypeFreeze:
+      return AppEventDelegate::ApplicationState::kFrozen;
+    case kSbEventTypeUnfreeze:
+      return AppEventDelegate::ApplicationState::kConcealed;
+    case kSbEventTypeStop:
+      return AppEventDelegate::ApplicationState::kStopped;
+    default:
+      NOTREACHED();
```
andrewsavage1: Consider adding an error message for this crash

cobalt/app/app_event_delegate.cc
```diff
@@ -0,0 +1,244 @@
...
+    switch (event->type) {
+      case kSbEventTypeStart:
+      case kSbEventTypePreload:
+        runner_->OnStart(event);
+        application_state_ = SbEventToTargetApplicationState(event->type);
+        return;
+      default:
+        LOG(WARNING) << "Received event " << event->type
+                     << " before start or preload. "
+                     << "An implicit preload event was inserted.";
```
andrewsavage1, line 108: Is this the same as the behavior of the app in c25? If not, it seems like it would be safer to maintain that behavior, wdyt?

cobalt/app/app_event_delegate.cc
```diff
@@ -0,0 +1,244 @@
...
+  switch (event->type) {
+    case kSbEventTypePreload:
+    case kSbEventTypeStart: {
+      // Redundant start/preload events are ignored, unless they contain a link.
+      const SbEventStartData* data =
+          static_cast<const SbEventStartData*>(event->data);
+      if (data && data->link && data->link[0] != '\0') {
+        LOG(WARNING) << "Received redundant "
+                     << (event->type == kSbEventTypeStart ? "start" : "preload")
+                     << " event with link after initialization. Treating as a "
```
andrewsavage1, line 131: This matches c25 behavior?

cobalt/app/app_event_delegate.cc
```diff
@@ -0,0 +1,244 @@
...
+
+void AppEventDelegate::TransitionToLifeCycleState(ApplicationState state) {
+  CHECK_GT(state, ApplicationState::kInitial);
+  CHECK_LE(state, ApplicationState::kStopped);
+  // Ensure all intermediate state changes are triggered.
+  while (application_state_ != state) {
+    CHECK_GT(application_state_, ApplicationState::kInitial);
+    CHECK_LE(application_state_, ApplicationState::kStopped);
+    ApplicationState old_state = application_state_;
+    if (old_state < state) {
```
andrewsavage1, line 195: It's strange to read a "less than" comparing states, but it does make the code much cleaner. If we're going to do this, we should at least add a comment in the definition of the ApplicationState enum stating that the order is important and future updates should maintain the invariant that being further down in the list=being closer to stopped. And if there's anything we can do to codify that (I can't think of anything offhand), even better

andrewsavage1:
Mostly lgtm, just want to make sure we aren't changing behavior from c25

cobalt/app/app_event_delegate.cc
jellefoks: No, it is not the same. On C25 the event would be dispatched to the rest of Cobalt with basically undefined behavior. This is much more robust IMHO. Similar the handling of preload and start in the main switch statement, around lines 115-135, where redundant start events are converted to deep links and start events as reasonable for those cases. Those two pieces of logic here is the core of the robustness improvement besides the  sequencing guarantee from TransitionToLifeCycleState.

cobalt/app/app_event_delegate.cc
jellefoks: No, this is a certain robustness improvement because it replaces undefined behavior in C25 (and here without this change) with reasonable defined behavior that adheres to the lifecycle contract that the rest of Cobalt code expects.

cobalt/app/app_event_delegate.cc
jellefoks: Done.

cobalt/app/app_event_delegate.cc
jellefoks: Done, and nummbering enforced with initializers. Makes it more clear if the number is reported in crash reports as well.

jellefoks:
> Mostly lgtm, just want to make sure we aren't changing behavior from c25

This actually is better than C25 because it removes relying on the platform to send events in the correct ordering for Cobalt to function in a deterministic way. This enforces the determinism and avoids undefined behavior. The added tests show and confirm the behavior.

cobalt/app/app_event_delegate.cc
andrewsavage1: I just worry that there could be some unintended side effects i.e. of platforms/clients depending on the broken behavior in Cobalt 25. It's probably okay as we'll be throughly testing lifecycle for EAP anyways.

Separately, I'm conflicted about the pattern of taking over control from the platform and implicitly injecting events that we'd generally expect from the platform. It makes sense to be defensive and not crash if the platform sends events in an unexpected order, but to me it seems better to not do anything other than log a warning when we get an unexpected event (i.e. I don't feel it should be respected).

I'm fine with approving this as we want to test out this functionality quite soon, but maybe worth thinking about for 27.lts/the future

PR #9426, opened 2026-03-09
media: Log encoded video/audio budgets as histogram

yell0wd0g:
This CL/PR adds UMA-like logging of each video playback 
demuxer's video/audio budgets, which is tantamount to say 
the amount of memory that will be allocated for either of 
those stream types' data pools. 

The numbers are retrieved from //starboard locations like 
e.g. starboard/android/shared/media_get_video_buffer_budget.cc,
and are quantities like e.g. 30MB for video 1080p on Linux.

This was tested locally, see the result e.g. here:
https://i.imgur.com/U4ExHUg.png (ignore the typo).

Bug: 487768963

### DIFF
--- a/media/starboard/decoder_buffer_memory_info.cc
+++ b/media/starboard/decoder_buffer_memory_info.cc
@@ -26,7 +26,7 @@ constexpr int kBytesPerMegabyte = 1024 * 1024;
 
 int GetAudioDecoderBufferLimitBytes() {
   // TODO(mcasas): This value should always be the same on every call, enforce.
-  const auto audio_budget_in_bytes = SbMediaGetAudioBufferBudget()
+  const auto audio_budget_in_bytes = SbMediaGetAudioBufferBudget();
   UMA_HISTOGRAM_MEMORY_MEDIUM_MB("Media.Starboard.AudioBufferBudget",
                                  audio_budget_in_bytes / kBytesPerMegabyte);
   return audio_budget_in_bytes;
@@ -38,7 +38,6 @@ int GetVideoDecoderBufferLimitBytes(VideoCodec codec,
   const auto video_budget_in_bytes = SbMediaGetVideoBufferBudget(
       MediaVideoCodecToSbMediaVideoCodec(codec), resolution.width(),
       resolution.height(), bits_per_pixel);
-
   UMA_HISTOGRAM_MEMORY_MEDIUM_MB("Media.Starboard.VideoBufferBudget",
                                  video_budget_in_bytes / kBytesPerMegabyte);
   return video_budget_in_bytes;
--- a/media/starboard/decoder_buffer_memory_info.cc
+++ b/media/starboard/decoder_buffer_memory_info.cc
@@ -14,6 +14,7 @@
 
 #include "media/starboard/decoder_buffer_memory_info.h"
 
+#include "base/metrics/histogram_macros.h"
 #include "media/base/video_codecs.h"
 #include "media/starboard/starboard_utils.h"
 #include "starboard/media.h"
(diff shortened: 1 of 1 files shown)


media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -57,24 +60,36 @@ int GetBitsPerPixel(const VideoDecoderConfig& video_config) {
 
 size_t GetDemuxerStreamAudioMemoryLimit(
     const AudioDecoderConfig* /*audio_config*/) {
-  return DecoderBuffer::Allocator::GetInstance()->GetAudioBufferBudget();
+  const auto audio_budget_in_bytes =
```
osagie98: Can this be made static? Since the audio budget doesn't consider the audio configuration, it should return the same value every time. We may be able to do. the same with the default video budget on line 74 as well.

yell0wd0g: Gotcha. I had to rebase this PR quite a bit since the code moved
to `media/starboard/decoder_buffer_memory_info`, let's continue
there?

yell0wd0g (commented):
@osagie98 @jasonzhangxx PTAL

media/base/starboard/demuxer_memory_limit_starboard.cc
yell0wd0g: I could do e.g.
```
  static const auto audio_budget_in_bytes = SbMediaGetAudioBufferBudget();
  DCHECK_EQ(audio_budget_in_bytes, SbMediaGetAudioBufferBudget());
```

which would enforce that `SbMediaGetAudioBufferBudget()` always 
returns the same value, but only on debug/devel builds (those with 
`DCHECK_IS_ON()` defined). WDYT?

However I'm not super sure this would always work because IIUC 
`SbMediaGetAudioBufferBudget()` is implemented inside Starboard
which means that partners could do unpredictable things.

Let me know what you think @osagie98 .

xiaomings: Using a non-static variable is probably better here:
1. `SbMediaGetAudioBufferBudget()` can technically change during the app lifetime.
2. The cost of calling `SbMediaGetAudioBufferBudget()` is probably negligible, and not worth the extra optimization.

media/starboard/decoder_buffer_memory_info.cc
```diff
@@ -14,23 +14,33 @@
...
 #include "starboard/media.h"
 #include "ui/gfx/geometry/size.h"
 
 namespace media {
 
+constexpr int kBytesPerMegabyte = 1024 * 1024;
+
 int GetAudioDecoderBufferLimitBytes() {
-  return SbMediaGetAudioBufferBudget();
+  // TODO(mcasas): This value should always be the same on every call, enforce.
```
xiaomings, line 28: Consider removing the comment as technically the return value doesn't have to always be the same.

media/starboard/decoder_buffer_memory_info.cc
```diff
@@ -14,23 +14,33 @@
...
 int GetVideoDecoderBufferLimitBytes(VideoCodec codec,
                                     const gfx::Size& resolution,
                                     int bits_per_pixel) {
-  return SbMediaGetVideoBufferBudget(MediaVideoCodecToSbMediaVideoCodec(codec),
-                                     resolution.width(), resolution.height(),
-                                     bits_per_pixel);
+  const auto video_budget_in_bytes = SbMediaGetVideoBufferBudget(
+      MediaVideoCodecToSbMediaVideoCodec(codec), resolution.width(),
+      resolution.height(), bits_per_pixel);
+  UMA_HISTOGRAM_MEMORY_MEDIUM_MB("Media.Starboard.VideoBufferBudget",
```
xiaomings, line 41: Not sure if we should move both to http://shortn/_60s46qFvhY.

Also, the name of the UMA metrics look good enough if it can be revised later, otherwise we may want to design it better with combination of codec/resolution.

xiaomings: Also, depending on what we want to track, the accurate value for memory allocated is still held by DecoderBufferAllocator.  Under the current implementation, if a 1080p video is played after a 4k video, the memory budget might be 30 MB, but the actual memory pool might be 180 MB (audio and video combined).

yell0wd0g: Gotcha, so should I move these calls to media/starboard/decoder_buffer_allocator.cc 's `Allocate()` ? And track the `size` param value? Happy to pivot to that.

https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:media/starboard/decoder_buffer_allocator.cc;l=98?q=DecoderBufferAllocator&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain&start=1

DecoderBufferAllocator is used for both Audio and Video, correct? (Just
making sure).

PR #9433, opened 2026-03-10
tvos: Extend IsAppleTV4K to cover 2nd and 3rd gen tvOS devices

abhijeetk:
IsAppleTV4KPriv() only matched AppleTV6-2 (1st gen),causing AppleTV11-1
(2nd gen) and AppleTV14-1 (3rd gen) to be misidentified as non-4K
devices.

Bug: 487326366

### DIFF
--- a/starboard/tvos/shared/media/playback_capabilities.mm
+++ b/starboard/tvos/shared/media/playback_capabilities.mm
@@ -18,11 +18,14 @@
 #import <UIKit/UIKit.h>
 #import <VideoToolbox/VideoToolbox.h>
 
+#include <algorithm>
 #include <mutex>
+#include <string_view>
 #include <vector>
 
 #include "starboard/common/log.h"
 #include "starboard/common/once.h"
+#include "starboard/common/system_property.h"
 #include "starboard/system.h"
 #include "starboard/tvos/shared/observer_registry.h"
 #include "starboard/tvos/shared/uikit_media_session_client.h"
@@ -42,11 +45,12 @@ bool IsAppleTVHDPriv() {
 }
 
 bool IsAppleTV4KPriv() {
-  char model_name[128];
-  bool succeeded = SbSystemGetProperty(kSbSystemPropertyModelName, model_name,
-                                       sizeof(model_name));
-  SB_DCHECK(succeeded);
-  return succeeded && strcmp(model_name, "AppleTV6-2") == 0;
+  static constexpr std::string_view kAppleTV4KModels[] = {
+      "AppleTV6-2", "AppleTV11-1", "AppleTV14-1"};
+  const std::string model =
+      starboard::GetSystemPropertyString(kSbSystemPropertyModelName);
+  return std::ranges::find(kAppleTV4KModels, model) !=
+         std::ranges::end(kAppleTV4KModels);
 }
 
 SbMediaAudioConnector GetConnectorFromAVAudioSessionPort(


abhijeetk:
cc: @Gyuyoung @jkim-julie

starboard/tvos/shared/media/playback_capabilities.mm
```diff
@@ -46,7 +46,9 @@ bool IsAppleTV4KPriv() {
   bool succeeded = SbSystemGetProperty(kSbSystemPropertyModelName, model_name,
                                        sizeof(model_name));
   SB_DCHECK(succeeded);
-  return succeeded && strcmp(model_name, "AppleTV6-2") == 0;
+  return succeeded && (strcmp(model_name, "AppleTV6-2") == 0 ||
+                       strcmp(model_name, "AppleTV11-1") == 0 ||
+                       strcmp(model_name, "AppleTV14-1") == 0);
```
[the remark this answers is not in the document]
rakuco: I'd go further and make this more C++-y:

```c++
#include <algorithm>
#include <array>
#include <string_view>

bool IsAppleTV4KPriv() {
  static constexpr auto kAppleTV4KModels = std::to_array<std::string_view>({
    "AppleTV6-2",
    // ...
  });

  // ...

  return succeeded && std::ranges::find(kAppleTV4KModels, model_name) != kAppleTV4KModels.end();
}
```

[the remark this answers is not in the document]
abhijeetk: Done.

abhijeetk:
> informal lgtm modulo the readability improvements
> 
> I'm surprised this was actually happening, does it basically mean the more recent Apple TV devices never got 4K video? Or was it limited to specific codecs?

I think, new gen2/3 devices always use HW path.

starboard/tvos/shared/media/playback_capabilities.mm
```diff
@@ -42,11 +45,15 @@ bool IsAppleTVHDPriv() {
 }
 
 bool IsAppleTV4KPriv() {
+  static constexpr auto kAppleTV4KModels = std::to_array<std::string_view>(
+      {"AppleTV6-2", "AppleTV11-1", "AppleTV14-1"});
+
   char model_name[128];
   bool succeeded = SbSystemGetProperty(kSbSystemPropertyModelName, model_name,
                                        sizeof(model_name));
   SB_DCHECK(succeeded);
```
yell0wd0g: This pattern (DCHECK()ing a value and then using it) is odd :)
If it's always true, which is what the DCHECK implies, then
we shouldn't add it to the return. I'd suggest removing the
`SB_DCHECK()`.

starboard/tvos/shared/media/playback_capabilities.mm
```diff
@@ -42,11 +45,15 @@ bool IsAppleTVHDPriv() {
 }
 
 bool IsAppleTV4KPriv() {
+  static constexpr auto kAppleTV4KModels = std::to_array<std::string_view>(
+      {"AppleTV6-2", "AppleTV11-1", "AppleTV14-1"});
+
   char model_name[128];
```
yell0wd0g: Can we use std::string or std::array here?

rakuco: Since this interacts with `SbSystemGetProperty()` I don't think it would help much, but an array or a `base::span` could work.

starboard/tvos/shared/media/playback_capabilities.mm
```diff
@@ -42,11 +45,15 @@ bool IsAppleTVHDPriv() {
 }
 
 bool IsAppleTV4KPriv() {
+  static constexpr auto kAppleTV4KModels = std::to_array<std::string_view>(
+      {"AppleTV6-2", "AppleTV11-1", "AppleTV14-1"});
```
fayev-g: nit for future clean up: I do notice in system_get_property.mm we replace the `,` with `-`. Is is possible to just use Apple's syntax or is it necessary elsewhere?

abhijeetk: Ack! I will handle it in separate PR.

starboard/tvos/shared/media/playback_capabilities.mm
abhijeetk: Removed unwanted SB_DCHECK.

starboard/tvos/shared/media/playback_capabilities.mm
abhijeetk: - Used plain std::string_view array
- For std::string_view(model_name) in find() : SbSystemGetProperty fills a raw char[128] buffer, so wrapping it explicitly makes the intent clear.

rakuco: Note that Miguel was asking about `model_name`, not `kAppleTV4KModels`. Using a plain array of `std::string_view`s works, but is the same in this case.

starboard/tvos/shared/media/playback_capabilities.mm
```diff
@@ -42,11 +44,16 @@ bool IsAppleTVHDPriv() {
...
+      "AppleTV6-2", "AppleTV11-1", "AppleTV14-1"};
+
   char model_name[128];
   bool succeeded = SbSystemGetProperty(kSbSystemPropertyModelName, model_name,
                                        sizeof(model_name));
-  SB_DCHECK(succeeded);
-  return succeeded && strcmp(model_name, "AppleTV6-2") == 0;
+
+  return succeeded &&
+         std::ranges::find(kAppleTV4KModels, std::string_view(model_name)) !=
```
rakuco: You don't need to explicitly create an `std::string_view` here, `find()` will compare the `std::string_view`s in the array to `const char*`, which also works.

abhijeetk: Done. Please take look at latest patch.

starboard/tvos/shared/media/playback_capabilities.mm
abhijeetk: Ahh, I missed it. PTAL at latest patch.

abhijeetk: > Can we use std::string or std::array here?

We can use std::string with `GetSystemPropertyString` (instead of using raw `SbSystemGetProperty`)
```
  const std::string model =
      starboard::GetSystemPropertyString(kSbSystemPropertyModelName);
```

rakuco: Oh, that's even better!

abhijeetk: Done.

starboard/tvos/shared/media/playback_capabilities.mm
abhijeetk: Created bug b/492055753 to track this request.

fayev-g: Thanks so much!

PR #9453, opened 2026-03-10
Support cobalt_browsertests on Evergreen

linxinan-yt:
This PR refactors SbEventHandle() of cobalt_browsertests, to be compatible with Evergreen build.

1. Move initialization under SbEventHandle(), as in Evergreen mode the test entrance is SbEventHandle() not main().
2. Replace TerminateCurrentProcessImmediately with _Exit() to avoid crashing in Starboard turn down after the test has done.  

Follow up PRs will try to build the test runner script into the loader.py.

TEST: out/evergreen-x64_devel/cobalt_browsertests_loader.py \
  --single-process-tests \
  --no-sandbox --single-process \
  --no-zygote --ozone-platform=starboard \
  --gtest_filter=ContentMainRunnerImplBrowserTest.StartupSequence
Issue: 463991461


cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -12,8 +12,13 @@
 // implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
+#include <stdlib.h>
```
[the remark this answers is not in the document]
linxinan-yt: Done.

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -49,38 +56,68 @@ class StarboardTestLauncherDelegate : public content::TestLauncherDelegate {
...
+
+      std::vector<char*> new_argv;
+      for (int i = 0; i < argc; ++i) {
+        new_argv.push_back(argv[i]);
+      }
+      char single_process_tests_arg[] = "--single-process-tests";
+      char single_process_arg[] = "--single-process";
+      char no_sandbox_arg[] = "--no-sandbox";
+      char no_zygote_arg[] = "--no-zygote";
+      char ozone_platform_arg[] = "--ozone-platform=starboard";
```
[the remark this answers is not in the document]
linxinan-yt: Those flags were removed from the CPP file.

linxinan-yt:
> LGTM Should we have someone from 3P review this?

Yes. That is the plan.

linxinan-yt:
@y4vor @niranjanyardi 

This PR is blocking my following PRs to run BR tests on RDK. May I get a review from you? Many thanks.

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -49,38 +50,53 @@ class StarboardTestLauncherDelegate : public content::TestLauncherDelegate {
...
-  }
-}
+      // A manager for singleton destruction.
+      base::AtExitManager at_exit;
 
-int main(int argc, char** argv) {
-  base::CommandLine::Init(argc, argv);
-  testing::InitGoogleTest(&argc, argv);
+      // TODO(b/433354983): Support more platforms.
+      ui::LinuxUi::SetInstance(ui::GetDefaultLinuxUi());
```
niranjanyardi, line 73: can we move this to somewhere under starboard rather than here ? this seems platform specific code and may not run on all platforms(think  3P AOSP)

idk if @y4vor or @jellefoks  have other thoughts

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -49,38 +50,53 @@ class StarboardTestLauncherDelegate : public content::TestLauncherDelegate {
...
 
-  // A manager for singleton destruction.
-  base::AtExitManager at_exit;
+      StarboardTestLauncherDelegate delegate;
```
niranjanyardi: Instead of StarboardTestLauncherDelegate consider renaming iot to CobaltBrowserTestLauncherDelegate as it derives from [content](content::TestLauncherDelegate)

feel free to address in an upcoming PR

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -49,38 +50,53 @@ class StarboardTestLauncherDelegate : public content::TestLauncherDelegate {
...
+      // Call std::_Exit() from <cstdlib> to immediately terminate the process
+      // without executing any C++ destructors or AtExitManager callbacks.
+      // Chromium browser tests intentionally leak state in single-process mode,
+      // which causes memory access violations during standard teardown by
+      // Starboard.
+      //
+      // Note: We cannot use standard _exit() (lowercase) or Chromium's
+      // base::Process::TerminateCurrentProcessImmediately (which calls _exit)
+      // because Starboard's Musl port specifically maps _exit() back to exit(),
+      // which runs the problematic teardown logic anyway. std::_Exit()
```
niranjanyardi, line 89: why is this a problem ? i tried reading through the bug

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
linxinan-yt: For android browser tests, it has a different entrance and does not use this file.
I am not sure 3p though. Do you have any recommendations under starboard?

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
linxinan-yt: The problem is that browser tests intentionally leak sth after test is done. So we could not use the production way to exit starboard, as one may see test passed but the test binary returned exit code -11. This is the reason we need some powerful exit function to just stop the test binary once test launcher completes. LGTY?

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -49,38 +50,53 @@ class StarboardTestLauncherDelegate : public content::TestLauncherDelegate {
...
+      // (uppercase) bypasses this and invokes the raw SYS_exit_group syscall.
+      std::_Exit(test_result_code);
```
y4vor, line 93: This is bypassing Starboard and might be unpredictable on a 3P device.
I think just ignore the asan results on specific tests. It is also possible to configure the run to exclude known leaks.

I think you should just let it exit normally. And deal with ASAN separtly.

linxinan-yt: Thanks for the pointer. Is there a way I can specifically turn off ASAN for browser tests?

andrewsavage1: You can try to use `export ASAN_OPTIONS=exitcode=0` to ignore the asan errors while keeping it on. Note that this ignores all errors instead of creating a list of known errors to exclude, which may be a better option

linxinan-yt: I see. So all evergreen builds have turned on ASAN, correct?

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
linxinan-yt: I will handle this with y4vor's recommendations in the follow-up PR.

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
linxinan-yt: Make sense. Thanks.

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
linxinan-yt: I tested it on RDK and it worked. I think the `ASAN_OPTIONS=exitcode=0` should go to the runner script, which I have a follow up PR #9524. I have added a TODO here to remind myself.

PR #9405, opened 2026-03-08
cobalt/test: Filter browser test data to reduce archive size

Awallky:
This change introduces a pattern-based filtering mechanism for browser
test data used by cobalt_browsertests. It replaces the blanket
inclusion of upstream content/test/data and media/test/data with
a curated subset defined by specific inclusion and exclusion patterns.

The filtering is implemented via a new GN template, cobalt_test_data.gni,
which leverages an expand_filelist.py script. This significantly
reduces the content included in test archives and Docker images by
omitting large, unnecessary directories such as meet_effects.

Furthermore, the Dockerfile is updated to utilize a multi-stage build
process, allowing for the pruning of unstripped libraries and other
heavy build artifacts from the final image. The artifact collection
script, collect_test_artifacts.py, is also enhanced to prefer stripped
libraries and include only essential tools.

These optimizations lead to a substantial reduction in the size of
test bundles, improving build and deployment efficiency. 

Test: `vpython3 cobalt/build/archive_test_artifacts.py       --source-dir .       --out-dir out/android-arm_devel/       --destination-dir cobalt/testing/browser_tests/tools/       --targets cobalt/testing/browser_tests:cobalt_browsertests       --compression gz       --archive-per-target       --use-android-deps-path       --flatten-deps && docker build -f cobalt/testing/browser_tests/tools/Dockerfile -t cobalt_browsertests_local:latest . && time docker run --rm -e ANDROID_SERIAL=$ANDROID_SERIAL --network host -v $(pwd)/test_results:/results cobalt_browsertests_local:latest android-arm_devel --gtest_output="xml:/results/results.xml" --gtest-filter=*Cobalt* --init-command=""`

Test: `vpython3 cobalt/testing/browser_tests/tools/expand_filelist_test.py`
Test: `vpython3 cobalt/testing/browser_tests/tools/test_run_tests_template.py`
Bug: 473909665
Bug: 483488213

PR #9466, opened 2026-03-11
cobalt/test: Rename test bundle data for ios to have ios in name

Awallky:
[#9405](https://github.com/youtube/cobalt/pull/9405) is adding browser test data dependency generation tooling for all platforms, excluding ios. To reduce the delta there and avoid naming conflicts of the two files, this change simply renames the data dependency for ios, since it is so drastically different from the other file list.

Bug: 473909665
Bug: 483488213


fayev-g:
lgtm - also just a reminder for posterity that the is_ios gn flag also actually applies to tvOS

fayev-g (approved):

PR #9468, opened 2026-03-11
media: Add defensive check for decode-to-texture

johnedocampo:
Prevent MediaIsVideoSupported from reporting support for
decode-to-texture video on Android. This check ensures
that unsupported configurations are correctly rejected, avoiding
potential playback issues. Once decode-to-texture is implemented,
we can revert this code.

Bug: 490494472

### DIFF
--- a/starboard/android/shared/media_is_video_supported.cc
+++ b/starboard/android/shared/media_is_video_supported.cc
@@ -49,12 +49,16 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
            ->IsHDRTransferCharacteristicsSupported(transfer_id)) {
     return false;
   }
-  // While not necessarily true, for now we assume that all Android devices
-  // can play decode-to-texture video just as well as normal video.
 
   bool must_support_tunnel_mode = false;
   if (mime_type) {
     if (!mime_type->is_valid()) {
+      return false;
+    }
+
+    // TODO(b/491832955): Remove once decode-to-texture is supported.
+    if (mime_type->ValidateBoolParameter("decode-to-texture") &&
+        mime_type->GetParamBoolValue("decode-to-texture", false)) {
       return false;
     }
 


starboard/android/shared/media_is_video_supported.cc
```diff
@@ -49,15 +49,19 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
...
-  // While not necessarily true, for now we assume that all Android devices
-  // can play decode-to-texture video just as well as normal video.
 
   bool must_support_tunnel_mode = false;
   if (mime_type) {
     if (!mime_type->is_valid()) {
       return false;
     }
 
+    // TODO (b/491832955) : Remove once decode-to-texture is supported.
```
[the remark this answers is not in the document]
johnedocampo: Done.

starboard/android/shared/media_is_video_supported.cc
```diff
@@ -49,15 +49,19 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
...
+    if (mime_type->ValidateBoolParameter("decode-to-texture") &&
+        mime_type->GetParamBoolValue("decode-to-texture", true)) {
+      return false;
+    }
```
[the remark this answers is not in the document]
johnedocampo: Done.

johnedocampo:
> nit: I think the change is also needed on main.

Will do.

starboard/android/shared/media_is_video_supported.cc
```diff
@@ -49,15 +49,19 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
...
 
   bool must_support_tunnel_mode = false;
   if (mime_type) {
     if (!mime_type->is_valid()) {
       return false;
     }
 
+    // TODO(b/491832955): Remove once decode-to-texture is supported.
+    if (mime_type->ValidateBoolParameter("decode-to-texture") &&
+        mime_type->GetParamBoolValue("decode-to-texture", false)) {
```
borongc, line 61: Should `decode-to-texture` be rejected regardless of `true` or `false`? As long as player asks for it, we should reject it?

borongc: This is nit, as `decode-to-texture=false` is no-op.

borongc (approved):

starboard/android/shared/media_is_video_supported.cc
johnedocampo: I think it should be okay to only treat `decode-to-texture=true` here. If `decode-to-texture=false` is reported, we should be okay.

starboard/android/shared/media_is_video_supported.cc
```diff
@@ -49,15 +49,19 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
...
-  // can play decode-to-texture video just as well as normal video.
 
   bool must_support_tunnel_mode = false;
   if (mime_type) {
     if (!mime_type->is_valid()) {
       return false;
     }
 
+    // TODO(b/491832955): Remove once decode-to-texture is supported.
+    if (mime_type->ValidateBoolParameter("decode-to-texture") &&
```
johnxwork, line 60: Just check: we didn't miss a `!` negation sign here, right? This is saying: if the parameter is validated, and the bool value is available => return false?

johnedocampo: This check says: if the `decode-to-texture` parameter is boolean, and that boolean is true, return false. We should be good here.

PR #9470, opened 2026-03-11
Add workflows for 27.lts

oxve:
Rename workflows from main to 27.lts. Remove unnecessary workflows.

Bug: 484351320


oxve:
Bypassing on-device tests due to build breakage

PR #9472, opened 2026-03-11
Add UMA histogram for native splash screen

tasunnn:
Introduce the `Cobalt.SplashScreen.FetchedFromCache` UMA boolean histogram. This metric records whether the native splash screen resource was successfully retrieved from the application's cache. If on error, the reason for failure will be specified and splash content will fall back to builtin depending on the circumstances. This helps monitor and gather relevant metrics related to splash screen caching.

Bug: 490510288

### DIFF
--- a/cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
+++ b/cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
@@ -385,18 +385,18 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
     SendResponse(state);
   }
 
-  void SendResponse(
-      std::optional<SplashScreenFetchedState> state = std::nullopt,
-      int http_status = net::HTTP_OK) {
+  void SendResponse(SplashScreenFetchedState state,
+                    int http_status = net::HTTP_OK) {
+    UMA_HISTOGRAM_ENUMERATION("Cobalt.SplashScreen.FetchedFromCache", state);
+    SendResponse(http_status);
+  }
+
+  void SendResponse(int http_status = net::HTTP_OK) {
     if (!client_) {
       LOG(ERROR) << "URLLoaderClient is not connected.";
       return;
     }
 
-    if (state.has_value()) {
-      UMA_HISTOGRAM_ENUMERATION("Cobalt.SplashScreen.FetchedFromCache",
-                                state.value());
-    }
     auto response_head = network::mojom::URLResponseHead::New();
     response_head->mime_type = mime_type_;
     response_head->content_length = content_.size();


tools/metrics/histograms/metadata/cobalt/histograms.xml
```diff
@@ -92,6 +92,18 @@ Always run the pretty print utility on this file after editing:
   </summary>
 </histogram>
 
+<histogram name="Cobalt.SplashScreen.FetchedFromCache" enum="Boolean"
```
linxinan-yt: Boolean is fine now but hard to expand in the future. Later we may want to know more from the cache, e.g. video v.s. image, music v.s. default. 
I would recommend we use `enum class` like Cobalt.Finch.ConfigState, in tools/metrics/histograms/metadata/cobalt/enums.xml.

E.g. 

```
<enum name="VariationsConfigState">
  <int value="0" label="built-in"/>
  <int value="1" label="cache"/>
</enum> 
```

It could also record the case that we failed to extract the cache.

linxinan-yt:
Is it possible to add a check in the browser tests to make sure the histogram is correctly recorded?

linxinan-chops (approved):

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -335,11 +350,14 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
   void SendBlobContent(uint64_t expected_size, std::vector<uint8_t> content) {
     if (content.size() != expected_size) {
       return DisconnectCacheAndSendFallback(
-          "Failed to read splash cache. Fallback to builtin.");
+          "Failed to read splash cache. Fallback to builtin.",
```
borongc, line 360: It seems like this is reading cache error, not `On Error` in general.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
borongc, line 253: This is another error (should not happen, but if it does, something is wrong).

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -72,6 +73,14 @@ const char kH5vccContentSecurityPolicy[] =
     "connect-src 'self' blob: data: %s:;";
 }  // namespace
 
+// An enum for UMA histogram to indicate the state of retrieving splash screen
+enum class SplashScreenFetchedState {
+  kBuiltIn = 0,
+  kCache = 1,
+  kError = 2,
+  kMaxValue = kError,
+};
```
yell0wd0g, line 85: Quick drive-by: It's usually good for enums tracking errors
to have an entry that means "no error" to have a reference
of how many times the "thing" succeeded, when looking at 
histograms server-side. Otherwise the error counts would not
have a reference of magnitude to compare against.

linxinan-yt: What about 

```
kOkCache = 0
kOkBuiltIn = 1, // no cache matched, falllback to built in splash.
kErrorEmptyCache = 2, // cache found but empty content.
....
```

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Tag it with `kErrorCacheReading`?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -348,15 +372,19 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
   }
 
   void DisconnectCacheStorage() { cache_storage_remote_.reset(); }
 
-  void DisconnectCacheAndSendFallback(const std::string& message) {
+  void DisconnectCacheAndSendFallback(const std::string& message,
+                                      SplashScreenFetchedState state) {
     LOG(ERROR) << message;
     DisconnectCacheStorage();
+    UMA_HISTOGRAM_ENUMERATION("Cobalt.SplashScreen.FetchedFromCache", state);
```
borongc, line 390: As we have multiple `UMA_HISTOGRAM_ENUMERATION("Cobalt.SplashScreen.FetchedFromCache", state)`, and they are all before `SendResponse()`, it may make more sense that we just log UMA inside `SendResponse()` to reduce these repeating codes.

tools/metrics/histograms/metadata/cobalt/enums.xml
```diff
@@ -32,6 +32,17 @@ https://chromium.googlesource.com/chromium/src.git/+/HEAD/tools/metrics/histogra
...
 
+<enum name="SplashScreenFetchedState">
+  <int value="0" label="Success: Content from builtin"/>
+  <int value="1" label="Success: Content from cache"/>
+  <int value="2"
+      label="Error: content from cache is empty, fallback to builtin."/>
+  <int value="3"
+      label="Error: Content from cache is too large, fallback to builtin."/>
+  <int value="4" label="Error: Failed to read cache, fallback to builtin."/>
+  <int value="5" label="Error: Resource not found."/>
```
linxinan-yt: `Builtin resource not found`.  Let's clarify this is built in resource not cache.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -245,10 +257,13 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
       ReadSplashCache(key);
       return;
     }
+
     if (content_.empty()) {
       SendNotFoundResponse(key);
       return;
     }
+    UMA_HISTOGRAM_ENUMERATION("Cobalt.SplashScreen.FetchedFromCache",
```
linxinan-yt: This is for the HTML. I think we can remove this, as this UMA is for the splash video/image.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -308,21 +322,27 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
   void OnCacheMatched(const std::string& cache_name,
                       blink::mojom::MatchResultPtr result) {
     if (!result->is_response()) {
-      return DisconnectCacheAndSendFallback(base::StringPrintf(
-          "Failed to match splash video from cache %s, error: %d",
-          cache_name.c_str(), static_cast<int>(result->get_status())));
+      return DisconnectCacheAndSendFallback(
+          base::StringPrintf(
+              "Failed to match splash video from cache %s, error: %d",
```
linxinan-yt, line 329: Since you touch this line, maybe reword it a little bit, as "no cache" is not an error state.

linxinan-yt: Synced offline, I will file a follow up PR to refine it.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -349,23 +372,31 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
...
+  void DisconnectCacheAndSendFallback(const std::string& message,
+                                      SplashScreenFetchedState state) {
     LOG(ERROR) << message;
     DisconnectCacheStorage();
-    SendResponse();
+    SendResponse(state);
   }
 
-  void SendResponse(int http_status = net::HTTP_OK) {
+  void SendResponse(
```
linxinan-yt: I would recommend overloading SendResponse(). So you don't have to change the current SendResponse method, but

```
void SendResponse(std::optional<SplashScreenFetchedState> state, int http_status = net::HTTP_OK) {
if (state.has_value()) {
      UMA_HISTOGRAM_ENUMERATION("Cobalt.SplashScreen.FetchedFromCache",
                                state.value());
    }
SendResponse(http_status);
```

PR #9480, opened 2026-03-12
tvos: Refactor IsAppleTVHDPriv to use GetSystemPropertyString

abhijeetk:
Adapt IsAppleTVHDPriv() to use starboard::GetSystemPropertyString,
consistent with the approach introduced for IsAppleTV4KPriv() in
#9433 . Replaces the manual char buffer and SbSystemGetProperty
call with the shared helper, removing the need for SB_DCHECK and
strcmp.

Bug: 487326366


abhijeetk:
cc : @Gyuyoung @jkim-julie

PR #9481, opened 2026-03-12
tvos: Remove always-false kEnableHdrWithSoftwareVp9 flag

abhijeetk:
The flag kEnableHdrWithSoftwareVp9 was hardcoded to false, making the
condition is_hdr && !kEnableHdrWithSoftwareVp9 always equivalent to
is_hdr. Remove the redundant flag and simplify the condition to
eliminate dead code.

Bug: 487326366

### DIFF
--- a/starboard/tvos/shared/media_is_video_supported.mm
+++ b/starboard/tvos/shared/media_is_video_supported.mm
@@ -68,8 +68,6 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
     }
     if (video_codec == kSbMediaVideoCodecVp9) {
 #if defined(INTERNAL_BUILD)
-      const bool kEnableHdrWithSoftwareVp9 = false;
-
       if (is_hdr) {
         if (transfer_id == kSbMediaTransferIdSmpteSt2084 ||
             transfer_id == kSbMediaTransferIdAribStdB67) {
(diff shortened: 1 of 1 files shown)


starboard/tvos/shared/media_is_video_supported.mm
```diff
@@ -104,6 +102,7 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
         experimental_allowed = experimental_value == "allowed";
       }
 
+      constexpr bool kEnableHdrWithSoftwareVp9 = false;
       if (experimental_allowed) {
         if (is_hdr && !kEnableHdrWithSoftwareVp9) {
```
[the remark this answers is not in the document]
abhijeetk: Accepted.

abhijeetk:
cc : @Gyuyoung @jkim-julie

starboard/tvos/shared/media_is_video_supported.mm
```diff
@@ -105,6 +103,7 @@ bool MediaIsVideoSupported(SbMediaVideoCodec video_codec,
       }
 
       if (experimental_allowed) {
+        constexpr bool kEnableHdrWithSoftwareVp9 = false;
```
rakuco: Isn't it better to just remove this altogether? It's always false so you could just check for `is_hdr` below.

abhijeetk: Yes make sense. Implemented changes.

PR #9484, opened 2026-03-12
android: Add experiment setup for video decoder threading

xiaomings:
This change adds experiment setup code for https://github.com/youtube/cobalt/pull/9711, so it can be experimented in production.

Bug: 329686979


starboard/android/shared/player_components_factory.cc
```diff
@@ -621,6 +619,21 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
                    << " of " << flush_delay_usec << "us during Flush().";
     }
 
+    if (creation_parameters.audio_codec() != kSbMediaAudioCodecNone) {
+      // `use_dual_threads_for_video` has to be disabled if the libopus audio
```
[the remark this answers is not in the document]
xiaomings: Done

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +441,132 @@ void MediaDecoder::DecoderThreadFunc() {
...
 
-  {
+  if (use_dual_threads_) {
+    ScopedLock scoped_lock(mutex_);
+    video_input_condition_variable_.Signal();
+    video_output_condition_variable_.Signal();
+  } else {
     ScopedLock scoped_lock(mutex_);
     condition_variable_.Signal();
   }
```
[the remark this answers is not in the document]
xiaomings: Done

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +441,132 @@ void MediaDecoder::DecoderThreadFunc() {
...
+
+    if (can_process_output) {
+      auto& dequeue_output_result = dequeue_output_results.front();
+      if (dequeue_output_result.index < 0) {
+        host_->RefreshOutputFormat(media_codec_bridge_.get());
+      } else {
+        host_->ProcessOutputBuffer(media_codec_bridge_.get(),
+                                   dequeue_output_result);
+      }
+      dequeue_output_results.erase(dequeue_output_results.begin());
```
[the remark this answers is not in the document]
xiaomings: The vector is usually pretty small, so I'll just keep the current implementation as is.

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +441,132 @@ void MediaDecoder::DecoderThreadFunc() {
...
+  decoder->OutputThreadFunc();
+  JNIState::GetVM()->DetachCurrentThread();
+  return NULL;
+}
+
+void MediaDecoder::OutputThreadFunc() {
+  SB_DCHECK(error_cb_);
+  SB_DCHECK_EQ(media_type_, kSbMediaTypeVideo);
+
+  ::starboard::shared::pthread::ThreadSetPriority(kSbThreadPriorityHigh);
```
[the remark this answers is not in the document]
xiaomings: Done

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
+  decoder->OutputThreadFunc();
+  JNIState::GetVM()->DetachCurrentThread();
+  return NULL;
+}
+
+void MediaDecoder::OutputThreadFunc() {
+  SB_DCHECK(error_cb_);
+  SB_DCHECK_EQ(media_type_, kSbMediaTypeVideo);
+
+  std::vector<DequeueOutputResult> dequeue_output_results;
```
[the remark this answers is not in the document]
xiaomings: The vector is usually pretty small, so I'll just keep the current implementation as is.

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
+
+    if (can_process_output) {
+      auto& dequeue_output_result = dequeue_output_results.front();
+      if (dequeue_output_result.index < 0) {
+        host_->RefreshOutputFormat(media_codec_bridge_.get());
+      } else {
+        host_->ProcessOutputBuffer(media_codec_bridge_.get(),
+                                   dequeue_output_result);
+      }
+      dequeue_output_results.erase(dequeue_output_results.begin());
```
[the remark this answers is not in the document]
xiaomings: The vector is usually pretty small, so I'll just keep the current implementation as is.

kjyoun:
do you want to land this change on 26.android ?

IIUC, now 26.android accepts only critical changes

xiaomings:
@gemini-cli /review

starboard/android/shared/media_decoder.cc
```diff
@@ -230,10 +238,22 @@ void MediaDecoder::WriteInputBuffers(const InputBuffers& input_buffers) {
...
   }
 
-  if (decoder_thread_ == 0) {
-    pthread_create(&decoder_thread_, nullptr,
-                   &MediaDecoder::DecoderThreadEntryPoint, this);
-    SB_DCHECK_NE(decoder_thread_, 0);
+  if (use_dual_threads_) {
+    SB_DCHECK_EQ(media_type_, kSbMediaTypeVideo);
+    if (video_input_thread_ == 0) {
+      pthread_create(&video_input_thread_, nullptr,
```
kjyoun, line 229: can you implement using starboard/common/thread.h (or JobThread) ?

It handles platform-specific thing under the hood (e.g. DetachCurrentThread in Android), so that this PR does not handle it.

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
+  pthread_setname_np(pthread_self(), "VidDecOut");
+  ::starboard::shared::pthread::ThreadSetPriority(kSbThreadPriorityHigh);
+
+  decoder->OutputThreadFunc();
+  JNIState::GetVM()->DetachCurrentThread();
+  return NULL;
+}
+
+void MediaDecoder::OutputThreadFunc() {
+  SB_DCHECK(error_cb_);
```
kjyoun, line 500: nit: why DCHECK error_cb_ ?

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
 
+// static
+void* MediaDecoder::InputThreadEntryPoint(void* context) {
+  SB_CHECK(context);
+  MediaDecoder* decoder = static_cast<MediaDecoder*>(context);
+  pthread_setname_np(pthread_self(), "VidDecIn");
+
+  decoder->InputThreadFunc();
+  JNIState::GetVM()->DetachCurrentThread();
+  return NULL;
```
kjyoun: nullptr

starboard/android/shared/media_decoder.cc
```diff
@@ -745,7 +910,11 @@ void MediaDecoder::OnMediaCodecOutputFormatChanged() {
 
   ScopedLock scoped_lock(mutex_);
   dequeue_output_results_.push_back(dequeue_output_result);
-  condition_variable_.Signal();
+  if (use_dual_threads_) {
```
kjyoun, line 889: This pattern is repeated multiple times

Making this block into separate method would make code more concise.


```
void MediaDecoder::NotifyInputAvailable() {
 if (use_dual_threads_) {
    video_input_condition_variable_.Signal();
    return;
  }

  condition_variable_.Signal();
}

void MediaDecoder::NotifyOutputAvailable() {
 if (use_dual_threads_) {
    video_output_condition_variable_.Signal();
    return;
  }

  condition_variable_.Signal();
}
```

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
+  SB_CHECK(context);
+  MediaDecoder* decoder = static_cast<MediaDecoder*>(context);
+  pthread_setname_np(pthread_self(), "VidDecIn");
+
+  decoder->InputThreadFunc();
+  JNIState::GetVM()->DetachCurrentThread();
+  return NULL;
+}
+
+void MediaDecoder::InputThreadFunc() {
```
kjyoun, line 445: DecodeThreadFunc and RenderThreadFunc 

instead InputThreadFunc and OutputThreadFunc?

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
+  MediaDecoder* decoder = static_cast<MediaDecoder*>(context);
+  pthread_setname_np(pthread_self(), "VidDecOut");
+  ::starboard::shared::pthread::ThreadSetPriority(kSbThreadPriorityHigh);
+
+  decoder->OutputThreadFunc();
+  JNIState::GetVM()->DetachCurrentThread();
+  return NULL;
+}
+
+void MediaDecoder::OutputThreadFunc() {
```
kjyoun, line 497: What do you think about using JobThread? 

Then loop becomes simpler and we don't need to use condition variable


````
ctor or init
std::unique_ptr<OutputThread> output_thread_ = ...


dtor or exit
output_thread_.reset();

when output is available
outptut_thread_.Schedule([this] {
   this.Render();
});


void MediaDecoder::Render() {
   bool can_process_output = !dequeue_output_results.empty();

    if (can_process_output) {
      auto& dequeue_output_result = dequeue_output_results.front();
      if (dequeue_output_result.index < 0) {
        host_->RefreshOutputFormat(media_codec_bridge_.get());
      } else {
        host_->ProcessOutputBuffer(media_codec_bridge_.get(),
                                   dequeue_output_result);
      }
      dequeue_output_results.erase(dequeue_output_results.begin()); 
    } else {
      ScopedLock scoped_lock(mutex_);
      CollectPendingOutputData_Locked(&dequeue_output_results);
    }

    bool ticked = host_->Tick(media_codec_bridge_.get());

    if (!ticked && dequeue_output_results.empty()) {
      ScopedLock scoped_lock(mutex_);
      CollectPendingOutputData_Locked(&dequeue_output_results);
      if (dequeue_output_results.empty()) {
        // TODO(b/329686979): Allow Tick() to return the time to next frame
        // release to dynamically adjust the wait duration.
        video_output_condition_variable_.WaitTimed(8'000);
      }
}

```

kjyoun (commented):
Can you break down this PR into 2?

one for making media_decoder for 2 thread
the one for only plumbing.

Then we can more focus on important stuff in this PR.

starboard/android/shared/media_decoder.cc
```diff
@@ -414,17 +443,133 @@ void MediaDecoder::DecoderThreadFunc() {
...
-    condition_variable_.Signal();
+    if (use_dual_threads_) {
+      video_input_condition_variable_.Signal();
+      video_output_condition_variable_.Signal();
+    } else {
+      condition_variable_.Signal();
+    }
+  }
+
+  if (video_input_thread_ != 0) {
```
kjyoun, line 551: Though this works on Android and this is Android code, this is not recommended practice

You can avoid this, once we use starboard::Thread or JobThread
FYI, main branch is refactored to use sb:Thread or sb::JobThread, instead of direct access to pthread_t

Per Gemini
> You should not compare a pthread_t (often referred to as a thread ID) to 0 or any other integer value if you want your code to be portable and robust
>
> While it might work on specific platforms (like Linux, where pthread_t is often an unsigned long), the POSIX standard defines pthread_t as an opaque type. This means it could be a structure, a pointer, or a large integer, depending on the implementation

starboard/android/shared/media_decoder.cc
xiaomings: As the PR doesn't alter the way the existing code handling thread management, I added a TODO, so it could be handled by a future PR.

starboard/android/shared/media_decoder.cc
xiaomings: It's used to check whether Initialize() is called before the threading functions.  I added comments to all three such CHECKs.

starboard/android/shared/media_decoder.cc
xiaomings: Done

starboard/android/shared/media_decoder.cc
xiaomings: I had thought about this approach.  It doesn't work elegantly with logic like
```
    if (use_dual_threads_) {
      video_input_condition_variable_.Signal();
      video_output_condition_variable_.Signal();
    } else {
      condition_variable_.Signal();
    }
```

The current one could also be slightly easier if we had to remove one branch later or move them into separate classes.

starboard/android/shared/media_decoder.cc
xiaomings: They are named input and output thread funcs as one processes input buffers and one processes output buffers.

DecoderThreadFunc can be ambiguous as the whole class is named `MediaDecoder`.  The rendering in output thread Tick() is indirect and not explicit.

starboard/android/shared/media_decoder.cc
xiaomings: We could refactor this in future using `//starboard/common/thread.h` as you suggested above, but JobThread is traditionally too heavy for such a thread as we try to manually fine tune all the processing.

starboard/android/shared/media_decoder.cc
xiaomings: I will keep this as is, and refactored using starboard::Thread in future.

xiaomings:
> Can you break down this PR into 2?
> 
> one for making media_decoder for 2 thread the one for only plumbing.
> 
> Then we can more focus on important stuff in this PR.

I can see benefits doing this with either way.  Given the experiment handling is pretty straight forward here, I will keep this as is.

xiaomings:
> > Can you break down this PR into 2?
> > one for making media_decoder for 2 thread the one for only plumbing.
> > Then we can more focus on important stuff in this PR.
> 
> I can see benefits doing this with either way. Given the experiment handling is pretty straight forward here, I will keep this as is.

On second thought, I think splitting it might be better for experiment tear down in future, i.e. we can revert the experiment setup PR, and manually flip the feature flag to enable it.

xiaomings:
> > > Can you break down this PR into 2?
> > > one for making media_decoder for 2 thread the one for only plumbing.
> > > Then we can more focus on important stuff in this PR.
> > 
> > 
> > I can see benefits doing this with either way. Given the experiment handling is pretty straight forward here, I will keep this as is.
> 
> On second thought, I think splitting it might be better for experiment tear down in future, i.e. we can revert the experiment setup PR, and manually flip the feature flag to enable it.

https://github.com/youtube/cobalt/pull/9711 is the PR for the feature only, once it's merged, I will re-purpose this PR for the experiment setup.

kjyoun (approved):

starboard/android/shared/player_components_factory.cc
```diff
@@ -630,6 +629,26 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
                    << " of " << flush_delay_usec << "us during Flush().";
     }
 
+    if (experimental_features.use_dual_threads_for_video.value_or(false) &&
```
kjyoun, line 641: this-if block is hard-to-read and similar logic is used above: https://screenshot.googleplex.com/6qJGXf6q5XdzrNB

how about moving this logic into a stand-alone method?

kjyoun (approved):
LGTM % suggestion for readability

starboard/android/shared/player_components_factory.cc
xiaomings: As discussed offline, this comment refers to the decision to decide whether libopus decoder should be used.  I abstracted it into a function.  I didn't use a member function so we don't have to capture `this` above.

kjyoun (approved):

johnxwork (approved):

starboard/shared/starboard/experimental_features.h
```diff
@@ -37,6 +37,7 @@ struct ExperimentalFeatures {
   std::optional<int> video_decoder_initial_preroll_count;
   std::optional<int> video_renderer_min_decoded_frames;
   std::optional<int> video_renderer_min_input_buffers;
+  std::optional<bool> use_dual_threads_for_video;
```
jasonzhangxx, line 40: For bool, we usually don't use optional, like the other bool values.

xiaomings: Is it possible for you to elaborate on this?

`std::optional<bool>` is used here (and we may consider using it for the other bool values), so the Starboard implementation can have a default value when Chrobalt and the backend don't set a value for this particular setting.  If `bool` is used instead, then the default value will be set in Chrobalt, which is less optimal than having it set in the Starboard implementation.

PR #8979, opened 2026-02-05
android: Migrate H5vccPlatformService to JNI

zhongqiliang:
This change migrates the H5vccPlatformService implementation on Android
from the legacy JavaBridge mechanism to direct JNI calls. It removes
the associated JavaBridge classes and JavaScript polyfill.

The Mojo/IDL-based implementation of H5vccPlatformService is now
universally enabled for all Cobalt Android targets, including Android TV,
which streamlines the platform service codepath and ensures consistency.

Bug: 403638702

PR #9486, opened 2026-03-12
3p: Remove h5vcc_platform_service_on_java_bridge

hlwarriner:
This JavaScript file being executed on Evergreen builds and may be causing issues.

It was removed from all builds in main by #8979 and should never have been used for Evergreen, anyway.

Issue: 403638702


andrewsavage1:
Note that I do not expect this to be causing issues (it should be a no-op for non-android platforms) but I'd rather remove it to be safe

PR #9496, opened 2026-03-12
net: Configure asynchronous DNS and DoH

zhongqiliang:
This change configures the network service to use Chromium's fast asynchronous
DNS client. It also sets up DNS-over-HTTPS (DoH) using Google Public DNS
with kAutomatic mode, which prefers DoH but gracefully falls back to
standard DNS if the DoH server is unreachable. This enhances DNS
resolution performance and security.

Bug: 492247862

### DIFF
--- a/cobalt/browser/cobalt_browser_main_parts.cc
+++ b/cobalt/browser/cobalt_browser_main_parts.cc
@@ -33,6 +33,7 @@
 #include "content/public/browser/resource_coordinator_service.h"
 #include "content/public/browser/storage_partition.h"
 #include "net/dns/public/dns_over_https_config.h"
+#include "net/dns/public/doh_provider_entry.h"
 #include "net/dns/public/secure_dns_mode.h"
 #include "services/network/public/mojom/cookie_manager.mojom.h"
 #include "services/network/public/mojom/network_service.mojom.h"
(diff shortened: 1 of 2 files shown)


cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,9 +107,30 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+  ConfigureAsyncDnsAndDoH();
+
   return ShellBrowserMainParts::PreMainMessageLoopRun();
 }
 
+void CobaltBrowserMainParts::ConfigureAsyncDnsAndDoH() {
+  if (auto* network_service = content::GetNetworkService()) {
+    // Use Google Public DNS.
+    auto doh_config = net::DnsOverHttpsConfig::FromString(
+        "https://dns.google/dns-query{?dns}");
```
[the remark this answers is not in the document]
zhongqiliang: added a flag to overwrite this value.

johnxwork:
Can you capture netlog before and after applying the switch and paste to the bug so that we can compare network activity differences?

ckim1346:
Looping in @jellefoks -- I remember discussing investigating potential improvements if we rely more on the system resolver for DNS similar to C25. But IIUC, this PR will attempt to use DoH first and then fall back to the system resolver. Do we have an experiment or local testing data to see if this will improve performance?

zhongqiliang:
> Looping in @jellefoks -- I remember discussing investigating potential improvements if we rely more on the system resolver for DNS similar to C25. But IIUC, this PR will attempt to use DoH first and then fall back to the system resolver. Do we have an experiment or local testing data to see if this will improve performance?

b/492247862#comment6

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -103,9 +110,48 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
+    // By default, we configure it to use Google Public DNS, but this can be
+    // overridden via the --doh-url command line switch to support environments
+    // with specific network policies or different privacy requirements.
+    std::string doh_url = "https://dns.google/dns-query{?dns}";
+    auto* command_line = base::CommandLine::ForCurrentProcess();
+    if (command_line->HasSwitch(switches::kDnsOverHttpsUrl)) {
+      doh_url = command_line->GetSwitchValueASCII(switches::kDnsOverHttpsUrl);
+    }
+
+    auto doh_config = net::DnsOverHttpsConfig::FromString(doh_url);
```
jellefoks: `DnsOverHttpsConfig::FromString` treats the incoming `StringPiece` as json, so I'm not sure if this does anything.

Seconds, can we somehow use the full configs from `DohProviderEntry::GetList()` instead?  (`net/dns/public/doh_provider_entry.cc`)

zhongqiliang: Gemini suggested to not use all from the list, because some are not accessible globally.

In the new code, CleanBrowsingFamily, Cloudflare, Google, OpenDNS are included by default.

PR #9498, opened 2026-03-12
android: add missing sendResponse() call to PlatformError.onDismiss()

ckim1346:
`PlatformError.onDismiss()` was only setting the dialog to null without calling `sendResponse()`. Updated the function to correctly call the missing function. 

Bug: 484078824

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
@@ -171,6 +171,7 @@ public class PlatformError
   @Override
   public void onDismiss(DialogInterface dialogInterface) {
     mDialog = null;
+    sendResponse(mResponse, mData);
   }
 
   /** Informs Starboard when the error is dismissed. */


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -171,6 +171,7 @@ public void onClick(DialogInterface dialogInterface, int whichButton) {
   @Override
   public void onDismiss(DialogInterface dialogInterface) {
     mDialog = null;
+    sendResponse(mResponse, mData);
```
zhongqiliang, line 174: gemini suggested: this should be called before mDialog = null;

ckim1346: Did it say why? C25 we had it set to null first: [onDismiss()](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java;drc=b0dd715816b07fa8e41e92db60f56979b7ab4ef3;l=129)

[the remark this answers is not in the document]
ckim1346: `onClick()` doesn't call `sendResponse()` anywhere... it only sets the value of `mResponse`

zhongqiliang: It printed with that order. I checked the code, the order is not a problem.

PR #9506, opened 2026-03-13
cobalt: Consolidate Crashpad initialization in CobaltMainDelegate

rakuco:
With #9233, a separate Crashpad initialization code path was added to `CobaltMainDelegate::PreSandboxStartup()` specifically for Android TV that called into the existing one in `ShellMainDelegate::PreSandboxStartup()`.

As other platforms are supposed to adopt `CobaltCrashReporterClient` (which is initialized in CobaltMainDelegate), the different code paths and `#ifdef` soup can get confusing quickly.

Attempt to make things simpler to follow:

* Remove `ShellMainDelegate::ZygoteForked()`, which was Linux-specific and never called since Cobalt always launches in single-process mode.
* Merge `ShellMainDelegate::PreSandboxStartup()` into the CobaltMainDelegate implementation:
  * Initialize crash keys and call InitializeResourceBundle() there.
  * For now, keep the existing behavior of unconditionally initializing Crashpad on Android TV but only doing so on other platforms if the kEnableCrashReporter command-line flag is passed. tvOS will do this unconditionally in the future, but Evergreen does this in a completely different manner with a wrapper in Starboard.
  * Compared to the ShellMainDelegate implementation, the only difference in the Crashpad initialization sequence is that SetUploadConsent(true) is always being called.

Bug: 477518757

### DIFF
--- a/.github/config/android-arm.json
+++ b/.github/config/android-arm.json
@@ -10,7 +10,7 @@
   "test_yts": true,
   "test_yts_playback": true,
   "test_browser": false,
-  "test_yts_wpt": false,
+  "test_yts_wpt": true,
   "test_root_target": "//cobalt/android:cobalt_apk",
   "test_attempts": "3",
   "e2e_test_targets": [
@@ -99,7 +99,7 @@
   ],
   "yts_wpt_test_targets": [
     {
-      "target": "testing/container/wpt:wpt_yts_test",
+      "target": "testing/container/wpt:yts_wpt_test_android_arm_devices",
       "test_attempts": "3"
     }
   ],
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+"""Script to automatically roll LTS branch."""
+import argparse
+import re
+import subprocess
+
+
+def get_out(cmd):
+  res = subprocess.run(cmd, capture_output=True, text=True, check=True)
+  return res.stdout.strip()
+
+
+def get_commits(start, target):
+  cmd = [
+      'git', 'rev-list', '--oneline', '--reverse', 'main', f'^{target}',
+      f'{start}^..main'
+  ]
+  return get_out(cmd).splitlines()
+
+
+def cherry_pick(sha):
+  ps = get_out(['git', 'show', '-s', '--format=%P', sha]).split()
+  cmd = ['git', 'cherry-pick', '--no-commit']
+  if len(ps) > 1:
+    cmd.append('--mainline=1')
+  subprocess.run(cmd + [sha], check=True)
+
+
+def commit_pick(sha, num, title):
+  info = get_out(['git', 'log', '-1', '--format=%an <%ae>%n%ad%n%b', sha])
+  auth, date, body = info.split('\n', 2)
+  msg = f'Cherry pick PR #{num}: {title}\n\n'
+  msg += f'Refer to original PR: #{num}\n\n{body}'
+  cmd = [
+      'git', 'commit', '--no-verify', f'--author={auth}', f'--date={date}',
+      '-m', msg
+  ]
+  subprocess.run(cmd, check=True)
+
+
+def main():
+  p = argparse.ArgumentParser()
+  p.add_argument('--target-branch', required=True)
+  p.add_argument('--start-commit', required=True)
+  args = p.parse_args()
+
+  links = []
+  for line in get_commits(args.start_commit, args.target_branch):
+    match = re.match(r'^(\w+) (.*) \(#(\d+)\)$', line)
+    if match:
+      sha, title, num = match.groups()
+      if not get_out([
+          'git', 'log', '-1', f'--grep=^Cherry pick PR #{num}:',
+          args.target_branch
+      ]):
+        cherry_pick(sha)
+        commit_pick(sha, num, title)
+        links.append(f'- #{num}')
+
+  if links:
+    print('\n'.join(links))
+
+
+if __name__ == '__main__':
+  main()
--- a/.github/workflows/autoroll_lts.yaml
+++ b/.github/workflows/autoroll_lts.yaml
@@ -1,0 +1,91 @@
+name: Autoroll LTS
+
+on:
+  schedule:
+    - cron: '0 * * * *' # Hourly.
+  workflow_dispatch:
+    inputs:
+      start_commit:
+        description: 'The commit to start rolling from, should point to the last release chore'
+        required: true
+        default: '7e6524981fdd6ab3c87bc55785343d40116a05e5'
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  autoroll:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Git user
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+      - name: Cherry Pick merged PRs
+        id: autoroll
+        env:
+          TARGET_BRANCH: 27.lts
+          START_COMMIT: ${{ inputs.start_commit }}
+        run: |
+          set -x
+
+          # Ensure we have local branches for both
+          git fetch origin main:main || git fetch origin main
+          git fetch origin "${TARGET_BRANCH}:${TARGET_BRANCH}" || git fetch origin "${TARGET_BRANCH}"
+
+          git checkout -B "autoroll-to-${TARGET_BRANCH}" "${TARGET_BRANCH}" --no-track
+
+          # Run the Python script to perform cherry-picks and get PR links from stdout
+          PR_LINKS=$(python3 .github/scripts/autoroll_lts.py \
+            --target-branch "${TARGET_BRANCH}" \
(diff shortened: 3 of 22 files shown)


cobalt/app/cobalt_main_delegate.cc
```diff
@@ -161,18 +166,35 @@ std::variant<int, content::MainFunctionParams> CobaltMainDelegate::RunProcess(
...
-      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
-          switches::kProcessType);
-  CobaltCrashReporterClient::Create();
-  crash_reporter::InitializeCrashpad(process_type.empty(), process_type);
-  crash_reporter::SetUploadConsent(true);
+#if BUILDFLAG(IS_ANDROIDTV)
+  const bool initialize_crashpad = true;
+#else
+  const bool initialize_crashpad =
+      base::CommandLine::ForCurrentProcess()->HasSwitch(
```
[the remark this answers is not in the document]
rakuco: This doesn't look like that big of an improvement, especially when one of the `#ifdef` code paths doesn't call `base::CommandLine::ForCurrentProcess()`.

rakuco:

cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -304,22 +258,6 @@ std::variant<int, MainFunctionParams> ShellMainDelegate::RunProcess(
 #endif
 }
 
-#if BUILDFLAG(IS_LINUX)
-void ShellMainDelegate::ZygoteForked() {
```
hlwarriner, line 308: Is there a problem with keeping the `ZygoteForked()` path? I think there's a reasonable chance we want to use Zygotes for something in the somewhat near future. And its removal in this PR seems somewhat independent from the consolidation of Crashpad initialization between `CobaltMainDelegate::PreSandboxStartup()` and `ShellMainDelegate::PreSandboxStartup()`.

cobalt/app/cobalt_main_delegate.cc
```diff
@@ -161,18 +166,35 @@ std::variant<int, content::MainFunctionParams> CobaltMainDelegate::RunProcess(
...
 
-#if BUILDFLAG(IS_ANDROIDTV)
 void CobaltMainDelegate::PreSandboxStartup() {
-  std::string process_type =
-      base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
-          switches::kProcessType);
-  CobaltCrashReporterClient::Create();
-  crash_reporter::InitializeCrashpad(process_type.empty(), process_type);
-  crash_reporter::SetUploadConsent(true);
+#if BUILDFLAG(IS_ANDROIDTV)
```
hlwarriner, line 166: @alunev , I don't think the changes in this PR should impact hang reporting but please take a look to confirm.

cobalt/shell/app/shell_main_delegate.h
```diff
@@ -46,13 +46,9 @@ class ShellMainDelegate : public ContentMainDelegate {
   std::optional<int> BasicStartupComplete() override;
   bool ShouldCreateFeatureList(InvokedIn invoked_in) override;
   bool ShouldInitializeMojo(InvokedIn invoked_in) override;
-  void PreSandboxStartup() override;
```
hlwarriner, line 49: @Awallky , please take a look at this PR from a Cobalt / Cobalt shell abstraction perspective. I'd like to make sure you consider the consolidation of logic to be an improvement.

cobalt/shell/app/shell_main_delegate.cc
rakuco: I've added it back. I'd removed it in the first place because the code was basically never used since `kEnableCrashReporter` is never being passed by the ports.

cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -227,35 +218,6 @@ bool ShellMainDelegate::ShouldInitializeMojo(InvokedIn invoked_in) {
...
-// TODO(crbug.com/40188745): Implement crash reporter integration for Fuchsia.
-#if !BUILDFLAG(IS_ANDROIDTV)
-  if (base::CommandLine::ForCurrentProcess()->HasSwitch(
-          switches::kEnableCrashReporter)) {
-    std::string process_type =
-        base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
-            switches::kProcessType);
-    crash_reporter::SetCrashReporterClient(g_shell_crash_client.Pointer());
-    // Reporting for sub-processes will be initialized in ZygoteForked.
-    if (process_type != switches::kZygoteProcess) {
```
rakuco, line 242: @hlwarriner one thing that's occurred to me: would you like me to add this check to `CobaltMainDelegate::PreSandboxStartup()` to have the code ready in case support for zygotes is added to Cobalt?

hlwarriner: Sure, that sounds good to me. Thanks.

rakuco: done

cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -97,11 +93,6 @@ enum class LoggingDest {
 #endif
 };
 
-#if !BUILDFLAG(IS_ANDROIDTV)
-base::LazyInstance<content::ShellCrashReporterClient>::Leaky
```
alunev, line 101: thanks for looking into this. It seems like Cobalt's `ShellCrashReporterClient` is not used anywhere inside cobalt/ now. Seems safe to remove along with this refactoring, could you check if it is?

cobalt/shell/app/shell_main_delegate.h
Awallky: Sorry, just saw this. I'll have a review out today.

cobalt/shell/app/shell_main_delegate.cc
```diff
@@ -227,35 +218,6 @@ bool ShellMainDelegate::ShouldInitializeMojo(InvokedIn invoked_in) {
...
-#endif
-    }
-  }
-#endif  // !BUILDFLAG(IS_ANDROIDTV)
-
-#if !BUILDFLAG(IS_ANDROIDTV)
-  crash_reporter::InitializeCrashKeys();
-#endif  // !BUILDFLAG(IS_ANDROIDTV)
-
-  InitializeResourceBundle();
```
Awallky, line 256: I don't know that we want to remove this. It looks like this method enables the renderer to be able to access the pak resource bundle, which is used for various things like devtools, among others.

It is possible, since we run in single process mode, this is not necessary. If we are going to remove this, I'd like to be sure this is not breaking any existing functionality, particularly devtools, before removing it.

Awallky: Oh, whoops. This is just ported to another call.

rakuco: Note that this call is not being removed, I've just moved it to `CobaltMainDelegate`.

cobalt/shell/app/shell_main_delegate.cc
rakuco: I've looked and indeed it doesn't seem to be used anywhere now. I've `git rm`'ed it from the tree since:
* `GetCrashDumpLocation()` and `EnableBreakpadForProcess()` were only called by `components/crash/core/app/breakpad_linux.cc`, which is only built when `is_castos` is true.
* `GetProductInfo()` is implemented by `CobaltCrashReporterClient`

cobalt/app/cobalt_main_delegate.cc
alunev: lgtm, I've left one comment below.

alunev (approved):
lgtm, left one comment about further cleanup of cobalt's version of ShellCrashReporterClient

rakuco:
> @rakuco I think my main concern is removing the bundle init, which browsertests may depend on. Can we keep that in place and move everything else?

@Awallky done, I've also moved crash key initialization back to `ShellMainDelegate` since it's independent from the Crashpad initialization sequence and they may end up being used in the browser tests code path.

With that said, can you elaborate on what should go into CobaltMainDelegate vs ShellMainDelegate? Sometimes it's not totally clear what should go where (e.g. why don't browser tests use CobaltMainDelegate?), and the fact that `CobaltMainDelegate`'s constructor takes an `is_content_browsertests` parameter but it's always false makes it even more confusing.

Awallky:
@rakuco Great question!

I agree it is an not ideal architecture. Historically, cobalt based on the original content shell. While it was based on content shell, Cobalt devs made changes to the CobaltMainDelegate to avoid changes to upstream Chromium. After decoupling cobalt from content shell, porting to our own version of it, the need for two distinct layers is unnecessary. As you mentioned, this can make it more difficult to parse through the layers of the application.

IMO merging the two layers would be a significant improvement.
This will require some changes to not just the app but browser tests. It just has not been prioritized over other efforts to this point.

rakuco:
@Awallky thanks for the explanation! Is there a tracking bug for this? I can file one if not.

In the meantime, if this PR looks fine now can either you or @hlwarriner merge it? I don't have write access to the repo myself.

Awallky:
@rakuco I don't think we have a bug filed for it yet. Feel free to file one.
LGTM. Happy to merge this change.

rakuco:
@Awallky @fayev-g I thought these arm-hardfp-rdk_e2e_tests were not required to pass, but they seem to be preventing the PR from being merged?

hlwarriner:
> * code owner review from youtube/cobalt-embedder-owners

Merging is actually blocked by code owner review from youtube/cobalt-embedder-owners. Approval from @Awallky would do it.

PR #9515, opened 2026-03-13
evergreen: Set version annotation in crash reports

hlwarriner:
ReadEvergreenVersion() is extracted into its own library so that it can be used in both Evergreen Full and Lite configurations.

Unlike in previous branches, the annotation is now set from below Cobalt just before an installation is loaded (both for Full and Lite). This enables attribution of earlier crashes to the correct version.

The changes were tested end-to-end by creating a manifest file in the expected installation slot and forcing a crash. It's not straightforward to add unit tests for the new behaviors...I may need to revive go/starboard-test-doubles.

Issue: 485307186

### DIFF
--- a/starboard/loader_app/BUILD.gn
+++ b/starboard/loader_app/BUILD.gn
@@ -182,7 +182,7 @@ static_library("read_evergreen_version") {
   ]
   deps = [
     "//starboard/common",
-    "//third_party/jsoncpp:jsoncpp",
+    "//third_party/jsoncpp",
   ]
   configs += [ "//third_party/jsoncpp:jsoncpp_config" ]
 }
@@ -195,7 +195,7 @@ source_set("read_evergreen_version_test") {
     "//starboard/common",
     "//testing/gmock",
     "//testing/gtest",
-    "//third_party/jsoncpp:jsoncpp",
+    "//third_party/jsoncpp",
   ]
   configs += [ "//third_party/jsoncpp:jsoncpp_config" ]
 }


starboard/loader_app/read_evergreen_version.cc
```diff
@@ -0,0 +1,68 @@
...
+  int read_size = manifest_file.ReadAll(file_data.data(), file_size);
+  if (read_size < 0) {
+    SB_LOG(WARNING) << "Error while reading from the manifest file.";
+    return false;
+  }
+
+  Json::Reader reader;
+  Json::Value obj;
+  if (!reader.parse(std::string(file_data.data(), file_size), obj) ||
+      !obj[kVersionKey]) {
```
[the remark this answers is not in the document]
hlwarriner: done

starboard/loader_app/read_evergreen_version_test.cc
```diff
@@ -0,0 +1,97 @@
...
+
+class ReadEvergreenVersionTest : public testing::Test {
+ protected:
+  virtual void SetUp() {
+    dir_.resize(kSbFileMaxPath);
+    ASSERT_TRUE(
+        SbSystemGetPath(kSbSystemPathTempDirectory, dir_.data(), dir_.size()));
+    manifest_path_.resize(kSbFileMaxName);
+    snprintf(manifest_path_.data(), kSbFileMaxPath, "%s%s%s", dir_.data(),
+             kSbFileSepString, kManifestFileName);
```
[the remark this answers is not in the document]
hlwarriner: done

starboard/loader_app/read_evergreen_version.cc
```diff
@@ -0,0 +1,68 @@
...
+    SB_LOG(WARNING) << "Failed to parse version from the manifest file at the "
+                       "installation path.";
+    return false;
+  }
+
+  snprintf(version, version_length, "%s", obj[kVersionKey].asString().c_str());
+  return true;
```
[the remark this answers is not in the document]
hlwarriner: This is good feedback but I'd actually like to return `false` from the function in the event of truncation, which changes the contract of the function. I think it would be cleaner to make this change separately so added a todo comment.

starboard/loader_app/BUILD.gn
```diff
@@ -174,6 +175,31 @@ source_set("drain_file_test") {
...
 }
 
+static_library("read_evergreen_version") {
+  sources = [
+    "read_evergreen_version.cc",
+    "read_evergreen_version.h",
+  ]
+  deps = [
+    "//starboard/common",
+    "//third_party/jsoncpp:jsoncpp",
```
andrewsavage1: nit: use shorthand

starboard/loader_app/BUILD.gn
```diff
@@ -174,6 +175,31 @@ source_set("drain_file_test") {
...
+
+source_set("read_evergreen_version_test") {
+  testonly = true
+  sources = [ "read_evergreen_version_test.cc" ]
+  deps = [
+    ":read_evergreen_version",
+    "//starboard/common",
+    "//testing/gmock",
+    "//testing/gtest",
+    "//third_party/jsoncpp:jsoncpp",
```
andrewsavage1: shorthand

starboard/loader_app/read_evergreen_version.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef STARBOARD_LOADER_APP_READ_EVERGREEN_VERSION_H_
+#define STARBOARD_LOADER_APP_READ_EVERGREEN_VERSION_H_
+
+#include <vector>
+
+namespace loader_app {
+
+// The max length of Evergreen version string.
+extern const int kMaxEgVersionLength;
```
andrewsavage1, line 23: Can't we just define it here too so it doesn't have to be extern (or is that only possible with constexpr)?

andrewsavage1 (approved):
Can you update the title to make it more clear that this is more of a refactor than a behavioral change (assuming I'm understanding it correctly)?

starboard/loader_app/BUILD.gn
hlwarriner: done

starboard/loader_app/BUILD.gn
hlwarriner: done

starboard/loader_app/read_evergreen_version.h
hlwarriner: What's the reason you want to make that change?

This pattern is used elsewhere under `//starboard/loader_app` and is one of the two patterns recommended internally by the C++ Library team (search for "Constants in Header Files"). That guide does also recommend using  `constexpr` in the .cc definition to ensure "each variable is a constant, is compile-time initialized, and has a trivial destructor." But it notes that some older toolchains have issues with this use of `constexpr`, so it's probably best if we omit that.

The other pattern recommended in that guide is, as you suggested, to define it in the header like `inline constexpr int kMaxEgVersionLength = xyz`. It seems like the main reason to choose this alternative is to make the value accessible at compile time, but I don't think we need that. So I'd prefer to keep the current declaration/definition approach unless you feel strongly.

hlwarriner:
> Can you update the title to make it more clear that this is more of a refactor than a behavioral change (assuming I'm understanding it correctly)?

I think the most important aspect actually is the behavior change: without this PR, we aren't setting the version annotation in Evergreen's crash reports. The refactoring is done to enable us to make the behavior change.

starboard/loader_app/read_evergreen_version.h
andrewsavage1: Yeah, we're not able to use constexpr for our platform level toolchains, so let's not do this. I just suggested it initially because I think it's more readable to have the constant defined and declared in the same place

starboard/loader_app/loader_app.cc
```diff
@@ -104,6 +130,9 @@ void LoadLibraryAndInitialize(const std::string& alternative_content_path,
   } else {
     content_path = alternative_content_path.c_str();
   }
+
+  InsertVersionAnnotationFromManifest(content_dir);
```
andrewsavage1, line 134: Should we be calling this in evergreen full configs? Wouldn't slot_management handle it?

andrewsavage1:
> > Can you update the title to make it more clear that this is more of a refactor than a behavioral change (assuming I'm understanding it correctly)?
> 
> I think the most important aspect actually is the behavior change: without this PR, we aren't setting the version annotation in Evergreen's crash reports. The refactoring is done to enable us to make the behavior change.

Got it, I had incorrectly assumed that slot_management.cc was already doing this—keep the title.

starboard/loader_app/loader_app.cc
hlwarriner: So this `LoadLibraryAndInitialize()` function is used for Evergreen-Lite, and also as a fallback for Evergreen-Full in the event that the Installation Manager can't be initialized.

`LoadSlotManagedLibrary()` is used for Evergreen-Full.

And this PR sets the version annotation from both functions.

Hope that helps - feel free to reopen if it's not clear or you still have concerns.

PR #9521, opened 2026-03-14
Pass operation result to unpacker

yuying-y:
Modify the update client's unpacking and installation pipeline to pass the
full OperationResult object. Cache is bypassed because Cobalt update doesn't use temporary file.

Issue: 479816505

### DIFF
--- a/components/update_client/op_install.cc
+++ b/components/update_client/op_install.cc
@@ -280,4 +280,3 @@ base::OnceClosure InstallOperation(
 }
 
 }  // namespace update_client
-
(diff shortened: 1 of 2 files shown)


components/update_client/op_install.cc
```diff
@@ -243,13 +272,12 @@ base::OnceClosure InstallOperation(
...
-#endif
                              std::move(callback), event_adder, crx_file),
               std::move(install_params), installer, progress_callback),
           crx_file, std::move(unzipper), pk_hash, crx_format));
+#endif
   return base::DoNothing();
 }
 
 }  // namespace update_client
+
```
andrewsavage1: nit: remove

components/update_client/unpacker.h
```diff
@@ -11,6 +11,15 @@
 #include <string>
 #include <vector>
 
+// These headers cannot be guarded by BUILDFLAG(IS_STARBOARD) because
+// they define BUILDFLAG(IS_STARBOARD)
+#include "build/build_config.h"
+#include "build/buildflag.h"
```
andrewsavage1: You should only need to include build_config.h

PR #9524, opened 2026-03-14
Add new template for browsertest in Evergreen

linxinan-chops:
Unify the entrance of browser tests on Evergreen.

Also updated the main file to align with Starboard on "exit".

Test: out/evergreen-x64_devel/cobalt_browsertests_loader.py \
   --gtest_filter=H5vccSchemeURLLoaderFactoryCacheBrowserTest.SplashContentTooLarge
Issue: 463991461


starboard/build/util/generate_runfile.py
```diff
@@ -26,13 +26,45 @@
...
+command = [sys.executable, run_browser_tests, elf_loader] + library_args + sys.argv[1:]
+try:
+    result = subprocess.run(command, check=False)
+    sys.exit(result.returncode)
+except subprocess.CalledProcessError:
+    # A subprocess failed, so don't log the python traceback.
+    raise SystemExit(1)
+except Exception as e:
+    print("An unexpected error occurred: " + str(e), file=sys.stderr)
+    sys.exit(1)
```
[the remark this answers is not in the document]
linxinan-yt: Simply copied the existing script. If necessary, I will file follow up PR to update both.

cobalt/testing/browser_tests/cobalt_browser_tests_main.cc
```diff
@@ -67,30 +71,42 @@ SB_EXPORT void SbEventHandle(const SbEvent* event) {
...
-      // because Starboard's Musl port specifically maps _exit() back to exit(),
-      // which runs the problematic teardown logic anyway. std::_Exit()
-      // (uppercase) bypasses this and invokes the raw SYS_exit_group syscall.
-      // TODO(b/463991461): Consider ASAN_OPTIONS=exitcode=0 to avoid the need
-      // for std::_Exit() here.
-      std::_Exit(test_result_code);
+      // std::_Exit() (uppercase) bypasses the C library entirely and invokes
+      // the raw SYS_exit_group syscall, escaping the sandbox and terminating
+      // cleanly.
+      std::_Exit(s_test_result_code);
```
linxinan-yt, line 109: @andrewsavage1 

I tried ASAN_OPTIONS, but seems not crash is not only from ASAN if I go through the normal exit of starboard for single process browser tests. Got SIGABRT, https://paste.googleplex.com/4541697771044864. 

For now, I think _Exit() is the best I can have. WDYT?

starboard/build/util/generate_runfile.py
```diff
@@ -26,13 +26,47 @@
...
 """
 
+_BROWSERTEST_TEMPLATE = """#!/usr/bin/env python3
+import os
+import subprocess
+import sys
+
+# The run_browser_tests.py script is copied to the output directory as
+# 'cobalt_browsertests_runner' at the build time.
+run_browser_tests = os.path.abspath(
```
andrewsavage1: Why abspath?

starboard/build/util/generate_runfile.py
```diff
@@ -26,13 +26,47 @@
...
+        'cobalt_browsertests_runner'
+    )
+)
+
+elf_loader = os.path.join(os.path.dirname(__file__), 'elf_loader_sandbox')
+library_args = ['--evergreen_content=.', '--evergreen_library={library}.so']
+
+# Pass the elf_loader as the "binary" argument to the runner script,
+# followed by the required library args, and then the rest of sys.argv
+command = [sys.executable, run_browser_tests, elf_loader] + library_args + sys.argv[1:]
```
andrewsavage1: The only difference between the script generation is these two arguments in the beginning of the list. Maybe there's a good way to reuse the existing template? Alternatively, it seems like this would only be used for a single target (is that correct?), in which case we could just use a python file directly that's copied to the out directory instead of a template

linxinan-yt: You are right. I probably don't need to copy _Template. Refactor it and allow browser test to inject its binary.

starboard/build/util/generate_runfile.py
linxinan-yt: Nice catch. Removed.

starboard/build/util/generate_runfile.py
```diff
@@ -12,7 +12,7 @@
 import sys
 
 command = [
-    os.path.join(os.path.dirname(__file__), 'elf_loader_sandbox'),
+    {runner_args}os.path.join(os.path.dirname(__file__), 'elf_loader_sandbox'),
     '--evergreen_content=.', '--evergreen_library={library}.so'
 ] + sys.argv[1:]
 try:
```
andrewsavage1, line 23: I think this may read cleaner if you change this a bit like so:

```
if is_browsertest:
  command = [sys.executable, os.path.join(os.path.dirname(__file__), 'cobalt_browsertests_runner'] + command
```

And then:
```
is_browsertest = args.library == 'libcobalt_browsertests'
_TEMPLATE.format(library=args.library, is_browsertest=is_browsertest)
```

linxinan-yt: True. This is ugly. Thanks for catching it.
Done.

andrewsavage1 (approved):

andrewsavage1:
Looks good, thanks for the iteration :)

PR #9531, opened 2026-03-16
Track memory usage of MediaCodec output buffers

jasonzhangxx:
Bug: 487768963


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -455,6 +485,7 @@ public void onOutputBufferAvailable(
                       info.offset,
                       info.presentationTimeUs,
                       info.size);
```
yell0wd0g, line 487: I think this code is pretty nifty. Have you tried this in
a few AndroidTV devices and does this size correlate with
the amount of bytes a texture (or 2 textures, since it'll
be NV12) would occupy? I would be wary that the driver
doesn't populate this number accurately.

PR #9538, opened 2026-03-16
Revert "Fix TTS on Starboard platforms"

isarkis:
Reverts youtube/cobalt#9360


isarkis:
We are bisecting trying to figure out which PR broke E2E tests on 26.eap, do not review this PR.

andrewsavage1:
@niranjanyardi I know you're already looking at it and confirmed that this was the PR that broke e2e tests on RDK, did you communicate that with engprod?

PR #9554, opened 2026-03-17
viz: Fix bad_optional_access crash on VizCompositorThread

jellefoks:
Fix a SIGABRT crash occurring during Reveal/Conceal transitions on WebOS platforms. The crash was caused by unprotected .value() calls on empty absl::optional members.

Key fixes:
- SurfaceAggregator: Add has_value() check before accessing per_quad_damage_rect in PrewalkRenderPass. This prevents crashes when damage_type is kFrame but quads like CompositorRenderPass lack damage.
- SkiaOutputSurfaceImpl: Add null check for root_ddl_recorder_ in BeginPaintCurrentFrame to handle cases where paint starts before Reshape().
- OverlayProcessorSurfaceControl: Safely handle empty output_surface_plane in AdjustOutputSurfaceOverlay.

These changes improve robustness during rapid application lifecycle state changes.

Bug: 493293791


jellefoks:
Did not solve the issue

PR #9555, opened 2026-03-17
Refactor cobalt_browsertest on Linux

linxinan-yt:
1. Simplify the entrance of cobalt_browsertests to bin/run_cobalt_browsertests like Android build.
2. Switch the build target in CI to cobalt_browsertests. The runner python script now is only an implementation and invisible to users and CI.
3. Correct the CI config to remove cobalt_browsertests from bots running unit tests. cobalt_browsertests now only runs on its dedicated workflow.

Test: out/linux-x64x11_devel/bin/run_cobalt_browsertests --gtest_filter=H5vccSchemeURLLoaderFactoryBrowserTest.* 
Issue: 463991461

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -223,7 +223,6 @@ jobs:
           echo "browser_test_targets=${browser_test_targets}" >> $GITHUB_OUTPUT
 
           # Also export just the target strings for the results matrix.
-          PLATFORM="${{ inputs.platform }}"
           browser_test_targets_json=$(echo "${browser_test_targets}" | jq -c 'map(.target)')
           echo "browser_test_targets_json=${browser_test_targets_json}" >> $GITHUB_OUTPUT
 


.github/workflows/main.yaml
```diff
@@ -226,8 +226,8 @@ jobs:
           PLATFORM="${{ inputs.platform }}"
           if [[ "$PLATFORM" == *linux* ]]; then
              # For Linux host, the target is in the primary 'targets' list.
-             if cat ${GITHUB_WORKSPACE}/.github/config/${{ inputs.platform }}.json | jq -e '.targets[] | select(contains("cobalt_browsertests_runner"))' > /dev/null; then
-               browser_test_targets_json='["cobalt/testing/browser_tests:cobalt_browsertests_runner"]'
+             if cat ${GITHUB_WORKSPACE}/.github/config/${{ inputs.platform }}.json | jq -e '.targets[] | select(contains("cobalt_browsertests"))' > /dev/null; then
+               browser_test_targets_json='["cobalt/testing/browser_tests:cobalt_browsertests"]'
```
oxve: We should make this a configuration flag in the json instead. Can you add a TODO?

testing/test.gni
```diff
@@ -1065,7 +1065,15 @@ template("test") {
         data += [ "//testing/location_tags.json" ]
       }
 
-      executable_args = [
+      executable_args = []
```
oxve, line 1068: Since this is 3rd party code, should it have an is_cobalt somewhere?

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -73,7 +73,7 @@ runs:
           # Determine the browser test target for this platform.
           BROWSERTEST_TARGET="cobalt/testing/browser_tests:cobalt_browsertests"
           if [[ "${PLATFORM}" == *linux* ]]; then
```
Awallky: You don't need this if block at all now.

linxinan-yt: True. Thanks.

.github/workflows/main.yaml
linxinan-yt: Ack.

testing/test.gni
linxinan-yt: You are right. I am thinking to push this change to chromium, but it may take some time.

PR #9583, opened 2026-03-17
media: Suppress repeated key events

sideb0ard:
Suppress repeated DPAD CENTER key events in CobaltActivity.java. This prevents main thread flooding from excessive input, improving microphone response when using a softmic.

Bug: 483713292

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -327,10 +327,10 @@ public abstract class CobaltActivity extends Activity {
   @Override
   public boolean onKeyDown(int keyCode, KeyEvent event) {
     if (keyCode == KeyEvent.KEYCODE_DPAD_CENTER && event.getRepeatCount() > 0)  {
-        // It's a repeat, consume it and do not propagate. We found this was
-        // flooding our main thread with key events during soft mic usage.
-        // https://b.corp.google.com/issues/483713292#comment46
-        return true;
+      // It's a repeat, consume it and do not propagate. We found this was
+      // flooding our main thread with key events during soft mic usage.
+      // https://b.corp.google.com/issues/483713292#comment46
+      return true;
     }
     // If input is a from a gamepad button, it shouldn't be dispatched to IME which incorrectly
     // consumes the event as a VKEY_UNKNOWN


kjyoun:
nit: can you update title per go/cobalt-style#commit-message

e.g.) media: Speed up microphone calls

media/audio/android/audio_manager_android.cc
```diff
@@ -39,6 +39,13 @@ using media_audio_android::InitializeStubs;
 using media_audio_android::kModuleAaudio;
 using media_audio_android::StubPathMap;
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+const int kDefaultOutputSampleRate = 48000;
```
kjyoun: s/const/constexpr/

media/audio/android/audio_manager_android.cc
```diff
@@ -39,6 +39,13 @@ using media_audio_android::InitializeStubs;
...
+const int kCobaltMicrophoneSampleRate = 16000;
```
kjyoun: nit: can you add digit separate for readability?

e.g.)
48'000;
16'000;

media/audio/android/opensles_input.cc
```diff
@@ -9,6 +9,11 @@
 #include "media/audio/android/audio_manager_android.h"
 #include "media/base/audio_bus.h"
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+#include "base/threading/platform_thread.h"
+constexpr int kCobaltSampleRate = 16000;
```
kjyoun: nit: 16'000

kjyoun:
Can you share some data about how much these changes improve latency? (e.g. on your local dev. environment)

sideb0ard:
> Can you share some data about how much these changes improve latency? (e.g. on your local dev. environment)

It certainly doesn't fix the issue yet, but it reduces latency from ~7 seconds to ~2 seconds.  

(I'll continue to work on it this week, but wanted to get something up so review can start as there  is a push for me to get something submitted before i'm out next week so we can see how it changes watchtime).

sideb0ard:
> Do we really all changes that this PR adopted?
> * consuming DPAD key

yeah,  you're totally right. I've pared this back to just the DPAD change, which is the only change I'm certain we need.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -326,6 +326,12 @@ protected boolean dispatchKeyEventToIme(int keyCode, int action) {
 
   @Override
   public boolean onKeyDown(int keyCode, KeyEvent event) {
+    if (keyCode == KeyEvent.KEYCODE_DPAD_CENTER) {
```
kjyoun: By combining check, we can avoid one-level of nesting.

go/tott-nesting


```
if (keyCode == KeyEvent.KEYCODE_DPAD_CENTER 
    && event.getRepeatCount() > 0) {
    // It's a repeat, consume it and do not propagate
    return true;
}
```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -326,6 +326,12 @@ protected boolean dispatchKeyEventToIme(int keyCode, int action) {
...
+      if (event.getRepeatCount() > 0) {
+        // It's a repeat, consume it and do not propagate
```
kjyoun: can you add comment about why we do this?

e.g.)
// We consume DPAD center to avoid latency. For details, see http://b/483713292#comment46

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
sideb0ard: done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
sideb0ard: done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -326,11 +326,11 @@ protected boolean dispatchKeyEventToIme(int keyCode, int action) {
 
   @Override
   public boolean onKeyDown(int keyCode, KeyEvent event) {
-    if (keyCode == KeyEvent.KEYCODE_DPAD_CENTER) {
-      if (event.getRepeatCount() > 0) {
-        // It's a repeat, consume it and do not propagate
+    if (keyCode == KeyEvent.KEYCODE_DPAD_CENTER && event.getRepeatCount() > 0)  {
+        // It's a repeat, consume it and do not propagate. We found this was
```
kjyoun: I think that you should use 2-space indentation here, instead of 4.

Looks like formatter did not work on this file.

kjyoun:
super minor nit-picking:  for title: media: "S"uppress repeated key events  ?

per go/cobalt-style#commit-message

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
sideb0ard: done

PR #9601, opened 2026-03-18
tvOS: Handle missing UIWindow in GetPlatformWindow()

Gyuyoung:
Browser tests on tvOS may run without a connected UIWindowScene,
as the test shell does not participate in the normal UIKit scene
lifecycle.

GetPlatformWindow() previously assumed exactly one scene was always
available, leading to an assert when connectedScenes was empty. This
change updates GetPlatformWindow() to return nil in such cases and
modifies call sites to safely handle the absence of a UIWindow.

This prevents crashes in browser test environments that do not have
an active UIWindowScene.

Bug: 489556129

### DIFF
--- a/starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
+++ b/starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
@@ -132,6 +132,11 @@ const int kRequiredBuffersInDisplayLayer = 16;
 UIWindow* GetPlatformWindow() {
   NSSet<UIScene*>* connected_scenes =
       UIApplication.sharedApplication.connectedScenes;
+  if (connected_scenes.count == 0) {
+    return nil;
+  }
+  // We don't expect multiple scenes on tvOS. This will assert if that
+  // assumption is violated.
   SB_CHECK_EQ(connected_scenes.count, 1U);
   UIWindowScene* scene =
       base::apple::ObjCCastStrict<UIWindowScene>(connected_scenes.anyObject);


Gyuyoung:
CC'ing @abhijeetk @jkim-julie

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
```diff
@@ -132,7 +132,10 @@ - (BOOL)resourceLoader:(AVAssetResourceLoader*)resourceLoader
...
   NSSet<UIScene*>* connected_scenes =
       UIApplication.sharedApplication.connectedScenes;
-  SB_CHECK_EQ(connected_scenes.count, 1U);
+  if (connected_scenes.count == 0) {
+    return nil;
+  }
+
   UIWindowScene* scene =
       base::apple::ObjCCastStrict<UIWindowScene>(connected_scenes.anyObject);
   return scene.keyWindow;
```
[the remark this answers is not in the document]
Gyuyoung: Applied.

rakuco:
> Browser tests on tvOS may run without a connected UIWindowScene,
> as the test shell does not participate in the normal UIKit scene
> lifecycle.

The question I have is: should browser tests run without a connected UIWindowScene in the first place? Maybe the launch process that needs to change instead: either the launcher code should be integrated with `base`'s launcher infrastructure (i.e. `base/test/launcher/unit_test_launcher.cc` and others), or something similar to #8732 needs to happen and `base::RunTestsFromIOSApp()` needs to be called from the existing launcher.

cc @Awallky

Gyuyoung:
> > Browser tests on tvOS may run without a connected UIWindowScene,
> > as the test shell does not participate in the normal UIKit scene
> > lifecycle.
> 
> The question I have is: should browser tests run without a connected UIWindowScene in the first place? Maybe the launch process that needs to change instead: either the launcher code should be integrated with `base`'s launcher infrastructure (i.e. `base/test/launcher/unit_test_launcher.cc` and others), or something similar to #8732 needs to happen and `base::RunTestsFromIOSApp()` needs to be called from the existing launcher.
> 
> cc @Awallky

Thank you for the comment. As you know, cobalt_browsertests runs in multiple processes by default. When running the tests, two processes were created. I believe the first one was the launcher process and the other one was the test process.

The issue was that the launcher process was connected to a UIWindowScene, but the second process for running the test was not connected to the scene, even though both were started through base::RunTestsFromIOSApp(). I am not sure yet why the second process is not connected to the scene.

FYI, the call sequence for both processes is as follows:

  1. main() : //cobalt/testing/browser_tests/content_test_launcher.cc
  2. LaunchTest() : //content/public/test/test_launcher.cc
  3. RunTestsFromIOSApp() : //base/test/test_support_ios.mm

However, only the launcher process was connected to the scene, while the second process was not.

Also, there was no assert when running a test with the `--single-process-tests` option, because in that case the single process created was connected to the scene.

I'm not sure whether browser tests are expected to run without a connected UIWindowScene in the test process. Please let me know if you have any clues about why the second process is not connected to the scene, or if this behavior is expected.

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
```diff
@@ -132,7 +132,12 @@ - (BOOL)resourceLoader:(AVAssetResourceLoader*)resourceLoader
 UIWindow* GetPlatformWindow() {
   NSSet<UIScene*>* connected_scenes =
       UIApplication.sharedApplication.connectedScenes;
-  SB_CHECK_EQ(connected_scenes.count, 1U);
+  if (connected_scenes.count == 0) {
+    return nil;
+  }
+  // We don't expect multiple scenes on tvOS. This will assert in debug builds
+  // if that assumption is violated.
+  SB_DCHECK_EQ(connected_scenes.count, 1U);
```
rakuco: I'd keep this a `CHECK` rather than a `DCHECK`.

rakuco (commented):
Thanks for the reminder, I forgot that we actually supported launching a separate process in browser tests.

Can you just check if other implementations of the `GetPlatformWindow()` code upstream (there are a few different functions that try to get the main scene or platform window) also allow no connected scenes? If so, I think there isn't much to do and this can go in as-is.

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
Gyuyoung: > I'd keep this a `CHECK` rather than a `DCHECK`.

ok, fixed.

Gyuyoung:
> Thanks for the reminder, I forgot that we actually supported launching a separate process in browser tests.
> 
> Can you just check if other implementations of the `GetPlatformWindow()` code upstream (there are a few different functions that try to get the main scene or platform window) also allow no connected scenes? If so, I think there isn't much to do and this can go in as-is.

I could only find GetPlatformWindow() in `//chrome/browser/headless/headless_mode_browsertest_linux.cc`[1]. But it was for linux port, so it didn't handle any scene. Otherwise, the scene has only been processed by //ios/chrome/*. But, it looks like they've only handled the scenes for the other purposes related to UI stuff. And, I only found such a check(e.g. CHECK_EQ(controller.state.connectedScenes.count, 0u)) in destructors before removing observers. So, it's difficult to say that the other `GetPlatformWindow()` allows no connected scenes. This appears to be the first case about it.

[1] https://github.com/youtube/cobalt/blob/main/chrome/browser/headless/headless_mode_browsertest_linux.cc#L59

Gyuyoung:
> > Thanks for the reminder, I forgot that we actually supported launching a separate process in browser tests.
> > Can you just check if other implementations of the `GetPlatformWindow()` code upstream (there are a few different functions that try to get the main scene or platform window) also allow no connected scenes? If so, I think there isn't much to do and this can go in as-is.
> 
> I could only find GetPlatformWindow() in `//chrome/browser/headless/headless_mode_browsertest_linux.cc`[1]. But it was for linux port, so it didn't handle any scene. Otherwise, the scene has only been processed by //ios/chrome/*. But, it looks like they've only handled the scenes for the other purposes related to UI stuff. And, I only found such a check(e.g. CHECK_EQ(controller.state.connectedScenes.count, 0u)) in destructors before removing observers. So, it's difficult to say that the other `GetPlatformWindow()` allows no connected scenes. This appears to be the first case about it.
> 
> [1] https://github.com/youtube/cobalt/blob/main/chrome/browser/headless/headless_mode_browsertest_linux.cc#L59

@rakuco Any comment?

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
```diff
@@ -218,7 +223,9 @@ - (BOOL)resourceLoader:(AVAssetResourceLoader*)resourceLoader
     [display_view removeFromSuperview];
 
     // Clear preferred display criteria.
-    GetPlatformWindow().avDisplayManager.preferredDisplayCriteria = nil;
```
rakuco, line 221: Does it work if you keep this line instead? This is Objective-C, so it's safe to send messages to `nil` (i.e. this is shorthand for `[[nil avDisplayManager] setPreferredDisplayCriteria:nil]`).

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
```diff
@@ -405,7 +412,14 @@ - (BOOL)resourceLoader:(AVAssetResourceLoader*)resourceLoader
 }
 
 void AVSBVideoRenderer::UpdatePreferredDisplayCriteria() {
-  AVDisplayManager* avDisplayManager = GetPlatformWindow().avDisplayManager;
```
rakuco, line 408: Same thing about not needing to change this.

rakuco (commented):
So what's happening here is that the [scene initialization code](https://github.com/youtube/cobalt/blob/e551786cc03840e5d3045ac9e6d09b650008924c/base/test/test_support_ios.mm#L106) is only called in the launcher process, so `GetPlatformWindow()` never finds any connected scenes when called from a forked process like in the browser tests. I think this is a corner case since according to the platform we were not supposed to be running multiple processes anyway.

I also think that this approach is good enough, if there's a connected scene it means this is the main app running on the device and there will be a valid `AVDisplayManager`.

I've added some inline comments that can possibly simplify this patch further.

rakuco:
> I also think that this approach is good enough, if there's a connected scene it means this is the main app running on the device and there will be a valid `AVDisplayManager`.

Just for the record: I'm testing some different approaches that could make a new process have an associated `UIWindow`. It's a valid improvement in the sense that it also ensures the view from `shell_platform_delegate_ios.mm` is also loaded in browser tests, but it doesn't help much in this specific case since AFAICS in the simulator the "Match Content" settings that the code in this PR is related to do not work anyway.

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
Gyuyoung: ok, I see. Then, let's keep it as it is.

starboard/tvos/shared/media/av_sample_buffer_video_renderer.mm
Gyuyoung: Done.

Gyuyoung (commented):
Fixed all and the failing tests still passed with this change.

rakuco (approved):
lgtm

fayev-g (approved):

PR #9617, opened 2026-03-18
android: Remove tunnel mode preroll timeout fallback

jasonzhangxx:
Remove the preroll timeout fallback for tunnel mode video decoding.
The previous preroll completion logic in tunnel mode, which included a
timeout, was unreliable. This change introduces a more robust
mechanism that relies on the first frame being rendered and a
specified number of frames being processed to signal preroll
completion.

Bug: 487397946

### DIFF
--- a/starboard/android/shared/video_decoder.cc
+++ b/starboard/android/shared/video_decoder.cc
@@ -239,9 +239,10 @@ const int64_t kNeedMoreInputCheckIntervalInTunnelMode = 50'000;  // 50ms
 
 const int kInitialPrerollFrameCount = 8;
 const int kNonInitialPrerollFrameCount = 1;
-
-const int kSeekingPrerollPendingWorkSizeInTunnelMode =
-    16 + kInitialPrerollFrameCount;
+// According to b/487397946#comment3, after the first non-DECODE_ONLY frame is
+// rendered, the rest of the playback should play without frame drops. So,
+// tunnel mode prerolling only needs 1 frame.
+const int kTunnelModePrerollFrameCount = 1;
 const int kDefaultMaxPendingInputsSize = 128;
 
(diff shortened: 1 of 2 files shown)


starboard/android/shared/video_decoder.cc
```diff
@@ -239,9 +239,7 @@ const int64_t kNeedMoreInputCheckIntervalInTunnelMode = 50'000;  // 50ms
 
 const int kInitialPrerollFrameCount = 8;
 const int kNonInitialPrerollFrameCount = 1;
-
-const int kSeekingPrerollPendingWorkSizeInTunnelMode =
-    16 + kInitialPrerollFrameCount;
+const int kTunnelModePrerollFrameCount = 1;
```
kjyoun, line 245: can you add comment about why you chose 1?

It will help future developer understand the context.

kjyoun (approved):
LGTM % comment.

I am now really think that we had better separate video decoder(e.g. TunnelModeVideoDecoder) to tunnel mode, instead of making `VideoDecoder` support both modes of non-tunnel-mode and tunnel-mode. WDYT?

jasonzhangxx:
> LGTM % comment.
> 
> I am now really think that we had better separate video decoder(e.g. TunnelModeVideoDecoder) to tunnel mode, instead of making `VideoDecoder` support both modes of non-tunnel-mode and tunnel-mode. WDYT?

When working on the TM2 prototype, I used separate audio/video renderers. Tunnel and non-tunnel still share a significant amount of code. So, my understanding is that we can have better abstraction for decoders and MediaCodec-related code to simplify the codebase and improve its overall cleanliness. We can discuss it later, but let's not start it on 26.android.

starboard/android/shared/video_decoder.cc
jasonzhangxx: Done

PR #9621, opened 2026-03-18
android: exit Cobalt upon StarboardBridge.java::requestStop()

tasunnn:
android: Exit Cobalt when requestStop called

Ensure Cobalt explicitly exits the Android process when StarboardBridge.requestStop() is invoked. Previously, the application process only suspends to the background since the `applicationStopped` flag is unset. So this change sets the corresponding boolean values by calling `applicationStopping` to indicate application shutdown and adds `System.exit()` to guarantee a full application exit.

Bug: 470853890


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java
```diff
@@ -295,6 +295,7 @@ public void requestStop(int errorLevel) {
     if (currentActivity != null) {
       currentActivity.finishAndRemoveTask();
```
johnxwork, line 297: FYI The purpose of this line is for a graceful shutdown, which didn't seem to work properly according to your local debugging.

tasunnn: yeah locally the app didn't shut down but instead was suspended in background

tasunnn: actually, I just took a closer look at what the line is doing. I think it does trigger lifecycle to shut down (onPause -> onStop -> onDestroy), but it's not passing this [check](https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java;l=205?q=starboardBr&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F26.android) in onDestroy to exit app since requestStop() does not trigger [afterStopped()](https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java;l=259?q=starboardBr&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F26.android). Theoretically, if we set applicationStopped = true in requestStop(), it could be a graceful shutdown.

johnxwork: That's an amazing find. Do you want to do applicationStopping() to flip the other boolean `applicationStarted` as well?

johnxwork (approved):

PR #9636, opened 2026-03-19
android: Add InputFrameRateEstimator

jasonzhangxx:
Introduce a InputFrameRateEstimator for Android tunneled
playback. The new estimator uses input timestamps (decoding order)
and a sliding window to determine frame rate, as output timestamps
are unreliable in tunneled mode. This enables accurate frame rate
estimation when output timing is managed by the display subsystem.

Refactor the base FrameRateEstimator to support both input and
output timestamp tracking. Dynamically select the appropriate
estimator in MediaCodecBridge based on tunneling state.

Propagate playback rate changes from the video renderer down to the
MediaCodecBridge to ensure the Android MediaCodec is configured with
the correct operating rate during tunneled playback. Set realtime
priority for the MediaCodec when using tunneled playback. These
changes are vital for smooth playback and resource management.

Bug: 487397946


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -132,6 +136,88 @@ public void onNewFrame(long presentationTimeUs) {
     }
   }
 
+  // This estimator is designed to use input timestamps which are in decoding order.
+  // Once it's verfied to work, we can remove |FrameRateEstimator| and always use this
+  // estimator.
+  private class TunnelFrameRateEstimator extends FrameRateEstimator {
```
kjyoun: I think you should define `FrameRateEstimator` as interface and make this method implement it.


Inheriting concrete class is not recommended[1]. In this case, I think that there is really no stong reason of inheriting concrete class.

This class override most meaningful methods of `FrameRateEstimator`,
Basically, it use `FrameRateEstimator` as a interface

---
[1]
- https://share.google/aimode/OekagItBjburXySc9
- [go/totw/labs/concrete-abstractions](https://goto.google.com/totw/labs/concrete-abstractions)
- [go/code-health-prefer-composition](https://goto.google.com/code-health-prefer-composition)

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -132,6 +136,88 @@ public void onNewFrame(long presentationTimeUs) {
...
+    private static final int WINDOW_SIZE = 32;
```
kjyoun: How about this class into separate file?

MediaCodecBridge.java is already 1000+ line long.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -94,13 +95,14 @@ public static final class MimeTypes {
     public static final String VIDEO_AV1 = "video/av01";
   }
 
+  // This estimator is designed to use output timestamps which are in presentation order.
   private class FrameRateEstimator {
```
kjyoun: I found starboard/shared/starboard/player/filter/video_frame_rate_estimator.h and this FrameRateEstimator here (and we are adding another Frame estimator).

It would make hard to figure out how/where frame rate is calculated.

I understand there can be different way to calculate frame rate, so that we may need multiple impl.
But, can't we make such a decision in a single place ?

jasonzhangxx: The video_frame_rate_estimator.h is used in c++ and based on output frames, so it's not suitable for tunnel mode player. There's b/466170081 for consolidating multiple fps estimator implementations. I believe we need further discussion for the future plan. And this PR would not address that issue.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: I renamed `TunnelFrameRateEstimator` to `InputFrameRateEstimator`, as the ultimate goal is to use it to replace `FrameRateEstimator`. After `TunnelFrameRateEstimator` is verified, `FrameRateEstimator` will be removed, so I don't think the design tips are applicable here.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: I added an interface when moving the code to a separated file.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecFrameRateEstimator.java
```diff
@@ -0,0 +1,169 @@
+// Copyright 2017 The Cobalt Authors. All Rights Reserved.
```
borongc: 2026?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -614,6 +574,10 @@ public static void createVideoMediaCodecBridge(
       mediaFormat.setFeatureEnabled(CodecCapabilities.FEATURE_TunneledPlayback, true);
       mediaFormat.setInteger(MediaFormat.KEY_AUDIO_SESSION_ID, tunnelModeAudioSessionId);
       Log.d(TAG, "Enabled tunnel mode playback on audio session " + tunnelModeAudioSessionId);
+
+      // TODO: KEY_PRIORITY might be also needed for non tunnel playback.
```
borongc: Should we open a ticket to track this? And maybe track the conversation that why |KEY_PRIORITY| is needed somewhere in the bug, and link it here.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -455,13 +418,9 @@ public void onOutputBufferAvailable(
...
                       info.size);
-              if (mFrameRateEstimator != null) {
-                mFrameRateEstimator.onNewFrame(info.presentationTimeUs);
-                int fps = mFrameRateEstimator.getEstimatedFrameRate();
-                if (fps != FrameRateEstimator.INVALID_FRAME_RATE && mFps != fps) {
-                  mFps = fps;
-                  updateOperatingRate();
-                }
+              if (!mIsTunnelingPlayback && mFrameRateEstimator != null) {
+                mFrameRateEstimator.onNewOutput(info.presentationTimeUs);
```
borongc, line 422: Should we add a comment here and for tunnel mode as well below to mention why this use output as the estimator, and why tunnel mode uses intput as the estimator?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -94,44 +95,6 @@ public static final class MimeTypes {
     public static final String VIDEO_AV1 = "video/av01";
   }
 
-  private class FrameRateEstimator {
```
borongc, line 97: To better to check what is changed in this file, maybe have a separate PR to move `FrameRateEstimator` to `MediaCodecFrameRateEstimator`, and then add `onNewOutput()` for tunnel mode after that PR.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecFrameRateEstimator.java
jasonzhangxx: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: The change is not too large, let's keep it in one PR.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: A comment is added in queue input functions. Once the new frame estimator is verified to work, the old estimator will be removed. And I'll clean the estimator code at that time.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -895,6 +871,12 @@ private int queueInputBuffer(
       Log.e(TAG, "Failed to queue input buffer", e);
       return MediaCodecStatus.ERROR;
     }
+
+    // Avoid update frame rate for inputs when tunnel mode is not enabled for better performance.
```
borongc, line 875: Consider to use a helper function for both L875--L879 and L941--L946.

starboard/shared/starboard/player/filter/filter_based_player_worker_handler.cc
```diff
@@ -371,6 +371,9 @@ HandlerResult FilterBasedPlayerWorkerHandler::SetPlaybackRate(
   }
 
   media_time_provider_->SetPlaybackRate(playback_rate_);
+  if (video_renderer_) {
```
borongc, line 374: I saw for non-tunnel mode playback, video playback rate is updated during VideoRenderAlgorithm::Render(), is there any reason we cannot do the same thing for tunnel mode playbacks?

jasonzhangxx: Tunnel mode player doesn't have output frames and VideoRenderAlgorithm::Render() is only called at prerolling and end of stream.

Once the new way to set playback rate are verified, the old way (via VideoRenderAlgorithm) can be removed.

borongc: Right, but this change also impacts for non-tunnel playbacks. Just want to know what's the risk and impact for this change.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
jasonzhangxx: Will do that in the cleanup change of removing old estimator. Right now it needs two helper functions, one for onInput and one for onOutput. I don't think that could make the code shorter.

borongc (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
kjyoun: Thanks for moving FrameEstimator into separate class.
The code becomes more readable.

PR #9637, opened 2026-03-20
Crashpad launch with android system Linker Loader (API 29+)

alunev:
Implement a storage-efficient loading mechanism for crashpad_handler on modern Android TV devices (API 29+).
This fixes crashpad handler launch on Androids that do not extract native libs to filesystem and can launch handler from uncompressed apk binary. Before it was failing since hang callback expected the handler binary to be extracted to .../lib/arm/

- Leverages /system/bin/linker to launch uncompressed binaries directly from APK offsets.
- Isolated via BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROIDTV).
- Verified zero-extraction and successful capture on Sabrina hardware.

Issue: 484122527

### DIFF
--- a/components/crash/core/app/BUILD.gn
+++ b/components/crash/core/app/BUILD.gn
@@ -232,6 +232,8 @@ if (is_mac || is_android || is_linux || is_chromeos) {
     ]
 
     if (is_cobalt && is_androidtv) {
+      # This is change is for Cobalt on AndroidTV. It enables crashpad handler
+      # launch with android linker and with java crashpad handler.
       ldflags = [ "-Wl,--export-dynamic" ]
     }
 


components/crash/core/app/crashpad_android.cc
```diff
@@ -316,7 +316,13 @@ bool GetHandlerTrampoline(std::string* handler_trampoline,
 
   std::string local_handler_trampoline(local_handler_library, 0,
                                        libdir_end + 1);
+
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROIDTV)
+  local_handler_trampoline += "libchrome_crashpad_handler.so";
```
johnxwork, line 325: The library name is different between Android and Linux?

components/crash/core/app/BUILD.gn
```diff
@@ -230,6 +230,10 @@ if (is_mac || is_android || is_linux || is_chromeos) {
       "//third_party/crashpad/crashpad/handler",
     ]
 
+    if (is_android) {
+      ldflags = [ "-Wl,--export-dynamic" ]
```
johnxwork, line 237: Could you measure if `export-dynamic` increases lib size by any bite?

components/crash/core/app/BUILD.gn
```diff
@@ -230,6 +230,10 @@ if (is_mac || is_android || is_linux || is_chromeos) {
       "//third_party/crashpad/crashpad/handler",
     ]
 
+    if (is_android) {
```
hlwarriner: I think we should guard this customization with `is_cobalt &&  is_androidtv`.

This would
* be more consistent with guards in the C++ source code
* make it clear this is a cobalt customization
* exclude 3P AOSP; we may want to eventually follow a similar approach for 3P AOSP but this PR probably shouldn't partially make that change.

components/crash/core/app/crashpad_android.cc
hlwarriner, line 325: Is `libchrome_crashpad_handler.so` actually a "trampoline"? Or is it a "standalone executable containing all of the crash dumping code" (from https://chromium.googlesource.com/chromium/src/+/main/docs/android_native_libraries.md#crashpad-packaging)?

third_party/crashpad/crashpad/client/crashpad_client_linux.cc
```diff
@@ -115,8 +115,12 @@ std::vector<std::string> BuildArgsToLaunchWithLinker(
   } else {
     argv.push_back("/system/bin/linker");
   }
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROIDTV)
+  argv.push_back(handler_trampoline);
+#else
   argv.push_back(handler_trampoline);
   argv.push_back(handler_library);
+#endif
```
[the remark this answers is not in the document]
alunev: I think I'll keep it verbose for readability and easy rebase.

alunev:
> Re b/484122527#comment14, does this PR implement just one of the `Android Q and above` or `Android N to P` approaches, or both?

This is Q+ only for now. Java fallback to follow. But it doesn't break [L through M](https://chromium.googlesource.com/chromium/src/+/main/docs/android_native_libraries.md#crashpad-packaging) - those are running extracted binary and are ok now. And it doesn't break L through P (Java fallback) - those are broken already and remain broken till java fallback is fixed.

components/crash/core/app/BUILD.gn
alunev: thanks, changed condition

components/crash/core/app/crashpad_android.cc
alunev: in this case libchrome_crashpad_handler.so is a "standalone executable containing all of the crash dumping code". Basically we don't need trampoline since our handler is standalone and complete (still jus ~1M). The var is still `local_handler_trampoline` to compromise with rebasability later. Added comment to clarify.

hlwarriner:
> > Re b/484122527#comment14, does this PR implement just one of the `Android Q and above` or `Android N to P` approaches, or both?
> 
> This is Q+ only for now. Java fallback to follow. But it doesn't break [L through M](https://chromium.googlesource.com/chromium/src/+/main/docs/android_native_libraries.md#crashpad-packaging) - those are running extracted binary and are ok now. And it doesn't break L through P (Java fallback) - those are broken already and remain broken till java fallback is fixed.

Sounds good. Thanks for explaining!

components/crash/core/app/crashpad_android.cc
alunev: I've added comment to clarify, it's actually a different lib: `libchrome_crashpad_handler.so` is full handler that can run standalone. `libcrashpad_handler_trampoline.so` is a 1kb launcher that is designed to load crashpad handler from the main chromium lib when crashpad handler is linked into it (in Cobalt it isn't).

components/crash/core/app/BUILD.gn
```diff
@@ -230,6 +231,10 @@ if (is_mac || is_android || is_linux || is_chromeos) {
       "//third_party/crashpad/crashpad/handler",
     ]
 
+    if (is_cobalt && is_androidtv) {
+      ldflags = [ "-Wl,--export-dynamic" ]
```
hlwarriner, line 237: Please add a comment explaining why this is needed for Cobalt's AndroidTV but not upstream Android platforms.

components/crash/core/app/crashpad_android.cc
hlwarriner: Ok. I get the rebasability benefit but I feel there's a significant readability cost to calling it a trampoline when it really isn't one. I think it's ok for now - clearly there are existing confusing naming conventions used in the upstream code, like naming the executable as a `.so` - but can you please add a todo comment (format: `TODO: b/484122527 - Cobalt: ...`) to consider refactoring this code in the future to address the trampoline naming? An ideal long term outcome might look like adding another packaging option to https://chromium.googlesource.com/chromium/src/+/main/docs/android_native_libraries.md#crashpad-packaging and upstreaming it.

components/crash/core/app/BUILD.gn
alunev: it's about 10kb

 | Metric | With `export-dynamic` | Without `export-dynamic` | Delta |                                                                             
| :--- | :--- | :--- | :--- |
| **Dynamic Symbols** | 304 | 218 | **+86 symbols** |                                                                                             │
| **Binary Size (Bytes)** | 1,113,196 | 1,102,556 | **+10,640 bytes (~10 KB)** |

components/crash/core/app/crashpad_android.cc
alunev: resolving based on lgtms and discussion in comments

components/crash/core/app/BUILD.gn
alunev: resolving based on lgtms and discussion in comments

components/crash/core/app/BUILD.gn
alunev: resolving based on lgtms and discussion in comments

components/crash/core/app/crashpad_android.cc
alunev: added todo and bug, thanks.

components/crash/core/app/BUILD.gn
alunev: thx, added

PR #9641, opened 2026-03-20
tvOS: Initialize Application for browser tests

Gyuyoung:
On tvOS, the H5vccSchemeURLLoaderFactory browser test invokes
Application::Get() internally. Without an initialized Application
instance during browser test setup, this leads to an assertion failure.
This change creates a dummy Application instance for testing
specifically on tvOS, preventing the crash and allowing the test
to execute correctly.

Bug: 489556129

### DIFF
--- a/cobalt/testing/browser_tests/content_test_launcher.cc
+++ b/cobalt/testing/browser_tests/content_test_launcher.cc
@@ -30,6 +30,11 @@
 #include "ui/base/buildflags.h"
 #include "ui/base/ui_base_switches.h"
 
+#if BUILDFLAG(IS_IOS_TVOS)
+#include "starboard/common/command_line.h"             // nogncheck
+#include "starboard/tvos/shared/application_darwin.h"  // nogncheck
+#endif
+
 namespace content {
 
 class ContentBrowserTestSuite : public ContentTestSuiteBase {
@@ -52,7 +57,21 @@ class ContentBrowserTestSuite : public ContentTestSuiteBase {
 #if BUILDFLAG(IS_ANDROID)
     RegisterInProcessThreads();
 #endif
+
+#if BUILDFLAG(IS_IOS_TVOS)
+    // Some tests indirectly invoke starboard::Application::Get(), and on tvOS
+    // the Application instance needs to be created manually from the Cobalt
+    // layer.
+    starboard_application_ = std::make_unique<starboard::ApplicationDarwin>(
+        std::make_unique<starboard::CommandLine>(
+            base::CommandLine::ForCurrentProcess()->argv()));
+#endif
   }
+
+#if BUILDFLAG(IS_IOS_TVOS)
+ private:
+  std::unique_ptr<starboard::ApplicationDarwin> starboard_application_;
+#endif
 };
 
 class ContentTestLauncherDelegate : public TestLauncherDelegate {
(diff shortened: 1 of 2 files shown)


Gyuyoung:

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -39,6 +44,13 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
   ~H5vccSchemeURLLoaderFactoryBrowserTest() override = default;
 
   void SetUp() override {
+#if BUILDFLAG(IS_IOS_TVOS)
+    // On tvOS, some tests require an Application instance. For testing
+    // purposes, a dummy application is created here.
+    auto starboard_application = std::make_unique<starboard::ApplicationDarwin>(
+        std::make_unique<starboard::CommandLine>(0, nullptr));
```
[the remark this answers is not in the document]
Gyuyoung: Done.

Gyuyoung:
> LGTM % @rakuco comment. We should just add this to init in `ContentBrowserTestSuite::Initialize` in `cobalt/testing/browser_tests/content_test_launcher.cc`.

yes, I also think it's better to move this there. Fixed. Thanks.

cobalt/testing/browser_tests/content_test_launcher.cc
```diff
@@ -52,7 +57,19 @@ class ContentBrowserTestSuite : public ContentTestSuiteBase {
 #if BUILDFLAG(IS_ANDROID)
     RegisterInProcessThreads();
 #endif
+
+#if BUILDFLAG(IS_IOS_TVOS)
+    // On tvOS, some tests require an Application instance. For testing
+    // purposes, a dummy application is created here.
+    starboard_application_ = std::make_unique<starboard::ApplicationDarwin>(
+        std::make_unique<starboard::CommandLine>(0, nullptr));
```
rakuco: Even though this `starboard::CommandLine` instance is not really used for a lot on tvOS, I think it makes sense to initialize it with actual arguments. Since this is in `cobalt/`, you could do something like

```c++
std::make_unique<starboard::CommandLine>(
    base::CommandLine::ForCurrentProcess()->argv())
```

cobalt/testing/browser_tests/content_test_launcher.cc
```diff
@@ -52,7 +57,19 @@ class ContentBrowserTestSuite : public ContentTestSuiteBase {
 #if BUILDFLAG(IS_ANDROID)
     RegisterInProcessThreads();
 #endif
+
+#if BUILDFLAG(IS_IOS_TVOS)
+    // On tvOS, some tests require an Application instance. For testing
+    // purposes, a dummy application is created here.
```
rakuco:
```suggestion
    // Some tests indirectly invoke starboard::Application::Get(), and on tvOS the Application instance needs to be created manually from the Cobalt layer.
```

Gyuyoung: Done.

cobalt/testing/browser_tests/content_test_launcher.cc
Gyuyoung: ok, it looks better. Fixed.

Gyuyoung:
@fayev-g Could you please take a look at this?

PR #9642, opened 2026-03-20
Rename {is => use}_starboard GN variable (1/2)

hferreiro:
is_starboard is an inaccurate name, because it denotes whether the build uses
starboard, i.e. should it call SbFoo or should it call a system function
instead. Rename to use_starboard.

Bug: 494478521


johnxwork (approved):

andrewsavage1:
This is actually a point I brought up over a year ago and had a design discussion with the team on it. We chose the is_starboard pattern from that discussion because we treat Starboard as a full platform, in the same way you would use is_linux. I definitely agree it's strange in some places and the fact that it overlaps with is_linux currently is less than ideal, but this probably isn't something we should change right now.

PR #9670, opened 2026-03-20
Fix Chromium sync workflow: move LASTCHANGE.chromium update to main s…

andrewsavage1:
…ync stage

Bug: 484467425


andrewsavage1:
note this is untested

PR #9674, opened 2026-03-21
Add perf tests splash browser tests

linxinan-yt:
cobalt: Add splash screen performance tests

Measure content fetch time in H5vcc scheme browser tests.
This change instruments the H5vccSchemeURLLoaderFactoryCacheBrowserTest
to record the duration of fetching cached content. Performance metrics
are reported via Test::RecordProperty and printed to stdout, enabling
automated performance monitoring and regression detection for the
splash screen's content loading.

Perf results are stored in the xml file like:

```
<?xml version='1.0' encoding='UTF-8'?>
<testsuites tests="1" failures="0" disabled="0" errors="0" time="0" name="AllTests"><testsuite name="H5vccSchemeURLLoaderFactoryCacheBrowserTest" tests="1" failures="0" disabled="0" skipped="0" errors="0" time="0.287" timestamp="2026-03-25T12:16:30.345">
    <testcase name="LoadSplashVideoFromDefaultCache" value_param="" type_param="" file="../../cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc" line="372" status="run" result="completed" time="0.287" timestamp="2026-03-25T12:16:30.345" classname="H5vccSchemeURLLoaderFactoryCacheBrowserTest">
      <properties>
        <property name="CacheReadTime_ms" value="9" />
        <property name="CacheReadTime_Trace" value="default" />
      </properties>
    </testcase>
  </testsuite>
</testsuites>
```

Bug: 489501943

### DIFF
--- a/cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
+++ b/cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
@@ -51,6 +51,7 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
     if (shell()) {
       EXPECT_TRUE(NavigateToURL(shell(), GURL("about:blank")));
     }
+    H5vccSchemeURLLoaderFactory::SetResourceMapForTesting(nullptr);
     ContentBrowserTest::TearDownOnMainThread();
   }
 
@@ -135,8 +136,10 @@ class H5vccSchemeURLLoaderFactoryBrowserTest : public ContentBrowserTest {
                               is_4k_supported ? "true" : "false");
   }
 
+ protected:
+  std::unique_ptr<GeneratedResourceMap> test_resource_map_;
+
  private:
-  std::unique_ptr<GeneratedResourceMap> test_resource_map;
   base::test::ScopedFeatureList scoped_feature_list_;
 };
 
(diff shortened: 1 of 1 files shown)


cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
```diff
@@ -241,7 +244,25 @@ class H5vccSchemeURLLoaderFactoryCacheBrowserTest
...
+        testing::Test::RecordProperty("CacheReadTime_ms",
+                                      base::NumberToString(time_ms));
+        testing::Test::RecordProperty("CacheReadTime_Trace", cache_name);
+        // Keep printing to stdout for legacy perf parsers as well
+        perf_test::PrintResult("CacheReadTime", "", cache_name, time_ms, "ms",
+                               true);
+      } else {
+        ADD_FAILURE() << "Failed to parse time from result: " << result;
+      }
+    } else {
```
Awallky, line 265: Do we want this in an `else` block? Don't we always want to check the test results?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
Awallky, line 272: Unrelated to this change, but I think this would be better suited inside of a test SetUp/TearDown method.

Awallky:
@linxinan-chops Could you share a sample printout of the added perf logging?

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Nice catch. Done.

cobalt/shell/browser/h5vcc_scheme_url_loader_factory_browsertest.cc
linxinan-yt: Make sense. I need to promote test_resource_map a class member from a local variable then.

linxinan-yt:
> @linxinan-chops Could you share a sample printout of the added perf logging?

Attached in the PR description.

PR #9686, opened 2026-03-23
Disable TTS for executable(not plugin) RDK cobalt builds

niranjanyardi:
The TTS fix (b/491841704) broke the RDK executable build.
As a temporary workaround, this change disables speech synthesis
support when building as an executable.

Plugin builds continue to use the existing implementation.

Bug: 491841704
Bug: 495477543

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/speech/speech_synthesis_is_supported_executable.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/speech/speech_synthesis_is_supported_executable.cc
@@ -14,7 +14,7 @@
 
 #include "starboard/speech_synthesis.h"
 
-// TODO: (b/495477543) Fix speech synthesis for executable type.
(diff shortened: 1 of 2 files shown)


starboard/contrib/rdk/src/third_party/starboard/rdk/shared/speech/speech_synthesis_is_supported_executable.cc
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/speech_synthesis.h"
+
+// TODO: (b/495477543) Fix speech synthesis for executable type.
```
andrewsavage1: Follow go/todo-format here and below

PR #9691, opened 2026-03-23
android: Prevent use of media_codec_bridge_ in callbacks

jasonzhangxx:
Avoid a potential race condition during MediaDecoder destruction. Remove the 
usage of `media_codec_bridge_` in `OnMediaCodecOutputFormatChanged`,
which  was only used for logging purposes. This prevents a crash if the callback
is invoked after the bridge has been destroyed.

Bug: 495444400


starboard/android/shared/media_decoder.cc
borongc, line 739: If we need this log message, we may use `if (media_codec_bridge_)`.

borongc (approved):

jasonzhangxx:
Googler helped to verify the fix could resolve the crash (b/495444400#comment9).

johnxwork (approved):

PR #9697, opened 2026-03-24
fix: Guard ANGLE-extension DCHECKs behind Cobalt/Starboard build flags

denisyuji:
Wrap the ANGLE-extension-absent DCHECKs in `GLContextEGL::InitializeImpl` with `#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) && !BUILDFLAG(IS_STARBOARD)` guards.

On platforms using native EGL without ANGLE (such as RDK), extensions like `EGL_ANGLE_create_context_bind_generates_resource` and `EGL_ANGLE_robust_resource_initialization` are not available. Without these guards, the DCHECKs fire fatally during GL context creation, preventing Cobalt from running on such platforms.

Bug: 486053854

### DIFF
--- a/ui/gl/gl_context_egl.cc
+++ b/ui/gl/gl_context_egl.cc
@@ -228,7 +228,10 @@ bool GLContextEGL::InitializeImpl(GLSurface* compatible_surface,
     context_attributes.push_back(attribs.bind_generates_resource ? EGL_TRUE
                                                                  : EGL_FALSE);
   } else {
+    // TODO: b/486053854 - Cobalt: Reset unsupported attribs to EGL defaults
+  #if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
     DCHECK(attribs.bind_generates_resource);
+  #endif
   }
 
   if (gl_display_->ext->b_EGL_ANGLE_create_context_webgl_compatibility) {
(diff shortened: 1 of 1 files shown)


ui/gl/gl_context_egl.cc
```diff
@@ -228,15 +228,19 @@ bool GLContextEGL::InitializeImpl(GLSurface* compatible_surface,
     context_attributes.push_back(attribs.bind_generates_resource ? EGL_TRUE
                                                                  : EGL_FALSE);
   } else {
+  #if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) && !BUILDFLAG(IS_STARBOARD)
```
niranjanyardi: just #if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) is sufficeint - no need to add && !BUILDFLAG(IS_STARBOARD)

Also, can you add a TODO above the change following this format:
// TODO: b/87654321 - Cobalt: fix foo

The bug should contain all the research and the unmerged PR links

denisyuji: Thank you for the review, @niranjanyardi . I've made the changes above and I created b/486053854

niranjanyardi (approved):

sherryzy (approved):

PR #9700, opened 2026-03-24
Cobalt: Backport from PR #8130: Port storage flush calls to AppEventRunnerImpl lifecycle

jellefoks:
Ensures that cookies and LocalStorage are explicitly flushed when the application transitions to the 'Blurred' (Paused) or 'Frozen' states. This is critical for data persistence on platforms where the OS may kill the application shortly after these transitions.

Issue: 371433332
Bug: 451215656
Bug: 450936106
Bug: 444451968

### DIFF
--- a/cobalt/app/app_event_runner.cc
+++ b/cobalt/app/app_event_runner.cc
@@ -23,11 +23,13 @@
 #include "base/allocator/partition_allocator/memory_reclaimer.h"
 #include "base/at_exit.h"
 #include "base/command_line.h"
+#include "base/functional/callback_helpers.h"
 #include "base/logging.h"
 #include "base/memory/memory_pressure_listener.h"
 #include "base/no_destructor.h"
 #include "build/build_config.h"
 #include "cobalt/app/app_event_delegate.h"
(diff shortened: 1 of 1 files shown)


cobalt/app/app_event_runner.cc
```diff
@@ -134,6 +136,12 @@ class AppEventRunnerImpl : public AppEventRunner {
   void DoBlur() override {
     content::Shell::OnBlur();
 #if BUILDFLAG(IS_STARBOARD)
+    {
+      auto* client = cobalt::CobaltContentBrowserClient::Get();
+      if (client) {
+        client->FlushCookiesAndLocalStorage(base::DoNothing());
```
andrewsavage1, line 141: Is this specific to Starboard platforms? In main it looks like it happens for all platforms

andrewsavage1 (approved):

cobalt/app/app_event_runner.cc
jellefoks: Right. Even the `HandleFocusEvent()` under the buildflag here is suspicious and should be revisited for main. I don't think Android matters very much for this branch though.

PR #9706, opened 2026-03-24
cobalt: Set migrated cookie source type to kOther

jkim-julie:
This sets migrated cookie source type to
`net::CookieSourceType::kOther` on storage migration.
Before this change, it’s set to `net::CookieSourceType::kUnknown`
used for testing and it caused CHECK failure.

Bug: 495656832


jkim-julie:

rakuco:
cc'ing @andrewsavage1, although I'm not sure who'd be the best reviewer for this patch

Just curious about this:

> Before this change, it’s set to `net::CookieSourceType::kUnknown`
> used for testing and it caused CHECK failure.

I wonder why it's not failing on other platforms since this is platform-independent code.

jkim-julie:
> Just curious about this:
> 
> > Before this change, it’s set to `net::CookieSourceType::kUnknown`
> > used for testing and it caused CHECK failure.
> 
> I wonder why it's not failing on other platforms since this is platform-independent code.

In order to test data migration, I installed C25 and loaded the app in order to save some data such as cookie or localstorage. And then, I installed Chrobalt without removing C25 to keep the previous data. With this step, the data exists in the app storage and Chrobalt reads it and migrate it to Chromium style.

Since it's not a specific code to tvOS, I think if other platforms also have the same steps for migration, it can occur.

rakuco:
I see. The change looks OK to me (so informal lgtm), I'd just like someone more familiar with the area (and who tested this on other platforms) to confirm if this fixes a crash affecting all ports.

(cc @Awallky who may know more as well)

Awallky:
`kOther` seems to be the most appropriate value for this field.

**Note:** The `CHECK_IS_FOR_TEST` in `cookie_monster.cc` was just removed today - see [this CL](https://chromium-review.googlesource.com/c/chromium/src/+/7686830). While it won't be a concern after rebasing to M148, it will be a good fix until this change is cherry-picked to Cobalt. 

LGTM ,assuming you were able to verify that Cobalt still works with some local tests.

Awallky:
Also, I see this was not added to the current release branch. So, this is a future-looking fix for the next release and beyond.

rakuco:
> Also, I see this was not added to the current release branch. So, this is a future-looking fix for the next release and beyond.

BTW, should it target the current release branch as well? If this one lands, a cherry-pick PR will be created automatically, but it's still unclear if we need to act on it or not. I guess this makes sense as a 27.lts fix since that crash could be triggered on other platforms?

Awallky:
> BTW, should it target the current release branch as well? If this one lands, a cherry-pick PR will be created automatically, but it's still unclear if we need to act on it or not. I guess this makes sense as a 27.lts fix since that crash could be triggered on other platforms?

I do not think we need this for Cobalt 26. I checked and this entire enum was introduced after M114.  We definitely need it for the next release.

rakuco:
Sorry, I got confused. I thought 27.lts was the current release branch, but that's actually the next release branch IIUC.

I think someone just needs to enable auto-merge here then.

PR #9721, opened 2026-03-24
[tvOS] Add simple_packager.py for tvOS

fayev-g:
Add working packager from C25 to use for main branch tvOS Kokoro pipeline. 

Bug: 495840519


oxve:
Approving PR, I think this should be migrated to use the new packager as soon as possible though.

PR #9727, opened 2026-03-24
[POSIX] Support getuid

sacuff:
[POSIX] Support getuid

getuid has a few uses that are probably not needed for Cobalt. However,
it is a very simple function and we already support the similar geteuid.

Like geteuid, the standard says getuid must never fail, so this adds
just one test to check that it returns any value.

Bug: 495875401

### DIFF
--- a/starboard/elf_loader/exported_symbols.cc
+++ b/starboard/elf_loader/exported_symbols.cc
@@ -332,6 +332,7 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_WRAPPER(geteuid);
   REGISTER_WRAPPER(getifaddrs);
   REGISTER_WRAPPER(getpid);
+  REGISTER_WRAPPER(getuid);
   REGISTER_WRAPPER(getpriority);
   REGISTER_WRAPPER(getrlimit);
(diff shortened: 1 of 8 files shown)


cobalt/common/libc/unistd/getuid.cc
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <unistd.h>
+
+extern "C" {
+uid_t getuid(void) {
+  // A non-zero UID to return for the stub implementation of getuid(). The value
+  // itself is not important, as long as it is not 0 (root).
+  return 1;
```
andrewsavage1: We should probably add either a `NOTIMEPLEMENTED` (from //base/notimplemented.h) or `NOTREACHED` (from //base/notreached.h) here—the former just logs, the second crashes if it is called.

andrewsavage1 (commented):
Were we able to check if this function is ever called by adding a crash in it? And we checked that it was fairly difficult to remove it from the build?

sacuff:
Switched to actually supporting the syscall

andrewsavage1 (approved):

PR #9729, opened 2026-03-24
Log as Info for splash screen cache miss

tasunnn:
cobalt: Log splash screen built-in as info

Previously, when the splash screen cache missed and the built-in
splash screen was used as a fallback, an ERROR log was generated.
This state is a valid fallback mechanism and not an error
condition. Adjust the log level to INFO for this scenario to avoid
misleading error reports.

Bug: 490510288


cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
linxinan-yt: Since you touch the logging, I think we should have better wording here. Instead of "Failed...", make it more neutral? E.g. Did not find splash video... WDYT?

linxinan-chops (approved):
LGTM % comment

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
```diff
@@ -377,7 +377,11 @@ class H5vccSchemeURLLoader : public network::mojom::URLLoader {
 
   void DisconnectCacheAndSendFallback(const std::string& message,
                                       SplashScreenFetchedState state) {
-    LOG(ERROR) << message;
+    if (state == SplashScreenFetchedState::kOkBuiltIn) {
```
johnxwork, line 380: Do you mean to use `!=`?

Also if the message is generic enough and don't contain personal data, we can consider LOG(INFO) unconditionally.

johnxwork (approved):

cobalt/shell/browser/h5vcc_scheme_url_loader_factory.cc
tasunnn: kOkBuiltIn is the case for cache miss(not on error) but we still want to know why splash is not fetched from cache so this goes to INFO. But for other cases they fail to find in cache due to unexpected errors so we want them to print as ERROR

PR #9730, opened 2026-03-24
ci: Fix max output size error in gemini commit message workflow

oxve:
This fixes the 'Argument list too long' error caused by overflow in the GH output.

Bug: 465494923


zhongqiliang:
Did you test it?

zhongqiliang (approved):

PR #9737, opened 2026-03-25
build: tune visibility of //base for starboard

hferreiro:
Make sure that builds that use starboard don't use //base when building
with the starboard toolchain by changing the visibility of //base.

Bug: 495203133


hferreiro:
This fails because visibility is checked when loading the BUILD.gn file, even if there are no dependencies to //base. An alternative would be to use a source_set that pointed to a non-existing file but I'll close this in favor of a redirection to //copied_base/base, that fixes build issues in AOSP Arm with the minimal changes in upstream code.

PR #9746, opened 2026-03-25
[POSIX] Add wrappers for if_indextoname and if_nametoindex

sacuff:
Adds Starboard wrappers for `if_indextoname` and `if_nametoindex`.

A wrapper is required because of the possibility that `IF_NAMESIZE` differs between platforms, and therefore the `ifname` buffer could overflow if unchecked.

Bug: 495906870


andrewsavage1:
It looks like this would be pretty easy to support (exists in the posix spec, parameters aren't that hard to wrap)—should we do that instead?

sacuff:
Updated to register the symbol instead of removing it

andrewsavage1:
Make sure to update the PR description to reflect that we're adding this function, not removing its uses

starboard/nplb/posix_compliance/posix_if_indextoname_test.cc
```diff
@@ -0,0 +1,60 @@
...
+
+#include <errno.h>
+#include <net/if.h>
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace nplb {
+namespace {
+
+TEST(PosixIfIndextonameTest, InvalidIndex) {
```
jellefoks: If we had `if_nametoindex` or `if_nameindex`, a thorough test for this would (cooperating with `getifaddrs()` if no `if_nameindex`),  to get all the names, and then use `if_nametoindex` to get the indices, and then `if_indextoname` on all the indices found. 

It looks like we have neither `if_nametoindex` nor `if_nameindex` though.

But... I see in `ip_endpoint.cc`, which appears to be the only place where `if_indextoname` is used, that `if_nametoindex` is used there... but it does not show up in the leak detector so possibly dead code elimination in the linker removes it before it leaks.  

I think that we probably need to also add `if_nametoindex` to Starboard.

starboard/nplb/posix_compliance/posix_if_indextoname_test.cc
```diff
@@ -0,0 +1,60 @@
...
+  char* name = if_indextoname(0xFFFFFFFF, buf);
+  EXPECT_EQ(name, nullptr);
+  EXPECT_EQ(errno, ENXIO);
+}
+
+TEST(PosixIfIndextonameTest, ValidIndexLoopback) {
+  char buf[IF_NAMESIZE];
+  // Loopback is named "lo" on almost all Linux systems and "lo0" on Mac.
+  const char* loopback_name = "lo";
+  unsigned int index = if_nametoindex(loopback_name);
```
jellefoks: We don't have if_nametoindex, do we? These tests run on the Cobalt side, not the platform side, so we can only use what we have there.

How come the leak detector didn't detect a new leak of that symbol from here? Do we not run it against nplb? that would be an oversight. I see that musl implements it by calling ioctl SIOCGIFINDEX on a unix socket, and I don't think we support the former maybe also not even the latter.

sacuff: It shows up if I run the leak detector and specify `-t nplb` but by default it only builds and checks libcobalt. b/477257357 I assume is related

starboard/nplb/posix_compliance/posix_if_indextoname_test.cc
sacuff: Added `if_nametoindex` and a test using `gettifaddrs`

starboard/nplb/posix_compliance/posix_if_nametoindex_test.cc
```diff
@@ -0,0 +1,47 @@
...
+  EXPECT_EQ(index, 0u);
+}
+
+TEST(PosixIfNametoindexTest, EmptyName) {
+  errno = 0;
+  unsigned int index = if_nametoindex("");
+  EXPECT_EQ(index, 0u);
+}
+
+TEST(PosixIfNametoindexTest, ValidNameLoopback) {
```
jellefoks, line 33: I think in one of these we could do a roundtrip test between names and indices, so confirm that the index output and inputs match the same sequencing.

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
```diff
@@ -390,6 +391,36 @@ SB_EXPORT int __abi_wrap_getifaddrs(struct ifaddrs** ifap) {
...
 }
 
+SB_EXPORT unsigned int __abi_wrap_if_nametoindex(const char* ifname) {
+  if (ifname == nullptr) {
+    // No errors defined by the POSIX spec.
+    return 0;
+  }
+
+  char platform_buf[IF_NAMESIZE];
+  starboard::strlcpy(platform_buf, ifname, MUSL_IF_NAMESIZE);
```
jellefoks: Do we still need to do this copy if IF_NAMESIZE >= MUSL_IF_NAMESIZE?

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
```diff
@@ -390,6 +391,36 @@ SB_EXPORT int __abi_wrap_getifaddrs(struct ifaddrs** ifap) {
...
+
+  char platform_buf[IF_NAMESIZE];
+  char* res = if_indextoname(ifindex, platform_buf);
+  if (res == nullptr) {
+    return nullptr;
+  }
+  // If IF_NAMESIZE > MUSL_IF_NAMESIZE, this will truncate the interface name.
+  // At time of writing, that is fine since current uses of this function are in
+  // logging.
+  starboard::strlcpy(ifname, res, MUSL_IF_NAMESIZE);
```
jellefoks, line 436: Same here, we could skip this depending on the value if IF_NAMESIZE relative to MUSL_IF_NAMESIZE.

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
sacuff: Added a check to skip the copy in that case

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
sacuff: Added

starboard/nplb/posix_compliance/posix_if_nametoindex_test.cc
```diff
@@ -0,0 +1,60 @@
...
+#include <errno.h>
+#include <net/if.h>
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace nplb {
+namespace {
+
+TEST(PosixIfNametoindexTest, InvalidName) {
+  errno = 0;
```
andrewsavage1: Is it necessary to set the errno in these tests?

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
```diff
@@ -390,6 +391,48 @@ SB_EXPORT int __abi_wrap_getifaddrs(struct ifaddrs** ifap) {
...
+SB_EXPORT unsigned int __abi_wrap_if_nametoindex(const char* ifname) {
+  if (ifname == nullptr) {
+    // No errors defined by the POSIX spec.
+    return 0;
+  }
+
+  // ifname might be longer than the platform's max interface name length.
+  if (MUSL_IF_NAMESIZE > IF_NAMESIZE) {
+    char platform_buf[IF_NAMESIZE];
+    starboard::strlcpy(platform_buf, ifname, IF_NAMESIZE);
```
andrewsavage1, line 403: Should we have some kind of warning if platform_bug != ifname?

starboard/shared/modular/starboard_layer_posix_socket_abi_wrappers.cc
```diff
@@ -390,6 +391,48 @@ SB_EXPORT int __abi_wrap_getifaddrs(struct ifaddrs** ifap) {
...
+
+  // ifname buffer is larger than the platform's max interface name length,
+  // so it is fine to pass it directly to the platform's if_indextoname.
+  if (MUSL_IF_NAMESIZE >= IF_NAMESIZE) {
+    return if_indextoname(ifindex, ifname);
+  }
+
+  // ifname buffer is smaller than the platform's max interface name length. If
+  // IF_NAMESIZE > MUSL_IF_NAMESIZE, the copy will truncate the interface name.
+  // At time of writing, that is fine since current uses of this function are in
```
andrewsavage1, line 429: Maybe we should still log a warning if the name is truncated

andrewsavage1 (approved):

PR #9748, opened 2026-03-25
[tvOS] Add build_mac.sh Kokoro script with updated Chrobalt paths

fayev-g:
Script builds tvOS cobalt artifacts and packages them for GCS uploads. Update paths and references for Chrobalt structure.

Bug: 495549526


SortaCookie (approved):
Let's get this submitted so we can start testing.

PR #9751, opened 2026-03-25
Draft: Implement Java Fallback for crashpad handler

alunev:
Switch CrashpadMain.java to manual JNI method to avoid GEN_JNI collisions
This involves changes in buld files to branch between annotation-based and standard jni and also changes in the crashpad main java and cc classes.

BUG: 484122527

### DIFF
--- a/base/BUILD.gn
+++ b/base/BUILD.gn
@@ -209,6 +209,7 @@ buildflag_header("cobalt_for_google3_buildflags") {
   header = "cobalt_for_google3_buildflags.h"
   header_dir = "base/android"
   flags = [
+    "IS_COBALT=$is_cobalt",
     "IS_COBALT_ON_GOOGLE3=$is_cobalt_on_google3",
   ]
 }
--- a/components/crash/android/BUILD.gn
+++ b/components/crash/android/BUILD.gn
@@ -3,6 +3,7 @@
 # found in the LICENSE file.
 
 import("//build/config/android/rules.gni")
+import("//starboard/build/buildflags.gni")
 
 _jni_sources = [
   "java/src/org/chromium/components/crash/CrashKeys.java",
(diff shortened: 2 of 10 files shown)


components/crash/android/java/src/org/chromium/components/crash/browser/CrashpadMainCobalt.java
```diff
@@ -0,0 +1,55 @@
...
+    private static void loadNativeLibraries() {
+        try {
+            System.loadLibrary("chrome_crashpad_handler");
+        } catch (UnsatisfiedLinkError e) {
+            try {
+                System.loadLibrary("chrobalt");
+            } catch (UnsatisfiedLinkError eFallback) {
+                try {
+                    for (String library : NativeLibraries.LIBRARIES) {
+                        System.loadLibrary(library);
```
hlwarriner: Please add comments to explain these fallback strategies.

components/crash/core/app/crashpad_handler_jni.cc
```diff
@@ -0,0 +1,51 @@
...
+  std::vector<std::string> argv_strings;
+  base::android::AppendJavaStringArrayToStringVector(
+      env, base::android::JavaParamRef<jobjectArray>(env, j_argv),
+      &argv_strings);
+
+  std::vector<const char*> argv;
+  StringVectorToCStringVector(argv_strings, &argv);
+  
+  ALOGI("Native: Transitioning to CrashpadHandlerMain with %zu args", argv.size() - 1);
+  CrashpadHandlerMain(argv.size() - 1, const_cast<char**>(argv.data()));
```
hlwarriner, line 46: Please add a short implementation comment to explain why `argv.size() - 1`, instead of `argv.size()`, is passed.

components/crash/core/app/BUILD.gn
```diff
@@ -226,8 +228,13 @@ if (is_mac || is_android || is_linux || is_chromeos) {
   executable("chrome_crashpad_handler") {
     sources = [ "chrome_crashpad_handler.cc" ]
 
+    if (is_android) {
```
hlwarriner, line 231: nit: should probably be `is_androidtv`.

components/crash/android/BUILD.gn
```diff
@@ -87,31 +88,51 @@ source_set("unit_tests") {
...
-static_library("crashpad_main") {
-  sources = [ "crashpad_main.cc" ]
-
-  deps = [
-    ":java_handler_jni_headers",
-    "//base",
-    "//third_party/crashpad/crashpad/client",
-    "//third_party/crashpad/crashpad/handler",
-  ]
+if (is_cobalt && is_androidtv) {
```
hlwarriner, line 91: I would personally prefer more targeted branching around/within these individual targets so that it's easier to see where Cobalt and Chromium diverge.

For example, `java_handler_jni_headers` looks identical between the two.

base/BUILD.gn
```diff
@@ -209,6 +209,7 @@ buildflag_header("cobalt_for_google3_buildflags") {
   header = "cobalt_for_google3_buildflags.h"
   header_dir = "base/android"
   flags = [
+    "IS_COBALT=$is_cobalt",
```
hlwarriner, line 212: Is this for `BUILDFLAG(IS_COBALT)`? Don't we already get that from `cobalt/build/configs/BUILD.gn`?

components/crash/android/java/src/org/chromium/components/crash/browser/CrashpadMainCobalt.java
alunev: thanks, this file is built for Cobalt only so I've simplified it.

PR #9758, opened 2026-03-25
android: Adding retry param to URL after network dialog Retry

ckim1346:
This PR adds a URL param 'netdialog_retry=1' to the URL if the network dialog "Try Again" button is pressed. It also shifts away from reloading the web contents as it's suspected we could be losing params and instead reloads the URL (more similar to C25). 

Bug: 496219065

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
@@ -183,7 +183,7 @@ public class PlatformError
               // Otherwise, add a param to the URL to indicate a bootstrap request with a retry from the network dialog
               if (currentUrl.isEmpty()) {
                 Log.i(TAG, "Visible URL is empty; cannot append retry parameter. Reloading the WebContents");
-                webContents.getNavigationController().reload(true);
+                webContents.getNavigationController().reload(/*param=*/true);
               } else {
                 cobaltActivity.getActiveShell().loadUrl(addRetryUrlParam(currentUrl));
               }


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
vincentmo, line 105: Curious, what does sending a CANCELED response do?

vincentmo, line 102: It seems possible for this to be called while a dialog is already shown. Should there be a check for the presence of a dialog and make this a noop?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,26 @@ interface Natives {
     void sendResponse(@PlatformError.Response int response, long data);
   }
 
-  /** Reloads the web contents if available */
-  private void reloadWebContents(CobaltActivity cobaltActivity) {
+  /** Reloads the URL and adds a retry param */
+  private void loadUrlWithRetryParam(CobaltActivity cobaltActivity) {
     if (cobaltActivity != null) {
       WebContents webContents = cobaltActivity.getActiveWebContents();
       if (webContents != null) {
```
vincentmo: Also early return when null.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,26 @@ interface Natives {
...
-        webContents.getNavigationController().reload(true);
+        String currentUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
+
+        String newUrl = currentUrl;
+        if (!currentUrl.isEmpty()) {
```
vincentmo: When would currentUrl be empty? And what happens when loadUrl() is invoked w/ an empty value?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,26 @@ interface Natives {
...
 
-  /** Reloads the web contents if available */
-  private void reloadWebContents(CobaltActivity cobaltActivity) {
+  /** Reloads the URL and adds a retry param */
+  private void loadUrlWithRetryParam(CobaltActivity cobaltActivity) {
     if (cobaltActivity != null) {
       WebContents webContents = cobaltActivity.getActiveWebContents();
       if (webContents != null) {
-        webContents.getNavigationController().reload(true);
+        String currentUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
```
vincentmo: You could just rename this to newUrl and eliminate the need for a 2nd variable, only update it if it's not empty.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,26 @@ interface Natives {
...
+            Uri.Builder uriBuilder = parsedUri.buildUpon();
+            uriBuilder.appendQueryParameter("netdialog_retry", "1");
+            newUrl = uriBuilder.build().toString();
+          }
+        }
+
+        cobaltActivity.getActiveShell().loadUrl(newUrl);
       } else {
-        Log.e(TAG, "WebContents is null and not available to reload the application.");
+        Log.e(TAG, "WebContents is null and not available to reload the URL.");
```
vincentmo: When is this expected to happen?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,26 @@ interface Natives {
...
       if (webContents != null) {
-        webContents.getNavigationController().reload(true);
+        String currentUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
+
+        String newUrl = currentUrl;
+        if (!currentUrl.isEmpty()) {
+          Uri parsedUri = Uri.parse(currentUrl);
+          if (parsedUri.getQueryParameter("netdialog_retry") == null) {
+            Uri.Builder uriBuilder = parsedUri.buildUpon();
+            uriBuilder.appendQueryParameter("netdialog_retry", "1");
```
vincentmo: Can we make this a counter instead, where it increments each time a retry is attempted for that app lifecycle?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: We do check for this where it's raised from in [cobalt_web_contents_observer](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/browser/cobalt_web_contents_observer.cc;drc=772e242047f98d009a17f5f808fcb022eafccec4;l=101)

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: We actually saw a significant crash when not checking b/466112455.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Essentially signals something went wrong with the dialog (ie activity == null) and the response was not user-prompted

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: It could be empty before the initial navigation has been started to the Navigation Controller or we for some reason have an uninitialized GURL object but neither of those should be able to happen as we raise it from [CobaltWebContentsObserver::DidFinishNavigation](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/browser/cobalt_web_contents_observer.cc;drc=772e242047f98d009a17f5f808fcb022eafccec4;l=74) (and as a callback in DidStartNavigation). But what can break will break -- adding a check to just reload as we did previously if it returns empty.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: I considered that but every 'retry' essentially raises a new platform error dialog -- we would have to plumb it through JNI to where it's called on the native side. That's why we have the `Cobalt.Network.CumulativePlatformErrorRaised` UMA histogram

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,29 @@ interface Natives {
...
-  private void reloadWebContents(CobaltActivity cobaltActivity) {
+  /** Reloads the URL and adds a retry param */
+  private void loadUrlWithRetryParam(CobaltActivity cobaltActivity) {
     if (cobaltActivity != null) {
       WebContents webContents = cobaltActivity.getActiveWebContents();
       if (webContents != null) {
-        webContents.getNavigationController().reload(true);
+        String newUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
+
+        if (newUrl.isEmpty() || newUrl == null) {
```
zhongqiliang: when will newUrl be null? if it is null, you already crashed the app

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,29 @@ interface Natives {
...
+          Log.e(TAG, "Visible URL is empty; cannot append retry parameter. Reloading the WebContents");
+          webContents.getNavigationController().reload(true);
+          return;
+        }
+        else {
+          Uri parsedUri = Uri.parse(newUrl);
+          if (parsedUri.getQueryParameter("netdialog_retry") == null) {
```
zhongqiliang: why not build it once, and cache the result?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +192,29 @@ interface Natives {
...
+            Uri.Builder uriBuilder = parsedUri.buildUpon();
+            uriBuilder.appendQueryParameter("netdialog_retry", "1");
+            newUrl = uriBuilder.build().toString();
+          }
+          cobaltActivity.getActiveShell().loadUrl(newUrl);
+        }
```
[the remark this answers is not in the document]
zhongqiliang: read this.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Updated

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Same as above -- the dialog is a new instance on every retry so would have to store it at a higher level and if the URL changes out of scope of the network dialog, we'd have to rebuild it again anyway. This is more similar to what C25 did in [browser_module.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:cobalt/browser/browser_module.cc;l=1?q=browser_module.cc&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2F25.lts.1%2B)

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
vincentmo: Is it possible to read the current url, parse the url param, increment, then set?

kaidokert:
> Unittests plz.

In spirit i agree. But : we have no test framework set up and running for Java code right now.  No CI job runs Java tests. ( [we did for Cobalt 25 and earlier](https://github.com/youtube/cobalt/actions/workflows/gradle.yaml) )

I think it's unreasonable burden to expect Charlie to set this all up in this PR ?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: theoretically yes, but I don't see a clean/easy way to do it -- is this pretty important to have? number of times per session vs just aggregating the total number of times the bootstrap request is from a retry?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
yell0wd0g: This is still unaddressed? 

Same with l. 201.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +195,42 @@ interface Natives {
     void sendResponse(@PlatformError.Response int response, long data);
   }
 
-  /** Reloads the web contents if available */
-  private void reloadWebContents(CobaltActivity cobaltActivity) {
+  //TODO: Add unit tests for retry logic
+  /** Reloads the URL and adds a retry param */
+  private void loadUrlWithRetryParam(CobaltActivity cobaltActivity) {
     if (cobaltActivity != null) {
```
yell0wd0g: Can `cobaltActivity` be null? Under what conditions?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +195,42 @@ interface Natives {
...
       WebContents webContents = cobaltActivity.getActiveWebContents();
       if (webContents != null) {
-        webContents.getNavigationController().reload(true);
+        String newUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
+
+        if (newUrl.isEmpty()) {
+          Log.e(TAG, "Visible URL is empty; cannot append retry parameter. Reloading the WebContents");
```
yell0wd0g: `Log.e` implies it's an error, but seems to be a normal condition?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +195,42 @@ interface Natives {
...
+
+        Uri parsedUri = Uri.parse(newUrl);
+        String retryParam = parsedUri.getQueryParameter(RETRY_PARAM_KEY);
+        if (retryParam != null) {
+          try {
+            int retryCount = Integer.parseInt(retryParam) + 1;
+            // Replace the existing parameter value in the string
+            newUrl = newUrl.replaceFirst("(" + RETRY_PARAM_KEY + "=)\\d+", "$1" + retryCount);
+          } catch (NumberFormatException e) {
+            Log.w(TAG, "Invalid netdialog_retry parameter value: " + retryParam);
```
yell0wd0g: Isn't it enough to use `e.toString()` here?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,14 +195,42 @@ interface Natives {
...
+            uriBuilder.appendQueryParameter(RETRY_PARAM_KEY, "1");
+            newUrl = uriBuilder.build().toString();
+          }
+        } else {
+          Uri.Builder uriBuilder = parsedUri.buildUpon();
+          uriBuilder.appendQueryParameter(RETRY_PARAM_KEY, "1");
+          newUrl = uriBuilder.build().toString();
+        }
+
+        cobaltActivity.getActiveShell().loadUrl(newUrl);
```
yell0wd0g: Feels like this method does too many things: looks for the
existing RETRY_PARAM_KEY and either adds one if not present
or increases its associated integer, _then_ it reloads. This
shows in the name `loadUrlWithRetryParam()`. Can we rename
this method to AddOrIncreaseRetryURLParam(currentURL) and 
returns a new one which then the caller uses to do the
`cobaltActivity.getActiveShell().loadUrl(newUrl);` ? I think
that would clarify a bit the return paths.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Ideally cobaltActivity would not be null (and if it were, the app would likely be crashing/shutting down) but we did see webContents being null and causing crashes when not handled, so it's probably best to null check it?

yell0wd0g: If there's a legitimate circumstance in which a variable can be null,
then yeah. But CobaltActivity being null sounds fishy. I'd suggest
either a massive CHECK(false) (or Java equivalent, `Objects.requireNonNull`?)
or an explanation of why this is possible.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Talked with @vincentmo offline -- the incremental count of the param is a little tricky so I've reverted those changes in the interest of time. Do you still want separation of the function?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Updated to Log.i

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,16 +195,34 @@ interface Natives {
...
-  /** Reloads the web contents if available */
-  private void reloadWebContents(CobaltActivity cobaltActivity) {
-    if (cobaltActivity != null) {
-      WebContents webContents = cobaltActivity.getActiveWebContents();
-      if (webContents != null) {
-        webContents.getNavigationController().reload(true);
-      } else {
-        Log.e(TAG, "WebContents is null and not available to reload the application.");
-      }
+  //TODO: Add unit tests for retry URL param logic b/496219065
```
yell0wd0g: nit: `//TODO(b/496219065): ...`

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,16 +195,34 @@ interface Natives {
...
+  /** Reloads the URL and adds a retry param */
+  private void loadUrlWithRetryParam(CobaltActivity cobaltActivity) {
+    // cobaltActivity should not be null but could be if the Activity was stopped (e.g.
+    // backgrounded) and StarboardBridge cleared the Holder, but a pending dialog click was
+    // still processed.
```
yell0wd0g: Thanks!

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -191,16 +195,34 @@ interface Natives {
...
+    WebContents webContents = cobaltActivity.getActiveWebContents();
+    if (webContents == null) {
+      Log.e(TAG, "WebContents is null and not available to reload the URL.");
+      return;
+    }
+    String newUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
+
+    if (newUrl.isEmpty()) {
+      Log.i(TAG, "Visible URL is empty; cannot append retry parameter. Reloading the WebContents");
+      webContents.getNavigationController().reload(true);
```
yell0wd0g: Do we need to reload in this case?

Also in this case there's no added parameter, but the method
doc says "Reloads the URL and adds a retry param", should it
say "Reloads the URL unconditionally and adds or increases
a retry param if the URL is readable" or some such? Such is the
problem of detailing the internals in the method documentation
:-) -- instead the method blurb could say the "why" do we
reload and/or add or increase parameters in the URL, as opposed
to describing the implementation. WDYT?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: No longer logging as we went back to the non-incremental approach

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Updated to return early on null

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Updated

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: No problem

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: without reloading, users would just run into a black screen since the dialog is just dismissed.

 I've separated out the adding of the param and moved the reload/loadUrl logic to the `RETRY_BUTTON` case. Let me know what you think

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -165,7 +169,26 @@ public void onClick(DialogInterface dialogInterface, int whichButton) {
...
+            if (webContents == null) {
+              Log.e(TAG, "WebContents is null and not available to reload the URL.");
+            } else {
+              String currentUrl = webContents.getVisibleUrl() != null ? webContents.getVisibleUrl().getSpec() : "";
+
+              // Reloading the web contents as a fallback if the URL is empty to attempt a fresh navigation.
+              // Otherwise, add a param to the URL to indicate a bootstrap request with a retry from the network dialog
+              if (currentUrl.isEmpty()) {
+                Log.i(TAG, "Visible URL is empty; cannot append retry parameter. Reloading the WebContents");
+                webContents.getNavigationController().reload(true);
```
vincentmo: Could you add a (/\*param=\*/true) comment here so it's clear what function param 'true' refers to?

vincentmo (commented):
lgtm

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
ckim1346: Done

PR #9773, opened 2026-03-26
build: Don't assume that is_starboard is Linux-only

hferreiro:
Linux was the only starboard platform until now. This led to implicit
assumptions about the OS for `is_starboard`. Fix that to take into
account that AOSP platforms build on Android.

Bug: 495203133

### DIFF
--- a/base/BUILD.gn
+++ b/base/BUILD.gn
@@ -61,6 +61,10 @@ if (is_win) {
   import("//build/config/win/control_flow_guard.gni")
 }
 
+if (is_cobalt) {
+  import("//starboard/build/config/toolchain_variables.gni")
+}
+
 declare_args() {
   # Unsafe developer build. Has developer-friendly features that may weaken or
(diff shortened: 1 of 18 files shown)


copied_base/base/BUILD.gn
```diff
@@ -3812,13 +3815,9 @@ test("base_unittests") {
 
     # Android does not use test_launcher to run gtests.
     sources -= [
-      "process/process_unittest.cc",
```
andrewsavage1, line 3823: We try to make any changes guarded by our flags, i.e. with the `if (is_starboard) {` block you're editing above, to make it clear what code we own vs what code is from upstream. Maybe what we can do here is add these files to the list at line 3681 so we only remove them for linux in the above block, and remove them for android here (probably with a comment in the code clarifying we still don't run them)

hferreiro: Done.

PR #9780, opened 2026-03-27
Fix Skia.SkGlyphCache UMA metrics

tasunnn:
Bug: 496590506

### DIFF
--- a/cobalt/browser/metrics/cobalt_metrics_browsertest.cc
+++ b/cobalt/browser/metrics/cobalt_metrics_browsertest.cc
@@ -109,7 +109,7 @@ IN_PROC_BROWSER_TEST_F(CobaltMetricsBrowserTest, MAYBE_RecordsMemoryMetrics) {
   check_histogram("Memory.Experimental.Browser2.V8");
   check_histogram("Memory.Experimental.Browser2.V8.AllocatedObjects");
   check_histogram("Memory.Experimental.Browser2.Skia");
-  check_histogram("Memory.Experimental.Browser2.Skia.Small.SkGlyphCache");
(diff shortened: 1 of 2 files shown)


cobalt/browser/metrics/cobalt_metrics_browsertest.cc
```diff
@@ -109,7 +109,7 @@ IN_PROC_BROWSER_TEST_F(CobaltMetricsBrowserTest, MAYBE_RecordsMemoryMetrics) {
   check_histogram("Memory.Experimental.Browser2.V8");
   check_histogram("Memory.Experimental.Browser2.V8.AllocatedObjects");
   check_histogram("Memory.Experimental.Browser2.Skia");
-  check_histogram("Memory.Experimental.Browser2.Skia.Small.SkGlyphCache");
+  check_histogram("Memory.Experimental.Browser2.Small.Skia.SkGlyphCache");
```
Awallky, line 112: I think this will change the name of the metric to omit the *Small* part of the metric name. 
You can use `cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py` to confirm its name.

tasunnn: I think the name was changed in the emitter but not here. It was originally Skia.Small.SkGlyphCache https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc;l=139;drc=9769f6ae0c89dc2145d1ed4075bf224ef73868b7;bpv=1;bpt=0, but now the suffix is Skia.SkGlyphCache, so maybe need to change this test case as well since I believe the size is prepended to the suffix?

PR #9785, opened 2026-03-27
android: Re-write input after input buffer unavailable error

jasonzhangxx:
When MediaCodec::GetInputBuffer returns null, indicating an input buffer
is unavailable, re-insert the current pending input into the queue for
reprocessing. This addresses scenarios where input buffers might be
temporarily unavailable due to dirty callbacks immediately following a
MediaCodec flush. Re-queueing allows the system to use an available
buffer, preventing decoding stalls or errors.

Bug: 493983265

### DIFF
--- a/starboard/android/shared/media_decoder.cc
+++ b/starboard/android/shared/media_decoder.cc
@@ -501,9 +501,13 @@ bool MediaDecoder::ProcessOneInputBuffer(
     ScopedJavaLocalRef<jobject> byte_buffer(
         media_codec_bridge_->GetInputBuffer(dequeue_input_result.index));
     if (byte_buffer.is_null()) {
-      SB_LOG(ERROR) << "Unable to write to MediaCodec buffer, |byte_buffer| is"
+      SB_LOG(ERROR) << "Unable to get MediaCodec input buffer, |byte_buffer| is"
                     << " null.";
-      // TODO: Stop the decoding loop and call error_cb_ on fatal error.
+      // There could be dirty callbacks right after flush, thus the
+      // MediaCodec.InputBuffer could be unavailable. In that case, we should
+      // re-write the input buffer with a different MediaCodec.InputBuffer.
+      pending_inputs->emplace(pending_inputs->begin(), pending_input);
+      number_of_pending_inputs_++;
       return false;
     }
 


starboard/android/shared/media_decoder.cc
```diff
@@ -662,9 +501,12 @@ bool MediaDecoder::ProcessOneInputBuffer(
...
         media_codec_bridge_->GetInputBuffer(dequeue_input_result.index));
     if (byte_buffer.is_null()) {
-      SB_LOG(ERROR) << "Unable to write to MediaCodec buffer, |byte_buffer| is"
+      SB_LOG(ERROR) << "Unable to get MediaCodec input buffer, |byte_buffer| is"
                     << " null.";
-      // TODO: Stop the decoding loop and call error_cb_ on fatal error.
+      // There could be dirty callbacks of input buffers right after flush, thus
+      // the input buffer could be unavailable. In that case, we should re-write
+      // the input buffer with a different input buffer.
+      pending_inputs->insert(pending_inputs->begin(), pending_input);
```
borongc: Although it doesn't make difference, I feel `pending_inputs->emplace_front(pending_input)` is more explicitly clear.

borongc (approved):

starboard/android/shared/media_decoder.cc
jasonzhangxx: Done

starboard/android/shared/media_decoder.cc
```diff
@@ -662,9 +501,13 @@ bool MediaDecoder::ProcessOneInputBuffer(
     ScopedJavaLocalRef<jobject> byte_buffer(
         media_codec_bridge_->GetInputBuffer(dequeue_input_result.index));
     if (byte_buffer.is_null()) {
-      SB_LOG(ERROR) << "Unable to write to MediaCodec buffer, |byte_buffer| is"
+      SB_LOG(ERROR) << "Unable to get MediaCodec input buffer, |byte_buffer| is"
                     << " null.";
-      // TODO: Stop the decoding loop and call error_cb_ on fatal error.
+      // There could be dirty callbacks of input buffers right after flush, thus
```
xiaomings: Chromium seems to [handle a similar issue](https://source.chromium.org/chromium/chromium/src/+/main:media/base/android/java/src/org/chromium/media/MediaCodecBridge.java;l=666;bpv=1;bpt=0;drc=5a355bf85c39a9807e58d1470f786a0e272bfd68) in a different way.  Consider investigating their solution to see if we should adopt.

starboard/android/shared/media_decoder.cc
```diff
@@ -662,9 +501,13 @@ bool MediaDecoder::ProcessOneInputBuffer(
...
+      // the input buffer could be unavailable. In that case, we should re-write
+      // the input buffer with a different input buffer.
```
xiaomings: Is it possible for you to elaborate on "we should re-write the input buffer with a different input buffer"?  The implementation below retries with the same input buffer index.

starboard/android/shared/media_decoder.cc
```diff
@@ -662,9 +501,13 @@ bool MediaDecoder::ProcessOneInputBuffer(
     ScopedJavaLocalRef<jobject> byte_buffer(
         media_codec_bridge_->GetInputBuffer(dequeue_input_result.index));
     if (byte_buffer.is_null()) {
-      SB_LOG(ERROR) << "Unable to write to MediaCodec buffer, |byte_buffer| is"
+      SB_LOG(ERROR) << "Unable to get MediaCodec input buffer, |byte_buffer| is"
```
xiaomings, line 504: `OnMediaCodecOutputBufferAvailable()` guards against flushing, but `OnMediaCodecInputBufferAvailable()` currently doesn't.  Shall we add the guard there instead?

starboard/android/shared/media_decoder.cc
jasonzhangxx: Discussed offline with Xiaoming. Chromium's pipeline may have a different threading model, and it's unclear if their solution can fully prevent dirty callbacks.

starboard/android/shared/media_decoder.cc
jasonzhangxx: Discussed with Xiaoming offline; we'll keep this change minimal as it doesn't use an experiment guard.

starboard/android/shared/media_decoder.cc
jasonzhangxx: The comment is updated.

johnxwork (approved):

PR #9794, opened 2026-03-27
Add UMA metrics for media memory allocation

tasunnn:
media: Add UMA metrics for media buffer memory

This change implements an OutputMemoryTracker in the 
MediaCodecBridge to monitor the lifecycle of mediaCodec 
output buffers on Android and estimates total memory allocated.

Introduce new UMA metrics, `Memory.Media.EncodedBuffer.Allocated`
and `Memory.Media.DecodedBuffer.Allocated` to track the total 
memory allocated by the media pipeline for encoded and decoded
buffers. This provides critical insight into the memory footprint of
media source buffers, aiding performance and stability analysis.

Bug: 487768963

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
@@ -133,23 +133,23 @@ class MediaCodecBridge {
         globalTotalMemoryUsage.addAndGet(-oldSize);
       }
       int trackedSize = size;
-            if (mDecodesToSurface) {
-                try {
-                    if (mActiveFormat != null) {
-                        int width = mActiveFormat.width();
-                        int height = mActiveFormat.height();
-                        if (width > 0 && height > 0) {
-                            // This is an estimated size assuming YUV420 format
-                            // width * height * 1.5
-                            trackedSize = width * height * 3 / 2;
-                        }
-                    }
-                } catch (Exception e) {
-                    trackedSize = 0;
-                }
+      if (mDecodesToSurface) {
+        try {
+          if (mActiveFormat != null) {
+            int width = mActiveFormat.width();
+            int height = mActiveFormat.height();
+            if (width > 0 && height > 0) {
+                // This is an estimated size assuming YUV420 format
+                // width * height * 1.5
+                trackedSize = width * height * 3 / 2;
             }
+          }
+        } catch (Exception e) {
+            trackedSize = 0;
+        }
+      }
       mBufferSizes.put(index, trackedSize);
-      mTotalMemoryUsage += sitrackedSizeze;
+      mTotalMemoryUsage += trackedSize;
       globalTotalMemoryUsage.addAndGet(trackedSize);
     }
 


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -114,7 +116,63 @@ public void onNewFrame(long presentationTimeUs) {
     }
   }
 
+  private static AtomicLong globalTotalMemoryUsage = new AtomicLong(0L);
```
Awallky, line 119: Style Violation (Naming): The static field globalTotalMemoryUsage must be renamed to sGlobalTotalMemoryUsage to adhere to the Chromium Java style guide (https://chromium.googlesource.com/chromium/src/+/main/styleguide/java/java.md#fields) for static members.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -854,6 +920,7 @@ private boolean configureVideo(
       }
 
       maybeSetMaxVideoInputSize(format);
+      mDecodesToSurface = surface != null;
```
jasonzhangxx, line 923: This will be always true. There should be always a valid `surface`.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -114,7 +116,63 @@ public void onNewFrame(long presentationTimeUs) {
...
+  @CalledByNative
+  public static long getGlobalOutputMemoryUsage() {
+    return globalTotalMemoryUsage.get();
+  }
+
+  private class OutputMemoryTracker {
+    private final SparseIntArray mBufferSizes = new SparseIntArray();
+    private long mTotalMemoryUsage = 0;
+
+    public synchronized void add(int index, int size) {
```
Awallky, line 129: Inconsistent indentation was found in `OutputMemoryTracker`. Lines such as trackedSize = width * height * 3 / 2; use 4-space indentation, whereas the Chromium and Cobalt standard for Java is 2-space indentation (https://source.android.com/docs/setup/contribute/code-style#use-spaces-for-indentation).

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -114,7 +116,63 @@ public void onNewFrame(long presentationTimeUs) {
...
+          if (mActiveFormat != null) {
+            int width = mActiveFormat.width();
+            int height = mActiveFormat.height();
+            if (width > 0 && height > 0) {
+                // This is an estimated size assuming YUV420 format
+                // width * height * 1.5
+                trackedSize = width * height * 3 / 2;
+            }
+          }
+        } catch (Exception e) {
```
Awallky, line 147: We should catch more specific exceptions, if we decide they are necessary - https://chromium.googlesource.com/chromium/src/+/main/styleguide/java/java.md#broad-catches.

media/base/media_client.cc
```diff
@@ -8,6 +8,10 @@
 #include "media/base/decoder_buffer.h"
 #endif
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA) && BUILDFLAG(IS_ANDROID)
+#include "starboard/android/shared/media_codec_memory_metrics.h"
```
jasonzhangxx, line 12: We're not supposed to directly use starboard code in chrobalt. A starboard extension is recommended. But I'm currently not sure if that's still true for Android. I'll defer to @johnxwork @y4vor @kaidokert.

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
```diff
@@ -414,6 +412,21 @@ void CobaltMemoryMetricsEmitter::CollateResults() {
...
+  base::UmaHistogramMemoryKB("Memory.Media.EncodedBuffer.Allocated",
+                             static_cast<int>(encoded_memory_bytes / kKiB));
+#if BUILDFLAG(IS_ANDROID)
+  uint64_t decoded_memory_bytes =
+      media::MediaClient::GetMediaDecodedTotalAllocatedMemory();
+  base::UmaHistogramMemoryKB("Memory.Media.DecodedBuffer.Allocated",
+                             static_cast<int>(decoded_memory_bytes / kKiB));
+#endif  // BUILDFLAG(IS_ANDROID)
+
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
jasonzhangxx, line 428: If the goal is to collect memory info in this file, it's not necessary to retrieve buffered output frames from java code. We can get similar information in native code. The original idea is to reuse chromium uma java code, so the example pr tried to get that information in java code.

tasunnn:
> No new tests were added to verify the tracking logic. Changes involving MediaCodec state management and global memory counters are prone to edge-case bugs (e.g., missed releases or resets).
> 
> I recommend adding unit tests to `cobalt_unittests` or browsertests to `cobalt_browsertests` to verify that OutputMemoryTracker correctly increments and decrements counts across various codec states (flush, stop, release).

I think verifying the OutputMemoryTracker would require a java test but I could add unittest/browsertest to check for the histograms.

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
tasunnn: Yeah we could put the UMA inside the java code. I originally thought it would be more consistent with how other memory metrics are gathered if also collecting it in this file. But if this requires too many unnecessary steps in between, we could just directly collect info in the java file, which also could resolve the comment below about using starboard code in cobalt.

PR #9798, opened 2026-03-27
Cobalt: Fix application lifecycle and focus reporting issues

jellefoks:
This PR significantly hardens the Cobalt application lifecycle, specifically addressing synchronous blocking requirements during downward state transitions and improving the robustness of the message pump for nested loops.

#### Key Improvements

*   **MessagePumpUIStarboard nested loop support:** Implemented the previously missing `Run()` method (which formerly threw a `NOTREACHED()`) to support a full Chromium-style message loop. This is critical for enabling nested `base::RunLoop` calls, allowing the main OS thread to block during teardown while the UI thread continues to process asynchronous tasks (e.g., surface destruction). The loop blocks highly efficiently using `base::WaitableEvent::TimedWait()`.
*   **Synchronous State Transitions:** Refactored `AppEventDelegate` to explicitly block the Starboard OS thread during downward transitions (Blur, Conceal, Freeze) until the target state is reached and all asynchronous side effects are completed. This uses a nested `RunLoop` on the UI thread and dynamically instantiated `WaitableEvent`s on non-UI threads. Added a 5-second safety timeout (`kTransitionTimeout`) to prevent permanent Starboard thread hangs.
*   **Stability & Thread Safety:**
    *   **Deadlock Prevention:** The final `kStopped` state transition (which destroys the `SequenceManager`) is now safely executed *synchronously* outside of the nested `RunLoop`. Attempting to tear down the task queue from within one of its own tasks would cause a fatal deadlock.
    *   Replaced `ContentMain` with `RunContentProcess` using a process-lifetime static `ContentMainRunner` to resolve SIGSEGV issues during shutdown.
    *   Introduced `base::Lock` and `std::atomic` state variables in `AppEventDelegate` and `AppEventRunner` to serialize lifecycle event processing and ensure thread safety across multiple threads without lock contention.
    *   Improved `MessagePumpUIStarboard` destruction (calling `CancelAll()` and `AfterRun()`) to prevent `DCHECK` crashes in `SequenceManager` during teardown.
*   **Visibility & Focus Reporting:**
    *   Added explicit calls to `web_contents()->WasHidden()` and `WasShown()` in `ShellPlatformDelegate` to ensure the compositor and Aura correctly track visibility changes.
    *   **Deterministic Focus:** Implemented a robust focus deferral mechanism in `cobalt::Shell`. Focus requests arriving while the window is hidden are queued and applied precisely when Aura signals visibility via `OnVisibilityChanged`.
*   **Robust Integration Tests:** Rewrote `test_lifecycle.sh` and `test_preload.sh` to include precise process cleanup validation (handling zombie states) and a new `wait_for_state.sh` helper that polls for JS state via DevTools, eliminating race conditions previously handled by brittle `sleep` commands.

Bug: 492166511
Bug: 491841704
Bug: 475990372
Bug: 477170017
Bug: 474496110
Bug: 492693465
Bug: 494002968

### DIFF
--- a/cobalt/app/app_event_delegate.cc
+++ b/cobalt/app/app_event_delegate.cc
@@ -18,13 +18,19 @@
 #include <utility>
 
 #include "base/check.h"
+#include "base/functional/bind.h"
 #include "base/logging.h"
 #include "base/notreached.h"
+#include "base/run_loop.h"
+#include "base/threading/platform_thread.h"
 #include "cobalt/app/app_event_runner.h"
+#include "content/public/browser/browser_thread.h"
 
 namespace cobalt {
 
 namespace {
+constexpr base::TimeDelta kTransitionTimeout = base::Seconds(5);
+
 AppEventDelegate::ApplicationState SbEventToTargetApplicationState(
     SbEventType type) {
   switch (type) {
@@ -57,7 +63,9 @@ AppEventDelegate::ApplicationState SbEventToTargetApplicationState(
 
 AppEventDelegate::AppEventDelegate(std::unique_ptr<AppEventRunner> runner)
     : runner_(std::move(runner)) {
+  base::AutoLock lock(lock_);
   application_state_ = ApplicationState::kInitial;
+  target_state_ = ApplicationState::kInitial;
   if (!runner_) {
     // If a special runner wasn't provided, use the default.
     runner_ = AppEventRunner::Create();
@@ -67,25 +75,63 @@ AppEventDelegate::AppEventDelegate(std::unique_ptr<AppEventRunner> runner)
 AppEventDelegate::~AppEventDelegate() {}
 
 bool AppEventDelegate::IsRunning() const {
+  base::AutoLock lock(lock_);
+  return IsRunningLocked();
+}
+
+bool AppEventDelegate::IsRunningLocked() const {
   return application_state_ != ApplicationState::kInitial &&
          application_state_ != ApplicationState::kStopped;
 }
 
 bool AppEventDelegate::IsVisible() const {
-  return application_state_ == ApplicationState::kBlurred || IsFocused();
+  base::AutoLock lock(lock_);
+  return IsVisibleLocked();
+}
+
+bool AppEventDelegate::IsVisibleLocked() const {
+  return application_state_ == ApplicationState::kBlurred ||
+         application_state_ == ApplicationState::kStarted;
 }
 
 bool AppEventDelegate::IsFocused() const {
+  base::AutoLock lock(lock_);
+  return IsFocusedLocked();
+}
+
+bool AppEventDelegate::IsFocusedLocked() const {
   return application_state_ == ApplicationState::kStarted;
 }
 
 bool AppEventDelegate::IsFrozen() const {
-  return application_state_ == ApplicationState::kFrozen || !IsRunning();
+  base::AutoLock lock(lock_);
+  return IsFrozenLocked();
+}
+
+AppEventDelegate::ApplicationState AppEventDelegate::GetState() const {
+  base::AutoLock lock(lock_);
+  return application_state_;
+}
+
+bool AppEventDelegate::IsFrozenLocked() const {
+  return application_state_ == ApplicationState::kFrozen ||
+         application_state_ == ApplicationState::kStopped ||
+         application_state_ == ApplicationState::kInitial;
 }
 
 void AppEventDelegate::HandleEvent(const SbEvent* event) {
+  // Use a lock to ensure thread safety as HandleEvent might be called from
+  // different threads (e.g., Starboard thread, UI thread).
+  base::AutoLock lock(lock_);
+  HandleEventLocked(event);
+}
+
+void AppEventDelegate::HandleEventLocked(const SbEvent* event) {
+  lock_.AssertAcquired();
+
   // Drop events received after the application stops.
-  if (application_state_ == ApplicationState::kStopped) {
+  if (application_state_ == ApplicationState::kStopped ||
+      target_state_ == ApplicationState::kStopped) {
     LOG(WARNING) << "Received event " << event->type
                  << " after stopping. Event is ignored.";
     return;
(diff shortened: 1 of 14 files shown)


cobalt/app/app_event_runner.cc
```diff
@@ -319,7 +366,7 @@ class AppEventRunnerImpl : public AppEventRunner {
 #endif
 #endif
     main_runner_ = GetContentMainRunner();
-    return content::ContentMain(std::move(params));
+    return content::RunContentProcess(std::move(params), main_runner_);
```
jellefoks, line 316: This one line change is at the core of the crash fix.

cobalt/app/app_event_delegate.cc
```diff
@@ -187,72 +338,65 @@ void AppEventDelegate::TransitionToLifeCycleState(ApplicationState state) {
...
+
+  bool is_downward = state > application_state_;
+
+  if (is_transitioning_) {
+    return;
+  }
+  is_transitioning_ = true;
+  content::GetUIThreadTaskRunner({})->PostTask(
+      FROM_HERE, base::BindOnce(&AppEventDelegate::ExecuteNextStepOnUIThread,
+                                base::Unretained(this)));
```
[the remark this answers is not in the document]
jellefoks: Acknowledged. I have restructured TransitionToLifeCycleState to ensure that downward transitions always block the Starboard thread, even if another transition (like an upward one) is already in progress. The check for is_transitioning_ now only skips the redundant PostTask for the state machine loop, but the blocking logic is always executed for downward moves.

cobalt/app/app_event_delegate.cc
```diff
@@ -177,7 +248,87 @@ void AppEventDelegate::HandleEvent(const SbEvent* event) {
...
+        next_state = ApplicationState::kConcealed;
+        break;
+      case ApplicationState::kConcealed:
+        runner_->OnFreeze();
+        next_state = ApplicationState::kFrozen;
+        break;
+      case ApplicationState::kFrozen:
+        runner_->OnStop();
+        next_state = ApplicationState::kStopped;
+        break;
```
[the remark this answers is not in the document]
jellefoks: Acknowledged. I have replaced the unreachable kFrozen -> kStopped async transition with a NOTREACHED() to clarify that Stop is only handled through the synchronous path.

cobalt/app/app_event_delegate.cc
```diff
@@ -187,72 +337,64 @@ void AppEventDelegate::TransitionToLifeCycleState(ApplicationState state) {
...
-          application_state_ = ApplicationState::kStopped;
-          break;
-        case ApplicationState::kStopped:
-        default:
-          NOTREACHED();
-          break;
+
+  target_state_ = state;
+
+  bool is_downward = state > application_state_;
```
andrewsavage1, line 343: "downward" doesn't seem like a very intuitive term to me—something like advancing/regressing or even forward/backward make more sense to me, as a conceptual "end" instead of "bottom" seems more like the stopped state

cobalt/app/app_event_delegate.cc
```diff
@@ -67,25 +75,63 @@ AppEventDelegate::AppEventDelegate(std::unique_ptr<AppEventRunner> runner)
...
+bool AppEventDelegate::IsFrozenLocked() const {
+  return application_state_ == ApplicationState::kFrozen ||
+         application_state_ == ApplicationState::kStopped ||
+         application_state_ == ApplicationState::kInitial;
 }
 
 void AppEventDelegate::HandleEvent(const SbEvent* event) {
+  // Use a lock to ensure thread safety as HandleEvent might be called from
+  // different threads (e.g., Starboard thread, UI thread).
+  base::AutoLock lock(lock_);
```
andrewsavage1, line 125: In the future, should we instead change application_state_ to an atomic? I don't know if there are any performance or deadlock concerns with the AutoLock approach here?

cobalt/app/app_event_delegate.h
```diff
@@ -39,6 +46,10 @@ class AppEventDelegate {
   // the invariant that states later in the enum are "further along" in the
   // lifecycle towards being stopped. Any updates to this enum must preserve
   // this monotonic ordering to ensure correct state machine transitions.
+  //
+  // The enum is ordered such that:
+  //   application_state_ < target_state_ => Moving towards Stopped
+  //   application_state_ > target_state_ => Moving towards Started
```
andrewsavage1, line 52: Whatever terminology we decide on (downward or otherwise), we should probably mention that term here

andrewsavage1 (approved):
Just minor comments, the functionality lgtm

cobalt/app/app_event_delegate.cc
jellefoks: This function also handles input events, deep links, etc. This is called as a result of input event handling (SbEventHandle calls from the platform). We should not be calling this ourselves, so there should be no risk for deadlocks resulting from this.

Just making the application_state_ atomic would not be sufficient to make this thread safe.

This lock exists because we should make this thread-safe for when the platforms call it from different threads. Our example modular/evergreen implementation already calls this from a different thread for start/preload event than for the other lifecycle changes, and I think different again for the input events.

This is not for a regression versus older releases, those do not have thread safety either. The locking overhead is negligible compared to the rate at which these occur (highest rate of calls would be for mouse/pad pointer position updates), so the insurance that this gives imho is very well worth it.

cobalt/app/app_event_delegate.cc
jellefoks: I will modify this in a followup PR to allow this to merge now, unblocking others that are running into the crash that this resolves.

cobalt/app/app_event_delegate.h
jellefoks: Acknowledged.

cobalt/app/app_event_delegate.cc
andrewsavage1: Sounds good

PR #9808, opened 2026-03-29
zlib: Fix GzipCompress invalid pointer crash

jellefoks:
GzipCompress crashed in a `CobaltMetricsLogUploaderTest` test 
due to a mismatched memory allocator issue.

The initial buffer was allocated using base::UncheckedMalloc,
which routes through PartitionAlloc. However, subsequent resizing
attempted to use the system realloc(), causing a SIGABRT.

This change avoids realloc() and correctly pairs base::UncheckedFree()
with base::UncheckedMalloc() to resolve the invalid pointer abort.

Bug: 486012022
Bug: 486003653
Bug: 502259405

### DIFF
--- a/third_party/zlib/google/compression_utils.cc
+++ b/third_party/zlib/google/compression_utils.cc
@@ -4,6 +4,10 @@
 
 #include "third_party/zlib/google/compression_utils.h"
 
+#if defined(COBALT)
+#include "build/build_config.h"
+#endif
+
 #include "base/bit_cast.h"
 #include "base/check_op.h"
 #include "base/process/memory.h"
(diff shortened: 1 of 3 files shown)


third_party/zlib/google/compression_utils.cc
```diff
@@ -57,20 +57,14 @@ bool GzipCompress(base::span<const uint8_t> input, std::string* output) {
           compressed_data, &compressed_data_size,
           base::bit_cast<const Bytef*>(input.data()), input_size, nullptr,
           nullptr) != Z_OK) {
-    free(compressed_data);
+    base::UncheckedFree(compressed_data);
     return false;
   }
 
-  Bytef* resized_data =
-      reinterpret_cast<Bytef*>(realloc(compressed_data, compressed_data_size));
```
jellefoks, line 65: This realloc is redundant, the compressed_data is already malloced at compressed_data_size. this was introduced a long time ago in a change (upstream [e25abf0](https://source.chromium.org/chromium/chromium/src/+/e25abf094f9cde0f500159c32628f3874d2df51b)) when migrating from std::vector to malloc/free, where this replaced a vector.resize() where the allocation was different.

third_party/zlib/google/compression_utils.cc
jellefoks, line 55: This was changed from a malloc to an uncheckedmalloc in upstream commit 
[1d6745b](https://source.chromium.org/chromium/chromium/src/+/1d6745b2ead8b5f23575773e51e3165f0f0a9a4b) without changing the corresponding realloc and free.

andrewsavage1:
@sherryzy this seems similar to a bug you found before?

third_party/zlib/google/compression_utils.cc
hlwarriner: Thanks for this comment - makes it much easier to review/understand the current PR.

third_party/zlib/google/compression_utils.cc
```diff
@@ -57,20 +57,14 @@ bool GzipCompress(base::span<const uint8_t> input, std::string* output) {
           compressed_data, &compressed_data_size,
           base::bit_cast<const Bytef*>(input.data()), input_size, nullptr,
           nullptr) != Z_OK) {
-    free(compressed_data);
+    base::UncheckedFree(compressed_data);
```
hlwarriner, line 65: I think we should guard the fixes with `BUILDFLAG(IS_COBALT)`. Otherwise, someone might accidentally drop them and revert to `free` when rebasing (assuming the upstream code still has `free` at that version).

hlwarriner, line 65: Not a high priority but do you plan to file a bug or upload a fix for the upstream code?

hlwarriner (approved):
Nice fix and cleanup - great to see removed test filter.

third_party/zlib/google/compression_utils.cc
jellefoks: Yeah, the upstream mysteriously is still broken.

andrewsavage1 (approved):
+1 to Holden's comments about guarding the changes, but giving 3p approval

third_party/zlib/google/compression_utils.cc
y4vor: +1 to add the `BUILDFLAG(IS_COBALT)`.

jellefoks: b/502259405

jellefoks: Done

PR #9817, opened 2026-03-30
cobalt: Make MetricsPollingState RefCounted

yell0wd0g:
Update MetricsPollingState to inherit from base::RefCounted instead of
base::RefCountedThreadSafe. This change clarifies that objects of this
type are not intended for concurrent access from multiple threads.

Introduce COBALT_THREAD_CHECKER to explicitly enforce single-threaded
access for MetricsPollingState and its derived classes, preventing
potential threading issues and aligning the object's access pattern
with its intended usage.

### DIFF
--- a/cobalt/browser/metrics/cobalt_metrics_service_client.cc
+++ b/cobalt/browser/metrics/cobalt_metrics_service_client.cc
@@ -20,6 +20,7 @@
 #include "base/command_line.h"
 #include "base/logging.h"
 #include "base/memory/raw_ptr.h"
+#include "base/memory/ref_counted.h"
 #include "base/notreached.h"
 #include "base/posix/file_descriptor_shuffle.h"
 #include "base/strings/string_number_conversions.h"
@@ -31,6 +32,7 @@
 #include "cobalt/browser/metrics/cobalt_memory_metrics_emitter.h"
 #include "cobalt/browser/metrics/cobalt_metrics_log_uploader.h"
 #include "cobalt/browser/switches.h"
+#include "cobalt/common/cobalt_thread_checker.h"
 #include "components/metrics/metrics_service.h"
 #include "components/metrics/metrics_state_manager.h"
 #include "components/prefs/pref_service.h"
@@ -43,11 +45,12 @@ namespace cobalt {
 
 // TODO(b/495528560): Fix ownership model in MetricsPollingState to separate CPU
 // and memory metrics polling
-class MetricsPollingState
-    : public base::RefCountedThreadSafe<MetricsPollingState> {
+class MetricsPollingState : public base::RefCounted<MetricsPollingState> {
  public:
   explicit MetricsPollingState(CobaltMetricsServiceClient* parent)
-      : parent_(parent) {}
+      : parent_(parent) {
+    COBALT_DETACH_FROM_THREAD(thread_checker_);
+  }
 
   // Parent pointer.
   raw_ptr<CobaltMetricsServiceClient> parent_;
@@ -59,6 +62,7 @@ class MetricsPollingState
   bool stop_logging = false;
 
   void RecordMetricsAfterDelay() {
+    CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
     if (stop_logging) {
       return;
     }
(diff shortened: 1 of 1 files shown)


Awallky (changes requested):
I think we have a couple of issues with a data race and concurrent accesses:

1. `scoped_refptr<MetricsPollingState>` instances are still being managed across multiple threads:
   - They are held as members of `CobaltMetricsServiceClient` on the main thread.
   - They are passed to a ThreadPool task_runner via `base::RetainedRef(this)` in `RecordMetricsAfterDelay`.
When the task on the task_runner completes, the reference count will be decremented on that background thread. Since `base::RefCounted` is not thread-safe, this results in a race condition on the reference count itself.

2. In the CobaltMetricsServiceClient destructor (running on the main thread), the fields `stop_logging` and `parent_` of the `MetricsPollingState` objects are modified:

```
   1 if (memory_state_) {
   2   memory_state_->stop_logging = true;
   3   memory_state_->parent_ = nullptr;
   4 }
```

At the same time, these fields are read on the background task_runner thread within `RecordMetricsAfterDelay` and `RequestMetrics`.

If we want to change this I think we will need a bit more of a refactor to support a `base::RefCounted` `MetricsPollingState`.

Awallky:
`base::SequenceBound<T>` seems like a good candidate for this as it manages its own lifetime and we do not need to expose a raw pointer or use RefCounts.

yell0wd0g:
> I think we have a couple of issues with a data race and concurrent accesses:
> 
> 1. `scoped_refptr<MetricsPollingState>` instances are still being managed across multiple threads:
>    
>    * They are held as members of `CobaltMetricsServiceClient` on the main thread.
>    * They are passed to a ThreadPool task_runner via `base::RetainedRef(this)` in `RecordMetricsAfterDelay`.
>      When the task on the task_runner completes, the reference count will be decremented on that background thread. Since `base::RefCounted` is not thread-safe, this results in a race condition on the reference count itself.
> 2. In the CobaltMetricsServiceClient destructor (running on the main thread), the fields `stop_logging` and `parent_` of the `MetricsPollingState` objects are modified:
> 
> ```
>    1 if (memory_state_) {
>    2   memory_state_->stop_logging = true;
>    3   memory_state_->parent_ = nullptr;
>    4 }
> ```
> 
> At the same time, these fields are read on the background task_runner thread within `RecordMetricsAfterDelay` and `RequestMetrics`.
> 
> If we want to change this I think we will need a bit more of a refactor to support a `base::RefCounted` `MetricsPollingState`.

I didn't follow all the code paths so apologies in advance, but I left the
Cobalt ThreadChecker that would be hit on all builds (not just debug/devel)
and it didn't trip on a linux monolithic build. Maybe ATV has a different
thread model?

yell0wd0g:
> > I think we have a couple of issues with a data race and concurrent accesses:
> > 
> > 1. `scoped_refptr<MetricsPollingState>` instances are still being managed across multiple threads:
> >    
> >    * They are held as members of `CobaltMetricsServiceClient` on the main thread.
> >    * They are passed to a ThreadPool task_runner via `base::RetainedRef(this)` in `RecordMetricsAfterDelay`.
> >      When the task on the task_runner completes, the reference count will be decremented on that background thread. Since `base::RefCounted` is not thread-safe, this results in a race condition on the reference count itself.
> > 2. In the CobaltMetricsServiceClient destructor (running on the main thread), the fields `stop_logging` and `parent_` of the `MetricsPollingState` objects are modified:
> > 
> > ```
> >    1 if (memory_state_) {
> >    2   memory_state_->stop_logging = true;
> >    3   memory_state_->parent_ = nullptr;
> >    4 }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > At the same time, these fields are read on the background task_runner thread within `RecordMetricsAfterDelay` and `RequestMetrics`.
> > If we want to change this I think we will need a bit more of a refactor to support a `base::RefCounted` `MetricsPollingState`.
> 
> I didn't follow all the code paths so apologies in advance, but I left the Cobalt ThreadChecker that would be hit on all builds (not just debug/devel) and it didn't trip on a linux monolithic build. Maybe ATV has a different thread model?

Actually I see now

```
[ RUN      ] CobaltMetricsServiceClientTest.GetMetricsServiceReturnsCreatedInstance
[INFO:cobalt/browser/metrics/cobalt_metrics_service_client.cc:203] Starting refresh timer for: 150 seconds
[FATAL:base/memory/ref_counted.h:64] DCHECK failed: CalledOnValidSequence().
```

Awallky:
@yell0wd0g 
> I didn't follow all the code paths so apologies in advance, but I left the
Cobalt ThreadChecker that would be hit on all builds (not just debug/devel)
and it didn't trip on a linux monolithic build.

Yeah I think it is a race, so it may not always occur. But with enough devices, we will eventually run into that issue.

PR #9841, opened 2026-03-31
Add C25->C26 Crashpad integration migration guide

hlwarriner:
A couple existing docs are also updated for accuracy.

Issue: 498170385

### DIFF
--- a/starboard/doc/eap/crashpad_integration.md
+++ b/starboard/doc/eap/crashpad_integration.md
@@ -1,4 +1,4 @@
-# Migration a platform's Crashpad integration
+# Migrating a platform's Crashpad integration
 
 Partners are required to complete some steps to integrate Crashpad into Cobalt.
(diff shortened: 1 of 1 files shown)


starboard/doc/crash_handlers.md
```diff
@@ -51,3 +51,8 @@ int main(int argc, char** argv) {
   ...
 }
 ```
+
+NOTE: platforms using Evergreen should not call
+`crashpad::InstallCrashpadHandler()` from their own application code. Starting
```
andrewsavage1, line 56: Maybe we should make a change to the actual code to change the name of this function or not expose it to partners applications (i.e. through gn visibility settings or an anonymous namespace)?

hlwarriner: Yeah that's a good idea. I filed b/498258600 to track it.

PR #9867, opened 2026-04-01
android: Wait for DRM system to open crypto session

osagie98:
Introduce a mechanism to synchronize the Android DRM system's
MediaCrypto session initialization. This addresses a race
condition where media decoders and DRM operations could attempt
to interact with an uninitialized MediaCrypto session.

The DrmSystem class now uses a condition variable to signal when
its MediaCrypto session is fully created. MediaCodecBridge is
updated to wait for this readiness before queuing secure buffers.
Similarly, MediaDrmBridge's key operations now block until the
DRM system is ready, ensuring proper state synchronization. This
improves stability for DRM-protected content playback,
particularly on devices with multi-threaded Widevine CDMs.

#vibe-coded

Bug: 389180717
Bug: 491355049


osagie98:
I've given a patch with a similar implementation in C25 to MediaTek (b/491355049), currently waiting for their feedback

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -34,6 +34,8 @@ using base::android::ToJavaByteArray;
 using base::android::ToJavaIntArray;
 using jni_zero::AttachCurrentThread;
 
+constexpr int64_t kDrmSystemReadyTimeoutUsec = 500'000;
```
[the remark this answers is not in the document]
osagie98: I'll keep these separate as we may want separate timeout periods for each

starboard/android/shared/media_codec_bridge.h
```diff
@@ -186,12 +192,16 @@ class MediaCodecBridge {
 
  private:
   // |MediaCodecBridge|s must only be created through its factory methods.
-  explicit MediaCodecBridge(Handler* handler);
+  explicit MediaCodecBridge(Handler* handler,
+                            DrmSystemReadyCb drm_system_ready_cb = {});
```
kjyoun: Looks like that we are always calling  Ctor with `drm_system_ready_cb` 

Then we don't need to make drm_system_ready_cb to optional param

starboard/android/shared/media_codec_bridge.h
```diff
@@ -186,12 +192,16 @@ class MediaCodecBridge {
...
   void Initialize(jobject j_media_codec_bridge);
 
   Handler* const handler_;
   base::android::ScopedJavaGlobalRef<jobject> j_media_codec_bridge_ = NULL;
 
+  DrmSystemReadyCb drm_system_ready_cb_;
```
kjyoun: const DrmSystemReadyCb ?

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -345,6 +349,16 @@ jint MediaCodecBridge::QueueSecureInputBuffer(
     const SbDrmSampleInfo& drm_sample_info,
     jlong presentation_time_microseconds,
     jboolean is_decode_only) {
+  if (!drm_system_ready_.load()) {
+    if (drm_system_ready_cb_) {
```
kjyoun: you can merge this check into one. Then we can remove one-level of nesting.

[go/tott/733](https://goto.google.com/tott/733)

starboard/android/shared/media_drm_bridge.cc
```diff
@@ -41,6 +41,8 @@ using base::android::ToJavaByteArray;
 
 using DrmOperationResult = MediaDrmBridge::OperationResult;
 
+constexpr int64_t kDrmSystemReadyTimeoutUsec = 500'000;
```
kjyoun: s/Usec/Us/

We usually use "Us" for usec suffix

https://source.corp.google.com/search?q=constexpr.*Us&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

starboard/android/shared/media_drm_bridge.cc
```diff
@@ -236,6 +238,13 @@ void MediaDrmBridge::CloseSession(std::string_view session_id) const {
 }
 
 const void* MediaDrmBridge::GetMetrics(int* size) {
+  if (!host_->WaitForDrmSystemReady(kDrmSystemReadyTimeoutUsec)) {
```
kjyoun: why do we need this WaitForDrmSystemReady only for GetMetrics, not for other methods?

starboard/android/shared/drm_system.cc
```diff
@@ -408,6 +410,13 @@ void DrmSystem::OnKeyStatusChange(
                                  drm_key_ids.data(), drm_key_statuses.data());
 }
 
+bool DrmSystem::WaitForDrmSystemReady(int64_t timeout_usec) {
+  std::unique_lock<std::mutex> lock(mutex_);
```
kjyoun: you don't need "<std::mutex>" here. compiler can deduct it. (CTAD)

starboard/android/shared/drm_system.cc
```diff
@@ -95,7 +95,9 @@ void DrmSystem::Run() {
   SB_CHECK(!enable_app_provisioning_);
 
   if (media_drm_bridge_->CreateMediaCryptoSession()) {
+    std::unique_lock<std::mutex> lock(mutex_);
```
kjyoun: you don't need "<std::mutex>" here. compiler can deduct it. (CTAD)


for locking, this is more efficient

{
  std::lock_guard lock(mutex_);
  created_media_crypto_session_.store(true);
}
// Call notify after releasing the lock.
created_media_crypto_session_cv_.notify_all();

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -345,6 +349,16 @@ jint MediaCodecBridge::QueueSecureInputBuffer(
...
+      if (!drm_system_ready_cb_(kDrmSystemReadyTimeoutUsec)) {
```
kjyoun: this part looks strange to me.

though this PR is to wait for MediaCrypto to be created, the code informs readiness(by calling callback) when QueueSecureInputBuffer is called. callback might be called eariler. (e.g. when the very moment MediaCrypto is created)

osagie98: The MediaCrypto is created when the DrmSystem is constructed, so it should always be available to the MediaCodecBridge. So I'm only interested in when the media crypto session is created, checking in `QueueSecureInputBuffer` allows us to delay any potential waits as long as possible.

starboard/android/shared/media_drm_bridge.cc
osagie98: I added to a couple more methods, I left our the opensession/provisioning methods as they either shouldn't need a mediacrypto or the session is expected not to have opened yet. I focused on GetMetrics as there is a related YTS failure b/389180717, and I wasn't sure if adding more methods is too heavy handed

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -345,6 +349,15 @@ jint MediaCodecBridge::QueueSecureInputBuffer(
     const SbDrmSampleInfo& drm_sample_info,
     jlong presentation_time_microseconds,
     jboolean is_decode_only) {
+  if (!drm_system_ready_.load()) {
+    SB_CHECK(drm_system_ready_cb_);
+    if (!drm_system_ready_cb_(kDrmSystemReadyTimeoutUs)) {
```
kjyoun: wait_drm_system_ready_cb_ ?

I originally thought that calling `drm_system_ready_cb_` means notifying drm system is ready

starboard/android/shared/drm_system.cc
```diff
@@ -192,7 +196,7 @@ void DrmSystem::GenerateSessionUpdateRequestWithAppProvisioning(
       request->GenerateWithAppProvisioning(media_drm_bridge_.get());
   switch (result.status) {
     case DRM_OPERATION_STATUS_SUCCESS:
-      created_media_crypto_session_ = true;
+      created_media_crypto_session_.store(true);
```
kjyoun: To achieve parity with line 98-102, I think that we should call notify_all here as well.

Then I would recommend to add method NotifyMediaCryptoSessionCreated and call it in both places

```

NotifyMediaCryptoSessionCreated() {
    {
      std::lock_guard lock(mutex_);
      created_media_crypto_session_.store(true);
    }
    created_media_crypto_session_cv_.notify_all();
}

if (media_drm_bridge_->CreateMediaCryptoSession()) {
   NotifyMediaCryptoSessionCreated();
}

  switch (result.status) {
    case DRM_OPERATION_STATUS_SUCCESS:
        NotifyMediaCryptoSessionCreated();


```

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -345,6 +349,15 @@ jint MediaCodecBridge::QueueSecureInputBuffer(
     const SbDrmSampleInfo& drm_sample_info,
     jlong presentation_time_microseconds,
     jboolean is_decode_only) {
+  if (!drm_system_ready_.load()) {
+    SB_CHECK(drm_system_ready_cb_);
```
kjyoun: why not move  SB_CHECK(drm_system_ready_cb_) to Ctor?

I think that we are always setting drm_system_ready_cb_

MediaCodecBridge::MediaCodecBridge(Handler* handler,
                                   DrmSystemReadyCb drm_system_ready_cb)
    : handler_(handler),
      drm_system_ready_cb_(std::move(drm_system_ready_cb)) {
SB_CHECK(drm_system_ready_cb_);
}

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -492,7 +505,11 @@ void MediaCodecBridge::OnMediaCodecFirstTunnelFrameReady(JNIEnv* env) {
   handler_->OnMediaCodecFirstTunnelFrameReady();
 }
 
-MediaCodecBridge::MediaCodecBridge(Handler* handler) : handler_(handler) {
+MediaCodecBridge::MediaCodecBridge(Handler* handler,
+                                   DrmSystemReadyCb drm_system_ready_cb)
+    : handler_(handler),
+      drm_system_ready_cb_(std::move(drm_system_ready_cb)),
+      drm_system_ready_(!drm_system_ready_cb_) {
```
kjyoun: you are setting drm system readiness  (drm_system_ready_) based on the existence of the callback.
this looks strange to me.

drm_system_ready_ should about the readiness of drm system, not about the existence the callback.

I think that you are dping this not to call callback when callback is not available. Then you should name it accordingly or add another member variable for it.

starboard/android/shared/media_drm_bridge.cc
```diff
@@ -321,6 +336,15 @@ bool MediaDrmBridge::IsCbcsSupported(JNIEnv* env) {
   return Java_MediaDrmBridge_isCbcsSchemeSupported(env) == JNI_TRUE;
 }
 
+bool MediaDrmBridge::IsDrmSystemReady(std::string_view caller_name) const {
+  if (!host_->WaitForDrmSystemReady(kDrmSystemReadyTimeoutUs)) {
+    SB_LOG(ERROR) << "Timed out waiting for DRM system to be ready for "
```
kjyoun: WARNING?

starboard/android/shared/media_drm_bridge.cc
```diff
@@ -236,6 +246,11 @@ void MediaDrmBridge::CloseSession(std::string_view session_id) const {
 }
 
 const void* MediaDrmBridge::GetMetrics(int* size) {
+  if (!IsDrmSystemReady(__FUNCTION__)) {
```
kjyoun: I am not sure adding IsDrmSystemReady across the methods is the best way, though I have no good suggestion for now.

Do we really this?

Isn't adding waiting logic in  MediaCodecBridge::QueueSecureInputBuffer enough?

osagie98: Possibly, I can't tell without a proper test, but we don't have the required devices. I can ask Amlogic to try two builds, one waiting in QueueSecureInputBuffer, and one waiting there and elsewhere, and compare

kjyoun: sounds great.

My concern would be how we can make sure the future change not break this condition.

The best way is to have a unit test to make sure multi-thread DRM is not broken, though I don't have a good idea about how to implement it.

starboard/android/shared/media_codec_bridge.cc
osagie98: It'll be null for non DRM playbacks. ctor check is ideal but I would need some other way to tell if the mediacodecbridge will support encrypted streams on creation, and I feel that passing a flag might be a less straightforward way to do it

starboard/android/shared/media_codec_bridge.cc
osagie98: I renamed related variables and functions to be more clear, the cb is really for the MediaCodecBridge to check if the DrmSystem's mediacryptosession has opened, and wait if it hasn't. Ideally, the DrmSystem will notify the MediaCodecBridge when the session opens, but I can't do it since the DrmSystem is created before the MediaCodecBridge. It's a little unclear what this callback does, hopefully the updated name is clearer.

starboard/android/shared/media_codec_bridge.cc
```diff
@@ -339,12 +345,19 @@ jint MediaCodecBridge::QueueInputBuffer(jint index,
       presentation_time_microseconds, flags, is_decode_only);
 }
 
-jint MediaCodecBridge::QueueSecureInputBuffer(
+int MediaCodecBridge::QueueSecureInputBuffer(
```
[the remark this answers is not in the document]
osagie98: Gemini cli sure likes changing things I didn't ask for

PR #9868, opened 2026-04-01
cobalt: Register dummy H5vccUpdater binders

sideb0ard:
Register dummy binders for H5vccUpdater and H5vccUpdaterSideloading Mojo
interfaces in builds where they are not natively supported (e.g.,
non-Evergreen or Linux-x64).

Previously, JavaScript probing for these interfaces would result in a
ReportBadMessage error if no binder was registered. This error caused the
BrowserInterfaceBroker (Mojo Master Pipe) to disconnect for the entire
frame. This disconnection prevented subsequent interface requests, such
as MediaDevicesDispatcherHost for getUserMedia(), from reaching the
browser process, leading to microphone recognition failures. Registering
dummy binders ensures these probes fail gracefully without severing the
Mojo connection.

Bug: 479656309

### DIFF
--- a/cobalt/browser/cobalt_browser_interface_binders.cc
+++ b/cobalt/browser/cobalt_browser_interface_binders.cc
@@ -110,6 +110,10 @@ void PopulateCobaltFrameBinders(
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterImpl::Create));
 #else
+  // Always register a binder for H5vccUpdater to prevent the browser
+  // from killing the Mojo connection if the renderer probes for this interface.
+  // If this is not registered, it disconnects the overall
+  // BrowserInterfaceBroker for the frame.
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(base::BindRepeating(
       [](content::RenderFrameHost*,
          mojo::PendingReceiver<h5vcc_updater::mojom::H5vccUpdater>) {
@@ -119,6 +123,8 @@ void PopulateCobaltFrameBinders(
 
 // Always register a binder for H5vccUpdaterSideloading to prevent the browser
 // from killing the Mojo connection if the renderer probes for this interface.
+// If this is not registered, it disconnects the overall BrowserInterfaceBroker
+// for the frame.
 // TODO(b/458483469): Remove the ALLOW_EVERGREEN_SIDELOADING check after
 // security review.
 #if BUILDFLAG(USE_EVERGREEN) && !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && \


hlwarriner:
Thanks for debugging this and proposing a fix. The behavior you describe in the PR description is surprising to me, since we expected the conditional binding added in #8774 to provide graceful handling on non-evergreen platforms (and I thought this was manually tested e2e). But yeah, what you describe actually is documented in https://chromium.googlesource.com/chromium/src/+/HEAD/docs/mojo_and_services.md#interface-brokers. I guess it may be a security feature.

cobalt/browser/cobalt_browser_interface_binders.cc
```diff
@@ -108,13 +109,30 @@ void PopulateCobaltFrameBinders(
...
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterImpl::Create));
+#else
+  binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(base::BindRepeating(
+      [](content::RenderFrameHost*,
+         mojo::PendingReceiver<h5vcc_updater::mojom::H5vccUpdater>) {
+        VLOG(1) << "Ignoring H5vccUpdater request for non-Evergreen build.";
+      }));
+#endif
+
+// Always register a binder for H5vccUpdaterSideloading to prevent the browser
```
hlwarriner, line 124: nit: please just make it clear, as you did in the PR description, that this is to prevent not just a Mojo disconnection for this particular interface (which we actually would want) but a disconnection of the overall BrowserInterfaceBroker for the frame.

hlwarriner, line 124: Thanks for the helpful comment. Please add a similar one for the H5vccUpdater dummy binder, or address that particular dummy binder in this comment as well.

cobalt/browser/cobalt_browser_interface_binders.cc
```diff
@@ -108,13 +109,30 @@ void PopulateCobaltFrameBinders(
 #if BUILDFLAG(USE_EVERGREEN)
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterImpl::Create));
+#else
+  binder_map->Add<h5vcc_updater::mojom::H5vccUpdater>(base::BindRepeating(
```
hlwarriner, line 117: With this fix, what does a JS web client actually observe now if it calls one of the updater APIs on a non-evergreen platform...does it match the behavior described in `third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.idl`?

hlwarriner (approved):

cobalt/browser/cobalt_browser_interface_binders.cc
sideb0ard: done

sideb0ard: done

cobalt/browser/cobalt_browser_interface_binders.cc
sideb0ard: Yeah, it matches - it gets a message saying " API not supported for this platform."

I tested via devtools - 
window.h5vcc.updater.getUpdaterChannel().then(val => console.log("SUCCESS: " + val)).catch(err => console.log("EXPECTED REJECTION: " + err));
Promise {<pending>}
VM191:1 EXPECTED REJECTION: API not supported for this platform.

hlwarriner: Great! Thanks for checking.

cobalt/browser/cobalt_browser_interface_binders.cc
```diff
@@ -108,13 +109,36 @@ void PopulateCobaltFrameBinders(
...
 // security review.
-#if !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
+#if BUILDFLAG(USE_EVERGREEN) && !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && \
+    ALLOW_EVERGREEN_SIDELOADING
   binder_map->Add<h5vcc_updater::mojom::H5vccUpdaterSideloading>(
       base::BindRepeating(&h5vcc_updater::H5vccUpdaterSideloadingImpl::Create));
-#endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD) && ALLOW_EVERGREEN_SIDELOADING
-#endif  // BUILDFLAG(USE_EVERGREEN)
+#else
+  binder_map->Add<
```
kjyoun, line 135: style nit: you don't use argument. Then this would be more concise

```
binder_map->Add<h5vcc_updater::mojom::H5vccUpdaterSideloading>(
    base::BindRepeating([](auto&&...) {
      VLOG(1) << "Ignoring H5vccUpdaterSideloading request.";
    }));
```

sideb0ard: Tried that but it wouldnt compile - base::BindRepeating does not support generic lambdas (using auto).
Gemini says:

 Why it failed:
  Chromium's base::Bind system uses template metaprogramming to "inspect" the lambda and extract its exact function signature (the RunType). When you use auto&&..., the lambda's operator() becomes a template.

  Since a template can represent infinite possible signatures, base::Bind doesn't know which one to pick to create the RepeatingCallback. It essentially says: "I see this is a template, but I need a concrete
  function signature to build the internal invoker."

PR #9870, opened 2026-04-01
Add UMA metric for media encoded buffer

tasunnn:
media: Add UMA metric for media encoded buffer memory

Introduce a new UMA histogram, "Memory.Media.AllocatedEncodedBuffer",
to track the memory consumption of media decoder buffers. This metric
will provide insight into media memory usage, aiding in performance
monitoring and optimization efforts, especially on platforms
utilizing Starboard media.

Test: 
```
autoninja -C out/linux-x64x11_debug cobalt_unittests && ./out/linux-x64x11_debug/cobalt_unittests 
--gtest_filter=CobaltMetricsServiceClientTest.*
```

Bug: 487768963

### DIFF
--- a/cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
+++ b/cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
@@ -591,9 +591,9 @@ TEST_F(CobaltMetricsServiceClientTest, RecordMediaMemoryMetricsHistogram) {
       histogram_tester.GetAllSamples("Memory.Media.AllocatedEncodedBuffer")
           .size(),
       1u);
-  EXPECT_GE(histogram_tester.GetBucketCount(
-                "Memory.Media.AllocatedEncodedBuffer", 2),
-            1);
+  EXPECT_GE(
+      histogram_tester.GetBucketCount("Memory.Media.AllocatedEncodedBuffer", 2),
+      1);
 #endif
 }
 


cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
```diff
@@ -414,6 +415,14 @@ void CobaltMemoryMetricsEmitter::CollateResults() {
   base::UmaHistogramMemoryLargeMB("Memory.Total.VmSize",
                                   static_cast<int>(vm_size_total_kb / kKiB));
 
+  // UMA metrics for media buffer memory usage
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
Awallky, line 419: Are there builds where we do not use starboard media stack?

tasunnn: I think all cobalt builds use it, have this here since media::MediaClient::GetMediaSourceTotalAllocatedMemory() is guarded by USE_STARBOARD_MEDIA

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
```diff
@@ -548,6 +551,41 @@ TEST_F(CobaltMetricsServiceClientTest, RecordMemoryMetricsRecordsHistogram) {
...
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  EXPECT_GE(
+      histogram_tester.GetAllSamples("Media.Memory.EncodedBuffer.Allocated")
+          .size(),
+      1u);
+  EXPECT_GE(histogram_tester.GetBucketCount(
+                "Media.Memory.EncodedBuffer.Allocated", 2),
+            1);
+  buffer = nullptr;
```
Awallky: I don't think this is necessary, since the buffer object should become null once we exit the scope of this test.

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
```diff
@@ -548,6 +551,41 @@ TEST_F(CobaltMetricsServiceClientTest, RecordMemoryMetricsRecordsHistogram) {
...
+  media::SetMediaClient(nullptr);
```
Awallky: Since this is cleanup for tests, can we instead add this to SetUp/TearDown methods for this test class? 
It looks like we currently implement the `SetUp` method already in this class. I don't think we necessarily need it in both.

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
Awallky: Ah, makes sense. Thanks for the info!

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
tasunnn: you are right, let me fix this

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
tasunnn: Ok I will add them to setup() and teardown() instead

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
```diff
@@ -414,6 +415,14 @@ void CobaltMemoryMetricsEmitter::CollateResults() {
...
+  uint64_t encoded_memory_bytes =
+      media::MediaClient::GetMediaSourceTotalAllocatedMemory();
+  base::UmaHistogramMemoryMB("Media.Memory.EncodedBuffer.Allocated",
```
yell0wd0g: micro-nit: There doesn't seem to be any other "Media.Memory.*"
UMAs, but there're plenty of "Memory.something". Should this 
be "Memory.Media.MediaSourceTotalAllocatedMemory" ?

yell0wd0g (approved):
LGTM with a naming comment.

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
tasunnn: there will be another Media.Memory.* histogram for decoded buffer but yeah it's probably better to follow the same naming convention for all memory metrics, I will change it to Memory.Media.* instead

yell0wd0g: Note that the CL description didn't update : )

tasunnn: just updated, thanks!

PR #9877, opened 2026-04-01
media: Add UMA metric for MediaCodec output buffer

tasunnn:
media: Add UMA metric for MediaCodec output buffer

Introduce a new UMA metric, `Memory.Media.EstimatedDecodedBuffer`, to track
the estimated memory consumed by Android MediaCodec output buffers.

This change integrates a new MediaCodecOutputTracker with MediaCodecBridge
instances. The tracker periodically reports the aggregated estimated memory
usage of all active decoded video frames based on their dimensions and the
number of active buffers, assuming YUV420 format.

The goal is to gain better visibility into the dynamic memory footprint of
decoded media on Android, aiding in performance analysis and optimization.

Bug: 487768963

### DIFF
--- a/.github/scripts/update_required_branch_checks.py
+++ b/.github/scripts/update_required_branch_checks.py
@@ -19,14 +19,14 @@ Requires PyGithub to run:
 """
 
 import argparse
-from typing import List
+import github
+import os
+import typing
 
-from github import Github
-
-# Issue a Personal Access Token with 'repo' permission on
-# https://github.com/settings/tokens.
-YOUR_GITHUB_TOKEN = ''
-assert YOUR_GITHUB_TOKEN != '', 'YOUR_GITHUB_TOKEN must be set.'
+_GITHUB_TOKEN = os.getenv('GITHUB_TOKEN', '')
+assert _GITHUB_TOKEN != '', (
+    'GITHUB_TOKEN must be set. Use `gh auth token` or issue a Personal Access'
+    'Token with \'repo\' permission on https://github.com/settings/tokens.')
 
 TARGET_REPO = 'youtube/cobalt'
 
@@ -38,6 +38,7 @@ EXCLUDED_CHECK_PATTERNS = [
     'assign-reviewer',
     'import/copybara',
     'upload-release-artifacts',
+    'generate_commit_message',
 
     # Excludes coverage and test reports.
     'linux-coverage',
@@ -65,7 +66,7 @@ MINIMUM_LTS_RELEASE = 19
 LATEST_LTS_RELEASE = 25
 
 
-def get_protected_branches() -> List[str]:
+def get_protected_branches() -> list[str]:
   branches = ['main']
   for i in range(MINIMUM_LTS_RELEASE, LATEST_LTS_RELEASE + 1)[:-1]:
     branches.append(f'{i}.lts.1+')
@@ -73,11 +74,11 @@ def get_protected_branches() -> List[str]:
 
 
 def initialize_repo_connection():
-  g = Github(YOUR_GITHUB_TOKEN)
+  g = github.Github(auth=github.Auth.Token(_GITHUB_TOKEN))
   return g.get_repo(TARGET_REPO)
 
 
-def get_checks_for_branch(repo, branch: str) -> None:
+def get_checks_for_branch(repo, branch: str) -> typing.Iterable[typing.Any]:
   # The 'merged' sort order is not listed in public docs but still works.
   # If this functionality is removed the alternative is to loop through all
   # PRs and use the 'merged_at' property to determine which is the latest one.
@@ -85,15 +86,15 @@ def get_checks_for_branch(repo, branch: str) -> None:
   prs = repo.get_pulls(
       state='closed', sort='merged', base=branch, direction='desc')
 
-  latest_pr = None
   for pr in prs:
     if pr.merged:
-      latest_pr = pr
-      break
+      latest_pr_commit = repo.get_commit(pr.head.sha)
+      checks = latest_pr_commit.get_check_runs()
+      checks_complete = all(c.status == 'completed' for c in checks)
+      if checks_complete:
+        return checks
 
-  latest_pr_commit = repo.get_commit(latest_pr.head.sha)
-  checks = latest_pr_commit.get_check_runs()
-  return checks
+  raise RuntimeError(f'Could not find any completed checks for branch {branch}')
 
 
 def should_include_run(check_run) -> bool:
@@ -103,14 +104,14 @@ def should_include_run(check_run) -> bool:
   return True
 
 
-def get_required_checks_for_branch(repo, branch: str) -> List[str]:
+def get_required_checks_for_branch(repo, branch: str) -> list[str]:
   checks = get_checks_for_branch(repo, branch)
   filtered_check_runs = [run for run in checks if should_include_run(run)]
   check_names = set(run.name for run in filtered_check_runs)
   return list(check_names)
 
 
-def print_checks(repo, branch_name: str, new_checks: List[str],
+def print_checks(repo, branch_name: str, new_checks: list[str],
                  print_unchanged: bool) -> None:
   branch = repo.get_branch(branch_name)
   current_checks = branch.get_required_status_checks().contexts
@@ -137,7 +138,7 @@ def print_checks(repo, branch_name: str, new_checks: List[str],
 
 
 def update_protection_for_branch(repo, branch: str,
-                                 check_names: List[str]) -> None:
+                                 check_names: list[str]) -> None:
   branch = repo.get_branch(branch)
   branch.edit_required_status_checks(contexts=check_names)
 


cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -114,7 +117,64 @@ public void onNewFrame(long presentationTimeUs) {
     }
   }
 
+  private static AtomicLong sGlobalTotalMemoryUsage = new AtomicLong(0L);
+  @CalledByNative
+  public static long getGlobalOutputMemoryUsage() {
+    return sGlobalTotalMemoryUsage.get();
+  }
+
+  private class OutputMemoryTracker {
```
jasonzhangxx: Let's move the memory tracker to a separated class to decouple it with MediaCodecBridge.

As the memory usage is estimated based on buffered output frames and output format, we can add two functions in MediaCodecBridge, `MediaFormat GetCurrentMediaFormat()` and `int SizeOfActiveOutputBuffers()`, with related code in MediaCodecBridge to provide these information. Then we can use the return value of these two functions to estimate the memory usage of buffered outputs. Also, to be able to access MediaCodecBridge instances, we need register the instance in `createVideoMediaCodecBridge()` to the new singleton OutputMemoryTracker class (I think right now we only care about the memory usage of video), and unregister the instance in `release()` function.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -662,6 +726,11 @@ public void release() {
       mNativeMediaCodecBridge = 0;
     }
 
+    // Collect output buffer total memory usage before releasing
+    RecordHistogram.recordMemoryMediumMBHistogram(
+        "Media.Memory.DecodedBuffer.Allocated",
+        (int)(sGlobalTotalMemoryUsage.get() / BYTES_PER_MIB));
```
jasonzhangxx: If we only report memory usage on release, my understanding is the reported value should always be 0.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,113 @@
+// Copyright 2017 The Cobalt Authors. All Rights Reserved.
```
Awallky: Please update the copyright year.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,113 @@
...
+  private final Set<MediaCodecBridge> mBridges =
+      Collections.synchronizedSet(Collections.newSetFromMap(new WeakHashMap<>()));
+
+  private final SequencedTaskRunner mTaskRunner =
+      PostTask.createSequencedTaskRunner(TaskTraits.BEST_EFFORT);
+  private final Runnable mReportRunnable =
+      new Runnable() {
+        @Override
+        public void run() {
+          synchronized (MediaCodecOutputTracker.this) {
```
Awallky: The mix of synchronized methods (register, unregister) and manual synchronized (MediaCodecOutputTracker.this) in the Runnable is inconsistent. Prefer using a single synchronization strategy (e.g., a final lock object).

See example usage here: [definition](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/compositor/CompositorView.java;drc=eb1b4c6a26bdb372fc55844b73f5b21a0f38b674;l=103), [usage1](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/compositor/CompositorView.java;drc=eb1b4c6a26bdb372fc55844b73f5b21a0f38b674;l=529), [usage2](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/compositor/CompositorView.java;drc=eb1b4c6a26bdb372fc55844b73f5b21a0f38b674;l=547), [usage3](https://source.chromium.org/chromium/chromium/src/+/main:chrome/android/java/src/org/chromium/chrome/browser/compositor/CompositorView.java;drc=eb1b4c6a26bdb372fc55844b73f5b21a0f38b674;l=756)

Awallky: Ensure that every method that reads from or writes to `mBridges` uses the same lock. Currently, your code inconsistently switches between synchronized methods (locking on this) and synchronized (mBridges).

tasunnn: that's a good point, I will update synchronization strategy to make it more consistent. But the reason for using multiple locks, specifically synchronized (mBridges) in getTotalOutputMemoryUsage(), is because I defined mBridges as a Collections.synchronizedSet(), this type requires the synchronized (mBridges) lock when iterating the entire set and hence using a different lock there instead of locking on this.

Awallky: In that case, it sounds like we can simplify things if we just use a regular `Set` and guard against races with the lock. WDYT?

tasunnn: yeah that's a good idea, just updated the code

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,118 @@
...
+ * Memory tracker for MediaCodecBridge instances to estimate video memory usage.
+ * Reports metrics periodically (1 minute) while bridges are active.
+ */
+public class MediaCodecOutputTracker {
+  private static MediaCodecOutputTracker sInstance;
+  private static final int BYTES_PER_MIB = 1024 * 1024;
+  private static final long DEFAULT_REPORT_INTERVAL_MS = 60000; // 1 minute
+  private static final Object mTrackerLock = new Object();
+
+  private final Set<MediaCodecBridge> mBridges = new HashSet<>();
```
Awallky: I recommend swapping the HashSet back to `WeakHashMap` in order to allow the bridges to be GC'd.
Using something like `Collections.newSetFromMap` should do the trick.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Memory tracker for MediaCodecBridge instances to estimate video memory usage.
+ * Reports metrics periodically (1 minute) while bridges are active.
+ */
+public class MediaCodecOutputTracker {
+  private static MediaCodecOutputTracker sInstance;
+  private static final int BYTES_PER_MIB = 1024 * 1024;
+  private static final long DEFAULT_REPORT_INTERVAL_MS = 60000; // 1 minute
+  private static final Object mTrackerLock = new Object();
```
Awallky: Per the style guide, we should replace the `m` with `s` for a static field name - https://source.android.com/docs/setup/contribute/code-style#follow-field-naming-conventions

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,118 @@
...
+
+  private final Set<MediaCodecBridge> mBridges = new HashSet<>();
+
+  private final SequencedTaskRunner mTaskRunner =
```
Awallky, line 38: Line wraps in Java should use 8-space indentation - https://source.android.com/docs/setup/contribute/code-style#use-spaces-for-indentation.

Also, Blocks should use 4-space indentation in Java.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  private void stopReporting() {
+    mIsReporting = false;
+  }
+
+  private long getTotalOutputMemoryUsage() {
+    synchronized (mTrackerLock) {
+      long totalMemory = 0;
+      for (MediaCodecBridge bridge : mBridges) {
```
Awallky: To avoid concurrent modification exceptions, instead of directly iterating over the mBridges, we should make a copy of it into a local array and iterate over the copy. 
This is because we could run into races where the `MediaCodecBridge` objects are GC'd while we still hold a reference to them and are actively iterating over them.

Awallky (approved):
This mostly LGTM % comments.
Once this lands, could we add some junit tests for the new class?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
tasunnn: I think bridges should be GC when release is called in MediaCodecBridge, the place where we also remove it from set, is a WeakHashMap still necessary in this case?

Awallky: > I think bridges should be GC when release is called in MediaCodecBridge

I don't think so. 
An item is only able to be GC'd if no strong references are left. The `HashSet` maintains a strong reference to the objects in the container. Since our `MediaCodecOutputTracker` class is a singleton, it will persist throughout the lifetime of the Cobalt process. This means the MediaCodeBridges would only be GC'd after Cobalt is killed. 

So, I think the WeakHashMap is necessary.

Awallky: Actually, in the success case, I think you are right. 
However, in the failure cases, it will not. In the exception cases, it would not. Having a weak reference would fix this leak for us.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -115,6 +116,19 @@ public void onNewFrame(long presentationTimeUs) {
   }
 
   private FrameRateEstimator mFrameRateEstimator = null;
+  private final AtomicInteger mActiveOutputBuffers = new AtomicInteger(0);
+  private volatile MediaFormatWrapper mActiveFormat = null;
```
jasonzhangxx, line 120: nit: `volatile` here is not necessary, but it's ok to keep it as is.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -115,6 +116,19 @@ public void onNewFrame(long presentationTimeUs) {
...
+
+  int getCurrentMediaFormatDimension() {
+    if (mActiveFormat == null) {
+      return 0;
+    }
+    return mActiveFormat.width() * mActiveFormat.height();
```
jasonzhangxx, line 126: nit: it is ok to keep it as is, but a function returning the active format directly would be a little better, in case if you may want to change the calculation of frame size, you'll not have to come back to modify MediaCodecBridge.java.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
tasunnn: that makes sense, just fixed it

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
```diff
@@ -0,0 +1,123 @@
...
+        synchronized (sTrackerLock) {
+          if (!mIsReporting || mBridges.isEmpty()) {
+            mIsReporting = false;
+            return;
+          }
+        }
+        reportMetrics();
+        postReportTask();
+      },
+      DEFAULT_REPORT_INTERVAL_MS);
```
jasonzhangxx, line 84: @Awallky, I don't recommend to periodically report all metrics, which may block threads and waste some performance. I recommend enabling metrics reporting only for non-gold builds and a select group of public users who would upload the metrics to our server to facilitate analysis, allowing optimal performance for the majority of users. This comment might not be directly related to this PR though.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
tasunnn: Yeah I agree, but MediaFormatWrapper is a private class in MediaCodecBridge, if we want to use width and height methods to do the calculations inside the tracker, we would then need to expose the class

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
Awallky: > I don't recommend to periodically report all metrics, which may block threads and waste some performance. 

I disagree that this is necessarily detrimental to performance. In Chromium, recording a histogram is designed to be nearly free. If a specific metric recording is found to be expensive (e.g., it requires a complex string calculation or a lock), that’s a bug in the instrumentation, not the system. 
Now, if you have reservations there, then I think we should consider how we are recording these metrics and whether there is a more efficient (in time or space) way to record these.

While we do not want to spend too much time blocking threads, this can easily be mitigated by only reporting metrics every 5 minutes, 10 minutes, or something configurable that we feel comfortable with. This recording period can be configured via a Finch feature flag parameter.

> I recommend enabling metrics reporting only for non-gold builds and a select group of public users who would upload the metrics to our server to facilitate analysis, allowing optimal performance for the majority of users. This comment might not be directly related to this PR though.

I agree, this one is outside of the scope of this PR. If you have serious reservations about using these in production, we should discuss offline with the relevant Cobalt-UMA folks.

tasunnn: @jasonzhangxx following up on concerns regarding this PR specifically, if we want to keep the current way of reporting metrics for this output buffer, we could adjust the time interval to be greater (right now it's 1 minute) or make it configurable to mitigate blocking. If we don't want to do the periodic reporting, is there a more efficient way that you suggest we could record the metrics here?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
tasunnn: sounds good made the change

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
tasunnn: Keep this as is

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecOutputTracker.java
tasunnn: synced offline, PR is fine as is, I changed default time interval to 5 minutes since 1 minute might be too frequent

johnxwork (approved):

PR #9878, opened 2026-04-01
android: Skip Java crashpad handler on old ATV

alunev:
Skip the Java fallback path for Crashpad's hang handler startup when
running Cobalt on Android TV. This avoids complex JNI/prefixing issues
on Android API levels 24-28. Modern Android (API 29+) is unaffected by
this change as it does not use the Java fallback.

Bug: 484122527

### DIFF
--- a/components/crash/core/app/crashpad_android.cc
+++ b/components/crash/core/app/crashpad_android.cc
@@ -565,6 +565,14 @@ class HandlerStarter {
     }
 
     if (use_java_handler_ || !handler_trampoline_.empty()) {
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROIDTV)
+      // Cobalt on Android TV skips the Java fallback handler for API 24-28.
+      // Reporting is not supported on these versions due to engineering cost
+      // and low traffic.
+      if (use_java_handler_) {
+        return database_path;
+      }
+#endif
       std::vector<std::string> env;
       if (!BuildEnvironmentWithApk(kUse64Bit, &env)) {
         return database_path;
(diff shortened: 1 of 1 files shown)


components/crash/core/app/crashpad_android.cc
```diff
@@ -565,6 +565,13 @@ class HandlerStarter {
     }
 
     if (use_java_handler_ || !handler_trampoline_.empty()) {
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROIDTV)
+      // For cobalt on atv we skip java fallback that is used on android api
+      // levels in [24...28]
+      if (use_java_handler_) {
+        return database_path;
+      }
+#endif
```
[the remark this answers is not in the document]
alunev: given use_java_handler_ is envolved I think it's more readable with duplication.

components/crash/core/app/crashpad_android.cc
```diff
@@ -565,6 +565,13 @@ class HandlerStarter {
     }
 
     if (use_java_handler_ || !handler_trampoline_.empty()) {
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROIDTV)
+      // For cobalt on atv we skip java fallback that is used on android api
+      // levels 24-28
```
hlwarriner: Please just briefly mention the _why_, for example, "Given the engineering cost and relatively low traffic we do not support Crashpad-based reporting on these older android api versions." Someone may read this comment and assume there is some other fallback that will provide support for these older versions.

alunev: thanks, added clarification.

PR #9157, opened 2026-02-18
Use git hash from chromium tag in LASTCHANGE file

andrewsavage1:
This ensures official builds (qa/gold) of Cobalt use the right version for devtools connections.

Bug: 484467425

PR #9879, opened 2026-04-02
cobalt/LASTCHANGE.chromium: correct the commit hash format

santoshmahto7:
The `build/util:chromium_git_revision` expect the value of LASTCHANGE
in thr `X-Y` format since it uses '-' to extract commit hash value.
Not using the above format broke the devtools frontend loading,
since while extracting the commit last character of commit hash was missed.

Bug: 496585663


andrewsavage1:
This is the content of my `//build/util/LASTCHANGE`, the file previously used by chromium (before https://github.com/youtube/cobalt/pull/9157) to insert the hash into the devtools link:
```
LASTCHANGE=e37b45a78bbeb89c4c8c4b48f0c0a91b319e8116
LASTCHANGE_YEAR=2026
```
It does not have the dash format. Why do we need that format now?

Whatever we do, we would also need to update the github workflow as was done in my previous PR.

Also, we should submit change to main and for now they'll be automatically be cherry picked to 27.lts—don't submit to 27.lts directly

santoshmahto7:
The content of /build/util/LASTCHANGE is also wrong. I think it is generated while sync since it is not part of git.

The issue was that last character of commit hash was missed while parsing by earlier `cobalt/LASTCHANGE.chromium`
The reason is as:
The commit has is extracted from `LASTCHANGE=hash` as:
` "LASTCHANGE=LASTCHANGE[:LASTCHANGE.find('-')]",`
https://source.chromium.org/chromium/chromium/src/+/refs/tags/138.0.7204.305:build/util/BUILD.gn;l=25

 So if '-' isn't found then `LASTCHANGE.find('-')` return -1 and above code becomes LASTCHANGE=LASTCHANGE[:-1] , i.e indexing from [0, last) but WITHOUT last character.

If you load the devtools  and notice the commit hash, the last value is missed.
btw, in my chromium repo (latest main)  
```
/srv/santoshR/chromium/src$ cat  build/util/LASTCHANGE
LASTCHANGE=63b2fd2c994a192cfbc79e490509fb281c7b9cbc-refs/heads/main@{#1513459}
LASTCHANGE_YEAR=2025
```

andrewsavage1:
Okay, let's also update the sync chromium branches workflow

niranjanyardi (approved):

andrewsavage1 (commented):
We can submit this as is to fix devtools but we should make sure this matches the python script output and update sync chromium branches job in the future

santoshmahto7:
> We can submit this as is to fix devtools but we should make sure this matches the python script output and update sync chromium branches job in the future

I will create a follow-up task to fix the "sync chromium action" separately. Thanks for letting it land

PR #9895, opened 2026-04-03
Support in-memory CRX updates

yuying-y:
Modify the update client's download and result handling to pass the pointer to the crx to the following operations when IN_MEMORY_UPDATES is defined. It also changes the macro guard for download_result.response in accordance with the definition of CrxDownloader::Result

Issue: 479816505

### DIFF
--- a/components/update_client/op_download.cc
+++ b/components/update_client/op_download.cc
@@ -44,8 +44,6 @@ constexpr int64_t kBackgroundDownloadSizeThreshold = 10'000'000; /*10 MB*/
 constexpr int64_t kBackgroundDownloadSizeThreshold = 0;
 #endif
 
-
-
 #if !BUILDFLAG(IS_STARBOARD)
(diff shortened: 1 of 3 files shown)


components/update_client/op_download.cc
```diff
@@ -195,7 +200,11 @@ void HandleAvailableSpace(
       crx_str,
 #endif
       base::BindOnce(&DownloadComplete, crx_downloader, cancellation,
+#if defined(IN_MEMORY_UPDATES)
```
y4vor, line 203: Why do we still have compile time flag IN_MEMORY_UPDATES?
We would not have different builds. If we need to keep this optional it should be a runtime flag.

yuying-y: There are multiple references in the code base. filed b/500107040 and added to the c27 hotlist for the migration.

PR #9783, opened 2026-03-27
build: redirect partner toolchains to copied_base

hferreiro:
Partner toolchains should use //copied_base/base instead of //base. This
PR adds a new GN variable `is_partner_toolchain` and uses it to redirect
`//base` to `//copied_base/base` when appropriate.

Additionally, simplify existing conditions by using the new variable.

Bug: 495203133

PR #9902, opened 2026-04-06
build: Replace is_native_target_build with is_native_target

hferreiro:
Replace `is_native_target_build` with the existing `is_native_target`
variable which is defined as current_toolchain == native_target. This
eliminates the need to manually set it in every toolchain definition.
Also rename the C++ buildflag from `IS_NATIVE_TARGET_BUILD` to
`IS_NATIVE_TARGET` for consistency.

Bug: 492403745


hferreiro:
@andrewsavage1 technically, this PR could use https://issuetracker.google.com/406511608 but it's not really that much related. Shall I create a new issue? Or assign to https://issuetracker.google.com/492403745?

cobalt/build/configs/initialize_variables.gni
```diff
@@ -39,9 +35,6 @@ declare_args() {
              !use_evergreen && !use_tsan
 }
 
-assert(!(is_starboard && is_native_target_build),
```
hferreiro, line 42: Is this assertion worth keeping?

andrewsavage1: Yes

andrewsavage1 (approved):
It seems like this is okay because is_native_target is never used in BUILDCONFIG.gn or GNI files imported from the buildconfig.

cobalt/build/configs/initialize_variables.gni
andrewsavage1: That said, we should probably avoid doing it in a GNI file imported by the buildconfig. Maybe it belongs in starboard/BUILD.gn

andrewsavage1:
> @andrewsavage1 technically, this PR could use https://issuetracker.google.com/406511608 but it's not really that much related. Shall I create a new issue? Or assign to https://issuetracker.google.com/492403745?

Using 492403745 seems okay

cobalt/build/configs/initialize_variables.gni
hferreiro: What about starboard/build/config/toolchain_variables.gni, where `is_native_target` is defined?

third_party/crashpad/crashpad/util/net/tls.gni
```diff
@@ -14,12 +14,16 @@
...
+  import("//starboard/build/config/toolchain_variables.gni")
+}
+
 declare_args() {
   # Cobalt comment: we're still exploring, in b/446904324, whether we can use
   # libcurl. For now will keep the c25 status quo and use raw sockets with
   # BoringSSL as the backend.
   if (crashpad_is_fuchsia || crashpad_is_android ||
-          (is_cobalt && is_native_target_build)) {
+          (is_cobalt && is_native_target)) {
```
hlwarriner, line 26: I wasn't a reviewer on #9783 but I feel `is_native_target` would be more clear and more consistent with related build variables if it were renamed to `is_native_target_toolchain` (same thing with the corresponding buildflag).

@andrewsavage1, wdyt?

If so, we can do a quick renaming after this current PR is merged.

hlwarriner (approved):
Nice cleanup!

cobalt/build/configs/initialize_variables.gni
hferreiro: Readded with a reworded message. Let me know if toolchain_variables.gni is a good place.

third_party/crashpad/crashpad/util/net/tls.gni
hferreiro: Agreed. It was my mistake not to do that in the first place. I can do a quick follow-up.

hlwarriner: Sounds good, thanks!

hferreiro: I wonder if it would make sense to rename to `is_native_toolchain`, and rename the `native_target` toolchains to use `native` instead?

hlwarriner: I don't feel strongly about that. @andrewsavage1?

andrewsavage1: Maybe the toolchain should be called `native_toolchain` to be in line with `host_toolchain`, `starboard_toolchain`, `default_toolchain`

cobalt/build/configs/initialize_variables.gni
andrewsavage1: Looks good to me

PR #9904, opened 2026-04-06
tvOS: Call contentKeySessionWithKeySystem without storageDirectoryAtURL

rakuco:
Follow-up to #9840. There is one call that only occurs in device builds that was triggering a warning and now causing the build to fail: the `storageDirectoryAtURL` parameter must not be nil when specified. Switch to the other method signature that does not take this parameter instead.

Bug: 432508012


rakuco:

fayev-g (approved):

PR #9911, opened 2026-04-06
In-memory crx verification

yuying-y:
Modify the unpacker and verifier to verify and unzip CRX files
directly from an in-memory buffer. Changes mostly ported from https://github.com/youtube/cobalt/pull/780

Issue: 479816505


components/crx_file/crx_verifier.cc
hlwarriner, line 1: Please mention in the PR description that the PR mostly ports forward changes from #780. That might be interesting context for a future member of the team who is looking at this PR for whatever reason.

PR #9448, opened 2026-03-10
media: Add Starboard microphone input support (#8554)

sideb0ard:
Cherry-picked from: https://github.com/youtube/cobalt/pull/8554

This change introduces the necessary components to enable microphone input functionality for Cobalt applications running on Starboard platforms.

It adds AudioManagerStarboard and AudioInputStreamStarboard to handle the creation and management of audio input streams through the Starboard microphone API. The build system is updated to include these new audio manager components.

For Linux-based Starboard implementations, a concrete implementation of the SbMicrophone API is provided, leveraging ALSA for discovery of available input devices. This replaces the previous stub implementations for microphone functionality, allowing for actual microphone interaction.

The core audio manager selection logic is updated to prioritize the Starboard audio manager when BUILDFLAG(USE_STARBOARD_MEDIA) is enabled.

The Starboard implementation of SyncSocket::Peek() is updated to be non-blocking.

Sample rate and number of channels are hard-coded and match Android implementation.

Tested locally on Linux-x64x11 and Evergreen-x64

Bug: 394365678

PR #9913, opened 2026-04-06
media: minor follow up to microphone implementation

sideb0ard:
Cherry-pick gemini review flagged a potential log flood.

Also address header include nit from previous PR (https://github.com/youtube/cobalt/pull/9448)

Bug: 394365678


base/sync_socket_posix.cc
```diff
@@ -174,9 +175,12 @@ size_t SyncSocket::Peek() {
   ssize_t number_chars = recv(handle_.get(), buffer, sizeof(buffer),
```
kjyoun, line 175: How about `HANDLE_INTR(recv...` ?

It would handle EINTR status

e.g. ) https://source.corp.google.com/h/github/youtube/cobalt/+/main:base/sync_socket_posix.cc;l=208;drc=e551786cc03840e5d3045ac9e6d09b650008924c

kjyoun:
> Also address header include nit from previous PR (https://github.com/youtube/cobalt/pull/9448)

Can you update bug description with why we are check WOULD_BLOCK?

It would help the person like me. I had to spend time to read doc to understand this PR.

e.g.)

```
    starboard: Suppress false error logs in SyncSocket::Peek()

    When no data is available on the socket, recv() correctly returns -1
    and sets errno to EAGAIN or EWOULDBLOCK, which is not an error. Previously,
    this was reported as an error, spamming the log.

    This change filters out these non-blocking statuses so errors are only
    logged on genuine socket failures.
```

PR #9918, opened 2026-04-07
ci: Use .netrc for GoB authentication

oxve:
This PR uses a .netrc file for authenticating with googlesource.com. This is a standard way to provide credentials for network logins and might be more compatible with some environments than .gitcookies or credential helpers.


oxve:
Closing in favor of PR 9917.

PR #9919, opened 2026-04-07
ci: Use http.extraheader for GoB authentication

oxve:
This PR uses http.extraheader with a Bearer token for authenticating with googlesource.com. This is often the most direct and reliable way to provide OAuth tokens to Git.


oxve:
Closing in favor of PR 9917.

PR #9921, opened 2026-04-07
Update pdfium submodule to match DEPS revision

hferreiro:
https://github.com/youtube/cobalt/pull/9833 updated pdfium's DEPS entry
but forgot to update the submodule. This results in git always showing
the submodule as dirty.

Bug: 481788398

### DIFF
--- a/.github/actions/depot_tools/action.yaml
+++ b/.github/actions/depot_tools/action.yaml
@@ -64,13 +64,8 @@ runs:
         set -x
         cd cobalt
 
-        CACHE_ARG=()
-        if [[ -n "${CACHE_PATH}" ]]; then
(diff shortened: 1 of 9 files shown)


hferreiro:
@andrewsavage1 sorry to repeat myself but, is there a way to bypass the Bug ID Check for this kind of fixes? I would reuse the bug number of https://github.com/youtube/cobalt/pull/9833 but it doesn't have one.

andrewsavage1:
> @andrewsavage1 sorry to repeat myself but, is there a way to bypass the Bug ID Check for this kind of fixes? I would reuse the bug number of #9833 but it doesn't have one.

@briantting shouldn't the rebase refresh PRs have bug numbers associated with them? Also, we'll need to fix the automated rebases if they're going to break submodules like this

@hferreiro I added a bug number we can use for this one.

PR #9923, opened 2026-04-07
build: Introduce android_toolchain for decoupling from default_toolchain

hferreiro:
For AOSP builds, the default_toolchain is a Linux/Evergreen toolchain
where is_android=false, but JNI and other Android-specific targets
use `default_toolchain` to mean the Android toolchain.

Introduce android_toolchain in
starboard/build/config/toolchain_variables.gni: a variable that
identifies which toolchain should be used for Android targets. For
Cobalt modular builds with separate toolchains it resolves to
`starboard_toolchain`; for all other builds it remains
`default_toolchain`.

Bug: 495203133


build/android/BUILD.gn
```diff
@@ -10,6 +10,7 @@ import("//build/config/coverage/coverage.gni")
 import("//build/config/gclient_args.gni")
 import("//build/config/python.gni")
 import("//build_overrides/build.gni")
+import("//starboard/build/config/android_toolchain.gni")
```
andrewsavage1, line 13: Looks like you forgot to add this file to the commit?

hferreiro: Ups, fixed

third_party/jni_zero/jni_zero.gni
```diff
@@ -38,7 +39,7 @@ if (target_os == "android") {
   jni_headers_dir = "$root_build_dir/gen/jni_headers"
 } else {
   # Chrome OS builds cannot share gen/ directories because is_android=false
-  # within default_toolchain.
```
hferreiro, line 41: Not very relevant, but it's not clear to me if this comment is correct. I guess jni_zero is also used by ChromeOS, and in that context, the name `android_toolchain` doesn't make sense.

tools/android/forwarder2/BUILD.gn
```diff
@@ -4,17 +4,18 @@
 
 import("//build/symlink.gni")
 import("//testing/test.gni")
+import("//starboard/build/config/android_toolchain.gni")
```
hferreiro, line 6: Is there any guidelines on how to format this? (i.e. unconditionally imported but not alphabetically sorted here).

base/BUILD.gn
```diff
@@ -1394,10 +1394,10 @@ component("base") {
...
     ]
     deps += [
-      ":base_minimal_jni($default_toolchain)",
-      ":command_line_jni($default_toolchain)",
-      ":library_loader_jni($default_toolchain)",
-      ":tasks_minimal_jni($default_toolchain)",
+      ":base_minimal_jni($android_toolchain)",
+      ":command_line_jni($android_toolchain)",
+      ":library_loader_jni($android_toolchain)",
+      ":tasks_minimal_jni($android_toolchain)",
```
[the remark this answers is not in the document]
hferreiro: Obsolete comment.

starboard/build/config/android_toolchain.gni
```diff
@@ -0,0 +1,27 @@
...
+
+# Identifies the toolchain that should be used for Android-specific targets
+# (JNI code generation, Android tools, etc.).
+# This variable is meant to replace `default_toolchain` in upstream Android GN
+# files where it is used to avoid duplicate builds by secondary toolchains.
+# In Cobalt modular builds with separate toolchains, the `default_toolchain` is
+# a Linux/Evergreen toolchain where is_android=false. Android targets must be
+# evaluated in the starboard toolchain where current_os="android".
+if (is_cobalt && build_with_separate_cobalt_toolchain &&
+    !(target_os == "android" && !use_evergreen)) {
```
hferreiro: @andrewsavage1 This last condition is worth a comment, but I wanted to double-check first. The issue is that when building AndroidTV, in the host toolchain `is_androidtv=false` and then `build_with_separate_cobalt_toolchain=true`.

build/android/BUILD.gn
```diff
@@ -273,7 +274,7 @@ group("stack_tools") {
 
 # GN evaluates each .gn file once per toolchain, so restricting to default
```
andrewsavage1: We should update this comment too

starboard/build/config/android_toolchain.gni
andrewsavage1: Ah, we should probably change the default value of build_with_separate_cobalt_toolchain to use target_os and use_evergreen directly instead of is_androidtv, as it would make more sense there than it does here

starboard/build/config/android_toolchain.gni
```diff
@@ -0,0 +1,27 @@
...
+  android_toolchain = starboard_toolchain
```
andrewsavage1, line 23: I think for the best upstreaming chance we should structure this like the `v8/gni/snapshot_toolchain.gni` flow, i.e. declare_args the toolchain with the empty string to allow overrides, then set it based on conditionals like you already do if it was not overridden via toolchain_args/args.gn

tools/android/forwarder2/BUILD.gn
andrewsavage1: Probably just make it alphabetical to appease any future formatters/linters—hopefully we can upstream it and not have to worry about it

starboard/build/config/android_toolchain.gni
hferreiro: I can do that only for the upstream CL. Would there be any benefit in doing it like that here?

andrewsavage1: No, it's fine to do it just for the upstream CL

third_party/jni_zero/jni_zero.gni
hferreiro: @andrewsavage1 I'll revert this change.

tools/android/forwarder2/BUILD.gn
hferreiro: Done.

starboard/build/config/android_toolchain.gni
hferreiro: Done.

build/android/BUILD.gn
hferreiro: Done.

cobalt/build/configs/initialize_variables.gni
```diff
@@ -29,7 +29,10 @@ declare_args() {
 
   using_old_compiler = false
 
-  build_with_separate_cobalt_toolchain = !is_androidtv && target_os != "ios"
+  build_with_separate_cobalt_toolchain =
+      # Avoid using `is_androidtv` which uses `current_os`, so that this is also
+      # false for the host toolchain.
+      !(target_os == "android" && !use_evergreen) && target_os != "ios"
```
andrewsavage1: nit: I think this would be slightly easier to read as `(target_os != "android" || use_evergreen) && target_os != "ios"`

andrewsavage1 (approved):

cobalt/build/configs/initialize_variables.gni
hferreiro: Done.

PR #9924, opened 2026-04-07
player: Stop defining kLogIntervalUs in release builds

rakuco:
This constant is only used inside a `!BUILDFLAG(COBALT_IS_RELEASE_BUILD)` block, which makes e.g. a gold build on tvOS fail with

    ../../starboard/shared/starboard/media/decoder_state_tracker.cc:47:19: error: unused variable 'kLogIntervalUs' [-Werror,-Wunused-const-variable]

Bug: 432508012


rakuco:

PR #9941, opened 2026-04-08
tvos: Add browser tests for IFA

Gyuyoung:
Add browser tests to verify the GetAdvertisingID and GetLimitAdTracking
Web APIs for IFA on tvOS.

The native tvOS APIs to retrieve advertising ID and tracking authorization
status do not function correctly on the simulator. To enable testing in
this environment, test hooks are introduced to simulate API responses,
allowing comprehensive validation of the Web APIs.

Bug: 453538923

### DIFF
--- a/cobalt/browser/h5vcc_system/h5vcc_system_impl_base.h
+++ b/cobalt/browser/h5vcc_system/h5vcc_system_impl_base.h
@@ -38,6 +38,12 @@ class H5vccSystemImpl : public content::DocumentService<mojom::H5vccSystem> {
   static void Create(content::RenderFrameHost* render_frame_host,
                      mojo::PendingReceiver<mojom::H5vccSystem> receiver);
 
+#if BUILDFLAG(IS_IOS_TVOS)
+  static void ResetTestState();
+  static void SetTrackingAuthorizationForTest(bool authorized);
+  static void SetAdvertisingIdForTest(const std::string& id);
+#endif
+
   H5vccSystemImpl(const H5vccSystemImpl&) = delete;
   H5vccSystemImpl& operator=(const H5vccSystemImpl&) = delete;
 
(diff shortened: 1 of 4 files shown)


Gyuyoung:
CC'ing @abhijeetk and @jkim-julie.

rakuco (commented):
Hmm, these `#ifdef`s in platform-independent files don't look great, and I don't think these tests even need to be tvOS-specific since these APIs are also implemented for other ports.

I'm also not sure the tests are that meaningful in this form, as they're testing how the Blink side reacts to the backend, but the backend itself is basically being replaced by a mock.

This looks like the kind of thing that would be better served by a web test that replaces the browser side of the API with a JS mock, but I understand we're unfortunately not there with web tests on tvOS. Perhaps trying to fix the tests on YTS would make more sense (dunno if we are even able to do that though)?

cc @andrewsavage1 too

Gyuyoung:
> Hmm, these `#ifdef`s in platform-independent files don't look great, and I don't think these tests even need to be tvOS-specific since these APIs are also implemented for other ports.

> I'm also not sure the tests are that meaningful in this form, as they're testing how the Blink side reacts to the backend, but the backend itself is basically being replaced by a mock.
> 
> This looks like the kind of thing that would be better served by a web test that replaces the browser side of the API with a JS mock, but I understand we're unfortunately not there with web tests on tvOS. Perhaps trying to fix the tests on YTS would make more sense (dunno if we are even able to do that though)?

Yes, I agree with your point. The main reason I added these tests was that I wasn't sure when the YTS tests could be updated, and the tvOS port does not support web tests yet.

Given the current limitations, I thought it might still be helpful to have a test that at least verifies the Web API code path from Blink to Cobalt, even though it relies on mocks. But, if others also think the tests are not meaningful, let me abandon this PR :) 

> cc @andrewsavage1 too

rakuco:
The YTS code was open-sourced a few hours ago :tada: We could try to update the IFA tests there: https://github.com/youtube/yts/blob/main/test_modules/web_tests/ifa_test.ts

(getting the new YTS to run on the Apple TV is still WIP though)

Gyuyoung:
Nobody has commented. So I close this PR and am trying to update the yts test. Thanks.

PR #9943, opened 2026-04-08
Add wrapper for getrandom

sacuff:
This PR adds an ABI wrapper for the `getrandom` function.

`getrandom` is usually implemented as a syscall, so we want to use the platform implementation. The wrapper handles translation of the `flags` bits between musl and the platform.

Bug: 500412520


sacuff:
Wondering if the test should move up to the `nplb` dir since it's not actually a POSIX function

andrewsavage1:
Note we'll definitely need a wrapper for getrandom, at the very least for the flags it uses—we'll need to translate between musl and the platform layer

jellefoks (approved):
Nice!

starboard/shared/modular/starboard_layer_posix_getrandom_abi_wrappers.cc
```diff
@@ -0,0 +1,36 @@
...
+#include <errno.h>
+#include <sys/random.h>
+
+ssize_t __abi_wrap_getrandom(void* buf, size_t buflen, unsigned flags) {
+  unsigned platform_flags = 0;
+  if (flags & MUSL_GRND_NONBLOCK) {
+    platform_flags |= GRND_NONBLOCK;
+    flags &= ~MUSL_GRND_NONBLOCK;
+  }
+  if (flags & MUSL_GRND_RANDOM) {
```
jellefoks, line 26: Can you add a test that uses this flag?

jellefoks (changes requested):

starboard/shared/modular/starboard_layer_posix_getrandom_abi_wrappers.cc
sacuff: It's used alongside `GRND_NONBLOCK` in `HandlesLargeBufferNonBlocking` but I added one that uses it alone

sacuff:
Symbol is no longer leaking after #10144. I'll leave this PR up in case we do still want the wrapper.

starboard/nplb/posix_compliance/getrandom_test.cc
```diff
@@ -0,0 +1,149 @@
...
+  EXPECT_NE(buf1, buf2) << "Two random buffers should not be identical.";
+
+  bool all_zero = true;
+  for (char c : buf1) {
+    if (c != 0) {
+      all_zero = false;
+      break;
+    }
+  }
+  EXPECT_FALSE(all_zero) << "Random buffer should not be all zeros.";
```
jellefoks, line 46: Technically, both of these could occasionally occur... But with 256 bytes, it likely it will be a true negative much more often than a false positive. But maybe leaving a clarifying comment here to the reader that it was considered is good form.

starboard/nplb/posix_compliance/getrandom_test.cc
```diff
@@ -0,0 +1,149 @@
...
+    GTEST_SKIP() << "getrandom() not supported on this platform.";
+  }
+  ASSERT_EQ(result1, kBufSize) << "getrandom failed: " << strerror(errno);
+
+  ssize_t result2 = getrandom(buf2.data(), buf2.size(), 0);
+  ASSERT_EQ(result2, kBufSize) << "getrandom failed: " << strerror(errno);
+
+  EXPECT_NE(buf1, buf2) << "Two random buffers should not be identical.";
+
+  bool all_zero = true;
```
jellefoks: Nit: in C++20, this can be a one-liner: 'bool all_zero = std::ranges::all_of(buf1, [](char c) { return c == 0; });' (include <algorithm>)

jellefoks (approved):
Just small nits.

y4vor (approved):
Please update starboard/CHANGELOG.md too.

y4vor (approved):

PR #9947, opened 2026-04-08
[POSIX] Add sched_setscheduler, sched_getscheduler, sched_getparam, sched_set_param to Starboard

sacuff:
This PR primarily implements a wrapper for `sched_setscheduler` using musl types and enums for `sched_param`, which already exised to support pthread. Two missing scheduling policies (`SCHED_BATCH` and `SCHED_IDLE`) are added to the ABI wrapper definitions based on the existing musl values.

Wrappers for `sched_getscheduler`, `sched_getparam`, and `sched_set_param` are also added, mostly in order to support test setup and teardown for `sched_setscheduler`.

Tests generated by Gemini and artisanally corrected.

Bug: 500417953


starboard/shared/modular/starboard_layer_posix_sched_abi_wrappers.cc
```diff
@@ -43,3 +46,44 @@ int __abi_wrap_sched_getaffinity(musl_pid_t pid,
...
+                                  int policy,
+                                  const musl_sched_param* param) {
+  if (!param || pid < 0) {
+    errno = EINVAL;
+    return -1;
+  }
+
+  int native_policy;
+  switch (policy) {
+    case MUSL_SCHED_FIFO:
```
jellefoks, line 82: These policy values should at least be covered by a 'doesn't crash' or better 'is supported' test. Even for scheduler policies that non-root apps may not have (allow it to return EPERM  not permitted). Perhaps consider which policies Chrome requests and make sure that for those it doesn't return EPERM, unless of course Chrome handles/allows EPERM).

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,88 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
```
jellefoks, line 1: We may also need sched_getscheduler, or our tests could assume that 'SCHED_OTHER' is the default. 

Regardless, after calling sched_setscheduler, it changes the process, so the tests have to reset it back to something sane (at least if it didn't return an error).

starboard/shared/modular/starboard_layer_posix_sched_abi_wrappers.cc
```diff
@@ -43,3 +46,44 @@ int __abi_wrap_sched_getaffinity(musl_pid_t pid,
...
+      native_policy = SCHED_OTHER;
+      break;
+#ifdef SCHED_BATCH
+    case MUSL_SCHED_BATCH:
+      native_policy = SCHED_BATCH;
+      break;
+#endif
+#ifdef SCHED_IDLE
+    case MUSL_SCHED_IDLE:
+      native_policy = SCHED_IDLE;
```
jellefoks, line 98: I think this one could be an important one to make sure we test because it's a non-realtime one.

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,88 @@
...
+
+TEST(PosixSchedSetSchedulerTest, SchedSetSchedulerFailsWithInvalidPriority) {
+  int min_priority = sched_get_priority_min(SCHED_FIFO);
+  int max_priority = sched_get_priority_max(SCHED_FIFO);
+
+  if (min_priority == -1 || max_priority == -1) {
+    GTEST_SKIP() << "Could not get priority range for SCHED_FIFO";
+  }
+
+  struct sched_param param;
```
jellefoks, line 172: To be thorough, we would like to check that setting to min priority is allowed. But, there is a problem with that because non-root processes are not allowed to increased their own priority, so we could only test that if we could spawn child processes to do that in, but we don't have fork() or clone(), so we can't test that. But I think we should at least mention here that it's not tested for that reason.

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,88 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace nplb {
+namespace {
+
+TEST(PosixSchedSetSchedulerTest, SchedSetSchedulerSuccess) {
+  struct sched_param param;
+  param.sched_priority = 0;
+  // SCHED_OTHER usually requires priority 0.
+  int result = sched_setscheduler(0, SCHED_OTHER, &param);
```
jellefoks, line 64: Let's also test sched_idle (even though it's technically Linux, not POSIX).

And maybe also just try to set the others SCHED_FIFO, SCHED_RR, and for those allow EPERM (because regular platforms usually don't allow realtime for regular processes), and SCHED_BATCH if it's defined for the platform.

jellefoks (changes requested):

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
sacuff: Added this (needed sched_getparam too)

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
sacuff: Added more cases. Oddly enough `SCHED_BATCH` seems to be working with max priority for me.

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
sacuff: Added a comment at the top

starboard/tools/api_leak_detector/api_leak_detector.py
```diff
@@ -234,8 +234,11 @@
     'recvmsg',
     'rename',
     'sched_getaffinity',
+    'sched_getparam',
```
andrewsavage1, line 238: This should be added to the exported symbols in the elf loader

starboard/nplb/posix_compliance/posix_sched_getparam_test.cc
```diff
@@ -0,0 +1,47 @@
...
+TEST(PosixSchedGetParamTest, SchedGetParamSuccess) {
+  struct sched_param param;
+  int result = sched_getparam(0, &param);
+
+  EXPECT_EQ(result, 0) << "sched_getparam failed: " << strerror(errno);
+}
+
+TEST(PosixSchedGetParamTest, SchedGetParamFailsWithInvalidPid) {
+  struct sched_param param;
+  int result = sched_getparam(-1, &param);
```
andrewsavage1: I don't think this is a valid test, as sched_getparam has undefined behavior for negative values of pid

starboard/nplb/posix_compliance/posix_sched_getscheduler_test.cc
```diff
@@ -0,0 +1,36 @@
...
+namespace nplb {
+namespace {
+
+TEST(PosixSchedGetSchedulerTest, SchedGetSchedulerSuccess) {
+  int policy = sched_getscheduler(0);
+  EXPECT_GE(policy, 0) << "sched_getscheduler failed: " << strerror(errno);
+}
+
+TEST(PosixSchedGetSchedulerTest, SchedGetSchedulerFailsWithInvalidPid) {
+  int result = sched_getscheduler(-1);
```
andrewsavage1: Same as above

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,188 @@
...
+  EXPECT_GE(result, 0) << "sched_setscheduler failed: " << strerror(errno);
+#else
+  GTEST_SKIP() << "SCHED_IDLE is not supported on this system.";
+#endif  // defined(SCHED_IDLE)
+}
+
+TEST_F(PosixSchedSetSchedulerTest, SchedSetSchedulerFailsWithInvalidPid) {
+  struct sched_param param;
+  param.sched_priority = 0;
+  int result = sched_setscheduler(-1, SCHED_OTHER, &param);
```
andrewsavage1: Same as above

starboard/nplb/posix_compliance/posix_sched_getparam_test.cc
sacuff: Switched to `numeric_limits<pid_t>::max()`

starboard/tools/api_leak_detector/api_leak_detector.py
```diff
@@ -235,8 +235,11 @@
...
```
jellefoks, line 238: If we're adding getparam, we may as well also add the complement setparam.

starboard/nplb/posix_compliance/posix_sched_getparam_test.cc
```diff
@@ -0,0 +1,49 @@
...
+  int result = sched_getparam(0, &param);
+
+  EXPECT_EQ(result, 0) << "sched_getparam failed: " << strerror(errno);
+}
+
+TEST(PosixSchedGetParamTest, SchedGetParamFailsWithInvalidPid) {
+  struct sched_param param;
+  int result = sched_getparam(std::numeric_limits<pid_t>::max(), &param);
+
+  EXPECT_EQ(result, -1);
```
jellefoks, line 39: Also test the priority value is in the range in between sched_get_priority_min and sched_get_priority_max?

starboard/nplb/posix_compliance/posix_sched_getscheduler_test.cc
```diff
@@ -0,0 +1,38 @@
...
+#include <limits>
+
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace nplb {
+namespace {
+
+TEST(PosixSchedGetSchedulerTest, SchedGetSchedulerSuccess) {
+  int policy = sched_getscheduler(0);
+  EXPECT_GE(policy, 0) << "sched_getscheduler failed: " << strerror(errno);
```
jellefoks: Also test that the return value is one of the values SCHED_FIFO/SCHED_RR or SCHED_OTHER (it would be wrong if we're in BATCH or IDLE during nplb).

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,191 @@
...
+
+  if (min_priority == -1 || max_priority == -1) {
+    GTEST_SKIP() << "Could not get priority range for SCHED_FIFO";
+  }
+
+  struct sched_param param;
+  param.sched_priority = max_priority;
+  int result = sched_setscheduler(0, SCHED_FIFO, &param);
+  if (result == -1 && errno == EPERM) {
+    GTEST_SKIP() << "Insufficient permissions to set scheduling policy.";
```
jellefoks, line 85: Just a question: Do we have any platforms where this, and setting RR does not end up being skipped? Any platforms where after setting batch (if successful), we can't get back to the original policy?

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,191 @@
...
+ protected:
+  void SetUp() override {
+    original_policy = sched_getscheduler(0);
+    ASSERT_GE(original_policy, 0)
+        << "sched_getscheduler failed: " << strerror(errno);
+    ASSERT_GE(sched_getparam(0, &original_param), 0)
+        << "sched_getparam failed: " << strerror(errno);
+  }
+  void TearDown() override {
+    sched_setscheduler(0, original_policy, &original_param);
```
jellefoks: We should at least log something if this fails. I'm not sure if this is always possible, it's possible that for example after going to IDLE, that you need special permissions to go back to FIFO.

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  EXPECT_GE(result, 0) << "sched_setscheduler failed: " << strerror(errno);
+#else
+  GTEST_SKIP() << "SCHED_BATCH is not supported on this system.";
+#endif  // defined(SCHED_BATCH)
+}
+
+TEST_F(PosixSchedSetSchedulerTest, SchedSetSchedulerSuccessForIdle) {
+#if defined(SCHED_IDLE)
```
jellefoks: This define SCHED_IDLE is defined in musl, so the only place where this can happen is on the monolythic build(s). If this ever happens on a modular or evergreen build, we should fail to compile here.

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
```diff
@@ -0,0 +1,191 @@
...
+  int result = sched_setscheduler(0, SCHED_RR, &param);
+  if (result == -1 && errno == EPERM) {
+    GTEST_SKIP() << "Insufficient permissions to set scheduling policy.";
+  }
+
+  EXPECT_GE(result, 0) << "sched_setscheduler failed: " << strerror(errno);
+}
+
+TEST_F(PosixSchedSetSchedulerTest, SchedSetSchedulerSuccessForBatch) {
+#if defined(SCHED_BATCH)
```
jellefoks: This define SCHED_IDLE is defined in musl, so the only place where this can happen is on the monolythic build(s). If this ever happens on a modular or evergreen build, we should fail to compile here.

jellefoks (changes requested):

starboard/nplb/posix_compliance/posix_sched_setscheduler_test.cc
sacuff: Don't think so. Maybe we might as well remove the RR and FIFO tests? Not sure who would be running YT on a realtime platform

jellefoks: I guess the only way to know is to keep the test in, and we'll see if any platforms don't skip it. There may be a benefit for media to have that capability.

jellefoks (approved):

y4vor (approved):

PR #9951, opened 2026-04-08
evergreen: Reload web contents on network reconnect

andrewsavage1:
On network reconnection (kSbEventTypeOsNetworkConnected), automatically
reload web contents that are currently displaying an error page or
represent an initial failed load. This improves the user experience
by recovering from network interruptions without requiring manual
intervention.

Additionally, implement a Starboard-specific network change notifier
factory to accurately reflect the initial network state. This ensures
Chromium's internal network stack is correctly initialized based on
the Starboard platform's connectivity status.

Bug: 495438901


andrewsavage1:
Tested on an EAP device, seems to resolve the linked bug

jellefoks (approved):
Beautiful solution!

PR #9961, opened 2026-04-09
cobalt: Record lifecycle state as crash annotation

hlwarriner:
This ports forward this behavior from C25, leveraging Cobalt's new lifecycle state machine. The behavior is only added for platforms that use Starboard and that implement the CrashHandler Starboard extension: so, 3P platforms.

The CrashHandler Starboard extension implementation may be injected into cobalt::AppEventDelegate as a dependency so that unit tests can provide a mock.

#vibe-coded

Issue: 486236529

### DIFF
--- a/cobalt/app/app_event_delegate.cc
+++ b/cobalt/app/app_event_delegate.cc
@@ -432,6 +432,8 @@ const char* AppEventDelegate::GetStateString(ApplicationState state) {
       return "kFrozen";
     case ApplicationState::kStopped:
       return "kStopped";
+    default:
+      NOTREACHED();
   }
 }
 
--- a/cobalt/app/app_event_delegate.cc
+++ b/cobalt/app/app_event_delegate.cc
@@ -25,6 +25,8 @@
 #include "base/threading/platform_thread.h"
 #include "cobalt/app/app_event_runner.h"
 #include "content/public/browser/browser_thread.h"
+#include "starboard/extension/crash_handler.h"
+#include "starboard/system.h"
 
 namespace cobalt {
 
@@ -61,8 +63,11 @@ AppEventDelegate::ApplicationState SbEventToTargetApplicationState(
 }
 }  // namespace
 
-AppEventDelegate::AppEventDelegate(std::unique_ptr<AppEventRunner> runner)
-    : runner_(std::move(runner)) {
+AppEventDelegate::AppEventDelegate(
+    std::unique_ptr<AppEventRunner> runner,
+    const CobaltExtensionCrashHandlerApi* crash_handler_extension)
+    : runner_(std::move(runner)),
+      crash_handler_extension_(crash_handler_extension) {
   base::AutoLock lock(lock_);
   application_state_ = ApplicationState::kInitial;
   target_state_ = ApplicationState::kInitial;
@@ -70,6 +75,13 @@ AppEventDelegate::AppEventDelegate(std::unique_ptr<AppEventRunner> runner)
     // If a special runner wasn't provided, use the default.
     runner_ = AppEventRunner::Create();
   }
+  if (!crash_handler_extension_) {
+    // If a special extension implementation wasn't provided, use the default.
+    crash_handler_extension_ =
+        static_cast<const CobaltExtensionCrashHandlerApi*>(
+            SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+  }
+  SetApplicationStateAnnotation(application_state_);
 }
 
 AppEventDelegate::~AppEventDelegate() {}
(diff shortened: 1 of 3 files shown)


cobalt/app/app_event_delegate.cc
```diff
@@ -145,6 +157,7 @@ void AppEventDelegate::HandleEventLocked(const SbEvent* event) {
         runner_->OnStart(event);
         application_state_ = SbEventToTargetApplicationState(event->type);
         target_state_ = application_state_;
+        SetApplicationStateAnnotation(application_state_);
```
jellefoks: Maybe a better pattern would be to stop assigning application_state_ directly in different places in the code, and make this a `SetApplicationState()` that has the desired side effect for the crash handler in one place.

cobalt/app/app_event_delegate.cc
```diff
@@ -317,6 +331,7 @@ void AppEventDelegate::ExecuteNextStepOnUIThreadLocked(
   }
 
   application_state_ = next_state;
+  SetApplicationStateAnnotation(application_state_);
```
jellefoks: Hmnm.

This is feature creep, but please give this some thought and maybe a followup task?

While this is technically correct, it may leave out something important that we may want to know: When the app fails during lifecycle state transitions itself. The report will have just the pre- (of mid-) transition state without any indication that we are transitioning.

I wonder if we can add information that to the crash report. e.g. I wonder if we can report something like "(target_state_ << 4) | current_state" while transitioning and use that as a signal that we're undergoing transitions. e.g. 0x11 means we are in stable started state, but 0x31 means we're in started state but we've initiated concealement, and 0x41 means we're frozen but we're trying to resume, etc.

cobalt/app/app_event_delegate.cc
```diff
@@ -61,15 +63,25 @@ AppEventDelegate::ApplicationState SbEventToTargetApplicationState(
...
+    : runner_(std::move(runner)),
+      crash_handler_extension_(crash_handler_extension) {
   base::AutoLock lock(lock_);
   application_state_ = ApplicationState::kInitial;
   target_state_ = ApplicationState::kInitial;
   if (!runner_) {
     // If a special runner wasn't provided, use the default.
     runner_ = AppEventRunner::Create();
   }
+  if (!crash_handler_extension_) {
```
jellefoks, line 71: Now that we're expanding on this pattern, I realize that this isn't the common way to do this, and turns ugly quickly.

It should probably use a factory method here (a public `static AppEventDelegate* CreateAppEventDelegate() {}`, after adding a private `SetAppEventRunner()` and `SetCrashHandlerExtension()`,  with the `CreateFoo` overloads that are needed for testing.

Even better if the AppEventRunner and CrashHandlerAPI are virtual interfaces, because then it's also a clean dependency injection that we could even default to a stub and not need `if (!foo)` safety checks anywhere for.

The constructor then becomes a lot simpler, and the code more undestandable because it uses common practice...

WDYT?

cobalt/app/app_event_delegate.h
```diff
@@ -94,6 +101,7 @@ class AppEventDelegate {
   mutable base::Lock lock_;
 
   std::unique_ptr<AppEventRunner> runner_;
+  const CobaltExtensionCrashHandlerApi* crash_handler_extension_ = nullptr;
```
jellefoks: Can we make this a `raw_ptr<T>`, as a start of buliding a habit of avoiding actual raw pointers.

https://chromium.googlesource.com/chromium/src/+/main/base/memory/raw_ptr.md

jellefoks (approved):
Great! None of these comments are blockers, but please do give them some consideration.

cobalt/app/app_event_delegate.h
hlwarriner: Good idea - done.

cobalt/app/app_event_delegate.cc
hlwarriner: Done. Taking this a step further, it would also be nice to enforce that updates to `application_state_` can only be done through `SetApplicationState()`, to ensure the annotation is also updated. I added a todo comment for this.

cobalt/app/app_event_delegate.cc
hlwarriner: I think this would be a good change from a code clarity perspective but also to better decouple AppEventDelegate from its dependencies. Since it's not trivial and we're under time pressure to land the crash annotation behavior, I added a todo comment and will save it for a future refactoring.

cobalt/app/app_event_delegate.cc
hlwarriner: This is an interesting idea but I agree it's feature creep. The critical business need is maintain feature parity with C25 by capturing application state so we can determine whether the application was visible to the user. You're right that we may miss some edge cases during transitions, but let's save that for future work. I'll file a feature request.

hlwarriner:
> Great! None of these comments are blockers, but please do give them some consideration.

Thanks for the thoughtful comments. PTAL.

PR #9989, opened 2026-04-10
Use overlay to manage runner cache for cobalt checkout

linxinan-chops:
Install fuse-overlayfs to manage the cobalt checkout and file fetched by gclient.
As changed files are stored in the upper path via overlay, we can simply restore the runner cache to its initial state(checked out at TOT of Cobalt). So that the next run could start from the same clean cache.

Also correct the runner for cobalt browser tests job, which should also use chrobalt-linux-runner like the unit test jobs.  

Bug: 501486531


.github/actions/overlay_workspace/action.yaml
```diff
@@ -0,0 +1,126 @@
...
+        WORK_DIR="/runner-cache/workspaces/${GITHUB_RUN_ID}_${GITHUB_RUN_ATTEMPT}/work"
+
+        # MERGED_DIR: The directory the CI job actually sees and works in. It seamlessly combines LOWER_DIR and UPPER_DIR.
+        MERGED_DIR="${GITHUB_WORKSPACE}/${{ inputs.path }}"
+
+        # Failsafe cleanup: Unmount and delete any orphaned workspaces from cancelled jobs
+        # Since this node only runs one pod at a time, we can safely wipe all workspaces.
+        echo "Cleaning up any orphaned overlay workspaces..."
+        fusermount3 -u "$MERGED_DIR" || fusermount -u "$MERGED_DIR" || umount "$MERGED_DIR" || true
+        rm -rf /runner-cache/workspaces/*
```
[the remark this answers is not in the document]
linxinan-yt: Our node only runs one runner each time. This is WAI.

.github/actions/overlay_workspace/action.yaml
```diff
@@ -0,0 +1,126 @@
...
+        # Add GoB Authentication
+        if ! curl -s -m 2 -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/email" > /dev/null; then
+          echo "Metadata server not reachable. Skipping GoB authentication setup."
+        else
+          SERVICE_ACCOUNT=$(curl -s -f -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/email")
+          ACCESS_TOKEN=$(curl -s -f -H "Metadata-Flavor: Google" "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token" | jq -r .access_token)
+          if [[ -n "${SERVICE_ACCOUNT}" && -n "${ACCESS_TOKEN}" ]]; then
+            echo "::add-mask::${ACCESS_TOKEN}"
+            echo "Authenticating with ${SERVICE_ACCOUNT}"
+            GITCOOKIES_PATH="/tmp/gitcookies"
```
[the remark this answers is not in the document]
linxinan-yt: Only single job pre node in our CI.

.github/actions/overlay_workspace/action.yaml
```diff
@@ -0,0 +1,126 @@
...
+          mkdir -p "$LOWER_DIR"
+          cd "$LOWER_DIR"
+
+          git clone https://github.com/youtube/cobalt.git src
+          # Clone depot_tools
+          git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git depot_tools
+          export PATH="$LOWER_DIR/depot_tools:$PATH"
+
+          cd src
+          gclient config --name=src https://github.com/youtube/cobalt.git
```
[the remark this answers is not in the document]
linxinan-yt: Done.

.github/actions/teardown_overlay/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Teardown OverlayFS Workspace
+description: Unmounts the OverlayFS and deletes the upper layer changes.
+
+runs:
+  using: "composite"
+  steps:
+    - name: Teardown Workspace
+      shell: bash
+      run: |
+        MERGED_DIR="${GITHUB_WORKSPACE}/cobalt"
```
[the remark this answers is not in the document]
linxinan-yt: Done.

.github/actions/teardown_overlay/action.yaml
```diff
@@ -0,0 +1,20 @@
+name: Teardown OverlayFS Workspace
+description: Unmounts the OverlayFS and deletes the upper layer changes.
+
```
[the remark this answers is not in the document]
linxinan-yt: Done.

.github/actions/overlay_workspace/action.yaml
```diff
@@ -0,0 +1,146 @@
...
+          git clone https://github.com/youtube/cobalt.git src
+          # Clone depot_tools
+          git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git depot_tools
+          export PATH="$LOWER_DIR/depot_tools:$PATH"
+
+          gclient config \
+            --name=src \
+            --custom-var="rbe_instance=\"projects/cobalt-actions-prod/instances/default_instance\"" \
+            "https://github.com/youtube/cobalt.git"
+          gclient sync -D --no-history
```
linxinan-yt, line 86: TODO: Should move gob authentication earlier than this step

.github/workflows/main.yaml
```diff
@@ -634,7 +657,7 @@ jobs:
         ((inputs.nightly == 'true' || github.event_name == 'schedule') && vars.RUN_ODT_TESTS_ON_NIGHTLY != 'False') ||
         (github.event_name == 'push' && vars.RUN_ODT_TESTS_ON_POSTSUBMIT != 'False')
       )
-    runs-on: [self-hosted, odt-runner]
+    runs-on: [self-hosted, chrobalt-linux-runner]
```
oxve, line 662: Why are you changing this runner? `chrobalt-linux-runner` is a pretty beefy machine that is not needed for running odt.

.github/workflows/main.yaml
```diff
@@ -482,6 +499,12 @@ jobs:
         uses: ./.github/actions/api_leak_detector
         if: inputs.run_api_leak_detector
 
+      - name: Teardown OverlayFS
```
oxve: According to this port the step should be allowed to finish running: https://github.com/orgs/community/discussions/26311#discussioncomment-3251359

Is it a problem if it doesn't? Given no official docs and the reports of bash not respecting this I don't think we should rely on it.

.github/actions/overlay_workspace/action.yaml
```diff
@@ -0,0 +1,146 @@
...
+        # Failsafe cleanup: Unmount and delete any orphaned workspaces from cancelled jobs
+        # Since this node only runs one pod at a time, we can safely wipe all workspaces.
+        echo "Cleaning up any orphaned overlay workspaces..."
+        fusermount3 -u "$MERGED_DIR" || fusermount -u "$MERGED_DIR" || umount "$MERGED_DIR" || true
+        rm -rf /runner-cache/workspaces/*
+
+        # Ensure the directories exist
+        mkdir -p "$UPPER_DIR" "$WORK_DIR" "$MERGED_DIR"
+
+        # Add GoB Authentication
```
oxve, line 48: Are you planning on keeping the code below in this action or can we rely on the logic as-is in the depot tools action?

If we can we should avoid the duplication.

.github/workflows/main.yaml
```diff
@@ -433,18 +434,34 @@ jobs:
           fetch-depth: 1
           filter: blob:none
           sparse-checkout: ${{ env.CI_ESSENTIALS }}
+      - name: Prepare OverlayFS Workspace
+        id: overlay_workspace
+        if: runner.os == 'Linux'
```
oxve: Can we put this in config instead?

.github/workflows/main.yaml
linxinan-yt: Not sure, but before I changed it I always got this error: https://github.com/youtube/cobalt/actions/runs/24270254516/job/70877449991

.github/workflows/main.yaml
linxinan-yt: We can reset the overlay at the beginning of each runner. This is proposed by Gemini, which I did not get time to go through last Friday.

.github/workflows/main.yaml
linxinan-yt: I see. Since tvOS deos not run webtests, I guess I can remove `if runner.os ...`, correct?

.github/actions/overlay_workspace/action.yaml
linxinan-yt: I would keep the authentication here, so that all our traffic is authenticated. I will file follow up PR to remove them from the depot_tools/action.yaml, since it will be used by TVOS only.

.github/workflows/main.yaml
linxinan-yt: Let me try to restore it and see.

.github/workflows/main.yaml
linxinan-yt: My bad, this is not webtest, but build. Which config do you mean?

linxinan-yt: Synced offline. Step "Prepare OverlayFS Workspace" should better integrate with existing steps, like depot tools checkout. I will leave a TODO here for followup.

.github/workflows/main.yaml
linxinan-yt: Synced offline, browser test should runs on chrobalt-linux-runner, same with unit tests.

.github/workflows/main.yaml
oxve: Sgtm. IMO we should always prefer gating by feature and not by platform name. That way the configs are the source of truth for build feature gating.

Hence, I think we should add a config flag under .github/config for platforms that have runner cache available. Or some other mechanism to check for the cache folder.

Gating by platform somewhere in the build config has a tendency to turn into spaghetti rather quickly.

.github/actions/overlay_workspace/action.yaml
oxve: Synced offline. We will integrate this into the existing actions once verified.

For the record, my question was about the rest of code in this file, not just the authentication bits.

oxve (approved):

.github/actions/overlay_workspace/action.yaml
linxinan-yt: Ack. Yes, they should be moved or refactored later. As we are new to overlay, I am not 100% sure how it work in production. If not going well, I may have to revert and I don't the revert too complicated. So let's land this POC first. If it work well, I can start a follow up PR to integrate the rest code.

.github/workflows/main.yaml
linxinan-yt: Ack. I will file follow up PR to those config files. For this one I will use `${{ contains(matrix.platform, 'tvos') }}` as a workaround.

PR #10006, opened 2026-04-13
cobalt: Add LevelDB errors to Local Storage UMA

zhongqiliang:
The previous UMA metrics for Local Storage recreate or reset events
did not provide specific error codes. This made it challenging to
diagnose the root causes of database issues.

This change introduces new UMA histograms to capture detailed LevelDB
error statuses when failures occur during Local Storage database open,
read operations, schema version checks, and commit operations. These
specific error codes will provide better insight into the reasons
behind database recreation attempts, improving debuggability and
monitoring.

Bug: 488465561


components/services/storage/dom_storage/local_storage_impl.cc
```diff
@@ -671,6 +671,9 @@ void LocalStorageImpl::OnDatabaseOpened(leveldb::Status status) {
                                 leveldb_env::LEVELDB_STATUS_MAX);
 
   if (!status.ok()) {
+    base::UmaHistogramEnumeration("Cobalt.LocalStorage.DatabaseOpenError",
```
johnxwork, line 677: You would need to add something like `#if BUILDFLAG(IS_COBALT)` since we are modifying Chromium code here.

johnxwork (commented):
Did you verify that you can see the data locally with the script in cobalt/tools/ to pull UMA?

components/services/storage/dom_storage/local_storage_impl.cc
zhongqiliang: done

johnxwork (approved):

PR #10017, opened 2026-04-14
cobalt/metrics: Add interfaces and stubs for granular memory tracking.

Awallky:
This commit introduces the `DetailedMetricsDelegate` interface and necessary
stubs in `OSMetrics` to support custom memory categorization in Cobalt.
It also includes a Starboard wrapper for `prctl` to support VMA naming.
This is based on the design specified in http://shortn/_Pw4G7chyR2.

- Add `DetailedMetricsDelegate` interface.
- Add stubs in `OSMetrics` for delegate interaction.
- Add Starboard wrapper for `prctl` (VMA naming).

Bug: 494004530

### DIFF
--- a/services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate.h
+++ b/services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate.h
@@ -1,6 +1,16 @@
-// Copyright 2026 The Chromium Authors
-// Use of this source code is governed by a BSD-style license that can be
-// found in the LICENSE file.
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
 
 #ifndef SERVICES_RESOURCE_COORDINATOR_PUBLIC_CPP_MEMORY_INSTRUMENTATION_DETAILED_METRICS_DELEGATE_H_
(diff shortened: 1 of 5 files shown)


services/resource_coordinator/public/cpp/memory_instrumentation/os_metrics_linux.cc
```diff
@@ -591,7 +535,8 @@ void OSMetrics::SetProcSmapsForTesting(FILE* f) {
 
 bool OSMetrics::FillOSMemoryDump(base::ProcessHandle handle,
                                  const MemDumpFlagSet& flags,
-                                 mojom::RawOSMemDump* dump) {
+                                 mojom::RawOSMemDump* dump,
```
andrewsavage1: We should guard changes to upstream code

services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate.h
```diff
@@ -0,0 +1,52 @@
+// Copyright 2026 The Chromium Authors
```
andrewsavage1: Where does this file come from? It's new but we're giving it a chromium header?

Awallky: Ah, yes. This is an artifact from porting it from this upstream CL - https://chromium-review.git.corp.google.com/c/chromium/src/+/7732681/12. I will modify this to a Cobalt license header. We can always update this when it lands upstream.

services/resource_coordinator/public/cpp/memory_instrumentation/os_metrics_linux.cc
Awallky: I reverted this and added a separate cobalt-specific method for this with the detailed metrics delegate as an extra param.

services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate.h
Awallky: I changed this to Cobalt's license header instead of Chromium's.

services/resource_coordinator/public/cpp/memory_instrumentation/os_metrics_linux.cc
Awallky: Done.

andrewsavage1 (approved):

PR #10028, opened 2026-04-14
build: fix condition for removing NetworkChangeNotifier on hermetic builds

rogerzanoni:
The base/network_change_notifier_linux.* files are added to the sources only for linux builds, but should be removed for hermetic builds. Add the same is_linux guard to check if the files should be removed from source to fix the following gn gen issue:

ERROR at //net/BUILD.gn:1334:7: Item not found
      "base/network_change_notifier_linux.cc"

Bug: 495203133

### DIFF
--- a/net/BUILD.gn
+++ b/net/BUILD.gn
@@ -1319,21 +1319,12 @@ component("net") {
 
   if (is_linux) {
     sources += [
-      "base/network_change_notifier_linux.cc",
-      "base/network_change_notifier_linux.h",
       "proxy_resolution/proxy_config_service_linux.cc",
       "proxy_resolution/proxy_config_service_linux.h",
     ]
     if (use_glib) {
       deps += [ "//ui/base/glib" ]
     }
-  }
-
-  if (is_cobalt_hermetic_build) {
-    sources -= [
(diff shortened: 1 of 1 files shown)


net/BUILD.gn
```diff
@@ -1329,7 +1329,7 @@ component("net") {
     }
   }
 
-  if (is_cobalt_hermetic_build) {
+  if (is_linux && is_cobalt_hermetic_build) {
```
hferreiro: Perhaps it makes sense to unify this with the block that adds address_map_cache_linux.*. That is, remove the 4 files in a single block.

hferreiro:
I'm not sure if net/android/network_change_notifier_android.* should be omitted too.

rogerzanoni:
Changed net/BUILD.gn to only add it if it's a linux and non-hermetic build instead of keeping the block that was removing it

net/BUILD.gn
rogerzanoni: Fixed, I'm only adding network_change_notifier_linux.cc if is_linux && !is_cobalt_hermetic_build

haozheng-cobalt (approved):

PR #9979, opened 2026-04-09
media: Support decode-to-texture mode for video

borongc:
Enable video playback using the decode-to-texture output mode. This
introduces a new GL texture image backing for Starboard decode targets,
allowing GPU-side rendering and composition of video frames.

The StarboardRendererWrapper now orchestrates the creation of ClientSharedImage
from SbDecodeTargets received from the Starboard player. This involves
communicating with the GPU thread to create and manage the necessary GL
textures and shared image representations. The SbPlayerBridge dynamically
registers its graphics context provider when in decode-to-texture mode.

This provides a flexible mechanism for integrating video output with
Cobalt's rendering pipeline, enabling advanced compositing effects.

The unit tests were generated with assistance from `gemini-cli` and refined
for correctness.

#vibe-coded

Issue: 500367083

PR #10045, opened 2026-04-14
evergreen: Render OpenGL texture in red channel

borongc:
This change updates the OpenGL texture format from GL_ALPHA to GL_RED_EXT
when creating and updating single-channel textures for video frames.
Previously, single-plane video data was incorrectly interpreted as alpha,
leading to improper rendering within Chromium display pipeline.
Using GL_RED_EXT ensures that the texture data is correctly
interpreted as brightness, facilitating proper color mapping and display
of video content.

This is required for decode-to-texture mode on 3P devices.
See https://github.com/youtube/cobalt/pull/9979.

Issue: 501495783

### DIFF
--- a/starboard/CHANGELOG.md
+++ b/starboard/CHANGELOG.md
@@ -7,6 +7,13 @@ since the version previous to it.
 
 **NOTE: Starboard versions 16 and older are no longer supported.**
 
+## Version 18
+
+### Support decode-to-texture mode
+Implementations of `DecodeTarget` for YUV planes should ensure they use
+`GL_RED_EXT` or `GL_LUMINANCE` instead of `GL_ALPHA` to avoid green-screen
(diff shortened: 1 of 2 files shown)


jasonzhangxx (approved):
LGTM, but needs @y4vor approval as it will require partners to make a similar change at their side.

andrewsavage1:
If this is a change that partners will have to make, please update //starboard/CHANGELOG.md as well so we remember to add it to release notes.

borongc:
> If this is a change that partners will have to make, please update //starboard/CHANGELOG.md as well so we remember to add it to release notes.

Done.

andrewsavage1 (approved):
LGTM. @y4vor you should take a look and see if there's any other way you'd like to track these changes to ensure it gets to partners, but @borongc feel free to submit and we can follow up on that later

PR #10060, opened 2026-04-15
DRAFT: Android: Use dependency injection for text to speech settings …

hferreiro:
…change event

Bug: 492704919

### DIFF
--- a/starboard/android/shared/BUILD.gn
+++ b/starboard/android/shared/BUILD.gn
@@ -319,7 +319,6 @@ static_library("starboard_platform") {
     "//cobalt/android:jni_headers",
     "//cobalt/browser:cobalt_crash_annotations",
     "//cobalt/browser/client_hint_headers",
-    "//cobalt/browser/h5vcc_accessibility:manager",
     "//cobalt/browser/h5vcc_runtime:deep_link_manager",
     "//starboard/common",
     "//starboard/extension:feature_config_header",
--- a/starboard/android/shared/text_to_speech_helper.cc
+++ b/starboard/android/shared/text_to_speech_helper.cc
@@ -14,8 +14,8 @@
 
 #include "starboard/android/shared/text_to_speech_helper.h"
 
-#include "cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_manager.h"
 #include "starboard/android/shared/starboard_bridge.h"
+#include "starboard/shared/starboard/application.h"
 
 // Must come after all headers that specialize FromJniType() / ToJniType().
 #include "cobalt/android/jni_headers/CobaltTextToSpeechHelper_jni.h"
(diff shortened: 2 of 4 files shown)


hferreiro:
@haozheng-cobalt I've read in some comments that Android doesn't use Starboard's event handling. If this approach is not possible, do you have another idea on how to fix this layering violation?

starboard/android/shared/BUILD.gn
haozheng-cobalt, line 306: We should gate all AOSP change in the common Android file using a gn arg so that the change won't impact Android TV platform which is not AOSP platform, do we have a gn arg such as is_aosp_build? 

sth like, we should pick better names, these are only suggestion.
```
if (is_aosp_build) {
  public_deps -= [ "//cobalt/browser/h5vcc_accessibility:manager" ]
  defines += [ "AOSP_BUILD=1" ]
}

File: starboard/android/shared/text_to_speech_helper.cc [hunk 1]
```diff
@@ -14,8 +14,8 @@
 
 #include "starboard/android/shared/text_to_speech_helper.h"
 
-#include "cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_manager.h"
```
haozheng-cobalt: Similarly, we should gate these in a AOSP macro.


#if !defined(AOSP_BUILD)
#include "cobalt/browser/h5vcc_accessibility/h5vcc_accessibility_manager.h"
#endif  // !defined(AOSP_BUILD)

starboard/android/shared/text_to_speech_helper.cc
```diff
@@ -41,8 +41,8 @@ bool CobaltTextToSpeechHelper::IsTextToSpeechEnabled(JNIEnv* env) const {
 }
 
 void CobaltTextToSpeechHelper::SendTextToSpeechChangeEvent(bool enabled) const {
-  cobalt::browser::H5vccAccessibilityManager::GetInstance()
-      ->OnTextToSpeechStateChanged(enabled);
+  Application::Get()
+      ->InjectAccessibilityTextToSpeechSettingsChangedEvent(enabled);
```
haozheng-cobalt, line 45: We can simply return; in this function, also remember to gate all AOSP change in a macro.


void CobaltTextToSpeechHelper::SendTextToSpeechChangeEvent(bool enabled) const {
#if !defined(AOSP_BUILD)
  cobalt::browser::H5vccAccessibilityManager::GetInstance()
      ->OnTextToSpeechStateChanged(enabled);
#endif // !defined(AOSP_BUILD)
}

haozheng-cobalt:
> @haozheng-cobalt I've read in some comments that Android doesn't use Starboard's event handling. If this approach is not possible, do you have another idea on how to fix this layering violation?

As @andrewsavage1 pointed out https://b.corp.google.com/issues/492704919#comment2.
P1 goal right now is to make the app run, most of these Cobalt deps in Starboard are not core functionalities, you can just comment out the deps in the BUILD file and use a AOSP macro to make these functions no-ops.

Make sure to create a follow-up ticket to remind us to get back fixing these stubbed out functions later, and add a // TODO([b/xxxx]):comment on the stubbed out functions.

PR #10064, opened 2026-04-15
android: Fix crash at VideoSurfaceTextureBridge::CreateSurface

borongc:
Update JNI calls for SurfaceTexture and Surface objects to use a
persistent global reference. Previously, calls used raw jobject or
temporary local references, which could lead to the underlying Java
object being garbage collected or invalid JNI references. This caused
crashes, particularly during Surface creation.

Ensure Java object references are correctly managed to prevent use-
after-free or null pointer issues in JNI interactions. Also, correct
the SB_CHECK to validate the created surface object itself.

#vibe-coded

Issue: 417242469
Issue: 500429244

### DIFF
--- a/starboard/android/shared/decode_target.h
+++ b/starboard/android/shared/decode_target.h
@@ -32,8 +32,12 @@ class DecodeTarget final : public SbDecodeTargetPrivate {
 
   bool GetInfo(SbDecodeTargetInfo* out_info) final;
 
-  jobject surface_texture() const { return surface_texture_.obj(); }
-  jobject surface() const { return surface_.obj(); }
+  const base::android::ScopedJavaGlobalRef<jobject>& surface_texture() const {
+    return surface_texture_;
+  }
+  const base::android::ScopedJavaGlobalRef<jobject>& surface() const {
+    return surface_;
+  }
 
   void set_dimension(const Size& size) {
     info_.planes[0].width = size.width;
--- a/starboard/android/shared/decode_target.cc
+++ b/starboard/android/shared/decode_target.cc
@@ -85,9 +85,8 @@ void DecodeTarget::CreateOnContextRunner() {
 
   // We will also need an Android Surface object in order to obtain a
   // ANativeWindow object that we can pass into the AMediaCodec library.
-  surface_ =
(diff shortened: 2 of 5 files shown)


starboard/android/shared/decode_target.h
```diff
@@ -33,6 +33,10 @@ class DecodeTarget final : public SbDecodeTargetPrivate {
   bool GetInfo(SbDecodeTargetInfo* out_info) final;
 
   jobject surface_texture() const { return surface_texture_.obj(); }
+  const base::android::ScopedJavaGlobalRef<jobject>&
+  surface_texture_global_ref() const {
+    return surface_texture_;
+  }
   jobject surface() const { return surface_.obj(); }
```
osagie98: I'm reviewing on mobile so I can't check the rest of the codebase, is this function still used? If so, can we replace any references with surface_texture_global_ref()?

osagie98 (approved):

starboard/android/shared/decode_target.h
borongc: It's used in L761 in starboard/android/shared/video_decoder.cc.

starboard/android/shared/decode_target.h
```diff
@@ -33,6 +33,10 @@ class DecodeTarget final : public SbDecodeTargetPrivate {
   bool GetInfo(SbDecodeTargetInfo* out_info) final;
 
   jobject surface_texture() const { return surface_texture_.obj(); }
+  const base::android::ScopedJavaGlobalRef<jobject>&
+  surface_texture_global_ref() const {
```
jasonzhangxx: My understanding is we'll no longer need `jobject surface_texture()` and `jobject surface()`. In that case, we can add new functions, like `base::android::ScopedJavaGlobalRef<jobject>&
  surface_texture()`, to replace the old ones.

starboard/android/shared/decode_target.h
borongc: Refactored it.

starboard/android/shared/decode_target.h
borongc: Done.

PR #10068, opened 2026-04-15
cobalt/metrics: refractor MetricsPollingState class

tasunnn:
Refactor the MetricsPollingState and its derived classes to improve
ownership and thread-safety.
- Make MetricsPollingState objects base::SequenceBound
- Move ownership of metrics emitters to MetricPollingState subclasses
- Maintain state of ProcessMetrics instance inside cpu emitter class
- Emitters are now created once with the initialization of 
CobaltMetricsServiceClient object instead of creating every time when 
metrics are requested.

Bug: 495528560

### DIFF
--- a/cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
+++ b/cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
@@ -25,10 +25,9 @@ namespace cobalt {
 
 // This class fetches CPU metrics for the current process and its threads and
 // records the average per-core CPU usage as a percentage in a UMA histogram.
-// It relies on a persistent base::ProcessMetrics object (tracked by
-// CobaltMetricsServiceClient::CpuPollingState) to maintain stateful CPU usage
-// tracking across multiple calls. This class is not thread-safe and must be
-// called from the same Sequence it was created on.
+// It relies on a persistent base::ProcessMetrics object to maintain stateful
+// CPU usage tracking across multiple calls. This class is not thread-safe and
+// must be called from the same Sequence it was created on.
 class CobaltCpuMetricsEmitter
     : public base::RefCountedThreadSafe<CobaltCpuMetricsEmitter> {
  public:
@@ -37,7 +36,7 @@ class CobaltCpuMetricsEmitter
   CobaltCpuMetricsEmitter(const CobaltCpuMetricsEmitter&) = delete;
   CobaltCpuMetricsEmitter& operator=(const CobaltCpuMetricsEmitter&) = delete;
 
-  void FetchAndEmitCpuMetrics(base::ProcessMetrics* process_metrics);
+  void FetchAndEmitCpuMetrics();
 
  protected:
   // TODO(b/492251096): add tests for CPU metrics emitter class
@@ -45,6 +44,8 @@ class CobaltCpuMetricsEmitter
 
  private:
   friend class base::RefCountedThreadSafe<CobaltCpuMetricsEmitter>;
+
+  std::unique_ptr<base::ProcessMetrics> process_metrics_;
 
   SEQUENCE_CHECKER(sequence_checker_);
 };
--- a/cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
+++ b/cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
@@ -211,7 +211,9 @@ static const char* MetricSizeToVersionSuffix(
 
 }  // namespace
 
-CobaltMemoryMetricsEmitter::CobaltMemoryMetricsEmitter() = default;
+CobaltMemoryMetricsEmitter::CobaltMemoryMetricsEmitter() {
+  DETACH_FROM_SEQUENCE(sequence_checker_);
+}
 
 void CobaltMemoryMetricsEmitter::FetchAndEmitProcessMemoryMetrics() {
   DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
--- a/cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
+++ b/cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
@@ -212,6 +212,9 @@ static const char* MetricSizeToVersionSuffix(
 }  // namespace
 
 CobaltMemoryMetricsEmitter::CobaltMemoryMetricsEmitter() {
+  // The emitter is created on the main thread but will be used
+  // on a background sequence maintained by base::SequenceBound
+  // in CobaltMetricsServiceClient.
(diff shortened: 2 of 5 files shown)


cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -153,20 +126,16 @@ void CobaltMetricsServiceClient::Initialize() {
...
 void CobaltMetricsServiceClient::StartMemoryMetricsLogger() {
   CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  memory_state_ = base::MakeRefCounted<MemoryPollingState>(this);
-  memory_state_->task_runner = base::ThreadPool::CreateSequencedTaskRunner({});
-  memory_state_->task_runner->PostTask(
-      FROM_HERE, base::BindOnce(&MemoryPollingState::RecordMetricsAfterDelay,
-                                base::RetainedRef(memory_state_)));
+  memory_state_.emplace(base::ThreadPool::CreateSequencedTaskRunner({}),
+                        CreateMemoryMetricsEmitter());
+  memory_state_.AsyncCall(&MemoryPollingState::RecordMetricsAfterDelay);
```
Awallky, line 128: I noticed that we have two different methods to control their polling frequency. However, AFAICT they both point to an underlying `memory_instrumentation::GetDelayForNextMemoryLog` for the delay. 
A better solution would be to add at least one Finch flag and a feature param for the delay. 

I have a flag & param combo for memory metrics I have yet to land - https://github.com/youtube/cobalt/pull/9944. But it would be good to make it so they both use a shared utility class if we want them to have the same frequency or refactor things so they both share some parent class's delay method.
Feel free to defer this to another PR.

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
```diff
@@ -25,10 +25,9 @@ namespace cobalt {
...
 // This class fetches CPU metrics for the current process and its threads and
 // records the average per-core CPU usage as a percentage in a UMA histogram.
-// It relies on a persistent base::ProcessMetrics object (tracked by
-// CobaltMetricsServiceClient::CpuPollingState) to maintain stateful CPU usage
-// tracking across multiple calls. This class is not thread-safe and must be
-// called from the same Sequence it was created on.
+// It relies on a persistent base::ProcessMetrics object to maintain stateful
+// CPU usage tracking across multiple calls. This class is not thread-safe and
+// must be called from the same Sequence it was created on.
 class CobaltCpuMetricsEmitter
```
Awallky, line 31: `CobaltCpuMetricsEmitter` looks to be mostly untested.  Can we add some unit or browser tests for this?

cobalt/browser/metrics/cobalt_metrics_service_client.cc
Awallky: Can you check if this is till needed?

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
```diff
@@ -211,7 +211,9 @@ static const char* MetricSizeToVersionSuffix(
 
 }  // namespace
 
-CobaltMemoryMetricsEmitter::CobaltMemoryMetricsEmitter() = default;
+CobaltMemoryMetricsEmitter::CobaltMemoryMetricsEmitter() {
+  DETACH_FROM_SEQUENCE(sequence_checker_);
```
Awallky, line 218: Could you add a brief comment explaining that the object is created on the main thread but intended for use on a background sequence via `SequenceBound`?

cobalt/browser/metrics/cobalt_metrics_service_client.h
```diff
@@ -171,8 +172,8 @@ class CobaltMetricsServiceClient : public metrics::MetricsServiceClient {
   base::TimeDelta min_idle_refresh_interval_ = kMinIdleRefreshInterval;
 
   // State objects for background metrics collection.
-  scoped_refptr<MemoryPollingState> memory_state_;
-  scoped_refptr<CpuPollingState> cpu_state_;
+  base::SequenceBound<MemoryPollingState> memory_state_;
```
Awallky, line 175: `cobalt_metrics_service_client_test.cc` was not updated to reflect the new asynchronous architecture. It continues to use mocks that bypass the `SequenceBound` logic, meaning the actual polling mechanism is unverified.

Can we add some test updates to this either in this PR or track it for a later PR?

Awallky (changes requested):
This mostly looks good and is a quality improvement. 
Can you just verify that memory and CPU metrics still fire as expected and address the comments I left?

cobalt/browser/metrics/cobalt_cpu_metrics_emitter.h
tasunnn: Yes, this will be in another PR

cobalt/browser/metrics/cobalt_metrics_service_client.cc
tasunnn: yeah some of them are no longer needed, will remove them

cobalt/browser/metrics/cobalt_metrics_service_client.h
tasunnn: sure, will include this in PR with tests for cpu emitter class

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
tasunnn: done

cobalt/browser/metrics/cobalt_metrics_service_client.cc
tasunnn: That's a good point, I think if we want them to be collecting at the same frequency, we might need to consolidate the two polling states to a single one so they won't do it asynchronously on two sequences, which in that way we can't guarantee memory and cpu metrics recording happen at the same time.

Awallky: Since we are pulling things from different sources, I don't think they necessarily need to be synchronized. However, we should probably at least  add some distinct feature flag params, one for memory metrics and one for cpu metrics, that we use instead of the memory instrumentation. While we may decide to default them to the same value, we could later run experiments for each to see their effect on watch time for an optimal polling frequency.
So, can you add one here? You could re-use what is in the linked PR?

tasunnn: Sure, I will add one for cpu metrics too, deferring it to a new PR after this one lands to keep PR size small.

tasunnn:
Verified locally that both memory and cpu UMA fire as normal after the refractor.

Awallky (approved):
LGTM % comment

PR #10075, opened 2026-04-15
ci: Add max-commits option to autoroll workflow

oxve:
Adds a --max-commits flag to autoroll_lts.py to limit the number of PRs included in a single roll.
Updates the GitHub workflow to support this new parameter.

Bug: 484351320

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -81,11 +81,15 @@ def main():
   p.add_argument('--target-branch', required=True)
   p.add_argument('--start-commit')
   p.add_argument('--origin-branch', default='main')
+  p.add_argument('--max-commits', type=int)
   args = p.parse_args()
 
   links = []
   target_prs = get_pr_set(args.target_branch, args.origin_branch)
   autoroll_prs = get_pr_set('HEAD', args.origin_branch)
+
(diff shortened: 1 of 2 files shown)


.github/scripts/autoroll_lts.py
```diff
@@ -103,8 +107,15 @@ def main():
 
       # If the PR is not on the current (autoroll) branch, cherry-pick it.
       if num not in autoroll_prs:
+        if (args.max_commits is not None and
+            current_pr_commits + commits_added >= args.max_commits):
+          print(
+              f"::warning::Reached max commits limit ({args.max_commits}).",
+              file=sys.stderr)
+          break
```
linxinan-yt: Instead of break, can we just truncate it? E.g. 

```
cherry_pick(sha, min(args.max_commits, num), title)
```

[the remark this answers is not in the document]
oxve: `get_pr_set` runs `git log --reverse` so it starts with the oldest commits first, as it must to avoid cherry pick conflicts.

oxve: That num is the PR number. I've updated the code to reflect. I also refactored and moved the exit condition to the top of the loop to make clearer.

linxinan-chops (approved):

PR #10080, opened 2026-04-15
Update sync_chromium_branches workflow

briantting:
The cherry pick PRs generated by the sync_chromium_branches workflow need the correct git user and token to satisfy CLA requirements.

Bug: 481788398

### DIFF
--- a/.github/workflows/sync_chromium_branches.yaml
+++ b/.github/workflows/sync_chromium_branches.yaml
@@ -84,6 +84,12 @@ jobs:
           ref: cherry-pick-main-sync
           fetch-depth: 16
           filter: blob:none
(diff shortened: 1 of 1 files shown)


SortaCookie (approved):

.github/workflows/sync_chromium_branches.yaml
oxve, line 70: I think the git config below must be done before this line.

oxve (approved):

.github/workflows/sync_chromium_branches.yaml
briantting: whoops i missed this comment, even though we don't care about the commit in line 70? The line 70 commit gets force pushed since its updating chromium_m138 branch. The commit that will be in the PR with the CLA check is the cherry pick done on line 97.

PR #10089, opened 2026-04-15
media: Move GetAudioRendererParams to AudioRendererSink

jasonzhangxx:
Move the logic for determining audio renderer buffer parameters from
PlayerComponents::Factory to AudioRendererSink. This refactors
GetAudioRendererParams into a virtual method on AudioRendererSink,
with specific implementations in AudioRendererSinkImpl and
AudioRendererSinkAndroid.

This change encapsulates platform-specific audio buffering logic directly
within the respective AudioRendererSink implementations. It removes
platform-specific details, such as Android's tunnel mode calculations and
AudioOutputManager interactions, from the generic PlayerComponents::Factory.
This improves modularity and ensures each audio sink can correctly calculate
its own required buffer sizes.

Bug: 500811542

### DIFF
--- a/starboard/android/shared/player_components_factory.cc
+++ b/starboard/android/shared/player_components_factory.cc
@@ -135,6 +135,50 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
     return tunnel_mode_audio_session_id_ != -1;
   }
 
+  void GetAudioRendererParams(const AudioStreamInfo& audio_stream_info,
+                              int* max_cached_frames,
+                              int* min_frames_per_append) const override {
+    SB_CHECK(max_cached_frames);
+    SB_CHECK(min_frames_per_append);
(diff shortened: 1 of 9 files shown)


borongc:
nit: this moves non-ATV platform as well, the title should be `starboard: ` or `refactor: `.

starboard/android/shared/player_components_factory.cc
```diff
@@ -57,6 +57,10 @@ namespace {
 using base::android::AttachCurrentThread;
 using features::FeatureList;
 
+int AlignUp(int value, int alignment) {
```
borongc: As gemini suggested, this `AlignUp` is the same as in `starboard/shared/starboard/player/filter/audio_renderer_sink_impl.cc`, we can just define it in AudioRendererSink?

jasonzhangxx: Done

borongc (approved):

PR #10102, opened 2026-04-16
android: Fix OnFrameAvailable race condition

borongc:
Introduce thread-safe handling for the host object within
VideoSurfaceTextureBridge and its Java counterpart, VideoSurfaceTexture.
This change guards against concurrent access to the host pointer during
attachment, detachment, and frame availability callbacks.

Previously, a race condition could occur where OnFrameAvailable was
called after the native MediaCodecVideoDecoder host was destroyed. This
resulted in a use-after-free crash when invoking host->OnFrameAvailable()
on an invalid pointer. The new locking mechanism ensures the host
pointer is valid or null when accessed, preventing crashes.

#vibe-coded

Issue: 417242469
Issue: 503034869

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java
@@ -16,6 +16,7 @@ package dev.cobalt.media;
 
 import android.graphics.SurfaceTexture;
 import android.view.Surface;
+import androidx.annotation.GuardedBy;
 import org.jni_zero.CalledByNative;
 import org.jni_zero.JNINamespace;
 import org.jni_zero.NativeMethods;
@@ -26,6 +27,11 @@ import org.jni_zero.NativeMethods;
  */
 @JNINamespace("starboard")
 public class VideoSurfaceTexture extends SurfaceTexture {
+  private final Object mLock = new Object();
+
+  @GuardedBy("mLock")
+  private long mNativeVideoSurfaceTextureBridge;
+
   @CalledByNative
   VideoSurfaceTexture(int texName) {
     super(texName);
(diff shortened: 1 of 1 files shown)


starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -64,10 +67,11 @@ class VideoSurfaceTextureBridge {
       const base::android::JavaRef<jobject>& surface_texture,
       const base::android::JavaParamRef<jfloatArray>& mtx);
 
-  void OnFrameAvailable(JNIEnv*) { host_->OnFrameAvailable(); }
+  void OnFrameAvailable(JNIEnv*);
 
  private:
-  const raw_ref<Host> host_;
+  mutable base::Lock host_lock_;
```
osagie98: Should we use `std::mutex` with `std::lock_guard` instead? I don't think we're enforcing using the std library for locks but we may eventually

borongc: Done.

osagie98 (approved):

starboard/android/shared/video_surface_texture_bridge.h
```diff
@@ -33,15 +35,17 @@ class VideoSurfaceTextureBridge {
...
+  VideoSurfaceTextureBridge() = default;
   ~VideoSurfaceTextureBridge() = default;
 
   // Disallow copy and assign.
   VideoSurfaceTextureBridge(const VideoSurfaceTextureBridge&) = delete;
   VideoSurfaceTextureBridge& operator=(const VideoSurfaceTextureBridge&) =
       delete;
 
+  void DetachHost();
+  void AttachHost(Host* host);
```
jasonzhangxx: @borongc, can you verify if these change are needed to prevent callback being invoked after video decoder is destroyed? My understanding is `VideoSurfaceTextureBridge`'s lifecycle is fully controlled by `MediaCodecVideoDecoder`, not like `VideoSurfaceTexture.java` which has different lifecycle, so it should be ok to not have that extra protection in `VideoSurfaceTextureBridge`.

Awallky (approved):
//cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceTexture.java LGTM

starboard/android/shared/video_surface_texture_bridge.h
borongc: Right, java-side lock is sufficient to prevent callback after video decoder is destroyed. Removed these two functions.

jasonzhangxx (approved):

PR #10107, opened 2026-04-16
Modifications to the mac_build.sh file.

SortaCookie:
We mostly copy over seemingly relevant operations from dind_build.sh. Also we removed the gn installation and sccache from the mac_setup step.

Bug: 495549526

### DIFF
--- a/cobalt/devinfra/kokoro/bin/build_mac.sh
+++ b/cobalt/devinfra/kokoro/bin/build_mac.sh
@@ -20,10 +20,40 @@ pipeline () {
   # Run mac specific setup steps.
   setup_mac
 
+  local gclient_root="${KOKORO_ARTIFACTS_DIR}/git"
+  git config --global --add safe.directory "${gclient_root}/src"
+  local git_url="$(git -C "${gclient_root}/src" remote get-url origin)"
+
+  # Set up gclient and run sync.
+  ##############################################################################
+  cd "${gclient_root}"
+  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git tools/depot_tools --filter=blob:none
+  export PATH="${PATH}:${gclient_root}/tools/depot_tools"
+  gclient config --name=src "${git_url}"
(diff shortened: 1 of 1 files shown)


cobalt/devinfra/kokoro/bin/build_mac.sh
```diff
@@ -20,10 +20,40 @@ pipeline () {
...
   setup_mac
 
+  local gclient_root="${KOKORO_ARTIFACTS_DIR}/git"
+  git config --global --add safe.directory "${gclient_root}/src"
+  local git_url="$(git -C "${gclient_root}/src" remote get-url origin)"
+
+  # Set up gclient and run sync.
+  ##############################################################################
+  cd "${gclient_root}"
+  git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git tools/depot_tools
```
fayev-g: Comparing to what we do with github workflows, it might be worth adding `--filter=blob:none` and fetch depth 0 to optimize this clone

cobalt/devinfra/kokoro/bin/build_mac.sh
```diff
@@ -20,10 +20,40 @@ pipeline () {
...
+  export PATH="${PATH}:${gclient_root}/tools/depot_tools"
+  gclient config --name=src --custom-var='rbe_instance="projects/cobalt-actions-prod/instances/default_instance"' "${git_url}"
```
fayev-g: RBE isn't supported on mac so you can just remove this

cobalt/devinfra/kokoro/bin/build_mac.sh
oxve: +1

Fetch depth is 0 (full) by default which is what we want here. Feel free to add it in or not.

`--filter=blob:none` makes git lazy load blobs (historical file contents) as required. It's a good idea in any case.

oxve (approved):

PR #10110, opened 2026-04-16
ci: Unify trigger rules across all main workflows

oxve:
Also add permissions block where missing.

Bug: 503427712

### DIFF
--- a/.github/workflows/android.yaml
+++ b/.github/workflows/android.yaml
@@ -20,6 +20,8 @@ on:
         type: boolean
         default: false
 
+permissions: read-all
+
 jobs:
   chromium_android-arm64:
     uses: ./.github/workflows/main.yaml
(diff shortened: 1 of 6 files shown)


arjungm (approved):

isarkis (changes requested):
Why do we need to rebuild pipelines without on-device tests?

oxve:
I've addressed the feedback by removing `ready_for_review` from `linux.yaml` and `tvos.yaml`, as these pipelines do not have on-device tests. To ensure browser tests still run for Linux PRs, I've ungated `browser-test-on-host` in `main.yaml` so it runs on draft PRs (consistent with other host-based tests). `evergreen.yaml` and `raspi-2-modular.yaml` retain the trigger because they include gated on-device tests.

isarkis (approved):

PR #10112, opened 2026-04-16
cobalt: Fix unused function warning in app_event_runner.cc on Android

kjyoun:
The GetContentMainRunner() function and associated main_runner_ member variable were unused on Android because they were only used in a block excluded by #if !BUILDFLAG(IS_ANDROID).

> cobalt/app/app_event_runner.cc:58:29: warning: unused function 'GetContentMainRunner'

This warning is introduced by #9698

This change wraps the definition, declaration, usage, and related headers in #if !BUILDFLAG(IS_ANDROID) to fix the build warning.

Issue: 503436954


jellefoks (approved):
Good catch!

PR #10115, opened 2026-04-17
Don't refresh job status when labels other than runtest are added

andrewsavage1:


.github/workflows/main.yaml
```diff
@@ -26,7 +26,7 @@ on:
 
 concurrency:
   # Cancel existing jobs on the same workflow, platform, and branch (or sha if merged).
-  group: ${{ github.workflow }}-${{ inputs.platform }} @ ${{ github.head_ref || github.sha }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ inputs.platform }} @ ${{ github.head_ref || github.sha }}-${{ github.ref }}-${{ (github.event.action == 'labeled' && github.event.label.name != 'runtest') && 'ignored' || 'active' }}
```
oxve, line 29: Changing this doesn't affect triggering jobs, only cancelling existing ones.

I experimented with this a while ago and depending on timing it would either work as expected, or all jobs on the PR would be cancelled or skipped.

PR #10118, opened 2026-04-17
aosp: temporarily disable some Cobalt features

hferreiro:
On Android, //starboard depends on several //cobalt/browser/ targets.
This is a layering violation that additionally breaks the AOSP build,
since it forces the building of large parts of Chromium with
//copied_base/base instead of //base.
    
Disable client hint headers, deep link manager and accessibility manager
until the correct layering is implemented.

Bug: 492704919
Bug: 495203133


starboard/android/shared/BUILD.gn
```diff
@@ -312,9 +312,6 @@ static_library("starboard_platform") {
     ":starboard_base_symbolize",
     "//cobalt/android:jni_headers",
     "//cobalt/browser:cobalt_crash_annotations",
-    "//cobalt/browser/client_hint_headers",
```
haozheng-cobalt, line 315: Plz do not modify the list for common platforms, just remove the disabled features for AOSP by doing:

if (is_androidtv) {
    public_deps **-=** [ ... ]
}

starboard/android/shared/starboard_bridge.cc
```diff
@@ -28,6 +26,12 @@
 #include "starboard/common/time.h"
 #include "starboard/shared/starboard/audio_sink/audio_sink_internal.h"
 
+// TODO(b/492704919): enable on AOSP when the layering violation is fixed.
+#if BUILDFLAG(IS_ANDROIDTV)
```
haozheng-cobalt: should use #if **!**BUILDFLAG(IS_ANDROIDTV) to remove the features.

starboard/android/shared/starboard_bridge.cc
```diff
@@ -121,6 +125,8 @@ void JNI_StarboardBridge_InitializePlatformAudioSink(JNIEnv* env) {
 void JNI_StarboardBridge_HandleDeepLink(JNIEnv* env,
                                         const JavaParamRef<jstring>& jurl,
                                         jboolean applicationStarted) {
+  // TODO(b/492704919): enable on AOSP when the layering violation is fixed.
+#if BUILDFLAG(IS_ANDROIDTV)
```
haozheng-cobalt: same as above, use negation to exclude the code.

haozheng-cobalt (changes requested):

starboard/android/shared/starboard_bridge.cc
hferreiro: Do you mean you prefer:
```
#if !BUILDFLAG(IS_ANDROIDTV)
return;
#else
[original code]
#endif 
```

starboard/android/shared/BUILD.gn
hferreiro: Done.

hferreiro:
In the latest commit, I switched from `is_androidtv` to `is_partner_toolchain`, which I'd say is more correct.

build/build_config.h
```diff
@@ -349,6 +349,12 @@
 #define BUILDFLAG_INTERNAL_IS_STARBOARD_TOOLCHAIN() (0)
 #endif
 
+#if defined(ENABLE_BUILDFLAG_IS_PARTNER_TOOLCHAIN)
```
haozheng-cobalt, line 354: link

hferreiro: What do you mean?

haozheng-cobalt: Sorry, was about to share this line of code to the team for review. This is good to me.

haozheng-cobalt (approved):

niranjanyardi:
> In the latest commit, I switched from `is_androidtv` to `is_partner_toolchain`, which I'd say is more correct.

i dont think we should add another flag to the buildconfig - if existing flags work.

instead of is_android_tv maybe you can use is_starboard && is_android . 

i dont think we should define a new flag

build/build_config.h
niranjanyardi, line 354: i dont think we should add a new flag here if exsting flags are finxing the issue .  maybe IS_STARBOARD && IS_ANDROIDTV are good for your use case.

highly recommend getting Andrew's review on this if you really want to add it .

We are trying to intentionally keep the number of global buildflags to as much as needed

andrewsavage1: I think we should add this buildflag with a TODO to deprecate it once the usages of it (removing the dependency on cobalt in the starboard android impl) is addressed

One option here is to have separate aosp/atv starboard_bridge/text_to_speech_helper source files. That may be the preferred way to remove this macro if the dependency can't be removed cleanly in shared code

andrewsavage1 (approved):

build/build_config.h
hferreiro: Done.

PR #10124, opened 2026-04-17
feat(starboard): Add prctl wrapper for VMA naming 

Awallky:
- Implement architecture-independent prctl wrappers in the modular
  Starboard and Cobalt layers.
- Define ABI-stable MUSL_PR_* constants to ensure compatibility between
  the musl-based Cobalt layer and the Starboard platform layer.
- Add support for prctl operation for anonymous memory region naming (PR_SET_VMA).
- Update modular build configuration to include the new wrapper files.

Bug: 443798603


starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
```diff
@@ -17,11 +17,14 @@
 #include <errno.h>
 #include <linux/prctl.h>
 #include <signal.h>
-#include <stdarg.h>
+#include <stdio.h>
 #include <sys/prctl.h>
+#include <unistd.h>
```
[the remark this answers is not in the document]
Awallky: It is already added here. I think we already meet the requirements here.

starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
```diff
@@ -274,10 +287,18 @@ int __abi_wrap_prctl(int op, ...) {
...
+      pid = (pid == (long)MUSL_PR_SET_PTRACER_ANY) ? PR_SET_PTRACER_ANY : pid;
       ret = prctl(platform_op, pid);
       break;
     }
+    case PR_SET_VMA: {
+      unsigned long arg2 = va_arg(args, unsigned long);
+      unsigned long arg3 = va_arg(args, unsigned long);
+      unsigned long arg4 = va_arg(args, unsigned long);
+      unsigned long arg5 = va_arg(args, unsigned long);
+      ret = prctl(platform_op, arg2, arg3, arg4, arg5);
```
jellefoks: Since we know that this is PR_SET_VMA, we also know that these are `attr, addr, size, val`.

And attr needs a translation for MUSL_PR_SET_VMA_ANON_NAME to the platform specific value.

starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
```diff
@@ -18,10 +18,18 @@
...
 #include <signal.h>
 #include <stdarg.h>
+#include <stdio.h>
 #include <sys/prctl.h>
+#include <unistd.h>
 
 #include "starboard/common/log.h"
 
+#ifndef PR_SET_VMA
+#define PR_SET_VMA 0x53564d41
```
jellefoks, line 34: Maybe add a clarifying comment similar to the one at sandbox/linux/system_headers/linux_prctl.h

starboard/elf_loader/exported_symbols.cc
```diff
@@ -235,6 +235,7 @@ ExportedSymbols::ExportedSymbols() {
 
   // POSIX APIs
   REGISTER_SYMBOL(aligned_alloc);
+  REGISTER_SYMBOL(atexit);
```
jellefoks: It's surprising that you added this, since we already use musl's `exit()` and `atexit()`.

Let's a least file a followup ticket for this with a clear description why this was added here, because we should not actually need this, nor will this always do what we want. for example, we may be loaded as a plugin and the platform's atexit() may not actually run when we are stopped. We also don't call `exit()` from Cobalt, it's not even in Starboard. 

Besides that, any non-crashing exits go through the stop lifecycle event, so we can trigger the atexit handler there. Maybe that's the bug, that we're not calling musl's `exit()` and/or musl's `exit()` doesn't call musl's `atexit()` for Starboard platforms.

starboard/nplb/posix_compliance/posix_prctl_test.cc
```diff
@@ -12,18 +12,138 @@
...
+  if (option == PR_SET_VMA) {
+    arg3 = va_arg(args, unsigned long);
+    arg4 = va_arg(args, unsigned long);
+    arg5 = va_arg(args, unsigned long);
+  }
+  va_end(args);
+
+  int result = prctl(option, arg2, arg3, arg4, arg5);
+
+  if (result == -1 && option == PR_SET_VMA && arg2 == PR_SET_VMA_ANON_NAME) {
```
jellefoks: Instead of copying the code from the wrapper here, the actual code in the wrapper should be tested. There is no guarantee that this code is exactly the same and will always behave exactly the same as the code in the other file.

starboard/shared/modular/cobalt_layer_posix_prctl_abi_wrappers.cc
```diff
@@ -39,6 +81,64 @@ int prctl(int op, ...) {
...
+          }
+        }
+
+        if (fd != -1) {
+          char buf[256];
+          int len = snprintf(buf, sizeof(buf), "0x%lx 0x%lx %s\n", arg3,
+                             arg3 + arg4, (const char*)arg5);
+          if (len > 0) {
+            size_t write_len = std::min((size_t)len, sizeof(buf) - 1);
+            if (write(fd, buf, write_len) != -1) {
```
jellefoks: Now that this data is written to the file, when/from where is it used? And since it's a randomized filename, how does the user know which file to use?

starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
Awallky: Done.

starboard/elf_loader/exported_symbols.cc
Awallky: Ah, that is my mistake. It was a large rebase from the last commit a few months ago. Likely just an artifact from it. I have since removed it.

Awallky: Everything works as intended without the addition.

starboard/shared/modular/cobalt_layer_posix_prctl_abi_wrappers.cc
Awallky: Good point! I added a log statement so users can find the file by observing the SB logs.

```
+ SB_LOG(INFO) << "Created VMA tags fallback file: "
+                          << g_vma_tags_file_path;
```

WDYT?

starboard/nplb/posix_compliance/posix_prctl_test.cc
Awallky: Yeah good point. I have removed the wrapper code and directly used the real thing.

starboard/shared/modular/cobalt_layer_posix_prctl_abi_wrappers.cc
jellefoks: Great

jellefoks (approved):

starboard/shared/modular/starboard_layer_posix_prctl_abi_wrappers.cc
Awallky: Fixed.

y4vor (approved):

PR #10125, opened 2026-04-17
[debug] Additional logging to pinpoint Samsung issue

sideb0ard:


media/webrtc/audio_processor.cc
kjyoun, line 635: Can you add log here to confirm that output_sample_rate is changed properly?

```
LOG(INFO) << "SAMSUNG DEBUG - output_sample_rate is " << output_sample_rate;
```

sideb0ard: done.

PR #10129, opened 2026-04-17
starboard: Migrate MediaCodecDecoder to use sb::Thread

kjyoun:
Migrate MediaCodecDecoder's video input and output threads from raw
pthreads to the Starboard Thread abstraction. This change implements
VideoInputThread and VideoOutputThread as nested classes inheriting
from Thread, replacing pthread_t handles with std::unique_ptr<Thread>.

This improves consistency with the existing decoder_thread_, simplifies
thread lifecycle management by using Thread::Start() and Thread::Join(),
and removes redundant base::android::DetachFromVM() calls, as these are
now handled by sb::Thread's TerminateOnThread on Android. The change
also cleans up unnecessary headers and updates back-pointers to be const.

Bug: 468356861

### DIFF
--- a/starboard/android/shared/media_decoder.cc
+++ b/starboard/android/shared/media_decoder.cc
@@ -90,7 +90,32 @@ class MediaCodecDecoder::DecoderThread : public Thread {
   }
 
  private:
-  MediaCodecDecoder* decoder_;
+  MediaCodecDecoder* const decoder_;
+};
+
+class MediaCodecDecoder::VideoInputThread : public Thread {
+ public:
+  explicit VideoInputThread(MediaCodecDecoder* decoder)
+      : Thread("VidDecIn"), decoder_(decoder) {}
+
+  void Run() override { decoder_->InputThreadFunc(); }
+
+ private:
+  MediaCodecDecoder* const decoder_;
+};
+
+class MediaCodecDecoder::VideoOutputThread : public Thread {
+ public:
+  explicit VideoOutputThread(MediaCodecDecoder* decoder)
+      : Thread("VidDecOut"), decoder_(decoder) {}
+
+  void Run() override {
+    SbThreadSetPriority(kSbThreadPriorityHigh);
+    decoder_->OutputThreadFunc();
+  }
+
+ private:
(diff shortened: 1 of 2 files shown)


starboard/android/shared/media_decoder.cc
```diff
@@ -293,13 +318,11 @@ void MediaCodecDecoder::WriteInputBuffers(const InputBuffers& input_buffers) {
...
-                     &MediaCodecDecoder::InputThreadEntryPoint, this);
-      pthread_create(&video_output_thread_, nullptr,
-                     &MediaCodecDecoder::OutputThreadEntryPoint, this);
-      SB_DCHECK_NE(video_input_thread_, 0);
-      SB_DCHECK_NE(video_output_thread_, 0);
+    if (!video_input_thread_) {
+      video_input_thread_ = std::make_unique<VideoInputThread>(this);
+      video_input_thread_->Start();
+      video_output_thread_ = std::make_unique<VideoOutputThread>(this);
+      video_output_thread_->Start();
```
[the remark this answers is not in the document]
kjyoun: will stick current code. 

create-start pair is more readable

starboard/android/shared/media_decoder.cc
```diff
@@ -90,7 +90,32 @@ class MediaCodecDecoder::DecoderThread : public Thread {
   }
 
  private:
-  MediaCodecDecoder* decoder_;
+  MediaCodecDecoder* const decoder_;
+};
+
+class MediaCodecDecoder::VideoInputThread : public Thread {
+ public:
+  explicit VideoInputThread(MediaCodecDecoder* decoder)
```
xiaomings: Consider passing the thread function as an `std::function<>` and the thread priority as ctor parameter so we only need one class?

kjyoun: Ack. 

Will consider it

kjyoun: Done.

Yeah, code becomes more concise. thanks for the suggestion!

xiaomings (approved):

PR #10130, opened 2026-04-17
Clean up OverlayFS workspace action

linxinan-chops:
Refactor OverlayFS workspace action to concentrate on mounting the layered workspace.

The checkout and `gclient sync` should leave to the corresponding action steps.

Bug: 501486531

### DIFF
--- a/.github/actions/checkout/action.yaml
+++ b/.github/actions/checkout/action.yaml
@@ -84,8 +84,6 @@ runs:
         if [ ! -d "${DEST_PATH}/.git" ]; then
           git clone \
             --filter=blob:none \
-            --reference-if-able "${CACHE_PATH}" \
-            --dissociate \
             "${depth_arg[@]}" \
             "${REPO_URL}" \
             "${DEST_PATH}"
--- a/.github/actions/checkout/action.yaml
+++ b/.github/actions/checkout/action.yaml
@@ -81,15 +81,19 @@ runs:
           depth_arg+=(--depth "${FETCH_DEPTH}")
         fi
 
-        git clone \
-          --filter=blob:none \
-          --reference-if-able "${CACHE_PATH}" \
-          --dissociate \
-          "${depth_arg[@]}" \
-          "${REPO_URL}" \
-          "${DEST_PATH}"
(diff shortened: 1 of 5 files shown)


.github/actions/checkout/action.yaml
```diff
@@ -81,15 +81,19 @@ runs:
...
+            "${REPO_URL}" \
+            "${DEST_PATH}"
+        else
+          echo "Repository already exists at ${DEST_PATH}. Skipping clone."
+        fi
 
         git config --global --add safe.directory "${GITHUB_WORKSPACE}/${DEST_PATH}"
 
-        git -C "${DEST_PATH}" fetch "${deptch_arg[@]}" origin "${REF}"
+        git -C "${DEST_PATH}" fetch "${depth_arg[@]}" origin "${REF}"
```
oxve: Thanks. Been looking at that line for a while now.

.github/actions/checkout/action.yaml
```diff
@@ -81,15 +81,19 @@ runs:
...
-          --filter=blob:none \
-          --reference-if-able "${CACHE_PATH}" \
-          --dissociate \
-          "${depth_arg[@]}" \
-          "${REPO_URL}" \
-          "${DEST_PATH}"
+        if [ ! -d "${DEST_PATH}/.git" ]; then
+          git clone \
+            --filter=blob:none \
+            --reference-if-able "${CACHE_PATH}" \
```
oxve: Let's clean this up later. The old cache version just adds unnecessary complexity now.

.github/actions/checkout/action.yaml
```diff
@@ -81,15 +81,19 @@ runs:
...
+            --dissociate \
```
oxve: We should remove these now that we have the overlay instead of the bare .git folder. It doesn't break anything but it's unnecessary clutter now.

```
            --reference-if-able "${CACHE_PATH}" \
            --dissociate \
```

oxve (approved):
Looks nice and clean!

.github/actions/checkout/action.yaml
linxinan-yt: Cool. Thanks for pointing it.

PR #10131, opened 2026-04-17
cobalt/metrics: Add SmapsCategorizer and DetailedMetricsDelegate stubs

Awallky:
Introduce SmapsCategorizer and DetailedMetricsDelegate to enable
Cobalt-specific detailed memory metrics collection.

Note: To this point in the PR chain, the changes effectively do nothing. See PR #10140 for the full diff of the refactor.  The focus here should be on the performance implications of the smaps categorizer and the delegate interface. 

The SmapsCategorizer provides a mechanism to parse /proc/self/smaps
on a background thread, extracting granular memory usage per region.
The DetailedMetricsDelegate serves as an extensible interface for
downstream consumers to process and aggregate this raw smaps data,
allowing for fine-grained analysis of Cobalt's memory footprint.

Bug: 494004530

### DIFF
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -47,11 +47,11 @@ source_set("lib") {
       "//services/resource_coordinator/public/cpp/memory_instrumentation",
       "//services/tracing/public/cpp",
     ]
+    deps -= [ services/resource_coordinator/BUILD.gn ]
     public_deps -= [
       "//services/metrics/public/cpp:metrics_cpp",
       "//services/metrics/public/cpp:ukm_builders",
       "//services/metrics/public/mojom",
-      "//third_party/perfetto:libperfetto",
     ]
   }
 }
@@ -70,8 +70,6 @@ source_set("tests") {
     "public/cpp/memory_instrumentation/tracing_integration_unittest.cc",
     "public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc",
   ]
-
-
 
   deps = [
     ":lib",
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -38,6 +38,20 @@ source_set("lib") {
     "//services/resource_coordinator/public/cpp:resource_coordinator_cpp",
     "//third_party/perfetto:libperfetto",
   ]
+
+  if (is_cobalt) {
+    public_deps += [ "//services/resource_coordinator/public/mojom" ]
+    deps = [
+      "//services/resource_coordinator/public/cpp/memory_instrumentation",
+      "//services/tracing/public/cpp",
+    ]
+    public_deps -= [
+      "//services/metrics/public/cpp:metrics_cpp",
+      "//services/metrics/public/cpp:ukm_builders",
+      "//services/metrics/public/mojom",
+      "//third_party/perfetto:libperfetto",
+    ]
+  }
 }
 
 source_set("tests") {
@@ -54,6 +68,8 @@ source_set("tests") {
     "public/cpp/memory_instrumentation/tracing_integration_unittest.cc",
     "public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc",
   ]
+
+
 
   deps = [
     ":lib",
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -47,7 +47,6 @@ source_set("lib") {
       "//services/resource_coordinator/public/cpp/memory_instrumentation",
       "//services/tracing/public/cpp",
     ]
-    deps -= [ services/resource_coordinator/BUILD.gn ]
     public_deps -= [
       "//services/metrics/public/cpp:metrics_cpp",
       "//services/metrics/public/cpp:ukm_builders",
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -39,9 +39,11 @@ source_set("lib") {
     "//third_party/perfetto:libperfetto",
   ]
 
+  deps = []
+
   if (is_cobalt) {
     public_deps += [ "//services/resource_coordinator/public/mojom" ]
-    deps = [
+    deps += [
       "//services/resource_coordinator/public/cpp/memory_instrumentation",
       "//services/tracing/public/cpp",
     ]
--- a/services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
+++ b/services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
@@ -1,0 +1,265 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.h"
+
+#include "services/resource_coordinator/public/cpp/memory_instrumentation/os_metrics.h"
+#include "base/functional/bind.h"
+#include "base/functional/callback.h"
+#include "base/files/file.h"
+#include "base/files/file_path.h"
+#include "base/metrics/histogram_functions.h"
+#include "base/strings/string_number_conversions.h"
+#include "base/strings/string_util.h"
+#include "base/task/thread_pool.h"
+#include "base/threading/scoped_blocking_call.h"
+#include "third_party/abseil-cpp/absl/strings/ascii.h"
+#include "third_party/abseil-cpp/absl/strings/match.h"
+#include "third_party/abseil-cpp/absl/strings/numbers.h"
+#include "third_party/abseil-cpp/absl/strings/string_view.h"
+
+namespace memory_instrumentation {
+
+namespace {
+
+enum class SmapsReliability {
+  kSuccess = 0,
+  kRollupMissing = 1,
+  kAccessDenied = 2,
+  kUnreliable = 3,
+  kMaxValue = kUnreliable,
+};
+
+// Helper to parse a line like "Pss:                328 kB"
+bool ParseSmapsLine(absl::string_view line, uint64_t* value_kb) {
+  size_t colon_pos = line.find(':');
+  if (colon_pos == absl::string_view::npos) return false;
+
+  absl::string_view value_part = line.substr(colon_pos + 1);
+  // Trim leading spaces
+  while (!value_part.empty() && absl::ascii_isspace(value_part.front())) {
+    value_part.remove_prefix(1);
+  }
+  
+  // Find the end of the numeric part
+  size_t space_pos = value_part.find(' ');
+  if (space_pos != absl::string_view::npos) {
+    value_part = value_part.substr(0, space_pos);
+  }
+
+  return absl::SimpleAtoi(value_part, value_kb);
+}
+
+}  // namespace
+
+SmapsCategorizer::SmapsCategorizer(base::WeakPtr<DetailedMetricsDelegate> delegate)
+    : delegate_(std::move(delegate)),
+      task_runner_(base::SequencedTaskRunner::GetCurrentDefault()) {
+}
+
+SmapsCategorizer::~SmapsCategorizer() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+}
+
+void SmapsCategorizer::RequestDump(base::OnceClosure callback) {
+  if (!task_runner_->RunsTasksInCurrentSequence()) {
+    task_runner_->PostTask(
+        FROM_HERE,
+        base::BindOnce(&SmapsCategorizer::RequestDump,
+                       weak_ptr_factory_.GetWeakPtr(), std::move(callback)));
+    return;
+  }
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  pending_callbacks_.push_back(std::move(callback));
+  if (is_scanning_) return;
+
+  is_scanning_ = true;
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::MayBlock(), base::TaskPriority::BEST_EFFORT},
+      base::BindOnce(&SmapsCategorizer::PerformScanOnBackgroundThread, delegate_.get()),
+      base::BindOnce(&SmapsCategorizer::OnScanComplete,
+                     weak_ptr_factory_.GetWeakPtr()));
+}
+
+void SmapsCategorizer::OnScanComplete(bool success) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  is_scanning_ = false;
+  
+  base::UmaHistogramExactLinear("Memory.SmapsCategorizer.CoalescedRequestsCount",
+                                 static_cast<int>(pending_callbacks_.size()), 50);
+  
+  for (auto& cb : pending_callbacks_) std::move(cb).Run();
+  pending_callbacks_.clear();
+}
+
+// static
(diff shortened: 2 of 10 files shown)


services/resource_coordinator/BUILD.gn
```diff
@@ -38,6 +38,20 @@ source_set("lib") {
     "//services/resource_coordinator/public/cpp:resource_coordinator_cpp",
     "//third_party/perfetto:libperfetto",
   ]
+
+  if (is_cobalt) {
+    public_deps += [ "//services/resource_coordinator/public/mojom" ]
+    deps = [
```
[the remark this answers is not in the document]
Awallky: Fixed. Added `deps = []` before this block.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,265 @@
...
+    absl::string_view processable = chunk.substr(0, last_newline + 1);
+    while (!processable.empty()) {
+      size_t newline_pos = processable.find('\n');
+      absl::string_view line = processable.substr(0, newline_pos);
+      processable.remove_prefix(newline_pos + 1);
+
+      if (line.empty())
+        continue;
+
+      if (absl::ascii_isxdigit(line[0])) {
```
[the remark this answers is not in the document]
Awallky: Done.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,265 @@
...
+  base::ScopedBlockingCall scoped_blocking_call(FROM_HERE,
+                                                base::BlockingType::MAY_BLOCK);
+
+  if (!delegate || !file.IsValid())
+    return false;
+
+  char buffer[4096];
+  size_t buffer_offset = 0;
+
+  absl::string_view current_name;
```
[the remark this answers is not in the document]
Awallky: Done.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,265 @@
...
+    return;
+  }
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  pending_callbacks_.push_back(std::move(callback));
+  if (is_scanning_) return;
+
+  is_scanning_ = true;
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::MayBlock(), base::TaskPriority::BEST_EFFORT},
+      base::BindOnce(&SmapsCategorizer::PerformScanOnBackgroundThread, delegate_.get()),
```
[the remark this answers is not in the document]
Awallky: The raw pointer is assumed to be safe as the delegate is expected to outlive the task. I avoided changing the interface to  scoped_refptr  to keep the changes minimal.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  return absl::SimpleAtoi(value_part, value_kb);
+}
+
+}  // namespace
+
+SmapsCategorizer::SmapsCategorizer(base::WeakPtr<DetailedMetricsDelegate> delegate)
+    : delegate_(std::move(delegate)),
+      task_runner_(base::SequencedTaskRunner::GetCurrentDefault()) {
```
yell0wd0g: nit: Since we have this |task_runner_|, |sequence_checker_| is
redundant: I'd remove the latter entirely and just 
`DCHECK(task_runner_->RunsTasksInCurrentSequence())` instead
(if that looks unclear you could make a private method and
name it more clearly).

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,262 @@
...
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  pending_callbacks_.push_back(std::move(callback));
+  if (is_scanning_) return;
+
+  is_scanning_ = true;
+  // NOTE: Passing raw pointer is assumed safe here as DetailedMetricsDelegate
+  // is expected to outlive this task.
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::MayBlock(), base::TaskPriority::BEST_EFFORT},
+      base::BindOnce(&SmapsCategorizer::PerformScanOnBackgroundThread, delegate_.get()),
```
yell0wd0g: I have a slightly different concern but in the same area, i.e.
`delegate_` is a WeakPtr, and since it's called from a thread
of this `base::ThreadPool` (l.146 `OnSmapsEntry` IIUC), you 
need to have this naked pointer to avoid DCHECKing the WeakPtr
dereference. 

To avoid that naked pointer, could you just pass to
`PerformScanOnBackgroundThread()`, ScanSmapsFile()`, 
`ScanSmaps()` a bound `OnSmapsEntry()` (using `delegate_`
directly as a WeakPtr) and then create that bound call
with `BindPostTaskToCurrentDefault`? I.e.

```

  auto bound_on_smaps_entry = 
    base::BindPostTaskToCurrentDefault(
        base::BindOnce(&DetailedMetricsDelegate::OnSmapsEntry,
                       delegate_));

  // Then the post-task-and-reply-with-result...
  base::ThreadPool::PostTaskAndReplyWithResult(
      FROM_HERE, {base::MayBlock(), base::TaskPriority::BEST_EFFORT},
      std::move(bound_on_smaps_entry),
      base::BindOnce(&SmapsCategorizer::OnScanComplete,
                     weak_ptr_factory_.GetWeakPtr()));
```

For the signatures of eg `PerformScanOnBackgroundThread()` and
friends, we could either deduce the arguments of `OnSmapsEntry`
via some probably very verbose decltype or just make a signature
ourselves in the class declaration via `using Bla = base::OnceCallback()` 
and then 

```
bool SmapsCategorizer::PerformScanOnBackgroundThread(Bla bound_callback) 
```

It's conceptually more complex having to think about so many 
jumps and redirections :-) but should be more robust long-term.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.h
```diff
@@ -0,0 +1,73 @@
...
+  explicit SmapsCategorizer(base::WeakPtr<DetailedMetricsDelegate> delegate);
+  ~SmapsCategorizer();
+
+  SmapsCategorizer(const SmapsCategorizer&) = delete;
+  SmapsCategorizer& operator=(const SmapsCategorizer&) = delete;
+
+  // Requests a memory dump. If a scan is already in progress, the callback
+  // will be queued and called when the current scan completes.
+  // Must be called on the same sequence the categorizer was created on.
+  void RequestDump(base::OnceClosure callback);
```
yell0wd0g: QQ: If `callback` is meant to mean "it's done" (seems so from
the implementation), could we `s/callback/done/`?

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,262 @@
...
+    : delegate_(std::move(delegate)),
+      task_runner_(base::SequencedTaskRunner::GetCurrentDefault()) {
+}
+
+SmapsCategorizer::~SmapsCategorizer() {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+}
+
+void SmapsCategorizer::RequestDump(base::OnceClosure callback) {
+  if (!task_runner_->RunsTasksInCurrentSequence()) {
```
yell0wd0g: This is a common pattern to allow `RequestDump()` to be called
from any TaskRunner. Is it really needed? Either way it'd be great
to call it out in the class comment (currently says "Use a 
sequence-affine front-end to handle request coalescing and a 
background ThreadPool task for I/O and parsing.")

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,262 @@
...
+        FROM_HERE,
+        base::BindOnce(&SmapsCategorizer::RequestDump,
+                       weak_ptr_factory_.GetWeakPtr(), std::move(callback)));
+    return;
+  }
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  pending_callbacks_.push_back(std::move(callback));
+  if (is_scanning_) return;
+
+  is_scanning_ = true;
```
yell0wd0g: QQ could we do here
```
  const bool is_scanning = !pending_callbacks_.empty();
  pending_callbacks_.push_back(std::move(callback));
  if (is_scanning) return;
```

Essentially using `!pending_callbacks_.empty()` as `is_scanning_`,
which would render the latter unnecessary?

yell0wd0g (approved):
LGTM with some Qs / suggestions.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: I think wee actually want this here, since developers might call  `RequestDump`  from a random thread, which would break this  `PostTaskAndReply`  flow and cause  `weak_ptr_factory_`  to crash.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: We can't guarantee this isn't called somewhere else in the future, especially for upstream. So, yeah I think we should keep it as-is. Will add a comment.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: Done. Though I moved it to a method so it is more readable.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: So I went a bit of a different route so we did not need to rely on the pointer at all:

I refactored `SmapsCategorizer` for thread safety and re-entrancy.
    
    - Implement "Accumulate and Return" pattern for SmapsCategorizer background parsing to avoid passing raw DetailedMetricsDelegate* to background thread.
    - Implement "swap" (copy smaps summary data) pattern in `OnScanComplete` to ensure re-entrancy safety when executing callbacks.

WDYT?

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,274 @@
...
+// Helper to parse a line like "Pss:                328 kB"
+bool ParseSmapsLine(absl::string_view line, uint64_t* value_kb) {
+  size_t colon_pos = line.find(':');
+  if (colon_pos == absl::string_view::npos) return false;
+
+  absl::string_view value_part = line.substr(colon_pos + 1);
+  // Trim leading spaces
+  while (!value_part.empty() && absl::ascii_isspace(value_part.front())) {
+    value_part.remove_prefix(1);
+  }
```
yell0wd0g: nit: Could we use base/strings/string_tokenizer.h for this?

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
yell0wd0g: I don't understand, in l.88 with the current code
we're guaranteed to be on |task_runner_|, which is
where |weak_ptr_factory_| is bound.  Could you 
elaborate on the scenario plz?

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
yell0wd0g: Is there actually an actual path that calls from a different
thread/TaskRunner? If there currently are not such paths,
the preferred way is to leave a CHECK and let the developers
of the future extend this class. Another version of YAGNI.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,274 @@
...
+  base::ScopedBlockingCall scoped_blocking_call(FROM_HERE,
+                                                base::BlockingType::MAY_BLOCK);
+
+  if (!file.IsValid())
+    return std::nullopt;
+
+  ParsedSmapsResults results;
+  results.reserve(128);
+
+  char buffer[4096];
```
yell0wd0g, line 133: nit: I always feel uneasy when I see naked arrays/strings.
I scanned the code and I couldn't quite point why or how 
to make it better, but in general parsing things by hand
should be avoided. Could we let Gemini take a pass at this
after landing and see if using base:: Tokenizer classes
looks good?

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.h
```diff
@@ -0,0 +1,88 @@
...
+#include "base/functional/callback_forward.h"
+#include "base/component_export.h"
+#include "base/memory/weak_ptr.h"
+#include "base/sequence_checker.h"
+#include "base/task/sequenced_task_runner.h"
+#include "services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate.h"
+
+namespace memory_instrumentation {
+
+struct ParsedSmapsEntry {
```
yell0wd0g: nit: Technically this should be a class (public members is
fine) since it has class members.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,274 @@
...
+  callbacks.swap(pending_callbacks_);
+  
+  for (auto& cb : callbacks) {
+    std::move(cb).Run();
+  }
+}
+
+// static
+std::optional<ParsedSmapsResults> SmapsCategorizer::PerformScanOnBackgroundThread() {
+  return ScanSmapsFile(base::FilePath("/proc/self/smaps"));
```
yell0wd0g: Can we bind directly ScanSmapsFile in l.95?

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,274 @@
...
+  if (!file.IsValid()) {
+    base::UmaHistogramEnumeration("Memory.SmapsCategorizer.Reliability",
+                                  SmapsReliability::kAccessDenied);
+    return std::nullopt;
+  }
+  return ScanSmaps(std::move(file));
+}
+
+// static
+std::optional<ParsedSmapsResults> SmapsCategorizer::ScanSmaps(
```
yell0wd0g, line 122: Can we make this static method a file-static function?
Maybe there are plans for the future, if so leave a
TODO please.

yell0wd0g (commented):
LGTM _if_ `SmapsCategorizer::RequestDump()` trampoline 
question is solved definitively.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: I think `base::SplitStringPiece` does the job just as well for us.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: Done.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.h
Awallky: Done.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: Done.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
Awallky: Yeah, this is used in a later PR. I left a TODO indicating this.

PR #10132, opened 2026-04-17
cobalt/metrics: Fix OSMemDump constructor in unit tests

Awallky:
Implement granular memory tracking and SmapsCategorizer in Cobalt
- Added interfaces and stubs for granular memory tracking.
- Implemented SmapsCategorizer for thread-safe, high-performance smaps parsing.
- Added unit tests for SmapsCategorizer and DetailedMetricsDelegate, with Cobalt-specific tracing observers safely guarded.

Note: A follow-up will be added to run these tests in CI. This serves as a follow-up to #10131, which introduced the units under test.

Test: `run_cobalt_memory_instrumentation_unittests --gtest-filter="TracingObserverProtoTest.*:SmapsCategorizerTest.*:DetailedMetricsDelegateTest.*:SmapsMetricsTest.*"`
Bug: 506209440

### DIFF
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -38,6 +38,20 @@ source_set("lib") {
     "//services/resource_coordinator/public/cpp:resource_coordinator_cpp",
     "//third_party/perfetto:libperfetto",
   ]
+
+  if (is_cobalt) {
+    public_deps += [ "//services/resource_coordinator/public/mojom" ]
+    deps = [
+      "//services/resource_coordinator/public/cpp/memory_instrumentation",
+      "//services/tracing/public/cpp",
+    ]
+    public_deps -= [
+      "//services/metrics/public/cpp:metrics_cpp",
+      "//services/metrics/public/cpp:ukm_builders",
+      "//services/metrics/public/mojom",
+      "//third_party/perfetto:libperfetto",
+    ]
+  }
 }
 
 source_set("tests") {
@@ -54,6 +68,8 @@ source_set("tests") {
     "public/cpp/memory_instrumentation/tracing_integration_unittest.cc",
     "public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc",
   ]
+
+
 
   deps = [
     ":lib",
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -83,3 +83,27 @@ source_set("tests") {
     "//testing/gtest",
   ]
 }
+
+if (is_cobalt) {
+  import("//testing/test.gni")
+
+  test("cobalt_memory_instrumentation_unittests") {
+    testonly = true
+    sources = [
+      "public/cpp/memory_instrumentation/detailed_metrics_delegate_unittest.cc",
+      "public/cpp/memory_instrumentation/smaps_categorizer_unittest.cc",
+      "public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc",
+    ]
+    deps = [
+      ":lib",
+      "//base",
+      "//base/test:test_support",
+      "//mojo/core/test:run_all_unittests",
+      "//services/resource_coordinator/public/cpp:resource_coordinator_cpp",
+      "//services/resource_coordinator/public/cpp/memory_instrumentation:memory_instrumentation",
+      "//services/tracing:test_utils",
+      "//testing/gmock",
+      "//testing/gtest",
+    ]
+  }
+}
--- a/services/resource_coordinator/BUILD.gn
+++ b/services/resource_coordinator/BUILD.gn
@@ -69,7 +69,12 @@ source_set("tests") {
     "public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc",
   ]
 
-
+  if (is_cobalt) {
+    sources += [
+      "public/cpp/memory_instrumentation/detailed_metrics_delegate_unittest.cc",
+      "public/cpp/memory_instrumentation/smaps_categorizer_unittest.cc",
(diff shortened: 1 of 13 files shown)


services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,265 @@
...
+  base::ScopedBlockingCall scoped_blocking_call(FROM_HERE,
+                                                base::BlockingType::MAY_BLOCK);
+
+  if (!delegate || !file.IsValid())
+    return false;
+
+  char buffer[4096];
+  size_t buffer_offset = 0;
+
+  absl::string_view current_name;
```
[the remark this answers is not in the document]
Awallky: Fixed in a different PR.

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,265 @@
...
+    absl::string_view processable = chunk.substr(0, last_newline + 1);
+    while (!processable.empty()) {
+      size_t newline_pos = processable.find('\n');
+      absl::string_view line = processable.substr(0, newline_pos);
+      processable.remove_prefix(newline_pos + 1);
+
+      if (line.empty())
+        continue;
+
+      if (absl::ascii_isxdigit(line[0])) {
```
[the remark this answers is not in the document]
Awallky: Fixed in a different PR

services/resource_coordinator/public/cpp/memory_instrumentation/smaps_categorizer.cc
```diff
@@ -0,0 +1,265 @@
...
+// static
+bool SmapsCategorizer::PerformScanOnBackgroundThread(
+    DetailedMetricsDelegate* delegate) {
+#if BUILDFLAG(IS_COBALT)
+  base::File file(base::FilePath("/proc/self/smaps"),
+                    base::File::FLAG_OPEN | base::File::FLAG_READ);
+  return ScanSmaps(std::move(file), delegate);
+#else
+  return ScanSmapsFile(base::FilePath("/proc/self/smaps"), delegate);
+#endif
```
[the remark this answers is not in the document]
Awallky: Fixed in a different PR

services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate_unittest.cc
```diff
@@ -0,0 +1,44 @@
+// Copyright 2026 The Chromium Authors
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
niranjanyardi: should this be a cobalt copyright ?

services/resource_coordinator/public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc
```diff
@@ -131,6 +131,15 @@ memory_instrumentation::mojom::OSMemDump GetFakeOSMemDump(
     uint32_t resident_set_kb,
     uint32_t private_footprint_kb,
     uint32_t shared_footprint_kb) {
+#if BUILDFLAG(IS_COBALT)
```
niranjanyardi, line 134: is this for 3P or all platforms  ? do you need IS_STARBOARD here ? unless you want it to apply to apple, android .

services/resource_coordinator/BUILD.gn
```diff
@@ -83,3 +90,15 @@ source_set("tests") {
     "//testing/gtest",
   ]
 }
+
+if (is_cobalt) {
+  import("//testing/test.gni")
```
niranjanyardi, line 88: what error do you get when remove this import ?

services/resource_coordinator/BUILD.gn
```diff
@@ -83,3 +90,15 @@ source_set("tests") {
...
+
+  test("cobalt_memory_instrumentation_unittests") {
```
niranjanyardi, line 90: why are we creating a new test ? there should be an existing target which uses the tests source set ... right ? why can't we reuse that ?

niranjanyardi (commented):
@briantting  do we need to add

services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate_unittest.cc
Awallky: Ah, yes thanks for flagging! I originally had this implemented in an upstream prototype CL and ported it to Cobalt. But until it lands upstream, let's make it a Cobalt license.

services/resource_coordinator/public/cpp/memory_instrumentation/tracing_observer_proto_unittest.cc
Awallky: Ideally this would be applicable to all platforms. Since this is from upstream, I don't want to modify the platform-specific changes here.

services/resource_coordinator/BUILD.gn
Awallky: We do not currently run any tests under `//services` in our CI. 
I want to provide coverage for the changes we have added to it while minimizing additional test time for presubmit CI. Every test added leads to longer pre-submit wait times. This seems like a good compromise between the two.

Awallky: I realized I had not reduced the test set with the initial change. I have since patched that to reduce our set of test suites in this target.

niranjanyardi (approved):

PR #10142, opened 2026-04-17
Use feature flag for cobalt metrics collection

tasunnn:
This change creates two feature params, `kCpuMetricsIntervalParam` 
and `kMemoryMetricsIntervalParam` to provide independent control 
over the collection interval of memory and CPU metrics. Default values
for both flags are 5 minutes and can be manually set when the feature flag
`kCobaltMetricsIntervalFeature` is enabled. 

It allows for easier tuning of the frequency to collect different cobalt metrics, 
could later be used in experiments to determine the optimal value.

Bug: 495528560

### DIFF
--- a/cobalt/browser/metrics/cobalt_metrics_browsertest.cc
+++ b/cobalt/browser/metrics/cobalt_metrics_browsertest.cc
@@ -36,9 +36,13 @@ class CobaltMetricsBrowserTest : public content::ContentBrowserTest {
   void SetUpCommandLine(base::CommandLine* command_line) override {
     content::ContentBrowserTest::SetUpCommandLine(command_line);
     // Set a short interval for memory metrics to verify periodic recording.
-    command_line->AppendSwitchASCII("
-        enable-features",
-        "CobaltMetricsInterval:memory-metrics-interval/1, CobaltMetricsInterval:cpu-metrics-interval/1");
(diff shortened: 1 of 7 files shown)


cobalt/browser/metrics/cobalt_metrics_browsertest.cc
```diff
@@ -38,7 +37,9 @@ class CobaltMetricsBrowserTest : public content::ContentBrowserTest {
     content::ContentBrowserTest::SetUpCommandLine(command_line);
     // Set a short interval for memory metrics to verify periodic recording.
     command_line->AppendSwitchASCII(
-        "enable-features", "CobaltMetricsInterval:cobalt-metrics-interval/1");
+        "enable-features",
+        "CobaltMetricsInterval:memory-metrics-interval/1, "
```
Awallky: I don't know if the `base::FeatureList` parsing supports spaces after commas. Can you remove the space between these?

Awallky (approved):
LGTM

For future changes, I recommend adding more context with more detailed commit messages for each commit in the PR.

cobalt/browser/metrics/cobalt_metrics_browsertest.cc
tasunnn: done, thanks!

PR #10150, opened 2026-04-19
aosp: skip bionic futex path on hermetic builds

rogerzanoni:
aosp: skip bionic futex path on hermetic builds

ABSL_INTERNAL_HAVE_FUTEX for Android was being guarded only by checking 
for __BIONIC__, so ABSL_INTERNAL_HAVE_FUTEX was being defined for
hermetic builds.

ABSL_INTERNAL_HAVE_FUTEX was enabling a block with platform code that calls
the futex() syscall, causing build issues:

  third_party/abseil-cpp/absl/synchronization/internal/futex.h:
      error: use of undeclared identifier '__NR_futex'
      error: use of undeclared identifier 'FUTEX_WAKE'

This fix ensures that ABSL_INTERNAL_HAVE_FUTEX is only defined for
non-hermetic builds for Android.

Bug: 505627880

### DIFF
--- a/third_party/abseil-cpp/absl/synchronization/internal/futex.h
+++ b/third_party/abseil-cpp/absl/synchronization/internal/futex.h
@@ -42,7 +42,7 @@
 
 #ifdef ABSL_INTERNAL_HAVE_FUTEX
 #error ABSL_INTERNAL_HAVE_FUTEX may not be set on the command line
-#elif defined(__BIONIC__)
+#elif defined(__BIONIC__) && !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
 // Bionic supports all the futex operations we need even when some of the futex
 // definitions are missing.
 #define ABSL_INTERNAL_HAVE_FUTEX


haozheng-cobalt (changes requested):
How did you catch this issue in the first place?
I think AOSP should be hermetic build.

third_party/abseil-cpp/absl/synchronization/internal/futex.h
```diff
@@ -22,12 +22,13 @@
 #include <unistd.h>
 #endif
 
-#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD)
+// Bionic needs these headers, exclude them only for musl-based builds
+#if !BUILDFLAG(IS_COBALT_HERMETIC_BUILD) || defined(__BIONIC__)
```
niranjanyardi: what's the error causing this ?

can you please post it in the bug / issue ?
I'm trying to understand if this is a cobalt dependency or starboard dep.

rogerzanoni: Thanks for reviewing it, added the build log to a [new bug](https://partnerissuetracker.corp.google.com/issues/505627880)

andrewsavage1: Looks like this is a starboard dependency, so I'm okay with this change, but maybe we can replace `defined(__BIONIC__)` with i.e. `BUILDFLAG(IS_ANDROID) && BUILDFLAG(IS_PARTNER_TOOLCHAIN)`

hferreiro: Though `&& BUILDFLAG(IS_PARTNER_TOOLCHAIN)` isn't strictly needed since this would be needed for AndroidTV too if it were to be built with an hermetic toolchain in the future, no?

andrewsavage1: If ATV were to build with a hermetic toolchain then I believe BUILDFLAG(IS_PARTNER_TOOLCHAIN) should be true

hferreiro: OK true. What I meant is that the change is required for Android, independently of any other build setup.

andrewsavage1: If Android switched to building hermetically like Evergreen, we actually want to make sure not to include the futex code in it for the Cobalt layer so as not to require it in the Starboard interface

rogerzanoni: @andrewsavage1 it makes sense. I will change it to prevent defining ABSL_INTERNAL_HAVE_FUTEX if it's an hermetic build

rogerzanoni:
@haozheng-cobalt @andrewsavage1 @niranjanyardi Thanks for the review, I changed the original fix to instead of allowing including the futex/syscall platform headers, prevent defining ABSL_INTERNAL_HAVE_FUTEX that was causing the internal implementation of absl Futex to be included in the build.

andrewsavage1:
> @haozheng-cobalt @andrewsavage1 @niranjanyardi Thanks for the review, I changed the original fix to instead of allowing including the futex/syscall platform headers, prevent defining ABSL_INTERNAL_HAVE_FUTEX that was causing the internal implementation of absl Futex to be included in the build.

This seems much better, thanks Roger

andrewsavage1 (approved):

rogerzanoni:
@haozheng-cobalt I think merging is blocked because you requested changes, I force-pushed the branch and can't see the comments, could you please review it again?

PR #10151, opened 2026-04-19
aosp: fix libunwind build for android

rogerzanoni:
Put starboard's implementation of evergreen_dl_iterate_phdr under the _LIBUNWIND_USE_DL_ITERATE_PHDR guard; Elf_Phdr type isn't defined for bionic, and android builds should use dl_unwind_find_exidx instead.

Bug: 495203133

### DIFF
--- a/third_party/llvm-project/libunwind/src/AddressSpace.hpp
+++ b/third_party/llvm-project/libunwind/src/AddressSpace.hpp
@@ -490,7 +490,7 @@ static int findUnwindSectionsByPhdr(struct dl_phdr_info *pinfo,
 
 #endif  // defined(_LIBUNWIND_USE_DL_ITERATE_PHDR)
 
-#if defined(STARBOARD_IMPLEMENTATION)
+#if defined(STARBOARD_IMPLEMENTATION) && defined(_LIBUNWIND_USE_DL_ITERATE_PHDR)
 int evergreen_dl_iterate_phdr(
                  int (*callback) (struct dl_phdr_info *info,
                                   size_t size, void *data),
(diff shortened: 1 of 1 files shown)


haozheng-cobalt (changes requested):
Why change the existing Macro guard? Would it be better to append the condition with BIONIC check?

rogerzanoni:
> Why change the existing Macro guard? Would it be better to append the condition with \_\_BIONIC\_\_ check?

@haozheng-cobalt moved it inside the block guarded by `_LIBUNWIND_USE_DL_ITERATE_PHDR` because it was previously only checking for `STARBOARD_IMPLEMENTATION`, and `evergreen_dl_iterate_phdr` uses the `Elf_Phdr` which is only typedef'd if _LIBUNWIND_USE_DL_ITERATE_PHDR is defined.

And in https://github.com/youtube/cobalt/blob/main/third_party/llvm-project/libunwind/src/config.h there's already a `\_\_BIONIC\_\_ ` that defines `_LIBUNWIND_USE_DL_UNWIND_FIND_EXIDX`:

```cpp
#elif defined(__BIONIC__) && defined(_LIBUNWIND_ARM_EHABI)
  // For ARM EHABI, Bionic didn't implement dl_iterate_phdr until API 21. After
  // API 21, dl_iterate_phdr exists, but dl_unwind_find_exidx is much faster.
  #define _LIBUNWIND_USE_DL_UNWIND_FIND_EXIDX 1
#elif defined(_AIX)
```

so when `_LIBUNWIND_USE_DL_UNWIND_FIND_EXIDX` is defined, AddressSpace.hpp will use `dl_unwind_find_exidx`:

 ```cpp
 #elif defined(_LIBUNWIND_USE_DL_UNWIND_FIND_EXIDX)
  int length = 0;
  info.arm_section =
      (uintptr_t)dl_unwind_find_exidx((_Unwind_Ptr)targetAddr, &length);
  info.arm_section_length = (size_t)length * sizeof(EHABIIndexEntry);
  if (info.arm_section && info.arm_section_length)
    return true;
#elif defined(_LIBUNWIND_USE_DL_ITERATE_PHDR)
```
I thought that would be appropriate because of the comments in config.h, but now I'm thinking about the evergreen flags, maybe it would be better to change config.h to guard to

 ```cpp
#elif defined(__BIONIC__) && defined(_LIBUNWIND_ARM_EHABI) && !BUILDFLAG(USE_EVERGREEN)
```

so it will fallback to the default configuration that defines `_LIBUNWIND_USE_DL_ITERATE_PHDR` (and the types `evergreen_dl_iterate_phdr` needs).

This seems more in line with the current implementation, wdyt?

(Tested it locally, it fixes the libunwind build issue)

andrewsavage1:
@y4vor or @hlwarriner could one of you take a look as y'all know more about this?

hlwarriner:
> Put starboard's implementation of evergreen_dl_iterate_phdr under the _LIBUNWIND_USE_DL_ITERATE_PHDR guard; Elf_Phdr type isn't defined for bionic, and android builds should use dl_unwind_find_exidx instead.
> 
> Bug: 495203133

I imagine that for Evergreen on AOSP, since the Cobalt shared library is loaded by the Portable ELF Loader instead of the system loaader, we would also need a custom implementation of `dl_unwind_find_exidx`? @y4vor WDYT

A way to test that this all works as expected would be to run a crashing unit test target in Evergreen mode on an AOSP device and ensure that a symbolized stack trace is printed.

hferreiro:
When we worked on running Cobalt `main` with the v25 loader, I had something quickly done to support that: https://github.com/hferreiro/cobalt/commit/2a14ef511c616baef9b712f0d55dda408fd4e3e9

rogerzanoni:
Now the fix adds a custom evergreen_dl_unwind_find_exidx implementation and gates the evergreen_dl_iterate_phdr block with _LIBUNWIND_USE_DL_ITERATE_PHDR to fix the build issues with undeclared Elf_Phdr type

third_party/llvm-project/libunwind/src/AddressSpace.hpp
```diff
@@ -510,6 +510,32 @@ int evergreen_dl_iterate_phdr(
 }
 #endif
 
+#if defined(STARBOARD_IMPLEMENTATION) && \
+    defined(_LIBUNWIND_USE_DL_UNWIND_FIND_EXIDX)
```
y4vor, line 514: I assume AOSP would define _LIBUNWIND_USE_DL_UNWIND_FIND_EXIDX.

y4vor (approved):

third_party/llvm-project/libunwind/src/AddressSpace.hpp
rogerzanoni: Than for checking it @y4vor, it does define it for bionic in https://github.com/youtube/cobalt/blob/main/third_party/llvm-project/libunwind/src/config.h#L42

haozheng-cobalt (approved):

PR #10153, opened 2026-04-19
jni: add jni_zero namespace to AttachCurrentThread() calls

rogerzanoni:
copied_base is already using jni_zero, but AttachCurrentThread() calls aren't using the jni_zero namespace or are using the wrong namespace.

Bug: 495203133


rogerzanoni:
Abandoning in favor of https://github.com/youtube/cobalt/pull/10177

PR #10155, opened 2026-04-20
cobalt: Fix flaky CobaltMetricsServiceClientTest

kjyoun:
Reset the default client created in SetUp() at the start of tests that create their own client. This prevents the background memory metrics logger from the initial client (which uses a random delay) from firing unexpectedly and interfering with the test's assertions.

Flaky test was introduced in #9944 

Bug: 504450672

### DIFF
--- a/cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
+++ b/cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
@@ -835,6 +835,7 @@ TEST_F(CobaltMetricsServiceClientTest, MetricsIntervalDefaultProductionTest) {
   // Reset the default client created in SetUp to prevent its background
   // metrics loggers from interfering with this test's results.
   client_.reset();
+  base::RunLoop().RunUntilIdle();
 
   // Verify that the feature is disabled by default.
   EXPECT_FALSE(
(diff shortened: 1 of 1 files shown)


cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
```diff
@@ -786,6 +786,11 @@ TEST_F(CobaltMetricsServiceClientTest,
 }
 
 TEST_F(CobaltMetricsServiceClientTest, CobaltMetricsIntervalFeatureTest) {
+  // Reset the default client created in SetUp to prevent its background
```
Awallky: Can you actually add this to the `SetUp()` method in the `CobaltMetricsServiceClientTest`?

Awallky (approved):
LGTM % comment

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
kjyoun: we cannot simply adding this logic to Setup, since it would break other tests expecting client_ to be created.

Instead, I created another test class, not creating client_.

cobalt/browser/metrics/cobalt_metrics_service_client_test.cc
```diff
@@ -413,13 +399,41 @@ class CobaltMetricsServiceClientTest : public ::testing::Test {
...
+    // This simulates the two-phase initialization of the static Create().
+    client_ = std::make_unique<TestCobaltMetricsServiceClient>(
+        metrics_state_manager_.get(), std::move(synthetic_trial_registry),
+        &prefs_);
+    client_->CallInitialize();  // This will use the overridden factory methods.
+
+    ASSERT_THAT(client_->mock_metrics_service(), NotNull());
+    ASSERT_THAT(client_->mock_log_uploader(), NotNull());
+  }
+};
```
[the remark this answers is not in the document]
kjyoun: All those changes are intended.

Test which needs client_ should use ClientTest

Awallky (approved):
LGTM
Thanks for the fix!

PR #10156, opened 2026-04-20
cobalt: Force system host resolver to use IPv4

zhongqiliang:
This change introduces a new feature, kUseIPv4, to explicitly instruct the
system host resolver to prioritize or exclusively use IPv4 for DNS
resolution. This is particularly useful in environments where IPv6
connectivity is unstable, misconfigured, or not desired, helping to
mitigate potential network issues and improve connection reliability.

The feature can be enabled from the Android Java layer by passing a
specific Java switch, which then translates into a C++ command line flag
(--enable-features=UseIPv4) that activates the feature within Cobalt's
networking stack.

[go/cobalt-dns-resolution-hang-investigation](https://goto.google.com/cobalt-dns-resolution-hang-investigation)

Bug: 492247862


cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -14,10 +14,27 @@
...
+
 /**
  * Defines the constant names for feature switches used in Kimono.
  */
 public class JavaSwitches {
   public static final String ENABLE_QUIC = "EnableQUIC";
   public static final String DISABLE_STARTUP_GUARD = "DisableStartupGuard";
+
+  /** flag to force use IPv4 for system host resolution. */
+  public static final String USE_IPV4 = "UseIPv4";
```
johnxwork: Can we start using all lower case in the switch names? That will reduce the risk of typos.

johnxwork (approved):

net/dns/host_resolver_system_task.cc
johnxwork, line 1: The changes here only guarantees that the host resolver uses IPv4 family. Can we also modify other net requests to all use IPv4 only? Since the switch has a generic name of "UseIPv4", not "UseIPv4ForDNS"

johnxwork (changes requested):

net/dns/host_resolver_system_task.cc
zhongqiliang: renamed to UseIPv4ForDNS

johnxwork (approved):

PR #10157, opened 2026-04-20
[POSIX] Disable use of memfd_create on Cobalt

sacuff:
`memfd_create` is not supported in Cobalt and the sycall wrapper is consequently defined as `(errno = ENOSYS, -1)`. This PR does not attempt to change that, but it does add the symbol to the allowlist in the API leak detector.

There are two existing uses of memfd_create in C27:

  * //third_party/crashpad/crashpad/util/file/file_io_posix.cc
  * //base/allocator/partition_allocator/src/partition_alloc/partition_address_space.cc

Crashpad has fallback handling for ENOSYS. PartitionAlloc on the other hand relies on `memfd_create` to enable the ShadowMetadata feature and appears to crash if it fails. This disables the build flag for ShadowMetadata on Cobalt.

Bug: 500418451


base/allocator/partition_allocator/partition_alloc.gni
```diff
@@ -364,7 +364,7 @@ declare_args() {
 declare_args() {
   # Shadow metadata is still under development and only supports Linux
   # for now.
-  enable_shadow_metadata = is_linux && has_64_bit_pointers
+  enable_shadow_metadata = is_linux && has_64_bit_pointers && !is_cobalt
```
andrewsavage1: Set this in one of our args.gn files instead, that's generally how we override gn arguments

y4vor: cobalt/build/configs/linux_common.gn

y4vor (approved):

PR #10160, opened 2026-04-20
media: Allow audio writing on pause to reduce latency

jasonzhangxx:
Introduce an experimental feature to allow the audio sink to continue
writing audio data even when the media pipeline is paused. This is
particularly beneficial during seeking, where the player pauses until
both audio and video are prerolled. Pre-writing audio ensures the
audio pipeline is ready for immediate playback resumption,
reducing perceived startup latency.

Bug: 500811542

### DIFF
--- a/starboard/android/shared/audio_sink_min_required_frames_tester.cc
+++ b/starboard/android/shared/audio_sink_min_required_frames_tester.cc
@@ -133,7 +133,7 @@ void MinRequiredFramesTester::TesterThreadFunc() {
             GetSampleSize(task.sample_type),
         &MinRequiredFramesTester::UpdateSourceStatusFunc,
         &MinRequiredFramesTester::ConsumeFramesFunc,
-        &MinRequiredFramesTester::ErrorFunc, 0, -1, false, false, this);
+        &MinRequiredFramesTester::ErrorFunc, 0, -1, false, false, false, this);
     {
       std::unique_lock lock(mutex_);
       bool notified = test_complete_cv_.wait_for(
--- a/starboard/android/shared/audio_track_audio_sink_type.cc
+++ b/starboard/android/shared/audio_track_audio_sink_type.cc
@@ -134,6 +134,7 @@ AudioTrackAudioSink::AudioTrackAudioSink(
     int64_t start_time,
     int tunnel_mode_audio_session_id,
     bool is_web_audio,
+    bool allow_audio_writing_on_pause,
     bool pause_using_audio_track_state,
     void* context)
     : type_(type),
@@ -146,12 +147,13 @@ AudioTrackAudioSink::AudioTrackAudioSink(
       consume_frames_func_(consume_frames_func),
       error_func_(error_func),
       start_time_(start_time),
-      pause_using_audio_track_state_(pause_using_audio_track_state),
       max_frames_per_request_(
           tunnel_mode_audio_session_id == -1
               ? kMaxFramesPerRequest
               : GetMaxFramesPerRequestForTunnelMode(sampling_frequency_hz_)),
       context_(context),
+      allow_audio_writing_on_pause_(allow_audio_writing_on_pause),
+      pause_using_audio_track_state_(pause_using_audio_track_state),
       bridge_(kSbMediaAudioCodingTypePcm,
               sample_type,
               channels,
@@ -325,7 +327,10 @@ void AudioTrackAudioSink::AudioThreadFunc() {
       }
     }
 
-    if (!is_playing || frames_in_buffer == 0) {
+    // When |allow_audio_writing_on_pause| feature is enabled, continue writing
+    // audio regardless of whether audio is playing.
+    if ((!is_playing && !allow_audio_writing_on_pause_) ||
+        frames_in_buffer == 0) {
       usleep(10'000);
       continue;
     }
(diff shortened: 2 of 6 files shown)


starboard/android/shared/audio_track_audio_sink_type.h
```diff
@@ -156,10 +158,12 @@ class AudioTrackAudioSink : public SbAudioSinkImpl {
   const ConsumeFramesFunc consume_frames_func_;
   const SbAudioSinkPrivate::ErrorFunc error_func_;
   const int64_t start_time_;  // microseconds
-  const bool pause_using_audio_track_state_;
   const int max_frames_per_request_;
   void* const context_;
 
+  const bool allow_audio_writing_on_pause_;
+  const bool pause_using_audio_track_state_;
```
borongc, line 165: Are there any reasons we want to move this variable?

jasonzhangxx: Consolidate the experimental attributes for improved management.

starboard/android/shared/player_components_factory.cc
```diff
@@ -418,7 +420,15 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
...
           << "AudioTrackState while pausing playback.";
+
+      // TODO: b/500811542 - Connect to H5VCC.
+      const bool allow_audio_writing_on_pause =
+          experimental_features.allow_audio_writing_on_pause;
+      SB_LOG_IF(INFO, allow_audio_writing_on_pause)
+          << "allow_audio_writing_on_pause is set to true.";
+
       const bool force_platform_opus_decoder = force_platform_opus_decoder_;
+
```
borongc: nit: extra line

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -325,7 +327,8 @@ void AudioTrackAudioSink::AudioThreadFunc() {
       }
     }
 
-    if (!is_playing || frames_in_buffer == 0) {
+    if ((!is_playing && !allow_audio_writing_on_pause_) ||
```
borongc, line 332: It's good to add comment here that what `allow_audio_writing_on_pause_` is for, and why we want to do it.

borongc (approved):

starboard/android/shared/audio_track_audio_sink_type.cc
jasonzhangxx: Done

starboard/android/shared/player_components_factory.cc
jasonzhangxx: Removed

starboard/android/shared/audio_track_audio_sink_type.h
```diff
@@ -76,6 +76,7 @@ class AudioTrackAudioSinkType : public SbAudioSinkPrivate::Type {
       int64_t start_time,
       int tunnel_mode_audio_session_id,
       bool is_web_audio,
+      bool allow_audio_writing_on_pause,
```
kjyoun, line 79: How about defining CreateOption and use it instead?

This Create mehtod  already has 14 input argument.
Usinng struct CreateOptions, instead  of adding a new one (to 15), we can reduce it to 14 and be future-proof. 

```
struct CreateOption {
  bool allow_audio_writing_on_pause = false;
  bool pause_using_audio_track_state = false;'
}

SbAudioSink Create(
  ...
   const CreateOption& options
   void *context);
```

starboard/android/shared/audio_sink_min_required_frames_tester.cc
```diff
@@ -133,7 +133,7 @@ void MinRequiredFramesTester::TesterThreadFunc() {
             GetSampleSize(task.sample_type),
         &MinRequiredFramesTester::UpdateSourceStatusFunc,
         &MinRequiredFramesTester::ConsumeFramesFunc,
-        &MinRequiredFramesTester::ErrorFunc, 0, -1, false, false, this);
+        &MinRequiredFramesTester::ErrorFunc, 0, -1, false, false, false, this);
```
kjyoun, line 136: use comment param?
```
/*allow_audio_writing_on_pause=*/false,
```

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -504,12 +509,13 @@ SbAudioSink AudioTrackAudioSinkType::Create(
...
   const int kTunnelModeAudioSessionId = -1;
   const bool kIsWebAudio = true;
+  const bool kAllowAudioWritingOnPause = false;
   const bool kPauseUsingAudioTrackState = false;
   return Create(channels, sampling_frequency_hz, audio_sample_type,
                 audio_frame_storage_type, frame_buffers, frames_per_channel,
                 update_source_status_func, consume_frames_func, error_func,
                 kStartTime, kTunnelModeAudioSessionId, kIsWebAudio,
-                kPauseUsingAudioTrackState, context);
+                kAllowAudioWritingOnPause, kPauseUsingAudioTrackState, context);
```
kjyoun, line 518: do we need to define constant here?

why not using param comment?
```
/*allow_audio_writing_on_pause*/false,
````

I think that it is better to replace this one-time contants with param comments.

```
 const int64_t kStartTime = 0;
  // Disable tunnel mode.
  const int kTunnelModeAudioSessionId = -1;
  const bool kIsWebAudio = true;
  const bool kAllowAudioWritingOnPause = false;
  const bool kPauseUsingAudioTrackState = false;
````

kjyoun:
Do you happen to do any data from local testing?

It will be great, if we have some data from local testing

starboard/android/shared/audio_track_audio_sink_type.h
jasonzhangxx: No, I don't think `CreateOption` help here. Passing ExperimentalFeatures could be an option, but let's not have that change in this pr.

jasonzhangxx:
> Do you happen to do any data from local testing?
> 
> It will be great, if we have some data from local testing

I have very basic numbers in b/500811542, but they doesn't make much sense as the samples are too small and only on a few devices.

starboard/android/shared/audio_sink_min_required_frames_tester.cc
jasonzhangxx: That can be done in follow-up change if we decide to do that. Let's not break the consistency in this pr.

starboard/android/shared/audio_track_audio_sink_type.cc
jasonzhangxx: Same as the other comment. That can be done in a follow-up change if we decide to do that. Let's not break the consistency in this pr.

PR #10161, opened 2026-04-20
android: Add startup diagnosis for network errors

zhongqiliang:
Introduce a mechanism to capture and report network navigation errors to the
StartupGuard on Android. Previously, if a navigation error occurred
during startup, StartupGuard would only report a generic crash.

This change enhances the StartupGuard crash message to include specific
network error codes (e.g., -2 for ERR_FAILED) when a DidFinishNavigation
failure triggers the StartupGuard timeout. This provides critical
diagnostic information to pinpoint the root cause of startup issues.

Bug: 492247862

### DIFF
--- a/cobalt/browser/cobalt_web_contents_observer.cc
+++ b/cobalt/browser/cobalt_web_contents_observer.cc
@@ -89,7 +89,8 @@ void CobaltWebContentsObserver::DidFinishNavigation(
                              -net_error_code);
     LOG(INFO) << "DidFinishNavigation: Raising platform error with code: "
               << net::ErrorToString(net_error_code);
-    SetStartupDiagnosisInfo("navigation_error", net::ErrorToString(net_error_code).c_str());
+    SetStartupDiagnosisInfo("navigation_error",
+                            net::ErrorToString(net_error_code).c_str());
     RaisePlatformError();
   } else if (net_error_code == net::OK) {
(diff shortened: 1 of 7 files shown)


ckim1346 (approved):

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -89,6 +89,10 @@ void CobaltWebContentsObserver::DidFinishNavigation(
                              -net_error_code);
     LOG(INFO) << "DidFinishNavigation: Raising platform error with code: "
               << net::ErrorToString(net_error_code);
+#if BUILDFLAG(IS_ANDROID)
+    starboard::StarboardBridge::GetInstance()->SetStartupDiagnosisInfo(
+        "navigation_error", net::ErrorToString(net_error_code).c_str());
```
johnxwork: Is it possible for `net::ErrorToString(net_error_code).c_str()` to return null ptr? We don't have null check in this newly added call stack.

johnxwork (approved):

cobalt/browser/cobalt_web_contents_observer.cc
zhongqiliang: No. https://source.corp.google.com/h/github/youtube/cobalt/+/main:net/base/net_errors.cc;l=24?q=net::ErrorToString&sq=git:github%2Fyoutube%2Fcobalt@main

PR #10164, opened 2026-04-20
Remove On Device Ad Auction from C27

haozheng-cobalt:
Bug: 504691422


linxinan-chops:
https://privacysandbox.google.com/private-advertising/protected-audience#interest-group-detail

protected-audience was phased out. I would recommend to land it on chromium.

haozheng-cobalt:
do not change files under Fenced Frames, exclude the entire folder.

PR #10169, opened 2026-04-20
Remove USE_M114_FUNCTIONS

yuying-y:
Cobalt customizations are integrated to main. These m114 functions are removed.

Issue: 455004672

### DIFF
--- a/components/update_client/BUILD.gn
+++ b/components/update_client/BUILD.gn
@@ -201,11 +201,6 @@ static_library("update_client") {
   }
 
   if (is_starboard) {
-    # TODO(b/455004672): Investigate the legacy functions from m114.
-    # Some functions are ported from m114 because Cobalt customizations
-    # depend on them.
-    defines = [ "USE_M114_FUNCTIONS" ]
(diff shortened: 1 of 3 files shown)


components/update_client/BUILD.gn
```diff
@@ -201,11 +201,6 @@ static_library("update_client") {
   }
 
   if (is_starboard) {
-    # TODO(b/455004672): Investigate the legacy functions from m114.
-    # Some functions are ported from m114 because Cobalt customizations
```
hlwarriner, line 205: I guess these m114 functions are no longer needed, in the main branch? Please update the PR description with a very short explanation of why it's ok to remove these.

For example, "Cobalt customizations have been updated to no longer depend on these functions - see PR x."

hlwarriner (approved):

components/update_client/BUILD.gn
yuying-y: done. Thanks.

y4vor (approved):

PR #10183, opened 2026-04-21
[POSIX] Add inotify API

sacuff:
This adds wrappers for `inotify_init`, `inotify_init1`, `inotify_add_watch`, and `inotify_rm_watch` Linux syscalls. Chromium uses these functions in several places, e.g. to watch for changes to the DNS config or timezone.

Bug: 500417827


starboard/elf_loader/exported_symbols.cc
```diff
@@ -335,6 +336,10 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_WRAPPER(getuid);
   REGISTER_WRAPPER(getpriority);
   REGISTER_WRAPPER(getrlimit);
+  REGISTER_WRAPPER(inotify_init);
```
jellefoks: Instead of adding this one to Starboard we should just implement ` inotify_init() { return inotify_init(0); }` see the man page 'If flags is 0, then inotify_init1() is the same as inotify_init()'.

starboard/nplb/posix_compliance/posix_inotify_test.cc
jellefoks, line 1: There doesn't seem to be any tests that validate that a mutation gets signaled. We should be able to make a temp folder/file and test the mutation notifications on it.

jellefoks (changes requested):

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Added a test for modifying a file

third_party/musl/src/starboard/inotify.c
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <sys/inotify.h>
+
+int inotify_init(void) {
```
andrewsavage1, line 17: If we're implementing this one in musl, shouldn't only inotify_init1 be added to the sb interface?

sacuff: This should go in third_party/musl/src/linux/inotify.c then?

andrewsavage1: I'm not sure I understand—I'm saying that because we're introducing inotify_init in musl, it does not cross the starboard boundary, only inotify_init1 does, so I would expect inotify_init1 to be an exported symbol but not inotify_init

sacuff: Per chat discussion `inotify_init` removed from exported_symbols.cc, changelog, and the leak detector allowlist now

starboard/nplb/posix_compliance/posix_inotify_test.cc
```diff
@@ -0,0 +1,154 @@
...
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace {
+
+class PosixInotifyTest : public ::testing::Test {
+ protected:
+  // The spec requires that all watches are freed when the inotify file
+  // descriptor is closed, so we only need to track the fds to close.
+  std::vector<int> fds_to_close_;
+  PosixInotifyTest() { errno = 0; }
```
andrewsavage1: I think you'd want to do these steps in SetUp/TearDown instead of constructor/destructor?

starboard/nplb/posix_compliance/posix_inotify_test.cc
```diff
@@ -0,0 +1,154 @@
...
+
+  EXPECT_EQ(inotify_rm_watch(fd, -1), -1);
+  EXPECT_EQ(errno, EINVAL);
+}
+
+TEST_F(PosixInotifyTest, WatchedFileTriggersNotification) {
+  nplb::ScopedRandomFile file;
+  ASSERT_TRUE(nplb::FileExists(file.filename().c_str()));
+
+  int inotify_fd = inotify_init1(0);
```
andrewsavage1, line 250: Is this ever closed?

I'm not sold on the fds_to_close_ pattern in general, it doesn't necessarily seem easier to write a test with that as opposed to just calling close on the fd

starboard/shared/modular/starboard_layer_posix_inotify_abi_wrappers.cc
```diff
@@ -0,0 +1,156 @@
...
+      MUSL_IN_MOVED_TO | MUSL_IN_CREATE | MUSL_IN_DELETE | MUSL_IN_DELETE_SELF |
+      MUSL_IN_MOVE_SELF | MUSL_IN_UNMOUNT | MUSL_IN_Q_OVERFLOW |
+      MUSL_IN_IGNORED | MUSL_IN_ONLYDIR | MUSL_IN_DONT_FOLLOW |
+      MUSL_IN_EXCL_UNLINK | MUSL_IN_MASK_ADD | MUSL_IN_ISDIR | MUSL_IN_ONESHOT;
+
+#ifdef IN_MASK_CREATE
+  all_masks |= MUSL_IN_MASK_CREATE;
+#endif
+
+  return (musl_mask & ~all_masks) == 0;
```
andrewsavage1: Consider adding the mask checks to a separate function, as it doesn't seem to rely on the result of this function and isn't obvious that it would happen in this function

starboard/nplb/posix_compliance/posix_inotify_test.cc
```diff
@@ -0,0 +1,154 @@
...
+
+TEST_F(PosixInotifyTest, InotifyInit1ValidFlags) {
+  int fd = inotify_init1(IN_NONBLOCK);
+  fds_to_close_.push_back(fd);
+  EXPECT_EQ(errno, 0);
+  EXPECT_GE(fd, 0);
+}
+
+TEST_F(PosixInotifyTest, InotifyInit1InvalidFlags) {
+  int fd = inotify_init1(0xFFFF);
```
andrewsavage1: It's not obvious to me why this would be an invalid flag

starboard/nplb/posix_compliance/posix_inotify_test.cc
```diff
@@ -0,0 +1,154 @@
...
+  EXPECT_EQ(fd, -1);
+  EXPECT_EQ(errno, EINVAL);
+}
+
+TEST_F(PosixInotifyTest, InotifyAddWatchValidMask) {
+  int fd = inotify_init();
+  fds_to_close_.push_back(fd);
+  ASSERT_GE(fd, 0);
+
+  int wd = inotify_add_watch(fd, "/dev/null", IN_MODIFY);
```
andrewsavage1: @jellefoks @y4vor are we safe to assume /dev/null exists on target devices? We may want to instead create a scoped temporary file to monitor

starboard/nplb/posix_compliance/posix_inotify_test.cc
```diff
@@ -0,0 +1,154 @@
...
+#include <fcntl.h>
+#include <limits.h>
+#include <string.h>
+#include <sys/inotify.h>
+#include <unistd.h>
+
+#include "starboard/nplb/file_helpers.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace {
```
andrewsavage1, line 25: We should additionally add tests for EEXIST, ENOENT, ENOTDIR, and EINVAL (for `IN_MASK_ADD | IN_MASK_CREATE`) for inotify_add_watch, EINVAL for a fd not opened with inotify for inotify_rm_watch

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Constructor/destructor is slightly preferred but they're effectively equivalent in this case - https://google.github.io/googletest/faq.html#CtorVsSetUp

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Yeah this one is missing a close

The idea was to have the destructor close fds so they get cleaned up even when an `ASSERT` fails, otherwise they'd still be open until the test process exits. I figure it's fine to just have `close` on the happy paths if that's clearer from a test maintenance perspective. It's probably not an issue to have a few stray fds handing around for a couple seconds.

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Test fixture has been removed now

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Test fixture has been removed now

starboard/shared/modular/starboard_layer_posix_inotify_abi_wrappers.cc
sacuff: Separated it

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Switched to `~(allvalidflags)`

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Replaced all /dev/null with scoped files

starboard/nplb/posix_compliance/posix_inotify_test.cc
sacuff: Added (and also EINVAL for non-inotify fd for inotify_add_watch)

y4vor (approved):

starboard/tools/api_leak_detector/api_leak_detector.py
```diff
@@ -134,6 +134,11 @@
     'getuid',
     'gmtime_r',
     'if_indextoname',
+    'if_nametoindex',
+    'inotify_add_watch',
+    'inotify_init',
```
jellefoks: Given that inotify_init is implemented in musl by calling inotify_init1, doesn't that mean it should *not* be listed here, because we're expecting it to never be an unknown symbol.

starboard/CHANGELOG.md
```diff
@@ -44,6 +44,10 @@ rendering issues caused by Chromium's shader expectations.
 * `getuid`
 * `if_indextoname`
 * `if_nametoindex`
+* `inotify_init`
```
jellefoks: Same here, with inotify_init implemented in musl, it's *not* actually added to Starboard. It's available to be used from Cobalt code, but it's not provided by the platform, because it's routed through inotify_init1.

jellefoks (approved):
Nice!

PR #10186, opened 2026-04-21
Bring sync chromium branches logic into autoroll.

briantting:
Generalize the autoroll workflow to support various source and target branches. Specifically this allows the cherry pick logic that brought m138 refresh commits from chromium/m138 to main to be implemented.

Bug: 505073626

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -30,10 +30,10 @@ def get_out(cmd):
   return res.stdout
 
 
-def get_commits(origin, target, start):
-  cmd = ['git', 'rev-list', '--oneline', '--reverse', origin, f'^{target}']
+def get_commits(source, target, start):
+  cmd = ['git', 'rev-list', '--oneline', '--reverse', source, f'^{target}']
   if start:
-    cmd.append(f'{start}^..{origin}')
+    cmd.append(f'{start}^..{source}')
   return get_out(cmd).splitlines()
 
 
@@ -70,7 +70,18 @@ def cherry_pick(sha, num, title):
   ps = get_out(['git', 'show', '-s', '--format=%P', sha]).strip().split()
   if len(ps) > 1:
     cmd.append('--mainline=1')
-  subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
+  result = subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
+
+  if result.returncode != 0:
+    status_output = get_out(['git', 'status'])
+    files_to_remove = []
+    for line in status_output.splitlines():
+      if 'deleted by us:' in line:
+        filename = line.split('deleted by us:')[1].strip()
+        files_to_remove.append(filename)
+    if files_to_remove:
+      cmd = ['git', 'rm'] + files_to_remove
+      subprocess.run(cmd, check=True, stdout=sys.stderr)
 
   cmd = [
       'git', 'commit', '--no-verify', f'--author={author}', f'--date={date}',
@@ -81,28 +92,33 @@ def cherry_pick(sha, num, title):
 
 def main():
   p = argparse.ArgumentParser()
+  p.add_argument('--source-branch', required=True)
   p.add_argument('--target-branch', required=True)
   p.add_argument('--start-commit')
-  p.add_argument('--origin-branch', default='main')
   p.add_argument('--max-commits', type=int, default=1000)
+  p.add_argument(
+      '--all-commits',
+      action='store_true',
+      help='Process all commits, not just PRs')
   args = p.parse_args()
 
-  links = []
-  target_prs = get_pr_set(args.target_branch, args.origin_branch)
-  autoroll_prs = get_pr_set('HEAD', args.origin_branch)
+  # All cherry picked PRs on the target branch since the branch point.
+  target_prs = get_pr_set(args.target_branch, args.source_branch)
+  # All cherry picked PRs on the autoroll branch since the branch point.
+  autoroll_prs = get_pr_set('HEAD', args.source_branch)
+  commits_added = []
 
-  # Get the number of unmerged commits on the autoroll branch.
-  commits_added = len(autoroll_prs - target_prs)
-
-  for line in get_commits(args.origin_branch, args.target_branch,
+  for line in get_commits(args.source_branch, args.target_branch,
                           args.start_commit):
-    if commits_added >= args.max_commits:
+    if len(commits_added) >= args.max_commits:
       print(f"Reached commit limit ({args.max_commits}).", file=sys.stderr)
       break
 
     match = re.match(r'^(\w+) (.*) \(#(\d+)\)$', line)
     if match:
       sha, title, pr_num = match.groups()
+
+      # Skip if the PR is in the skip list.
       if any(
           skip_sha.startswith(sha)
           for skip_sha in _SKIP_LIST.get(args.target_branch, [])):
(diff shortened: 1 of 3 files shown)


.github/scripts/autoroll_lts.py
```diff
@@ -69,7 +69,18 @@ def cherry_pick(sha, num, title):
   ps = get_out(['git', 'show', '-s', '--format=%P', sha]).strip().split()
   if len(ps) > 1:
     cmd.append('--mainline=1')
-  subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
+  result = subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
+
+  if result.returncode != 0:
```
oxve: How will this handle "regular" file conflicts?

.github/scripts/autoroll_lts.py
```diff
@@ -30,26 +30,31 @@ def get_out(cmd):
...
   return get_out(cmd).splitlines()
 
 
-def get_pr_set(branch, exclude_branch):
-  prs = set()
+def get_change_id_set(branch, exclude_branch, identifier_type):
+  if identifier_type == 'pr':
+    pattern = r'PR #(\d+)'
+  else:
+    pattern = r'commit ([0-9a-f]+)'
```
oxve: This pattern is probably a bit too permissive. Can we change it to match the entire string `cherry picked from commit abcd1234`

Also setting a minimum length on the commit hash? 4 is the minimum git allows (if unambiguous).

.github/workflows/autoroll_lts.yaml
```diff
@@ -5,7 +5,10 @@ on:
     - cron: '0 * * * *' # Hourly.
   workflow_dispatch:
     inputs:
-      start_commit:
+      main_start_commit:
+        description: 'The commit to start rolling from, should point to the last release chore'
+        type: string
+      m138_start_commit:
         description: 'The commit to start rolling from, should point to the last release chore'
```
oxve: Should this be moved to the matrix config below?

For the main rolling I don't actually envision using this. I replaced it with the skip list which should handle all cases.

.github/workflows/autoroll_lts.yaml
```diff
@@ -41,15 +52,18 @@ jobs:
...
         run: |
           set -x
 
-          # Copy the script from the main branch before switching branches.
+          # Copy the script from the main branch.
+          if [ "${{ matrix.branch.source }}" != "main" ]; then
+            git fetch --depth=1 --no-recurse-submodules --filter=blob:none origin main:script
+            git checkout script
+          fi
           cp .github/scripts/autoroll_lts.py "${RUNNER_TEMP}"/
```
oxve: I think this is somewhat overly complex for what it's doing. It can be replaced with a single command.

From:
```
          # Copy the script from the main branch.
          if [ "${{ matrix.branch.source }}" != "main" ]; then
            git fetch --depth=1 --no-recurse-submodules --filter=blob:none origin main:script
            git checkout script
          fi
          cp .github/scripts/autoroll_lts.py "${RUNNER_TEMP}"/
```

To:
```
          # Copy the script from the main branch.
          git show main:.github/scripts/autoroll_lts.py > "${RUNNER_TEMP}/autoroll_lts.py"
```

.github/workflows/autoroll_lts.yaml
```diff
@@ -59,23 +73,30 @@ jobs:
...
+              --target-branch "${TARGET_BRANCH}" \
+              --start-commit "${{ inputs.main_start_commit }}" \
+              --max-commits "${MAX_COMMITS}")
+          else
+            PR_LINKS=$(python3 "${RUNNER_TEMP}/autoroll_lts.py" \
+              --source-branch "${SOURCE_BRANCH}" \
+              --target-branch "${TARGET_BRANCH}" \
+              --start-commit "${{ inputs.m138_start_commit }}" \
+              --max-commits "${MAX_COMMITS}" \
+              --identifier-type commits)
```
oxve: Can you put this in the matrix config too? It seems like something we'd like to configure per matrix job.

.github/workflows/autoroll_lts.yaml
```diff
@@ -59,23 +73,30 @@ jobs:
             git checkout --no-recurse-submodules -B "autoroll-to-${TARGET_BRANCH}" "${TARGET_BRANCH}" --no-track
```
oxve: The branch name only contains the target branch. If we end up autorolling multiple branches to the same target we would be in for a surprise or two.

Can you update `autoroll-${SOURCE_BRANCH}-to-${TARGET_BRANCH}` to future proof?

.github/workflows/autoroll_lts.yaml
```diff
@@ -59,23 +73,30 @@ jobs:
...
         env:
-          TARGET_BRANCH: 27.lts
-          START_COMMIT: ${{ inputs.start_commit }}
+          SOURCE_BRANCH: ${{ matrix.branch.source }}
+          TARGET_BRANCH: ${{ matrix.branch.target }}
           MAX_COMMITS: ${{ inputs.max_commits || 10 }}
         run: |
           set -x
 
           # Run the Python script to perform cherry-picks and get all PR links.
```
oxve, line 77: If you move the start commit and the identifier-type to the matrix config the two if cases would be the same.

But, if you don't do those refactors I think this should follow the FLAGS pattern used here: https://github.com/youtube/cobalt/blob/main/.github/actions/upload_test_artifacts/action.yaml#L59-L67

.github/scripts/autoroll_lts.py
```diff
@@ -68,7 +68,18 @@ def cherry_pick(sha, num, title):
   ps = get_out(['git', 'show', '-s', '--format=%P', sha]).strip().split()
   if len(ps) > 1:
     cmd.append('--mainline=1')
-  subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
+  result = subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
+
+  if result.returncode != 0:
```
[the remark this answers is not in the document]
oxve: This gemini review comment above should probably be fixed.

.github/scripts/autoroll_lts.py

.github/workflows/autoroll_lts.yaml
briantting: done

.github/workflows/autoroll_lts.yaml
briantting: done.

.github/workflows/autoroll_lts.yaml
briantting: done.

.github/workflows/autoroll_lts.yaml
briantting: done.

.github/scripts/autoroll_lts.py
```diff
@@ -147,10 +148,16 @@ def cherry_pick(sha, num, title):
...
-         f'{body_section}'
-         f'(cherry picked from commit {sha})')
+  if num is not None:
+    msg = (f'Cherry pick PR #{num}: {title}\n\n'
+           f'Refer to original PR: #{num}\n\n'
+           f'{body_section}'
+           f'(cherry picked from commit {sha})')
+  else:
+    msg = (f'Cherry pick commit {sha}: {title}\n\n'
+           f'Refer to original commit: #{sha}\n\n'
```
oxve: I think the hash is wrong here. According to the docs a commit sha should be written as-is to link to the original: https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/autolinked-references-and-urls#commit-shas

.github/scripts/autoroll_lts.py
```diff
@@ -181,50 +188,82 @@ def cherry_pick(sha, num, title):
...
+        # Skip if commit in autoroll branch.
+        if any(
+            commit_hash.startswith(sha) for commit_hash in autoroll_change_ids):
+          commits_added.append(f'- {sha}')
+          continue
+
+        # Cherry pick commit.
+        if cherry_pick(sha, None, title):
+          autoroll_change_ids.add(sha)
+          commits_added.append(f'- {sha}')
```
oxve: This else branch looks a lot like the if branch. I think if you make the space and the PR number optional in the regex you could combine the two into one without making the code less readable.

oxve (approved):

.github/scripts/autoroll_lts.py
briantting: done

.github/scripts/autoroll_lts.py
briantting: done

PR #4437, opened 2024-11-14
[media] Support progressive playback

borongc:
This PR brings the progressive demuxer from Cobalt C25 to support progressive videos.

ProgressiveDemuxer is only enabled with `media_use_ffmpeg=false` and `use_starboard_media=true`, as Cobalt needs this for evergreen.

We should re-enable unit-tests for ProgressDemuxer (b/380335617) after this PR.

b/380285071

PR #10188, opened 2026-04-22
media: Remove Progressive playback implementation

osagie98:
Removes the Starboard `ProgressiveDemuxer` implementation, as well `kCobaltProgressivePlayback` media switch. Cobalt now rejects all queries for progressive content type support. When forced to create a `Demuxer` for a progressive content type, the `DemuxerManager` returns an error `DEMUXER_ERROR_NO_SUPPORTED_STREAMS`.

Additionally, the `HTMLMediaElementTest` `CanHandleCobaltProgressiveSupportQueries` has been renamed to `RejectsCobaltProgressiveSupportQueries`, and tests that `HTMLMediaElement` rejects all queries for progressive content type support.

This effectively reverts https://github.com/youtube/cobalt/pull/4437

Bug: 380854966


media/filters/demuxer_manager.cc
```diff
@@ -356,12 +347,8 @@ PipelineStatus DemuxerManager::CreateDemuxer(
...
     SetDemuxer(CreateFFmpegDemuxer());
 #elif BUILDFLAG(USE_STARBOARD_MEDIA)
-    if (base::FeatureList::IsEnabled(media::kCobaltProgressivePlayback)) {
-      SetDemuxer(CreateProgressiveDemuxer());
-    } else {
-      LOG(INFO) << "Cobalt progressive playback is disabled via base features.";
-      return DEMUXER_ERROR_PROGRESSIVE_DISABLED;
-    }
+    LOG(INFO) << "Progressive streams are unsupported.";
+      return DEMUXER_ERROR_NO_SUPPORTED_STREAMS;
```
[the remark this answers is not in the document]
osagie98: This is changed to `DEMUXER_ERROR_NO_SUPPORTED_STREAMS` as `DEMUXER_ERROR_PROGRESSIVE_DISABLED` implies that progressive support can be enabled, which it can't after this PR

media/filters/demuxer_manager.cc
```diff
@@ -356,12 +347,8 @@ PipelineStatus DemuxerManager::CreateDemuxer(
...
     SetDemuxer(CreateFFmpegDemuxer());
 #elif BUILDFLAG(USE_STARBOARD_MEDIA)
-    if (base::FeatureList::IsEnabled(media::kCobaltProgressivePlayback)) {
-      SetDemuxer(CreateProgressiveDemuxer());
-    } else {
-      LOG(INFO) << "Cobalt progressive playback is disabled via base features.";
-      return DEMUXER_ERROR_PROGRESSIVE_DISABLED;
-    }
+    LOG(INFO) << "Progressive streams are unsupported.";
+    return DEMUXER_ERROR_NO_SUPPORTED_STREAMS;
```
borongc, line 351: We should check if this changes error-reporting to the metrics.

kjyoun (approved):

media/filters/demuxer_manager.cc
kjyoun: agree that we need to check it, though it would be a separate PR and bug, when the error is not reported.

osagie98: They both map to a network state [format error](https://source.corp.google.com/h/github/youtube/cobalt/+/main:third_party/blink/renderer/platform/media/web_media_player_impl.cc;l=387-397;drc=dd10cf81e71a27d4832126f7fcd0afbadc50111a), so I don't think it makes a difference for qoe metrics, however it will affect UMA.

borongc (approved):

borongc:
plz update PR descriptions. Other than this, LGTM.

media/starboard/BUILD.gn
```diff
@@ -18,7 +18,6 @@ import("//starboard/build/buildflags.gni")
 
 assert(use_starboard_media, "This file is only for use_starboard_media builds")
 
-import("//media/media_options.gni")
```
kjyoun, line 21: why do you need this change? 

is it related with removing progressive playback?

media/starboard/BUILD.gn
```diff
@@ -116,17 +94,14 @@ source_set("starboard") {
     "//mojo/public/cpp/bindings:bindings",
     "//starboard:starboard_group",
     "//starboard/common",
+    "//third_party/abseil-cpp:absl",
```
kjyoun, line 97: is this intended?

media/starboard/BUILD.gn
```diff
@@ -116,17 +94,14 @@ source_set("starboard") {
...
     "//url",
   ]
 
   if (is_android) {
     deps += [ "//media/base/android" ]
   }
 
-  if (!media_use_ffmpeg) {
```
kjyoun, line 126: removal of this code is intended?

kjyoun (commented):
I found some changes in build.gn, which seems to be not related with this PR.

are they intended?

media/starboard/BUILD.gn
osagie98: This is essentially a revert of https://github.com/youtube/cobalt/pull/4437. That PR imports the media_options.gni file to add checks for progressive support (`if (!media_use_ffmpeg)`) so I removed it

media/starboard/BUILD.gn
osagie98: Yes, it used to be in a progressive-only block as abseil was only used in progressive files. Now abseil also used in other media files, so I moved it here.

media/starboard/BUILD.gn
osagie98: Yes

media/starboard/BUILD.gn
kjyoun: thanks for clarification

can you update description to mention this revert https://github.com/youtube/cobalt/pull/4437 ?

It will help us tracking

osagie98: Sure, done

hlwarriner (approved):
LGTM from `Cobalt Web API` perspective

third_party/blink/renderer/core/html/media/html_media_element.cc
```diff
@@ -415,11 +415,9 @@ MIMETypeRegistry::SupportsType HTMLMediaElement::GetSupportsType(
...
   // Interrupt Chromium's IsTypeSupported() from here for better performance.
   MIMETypeRegistry::SupportsType result;
-  if (!base::FeatureList::IsEnabled(media::kCobaltProgressivePlayback) &&
-      IsProgressiveFormat(content_type)) {
-    LOG(INFO) << "Content type \'" << content_type.Raw()
-              << "\' is unsupported as Cobalt progressive playback is disabled "
-                 "via base features.";
+  if (IsProgressiveFormat(content_type)) {
+    LOG(INFO) << "Progressive content type \'" << content_type.Raw()
+              << "\' is unsupported.";
```
yell0wd0g, line 420: Just a thought that this tripped some people (they were confused
as to why it was so) so perhaps we could add here sth like
```
  Document& document = GetDocument(); 
  String message = "Cobalt does not support progressive video playback.";
  document.AddConsoleMessage(MakeGarbageCollected<ConsoleMessage>(
            mojom::ConsoleMessageSource::kJavaScript,
            mojom::ConsoleMessageLevel::kWarning, message));

```

WDYT?

osagie98: I like it, however when I just tested it, `GetSupportsType()` can't call `GetDocument()` since it's a static function. We would need to get a reference to the `HTMLMediaElement` object or log it somewhere else, like in `MediaSource::IsTypeSupported()`

PR #10189, opened 2026-04-22
tvos: Implement deep link URL handling

jkim-julie:
Configure the tvOS application to register a 'YouTube' URL scheme.
Incoming deep link URLs from scene:openURLContexts (for running apps)
and scene:willConnectToSession (for cold starts) are now routed to the
DeepLinkManager. This enables launching the application and navigating
to specific content using deep links, for example, from the command
line.

Bug: 452302320

### DIFF
--- a/cobalt/app/tvos/main.mm
+++ b/cobalt/app/tvos/main.mm
@@ -26,12 +26,14 @@
 #include "base/task/thread_pool.h"
 #include "cobalt/app/cobalt_main_delegate.h"
 #include "cobalt/app/cobalt_switch_defaults.h"
+#include "cobalt/browser/h5vcc_runtime/deep_link_manager.h"
 #include "cobalt/shell/browser/shell.h"
 #include "components/crash/core/app/crashpad.h"
 #include "content/public/app/content_main.h"
 #include "content/public/app/content_main_runner.h"
 #include "content/public/browser/render_view_host.h"
 #include "content/public/browser/render_widget_host.h"
+#include "net/base/apple/url_conversions.h"
 #include "starboard/common/command_line.h"
 #include "starboard/tvos/shared/application_darwin.h"
 
@@ -48,6 +50,27 @@ static const char** g_argv = nullptr;
   return content::Shell::windows()[0]->web_contents();
 }
 
+- (void)handleDeepLink:(NSSet<UIOpenURLContext*>*)URLContexts {
+  // URLContexts can contain multiple items, but since DeepLinkManager
+  // handles one link at a time, only a single context is used.
+  UIOpenURLContext* context = URLContexts.anyObject;
+  if (!context) {
+    return;
+  }
+  NSURL* url = context.URL;
+  if (!url) {
+    return;
+  }
+
+  GURL parsedURL = net::GURLWithNSURL(url);
+
+  if (!parsedURL.is_valid() || parsedURL.scheme().length() == 0) {
+    return;
+  }
+
+  cobalt::browser::DeepLinkManager::GetInstance()->OnDeepLink(parsedURL.spec());
+}
+
 - (void)scene:(UIScene*)scene
     willConnectToSession:(UISceneSession*)session
                  options:(UISceneConnectionOptions*)connectionOptions {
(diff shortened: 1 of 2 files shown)


cobalt/app/tvos/main.mm
```diff
@@ -48,6 +49,21 @@ @implementation CobaltAppSceneDelegate
   return content::Shell::windows()[0]->web_contents();
 }
 
+- (void)handleDeepLink:(NSSet<UIOpenURLContext*>*)URLContexts {
+  // URLContexts can contain multiple items, but this uses the first one since
+  // DeepLinkManager processes one link at a time.
+  UIOpenURLContext* context = URLContexts.allObjects.firstObject;
```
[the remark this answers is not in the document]
jkim-julie: Done.

cobalt/app/tvos/main.mm
```diff
@@ -48,6 +49,21 @@ @implementation CobaltAppSceneDelegate
...
+  if (!context) {
+    return;
+  }
+  NSURL* url = context.URL;
+  if (!url) {
+    return;
+  }
+  cobalt::browser::DeepLinkManager::GetInstance()->OnDeepLink(
+      url.absoluteString.UTF8String);
```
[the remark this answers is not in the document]
jkim-julie: Done.

jkim-julie:

cobalt/shell/app/tvos-Info.plist
```diff
@@ -33,5 +33,16 @@
   <string>This identifier will be used to deliver personalized ads to you.</string>
   <key>YTApplicationURL</key>
   <string>https://www.youtube.com/tv</string>
+  <key>CFBundleURLTypes</key>
+  <array>
+    <dict>
+      <key>CFBundleURLName</key>
+      <string>dev.cobalt.tvos</string>
```
rakuco: Shouldn't this be `${BUNDLE_IDENTIFIER}`?

cobalt/app/tvos/main.mm
```diff
@@ -48,6 +49,21 @@ @implementation CobaltAppSceneDelegate
...
+  UIOpenURLContext* context = URLContexts.anyObject;
+  if (!context) {
+    return;
+  }
+  NSURL* url = context.URL;
+  if (!url) {
+    return;
+  }
+  cobalt::browser::DeepLinkManager::GetInstance()->OnDeepLink(
+      url.absoluteString.UTF8String ?: "");
```
rakuco: I think there should be more validation here, similar to e.g. https://source.chromium.org/chromium/chromium/src/+/main:ios/chrome/app/startup/chrome_app_startup_parameters.mm;l=155;drc=4fea9fb486691d4d62c1005aac1251cb8b77c3cb:
- convert from NSURL to a GURL
- return early if the GURL is not valid
- pass `GURL::spec()` to `OnDeepLink()`

cobalt/shell/app/tvos-Info.plist
```diff
@@ -33,5 +33,16 @@
   <string>This identifier will be used to deliver personalized ads to you.</string>
   <key>YTApplicationURL</key>
   <string>https://www.youtube.com/tv</string>
+  <key>CFBundleURLTypes</key>
```
rakuco, line 36: @fayev-g do you know how this works internally with C25? Does the official app handle this schema and pass on these deep links by default as well? Just double checking to make sure there are no security concerns involved.

cobalt/shell/app/tvos-Info.plist
jkim-julie: Right! Done.

cobalt/app/tvos/main.mm
jkim-julie: Done.

rakuco (approved):
informal lgtm pending the question about unconditionally including these new entries in the plist

cobalt/shell/app/tvos-Info.plist
fayev-g: Yes, this schema looks like it was supported in C25. The values are slightly different but they will be overridden in the prod build anyway, so this is fine

fayev-g (approved):

cobalt/shell/app/tvos-Info.plist
```diff
@@ -33,5 +33,16 @@
...
+  <array>
+    <dict>
+      <key>CFBundleURLName</key>
+      <string>${BUNDLE_IDENTIFIER}</string>
+      <key>CFBundleURLSchemes</key>
+      <array>
+        <string>YouTube</string>
```
yell0wd0g, line 43: Not sure but we might need to add "YouYubeTV" here as well?
For the other app. @fayev-g would know.

yell0wd0g (approved):
RS LGTM if @fayev-g is happy.

cobalt/shell/app/tvos-Info.plist
jkim-julie: I think it makes sense.
Added "YouTubeTV" for CFBundleURLSchemes.

PR #10205, opened 2026-04-23
Remove `__gcov_dump` and `__gcov_flush`

sacuff:
Bug: 505819427

### DIFF
--- a/third_party/abseil-cpp/absl/log/internal/log_message.cc
+++ b/third_party/abseil-cpp/absl/log/internal/log_message.cc
@@ -60,6 +60,7 @@
 #include "absl/time/clock.h"
 #include "absl/time/time.h"
 #include "absl/types/span.h"
+#include "build/build_config.h"
 
 extern "C" ABSL_ATTRIBUTE_WEAK void ABSL_INTERNAL_C_SYMBOL(
     AbslInternalOnFatalLogMessage)(const absl::LogEntry&) {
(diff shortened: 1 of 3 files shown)


third_party/abseil-cpp/absl/log/internal/log_message.cc
```diff
@@ -373,7 +374,7 @@ void LogMessage::FailWithoutStackTrace() {
   __debugbreak();
 #endif
 
-#ifdef __ELF__
+#if defined(__ELF__) && !BUILDFLAG(IS_STARBOARD)
   // For b/8737634, flush coverage if we are in coverage mode.
   if (&__gcov_dump != nullptr) {
```
andrewsavage1, line 379: @y4vor any thoughts on this pattern of checking that the function actually exists while weakly linking it? I wonder if it's cleaner to just allow the leak as a weak symbol instead, though we wouldn't know if someone's using it without this check in the future

andrewsavage1 (approved):

third_party/abseil-cpp/absl/log/internal/log_message.cc
sacuff: FWIW this pattern also came up for b/505845827 with https://github.com/llvm/llvm-project/blob/8b87c82fdc8c0e3dbd01602ba9d91832d7e3db3e/libcxxabi/src/cxa_thread_atexit.cpp#L116

andrewsavage1: Feel free to merge without waiting for yavor to comment here

PR #10208, opened 2026-04-23
media: Add AV2 to SbMediaVideoCodec

osagie98:
This change introduces the AV2 video codec to the Starboard API.
Adding `kSbMediaVideoCodecAv2` to the `SbMediaVideoCodec` enum prepares
Starboard for future integration of AV2 decoding
capabilities. Related helper functions and NPLB tests are updated
to recognize the new codec.

Also added a `BUILDFLAG` `COBALT_MEDIA_ENABLE_AV2_SUPPORT` that can be enabled once AV2 is supported in Chromium.

Bug: 505744317


starboard/common/media.cc
```diff
@@ -624,6 +624,10 @@ const char* GetMediaVideoCodecName(SbMediaVideoCodec codec) {
       return "vp8";
     case kSbMediaVideoCodecVp9:
       return "vp9";
+#if SB_API_VERSION >= 18
```
kjyoun: cant we just use `COBALT_MEDIA_ENABLE_AV2_SUPPORT`?

I think that its more readable. 
and we can enable it only in SB_API_VERSION >= 18

```
   # TODO(b/505744317): Enable AV2 once supported in Chromium.
   # This flasg should be enabled in SB_API_VERSION >= 18
    "COBALT_MEDIA_ENABLE_AV2_SUPPORT=0",

```


WDYT?

starboard/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Implementations of `DecodeTarget` for YUV planes should ensure they use
 `GL_RED_EXT` or `GL_LUMINANCE` instead of `GL_ALPHA` to avoid green-screen
 rendering issues caused by Chromium's shader expectations.
 
+### Added AV2 to SbMediaVideoCodec
+This prepares Cobalt for future AV2 playback support.
+
```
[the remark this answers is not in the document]
osagie98: Not useful, it's already under version 18

media/starboard/BUILD.gn
```diff
@@ -57,6 +57,9 @@ source_set("starboard") {
 
     # TODO(b/375241265): Re-enable UMA metrics
     "COBALT_MEDIA_ENABLE_UMA_METRICS=0",
+
+    # TODO(b/505744317): Enable AV2 once supported in Chromium.
+    "COBALT_MEDIA_ENABLE_AV2_SUPPORT=0",
```
xiaomings: Let's get a test stream to see if we can trivially enable av2 directly in Chrobalt for C27.

starboard/nplb/player_create_test.cc
```diff
@@ -399,6 +402,9 @@ TEST_F(SbPlayerTest, MultiPlayer) {
     case kVideoCodecs[6]:
     case kVideoCodecs[7]:
     case kVideoCodecs[8]:
+#if SB_API_VERSION >= 18
+    case kVideoCodecs[9]:
+#endif  // SB_API_VERSION >= 18
```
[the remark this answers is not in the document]
osagie98: If we'd like to make this change, it'd be in a different PR as it's outside of the scope

starboard/nplb/media_buffer_test.cc
```diff
@@ -144,6 +147,9 @@ TEST(SbMediaBufferTest, VideoCodecs) {
     case kVideoCodecs[6]:
     case kVideoCodecs[7]:
     case kVideoCodecs[8]:
+#if SB_API_VERSION >= 18
+    case kVideoCodecs[9]:
+#endif  // SB_API_VERSION >= 18
```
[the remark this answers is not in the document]
osagie98: Same as in player_create_test.cc

starboard/common/media.cc
osagie98: I use both in starboard_utils.cc, would that work?

kjyoun: Why both, if one is enough?  

one flag is more readable/manageable than two flags.

for one flag, I would prefer `COBALT_MEDIA_ENABLE_AV2_SUPPORT`

Then the change is prettty self-explaininig


```
#if COBALT_MEDIA_ENABLE_AV2_SUPPORT
    case kSbMediaVideoCodecAv2:
      return "av2";
#endif
````

and can't we use BUILDFLAG here?  Then it make code more safe
It would generate compile error, if the given build flag is not defined.


```
#if BUILDFLAG(COBALT_MEDIA_ENABLE_AV2_SUPPORT)
```

per Gemini
---

2. The Power of BUILDFLAG
Switching to BUILDFLAG is a significant upgrade over standard #if or #ifdef macros. Here is why:

Typo Protection: If you use #if DEFINED_FLAG and accidentally type #if DEFINED_FLAG_TYPO, the compiler treats the undefined macro as 0 (false) and quietly skips the code. With BUILDFLAG(NAME), the build will fail immediately if the flag hasn't been explicitly defined in the build header.

Explicit State: It forces the flag to be either 0 or 1, rather than just "existing" or "not existing." This makes the intent of the code much more transparent.

File: media/starboard/BUILD.gn [hunk 6]
```diff
@@ -18,7 +18,7 @@ import("//starboard/build/buildflags.gni")
 
 assert(use_starboard_media, "This file is only for use_starboard_media builds")
 
-import("//starboard/build/buildflags.gni")
```
osagie98: This change is intentional as it was imported twice

starboard/common/media.cc
osagie98: Added a buildflag. We should probably do the same with other cobalt media guards. I also removed the starboard api check since we don't seem to be using it elsewhere

media/starboard/BUILD.gn
osagie98: Spoke with the codecs team, there's no test stream available yet since the bitstream structure isn't finalized yet. AV2 hasn't been added to Chromium yet as well.

media/starboard/starboard_utils.cc
```diff
@@ -97,6 +98,10 @@ SbMediaVideoCodec MediaVideoCodecToSbMediaVideoCodec(VideoCodec codec) {
       return kSbMediaVideoCodecH265;
     case VideoCodec::kAV1:
       return kSbMediaVideoCodecAv1;
+#if BUILDFLAG(COBALT_MEDIA_ENABLE_AV2_SUPPORT)
```
borongc, line 101: We should check if supporting av2 in starboard layer is sufficient, because Chromium also checks codecs (e.g., in media/base/supported_types.cc). If we don't have av2 codec there, it will still be rejected by chromium before check IsTypeSupported from Starboard.

osagie98: Chromium doesn't have any AV2 support yet, so the VideoCodec::kAv2 is a placeholder since that enum doesn't exist. It should be there in the next few months, we're just a bit ahead. Once we pull those changes we can enable the flag for testing

kjyoun (approved):

y4vor (approved):

PR #10214, opened 2026-04-23
Use a little `nm` in API leak detector

sacuff:
During the grep step in the leak detector, object files can incorrectly match the searched symbol in the (so far rare) case that the symbol appears in some other data. For example, `dlsym` matches native_library_starboard.o because there is a string mentioning that dlsym is unsupported, and putil.o because it defines a separate `uprv_dlsym_func` symbol.

This adds a post-grep `nm` step to double check the symbol matches. That does make the `--only-grep` flag a bit misleading now, but execution time does not seem to be significantly affected.

Also adds a `--skip-build` flag to shave off the roughly 10 seconds that it takes autoninja to exit if the build is already up to date.

Bug: 450956860

### DIFF
--- a/starboard/tools/api_leak_detector/api_leak_detector.py
+++ b/starboard/tools/api_leak_detector/api_leak_detector.py
@@ -458,11 +458,11 @@ def FindLeakLocations(leaked_symbols, config_path, only_grep, use_ripgrep):
     # similarly named symbol. Unfortunately this doesn't work for Rust libs, so
     # we just assume those are matches.
     leaking_files[_UNKNOWN_LIBRARIES][symbol] = set()
-    for i, obj_file in enumerate(obj_files_to_check):
+    for obj_file in obj_files_to_check:
       if os.path.splitext(obj_file)[1] == '.rlib' or symbol in ProcessNmOutput(
           RunCommand(['nm', '-u', '-C', obj_file])):
         leaking_files[_UNKNOWN_LIBRARIES][symbol].add(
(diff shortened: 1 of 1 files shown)


sacuff:
This has an interesting effect of removing Rust's liblibc_lib.rlib from the output, since it doesn't seem to actually contain any symbols beyond a few types:

```
nm -C out/evergreen-x64_devel/obj/third_party/rust/libc/v0_2/lib/liblibc_lib.rlib 

lib.rmeta:
nm: lib.rmeta: no symbols

liblibc_lib.libc.7a35e47d60bac5e5-cgu.0.rcgu.o:
0000000000000000 T libc::unix::linux_like::linux::gnu::<impl libc::unix::linux_like::linux::gnu::b64::x86_64::siginfo_t>::si_pid
0000000000000000 T libc::unix::linux_like::linux::gnu::<impl libc::unix::linux_like::linux::gnu::b64::x86_64::siginfo_t>::si_uid
0000000000000000 T libc::unix::linux_like::linux::gnu::<impl libc::unix::linux_like::linux::gnu::b64::x86_64::siginfo_t>::si_stime
0000000000000000 T libc::unix::linux_like::linux::gnu::<impl libc::unix::linux_like::linux::gnu::b64::x86_64::siginfo_t>::si_utime
0000000000000000 T libc::unix::linux_like::linux::gnu::<impl libc::unix::linux_like::linux::gnu::b64::x86_64::siginfo_t>::si_status
0000000000000000 T libc::unix::linux_like::linux::_IO
```

andrewsavage1 (approved):

PR #10215, opened 2026-04-23
net: Enable built-in Async DNS client

zhongqiliang:
This change introduces a new feature flag, kEnableAsyncDNSClient, to
control the activation of Chromium's built-in asynchronous DNS client.
Enabling this flag configures the HostResolverManager to use the
insecure DNS client, leveraging Chromium's native asynchronous DNS
resolution capabilities. This can improve DNS lookup efficiency and
resilience. An Android Java switch is added to allow enabling this
feature via command-line arguments on the Android platform.

Bug: 505401635


services/network/network_service.cc
```diff
@@ -469,9 +469,17 @@ void NetworkService::Initialize(mojom::NetworkServiceParamsPtr params,
 
   dns_config_change_manager_ = std::make_unique<DnsConfigChangeManager>();
 
-  host_resolver_manager_ = std::make_unique<net::HostResolverManager>(
-      net::HostResolver::ManagerOptions(),
-      net::NetworkChangeNotifier::GetSystemDnsConfigNotifier(), net_log_);
+  if (base::FeatureList::IsEnabled(net::features::kHappyEyeballsV3)) {
+    net::HostResolver::ManagerOptions options;
+    options.insecure_dns_client_enabled = true;
```
johnxwork, line 479: This is a different setting than HappyeyeballsV3. We shouldn't group them together. You can see they are parallel:
- [SetInsecureDnsClientEnabled](https://source.corp.google.com/h/github/youtube/cobalt/+/main:net/dns/host_resolver_manager.h;drc=524126904340bf2104b6615d6bb6352ba6d7d844;l=191)
- [SetIsHappyEyeballsV3Enabled](https://source.corp.google.com/h/github/youtube/cobalt/+/main:net/dns/host_resolver_manager.h;drc=524126904340bf2104b6615d6bb6352ba6d7d844;l=202)

Also, is this something we can just specify at the embedder level?

zhongqiliang: WIthout set insecure_dns_client_enabled to true, the HappyEyeballs feature is not applied, it still fallback to Android 's getaddrinfo.

zhongqiliang: Comment from the author:
```
Hi Colin, you probably need insecure_dns_client_enabled = true since the default value is false. Chrome usually set it to true [1] so HEv3 works in Chrome if the feature flag is enabled.

(BTW, we paused the HEv3 project since it took too many time than we initially thought, and the feature is still premature, thank you for trying to use it anyway)

[1] https://source.chromium.org/chromium/chromium/src/+/main:services/network/network_context.cc;l=2191;drc=5d648a203bc1df8ff70266c6b3d1b8c9f90208f0
```

PR #10222, opened 2026-04-24
starboard/tests: wire opus_tests target

denisyuji:
Split per target for easier review.

Includes:
- target-specific Starboard wiring commit(s)
- target-specific CI commit that enables target resolution and adds that same target to disabled in the same commit

Bug: 486053350

PR #10223, opened 2026-04-24
starboard/tests: wire test_opus_encode target

denisyuji:

Includes:

Bug: 486053350

### DIFF
--- a/third_party/opus/BUILD.gn
+++ b/third_party/opus/BUILD.gn
@@ -552,6 +552,9 @@ test("test_opus_encode") {
   ]
 
   deps = [ ":opus" ]
+  if (is_starboard) {
+    deps += [ ":sbeventhandle_opus_tests" ]
+  }
 }
 
 test("test_opus_decode") {
@@ -565,6 +568,9 @@ test("test_opus_decode") {
   ]
 
   deps = [ ":opus" ]
+  if (is_starboard) {
+    deps += [ ":sbeventhandle_opus_tests" ]
+  }
 }
 
 test("test_opus_padding") {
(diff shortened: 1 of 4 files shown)


third_party/opus/starboard/sbeventhandle_main_no_loader_args.cc
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/client_porting/wrap_main/wrap_main.h"
+
+int main(int argc, char** argv);
```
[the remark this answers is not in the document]
denisyuji: done

third_party/opus/BUILD.gn
```diff
@@ -536,13 +536,22 @@ test("test_opus_encode") {
     "src/tests/test_opus_encode.c",
   ]
 
+  if (is_starboard) {
```
andrewsavage1: It looks like these changes are the same as it is in other targets in #10224 and #10222. I recommend putting it in a separate static_library or source_set and having the different targets depend on it instead of making the same changes in each one separately

denisyuji: @andrewsavage1 , makes sense. I created one `source_set("sbeventhandle_opus_tests")` defined once now.
Note in these PRs that there is a chain of PRs,to be merged in this order:

- ci/tests: wire and quarantine test_opus_encode 
- ci/tests: wire and quarantine test_opus_decode
- ci/tests: wire and quarantine test_opus_padding

When I change one PR branch, I rebase the downstream PR branches - that's why the same lines are replicated in different PRs.

PR #10224, opened 2026-04-24
starboard/tests: wire test_opus_decode target

denisyuji:

Includes:

Bug: 486053350

### DIFF
--- a/third_party/opus/BUILD.gn
+++ b/third_party/opus/BUILD.gn
@@ -552,6 +552,9 @@ test("test_opus_encode") {
   ]
 
   deps = [ ":opus" ]
+  if (is_starboard) {
+    deps += [ ":sbeventhandle_opus_tests" ]
+  }
 }
 
 test("test_opus_decode") {
--- a/third_party/opus/BUILD.gn
+++ b/third_party/opus/BUILD.gn
@@ -568,6 +568,9 @@ test("test_opus_decode") {
   ]
 
   deps = [ ":opus" ]
+  if (is_starboard) {
+    deps += [ ":sbeventhandle_opus_tests" ]
+  }
 }
 
 test("test_opus_padding") {
--- a/cobalt/build/testing/targets/evergreen-arm-hardfp-rdk/test_targets.json
+++ b/cobalt/build/testing/targets/evergreen-arm-hardfp-rdk/test_targets.json
@@ -19,7 +19,6 @@
       "gl:ozone_gl_unittests_loader",
       "third_party/blink/renderer/platform/heap:blink_heap_unittests_loader",
       "third_party/ipcz/src:ipcz_tests_loader",
-      "third_party/opus:test_opus_decode_loader",
       "third_party/opus:test_opus_padding_loader"
     ]
   },
@@ -106,5 +105,9 @@
   {
     "target": "third_party/opus:test_opus_encode_loader",
     "executable": "out/evergreen-arm-hardfp-rdk_devel/test_opus_encode_loader.py"
+  },
+  {
+    "target": "third_party/opus:test_opus_decode_loader",
+    "executable": "out/evergreen-arm-hardfp-rdk_devel/test_opus_decode_loader.py"
   }
 ]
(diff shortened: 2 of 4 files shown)


third_party/opus/starboard/sbeventhandle_main_no_loader_args.cc
```diff
@@ -0,0 +1,30 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/client_porting/wrap_main/wrap_main.h"
+
+int main(int argc, char** argv);
```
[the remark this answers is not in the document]
denisyuji: done

third_party/opus/BUILD.gn
```diff
@@ -556,6 +560,12 @@ test("test_opus_decode") {
   ]
 
   deps = [ ":opus" ]
+  if (is_cobalt_hermetic_build) {
+    deps += [
+      "//starboard/common",
+      "//starboard/elf_loader:sabi_string",
+    ]
+  }
```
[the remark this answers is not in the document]
denisyuji: done

third_party/opus/starboard/sbeventhandle_main_no_loader_args.cc
niranjanyardi, line 5: is this file still needed ? given you already amde changes to run_all_unittests

denisyuji: @niranjanyardi ,  good question and yes, this file is still needed for the Opus test targets.

The `run_all_unittests` changes cover gtest-based binaries, but `test_opus_api`, `test_opus_encode`, and `test_opus_decode` use their own `main` and fail if loader args are passed through.
 This wrapper strips loader args only for those Opus tests.

I wouldn’t put it in `storage/common/run_all_unittests.cc`, because that file is a storage gtest entrypoint (`InitAndRunAllTests/RUN_ALL_TESTS`), while the Opus case needs a wrapper that calls the Opus test main with filtered args.

denisyuji: Based on the feedback above, I created a [common branch/commit](https://github.com/youtube/cobalt/pull/10224/changes/b8664a82dc446f1ae7b4a667925684e136fbffa8) introducing the code that is necessary for all the Opus tests.

 I also renamed the `third_party/opus/starboard/sbeventhandle_main_no_loader_args.cc` to `third_party/opus/starboard/sbeventhandle_opus_tests.cc` to make it clear it covers the `opus` tests.

denisyuji (commented):
Adressing feedbacks

third_party/opus/BUILD.gn
```diff
@@ -548,6 +548,9 @@ test("test_opus_encode") {
 test("test_opus_decode") {
   sources = [ "src/tests/test_opus_decode.c" ]
 
+  if (is_starboard) {
```
andrewsavage1: nit: minimize conditionals by putting the sources += in this block instead

third_party/opus/BUILD.gn
```diff
@@ -556,6 +559,12 @@ test("test_opus_decode") {
   ]
 
   deps = [ ":opus" ]
+  if (is_starboard) {
+    deps += [
+      "//starboard/common",
+      "//starboard/elf_loader:sabi_string",
```
andrewsavage1: Why is this needed?

andrewsavage1 (approved):

third_party/opus/BUILD.gn
denisyuji: @andrewsavage1 , you are right, thanks for the review. See my comment on [10222#discussion_r3308446420]( https://github.com/youtube/cobalt/pull/10222#discussion_r3308446420). I dropped these lines.

third_party/opus/BUILD.gn
denisyuji: @andrewsavage1 , makes sense. I created one `source_set("sbeventhandle_opus_tests")` defined once now.

PR #10229, opened 2026-04-24
tvOS: Remove some deprecated and empty app lifecycle functions

rakuco:
These functions are deprecated and never called since this code uses UIKit scenes.

Bug: 506073148


rakuco:

fayev-g (approved):

PR #6276, opened 2025-07-02
media: Add BASE feature for Android Overlay

jxqu2048:
Implements a BASE Feature, `CobaltUsingAndroidOverlay`, for SbPlayer using `AndroidOverlay`, and this is disabled by default.

This feature is used to enable `StarboardRenderer` to create `AndroidOverlay`, so SbPlayer can use the SurfaceView of `AndroidOverlay`, instead of VideoSurfaceView from StarboardBridge.

This ensures the life cycle of SurfaceView is managed by StarboardRenderer, instead of in StarboardBridge.

Issue: 428017363

PR #5271, opened 2025-04-01
Stop creating a11y node provder for Chrobalt

haozheng-cobalt:
This is a better Chromium a11y tree bypass than before.
Before we only bypassed navigation for WebViews, and we would ran into focus issues with videos which are VideoSurfaceView. This PR stops creating creating a11y node provder for Chrobalt.

Test: https://paste.googleplex.com/4565574762627072

b/391708407

PR #10233, opened 2026-04-24
android: Use Lowercase for AndroidOverlay switch

borongc:
Standardize the string used for the Cobalt Android overlay switch
to use lowercase and hyphens. This aligns with common naming
conventions for flags and properties, improving consistency and
readability across the codebase.

Issue: 505895356

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -77,6 +77,7 @@ public abstract class CobaltActivity extends Activity {
   private static final String META_DATA_ENABLE_SPLASH_SCREEN = "cobalt.ENABLE_SPLASH_SCREEN";
   private static final String META_DATA_ENABLE_FEATURES = "cobalt.ENABLE_FEATURES";
   private static final String YOUTUBE_URL = "https://www.youtube.com/tv";
+  private static final String COBALT_USING_ANDROID_OVERLAY = "cobalt-using-android-overlay";
 
   // This key differs in naming format for legacy reasons
   public static final String COMMAND_LINE_ARGS_KEY = "commandLineArgs";
@@ -113,7 +114,6 @@ public abstract class CobaltActivity extends Activity {
   private Boolean mIsKeepScreenOnEnabled = false;
 
   private boolean mIsCobaltUsingAndroidOverlay;
-  private static final String COBALT_USING_ANDROID_OVERLAY = "CobaltUsingAndroidOverlay";
 
   private boolean mEnableSplashScreen;
   private String mStartDeepLink;


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -113,7 +113,7 @@ public abstract class CobaltActivity extends Activity {
   private Boolean mIsKeepScreenOnEnabled = false;
 
   private boolean mIsCobaltUsingAndroidOverlay;
-  private static final String COBALT_USING_ANDROID_OVERLAY = "CobaltUsingAndroidOverlay";
```
zhongqiliang, line 116: Should this be moved into JavaSwitches.java?

borongc: No, this is for BASE without explicitly passing it from C++ side.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -113,7 +113,7 @@ public abstract class CobaltActivity extends Activity {
...
+  private static final String COBALT_USING_ANDROID_OVERLAY = "cobalt-using-android-overlay";
```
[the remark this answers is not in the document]
borongc: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: Is this flag passed from command line argument or is it passed from Kimono code?
If it is from command line argument, any doc for when and how to use it?

borongc: It's passed from `command line argument`.

To use this feature, we can use:
```
$ adb shell am start --esa commandLineArgs '--enable-features=CobaltUsingAndroidOverlay,--cobalt-using-android-overlay' dev.cobalt.coat/dev.cobalt.app.MainActivity
```

zhongqiliang: Is this only for local test? Or is it used on production? If it is used on production, how does the flag being passed in?
I am asking because this is the first command line flag that stays in this location. We need to set the correct pattern for people to follow.

borongc: C26 uses Java switch to run experiments with this feature in prod, but we see some regressions.

For C27, I don't have a plan to re-run experiments again, so just keep it as a local test feature.

zhongqiliang: Do you think, we should let GEMINI to refactor COMMAND_LINE_ARGS_KEY and COBALT_USING_ANDROID_OVERLAY into a separate class, should be a subclass of org.chromium.base.CommandLine.

borongc: No, the actual right way for this is to add a content feature from C++ layer, and pass it via ContentFeatureList in content/public/android/java/src/org/chromium/content_public/browser/ContentFeatureList.java, so we can use it directly in java without defining any extra names. See b/428017363 for more info.

It can be simplified as just:
```
$ adb shell am start --esa commandLineArgs '--enable-features=CobaltUsingAndroidOverlay' dev.cobalt.coat/dev.cobalt.app.MainActivity
```

However, there are historical discussion (https://github.com/youtube/cobalt/pull/6276) that we may not want to add this via ContentFeatureList, resulting in we added one from BASE the other from Java side.

For me, I am happy to use ContentFeatureList for this, as that's the Chromium way to add it. However, from Cobalt upstream side, this feature may or may not be the final feature we should keep, so adding it from both C++ and Java side for local test or experiments is fine.

If you would like to refactor this to two classes, feel free to do it by the following PR, but I think it is out-of-scope of this PR.

zhongqiliang: It is okey to submit as it is now. 
But when it needs the second flag, I want to refactor, can we extend from ContentFeatureList to put Cobalt features?

zhongqiliang (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
borongc: > can we extend from ContentFeatureList to put Cobalt features?

We could, but there is no build flag we can gate it from Java side, that's why https://github.com/youtube/cobalt/pull/6276 didn't merge to main.

borongc: We have an existing PR using ContentFeatureList: https://github.com/youtube/cobalt/pull/5271.

PR #10245, opened 2026-04-27
Add fallback mechanism for kSbSystemPropertyCertificationScope

abellin7:
Read CERT_SCOPE defined in /etc/device.properties if the certification scope is NOT found by existing methods. This unifies the management of certification scope and eliminates the need to manually set the COBALT_CERT_SCOPE_FILE_NAME environment variable.

Bug: 477465556


jonastsai:
Overall it looks okay. but please consider the gemini commit not to use do{} while(false).

PR #10247, opened 2026-04-27
build: add more //base => //copied_base/base redirections

hferreiro:
Partner toolchains should use //copied_base/base instead of //base.
The updated targets were additionally creating conflicting ninja rules, as both versions generate their files at toolchain_dir/gen/base/.

This PR redirects the //base targets to //copied_base/base, but keeps the generated file paths, as they are now unique per toolchain.

Bug: 495203133


andrewsavage1:
I'm surprised this is an issue—I would expect because we're only building copied_base with the starboard and native_target toolchains that generated files would be output to `<out>/starboard/gen` and `<out>/native_target/gen`.

hferreiro:
yes, the problem is that both copied_base and base create rules that output to `<out>/starboard/gen/base/..`

hferreiro:
@andrewsavage1 switched approach and updated the PR description. PTAL.

andrewsavage1 (approved):

PR #10250, opened 2026-04-27
android: Support video frame VSP adjustment

jasonzhangxx:
This change introduces an experimental feature to support video frame
presentation timestamp (PTS) adjustment for varied speed playback (VSP).
Platforms that do not natively support direct video VSP can now rely on the
video renderer to synchronize video frames with the audio clock when playback
speeds deviate from 1.0x. This ensures smoother and more accurate playback.

Key changes include new MediaTimeProvider methods to expose audio clock
information and enable timestamp conversion, implemented in AudioRendererPcm
using a refactored AudioFrameTracker. The VideoRendererImpl now, when the
experimental flag is enabled, delays video frames and adjusts their PTS
based on the audio clock before submission. This allows the audio sink to
operate at a 1.0x playback rate while the video renderer manages the
effective VSP.

Bug: 454305379

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecFrameRateEstimator.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecFrameRateEstimator.java
@@ -90,7 +90,6 @@ public final class MediaCodecFrameRateEstimator {
     private static final int MINIMUM_REQUIRED_TUNNEL_FRAMES = 8;
     private static final long INVALID_FRAME_TIMESTAMP = -1;
     private final long[] mSortedTimestampsUs = new long[WINDOW_SIZE];
-    private long mMinDeltaUs = INVALID_FRAME_TIMESTAMP;
     private int mCount = 0;
 
     @Override
@@ -99,25 +98,16 @@ public final class MediaCodecFrameRateEstimator {
         return INVALID_FRAME_RATE;
       }
 
-      if (mMinDeltaUs == INVALID_FRAME_TIMESTAMP) {
-        for (int i = 1; i < mCount; i++) {
-          long delta = mSortedTimestampsUs[i] - mSortedTimestampsUs[i - 1];
-          if (mMinDeltaUs == INVALID_FRAME_TIMESTAMP || delta < mMinDeltaUs) {
-            mMinDeltaUs = delta;
-          }
-        }
-      }
-
       int validTimestamps = 1;
+      long previousDelta = mSortedTimestampsUs[1] - mSortedTimestampsUs[0];
       for (; validTimestamps < mCount; validTimestamps++) {
         long delta =
             mSortedTimestampsUs[validTimestamps] - mSortedTimestampsUs[validTimestamps - 1];
-        if (delta < mMinDeltaUs) {
-          mMinDeltaUs = delta;
-        } else if (delta > 2 * mMinDeltaUs) {
+        if (delta > 1.5 * previousDelta) {
           // The delta is too large.
           break;
         }
+        previousDelta = delta;
       }
 
       if (validTimestamps < MINIMUM_REQUIRED_TUNNEL_FRAMES) {
@@ -134,7 +124,6 @@ public final class MediaCodecFrameRateEstimator {
     @Override
     public void reset() {
       mCount = 0;
-      mMinDeltaUs = INVALID_FRAME_TIMESTAMP;
       Arrays.fill(mSortedTimestampsUs, 0);
     }
 
--- a/starboard/android/shared/player_components_factory.cc
+++ b/starboard/android/shared/player_components_factory.cc
@@ -70,8 +70,10 @@ bool UseLibopusDecoder(SbMediaAudioCodec codec,
 // This class allows us to force int16 sample type when tunnel mode is enabled.
 class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
  public:
-  explicit AudioRendererSinkAndroid(int tunnel_mode_audio_session_id = -1,
-                                    bool allow_audio_writing_on_pause = false)
+  explicit AudioRendererSinkAndroid(
+      int tunnel_mode_audio_session_id = -1,
+      bool allow_audio_writing_on_pause = false,
+      bool enable_video_renderer_vsp_adjustment = false)
       : AudioRendererSinkImpl(
             [=](int64_t start_media_time,
                 int channels,
@@ -96,14 +98,17 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
                   /*is_web_audio=*/false, allow_audio_writing_on_pause,
                   context);
             }),
-        tunnel_mode_audio_session_id_(tunnel_mode_audio_session_id) {}
+        tunnel_mode_audio_session_id_(tunnel_mode_audio_session_id),
+        enable_video_renderer_vsp_adjustment_(
+            enable_video_renderer_vsp_adjustment) {}
 
   bool AllowOverflowAudioSamples() const override {
     return tunnel_mode_audio_session_id_ != -1;
   }
 
   bool AllowDirectPlaybackRateSetting() const override {
-    return tunnel_mode_audio_session_id_ != -1;
+    return tunnel_mode_audio_session_id_ != -1 &&
+           !enable_video_renderer_vsp_adjustment_;
   }
 
   void GetAudioRendererParams(const AudioStreamInfo& audio_stream_info,
@@ -163,6 +168,7 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
   }
 
   const int tunnel_mode_audio_session_id_;
+  const bool enable_video_renderer_vsp_adjustment_;
 };
 
 class AudioRendererSinkCallbackStub : public AudioRendererSink::RenderCallback {
@@ -425,6 +431,11 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
       SB_LOG_IF(INFO, allow_audio_writing_on_pause)
           << "allow_audio_writing_on_pause is set to true.";
 
+      const bool enable_video_renderer_vsp_adjustment =
+          experimental_features.enable_video_renderer_vsp_adjustment;
+      SB_LOG_IF(INFO, enable_video_renderer_vsp_adjustment)
+          << "enable_video_renderer_vsp_adjustment is set to true.";
+
       const bool force_platform_opus_decoder = force_platform_opus_decoder_;
       auto decoder_creator =
           [enable_flush_during_seek, force_platform_opus_decoder, job_queue](
@@ -455,7 +466,8 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
 
       components.audio.renderer_sink =
           std::make_unique<AudioRendererSinkAndroid>(
-              tunnel_mode_audio_session_id, allow_audio_writing_on_pause);
+              tunnel_mode_audio_session_id, allow_audio_writing_on_pause,
+              enable_video_renderer_vsp_adjustment);
     }
 
     if (creation_parameters.video_codec() != kSbMediaVideoCodecNone) {
@@ -465,6 +477,13 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
       SB_LOG_IF(INFO, max_video_input_size > 0)
           << "The maximum size in bytes of a buffer of data is "
           << max_video_input_size;
+
+      if (experimental_features.enable_video_renderer_vsp_adjustment &&
+          !experimental_features.allow_audio_writing_on_pause) {
+        return Failure(
+            "Video renderer vsp adjustment needs to be enabled with audio "
+            "writing on pause.");
+      }
 
       if (tunnel_mode_audio_session_id == -1) {
         force_secure_pipeline_under_tunnel_mode = false;
--- a/starboard/shared/starboard/player/filter/audio_frame_tracker.cc
+++ b/starboard/shared/starboard/player/filter/audio_frame_tracker.cc
@@ -15,6 +15,7 @@
 #include "starboard/shared/starboard/player/filter/audio_frame_tracker.h"
 
 #include <queue>
+#include <sstream>
 
 #include "starboard/common/check_op.h"
 #include "starboard/common/log.h"
@@ -25,8 +26,10 @@ namespace starboard {
 
 void AudioFrameTracker::Reset() {
   frame_records_.clear();
-  frames_played_adjusted_to_playback_rate_ = 0;
+  last_playback_rate_ = 1.0;
   overflowed_frames_ = 0;
+  current_original_frame_head_position_ = 0;
+  current_adjusted_frame_head_position_ = 0;
 }
 
 void AudioFrameTracker::AddFrames(int number_of_frames, double playback_rate) {
@@ -40,7 +43,8 @@ void AudioFrameTracker::AddFrames(int number_of_frames, double playback_rate) {
 
   if (frame_records_.empty() ||
       frame_records_.back().playback_rate != playback_rate) {
-    FrameRecord record = {number_of_frames, playback_rate};
+    FrameRecord record = {number_of_frames, 0 /* number_of_played_frames */,
+                          playback_rate};
     frame_records_.push_back(record);
   } else {
     frame_records_.back().number_of_frames += number_of_frames;
@@ -50,20 +54,40 @@ void AudioFrameTracker::AddFrames(int number_of_frames, double playback_rate) {
 void AudioFrameTracker::RecordPlayedFrames(int number_of_frames) {
   while (number_of_frames > 0 && !frame_records_.empty()) {
     FrameRecord& record = frame_records_.front();
-    if (record.number_of_frames > number_of_frames) {
-      frames_played_adjusted_to_playback_rate_ +=
-          static_cast<int>(number_of_frames * record.playback_rate);
-      record.number_of_frames -= number_of_frames;
+    int64_t unplayed_frames =
+        record.number_of_frames - record.number_of_played_frames;
+    if (unplayed_frames >= number_of_frames) {
+      record.number_of_played_frames += number_of_frames;
       number_of_frames = 0;
     } else {
-      number_of_frames -= record.number_of_frames;
-      frames_played_adjusted_to_playback_rate_ +=
-          static_cast<int>(record.number_of_frames * record.playback_rate);
+      number_of_frames -= unplayed_frames;
+      // Note that since |current_original_frame_head_position_| is an integer,
+      // multiplying by a non-integer |playback_rate| can lead to truncation
+      // errors. These small errors accumulate every time a record is removed,
+      // potentially leading to a slight drift in the original timeline over
+      // long periods of playback. Remove record after it's fully played to
+      // minimize that drift.
+      current_original_frame_head_position_ +=
+          record.number_of_frames * record.playback_rate;
+      current_adjusted_frame_head_position_ += record.number_of_frames;
       frame_records_.erase(frame_records_.begin());
     }
   }
-
   overflowed_frames_ += number_of_frames;
+}
+
+int64_t AudioFrameTracker::GetOverflowedFrames() const {
+  // Overflowed frames calculation uses last playback rate, so any playback rate
+  // change after audio end of stream is ignored.
+  return overflowed_frames_ * last_playback_rate_;
+}
+
+int64_t AudioFrameTracker::GetTotalOriginalFrames() const {
+  int64_t ret = current_original_frame_head_position_;
+  for (auto& record : frame_records_) {
+    ret += record.number_of_frames * record.playback_rate;
+  }
+  return ret;
 }
 
 int64_t AudioFrameTracker::GetFutureFramesPlayedAdjustedToPlaybackRate(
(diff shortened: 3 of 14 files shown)


starboard/shared/starboard/player/filter/audio_frame_tracker.cc
```diff
@@ -75,26 +99,89 @@ int64_t AudioFrameTracker::GetFutureFramesPlayedAdjustedToPlaybackRate(
...
+
+  if (number_of_frames > 0) {
+    // The original frame position is out of audio record range. The video
+    // stream could be longer than audio stream. Use last playback rate to
+    // calculate adjusted frame position.
+    converted_position += number_of_frames / last_playback_rate_;
+  }
+
+  return converted_position;
+}
```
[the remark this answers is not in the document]
jasonzhangxx: Playback rate cannot be 0 in this class.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -143,21 +145,74 @@ void VideoRendererImpl::WriteSamples(const InputBuffers& input_buffers) {
...
+    if ((*it)->timestamp() > audio_write_head) {
+      first_input_after_audio_head_it = it;
+      break;
+    } else if (is_vsp_adjustment_enabled_) {
+      // Adjust timestamps of frames that will be sent to the decoder.
+      int64_t adjusted_timestamp =
+          media_time_provider_->AdjustTimestampToAudioClock((*it)->timestamp());
+      SB_CHECK_GE(adjusted_timestamp, 0)
+          << "Failed to adjust frame at " << (*it)->timestamp();
+      (*it)->SetTimestamp(adjusted_timestamp);
```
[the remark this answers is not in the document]
jasonzhangxx: If the timestamp is behind audio write ahead, it should not get any error.

starboard/shared/starboard/player/filter/audio_frame_tracker.h
```diff
@@ -24,6 +24,7 @@
 
 namespace starboard {
 
+// TODO (b/507074472): add tests for AudioFrameTracker
```
borongc: nit: TODO: b/507074472 - add tests for AudioFrameTracker

starboard/shared/starboard/player/filter/video_renderer_internal_impl.h
```diff
@@ -109,6 +110,13 @@ class VideoRendererImpl : public VideoRenderer, private JobQueue::JobOwner {
   std::atomic<int32_t> input_buffers_sent_{0};
   int64_t seeking_to_time_ = 0;  // microseconds
 
+  // When experiment |enable_video_renderer_vsp_adjustment| is enabled and
+  // playabck rate is not 0.0x or 1.0x. Video renderer vsp adjustment will be
```
borongc, line 114: nit: typo `playabck` should be `playabck`

starboard/shared/starboard/player/filter/audio_frame_tracker.cc
```diff
@@ -75,26 +99,86 @@ int64_t AudioFrameTracker::GetFutureFramesPlayedAdjustedToPlaybackRate(
...
+  int64_t converted_position = current_adjusted_frame_head_position_;
+  int64_t number_of_frames =
+      frame_position - current_original_frame_head_position_;
+  if (number_of_frames <= 0) {
+    // We don't have records for past audio time, which should rarely happen. We
+    // use the playback rate of first record or last playback rate to give an
+    // estimation.
+    if (!frame_records_.empty()) {
+      converted_position +=
+          number_of_frames / frame_records_.front().playback_rate;
```
borongc, line 132: Should we check `frame_records_.front().playback_rate` > 0?

starboard/shared/starboard/player/filter/audio_frame_tracker.cc
```diff
@@ -75,26 +99,86 @@ int64_t AudioFrameTracker::GetFutureFramesPlayedAdjustedToPlaybackRate(
...
     } else {
-      number_of_frames -= record.number_of_frames;
-      frames_played +=
-          static_cast<int>(record.number_of_frames * record.playback_rate);
+      converted_position += number_of_frames / last_playback_rate_;
```
borongc, line 134: Same for `last_playback_rate_`.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -143,21 +145,74 @@ void VideoRendererImpl::WriteSamples(const InputBuffers& input_buffers) {
...
+  if (!experimental_features_.enable_video_renderer_vsp_adjustment ||
+      (!is_vsp_adjustment_enabled_ && seeking_)) {
+    write_to_decoder(input_buffers);
+    return;
+  }
+
+  // When the |enable_video_renderer_vsp_adjustment| feature is enabled and
+  // after playback is prerolled, we hold video frames until the audio head
+  // reaches their presentation time to ensure a smooth playback rate change.
+  int64_t audio_write_head = media_time_provider_->GetAudioWriteHead();
```
borongc, line 164: Passthrough and video-only stream use `0` here?

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecFrameRateEstimator.java
```diff
@@ -99,25 +98,16 @@ public int getEstimatedFrameRate() {
...
-        for (int i = 1; i < mCount; i++) {
-          long delta = mSortedTimestampsUs[i] - mSortedTimestampsUs[i - 1];
-          if (mMinDeltaUs == INVALID_FRAME_TIMESTAMP || delta < mMinDeltaUs) {
-            mMinDeltaUs = delta;
-          }
-        }
-      }
-
       int validTimestamps = 1;
+      long previousDelta = mSortedTimestampsUs[1] - mSortedTimestampsUs[0];
```
borongc, line 102: Is this robust?

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -209,8 +266,19 @@ void VideoRendererImpl::SetPlaybackRate(double playback_rate) {
   SB_DCHECK(BelongsToCurrentThread());
   SB_DCHECK_GE(playback_rate, 0);
 
+  if (experimental_features_.enable_video_renderer_vsp_adjustment &&
+      playback_rate != 0.0 && playback_rate != 1.0) {
+    // Start vsp adjustment of video input presentation timepstamps in video
+    // renderer after we see a none 0.0x or 1.0x playback rate.
+    is_vsp_adjustment_enabled_ = true;
```
borongc, line 277: After this is enabled, is this always to be enabled?

starboard/shared/starboard/player/input_buffer_internal.h
```diff
@@ -66,6 +66,10 @@ class InputBuffer : public RefCountedThreadSafe<InputBuffer> {
     return has_drm_info_ ? &drm_info_ : NULL;
   }
   void SetDecryptedContent(std::vector<uint8_t> decrypted_content);
+  // Input buffers are passed one-way from the renderer to subcomponents.
+  // Although modifying timestamp directly is not ideal, it's still ok if
+  // modifications are in renderers.
+  void SetTimestamp(int64_t timestamp) { timestamp_ = timestamp; }
```
borongc, line 71: How can we ensure the only use case is in renderers?

starboard/shared/starboard/player/filter/audio_frame_tracker.cc
```diff
@@ -50,22 +54,42 @@ void AudioFrameTracker::AddFrames(int number_of_frames, double playback_rate) {
...
-          static_cast<int>(record.number_of_frames * record.playback_rate);
+      number_of_frames -= unplayed_frame;
+      // Note that since |current_original_frame_head_position_| is an integer,
+      // multiplying by a non-integer |playback_rate| can lead to truncation
+      // errors. These small errors accumulate every time a record is removed,
+      // potentially leading to a slight drift in the original timeline over
+      // long periods of playback. Remove record after it's fully played to
+      // minimize that drift.
+      current_original_frame_head_position_ +=
+          record.number_of_frames * record.playback_rate;
```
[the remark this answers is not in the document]
borongc: This seems to be a reasonable review.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecFrameRateEstimator.java
jasonzhangxx: According to Gemini, the largest distance for two continous frames (in display order) is determined by  "Mini-GOP" (the group of frames between your "anchor" frames). The maximum Mini-GOP size for h264, vp9 and av1 are 16, 16 and 32. So, after the estimator reaches 32 frames, the delta between the first two frames should accurately reflect the actual FPS.

I updated the check at line 106, it was too large.

starboard/shared/starboard/player/filter/audio_frame_tracker.cc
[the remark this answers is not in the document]
jasonzhangxx: It is mentioned in the comment above. It could have accumulated drift of 1 frame every time playback rate changes. Considering that the audio usually has 44100 sample per second, a small amount of accumulated drift should be ok.

starboard/shared/starboard/player/filter/audio_frame_tracker.cc
jasonzhangxx: It cannot be 0, there's a `SB_CHECK()` in `AddFrames`. Play audio with playback rate 0 doesn't make sense.

starboard/shared/starboard/player/filter/audio_frame_tracker.cc
jasonzhangxx: Same as the previous comment.

starboard/shared/starboard/player/filter/audio_frame_tracker.h
jasonzhangxx: Done

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: Yes, right now it's only used by tunnel mode, tunnel mode doesn't support passthrough and video-only.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: Yes, it's mentioned in the comment in header file.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.h
jasonzhangxx: Done.

starboard/shared/starboard/player/input_buffer_internal.h
jasonzhangxx: It's a tricky function like `SetDecryptedContent` above. It actually can be used anywhere after it's added. But making `InputBuffer` virtual might negatively impact performance, although subtly. Since VSP is not used in most playback scenarios, this approach minimizes the impact by allowing timestamp changes in a less intrusive way.

Another approach would be to make InputBuffer virtual and add a ProxyInputBuffer that accepts the timestamp during instantiation.

johnxwork:
Just supporting the PR from ATV owners. The PR only touches ATV media code.

johnxwork (approved):

starboard/android/shared/player_components_factory.cc
```diff
@@ -70,8 +67,11 @@ bool UseLibopusDecoder(SbMediaAudioCodec codec,
 // This class allows us to force int16 sample type when tunnel mode is enabled.
 class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
  public:
```
borongc, line 72: This needs to be rebased.

starboard/android/shared/player_components_factory.cc
```diff
@@ -390,41 +395,49 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
...
+          << "kPauseUsingAudioTrackState is set to true, force using "
+          << "AudioTrackState while pausing playback.";
+
       // TODO: b/500811542 - Connect to H5VCC.
       const bool allow_audio_writing_on_pause =
           experimental_features.allow_audio_writing_on_pause;
       SB_LOG_IF(INFO, allow_audio_writing_on_pause)
           << "allow_audio_writing_on_pause is set to true.";
 
+      const bool enable_video_renderer_vsp_adjustment =
```
borongc, line 434: If this is only used in tunnel mode, can we CHECK(force_tunnel_mode) here?

borongc: Or, if this feature depends on both `allow_audio_writing_on_pause` and `force_tunnel_mode`, we should set the experiment flag to be true if the above flags to be true.

starboard/android/shared/player_components_factory.cc
```diff
@@ -470,6 +485,13 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
           << "The maximum size in bytes of a buffer of data is "
           << max_video_input_size;
 
+      if (experimental_features.enable_video_renderer_vsp_adjustment &&
```
borongc, line 481: With the comment above, we don't need this.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -143,21 +145,74 @@ void VideoRendererImpl::WriteSamples(const InputBuffers& input_buffers) {
...
 
-  input_buffers_sent_.fetch_add(static_cast<int32_t>(input_buffers.size()));
-  decoder_->WriteInputBuffers(input_buffers);
+  auto write_to_decoder = [&](const InputBuffers& buffers) {
+    input_buffers_sent_.fetch_add(static_cast<int32_t>(buffers.size()));
+    decoder_->WriteInputBuffers(buffers);
+  };
+
+  if (!experimental_features_.enable_video_renderer_vsp_adjustment ||
+      (!is_vsp_adjustment_enabled_ && seeking_)) {
```
borongc, line 156: `is_vsp_adjustment_enabled_ = true` only if `experimental_features_.enable_video_renderer_vsp_adjustment = true`.

We can just simplify this to `if (!is_vsp_adjustment_enabled_ && seeking_)`.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -143,21 +145,74 @@ void VideoRendererImpl::WriteSamples(const InputBuffers& input_buffers) {
...
-  SB_LOG_IF(WARNING, end_of_stream_written_.load())
-      << "Try to write EOS after EOS is reached";
-  if (end_of_stream_written_.load()) {
+  if (end_of_stream_written_.load() || pending_end_of_stream_) {
+    SB_LOG(WARNING) << "Try to write EOS after EOS is reached";
     return;
   }
+
+  if (!pending_input_buffers_.empty()) {
+    pending_end_of_stream_ = true;
```
borongc, line 216: In which case, we need this flag `pending_end_of_stream_`?

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -209,8 +266,19 @@ void VideoRendererImpl::SetPlaybackRate(double playback_rate) {
...
+  if (experimental_features_.enable_video_renderer_vsp_adjustment &&
+      playback_rate != 0.0 && playback_rate != 1.0) {
+    // Start vsp adjustment of video input presentation timepstamps in video
+    // renderer after we see a none 0.0x or 1.0x playback rate.
+    is_vsp_adjustment_enabled_ = true;
+  }
+
   if (sink_) {
-    sink_->SetPlaybackRate(playback_rate);
+    if (is_vsp_adjustment_enabled_) {
```
borongc, line 281: Consider to add comment like "Force the playback to be 0.0x or 1.0x if `is_vsp_adjustment_enabled_` is enabled".

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -382,6 +380,48 @@ int64_t AudioRendererPcm::GetCurrentMediaTime(bool* is_playing,
...
+  }
+  SB_CHECK(decoder_sample_rate_);
+
+  int64_t frame_position =
+      (timestamp - seeking_to_time_) * *decoder_sample_rate_ / 1'000'000LL;
+  int64_t adjusted_frame_position =
+      audio_frame_tracker_.ConvertOriginalFramePositionToAdjustedFrames(
+          frame_position);
+  if (adjusted_frame_position < 0) {
+    // Failed to adjust frame position, return the error code.
```
borongc: nit: return the negative value as a error code.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -143,21 +145,74 @@ void VideoRendererImpl::WriteSamples(const InputBuffers& input_buffers) {
...
+  auto first_input_after_audio_head_it = input_buffers.end();
+  for (auto it = input_buffers.begin(); it != input_buffers.end(); ++it) {
+    if ((*it)->timestamp() > audio_write_head) {
+      first_input_after_audio_head_it = it;
+      break;
+    } else if (is_vsp_adjustment_enabled_) {
+      // Adjust timestamps of frames that will be sent to the decoder.
+      int64_t adjusted_timestamp =
+          media_time_provider_->AdjustTimestampToAudioClock((*it)->timestamp());
+      SB_CHECK_GE(adjusted_timestamp, 0)
```
borongc, line 174: We crash the app immediately if there is a negative timestamp, how likely this would happen?

borongc: Or, we can consider to send a pipeline error back to WebApp without crashing it.

starboard/android/shared/player_components_factory.cc
jasonzhangxx: Rebased.

starboard/android/shared/player_components_factory.cc
jasonzhangxx: Yes, we'll set `enable_video_renderer_vsp_adjustment` flag only when  `allow_audio_writing_on_pause` is true and only when tunnel mode is used. Reporting an error here so that if the server configuration is wrong, we can easily identify the problem. CHECK() will lead to app termination immediately, it's not user friendly.

starboard/android/shared/player_components_factory.cc
jasonzhangxx: The error reporting is to quickly locating the problem if that happens.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
jasonzhangxx: Done.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: No, there's a case that `experimental_features_.enable_video_renderer_vsp_adjustment: false, is_vsp_adjustment_enabled_: false, seeking_: false`.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: It should not happen as `GetAudioWriteHead` is checked above. The `SB_CHECK` is to expose problem immediately if that unexpected case occurs. `SB_DCHECK` is better, but we're now encouraged to use `SB_CHECK`.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: If a playback only have a few video sample (for example, one video frame), `WriteEndOfStream()` might be called very early, even before audio is written. When there're pending inputs, we cannot write eos before all inputs are written into the decode.

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: Done.

starboard/shared/starboard/player/input_buffer_internal.h
jasonzhangxx: I updated the comment to explicitly mentioned that the set functions should not be used outside drm system and renderers.

fayev-g (approved):
For tvos/. Thanks!

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
borongc: Can we add this as a comment?

borongc (approved):

starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
jasonzhangxx: Done.

PR #10251, opened 2026-04-27
Fix CobaltMetricsBrowserTest.RecordCpuMetrics

tasunnn:
The CobaltMetricsBrowserTest.RecordsCpuMetrics test was failing or flaky
due to blocking operations on a thread where blocking is disallowed.
Introduce base::ScopedAllowBlockingForTesting to permit these operations,
ensuring the test can execute reliably without failures.

Bug: 507063989


Awallky (approved):
LGTM

Thanks for the fix!

PR #10258, opened 2026-04-28
Sync larboard 1.0.2 changes 

santoshmahto7:


base/memory/ref_counted.h
```diff
@@ -7,6 +7,7 @@
 
 #include <stddef.h>
 
+#include <limits>
```
niranjanyardi: Do we need this ? other changes look okay. this file was being compiled before without any errors

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/platform_configuration/configuration.gni
```diff
@@ -56,5 +56,5 @@ size_config_path =
   "$rdk_starboard_root/third_party/starboard/rdk/shared/platform_configuration:size"
 
 if (sb_is_evergreen_compatible) {
-  sb_evergreen_compatible_use_libunwind = true
+  sb_evergreen_compatible_use_libunwind = false
```
niranjanyardi: @hlwarriner / @y4vor  is it okay to  disable this ?

jonastsai (approved):
This is a PR to sync the changes from RDK. Let's keep its original change if possible

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_path.cc
```diff
@@ -263,7 +231,53 @@ bool GetTemporaryDirectory(char* out_path, int path_size) {
 }
 }  // namespace
 
+namespace third_party {
+namespace starboard {
+namespace rdk {
+namespace shared {
+namespace system {
+// Gets the path to the content directory.
+bool GetContentDirectory(char* out_path, int path_size) {
```
jonastsai, line 240: I think we have modify this function before to meet Cobalt development need. We should review and keep the existing change.

jonastsai, line 240: I think we have modify this function before to meet Cobalt development need. We should review and keep the existing change.

jonastsai (changes requested):

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_path.cc
santoshmahto7: The function is only moved to a different namespace - no code has been changed. This is done since this function is now made available to other files also, by adding a declaration in the header file (system_get_path.h) with a proper namespace. before this function was private to this file only.
See the added header file: https://github.com/youtube/cobalt/pull/10258/changes#diff-1e993980b390a2e9cf19d5a6c3879ef7b5847d2a715c834abb0f8b6d1725871b

santoshmahto7: see above comments

santoshmahto7:
Since this is a merge, let's ignore the Gemini improvement suggestion, which can be handled separately in a different commit, so that we have fewer differences compared to larboard branch

niranjanyardi (approved):

PR #10265, opened 2026-04-28
Remove SbThreadSetPriority API and update usages

niranjanyardi:
This PR removes the SbThreadSetPriority API and updates usages to use SetCurrentThreadPriority unconditionally.


cobalt/site/docs/reference/starboard/modules/thread.md
```diff
@@ -238,7 +238,7 @@ the context returned from `SbThreadSamplerFreeze`.
 bool SbThreadSamplerThaw(SbThreadSampler sampler)
 ```
 
-### SbThreadSetPriority
+### SbThreadSetPriority (deprecated)
```
niranjanyardi, line 241: just delete it

PR #10278, opened 2026-04-29
android: Add Missing KeyEvent.ACTION_UP to VolumeStateReceiever

ckim1346:
This PR introduces a missing `KeyEvent.ACTION_UP` to VolumeStateReceiver. When the IME Adapter was only dispatching `KeyEvent.ACTION_DOWN`, YTS guided volume key tests were failing. This PR now allows the test to pass by passing the action up to the dispatch, allowing it to reach the JavaScript layer.

Bug: 452811427


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/VolumeStateReceiver.java
```diff
@@ -23,6 +27,52 @@ final class VolumeStateReceiver extends BroadcastReceiver {
...
+      + "      bubbles: true,"
+      + "      cancelable: true"
+      + "    });"
+      + "    Object.defineProperties(event, {"
+      + "      keyCode: { value: k, enumerable: true },"
+      + "      which: { value: k, enumerable: true }"
+      + "    });"
+      + "    target.dispatchEvent(event);"
+      + "  }"
+      + "})(%d, '%s');";
```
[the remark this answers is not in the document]
ckim1346: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/VolumeStateReceiver.java
```diff
@@ -23,6 +27,51 @@ final class VolumeStateReceiver extends BroadcastReceiver {
...
+        if (mapping.androidKeyCode == code) {
+          return mapping;
+        }
+      }
+      return null;
+    }
+  }
+
+  // Reusable parameterized JS injection script evaluated against DOM window targets
+  private static final String JS_EVENT_TEMPLATE =
```
zhongqiliang: I would avoid to call evaluateJavaScript() if possible. It is hard to find when it fails to run into javascript error. We need to measure how efficient is this compare with imeAdapter.dispatchKeyEvent(). 
Note, this button event is a continuous event, which means when user press the volume button, onReceive() will be triggered multiple times in a row. If evaluateJavaScript() is inefficient, it may impact the UI.

Can you find why imeAdapter.dispatchKeyEvent() does not work anymore?
Is there anything we can do with this new API? https://docs.google.com/document/d/1_hLZ96jaTrBxb5Wz-t6SZyLLYEIXQF7U-dudtsAnv54/edit?tab=t.0#heading=h.xzptrog8pyxf

johnxwork (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/VolumeStateReceiver.java
ckim1346: Chromium processes volume inputs as "system keys": the IME pipeline checks if there is an active, focused editable element (like an  <input>  or  contenteditable  tag) on the page (which the volume buttons don't have). Blink then swallows the event, preventing it from bubbling up to standard window or document DOM listeners in JavaScript. This was fine when we were using the custom starboard way of handling, but now we can't get it through.

I can try with the new API but looks like it isn't available yet -- b/398284486#comment9. Should we unblock this YTS test for now?

ckim1346: Gemini's thoughts on performance: 

### 1. Is the JS payload large?                                                                                                                
                                                                                                                                                 
  No, the payload is extremely small. It is just a few lines of JavaScript that construct and dispatch a  KeyboardEvent . The string size is negligible (less than 500 bytes), so transferring it from Java to C++ takes minimal time.                                                      
  ### 2. Could it impact UI or performance if held down?                                                                                         
                                                                                                                                                 
  Yes, it could, but it is unlikely to be noticeable. Here is why:                                                                               
                                                                                                                                                 
  • Context Switching Overhead: When you hold a volume button, Android generates repeated key events (typically at a rate of about 10 to 20 events per second). Calling  evaluateJavaScript  at this frequency requires moving from the Java UI thread, through the JNI bridge into Chromium's C++ message loop, and finally executing in the V8 JavaScript engine.                                                                
  • Comparison to Native Pipeline: This path is less efficient than Chromium's native input pipeline, which is highly optimized for high-frequency events (like touch moves).                                                                                                           
  • Likely Impact: For a frequency of 10-20 Hz (typical for key repeats), modern TV hardware should easily handle this without causing UI jank or blocking the main thread. It is far less demanding than processing continuous touch or mouse-move events at 60Hz or 120Hz.
  
If performance ever became an issue, we could optimize it in CobaltActivity.java by throttling the repeat events on the Java side before sending them to JavaScript.

zhongqiliang: "Blink then swallows the event, preventing it from bubbling up to standard window or document DOM listeners in JavaScript." Does it mean the regression is from Blink changes? Can you find where is the change? If there is a key mapping removed there, would it be better to add it back there for cobalt repo?

b/398284486#comment9 says the API will be available soon on Android 16. Please test it out when it is ready.

ckim1346: The regression wasn't due to changes to the dispatch -- in C25, we didn't use `imeAdapter.dispatchKeyEvent()`, we similarly did a custom cobalt injection into the cobalt application event queue (see C25 [VolumeStateReceiver.java](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/android/apk/app/src/main/java/dev/cobalt/coat/VolumeStateReceiver.java;drc=175c351adb738f2077a54587bc972879b17aaee6;l=41) and C25 [android_main.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/25.lts.1+:starboard/android/shared/android_main.cc;drc=74f8c87cb0991061d7f86247fe797010b1a15c30;l=402)). Someone updated the `nativeVolumeChanged()` and `nativeMuteChanged()` calls to use the IME dispatcher when we migrated to C26, but the chromium input stack blocks the volume keys from reaching the JS layer hence why volume key event YTS tests have been broken since C26.

zhongqiliang: Why does chromium block the volume keys? Any comment in the code tells why?
How difficult is it to remove the block?

ckim1346: Thanks Colin for pushing -- turns out we were missing the `KeyEvent.ACTION_UP` in the new VolumeStateReceiver, which is why we couldn't capture a signal for the YTS test. Updated the PR (now only a one-line fix!). Will have to keep looking into getting the other system-level key events in the Chrobalt application layer separately.

zhongqiliang (approved):

PR #10279, opened 2026-04-29
media: Refactor Video BufferClamp impl

johnedocampo:
This change refactors the implementation of the MSE video buffer size
clamping. Previously, an H5vcc setting was translated into a command
line switch, which was then read and propagated through several media
components.

The new approach removes the command line switch and directly applies
the H5VCC setting to a global variable in DemuxerMemoryLimitStarboard. We
also move the H5VCC initialization from cobalt_content_renderer_client.cc to
h_5_vcc_settings.cc.

Bug: 460460519


media/filters/source_buffer_state_unittest.cc
```diff
@@ -382,4 +382,23 @@ TEST_F(SourceBufferStateTest, TrackIdsSwappedInSecondInitSegment) {
...
 }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+TEST_F(SourceBufferStateTest, SetVideoBufferSizeClampMb) {
+  SourceBufferState::SetVideoBufferSizeClampMb(10);
+
+  std::unique_ptr<SourceBufferState> sbs =
+      CreateAndInitSourceBufferState("vp9");
+
+  std::unique_ptr<MediaTracks> tracks(new MediaTracks());
```
kjyoun: auto tracks = std::make_unique<MediaTracks>(); ?

media/filters/source_buffer_state_unittest.cc
```diff
@@ -382,4 +382,23 @@ TEST_F(SourceBufferStateTest, TrackIdsSwappedInSecondInitSegment) {
   AppendDataAndReportTracks(sbs, std::move(tracks2));
 }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+TEST_F(SourceBufferStateTest, SetVideoBufferSizeClampMb) {
```
kjyoun: How about this?

go/unit-testing-practices?polyglot=cpp#structure


TEST_F(SourceBufferStateTest, SetVideoBufferSizeClampMb) {
// Arrange
 std::unique_ptr<SourceBufferState> sbs =
      CreateAndInitSourceBufferState("vp9");
auto tracks..
AddVideoTrack(tracks, VideoCodec::kVP9, 1);
ASSERT_FOUND_CODEC_NAME(Video, "vp9");
 ASSERT_CALL(*this, MediaTracksUpdatedMock(_));


// Act
SourceBufferState::SetVideoBufferSizeClampMb(10);

// Assert
EXPECT_TRUE(AppendDataAndReportTracks(sbs, std::move(tracks)));
}

media/filters/source_buffer_state.cc
```diff
@@ -38,6 +38,7 @@ namespace {
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   std::atomic<int> g_max_pending_bytes_per_parse{StreamParser::kMaxPendingBytesPerParse};  // 128KiB
+  std::atomic<int> g_video_buffer_size_clamp_mb{std::numeric_limits<int>::max()};
```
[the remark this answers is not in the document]
johnedocampo: Done. Left a comment as to why we set this to 0.

media/filters/source_buffer_state.cc
```diff
@@ -824,15 +830,10 @@ void SourceBufferState::SetStreamMemoryLimits() {
...
-    MEDIA_LOG(INFO, media_log_)
-        << "Custom video per-track SourceBuffer size clamp limit(MiB)="
-        << (video_buf_size_limit_clamp / 1024 / 1024);
-    for (const auto& it : video_streams_)
-      it.second->SetStreamMemoryLimitClamp(video_buf_size_limit_clamp);
-  }
+ if (g_video_buffer_size_clamp_mb > 0) {
+  for (const auto& it : video_streams_)
+    it.second->ApplyStreamMemoryLimitCeiling(g_video_buffer_size_clamp_mb * 1024 * 1024);
+ }
```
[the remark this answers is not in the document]

media/filters/source_buffer_state_unittest.cc
```diff
@@ -382,4 +382,23 @@ TEST_F(SourceBufferStateTest, TrackIdsSwappedInSecondInitSegment) {
...
+      CreateAndInitSourceBufferState("vp9");
+
+  std::unique_ptr<MediaTracks> tracks(new MediaTracks());
+  AddVideoTrack(tracks, VideoCodec::kVP9, 1);
+
+  EXPECT_FOUND_CODEC_NAME(Video, "vp9");
+  EXPECT_CALL(*this, MediaTracksUpdatedMock(_));
+  EXPECT_TRUE(AppendDataAndReportTracks(sbs, std::move(tracks)));
+  
+  // Reset for other tests just in case.
```
kjyoun: for this cleanup operation,

how about moving it to TearDown?

media/filters/source_buffer_state.cc
```diff
@@ -824,15 +831,11 @@ void SourceBufferState::SetStreamMemoryLimits() {
...
- size_t video_buf_size_limit_clamp =
-      GetMSEBufferSizeLimitIfExists(switches::kMSEVideoBufferSizeLimitClampMb);
-  if (video_buf_size_limit_clamp) {
-    MEDIA_LOG(INFO, media_log_)
-        << "Custom video per-track SourceBuffer size clamp limit(MiB)="
-        << (video_buf_size_limit_clamp / 1024 / 1024);
-    for (const auto& it : video_streams_)
-      it.second->SetStreamMemoryLimitClamp(video_buf_size_limit_clamp);
-  }
+ if (g_video_buffer_size_clamp_mb > 0) {
```
borongc: should this begin with two spaces?

media/filters/source_buffer_state.cc
```diff
@@ -824,15 +831,11 @@ void SourceBufferState::SetStreamMemoryLimits() {
...
+   for (const auto& it : video_streams_)
```
borongc: Same for this code block, seems to have wrong indentation.

media/filters/source_buffer_state_unittest.cc
```diff
@@ -382,4 +382,23 @@ TEST_F(SourceBufferStateTest, TrackIdsSwappedInSecondInitSegment) {
...
+  std::unique_ptr<SourceBufferState> sbs =
+      CreateAndInitSourceBufferState("vp9");
+
+  std::unique_ptr<MediaTracks> tracks(new MediaTracks());
+  AddVideoTrack(tracks, VideoCodec::kVP9, 1);
+
+  EXPECT_FOUND_CODEC_NAME(Video, "vp9");
+  EXPECT_CALL(*this, MediaTracksUpdatedMock(_));
+  EXPECT_TRUE(AppendDataAndReportTracks(sbs, std::move(tracks)));
+
```
borongc: Extra two spaces.

As this is Chromium's code, our `pre-commit` doesn't apply to this file.

media/filters/source_buffer_state_unittest.cc
```diff
@@ -382,4 +382,23 @@ TEST_F(SourceBufferStateTest, TrackIdsSwappedInSecondInitSegment) {
   AppendDataAndReportTracks(sbs, std::move(tracks2));
 }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
borongc: For our unit tests (or added unit tests), we should consider to use an another file (like `//media/filters/starboard/source_buffer_state_unittest.cc`), so we don't add extra complexity to maintain both Chromium code and unittest code when upstream.

Also, if this test is needed in general, you can upstream first. If this is Cobalt-specific, then it's more ideal to put it to another single, new, Cobalt-owned file (like `//media/filters/starboard/source_buffer_state_unittest.cc`).

media/filters/source_buffer_stream.h
```diff
@@ -186,9 +186,10 @@ class MEDIA_EXPORT SourceBufferStream {
   }
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  void set_memory_limit_clamp(size_t memory_limit_clamp) {
-    memory_limit_clamp_ = memory_limit_clamp;
-    memory_limit_ = std::min(memory_limit_, memory_limit_clamp);
+// Will set |memory_limit_ceiling_| and will also clamp |memory_limit_| if memory_limit_ceiling_ < |memory_limit_|.
```
borongc: Two spaces are needed.

johnedocampo: Done.

media/filters/source_buffer_state_unittest.cc
johnedocampo: Thank you for the suggestion; Created //media/filters/starboard/source_buffer_state_unittest.cc.

This is a Cobalt-Specific test to help test the new setter method added for H5VCC experimentation. I've copied over the necessary components from the chromium tests to run this test for Cobalt.

media/filters/source_buffer_state_unittest.cc
johnedocampo: Done.

media/filters/source_buffer_state.cc
johnedocampo: Fixed; updated indentation.

media/filters/source_buffer_state.cc
johnedocampo: Fixed.

media/filters/source_buffer_state_unittest.cc
johnedocampo: Done; this code was originally copying how Chromium made their tests, but since this has been split to media/filters/starboard/starboard_source_buffer_state_unittest.cc, updated the code to the suggestion.

media/filters/source_buffer_state_unittest.cc
johnedocampo: Done.

media/filters/source_buffer_state_unittest.cc
johnedocampo: Done.

media/filters/source_buffer_state.cc
```diff
@@ -824,14 +831,15 @@ void SourceBufferState::SetStreamMemoryLimits() {
   }
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
- size_t video_buf_size_limit_clamp =
-      GetMSEBufferSizeLimitIfExists(switches::kMSEVideoBufferSizeLimitClampMb);
-  if (video_buf_size_limit_clamp) {
+  if (g_video_buffer_size_clamp_mb > 0) {
+    size_t ceiling_in_bytes =
+        static_cast<size_t>(g_video_buffer_size_clamp_mb) * 1024 * 1024;
```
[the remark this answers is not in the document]
johnedocampo: This shouldn't be an issue; we do not expect this value to go that high.

media/filters/starboard/starboard_source_buffer_state_unittest.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "media/filters/frame_processor.h"
+#include "testing/gmock/include/gmock/gmock.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace media {
+using testing::_;
+using testing::DoAll;
+using testing::InvokeWithoutArgs;
+using testing::Return;
+
```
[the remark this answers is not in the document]
johnedocampo: This macro comes from "media/base/mock_media_log.h".

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -277,6 +236,10 @@ ExperimentalFeatures ProcessH5vccSettings(
           settings, kH5vccSettingsKeyMediaUseDualThreadsForVideo)) {
     parsed.use_dual_threads_for_video = *val != 0;
   }
+  if (auto* val = GetSettingValue<int64_t>(
+          settings, kH5vccSettingsKeyMediaVideoBufferSizeClampMb)) {
+    ::media::SourceBufferState::SetVideoBufferSizeClampMb(*val);
```
kjyoun: can you move this operation to h_5_vcc_settings.cc ?

This place seems to fit better there. It already sets SourceBufferState. while this method is just for basically parsing and setting `parsed`

https://source.corp.google.com/h/github/youtube/cobalt/+/main:third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc;l=164;drc=413a7ea996f629f97c25eb13dad07c76d0b2d353
```
 if (name == "Media.ExperimentalMaxPendingBytesPerParse") {
    return ProcessSettingAsPositiveInt(
        script_state, exception_context, name, *value, [](int int_value) {
          ::media::SourceBufferState::SetMaxPendingBytesPerParseOverride(
              int_value);
          return true;
#else   // BUILDFLAG(USE_STARBOARD_MEDIA)
          return false;
        });
  }
```

johnedocampo: That's my bad, thanks for the catch. Done.

media/filters/starboard/starboard_source_buffer_state_unittest.cc
```diff
@@ -0,0 +1,153 @@
...
+#include "media/base/mock_filters.h"
+#include "media/base/mock_media_log.h"
+#include "media/base/stream_parser.h"
+#include "media/filters/chunk_demuxer.h"
+#include "media/filters/frame_processor.h"
+#include "testing/gmock/include/gmock/gmock.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace media {
+using testing::_;
```
kjyoun: nit: move using to line 42 inside anonymous namespace


https://screenshot.googleplex.com/7mwxezKJTR9AzcF

media/filters/starboard/starboard_source_buffer_state_unittest.cc
```diff
@@ -0,0 +1,153 @@
...
+}  // namespace
+
+class StarboardSourceBufferStateTest : public ::testing::Test {
+ public:
+  StarboardSourceBufferStateTest() : mock_stream_parser_(nullptr) {}
+
+  void TearDown() override { SourceBufferState::SetVideoBufferSizeClampMb(0); }
+
+  std::unique_ptr<SourceBufferState> CreateSourceBufferState() {
+    std::unique_ptr<FrameProcessor> frame_processor =
```
kjyoun: nit: auto 

you (or Gemini) already used make_unique :)

media/filters/starboard/starboard_source_buffer_state_unittest.cc
```diff
@@ -0,0 +1,153 @@
...
+        std::make_unique<FrameProcessor>(base::DoNothing(), &media_log_);
+    mock_stream_parser_ = new testing::StrictMock<MockStreamParser>();
+    return base::WrapUnique(new SourceBufferState(
```
kjyoun: nit: why not make_unique ?

media/filters/source_buffer_stream.cc
```diff
@@ -1865,24 +1865,16 @@ bool SourceBufferStream::UpdateVideoConfig(const VideoDecoderConfig& config,
     // Dynamically increase |memory_limit_| on video config changes.
     size_t new_memory_limit =
         GetDemuxerStreamVideoMemoryLimit(DemuxerType::kChunkDemuxer, &config);
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+    new_memory_limit = std::min(new_memory_limit, memory_limit_ceiling_);
```
kjyoun: why not we just override memory_limit?

```
new_memory_limit = std::min(new_memory_limit, memory_limit_ceiling_);
```

With this "ceiling", it is more complicated to tell what is the actual limit applied.

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -63,14 +67,28 @@ size_t GetDemuxerStreamAudioMemoryLimit(
...
-                                         video_config->visible_rect().size(),
-                                         GetBitsPerPixel(*video_config));
+  size_t ceiling = g_video_buffer_size_clamp_bytes.load();
+  if (ceiling > 0) {
+    limit = std::min(limit, ceiling);
+  }
+  return limit;
+}
+
+void SetVideoBufferSizeClamp(int size_in_mb) {
```
kjyoun: `size_mb` ?

Per Gemini
----

Both are perfectly acceptable, but **`size_mb`** is generally the better choice for standard C++ development. 

Here is why, along with a few nuances to help you decide.

### Why `size_mb` Wins
* **Concision:** C++ code can get verbose quickly. Dropping the `"_in"` keeps the variable name short and sweet without losing any meaning. Every C++ developer reading `size_mb` will instantly know it means "size in megabytes."
* **Standard Library Consistency:** The C++ Standard Library (STL) and popular libraries lean toward minimalism. For example, `std::chrono` uses suffixes like `ms` (milliseconds) or `ns` (nanoseconds), not `_in_ms`. 

---

### When to Consider `size_in_mb`
* **Strict Style Guides:** Some team style guides strictly mandate conversational variable names to completely eliminate ambiguity for junior developers. 
* **Complex Conversions:** If your codebase is doing heavy math converting between bytes, kilobytes, and megabytes in the same function, sometimes the explicit `_in_` can act as a mental speed bump to prevent logic errors.

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -63,14 +67,28 @@ size_t GetDemuxerStreamAudioMemoryLimit(
...
+  } else {
+    limit = GetVideoDecoderBufferLimitBytes(video_config->codec(),
+                                            video_config->visible_rect().size(),
+                                            GetBitsPerPixel(*video_config));
   }
 
-  return GetVideoDecoderBufferLimitBytes(video_config->codec(),
-                                         video_config->visible_rect().size(),
-                                         GetBitsPerPixel(*video_config));
+  size_t ceiling = g_video_buffer_size_clamp_bytes.load();
```
kjyoun: if `g_video_buffer_size_clamp_bytes` is set to infinite by default. 

Logic becomes simpler.

````
// Initialization
std::atomic<size_t> g_video_buffer_size_clamp_bytes{std::numeric_limits<size_t>::max()};

// Usage inside your function:
size_t GetDemuxerStreamVideoMemoryLimit(...) {
  size_t limit = /* ... calculation ... */;

  // No 'if' statement required! If clamp isn't set, std::min returns limit.
  return std::min(limit, g_video_buffer_size_clamp_bytes.load());
}
```

kjyoun (approved):

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Done; the clamp is now set to infinite by default.

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Done; changed to size_mb

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -55,6 +60,12 @@ int GetBitsPerPixel(const VideoDecoderConfig& video_config) {
 
 }  // namespace
 
+void SetVideoBufferSizeClamp(int size_mb) {
```
borongc, line 69: Add a CHECK to ensure size_mb > 0.

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -28,6 +30,9 @@
 namespace media {
 namespace {
 
+std::atomic<size_t> g_video_buffer_size_clamp_bytes{
```
borongc, line 35: consider adding a comment that this is process-wide, and current is set by h5vcc flag.

media/filters/source_buffer_stream.cc
```diff
@@ -1872,17 +1872,6 @@ bool SourceBufferStream::UpdateVideoConfig(const VideoDecoderConfig& config,
     }
   }
 
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
borongc, line 1875: Note that with this refactor, if `memory_limit_overridden_ == true`, the update logic won't work.

We should consider sanity check `memory_limit_overridden_` is not true, that is, the experiment doesn't work with `kMSEAudioBufferSizeLimitMb` and/or `kMSEVideoBufferSizeLimitMb` enabled.

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Done.

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Done.

media/filters/source_buffer_stream.cc
johnedocampo: Done. For the sanity check, I added a check that will log a WARNING if `memory_limit_overridden_` and if the experiment is active. This warning is only logged when `kMSEVideoBufferSizeLimitMb` is enabled, as this check is only performed in `SourceBufferStream::UpdateVideoConfig`.

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -55,6 +62,17 @@ int GetBitsPerPixel(const VideoDecoderConfig& video_config) {
...
 
+size_t GetVideoBufferSizeClamp() {
+  return g_video_buffer_size_clamp_bytes.load();
+}
+
+void SetVideoBufferSizeClamp(int size_mb) {
+  // We convert the value from MBs to bytes, as the values returned by
+  // GetVideoDecoderBufferLimitBytes's return value is in bytes.
+  CHECK(size_mb > 0);
+  g_video_buffer_size_clamp_bytes = static_cast<size_t>(size_mb) * 1024 * 1024;
```
Awallky, line 75: Multiplying  `size_mb`  by  1024 * 1024  can overflow the 32-bit  `size_t`.
      • For example, setting a clamp of  4096  (4GB) wraps to  0 , which clamps the video buffer limit to  0  and will break playback.
      • Recommendation: Enforce a hard upper limit (e.g.,  `CHECK_LT(size_mb, 4096)` ) or use  `base::CheckedNumeric`.

Awallky (approved):
//cobalt LGTM
Added a comment for one potential bug.

media/filters/source_buffer_stream.cc
borongc: Just want to say we don't have a way to verify if the experiment is active, `memory_limit_overridden_` should not be true.

This will be different, so it's not really a simple refactor for this PR.

borongc: I don't want to block this PR, but just want to callout if memory_limit_ is somehow overridden for the active experiment, there is no way we can debug it (although it's unlikely).

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Done; added `CHECK_LT(size_mb, 4096)`.

andrewsavage1 (approved):
web api changes LGTM

media/filters/source_buffer_stream.cc
johnedocampo: This is a valid issue. I can merge this PR in and maybe we can supply a follow-up PR to disable memory_limit_overridden_ if the experiment is enabled or if the experiment becomes enabled.

PR #10285, opened 2026-04-30
tvOS: Report compressed memory as swap on tvOS

Gyuyoung:
tvOS does not provide a direct API to measure swap memory usage. Instead,
it utilizes memory compression under pressure. This change updates
MeasureUsedSwapMemory to return the process's compressed memory bytes
as a proxy for swap memory on tvOS. This provides a more meaningful
metric than always reporting zero, addressing a previous TODO.

Bug: 497682329

### DIFF
--- a/cobalt/browser/performance/performance_impl.cc
+++ b/cobalt/browser/performance/performance_impl.cc
@@ -61,16 +61,19 @@ void PerformanceImpl::MeasureUsedCpuMemory(
 
 void PerformanceImpl::MeasureUsedSwapMemory(
     MeasureUsedSwapMemoryCallback callback) {
-#if BUILDFLAG(IS_IOS_TVOS)
-  // TODO: b/497682329 - vm_swap_bytes does not exist on tvOS.
-  std::move(callback).Run(0);
-#else
   auto process_metrics = base::ProcessMetrics::CreateProcessMetrics(
       base::GetCurrentProcessHandle());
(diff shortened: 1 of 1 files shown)


Gyuyoung:
I'm not sure we can implement MeasureUsedSwapMemory() using `compressed_byte`, since tvOS doesn't support retrieving swap memory. If you guys think using `compressed_byte` doesn't make sense, I will close this PR. wdyt?

rakuco (commented):
I haven't had time to review this in depth yet, my concern so far is that reporting compressed memory as swap might be misleading, not sure if this should be reported with a different name altogether (and if Kabuki's willing to adjust as well).

fayev-g:
> I haven't had time to review this in depth yet, my concern so far is that reporting compressed memory as swap might be misleading, not sure if this should be reported with a different name altogether (and if Kabuki's willing to adjust as well).

Yeah, I don't think we should report something inconsistent/different here. Kabuki specifically relies on consistent telemetry when it comes to these things, and it wouldn't be obvious from the API that tvOS reports something slightly different. (I know similar data pollution/API changes is a huge no-no between cobalt versions for Kabuki. This compressed memory seems to have never been reported in prev cobalt versions though so I think we have the opportunity to set precedent.)

I seems like the recommended way to handle a no-op for a platform would be to update the mojom API to be able to return an optional nullable object. `MeasureUsedSwapMemory() => (uint64? bytes);` And then return null.

PR #10287, opened 2026-04-30
android: Enable PauseUsingAudioTrackState by default

borongc:
Enable the PauseUsingAudioTrackState feature by default to prevent
dropped frames when resuming playback after a pause.

Previously, SbPlayer stopped updating audio timestamps when the player
was paused. This led to media time drifts because Android's AudioTrack
may still present audio data committed before the pause. These drifts
caused dropped video frames upon resumption as the player attempted to
resync.

Continuous polling of audio timestamps during pause ensures accurate
media time is maintained. Field experiments confirmed significant
improvements in QoE metrics with negligible CPU impact (~1-2%).

Issue: 349854301

### DIFF
--- a/starboard/android/shared/audio_track_audio_sink_type.cc
+++ b/starboard/android/shared/audio_track_audio_sink_type.cc
@@ -134,7 +134,6 @@ AudioTrackAudioSink::AudioTrackAudioSink(
     int tunnel_mode_audio_session_id,
     bool is_web_audio,
     bool allow_audio_writing_on_pause,
-    bool pause_using_audio_track_state,
     void* context)
     : type_(type),
       channels_(channels),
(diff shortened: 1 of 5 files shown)


starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -509,12 +489,11 @@ SbAudioSink AudioTrackAudioSinkType::Create(
   const int kTunnelModeAudioSessionId = -1;
   const bool kIsWebAudio = true;
   const bool kAllowAudioWritingOnPause = false;
-  const bool kPauseUsingAudioTrackState = false;
```
osagie98, line 512: Previously, `pause_using_audio_track_state_` was false for web audio. Now that the param is removed, it looks like web audio will also now have this new behavior, if I'm reading the rest of the code correctly, is that intended?

borongc: It is, the functionality of `pause_using_audio_track_state` is working in prod. It is low-risk to enable it in WebAudio as well.

Also, this feature is a low-risk change as well.

osagie98 (approved):

PR #10288, opened 2026-04-30
Android: Update Platform Error to append retry param and add tests

vincentmo:
android tv: Track network dialog retry attempts

The RaisePlatformError mechanism was updated to propagate the
navigation URL, allowing PlatformError to use it for retries, increasing the url param coverage.

The 'netdialog_retry' URL parameter for Android TV's Network Dialog
was previously hardcoded to '1'. This change introduces a monotonically increasing counter to track
retry attempts throughout the app instance's lifecycle. The
addRetryUrlParam logic now correctly appends or updates this dynamic
parameter, handling existing query parameters and URL fragments.

New Robolectric tests for PlatformError dialog.

Bug: 496219065

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
@@ -27,11 +27,14 @@ import android.os.Looper;
 import android.provider.Settings;
 import android.view.KeyEvent;
 import androidx.annotation.IntDef;
+import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
 import dev.cobalt.shell.StartupGuard;
 import dev.cobalt.util.Holder;
 import dev.cobalt.util.Log;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.chromium.content_public.browser.WebContents;
 import org.jni_zero.NativeMethods;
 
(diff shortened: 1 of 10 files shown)


starboard/android/shared/system_platform_error.cc
```diff
@@ -63,7 +63,7 @@ bool SbSystemRaisePlatformError(SbSystemPlatformErrorType type,
                : nullptr;
 
   starboard::StarboardBridge::GetInstance()->RaisePlatformError(
-      env, jni_error_type, reinterpret_cast<jlong>(send_response_callback));
+      env, jni_error_type, reinterpret_cast<jlong>(send_response_callback), "");
```
zhongqiliang, line 66: @y4vor how will SbSystemRaisePlatformError() be used? Will the request url be passed into user_data?

zhongqiliang (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -67,20 +69,26 @@ public class PlatformError
   private static final int DISMISS_BUTTON = 3;
 
   private static final String RETRY_PARAM_KEY = "netdialog_retry";
-  private static final String RETRY_PARAM_VALUE = "1";
+  private static final AtomicInteger sRetryCount = new AtomicInteger(0);
```
Awallky, line 73: Since this is static, it means this value will persist across different error events and different instances of `PlatformError`. Should we reset this at any point in time or make this non-static?

vincentmo: The lifecycle of the retry counter is tied to the app instance and should reset only when it's restarted so it's ok to be static and doesn't need to reset.

starboard/android/shared/system_platform_error.cc
vincentmo: AFAICT from inspecting the source, this is implemented but never actually called.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -217,18 +232,55 @@ interface Natives {
...
+    if (!retryParamAdded) {
+      uriBuilder.appendQueryParameter(RETRY_PARAM_KEY, String.valueOf(count));
+    }
+
+    String result = uriBuilder.build().toString();
+    Log.i(TAG, "Reloading URL with retry param: " + result);
+    return result;
+  }
+
+  @androidx.annotation.VisibleForTesting
```
Awallky: To keep styling consistent, such as CobaltActivity#appendArgsFromMetaData, you can import this at the top and simplify here to `@VisibleForTesting`.

Awallky (approved):
LGTM % comment

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
vincentmo: Done, thanks.

PR #10290, opened 2026-04-30
Fix for rdk deploy script

jellefoks:
This is a very important PR for our team

It especially makes sure that b/9999999 is addressed.

Bug: 123456789
Issue: 9876543212

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/arm/scripts/deploy_rdk.py
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/arm/scripts/deploy_rdk.py
@@ -155,9 +155,9 @@ def package_and_deploy(
     archive_name = "archive.tar.gz"
 
     if mode == "plugin":
-        tar_cmd = ["tar", "-czf", archive_name, "-C", str(out_dir), "-T", deps_file.name, "libloader_app.so"]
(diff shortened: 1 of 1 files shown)


starboard/contrib/rdk/src/third_party/starboard/rdk/arm/scripts/deploy_rdk.py
```diff
@@ -155,9 +155,9 @@ def package_and_deploy(
     archive_name = "archive.tar.gz"
 
     if mode == "plugin":
-        tar_cmd = ["tar", "-czf", archive_name, "-C", str(out_dir), "-T", deps_file.name, "libloader_app.so"]
```
niranjanyardi: instead of deps_file.name  - i verified that deps_file works - can you please that change here and below

niranjanyardi (approved):

starboard/contrib/rdk/src/third_party/starboard/rdk/arm/scripts/deploy_rdk.py
andrewsavage1: Seems like we could just use https://github.com/youtube/cobalt/pull/10300/changes?

PR #10302, opened 2026-04-30
Add browsertest to monitor memory footprint in CI

tasunnn:
This PR introduces a new browser performance test `CobaltMemoryThresholdBrowserTest`
to collect the following metrics: 
- private memory footprint 
- peak RSS
- V8 heap size
- peak GPU memory usage during page load 

This will be the first part in effort to incorporate browser performance test for 
memory monitoring in GitHub CI. It adds two tests, which loads a blank 
page and a mock YT homepage, to gather baseline memory usage for an initial
assessment of the test's functionality on CI. Currently, the test will only 
gather metrics and upload results to gcs.

mock YT homepage looks like this:
mmock_yt

Bug: 501202622


cobalt/testing/browser_tests/BUILD.gn
```diff
@@ -219,6 +219,7 @@ test("cobalt_browsertests") {
     "//cobalt/browser/h5vcc_runtime_browsertest.cc",
     "//cobalt/browser/h5vcc_system_browsertest.cc",
     "//cobalt/browser/metrics/cobalt_metrics_browsertest.cc",
+    "//cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc",
```
Awallky, line 223: I would prefer we separate this into its own test target, since this is not a browser test. This is a performance test.

tasunnn: that's a good point, just added a new target for it, thanks

tasunnn: if we separate them, also need to add this new target to CI as well

Awallky: Yes definitely. However, we still do not have the necessary infra to process and detect regressions. So, we need infra changes regardless. I also think that adding another test target should be straightfoward, since we already run browsertests in CI. 

If we have everything ready to run in infra, and this is blocking, we can alternatively leave a TODO to refactor this to its own perf test target.

tasunnn: yes just discussed with Xinan offline, I will follow up with another PR to separate the target

cobalt/testing/browser_tests/data/mock_homepage_tv.html
```diff
@@ -0,0 +1,122 @@
...
+        <!-- Cards will be inserted here -->
+      </div>
+    </div>
+  </div>
+
+  <script>
+    // Simulate large DOM with multiple images
+    const grid = document.getElementById('video-grid');
+    const images = [
+      'mock_thumbnail.png',
```
Awallky: I think these are actually JPEGs. Can you rename them to match their format?

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,159 @@
...
+
+IN_PROC_BROWSER_TEST_P(CobaltMemoryThresholdBrowserTest,
+                       BlankPageMemoryFootprint) {
+  base::HistogramTester histogram_tester;
+
+  // Navigate to a blank page.
+  EXPECT_TRUE(
+      content::NavigateToURL(shell()->web_contents(), GURL("about:blank")));
+
+  // Wait for a few seconds for metrics to be recorded.
```
Awallky: We shouldn't do this to wait for metrics to be recorded. Putting sleeps to wait for things can lead to flakes. We should instead, block on a callback to be invoked from our memory metrics emitter - cobalt/browser/metrics/cobalt_memory_metrics_emitter.h - set_callback_for_testing.

This callback is invoked when we record our metrics. If we block on this callback invocation, we don't need the sleep/timed wait here.

Awallky (changes requested):
Left a comment about how to wait for metrics changes before running our "tests". 
This mostly looks good!

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: The metrics in this test are not coming from CobaltMemoryMetricsEmitter, using set_callback_for_testing won't guarantee recording has finished when it's unblocked? Currently, waiting for global memory dump is done with a callback in the test, and the sleep here is only for "Memory.GPU.PeakMemoryUsage2.PageLoad" to log but this is triggered by page load not by CobaltMemoryMetricsEmitter

cobalt/testing/browser_tests/data/mock_homepage_tv.html
tasunnn: good catch, will rename it

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
linxinan-yt: > but this is triggered by page load

Whenever possible we should avoid hard coded sleep second in browser tests. Note your test may not run only on Linux host. It could be a low profile ATV or EG devices. 
If triggered by page load, you may consider WaitForLoadStop() or WaitForNavigationFinished(). See content/public/test/browser_test_utils.cc.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,159 @@
...
+namespace cobalt {
+
+class CobaltMemoryThresholdBrowserTest
+    : public content::ContentBrowserTest,
+      public testing::WithParamInterface<int> {
+ public:
+  CobaltMemoryThresholdBrowserTest() = default;
+  ~CobaltMemoryThresholdBrowserTest() override = default;
+
+  void OnMemoryDumpReceived(
```
linxinan-yt, line 46: If you create a new test target for memory or performance, I would recommend you to move this helpers to another file with headers. So that others can reuse them in their tests.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,159 @@
...
+  base::RunLoop run_loop_delay;
+  base::SingleThreadTaskRunner::GetCurrentDefault()->PostDelayedTask(
+      FROM_HERE, run_loop_delay.QuitClosure(), base::Seconds(5));
+  run_loop_delay.Run();
+
+  CollectMemoryMetrics(histogram_tester, "BlankPage");
+}
+
+IN_PROC_BROWSER_TEST_P(CobaltMemoryThresholdBrowserTest,
+                       YouTubeHomepageMemoryFootprint) {
```
linxinan-yt: You are using Youtube page. Suggest MockYoutubeHomePageMemoryFootprint.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,159 @@
...
+  // Wait for a few seconds for the page to load and stabilize.
+  base::RunLoop run_loop_delay;
+  base::SingleThreadTaskRunner::GetCurrentDefault()->PostDelayedTask(
+      FROM_HERE, run_loop_delay.QuitClosure(), base::Seconds(5));
+  run_loop_delay.Run();
+
+  CollectMemoryMetrics(histogram_tester, "YouTubeHomepage");
+}
+
+// Instantiate suite to repeat test run for five times.
```
linxinan-yt: This comment does not provide business value, as the below code has stated the same thing. Instead you should explain why you need this test to run 5 times at the binary level.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: I think [NavigateToURL()](https://source.corp.google.com/h/github/youtube/cobalt/+/main:content/public/test/browser_test_utils.cc;l=611?q=content%2Fpublic%2Ftest%2Fbrowser_test_utils.cc&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) already blocks until it finishes loading. But this only ensures page load finishes and we still can't guarantee that the UMA finishes recording.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,159 @@
...
+                           true);
+    perf_test::PrintResult("V8HeapSize", "", scenario_name,
+                           static_cast<size_t>(v8_heap_size_kb), "kb", true);
+  }
+
+  void CollectMemoryMetrics(base::HistogramTester& histogram_tester,
+                            const std::string& scenario_name) {
+    base::RunLoop run_loop;
+
+    // Gather metrics from memory dump.
```
linxinan-yt: So we only need to collect v8's memory dump?

tasunnn: for allocator yes, only v8 right now will be collected. But we will always have memory metrics from os dump by default, the first argument passed in only specifies allocator dumps to collect

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: done, thanks

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
linxinan-yt: You can explain this in the comment. Rather than explain the code :)

tasunnn: done thanks

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,159 @@
...
+    uint64_t peak_resident_set_kb = 0;
+    uint64_t v8_heap_size_kb = 0;
+
+    for (const auto& process_dump : dump->process_dumps()) {
+      // OS memory dump
+      private_footprint_kb += process_dump.os_dump().private_footprint_kb;
+      peak_resident_set_kb += process_dump.os_dump().peak_resident_set_kb;
+
+      // Allocator dump
+      std::optional<uint64_t> v8_size = process_dump.GetMetric("v8", "size");
```
sherryzy: Please see my comment in the design doc. I don't believe we need to include v8_size specifically in these memory tests. While V8 size is useful for detecting memory leaks on Kabuki, the value for the V8 heap largely depends on the JavaScript content. Since we are focusing on detecting memory regressions for Cobalt changes in CI, including the V8 heap size will not be very useful and may even cause confusion.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
Awallky: If you set up the tests to exercise cobalt's memory metrics emitter, you can then block on the callback, which ensures you can collect the histogram data. You would want to reduce the time it emits to 1 second.

And then you can extract the data using a HistogramTester e.g.:

```
  base::HistogramTester histogram_tester;

  // Perform actions in the browser that are expected to record a histogram.
  // For example, navigate to a page or interact with a feature.
  // e.g., NavigateToURL(browser(), test_url);

  LOG(INFO) << "Histogram 'HistogramName' samples:";
  for (const auto& bucket : samples) {
    LOG(INFO) << "  Sum: " << histogram_tester.GetTotalSum("HistogramName") << ", Count: " << histogram_tester.GetBucketCount("HistogramName", 10);
    // You can now use bucket.min and bucket.count for checks or further logic
    // Example: EXPECT_TRUE(bucket.min == 10 || bucket.min == 20 || bucket.min == 30);
  }
```

You can emit the same information we use in production with the same mechanism we use to collect it within Cobalt. You can also add some expectations that the histogram data is non-zero.

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: deleted it, thanks

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: yeah maybe will defer this to later when separating the targets

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: the Memory.GPU.PeakMemoryUsage2.PageLoad UMA is not recorded by the memory emitter so I don't think using the emitter's callback could correctly block for the metrics we are collecting. But I found a [ScopedHistogramSampleObserver](https://source.chromium.org/chromium/chromium/src/+/main:base/metrics/statistics_recorder.h;drc=df329a4bab484c65c08a05de98e8c61fbf220404;l=93) in chromium and looks like we could use this

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,152 @@
...
+            {},
+            base::BindOnce(
+                &CobaltMemoryThresholdBrowserTest::OnMemoryDumpReceived,
+                base::Unretained(this), run_loop.QuitClosure(), scenario_name));
+    run_loop.Run();
+
+    // Verify and log GPU Peak Memory from UMA.
+    WaitForHistogram(histogram_tester, "Memory.GPU.PeakMemoryUsage2.PageLoad");
+    std::vector<base::Bucket> buckets =
+        histogram_tester.GetAllSamples("Memory.GPU.PeakMemoryUsage2.PageLoad");
```
Awallky, line 109: Can we add an assertion here that the buckets size is > 0?

Awallky (approved):
LGTM % comment

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
tasunnn: done

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,153 @@
...
+    perf_test::PrintResult("PeakResidentSet", "", scenario_name,
+                           static_cast<size_t>(peak_resident_set_kb), "kb",
+                           true);
+  }
+
+  void WaitForHistogram(base::HistogramTester& histogram_tester,
+                        const std::string& histogram_name) {
+    base::RunLoop run_loop;
+    // Listen using a histogram observer to avoid flakiness in test
+    auto histogram_observer = std::make_unique<
```
linxinan-yt, line 82: Awesome, you found it!

cobalt/browser/metrics/cobalt_memory_threshold_browsertest.cc
```diff
@@ -0,0 +1,137 @@
...
+  void OnMemoryDumpReceived(
+      base::OnceClosure quit,
+      const std::string& scenario_name,
+      bool success,
+      std::unique_ptr<memory_instrumentation::GlobalMemoryDump> dump) {
+    // ScopedClosureRunner guarantees quit.Run() is called when this function ends,
+    // no matter how it exits (failure, etc.)
+    base::ScopedClosureRunner quit_runner(std::move(quit)); 
+    ASSERT_TRUE(success);
+    ASSERT_TRUE(dump);
```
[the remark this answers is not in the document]
linxinan-yt: I think this suggestions make sense. If a developer received this failure, they will just see some crash message(EXPECT() will provide the expected value though).

It is helpful to give them some hints what is going wrong.

linxinan-yt (approved):
LGTM % comment

yell0wd0g (commented):
Drive-by: Love the test (and the thumbnails :) 
Thanks for this!

PR #10312, opened 2026-05-02
android: Enable Flush AudioTrack during Seek

borongc:
Previously, seeking operations would cause the Android AudioTrack to be
destroyed and recreated. This can lead to audio glitches or delays
during playback.

This change introduces a Starboard extension to allow platforms to
implement an audio sink flush mechanism. For Android, this is
implemented to flush the underlying AudioTrack during a seek operation
rather than stopping and recreating it.

This new behavior is enabled via the
kForceFlushAudioTrackDuringReset feature flag, providing a smoother
and more efficient audio reset during seeks. The internal audio
renderer logic is also updated to correctly handle the sink's state
when it is flushed but not stopped.

Issue: 330793785

### DIFF
--- a/starboard/android/shared/audio_track_audio_sink_type.cc
+++ b/starboard/android/shared/audio_track_audio_sink_type.cc
@@ -189,7 +189,7 @@ AudioTrackAudioSink::AudioTrackAudioSink(
   SB_DCHECK(frame_buffer_);
   SB_CHECK(bridge_);
 
-  SB_LOG(INFO) << "Creating audio sink starts at " << start_time_;
+  SB_LOG(INFO) << "Creating audio sink starts at " << start_time_.load();
 }
 
 AudioTrackAudioSink::~AudioTrackAudioSink() {
@@ -227,6 +227,11 @@ int AudioTrackAudioSink::GetStartThresholdInFrames() {
   return bridge_->GetStartThresholdInFrames();
 }
 
+bool AudioTrackAudioSink::Flush() {
+  flush_requested_ = true;
+  return true;
+}
+
 // TODO: Break down the function into manageable pieces.
 void AudioTrackAudioSink::AudioThreadFunc() {
   JNIEnv* env = AttachCurrentThread();
@@ -244,6 +249,19 @@ void AudioTrackAudioSink::AudioThreadFunc() {
       features::kReleaseVideoFramesAfterAudioStarts);
 
   while (!quit_) {
+    if (flush_requested_) {
+      bridge_->PauseAndFlush();
+      frames_in_audio_track = 0;
+      accumulated_written_frames = 0;
+      last_playback_head_event_at = -1;
+      // Set to -1 to ignore the first playback head position read after a
+      // flush. This avoids using a stale value from before the reset, which
+      // could lead to incorrect frame consumption calculations.
+      last_playback_head_position = -1;
+      flush_requested_ = false;
+      continue;
+    }
+
     int playback_head_position = 0;
     int64_t frames_consumed_at = 0;
     if (bridge_->GetAndResetHasAudioDeviceChanged(env)) {
@@ -268,10 +286,16 @@ void AudioTrackAudioSink::AudioThreadFunc() {
     if (should_update_media_time) {
       playback_head_position =
           bridge_->GetAudioTimestamp(&frames_consumed_at, env);
-      SB_DCHECK_GE(playback_head_position, last_playback_head_position);
 
-      int frames_consumed =
-          playback_head_position - last_playback_head_position;
+      int frames_consumed = 0;
+      if (last_playback_head_position == -1) {
+        last_playback_head_position = playback_head_position;
+      } else {
+        frames_consumed =
+            std::max(0, static_cast<int>(playback_head_position -
+                                         last_playback_head_position));
+      }
+
       int64_t now = CurrentMonotonicTime();
 
       if (last_playback_head_event_at == -1) {
@@ -367,8 +391,8 @@ void AudioTrackAudioSink::AudioThreadFunc() {
         std::vector<uint8_t> silence_buffer(channels_ *
                                             GetBytesPerSample(sample_type_) *
                                             silence_frames_per_append);
-        int64_t sync_time =
-            start_time_ + GetFramesDurationUs(accumulated_written_frames);
+        int64_t sync_time = start_time_.load() +
+                            GetFramesDurationUs(accumulated_written_frames);
         // Not necessary to handle error of WriteData(), as the audio has
         // reached the end of stream.
         WriteData(env, silence_buffer.data(), silence_frames_per_append,
@@ -380,7 +404,7 @@ void AudioTrackAudioSink::AudioThreadFunc() {
     }
     SB_DCHECK_GT(expected_written_frames, 0);
     int64_t sync_time =
-        start_time_ + GetFramesDurationUs(accumulated_written_frames);
+        start_time_.load() + GetFramesDurationUs(accumulated_written_frames);
     SB_DCHECK_LE(start_position + expected_written_frames, frames_per_channel_)
         << "start_position: " << start_position
         << ", expected_written_frames: " << expected_written_frames
--- a/starboard/android/shared/audio_track_audio_sink_type.h
+++ b/starboard/android/shared/audio_track_audio_sink_type.h
@@ -151,6 +151,8 @@ class AudioTrackAudioSink : public SbAudioSinkImpl {
   void SetVolume(double volume) override;
   int GetUnderrunCount();
   int GetStartThresholdInFrames();
+  bool Flush();
+  void SetStartTime(int64_t start_time) { start_time_.store(start_time); }
 
  private:
   class AudioTrackOutThread;
@@ -171,7 +173,7 @@ class AudioTrackAudioSink : public SbAudioSinkImpl {
   const raw_ptr<void> frame_buffer_;
   const int frames_per_channel_;
   const AudioTrackAudioSinkType::Callbacks callbacks_;
-  const int64_t start_time_;  // microseconds
+  std::atomic<int64_t> start_time_;  // microseconds
   const int max_frames_per_request_;
   const raw_ptr<void> context_;
 
@@ -181,6 +183,7 @@ class AudioTrackAudioSink : public SbAudioSinkImpl {
   const std::unique_ptr<AudioTrackBridge> bridge_;
 
   volatile bool quit_ = false;
+  std::atomic_bool flush_requested_{false};
   // Guaranteed to be non-null.
   const std::unique_ptr<Thread> audio_out_thread_;
 
--- a/starboard/android/shared/player_components_factory.cc
+++ b/starboard/android/shared/player_components_factory.cc
@@ -73,7 +73,8 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
   explicit AudioRendererSinkAndroid(
       int tunnel_mode_audio_session_id = -1,
       bool allow_audio_writing_on_pause = false,
-      bool enable_video_renderer_vsp_adjustment = false)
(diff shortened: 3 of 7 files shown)


starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -248,34 +250,39 @@ void AudioRendererPcm::Seek(int64_t seek_to_time) {
...
-  if (resampler_) {
-    resampler_.reset();
-    time_stretcher_.FlushBuffers();
-  }
+    // If the sink is flushed but not stopped, the callbacks might still be
+    // called by the audio thread, so the following modifications must be
+    // protected by the lock.
+    if (resampler_) {
+      if (sink_stopped) {
+        resampler_.reset();
```
sideb0ard: If we don't reset the resampler_, won't it still have some audio data left which will cause a glitch?

starboard/android/shared/audio_sink_flush.cc
```diff
@@ -0,0 +1,43 @@
...
+#include "starboard/extension/audio_sink_flush.h"
+
+namespace starboard {
+namespace {
+
+bool AudioSinkFlush(SbAudioSink audio_sink) {
+  if (!SbAudioSinkIsValid(audio_sink)) {
+    return false;
+  }
+  auto* track_sink = static_cast<AudioTrackAudioSink*>(audio_sink);
```
sideb0ard: Would it be worth adding a CHECK here to ensure its an AudioTrackAudioSink?

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
borongc: Done. I reset `resampler_` as it may have some audio data left.

starboard/android/shared/audio_sink_flush.cc
borongc: Added.

sideb0ard (approved):

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -622,7 +624,11 @@ void AudioRendererPcm::ProcessAudioData() {
     scoped_refptr<DecodedAudio> decoded_audio =
         decoder_->Read(&decoded_audio_sample_rate);
     SB_DCHECK(decoded_audio);
-    if (!audio_renderer_sink_->HasStarted()) {
+    // `decoder_sample_rate_` is reset to nullopt during initialization and
+    // Seek. We use it instead of `audio_renderer_sink_->HasStarted()` to
+    // trigger `OnFirstOutput` because the sink might still be running if it was
+    // flushed (rather than stopped) during a seek.
+    if (!decoder_sample_rate_) {
```
jasonzhangxx: Let's still use `HasStarted()` here. But `AudioRendererSinkAndroid::HasStarted()` needs to be override to ensure that `HasStarted()` returns false before `Start()` is called.

starboard/android/shared/player_components_factory.cc
```diff
@@ -71,7 +71,8 @@ bool UseLibopusDecoder(SbMediaAudioCodec codec,
 class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
  public:
   explicit AudioRendererSinkAndroid(int tunnel_mode_audio_session_id = -1,
-                                    bool allow_audio_writing_on_pause = false)
+                                    bool allow_audio_writing_on_pause = false,
+                                    bool allow_flush_during_seek = false)
```
jasonzhangxx: nit: rebase will has conflict.

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -469,6 +479,11 @@ int AudioTrackAudioSink::GetStartThresholdInFrames() {
   return bridge_->GetStartThresholdInFrames();
 }
 
+bool AudioTrackAudioSink::Flush() {
```
jasonzhangxx: nit: `AudioTrackAudioSink::Flush()` can't fail, so it's not necessary to return a bool value. But it's ok to keep it as is.

starboard/android/shared/player_components_factory.cc
```diff
@@ -150,6 +152,26 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
...
+  void Start(int64_t media_start_time,
+             int channels,
+             int sampling_frequency_hz,
+             SbMediaAudioSampleType audio_sample_type,
+             SbMediaAudioFrameStorageType audio_frame_storage_type,
+             SbAudioSinkFrameBuffers frame_buffers,
+             int frames_per_channel,
+             RenderCallback* render_callback) override {
+    if (allow_flush_during_seek_ && audio_sink() &&
+        audio_sink()->IsType(SbAudioSinkImpl::GetPreferredType())) {
```
jasonzhangxx: We can add a function for this check as it's used in Start() and Reset(), and need to be exactly same.

starboard/android/shared/player_components_factory.cc
```diff
@@ -150,6 +152,26 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
...
+      SB_LOG(INFO) << "Audio track is already started, skipping Start().";
```
jasonzhangxx: Add a comment to mention that if the audio format could change and what we're going to do if it changes.

starboard/shared/starboard/player/filter/audio_renderer_sink_impl.h
```diff
@@ -50,6 +50,22 @@ class AudioRendererSinkImpl : public AudioRendererSink {
                               int* max_cached_frames,
                               int* min_frames_per_append) const override;
 
+ protected:
+  SbAudioSinkPrivate* audio_sink() const { return audio_sink_; }
```
jasonzhangxx: We can consider to make `audio_sink_` and `render_callback_` protected?

starboard/android/shared/player_components_factory.cc
```diff
@@ -150,6 +152,26 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
...
+      set_render_callback(render_callback);
```
jasonzhangxx: "media_start_time" needs update too.

starboard/android/shared/player_components_factory.cc
```diff
@@ -150,6 +152,26 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
...
+      return;
```
jasonzhangxx, line 192: It should be ok to use same `tunnel_mode_audio_session_id`, but please help verify if tunnel mode can still work.

starboard/android/shared/player_components_factory.cc
```diff
@@ -162,7 +184,21 @@ class AudioRendererSinkAndroid : public AudioRendererSinkImpl {
     return SbAudioSinkIsAudioSampleTypeSupported(audio_sample_type);
   }
 
+  void Reset() override {
+    if (allow_flush_during_seek_ && audio_sink() &&
+        audio_sink()->IsType(SbAudioSinkImpl::GetPreferredType())) {
+      auto* track_sink = static_cast<AudioTrackAudioSink*>(audio_sink());
+      if (track_sink->Flush()) {
+        SB_LOG(INFO) << "Flushing AudioTrack.";
```
jasonzhangxx, line 223: Do we need to reset playback rate here?

starboard/android/shared/audio_track_audio_sink_type.cc
```diff
@@ -236,6 +236,16 @@ void AudioTrackAudioSink::AudioThreadFunc() {
       features::kReleaseVideoFramesAfterAudioStarts);
 
   while (!quit_) {
+    if (flush_requested_) {
+      bridge_->PauseAndFlush();
```
jasonzhangxx, line 253: `audio_track_play_state` should be updated after audio sink is paused.

starboard/android/shared/player_components_factory.cc
borongc: Done.

starboard/shared/starboard/player/filter/audio_renderer_sink_impl.h
borongc: Done.

starboard/android/shared/player_components_factory.cc
borongc: We don't, `playback_rate_` is managed by `AudioRendererSinkImpl`, and it resumes with the previously set playback rate and volume.

starboard/android/shared/player_components_factory.cc
borongc: Done.

starboard/android/shared/player_components_factory.cc
borongc: Done.

starboard/android/shared/player_components_factory.cc
borongc: Done.

starboard/android/shared/audio_track_audio_sink_type.cc
borongc: Acked

starboard/android/shared/player_components_factory.cc
borongc: Verified. It works with tunnel mode.

starboard/android/shared/audio_track_audio_sink_type.cc
borongc: It doesn't. It will be updated with the next for loop iteration.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
borongc: It's more complicated here, as `HasStarted()` is used to many places, but we only need OnFirstWrite to be called after Seek().

starboard/android/shared/audio_track_audio_sink_type.h
```diff
@@ -151,6 +151,8 @@ class AudioTrackAudioSink : public SbAudioSinkImpl {
   void SetVolume(double volume) override;
   int GetUnderrunCount();
   int GetStartThresholdInFrames();
+  bool Flush();
+  void SetStartTime(int64_t start_time) { start_time_.store(start_time); }
```
jasonzhangxx, line 155: nit: Maybe `Restart()` is a better name?

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
jasonzhangxx: After `audio_renderer_sink_->Stop()` and before `audio_renderer_sink_->Start()` is called, `audio_renderer_sink_->HasStarted()` used to always return false. With the new change, `audio_renderer_sink_->HasStarted()`  would return true before `audio_renderer_sink_->Start()` is called, which might change some `AudioRendererPcm` behavior. Making `AudioRendererSinkAndroid::HasStarted()` returns false before `audio_renderer_sink_->Start()`  should have lower risk.

borongc: Done. The issue was after `AudioRendererSinkAndroid::Start()`, if flush is enabled, we should call:

```
      track_sink->SetPlaybackRate(playback_rate_);
      track_sink->SetVolume(volume_);
```
so AudioTrack can start (as the original Start() is ignored).

starboard/android/shared/audio_track_audio_sink_type.h
borongc: Keep it as Flush() as when it is called, it does Flush() instead of Reset() (which may be either flush or destroy the AudioTrackAudioSink).

jasonzhangxx (approved):

PR #10314, opened 2026-05-04
Fix javascript cache handling

y4vor:
cobalt: Retain compiled JS despite rejected cache

The V8 JavaScript compiler incorrectly discarded a successfully compiled
script if its associated cache entry was marked as 'rejected'. This led
to unnecessary recompilation of valid JavaScript code.

This change ensures that if the script compilation process
succeeds, the compiled script is always returned. If the original
cache entry was marked as rejected, it is now explicitly deleted
from the cache to prevent future issues with that specific entry.
This improves performance by reusing valid compiled scripts.

Issue: 506177754


lamdbui (approved):
Thanks, @y4vor . This seems like a good one to get in.

yuying-y (approved):

linxinan-chops:
Per the PR author, the CI error is unrelated. Merged the PR.

PR #10315, opened 2026-05-04
aosp: Break starboard -> base dep cycle

hferreiro:
On AOSP, the Starboard library is built from //starboard/android/shared, which depends on //base. //base in turn depends on //starboard via the is_starboard branch in copied_base/base/BUILD.gn, closing a dependency cycle that GN refuses during 'gn gen'.

Temporarily restrict the Starboard dependency to :starboard_headers_only for AOSP.

Bug: 495203133

### DIFF
--- a/copied_base/base/BUILD.gn
+++ b/copied_base/base/BUILD.gn
@@ -1765,6 +1765,14 @@ component("base") {
       "//starboard:starboard_group",
       "//starboard/common",
     ]
+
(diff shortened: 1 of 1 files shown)


hferreiro:
@andrewsavage1 perhaps this is worth a separate issue. I used the main one for the AOSP platform.

andrewsavage1 (approved):

PR #10317, opened 2026-05-04
android: Outbound deeplinks for YouTube and Netflix

ckim1346:
Integrate Chromium's external intent and navigation interception components to support outbound deep linking on Android. This allows Cobalt to launch external applications (e.g., YouTube TV, Netflix) when a user clicks on an app-specific URI (e.g., intent://, nflx://).

New components `InterceptNavigationDelegateClientImpl` and `ExternalNavigationDelegateImpl` are added to handle the interception and launching of these external URLs. The Android manifest is updated with package visibility queries to ensure Cobalt can discover and launch target applications on Android 11+.

Package names have to be added explicitly for intent launches (ie `com.google.android.youtube.tv`). Currently verified to support YouTube and Netflix, testing being done on other common packages (ie Spotify, Disney+, Hulu, and YouTubeTV/YouTube Music, etc.)

Bug: 406300169


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/InterceptNavigationDelegateClientImpl.java
```diff
@@ -0,0 +1,121 @@
...
+import org.chromium.content_public.browser.WebContentsObserver;
+
+/**
+ * Class that provides embedder-level information to InterceptNavigationDelegateImpl based off a
+ * Tab.
+ */
+public class InterceptNavigationDelegateClientImpl implements InterceptNavigationDelegateClient {
+    private WebContents mWebContents;
+    private RedirectHandler mRedirectHandler;
+    private WebContentsObserver mWebContentsObserver;
```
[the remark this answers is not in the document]
ckim1346: Removed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/InterceptNavigationDelegateClientImpl.java
```diff
@@ -0,0 +1,121 @@
...
+        mInterceptNavigationDelegate.associateWithWebContents(mWebContents);
+    }
+
+    public void destroy() {
+        if (mWebContentsObserver != null) {
+            mWebContentsObserver.observe(null);
+        }
+        mInterceptNavigationDelegate.associateWithWebContents(null);
+        mInterceptNavigationDelegate = null;
+    }
```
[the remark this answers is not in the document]
ckim1346: Removed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -690,7 +699,8 @@ protected boolean isDevelopmentBuild() {
 
   @Override
   protected void onNewIntent(Intent intent) {
-    getStarboardBridge().handleDeepLink(getIntentUrlAsString(intent));
+    String deepLink = getIntentUrlAsString(intent);
+    getStarboardBridge().handleDeepLink(deepLink);
```
zhongqiliang: This change is unnecessary.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+            if (context.getPackageManager().resolveActivity(intent, 0) != null) {
+                try {
+                    context.startActivity(intent);
+                    return true; // Handled
+                } catch (Exception e) {
+                    Log.e(TAG, "Failed to launch intent: " + intent, e);
+                }
+            }
+
+            // If all launch attempts failed, use the fallback URL (only for intent:// URIs).
```
zhongqiliang, line 98: any doc for the fallback logic?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    @Override
+    public boolean shouldDisableExternalIntentRequestsForUrl(GURL url) {
+        final String scheme = url.getScheme();
+
+        // Let the browser handle standard web schemes.
+        if (scheme.equals("http") || scheme.equals("https")) {
+            // Special case for development package to prevent escaping to YouTube TV app
+            if ("dev.cobalt.coat".equals(mContext.getPackageName())) {
```
zhongqiliang: This name is for CoAT, won 't work for Kimono

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+    public boolean canLoadUrlInCurrentTab() {
+        return true;
+    }
+
+    @Override
+    public void closeTab() {}
+
+    @Override
+    public boolean isIncognito() {
+        return mWebContents.isIncognito();
```
zhongqiliang: do we return true or false here?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+        return mWebContents;
+    }
+
+    @Override
+    public boolean hasValidTab() {
+        return true;
+    }
+
+    @Override
+    public boolean canCloseTabOnIncognitoIntentLaunch() {
```
zhongqiliang, line 208: What will this do? return true, does it close the "tab" that hosts cobalt?

cobalt/android/BUILD.gn
```diff
@@ -205,6 +209,7 @@ android_assets("coat_assets") {
 # TODO(b/450024477): just a shared library once jni migration is complete
 shared_library_with_jni("libchrobalt") {
   remove_uncalled_jni = true
+  add_stubs_for_missing_jni = true
```
zhongqiliang, line 212: why this? which one is missing?

cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2
```diff
@@ -144,4 +144,18 @@
     </application>
     {% block extra_root_definitions %}
     {% endblock %}
+    <queries>
```
zhongqiliang, line 148: how do you come up with the list? Is there a requirement from the product team?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
zhongqiliang: add unit test for this file

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Since Cobalt (to my knowledge, correct me if I'm wrong) doesn't support incognito mode, this would return false. `mWebContents.isIncognito()` should return whether or not the tab was incognito when it was created

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Only if [ExternalNavigationDelegateImpl.closeTab()](https://source.corp.google.com/h/github/youtube/cobalt/+/main:components/external_intents/android/java/src/org/chromium/components/external_intents/ExternalNavigationDelegate.java;drc=524126904340bf2104b6615d6bb6352ba6d7d844;l=87) is implemented (it's currently empty). But I agree that this should be `false` instead -- updated the code.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: Agreed, reverted.

cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2
ckim1346: No, I'm just testing out different applications locally to see if they would work. Do you have any concerns/changes in mind?

cobalt/android/BUILD.gn
ckim1346: Missing five according to the compiler, here's what it returned:

```
Failed JNI assertion!
We reference Java files which use JNI, but our native library does not depend on
the corresponding generate_jni().
To bypass this check, add stubs to Java with --add-stubs-for-missing-jni.
Excess Java files:
../../chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeFeatureMap.java
../../chrome/browser/flags/android/java/src/org/chromium/chrome/browser/flags/ChromeSessionState.java
../../components/browser_ui/notifications/android/java/src/org/chromium/components/browser_ui/notifications/NotificationFeatureMap.java
../../components/messages/android/java/src/org/chromium/components/messages/MessageFeatureMap.java
../../components/messages/android/java/src/org/chromium/components/messages/MessageWrapper.java
```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: [Android Intents with Chrome](https://developer.chrome.com/docs/android/intents) explains the fallback logic. I can also write a short design doc for the implementation for Chrobalt if that's more helpful

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Yes, I'm trying to prevent Kimono from also launching when we startup coAT. Without the check, `com.google.android.youtube.tv` also launches right after launching `dev.cobalt.coat`

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
yell0wd0g: nit:2026, here and in the next Java file.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+import org.chromium.base.Callback;
+import org.chromium.base.supplier.Supplier;
+import org.chromium.components.external_intents.ExternalNavigationDelegate;
+import org.chromium.content_public.browser.Visibility;
+import org.chromium.content_public.browser.WebContents;
+import org.chromium.ui.base.WindowAndroid;
+import org.chromium.url.GURL;
+
+/**
+ * Cobalt's implementation of the {@link ExternalNavigationDelegate}.
```
yell0wd0g: I guess this comment is a truism. What I'd like to see is
what's the reason for this class existence, the "why", and
the lifetime and threading models, if that makes sense.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+    @Override
+    public boolean willAppHandleIntent(Intent intent) {
+        return mContext.getPackageManager().resolveActivity(intent, 0) != null;
+    }
+
+    /**
+     * Handles a URL that is not a standard web URL (http or https).
+     * This can be called from the Activity on startup or from the navigation throttle.
+     * @return {@code true} if the URL was handled.
+     */
```
yell0wd0g, line 74: If the logic in this function resembles or is inspired
by some in Clank or Webview, consider leaving a link
or at least a reference to it by path.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,236 @@
...
+        final String scheme = url.getScheme();
+
+        // Let the browser handle standard web schemes.
+        if (scheme.equals("http") || scheme.equals("https")) {
+            // Special case for development package to prevent escaping to YouTube TV app
+            if ("dev.cobalt.coat".equals(mContext.getPackageName())) {
+                if (url.getHost().contains("youtube.com")) {
+                    return true; // Disable external intent requests, keep in Cobalt
+                }
+            }
```
yell0wd0g: nit: Instead of explaining the code, just give names, e.g.

```
   bool is_development_package = "dev.cobalt.coat".equals(mContext.getPackageName());
   bool is_this_or_that = url.getHost().contains("youtube.com");
   return is_development_package && is_this_or_that;
```

OTOH, I'm not sure I understand the logic, we don't want to
handle a YTTV intent in development builds?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/InterceptNavigationDelegateClientImpl.java
```diff
@@ -0,0 +1,119 @@
...
+import org.chromium.components.external_intents.InterceptNavigationDelegateClient;
+import org.chromium.components.external_intents.InterceptNavigationDelegateImpl;
+import org.chromium.components.external_intents.RedirectHandler;
+import org.chromium.content_public.browser.LoadUrlParams;
+import org.chromium.content_public.browser.WebContents;
+import org.chromium.content_public.browser.WebContentsObserver;
+
+/**
+ * Class that provides embedder-level information to InterceptNavigationDelegateImpl based off a
+ * Tab.
```
yell0wd0g: nit: We don't have Tabs (the class Tab), right? Clank does but we don't.

yell0wd0g (approved):
+1 to unit tests, here in this CL or in a later one.
If in a later one, please leave a `//TODO` somewhere
for your reviewers, to avoid confusion.

Code changes LGTM with some comments, leaving a +1
but let's wait for Colin to approve it too. Thanks!

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/InterceptNavigationDelegateClientImpl.java
ckim1346: I believe you're right -- updated to reflect that ours is based off of `WebContents` and `CobaltActivity`

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: On cobalt app start, without this logic, we immediately also follow through with a Kimono intent which also launches Kimono on top of Cobalt. The code still allows other YouTube intents in other formats (ie `intent://VI0YaqL68Ik/#Intent;scheme=vnd.youtube;package=com.google.android.youtube.tv;S.browser_fallback_url=market://details?id=com.google.android.youtube.tv;end` will still launch Kimono). Updated the names/comments but there might indeed be a better way to handle -- startup does tend to be tricky.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Updated

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: sure thing, updated the class description 👍

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Updated

ckim1346: added

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
zhongqiliang: Is it tested locally? 
That when it redirect to the other app, it closed the YouTube app, with this change, it did not close?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
zhongqiliang: verified locally? 
why not hardcode the return value?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
zhongqiliang: This is good enough.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
zhongqiliang: I do not see you update this code. 
Do you really need to check the package name here?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Yes, it just opens another app. So if it went to Netflix for example, then I close Netflix, you'll see YouTube app still open.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: I did verify locally, but I can hardcode a return value of `false` since Cobalt shouldn't be incognito.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Updated to check against the startup URL and start Deeplink instead

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,241 @@
...
+    public boolean shouldDisableExternalIntentRequestsForUrl(GURL url) {
+        final String scheme = url.getScheme();
+
+        // Let the browser handle standard web schemes.
+        if (scheme.equals("http") || scheme.equals("https")) {
+            // Special case for development package to prevent escaping to YouTube TV app
+            if ("dev.cobalt.coat".equals(mContext.getPackageName())) {
+                if (url.getHost().contains("youtube.com")) {
+                    return true; // Disable external intent requests, keep in Cobalt
+                }
```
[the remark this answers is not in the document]
ckim1346: Checking against startup URL and deeplink instead of hardcoded values instead

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Colin suggested checking the requested intent/URL/deeplink against the startup URL/Deeplink instead. Updated to reflect that

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
ckim1346: Updated to return false

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+
+            // If all launch attempts failed, use the fallback URL (only for intent:// URIs).
+            if (urlSpec.startsWith("intent:")) {
+                String fallbackUrl = intent.getStringExtra("browser_fallback_url");
+                if (fallbackUrl != null && !fallbackUrl.isEmpty()) {
+                    Intent fallbackIntent = new Intent(Intent.ACTION_VIEW, Uri.parse(fallbackUrl));
+                    fallbackIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                    try {
+                        context.startActivity(fallbackIntent);
+                    } catch (android.content.ActivityNotFoundException e) {
```
zhongqiliang, line 106: import this exception to make it shorter

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+                intent = new Intent(Intent.ACTION_VIEW, Uri.parse(urlSpec));
+            }
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+            // Try to launch the intent as-is (with package if specified).
+            if (context.getPackageManager().resolveActivity(intent, 0) != null) {
+                try {
+                    context.startActivity(intent);
+                    return true; // Handled
+                } catch (Exception e) {
```
zhongqiliang, line 93: the code pattern is different, you used ActivityNotFoundException in the code below.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+                    fallbackIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+                    try {
+                        context.startActivity(fallbackIntent);
+                    } catch (android.content.ActivityNotFoundException e) {
+                        Log.e(TAG, "Could not launch fallback URL: " + fallbackUrl, e);
+                    }
+                    return true; // Handled
+                }
+            }
+        } catch (java.net.URISyntaxException e) {
```
zhongqiliang, line 112: import the class

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/InterceptNavigationDelegateClientImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Override
+    public boolean wasTabLaunchedFromLongPressInBackground() {
+        return false;
+    }
+
+    @Override
+    public void closeTab() {
+      // TODO
```
zhongqiliang, line 93: What is this for?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    @Override
+    public boolean shouldLaunchWebApksOnInitialIntent() {
+        return false;
+    }
+
+    @Override
+    public void setPackageForTrustedCallingApp(Intent intent) {
+        assert false;
```
zhongqiliang, line 224: When will this method be called? this line would crash the app.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+ * generally tied to the lifetime of the active {@link WebContents} or the hosting {@link android.app.Activity}.
+ *
+ * Threading model: All methods in this class must be called on the main/UI thread.
+ */
+public class ExternalNavigationDelegateImpl implements ExternalNavigationDelegate {
+    private WebContents mWebContents;
+    private Context mContext;
+
+    public ExternalNavigationDelegateImpl(WebContents webContents, Context context) {
+        assert webContents != null;
```
zhongqiliang, line 49: remove this line, annotate WebContents with androidx.annotation.NonNull;

cobalt/android/BUILD.gn
zhongqiliang: This is probably not right. Why does it need to reference code under chrome/ ?
How many more components are pulled in for this feature?

cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2
zhongqiliang: The config is hardcoded in the APK, means it won't change until we update it and deploy a new version to the device.

If this feature is not required by the product team, I 'd not to submit the code until they ask for it.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+                if ((startupUrl != null && urlSpec.equals(startupUrl)) ||
+                    (startDeepLink != null && !startDeepLink.isEmpty() && urlSpec.equals(startDeepLink))) {
+                    return true;
+                }
+            }
+            return false;
+        }
+
+        // For any other scheme, we will try to handle it ourselves and stop the browser from trying.
+        return handleExternalURL(mContext, url);
```
zhongqiliang, line 143: There is a code structure problem. This override method shouldDisableExternalIntentRequestsForUrl() is supposed to return a true/false value. 

But inside of handleExternalURL(), it launches another activity from it. This code is not correct.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/ExternalNavigationDelegateImpl.java
```diff
@@ -0,0 +1,258 @@
...
+            } else {
+                // This is a simple URI scheme (e.g., "nflx://").
+                intent = new Intent(Intent.ACTION_VIEW, Uri.parse(urlSpec));
+            }
+            intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+
+            // Try to launch the intent as-is (with package if specified).
+            if (context.getPackageManager().resolveActivity(intent, 0) != null) {
+                try {
+                    context.startActivity(intent);
```
zhongqiliang, line 91: We need to find a security engineer to review this.
If the url is manipulated, can it hack the other app that is launched from here.

cobalt/android/BUILD.gn
ckim1346: `ExternalNavigationDelegateImpl` implements `org.chromium.components.external_intents.ExternalNavigationDelegate`. Chromium's `ExternalNavigaitonHandler.java` pulls in things like `"//components/messages/android:java",
  "//components/webapk/android/libs/client:java"`, in its gn build file [`components/external_intents/android/BUILD.gn`](https://source.corp.google.com/h/github/youtube/cobalt/+/main:components/external_intents/android/BUILD.gn;l=1?q=components%2Fexternal_intents%2Fandroid%2FBUILD&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) which pull in the /chrome and /components java dependencies which declare native JNI methods and throws the failed JNI assertion. That's why I added the `add_stubs_for_missing_jni = true` so that it only creates empty stubs for the JNI methods we don't need which prevents us from having to compile or link those dependencies

PR #10329, opened 2026-05-05
android: Maintain Cobalt API 24 compatibility

kjyoun:
Ensure continued support for legacy Android TV devices by updating the baseline target SDK.

This PR surgically removes hard linking dependencies and suppresses compiler warnings for unused SDK 26+ APIs (i.e. AAudio and  AImageReader) that break SDK 24 compatibility, preventing startup dynamic linker crashes (`UnsatisfiedLinkError: library "libaaudio.so" not found`) and compile-time `-Wunguarded-availability` issues.

Because Cobalt completely delegates its audio and video rendering pipelines to the Starboard layer, these advanced Chromium graphics and composition features (such as `AImageReader`) are completely inactive[1]. This change ensures the unused files compile cleanly targeting SDK 24, while their dead-code bytes are stripped out of the final binary by the linker.

TESTED=played a video on SDK 24 using xcid (`acid lease_android_emulator GENERIC_PHONE 24`)

Bug: 508072838
Bug: 422247849

---
[1] Codes accessing `AImageReader` or `AAudio` are compiled but stripped away, since there is no code path to them.

### DIFF
--- a/base/android/android_image_reader_compat.cc
+++ b/base/android/android_image_reader_compat.cc
@@ -4,14 +4,18 @@
 
 #include "base/android/android_image_reader_compat.h"
 
+#include "build/build_config.h"
+
 namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
+#if !BUILDFLAG(IS_COBALT)
   // Currently we want to enable AImageReader only for android P+ devices.
   if (__builtin_available(android 28, *)) {
     return true;
   }
+#endif
   return false;
 }
 
--- a/base/android/android_image_reader_compat.cc
+++ b/base/android/android_image_reader_compat.cc
@@ -5,12 +5,20 @@
 #include "base/android/android_image_reader_compat.h"
 
 #include "build/build_config.h"
+#include "build/buildflag.h"
 
 namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
-  // Cobalt does not use AImageReader, which breaks SDK 24 compatibility.
+  // Cobalt completely bypasses Chromium's native video decoding pipeline
+  // (which is the sole instantiator of AImageReader/ImageReaderGLOwner)
+  // and renders all media directly through the Starboard player.
+  //
+  // Disabling AImageReader here also safely prevents the GPU compositor
+  // from attempting to initialize advanced multi-threaded scheduling
+  // features like DrDc (which has a hard hardware dependency on AImageReader
+  // buffer sharing), ensuring maximum rendering stability on TV platforms.
   // See b/508072838.
 #if !BUILDFLAG(IS_COBALT)
   // Currently we want to enable AImageReader only for android P+ devices.
--- a/base/android/android_image_reader_compat.cc
+++ b/base/android/android_image_reader_compat.cc
@@ -10,10 +10,11 @@ namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
-  // COBALT: Disable AImageReader for Cobalt builds to avoid compatibility issues.
-  // Cobalt lowers default_min_sdk_version to 24, so we must guard this.
+  // COBALT: Cobalt does not use AImageReader. We disable it to maintain compatibility
+  // with Android 7.0 (API 24), as AImageReader NDK APIs require API 26+ (and are
+  // only enabled by Chromium at runtime for API 28+).
   // See b/508072838
-#if __ANDROID_MIN_SDK_VERSION__ >= 26
+#if !BUILDFLAG(IS_COBALT)
   // Currently we want to enable AImageReader only for android P+ devices.
   if (__builtin_available(android 28, *)) {
     return true;
--- a/base/android/android_image_reader_compat.cc
+++ b/base/android/android_image_reader_compat.cc
@@ -10,7 +10,10 @@ namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
-#if !BUILDFLAG(IS_COBALT)
+  // COBALT: Disable AImageReader for Cobalt builds to avoid compatibility issues.
+  // Cobalt lowers default_min_sdk_version to 24, so we must guard this.
+  // See b/508072838
+#if __ANDROID_MIN_SDK_VERSION__ >= 26
   // Currently we want to enable AImageReader only for android P+ devices.
   if (__builtin_available(android 28, *)) {
     return true;
--- a/base/android/android_image_reader_compat.cc
+++ b/base/android/android_image_reader_compat.cc
@@ -10,10 +10,8 @@ namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
-  // COBALT: Cobalt does not use AImageReader. We disable it to maintain compatibility
-  // with Android 7.0 (API 24), as AImageReader NDK APIs require API 26+ (and are
-  // only enabled by Chromium at runtime for API 28+).
-  // See b/508072838
+  // Cobalt does not use AImageReader, which breaks SDK 24 compatibility.
+  // See b/508072838.
 #if !BUILDFLAG(IS_COBALT)
   // Currently we want to enable AImageReader only for android P+ devices.
   if (__builtin_available(android 28, *)) {
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -81,7 +81,12 @@ if (is_android) {
       default_min_sdk_version = 29
     }
     if (is_cobalt) {
-      default_min_sdk_version = 26
+      # Cobalt maintains backward compatibility with Android 7.0 (API 24), consistent
+      # with Cobalt 25 and 26 releases. Setting this to 24 allows newer system APIs
+      # (like AAudio) to compile as weak symbols, preventing UnsatisfiedLinkError
+      # startup crashes on legacy TV devices where those libraries don't exist.
+      # See b/508072838, b/422247849
+      default_min_sdk_version = 24
     }
 
     # Static analysis can be either "on" or "off" or "build_server". This
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -82,10 +82,10 @@ if (is_android) {
     }
     if (is_cobalt) {
       # Cobalt maintains backward compatibility with Android 7.0 (API 24), consistent
-      # with Cobalt 25 and 26 releases. Setting this to 24 allows newer system APIs
-      # (like AAudio) to compile as weak symbols, preventing UnsatisfiedLinkError
-      # startup crashes on legacy TV devices where those libraries don't exist.
-      # See b/508072838, b/422247849.
+      # with Cobalt 25 and 26 releases. Setting this to 24 ensures the Android
+      # platform config compiles cleanly targeting API 24, while advanced system
+      # APIs (such as AAudio and AImageReader) are excluded from linking and safely
+      # disabled at runtime. See b/508072838, b/422247849.
       default_min_sdk_version = 24
     }
 
--- a/build/config/android/config.gni
+++ b/build/config/android/config.gni
@@ -85,7 +85,7 @@ if (is_android) {
       # with Cobalt 25 and 26 releases. Setting this to 24 allows newer system APIs
       # (like AAudio) to compile as weak symbols, preventing UnsatisfiedLinkError
       # startup crashes on legacy TV devices where those libraries don't exist.
-      # See b/508072838, b/422247849
+      # See b/508072838, b/422247849.
       default_min_sdk_version = 24
     }
 
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -108,9 +108,6 @@ public final class CommandLineOverrideHelper {
         // For details, see http://b/478022126#comment6.
         paramOverrides.add("UseAAudioInput");
 
-        // Disable AAudio, which breaks SDK 24 compatibility. See b/508072838.
-        paramOverrides.add("UseAAudioDriver");
-
         return paramOverrides;
     }
 
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -108,6 +108,10 @@ public final class CommandLineOverrideHelper {
         // For details, see http://b/478022126#comment6.
         paramOverrides.add("UseAAudioInput");
 
+        // Disable AAudio output to prevent UnsatisfiedLinkError crashes on older Android TVs (like Android 7.1).
+        // See b/508072838
+        paramOverrides.add("UseAAudioDriver");
+
         return paramOverrides;
     }
 
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -108,8 +108,7 @@ public final class CommandLineOverrideHelper {
         // For details, see http://b/478022126#comment6.
         paramOverrides.add("UseAAudioInput");
 
-        // Disable AAudio output to prevent UnsatisfiedLinkError crashes on older Android TVs (like Android 7.1).
-        // See b/508072838
+        // Disable AAudio, which breaks SDK 24 compatibility. See b/508072838.
         paramOverrides.add("UseAAudioDriver");
 
         return paramOverrides;
--- a/gpu/command_buffer/service/BUILD.gn
+++ b/gpu/command_buffer/service/BUILD.gn
@@ -566,6 +566,8 @@ target(link_target_type, "gles2_sources") {
     sources += [
       "ahardwarebuffer_utils.cc",
       "ahardwarebuffer_utils.h",
+      "image_reader_gl_owner.cc",
+      "image_reader_gl_owner.h",
       "ref_counted_lock.cc",
       "ref_counted_lock.h",
       "shared_image/ahardwarebuffer_image_backing_factory.cc",
@@ -590,15 +592,6 @@ target(link_target_type, "gles2_sources") {
       "texture_owner.cc",
       "texture_owner.h",
     ]
-
-    # Cobalt does not use ImageReaderGLOwner, which requires API 26+, breaking SDK 24 compatibility.
-    # See b/508072838.
-    if (!is_cobalt) {
-      sources += [
-        "image_reader_gl_owner.cc",
-        "image_reader_gl_owner.h",
-      ]
-    }
 
     # For AImageReader.
     libs = [ "mediandk" ]
@@ -687,15 +680,10 @@ if (is_android) {
   source_set("android_texture_owner_unittests") {
     testonly = true
     sources = [
+      "image_reader_gl_owner_unittest.cc",
       "surface_texture_gl_owner_unittest.cc",
       "surface_texture_transform_unittest.cc",
     ]
-
-    # Cobalt does not use ImageReaderGLOwner, which requires API 26+, breaking SDK 24 compatibility.
-    # See b/508072838.
-    if (!is_cobalt) {
-      sources += [ "image_reader_gl_owner_unittest.cc" ]
-    }
 
     # For AImageReader_getFormat()
     libs = [ "mediandk" ]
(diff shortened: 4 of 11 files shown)


gpu/command_buffer/service/texture_owner.cc
```diff
@@ -14,12 +14,16 @@
...
-#include "gpu/command_buffer/service/image_reader_gl_owner.h"
 #include "gpu/command_buffer/service/surface_texture_gl_owner.h"
 #include "gpu/command_buffer/service/texture_base.h"
 #include "ui/gl/scoped_binders.h"
 #include "ui/gl/scoped_make_current.h"
 
+#if BUILDFLAG(IS_ANDROID) && \
+    (__ANDROID_MIN_SDK_VERSION__ >= AIMAGEREADER_MIN_API)
+#include "gpu/command_buffer/service/image_reader_gl_owner.h"
+#endif
```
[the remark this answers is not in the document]
kjyoun: Done

gpu/command_buffer/service/texture_owner.cc
```diff
@@ -104,9 +108,12 @@ scoped_refptr<TextureOwner> TextureOwner::Create(
     case Mode::kAImageReaderInsecure:
     case Mode::kAImageReaderInsecureSurfaceControl:
     case Mode::kAImageReaderSecureSurfaceControl:
+#if BUILDFLAG(IS_ANDROID) && \
+    (__ANDROID_MIN_SDK_VERSION__ >= AIMAGEREADER_MIN_API)
```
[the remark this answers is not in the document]
kjyoun: Done

media/gpu/BUILD.gn
```diff
@@ -184,9 +188,13 @@ component("gpu") {
...
     }
     libs += [
-      "aaudio",
       "mediandk",
     ]
+    # AAudio is removed for Cobalt (or any SDK < 26 build) to prevent UnsatisfiedLinkError on older Android TVs.
+    # See b/508072838
+    if (default_min_sdk_version >= 26) {
+      libs += [ "aaudio" ]
+    }
```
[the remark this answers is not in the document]
kjyoun: this is intended

media/audio/BUILD.gn
```diff
@@ -281,7 +285,11 @@ source_set("audio") {
         "android/opensles_wrapper.cc",
       ]
     }
-    libs += [ "aaudio" ]
+    # AAudio is removed for Cobalt (or any SDK < 26 build) to prevent UnsatisfiedLinkError on older Android TVs.
+    # See b/508072838
+    if (default_min_sdk_version >= 26) {
+      libs += [ "aaudio" ]
+    }
```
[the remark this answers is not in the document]
kjyoun: this is intended

third_party/webrtc/sdk/android/BUILD.gn
```diff
@@ -1285,7 +1285,12 @@ if (current_os == "linux" || is_android) {
         "src/jni/audio_device/aaudio_wrapper.cc",
         "src/jni/audio_device/aaudio_wrapper.h",
       ]
-      libs = [ "aaudio" ]
+      libs = []
+      # AAudio is removed for Cobalt (or any SDK < 26 build) to prevent UnsatisfiedLinkError on older Android TVs.
+      # See b/508072838
+      if (default_min_sdk_version >= 26) {
+        libs += [ "aaudio" ]
+      }
```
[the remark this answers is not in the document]
kjyoun: this is intended to minimize the diff.

code path to use aaudio is turned off, so that code is not actually executed

gpu/command_buffer/service/texture_owner.cc
```diff
@@ -14,12 +14,16 @@
...
-#include "gpu/command_buffer/service/image_reader_gl_owner.h"
 #include "gpu/command_buffer/service/surface_texture_gl_owner.h"
 #include "gpu/command_buffer/service/texture_base.h"
 #include "ui/gl/scoped_binders.h"
 #include "ui/gl/scoped_make_current.h"
 
+#if BUILDFLAG(IS_ANDROID) && \
+    (__ANDROID_MIN_SDK_VERSION__ >= 26)
+#include "gpu/command_buffer/service/image_reader_gl_owner.h"
+#endif
```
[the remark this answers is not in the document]
kjyoun: Done

gpu/command_buffer/service/texture_owner.cc
```diff
@@ -104,9 +108,12 @@ scoped_refptr<TextureOwner> TextureOwner::Create(
     case Mode::kAImageReaderInsecure:
     case Mode::kAImageReaderInsecureSurfaceControl:
     case Mode::kAImageReaderSecureSurfaceControl:
+#if BUILDFLAG(IS_ANDROID) && \
+    (__ANDROID_MIN_SDK_VERSION__ >= 26)
       return new ImageReaderGLOwner(std::move(texture), mode,
                                     std::move(context_state),
                                     std::move(drdc_lock), type_for_metrics);
+#endif
```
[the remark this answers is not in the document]
kjyoun: This comment is not valid.

johnxwork (approved):

gpu/command_buffer/service/BUILD.gn
```diff
@@ -566,8 +569,6 @@ target(link_target_type, "gles2_sources") {
     sources += [
       "ahardwarebuffer_utils.cc",
       "ahardwarebuffer_utils.h",
-      "image_reader_gl_owner.cc",
```
borongc, line 569: This changes are not gated by is_cobalt or Cobalt build flags, just want to check how can we differentiate our changes with upstream codes?

build/config/android/config.gni
```diff
@@ -81,7 +81,12 @@ if (is_android) {
       default_min_sdk_version = 29
     }
     if (is_cobalt) {
-      default_min_sdk_version = 26
+      # Cobalt maintains backward compatibility with Android 7.0 (API 24), consistent
+      # with Cobalt 25 and 26 releases. Setting this to 24 allows newer system APIs
+      # (like AAudio) to compile as weak symbols, preventing UnsatisfiedLinkError
+      # startup crashes on legacy TV devices where those libraries don't exist.
+      # See b/508072838, b/422247849
+      default_min_sdk_version = 24
```
jasonzhangxx, line 89: @kjyoun, can we confirm that if chrome doesn't have any other dependency for api level 25/26?

gpu/command_buffer/service/BUILD.gn
kjyoun: I have added  COBALT:  comment blocks and the tracking bug number ( b/508072838 ) to explicitly mark these changes. 

I believe guarding this with the Android API version (`default_min_sdk_version`  /  `__ANDROID_MIN_SDK_VERSION__` ) is better than using `IS_COBALT`  directly. It clearly documents the technical rationale (supporting Android 7.0 / API 24) while keeping the code cleaner

build/config/android/config.gni
kjyoun: For context, this is not a new feature or behavior change, but rather a regression fix. The  default_min_sdk_version  was mistakenly(?) bumped to 26 in commit  2b6e73fd8a787ab873dcbdc55ce0245aca1be2d2  during the M138 migration. Cobalt 25/26 have historically maintained  compatibility with API 24.
 
 While I cannot 100% guarantee there are no other hidden API 24/25 dependencies in the codebase, this PR successfully resolves all the known startup blockers. A quick check by Gemini did not reveal any other obvious API conflicts.
  
 More importantly, both QA and I have empirically verified that the app builds successfully and runs without issues on physical Android  7.0 (SDK 24) devices. I think this is sufficient for now, and if any other edge-case blockers arise in the future, we can address them then, including a major product decision like dropping the support of SDK24/25.

base/android/android_image_reader_compat.cc
```diff
@@ -4,14 +4,21 @@
...
+#include "build/build_config.h"
+
 namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
+  // COBALT: Disable AImageReader for Cobalt builds to avoid compatibility issues.
+  // Cobalt lowers default_min_sdk_version to 24, so we must guard this.
+  // See b/508072838
+#if __ANDROID_MIN_SDK_VERSION__ >= 26
```
borongc: We can just disable it and gate by IS_COBALT, as we don't use ImageReader and this code path for Cobalt entirely.

gpu/command_buffer/service/BUILD.gn
borongc: I agree that using API is better for this case, but I'm asking what's the long-term way we will maintain this.

For example, if Chromium newer milestones deprecate some APIs, how can we ensure we track all of these differences?

base/android/android_image_reader_compat.cc
borongc: Also, most of ImageReader APIs are available since API 19 (https://developer.android.com/reference/android/media/ImageReader). Which are the issues for SDK < 26?

kjyoun: Gating by  IS_COBALT  is indeed the correct approach because Cobalt does not use the  AImageReader  path at all (it is not just an  SDK compatibility issue; our Starboard-based media pipeline simply has no need for it).

Updated the code and the comment


re: SDK version:
We are using the native NDK  AImageReader  APIs here, not the Java  ImageReader  class (which was indeed API 19).

https://developer.android.com/ndk/reference/group/media#aimagereader_newwithusage

gpu/command_buffer/service/BUILD.gn
kjyoun: Make sense. 

Switched to IS_COBALT. and updated comment.

base/android/android_image_reader_compat.cc
borongc: My bad, it should be AImageReader, which is available for SDK >= 24. So, I think we should be fine with target SDK to 24.

In this case, we should figure out which  AImageReader APIs cause the compiling issues.

gpu/command_buffer/service/BUILD.gn
```diff
@@ -593,6 +591,15 @@ target(link_target_type, "gles2_sources") {
       "texture_owner.h",
     ]
 
+    # ImageReader is disabled because it is not used and breaks SDK 24 compatibility.
+    # See b/508072838.
+    if (!is_cobalt) {
+      sources += [
+        "image_reader_gl_owner.cc",
```
borongc: Same as the comment in other thread, can we check which AImageReader APIs break SDK 24 compatibility?

borongc: The reason I asked this is due to image_reader_gl_owner is there in m114, but if this doesn't work in m138, we can just simply gate those non-compatible functions in `image_reader_gl_owner.{cc/h}`, as we don't use them, so it's fine.

This also helps us (media team) to track what's needed to use AImageReader for Starboard Android, as this could be a future topic for decode-to-texture mode.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -108,6 +108,9 @@ public static StringJoiner getDefaultDisableFeatureOverridesList() {
         // For details, see http://b/478022126#comment6.
         paramOverrides.add("UseAAudioInput");
 
+        // Disable AAudio, which breaks SDK 24 compatibility. See b/508072838.
```
yell0wd0g: Can we use here the
```
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```

build syntax used in WindowAndroid.java a bit further below?
Or maybe it makes sense to unconditionally disable it, case
in which feel free to disregard this comment.

base/android/android_image_reader_compat.cc
kjyoun: Yes, NDK  AImageReader  core features were introduced in API 24, but Chromium's  ImageReaderGLOwner  is implemented using APIs that require API level 26 (Android 8.0).
 
 The specific compilation/linker blockers are:
 
 -  AImageReader_newWithUsage https://developer.android.com/ndk/reference/group/media#aimagereader_newwithusage (API 26)
 -  AImage_getHardwareBuffer  https://developer.android.com/ndk/reference/group/media#aimage_gethardwarebuffer (API 26)
 -   AImageReader_acquireNextImageAsync  (API 26)
 -  AImage_deleteAsync  (API 26)"

gpu/command_buffer/service/BUILD.gn
kjyoun: m114 supports dynamic loading based on runtime SDK version

https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:base/android/android_image_reader_compat.cc;l=39-51;drc=dcf01765e6d562898a5dbd58faa5ffce256c7aaf

while this PR completely disabled it. 

Let me think how we can do simialar thing

base/android/android_image_reader_compat.cc
```diff
@@ -4,14 +4,20 @@
...
 
+#include "build/build_config.h"
+
 namespace base {
 namespace android {
 
 bool EnableAndroidImageReader() {
+  // Cobalt does not use AImageReader, which breaks SDK 24 compatibility.
+  // See b/508072838.
+#if !BUILDFLAG(IS_COBALT)
```
yell0wd0g, line 23: Do we need this `#if !...` ? IOW would the `builtin_available`
fail to compile/be parsed when building with previous SDKs?

gpu/command_buffer/service/texture_owner.cc
```diff
@@ -14,12 +14,17 @@
...
 #include "gpu/command_buffer/service/feature_info.h"
-#include "gpu/command_buffer/service/image_reader_gl_owner.h"
 #include "gpu/command_buffer/service/surface_texture_gl_owner.h"
 #include "gpu/command_buffer/service/texture_base.h"
 #include "ui/gl/scoped_binders.h"
 #include "ui/gl/scoped_make_current.h"
 
+// Cobalt does not use ImageReaderGLOwner, which requires API 26+, breaking SDK 24 compatibility.
+// See b/508072838.
+#if !BUILDFLAG(IS_COBALT)
```
yell0wd0g: Sorry for the delay in reviewing this. I kept looking and
but couldn't marshall my thoughts, but now I think I have
an idea.

Technically we don't use this not because we're Cobalt but
because the SDK version we use is too old. So would it be
more correct e.g.

```
  // Cobalt builds and maintains compat with Android SDK 24, but 
  // ImageReaderGLOwner uses SDK 28, see b/...
#if __ANDROID_MIN_SDK_VERSION__ <= 24
#include "gpu/command_buffer/service/image_reader_gl_owner.h"
#endif
```

Similarly for Java files 
```
if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
```
which is correctly used in 
ui/android/java/src/org/chromium/ui/base/WindowAndroid.java
but not in
cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
(I'd prefer the "correct version everywhere because maybe
one day we'll stop supporting SDKs before 28, for example,
and it would be hard to figure out all the APIs that we'd
be able to start using).

Finally I think the same pattern should be used in the
gn files, so instead of `if(is_cobalt)`, just use
`default_min_sdk_version` comparisons, if possible;
with a comment saying that this a cobalt specific 
provision. WDYT?

yell0wd0g (commented):
Apologies for the delay, needed to think a bit
about this CL. Thanks for tackling this! It's a
tough problem.

gpu/command_buffer/service/texture_owner.cc
borongc: We were told that we should use `IS_COBALT`, `IS_STARBOARD`, or `USE_STARBOARD_MEDIA` to gate our Cobalt changes to Chromium files. If we don't use them, this would be a new case.

gpu/command_buffer/service/BUILD.gn
borongc: This seems to be a better approach, where we can keep this as part of our Cobalt code to make it compiled.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
kjyoun: It turns out that we don't need to disable  UseAAudioDriver , since the AAudio library is not used regardless of this flag. There is no code path to use AAudio. Instead, Cobalt uses Starboard media.

gpu/command_buffer/service/BUILD.gn
kjyoun: It turns out Cobalt doesn't use  ImageReaderGLOwner . It is compiled, but stripped away from the final binary since there is no code path to it, as Cobalt uses Starboard media. Hence, we only have to suppress build warnings.

base/android/android_image_reader_compat.cc
kjyoun: Yes, I think we had better return  false  for Cobalt. Cobalt has no code path to  AImageReader , so all code leading to  AImageReader  is stripped away. So it is better to return  false  here, vs returning  true / false  based on the SDK version. Returning  true  is not correct, since Cobalt has no code path to use  AndroidImageReader . In Cobalt 26, we explicitly disabled  AImageReader  [1].
 
 [1] https://source.corp.google.com/h/github/youtube/cobalt/+/26.android:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java;l=87-91;drc=2845b06e703edab16577dbff202c37a2c0b0062d

gpu/command_buffer/service/texture_owner.cc
kjyoun: I also thought a lot about this. I originally had your stance (using version flag vs is_cobalt). 

Here is the reason that I use is_cobalt flag

- It's already well-known practice to guard cobalt-specific changes under chromium code. Searching by `BUILDFLAG(IS_COBALT)` is a very convenient way to find Cobalt-specific changes. [1]

- More importantly, using NDK  AAudio  and AImageReader is more product decision than simple SDK version. Cobalt delegates all media  rendering and audio output to the Starboard layer (`SbPlayer`, `SbAudioSink`). There is no code path to those NDK libraries[2].  IOW, `#if __ANDROID_MIN_SDK_VERSION__ <= 24` does not decide whether to use these APIs or not.  I think that, when Cobalt baseline upgrades to SDK 26+ in the future,  I think we need to "evaluate" cost/benefit of using those APIs and make decision. 
  -  Compared to this, I added simple if version check logic to `WindowAndroid.java`, since it is purely decided by  SDK version.

---
[1] https://source.corp.google.com/search?q=IS_COBALT%20lang:Cc&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

[2] The code apparently using those NDK APIs are compiled since they are listed in build.gn, but stripped away from the final binary, since there are no code path that actually use them.

borongc (approved):

base/android/android_image_reader_compat.cc
yell0wd0g: Gotcha thanks. My point was, could we not just write:
```
bool EnableAndroidImageReader() {
  // ...
  if (__builtin_available(android 28, *)) {
    return true;
  }
  return false;
}
```

without the `#if` ?

media/audio/BUILD.gn
```diff
@@ -281,7 +281,10 @@ source_set("audio") {
         "android/opensles_wrapper.cc",
       ]
     }
-    libs += [ "aaudio" ]
+    # Cobalt uses the Starboard layer instead of AAudio. AAudio also breaks SDK 24 compatibility. See b/508072838.
+    if (!is_cobalt) {
```
yell0wd0g, line 285: Not sure if I mentioned this, but could we do here:
```
  // Cobalt supports SDKs before 24 and those don't support AAudio.
  if (default_min_sdk_version > 24) {
    libs += [ "aaudio" ]
  }
```

IOW we don't use it because of the min SDK, not because of
Cobalt? Oh but I see you say Cobalt doesn't use it anyway.
Hmm.

yell0wd0g (approved):
Changes LGTM! Thanks!

base/android/android_image_reader_compat.cc
kjyoun: I think that returning false here regardless of SDK version is the right behavior (using #if).

Cobalt is not using AImageReader, regardess of SDK version.

media/audio/BUILD.gn
kjyoun: We don't use  AAudio  because Cobalt delegates all media rendering to the Starboard layer, regardless of the Android SDK version.
 
Simply raising the SDK version wouldn't make  AAudio  work; doing so would require a major architectural overhaul of our media pipeline.

Therefore, using  `if (default_min_sdk_version > 24)`  would be misleading, as it falsely suggests that using  AAudio  is purely an SDK compatibility issue for Cobalt. Using  `!is_cobalt`  correctly reflects that this is a Cobalt-specific architectural decision.

PR #10331, opened 2026-05-05
AOSP cpp17 fixes

rogerzanoni:
AOSP is going to use cpp17 for starboard/native targets, these changes address issues found while building with the platform.

The initial config for aosp is being reviewed at https://github.com/youtube/cobalt/pull/10313

Bug: 495203133


hferreiro:
I think you forgot to remove scoped_java_ref.cc

rogerzanoni:
> I think you forgot to remove scoped_java_ref.cc

I have another commit that drops it, I will cherry-pick it to this branch

andrewsavage1:
Please add a little bit more explanation to your description about why we need to make these changes, i.e. if it's due to copied_base assuming c++20 features please very explicitly call that out and explain why this change fixes it

rogerzanoni:
Updated the commit messages with a better explanation about the issues they fix, squashed two of them (the one that add type aliases to jni_zero types and one that was removing the copied_base ScopedJavaRef impl)

third_party/jni_zero/default_conversions.h
```diff
@@ -14,6 +14,8 @@
 
 namespace jni_zero {
 
+#ifdef JNI_ZERO_ENABLE_TYPE_CONVERSIONS
```
haozheng-cobalt, line 15: We should not change third_party/jni_zero like this. it's used across platforms for 1P ATV, the change should be restricted in //copied_base.

haozheng-cobalt (changes requested):

third_party/jni_zero/default_conversions.h
rogerzanoni: Thanks for reviewing it @haozheng-cobalt, for this particular one I think that it makes sense to make the change in jni_zero as we discussed, and adding the STARBOARD check do the code I'm using to gate `JNI_ZERO_ENABLE_TYPE_CONVERSIONS` in `third_party/jni_zero/type_conversions.h`, because event that aosp is the only target that currently is going to use cpp17 for startboartd, it may be a problem if other configurations build with cpp17 as well, and the change won't affect configurations that already use cpp20

haozheng-cobalt: Gotcha, will be better to move this guard to the beginning of this file, like at line5?

third_party/jni_zero/default_conversions.h
haozheng-cobalt, line 5: link

haozheng-cobalt (approved):

PR #10338, opened 2026-05-06
tvos: Remove obsolete MimeType validity check

abhijeetk:
MimeType was refactored in #10303 to use a factory method returning
std::optional but the tvOS video support check was not updated, causing
a build error. The validity check is now redundant and can be removed.

Bug: 479994435
Bug: 487326366


abhijeetk:
cc: @Gyuyoung @rakuco @jkim-julie

rakuco (approved):

kjyoun (approved):

yell0wd0g (approved):
RS LGTM!

PR #10341, opened 2026-05-06
starboard/tests: wire boringssl_pki_tests target

denisyuji:

Includes:

Bug: 486053350

### DIFF
--- a/cobalt/build/testing/targets/evergreen-arm-hardfp-rdk/test_targets.json
+++ b/cobalt/build/testing/targets/evergreen-arm-hardfp-rdk/test_targets.json
@@ -15,6 +15,13 @@
       "third_party/zlib:zlib_unittests_loader"
     ],
     "TODO(b/486053350): Crashes on missing SbEventHandle": [
+    "executable": "out/evergreen-arm-hardfp-rdk_devel/storage_unittests_loader.py",
+    "runs_on": "device",
+    "test_type": "gtest"
+  },
+  {
+    "target": "ui/views/examples:views_examples_unittests_loader",
+    "executable": "out/evergreen-arm-hardfp-rdk_devel/views_examples_unittests_loader.py"
       "gl:gl_unittests_loader",
       "gl:ozone_gl_unittests_loader",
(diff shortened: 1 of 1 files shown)


cobalt/testing/filters/evergreen-arm-hardfp-rdk/boringssl_pki_tests_loader_filter.json
```diff
@@ -0,0 +1,5 @@
+{
+  "failing_tests": [
+    "*"
```
niranjanyardi, line 4: same question : why are all tests filtered ?
There should be some passing subset

lamdbui: @denisyuji , any context for the filtering?

niranjanyardi: please add the bug 509955465 in the  PR description - seems like the analysis is in a separate bug

denisyuji: as per my tests, all the filtered tests failed on the first CI run. I will re-try the tests

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn
```diff
@@ -377,9 +377,15 @@ group("loader_app_rdk_plugin") {
 }
 
 if (sb_is_evergreen_compatible && current_toolchain == starboard_toolchain) {
-  target("shared_library", "loader_app") {
-    output_dir = root_build_dir
-    deps = [ "//starboard/loader_app:loader_app_sources" ]
-    data_deps = [ "//starboard/content/fonts:copy_font_data" ]
```
niranjanyardi: this is a deeper investigation :

https://b.corp.google.com/issues/481314284#comment19

lets keep it seperate from adding th test

niranjanyardi (changes requested):

denisyuji:
added on #10222

PR #10259, opened 2026-04-28
android: Add WebContentsObserver for mediaSession

jasonzhangxx:
android: Add WebContentsObserver for mediaSession

Attach a WebContentsObserver to CobaltMediaSession on Android.

Previously, the MediaSession was initialized when 
MediaSession.fromWebContents() was called, which has been changed
in m138.

By introducing a WebContentsObserver, CobaltMediaSession can now
react to mediaSessionCreated events, ensuring that the
MediaSessionObserver is correctly attached to any new media
sessions. This also adds handling for webContentsDestroyed to
ensure proper cleanup.

Bug: 506079374

PR #10345, opened 2026-05-06
Revert "android: Add WebContentsObserver for mediaSession"

Awallky:
Reverts youtube/cobalt#10259

Test: Load production app & ensure no crashes due to `java.lang.ClassNotFoundException: org/chromium/services/media_session/MediaImage`
Bug: 508673132


yell0wd0g (approved):
RS cklean revert

Awallky:
FYI @jasonzhangxx 
This fixes the crash locally on our production app (not CoAT). I am reverting this. We need a fix for the crash before re-landing this.

PR #10346, opened 2026-05-06
android: Fix memory leak and OOM crash in AudioTrackBridge

borongc:
The change addresses two memory-related issues in the Android AudioTrack
implementation that could lead to application crashes.

First, ensure the Java AudioTrackBridge object is properly destroyed if
the native audio data array allocation fails during creation. Previously,
this path would leak the object and its underlying AudioTrack.

Second, catch Exception and OutOfMemoryError when calling
AudioTrack.getTimestamp() on the audio hot path. This avoids catching the
overly broad Throwable (which can mask critical system failures) while still
protecting against platform-specific RuntimeExceptions and
OutOfMemoryErrors on low-resource devices. This allows the system to
log the error and fall back gracefully to the last known position rather
than crashing.

This is introduced by https://github.com/youtube/cobalt/pull/10291.

#vibe-coded

Issue: 510354897

### DIFF
--- a/starboard/android/shared/audio_track_bridge.cc
+++ b/starboard/android/shared/audio_track_bridge.cc
@@ -66,23 +66,6 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
   }
 
   JNIEnv* env = AttachCurrentThread();
-  ScopedJavaLocalRef<jobject> j_audio_track_bridge =
-      AudioOutputManager::GetInstance()->CreateAudioTrackBridge(
-          env, GetAudioFormatSampleType(coding_type, sample_type),
-          sampling_frequency_hz, channels, preferred_buffer_size_in_bytes,
-          tunnel_mode_audio_session_id, is_web_audio);
-
-  if (j_audio_track_bridge.is_null()) {
-    // One of the cases that this may hit is when output happened to be switched
-    // to a device that doesn't support tunnel mode.
-    // TODO: Find a way to exclude the device from tunnel mode playback, to
-    //       avoid infinite loop in creating the audio sink on a device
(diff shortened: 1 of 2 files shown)


osagie98 (approved):

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -106,6 +106,8 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
 
   if (j_audio_data.is_null()) {
     SB_LOG(WARNING) << "Failed to allocate |j_audio_data_|";
+    AudioOutputManager::GetInstance()->DestroyAudioTrackBridge(
```
kjyoun: Thanks for this fix!

Indeed, this is a regression from my recent refactoring (#10291). You can mention in description for the tracking.

For the fix, could we change the order of the code so that the  AudioTrackBridge  instance is only created after  j_audio_data  is allocated?
 
This way, we don't need to destroy the already-created instance on failure, which is more future-proof. Otherwise, it is easy to forget to call  DestroyAudioTrackBridge()  (as I did!).
 
 Here is the suggested change: https://github.com/youtube/cobalt/pull/10347/changes/7902ddc16563b91d91b9ce849f0409c01c8b9830

borongc: Done. Thanks for the suggestions.

kjyoun (approved):

yell0wd0g (approved):
RS changes, two things to consider in the future:
a) Submit two CLs instead of one with two fixes.
b) Can this be tested? If not (because it's hard
to inject platform errors), it'd be good to say
so in the commit description (maybe also tell how
you found/repro'ed the fix).

PR #10354, opened 2026-05-06
cobalt: Clean up Starboard media settings

xiaomings:
Group Starboard media-specific settings within H5vccSettings::set
under a single BUILDFLAG(USE_STARBOARD_MEDIA) block. This removes
redundant preprocessor checks inside individual lambdas.

Also add a comment to explain the fallback behavior when
USE_STARBOARD_MEDIA is not defined, ensuring consistency with other
experiments.

Bug: 454441375

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
@@ -143,9 +143,9 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
     const WTF::String& name,
     const V8UnionLongOrString* value,
     ExceptionState& exception_state) {
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
   const ExceptionContext& exception_context = exception_state.GetContext();
 
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
   if (name == "DecoderBuffer.EnableDecommitableAllocatorStrategy") {
     return ProcessSettingAsEnableOnly(
         script_state, exception_context, name, *value, [] {
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
@@ -145,27 +145,20 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
     ExceptionState& exception_state) {
   const ExceptionContext& exception_context = exception_state.GetContext();
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
   if (name == "DecoderBuffer.EnableDecommitableAllocatorStrategy") {
     return ProcessSettingAsEnableOnly(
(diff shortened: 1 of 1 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -145,27 +145,20 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
     ExceptionState& exception_state) {
   const ExceptionContext& exception_context = exception_state.GetContext();
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
xiaomings: @johnedocampo this changes the behavior slightly, as some settings are no longer rejected.  However, this shouldn't affect production.  Please help me review the PR.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
johnedocampo, line 124: With the new changes, this code is no longer reachable, we should be good to remove it.

johnedocampo, line 103: With the new changes, this if statement shouldn't be reachable anymore. I think it will be fine to remove this.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
johnedocampo: I think the change is fine. Currently, the experiment code will always return true on builds with `USE_STARBOARD_MEDIA`; we only return false on builds where the flag is false. I agree that the builds without `USE_STARBOARD_MEDIA` should be fine too since the flag is always enabled in production.

johnedocampo (approved):
Changes lgtm % comments to remove dead code. I am not aware of any builds where USE_STARBOARD_MEDIA isn't enabled and needs the experiment values, so I think the risk of these changes are extremely small.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -185,26 +178,23 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
...
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
           ::media::StreamParser::SetEnableIncrementalParseLookAhead(true);
           return true;
-#else   // BUILDFLAG(USE_STARBOARD_MEDIA)
-          return false;
-#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
         });
   }
+#else   // BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Fall back to Mojo if BUILDFLAG(USE_STARBOARD_MEDIA) isn't defined. The
```
yell0wd0g, line 194: Thanks for the comment! Help reviewing

yell0wd0g (approved):
LGTM! Love a good cleanup.

PR #10361, opened 2026-05-07
aosp: Define is_androidtv with is_starboard

hferreiro:
AOSP builds will support `use_evergreen=false`. Use `is_starboard` to tell them apart from AndroidTV.

Bug: 495203133


cobalt/build/configs/initialize_variables.gni
```diff
@@ -28,7 +28,9 @@ declare_args() {
   build_with_separate_cobalt_toolchain =
       # Avoid using `is_androidtv` which uses `current_os`, so that this is also
       # false for the host toolchain.
-      (target_os != "android" || use_evergreen) && target_os != "ios"
+      (target_os != "android" || is_starboard) && target_os != "ios"
```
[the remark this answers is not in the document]
hferreiro: Not an issue.

cobalt/build/configs/starboard.gni
```diff
@@ -12,19 +12,8 @@
...
-
-  declare_args() {
-    # Differentiates our 1P ATV platform from 3P AOSP.
-    is_androidtv = is_android && !use_evergreen
-  }
-}
-
 declare_args() {
-  is_starboard = is_cobalt && !is_androidtv && target_os != "ios" &&
+  is_starboard = is_cobalt && target_os != "android" && target_os != "ios" &&
```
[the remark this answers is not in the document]
hferreiro: In evergreen builds, `target_os == "linux"`.

cobalt/build/configs/starboard.gni
```diff
@@ -38,5 +27,10 @@ declare_args() {
   sb_is_evergreen_compatible = is_starboard
 }
 
+if (is_cobalt) {
+  # Differentiates our 1P ATV platform from 3P AOSP.
+  is_androidtv = is_android && !is_starboard
+}
```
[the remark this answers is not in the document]
hferreiro: `is_androidtv` shouldn't be overridden.

andrewsavage1 (approved):

andrewsavage1:
FYI @haozheng-cobalt

cobalt/build/configs/initialize_variables.gni
```diff
@@ -28,7 +28,9 @@ declare_args() {
...
+
+  use_evergreen = false
```
hferreiro, line 33: @andrewsavage1 Btw, I restored `use_evergreen` to here, as it was before https://github.com/youtube/cobalt/pull/9677. I'm not sure if I should do the same with `is_androidtv` and move it back to starboard/build/buildflags.gni.

cobalt/build/configs/initialize_variables.gni
yell0wd0g, line 31: This is getting a bit hard to read, with all the negatives, 
and it maybe is because of the negative definitions, i.e.
`build_with_separate_cobalt_toolchain` is defined as what
it isn't, which forces the reader to know all the different
platforms and configuration and then derive where this flag
is set. Could we transform the writing into more positive
statements like e.g. (I'm not sure the logic is correct):
```
build_... = 
    (target_os == "linux") || 
    (target_os == "android" && is_starboard) ||
    ...;
```
?
Same comment applies to `starboard.gni` below.

cobalt/build/configs/initialize_variables.gni
```diff
@@ -28,7 +28,9 @@ declare_args() {
   build_with_separate_cobalt_toolchain =
       # Avoid using `is_androidtv` which uses `current_os`, so that this is also
       # false for the host toolchain.
```
yell0wd0g, line 30: nit: Could you add a comment linking back to cobalt/doc/build_flavours.md plz? Thanks!

yell0wd0g (commented):
The idea here looks good, I have a comment on readability.

cobalt/build/configs/initialize_variables.gni
andrewsavage1: I think it's okay to keep is_androidtv where it is now

cobalt/build/configs/initialize_variables.gni
andrewsavage1: The reason I don't like this suggestion is because hermetic Starboard is supposed to be the default case, and we're noting that these specific OSes are exceptions to that rule

hferreiro: Perhaps I can just add a comment with that info.

cobalt/build/configs/initialize_variables.gni
hferreiro: Not sure where exactly. In particular this variable is not even in that file.

PR #10364, opened 2026-05-07
AOSP: add missing includes

rogerzanoni:
These are minor fixes that add missing includes that caused buid issues while building using the AOSP config (https://github.com/youtube/cobalt/pull/10313)

Bug: 495203133

### DIFF
--- a/starboard/android/shared/system_get_stack.cc
+++ b/starboard/android/shared/system_get_stack.cc
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <link.h>
 #include <unwind.h>
 
 #include "starboard/common/log.h"
--- a/copied_base/base/android/java_exception_reporter.cc
+++ b/copied_base/base/android/java_exception_reporter.cc
@@ -12,6 +12,7 @@
 #include "base/functional/bind.h"
 #include "base/functional/callback.h"
 #include "base/lazy_instance.h"
+#include "base/logging.h"
 
 using base::android::JavaParamRef;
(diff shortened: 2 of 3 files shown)


copied_base/base/base_paths_android.cc
```diff
@@ -5,6 +5,8 @@
 // Defines base::PathProviderAndroid which replaces base::PathProviderPosix for
 // Android in base/path_service.cc.
 
+#include "base/base_paths_android.h"
```
hferreiro: This should be base/base_paths.h".

andrewsavage1 (approved):

andrewsavage1:
You'll have to fix the lint pre-commit check, otherwise lgtm

copied_base/base/base_paths_android.cc
rogerzanoni: Removed this commit. The build issue was happening only in the previous version of the aosp arm configuration.

starboard/android/shared/system_get_stack.cc
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <unwind.h>
+#include <link.h>
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

haozheng-cobalt (approved):
Looks good to me, please make sure there is no issues with CIs.

haozheng-cobalt:
You could look at the log of the failing on-device test CIs, sometimes they are flaky, if the failure is not caused by your change, you could try to refresh the CI 2 more times.

PR #10365, opened 2026-05-07
aosp: fix minor warning issues

rogerzanoni:
Fixes minor warning issues that cause build errors due to -Werror in aosp.

aosp is currently setting treat_warnings_as_errors=false, but the fixes are needed in order to re-enable it.

Bug: 495203133

### DIFF
--- a/copied_base/base/process/current_process.cc
+++ b/copied_base/base/process/current_process.cc
@@ -45,6 +45,8 @@ const char* GetNameForProcessType(CurrentProcessType process_type) {
       return "Service: proxy_resolver.mojom.ProxyResolverFactory";
     case CurrentProcessType::PROCESS_SERVICE_CDM:
       return "Service: media.mojom.CdmService";
+    case CurrentProcessType::PROCESS_SERVICE_MEDIA_FOUNDATION:
+      return "Service: media.mojom.MediaFoundationServiceBroker";
     case CurrentProcessType::PROCESS_SERVICE_VIDEO_CAPTURE:
       return "Service: video_capture.mojom.VideoCaptureService";
     case CurrentProcessType::PROCESS_SERVICE_UNZIPPER:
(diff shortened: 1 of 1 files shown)


hferreiro:
I changed the initial configuration to include `treat_warnings_as_errors=false`, so perhaps you should ammend the description to say that this is part of trying to remove that.

copied_base/base/process/current_process.cc
```diff
@@ -45,6 +45,8 @@ const char* GetNameForProcessType(CurrentProcessType process_type) {
       return "Service: proxy_resolver.mojom.ProxyResolverFactory";
     case CurrentProcessType::PROCESS_SERVICE_CDM:
       return "Service: media.mojom.CdmService";
+    case CurrentProcessType::PROCESS_SERVICE_MEDIA_FOUNDATION:
```
andrewsavage1, line 48: Is this using an enum from //base or //copied_base or neither?

rogerzanoni: Neither, it's defined in third_party/perfetto/protos/perfetto/trace/track_event/chrome_process_descriptor.proto and current_process.h/cc use the generated headers. //base/process/current_process.cc already handles this enum value, //copied_base needs to handle it too

copied_base/base/process/current_process.cc
```diff
@@ -45,6 +45,8 @@ const char* GetNameForProcessType(CurrentProcessType process_type) {
...
+      return "Service: media.mojom.MediaFoundationServiceBroker";
```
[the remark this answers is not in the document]
rogerzanoni: I used the same name that //base_process/current_process.cc, I will keep the same name it uses to reduce the possibility of conflicts when backmerging patches to //copied_base

copied_base/base/process/current_process.cc
haozheng-cobalt: Yea, I see it's in //base too: https://source.corp.google.com/h/github/youtube/cobalt/+/main:base/process/current_process.cc;l=48?q=base%2Fprocess%2Fcurrent_process.cc%20&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

haozheng-cobalt (approved):

PR #10366, opened 2026-05-07
build: Extract evergreen_common.gn from evergreen-x64

hferreiro:
Create evergreen_common.gn with the shared evergreen settings and update the evergreen-{x64,arm64,arm-softfp,arm-hardfp-rdk} configs to import it.

Bug: 495203133


hferreiro:
@andrewsavage1 shall we do the same for linux-x64x11 and linux-x64x11-modular?

andrewsavage1 (approved):

PR #10380, opened 2026-05-07
Cobalt: Correct MediaImage JNI path for Google3

zhongqiliang:
Add conditional logic to resolve the MediaImage JNI class path.
When built for Cobalt within the Google3 environment, the Java
package for MediaImage is prefixed with "cobalt/". This ensures
the correct class is found, preventing failures in media session
APIs on Android for these specific builds. Also add the required
build flag dependency.

Bug: 508673132


Awallky (approved):
LGTM

y4vor (approved):

PR #10381, opened 2026-05-07
[WIP] Add cobalt_coat_junit_tests to Android GitHub CI

arjungm:
This change adds support for running Android JUnit tests (cobalt_coat_junit_tests) on host in GitHub CI. It introduces a new junit_test_on_host flag in platform configurations to control execution independently of standard gtests. A new step is added in on_host_tests/action.yaml to run these tests without sharding or gtest flags.

Bug: 431780245


borongc:
LGTM, leave @Awallky for the final review. Also, does `cobalt_coat_junit_tests ` the only junit tests we need? I found Chromium has others like `media_base_junit_tests`.

arjungm:
> LGTM, leave @Awallky for the final review. Also, does `cobalt_coat_junit_tests ` the only junit tests we need? I found Chromium has others like `media_base_junit_tests`.

fair point, I think I'll try and land this smaller scoped PR and then go back to find more of them and fix them if needed before adding them to the CI

Awallky (commented):
I don't see the on_host_tests being run for android at all. Am I missing where they are being run? It looks like they are skipped.

cobalt/build/testing/targets/android-arm64/test_targets.json
```diff
@@ -26,7 +26,8 @@
     "third_party/perfetto:trace_processor_minimal_smoke_tests",
     "third_party/zlib:zlib_unittests",
     "url:url_perftests",
-    "url:url_unittests"
+    "url:url_unittests",
+    "cobalt/android:cobalt_coat_junit_tests"
```
Awallky: nit: Can you put these in alphabetical order?

arjungm:
@Awallky @borongc @niranjanyardi @linxinan-chops I'm removing reviewers because I'm still iterating on this change, but I can't run CI steps without marking it "Ready For Review"

.github/actions/on_host_tests/action.yaml
```diff
@@ -28,92 +28,176 @@ runs:
       shell: bash
       run: |
         set -x
-        mkdir -p unit_test/
-        cd unit_test/
-        tar -I 'zstd -T0' -xf ../test_artifacts.tar.zstd
-    - name: Run Tests
+        tar -I 'zstd -T0' -xf test_artifacts.tar.zstd
+
+    - name: Parse Test Targets
```
oxve: There is a test target json action that should be used.

cobalt/build/testing/targets/android-arm/test_targets.json
oxve, line 1: This file should be refactored but I have some suggestions.

Since the build job processes the old test_targets list I think it should stay as-is, perhaps be renamed to better match the use (build_targets?). For the old executables section I think it makes sense to separate by test type (on_device, on_host, browser, etc).

cobalt/build/archive_test_artifacts.py
```diff
@@ -141,8 +146,23 @@ def create_archive(
       if os.path.exists(test_targets_json):
         target_deps.add(os.path.join(tar_root, 'test_targets.json'))
 
+      if is_junit_test:
+        if flatten_deps:
+          raise ValueError('Unsupported configuration for JUnit tests.')
+        wrapper_rel_path = os.path.join('bin', f'run_{target_name}')
```
oxve, line 152: The run file is in the deps for other test types. Can you investigate why (if not) it's not included here?

cobalt/build/archive_test_artifacts.py
```diff
@@ -111,12 +111,17 @@ def create_archive(
...
+
     # Paths are configured in test.gni:
     # https://github.com/youtube/cobalt/blob/main/testing/test.gni
     if use_android_deps_path:
       deps_file = os.path.join(
           out_dir, 'gen.runtime', target_path,
           f'{target_name}__test_runner_script.runtime_deps')
+      if not os.path.exists(deps_file) and is_junit_test:
+        deps_file = os.path.join(out_dir, 'gen.runtime', target_path,
+                                 f'{target_name}.runtime_deps')
```
oxve, line 124: Can you investigate using the same path as original android above? I think it should be possible to point to the same file here.

cobalt/build/archive_test_artifacts.py
```diff
@@ -141,8 +146,23 @@ def create_archive(
...
+        wrapper_rel_path = os.path.join('bin', f'run_{target_name}')
+        target_deps.add(os.path.join(out_dir, wrapper_rel_path))
+        target_deps.add(os.path.join(out_dir, 'bin/helper', target_name))
+        target_src_root_deps.update([
+            'third_party/catapult/common/py_trace_event',
+            'third_party/catapult/common/py_utils',
+            'third_party/catapult/dependency_manager',
+            'third_party/catapult/devil',
+            'third_party/catapult/third_party/six',
+        ])
```
oxve, line 161: Can you investigate if these dependencies can/should be declared in gn?

.github/actions/on_host_tests/action.yaml
```diff
@@ -28,92 +28,176 @@ runs:
...
 
-        for test_binary_path in $(cat out/${{ matrix.platform }}_${{ matrix.config }}/test_targets.json | jq -cr '.executables | join(" ")'); do
-          filename=$(basename "${test_binary_path}")
-          test_binary="${filename%%.*}"
+        if [[ -z "${EXECUTABLES}" ]]; then
+          echo "No on-host unit tests to run."
+        else
+          for executable in ${EXECUTABLES}; do
+            test_binary="${executable%%.*}"
+            test_binary_path="out/${{ matrix.platform }}_${{ matrix.config }}/${executable}"
```
oxve: From offline convo:

The executable should come from the gn target output. The test target json should be generated by a script that parses these values from the gn target.

.github/actions/upload_host_test_artifacts/action.yaml
oxve, line 1: Fold this back in to the existing upload action. I don't see this doing anything unique compared with the original action.

.github/actions/on_host_tests/action.yaml
```diff
@@ -28,92 +28,176 @@ runs:
...
-        # Don't fail if only asan fails.
-        export ASAN_OPTIONS=exitcode=0
-
-        # Make results dir available to the archiving step below.
-        results_dir="${GITHUB_WORKSPACE}/results"
-        echo "results_dir=${results_dir}" >> $GITHUB_ENV
-
-        # Test results (xml and logs) must be in a subfolder in results_dir
-        # to be picked up by the test result processor.
-        # This also prevents accidental overwriting when collecting results.
```
oxve, line 65: A bunch of comments with TODOs were removed here.

.github/actions/on_host_tests/action.yaml
```diff
@@ -28,92 +28,176 @@ runs:
...
+          done
+        fi
         echo "Finished running tests..."
         if [[ -n "${failed_suites}" ]]; then
           echo "Test suites failed:${failed_suites}"
-          # Fail the job so it's easy to retrigger.
+          exit 1
+        fi
+
+    - name: Run JUnit Tests
```
oxve, line 136: It looks to me like this step can be folded in with the previous step.

Perhaps extract command line flag generation to separate steps depending on test type.

.github/actions/build/action.yaml
```diff
@@ -110,7 +110,12 @@ runs:
           #     --json-output > "${DYN_TEST_TARGETS_JSON_FILE}"
 
 
-          test_targets_json=$(cat ${{ inputs.test_targets_json_file }} | jq -cr '.test_targets')
+          test_targets_json=$(cat ${{ inputs.test_targets_json_file }} | jq -cr '
```
oxve, line 113: As per offline dicussion:

This list should be a flat list of build_targets.

PR #10383, opened 2026-05-08
Run updater in Cobalt

yuying-y:
- Run updater in the browser. Same as in EAP.
- Fix a `DCHECK` sequence checker crash in `PersistedDataImpl`.
- Refactor `UpdaterModule` to use a singleton pattern to allow global access by the H5vcc updater APIs. Originally from PR #8182.
- Fix a `DCHECK` crash when a singleton already exists.
- Update `MakeCrxDownloaderFactory` signature in Starboard to accept an optional background downloader cache path

Bug: 479816505

### DIFF
--- a/cobalt/shell/browser/shell.cc
+++ b/cobalt/shell/browser/shell.cc
@@ -533,6 +533,15 @@ void Shell::PrimaryMainDocumentElementAvailable() {
 #if BUILDFLAG(IS_ANDROIDTV)
   starboard::StarboardBridge::GetInstance()->SetStartupMilestone(27);
 #endif
+#if BUILDFLAG(USE_EVERGREEN)
+  cobalt::updater::UpdaterModule* updater_module =
+      cobalt::updater::UpdaterModule::GetInstance();
+  if (updater_module) {
+    LOG(INFO) << "Mark the current installation as successful after the "
+                 "PrimaryMainDocumentElement is available.";
+    updater_module->MarkSuccessful();
+  }
+#endif
 }
 
 void Shell::DidFinishLoad(RenderFrameHost* render_frame_host,
@@ -862,16 +871,6 @@ WebContents* Shell::OpenURLFromTab(
   if (navigation_handle_callback && navigation_handle) {
     std::move(navigation_handle_callback).Run(*navigation_handle);
   }
-
-#if BUILDFLAG(USE_EVERGREEN)
-  cobalt::updater::UpdaterModule* updater_module =
-      cobalt::updater::UpdaterModule::GetInstance();
-  if (updater_module) {
-    // Mark the current installation as successful.
-    // TODO(b/458770751): investigate moving this to after load is finished
-    updater_module->MarkSuccessful();
-  }
-#endif
 
   return target;
 }
--- a/cobalt/shell/browser/shell.cc
+++ b/cobalt/shell/browser/shell.cc
@@ -70,6 +70,11 @@
 #include "third_party/blink/public/mojom/mediastream/media_stream.mojom.h"
 #include "third_party/blink/public/mojom/window_features/window_features.mojom.h"
 
+#if BUILDFLAG(USE_EVERGREEN)
+#include "cobalt/updater/updater_module.h"
+#include "content/public/browser/storage_partition.h"
+#endif
+
 #if BUILDFLAG(IS_ANDROID)
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 #include "starboard/android/shared/audio_permission_requester.h"
@@ -347,6 +352,18 @@ void Shell::FinishShellInitialization(Shell* shell) {
   GetPlatform()->SetSkipForTesting(shell->skip_for_testing());
 #endif  // BUILDFLAG(IS_ANDROID)
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+
+#if BUILDFLAG(USE_EVERGREEN)
+  // Create the updater module singleton.
+  auto* storage_partition =
+      raw_web_contents->GetPrimaryMainFrame()->GetStoragePartition();
+  if (storage_partition) {
+    LOG(INFO) << "Creating UpdaterModule singleton for Shell.";
+    cobalt::updater::UpdaterModule::CreateInstance(
+        storage_partition->GetURLLoaderFactoryForBrowserProcess(),
+        cobalt::updater::kDefaultUpdateCheckDelay);
+  }
+#endif
 }
 
 Shell* Shell::CreateShell(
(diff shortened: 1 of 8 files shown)


cobalt/shell/browser/shell.cc
```diff
@@ -347,6 +352,18 @@ void Shell::FinishShellInitialization(Shell* shell) {
   GetPlatform()->SetSkipForTesting(shell->skip_for_testing());
 #endif  // BUILDFLAG(IS_ANDROID)
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+
+#if BUILDFLAG(USE_EVERGREEN)
+  // Create the updater module singleton.
+  auto* storage_partition =
+      raw_web_contents->GetPrimaryMainFrame()->GetStoragePartition();
```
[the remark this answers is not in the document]
yuying-y: hallucination. Defined on line 328

cobalt/updater/updater_module.h
```diff
@@ -137,6 +151,9 @@ class UpdaterModule {
   void Finalize();
   void MarkSuccessfulImpl();
   void Update();
+
+  // Holds the single instance of UpdaterModule.
+  static base::NoDestructor<UpdaterModule>* updater_module_;
```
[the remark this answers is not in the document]
yuying-y: Following approved PR https://github.com/youtube/cobalt/pull/8182.

cobalt/updater/updater_module.cc
```diff
@@ -91,6 +91,25 @@ namespace updater {
...
+base::NoDestructor<UpdaterModule>* UpdaterModule::updater_module_ = nullptr;
+
+void UpdaterModule::CreateInstance(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    base::TimeDelta update_check_delay) {
+  DCHECK(!updater_module_)
+      << "UpdaterModule is already created. Use UpdaterModule::GetInstance() "
+         "to get the instance.";
+  updater_module_ = new base::NoDestructor<UpdaterModule>(
+      std::move(url_loader_factory), update_check_delay);
```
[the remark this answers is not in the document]
yuying-y: same as above

cobalt/updater/updater_module.cc
```diff
@@ -91,6 +91,25 @@ namespace updater {
 // The delay before the first update check.
 const base::TimeDelta kDefaultUpdateCheckDelay = base::Seconds(30);
 
+// static
+base::NoDestructor<UpdaterModule>* UpdaterModule::updater_module_ = nullptr;
```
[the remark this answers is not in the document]
yuying-y: same

cobalt/shell/browser/shell.cc
```diff
@@ -347,6 +352,18 @@ void Shell::FinishShellInitialization(Shell* shell) {
...
+  if (storage_partition) {
+    LOG(INFO) << "Creating UpdaterModule singleton for Shell.";
+    cobalt::updater::UpdaterModule::CreateInstance(
```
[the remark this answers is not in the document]
yuying-y: Only one shell is created. Following PR https://github.com/youtube/cobalt/pull/8182.

cobalt/updater/updater_module.cc
```diff
@@ -91,6 +91,28 @@ namespace updater {
...
+  }
+  updater_module_ = new base::NoDestructor<UpdaterModule>(
+      std::move(url_loader_factory), update_check_delay);
+}
+
+UpdaterModule* UpdaterModule::GetInstance() {
+  DCHECK(updater_module_) << "UpdaterModule is not created yet, and cannot be "
+                             "retrieved by UpdaterModule::GetInstance().";
+  return updater_module_->get();
+}
```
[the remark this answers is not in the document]
yuying-y: Removed NoDestructor. The rest of the comment is on an outdated snapshot.

cobalt/updater/updater_module.cc
```diff
@@ -91,6 +91,28 @@ namespace updater {
...
 const base::TimeDelta kDefaultUpdateCheckDelay = base::Seconds(30);
 
+// static
+UpdaterModule* UpdaterModule::updater_module_ = nullptr;
+
+void UpdaterModule::CreateInstance(
+    scoped_refptr<network::SharedURLLoaderFactory> url_loader_factory,
+    base::TimeDelta update_check_delay) {
+  if (updater_module_) {
+    LOG(WARNING)
```
andrewsavage1, line 101: Should we consider converting this LOG and the below DCHECK to both be CHECKs to ensure we catch this happening in development or are the edge cases for this complex enough that we want to be forgiving here?

cobalt/updater/updater_module.h
```diff
@@ -120,6 +125,13 @@ class UpdaterModule {
   void MarkSuccessful();
 
  private:
+  // Private constructor and destructor to enforce singleton pattern.
+  explicit UpdaterModule(scoped_refptr<network::SharedURLLoaderFactory>,
+                         base::TimeDelta update_check_delay);
+  ~UpdaterModule();
+
+  // TODO: b/454440974 Investigate whether singleton is necessary.
```
andrewsavage1: Can we create a separate bug for this? This bug seems unrelated

andrewsavage1 (commented):
Mostly seems good, just minor qs/nits

cobalt/updater/updater_module.cc
yuying-y: Changed both to LOG. We'd like to be forgiving here. The caller does the check.

cobalt/updater/updater_module.h
yuying-y: It's outdated. Singleton is the solution to allow access by both the browser client and h5vcc APIs. Removed.

andrewsavage1: Why is this different from c26? We are able to share instances between h5vcc APIs and the browser client, it would just require some work to connect mojo and the embedder appropriately

yuying-y: Using singleton is the approved solution in c26. The code is the same as c26. The original PR https://github.com/youtube/cobalt/pull/8182 was noted in the commit message. If you'd like to suggest another solution, it can happen in a follow-up PR.

andrewsavage1: My understanding was that a singleton was an acceptable solution because we wanted something quickly, but we filed a TODO in #8182 to investigate if we could move away from using a singleton. I don't think it has to be high priority and I don't think you necessarily have to take it on, but I think we should still file a bug noting that this is something we want to look at.

yuying-y: Done

andrewsavage1 (approved):

cobalt/updater/updater_module.h
andrewsavage1: Great, thank you

PR #10385, opened 2026-05-08
build: Deduplicate Cobalt GN platform configs

hferreiro:
Follow-up from https://github.com/youtube/cobalt/pull/10366 to factor shared GN args out of architecture-specific Cobalt config files.

Introduce shared Android, Linux, Linux modular, and Evergreen config files so
architecture-specific args only set target-specific values.

Move `enable_remoting` into common.gn so all Cobalt configs consistently disable
Chrome Remote Desktop, including AndroidTV, which previously differed from the other platforms.

Bug: 495203133


hferreiro:
@haozheng-cobalt can you confirm the `enable_remoting` change in AndroidTV?

andrewsavage1 (approved):

PR #10390, opened 2026-05-08
starboard/rdk: fix navigator.language to reflect device presentation language

denisyuji:
Currently, `navigator.language` is hardcoded to `en-us,en` in `GetShellLanguage()` because the WPEFramework process starts with `LANG=C` — systemd does not source `/etc/locale.conf` for service units, so localectl changes had no effect on the running process.

Fix:
- Added an RDK override of `SbSystemGetLocaleId()` that owns locale resolution end-to-end: it first queries `UserSettings.getPresentationLanguage` via Thunder JSON-RPC (the durable, always-fresh source), then falls back to `setlocale(LC_MESSAGES)`/`$LC_ALL`/`$LC_MESSAGES`/`$LANG`, normalising POSIX locales to BCP-47. Returns `""` (never `NULL`) when nothing is found, honoring the Starboard API's non-null contract.
- Added `UserSettings::GetPresentationLanguage()` to `rdkservices.h/cc` using the existing `ServiceLink` pattern.
- `GetShellLanguage()` in `shell_content_browser_client.cc` now simply defers to `SbSystemGetLocaleId()`, falling back to `"en-us,en"` only if Starboard reports nothing — removing duplicated normalisation logic from platform-agnostic Cobalt code.
- `CobaltImplementation::Configure()` no longer fetches the language itself; it only sets `$LANG` when the plugin configuration provides an explicit override, since `SbSystemGetLocaleId()` now reads UserSettings directly.
- Updated `BUILD.gn` to use the new RDK `system_get_locale_id_bcp47.cc` instead of the posix implementation.

Bug: 503631204

### DIFF
--- a/cobalt/shell/browser/shell_content_browser_client.cc
+++ b/cobalt/shell/browser/shell_content_browser_client.cc
@@ -94,6 +94,7 @@
 #include "services/network/public/mojom/network_context.mojom.h"
 #include "services/network/public/mojom/network_service.mojom.h"
 #include "services/network/public/mojom/permissions_policy/permissions_policy_feature.mojom-shared.h"
+#include "starboard/system.h"
 #include "third_party/blink/public/common/features.h"
 #include "third_party/blink/public/common/user_agent/user_agent_metadata.h"
 #include "third_party/blink/public/common/web_preferences/web_preferences.h"
@@ -224,6 +225,13 @@ SharedState& GetSharedState() {
 }
 
 std::string GetShellLanguage() {
+  // Defer to the Starboard layer, which is the single source of truth for
+  // the platform's presentation language and already returns a normalised
+  // BCP-47 tag.
+  const char* locale = SbSystemGetLocaleId();
+  if (locale && *locale) {
+    return locale;
+  }
   return "en-us,en";
 }
 
--- a/starboard/contrib/rdk/plugin/CobaltImplementation.cpp
+++ b/starboard/contrib/rdk/plugin/CobaltImplementation.cpp
@@ -291,6 +291,9 @@ private:
       }
 
       if (config.Language.IsSet() == true) {
+        // Explicit config override always wins; SbSystemGetLocaleId() reads
+        // UserSettings directly when LANG is unset, so no fallback is
+        // needed here.
         string lang = config.Language.Value();
         Core::SystemInfo::SetEnvironment(_T("LANG"), lang);
       }
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/rdkservices.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/rdkservices.cc
@@ -1626,6 +1626,17 @@ public:
     needs_refresh_ = true;
     subscriptions_.clear();
   }
+
+  bool GetPresentationLanguage(std::string& out) {
+    Core::JSON::String result;
+    uint32_t rc = link_.Get(kDefaultTimeoutMs, "getPresentationLanguage", result);
+    if (Core::ERROR_NONE != rc) {
+      SB_LOG(ERROR) << "UserSettings.getPresentationLanguage failed, rc=" << rc;
+      return false;
+    }
+    out = result.Value();
+    return !out.empty();
+  }
 };
 
 SB_ONCE_INITIALIZE_FUNCTION(UserSettingsImpl, GetUserSettings);
@@ -1769,6 +1780,10 @@ bool DeviceInfo::GetAudioConfiguration(int index, SbMediaAudioConfiguration* out
 
 bool DeviceInfo::GetBrandName(std::string& out) {
   return GetDeviceInfo()->GetBrandName(out);
+}
+
+bool UserSettings::GetPresentationLanguage(std::string& out) {
+  return GetUserSettings()->GetPresentationLanguage(out);
 }
 
 void TeardownJSONRPCLink() {
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
@@ -1,0 +1,99 @@
+//
+// Copyright 2020 Comcast Cable Communications Management, LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "starboard/system.h"
+
+#include <locale.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <string>
+
+#include "starboard/common/log.h"
+#include "third_party/starboard/rdk/shared/rdkservices.h"
+
+namespace {
+
+bool IsValidLocaleIdBcp47(const char* id) {
+  if (!id || *id == '\0') {
+    return false;
+  }
+  return strcmp(id, "C") != 0 && strcmp(id, "POSIX") != 0;
+}
+
+// Normalise a POSIX locale string to a BCP-47 language tag in place.
+// e.g. "fr_FR.UTF-8" → "fr-FR", "en_US@euro" → "en-US".
+void NormalizePosixLocaleToBcp47(std::string& locale) {
+  size_t at = locale.find('@');
+  if (at != std::string::npos) {
+    locale.resize(at);
+  }
+  size_t dot = locale.find('.');
+  if (dot != std::string::npos) {
+    locale.resize(dot);
+  }
+  for (char& c : locale) {
+    if (c == '_') {
+      c = '-';
+    }
+  }
+}
+
+}  // namespace
+
+// Cache the BCP-47 locale string so SbSystemGetLocaleId() returns a stable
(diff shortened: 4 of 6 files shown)


starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
```diff
@@ -0,0 +1,96 @@
...
+  if (cached[0] == '\x01')
+    return NULL;
+  if (cached[0])
+    return cached;
+
+  // 1. Check setlocale (works if the process environment was set before init).
+  const char* id = setlocale(LC_MESSAGES, NULL);
+  if (IsValidLocaleId(id)) {
+    strncpy(cached, id, sizeof(cached) - 1);
+    return cached;
```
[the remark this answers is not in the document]
denisyuji: fixed

cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -224,6 +224,26 @@ SharedState& GetSharedState() {
...
+    if (at != std::string::npos)
+      lang.erase(at);
+    if (!lang.empty() && lang != "C" && lang != "POSIX") {
+      for (char& c : lang) {
+        if (c == '_') c = '-';
+      }
+      return lang;
+    }
+  }
   return "en-us,en";
```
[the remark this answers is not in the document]
denisyuji: fixed

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/rdkservices.cc
```diff
@@ -1776,6 +1776,18 @@ bool DeviceInfo::GetBrandName(std::string& out) {
   return GetDeviceInfo()->GetBrandName(out);
 }
 
+bool UserSettings::GetPresentationLanguage(std::string& out) {
+  Core::JSON::String result;
+  uint32_t rc = ServiceLink(kUserSetingsCallsign)
```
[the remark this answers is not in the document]
denisyuji: fixed

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
```diff
@@ -0,0 +1,96 @@
...
+  if (f) {
+    char line[128];
+    while (fgets(line, sizeof(line), f)) {
+      if (strncmp(line, "LANG=", 5) == 0) {
+        char* val = line + 5;
+        // Trim trailing whitespace / newline.
+        size_t len = strlen(val);
+        while (len > 0 && (val[len - 1] == '\n' || val[len - 1] == '\r' ||
+                            val[len - 1] == ' '))
+          val[--len] = '\0';
```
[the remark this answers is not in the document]
denisyuji: fixed

cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -224,6 +227,41 @@ SharedState& GetSharedState() {
...
 std::string GetShellLanguage() {
+  // Look up the process locale and normalise to a BCP-47 language tag.
+  // On RDK, CobaltImplementation::Configure() sets $LANG from UserSettings
+  // before the Cobalt thread starts, so getenv("LANG") reflects the device
+  // presentation language. The /etc/locale.conf fallback lives in the
+  // Starboard layer (system_get_locale_id_bcp47.cc) and is not needed here.
+  auto is_valid = [](const char* s) {
+    return s && *s && strcmp(s, "C") != 0 && strcmp(s, "POSIX") != 0;
+  };
+  auto normalize = [](const char* s) {
```
andrewsavage1: I recommend using a string_view for some of this for less copying

```
auto normalize = [](std::string_view s) {
  auto end = s.find_first_of(".@");
  std::string out(s.substr(0, end));
  
  for (char& c : out) {
    if (c == '_') {
      c = '-';
    }
  }
  
  return out;
};
```

cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -224,6 +227,41 @@ SharedState& GetSharedState() {
 }
 
 std::string GetShellLanguage() {
+  // Look up the process locale and normalise to a BCP-47 language tag.
+  // On RDK, CobaltImplementation::Configure() sets $LANG from UserSettings
```
andrewsavage1: Avoid referencing platforms in shared cobalt code

andrewsavage1 (commented):
Letting someone else review the rdk-specific code as I'm less familiar with it

cobalt/shell/browser/shell_content_browser_client.cc
denisyuji: Makes sense, updated to take a `string_view` and only copy the trimmed portion

cobalt/shell/browser/shell_content_browser_client.cc
denisyuji: Good catch. I'll strip the RDK-specific explanation from the comment

.claude/settings.local.json
```diff
@@ -0,0 +1,7 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(gh pr *)"
+    ]
+  }
+}
```
niranjanyardi: hey ! @denisyuji  can you clean up this file and other claude related files ?

denisyuji: Sorry about this, I added all the untracked files in my repo in the last rebase. Fixed it

niranjanyardi (commented):
@jonastsai  can you pelase review the RDK side changes ?

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
```diff
@@ -0,0 +1,116 @@
...
+      NormalizePosixLocaleToBcp47(cached);
+      return cached;
+    }
+  }
+
+  // 3. RDK-specific fallback: read LANG= from /etc/locale.conf (written by
+  //    localectl or the platform locale manager). WPEFramework is typically
+  //    started with LANG=C so this file is not reflected in the process
+  //    environment even after localectl set-locale.
+  FILE* f = fopen("/etc/locale.conf", "r");
```
jonastsai: I cannot find /etc/locale.conf file on the RDK6 image.

starboard/contrib/rdk/plugin/CobaltImplementation.cpp
```diff
@@ -293,6 +295,13 @@ class CobaltImplementation:
       if (config.Language.IsSet() == true) {
         string lang = config.Language.Value();
         Core::SystemInfo::SetEnvironment(_T("LANG"), lang);
+      } else {
+        // Fall back to the platform's presentation language from UserSettings.
+        std::string lang;
+        if (third_party::starboard::rdk::shared::UserSettings::GetPresentationLanguage(lang)) {
```
jonastsai: The plug-in wrapper(like this file) should not get language setting from RDK starboard and set the value into ENV for the RDK starboard to read. The RDK starboard can do that without plugin wrapper.

You can implement the get language all in the SbSystemGetLocaleId().

if the LANG env will also be read by [shell_content_browser_client.cc‎](https://github.com/youtube/cobalt/pull/10390/changes#diff-fc39ec5b698bf5c5a4742b1b1fa6924a0b7c148b865c549eda1a7e05e87d77a7), we should consider to move GetPresentationLanguage() to plugin wrapper.

cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -224,6 +227,30 @@ SharedState& GetSharedState() {
 }
 
 std::string GetShellLanguage() {
```
jonastsai, line 227: what is the purpose of this function and SbSystemGetLocaleId()? Which one decide the final language displayed on the cobalt?

jonastsai (commented):
From the issue description, this issue doesn't existing in 26.eap. What is the difference between these two branch?

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
denisyuji: @jonastsai , this was used as a last resource. I changed it to just return a empty string, and the GetShellLanguage() will default to "en-us,en"

starboard/contrib/rdk/plugin/CobaltImplementation.cpp
denisyuji: You are right. I pulled that logic out of the plugin and put it straight into SbSystemGetLocaleId(), so Starboard owns it completely. Plugin only touches $LANG now if there's an explicit config override.

cobalt/shell/browser/shell_content_browser_client.cc
denisyuji: You are right. I changed the GetShellLanguage() now for just call SbSystemGetLocaleId() and falls back to "en-us,en" if it's empty

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_locale_id_bcp47.cc
```diff
@@ -0,0 +1,98 @@
...
+      strncpy(cached, id, sizeof(cached) - 1);
+      NormalizePosixLocaleToBcp47(cached);
+      return cached;
+    }
+  }
+
+  SB_LOG(WARNING) << "SbSystemGetLocaleId: no valid locale found";
+  cached[0] = '\x01';
+  return "";
+}
```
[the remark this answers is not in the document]
denisyuji: Good catch. The lazy-init pattern with the raw static char cached[64] buffer was indeed a data race on first concurrent calls. Fixed it

jonastsai:
Let's pause for a while as the language testing is passed now without this change. We need to check more about current status.

denisyuji:
Closing it since it is possible to fix this issue by adjusting dab-adapter's Cobalt launch

PR #10391, opened 2026-05-08
nplb: Migrate player test fixtures and suites to std::chrono

osagie98:
Transitions `SbPlayerTestFixture` and all dependent NPLB player tests
(`player_get_media_time_test.cc`, `player_write_sample_test.cc`,
`multiple_player_test.cc`, `player_get_audio_configuration_test.cc`,
`vertical_video_test.cc`) to standard C++ `<chrono>` types.

Updates fixture interfaces to accept and return `std::chrono` durations
(`microseconds`, `milliseconds`, `seconds`), eliminating raw `int64_t`
time manipulations and making test code more concise and type-safe.

#vibe-coded

Bug: 319926976


starboard/nplb/player_get_media_time_test.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "starboard/common/time.h"
 #include "starboard/nplb/player_test_fixture.h"
+#include "starboard/shared/starboard/media/media_time_delta.h"
```
kjyoun: is it allowed for nplb to use header file deep inside starboard/ ? other than common/ and testing/
(i.e. starboard/shared/starboard/media)

AFAIK, we want decoupling between nplb and starboard. so that it even has different namespaces. (starboard:: vs nplb::)

Even though there are some cases, I think that it is not recommended practice.

If this class is used only in nplb, why not defining this class under nplb?

starboard/shared/starboard/media/media_time_delta.h
```diff
@@ -0,0 +1,140 @@
...
+#define STARBOARD_SHARED_STARBOARD_MEDIA_MEDIA_TIME_DELTA_H_
+
+#include <cstdint>
+
+namespace starboard {
+
+// Represents a duration or elapsed amount of media time.
+// Internally represented in nanoseconds to provide high precision
+// while maintaining a small footprint (8 bytes).
+class MediaTimeDelta {
```
kjyoun: can't we use chrono instead?

https://en.cppreference.com/cpp/chrono/duration

starboard/shared/starboard/media/media_time_delta.h
```diff
@@ -0,0 +1,140 @@
...
+ public:
+  static constexpr int64_t kNanosecondsPerMicrosecond = 1000;
```
kjyoun: nit: use digit seperator for readability e.g)  1'000

I think we don't need these constance once we use digit seperator.
and digit separator is even more readable.

```
static constexpr MediaTimeDelta FromMicroseconds(int64_t usecs) {
    return MediaTimeDelta(usecs * kNanosecondsPerMicrosecond);
  }

vs

static constexpr MediaTimeDelta FromMicroseconds(int64_t usecs) {
    return MediaTimeDelta(msecs * 1'000'00);
  }
```

starboard/nplb/player_get_media_time_test.cc
osagie98: It may not, you're right that the placement is important. I originally put this class under starboard/common, which is allowed for nplb, but this class should only be used for media related files (starboard/player/filter, starboard/android/shared, etc). I may need to move it to a spot that makes sense for nplb and those other locations

starboard/shared/starboard/media/media_time_delta.h
osagie98: Chrono would be useful here, however it's [banned](https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++-features.md#chrono_banned) in Chromium, if that doesn't also apply to Starboard we can use it instead.

starboard/nplb/player_get_media_time_test.cc
osagie98: There are a [few examples](https://github.com/search?q=repo%3Ayoutube%2Fcobalt+path%3A%2F%5Estarboard%5C%2Fnplb%5C%2F%2F+starboard%2Fshared%2Fstarboard&type=code) of some files under `starboard/shared/starboard/` used in nplb

starboard/shared/starboard/media/media_time_delta.h
osagie98: Done

starboard/shared/starboard/media/media_time_delta.h
kjyoun: Chromium has alternative for duration, while we dont

starboard are already using it: https://source.corp.google.com/search?q=std::chrono&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

or you can replace existing usage of std::chrono with the class.

starboard/nplb/player_get_media_time_test.cc
kjyoun: As I mentioned in the original comment, 

> Even though there are some cases, I think that it is not recommended practice.

Now, starboard layer is almost solely for media impl. Other starboard APIs are replaced with Posix APIs.
border between starboard and media is very vague (gap is very small)

I would put this code into starboard/common, if it is used both in starboard:: and npbl::  
and it would make more sense, if we replace current chrono:: usage with this class.

CC: @y4vor for nplb accessing deep inside starboard layer

kjyoun: FYI, this is what I heard from @yavor, while doing namespace refactoring
> we want decoupling between nplb and starboard

As a result, I put them in different namespaces. (starboard:: vs nplb::)

starboard/nplb/player_get_media_time_test.cc
```diff
@@ -47,13 +49,13 @@ TEST_P(SbPlayerGetMediaTimeTest, SunnyDay) {
...
-    samples.AddAudioSamples(
-        0, player_fixture.ConvertDurationToAudioBufferCount(kDurationToPlay));
+    samples.AddAudioSamples(0, player_fixture.ConvertDurationToAudioBufferCount(
+                                   kDurationToPlay.InMicroseconds()));
     samples.AddAudioEOS();
   }
   if (player_fixture.HasVideo()) {
-    samples.AddVideoSamples(
-        0, player_fixture.ConvertDurationToVideoBufferCount(kDurationToPlay));
+    samples.AddVideoSamples(0, player_fixture.ConvertDurationToVideoBufferCount(
```
kjyoun: can you update `ConvertDurationToVideoBufferCount` to accept `MediaTimeDelta` directly?

For now, you can add helper method.
```
// Deprecated. Use ConvertDurationToVideoBufferCount(MediaTimeDelta duration) instead
// TODO: b/319926976 - Remove this once migration is completed
int ConvertDurationToVideoBufferCount(int64_t duration) const;

int ConvertDurationToVideoBufferCount(MediaTimeDelta duration) const; 
```

starboard/shared/starboard/media/media_time_delta.h
```diff
@@ -0,0 +1,125 @@
...
+#define STARBOARD_SHARED_STARBOARD_MEDIA_MEDIA_TIME_DELTA_H_
+
+#include <cstdint>
+
+namespace starboard {
+
+// Represents a duration or elapsed amount of media time.
+// Internally represented in nanoseconds to provide high precision
+// while maintaining a small footprint (8 bytes).
+class MediaTimeDelta {
```
kjyoun: For class name, I would recommend "Duration"
- it is already used across many libraries: std::chrono::duration, absl::Duration
- This class itself does not have any media-related characteristics. This class can be used anywhere.

kjyoun (commented):
I left some comments before your last commit, just in case that you might have missed them

starboard/shared/starboard/media/media_time_delta.h
osagie98: I can implement `MediaTimeDelta/Duration` with `std::chrono` internally, which allows us to maintain a friendly interface and avoid being so verbose with `chrono` directly. It'd also let us implement additional custom methods down the line

starboard/shared/starboard/media/media_time_delta.h
```diff
@@ -0,0 +1,125 @@
...
+    time_ns_ -= other.time_ns_;
+    return *this;
+  }
+
+  // Unary negation.
+  constexpr MediaTimeDelta operator-() const {
+    return MediaTimeDelta(-time_ns_);
+  }
+
+  // Scalar multiplication and division.
```
borongc: Should we consider the floating-point scalar operators for scaling durations (e.g., for playback rate adjustments, which uses double type)?

starboard/shared/starboard/media/media_time_delta.h
kjyoun: I think choices are
- use std::chrono::duration
- or define starboard::Duration and use it instead.

Implementing `starboard::Duration` using `std::chrono::duration` internally feels like an awkward hybrid approach. 

If we do go with `starboard::Duration`, we will definitely need to provide explicit conversion methods to std::chrono (similar to how absl::Duration does [1]) so we can easily interface with standard library APIs that require it (such as std::condition_variable::wait_for).

---
[1] e.g.) 
http://google3/third_party/absl/time/time.h;l=689-715;rcl=913847745

osagie98:
Ignore the earlier review ping for the non-media groups, I somehow added a commit that combined the last 58 commits in the codebase. Removed it now

starboard/nplb/player_get_media_time_test.cc
osagie98: Removed the class in favor of `std::chrono`

starboard/shared/starboard/media/media_time_delta.h
osagie98: Used `std::chrono`

starboard/nplb/player_get_media_time_test.cc
```diff
@@ -60,29 +65,31 @@ TEST_P(SbPlayerGetMediaTimeTest, SunnyDay) {
...
-                << std::abs(end_media_time - kDurationToPlay) << ".";
-  SB_DLOG(INFO) << "The expected total playing time should be "
-                << kDurationToPlay << ", the actual playing time is "
-                << end_system_time - start_system_time + start_media_time
-                << ", with difference "
-                << std::abs(end_system_time - start_system_time +
-                            start_media_time - kDurationToPlay)
-                << ".";
+  SB_DLOG(INFO) << "The expected media time should be "
+                << kDurationToPlay.count() << ", the actual media time is "
```
kjyoun: can you add time unit (sec or msec or usec) ?

starboard/nplb/player_write_sample_test.cc
```diff
@@ -216,29 +221,31 @@ TEST_P(SbPlayerWriteSampleTest, PartialAudio) {
...
-                << ", with difference "
-                << std::abs(end_system_time - start_system_time +
-                            start_media_time - kDurationToPlay)
-                << ".";
+  SB_DLOG(INFO) << "The expected media time should be "
+                << kDurationToPlay.count() << ", the actual media time is "
+                << end_media_time.count() << ", with difference "
+                << std::abs((end_media_time - kDurationToPlay).count()) << ".";
+  SB_DLOG(INFO)
+      << "The expected total playing time should be " << kDurationToPlay.count()
```
kjyoun: time unit ?

starboard/nplb/player_test_fixture.cc
```diff
@@ -585,10 +595,10 @@ void SbPlayerTestFixture::WriteEndOfStream(SbMediaType media_type) {
   }
 }
 
-void SbPlayerTestFixture::WaitAndProcessNextEvent(int64_t timeout) {
+void SbPlayerTestFixture::WaitAndProcessNextEvent(microseconds timeout) {
   SB_CHECK(thread_checker_.CalledOnValidThread());
 
-  auto event = callback_event_queue_.GetTimed(timeout);
+  auto event = callback_event_queue_.GetTimed(timeout.count());
```
kjyoun: can you add helper method to to accept chrono and use it?

https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/common/queue.h;l=77;drc=4fc306eed3e36a2605669871aa4aabd16da18bff


This is what Gemini suggests.
```
  // Deprecated. Use GetTimed(std::chrono::duration<Rep, Period> duration) 
  T GetTimed(int64_t duration_us) {
    return GetTimed(std::chrono::microseconds(duration_us));
  }

  // NEW: Accepting any type-safe std::chrono duration directly
  template <typename Rep, typename Period>
  T GetTimed(std::chrono::duration<Rep, Period> duration) {
    std::unique_lock lock(mutex_);
```

starboard/nplb/player_get_media_time_test.cc
```diff
@@ -21,6 +23,9 @@ namespace nplb {
 namespace {
 
 using ::starboard::FakeGraphicsContextProvider;
+using std::chrono::microseconds;
```
kjyoun: suggestion

WDYT about defining these common time aliases centrally in a shared starboard header?

then callsites only have to include this header when using chrono duration.
== We can centralize the #include <chrono> dependency inside a single common header instead of scattering it across individual test files.

`starboard/common/duration.h`
```cpp
#include <chrono>

namespace starboard {
using microseconds = std::chrono::microseconds;
using milliseconds = std::chrono::milliseconds;
}  // namespace starboard

Comment (osagie98) [reply starboard/nplb/player_get_media_time_test.cc]: Good idea, done

File: starboard/nplb/player_test_fixture.cc (hunk 9 above)
Comment (osagie98) [reply starboard/nplb/player_test_fixture.cc]: done

File: starboard/common/duration.h [hunk 11]
```diff
@@ -0,0 +1,33 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef STARBOARD_COMMON_DURATION_H_
+#define STARBOARD_COMMON_DURATION_H_
+
+#include <chrono>
+
+using std::chrono::duration_cast;
```
kjyoun, line 20: Can you put this alias inside starboard namespace  ?
(Otherwise, they polute global namespace.)

type alias under starboard namespace may need more codes in nplb namespace, but it is intended
With starboard:: prefix, we can clearly they come from starboard::

starboard/common/duration.h
```diff
@@ -0,0 +1,33 @@
...
+using std::chrono::microseconds;
+using std::chrono::milliseconds;
+using std::chrono::nanoseconds;
+using std::chrono::seconds;
+
+using std::chrono_literals::operator""h;
```
kjyoun, line 26: using literal in header file is not recommended. 

Gemini recommends to use std::chrono_literals directly in .cc file.


FYI, here is what Gemini told

---

Adding `using` declarations to the global namespace in a header file is a highly discouraged practice (header pollution). It forces those symbols into the global scope of every source file that includes the header, increasing the risk of naming collisions and defeating the purpose of namespaces.

Here is a breakdown of the best practices regarding moving them inside the `starboard::` namespace and using aliases in headers.

### 1. Adding them inside the `starboard::` namespace

Moving them inside the `starboard::` namespace is a **significant improvement** and is much better than leaving them in the global namespace. 

```cpp
namespace starboard {
using std::chrono::microseconds;
using std::chrono::milliseconds;
// ...
}  // namespace starboard
```

This scopes the aliases to `starboard::microseconds`, etc. 
* **Pros:** It prevents global namespace pollution. Code inside the `starboard` namespace can use `microseconds` directly, while code outside must use `starboard::microseconds` or explicitly import it.
* **Cons:** It still introduces symbols into the `starboard` namespace that might conflict with other things in that namespace in the future, though this is much less risky than global pollution.

---

### 2. Is it still not recommended to have aliases in header files?

It depends on the **type of alias** and **what** you are aliasing.

#### A. Type Aliases (Acceptable inside a namespace)
If the goal is to define "Starboard's standard representation of time" and you want to guarantee it is consistent across the codebase, using C++11 type aliases is generally acceptable in headers, provided they are inside a namespace:

```cpp
namespace starboard {
using microseconds = std::chrono::microseconds;
using milliseconds = std::chrono::milliseconds;
}  // namespace starboard
```
*Note: Type aliases (`using new_name = old_name`) are generally preferred in modern C++ for defining API types over using-declarations.*

#### B. Using Declarations for Standard Functions/Classes (Discouraged in headers)
Using declarations like `using std::chrono::duration_cast;` inside a header (even inside a namespace) should generally be avoided. It is cleaner to let headers use the fully qualified name (`std::chrono::duration_cast`) and let `.cc` files use `using` declarations locally if they want to avoid verbosity.

#### C. Literal Operators (Highly Discouraged in headers)
Bringing literal operators (like `operator""ms`) into a header, even inside a namespace, is **strongly discouraged**.

```cpp
// NOT recommended in headers
using std::chrono_literals::operator""ms; 
```

Literal operators are designed to be imported in the narrowest scope possible to avoid confusing syntax issues. 
* **Best Practice:** Developers who want to use literals (like `10ms` or `5s`) should explicitly write `using namespace std::chrono_literals;` (or `using namespace std::literals::chrono_literals;`) **inside their `.cc` files** or **inside specific function scopes** where they are used.
* Exposing them in headers can lead to unexpected parsing errors or conflicts if other libraries define similar literals.

---

### Recommendation

A cleaner and more robust design for `starboard/common/duration.h` would be:

1. **Define Type Aliases** for the core duration types inside the `starboard` namespace. This establishes them as part of the Starboard API.
2. **Do NOT include the literal operators** or utility functions like `duration_cast` as aliases in the header. Force callsites to use `std::chrono::duration_cast` or import `std::chrono_literals` locally.

```cpp
#ifndef STARBOARD_COMMON_DURATION_H_
#define STARBOARD_COMMON_DURATION_H_

#include <chrono>

namespace starboard {

// Define Starboard's standard duration type aliases.
using nanoseconds = std::chrono::nanoseconds;
using microseconds = std::chrono::microseconds;
using milliseconds = std::chrono::milliseconds;
using seconds = std::chrono::seconds;

}  // namespace starboard

#endif  // STARBOARD_COMMON_DURATION_H_
```

And then, in a `.cc` file that needs to use literals:

```cpp
#include "starboard/common/duration.h"

namespace starboard {
namespace {

void SomeFunction() {
  using namespace std::chrono_literals; // Import literals locally
  microseconds timeout = 500ms;
  // ...
}

}  // namespace
}  // namespace starboard
```

PR #10395, opened 2026-05-08
Use the latest siso to build Cobalt

linxinan-yt:
Use the latest siso to support remote build on mac.

1. The new siso support login via gcloud, therefore add corresponding steps before cobalt build.
2. Allow macOS to download rewrapper configs via CIPD

Bug: 507872651


y4vor (approved):

buildtools/reclient_cfgs/configure_reclient_cfgs.py
```diff
@@ -271,10 +271,6 @@ def main():
         if not GenerateReproxyCfg(args.reproxy_cfg_template, args.rbe_instance,
                                   rbe_project, args.use_luci_auth_credshelper):
             return 1
-# COBALT Disables CIPD fetch of rewrapper configs in favor of local rewrapper configs.
```
briantting, line 274: Happy to see this removed, this Cobalt specific customization was always a pain.
- we had needed `download_remoteexec_cfg` to be true
- but wanted to use local cfg files and avoid CIPD
- and the skip_remoteexec_cfg_fetch flag did not exist previously

This is deprecated twice over with the switch to siso which mostly doesn't need `download_remoteexec_cfg` to be true and with the addition of the skip_remoteexec_cfg_fetch flag.

Mac however needs `download_remoteexec_cfg` to be true and wants to use remote cfg files and CIPD so this deprecated customization is a hinderance.

briantting (approved):

buildtools/reclient_cfgs/configure_reclient_cfgs.py
linxinan-yt: Thanks for context!

PR #10402, opened 2026-05-11
aosp: Include platform base path keys for Starboard

hferreiro:
On AOSP, base_paths.h failed to include Android-specific keys. Rework the includes, so that platform-specific includes are always used, in addition to the Starboard one.

Bug: 495203133

### DIFF
--- a/base/base_paths.h
+++ b/base/base_paths.h
@@ -10,9 +10,6 @@
 
 #include "build/build_config.h"
 
-#if BUILDFLAG(IS_STARBOARD)
-#include "base/base_paths_starboard.h"
-#else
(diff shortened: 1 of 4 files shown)


base/base_paths.h
andrewsavage1, line 22: I understand why we would want this for //copied_base, but why is it necessary for //base too?

hferreiro: Yeah, it's not needed. I did it for consistency. It's also probably more resilient to future rebases unless we want to explicitly catch additions to the POSIX header.

hferreiro: Reverted the changes in base/.

andrewsavage1 (approved):

PR #10406, opened 2026-05-11
starboard: Extract helper for  media_codec_video_decoder and unit tests

kjyoun:
Refactor `MediaCodecVideoDecoder` by extracting several utility functions
into a new, dedicated `media_codec_video_decoder_helpers` library. This
improves code modularity and allows for direct unit testing of these helpers.

Added comprehensive unit tests in `media_codec_video_decoder_helpers_test.cc`
covering various scenarios for each helper function.

Issue: 512137801


starboard/android/shared/media_codec_video_decoder_helpers.cc
```diff
@@ -0,0 +1,212 @@
...
+
+  Size max_size = {height * frame_size.width / frame_size.height, height};
+  SB_LOG(INFO) << "Inferred max size=" << max_size
+               << ", frame resolution=" << frame_size;
+  return max_size;
+}
+
+bool Equal(const SbMediaMasteringMetadata& lhs,
+           const SbMediaMasteringMetadata& rhs) {
+  return std::memcmp(&lhs, &rhs, sizeof(SbMediaMasteringMetadata)) == 0;
```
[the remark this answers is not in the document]
kjyoun: will stick with existing codes

starboard/android/shared/BUILD.gn
```diff
@@ -313,6 +313,7 @@ static_library("starboard_platform") {
   ]
 
   deps = [
+    ":media_codec_video_decoder_helpers",
```
osagie98, line 316: I think you may be able to combine the helpers with the rest of the source files instead of making a separate module. As far as I can tell, this library will always be built, might as well simplify

starboard/android/shared/media_codec_video_decoder_helpers.h
```diff
@@ -0,0 +1,50 @@
...
+#include <string>
+
+#include "starboard/common/size.h"
+#include "starboard/configuration.h"
+#include "starboard/decode_target.h"
+#include "starboard/media.h"
+
+namespace starboard {
+
+bool IsSoftwareDecodeRequired(const std::string& max_video_capabilities);
```
osagie98: nit: IsSoftwareDecode**r**Required sounds a bit better, unless we already use IsSoftwareDecode somewhere else

osagie98 (approved):

starboard/android/shared/BUILD.gn
kjyoun: I agree this differs from the current convention, but I'd prefer to keep it as a separate module to avoid contributing to the giant monolithic `starboard_platform` target.

Keeping it separate enforces strict dependency scoping (ensuring these helpers only depend on  common  and media_util and makes the structural relationship with  media_codec_video_decoder  much more evident. I think this sets a healthier  precedent for modularizing Starboard going forward.

starboard/android/shared/media_codec_video_decoder_helpers.h
kjyoun: Done

PR #10408, opened 2026-05-11
Remove unused Starboard constants

y4vor:
Bug: 502581618

### DIFF
--- a/starboard/nplb/posix_compliance/posix_fdopendir_test.cc
+++ b/starboard/nplb/posix_compliance/posix_fdopendir_test.cc
@@ -58,7 +58,7 @@ TEST(PosixFdopendirTest, ManySunnyDay) {
   EXPECT_FALSE(path.empty());
   EXPECT_TRUE(FileExists(path.c_str())) << "Filename is " << path.c_str();
 
-  const int kMany = kSbFileMaxOpen / 2;  // Avoid hitting fd limits.
+  const int kMany = 64;
   std::vector<DIR*> directories(kMany, 0);
   std::vector<int> fds(kMany, -1);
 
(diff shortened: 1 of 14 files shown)


starboard/nplb/posix_compliance/posix_fdopendir_test.cc
```diff
@@ -53,27 +53,6 @@ TEST(PosixFdopendirTest, SunnyDayStaticContent) {
   }
 }
 
-TEST(PosixFdopendirTest, ManySunnyDay) {
-  std::string path = GetTempDir();
-  EXPECT_FALSE(path.empty());
-  EXPECT_TRUE(FileExists(path.c_str())) << "Filename is " << path.c_str();
-
-  const int kMany = kSbFileMaxOpen / 2;  // Avoid hitting fd limits.
```
niranjanyardi, line 61: kMany =64 in this file starboard/nplb/posix_compliance/posix_thread_local_value_test.cc 

Should we leave this test with kMany = 64 ?
I thin it's a valid test , even without the constant and tests the underlying POSIX API

niranjanyardi (approved):

andrewsavage1 (commented):
I can see these aren't used in the codebase currently—are they replaced by musl macros or do we just not think we'll need them in Chromium code because we would replace them with musl macros if necessary?

y4vor:
> I can see these aren't used in the codebase currently—are they replaced by musl macros or do we just not think we'll need them in Chromium code because we would replace them with musl macros if necessary?

Obviously we don't need them now. If we need a new constant we should start with a macro.

andrewsavage1 (approved):

linxinan-yt (approved):

PR #10412, opened 2026-05-11
[Prototype - DNS][tvOS] Create Chrobalt-compatible tvos_packager.py

fayev-g:
Bug: 512202351

Add tvos_packager.py to handle thin archives and Rust libraries for Chrobalt tvOS. tvOS needs a separate packager from Android/3P due to .a requirement in lieu of .so.

This script replaces the library packaging functionality of `simple_packager.py` and addresses issues with handling LLVM thin archives and Rust libraries:

- Thin Archive Support: `simple_packager.py` relied on Apple's system `ld` and `libtool`, which cannot process LLVM thin archives (resulting in "file too small" or format errors). `tvos_packager.py` uses Cobalt's custom `llvm-ar` with the `L` modifier to correctly extract and merge members from thin archives
- Rust Support: It parses `cobalt_executable.ninja` directly to extract all implicit dependencies, including Rust `.rlib` files, ensuring complete symbol coverage for modern Cobalt checkouts.
- Argument Compatibility: It accepts the same arguments as `simple_packager.py` for easy integration into existing Kokoro/CI workflows.

Currently, this script only handles generating and copying `cobalt.a` to the package directory, serving as a first step toward full package generation. Verified on local Mac machine.

### DIFF
--- a/cobalt/devinfra/kokoro/build/tvos/tvos_packager.py
+++ b/cobalt/devinfra/kokoro/build/tvos/tvos_packager.py
@@ -1,0 +1,173 @@
+#!/usr/bin/env python3
+#
+# Copyright 2026 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


fayev-g:
Closing in favor of moving into gn
https://github.com/youtube/cobalt/pull/10430
https://github.com/youtube/cobalt/pull/10431

PR #10414, opened 2026-05-11
media: Update kDefaultMaxSamplePerWrite

johnedocampo:
This PR updates `kDefaultMaxSamplePerWrite` from 1 to 12. After running an experiment between 1, 4, 8, 12, and 60, we saw that `kDefaultMaxSamplePerWrite` at 12 showed the best improvements.

Bug: 497891900

### DIFF
--- a/media/starboard/starboard_renderer.cc
+++ b/media/starboard/starboard_renderer.cc
@@ -53,7 +53,9 @@ using ::starboard::GetPlayerStateName;
 // buffer, this extra write during preroll can be eliminated.
 const int kPrerollGuardAudioBuffer = 1;
 
-constexpr int kDefaultMaxSamplePerWrite = 1;
+// The value of 12 was chosen after experimentation was done. For more
+// details, see b/497891900.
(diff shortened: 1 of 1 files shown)


media/starboard/starboard_renderer.cc
```diff
@@ -53,7 +53,7 @@ using ::starboard::GetPlayerStateName;
 // buffer, this extra write during preroll can be eliminated.
 const int kPrerollGuardAudioBuffer = 1;
 
-constexpr int kDefaultMaxSamplePerWrite = 1;
+constexpr int kDefaultMaxSamplePerWrite = 12;
```
kjyoun, line 58: can you add comments that 12 comes from the experiment?

kjyoun (approved):
LGTM % comment

media/starboard/starboard_renderer.cc
johnedocampo: Done; added a comment explaining where 12 comes from.

PR #10417, opened 2026-05-12
ci: Add aosp-arm

hferreiro:
Add a CI configuration for aosp-arm where a minimal target is built.

Bug: 495203133


andrewsavage1:
I should've thought of this when we were adding the platform, but I think we'll want aosp to follow a similar pattern to x64x11, raspi, and rdk, where they are the reference device underneath the evergreen build, so you end up with platforms like evergreen-arm-hardfp-rdk. I think we would want a platform aosp-arm for non-evergreen modular builds, but for evergreen builds would want it to be under evergreen-arm-softfp/evergreen-arm64.

It's fine to do this in the future as it may be annoying to consolidate them now, given that we already have builders for evergreen-arm-softfp and evergreen-arm64, but wanted to bring it up.

andrewsavage1:
@hferreiro could you get this green? You may just need to retry tests. I'd like to merge it

andrewsavage1 (approved):

oxve (approved):

PR #10427, opened 2026-05-12
Create script to print top components in sizediff

haozheng-cobalt:
An example to run this script with evergreen-arm-hardfp-rdk_gold between C26 and C27. https://paste.googleplex.com/6597918858543104

This is especially useful for 3P, because evergreen modular architecture is not incompatibility on SuperSize view portal due to lacking linker map.

Bug: 512518336


analyze_size_diff.py
johnxwork, line 1: Can you also put this script under cobalt/tools/?

johnxwork (approved):

PR #10448, opened 2026-05-13
build: guard crash_annotator_jni with !use_evergreen

rogerzanoni:
CrashAnnotatorImplFirstParty is generated (//cobalt/android/BUILD.gn) only when !use_evergreen, use the same logic when adding the source in //cobalt/browser, otherwise the build fails because cobalt/android/jni_headers/CrashAnnotatorImplFirstParty_jni.h isn't being generated.

Bug: 495203133

### DIFF
--- a/cobalt/browser/BUILD.gn
+++ b/cobalt/browser/BUILD.gn
@@ -211,7 +211,7 @@ source_set("cobalt_crash_annotations") {
     "//base",
     "//third_party/crashpad/crashpad/client",
   ]
-  if (is_android) {
+  if (is_android && !use_evergreen) {
     sources += [ "crash_annotator/crash_annotator_jni.cc" ]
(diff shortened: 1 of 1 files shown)


haozheng-cobalt (approved):
use_evergreen could be true or false for AOSP, right now use_evergreen for false is not implemented yet. And since [cobalt/android/BUILD.gn](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/BUILD.gn;l=173?q=cobalt%2Fandroid%2FBUILD.gn&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) uses `use_evergreen`, I think we can use it for now. This flag can be addressed later.

PR #10267, opened 2026-04-29
Make loader_app work on RDK7

santoshmahto7:
This fixes the build error and one seg fault issue by happening in libunwind API call
Bug: 507107324
Change-Id: Ifb78be2abd841aeabe838fd94e5c3a0eb3e759ae

PR #4826, opened 2025-02-03
Increase pread() offset range to fix symbolization on 32-bit systems

astolcenburg:
This allows reading symbols from libraries loaded above the 0x7fffffff address.

b/299486087

PR #4823, opened 2025-02-03
Fix IS_EVERGREEN_ADDRESS macro

astolcenburg:
Fixes IS_EVERGREEN_ADDRESS macro for values >= 0x80000000 on 32-bit systems

b/299486087

PR #10450, opened 2026-05-13
 starboard/rdk: fix stack trace symbolization on 32-bit systems

denisyuji:
## Description ##

Cherry-picks 3 commits from #10267 (`feature/rdk7-26.eap`) into `main`, and add additional fixes. This PR fixes symbolized stack traces on 32-bit RDK targets (arm-hardfp), where crashhandlers previously printed `<unknown> [0x...]` for every frame.

Four root causes addressed:

1. **libgcc_s symbol conflict** — `libgcc_s.so` overrides Cobalt's bundled   libunwind symbols, crashing the unwinder before any frames are collected.   Fix: `-fvisibility=hidden` on the `unwind_starboard` build config.

2. **IS_EVERGREEN_ADDRESS overflow** — casting 32-bit pointers to `uint64_t`   sign-extends addresses ≥ `0x80000000`, so no Evergreen frame is recognized.   Fix: cast to `uintptr_t` in `evergreen_info.h`.

3. **pread() offset limit** — without `_FILE_OFFSET_BITS=64`, pread fails for   libraries mapped above `0x7fffffff`, blocking ELF header reads in symbolize.cc.   Fix: add `#define _FILE_OFFSET_BITS 64`.

4. **No-op stubs in RDK BUILD.gn** — `log_raw_dump_stack` and `system_symbolize`   were wired to empty stubs, so nothing was ever logged even if unwinding worked.   Fix: replace with the real Linux/Starboard implementations.

Verified on an arm-hardfp RDK device: frames now show function names instead of
`<unknown>`.

## Validating the bugfix

To create a controlled crash to verify, after the fixes in `pr/fix-stack-tracing`, add an intentional SIGSEGV at the top of `SbRunStarboardMain` in:
 
```
starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc
```

Add the `InstallCrashSignalHandlers()` call and the null dereference right before the rest of the function body:

```cpp
#include "starboard/shared/signal/crash_signals.h"  // already included

int SbRunStarboardMain(int argc, char** argv, SbEventHandleCallback callback) {
tzset();
// ...

  starboard::InstallCrashSignalHandlers();   // <-- add this
  starboard::InstallSuspendSignalHandlers();
  starboard::InstallStopSignalHandlers();

  // Intentional crash to verify stack trace symbolization — remove after check
  volatile int* null_ptr = nullptr;
  *null_ptr = 0;

// ... rest of function
```

Build and deploy normally, then run `loader_app`. The crash handler fires on SIGSEGV and prints the stack trace to stderr. 

### Expected output (with fixes applied)

The named frames, e.g.:

```
[loader_app/26797:0602/214516.536348(UTC):FATAL:loader_app.cc(167)] Check failed: false. Failed to load library at '/data/out_cobalt/app/cobalt/lib/libcobalt.lz4'.
        (anonymous namespace)::LoadLibraryAndInitialize() [0x440087]
        SbEventHandle [0x440a89]
        starboard::Application::HandleEventAndUpdateState() [0x463c1f]
        starboard::Application::DispatchAndDelete() [0x463fc5]
        starboard::Application::DispatchStart() [0x46420b]
        starboard::Application::RunLoop() [0x4643db]
        starboard::Application::Run() [0x4646ef]
        SbRunStarboardMain [0x462093]
        __libc_start_main [0xf56e6a0f]
```

Bug: 507107324
Bug: 511316178

### DIFF
--- a/base/third_party/symbolize/symbolize.cc
+++ b/base/third_party/symbolize/symbolize.cc
@@ -52,6 +52,9 @@
 #include GLOG_BUILD_CONFIG_INCLUDE
 #endif  // GLOG_BUILD_CONFIG_INCLUDE
 
+// Allow pread() to use higher offsets on 32-bit systems.
+#define _FILE_OFFSET_BITS 64
+
 #include "symbolize.h"
 
 #include "utilities.h"
--- a/base/third_party/symbolize/symbolize.cc
+++ b/base/third_party/symbolize/symbolize.cc
@@ -52,9 +52,6 @@
 #include GLOG_BUILD_CONFIG_INCLUDE
 #endif  // GLOG_BUILD_CONFIG_INCLUDE
 
-// Allow pread() to use higher offsets on 32-bit systems.
-#define _FILE_OFFSET_BITS 64
-
 #include "symbolize.h"
 
 #include "utilities.h"
@@ -71,6 +68,13 @@
 #include "demangle.h"
 
 #include "build/build_config.h"
+
+#if BUILDFLAG(IS_STARBOARD)
+// Allow pread() to use higher offsets on 32-bit systems.
+#define _FILE_OFFSET_BITS 64
+#endif  // BUILDFLAG(IS_STARBOARD)
+
+
 // TODO: b/398296821 - Cobalt: to support google::Symbolize() when called from
 // the Evergreen library (as opposed to from below Starboard) we would need to
 // also make these customizations when building with the Cobalt toolchain,
--- a/starboard/elf_loader/evergreen_info.h
+++ b/starboard/elf_loader/evergreen_info.h
@@ -26,10 +26,10 @@
 #define EVERGREEN_FILE_PATH_MAX_SIZE 4096
 #define EVERGREEN_BUILD_ID_MAX_SIZE 128
 
-#define IS_EVERGREEN_ADDRESS(address, evergreen_info)                    \
-  (evergreen_info.base_address != 0 &&                                   \
-   reinterpret_cast<uint64_t>(address) >= evergreen_info.base_address && \
-   (reinterpret_cast<uint64_t>(address) - evergreen_info.base_address) < \
+#define IS_EVERGREEN_ADDRESS(address, evergreen_info)                     \
+  (evergreen_info.base_address != 0 &&                                    \
+   reinterpret_cast<uintptr_t>(address) >= evergreen_info.base_address && \
+   (reinterpret_cast<uintptr_t>(address) - evergreen_info.base_address) < \
        evergreen_info.load_size)
 
 #ifdef __cplusplus
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn
@@ -195,7 +195,7 @@ static_library("starboard_platform_sources") {
     "//starboard/shared/starboard/window_set_default_options.cc",
     "//starboard/shared/stub/decode_target_get_info.cc",
     "//starboard/shared/stub/decode_target_release.cc",
-    "//starboard/shared/stub/log_raw_dump_stack.cc",
+    "//starboard/shared/starboard/log_raw_dump_stack.cc",
     "//starboard/shared/stub/media_can_change_type.cc",
     "//starboard/shared/stub/microphone_close.cc",
     "//starboard/shared/stub/microphone_create.cc",
@@ -206,7 +206,7 @@ static_library("starboard_platform_sources") {
     "//starboard/shared/stub/microphone_read.cc",
     "//starboard/shared/stub/system_hide_splash_screen.cc",
     "//starboard/shared/stub/system_raise_platform_error.cc",
-    "//starboard/shared/stub/system_symbolize.cc",
+    "//starboard/shared/linux/system_symbolize.cc",
 
     # RDK implementation
     "accessibility_data.cc",
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc
@@ -81,6 +81,7 @@ int SbRunStarboardMain(int argc, char** argv, SbEventHandleCallback callback) {
   stack_size.rlim_cur = 2 * 1024 * 1024;
   setrlimit(RLIMIT_STACK, &stack_size);
 
+  starboard::InstallCrashSignalHandlers();
   starboard::InstallSuspendSignalHandlers();
   starboard::InstallStopSignalHandlers();
 
@@ -95,6 +96,7 @@ int SbRunStarboardMain(int argc, char** argv, SbEventHandleCallback callback) {
 
   starboard::UninstallStopSignalHandlers();
   starboard::UninstallSuspendSignalHandlers();
+  starboard::UninstallCrashSignalHandlers();
 
   return result;
 }
--- a/third_party/llvm-project/libunwind/BUILD.gn
+++ b/third_party/llvm-project/libunwind/BUILD.gn
@@ -135,10 +135,19 @@ config("unwind_starboard_config") {
   ]
 }
 
+config("unwind_hide_symbols_config") {
+  cflags = [ "-fvisibility=hidden" ]
+
+  defines = [
+    "_LIBUNWIND_HIDE_SYMBOLS",
+  ]
+}
+
 static_library("unwind_starboard") {
   sources = common_sources
 
   configs += [ ":unwind_starboard_config" ]
+  configs += [ ":unwind_hide_symbols_config" ]
   all_dependent_configs = [ ":common_unwind_dependents_config" ]
 
   public_deps = [


base/third_party/symbolize/symbolize.cc
```diff
@@ -52,6 +52,9 @@
 #include GLOG_BUILD_CONFIG_INCLUDE
 #endif  // GLOG_BUILD_CONFIG_INCLUDE
 
+// Allow pread() to use higher offsets on 32-bit systems.
```
niranjanyardi: this doesnt seem right . all partners are going to use this code. 32 and 64 bit . Check if there's a starboard API through which you can access this info

starboard/elf_loader/evergreen_info.h
```diff
@@ -28,8 +28,8 @@
 
 #define IS_EVERGREEN_ADDRESS(address, evergreen_info)                    \
   (evergreen_info.base_address != 0 &&                                   \
-   reinterpret_cast<uint64_t>(address) >= evergreen_info.base_address && \
```
niranjanyardi, line 31: can you explain what's the motivation for this change ?
Check this code on Cobalt 25 - was it working for RDK ?

base/third_party/symbolize/symbolize.cc
denisyuji: I will leave this question to @santoshmahto7 , since he is the author of the [original PR](https://github.com/youtube/cobalt/pull/10267/changes)

starboard/elf_loader/evergreen_info.h
denisyuji: same from above

denisyuji:
@astolcenburg , could you help us by [signing the CLA](https://cla.developers.google.com/clas) ?

base/third_party/symbolize/symbolize.cc
santoshmahto7: @niranjanyardi  The commit is cherry-picked from the C25 libunwind code, but your concern is valid. I will check  deeper and update.

jonastsai: please check the discussing at
b/299486087 and
https://github.com/youtube/cobalt/pull/4826

starboard/elf_loader/evergreen_info.h
jonastsai: please check the discussing at
b/299486087 and
https://github.com/youtube/cobalt/pull/4823

base/third_party/symbolize/symbolize.cc
santoshmahto7: On further checking, the above _FILE_OFFSET_BITS only affects 32-bit systems and allows reading files larger than 2GB (the maximum limit on 32-bit systems). On 64-bit, this has no effect.
This allows reading symbols from libraries loaded above the 0x7fffffff address (as per the commit message)

In RDK (or RPI), this is already done via build configs:
./starboard/contrib/rdk/src/third_party/starboard/rdk/shared/platform_configuration/BUILD.gn:      "_FILE_OFFSET_BITS=64",

So for RDK this is not required since it is already set. But I think this will be useful for other builds also

starboard/elf_loader/evergreen_info.h
santoshmahto7: This is also cherry-picked from C25.  I think using uintptr_t is more correct. here than  using 64bit (uint64_t) .

base/third_party/symbolize/symbolize.cc
niranjanyardi: @y4vor  can you ptal ?

jellefoks:
I took that PR, but still no stack trace when I make it crash

Just a `pid 6 has terminated by signal 11`, no stack trace in the journalctl.

Do I need to build/deploy other things to see stack traces? I rebuilt and deployed libcobalt and libloader_app.so and also re-ran chCobalt c26 and rebooted, but still see no stack traces in journalctl.

santoshmahto7:
> I took that PR, but still no stack trace when I make it crash
> 
> Just a `pid 6 has terminated by signal 11`, no stack trace in the journalctl.
> 
> Do I need to build/deploy other things to see stack traces? I rebuilt and deployed libcobalt and libloader_app.so and also re-ran chCobalt c26 and rebooted, but still see no stack traces in journalctl.

This PR doesn't specifically address the stacktrace print after crash, but just fixes the libwunwind issues. For the call stack in the journalctl,  additional investigation is required.

denisyuji:
> I took that PR, but still no stack trace when I make it crash
> 
> Just a `pid 6 has terminated by signal 11`, no stack trace in the journalctl.
> 
> Do I need to build/deploy other things to see stack traces? I rebuilt and deployed libcobalt and libloader_app.so and also re-ran chCobalt c26 and rebooted, but still see no stack traces in journalctl.

@jellefoks , can you try again after we pushed changes? I reproduced from a clean cobalt build now, to make sure the intermediate artifacts not mask the results

denisyuji:
> > I took that PR, but still no stack trace when I make it crash
> > Just a `pid 6 has terminated by signal 11`, no stack trace in the journalctl.
> > Do I need to build/deploy other things to see stack traces? I rebuilt and deployed libcobalt and libloader_app.so and also re-ran chCobalt c26 and rebooted, but still see no stack traces in journalctl.
> 
> @jellefoks ,see also in the PR message how I produced the fault to test.

denisyuji:
> > I took that PR, but still no stack trace when I make it crash
> > Just a `pid 6 has terminated by signal 11`, no stack trace in the journalctl.
> > Do I need to build/deploy other things to see stack traces? I rebuilt and deployed libcobalt and libloader_app.so and also re-ran chCobalt c26 and rebooted, but still see no stack traces in journalctl.
> 
> This PR doesn't specifically address the stacktrace print after crash, but just fixes the libwunwind issues. For the call stack in the journalctl, additional investigation is required.

Thanks, @santoshmahto7 . I added additional commits to the cherry picked ones and I've updated the title and description of the PR, since it is no longer a simple cherry-pick

starboard/elf_loader/evergreen_info.h
denisyuji: thanks @santoshmahto7, @jonastsai  and @niranjanyardi. The commit `97c197f13bb7089909e07e77348f051bea07ec87` (also cherry-picked in this PR) fixes it:

```
 -  reinterpret_cast<uint64_t>(address) >= evergreen_info.base_address
 +  reinterpret_cast<uintptr_t>(address) >= evergreen_info.base_address
```

santoshmahto7:
@denisyuji  Thanks for fixing the stack symbol issue. I have confirmed that those new commits are perfect to get stacktrace in journalctl or in log.

denisyuji:
## Cobalt + Starboard crash verification

Added an intentional null-dereference in `ShellBrowserMainParts::PreMainMessageLoopRun()` (deep inside `libcobalt.so`) to verify that a crash originating in the Cobalt browser layer is caught by the Starboard signal handler installed by this PR.

### Result

```
Caught signal: SIGSEGV (11)
        <unknown> [0x5b6fdca]     ← crash in ShellBrowserMainParts::PreMainMessageLoopRun()
        <unknown> [0x179834b]
        <unknown> [0x3a26a37]
        <unknown> [0x18f5d51]
        <unknown> [0x400df91]
        <unknown> [0x3a2671b]
        <unknown> [0x3a28387]
        <unknown> [0x1790ceb]
        <unknown> [0x45a6209]
        <unknown> [0x45a7aa7]
        <unknown> [0x45a778b]
        <unknown> [0x45a5cff]
        <unknown> [0x1790557]
        <unknown> [0x178ff7b]
        <unknown> [0x178e313]
        <unknown> [0x178e295]
        <unknown> [0x178e01f]
        SbEventHandle [0xa8f91d]
        starboard::Application::HandleEventAndUpdateState() [0xab2d1f]
        starboard::Application::DispatchAndDelete() [0xab30c5]
        starboard::Application::DispatchStart() [0xab330b]
        starboard::Application::RunLoop() [0xab34db]
        starboard::Application::Run() [0xab37ef]
        SbRunStarboardMain [0xab1097]
        __libc_start_main [0xf523fa0f]
```

### Observations

-  **SIGSEGV originating in `libcobalt.so` is caught** by the Starboard crash handler installed by this PR
-  **Starboard frames are fully symbolized** (`SbRunStarboardMain`, `Application::Run()`, etc.)
-  **Cobalt frames show as `<unknown>`** — this is an inherent limitation of the Evergreen architecture: `libcobalt.so` is loaded by the custom ELF loader at runtime and is not registered with the system dynamic linker, so `dladdr()` cannot resolve those addresses to symbols

The `<unknown>` Cobalt frames are not a regression from this PR — they are the same before and after. What this PR fixes is the Starboard frames, which were also `<unknown>` before `InstallCrashSignalHandlers()` was called. Symbolizing the Cobalt ELF-loaded frames would require additional integration between the ELF loader and the crash handler.

astolcenburg:
Hi @denisyuji , can you try with the unstripped version of the libcobalt.so library?

denisyuji:
> Hi @denisyuji , can you try with the unstripped version of the libcobalt.so library?

@astolcenburg , thanks for reviweing. See #10782 

#10450 (this PR) - Fixes the stacktrace issue with RDK cobalt
#10782 - Fixes the cobalt symbolization

You will need to apply both

santoshmahto7:
> Hi @denisyuji , can you try with the unstripped version of the libcobalt.so library?

 The cobalt symbols is missed because  //copied_base/.../symbolliz.cc doesn't have same code as in //base/.../symbolize.cc which does help to resolve evergreen library symbols 

I have verified that this PR enables stacktrace on crash on RDK. for cobalt symbols, another PR https://github.com/youtube/cobalt/pull/10782 is opened.

santoshmahto7 (approved):

niranjanyardi (approved):

base/third_party/symbolize/symbolize.cc
niranjanyardi: can we wrap this in BUILDFLAG(IS_STARBOARD) so that oit doesn't apply to upstream chromium code ?

niranjanyardi (approved):

niranjanyardi (approved):
look good to me apart from the 1 open comment about wrapping up the 3P code with is_starboard

base/third_party/symbolize/symbolize.cc
denisyuji: that makes sense. I guarded it with `defined(STARBOARD)`

niranjanyardi: BUILDFLAG(IS_STARBOARD) - please use this - we are trying to be consistent with using this pattern

base/third_party/symbolize/symbolize.cc
```diff
@@ -71,6 +68,13 @@
 #include "demangle.h"
 
 #include "build/build_config.h"
+
+#if BUILDFLAG(IS_STARBOARD)
+// Allow pread() to use higher offsets on 32-bit systems.
+#define _FILE_OFFSET_BITS 64
```
astolcenburg, line 74: To achieve the expected behavior, _FILE_OFFSET_BITS must be defined as 64 before any system headers are included.

Additionally, please note that several modules already unconditionally define _FILE_OFFSET_BITS as 64, for example:
https://github.com/youtube/cobalt/blob/main/third_party/ijar/zip.cc#L23

base/third_party/symbolize/symbolize.cc
denisyuji: Ok! I placed it after `build/build_config.h`, since `IS_STARBOARD` is defined in this header

PR #6641, opened 2025-07-30
[POSIX] Remove references to ioctl

andrewsavage1:
Fixed: 424001886

PR #10458, opened 2026-05-13
Media: add FIFO buffer for audio input stream

sideb0ard:
Add FIFO buffer to handle case when input stream is bursty.

Tested on Android and Linux implementations.

Bug: 483713292

### DIFF
--- a/media/audio/starboard/audio_input_stream_starboard.cc
+++ b/media/audio/starboard/audio_input_stream_starboard.cc
@@ -180,17 +180,18 @@ void AudioInputStreamStarboard::ReadAudio() {
       params_.frames_per_buffer() * params_.channels() * sizeof(int16_t);
   const size_t max_fifo_size_bytes = target_buffer_size_bytes * kMaxFifoBuffers;
   const size_t read_size_bytes =
-      std::max(size_t{kMinReadBytes}, target_buffer_size_bytes);
+      std::max(static_cast<size_t>(kMinReadBytes), target_buffer_size_bytes);
 
-  std::vector<uint8_t> temp_read_buf(read_size_bytes);
+  // Read directly into the FIFO to avoid extra allocation and copy.
+  size_t current_fifo_size = audio_fifo_.size();
+  audio_fifo_.resize(current_fifo_size + read_size_bytes);
   int bytes_read = SbMicrophoneRead(
-      microphone_, reinterpret_cast<char*>(temp_read_buf.data()),
+      microphone_,
+      reinterpret_cast<char*>(audio_fifo_.data() + current_fifo_size),
       read_size_bytes);
 
   if (bytes_read > 0) {
-    audio_fifo_.insert(audio_fifo_.end(), temp_read_buf.data(),
-                       temp_read_buf.data() + bytes_read);
-
+    audio_fifo_.resize(current_fifo_size + bytes_read);
     if (audio_fifo_.size() > max_fifo_size_bytes) {
       size_t bytes_to_drop = audio_fifo_.size() - max_fifo_size_bytes;
       audio_fifo_.erase(audio_fifo_.begin(),
@@ -198,14 +199,18 @@ void AudioInputStreamStarboard::ReadAudio() {
       DLOG(WARNING) << "Audio FIFO overflow, dropped " << bytes_to_drop
                     << " bytes.";
     }
-  } else if (bytes_read < 0) {
-    DLOG(WARNING) << "SbMicrophoneRead returned " << bytes_read << ".";
-    HandleError("SbMicrophoneRead");
-    return;
+  } else {
+    audio_fifo_.resize(current_fifo_size);
+    if (bytes_read < 0) {
+      DLOG(WARNING) << "SbMicrophoneRead returned " << bytes_read << ".";
+      HandleError("SbMicrophoneRead");
+      return;
+    }
   }
 
   bool data_pushed = false;
-  if (audio_fifo_.size() >= target_buffer_size_bytes) {
+  // Process all available complete buffers to minimize latency.
+  while (audio_fifo_.size() >= target_buffer_size_bytes) {
     std::copy(audio_fifo_.begin(),
               audio_fifo_.begin() + target_buffer_size_bytes,
               reinterpret_cast<uint8_t*>(buffer_.data()));
@@ -217,13 +222,11 @@ void AudioInputStreamStarboard::ReadAudio() {
 
     callback_->OnData(audio_bus_.get(), base::TimeTicks::Now(), 1.0, {});
     data_pushed = true;
-  }
-
-  // Adaptive scheduling: poll faster if we didn't have enough data to push.
-  if (data_pushed) {
     next_read_time_ =
         std::max(base::TimeTicks::Now(), next_read_time_ + buffer_duration_);
-  } else {
+  }
+
+  if (!data_pushed) {
     next_read_time_ = base::TimeTicks::Now() + buffer_duration_ / 2;
   }
 
--- a/media/audio/starboard/audio_input_stream_starboard.cc
+++ b/media/audio/starboard/audio_input_stream_starboard.cc
@@ -14,11 +14,17 @@
 
 #include "media/audio/starboard/audio_input_stream_starboard.h"
 
+#include <algorithm>
+
 #include "base/logging.h"
 #include "media/audio/audio_manager.h"
 #include "media/audio/audio_manager_base.h"
 
 namespace media {
+
+namespace {
+constexpr int kMinReadBytes = 2048;
+}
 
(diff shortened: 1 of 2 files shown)


media/webrtc/audio_processor.cc
```diff
@@ -620,6 +620,8 @@ AudioParameters AudioProcessor::GetDefaultOutputFormat(
 #if BUILDFLAG(IS_CASTOS) || BUILDFLAG(IS_CAST_ANDROID)
           std::min(media::WebRtcAudioProcessingSampleRateHz(),
                    input_format.sample_rate())
+#elif BUILDFLAG(IS_COBALT)
+          input_format.sample_rate()
```
kjyoun: I think that this is a separate change than changes in audio_input_stream_starboard.h

Can you make this as a separate PR?

base/sync_socket_posix.cc
```diff
@@ -174,7 +174,9 @@ size_t SyncSocket::Peek() {
   ssize_t number_chars = recv(handle_.get(), buffer, sizeof(buffer),
                               MSG_PEEK | MSG_TRUNC | MSG_DONTWAIT);
   if (number_chars < 0) {
-    PLOG(ERROR) << "recv failed in SyncSocket::Peek";
+    if (errno != EAGAIN && errno != EWOULDBLOCK) {
+      PLOG(ERROR) << "recv failed in SyncSocket::Peek";
+    }
```
kjyoun: Can you make this change into separate PR?

That being said, how can we verify this is a right change?

why not try again when EAGAIN or EWOULDBLOCK?

Can you add comment about this behavior?

another question --
why do we need starboard own impl.. not using chromium's impl. below (iotcl)?

media/audio/starboard/audio_input_stream_starboard.cc
```diff
@@ -14,12 +14,18 @@
...
+#include <algorithm>
+
 #include "base/logging.h"
 #include "media/audio/audio_manager.h"
 #include "media/audio/audio_manager_base.h"
 
 namespace media {
 
+namespace {
+constexpr int kMinReadSizeBytes = 2048;
```
kjyoun: super nit: `kMinReadBytes` ?

media/audio/starboard/audio_input_stream_starboard.cc
```diff
@@ -171,65 +176,62 @@ void AudioInputStreamStarboard::ReadAudio() {
...
-                       base::Unretained(this)),
-        next_check_time, base::subtle::DelayPolicy::kPrecise);
-  } else {
-    DLOG(WARNING) << "SbMicrophoneRead returned " << bytes_read
-                  << ". Dropping this buffer.";
+  } else if (bytes_read < 0) {
+    DLOG(WARNING) << "SbMicrophoneRead returned " << bytes_read << ".";
     HandleError("SbMicrophoneRead");
     return;
   }
```
kjyoun, line 209: don't we need to handle bytes_read == 0?

kjyoun (commented):
I think that this PR has 3 different changes, each of which deserves its own PR and review.

Can you break down?

super nit: can u change title to "media: Add... " per https://g3doc.corp.google.com/company/teams/cobalt/team/coding_style.md?cl=head#commit-message

base/sync_socket_posix.cc
sideb0ard: new PR at https://github.com/youtube/cobalt/pull/10470

> why do we need starboard own impl.. not using chromium's impl. below (iotcl)?

Starboard doesn't support ioctl
https://github.com/youtube/cobalt/pull/6641/changes

media/webrtc/audio_processor.cc
sideb0ard: done - https://github.com/youtube/cobalt/pull/10471

media/audio/starboard/audio_input_stream_starboard.cc
sideb0ard: We don't need an explicit  else if (bytes_read == 0)  branch anymore because it is handled correctly by the fall-through behavior of the new code.
  
Here is what happens when  bytes_read  is  0 :
  
1. It skips  if (bytes_read > 0)  (no data to add to FIFO).
2. It skips  else if (bytes_read < 0)  (no error).
3. It checks if the FIFO has enough data to push ( audio_fifo_.size() >= target_buffer_size_bytes ).
      • If it does (from previous reads), it pushes it and sets  data_pushed = true .
      • If it doesn't,  data_pushed  remains  false .
4. Finally, it hits the adaptive scheduling:
      • Since no data was read and (presumably) no data was pushed,  data_pushed  is false, and it schedules the next read in  buffer_duration_ / 2 .
  
This matches the old behavior for  bytes_read == 0  (polling faster to check for data again) but with the added benefit that we can still push a chunk if there was enough data left over in the FIFO from a previous burst.

media/audio/starboard/audio_input_stream_starboard.cc
sideb0ard: updated.

media/audio/starboard/audio_input_stream_starboard.h
```diff
@@ -74,7 +76,7 @@ class AudioInputStreamStarboard : public AgcAudioStream<AudioInputStream> {
   bool closing_ = false;
   std::unique_ptr<AudioBus> audio_bus_;
   std::vector<int16_t> buffer_;
-  int frames_in_buffer_ = 0;
+  std::vector<uint8_t> audio_fifo_;
```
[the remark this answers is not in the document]
sideb0ard: will stick with vector

media/audio/starboard/audio_input_stream_starboard.cc
```diff
@@ -171,65 +176,62 @@ void AudioInputStreamStarboard::ReadAudio() {
...
+        std::max(base::TimeTicks::Now(), next_read_time_ + buffer_duration_);
+  } else {
+    next_read_time_ = base::TimeTicks::Now() + buffer_duration_ / 2;
+  }
+
+  capture_thread_.task_runner()->PostDelayedTaskAt(
+      base::subtle::PostDelayedTaskPassKey(), FROM_HERE,
+      base::BindOnce(&AudioInputStreamStarboard::ReadAudio,
+                     base::Unretained(this)),
+      next_read_time_, base::subtle::DelayPolicy::kPrecise);
```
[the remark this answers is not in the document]
sideb0ard: refactored

media/audio/starboard/audio_input_stream_starboard.cc
```diff
@@ -171,65 +176,65 @@ void AudioInputStreamStarboard::ReadAudio() {
...
+    }
+  }
+
+  bool data_pushed = false;
+  // Process all available complete buffers to minimize latency.
+  while (audio_fifo_.size() >= target_buffer_size_bytes) {
+    std::copy(audio_fifo_.begin(),
+              audio_fifo_.begin() + target_buffer_size_bytes,
+              reinterpret_cast<uint8_t*>(buffer_.data()));
+    audio_fifo_.erase(audio_fifo_.begin(),
```
kjyoun, line 217: std::vector::erase(..) has overhead of O(n). Inside of the loop, it results in O(N^2)


How about using `base::circular_deque` which can handle erase O(1) ?

media/audio/starboard/audio_input_stream_starboard.cc
```diff
@@ -171,65 +176,65 @@ void AudioInputStreamStarboard::ReadAudio() {
...
-      frames_to_read * params_.channels() * sizeof(int16_t);
+  const size_t target_buffer_size_bytes =
+      params_.frames_per_buffer() * params_.channels() * sizeof(int16_t);
+  const size_t max_fifo_size_bytes = target_buffer_size_bytes * kMaxFifoBuffers;
+  const size_t read_size_bytes =
+      std::max(static_cast<size_t>(kMinReadBytes), target_buffer_size_bytes);
 
+  // Read directly into the FIFO to avoid extra allocation and copy.
+  size_t current_fifo_size = audio_fifo_.size();
+  audio_fifo_.resize(current_fifo_size + read_size_bytes);
```
kjyoun, line 187: I can see  you are calling resize() in several places. here and line 194, and line 203

Can we resize only once to the possible largest size?

calling resize() multiple times per each Read() does not look ideal.

media/audio/starboard/audio_input_stream_starboard.cc
kjyoun, line 172: The FIFO buffering logic here is getting quite complex and handles some tricky edge cases (like overflow and partial buffer accumulation). This logic deserves its own isolated unit tests.
 
How about extract this into a dedicated  class (e..g. AudioFifo) and cover it with unit tests, and have  AudioInputStreamStarboard simply use that class? 

It will  improve the readability of  ReadAudio()  and ensure our buffer math is bulletproof

PR #10459, opened 2026-05-13
Remove WebNN from Cobalt

haozheng-cobalt:
Bug: 506153165


johnxwork:
Please post the binary size saving in the description.

Another potentially good idea is to use a different macro from "IS_COBALT" since these removal serves a particular purpose of "no webnn".

The lines of code change is actually very manageable. That's great.

PR #10471, opened 2026-05-14
media: Disable WetRtc processing for Cobalt

sideb0ard:
Cobalt does not require WebRtc processing as it adds additional latency for the microphone path.
Normally this is used for echo cancellation and automatic gain control when handling two way calls, which we don't need or support.

Bug: 483713292

### DIFF
--- a/media/BUILD.gn
+++ b/media/BUILD.gn
@@ -209,7 +209,10 @@ test("media_unittests") {
     "//media/test:pipeline_integration_tests",
     "//media/test:run_all_unittests",
     "//media/video:unit_tests",
-    "//media/webrtc:unit_tests",
+    # Cobalt disabled WebRtc processing in 
+    # https://github.com/youtube/cobalt/pull/10471
+    # due to additional latency for the microphone path.
+    # "//media/webrtc:unit_tests",
   ]
 
   if (is_cobalt && use_starboard_media) {
--- a/media/BUILD.gn
+++ b/media/BUILD.gn
@@ -214,6 +214,9 @@ test("media_unittests") {
 
   if (is_cobalt && use_starboard_media) {
     deps += ["//media/starboard:unit_tests"]
+  }
+
+  if (is_cobalt) {
     # Cobalt disabled WebRtc processing in 
     # https://github.com/youtube/cobalt/pull/10471
     # due to additional latency for the microphone path.
--- a/media/BUILD.gn
+++ b/media/BUILD.gn
@@ -209,14 +209,15 @@ test("media_unittests") {
     "//media/test:pipeline_integration_tests",
     "//media/test:run_all_unittests",
     "//media/video:unit_tests",
-    # Cobalt disabled WebRtc processing in 
-    # https://github.com/youtube/cobalt/pull/10471
-    # due to additional latency for the microphone path.
-    # "//media/webrtc:unit_tests",
+    "//media/webrtc:unit_tests",
   ]
 
   if (is_cobalt && use_starboard_media) {
     deps += ["//media/starboard:unit_tests"]
+    # Cobalt disabled WebRtc processing in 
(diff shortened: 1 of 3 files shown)


media/webrtc/audio_processor.cc
kjyoun, line 618: WDYT about making `need_webrtc_audio_processing` to false?

Then `input_format.sample_rate()` will be selected

I think that we dont need to use WebRtcProcessing,  (C25 doesn't use it ?)


How about this change?

https://source.corp.google.com/h/github/youtube/cobalt/+/main:media/base/audio_processing.h;l=35-45;drc=524126904340bf2104b6615d6bb6352ba6d7d844

```
bool NeedWebrtcAudioProcessing() const {
#if IS_COBALT
  // Cobalt does not support WebRtcAudioProcessing, since it increases audio latency on low-end TV devices
  return false;
#endif

  }
```

sideb0ard: i dig it!

media/base/audio_processing.h
```diff
@@ -33,6 +33,11 @@ struct MEDIA_EXPORT AudioProcessingSettings {
   }
 
   bool NeedWebrtcAudioProcessing() const {
+#if BUILDFLAG(IS_COBALT)
+  // Cobalt does not support WebRtcAudioProcessing, since it increases audio
+  // latency on low-end TV devices
```
kjyoun, line 38: can you add link to a bug comment about this, if you have any?

kjyoun (approved):

media/base/audio_processing.h
sideb0ard: done.

media/BUILD.gn
```diff
@@ -209,7 +209,10 @@ test("media_unittests") {
     "//media/test:pipeline_integration_tests",
     "//media/test:run_all_unittests",
     "//media/video:unit_tests",
-    "//media/webrtc:unit_tests",
+    # Cobalt disabled WebRtc processing in 
+    # https://github.com/youtube/cobalt/pull/10471
+    # due to additional latency for the microphone path.
+    # "//media/webrtc:unit_tests",
```
borongc: maybe like this at L220:
```
    deps -= ["//media/webrtc:unit_tests"]
```

Would this work?

sideb0ard: yeah, thanks!

media/BUILD.gn
```diff
@@ -216,6 +216,13 @@ test("media_unittests") {
     deps += ["//media/starboard:unit_tests"]
   }
 
+  if (is_cobalt) {
```
borongc, line 219: One more, is this only for Cobalt, or both `is_cobalt && use_starboard_media`?

If it's the latter, we can do:

```
  if (is_cobalt && use_starboard_media) {
    deps += ["//media/starboard:unit_tests"]
    # Cobalt disabled WebRtc processing in 
    # https://github.com/youtube/cobalt/pull/10471
    # due to additional latency for the microphone path.
    deps -= ["//media/webrtc:unit_tests"]
  }
```

borongc (approved):

media/BUILD.gn
sideb0ard: i initially used that same scope (is_cobalt &&use_starboard_media), but as the check in media/base/audio_processing.h is only for IS_COBALT, i thought it would better to stick to same constraints.

borongc: Then, you can do:

```
  if (is_cobalt) {
    # Cobalt disabled WebRtc processing in 
    # https://github.com/youtube/cobalt/pull/10471
    # due to additional latency for the microphone path.
    deps -= ["//media/webrtc:unit_tests"]

    if (use_starboard_media) {
        deps += ["//media/starboard:unit_tests"]
    }    
  }
```

PR #10472, opened 2026-05-14
Disable V8 maglev

zhongqiliang:
Bug: 513264006


haozheng-cobalt:
Can you build app before and after the change to see how much binary is reduced from this change?

zhongqiliang:
The binary size reduced 0.4%, not a lot.

PR #10473, opened 2026-05-14
 Deprecate SbThreadGetId

niranjanyardi:
Deprecate `SbThreadGetId` from the Starboard API and replace internal usages across Starboard and adapted third-party dependencies (like WebRTC) with direct calls to linux kernel's `gettid()`.

### Key Architectural Refactors:

1. **Starboard gettid Fallback (`starboard/common/gettid.h`)**
   Consolidated compile-time overrides into a single platform-agnostic header:
   - **Bionic / Musl**: Resolves natively to `<unistd.h>` `gettid()` directly (fully optimized).
   - **glibc < 2.30 Fallback**: Calls raw `syscall(SYS_gettid)` on Linux to bypass host sysroot link errors.
   - **Apple Platforms (tvOS / macOS)**: Uses Darwin's native Mach thread port `pthread_mach_thread_np(pthread_self())` to avoid `syscall()` deprecation compilation errors on tvOS.
   - **Descriptive Compiler Error**: Fails with an actionable, highly descriptive `#error` message guiding port developers if both `SYS_gettid` and `__NR_gettid` are missing.

2. **Clean Transitive Header Imports**
   Cleaned up redundant direct `<sys/syscall.h>` and `<unistd.h>` inclusions (e.g., in `starboard/common/log.cc`) as `gettid.h` handles system and native headers transitively.

3. **POSIX Compliance Tests**
   Added `posix_gettid_test.cc` to the `nplb` compliance suite to verify uniqueness and correctness of thread IDs unconditionally on all platforms.

Issue: 502578513
Issue: 517239886


starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -698,6 +699,14 @@ musl_pid_t __abi_wrap_getpid() {
   return static_cast<musl_pid_t>(getpid());
 }
 
+musl_pid_t __abi_wrap_gettid() {
+#if defined(__ANDROID__)
+  return static_cast<musl_pid_t>(gettid());
+#else
+  return static_cast<musl_pid_t>(syscall(SYS_gettid));
+#endif
+}
```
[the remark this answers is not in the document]
niranjanyardi: mac is not starbordized

starboard/android/shared/BUILD.gn
```diff
@@ -254,7 +254,7 @@ static_library("starboard_platform") {
     "system_request_stop.cc",
     "text_to_speech_helper.cc",
     "text_to_speech_helper.h",
-    "thread_get_id.cc",
+    "thread_create_priority.cc",
```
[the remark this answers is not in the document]
niranjanyardi: Fixed. We resolved the merge conflicts correctly by completely removing both the deprecated `thread_get_id.cc` and the already-deleted `thread_create_priority.cc` / `thread_priority.cc` lines from `android/shared/BUILD.gn`, `stub/BUILD.gn`, and `tvos/shared/BUILD.gn`.

starboard/common/gettid.h
```diff
@@ -12,11 +12,13 @@
...
-#include "starboard/thread.h"
+#include "starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.h"
 
-SbThreadId SbThreadGetId() {
-  mach_port_t machTID = pthread_mach_thread_np(pthread_self());
-  return static_cast<SbThreadId>(machTID);
-}
+#if !defined(gettid)
+#define gettid() __abi_wrap_gettid()
+#endif
```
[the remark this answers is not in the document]
niranjanyardi: Fixed. We consolidated the macro into `starboard/common/gettid.h` as a clean forwarding header to `__abi_wrap_gettid()`. This keeps the common code (`log.cc`, `thread_checker.h`) completely clean and uniform. Since we only care about modular builds in this context, this dependency is appropriate and aligns with other Starboard common wrappers (like `allocator.h`, `file.h`).

starboard/nplb/posix_compliance/posix_gettid_test.cc
```diff
@@ -0,0 +1,95 @@
...
+#if defined(__linux__)
+  pid_t tid = gettid();
+  EXPECT_GT(tid, 0);
+
+  pid_t syscall_tid = syscall(SYS_gettid);
+  EXPECT_EQ(tid, syscall_tid);
+#else
+  // On non-linux platforms, our fallback wrapper returns 0.
+  EXPECT_EQ(gettid(), 0);
+#endif
```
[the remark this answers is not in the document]
niranjanyardi: We are fully ignoring Apple platforms in this branch as Apple is not a targeted Starboard platform for this project. We implemented a safe fallback returning 0 on non-Linux/non-Android platforms to prevent compile errors, which is sufficient for our modular build scope.

starboard/common/gettid.h
```diff
@@ -12,15 +12,21 @@
...
-// clang-format off
-#include "starboard/thread.h"
-// clang-format on
+#ifndef STARBOARD_COMMON_GETTID_H_
+#define STARBOARD_COMMON_GETTID_H_
 
+#include "build/build_config.h"
+
+#if BUILDFLAG(IS_STARBOARD) && !BUILDFLAG(IS_ANDROID)
 #include <sys/syscall.h>
```
niranjanyardi: this is the key part to review IMO.

IIUC the syscall gets routed  via musl back to starboard wrappers.

FYI @jellefoks

starboard/common/gettid.h
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 The Cobalt Authors. All Rights Reserved.
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
```
jellefoks, line 1: Don't change the copyright year on existing files.

starboard/common/gettid.h
```diff
@@ -12,15 +12,21 @@
...
+#include "build/build_config.h"
+
+#if BUILDFLAG(IS_STARBOARD) && !BUILDFLAG(IS_ANDROID)
 #include <sys/syscall.h>
-#include <sys/types.h>
 #include <unistd.h>
+#if !defined(gettid)
+#define gettid() syscall(SYS_gettid)
+#endif
+#elif BUILDFLAG(IS_ANDROID)
```
jellefoks: I don't think we have to treat Android special here, and above.

starboard/common/gettid.h
jellefoks: Let's only fall include syscall when we actually have to fall back to it, e.g. when unistd.h doesn't define gettid.

starboard/common/gettid.h
```diff
@@ -12,15 +12,21 @@
...
+#define STARBOARD_COMMON_GETTID_H_
 
+#include "build/build_config.h"
+
+#if BUILDFLAG(IS_STARBOARD) && !BUILDFLAG(IS_ANDROID)
 #include <sys/syscall.h>
-#include <sys/types.h>
 #include <unistd.h>
+#if !defined(gettid)
+#define gettid() syscall(SYS_gettid)
```
jellefoks: Let's put a comment here that clearly states that we're falling back to doing the direct syscall  because the unistd.h on the system doesn't define gettid().

jellefoks: We should ensure that `SYS_gettid` is defined (should be defined in sys/syscall.h, but we can make it clear here that it's missing for platforms where somehow it isn't), more so even that would make it very clear what this code needs from sys/syscall.h for ports that don't have sys/syscall.h.

jellefoks (changes requested):
Looks good in principle, I would just make tidy up the things that I commented on.

third_party/webrtc/rtc_base/platform_thread_types.cc
```diff
@@ -56,7 +57,7 @@ PlatformThreadId CurrentThreadId() {
 #elif defined(__EMSCRIPTEN__)
   return static_cast<PlatformThreadId>(pthread_self());
 #elif BUILDFLAG(IS_STARBOARD)
-  return SbThreadGetId();
+  return gettid();
```
jellefoks: Merge this with the   defined(WEBRTC_ANDROID) that already uses gettid()?

starboard/common/gettid.h
niranjanyardi: hmm, this is a new file. it's not really a copy of the earlier ones.

starboard/common/gettid.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Fallback for platforms where the toolchain/sysroot <unistd.h> does not
+// declare gettid() natively (e.g., glibc < 2.30).
+#if !defined(gettid)
+
+// TODO(b/517239886): Apple is not a Starboard platform and the Apple
+// dependencies on this file need to be cleaned up.
+#if BUILDFLAG(IS_IOS_TVOS)
+#include <pthread.h>
+inline pid_t gettid() {
```
niranjanyardi, line 30: this is based on the existing implementation in  starboard/tvos/shared/thread_get_id.cc which is removed in this PR.

If apple needs to use some starboard files for some reason - i think we should patch the starboard sites which call gettid() with pthread_mach_thread_np directly .
 
```
#include <pthread.h>

#include "starboard/thread.h"

SbThreadId SbThreadGetId() {
  mach_port_t machTID = pthread_mach_thread_np(pthread_self());
  return static_cast<SbThreadId>(machTID);
}
```

@fayev-g is this okay ?

note: i ran into this because of build errors; https://b.corp.google.com/issues/502578513#comment9 , comment10

starboard/common/gettid.h
niranjanyardi: done, removed , please re-review

starboard/common/gettid.h
niranjanyardi: done

third_party/webrtc/rtc_base/platform_thread_types.cc
niranjanyardi: done

starboard/common/gettid.h
niranjanyardi: added a comment, pelase review

niranjanyardi: in this case , i have ensured that a compile error throws up with a descriptve message, please review

starboard/common/gettid.h
fayev-g: sg. Thanks for moving that over

fayev-g (approved):
For tvos/

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -698,6 +699,14 @@ musl_pid_t __abi_wrap_getpid() {
   return static_cast<musl_pid_t>(getpid());
 }
 
+musl_pid_t __abi_wrap_gettid() {
+#if defined(__ANDROID__)
+  return static_cast<musl_pid_t>(gettid());
+#else
+  return static_cast<musl_pid_t>(syscall(SYS_gettid));
```
jellefoks, line 706: It's more efficient to call `gettid()` from <unistd.h>, because that does not always cross the syscall kernel boundary (and use TLS instead). We should only fallback to the syscall if for some reason unistd.h doesn't declare `gettid()`.

The SbThreadGetId implementation that this PR removes also was not as efficient as possible, so it we can improve that here, it may (should) help performance.

niranjanyardi: good point , done . the implementation now uses gettid whenever available

jellefoks (approved):
Great!

y4vor (approved):

PR #10475, opened 2026-05-14
[nplb] Fix signedness test of enum in C++17

andrewsavage1:
Bug: 452970985


starboard/sabi/x64/sysv/sabi-v18.json
```diff
@@ -9,7 +9,7 @@
     "floating_point_abi": "",
     "floating_point_fpu": "",
     "signedness_of_char": "signed",
-    "signedness_of_enum": "signed",
+    "signedness_of_enum": "unsigned",
```
y4vor, line 12: This part might be problematic for the Omaha config.
@yuying-y please take alook.

y4vor (approved):

PR #10476, opened 2026-05-14
[DRAFT] cobalt/metrics: Add thread based memory allocation hooks

Awallky:
IMPORTANT: STILL A DRAFT! IGNORE!

Cobalt's previous memory metrics lacked subsystem-level granularity,
bucketing allocations into monolithic categories like partition_alloc.
This change introduces an in-process allocation attribution system
to track gross allocation churn per subsystem (DOM, Layout, Media).

Key changes:
- Introduce MemoryContext enum and ScopedMemoryContext RAII object
  using POD thread_local storage to tag execution threads.
- Install lock-free, allocation-free observer hooks into PartitionAlloc
  to intercept allocations and increment static atomic counters.
- Implement CobaltMemoryAttributionManager with a decoupled 5-minute
  repeating timer to report allocation deltas to UMA histograms
  (Memory.Cobalt.AllocationVolume.*), guarding against timer delay outliers.
- Implement robust unit tests and browser integration tests
  (memory_attribution_browsertest.cc) verifying UMA delta reporting.
- Guard upstream modifications in media/starboard with is_cobalt flags.
- Register new histograms in tools/metrics/histograms/metadata/cobalt/histograms.xml.

Bug: 512925852


media/starboard/starboard_renderer.cc
```diff
@@ -21,6 +21,9 @@
 #include "base/strings/stringprintf.h"
 #include "base/task/sequenced_task_runner.h"
 #include "base/trace_event/trace_event.h"
+#if BUILDFLAG(IS_COBALT)
```
kjyoun, line 24: nit: you dont need IS_COBALT here,

since this file is already activated only for starboard.

media/starboard/starboard_renderer.cc
```diff
@@ -752,6 +755,10 @@ void StarboardRenderer::OnDemuxerStreamRead(
     int max_buffers,
     DemuxerStream::Status status,
     DemuxerStream::DecoderBufferVector buffers) {
+#if BUILDFLAG(IS_COBALT)
```
kjyoun, line 758: ditto

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -20,6 +20,9 @@
 #include "base/functional/bind.h"
 #include "base/logging.h"
 #include "build/build_config.h"
+#if BUILDFLAG(IS_COBALT)
```
kjyoun, line 23: no need for IS_COBALT

media/starboard/starboard_renderer.cc
Awallky: Done

media/starboard/decoder_buffer_allocator.cc
Awallky: Fixed.

PR #10478, opened 2026-05-14
media: Guarantee GPU task runner outlives StarboardRenderer

johnedocampo:
Currently in `StarboardRendererWrapper`, `renderer_` is declared before `gpu_task_runner_`. This means that during StarboardRendererWrapper's destruction, `gpu_task_runner_` gets destroyed before `renderer_`. 
  
Upcoming changes will require `StarboardRenderer` to interact with the GPU task runner throughout its entire lifetime (including during its destructor). Relying solely on member variable declaration order in `StarboardRendererWrapper` to prevent use-after-free crashes during teardown 
  is fragile and easily broken by future refactoring.

Bug: 429021006

### DIFF
--- a/media/mojo/services/starboard/starboard_renderer_wrapper.cc
+++ b/media/mojo/services/starboard/starboard_renderer_wrapper.cc
@@ -44,9 +44,8 @@ StarboardRendererWrapper::StarboardRendererWrapper(
                                traits.task_runner),
       video_geometry_setter_service_(traits.video_geometry_setter_service),
       overlay_plane_id_(traits.overlay_plane_id),
-      renderer_(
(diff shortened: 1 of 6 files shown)


media/mojo/services/starboard/starboard_renderer_wrapper.cc
```diff
@@ -58,7 +58,7 @@ StarboardRendererWrapper::StarboardRendererWrapper(
           ,
           std::move(traits.android_overlay_factory_cb)
 #endif  // BUILDFLAG(IS_ANDROID)
-      ) {
+              )) {
```
johnedocampo, line 61: This change was made by pre-commit.

PR #10481, opened 2026-05-14
"Posix": Register ioctl

sacuff:
This adds compile-time "dispatch" for ioctl operations, similar to existing handling that was created for TIOCGWINSZ, and implements a passthrough wrapper for FIONREAD, which is currently the only ioctl command needed by C27 (in the file_path_watcher_inotify lib).

Under this model, `ioctl(fd, op, arg)` is preprocessed to `ioctl_op(fd, arg)`, and individual `ioctl_op` symbols are handled as Starboard usually deals with them. For example, this PR retains the in-musl custom implementation of `ioctl_TIOCGWINSZ`, and registers an ABI wrapper for `ioctl_FIONREAD`.

New ioctl operations added after this PR will fail to build.

Bug: 500418052


jellefoks (approved):

starboard/shared/modular/cobalt_layer_posix_ioctl_abi_wrappers.cc
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <sys/ioctl.h>
+
+extern "C" {
+
+int __abi_wrap_ioctl_FIONREAD(int fd, int* arg);
+
+int ioctl_FIONREAD(int fd, int* arg) {
```
y4vor, line 21: why is the name not just `ioctl`?

sacuff: The macros in third_party/musl/include/sys/ioctl.h redirect `ioctl(fd, FIONREAD, arg)` to this so we can support specific operations and have unsupported ones caught at build time

y4vor: hm, this could be achieved by just excluding the unsupported operations in musl with #if !defined(STARBOARD)

y4vor: @jellefoks, what do you think?

jellefoks (approved):
Agreed. More minimal than pulling in all of ioctl, which we apparently don't need too luch of yet.

y4vor (approved):

PR #10483, opened 2026-05-14
Update Android setup guide for 27.lts release

johnxwork:
Bug: 508340849


arjungm (approved):
LGTM


There might be some redundancy with the setup-linux.md instructions but that's an optimization we can deal with after getting technical writer review. Thanks for getting this done!

cobalt/site/docs/development/setup-android.md
```diff
@@ -1,375 +1,228 @@
...
-    starboard messages
-
-    ```sh
-    adb logcat -s starboard:*
-    ```
-
-1.  To kill any existing running application process (even if it's no longer the
-    active app) use
+```bash
+python3 .agent/skills/cobalt-new-checkout/scripts/cobalt_new_checkout.py --non-interactive --internal --github-user "<your_github_username>"
```
oxve: `--internal` looks wrong for external docs

PR #10488, opened 2026-05-15
Prune Web Bluetooth, WebUSB, WebHID, Web Serial, and Web NFC

johnxwork:
This change systematically eliminates unused peripheral hardware modules from the Cobalt Android APK to reduce binary size by ~295 KB.

Specific changes:
1. third_party/blink/renderer/modules/BUILD.gn:
   - Excluded bluetooth, hid, webusb, serial, and nfc from sub_modules in Cobalt builds.

2. third_party/blink/renderer/bindings/modules/v8/BUILD.gn:
   - Excluded auto-generated V8 wrapper sources for bluetooth, hid, usb, serial, and ndef.

3. third_party/blink/renderer/modules/service_worker/BUILD.gn & service_worker_global_scope.cc:
   - Gated hid and webusb dependencies and event handler lookups behind BUILDFLAG(IS_COBALT).

4. third_party/blink/renderer/modules/cobalt_modules_stubs.cc:
   - Implemented dummy V8 WrapperTypeInfo stubs and C++ accessors for Bluetooth, USB, HID, and Serial to satisfy link-time supplement references.

More savings might be achieved by removing the //services and //device code for these APIs; those couldn't be removed earlier due to the transitive chain of deps (blink --> content --> services --> device), but after this CL it should be possible. 

Bug: 498257436


yell0wd0g (approved):
The delta seems small but at least the idioms used are
known and should be recognizable.

More savings might be achieved by removing the //services
and //device code for these APIs; those couldn't be 
removed earlier due to the transitive chain of deps
(blink --> content --> services --> device), but after
this CL it should be possible. The CL description doesn't
say anything about this,

PR #10489, opened 2026-05-15
Use BUILDFLAG(COBALT_IS_RELEASE_BUILD) in devtool delegate

johnxwork:
Previously, cobalt_content_browser_client.cc used #if defined(COBALT_IS_RELEASE_BUILD), which was always false because COBALT_IS_RELEASE_BUILD is not defined directly as a preprocessor macro. Instead, ENABLE_BUILDFLAG_COBALT_IS_RELEASE_BUILD is set by GN in Gold builds, which powers BUILDFLAG(COBALT_IS_RELEASE_BUILD).

This change updates the preprocessor checks so DevTools manager delegate creation is correctly disabled on Gold builds while remaining enabled on Cobalt QA builds.

Bug: 504725906

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -16,6 +16,8 @@
 
 #include <string>
 
+#include "build/build_config.h"
+
 #include "base/base_switches.h"
 #include "base/command_line.h"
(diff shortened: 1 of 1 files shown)


Awallky (approved):
Thanks for the cleanup!

PR #10491, opened 2026-05-15
Cleanup starboard/configuration.h

y4vor:
Issue: 502584355

### DIFF
--- a/starboard/configuration.h
+++ b/starboard/configuration.h
@@ -104,18 +104,6 @@ struct CompileAssert {};
   extern char _COMPILE_ASSERT_##msg[(expr) ? 1 : -1]
 #endif
 
-// Standard CPP trick to stringify an evaluated macro definition.
-#define SB_STRINGIFY(x) SB_STRINGIFY2(x)
-#define SB_STRINGIFY2(x) #x
-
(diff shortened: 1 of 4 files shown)


niranjanyardi (approved):

starboard/configuration.h
```diff
@@ -104,18 +104,6 @@ struct CompileAssert {};
   extern char _COMPILE_ASSERT_##msg[(expr) ? 1 : -1]
 #endif
 
-// Standard CPP trick to stringify an evaluated macro definition.
-#define SB_STRINGIFY(x) SB_STRINGIFY2(x)
-#define SB_STRINGIFY2(x) #x
-
-// An enumeration of values for the kSbPreferredByteOrder configuration
```
andrewsavage1, line 111: cc @sherryzy 

Previously we had http://github.com/youtube/cobalt/blob/25.lts.1%2B/cobalt/renderer/rasterizer/skia/skia/config/SkUserConfig.h which would optimize swizzles for certain platforms in a way that requires few upstream changes. That said, it looks like we only had platforms that supported the RGBA order. Either way, we should at least take a quick look to see if this is worth bringing to c27 in some way

andrewsavage1 (approved):

PR #10495, opened 2026-05-15
wip

andrewsavage1:


andrewsavage1:
Superseded by a 3-PR stack rebased onto current `main` (this branch was 949 commits behind, and the CI was refactored underneath it in ways that made part of this approach obsolete):

- #12744 — `ci: skip the whole pipeline for documentation-only pull requests`
- #12745 — `ci: add a ci-dry-run label to exercise CI orchestration without building`
- #12746 — `ci: prune per-platform builds with gn analyze`

Notable changes while rebasing:

- The custom gatekeeper job and the `junit_mini_parser` empty-glob fix were dropped — the new `validate-result` job already treats skipped `needs` as passing and guards the empty XML glob.
- `check_pruned` now reads a dedicated `build_pruned` marker instead of injecting a `pruned` key into `test_targets.json`, which is now a flat array consumed by the `test_targets_json` action.
- Fixed a correctness bug: `gn analyze` reports `"No dependency"` both for a source nothing depends on **and** for a path it has never heard of (test expectations, runtime data, Python scripts). gn is now only consulted when every changed file is a compiled source or GN file.
- Dry-run overrides are expressed as one generated platform config file rather than a conditional inside each of the ~29 `Set ...` steps.

PR #10498, opened 2026-05-15
android: prevent TrustStoreAndroid network thread blocking on Startup

ckim1346:
On startup, while `TrustStoreAndroid` is initializing, the thread that holds the startup JNI call for the Android Trust Store `net::android::GetUserAddedRoots()` is still holding onto `init_lock` for ~1.5-2 seconds while the navigation to https://www.youtube.com/tv begins. The BoringSSL TLS handshake is required to verify certificates; the NetworkService thread also then calls `TrustStoreAndroid::SyncGetIssuersOf()` to attempt to acquire init_lock_ as well. 

This PR decouples the locking and allows an early return if the background init is actively in progress, unblocking the thread from having to wait for the root certificate check. This only affect user-added roots in `TrustStoreAndroid`, the system roots (`TrustStoreChrome`) will still be verified.  

Bug: 510376425


net/cert/internal/trust_store_android.cc
```diff
@@ -9,6 +9,7 @@
 #include "base/task/task_traits.h"
 #include "base/task/thread_pool.h"
 #include "base/threading/scoped_blocking_call.h"
+#include "build/build_config.h"
```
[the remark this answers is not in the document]
ckim1346: it's already included on line 8

net/cert/internal/trust_store_android.cc
```diff
@@ -101,6 +102,29 @@ void TrustStoreAndroid::OnTrustStoreChanged() {
...
+  {
+    base::AutoLock lock(init_lock_);
+    if (impl_ && impl_->generation() == current_generation) {
+      return impl_;
+    }
+    if (is_initializing_) {
+      return impl_;
+    }
+    is_initializing_ = true;
+  }
```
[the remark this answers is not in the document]
ckim1346: fair point, updated

net/cert/internal/trust_store_android.h
```diff
@@ -62,6 +63,9 @@ class NET_EXPORT TrustStoreAndroid : public PlatformTrustStore,
 
   base::Lock init_lock_;
   scoped_refptr<Impl> impl_ GUARDED_BY(init_lock_);
+#if BUILDFLAG(IS_COBALT)
+  bool is_initializing_ GUARDED_BY(init_lock_) = false;
```
yell0wd0g, line 72: Every time I see a boolean flag to mark some state, I
automatically ask myself if it could be done with the
existing classes. 

Instead of adding this, it seems like the problem is
that MaybeInitializeAndGetImpl() could get blocked in
the AutoLock? If yes then could you use `AutoTryLock` in
the implementation? AutoTryLock is similar to AutoLock
but doesn't "block" when the lock is already acquired
elsewhere, instead is has a `is_acquired()` method that
can be called to see whether the lock is been acquired
or not. See e.g. [here](https://source.chromium.org/chromium/chromium/src/+/main:chrome/renderer/media/chrome_speech_recognition_client.cc;l=92-95;drc=6921036f402c9e2fecd9f01ea498bb836934a72a;bpv=1;bpt=1)

AutoTryLock: 
https://source.chromium.org/chromium/chromium/src/+/main:base/synchronization/lock.h;l=141;drc=6921036f402c9e2fecd9f01ea498bb836934a72a;bpv=1;bpt=1

yell0wd0g (commented):
The code is hard to read 😓 and there's no unit tests
so have a suggestion to hopefully simplify the net add.

net/cert/internal/trust_store_android.h
ckim1346: So trying to use `AutoTryLock`: if we try to return `impl_` if `!is_acquired` (ie the lock is acquired elsewhere), because impl_ is a scoped_refptr, we get a Wthread-safety-analysis warning as it's not guaranteed that the AutoTryLock will be able to hold it: 

```            
warning: reading variable 'impl_' requires holding mutex 'init_lock_' 
```

Using the bool also lets us do the slow JNI init outside of the lock, whereas with `AutoTryLock` I believe we have to do the init inside of the lock. And doing so, I only saw about a 1.4 second improvement in startup time locally (startup latency time went from about 2.5s down to 1s)

ckim1346: But agreed! Adding comments to improve readability and a few unit tests

ckim1346: Added, PTAL and LMK what you think :)

net/cert/internal/trust_store_android.cc
```diff
@@ -101,6 +102,53 @@ void TrustStoreAndroid::OnTrustStoreChanged() {
...
 TrustStoreAndroid::MaybeInitializeAndGetImpl() {
+#if BUILDFLAG(IS_COBALT)
+  // Allow bypassing the Android trust store (user-added roots) initialization
+  // lock by returning the current impl_ if initialization is already in
+  // progress. Otherwise, set `is_initializing_ = true` and release the lock so
+  // that we can perform the slow constructor work outside of the lock.
+  int current_generation;
+  {
+    base::AutoLock lock(init_lock_);
+    current_generation = generation_.load();
```
yell0wd0g, line 113: I see `generation_` is atomic, could this be done outside the lock
in l.110? And then it could be const.

net/cert/internal/trust_store_android.cc
```diff
@@ -101,6 +102,53 @@ void TrustStoreAndroid::OnTrustStoreChanged() {
...
+    // If another thread is already in the middle of performing the slow
+    // background initialization, return the existing implementation. This is a
+    // non-blocking bypass: we return what we currently have. This could be nullptr
+    // on first init, or the old valid impl_ during a runtime reload to ensure the
+    // network thread never stalls.
+    if (is_initializing_) {
+      return impl_;
+    }
+    
+    // If we are the first thread to detect that the store is out-of-date,
```
yell0wd0g: nit: "out-of-date" or "not existing" ?

net/cert/internal/trust_store_android.cc
```diff
@@ -101,6 +102,53 @@ void TrustStoreAndroid::OnTrustStoreChanged() {
...
+  // lock by returning the current impl_ if initialization is already in
+  // progress. Otherwise, set `is_initializing_ = true` and release the lock so
+  // that we can perform the slow constructor work outside of the lock.
+  int current_generation;
+  {
+    base::AutoLock lock(init_lock_);
+    current_generation = generation_.load();
+    
+    // Return if we already have a fully initialized store for this generation.
+    if (impl_ && impl_->generation() == current_generation) {
```
yell0wd0g: nit: This is short enough that you could have a named bool, i.e.
```
  const bool is_impl_available = impl_ && impl_->generation() == current_generation;
  if (is_impl_available)
    return impl_;
```

Not 100% sure but in Chromium, you can omit {} for single line 
conditionals and bodies. The file style might be different tho.

net/cert/internal/trust_store_android.cc
```diff
@@ -101,6 +102,53 @@ void TrustStoreAndroid::OnTrustStoreChanged() {
...
+    // release the lock so we can perform the init without blocking others.
+    is_initializing_ = true;
+  }
+
+  // Perform the slow constructor work (including JNI calls to Android system)
+  // OUTSIDE of the lock.
+  scoped_refptr<TrustStoreAndroid::Impl> new_impl;
+  {
+    SCOPED_UMA_HISTOGRAM_LONG_TIMER("Net.CertVerifier.AndroidTrustStoreInit");
+    new_impl = base::MakeRefCounted<TrustStoreAndroid::Impl>(current_generation);
```
yell0wd0g: micro-nit: `s/new_impl/tmp/`? Or perhaps `s/new_impl/tmp_impl/`?

yell0wd0g (approved):
I can only find cosmetic nits so LGTM w/ suggestions
to consider or ignore : )

Thanks for the persistence and for the unit tests!

net/cert/internal/trust_store_android.cc
ckim1346: If we do it outside the lock, there's a small possibility that the generation could change before we actually acquire the lock, causing us to read stale data and potentially overwrite a newer trust store with an older version especially since there are so many threads on startup.

net/cert/internal/trust_store_android.cc
ckim1346: updated to say "initialized or updated"

net/cert/internal/trust_store_android.cc
ckim1346: changed to a name bool but compiler didn't allow me to omit the {}

net/cert/internal/trust_store_android.cc
ckim1346: changed to `tmp_impl`

PR #10500, opened 2026-05-15
Remove peerconnection from Chrobalt

haozheng-cobalt:
This PR removes WebRTC P2P component from Chrobalt, and all essential local media and WebRTC-adjacent capture/rendering/encoding features are fully preserved., to be specific:
Standard Chromium platforms compile and execute the WebRTC AEC/audio routing pipeline with 100% backwards compatibility.
Cobalt/pruned platforms run high-speed local microphone capture, software noise suppression, and AGC with absolute safety, completely free of PeerConnection module references.

Reduced 2.8MB for android-arm_gold

```
haozheng@haozheng-usa:~/cobalt/src$ ls -l ~/cobalt/src/out/android-arm_gold/libchrobalt.so
-rwxr-xr-x 1 haozheng primarygroup 77620796 May 16 00:34 /usr/local/google/home/haozheng/cobalt/src/out/android-arm_gold/libchrobalt.so
```

```
haozheng@haozheng-usa:~/cobalt/src$ ls -l ~/cobalt/src/out/android-arm_gold/libchrobalt.so
-rwxr-xr-x 1 haozheng primarygroup 74835516 May 16 00:55 /usr/local/google/home/haozheng/cobalt/src/out/android-arm_gold/libchrobalt.so
```

Bug: 513664168


johnxwork (approved):

johnxwork:
You need better description and title

third_party/blink/renderer/bindings/bindings.gni
```diff
@@ -262,3 +262,12 @@ bindings_unittest_files = get_path_info(
           "modules/v8/v8_element_test.cc",
         ],
         "abspath")
+
+# Exclusion patterns used in Cobalt builds to omit interfaces.
+cobalt_bindings_exclude_patterns = [
```
yell0wd0g, line 267: QQ: Would it be possible to leave the bindings for PC stuff,
and just remove the implementations? By that I mean just
the third_party/blink/renderer/modules/BUILD.gn modifications.
This is because technically we should not remove parts of the
Web Platform, because it makes Cobalt different from other
browsers. But it might be that the generated files won't
compile cleanly and then you need to cut more.

yell0wd0g (approved):
This LGTM given the impact albeit the cut is deep and 
will need revisit upstream. Leaving a question.

third_party/blink/renderer/bindings/bindings.gni
haozheng-cobalt: I investigated whether we could keep the Web IDL bindings for PeerConnection while only omitting the C++ implementations to maintain web platform consistency.
Unfortunately, this isn't feasible without stubbing out all of the WebRTC classes:
1. When we leave the IDL files in `static_idl_files_in_modules`, the bindings generator produces all the V8 binding `.cc` files (e.g., `v8_rtc_peer_connection.cc`).
2. These generated V8 binding files directly call C++ implementation methods on classes like `RTCPeerConnection` (e.g., `RTCPeerConnection::Create(...)`).
3. If we exclude `third_party/blink/renderer/modules/peerconnection` from compilation, the final link step for `libcobalt.so` fails with a large number of unresolved external symbols because the C++ implementation symbols are missing.
Writing and maintaining dummy stub implementations for the entire WebRTC PeerConnection interface to satisfy the linker would introduce significant boilerplate. Therefore, filtering out the IDLs and the generated bindings along with the implementations (as done in this PR) is the cleanest approach to strip WebRTC from the build.

PR #10502, opened 2026-05-15
media: Remove AllocateOnDemand Experiment code

johnedocampo:
Remove currently unused H5VCC code to enable the AllocateOnDemand strategy. We also update |is_memory_pool_allocated_on_demand_| to be const, as there is now no way to change the value during the DecoderBufferAllocator's lifetime.

Bug: 463439461

### DIFF
--- a/media/starboard/decoder_buffer_allocator.h
+++ b/media/starboard/decoder_buffer_allocator.h
@@ -88,7 +88,6 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
 
   // Utility functions for h5vcc settings.
   // TODO(b/460292554): To be deprecated with h5vcc settings.
-  void SetAllocateOnDemand(bool enabled);
   static void EnableConfigurableDecommitStrategy(
       int block_size,
       int retain_blocks,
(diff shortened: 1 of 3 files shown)


media/starboard/decoder_buffer_allocator.h
```diff
@@ -97,7 +96,7 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
   void TryFlushAllocationLog_Locked() EXCLUSIVE_LOCKS_REQUIRED(mutex_);
 #endif  // !BUILDFLAG(COBALT_IS_RELEASE_BUILD)
 
-  bool is_memory_pool_allocated_on_demand_ GUARDED_BY(mutex_);
+  const bool is_memory_pool_allocated_on_demand_;
```
kjyoun, line 105: now this member variable is not used. 

Delete it?

kjyoun:
Sorry, I rebased this commit to ToT by mistake

media/starboard/decoder_buffer_allocator.h
johnedocampo: I don't think we can delete this variable, as it is assigned from [SbMediaIsBufferPoolAllocateOnDemand](https://github.com/youtube/cobalt/blob/main/starboard/media.h#L606).

kjyoun: thanks for clarification

kjyoun (approved):

johnxwork (approved):

PR #10509, opened 2026-05-16
rdk: Add RDK memory benchmark script and documentation

sherryzy:
Created a new memory benchmark script for RDK-based devices to measure usage across critical scenarios and establish a memory baseline. Updated README.md to document its usage, key features like smaps/top RSS tracking, and Mali GPU monitoring.

Bug: 513309842

### DIFF
--- a/cobalt/tools/performance/memory/rdk_memory_benchmark.sh
+++ b/cobalt/tools/performance/memory/rdk_memory_benchmark.sh
@@ -1,0 +1,237 @@
+#!/bin/bash
+
+# Configuration
+readonly PROCESS_NAME="loader_app"
+readonly INTERVAL=1
+readonly OUTPUT_FILE="memory_test_results.txt"
+
+# Initialize output file
+echo "Memory Test Started at $(date)" > "$OUTPUT_FILE"
+
+# Redirect stdout and stderr to both terminal and output file
+exec > >(tee -a "$OUTPUT_FILE") 2>&1
+
+# Define scenarios: name|command|duration|rounds
+# If rounds is omitted, it defaults to 1.
+SCENARIOS=(
+    "Default Page|./out_cobalt/loader_app|60|3"
+    "About Blank|./out_cobalt/loader_app --url=\"about:blank\"|30|1"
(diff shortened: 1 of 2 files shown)


Awallky (commented):
What's the motivation to use a shell script instead of python? Is it that RDKs do not natively include Python?

cobalt/tools/performance/memory/rdk_memory_benchmark.sh
```diff
@@ -0,0 +1,245 @@
+#!/bin/bash
+
+# Configuration
+PROCESS_NAME="loader_app"
```
Awallky: consider making these constants readonly, making them immutable:

```
readonly PROCESS_NAME="loader_app"
readonly INTERVAL=1
readonly OUTPUT_FILE="memory_test_results.txt"
```

Awallky (approved):
This mostly LGTM % comments.
I think the review bot's comments seem valid as well.

sherryzy:
@Awallky Thanks for reviewing. This script measures total RSS to benchmark Cobalt's memory usage, which I use for weekly tracking of our progress. I have also shared it with the client team so they can measure the impact of their experiments on total memory consumption. 

Regarding the choice between Python and shell, do you have a preference? I lean toward a shell script as it is more lightweight and avoids dependencies on Python. Since we only need to extract total RSS and GPU memory from system file for these benchmarks rather than a detailed breakdown, shell seems sufficient.

Awallky:
@sherryzy 

> Regarding the choice between Python and shell, do you have a preference? I lean toward a shell script as it is more lightweight and avoids dependencies on Python. Since we only need to extract total RSS and GPU memory from system file for these benchmarks rather than a detailed breakdown, shell seems sufficient.

I typically prefer Python mostly for readability & testability. 
However, given the simple use case we have here, I think it is reasonable to use a shell script. I would only recommend moving to a Python harness if we find that our fleet of RDK devices format their smaps files differently, which causes parsing failures. 
I.e. if we have to expand the script logic, making it more complex, we can consider switching to enable testability.

PR #10510, opened 2026-05-16
build: Restore missing binary size reduction flags

johnxwork:
This is the result of reviewing https://github.com/youtube/cobalt/pulls?page=2&q=label%3Alightweight+is%3Aclosed

This change restores those flags to ensure a minimal binary footprint across Cobalt builds:
- enable_plugins = false (from PR #7888)
- enable_pdf = false (from PR #7888)
- enable_printing = false (from PR #7888)
- use_vaapi = false (from PR #5515)

### DIFF
--- a/cobalt/build/configs/common.gn
+++ b/cobalt/build/configs/common.gn
@@ -72,6 +72,11 @@ enable_widevine = true
 
 generate_about_credits = false
 
+# Disable plugins, PDF, and printing support to reduce binary size (from PR #7888).
+enable_plugins = false
+enable_pdf = false
+enable_printing = false
+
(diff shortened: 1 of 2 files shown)


cobalt/build/configs/common.gn
```diff
@@ -72,6 +72,11 @@ enable_widevine = true
 
 generate_about_credits = false
 
+# Disable plugins, PDF, and printing support to reduce binary size (from PR #7888).
+enable_plugins = false
```
y4vor, line 76: These 3 flags don't work out of the box in m138. There are additional customization needed to be able to build.

PR #10511, opened 2026-05-16
Remove Topics API behind disable_privacy_sandbox_apis flag

linxinan-yt:
Since the Topics API was deprecated and will be removed from Chromium,
this CL introduces a new GN argument disable_privacy_sandbox_apis
(default false) to allow disabling deprecated Privacy Sandbox APIs at
compile time.
Following PRs will remove the rest deprecated Privacy Sandbox APIs to
reduce the Cobalt size.
NO impact to current 27 Cobalt as its default is false. A another switch
PR will be filed to make this change functional.

Note, those APIs will be removed from Chromium in future milestones. To avoid
rebase conflicts, C++ code changes are tagged with CHROMIUM_MILESTONE_LE_138.

This PR saves about 16k for gold build:

disable_privacy_sandbox_apis on:
```
ls -l out/android-arm_gold/apks/Cobalt.apk
-rw------- 1 linxinan primarygroup 96814250 May 18 23:35 out/android-arm_gold/apks/Cobalt.apk
```

disable_privacy_sandbox_apis off:
```
 ls -l out/android-arm_gold/apks/Cobalt.apk
-rw------- 1 linxinan primarygroup 96830634 May 18 23:21 out/android-arm_gold/apks/Cobalt.apk
```

Bug: 508657772


haozheng-cobalt (dismissed):
For smaller components less than 500kb, I feel like there is no need to create GN flags for them.

Were you be able to find a larger dependent component of this browsing_topcis API? In general it will be much efficient to cut out the bigger component than trimming of individual small APIs in it. I can run the analyzer script to see if there is a bigger dependent of this, and show you.

content/browser/loader/subresource_proxying_url_loader.cc
```diff
@@ -4,7 +4,12 @@
 
 #include "content/browser/loader/subresource_proxying_url_loader.h"
 
+#include "content/public/common/content_milestone_features.h"
+#include "content/public/common/buildflags.h"
+
+#if !BUILDFLAG(DISABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
```
haozheng-cobalt, line 10: Love the idea of CHROMIUM_MILESTONE_LE_138

content/browser/BUILD.gn
```diff
@@ -726,16 +725,6 @@ source_set("browser") {
     "browsing_data/storage_partition_code_cache_data_remover.h",
     "browsing_instance.cc",
     "browsing_instance.h",
-    "browsing_topics/browsing_topics_document_host.cc",
```
haozheng-cobalt, line 729: Use subtraction or filter_exclude, to minimize change.

if (is_cobalt) {
  sources =  filter_exclude(sources, [ "browsing_topics/*" ])
}

content/public/test/test_storage_partition.h
```diff
@@ -171,11 +173,13 @@ class TestStoragePartition : public StoragePartition {
 
   void DeleteStaleSessionData() override {}
 
+#if !BUILDFLAG(DISABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
   void set_browsing_topics_site_data_manager(
       BrowsingTopicsSiteDataManager* manager) {
     browsing_topics_site_data_manager_ = manager;
   }
   BrowsingTopicsSiteDataManager* GetBrowsingTopicsSiteDataManager() override;
+#endif
```
haozheng-cobalt: nit: add comment on the enclosing MACRO #endif // !BUILDFLAG(DISABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138

johnxwork (approved):

content/public/test/test_storage_partition.h
linxinan-yt: Nice catch. Thanks!

content/browser/BUILD.gn
linxinan-yt: Good idea. This could bring less conflicts in the rebase and leave the code level clean up to chromium's next milestone we use.

content/browser/renderer_host/render_frame_host_impl.cc
```diff
@@ -9264,13 +9266,22 @@ void RenderFrameHostImpl::DidChangeIframeAttributes(
...
+  }
+#else
   if (attributes->browsing_topics &&
       !base::FeatureList::IsEnabled(network::features::kBrowsingTopics)) {
     bad_message::ReceivedBadMessage(
         GetProcess(),
         bad_message::RFH_RECEIVED_INVALID_BROWSING_TOPICS_ATTRIBUTE);
     return;
   }
+#endif
```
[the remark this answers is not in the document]
linxinan-yt: @gemini-code-assist  please double check.

linxinan-yt:
> For smaller components less than 500kb, I feel like there is no need to create GN flags for them.
> 
> Were you be able to find a larger dependent component of this browsing_topcis API? In general it will be much efficient to cut out the bigger component than trimming of individual small APIs in it. I can run the analyzer script to see if there is a bigger dependent of this, and show you.

The flag is for entire PS APIs while topics is just the first one. I separated the PRs to make review and revert easy.

content/public/common/features.gni
```diff
@@ -10,4 +10,7 @@ declare_args() {
   # at some point once the feature is stable. See crbug.com/40418135.
   enable_screen_capture = is_linux || is_chromeos || (is_apple && use_blink) ||
                           is_win || is_desktop_android || is_fuchsia
+
+  # If true, Privacy Sandbox APIs (e.g. Topics API) are excluded from the build.
+  disable_privacy_sandbox_apis = false
```
yell0wd0g, line 15: Ok so IIUC this will be applied to this and subsequent
changes rounding up the PS APIs and implementations,
correct?
And in time (150? ) those parts will be all .. gone? Or 
do they define a similar gn var upstream? Just to clarify
the overall plan.

yell0wd0g (approved):
If my understanding is correct and these APIs will be
gone in 150 (we need some concrete milestone, rather
not "in the future") then this LGTM although is a big
cut. Happy to see the `CHROMIUM_MILESTONE_LE_138` 
applied to make rebases easier.

linxinan-yt:
> my understanding is correct and these APIs will be
> gone in 150 (we need some concrete milestone, rather

For topics API, yes. See their estimation https://groups.google.com/a/chromium.org/g/blink-dev/c/_R85yctz4Rs.

content/public/common/features.gni
linxinan-yt: I pushed once crrev.com/c/7801142, but the owner seems not happy about that.

I would leave disable_privacy_sandbox_apis here, because there are still some APIs not deprecated. See b/505811196#c8. So if we have more bandwidth we can investigate and disable them as well if not used by Cobalt.

PR #10515, opened 2026-05-18
android: Refine pending inputs handling in VideoRendererImpl

jasonzhangxx:
Reset the pending end-of-stream flag before calling WriteEndOfStream
while writing pending inputs. This ensures the internal state is
synchronized before the end-of-stream signal is processed by the
player, preventing potential inconsistencies in state tracking.

Bug: 513852955

### DIFF
--- a/starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
+++ b/starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
@@ -491,6 +491,7 @@ void VideoRendererImpl::WritePendingInputs() {
   if (pending_end_of_stream_ && pending_input_buffers_.empty()) {
     // All pending buffers have been sent to the decoder; now it's safe to
     // send the end-of-stream signal.
+    pending_end_of_stream_ = false;
     WriteEndOfStream();
   }
 }


starboard/shared/starboard/player/filter/video_renderer_internal_impl.cc
```diff
@@ -491,6 +491,7 @@ void VideoRendererImpl::WritePendingInputs() {
   if (pending_end_of_stream_ && pending_input_buffers_.empty()) {
     // All pending buffers have been sent to the decoder; now it's safe to
     // send the end-of-stream signal.
+    pending_end_of_stream_ = false;
```
kjyoun, line 494: I have no problem with this logic change, but we really need unit tests to protect it and verify the intended impact.

Since this looks like an optimization, if we don't have a test or benchmark to validate it, we'll have no way of knowing if a future change accidentally breaks or reverts it. I think that we shouldn't have to rely purely on manual code review to catch performance regressions.

jasonzhangxx: Yes, I totally agree we should add more unit tests. I created b/514405367 for that.

kjyoun (approved):

PR #10516, opened 2026-05-18
android: Remove redundant deps from cobalt_apk

hferreiro:
Remove unused dependencies from //cobalt/android:cobalt_apk. These dependencies are already present in the android_library targets where they are needed.

Bug: 492403745


andrewsavage1:
LGTM but @haozheng-cobalt or @Awallky ptal

Awallky (approved):
LGTM

PR #10517, opened 2026-05-18
android: Remove unneeded not_empty.txt asset placeholders

hferreiro:
Remove unneeded not_empty.txt placeholders from
//cobalt/android:coat_assets. Since the target was not using renaming_sources/destinations, both paths were flattened to assets/not_empty.txt. In addition, since the logo file was already present at assets/, none of them are actually needed.

Bug: 492403745

### DIFF
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -194,11 +194,7 @@ android_library("cobalt_apk_java") {
 }
 
 android_assets("coat_assets") {
-  sources = [
-    "apk/app/src/app/assets/not_empty.txt",
-    "apk/app/src/app/assets/test/not_empty.txt",
-    "apk/app/src/app/assets/web/cobalt_logo_1024.png",
-  ]
+  sources = [ "apk/app/src/app/assets/web/cobalt_logo_1024.png" ]
   disable_compression = true
 }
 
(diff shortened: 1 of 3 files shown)


cobalt/android/BUILD.gn
```diff
@@ -194,11 +194,7 @@ android_library("cobalt_apk_java") {
 }
 
 android_assets("coat_assets") {
-  sources = [
-    "apk/app/src/app/assets/not_empty.txt",
```
andrewsavage1, line 198: @johnxwork any context on why these were added? Seems fine to remove them now

andrewsavage1 (approved):

cobalt/android/BUILD.gn
johnxwork: I believe these were added to make sure the "test" folder is not optimized out during test binary packaging. The apk builder on Android would just not package empty directory, at least a few years ago.

andrewsavage1: Talked to John, these were apparently added for nplb tests, so you might run into this issue when trying to run nplb tests later. It could be that we don't need it anymore, but keep it in mind if you see bugs in apk packaging/apk content during nplb tests in the future.

johnxwork (approved):

PR #10524, opened 2026-05-18
Revert "Use BUILDFLAG(COBALT_IS_RELEASE_BUILD) in devtool delegate"

haozheng-cobalt:
Reverts youtube/cobalt#10489

App crashes due to this change.

https://paste.googleplex.com/6341895293894656

Bug: 504725906

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -32,7 +32,6 @@
 #include "base/strings/utf_string_conversions.h"
 #include "base/time/time.h"
 #include "base/timer/elapsed_timer.h"
-#include "build/build_config.h"
 #include "cobalt/browser/cobalt_browser_interface_binders.h"
 #include "cobalt/browser/cobalt_browser_main_parts.h"
 #include "cobalt/browser/cobalt_secure_navigation_throttle.h"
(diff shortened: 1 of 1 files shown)


haozheng-cobalt:
Gemini comments:


```
Root Cause: Recent Regression
  The crash was introduced by commit 4f8b62c ("Use BUILDFLAG(COBALT_IS_RELEASE_BUILD) in devtool delegate"), which was submitted
  yesterday.

  Before this change, CobaltContentBrowserClient::CreateDevToolsManagerDelegate() was using a faulty check:

   1 // OLD CODE (Faulty, always defined DevToolsManagerDelegate)
   2 #if defined(COBALT_IS_RELEASE_BUILD)
   3   return nullptr;
   4 #else
   5   return content::ShellContentBrowserClient::CreateDevToolsManagerDelegate();
   6 #endif
  COBALT_IS_RELEASE_BUILD was never defined as a direct macro, so the #else branch always ran, providing a valid delegate.

  The new code correctly identifies that you are running a Gold (Release) build (indicated by out/android-arm_gold in your prompt) and
  returns nullptr:

   1 // NEW CODE (Correctly identifies Gold build)
   2 #if BUILDFLAG(COBALT_IS_RELEASE_BUILD)
   3   return nullptr;  // Returns nullptr on Gold builds
   4 #else
   5   return content::ShellContentBrowserClient::CreateDevToolsManagerDelegate();
   6 #endif

  Why it Crashes
  While the delegate is now correctly nullptr on Gold builds, the Chromium code in DevToolsAgentHost::StartRemoteDebuggingServer (frame
  #0 in your trace) has a mandatory CHECK(delegate). Since DevTools is explicitly disabled on Gold builds, any attempt to start the
  DevTools server (e.g., via command line flags or default setup) will now trigger this fatal assertion.

  Recommendation
   1. If you need DevTools: Switch to a QA build (android-arm_qa), where COBALT_IS_RELEASE_BUILD is false and the delegate is provided.
   2. If you are testing a Gold build: Ensure that DevTools is not being requested. Note that on Gold builds, command-line arguments
      are often disabled or ignored to prevent enabling DevTools in production.

```

johnxwork (approved):

PR #10528, opened 2026-05-18
Add java switches for tuning compositor tile configurations

tasunnn:
Add `InterestAreaSizeInPixels` and `ReclaimDelayInSeconds` Java switches
to allow runtime experimentation with compositor offscreen interest
area size and tile reclaim delays. These switches map to the
SmallerInterestArea feature flag and will help in investigating
opportunities for memory reduction on Android devices.

Bug: 512154139

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
@@ -65,12 +65,12 @@ public class JavaSwitches {
     StringJoiner featureParams = new StringJoiner("/");
     if (javaSwitches.containsKey(JavaSwitches.INTEREST_AREA_SIZE_IN_PIXELS)) {
       String size = javaSwitches.get(JavaSwitches.INTEREST_AREA_SIZE_IN_PIXELS).replaceAll("[^0-9]", "");
-      featureParams.add("size_in_pixels/" + size); 
+      featureParams.add("size_in_pixels/" + size);
     }
 
     if (javaSwitches.containsKey(JavaSwitches.RECLAIM_DELAY_IN_SECONDS)) {
       String delay = javaSwitches.get(JavaSwitches.RECLAIM_DELAY_IN_SECONDS).replaceAll("[^0-9]", "");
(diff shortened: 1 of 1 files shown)


sherryzy (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -56,6 +62,22 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
       extraCommandLineArgs.add("--js-flags=--no-decommit-pooled-pages");
     }
 
+    StringJoiner featureParams = new StringJoiner("/");
```
Awallky, line 66: Do we need an import for `StringJoiner`?

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -56,6 +62,22 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
...
+    if (javaSwitches.containsKey(JavaSwitches.INTEREST_AREA_SIZE_IN_PIXELS)) {
+      String size = javaSwitches.get(JavaSwitches.INTEREST_AREA_SIZE_IN_PIXELS).replaceAll("[^0-9]", "");
```
Awallky, line 68: What is the rationale behind `replaceAll`?

Awallky (approved):
LGTM % comments

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
tasunnn: you are right, thanks!

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
tasunnn: this is just safety check in case of any spaces/unit appended to the string and if we couldn't guarantee it will be all digits

PR #10556, opened 2026-05-20
[cobalt] Restore GlobalFeatures setting reader for HangReporting

alunev:
cobalt: Restore GlobalFeatures hang reporting

Restore the logic in CobaltHangWatcherDelegate to check the
EnableHangReporting setting from GlobalFeatures. This serves as a
fallback or alternative to the Finch flag bridge, allowing embedders
to enable hang reporting on Cobalt 27.

Bug: 510831691

### DIFF
--- a/cobalt/browser/hang_watcher_delegate_impl.cc
+++ b/cobalt/browser/hang_watcher_delegate_impl.cc
@@ -19,7 +19,7 @@
 
 #include "base/feature_list.h"
 #include "cobalt/browser/features.h"
-
+#include "cobalt/browser/global_features.h"
 namespace cobalt {
 namespace browser {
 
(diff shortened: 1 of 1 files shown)


cobalt/browser/hang_watcher_delegate_impl.cc
```diff
@@ -30,6 +30,13 @@ void CobaltHangWatcherDelegate::Initialize() {
 }
 
 bool CobaltHangWatcherDelegate::IsHangReportingEnabled() {
+  const auto& settings = cobalt::GlobalFeatures::GetInstance()->GetSettings();
+  auto it = settings.find("EnableHangReporting");
+  if (it != settings.end()) {
+    if (const auto* val = std::get_if<int64_t>(&it->second)) {
+      return *val != 0;
+    }
+  }
```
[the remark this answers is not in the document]
alunev: fixed, thanks

johnxwork:
It feels a little strange to me that we are prioritizing GlobalFeature list. Adding Yijia to for Finch opinion.

yjzhang111:
@johnxwork Could you explain what prioritizing GlobalFeature list means?
Side note (unrelated to this particular PR) - Global Features is NOT Finch specific, it just so happens that everything in there is Finch/UMA related right now. It's intended to hold all features/objects that are scoped to the entire browser process. See link below
https://source.chromium.org/chromium/chromium/src/+/main:chrome/browser/browser_process.h;l=5%3D18

h5vcc setting is technically not Finch, even tho (I think?) it's heavily used by the media team to pass in feature flags and values as a workaround without Finch. This looks ok but do we anticipate to move this to Finch once it's ready or do we intend to keep it running as a Kabuki experiment (status quo)?

cobalt/browser/hang_watcher_delegate_impl.cc
```diff
@@ -30,6 +30,15 @@ void CobaltHangWatcherDelegate::Initialize() {
 }
 
 bool CobaltHangWatcherDelegate::IsHangReportingEnabled() {
```
johnxwork, line 32: The difference between "GlobalFeatures" and "FeatureList" is confusing. It certainly was for me. The bad naming does not need to be addressed by this PR. Might be helpful to add some comments in the code.

johnxwork (approved):

cobalt/browser/hang_watcher_delegate_impl.cc
alunev: Thanks, added clarification comment.

PR #10557, opened 2026-05-20
media: Simplify video decoder capability checks

kjyoun:
Add overloads for FindVideoDecoder and HasVideoDecoderFor in the
MediaCapabilitiesCache to allow omitting resolution and performance
parameters when they are not required.

These changes simplify several call sites in MediaCodecBridge and
PlayerComponentsFactory, replacing verbose zero-initialized parameter
lists with cleaner method calls. This also ensures the secure
pipeline check logic in PlayerComponentsFactory is correctly
implemented.

Issue: 414009070


starboard/android/shared/media_capabilities_cache.h
```diff
@@ -198,8 +198,16 @@ class MediaCapabilitiesCache {
                           int bitrate,
                           int fps);
 
+  bool HasTunnelModeVideoDecoderFor(const std::string& mime_type,
```
jasonzhangxx: It's a little weird that we add a function specifically for tunnel mode.

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -271,4 +271,132 @@ TEST_F(MediaCapabilitiesCacheTest, ClearCacheClearsAllValues) {
   EXPECT_FALSE(cache_->GetAudioConfiguration(2, &config));
   EXPECT_FALSE(cache_->HasAudioDecoderFor("audio/mp4", 192000));
 }
+
+TEST_F(MediaCapabilitiesCacheTest, HasTunnelModeVideoDecoderFor_Supported) {
```
jasonzhangxx: Thanks @kjyoun! We do need more tests to cover tunnel mode and `HasVideoDecoderFor`.

@johnedocampo to review the test code.

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -271,4 +271,132 @@ TEST_F(MediaCapabilitiesCacheTest, ClearCacheClearsAllValues) {
...
+  auto populate_caps =
```
johnedocampo: This lambda and the `populate_caps` lambda in the other tests look to be extremely similar, with the only difference being the configuration used to create the CodecCapability entry. To help reduce the code duplication, would it make sense to have a helper function generate a lambda for this call given a specific configuration?

starboard/android/shared/media_capabilities_cache.h
```diff
@@ -198,8 +198,16 @@ class MediaCapabilitiesCache {
...
+                                    bool must_support_secure);
+
   std::string FindAudioDecoder(const std::string& mime_type, int bitrate);
 
+  std::string FindVideoDecoder(const std::string& mime_type,
```
johnedocampo, line 207: Correct me if I am wrong, but it looks like with this definition, the original `FindVideoDecoder` is no longer needed in the codebase. Can we just remove the original definition then and only have this new one?

starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -432,11 +432,22 @@ bool MediaCapabilitiesCache::HasVideoDecoderFor(const std::string& mime_type,
...
+                           /*require_software_codec=*/false,
+                           must_support_tunnel_mode, frame_width, frame_height,
+                           bitrate, fps)
               .empty();
 }
 
+bool MediaCapabilitiesCache::HasTunnelModeVideoDecoderFor(
+    const std::string& mime_type,
+    bool must_support_secure) {
+  return HasVideoDecoderFor(mime_type, must_support_secure,
```
johnedocampo: While returning `HasVideoDecoderFor` is okay for now as it reroutes to `FindVideoDecoder`, I imagine situations where if `HasTunnelModeVideoDecoderFor` breaks, it could actually be the case that `HasVideoDecoderFor` is at issue. To decouple this connection, can we have this function instead return:

```
return !FindVideoDecoder(mime_type, must_support_secure, /*must_support_hdr=*/false,
                            /*must_support_tunnel_mode=*/true,
                            /*frame_width=*/0, /*frame_height=*/0,
                            /*bitrate=*/0, /*fps=*/0).empty();

```

This can help better distinguish issues in the future when debugging provider implementations.

starboard/android/shared/media_capabilities_cache_test.cc
johnedocampo: Yes I agree! Thanks for adding more tests here.

starboard/android/shared/media_capabilities_cache_test.cc
```diff
@@ -271,4 +271,132 @@ TEST_F(MediaCapabilitiesCacheTest, ClearCacheClearsAllValues) {
...
+
+  EXPECT_CALL(*mock_media_capabilities_provider_,
+              GetCodecCapabilities(testing::_, testing::_))
+      .WillOnce(testing::Invoke(populate_caps));
+
+  EXPECT_FALSE(cache_->HasTunnelModeVideoDecoderFor(
+      "video/x-vnd.on2.vp9", /*must_support_secure=*/false));
+}
+
+TEST_F(MediaCapabilitiesCacheTest, HasVideoDecoderFor_ResolutionLimits) {
```
johnedocampo, line 293: Side comment that could be tackled in another PR: This test currently checks for two false instances: when the resolution isn't supported and when the fps isn't supported. There can be an idea where we introduce a mechanism to test all permutations of the configurations to make sure each field is correctly checked. Again, just thinking out loud and is likely out of scope of this PR.

johnedocampo (commented):
Thanks for adding the new tests. Left some comments.

FWIW: I've had https://github.com/youtube/cobalt/pull/9318 in my backlog for some time which adds additional support for FindVideoDecoder and FindAudioDecoder (haven't had the moment to bring this one up for review). If you think it would be helpful for this PR, let me know and I can get that one ready for review. If not, I can just rebase after this one merges and adjust my PR accordingly.

starboard/android/shared/media_capabilities_cache_test.cc
kjyoun: ack

starboard/android/shared/media_capabilities_cache.cc
kjyoun: I think that the right approach here to make sure `HasTunnelModeVideoDecoderFor` not broken.

Here is Gemini comment that I got and I agree

---

3. "Coding to Avoid Bugs in Helpers" is an Anti-Pattern

  The reviewer's concern that "if  HasTunnelModeVideoDecoderFor  breaks, it could actually be the case that  HasVideoDecoderFor  is at issue" is a fallacious argument for decoupling.

 - Fix the root cause: If  HasVideoDecoderFor  is buggy, it should be fixed. Working around potential bugs in helper methods by duplicating code is an anti-pattern that leads to "spaghetti code" and maintenance nightmares.
  - Unit Tests are the solution: The way to ensure  HasVideoDecoderFor  doesn't break  HasTunnelModeVideoDecoderFor  is to have robust unit tests for both methods (which you have already added in this CL). If  HasVideoDecoderFor  breaks, its own unit tests should catch it.

starboard/android/shared/media_capabilities_cache.h
kjyoun: It is still used

https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/media_is_video_supported.cc;l=108-110;drc=385aa8bf6a63fe109c8907baaf094b8c3c526aa6

https://source.corp.google.com/h/github/youtube/cobalt/+/main:starboard/android/shared/media_capabilities_cache.cc;l=426-438;drc=77ffa0509e72d9ebbad8c1ae18397538498e5dd2

starboard/android/shared/media_capabilities_cache_test.cc
kjyoun: Good point. 

Added helper method, though I intentionally did not hide the entire lambda.

Per go/too-dry, it is beneficial to keep key setup details—like the mime-type mapping—visible inside the test block:
 
```
   video_caps["video/x-vnd.on2.vp9"] = ...
```

starboard/android/shared/media_capabilities_cache_test.cc
kjyoun: I added more test checks, though they don't cover all permutations, which needs to be a separate PR

starboard/android/shared/media_capabilities_cache.h
kjyoun: agree. it is weird that this generic media capabilities cache mention a specific feature.

I moved this helper method to `player_component_factory.cc`

starboard/android/shared/media_capabilities_cache.h
kjyoun: realized 9-param version is used only inside this file. 

I moved it to private method

jasonzhangxx (commented):
LGTM, I'll let @johnedocampo to continue the review of tests and make the final approve.

starboard/android/shared/media_capabilities_cache.cc
johnedocampo: Ack, let's keep it as-is.

johnedocampo (approved):

PR #10564, opened 2026-05-20
Fix workflow secrets error in composite action

oxve:
This PR fixes the 'Unrecognized named-value: secrets' error in the composite action by passing the secret as an input. Minimal change for release branch.

BUG=508469475
TAG=agy
CONV=b353fb67-a1fd-4829-890c-12c9bc6ae550

### DIFF
--- a/.github/actions/on_device_tests/action.yaml
+++ b/.github/actions/on_device_tests/action.yaml
@@ -1,6 +1,11 @@
 name: On Device Test
 description: Runs on-device tests.
 
+inputs:
+  DD_API_KEY:
+    description: 'DataDog API Key'
+    required: true
+
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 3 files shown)


.github/actions/on_device_tests/action.yaml
```diff
@@ -1,6 +1,11 @@
 name: On Device Test
 description: Runs on-device tests.
 
+inputs:
+  DD_API_KEY:
+    description: 'DataDog API Key'
+    required: true
```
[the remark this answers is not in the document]
oxve: Done

docker-compose.yml
```diff
@@ -189,9 +193,11 @@ services:
     scale: 0
 
   linux-x64x11-gcc:
-    <<:
-      - *common-definitions
-      - *build-volumes
+    <<: *common-definitions
+    volumes:
+      - ${COBALT_SRC:-.}:/code/
+      - android-debug-keystore:/root/.android/
```
arjungm, line 199: nit: do all these linux platforms need this mounted? if it's possible to remove them let's do that, but otherwise it probably isn't an issue

arjungm (approved):

docker-compose.yml
oxve: Probably not needed but they were mounted before so I'm not changing it. I'm not sure why docker suddenly decided that the old way of mounting the volumes isn't supported anymore.

PR #10566, opened 2026-05-20
starboard: use gold linker instead of default ld

santoshmahto7:
starboard/RDK: Use gold linker instead of the default ld.
    
The RDK platform now explicitly uses the gold linker instead of the
default linker. This change addresses linker errors[1] encountered
while building the loader_app using the RDK7 toolchain.
The gold linker is also known to be faster, uses less memory,
and more efficient than the default ld.
    
[1]
```
aarch64-rdk-linux/usr/lib/libEGL.so: .dynsym local symbol at index 3 (>= sh_info of 3)
aarch64-rdk-linux/usr/lib/libEGL.so: .dynsym local symbol at index 4 (>= sh_info of 3)
arch64-rdk-linux/usr/lib/libEGL.so: .dynsym local symbol at index 5 (>= sh_info of 3)
collect2: error: ld returned 1 exit status 
```

Bug: 510541109

Change-Id: I5fbb070079c6727e456da1e22f72bec68f6de0b5


niranjanyardi (approved):
please add a more detailed PR description as to why this was needed ( i got context from bug but ideally the PR description explains it )

santoshmahto7:
> please add a more detailed PR description as to why this was needed ( i got context from bug but ideally the PR description explains it )

Done!!

PR #10568, opened 2026-05-20
aosp: Remove arm_use_neon=false

hferreiro:
Follow-up from https://github.com/youtube/cobalt/pull/10454, in which neon was enabled for evergreen softfp builds.

For consistency, remove it from evergreen-arm-softfp too, so that the default is always used.

Bug: 453022760

### DIFF
--- a/cobalt/build/configs/aosp-arm/args.gn
+++ b/cobalt/build/configs/aosp-arm/args.gn
@@ -4,9 +4,5 @@ target_cpu = "arm"
 starboard_target_platform = "aosp-arm"
 arm_float_abi = "softfp"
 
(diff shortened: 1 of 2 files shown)


hferreiro:
@andrewsavage1 would it make sense to also remove from evergreen-arm-softfp/args.gn to use the default? Or set to true here?

andrewsavage1 (approved):

andrewsavage1:
> @andrewsavage1 would it make sense to also remove from evergreen-arm-softfp/args.gn to use the default? Or set to true here?

Yeah I think we should remove it there and always use the default

hferreiro:
@andrewsavage1 I've just removed it.

PR #10577, opened 2026-05-20
cobalt/build/configs: Disable compute_pressure on Linux-derivative builds

yell0wd0g:
It's enabled by default. Stumbled upon this while doing something 
else in .//content/browser/.../BUILD.gn.

Locally and for my linux-x64x11_qa build the size reduction is
from 814245184 to 813419008, so 826176Bytes, I expect the number
to much smaller in a gold-stripped build but we'll take it given
that it's trivial to remove, and we don't nor will use it. 

https://developer.mozilla.org/en-US/docs/Web/API/Compute_Pressure_API

Bug: 505827769


yell0wd0g:
@haozheng-cobalt PTAL

haozheng-cobalt (changes requested):
The ROI for this one-liner change is not small. Don't we want to disable this for all platforms?

yell0wd0g:
> The ROI for this one-liner change is not small. Don't we want to disable this for all platforms?

It's already disabled on androidtv and tvOS https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:services/device/public/cpp/compute_pressure/buildflags.gni;l=12?q=enable_compute_pressure&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain&start=1 .

haozheng-cobalt (approved):
Nice, thanks!

PR #10578, opened 2026-05-20
media: Implement Direct Renderer for single-process

borongc:
Introduce a Direct Renderer path to optimize media playback in
single-process mode. This allows the Starboard renderer client to
communicate directly with the renderer service via a C++ interface,
bypassing Mojo IPC overhead.

The feature is controlled by the CobaltUsingDirectRenderer feature
flag and the Media.UseDirectRenderer H5vcc setting, and requires the
single-process command-line switch.

Issue: 513254071


borongc:
Hand over to https://github.com/youtube/cobalt/pull/11262.

PR #10581, opened 2026-05-21
cobalt/testing: Fix browsertests for tvOS

Awallky:
Wrap granular memory browsertests under IS_LINUX or IS_ANDROID to fix macOS/tvOS compilation

Bug: 515200295


cobalt/testing/browser_tests/granular_memory_browsertest.cc
```diff
@@ -21,6 +21,8 @@
 
 namespace content {
 
+#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_ANDROID)
```
linxinan-yt: If the entire file is not for tvOS, why not add the conditional build to the build.gn file?

linxinan-yt:
FYI: you can attach kokoro:run label to trigger the a macOS build(still in experiment though).

cobalt/testing/browser_tests/granular_memory_browsertest.cc
Awallky: Great point! I modified to exclude it for `is_ios` builds

Awallky:
@rakuco Has a fix for browsertests and cobalt builds.

PR #10584, opened 2026-05-21
Remove Interest Groups from C27

haozheng-cobalt:
Disable the Privacy Sandbox Interest Group (Ad Auction/Fledge)
functionality by introducing the enable_interest_groups build flag,
which is set to false for Cobalt.

This change excludes the auction worklet service from the build and
wraps related storage, DevTools, and Blink modules in preprocessor
guards. Removing these unused components helps reduce binary size
and simplifies the Cobalt runtime.

Automation helper scripts results:
analyze_feature_disable_difficulty results: https://paste.googleplex.com/4650544823140352
json input to the SKILL.md: https://paste.googleplex.com/6339204429774848

Size savings:
This PR saves 1.1 MB on android-arm_gold.

Before:
```
haozheng@haozheng-usa:~/cobalt/src$ ls -l ~/cobalt/src/out/android-arm_gold/libchrobalt.so
-rwxr-xr-x 1 haozheng primarygroup 74188348 May 22 05:26 /usr/local/google/home/haozheng/cobalt/src/out/android-arm_gold/libchrobalt.so
```
After:
```
haozheng@haozheng-usa:~/cobalt/src$ ls -l ~/cobalt/src/out/android-arm_gold/libchrobalt.so
-rwxr-xr-x 1 haozheng primarygroup 73082428 May 22 05:20 /usr/local/google/home/haozheng/cobalt/src/out/android-arm_gold/libchrobalt.so
```

Bug: 504691422


content/browser/BUILD.gn
```diff
@@ -3766,6 +3766,21 @@ source_set("browser") {
...
+      "//content/services/auction_worklet/public/mojom",
+      "//content/browser/interest_group:interest_group_proto",
+    ]
+    _filtered_sources = filter_exclude(sources, [
+      "interest_group/*",
+      "devtools/auction_worklet_devtools_agent_host.cc",
+      "devtools/auction_worklet_devtools_agent_host.h",
+    ])
+    sources = []
+    sources = _filtered_sources
```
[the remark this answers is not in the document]
haozheng-cobalt: this is to avoid the warning of overriding non-empty gn variable.

content/browser/storage_partition_impl.cc
```diff
@@ -2759,7 +2769,12 @@ void StoragePartitionImpl::ClearDataImpl(
       std::move(cookie_deletion_filter), GetPath(), dom_storage_context_.get(),
       quota_manager_.get(), special_storage_policy_.get(),
       filesystem_context_.get(), GetCookieManagerForBrowserProcess(),
-      interest_group_manager_.get(), attribution_manager_.get(),
+#if BUILDFLAG(ENABLE_INTEREST_GROUPS)
+      interest_group_manager_.get(),
+#else
+      nullptr,
+#endif  // BUILDFLAG(ENABLE_INTEREST_GROUPS)
```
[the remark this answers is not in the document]
haozheng-cobalt: DataDeletionHelper::ClearDataOnUIThread handles a nullptr for the interest_group_manager parameter completely and safely

content/browser/BUILD.gn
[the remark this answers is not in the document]
haozheng-cobalt: gn args does not support overriding non empty variable

third_party/blink/renderer/bindings/bindings.gni
```diff
@@ -264,10 +266,35 @@ bindings_unittest_files = get_path_info(
...
+}
+
+if (!enable_interest_groups) {
+  cobalt_bindings_exclude_patterns += [
+    "*ad_auction*",
+    "*auction_ad*",
+    "*protected_audience*",
+    "*storage_interest_group*",
+    "*view_or_click_counts*",
+    "*_ads*",
```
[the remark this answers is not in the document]
haozheng-cobalt: this matches "v8_ads.cc" for now, safe and unique to FLEDGE APIs.

content/browser/devtools/protocol/browser_handler.cc
```diff
@@ -29,7 +29,10 @@
 #include "content/browser/devtools/devtools_manager.h"
 #include "content/browser/devtools/protocol/devtools_download_manager_delegate.h"
 #include "content/browser/gpu/gpu_process_host.h"
-#include "content/browser/interest_group/interest_group_manager_impl.h"
+#include "third_party/blink/public/common/buildflags.h"
+#if BUILDFLAG(ENABLE_INTEREST_GROUPS)
```
linxinan-yt, line 33: Can we use DISABLE_PRIVACY_SANDBOX_APIS to save one build flag? Because I have other APIs to remove by this flag. It is easier to use one flag to flip all privacy sandbox APIs later.

haozheng-cobalt: add a TODO for better naming as communicated offline.

linxinan-yt (approved):

third_party/blink/renderer/modules/ad_auction/shared_storage_worklet_global_scope.h
andrewsavage1, line 1: Why do we need this file?

third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_global_scope.idl
```diff
@@ -28,12 +28,6 @@ interface SharedStorageWorkletGlobalScope : WorkletGlobalScope {
   ]
   readonly attribute Crypto crypto;
 
-  [
```
andrewsavage1, line 31: I really dislike this method, I would much rather try to stub this out

andrewsavage1 (commented):
This change is really invasive. Is there a way to instead create stub classes for auctions/interest groups? Can you give a sense of what that might look like or if it's even possible?

haozheng-cobalt:
> This change is really invasive. Is there a way to instead create stub classes for auctions/interest groups? Can you give a sense of what that might look like or if it's even possible?

yes, I've actually considered this.  The reason I was not going for this approach is when heavy data types are used outside the class to be stubbed complicated changes will still be needed across codebase.  I'm not sure how practical it will be.

third_party/blink/renderer/modules/ad_auction/shared_storage_worklet_global_scope.h
haozheng-cobalt: yea, this is the tricky part because because the Interest groups is partial web IDL of storage partition: third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_global_scope.idl contains the interest group/ad auction definition which I think cannot be conditionally excluded. So I extracted the interest group/ad auction definition into it's separate files third_party/blink/renderer/modules/shared_storage/shared_storage_worklet_global_scope.idl and third_party/blink/renderer/modules/ad_auction/shared_storage_worklet_global_scope.h

PR #10593, opened 2026-05-21
cobalt 3p: enable compositor features for tile cache memory reduction

tasunnn:
This change enables three compositor features that are new to C27: 

 - ``base::Feature:smallerInterestArea``: reduces the default interest 
    area padding for offscreen prepaint from 3000px down to 500px default
- ``base::Feature:ReclaimPrepaintTilesWhenIdle``: actively evicts 
    prepaint tiles outside the visible viewport whenever the compositor enters an 
    idle state. 
- ``base::Feature:ReclaimOldPrepaintTiles``: purges prepaint tiles that 
    have not been accessed or viewed within a specified timeout (default 30s) 
     
From local testing, memory reduction is more significant on Android but will also 
enable them for 3p since they will still help relieve Cobalt's memory 
pressure (~5MB RSS). 

Bug: 512154139


cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -89,7 +89,9 @@ CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
       // creation, but egl extension required to support the attribute is
       // missing and causes errors. So Enable it by default.
       {::switches::kEnableFeatures,
-       "LimitImageDecodeCacheSize:mb/24, DefaultEnableANGLEValidation"},
+       "LimitImageDecodeCacheSize:mb/24, DefaultEnableANGLEValidation, "
```
yell0wd0g: I wonder if we could have each flag in a separate line?
So that git blame can separate the history of each.

yell0wd0g (approved):
Ship it! :)

cobalt/app/cobalt_switch_defaults_starboard.cc
tasunnn: yeah let me try it out and see if lint yells at me on that

yell0wd0g: Actually I saw that the comments in l.87-90 apply to the
`DefaultEnableANGLEValidation`, so if you could reshuffle
the comment and lines now, best-est.

yell0wd0g: Feel free to leave a link to b/444042898 in those comments :P
(Sorry I'm saddling you with cleaning this up)

tasunnn: looks like it works! yes I will move the comments and add bug number

PR #10607, opened 2026-05-22
Add Gemini review guide for binary size reduction and feature removal

haozheng-cobalt:
This pull request updates the SKILL.md documentation to establish a core philosophy of surgical minimality and upstream friendliness for feature removal. It adds detailed guidelines on auditing Web IDL and V8 bindings, addressing compiler warnings, enforcing transitive buildflag includes, and implementing the partial IDL interface pattern to handle cross-module dependencies. 

Bug: 515786375

### DIFF
--- a/cobalt/tools/binary_size/SKILL.md
+++ b/cobalt/tools/binary_size/SKILL.md
@@ -9,6 +9,15 @@ This skill guides an AI coding agent through the step-by-step process of disabli
 
 ---
 
+> [!IMPORTANT]
+> **Core Philosophy: Surgical Minimality & Upstream Friendliness**
+> The paramount principle of feature removal in Cobalt/Chrobalt is that **every change must be as minimal and upstream-friendly as possible to avoid future merge conflicts**.
+> 1. **Never delete code or headers unnecessarily**: Only gate or prune feature-specific files. Keep core/shared files (e.g., standard Chromium C++ files, standard headers) fully intact and untouched.
+> 2. **Never break transitive include assumptions**: Avoid removing standard C++ headers that are not strictly owned by the excluded feature. Upstream merges assume core files import their expected core headers.
+> 3. **Utilize dynamic pruning over static edits**: Use `bindings.gni` dynamic exclude patterns and GN `filter_exclude` rather than editing static lists like `generated_in_modules.gni` to keep downstream maintenance completely conflict-free.
+
+---
+
 ## 1. High-Level Workflow Overview
 
 The feature removal process follows a four-phase methodology:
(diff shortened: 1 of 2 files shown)


cobalt/tools/binary_size/SKILL.md
```diff
@@ -156,7 +185,47 @@ If the feature exposes any APIs to JavaScript (defined in `.idl` files), you mus
...
+   }
+   ```
+   > [!IMPORTANT]
+   > Make sure to list **all** wildcard variations matching any V8 binding files generated for the feature (e.g. checking for dictionaries, interfaces, unions, and enumerations) to prevent the `check_generated_file_list` validator from throwing unlisted source errors.
+4. Verify that the parent targets in `third_party/blink/renderer/bindings/BUILD.gn` and `third_party/blink/renderer/bindings/modules/v8/BUILD.gn` apply the exclusions generically:
+   ```gn
+   if (cobalt_bindings_exclude_patterns != []) {
+     _filtered_sources = filter_exclude(sources, cobalt_bindings_exclude_patterns)
+     sources = []
+     sources = _filtered_sources
```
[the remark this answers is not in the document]
haozheng-cobalt: GN strictly prohibits overwriting a populated list variable with another list directly.

cobalt/tools/binary_size/SKILL.md
```diff
@@ -52,19 +61,13 @@ buildflag_header("buildflags") {
...
-> If an include is required:
-> ```cpp
-> #include "third_party/blink/public/common/buildflags.h"
-> #if BUILDFLAG(ENABLE_<FEATURE_NAME>)
-> ...
-> #endif
-> ```
+> [!IMPORTANT]
+> **Strict Header Inclusion (Include What You Use - IWYU)**:
+> 1. **Always explicitly include the buildflags header**: If you modify any C++ source (`.cc`) or header (`.h`) file to add a feature gating preprocessor macro (`#if BUILDFLAG(ENABLE_<FEATURE>)`), you **must** explicitly include the generated buildflags header at the top of the file:
```
linxinan-yt, line 66: Does it have to be ENABLE_? If the feature is by default enabled in Chromium but has to be disabled in Cobalt, I feel like it should be BUILDFLAG(DISABLE_<FEATURE>). 
While if the feature is for cobalt and either null or false in Chromium, we should use ENABLE_. WDYT?

Plus, my two cents:
1. if you know the change is not necessary in next rebase(M150) attach CHROMIUM_MILESTONE_LE_138
2. The first PR of DISABLE_<FEATURE> or ENABLE_ should not change the default value. E.g. for Privacy sandbox API, the DISABLE_PRIVACY_SANDBOX is false in the surgical PR. And the build flag change should be in a follow up PR. This make the gardener easier to make a clean revert PR if anything goes wrong.

johnxwork (approved):

cobalt/tools/binary_size/SKILL.md
haozheng-cobalt: enable_ and disable_ both can work, I think it's more of a best practice question, if you look at `gn args out/some-build --list` almost all gn args are using use_ or enable_ in Cobalt, to simply the logic and be consistent in Cobalt's feature removal work, I suggest we stick to one convention: enable_.

haozheng-cobalt: Regarding your point about the MACRO `CHROMIUM_MILESTONE_LE_138`, I think it's a good idea but I have one concern, we cannot use this MACRO alone to gate sources code we also need to find the corresponding gn args to gate the deps or targets in the BUILD.gn or gni.

haozheng-cobalt: Agree with your second point, it will be easier to make a clean revert this way, thanks!

linxinan-yt: Yes CHROMIUM_MILESTONE_LE_138 is only for C++ code. Idl, py, java and gn are not covered yet. However, gn's logic is much easier, hence less hard to resolve the conflicts.  
And yes, I will see how to create a similar gn function to get the milestone number.

linxinan-yt (approved):

PR #10612, opened 2026-05-22
cobalt: Disable HTTP Cache

zhongqiliang:
Disable Http Cache by default.
Bug: 514683173


johnxwork:
Commented on the bug, I think turning the entire HTTP cache off can not be done without experiment. We don't have a corresponding JavaSwitch for it. Also we never disabled HTTP cache in c26. The issue lies deeper.

johnxwork (approved):

johnxwork (approved):

johnxwork:
Reminder to check for potentially needed unit test changes

PR #10614, opened 2026-05-23
nplb: Move WaitForFinish() to AbstractTestThread

rakuco:
Move the function added in #8732 to work around some hangs and freezes on tvOS out of `multiple_player_test.cc` and into the base class, and make Join() a private method so that callers must call the right function.

This fixes a hang in SbPlayerWriteSampleTest.SecondaryPlayerTest with the same cause described in #8732.

While here, remove `AbstractTestThread::GetThread()`. It is not used anywhere, and the handle it returns could be used by callers to invoke `pthread_join()` directly.

Bug: 465634083

### DIFF
--- a/starboard/nplb/posix_compliance/posix_thread_helpers.h
+++ b/starboard/nplb/posix_compliance/posix_thread_helpers.h
@@ -20,14 +20,9 @@
 #include <cstddef>
 #include <cstdint>
 
-#include "build/build_config.h"
 #include "starboard/configuration.h"
 #include "starboard/thread.h"
 #include "testing/gtest/include/gtest/gtest.h"
-
-#if BUILDFLAG(IS_IOS_TVOS)
-#include "starboard/tvos/shared/run_in_background_thread_and_wait.h"
-#endif  // BUILDFLAG(IS_IOS_TVOS)
 
 namespace nplb {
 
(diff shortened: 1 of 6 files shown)


rakuco:

starboard/nplb/posix_compliance/posix_thread_helpers.h
```diff
@@ -172,15 +177,27 @@ class AbstractTestThread {
...
+  // main thread (in AVSBVideoRenderer).
+  void WaitForFinish() {
+#if BUILDFLAG(IS_IOS_TVOS)
+    RunInBackgroundThreadAndWait([this] { Join(); });
+#else
+    Join();
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+  }
+
+  pthread_t GetThread() { return thread_; }
```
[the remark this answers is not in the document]
rakuco: fixed by removing `GetThread()` altogether

starboard/nplb/posix_compliance/posix_thread_helpers.h
```diff
@@ -172,15 +177,25 @@ class AbstractTestThread {
     return;
   }
 
+  // A wrapper for Join() that, on tvOS, invokes it on a background GCD queue.
+  // This must be done to avoid a deadlock when the main thread is blocked on
+  // Join() and the worker thread is blocked attempting to invoke code in the
+  // main thread (in AVSBVideoRenderer).
+  void WaitForFinish() {
```
kjyoun: Looks good to me!

My only suggestion is to move the implementation of  WaitForFinish()  (and  Join() ) to the  .cc  file.

I don't think we need to expose these implementation details in the header file. It is better to keep the implementation in the  .cc  file for better encapsulation.
 
The only benefit of keeping it in the header would be compiler inlining for performance, which is not necessary for test code

kjyoun (approved):

starboard/nplb/posix_compliance/posix_thread_helpers.h
rakuco: Done, thank you!

PR #10631, opened 2026-05-26
cobalt/metrics: Use target-specific buildflag COBALT_DETAILED_MEMORY_METRICS

Awallky:
Replaced the fragile global IS_COBALT macro inside services/resource_coordinator C++ codebase with a target-specific buildflag COBALT_DETAILED_MEMORY_METRICS.

    - Added a target-specific buildflag_header target in BUILD.gn, restricted to is_cobalt && (is_linux || is_android) to compile detailed Smaps telemetry on Linux/Android only.
    - Fixed a production bug by registering CobaltDetailedMetricsDelegate in `CobaltBrowserMainParts::PreMainMessageLoopRun` to ensure Smaps memory instrumentation is active in production.
    - Fixed a design bug in `queued_request.cc` by ensuring MEM_DUMP_DETAILED_STATS flag is always set for Cobalt builds, restoring periodic UMA detailed memory telemetry.
    - Added support for matching '/cobalt' monolithic executable mapping in C++ delegate and Python verification script to enable sizeable LibChrobaltRss telemetry on Linux monolithic platforms.

Bug: 494004530
Bug: 512576130


services/resource_coordinator/public/cpp/memory_instrumentation/BUILD.gn
```diff
@@ -45,6 +45,9 @@ component("memory_instrumentation") {
   }
 
   defines = [ "IS_RESOURCE_COORDINATOR_PUBLIC_MEMORY_INSTRUMENTATION_IMPL" ]
+  if (is_cobalt) {
```
jellefoks: I'm not a specialist on the build system, but it seems to me, if it works, that it would be better to have a '.gn' file with something like the snippet below, such as `third_party/angle/.gn` or `v8/.gn` do.

```
# This file is used by the GN meta build system to find the root of the source
# tree and to set startup options. For documentation on the values set in this
# file, run "gn help dotfile" at the command line.

import("//build/dotfile_settings.gni")

# The location of the build configuration file.
buildconfig = "//build/config/BUILDCONFIG.gn"
```

services/resource_coordinator/public/cpp/memory_instrumentation/BUILD.gn
```diff
@@ -45,6 +45,9 @@ component("memory_instrumentation") {
...
+    defines += [ "ENABLE_BUILDFLAG_IS_COBALT" ]
```
andrewsavage1: We should not do this. It is quite fragile, as IS_COBALT blocks outside of just the smap ones you care about could be enabled from this. Instead, if you want the changes to be enabled for chromium builds too, change the BUILDFLAG(IS_COBALT) to something else or remove it.

andrewsavage1 (dismissed):

services/resource_coordinator/public/cpp/memory_instrumentation/BUILD.gn
Awallky: Thanks for the suggestion! 
I think the changes in services/resource_coordinator/public/cpp/memory_instrumentation/BUILD.gn match that now.

services/resource_coordinator/public/cpp/memory_instrumentation/BUILD.gn
Awallky: Done. I refactored things to use a common detailed metrics dump flag.

johnxwork (approved):

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -120,6 +123,17 @@ int CobaltBrowserMainParts::PreCreateThreads() {
...
   StartMetricsRecording();
 
+#if BUILDFLAG(COBALT_DETAILED_MEMORY_METRICS)
+  static base::NoDestructor<CobaltDetailedMetricsDelegate> delegate;
+  if (memory_instrumentation::MemoryInstrumentation::GetInstance()) {
+    memory_instrumentation::MemoryInstrumentation::GetInstance()
+        ->SetDetailedMetricsDelegate(delegate.get());
+  } else {
+    LOG(WARNING) << "MemoryInstrumentation instance not available in "
+                    "PreMainMessageLoopRun!";
```
yell0wd0g: Do we need this LOG()? It's in the critical initialization path
and IIUC `MemoryInstrumentation::GetInstance()` is a singleton
that either is implemented per platform or it isn't, right?
So the LOG doesn't add much more than a build time indication.
If my understanding is right.

cobalt/testing/browser_tests/granular_memory_browsertest.cc
```diff
@@ -14,11 +14,13 @@
 #include "content/public/test/browser_test_utils.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/detailed_metrics_delegate.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/memory_instrumentation.h"
+#include "services/resource_coordinator/public/cpp/memory_instrumentation/memory_instrumentation_features.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/os_metrics.h"
 #include "services/resource_coordinator/public/mojom/memory_instrumentation/memory_instrumentation.mojom.h"
 #include "testing/gtest/include/gtest/gtest.h"
 #include "url/gurl.h"
 
+#if BUILDFLAG(COBALT_DETAILED_MEMORY_METRICS)
```
yell0wd0g: If the whole file is protected by this guard, maybe just not
build it if the GN associated flag is not defined?

cobalt/browser/metrics/cobalt_detailed_metrics_delegate.cc
```diff
@@ -62,6 +62,10 @@ void CobaltDetailedMetricsDelegate::OnSmapsEntry(
 
   if (name.empty()) {
     label = "anonymous_other";
+    // On Linux monolithic builds, the main executable mapping is named
+    // "/cobalt" rather than "libchrobalt.so" or "libcobalt.so".
+  } else if (absl::EndsWith(name, "/cobalt")) {
+    label = "lib_chrobalt";
```
yell0wd0g: Could we add this to the `kCobaltPatterns`? It's strange
to special case it. Instead of using a boolean to mean 
startswith/contains, how about just having the function 
to use directly, and bind it as a `base::FunctionRef<>`? 
That way there would be no chances for confusion and
the code here would be simpler.

yell0wd0g (approved):
The COBALT_DETAILED_MEMORY_METRICS part looks
straightforward enough. There are more bits to 
this CL that are unrelated though, right?

cobalt/testing/browser_tests/granular_memory_browsertest.cc
Awallky: Done.

cobalt/browser/metrics/cobalt_detailed_metrics_delegate.cc
Awallky: Done.

cobalt/browser/cobalt_browser_main_parts.cc
Awallky: It was more useful for sanity checks, but it seems to be working so I don't think it is necessary.
So, it is removed.
I'd rather a histogram if I really thought it was necessary to track anyways. :)

PR #10638, opened 2026-05-27
aosp: treat warnings as errors

rogerzanoni:
treat_warnings_as_errors was unset to make it easier to progress with other development tasks.

This set of changes make aost treat warnings as errors again and provide the fixes for the build errors that happen because of the warnings.

Bug: 495203133

### DIFF
--- a/cobalt/build/configs/aosp-arm/args.gn
+++ b/cobalt/build/configs/aosp-arm/args.gn
@@ -3,6 +3,3 @@ import("//cobalt/build/configs/evergreen-x64/args.gn")
 target_cpu = "arm"
 starboard_target_platform = "aosp-arm"
 arm_float_abi = "softfp"
-
-# TODO(b/380339614): Remove this flag after resolving build errors.
-treat_warnings_as_errors = false
--- a/copied_base/base/os_compat_android.cc
+++ b/copied_base/base/os_compat_android.cc
@@ -23,6 +23,9 @@
 #include "base/strings/string_piece.h"
 
 extern "C" {
+// TODO(b/374300500): when posix emulation is re-enabled, move
+// futimes to the emulation layer.
+#if __ANDROID_API__ < 26
 // There is no futimes() avaiable in Bionic, so we provide our own
 // implementation until it is there.
 int futimes(int fd, const struct timeval tv[2]) {
@@ -43,6 +46,7 @@ int futimes(int fd, const struct timeval tv[2]) {
   ts[1].tv_nsec = tv[1].tv_usec * 1000;
   return base::checked_cast<int>(syscall(__NR_utimensat, fd, NULL, ts, 0));
 }
+#endif // __ANDROID_API__ < 26
 
 #if !defined(__LP64__)
 // 32-bit Android has only timegm64() and not timegm().
--- a/starboard/android/shared/starboard_bridge.cc
+++ b/starboard/android/shared/starboard_bridge.cc
@@ -49,6 +49,8 @@ using jni_zero::JavaParamRef;
 using jni_zero::ScopedJavaGlobalRef;
 using jni_zero::ScopedJavaLocalRef;
 
+// TODO(b/492704919): enable on AOSP when the layering violation is fixed.
+#if !BUILDFLAG(IS_PARTNER_TOOLCHAIN)
 // Client Hint Header name constants
 constexpr char kAndroidOSExperienceHeader[] =
     "Sec-CH-UA-Co-Android-OS-Experience";
@@ -58,6 +60,7 @@ constexpr char kBuildFingerprintHeader[] =
     "Sec-CH-UA-Co-Android-Build-Fingerprint";
 constexpr char kYoutubeCertScopeHeader[] =
     "Sec-CH-UA-Co-Youtube-Certification-Scope";
+#endif  // !BUILDFLAG(IS_PARTNER_TOOLCHAIN)
 
 // Global pointer to hold the single instance of ApplicationAndroid.
 ApplicationAndroid* g_native_app_instance = nullptr;
--- a/starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
@@ -32,7 +32,17 @@ int __abi_wrap_readdir_r(musl_dir* dirp,
 
   struct dirent entry = {0};  // The type from platform toolchain.
   struct dirent* result = nullptr;
+// from POSIX.1-2008 readdir_r() was fully supported but emitting
+// a future deprecation warning. Marked as obsolecscent in POSIX.1-2024
+// TODO(b/374300500): add back posix emulation
+#if defined(_POSIX_VERSION) && (_POSIX_VERSION < 202406L) && defined(__clang__)
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
   int retval = readdir_r(dirp->dir, &entry, &result);
+#if defined(_POSIX_VERSION) && (_POSIX_VERSION < 202405L) && defined(__clang__)
+#pragma clang diagnostic pop
+#endif
   if (retval != 0) {
     return retval;
   }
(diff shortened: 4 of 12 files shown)


starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.cc
```diff
@@ -1,4 +1,4 @@
-// Copyright 2024 The Cobalt Authors. All Rights Reserved.
+// Copyright 2024 The Cobalt Authors. All Rights Rese>=ved.
```
andrewsavage1: revert

starboard/shared/modular/starboard_layer_posix_getrandom_abi_wrappers.cc
```diff
@@ -18,6 +18,10 @@
 #include <sys/random.h>
 
 ssize_t __abi_wrap_getrandom(void* buf, size_t buflen, unsigned flags) {
+#if defined(__ANDROID_API__) && __ANDROID_API__ < 28
```
andrewsavage1: This is functionality that we need to support. We should have an android-specific wrapper (similar to the open wrapper for android assets in C25). For now it can simply be implemented with a TODO to see if it needs to be fully implemented

starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.cc
```diff
@@ -492,7 +492,11 @@ int __abi_wrap_pthread_setname_np(musl_pthread_t thread, const char* name) {
 int __abi_wrap_pthread_getname_np(musl_pthread_t thread,
                                   char* name,
                                   size_t len) {
+#if __ANDROID_API__ >= 26
```
andrewsavage1: We would similarly want to implement an android-specific wrapper

starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
```diff
@@ -30,28 +30,33 @@ int __abi_wrap_readdir_r(musl_dir* dirp,
     return -1;
   }
 
-  struct dirent entry = {0};  // The type from platform toolchain.
-  struct dirent* result = nullptr;
-  int retval = readdir_r(dirp->dir, &entry, &result);
-  if (retval != 0) {
-    return retval;
+  errno = 0;
```
andrewsavage1: Can you explain why this change is necessary?

andrewsavage1 (commented):
The changes to starboard wrappers are concerning

starboard/android/shared/system_get_stack.cc
```diff
@@ -14,6 +14,7 @@
 
 #include <link.h>
 #include <unwind.h>
+#include <link.h>
```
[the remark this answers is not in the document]
rogerzanoni: Removed the commit iadding the link, it was fixed by another change.

rogerzanoni:
I will remove some of the wrapper changes; I checked the other platforms, they usually add -Wno-deprecated-declarations to cflags.

starboard/shared/modular/starboard_layer_posix_getrandom_abi_wrappers.cc
rogerzanoni: Thanks, I will add a TODO, and check if we need a follow-up commit with a wrapper.

starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
rogerzanoni: I was changing readdir_r wrapper to use readdir() while mimicking readdir_r behavior; readdir_r posix spec states that it handles errors using the return value and an output var, and doesn't mention errno. But readdir() does change errno, and I was using its value to set musl_result (the output var of the readdir_r signature). There's one case of readdir() that's problematic, "If  the  end of the directory stream is reached, NULL is returned and errno is not changed.", in this case I could be returning an errno set by other call, that's why I was zeroing it. But I will remove this readdir wrappers change, and add a cflag to suppress deprecated warnings as some other platforms do.

starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
```diff
@@ -30,28 +30,33 @@ int __abi_wrap_readdir_r(musl_dir* dirp,
...
-    *musl_result = nullptr;
-  } else {
-    *musl_result = musl_entry;
+  if (starboard::strlcpy(musl_entry->d_name, result_platform->d_name,
+                         sizeof(musl_entry->d_name)) >=
+      sizeof(musl_entry->d_name)) {
+    SB_LOG(WARNING) << "Truncated d_name in readdir_r wrapper."
+                    << " src_size=" << sizeof(result_platform->d_name)
+                    << " dst_size=" << sizeof(musl_entry->d_name);
   }
```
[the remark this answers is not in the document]
rogerzanoni: Removed the readdir wrappers change

starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.cc
rogerzanoni: Fixed

cobalt/build/configs/aosp-arm/args.gn
```diff
@@ -4,5 +4,4 @@ target_cpu = "arm"
 starboard_target_platform = "aosp-arm"
 arm_float_abi = "softfp"
 
-# TODO(b/380339614): Remove this flag after resolving build errors.
-treat_warnings_as_errors = false
+treat_warnings_as_errors = true
```
hferreiro: You can just remove it, it default to `true`.

starboard/shared/modular/starboard_layer_posix_getrandom_abi_wrappers.cc
rogerzanoni: Added a TODO to handle it on b/374300500 and instead of preventing the call, just pushed a diagnostic pragma to prevent the warning.

starboard/shared/modular/starboard_layer_posix_pthread_abi_wrappers.cc
rogerzanoni: Added a TODO to handle it on b/374300500 and instead of preventing the call, just pushed a diagnostic pragma to prevent the warning.

starboard/shared/modular/starboard_layer_posix_directory_abi_wrappers.cc
rogerzanoni: Removed this change and added a todo + pragma to prevent the warning. Instead of the above change, I will push a PR for re-enabling posix emulation for aosp

cobalt/build/configs/aosp-arm/args.gn
rogerzanoni: Fixed

copied_base/base/os_compat_android.cc
```diff
@@ -23,6 +23,7 @@
 #include "base/strings/string_piece.h"
 
 extern "C" {
+#if __ANDROID_API__ < 26
```
andrewsavage1, line 28: For older android api versions, would we want to create a emulation layer near the starboard wrappers as we probably would the other missing functions (like pthread_getname_np below)?

andrewsavage1 (approved):

copied_base/base/os_compat_android.cc
rogerzanoni: Yes, I think we should move it to an emulation layer for consistency. Upstream has it with the same guards, but since we are handling this on copied_base we can drop os_compat_android and handle it on posix_emu

andrewsavage1: In that case, let's add a TODO here as well

rogerzanoni: Done

starboard/android/shared/starboard_bridge.cc
```diff
@@ -49,6 +49,8 @@ using jni_zero::JavaParamRef;
 using jni_zero::ScopedJavaGlobalRef;
 using jni_zero::ScopedJavaLocalRef;
 
+// TODO(b/492704919): enable on AOSP when the layering violation is fixed.
+#if !BUILDFLAG(IS_PARTNER_TOOLCHAIN)
```
hferreiro, line 53: Perhaps it makes more sense to use `IS_STARBOARD` for code not involving the native toolchain?

rogerzanoni: IS_PARTNER_TOOLCHAIN was used to match the guarded uses of those variables in the JNI_StarboardBridge_* functions

rogerzanoni:
@andrewsavage1 could you please merge it? All checks are ok but the option to merge it isn't available to me

PR #10640, opened 2026-05-27
android: Fix Android ShellManagerTest compilation

arjungm:
Update ShellManagerTest to use the fully qualified class name for
ContentViewRenderView. This resolves a compilation error in Cobalt
Android JUnit tests caused by an ambiguous class reference.

Bug: 431780245

### DIFF
--- a/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ShellManagerTest.java
+++ b/cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ShellManagerTest.java
@@ -23,7 +23,7 @@ import android.widget.FrameLayout;
 
 import dev.cobalt.shell.Shell;
 import dev.cobalt.shell.ShellManager;
-import org.chromium.components.embedder_support.view.ContentViewRenderView;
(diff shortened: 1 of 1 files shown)


cobalt/android/apk/app/src/test/java/dev/cobalt/coat/ShellManagerTest.java
```diff
@@ -44,7 +44,7 @@ public class ShellManagerTest {
   private Activity mActivity;
   private ShellManager mShellManager;
   private FrameLayout mRootView;
-  @Mock private ContentViewRenderView mMockContentViewRenderView;
+  @Mock private dev.cobalt.shell.ContentViewRenderView mMockContentViewRenderView;
```
Awallky: Can you import this instead of using the full package here (and replace the org.chromium... version)?

arjungm: done

Awallky (approved):
LGTM
Thanks!

PR #10641, opened 2026-05-27
Report kokoro test results to resultstore

linxinan-yt:
Bug: 516781914


linxinan-yt:
We don't need to report result to resultstore now. Instead we run the test artifacts from kokoro on GHA. Close.

PR #10643, opened 2026-05-27
h5vcc: Re-enable passing web tests

andrewsavage1:
And fix the logic in the failing sideloader tests.

Bug: 515122610

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
@@ -116,7 +116,7 @@ ScriptPromise<IDLBoolean> H5vccUpdater::getAllowSelfSignedPackages(
   auto* resolver = MakeGarbageCollected<ScriptPromiseResolver<IDLBoolean>>(
       script_state, exception_state.GetContext());
 
-  EnsureReceiverIsBound();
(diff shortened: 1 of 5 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
```diff
@@ -116,7 +116,7 @@ ScriptPromise<IDLBoolean> H5vccUpdater::getAllowSelfSignedPackages(
   auto* resolver = MakeGarbageCollected<ScriptPromiseResolver<IDLBoolean>>(
       script_state, exception_state.GetContext());
 
-  EnsureReceiverIsBound();
+  EnsureSideloadingReceiverIsBound();
```
hlwarriner, line 119: Do you plan to fix this in `26.eap`, too?

hlwarriner (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_updater/h_5_vcc_updater.cc
andrewsavage1: Probably the only thing that sideloading will be used for is as a testing ground for bringing it to main—that's the intern project that @niranjanyardi and @yuying-y are leading, so I'll let them hand it off or bring it over

PR #10650, opened 2026-05-27
evergreen: Suspend and Resume UpdaterModule on Freeze and Unfreeze

yuying-y:
Coordinate the lifecycle transition of the updater during App Freeze and Unfreeze 

- coordinate Freeze (kSbEventTypeFreeze) and Unfreeze (kSbEventTypeUnfreeze) events inside AppEventDelegate::HandleEventLocked.
- implemented synchronous blocking in UpdaterModule::Suspend using a base::WaitableEvent. This forces the calling thread to wait until the updater thread completely finalizes its operations
- converted is_updater_running_ to std::atomic<bool> to avoid data races.

Issue: 479816505

### DIFF
--- a/cobalt/app/app_event_delegate.cc
+++ b/cobalt/app/app_event_delegate.cc
@@ -23,11 +23,16 @@
 #include "base/notreached.h"
 #include "base/run_loop.h"
 #include "base/threading/platform_thread.h"
+#include "build/build_config.h"
 #include "cobalt/app/app_event_runner.h"
 #include "cobalt/browser/lifecycle/cobalt_lifecycle_manager.h"
 #include "content/public/browser/browser_thread.h"
 #include "starboard/extension/crash_handler.h"
 #include "starboard/system.h"
+
+#if BUILDFLAG(USE_EVERGREEN)
+#include "cobalt/updater/updater_module.h"
+#endif
 
 namespace cobalt {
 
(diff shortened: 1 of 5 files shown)


cobalt/app/app_event_delegate.cc
```diff
@@ -23,6 +23,7 @@
 #include "base/notreached.h"
 #include "base/run_loop.h"
 #include "base/threading/platform_thread.h"
+#include "build/build_config.h"
```
jellefoks: this is the only line in this file. leftover from an earlier iteration?

cobalt/updater/updater_module.cc
```diff
@@ -204,14 +205,22 @@ UpdaterModule::~UpdaterModule() {
 }
 
 void UpdaterModule::Suspend() {
-  if (is_updater_running_) {
-    is_updater_running_ = false;
+  if (is_updater_running_.exchange(false)) {
```
jellefoks, line 207: This means that the updater suspension executes sequentially with the other suspending work. suspending and resuming would be faster if this would have separate methods to trigger the suspend and wait for completion calls.

cobalt/updater/updater_module.cc
```diff
@@ -204,14 +205,22 @@ UpdaterModule::~UpdaterModule() {
...
+    base::WaitableEvent event(base::WaitableEvent::ResetPolicy::MANUAL,
```
jellefoks, line 208: Hmm it's possible that this results in a DCHECK in devel builds because it may not like using WaitableEvent in UI threads.

cobalt/updater/updater_module.cc
```diff
@@ -206,12 +207,19 @@ UpdaterModule::~UpdaterModule() {
...
+          FROM_HERE, base::BindOnce(
+                         [](UpdaterModule* module, base::WaitableEvent* event) {
+                           module->Finalize();
+                           event->Signal();
+                         },
+                         base::Unretained(this), base::Unretained(&event)));
+      event.Wait();
     }
   }
 }
```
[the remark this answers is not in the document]
yuying-y: Done. Applied the code review suggestion: updated Suspend() to atomically exchange is_updater_running_ to false and safely check the return of PostTask() to prevent any potential deadlocks.

cobalt/app/app_event_delegate.cc
```diff
@@ -203,9 +208,31 @@ void AppEventDelegate::HandleEventLocked(const SbEvent* event) {
...
+    {
+      cobalt::updater::UpdaterModule* updater_module =
+          cobalt::updater::UpdaterModule::GetInstance();
+      if (updater_module) {
+        updater_module->Resume();
+      }
+    }
+#endif
       TransitionToLifeCycleState(SbEventToTargetApplicationState(event->type));
       break;
```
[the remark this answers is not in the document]
yuying-y: Done. Migrated the suspend/resume logic to DoFreeze() and DoUnfreeze() inside AppEventRunner. In DoUnfreeze(), the main engine unfreeze sequence content::Shell::OnUnfreeze() is executed first before resuming the updater, completely resolving this sequencing concern.

cobalt/app/app_event_delegate.cc
yuying-y: Yes, this was a leftover from the previous delegate-level compilation setup. It has been successfully removed in the latest commit.

cobalt/updater/updater_module.cc
yuying-y: You are completely correct that triggering the finalization asynchronously and waiting for it later (e.g. at the very end of the suspension transitions) would allow overlapping the updater shutdown with other suspending tasks like Flushing Storage/Cookies or surface destruction.

However, because the updater's Finalize() runs on a dedicated background thread and performs very fast cleanup operations (unregistering observers, committing cached configurator preference variables, and slot file cleanup) that usually take less than a few milliseconds, the overhead of sequential execution here is negligible. Keeping the trigger and synchronization unified in a single blocking Suspend() method simplifies the lifecycle API and prevents any possibility of incomplete transitions or state tracking complexity. We can certainly explore split-phase execution if profiling metrics on low-end devices show suspend latency bottlenecks.

jellefoks (approved):

PR #10651, opened 2026-05-28
media: Resolve video pipeline backpressure leak on Android

kjyoun:
This PR fixes a bug where `MediaCodecVideoDecoder` signaled NEED_MORE_INPUT unconditionally. By introducing the unified `SignalIfNeedMoreInput()` method, this PR ensures all NEED_MORE_INPUT calls are only dispatched after a proper buffer-fullness check.

Issue: 515102461
Issue: 517914191

### DIFF
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -966,8 +966,8 @@ void MediaCodecVideoDecoder::WriteInputBuffersInternal(
   }
 
   media_decoder_->WriteInputBuffers(input_buffers);
-  bool need_more_input_signaled = SignalIfNeedMoreInput();
-  if (!need_more_input_signaled && tunnel_mode_audio_session_id_) {
+  bool need_more_input = SignalIfNeedMoreInput();
+  if (!need_more_input && tunnel_mode_audio_session_id_) {
     // In tunnel mode playback when need data is not signaled above, it is
     // possible that the VideoDecoder won't get a chance to send kNeedMoreInput
     // to the renderer again.  Schedule a task to check back.
@@ -1018,11 +1018,11 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
       dequeue_output_result, media_codec_bridge,
       std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this));
 
-  bool need_more_input_signaled = false;
+  bool need_more_input = false;
   if (!is_end_of_stream) {
-    need_more_input_signaled = SignalIfNeedMoreInput(video_frame);
+    need_more_input = SignalIfNeedMoreInput(video_frame);
   }
-  if (!need_more_input_signaled) {
+  if (!need_more_input) {
     decoder_status_cb_(kBufferFull, std::move(video_frame));
   }
 }
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -993,7 +993,8 @@ void MediaCodecVideoDecoder::WriteInputBuffersInternal(
 
 void MediaCodecVideoDecoder::ProcessOutputBuffer(
     MediaCodec* media_codec_bridge,
-    const DequeueOutputResult& dequeue_output_result) {
+    const DequeueOutputResult& dequeue_output_result,
+    int number_of_pending_inputs) {
   SB_DCHECK(decoder_status_cb_);
   SB_DCHECK_GE(dequeue_output_result.index, 0);
 
@@ -1018,14 +1019,13 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
   }
 
   if (fix_need_more_input_backpressure_) {
-    auto video_frame = make_scoped_refptr<VideoFrameImpl>(
-        dequeue_output_result, media_codec_bridge,
-        std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this));
-
-    Schedule([this, is_end_of_stream, video_frame = std::move(video_frame)]() {
-      ProcessOutputBufferWithBackpressure(is_end_of_stream,
-                                          std::move(video_frame));
-    });
+    bool need_more_input =
+        !is_end_of_stream && number_of_pending_inputs < kMaxPendingInputsSize;
+    decoder_status_cb_(
+        need_more_input ? kNeedMoreInput : kBufferFull,
+        make_scoped_refptr<VideoFrameImpl>(
+            dequeue_output_result, media_codec_bridge,
+            std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this)));
     return;
   }
 
@@ -1034,20 +1034,6 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
       new VideoFrameImpl(
           dequeue_output_result, media_codec_bridge,
           std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this)));
-}
-
-void MediaCodecVideoDecoder::ProcessOutputBufferWithBackpressure(
-    bool is_end_of_stream,
-    scoped_refptr<VideoFrame> video_frame) {
-  SB_CHECK(BelongsToCurrentThread());
-  if (!decoder_status_cb_) {
-    return;
-  }
-  bool need_more_input =
-      !is_end_of_stream && media_decoder_ &&
-      media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
-  decoder_status_cb_(need_more_input ? kNeedMoreInput : kBufferFull,
-                     std::move(video_frame));
 }
 
 void MediaCodecVideoDecoder::OnEndOfStreamWritten(
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -381,6 +381,9 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
           kMediaEnableLowLatency)),
       enable_ndk_video_(
           pipeline_config.experimental_features.GetBool(kMediaNdkVideo)),
+      fix_need_more_input_backpressure_(
+          pipeline_config.experimental_features.GetBool(
+              kMediaFixNeedMoreInputBackpressure)),
       is_video_frame_tracker_enabled_(android_get_device_api_level() >= 34 ||
                                       tunnel_mode_audio_session_id_),
       media_codec_factory_(std::move(media_codec_factory)),
@@ -1014,9 +1017,12 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
     }
   }
 
-  bool need_more_input =
-      !is_end_of_stream && media_decoder_ &&
-      media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
+  bool need_more_input = !is_end_of_stream;
+  if (need_more_input && fix_need_more_input_backpressure_) {
+    need_more_input =
+        media_decoder_ &&
+        media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
+  }
   decoder_status_cb_(
       need_more_input ? kNeedMoreInput : kBufferFull,
       make_scoped_refptr<VideoFrameImpl>(
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -133,17 +133,15 @@ const int kNonInitialPrerollFrameCount = 1;
 // tunnel mode prerolling only needs 1 frame.
 const int kTunnelModePrerollFrameCount = 1;
 // The maximum number of pending inputs allowed in the decoder queue.
-// We set this to 512 frames (approx 8.5 seconds of 60fps video or 17 seconds
-// of 30fps video) to provide a robust buffer safety cushion that survives
-// V8 JavaScript main-thread congestion (which typically lasts 2-5 seconds
-// during app startup or heavy page transitions) without video starvation,
-// while keeping C++ heap memory usage extremely low (~25MB for 4K streams).
-constexpr int kMaxPendingInputsSize = 512;
+// We set this to 128 frames (approx 2.1 seconds of 60fps video or 4.2 seconds
+// of 30fps video) to provide a buffer safety cushion that helps survive
+// V8 JavaScript main-thread congestion without video starvation,
+constexpr int kMaxPendingInputsSize = 128;
 
 // VideoFrameTracker tracks frames in the entire media pipeline (decoder queue,
 // codec, and renderer). We set its capacity to accommodate the maximum input
-// queue size (512) plus a margin of 100 frames for frames in the codec and
-// renderer.
+// queue size (`kMaxPendingInputsSize`) plus a margin of 100 frames for frames
+// in the codec and renderer.
 constexpr int kVideoFrameTrackerCapacity = kMaxPendingInputsSize + 100;
 
 const int kFpsGuesstimateRequiredInputBufferCount = 3;
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -1017,17 +1017,37 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
     }
   }
 
-  bool need_more_input = !is_end_of_stream;
-  if (need_more_input && fix_need_more_input_backpressure_) {
-    need_more_input =
-        media_decoder_ &&
-        media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
+  if (fix_need_more_input_backpressure_) {
+    auto video_frame = make_scoped_refptr<VideoFrameImpl>(
+        dequeue_output_result, media_codec_bridge,
+        std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this));
+
+    Schedule([this, is_end_of_stream, video_frame = std::move(video_frame)]() {
+      ProcessOutputBufferWithBackpressure(is_end_of_stream,
+                                          std::move(video_frame));
+    });
+    return;
   }
+
   decoder_status_cb_(
-      need_more_input ? kNeedMoreInput : kBufferFull,
-      make_scoped_refptr<VideoFrameImpl>(
+      is_end_of_stream ? kBufferFull : kNeedMoreInput,
+      new VideoFrameImpl(
           dequeue_output_result, media_codec_bridge,
           std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this)));
+}
+
+void MediaCodecVideoDecoder::ProcessOutputBufferWithBackpressure(
+    bool is_end_of_stream,
+    scoped_refptr<VideoFrame> video_frame) {
+  SB_CHECK(BelongsToCurrentThread());
+  if (!decoder_status_cb_) {
+    return;
+  }
+  bool need_more_input =
+      !is_end_of_stream && media_decoder_ &&
+      media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
+  decoder_status_cb_(need_more_input ? kNeedMoreInput : kBufferFull,
+                     std::move(video_frame));
 }
 
 void MediaCodecVideoDecoder::OnEndOfStreamWritten(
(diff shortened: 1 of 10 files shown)


starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -945,11 +956,17 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
...
     }
   }
-  decoder_status_cb_(
-      is_end_of_stream ? kBufferFull : kNeedMoreInput,
-      new VideoFrameImpl(
-          dequeue_output_result, media_codec_bridge,
-          std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this)));
+  auto video_frame = make_scoped_refptr<VideoFrameImpl>(
+      dequeue_output_result, media_codec_bridge,
+      std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this));
```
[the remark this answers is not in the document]
kjyoun: This commit is incorrect. No problem in compile

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -943,11 +954,17 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
...
+
+  bool need_more_input = false;
+  if (!is_end_of_stream) {
+    need_more_input = SignalIfNeedMoreInput(video_frame);
+  }
+  if (!need_more_input) {
+    decoder_status_cb_(kBufferFull, std::move(video_frame));
+  }
```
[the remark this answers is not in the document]
kjyoun: gist of this PS is to use single-point for NEED_MORE_INPUT signaling.

Hence this proposal breaks the very goal of this PR.

Instead, I changed the method to accept const reference, so that we can avoid ref count arithmetics

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -1168,4 +1184,21 @@ void MediaCodecVideoDecoder::ResetInternal(bool skip_flush) {
...
+    return false;
+  }
+
+  bool need_more_input =
+      media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
+  if (need_more_input) {
+    decoder_status_cb_(kNeedMoreInput, std::move(frame));
+  }
+  return need_more_input;
+}
```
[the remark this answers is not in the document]
kjyoun: ditto

starboard/android/shared/media_codec_video_decoder.h
```diff
@@ -263,6 +263,8 @@ class MediaCodecVideoDecoder : public VideoDecoder,
   const size_t initial_number_of_preroll_frames_;
   size_t number_of_preroll_frames_;
 
+  bool SignalIfNeedMoreInput(scoped_refptr<VideoFrame> frame = nullptr);
```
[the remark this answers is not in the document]
kjyoun: ditto

xiaomings:
My apologies, I didn't realize there was already an attempt to resolve a similar issue!  I opened a PR here: https://github.com/youtube/cobalt/pull/10654, to resolve from a different angle.

https://github.com/youtube/cobalt/pull/10654 is trying to resolve the issue that the `decoding_buffers_` in `SbPlayerBridge` can grow to very large (>1000), which causes extra time on adding and removing.

I saw this PR increases the pending element count to 512 (8.5 seconds of 60fps).  Unless we really need this amount of buffering to avoid any hiccups, I would re-consider this, as it has runtime cost involved.

kjyoun:
> My apologies, I didn't realize there was already an attempt to resolve a similar issue! I opened a PR here: #10654, to resolve from a different angle.
> 
> #10654 is trying to resolve the issue that the `decoding_buffers_` in `SbPlayerBridge` can grow to very large (>1000), which causes extra time on adding and removing.
> 
> I saw this PR increases the pending element count to 512 (8.5 seconds of 60fps). Unless we really need this amount of buffering to avoid any hiccups, I would re-consider this, as it has runtime cost involved.


> My apologies, I didn't realize there was already an attempt to resolve a similar issue! I opened a PR here: #10654, to resolve from a different angle.
> 
> #10654 is trying to resolve the issue that the `decoding_buffers_` in `SbPlayerBridge` can grow to very large (>1000), which causes extra time on adding and removing.
> 
> I saw this PR increases the pending element count to 512 (8.5 seconds of 60fps). Unless we really need this amount of buffering to avoid any hiccups, I would re-consider this, as it has runtime cost involved.

Left comment in https://github.com/youtube/cobalt/pull/10654/changes#r3319154270

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -943,11 +954,17 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
...
     }
   }
-  decoder_status_cb_(
-      is_end_of_stream ? kBufferFull : kNeedMoreInput,
-      new VideoFrameImpl(
-          dequeue_output_result, media_codec_bridge,
-          std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this)));
+  auto video_frame = make_scoped_refptr<VideoFrameImpl>(
+      dequeue_output_result, media_codec_bridge,
+      std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this));
```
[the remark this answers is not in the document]
kjyoun: this is not correct.

starboard/android/shared/video_frame_tracker.cc
```diff
@@ -87,6 +87,10 @@ void VideoFrameTracker::OnInputBuffer(int64_t timestamp) {
       static_cast<size_t>(max_pending_frames_size_)) {
     // OnFrameRendered() is only available after API level 23.  Cap the size
     // of |frames_to_be_rendered_| in case OnFrameRendered() is not available.
+    SB_LOG(WARNING) << "VideoFrameTracker capacity exceeded ("
+                    << max_pending_frames_size_
+                    << " frames). Evicting oldest frame with timestamp: "
+                    << frames_to_be_rendered_.front();
```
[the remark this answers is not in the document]
kjyoun: this log would not happen in normal condition

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -106,7 +106,19 @@ const int kNonInitialPrerollFrameCount = 1;
...
 // tunnel mode prerolling only needs 1 frame.
 const int kTunnelModePrerollFrameCount = 1;
-const int kMaxPendingInputsSize = 128;
+// The maximum number of pending inputs allowed in the decoder queue.
+// We set this to 512 frames (approx 8.5 seconds of 60fps video or 17 seconds
+// of 30fps video) to provide a robust buffer safety cushion that survives
+// V8 JavaScript main-thread congestion (which typically lasts 2-5 seconds
+// during app startup or heavy page transitions) without video starvation,
+// while keeping C++ heap memory usage extremely low (~25MB for 4K streams).
+constexpr int kMaxPendingInputsSize = 512;
```
xiaomings: We should re-consider setting this to 512.  Even if the system is at load, we don't need to cache 8.5 seconds of buffers here.  This will inflate decoding_buffers_ in SbPlayerBridge and have performance side effects.

kjyoun: Ack.

will figure out the best default value and ask review

kjyoun: I reverted back to 128, the original value

My concern was that with this bug, kMaxPendingInputsSize was not respected so that  starboard pipeline could have been holding far more than the limit (then hard-enforcing 128 could make regressions)

But, with local test, I verified that pending input level is just a little higher than 128. (until very recently [this change](https://source.corp.google.com/h/github/youtube/cobalt/+/f783ca8af2a038054d929ad61e76a090a85a548b)

https://b.corp.google.com/issues/517914191#comment3

Then it would be no harm to hard-enforce 128 from now on.

kjyoun:
reverted max pending input to 128 and added unit tests

PTAL

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -949,11 +958,17 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
       }
     }
   }
-  decoder_status_cb_(
-      is_end_of_stream ? kBufferFull : kNeedMoreInput,
-      new VideoFrameImpl(
-          dequeue_output_result, media_codec_bridge,
-          std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this)));
+  auto video_frame = make_scoped_refptr<VideoFrameImpl>(
```
xiaomings: super nit: if we had to add empty line around this statement, it should be added before this statement instead of added after.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -949,11 +958,17 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
...
+      dequeue_output_result, media_codec_bridge,
+      std::bind(&MediaCodecVideoDecoder::OnVideoFrameRelease, this));
+
+  bool need_more_input = false;
+  if (!is_end_of_stream) {
+    need_more_input = SignalIfNeedMoreInput(video_frame);
+  }
+  if (!need_more_input) {
+    decoder_status_cb_(kBufferFull, std::move(video_frame));
+  }
```
xiaomings: Consider just using `media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize` here, instead of consolidating the common code into SignalIfNeedMoreInput().  We should also remove `SignalIfNeedMoreInput()`, for the following reasons:
1. It's slightly confusing that `decoder_status_cb_` is called by `SignalIfNeedMoreInput()` when we need more buffer, and is called locally when we don't need more buffer.
2. The original logic makes it clear that when `is_end_of_stream` is true, we don't need more buffers, e.g. `SignalIfNeedMoreInput()` should return false here.  However, the current version of `SignalIfNeedMoreInput()` doesn't capture this information, and we are relying on the fact that `SignalIfNeedMoreInput()` should never return true when `is_end_of_stream` is also true.  While this is true, it's quite implicit.

Note that it's often ok to call a callback in such a function, when under a more obvious situation.  If we really want to consolidate, we should create a function with elegant interface to manage the call to `decoder_status_cb_` under all situations.

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Adopted `media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize`

I tried `a function with elegant interface` but it made this PR complicated, so that I gave up

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -994,9 +1004,13 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
       }
     }
   }
+
+  bool need_more_input =
```
borongc: Should this be gated by an experimental flag? I feel like there is a risk for this change.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -994,9 +1004,13 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
...
+      !is_end_of_stream &&
+      media_decoder_->GetNumberOfPendingInputs() < kMaxPendingInputsSize;
```
xiaomings: `media_decoder_.reset()` call in this file will set the internal pointer of `media_decoder_` to nullptr before destroying the object.  There could be a race condition that the callback is called between them, and hit the nullptr.

starboard/android/shared/video_frame_tracker.cc
```diff
@@ -87,6 +87,10 @@ void VideoFrameTracker::OnInputBuffer(int64_t timestamp) {
       static_cast<size_t>(max_pending_frames_size_)) {
     // OnFrameRendered() is only available after API level 23.  Cap the size
     // of |frames_to_be_rendered_| in case OnFrameRendered() is not available.
+    SB_LOG(WARNING) << "VideoFrameTracker capacity exceeded ("
```
xiaomings, line 90: How frequent would this happen?  Will this cause a log spam?

starboard/android/shared/media_codec_video_decoder_test.cc
```diff
@@ -188,5 +191,71 @@ TEST_F(MediaCodecVideoDecoderTest, BasicDecodingFlow) {
...
+        if (frame) {
+          received_status = status;
+          std::lock_guard local_lock(local_mutex);
+          output_received = true;
+          local_cv.notify_all();
+        }
+      },
+      [](SbPlayerError error, const std::string& msg) {});
+
+  const int kMaxPendingInputs = 128;
```
xiaomings: nit: constexpr

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -131,7 +131,17 @@ const int kNonInitialPrerollFrameCount = 1;
 // rendered, the rest of the playback should play without frame drops. So,
 // tunnel mode prerolling only needs 1 frame.
 const int kTunnelModePrerollFrameCount = 1;
-const int kMaxPendingInputsSize = 128;
+// The maximum number of pending inputs allowed in the decoder queue.
+// We set this to 128 frames (approx 2.1 seconds of 60fps video or 4.2 seconds
+// of 30fps video) to provide a buffer safety cushion that helps survive
+// V8 JavaScript main-thread congestion without video starvation,
```
xiaomings: nit: Consider using trailing '.' instead of ','.

starboard/android/shared/media_codec_video_decoder.cc
xiaomings: +1

starboard/android/shared/video_frame_tracker.cc
kjyoun: Yes, it will add another log, on top of log existing log spamming

https://b.corp.google.com/issues/515102461#comment1

This log would be more helpful than existing log spamming, since this log tells why/when the problem happens.

In this case, I think that we had better spam the log.
I found this, only because the log is spammed. Other it would be silently ignored, though this has significant impact of corrupting "Dropped frame" count in the rasta.

starboard/android/shared/media_codec_video_decoder_test.cc
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Thanks for the catch.

Updated the code

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: @xiaomings 

How about reviving https://github.com/youtube/cobalt/pull/10654, instead of this?

Basically, this PR is becoming https://github.com/youtube/cobalt/pull/10654. then it would be more convenient to revive it, since it has this feature under the experiment.

kjyoun: after the chat with Xiaoming, decided to add experiment flag on this PR.

Add the flag. Take a look.

starboard/android/shared/video_frame_tracker.cc
borongc, line 90: This is only on ATV 14. Do you have any local test showing adding this backpressure will lead to this?

kjyoun: The underlying issue (signaling `kNeedMoreInput` without checking the number of pending frames) exists across all Android OS versions, but it is only "visible" on Android 14+ because `VideoFrameTracker` is enabled exclusively for API level 34+ (`android_get_device_api_level() >= 34`).

* **On Android TV 14+:** `VideoFrameTracker` actively tracks end-to-end frame delivery and logs warnings when pending frames overflow its capacity.
* **On earlier Android versions:** `VideoFrameTracker` is inactive, so excessive pending frames accumulate silently in the decoder without warnings.

Local testing (toggling the experimental flag on/off on an ATV 14 device) confirmed that enabling this backpressure fix successfully keeps the pending input frame count <= 128 (`kMaxPendingInputsSize`) and resolves the `VideoFrameTracker` capacity overflow warnings.

---
The missing backpressure check in `MediaCodecVideoDecoder` has been present for a long time (on C25/26), but it did not cause noticeable issues until we increased `Media.MaxSamplesPerWrite` this May (see b/497891900#comment13). 

Though this is a bug fix, I agree that gating the fix behind the experimental feature flag is the safest route to verify stability and prevent regressions,  because this is a long-standing behavior,

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -1003,9 +1016,16 @@ void MediaCodecVideoDecoder::ProcessOutputBuffer(
       }
     }
   }
+
+  bool need_more_input = !is_end_of_stream;
+  if (need_more_input && fix_need_more_input_backpressure_) {
+    need_more_input =
+        media_decoder_ &&
```
xiaomings: Technically there is still a race condition here (sorry that the same issue occurs in my PR).  Feel free to go ahead with the experiment, but we should revisit and resolve this eventually.

xiaomings (approved):

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Great catch..

I dont want to add another bug while fixing old bug.

I addressed it. PTAL

xiaomings (approved):

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Met crash with new change. Investigating

--

  ### Root Cause of the Crash

  The tombstone stack trace reveals an object lifetime violation (Use-After-Free / Null Dereference) caused by storing video_frame inside
  JobQueue:

    02cfff97  starboard::(anonymous namespace)::VideoFrameImpl::~VideoFrameImpl()
    02cec335  starboard::JobQueue::Job::Reset()
    02d0acc9  starboard::JobQueue::RemoveJobsByOwner()
    02cfef19  starboard::MediaCodecVideoDecoder::ResetInternal()
    ──────
  ### Why It Crashed

  1. In media_codec_video_decoder.cc, media_codec_video_decoder.cc is called first. This destroys media_decoder_ and its underlying
  MediaCodecBridge.
  2. Next, ResetInternal() calls CancelPendingJobs(), which purges the JobQueue.
  3. Purging the JobQueue destroys the lambda capturing video_frame, triggering media_codec_video_decoder.cc:
 ```
    ~VideoFrameImpl() {
      if (!released_) {
        media_codec_bridge_->ReleaseOutputBuffer(...); // <-- CRASH: media_codec_bridge_ is already dead!
      }
    }
```
  4. Because media_codec_video_decoder.cc holds a raw MediaCodec* media_codec_bridge_ pointer, holding VideoFrameImpl inside JobQueue
  causes it to outlive the hardware codec during seeks/resets.
  ──────
  ### The Solution: Keep ProcessOutputBuffer Synchronous

  JobQueue is fundamentally unsafe for holding VideoFrameImpl because the queue does not participate in codec flush notifications
  (OnFlushing()).

  The proper fix is to keep ProcessOutputBuffer synchronous and pass num_pending_inputs directly through the Host::ProcessOutputBuffer
  callback interface (the 5-file fix we explored earlier):

  1. Synchronous Frame Delivery: No Schedule(), so VideoFrameImpl is never stored in JobQueue.
  2. Zero Lifetime Bugs: Frames are managed exclusively by VideoRendererImpl, which flushes frames before codec teardown.
  3. Thread-Safe: No cross-thread reads of media_decoder_.
  ──────
  ### Work Summary

  Diagnosed the tombstone crash log from the Android device. Explained that capturing VideoFrameImpl in JobQueue causes ~VideoFrameImpl()
  to access a dangling media_codec_bridge_ during ResetInternal(), and recommended reverting Schedule() in favor of the synchronous
  callback parameter fix.

kjyoun: I changed code to pass `number_of_pending_inputs` through argument.

Although it requires some parameter plumbing, this approach is much safer than modifying the destruction order of the media pipeline. When I called `Schedule` method, `ProcessOutputBuffer` became async and susceptible on destruction order correctness.

https://github.com/youtube/cobalt/pull/10651/changes/14483985351749ab8eef2d42cb5e283059e3bf51

@xiaomings 
WDYT?

xiaomings (approved):

PR #10669, opened 2026-05-28
android: Fix crash in audio sink frame tester

borongc:
The AudioSinkMinRequiredFramesTester experienced a crash due to
unsynchronized access to the audio sink between the tester thread and
the frame consumption logic. This led to potential null pointer
dereferences when the sink was released.

Ensure the audio sink lifecycle is thread-safe to prevent race
conditions when querying underrun counts or start thresholds during
resource teardown.

This is introduced by https://github.com/youtube/cobalt/pull/10291.

Issue: 517558884

### DIFF
--- a/starboard/android/shared/audio_sink_min_required_frames_tester.cc
+++ b/starboard/android/shared/audio_sink_min_required_frames_tester.cc
@@ -145,14 +145,19 @@ void AudioSinkMinRequiredFramesTester::TesterThreadFunc() {
       wait_timeout = !notified;
     }
 
-    // Get start threshold before release the audio sink.
-    int start_threshold =
-        audio_sink_ ? audio_sink_->GetStartThresholdInFrames() : 0;
-
+    int start_threshold = 0;
+    std::unique_ptr<AudioTrackAudioSink> sink_to_destroy;
+    {
+      std::lock_guard lock(mutex_);
+      if (audio_sink_) {
+        start_threshold = audio_sink_->GetStartThresholdInFrames();
+      }
+      sink_to_destroy = std::move(audio_sink_);
+    }
     // |min_required_frames_| is shared between two threads. Release audio sink
     // to end audio sink thread before access |min_required_frames_| on this
     // thread.
-    audio_sink_.reset();
+    sink_to_destroy.reset();
 
     if (wait_timeout) {
       SB_LOG(ERROR) << "Audio sink min required frames tester timeout.";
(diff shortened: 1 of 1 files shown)


starboard/android/shared/audio_sink_min_required_frames_tester.cc
```diff
@@ -243,16 +248,23 @@ void AudioSinkMinRequiredFramesTester::ConsumeFrames(int frames_consumed) {
...
+      return;
+    }
+    if (last_underrun_count_ == -1) {
+      // |last_underrun_count_| is unknown, record the current underrun count
+      // and start to observe the underrun count.
+      last_underrun_count_ = audio_sink_->GetUnderrunCount();
+      last_total_consumed_frames_ = total_consumed_frames_;
+      return;
+    }
+    underrun_count = audio_sink_->GetUnderrunCount();
```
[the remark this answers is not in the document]
borongc: Done.

jasonzhangxx (approved):

starboard/android/shared/audio_sink_min_required_frames_tester.cc
```diff
@@ -145,14 +145,19 @@ void AudioSinkMinRequiredFramesTester::TesterThreadFunc() {
...
+      if (audio_sink_) {
+        start_threshold = audio_sink_->GetStartThresholdInFrames();
+      }
+      sink_to_destroy = std::move(audio_sink_);
+    }
     // |min_required_frames_| is shared between two threads. Release audio sink
     // to end audio sink thread before access |min_required_frames_| on this
     // thread.
-    audio_sink_.reset();
+    sink_to_destroy.reset();
```
kjyoun, line 160: How about adding Stop() method to AudioSinkThread so that call site can call it before destroying it.
(Then we dont need to add mutex)

WDYT about this ?

https://github.com/youtube/cobalt/pull/10676/changes

borongc: per gemini:

---

While it improves lifecycle management, `it fails to fix the race condition` that causes the crash in the audio sink tester because it lacks thread synchronization:

1. `No Mutex Protection`: It does not add locks around the shared  `audio_sink_`  pointer.
2. `Concurrent Access`: The tester thread can still modify/reset the  `audio_sink_`  pointer while the audio callback thread is actively reading it to check underrun counts.
3. `Potential Crash`: This data race can still lead to null pointer dereferences or use-after-free crashes if the callback executes at the exact moment the sink is being released.
  
In short, PR 10676 addresses thread joining but misses the necessary `locking` to make access to the shared object pointer thread-safe.

---

I think https://github.com/youtube/cobalt/pull/10676 can be a clean up PR, and you can work on it after this PR.

PR #10675, opened 2026-05-28
android: Always resample audio samples to float

jasonzhangxx:
The AudioTimeStretcher component requires audio data to be in a
float32 interleaved format. This update ensures that the audio
renderer always resamples decoded audio to this format, which
prevents compatibility issues and eliminates the need for
additional conversions later in the pipeline.

Bug: 515433551

### DIFF
--- a/starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
+++ b/starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
@@ -578,13 +578,17 @@ void AudioRendererPcm::OnFirstOutput(
   buffered_frames_to_start_ = std::min(
       destination_sample_rate / 5, max_cached_frames_ - min_frames_per_append_);
 
+  // |time_stretcher_| only supports kSbMediaAudioSampleTypeFloat32 and
+  // kSbMediaAudioFrameStorageTypeInterleaved, so we resample the audio to that
+  // format to avoid unnecessary extra conversion.
   if (decoded_sample_rate != destination_sample_rate ||
-      decoded_sample_type != sink_sample_type_ ||
+      decoded_sample_type != kSbMediaAudioSampleTypeFloat32 ||
       decoded_storage_type != kSbMediaAudioFrameStorageTypeInterleaved) {
     resampler_ = AudioResampler::Create(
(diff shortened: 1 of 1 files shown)


starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
borongc, line 703: Do we still need this check?

jasonzhangxx: Technically, no, but to be cautious I prefer to keep it for now as this change is not guarded behind an experiment.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -581,13 +578,17 @@ void AudioRendererPcm::OnFirstOutput(
   buffered_frames_to_start_ = std::min(
       destination_sample_rate / 5, max_cached_frames_ - min_frames_per_append_);
 
+  // |time_stretcher_| only supports kSbMediaAudioSampleTypeFloat32 and
+  // kSbMediaAudioFrameStorageTypeInterleaved, so we resample the audio to that
+  // format to avoid unnecessary extra conversion.
   if (decoded_sample_rate != destination_sample_rate ||
-      decoded_sample_type != sink_sample_type_ ||
+      decoded_sample_type != kSbMediaAudioSampleTypeFloat32 ||
```
xiaomings, line 585: Shall we run an experiment on this?

jasonzhangxx: I don't think that's necessary. Except for tunnel mode, on Android, |sink_sample_type_| is always kSbMediaAudioSampleTypeFloat32.

borongc: Would this impact non-Android platforms?

jasonzhangxx: Behavior of linux reference implementation may be changed a little bit, but the samples will be converted to float before sending to time stretcher anyway.

xiaomings (approved):

PR #10677, opened 2026-05-28
Prune WebXR to reduce android binary size

ckim1346:
This PR decouples Cobalt's dependency on WebXR, which is important in pruning Dawn/WebGPU. This is Step 2 of 3 (See step 1 #10644) to safely remove the heavy Dawn/WebGPU dependencies and achieve a significant binary size reduction on Android (~800 KB total). Pruning WebXR on top of WebNN shows an additional binary size decrease of ~278.5 KB (93,093,829 to 92,815,301 bytes), bringing total cumulative savings to ~557 KB.  It also addresses a crash that happened https://github.com/youtube/cobalt/pull/10649#issuecomment-4566124338 that occurred due to a uncaught `nullptr` in `cobalt_module_stubs.cc`'s `static const WrapperTypeInfo g_dummy_wrapper_type_info` which has been replaced with a `DummyInstallInterfaceTemplateFunc`.                            
                                                                                                               
  #### Key Changes:                                                                                            
                                                                                                               
  1. Build Graph Pruning (GN Subtraction):                                                                     
      • Subtracted  //third_party/blink/renderer/modules/xr  from  sub_modules  in  modules/BUILD.gn .         
      • Gated the  xr  dependency behind  !is_cobalt  in  third_party/blink/renderer/modules/video_rvfc/BUILD.          
      • Subtracted the  xr  dependency in  third_party/blink/renderer/modules/webgl/BUILD.gn  for Cobalt builds
      to ensure WebGL remains fully functional without WebXR.                                                                           
  2. Bindings Filtering:                                                                                       
      • Added  cobalt_webxr_exclude_patterns  to  bindings.gni  to target  *v8_xr.*  and  *v8_xr_* .           
      • Filtered out generated WebXR V8 bindings in  bindings/modules/v8/BUILD.gn  (concatenated with          
      cobalt_webnn_exclude_patterns ) to prevent their compilation into  libv8.a .
  3. Lightweight C++ Stubbing & Best Practices:
      • Appended WebXR stubs (V8 wrappers and C++ entry points like  XRSystem::From()  returning  nullptr ) to 
      cobalt_modules_stubs.cc .

Bug: 512589073

### DIFF
--- a/third_party/blink/renderer/bindings/bindings.gni
+++ b/third_party/blink/renderer/bindings/bindings.gni
@@ -276,8 +276,3 @@ cobalt_webnn_exclude_patterns = [
   "*v8_ml.*",
   "*v8_ml_*",
 ]
-
-cobalt_webxr_exclude_patterns = [
-  "*v8_xr.*",
-  "*v8_xr_*",
-]
--- a/third_party/blink/renderer/bindings/bindings.gni
+++ b/third_party/blink/renderer/bindings/bindings.gni
@@ -276,3 +276,8 @@ cobalt_webnn_exclude_patterns = [
   "*v8_ml.*",
   "*v8_ml_*",
 ]
+
+cobalt_webxr_exclude_patterns = [
+  "*v8_xr.*",
+  "*v8_xr_*",
+]
--- a/third_party/blink/renderer/bindings/modules/v8/BUILD.gn
+++ b/third_party/blink/renderer/bindings/modules/v8/BUILD.gn
@@ -34,9 +34,8 @@ blink_modules_sources("v8") {
   }
 
   if (is_cobalt) {
-    _filtered_cobalt_sources = filter_exclude(
-        sources,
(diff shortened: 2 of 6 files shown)


yell0wd0g (commented):
In general for a relanding I'd expect:
- The title to say "Reland: abc"
- Two or more commits, with the first one being the original
landed, then reverted PR/CL.
- A description that states the differences clearly,
then the original PR/CL description verbatim. See e.g.
https://chromium-review.git.corp.google.com/c/chromium/src/+/4808891

ckim1346:
> In general for a relanding I'd expect:
> 
> * The title to say "Reland: abc"
> * Two or more commits, with the first one being the original
>   landed, then reverted PR/CL.
> * A description that states the differences clearly,
>   then the original PR/CL description verbatim. See e.g.
>   https://chromium-review.git.corp.google.com/c/chromium/src/+/4808891

Gotcha, closing this PR and opening a new one: #10693

PR #10680, opened 2026-05-28
media: Support changeType in media pipeline

johnedocampo:
Implementing MSE changeType() allows media players to switch containers
or codecs mid-stream without a full pipeline reset. This is essential
for features like adaptive bitrate switching across different formats.

This PR adds:

* **Blink MSE `SourceBuffer.changeType()`**: Added Starboard (`BUILDFLAG(USE_STARBOARD_MEDIA)`) support to pass full raw MIME strings directly to `WebSourceBuffer` without splitting container and codec parameters.
* **Hardware Capability Validation**: Integrated `SbMediaCanChangeType(current_mime, target_mime)` into `ChunkDemuxer::CanChangeType()` to validate platform decoder support before committing to format switches.
* **Decoder Config Synchronization**: Updated `SourceBufferState` and `ChunkDemuxer::ChangeType()` to maintain and update the active `mime_type_`, binding container MIME strings directly to `AudioDecoderConfig` and `VideoDecoderConfig`.
* **Starboard Platform Integration**: Added platform interface stubs and Android build targets for `SbMediaIsChangeTypeTransitionSupported` / `SbMediaCanChangeType`.

A follow-up PR is planned to provide support for Starboard Android builds.

Bug: 132743206

### DIFF
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -78,17 +78,6 @@ bool AudioDecoderConfig::IsValidConfig() const {
 // the config/mime_type is identical to the current config/mime_type.
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
-  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
-  // is intentionally excluded from Matches():
-  // 1. Matches() checks whether underlying hardware decoders require 
-  //    re-initialization based on elementary stream properties (codec,
-  //    profile, resolution, extra_data).
-  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
-  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
-  //    boundaries.
-  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
-  //    container MIME strings differ while elementary video/audio parameters
-  //    remain identical.
   return (
       (codec() == config.codec()) &&
       (bytes_per_channel() == config.bytes_per_channel()) &&
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -88,16 +88,6 @@ bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
   //    boundaries.
   // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
   //    container MIME strings differ while elementary video/audio parameters
-  //    remain identical.  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
-  // is intentionally excluded from Matches():
-  // 1. Matches() checks whether underlying hardware decoders require 
-  //    re-initialization based on elementary stream properties (codec,
-  //    profile, resolution, extra_data).
-  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
-  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
-  //    boundaries.
-  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
-  //    container MIME strings differ while elementary video/audio parameters
   //    remain identical.
   return (
       (codec() == config.codec()) &&
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -78,6 +78,27 @@ bool AudioDecoderConfig::IsValidConfig() const {
 // the config/mime_type is identical to the current config/mime_type.
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
+  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require 
+  //    re-initialization based on elementary stream properties (codec,
+  //    profile, resolution, extra_data).
+  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
+  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
+  //    boundaries.
+  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
+  //    container MIME strings differ while elementary video/audio parameters
+  //    remain identical.  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require 
+  //    re-initialization based on elementary stream properties (codec,
+  //    profile, resolution, extra_data).
+  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
+  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
+  //    boundaries.
+  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
+  //    container MIME strings differ while elementary video/audio parameters
+  //    remain identical.
   return (
       (codec() == config.codec()) &&
       (bytes_per_channel() == config.bytes_per_channel()) &&
--- a/media/base/audio_decoder_config.h
+++ b/media/base/audio_decoder_config.h
@@ -127,7 +127,6 @@ class MEDIA_EXPORT AudioDecoderConfig {
   void set_mime_type(std::string_view mime_type) { mime_type_ = mime_type; }
   const std::string& mime_type() const { return mime_type_; }
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
-
  private:
   // WARNING: When modifying or adding any parameters, update the following:
   // - AudioDecoderConfig::AsHumanReadableString()
--- a/media/base/audio_decoder_config.h
+++ b/media/base/audio_decoder_config.h
@@ -127,6 +127,7 @@ class MEDIA_EXPORT AudioDecoderConfig {
   void set_mime_type(std::string_view mime_type) { mime_type_ = mime_type; }
   const std::string& mime_type() const { return mime_type_; }
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+
  private:
   // WARNING: When modifying or adding any parameters, update the following:
   // - AudioDecoderConfig::AsHumanReadableString()
--- a/media/base/video_decoder_config.cc
+++ b/media/base/video_decoder_config.cc
@@ -88,6 +88,17 @@ bool VideoDecoderConfig::IsValidConfig() const {
 // the config/mime_type is identical to the current config/mime_type.
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 bool VideoDecoderConfig::Matches(const VideoDecoderConfig& config) const {
+  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require 
+  //    re-initialization based on elementary stream properties (codec,
+  //    profile, resolution, extra_data).
+  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
+  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
+  //    boundaries.
+  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
+  //    container MIME strings differ while elementary video/audio parameters
+  //    remain identical.
   return codec() == config.codec() && profile() == config.profile() &&
          alpha_mode() == config.alpha_mode() &&
          video_transformation() == config.video_transformation() &&
--- a/media/base/video_decoder_config.cc
+++ b/media/base/video_decoder_config.cc
@@ -88,17 +88,6 @@ bool VideoDecoderConfig::IsValidConfig() const {
 // the config/mime_type is identical to the current config/mime_type.
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 bool VideoDecoderConfig::Matches(const VideoDecoderConfig& config) const {
-  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
-  // is intentionally excluded from Matches():
-  // 1. Matches() checks whether underlying hardware decoders require 
-  //    re-initialization based on elementary stream properties (codec,
-  //    profile, resolution, extra_data).
-  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
-  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
-  //    boundaries.
-  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
-  //    container MIME strings differ while elementary video/audio parameters
-  //    remain identical.
   return codec() == config.codec() && profile() == config.profile() &&
          alpha_mode() == config.alpha_mode() &&
          video_transformation() == config.video_transformation() &&
--- a/media/base/video_decoder_config.h
+++ b/media/base/video_decoder_config.h
@@ -169,7 +169,6 @@ class MEDIA_EXPORT VideoDecoderConfig {
   void set_mime_type(std::string_view mime_type) { mime_type_ = mime_type; }
   const std::string& mime_type() const { return mime_type_; }
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
-
  private:
   VideoCodec codec_ = VideoCodec::kUnknown;
   VideoCodecProfile profile_ = VIDEO_CODEC_PROFILE_UNKNOWN;
--- a/media/base/video_decoder_config.h
+++ b/media/base/video_decoder_config.h
@@ -169,6 +169,7 @@ class MEDIA_EXPORT VideoDecoderConfig {
   void set_mime_type(std::string_view mime_type) { mime_type_ = mime_type; }
   const std::string& mime_type() const { return mime_type_; }
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+
  private:
   VideoCodec codec_ = VideoCodec::kUnknown;
   VideoCodecProfile profile_ = VIDEO_CODEC_PROFILE_UNKNOWN;
--- a/media/filters/BUILD.gn
+++ b/media/filters/BUILD.gn
@@ -84,7 +84,7 @@ source_set("filters") {
   ]
 
   if (use_starboard_media) {
-        deps += [ "//starboard:starboard_headers_only" ]
+    deps += [ "//starboard:starboard_headers_only" ]
   }
 
   libs = []
--- a/media/filters/BUILD.gn
+++ b/media/filters/BUILD.gn
@@ -83,6 +83,10 @@ source_set("filters") {
     "//ui/gfx/geometry:geometry",
   ]
 
+  if (use_starboard_media) {
+        deps += [ "//starboard:starboard_headers_only" ]
+  }
+
   libs = []
 
   if (proprietary_codecs) {
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -876,7 +876,6 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(it != id_to_mime_map_.end());
   const std::string& mime_type = it->second;
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
-
   source_state->Init(base::BindOnce(&ChunkDemuxer::OnSourceInitDone,
                                     base::Unretained(this), id),
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -32,6 +32,7 @@
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 #include "base/containers/contains.h"
 #include "base/strings/string_split.h"
+#include "starboard/media.h"
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
 namespace {
@@ -1303,6 +1304,22 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
   if (!supports_change_type_) {
     return false;
   }
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto iter = id_to_mime_map_.find(id);
+  std::string current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
+  std::string target_mime = content_type;
+  if (!codecs.empty()) {
+    target_mime += "; codecs=\"" + codecs + "\"";
+  }
+
+  if (!SbMediaIsChangeTypeTransitionSupported(current_mime.c_str(),
+                                              target_mime.c_str())) {
+    LOG(INFO) << "Codec transition unsupported natively on hardware: "
+              << current_mime << " -> " << target_mime;
+    return false;
+  }
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
   // CanChangeType() doesn't care if there has or hasn't been received a first
   // initialization segment for the source buffer corresponding to |id|.
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -32,7 +32,7 @@
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 #include "base/containers/contains.h"
 #include "base/strings/string_split.h"
-#include "starboard/media.h"
+#include "starboard/media.h"  // nogncheck
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
 namespace {
@@ -872,11 +872,6 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(*insert_result.first == id);
   CHECK(insert_result.second);  // Only true if insertion succeeded.
 
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
-  auto it = id_to_mime_map_.find(id);
-  CHECK(it != id_to_mime_map_.end());
-  const std::string& mime_type = it->second;
-#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
   source_state->Init(base::BindOnce(&ChunkDemuxer::OnSourceInitDone,
                                     base::Unretained(this), id),
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
@@ -1305,21 +1300,6 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
     return false;
   }
 
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
-  auto iter = id_to_mime_map_.find(id);
-  std::string current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
-  std::string target_mime = content_type;
-  if (!codecs.empty()) {
-    target_mime += "; codecs=\"" + codecs + "\"";
-  }
-
-  if (!SbMediaCanChangeType(current_mime.c_str(), target_mime.c_str())) {
-    LOG(INFO) << "Codec transition unsupported natively on hardware: "
-              << current_mime << " -> " << target_mime;
-    return false;
-  }
-#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
-
   // CanChangeType() doesn't care if there has or hasn't been received a first
   // initialization segment for the source buffer corresponding to |id|.
 
@@ -1349,10 +1329,30 @@ void ChunkDemuxer::ChangeType(const std::string& id,
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 bool ChunkDemuxer::CanChangeType(const std::string& id,
-                                 const std::string& mime_type) {
+                                 const std::string& target_mime) {
+  std::string current_mime;
+  {
+    base::AutoLock auto_lock(lock_);
+    DCHECK(IsValidId_Locked(id));
+
+    auto itr = source_state_map_.find(id);
+    if (itr == source_state_map_.end()) {
+      LOG(INFO) << "CanChangeType failed: id '" << id
+                << "' not found in source_state_map_.";
+      return false;
+    }
+    current_mime = itr->second->mime_type();
+  }
+
+  if (current_mime.empty() ||
+      !SbMediaCanChangeType(current_mime.c_str(), target_mime.c_str())) {
+    LOG(INFO) << "Codec transition unsupported: current_mime='" << current_mime
+              << "' -> target_mime='" << target_mime << "'";
+    return false;
+  }
   std::string content_type;
   std::string codecs;
-  ParseMimeType(mime_type, &content_type, &codecs);
+  ParseMimeType(target_mime, &content_type, &codecs);
   return CanChangeType(id, content_type, codecs);
 }
 
@@ -1364,8 +1364,6 @@ void ChunkDemuxer::ChangeType(const std::string& id,
   DCHECK(state_ == INITIALIZING || state_ == INITIALIZED) << state_;
   DCHECK(IsValidId_Locked(id));
 
-  id_to_mime_map_[id] = mime_type;
-
   std::string content_type;
   std::string codecs;
   ParseMimeType(mime_type, &content_type, &codecs);
@@ -1373,8 +1371,12 @@ void ChunkDemuxer::ChangeType(const std::string& id,
   std::unique_ptr<media::StreamParser> stream_parser(
       CreateParserForTypeAndCodecs(content_type, codecs, media_log_));
   DCHECK(stream_parser);
-  source_state_map_[id]->ChangeType(std::move(stream_parser),
-                                    ExpectedCodecs(content_type, codecs));
+
+  auto itr = source_state_map_.find(id);
+  DCHECK(itr != source_state_map_.end());
+  itr->second->set_mime_type(mime_type);
+  itr->second->ChangeType(std::move(stream_parser),
+                          ExpectedCodecs(content_type, codecs));
 }
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -1313,8 +1313,7 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
     target_mime += "; codecs=\"" + codecs + "\"";
   }
 
-  if (!SbMediaIsChangeTypeTransitionSupported(current_mime.c_str(),
-                                              target_mime.c_str())) {
+  if (!SbMediaCanChangeType(current_mime.c_str(), target_mime.c_str())) {
     LOG(INFO) << "Codec transition unsupported natively on hardware: "
               << current_mime << " -> " << target_mime;
     return false;
@@ -1347,6 +1346,37 @@ void ChunkDemuxer::ChangeType(const std::string& id,
   source_state_map_[id]->ChangeType(std::move(stream_parser),
                                     ExpectedCodecs(content_type, codecs));
 }
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+bool ChunkDemuxer::CanChangeType(const std::string& id,
+                                 const std::string& mime_type) {
+  std::string content_type;
+  std::string codecs;
+  ParseMimeType(mime_type, &content_type, &codecs);
+  return CanChangeType(id, content_type, codecs);
+}
+
+void ChunkDemuxer::ChangeType(const std::string& id,
+                              const std::string& mime_type) {
+  DVLOG(1) << __func__ << " id=" << id << " mime_type=" << mime_type;
+
+  base::AutoLock auto_lock(lock_);
+  DCHECK(state_ == INITIALIZING || state_ == INITIALIZED) << state_;
+  DCHECK(IsValidId_Locked(id));
+
+  id_to_mime_map_[id] = mime_type;
+
+  std::string content_type;
+  std::string codecs;
+  ParseMimeType(mime_type, &content_type, &codecs);
+
+  std::unique_ptr<media::StreamParser> stream_parser(
+      CreateParserForTypeAndCodecs(content_type, codecs, media_log_));
+  DCHECK(stream_parser);
+  source_state_map_[id]->ChangeType(std::move(stream_parser),
+                                    ExpectedCodecs(content_type, codecs));
+}
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
 double ChunkDemuxer::GetDuration() {
   base::AutoLock auto_lock(lock_);
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -871,6 +871,10 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(*insert_result.first == id);
   CHECK(insert_result.second);  // Only true if insertion succeeded.
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  const std::string& mime_type = id_to_mime_map_[id];
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+
   source_state->Init(base::BindOnce(&ChunkDemuxer::OnSourceInitDone,
                                     base::Unretained(this), id),
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -872,7 +872,9 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(insert_result.second);  // Only true if insertion succeeded.
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  const std::string& mime_type = id_to_mime_map_[id];
+  auto it = id_to_mime_map_.find(id);
+  CHECK(it != id_to_mime_map_.end());
+  const std::string& mime_type = it->second;
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
   source_state->Init(base::BindOnce(&ChunkDemuxer::OnSourceInitDone,
--- a/media/filters/chunk_demuxer.cc
+++ b/media/filters/chunk_demuxer.cc
@@ -1329,52 +1329,51 @@ void ChunkDemuxer::ChangeType(const std::string& id,
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 bool ChunkDemuxer::CanChangeType(const std::string& id,
-                                 const std::string& target_mime) {
-  std::string current_mime;
+                                 const std::string& target_mime_type) {
+  std::string current_mime_type;
   {
     base::AutoLock auto_lock(lock_);
     DCHECK(IsValidId_Locked(id));
 
     auto itr = source_state_map_.find(id);
-    if (itr == source_state_map_.end()) {
-      LOG(INFO) << "CanChangeType failed: id '" << id
-                << "' not found in source_state_map_.";
-      return false;
-    }
-    current_mime = itr->second->mime_type();
+    current_mime_type = itr->second->mime_type();
   }
 
-  if (current_mime.empty() ||
-      !SbMediaCanChangeType(current_mime.c_str(), target_mime.c_str())) {
-    LOG(INFO) << "Codec transition unsupported: current_mime='" << current_mime
-              << "' -> target_mime='" << target_mime << "'";
+  if (!SbMediaCanChangeType(current_mime_type.c_str(), 
+                            target_mime_type.c_str())) {
+    LOG(INFO) << "Codec transition unsupported: current_mime_type='" 
+              << current_mime_type << "' -> target_mime_type='" 
+              << target_mime_type << "'";
     return false;
   }
+
   std::string content_type;
   std::string codecs;
-  ParseMimeType(target_mime, &content_type, &codecs);
+  ParseMimeType(target_mime_type, &content_type, &codecs);
   return CanChangeType(id, content_type, codecs);
 }
 
 void ChunkDemuxer::ChangeType(const std::string& id,
-                              const std::string& mime_type) {
-  DVLOG(1) << __func__ << " id=" << id << " mime_type=" << mime_type;
-
-  base::AutoLock auto_lock(lock_);
-  DCHECK(state_ == INITIALIZING || state_ == INITIALIZED) << state_;
-  DCHECK(IsValidId_Locked(id));
+                              const std::string& target_mime_type) {
+  DVLOG(1) << __func__ << " id=" << id << " target_mime_type=" 
+           << target_mime_type;
 
   std::string content_type;
   std::string codecs;
-  ParseMimeType(mime_type, &content_type, &codecs);
+  ParseMimeType(target_mime_type, &content_type, &codecs);
+
+  base::AutoLock auto_lock(lock_);
+
+  DCHECK(state_ == INITIALIZING || state_ == INITIALIZED) << state_;
+  DCHECK(IsValidId_Locked(id));
 
   std::unique_ptr<media::StreamParser> stream_parser(
       CreateParserForTypeAndCodecs(content_type, codecs, media_log_));
+  // Caller should query CanChangeType() first to protect from failing this.
   DCHECK(stream_parser);
 
   auto itr = source_state_map_.find(id);
-  DCHECK(itr != source_state_map_.end());
-  itr->second->set_mime_type(mime_type);
+  itr->second->set_mime_type(target_mime_type);
   itr->second->ChangeType(std::move(stream_parser),
                           ExpectedCodecs(content_type, codecs));
 }
--- a/media/filters/chunk_demuxer.h
+++ b/media/filters/chunk_demuxer.h
@@ -419,6 +419,11 @@ class MEDIA_EXPORT ChunkDemuxer : public Demuxer {
                   const std::string& content_type,
                   const std::string& codecs);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  bool CanChangeType(const std::string& id, const std::string& mime_type);
+  void ChangeType(const std::string& id, const std::string& mime_type);
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+
   // If the buffer is full, attempts to try to free up space, as specified in
   // the "Coded Frame Eviction Algorithm" in the Media Source Extensions Spec.
   // Returns false iff buffer is still full after running eviction.
--- a/media/filters/chunk_demuxer.h
+++ b/media/filters/chunk_demuxer.h
@@ -422,8 +422,9 @@ class MEDIA_EXPORT ChunkDemuxer : public Demuxer {
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
   // Starboard-specific implementations of CanChangeType() and ChangeType()
   // that accept in the full mime_type string from the web app.
-  bool CanChangeType(const std::string& id, const std::string& target_mime);
-  void ChangeType(const std::string& id, const std::string& mime_type);
+  bool CanChangeType(const std::string& id, 
+                     const std::string& target_mime_type);
+  void ChangeType(const std::string& id, const std::string& target_mime_type);
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
 
   // If the buffer is full, attempts to try to free up space, as specified in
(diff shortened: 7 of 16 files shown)


media/base/audio_decoder_config.cc
```diff
@@ -89,7 +94,13 @@ bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
        config.should_discard_decoder_delay()) &&
       (target_output_channel_layout() ==
        config.target_output_channel_layout()) &&
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      (target_output_sample_format() == config.target_output_sample_format()) &&
+      (from_changeType() == config.from_changeType()) &&
+      (mime_type() == config.mime_type()));
+#else  // BUILDFLAG(USE_STARBOARD_MEDIA)
       (target_output_sample_format() == config.target_output_sample_format()));
+#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/base/video_decoder_config.cc
```diff
@@ -94,7 +100,13 @@ bool VideoDecoderConfig::Matches(const VideoDecoderConfig& config) const {
          extra_data() == config.extra_data() &&
          encryption_scheme() == config.encryption_scheme() &&
          color_space_info() == config.color_space_info() &&
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+         hdr_metadata() == config.hdr_metadata() && level() == config.level() &&
+         from_changeType() == config.from_changeType() &&
+         mime_type() == config.mime_type();
+#else  // BUILDFLAG(USE_STARBOARD_MEDIA)
          hdr_metadata() == config.hdr_metadata() && level() == config.level();
+#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/source_buffer_state.cc
```diff
@@ -177,6 +183,14 @@ void SourceBufferState::ChangeType(
...
 
   stream_parser_ = std::move(new_stream_parser);
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  for (auto& stream : audio_streams_) {
+    stream.second->SetMimeType(new_mime_type);
+  }
+  for (auto& stream : video_streams_) {
+    stream.second->SetMimeType(new_mime_type);
+  }
+#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: It does not.

media/filters/chunk_demuxer.h
```diff
@@ -191,7 +192,7 @@ class MEDIA_EXPORT ChunkDemuxerStream : public DemuxerStream {
   GetPendingBuffers_Locked() EXCLUSIVE_LOCKS_REQUIRED(lock_);
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  const std::string mime_type_;
+  std::string mime_type_ GUARDED_BY(lock_);
```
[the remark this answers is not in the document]
johnedocampo: This comment is false.

media/filters/chunk_demuxer.cc
```diff
@@ -1295,6 +1302,22 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
     return false;
   }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto iter = id_to_mime_map_.find(id);
+  std::string current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
```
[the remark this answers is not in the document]
johnedocampo: We only call the starboard-specific `AddID` function, which will always populate the map.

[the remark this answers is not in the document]
johnedocampo: filed b/519741453

media/filters/source_buffer_state.cc
```diff
@@ -666,6 +680,14 @@ bool SourceBufferState::OnNewConfigs(std::unique_ptr<MediaTracks> tracks) {
       }
 
       track->set_id(stream->media_track_id());
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      // Currently, |PENDING_PARSER_RECONFIG| is only set when a changeType
+      // call occurs, so we're safe to update the config's value.
+      if (state_ == PENDING_PARSER_RECONFIG) {
+        audio_config.set_is_change_type_transition(true);
+      }
```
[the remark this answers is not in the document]
johnedocampo: This comment is not true. `state_` is only set at the very end of `OnNewConfigs`. This should be okay.

media/filters/source_buffer_state.cc
```diff
@@ -751,6 +773,14 @@ bool SourceBufferState::OnNewConfigs(std::unique_ptr<MediaTracks> tracks) {
       }
 
       track->set_id(stream->media_track_id());
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      // Currently, |PENDING_PARSER_RECONFIG| is only set when a changeType
+      // call occurs, so we're safe to update the config's value.
+      if (state_ == PENDING_PARSER_RECONFIG) {
+        video_config.set_is_change_type_transition(true);
+      }
```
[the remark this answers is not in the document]
johnedocampo: ditto

media/base/video_decoder_config.h
```diff
@@ -164,6 +165,14 @@ class MEDIA_EXPORT VideoDecoderConfig {
   // useful for decryptors that decrypts an encrypted stream to a clear stream.
   void SetIsEncrypted(bool is_encrypted);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  void set_is_change_type_transition(bool value) { is_change_type_transition_ = value; }
+  bool is_change_type_transition() const { return is_change_type_transition_; }
+
+  void set_mime_type(const std::string& mime_type) { mime_type_ = mime_type; }
+  const std::string& mime_type() const { return mime_type_; }
```
osagie98: Consider returning a `std::string_view` here. Also check anywhere else in the PR where you use a `const std::string&` to see if you can use `std::string_view` there instead

media/filters/chunk_demuxer.cc
```diff
@@ -1295,6 +1302,22 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
...
+  std::string target_mime = content_type;
+  if (!codecs.empty()) {
+    target_mime += "; codecs=\"" + codecs + "\"";
```
osagie98: If it doesn't require any big changes, `base::StrCat` is preferable to adding strings since it's more memory efficient. I think this is fine here but it's more important in a loop

https://www.chromium.org/developers/chromium-string-usage/#guidelines-for-string-use-in-our-codebase

media/filters/source_buffer_state.cc
```diff
@@ -666,6 +680,14 @@ bool SourceBufferState::OnNewConfigs(std::unique_ptr<MediaTracks> tracks) {
       }
 
       track->set_id(stream->media_track_id());
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      // Currently, |PENDING_PARSER_RECONFIG| is only set when a changeType
```
osagie98: If possible, it would be good to `DCHECK` this

media/filters/chunk_demuxer.cc
johnedocampo: Done.

media/base/video_decoder_config.h
johnedocampo: Done, updated the PR to make use of std::string_view as best as possible.

media/filters/source_buffer_state.cc
johnedocampo: Can you elaborate on what the DCHECK here is? Are we trying to do `DCHECK(state==PENDING_PARSER__RECONFIG)`?

osagie98: Misread this comment before. I just want to make sure that the config is never updated when we don't expect it to (i.e. when the state is something other than `PENDING_PARSER_RECONFIG`), but the way it's written should be fine.

media/base/audio_decoder_config.h
```diff
@@ -16,6 +16,11 @@
 #include "media/base/encryption_scheme.h"
 #include "media/base/media_export.h"
 #include "media/base/sample_format.h"
+#include "media/media_buildflags.h"
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+#include <string_view>
```
osagie98: For simplicity's sake, the code should compile fine if you don't include `string_view` (very likely one of the headers already does) - that would reduce the delta with chrome code though it's technically against the style guide. That also avoids the awkward placement where the include is placed after header files, since it uses a BUILDFLAG

media/base/video_decoder_config.cc
```diff
@@ -5,6 +5,7 @@
 #include "media/base/video_decoder_config.h"
 
 #include <iomanip>
+#include "media/media_buildflags.h"
```
osagie98: This should be with the other header includes

osagie98 (approved):
LGTM, just some minor include comments

media/base/audio_decoder_config.cc
```diff
@@ -5,6 +5,7 @@
 #include "media/base/audio_decoder_config.h"
 
 #include "base/logging.h"
+#include "media/media_buildflags.h"
```
borongc: is this needed? if yes, should this be gated?

media/base/video_decoder_config.cc
```diff
@@ -115,7 +127,14 @@ std::string VideoDecoderConfig::AsHumanReadableString() const {
     << ", encryption scheme: " << encryption_scheme()
     << ", rotation: " << VideoRotationToString(video_transformation().rotation)
     << ", flipped: " << video_transformation().mirrored
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
borongc, line 118: Maybe:

```
    << ", mime: \"" << mime_type() << "\"" 
    << ", is_change_type_transition: "
    << base::ToString(is_change_type_transition())
    << ", color space: " << color_space_info().ToGfxColorSpace().ToString();
```

media/mojo/mojom/video_decoder_config_mojom_traits.h
```diff
@@ -9,13 +9,18 @@
 
 #include "media/base/ipc/media_param_traits.h"
 #include "media/base/video_decoder_config.h"
+#include "media/media_buildflags.h"
```
borongc: why we need this?

borongc:
nit: This is a general change, not just for "Starboard" media pipeline, consider to change the title as if other media pipelines can use this as well.

media/base/audio_decoder_config.cc
borongc: `USE_STARBOARD_MEDIA` should be a global macro, same for all `#include "media/media_buildflags.h"` in this PR.

media/filters/BUILD.gn
```diff
@@ -83,6 +83,10 @@ source_set("filters") {
     "//ui/gfx/geometry:geometry",
   ]
 
+  if (use_starboard_media) {
+        deps += [ "//starboard:starboard_headers_only" ]
```
borongc: extra indents?

media/filters/chunk_demuxer.cc
```diff
@@ -1298,6 +1306,22 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
...
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto iter = id_to_mime_map_.find(id);
+  std::string current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
+  std::string target_mime =
+      codecs.empty()
+          ? content_type
+          : base::StrCat({content_type, "; codecs=\"", codecs, "\""});
+
+  if (!SbMediaCanChangeType(current_mime.c_str(),
```
borongc: This seems to be the only Starboard API in this PR, we should make a comment here like "query supports via platform api; that is, if this PR is upstram, this is the only part that Cobalt needs.

media/base/audio_decoder_config.h
johnedocampo: I think we should keep the inclusion here. While I agree that the code looks awkward here, with this code we do not rely on the other header files having the include.  Doing so may introduce sudden build issues when the code is updated.

media/base/video_decoder_config.cc
johnedocampo: Ack; code is removed and no longer there.

media/base/audio_decoder_config.cc
johnedocampo: Thank you for letting me know. I've removed all `#include "media/media_buildflags.h" includes from this PR.

media/base/video_decoder_config.cc
johnedocampo: Done.

media/mojo/mojom/video_decoder_config_mojom_traits.h
johnedocampo: Originally I thought we needed this for the `USE_STARBOARD_MEDIA` buildflag, but this is not the case. Removed.

media/filters/BUILD.gn
johnedocampo: Done.

johnedocampo:
> nit: This is a general change, not just for "Starboard" media pipeline, consider to change the title as if other media pipelines can use this as well.

Done.

media/filters/chunk_demuxer.cc
johnedocampo: Done. Added a comment about SbMediaCanChangeType.

media/starboard/starboard_renderer.cc
```diff
@@ -726,12 +728,14 @@ void StarboardRenderer::UpdateDecoderConfig(DemuxerStream* stream) {
 
   if (stream->type() == DemuxerStream::AUDIO) {
     const AudioDecoderConfig& decoder_config = stream->audio_decoder_config();
-    player_bridge_->UpdateAudioConfig(decoder_config, stream->mime_type());
+    player_bridge_->UpdateAudioConfig(decoder_config,
```
borongc, line 728: Is this by `pre-commit` formatting?

johnedocampo: Yes, it is.

media/filters/source_buffer_state.cc
```diff
@@ -666,6 +680,14 @@ bool SourceBufferState::OnNewConfigs(std::unique_ptr<MediaTracks> tracks) {
...
+      // call occurs, so we're safe to update the config's value.
+      if (state_ == PENDING_PARSER_RECONFIG) {
+        audio_config.set_is_change_type_transition(true);
+      }
+      audio_config.set_mime_type(stream->mime_type());
```
borongc: As `stream` already has new mime attributes, do we need mime in both AudioDecoderConfig and VideoDecoderConfig?

johnedocampo: Addressed;

Discussed offline, I have refactored the code to be able to update the mime string through Mojo Adapter.

media/starboard/starboard_renderer.cc
```diff
@@ -821,17 +867,29 @@ void StarboardRenderer::OnDemuxerStreamRead(
...
+        pending_video_config_ = config;
+        LOG(INFO) << "Pending Video config change recorded. "
+                     "is_change_type_transition=true";
+      } else {
+        client_->OnVideoConfigChange(config);
+        client_->OnVideoNaturalSizeChange(config.visible_rect().size());
+        paint_video_hole_frame_cb_.Run(config.visible_rect().size());
+        UpdateDecoderConfig(stream);
+      }
     }
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/chunk_demuxer.cc
```diff
@@ -1298,6 +1306,23 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
...
+          ? content_type
+          : base::StrCat({content_type, "; codecs=\"", codecs, "\""});
+
+  // This function queries ChangeType support via platform API.
+  // if (!SbMediaCanChangeType(current_mime.c_str(),
+  //                                             target_mime.c_str())) {
+  //   LOG(INFO) << "Codec transition unsupported natively on hardware: "
+  //             << current_mime << " -> " << target_mime;
+  //   return false;
+  // }
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/source_buffer_state.h
```diff
@@ -58,7 +62,12 @@ class MEDIA_EXPORT SourceBufferState {
   // first ensure that ResetParserState() was done to flush any pending frames
   // from the old stream parser.
   void ChangeType(std::unique_ptr<StreamParser> new_stream_parser,
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+                  std::string_view new_expected_codecs,
+                  std::string_view new_mime_type);
+#else  // BUILDFLAG(USE_STARBOARD_MEDIA)
                   const std::string& new_expected_codecs);
+#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: Done; redid declaration to reduce complexity.

media/filters/source_buffer_state.cc
```diff
@@ -166,7 +166,13 @@ void SourceBufferState::Init(StreamParser::InitCB init_cb,
 
 void SourceBufferState::ChangeType(
     std::unique_ptr<StreamParser> new_stream_parser,
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+    std::string_view new_expected_codecs,
+    std::string_view new_mime_type) {
+#else  // BUILDFLAG(USE_STARBOARD_MEDIA)
     const std::string& new_expected_codecs) {
+#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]

media/filters/chunk_demuxer.cc
```diff
@@ -1298,6 +1306,23 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
     return false;
   }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto iter = id_to_mime_map_.find(id);
+  std::string current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
+  std::string target_mime =
+      codecs.empty()
+          ? content_type
+          : base::StrCat({content_type, "; codecs=\"", codecs, "\""});
```
xiaomings: We should keep the mime type passed from the webapp as is, without parsing and reconstructing it.  Please see the Starboard specific `AddId()` in this file for more details.

media/filters/chunk_demuxer.cc
```diff
@@ -1298,6 +1306,23 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
...
+
+  // This function queries ChangeType support via platform API.
+  if (!SbMediaCanChangeType(current_mime.c_str(),
+                                              target_mime.c_str())) {
```
xiaomings: The indent is off.

media/filters/chunk_demuxer.cc
```diff
@@ -1298,6 +1306,23 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
...
+    LOG(INFO) << "Codec transition unsupported natively on hardware: "
```
xiaomings: Consider just reporting the failing using something like "SbMediaCanChangeType() ... returns false".  The log information contains too much details and may not be accurate.

Also, the comment "// This function queries ChangeType support via platform API." is redundant.

media/filters/chunk_demuxer.cc
```diff
@@ -1298,6 +1306,23 @@ bool ChunkDemuxer::CanChangeType(const std::string& id,
     return false;
   }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto iter = id_to_mime_map_.find(id);
+  std::string current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
+  std::string target_mime =
```
xiaomings: new_mime, to be consistent with other variable names.

media/filters/source_buffer_state.cc
```diff
@@ -666,6 +682,13 @@ bool SourceBufferState::OnNewConfigs(std::unique_ptr<MediaTracks> tracks) {
       }
 
       track->set_id(stream->media_track_id());
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      // Currently, |PENDING_PARSER_RECONFIG| is only set when a changeType
+      // call occurs, so we're safe to update the config's value.
+      if (state_ == PENDING_PARSER_RECONFIG) {
+        audio_config.set_is_change_type_transition(true);
+      }
+#endif
```
xiaomings: nit: add a matching comment.

media/filters/source_buffer_state.cc
```diff
@@ -751,6 +774,13 @@ bool SourceBufferState::OnNewConfigs(std::unique_ptr<MediaTracks> tracks) {
       }
 
       track->set_id(stream->media_track_id());
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      // Currently, |PENDING_PARSER_RECONFIG| is only set when a changeType
+      // call occurs, so we're safe to update the config's value.
+      if (state_ == PENDING_PARSER_RECONFIG) {
+        video_config.set_is_change_type_transition(true);
+      }
+#endif
```
xiaomings: nit: add a matching comment

media/starboard/starboard_renderer.h
```diff
@@ -153,6 +153,8 @@ class MEDIA_EXPORT StarboardRenderer : public Renderer,
   void CreatePlayerBridge();
   void ApplyPendingBounds();
   void UpdateDecoderConfig(DemuxerStream* stream);
+  void ApplyPendingVideoConfig();
+  void ApplyPendingAudioConfig();
```
xiaomings: nit: whenever there are names for both audio and video, put the audio one before the video one.

media/starboard/starboard_renderer.h
```diff
@@ -231,6 +233,13 @@ class MEDIA_EXPORT StarboardRenderer : public Renderer,
 
   std::optional<gfx::Rect> output_rect_;
 
+  // Stored video/audio configs to store the new config from a
+  // SourceBuffer.changeType() call. Configs from changeType() are not applied
+  // to StarboardRenderer when DemuxerStream::kConfigChanged occurs, but when
+  // the first sample of the updated player config is applied.
```
xiaomings: The comment explains what the implementation is doing, consider elaborating on why we need it this way.

media/starboard/starboard_renderer.cc
```diff
@@ -821,17 +865,32 @@ void StarboardRenderer::OnDemuxerStreamRead(
...
   } else if (status == DemuxerStream::kConfigChanged) {
     if (stream == audio_stream_) {
-      client_->OnAudioConfigChange(stream->audio_decoder_config());
+      const AudioDecoderConfig& config = stream->audio_decoder_config();
+      if (config.is_change_type_transition()) {
+        pending_audio_config_ = config;
+        LOG(INFO) << "Pending Audio config change recorded. "
+                     "is_change_type_transition=true";
```
xiaomings: Consider logging something like "Pending audio config change due to changeType".  `is_change_type_transition=true` is slightly redundant as we don't log the path when is_change_type_transition is false.

media/filters/chunk_demuxer.cc
johnedocampo: Done; updated the `changeType_locked` function in source_buffer.cc to now send the mime_type received from the webapp. I've also updated the declarations for `CanChangeType` and `ChangeType` in `web_source_buffer_impl.cc` and `chunk_demuxer.cc` to properly pass and receive the unedited mime_type.

media/filters/chunk_demuxer.cc
johnedocampo: Addressed.

media/filters/chunk_demuxer.cc
johnedocampo: Done. Removed the redundant comment as well.

media/filters/chunk_demuxer.cc
johnedocampo: Done.

media/filters/source_buffer_state.cc

media/filters/source_buffer_state.cc

media/starboard/starboard_renderer.h
johnedocampo: Done.

media/starboard/starboard_renderer.h
johnedocampo: Done; updated comment with explanation.

media/starboard/starboard_renderer.cc
johnedocampo: Done; updated comments for both audio and video.

media/starboard/starboard_renderer.cc
```diff
@@ -325,6 +325,8 @@ void StarboardRenderer::Flush(base::OnceClosure flush_cb) {
   player_bridge_->PrepareForSeek();
   // Reset |playback_rate_| after PrepareForSeek().
   playback_rate_ = 0.0;
+  pending_video_config_.reset();
+  pending_audio_config_.reset();
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/chunk_demuxer.cc
```diff
@@ -1276,6 +1284,62 @@ void ChunkDemuxer::Remove(const std::string& id,
...
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+bool ChunkDemuxer::CanChangeType(const std::string& id,
+                                 const std::string& new_mime_type) {
+  std::string current_mime;
+  {
+    base::AutoLock auto_lock(lock_);
+    auto iter = id_to_mime_map_.find(id);
+    current_mime = iter != id_to_mime_map_.end() ? iter->second : "";
+  }
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/chunk_demuxer.cc
```diff
@@ -1276,6 +1283,58 @@ void ChunkDemuxer::Remove(const std::string& id,
   host_->OnBufferedTimeRangesChanged(GetBufferedRanges_Locked());
 }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+bool ChunkDemuxer::CanChangeType(const std::string& id,
```
borongc, line 1268: This is mostly similar with Chromium's ChunkDemuxer::ChangeType, do we need this to be standalone here?

media/filters/chunk_demuxer.h
```diff
@@ -29,6 +29,10 @@
 #include "media/filters/source_buffer_stream.h"
 #include "media/filters/stream_parser_factory.h"
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
borongc, line 33: This should be moved to below L14 as this is C++ header.

media/filters/chunk_demuxer.h
```diff
@@ -299,9 +304,12 @@ class MEDIA_EXPORT ChunkDemuxer : public Demuxer {
 #endif
 
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  // Special version of AddId() that retains the |mime_type| from the web app.
+  // Special versions of AddId(), CanChangeType(), and ChangeType() that retain
+  // the |mime_type| from the web app.
   [[nodiscard]] Status AddId(const std::string& id,
                              const std::string& mime_type);
+  bool CanChangeType(const std::string& id, const std::string& new_mime);
```
borongc: Note these two function don't align the order in their .cc

media/mojo/clients/mojo_demuxer_stream_impl.cc
```diff
@@ -93,13 +93,23 @@ void MojoDemuxerStreamImpl::OnBufferReady(
...
+#else   // BUILDFLAG(USE_STARBOARD_MEDIA)
     std::move(callback).Run(Status::kConfigChanged, {}, audio_config,
                             video_config);
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
     return;
   }
 
   if (status == Status::kAborted) {
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+    std::move(callback).Run(Status::kAborted, {}, audio_config, video_config, /*mime_type=*/std::nullopt);
```
borongc: nit: max 80 chars per line

media/starboard/starboard_renderer.cc
```diff
@@ -752,6 +752,39 @@ void StarboardRenderer::UpdateDecoderConfig(DemuxerStream* stream) {
...
+  DCHECK(pending_video_config_.has_value());
+  DCHECK(video_stream_);
+
+  LOG(INFO)
+      << "Applying pending Video config change from changeType transition.";
+
+  client_->OnVideoConfigChange(*pending_video_config_);
+  client_->OnVideoNaturalSizeChange(
+      pending_video_config_->visible_rect().size());
+  paint_video_hole_frame_cb_.Run(pending_video_config_->visible_rect().size());
```
borongc: This is in UpdateDecoderConfig, do we need to call it here?

media/starboard/starboard_renderer.h
```diff
@@ -231,6 +233,15 @@ class MEDIA_EXPORT StarboardRenderer : public Renderer,
 
   std::optional<gfx::Rect> output_rect_;
 
+  // Stored video/audio configs to store the new config from a
```
borongc: nit: consider to move this to L298, as these are newly added members.

third_party/blink/renderer/modules/mediasource/source_buffer.cc
```diff
@@ -1073,6 +1073,17 @@ void SourceBuffer::ChangeType_Locked(
   DCHECK(!updating_);
   source_->AssertAttachmentsMutexHeldIfCrossThreadForDebugging();
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  if (!MediaSource::IsTypeSupportedInternal(
+          GetExecutionContext(), type,
+          false /* allow underspecified codecs in |type| */) ||
+      !web_source_buffer_->CanChangeType(type)) {
```
borongc: This is the only different line compared with L1098 --  1106. Consider to simplify them.

borongc:
Could you help to sanity check if this PR works with `kCobaltBypassMojoForMedia` is enabled in media/base/media_switches.cc?

media/filters/chunk_demuxer.cc
johnedocampo: We do need this to be standalone here as we prefer to pass the raw mime_type string from the webapp to starboard. See https://github.com/youtube/cobalt/pull/10680#discussion_r3497180914 for more context.

third_party/blink/renderer/modules/mediasource/source_buffer.cc
johnedocampo: Done; simplified the code.

media/starboard/starboard_renderer.h
johnedocampo: Done.

media/starboard/starboard_renderer.cc
johnedocampo: Discussed offline, we will remove the `paint_video_hole_frame_cb_` callback here as it is called subsequently in `UpdateDecoderConfig`.

media/mojo/clients/mojo_demuxer_stream_impl.cc
johnedocampo: Fixed.

media/filters/chunk_demuxer.h
johnedocampo: I tried to do this, but I got build errors due to the file that defines `BUILDFLAG` not being included before this buildflag. Will keep as-is. 

Also, added an additional include to explicitly support the use of the BUILDFLAG here.

media/filters/chunk_demuxer.h
johnedocampo: I checked the cc file, and currently I am seeing that `CanChangeType` precedes `ChangeType`. Please let me know if you meant something else is amiss with this change.

media/base/audio_decoder_config.cc
```diff
@@ -111,7 +111,12 @@ std::string AudioDecoderConfig::AsHumanReadableString() const {
     << ", target_output_channel_layout: "
     << ChannelLayoutToString(target_output_channel_layout())
     << ", target_output_sample_format: "
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+    << SampleFormatToString(target_output_sample_format())
+    << ", is_change_type_transition: " << base::ToString(is_change_type_transition());
```
xiaomings: nit: if is_change_type_transition_ declaration is moved before target_output_channel_layout_, and is also logged here before target_output_sample_format(), we can simplify the code by saving the #else branch.

media/starboard/starboard_renderer.cc
```diff
@@ -822,17 +860,32 @@ void StarboardRenderer::OnDemuxerStreamRead(
     }
   } else if (status == DemuxerStream::kConfigChanged) {
     if (stream == audio_stream_) {
-      client_->OnAudioConfigChange(stream->audio_decoder_config());
+      const AudioDecoderConfig& config = stream->audio_decoder_config();
```
xiaomings: audio_decoder_config() returns by value and AudioDecoderConfig has move ctor, if we use `AudioDecoderConfig config = stream->audio_decoder_config();` here, we can use `pending_audio_config_ = std::move(config);` below.

The same to `pending_video_config_`.

media/base/video_decoder_config.h
```diff
@@ -190,6 +195,12 @@ class MEDIA_EXPORT VideoDecoderConfig {
   VideoColorSpace color_space_info_;
   std::optional<gfx::HDRMetadata> hdr_metadata_;
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Designates whether the received config was caused by a
+  // SourceBuffer.changeType() call.
+  bool is_change_type_transition_ = false;
+
```
xiaomings: nit: Move the empty line below the #endif.

media/filters/chunk_demuxer.h
```diff
@@ -299,9 +305,12 @@ class MEDIA_EXPORT ChunkDemuxer : public Demuxer {
...
+  void ChangeType(const std::string& id, const std::string& new_mime);
```
xiaomings: nit: new_mime_type

media/filters/chunk_demuxer.cc
```diff
@@ -32,6 +32,7 @@
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 #include "base/containers/contains.h"
 #include "base/strings/string_split.h"
+#include "starboard/media.h"  // nogncheck
```
xiaomings: Feel free to keep this as is.  However, it's reasonable for //media to use a starboard header, and there could be a better way to resolve this.

johnedocampo:
> Could you help to sanity check if this PR works with `kCobaltBypassMojoForMedia` is enabled in media/base/media_switches.cc?

Tested with kCobaltBypassMojoForMedia, I noticed locally that tests started to fail. Filed b/540097677 to track this.

third_party/blink/renderer/platform/media/web_source_buffer_impl.cc
```diff
@@ -206,6 +206,18 @@ void WebSourceBufferImpl::Remove(double start, double end) {
...
+  return demuxer_->CanChangeType(id_, mime_type.Utf8());
+}
+
+void WebSourceBufferImpl::ChangeType(const WebString& mime_type) {
+  // Caller must first call ResetParserState() to flush any pending frames.
+  DCHECK(!demuxer_->IsParsingMediaSegment(id_));
+
+  demuxer_->ChangeType(id_, mime_type.Utf8());
+}
+#else
```
xiaomings: nit: comment

third_party/blink/renderer/modules/mediasource/source_buffer.cc
```diff
@@ -1072,22 +1072,27 @@ void SourceBuffer::ChangeType_Locked(
   DCHECK(source_);
   DCHECK(!updating_);
   source_->AssertAttachmentsMutexHeldIfCrossThreadForDebugging();
-
```
xiaomings, line 1075: nit: restore empty line

media/mojo/clients/mojo_demuxer_stream_impl.cc
```diff
@@ -110,7 +121,11 @@ void MojoDemuxerStreamImpl::OnBufferReady(
     mojom::DecoderBufferPtr mojo_buffer =
         mojo_decoder_buffer_writer_->WriteDecoderBuffer(std::move(buffer));
     if (!mojo_buffer) {
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      std::move(callback).Run(Status::kAborted, {}, audio_config, video_config, /*mime_type=*/std::nullopt);
```
xiaomings: The line length exceeds 80 characters.

media/mojo/services/mojo_demuxer_stream_adapter.cc
```diff
@@ -106,14 +106,24 @@ void MojoDemuxerStreamAdapter::OnBufferReady(
...
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
   DVLOG(3) << __func__
            << ": status=" << ::media::DemuxerStream::GetStatusName(status)
            << ", batch_buffers.size=" << batch_buffers.size();
   DCHECK(read_cb_);
   DCHECK_NE(type_, UNKNOWN);
 
   if (status == kConfigChanged) {
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+    if (mime_type.has_value()) {
```
xiaomings: Can this ever be std::nullopt in this branch?

media/starboard/starboard_renderer.h
```diff
@@ -292,6 +294,15 @@ class MEDIA_EXPORT StarboardRenderer : public Renderer,
   static inline constexpr const char* kSbPlayerCapabilityChangedErrorMessage =
       "MEDIA_ERR_CAPABILITY_CHANGED";
 
+  // Stored video/audio configs to store the new config from a
+  // SourceBuffer.changeType() call. Configs from changeType() are not applied
+  // to StarboardRenderer when DemuxerStream::kConfigChanged occurs, but when
+  // the first sample of the updated player config is applied. We do this to
+  // prevent in-flight samples from being written with the incorrect decoder
+  // configuration.
```
xiaomings: Is it possible for you to elaborate on this?

media/filters/chunk_demuxer.cc
```diff
@@ -1276,6 +1283,58 @@ void ChunkDemuxer::Remove(const std::string& id,
...
+                              const std::string& new_mime) {
+  DVLOG(1) << __func__ << " id=" << id << " new_mime=" << new_mime;
+
+  base::AutoLock auto_lock(lock_);
+
+  DCHECK(state_ == INITIALIZING || state_ == INITIALIZED) << state_;
+  CHECK(IsValidId_Locked(id));
+
+  std::string type, codecs;
+  CHECK(ParseMimeType(new_mime, &type, &codecs));
```
xiaomings: Consider making the function prototype branching on `BUILDFLAG(USE_STARBOARD_MEDIA)` with the `ChangeType()` below, and embed the code into that function, to minimize rebase effort.

The same to `CanChangeType()`.

Feel free to keep this as is if the result doesn't look clean.  However, if the result looks clean, it's better to be implemented in that way.

media/base/audio_decoder_config.cc
johnedocampo: Done; simplified the code.

media/starboard/starboard_renderer.cc
johnedocampo: Done. Changed the code to make use of std::move().

media/base/video_decoder_config.h
johnedocampo: Done.

media/filters/chunk_demuxer.h
johnedocampo: Done.

third_party/blink/renderer/platform/media/web_source_buffer_impl.cc
johnedocampo: Done.

third_party/blink/renderer/modules/mediasource/source_buffer.cc
johnedocampo: Done.

media/mojo/clients/mojo_demuxer_stream_impl.cc
johnedocampo: Fixed.

media/mojo/services/mojo_demuxer_stream_adapter.cc
johnedocampo: Nice catch, this can never be std::nullopt in this branch. Removed the if statement.

media/starboard/starboard_renderer.h
johnedocampo: Added additional comments here about why we store the pending sample. Please let me know if you think there is more to be said here.

media/filters/chunk_demuxer.cc
johnedocampo: Yea, the `// nogncheck` here is due to the `linux-x64x11-no-starboard` causing build errors without it. To my understanding is that `gn` does not recognize build macros, causing the error. Will leave this as-is.

media/filters/chunk_demuxer.cc
johnedocampo: Done. embed the `BUILDFLAG(USE_STARBOARD_MEDIA)` code into the original functions.

media/starboard/starboard_renderer.h
```diff
@@ -292,6 +294,19 @@ class MEDIA_EXPORT StarboardRenderer : public Renderer,
...
+  // Stored audio/video configs to store the new config from a
+  // SourceBuffer.changeType() call. Configs from changeType() are not applied
+  // to StarboardRenderer when DemuxerStream::kConfigChanged occurs, but when
+  // the first sample of the updated player config is applied. It is possible
+  // that when a SourceBuffer.changeType() call is made, we are still reading
+  // buffers from the original mime_type configuration. If we let the config
+  // switch immediately from when StarboardRenderer receives the kConfigChanged
+  // status from the changeType call, the current samples being read will
+  // be incorrectly labelled, potentially causing issues to the SbPlayer
+  // implementation.
```
xiaomings: We probably should no longer store mime types in ChunkDemuxerStream, as they become a per config information instead of a per stream information.  See https://lbshell-internal-review.git.corp.google.com/c/cobalt_src/+/152200 for an example.

We should also revisit whether the Matches() functions of AudioDecoderConfig and VideoDecoderConfig should be updated to reflect this new change.  Once these concepts are sorted out, we probably can get an more accurate representation of the config on reading and these pending configs may no longer be needed.

johnedocampo: I've updated the PR to now store mime types in the configs. Since this is a big change that can be separated from changeType, I will make it into its own PR. Will put this back into draft for now while I spin that PR up.

johnedocampo: Done; Merged https://github.com/youtube/cobalt/pull/11732.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,47 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
+      !SbMediaCanChangeType(current_mime.c_str(), target_mime.c_str())) {
+    LOG(INFO) << "Codec transition unsupported: current_mime='" << current_mime
+              << "' -> target_mime='" << target_mime << "'";
+    return false;
+  }
+  std::string content_type;
+  std::string codecs;
+  ParseMimeType(target_mime, &content_type, &codecs);
+  return CanChangeType(id, content_type, codecs);
+}
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/chunk_demuxer.cc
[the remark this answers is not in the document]
johnedocampo: gn currently does not respect BUILDFLAG arguments. Builds where starboard is not used will always throw a compile error despite the include being guarded by a buildflag. Will keep.

media/filters/source_buffer_state.h
```diff
@@ -57,6 +57,11 @@ class MEDIA_EXPORT SourceBufferState {
             const StreamParser::EncryptedMediaInitDataCB&
                 encrypted_media_init_data_cb);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  const std::string& mime_type() const { return mime_type_; }
+  void set_mime_type(std::string_view mime_type) { mime_type_ = std::string(mime_type); }
```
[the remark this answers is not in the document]
johnedocampo: This is not true; mime_type_ is defined in SourceBufferState.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,51 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
+    DCHECK(IsValidId_Locked(id));
+
+    auto itr = source_state_map_.find(id);
+    if (itr != source_state_map_.end()) {
+      current_mime = itr->second->mime_type();
+    }
+  }
+
+  if (current_mime.empty() ||
+      !SbMediaCanChangeType(current_mime.c_str(), target_mime.c_str())) {
```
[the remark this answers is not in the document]
johnedocampo: This is not an issue; in starboard builds, mime_type_ is properly initialized on startup.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,54 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
```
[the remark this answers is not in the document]
johnedocampo: This shouldn't be an issue.

In Starboard configurations, `WebMediaSourceImpl::AddSourceBuffer` will call `ChunkDemuxer::AddId(id, mime_type)`, that passes the full mime_type_ down. This ensures that mime_type_ will be correctly initialized for `SourceBufferState`.

media/filters/source_buffer_state.h
```diff
@@ -185,6 +185,12 @@ class MEDIA_EXPORT SourceBufferState {
 
   void SetParseWarningCallback(SourceBufferParseWarningCB parse_warning_cb);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Set and get the current |mime_type_| string for the SourceBuffer.
+  void set_mime_type(std::string_view mime_type) { mime_type_ = std::string(mime_type); }
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,54 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
+  std::string codecs;
+  ParseMimeType(mime_type, &content_type, &codecs);
+
+  std::unique_ptr<media::StreamParser> stream_parser(
+      CreateParserForTypeAndCodecs(content_type, codecs, media_log_));
+  DCHECK(stream_parser);
+
+  source_state_map_[id]->set_mime_type(mime_type);
+  source_state_map_[id]->ChangeType(std::move(stream_parser),
+                                    ExpectedCodecs(content_type, codecs));
```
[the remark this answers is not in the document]
johnedocampo: Done; changed code to use an itr.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,54 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
+  std::string current_mime;
+  {
+    base::AutoLock auto_lock(lock_);
+    DCHECK(IsValidId_Locked(id));
+
+    auto itr = source_state_map_.find(id);
+    if (itr != source_state_map_.end()) {
+      current_mime = itr->second->mime_type();
+    }
+  }
```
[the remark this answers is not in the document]
johnedocampo: Done; refactored the code to perform an early exit and also added a log to inform developers about this case being met.

media/filters/BUILD.gn
xiaomings: The indent still seems off to me.

media/filters/chunk_demuxer.h
```diff
@@ -419,6 +419,13 @@ class MEDIA_EXPORT ChunkDemuxer : public Demuxer {
                   const std::string& content_type,
                   const std::string& codecs);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Starboard-specific implementations of CanChangeType() and ChangeType()
+  // that accept in the full mime_type string from the web app.
+  bool CanChangeType(const std::string& id, const std::string& target_mime);
+  void ChangeType(const std::string& id, const std::string& mime_type);
```
xiaomings: Consider unifying the variable names `target_mime` and `mime_type`.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,59 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
+  {
+    base::AutoLock auto_lock(lock_);
+    DCHECK(IsValidId_Locked(id));
+
+    auto itr = source_state_map_.find(id);
+    if (itr == source_state_map_.end()) {
+      LOG(INFO) << "CanChangeType failed: id '" << id
+                << "' not found in source_state_map_.";
+      return false;
+    }
```
xiaomings: Consider removing this as it's DCHECKed above.

xiaomings: Consider removing this as it's DCHECKed above.

third_party/blink/renderer/modules/mediasource/source_buffer.cc
```diff
@@ -1078,16 +1078,22 @@ void SourceBuffer::ChangeType_Locked(
...
   // of impl's StreamParserFactory (since it returns true iff a stream parser
   // can be constructed with |type|). See https://crbug.com/535738.
   if (!MediaSource::IsTypeSupportedInternal(
           GetExecutionContext(), type,
           false /* allow underspecified codecs in |type| */) ||
+#if !BUILDFLAG(USE_STARBOARD_MEDIA)
       !web_source_buffer_->CanChangeType(content_type.GetType(), codecs)) {
+#else  // !BUILDFLAG(USE_STARBOARD_MEDIA)
+      !web_source_buffer_->CanChangeType(type)) {
+#endif  // !BUILDFLAG(USE_STARBOARD_MEDIA)
```
xiaomings: Reverse the branches and use `#if BUILDFLAG(USE_STARBOARD_MEDIA)` instead.

media/filters/chunk_demuxer.cc
```diff
@@ -1326,6 +1327,59 @@ void ChunkDemuxer::ChangeType(const std::string& id,
...
+                              const std::string& mime_type) {
+  DVLOG(1) << __func__ << " id=" << id << " mime_type=" << mime_type;
+
+  base::AutoLock auto_lock(lock_);
+  DCHECK(state_ == INITIALIZING || state_ == INITIALIZED) << state_;
+  DCHECK(IsValidId_Locked(id));
+
+  std::string content_type;
+  std::string codecs;
+  ParseMimeType(mime_type, &content_type, &codecs);
```
xiaomings: Consider moving this above the lock, and also keep the code similar to the original changeType() line by line.

media/filters/BUILD.gn
johnedocampo: Addressed the indent.

media/filters/chunk_demuxer.h
```diff
@@ -419,6 +419,14 @@ class MEDIA_EXPORT ChunkDemuxer : public Demuxer {
                   const std::string& content_type,
                   const std::string& codecs);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Starboard-specific implementations of CanChangeType() and ChangeType()
+  // that accept in the full mime_type string from the web app.
+  bool CanChangeType(const std::string& id, 
+                     const std::string& target_mime_type);
```
johnedocampo, line 426: Done; unified the names to `target_mime_type`.

media/filters/chunk_demuxer.h
johnedocampo: Done; unified the names to `target_mime_type`.

third_party/blink/renderer/modules/mediasource/source_buffer.cc
johnedocampo: Done; reversed the branches.

media/filters/chunk_demuxer.cc
johnedocampo: Done; moved `ParseMimeType` code to above the lock, and updated the code to be similar to the original changeType().

media/filters/chunk_demuxer.cc
johnedocampo: Done; removed the code.

xiaomings (approved):

third_party/blink/renderer/platform/media/web_source_buffer_impl.cc
```diff
@@ -206,6 +206,18 @@ void WebSourceBufferImpl::Remove(double start, double end) {
...
+void WebSourceBufferImpl::ChangeType(const WebString& mime_type) {
+  // Caller must first call ResetParserState() to flush any pending frames.
+  DCHECK(!demuxer_->IsParsingMediaSegment(id_));
+
+  demuxer_->ChangeType(id_, mime_type.Utf8());
+}
+#else  // BUILDFLAG(USE_STARBOARD_MEDIA)
 bool WebSourceBufferImpl::CanChangeType(const WebString& content_type,
                                         const WebString& codecs) {
   return demuxer_->CanChangeType(id_, content_type.Utf8(), codecs.Utf8());
```
andrewsavage1, line 223: Wouldn't it be simpler to make the change just to these functions instead of changing the full pipeline? i.e. here just ignore the codecs argument?

johnedocampo: We want to keep the full `mime_type` string when passing it down to Starboard. Placing it in a `ContentType` object deconstructs the string, which could potentially lead to losing information if we try to reconstruct it.

johnedocampo: More context about this can be found in this earlier comment in the PR: https://github.com/youtube/cobalt/pull/10680#discussion_r3497180914

andrewsavage1: Got it, okay let's merge as is

andrewsavage1 (approved):

PR #10683, opened 2026-05-28
Remove dlopen and dlsym by removing tflite

sacuff:
This removes symbols leaks of dlopen and dlsym, which we do not support in Starboard.

The source of the undefined symbols in C27 is the tflite library, which is primarily a dependency of AI features in upstream Chromium that Cobalt does not use. For example, language detection and auto translation.

This PR refactors build rules to exclude tflite from the Cobalt build, including by removing the Blink language detection library, and is ironically mostly generated by Gemini.

As a bonus, binary size is reduced by ~450K

Bug: 500419612


components/autofill/core/browser/BUILD.gn
```diff
@@ -865,7 +865,7 @@ static_library("browser") {
     "//components/infobars/core",
     "//components/keyed_service/core",
     "//components/language/core/browser",
-    "//components/language_detection/core",
```
y4vor, line 868: The way we do the customizations is to leave the dependency but to have a section after. This helps with future rebases:

if (is_starboard) {
   deps -= [ "//components/language_detection/core" ]
   deps += [ "//components/language_detection/core:constants" ]
}

components/language_detection/core/BUILD.gn
```diff
@@ -80,3 +80,8 @@ source_set("unit_tests") {
     "//components/translate/core/common:common",
   ]
 }
+
+source_set("constants") {
```
y4vor: guard with if (is_cobalt)

components/translate/core/browser/BUILD.gn
```diff
@@ -56,7 +56,7 @@ static_library("browser") {
     "//components/keyed_service/core",
     "//components/language/core/browser",
     "//components/language/core/common",
-    "//components/language_detection/core",
+    "//components/language_detection/core:constants",
```
y4vor: same technique here and below

third_party/blink/renderer/bindings/scripts/bind_gen/__init__.py
```diff
@@ -49,7 +49,8 @@ def init(web_idl_database_path,
          root_gen_dir,
          component_reldirs,
          enable_style_format=False,
-         enable_code_generation_tracing=False):
+         enable_code_generation_tracing=False,
+         union_name_config_path=None):
```
y4vor: why do we need to modify the bind_gen?

third_party/blink/renderer/bindings/BUILD.gn
```diff
@@ -15,11 +15,17 @@ import("//third_party/blink/renderer/bindings/idl_in_core.gni")
 import("//third_party/blink/renderer/bindings/idl_in_extensions_chromeos.gni")
 import("//third_party/blink/renderer/bindings/idl_in_extensions_webview.gni")
 import("//third_party/blink/renderer/bindings/idl_in_modules.gni")
+import("//third_party/blink/public/public_features.gni")
 
 visibility = [ "//third_party/blink/renderer/*" ]
 
 bindings_gen_dir = get_path_info(".", "gen_dir")
 
+_union_name_config = "union_name_map.conf"
```
y4vor: is_cobalt

components/autofill/core/browser/BUILD.gn
```diff
@@ -865,7 +865,7 @@ static_library("browser") {
...
+    "//components/language_detection/core:constants",
```
yell0wd0g: You probably have looked at this but wouldn't it be better
to remove autofill completely? I see that this comes from
our copy of content_shell:
```
$ gn path out/tvos-arm64-simulator_qa/ cobalt components/autofill/core/browser
//cobalt:cobalt --[private]-->
//cobalt/shell:cobalt_shell_lib --[public]-->
//cobalt/shell:cobalt_shell_lib_deps --[public]-->
//components/custom_handlers:custom_handlers --[private]-->
//components/permissions:permissions --[private]-->
//components/content_settings/browser:browser --[private]-->
//components/browsing_data/core:core --[private]-->
//components/autofill/core/browser:browser
```

So perhaps taking a deeper cut, at the
`//cobalt/shell:cobalt_shell_lib`(`_deps`) level,
while harder, might be cleaner in the long run?

third_party/blink/renderer/bindings/generated_in_modules.gni
```diff
@@ -3669,3 +3669,139 @@ generated_interface_extra_sources_for_testing_in_modules = [
...
+  generated_enumeration_sources_in_modules = []
+  generated_enumeration_sources_in_modules = _filtered_enumeration
+
+  _filtered_interface = filter_exclude(generated_interface_sources_in_modules, _ai_generated_files)
+  generated_interface_sources_in_modules = []
+  generated_interface_sources_in_modules = _filtered_interface
+
+  _filtered_union = filter_exclude(generated_union_sources_in_modules, _ai_generated_files)
+  generated_union_sources_in_modules = []
+  generated_union_sources_in_modules = _filtered_union
```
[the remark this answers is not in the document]
sacuff: This is not correct

y4vor (approved):

y4vor (approved):

components/translate/core/common/BUILD.gn
```diff
@@ -27,6 +27,10 @@ static_library("common") {
     "//third_party/metrics_proto",
     "//url",
   ]
+
+  if (is_cobalt) {
+    deps -= [ "//components/language_detection/core:core" ]
+  }
```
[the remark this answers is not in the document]
sacuff: Correct but this is consistent with the use of `core:core` above (from upstream)

components/autofill/core/browser/BUILD.gn
sacuff: Filed b/520048691 to follow up

third_party/blink/public/public_features.gni
```diff
@@ -17,3 +17,4 @@ declare_args() {
   use_webrtc_peer_connection = !is_cobalt
 }
 
+
```
andrewsavage1: nit: revert

third_party/blink/renderer/bindings/bindings.gni
```diff
@@ -297,3 +297,13 @@ cobalt_webgpu_exclude_patterns = [
   "*v8_union_*gpu*",
 ]
 
+cobalt_ai_exclude_patterns = [
```
andrewsavage1: nit: wrap in is_cobalt

andrewsavage1 (approved):

PR #10685, opened 2026-05-28
Add java switches for disabling optimizing compilers

tasunnn:
This PR adds java switches to disable v8 optimizing compilers 
(i.e. maglev, turbofan, and sparkplug) since significant v8 runtime memory 
savings are observed in local runs, see [b/512163617#comment10](https://b.corp.google.com/issues/512163617#comment10)

This will allow us to set up experiment to further evaluate memory
and potential performance impact from disabling those compilers, which 
could also aid investigation of excluding them entirely from cobalt 
build to reduce binary size.

Bug: 512163617


yell0wd0g (approved):
RS LGTM since it seems to just add in the style of the file.

PR #10686, opened 2026-05-28
refactor(ci): add test subtypes and runs_on to test_targets.json

arjungm:
Refactored test_targets.json files for all platforms to include runs_on and test_type fields. Updated GHA actions to leverage these new fields for safer execution and modular archiving.

Bug: 431780245


.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -31,6 +28,20 @@ runs:
...
+        TEST_TARGETS_JSON_FILE: ${{ inputs.test_targets_json_file }}
+      run: |
+        has_host_gunit_tests=$(jq '[.[] | select(.target != null and .runs_on == "host" and .test_type == "gunittest")] | length > 0' "${TEST_TARGETS_JSON_FILE}")
+        has_device_gunit_tests=$(jq '[.[] | select(.target != null and .runs_on == "device" and .test_type == "gunittest")] | length > 0' "${TEST_TARGETS_JSON_FILE}")
+        has_browsertest_or_webtest=$(jq '[.[] | select(.target != null and (.test_type == "browsertest" or .test_type == "webtest"))] | length > 0' "${TEST_TARGETS_JSON_FILE}")
+
+        echo "has_host_gunit_tests=${has_host_gunit_tests}" >> $GITHUB_OUTPUT
+        echo "has_device_gunit_tests=${has_device_gunit_tests}" >> $GITHUB_OUTPUT
+        echo "has_browsertest_or_webtest=${has_browsertest_or_webtest}" >> $GITHUB_OUTPUT
+      shell: bash
```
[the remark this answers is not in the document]
arjungm: this is not an issue because regardless of whether the browsertest runs on host or on device, it will be put into its own archive

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -66,19 +97,42 @@ runs:
...
           fi
+        else
+          FLAGS="--compression zstd --compression-level=9"
         fi
 
+        TARGETS=$(jq -cr '[.[] | select(.target != null and (.test_type == "browsertest" or .test_type == "webtest")) | .target] | join(",")' "${TEST_TARGETS_JSON_FILE}")
+
         if [[ -n "${TARGETS}" ]]; then
           time vpython3 -u ./cobalt/build/archive_test_artifacts.py \
             --source-dir ${GITHUB_WORKSPACE}/cobalt/src \
```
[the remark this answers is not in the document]
arjungm: this is invalid comment because the browsertest would only be included in the "Archive Host Test Artifacts" step if it was also present in the test_targets.json file with the "runs_on: host" key

arjungm:
The copybara check is a false negative and the tvos simulator is taking more than 2 hours to complete which is also a false negative issue 

Aside from all that I do think this PR is ready for a review. PTAL when you can @oxve @linxinan-chops

.github/actions/test_targets_json/action.yaml
```diff
@@ -11,9 +11,12 @@ runs:
     - name: Test target count
       id: parse-json
       run: |
-        test_targets_json="$(jq -cr '[.[] | select(.target != null) | .target]' test_targets.json)"
+        test_targets_json="$(jq -cr '[.[] | select(.target != null and .test_type == "gunittest") | .target]' test_targets.json)"
```
oxve: Would it make sense to make `test_type` and `runs_on` input parameters? Afaik there aren't any places where you would want/need to specify multiple.

I can see the outputs list of this action growing.

.github/actions/on_device_tests/action.yaml
```diff
@@ -104,8 +103,6 @@ runs:
       shell: bash
     - name: Download ${{ matrix.platform }} Test Results from GCS
       if: always()
-      env:
-        TEST_TARGETS_JSON: ${{ inputs.test_targets_json }}
```
oxve, line 108: Using environment variables for template variables is mandatory to avoid env injection.

Our pre-commit check unfortunately only supports validating this on workflows, not on composable actions like this file.

.github/actions/test_targets_json/action.yaml
oxve, line 1: Apparently gunit is a completely different framework from gtest:
https://github.com/cpp-testing/GUnit vs https://github.com/google/googletest

Rename to just `"gtest"`?

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -16,9 +16,6 @@ inputs:
   test_targets_json_file:
     description: "The path to test targets json file."
     required: true
-  archive_browsertests:
-    description: "Indicates if browser tests should be archived."
-    default: "false"
```
oxve, line 21: nice!

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -31,6 +28,21 @@ runs:
     - name: Create Artifacts Directory
       run: mkdir -p "${GITHUB_WORKSPACE}/artifacts/"
```
oxve, line 29: This directory is created 4 more times below. Can you remove the line from the steps below?

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -39,54 +51,87 @@ runs:
...
 
-        # Configure archiver to gz compression for android, raspi, and rdk as the devices don't have zstd installed.
-        FLAGS=""
-        if [[ "${PLATFORM}" == *android* ]]; then
-          FLAGS="--compression gz --archive-per-target --use-android-deps-path --flatten-deps"
-        elif [[ "${PLATFORM}" == *raspi* || "${PLATFORM}" == *rdk* ]]; then
-          FLAGS="--compression gz --archive-per-target --flatten-deps"
-        else
-          FLAGS="--compression zstd --compression-level=9"
+        TARGETS=$(jq -cr '[.[] | select(.target != null and .runs_on == "host" and .test_type == "gunittest") | .target] | join(",")' "${TEST_TARGETS_JSON_FILE}")
```
oxve: Could you move the target calculation to a separate step like you did with the flags?

Instead of using the bash if you would be able to use the built-in `if:` instead, making each packaging step a single call to the packager.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -39,54 +51,87 @@ runs:
...
       env:
         TEST_TARGETS_JSON_FILE: ${{ inputs.test_targets_json_file }}
         PLATFORM: ${{ inputs.platform }}
         CONFIG: ${{ inputs.config }}
-        ARCHIVE_BROWSERTESTS: ${{ inputs.archive_browsertests }}
+        ARCHIVE_FLAGS: ${{ steps.compute-flags.outputs.flags }}
       run: |
         set -x
         mkdir -p "${GITHUB_WORKSPACE}/artifacts/"
         cd cobalt/src/
```
oxve: This line is also superfluous.

All it's used for below is in the path to the packager (`./cobalt/build/archive_test_artifacts.py`). The packager itself is path agnostic.

.github/actions/on_device_tests/action.yaml
arjungm: I changed the on-device-tests action to accept the pre-parsed list of test targets as an input, instead of a json file and then parsing on that. I think that obviates having to use environment variables here, but I could be wrong?

.github/actions/upload_test_artifacts/action.yaml
arjungm: Are you asking if I could make a single packager call in the action? I don't think it'll be possible because eventually Android will request on-host junit and on-device gtest packaging, which would necessitate computing targets and calling the packager twice and I think that'll be a bit confusing:

```
* compute targets -> 3 outputs (host, device, browser/web)
* package host
* package device
* package browser/web
```

or 

```
* compute targets (host)
* compute targets (device)
* compute targets (browser/web)
* package -> check each output and call for each one
```

or

```
* compute targets -> 3 outputs
* package targets -> 3 calls
```

How are you envisioning I factor this part out?

arjungm: I moved targets calculation to a single preceding step, and now it looks like:

```
* compute targets -> 3 outputs
* package host
* package device
* package browser/web
```

I think it looks like we should combine all three package steps together, but I think at minimum we need package-host-device and package-browser-web. However I think we'll need to revisit that coupling when junit tests are added, so I'm in favor of keeping them separate for this PR, and we can refactor it in a follow-up.

oxve (approved):

.github/actions/on_device_tests/action.yaml
oxve: The risk is that someone manages to change to value of `inputs.test_targets` to be a bash string and inject a bash command in (e.g.) the readarray command below. The template variables are injected into the bash script as-is whereas environment variables get properly escaped automatically.

arjungm: it should be fixed in the current PR now

PR #10688, opened 2026-05-29
starboard: Use preallocated pool for `VideoFrameImpl` and decoded audio buffers

kjyoun:
Introduce thread-safe memory pools to recycle allocations and reduce heap fragmentation during playback.

Specifically:
- Implement `FixedSizeMemoryPool` and its wrapper `ObjectPool` (which supports fallback to heap allocation).
- Update `VideoFrameImpl` on Android to use `ObjectPool`, eliminating 60Hz heap allocations during playback. This is controlled by the `kEnableVideoFrameImplMemoryPool` feature flag.
- Update `Buffer` (used for decoded audio buffers) to use `FixedSizeMemoryPool` for allocations up to 32KB, controlled by the `kDecodedAudioBufferPool` feature flag.
- Add unit tests for `FixedSizeMemoryPool`, `ObjectPool`, and `LazyInitializer`.

Local test shows positive result:  http://b/517947010#comment3

Issue: 517947010


starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,162 @@
...
+    SB_CHECK_LE(size, kBufferSize);
+    if (IsFromPool(ptr)) {
+      std::lock_guard lock(mutex_);
+      if (free_list_.size() < kPoolSize) {
+        free_list_.push_back(ptr);
+        return;
+      }
+    }
+    delete[] ptr;
+  }
```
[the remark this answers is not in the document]
kjyoun: Critical catch! Deleting a pooled pointer is indeed a severe heap corruption hazard. I have fixed this by replacing the fallback deletion logic with a strict assertion check `SB_CHECK_LT(free_list_.size(), kPoolSize)` and only deleting non-pooled pointers in the `else` branch. This completely eliminates the vulnerability.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,162 @@
...
+  BufferPool(const BufferPool&) = delete;
+  void operator=(const BufferPool&) = delete;
+};
+
+SB_ONCE_INITIALIZE_FUNCTION(BufferPool, BufferPool::Get)
+
+}  // namespace
+
+Buffer::Buffer(int size) : size_(size) {
+  size_t total_size = static_cast<size_t>(size) + kPaddingSize * 2;
```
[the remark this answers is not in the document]
kjyoun: Great security catch! I have added an explicit `SB_CHECK_GE(size, 0);` assertion at the start of the constructor to ensure `size` is always non-negative and prevent any potential wrap-around or under-allocation.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,162 @@
...
+  size_t total_size = static_cast<size_t>(size_) + kPaddingSize * 2;
+  if (kUseBufferPool && total_size <= kBufferSize) {
+    data_ = BufferPool::Get()->Allocate(total_size);
+    is_pooled_ = true;
+  } else {
+    data_ = new uint8_t[total_size];
+    is_pooled_ = false;
+  }
+  memcpy(data_, that.data_, total_size);
+}
```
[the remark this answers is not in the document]
kjyoun: Excellent catch! Yes, copy-constructing a default-constructed or moved-from `Buffer` (with null data) is perfectly valid, and calling `memcpy` on it is undefined behavior. I have added a null guard at the beginning of the copy constructor to handle this safely without hitting `memcpy`.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,162 @@
...
+constexpr size_t kPoolSize = 128;
+
+#if defined(NDEBUG)
+constexpr int kPaddingSize = 0;
+#else   // defined(NDEBUG)
+constexpr uint8_t kPadding = 0x78;
+constexpr int kPaddingSize = 32;
+#endif  // defined(NDEBUG)
+
+class BufferPool {
```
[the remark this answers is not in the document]
kjyoun: Done. Added a detailed class comment for `BufferPool` explaining its purpose, lifetime, and thread-safety/synchronization model.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,190 @@
...
+    is_pooled_ = true;
+  } else {
+    data_ = new uint8_t[total_size];
+    is_pooled_ = false;
+  }
+#if !defined(NDEBUG)
+  memset(data_, kPadding, kPaddingSize);
+  memset(data_ + kPaddingSize + size_, kPadding, kPaddingSize);
+#endif  // !defined(NDEBUG)
+}
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/shared/starboard/player/buffer_pool.cc
```diff
@@ -0,0 +1,71 @@
...
+  if (!free_list_.empty()) {
+    uint8_t* ptr = free_list_.back();
+    free_list_.pop_back();
+    return ptr;
+  }
+
+  SB_LOG(WARNING)
+      << "BufferPool is exhausted! Fallback to heap allocation for size: "
+      << size;
+  return nullptr;
```
xiaomings: The behavior doesn't align to the above log and the comment in the header file.

starboard/shared/starboard/player/buffer_pool.h
```diff
@@ -0,0 +1,79 @@
...
+  size_t buffer_size() const { return buffer_size_; }
+  size_t pool_size() const { return pool_size_; }
+  size_t free_list_size() const {
+    std::lock_guard<std::mutex> lock(mutex_);
+    return free_list_.size();
+  }
+
+ private:
+  const size_t buffer_size_;
+  const size_t pool_size_;
```
xiaomings: Consider using something like max_number_of_buffers_, `pool_size_` is slightly ambiguous as it can be understood as the size in bytes for the whole memory pool.

starboard/shared/starboard/player/buffer_pool.h
```diff
@@ -0,0 +1,79 @@
...
+  // Allocate a buffer. If size exceeds buffer_size or pool is empty,
+  // it falls back to standard heap allocation.
+  uint8_t* Allocate(size_t size);
+
+  // Free a buffer. If it was allocated from the pool, returns it.
+  // Otherwise, deletes it.
+  void Free(uint8_t* ptr, size_t size);
+
+  // Helper to check if a pointer belongs to this pool.
+  bool IsFromPool(uint8_t* ptr) const;
```
xiaomings: const uint8_t*

starboard/shared/starboard/player/buffer_pool.h
```diff
@@ -0,0 +1,79 @@
...
+    std::lock_guard<std::mutex> lock(mutex_);
+    return free_list_.size();
+  }
+
+ private:
+  const size_t buffer_size_;
+  const size_t pool_size_;
+  std::unique_ptr<uint8_t[]> pool_storage_;
+  std::vector<uint8_t*> free_list_;
+  mutable std::mutex mutex_;
```
xiaomings: nit: move `mutex_` before the variables it guards.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,149 @@
...
+#if BUILDFLAG(IS_ANDROID)
+#include "starboard/shared/starboard/features.h"
+#endif
+
+namespace starboard {
+namespace {
+
+// kBufferSize (32KB) is chosen to be large enough to hold the maximum typical
+// float32 PCM audio frame payloads:
+// - Stereo Float32 PCM at 48kHz (20ms packets) requires 7,680 bytes.
```
xiaomings: The frequency of the audio is irrelevant here.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,149 @@
...
+  size_t total_size = static_cast<size_t>(size) + kPaddingSize * 2;
+  data_ = nullptr;
+  if (UseBufferPool() && total_size <= kBufferSize) {
+    data_ = GetBufferPool()->Allocate(total_size);
+  }
+  is_pooled_ = (data_ != nullptr);
+  if (!data_) {
+    data_ = new uint8_t[total_size];
+  }
+#if !defined(NDEBUG)
```
xiaomings: Not sure if this is on in QA build, if not, consider using !BUILDFLAG(COBALT_IS_RELEASE_BUILD).

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,149 @@
+// Copyright 2023 The Cobalt Authors. All Rights Reserved.
```
kjyoun: 2026

kjyoun: Done

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: sampling rate(== frequency) is relevant in calculating buffer size, which should be > chunk size

chunk size = (sample size in bytes) * # of channels * sample rate/sec * chunk duration.

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: `!defined(NDEBUG)` is disabled on QA build.

Replaced with  !BUILDFLAG(COBALT_IS_RELEASE_BUILD).

starboard/shared/starboard/player/buffer_pool.cc
kjyoun: Done

starboard/shared/starboard/player/buffer_pool.h
kjyoun: Done

starboard/shared/starboard/player/buffer_pool.h
kjyoun: Done

starboard/shared/starboard/player/buffer_pool.h
kjyoun: Renamed to max_buffer_count_

kjyoun:
@xiaomings 

I refactored code so that we can re-use object pool for any object,
and applied to `VideoFrameImpl` and decoded audio buffer.

PTAL

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,102 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/shared/starboard/player/buffer_internal.h"
+
+#include <algorithm>
+#include <cstring>
+#include <utility>
```
xiaomings: nit: are they used?

starboard/shared/starboard/player/fixed_size_memory_pool_test.cc
```diff
@@ -0,0 +1,146 @@
...
+
+  std::vector<std::thread> threads;
+  for (int t = 0; t < kNumThreads; ++t) {
+    threads.emplace_back([&pool]() {
+      for (int i = 0; i < kIterations; ++i) {
+        std::vector<void*> allocated;
+        // Allocate up to 2 blocks per thread.
+        for (size_t j = 0; j < 2; ++j) {
+          void* ptr = pool.Allocate();
+          if (ptr) {
```
xiaomings: ASSERT_NE(ptr, nullptr)?  Since we have 4 threads allocating up to 2 blocks per thread from a 10 block pool, it should never return nullptr.

starboard/shared/starboard/player/fixed_size_memory_pool.h
```diff
@@ -0,0 +1,68 @@
...
+  // Returns a pointer to the allocated block, or nullptr if the pool is
+  // exhausted.
+  void* Allocate();
+
+  // Returns a block of memory back to the pool.
+  // |ptr| must have been allocated from this pool and not yet freed.
+  void Free(void* ptr);
+
+  // Helper to check if a pointer belongs to this pool.
+  bool IsFromPool(void* ptr) const;
```
xiaomings: const void*

starboard/shared/starboard/player/fixed_size_memory_pool.cc
```diff
@@ -0,0 +1,95 @@
...
+#include <limits>
+
+#include "starboard/common/check_op.h"
+#include "starboard/common/log.h"
+#include "starboard/common/pointer_arithmetic.h"
+
+namespace starboard {
+
+FixedSizeMemoryPool::FixedSizeMemoryPool(size_t block_size, size_t capacity)
+    : block_size_([&] {
```
xiaomings: Unless there are already precedence in Chromium code, I would recommend against using IILE for ctor initializer.  This make it less readable.

Consider adding a local helper function like `void* AllocatePoolStorage(size_t aligned_block_size, size_t capacity)`.

starboard/shared/starboard/player/fixed_size_memory_pool.h
```diff
@@ -0,0 +1,68 @@
...
+
+  size_t block_size() const { return block_size_; }
+  size_t capacity() const { return capacity_; }
+  size_t free_list_size() const;
+
+ private:
+  const size_t block_size_;
+  const size_t capacity_;
```
xiaomings: Consider using something like `total_blocks_`, `capacity_` is slightly ambiguous.

starboard/shared/starboard/player/fixed_size_memory_pool.h
```diff
@@ -0,0 +1,68 @@
...
+  void* const pool_storage_;
```
xiaomings: std::unique_ptr<uint8_t[]>?

starboard/shared/starboard/player/fixed_size_memory_pool.cc
```diff
@@ -0,0 +1,95 @@
...
+  SB_CHECK(IsFromPool(ptr));
+  std::lock_guard lock(mutex_);
+  SB_CHECK_LT(free_list_.size(), capacity_);
+  free_list_.push_back(ptr);
+}
+
+bool FixedSizeMemoryPool::IsFromPool(void* ptr) const {
+  if (!pool_storage_) {
+    return false;
+  }
```
xiaomings: This check isn't necessary, consider removing.

starboard/shared/starboard/player/object_pool.h
```diff
@@ -0,0 +1,61 @@
...
+  // Returns a recycled block from the pool if available, or allocates a new one
+  // from heap.
+  void* Allocate();
+
+  // Returns a block of memory back to the pool for future reuse
+  // if it belongs to the pool. Otherwise, deletes it.
+  void Free(void* ptr);
+
+  // Helper to check if a pointer belongs to the pre-allocated pool.
+  bool IsFromPreallocatedPool(void* ptr) const;
```
xiaomings: This function is a bit too delicate to be part of the public interface. The name `ObjectPool` is also slightly misleading as it's not a pool for objects, but more of a `FixedSizeMemoryPool` with heap fallback.

I suggest replacing both `FixedSizeMemoryPool` and `ObjectPool` with a single unified class.  It would maintain a fixed block size and fixed capacity internally. Its `Allocate` method should accept a `size` and prioritize allocations from the pool. If an allocation cannot be fulfilled by the pool, either because the pool is full or the requested size is too large, it should automatically fall back to the heap internally.

This allows us to keep the pool check private, abstracting all allocation details away from the callers.

kjyoun: `IsFromPreallocatedPool` is for testing.

I rewrote test not to use it and remove it from public interface.

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Indeed, we dont need them

Removed

starboard/shared/starboard/player/fixed_size_memory_pool.h
kjyoun: Done

starboard/shared/starboard/player/fixed_size_memory_pool.cc
kjyoun: Done

starboard/shared/starboard/player/fixed_size_memory_pool_test.cc
kjyoun: Done

starboard/shared/starboard/player/fixed_size_memory_pool.cc
kjyoun: Done

starboard/shared/starboard/player/fixed_size_memory_pool.h
kjyoun: thanks for suggestions

renamed

starboard/shared/starboard/player/fixed_size_memory_pool.h
kjyoun: great suggestion. thanks!

starboard/shared/starboard/player/object_pool.h
kjyoun: You are right. I dont need to separate `FixedSizeMemoryPool` and `ObjectPool`

Merged them into `MemoryPool`

starboard/shared/starboard/player/memory_pool.cc
```diff
@@ -0,0 +1,163 @@
...
+                                               size_t total_blocks) {
+  SB_CHECK_LE(block_size, std::numeric_limits<size_t>::max() / total_blocks);
+  std::unique_ptr<uint8_t[]> storage(new uint8_t[block_size * total_blocks]);
+  SB_CHECK(storage);
+  return storage;
+}
+
+// When set, periodically logs the fullness status of the memory pool at the
+// specified interval in milliseconds. std::nullopt disables logging.
+constexpr std::optional<int64_t> kStatusLogIntervalMs = 5'000;
```
xiaomings: It's currently set.  Shall we actually log this in production?

nit 1: constants should be moved before functions.
nit 2: shall we use microseconds here just to be consistent?

starboard/shared/starboard/player/memory_pool.cc
```diff
@@ -0,0 +1,163 @@
...
+                        << name_ << ", requested_size(KiB)=" << (size / 1024)
+                        << ", block_size(KiB)=" << (block_size_ / 1024);
+      } else {
+        SB_LOG(WARNING) << "MemoryPool exhausted (fallback will occur): name="
+                        << name_ << ", block_size(KiB)=" << (block_size_ / 1024)
+                        << ", total_blocks=" << total_blocks_;
+      }
+      last_logged_us_ = now;
+    }
+  }
```
xiaomings: Consider removing or guarding it using kStatusLogIntervalMs so this won't hit production.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -39,23 +39,47 @@
...
 
+LazyInitializer<MemoryPool, /*NoDestruct=*/true> g_pool;
+MemoryPool* GetPool();
+
 class VideoFrameImpl : public VideoFrame {
  public:
   typedef std::function<void()> VideoFrameReleaseCallback;
 
+  void* operator new(size_t size) {
+    if (FeatureList::IsEnabled(features::kVideoFrameImplMemoryPool) &&
```
xiaomings: Consider caching this in a static variable so we don't have to query this frequently.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -39,23 +39,47 @@
...
+        size == sizeof(VideoFrameImpl)) {
```
xiaomings: If this is for potential sub-classes, consider marking VideoFrameImpl as final.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,94 @@
...
+constexpr size_t kPoolSize = 40;
+
+LazyInitializer<MemoryPool, /*NoDestruct=*/true> g_buffer_pool;
+
+MemoryPool* GetPool() {
+  return g_buffer_pool.Get("DecodedAudioBuffer", kBufferSize, kPoolSize);
+}
+
+bool UseBufferPool() {
+#if BUILDFLAG(IS_ANDROID)
```
xiaomings: Consider caching it somehow.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,94 @@
...
+// - Stereo Float32 PCM at 48kHz (20ms packets) requires 7,680 bytes.
+// - 5.1 Surround Float32 PCM at 48kHz (20ms packets) requires 23,040 bytes.
+// - Opus software decoder initial frame safety buffer requires 76,800 bytes.
+// - Large renderer pull sizes (up to 66,560 bytes) driven by platform-specific
+//   audio sink minimum buffer requirements (e.g., 96KB AudioTrack on Android).
+constexpr size_t kBufferSize = 80 * 1024;
+
+// kPoolSize (40) provides a safe margin. Real-world playback tests show a
+// maximum of ~26 buffers in-flight concurrently (leaving 14+ free), while
+// consuming only ~1.25MB of RAM total (32KB * 40).
```
xiaomings: 32KB is mentioned in the comment, but kBufferSize is set to 80KB.

starboard/shared/starboard/player/memory_pool.h
```diff
@@ -0,0 +1,83 @@
...
+  // Otherwise, it is freed from the heap (using ::operator delete).
+  void Free(void* ptr);
+
+  // Helper to check if a pointer belongs to this pool.
+  bool IsFromPool(const void* ptr) const;
+
+  std::string_view name() const { return name_; }
+  size_t block_size() const { return block_size_; }
+  size_t total_blocks() const { return total_blocks_; }
+  size_t free_list_size() const;
```
xiaomings: If they are for testing, consider adding comments on the beginning of the function group mention that.

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Updated comments

starboard/shared/starboard/player/memory_pool.h
kjyoun: Renamed to `GetInfoForTesting()`

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Done

starboard/shared/starboard/player/memory_pool.cc
kjyoun: Done

starboard/shared/starboard/player/memory_pool.cc
kjyoun: We need this warning message to detect when/why buffer pool is not used.

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

kjyoun:
I renamed memory_pool.h with fixed_block_pool.h to avoid conflict with starboard/common/memory_pool.

cobalt/build/android-coat/package.json
```diff
@@ -1,28 +1,58 @@
...
-              "apks/Cobalt.apk"
-          ],
-          "rename_files": [
-              {
-                  "from_file": "apks/Cobalt.apk",
-                  "to_file": "Cobalt.apk"
-              }
-          ],
-          "archive_type": "ARCHIVE_TYPE_FILES"
+            "files": [
```
borongc, line 4: Why this change needs to touch this file?

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -39,23 +39,51 @@
...
 namespace {
 
+using features::FeatureList;
 using jni_zero::AttachCurrentThread;
 using jni_zero::JavaRef;
 using std::placeholders::_1;
 using std::placeholders::_2;
 
-class VideoFrameImpl : public VideoFrame {
+features::CachedFeature g_use_video_frame_pool_cached(
```
borongc: As `features::CachedFeature` is only used on ATV, why not just make them for ATV only, so you don't need to touch `//starboard/linux/shared/BUILD.gn`. IIRC, starboard feature is only for ATV as well.

cobalt/build/android-coat/package.json
kjyoun: for some reason, brian's commit is included in this commit, while rebasing.

https://github.com/youtube/cobalt/pull/10688/changes/4889d4f5d4928efa94c00dfd518720f359853758

Will remove this.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,106 @@
...
+// - 5.1 Surround Float32 PCM at 48kHz (20ms packets) requires 23,040 bytes.
+// - Opus software decoder initial frame safety buffer requires 76,800 bytes.
+// - Large renderer pull sizes (up to 66,560 bytes) driven by platform-specific
+//   audio sink minimum buffer requirements (e.g., 96KB AudioTrack on Android).
+constexpr size_t kBufferKiB = 80;
+
+// kPoolSize (40) provides a safe margin. Real-world playback tests show a
+// maximum of ~26 buffers in-flight concurrently (leaving 14+ free), while
+// consuming ~3.12MiB of RAM total (80KiB * 40).
+constexpr size_t kPoolSize = 40;
```
borongc: Besides this is from a local test, how can we ensure this is truly a risk-free chosen value?

starboard/android/shared/media_codec_video_decoder.cc
borongc: Also, doing this `CachedFeature` to avoid calling FeatureList::IsEnabled involves many code complexity.

Why not just connect with experiment feature, and make this as a global variable in starboard, so you don't need `CachedFeature`?

starboard/shared/starboard/player/buffer_internal.h
```diff
@@ -21,7 +21,6 @@
 #include <utility>
 
 #include "starboard/common/check_op.h"
-#include "starboard/common/log.h"
```
borongc, line 24: while we may not use SB_LOG in this file, removing this header seems to be unrelated to this PR.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,106 @@
...
+
+#include "starboard/shared/starboard/player/buffer_internal.h"
+
+#include "build/build_config.h"
+#include "starboard/shared/starboard/cached_feature.h"
+#include "starboard/shared/starboard/feature_list.h"
+#include "starboard/shared/starboard/player/fixed_block_pool.h"
+#include "starboard/shared/starboard/player/lazy_initializer.h"
+
+#if BUILDFLAG(IS_ANDROID)
```
borongc: I think the original design of `//starboard/shared/starboard` should not have platform-specific codes.

starboard/shared/starboard/feature_list_test.cc
```diff
@@ -146,4 +146,5 @@ TEST_F(StarboardFeatureListTest, CanAccessParams) {
     }
   }
 }
+
```
borongc: nit: new line

starboard/shared/starboard/player/buffer_internal.cc
borongc: Also, did you check various playback speeds (e.g., 2x)?

starboard/shared/starboard/feature_list_test.cc
kjyoun: Done

starboard/shared/starboard/player/buffer_internal.h
kjyoun: This is removed

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: There is no crash risk because if the pool is exhausted (or for >80KB frames), it gracefully falls back to normal heap allocation.
 
Verified current pool size can handle various speed (x2 to x.25).
Any temporary overflow, even if it would happen, will safely use the heap fallback, which we've already seen work smoothly for oversized frames.

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Valid point.

Removed dependencies on starboard::feature, which introduced in android dependencies

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: `CachedFeature` is indeed overengineering. removed it

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -52,10 +55,32 @@ using jni_zero::JavaRef;
 using std::placeholders::_1;
 using std::placeholders::_2;
 
-class VideoFrameImpl : public VideoFrame {
+std::atomic<bool> g_video_frame_pool_enabled{false};
+
+LazyInitializer<FixedBlockPool, /*NoDestruct=*/true> g_pool;
+FixedBlockPool* GetPool();
```
xiaomings: It seems that `LazyInitializer` is always used by setting NoDestruct to true.  In that case, shall we consider removing the class and using:

FixedBlockPool* GetPool() {
  static starboard::NoDestructor<FixedBlockPool> pool(
    "VideoFrameImpl", sizeof(VideoFrameImpl), kPoolSize);
  return pool.get();
}

starboard/shared/starboard/player/buffer_internal.h
```diff
@@ -81,7 +86,21 @@ class Buffer {
     return size_ == 0 ? nullptr : data_ + kPaddingSize;
   }
 
+  static void SetPoolEnabled(bool enabled);
+
+  struct PoolState {
+    size_t free_blocks;
+    size_t total_blocks;
+  };
```
xiaomings: nit: type definition before all functions.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,105 @@
...
+  auto info = GetSmallPool()->GetInfoForTesting();
+  return PoolState{info.free_blocks, info.total_blocks};
+}
+
+Buffer::PoolState Buffer::GetLargePoolStateForTesting() {
+  auto info = GetLargePool()->GetInfoForTesting();
+  return PoolState{info.free_blocks, info.total_blocks};
+}
+
+void Buffer::SetPoolEnabled(bool enabled) {
```
xiaomings, line 54: nit: Function not defined in declaration order.

starboard/shared/starboard/player/fixed_block_pool.cc
```diff
@@ -0,0 +1,203 @@
...
+  SB_CHECK(IsAligned(ptr_val - start_val, block_size_))
+      << "Pointer is misaligned inside pool range: " << ptr;
+}
+
+size_t FixedBlockPool::GetBlockIndex(const void* ptr) const {
+  uintptr_t ptr_val = reinterpret_cast<uintptr_t>(ptr);
+  uintptr_t start_val = reinterpret_cast<uintptr_t>(pool_storage_.get());
+
+  return (ptr_val - start_val) / block_size_;
+}
```
xiaomings: nit: Functions not defined in declaration order.

starboard/shared/starboard/player/fixed_block_pool.h
```diff
@@ -0,0 +1,104 @@
...
+  // If |size| is <= block_size() and the pool has free blocks, allocates from
+  // the pool. Otherwise, falls back to heap allocation (using ::operator new).
+  void* Allocate(size_t size);
+
+  // Frees a block of memory.
+  // If |ptr| belongs to the pool, it is returned to the pool.
+  // Otherwise, it is freed from the heap (using ::operator delete).
+  void Free(void* ptr);
+
+  bool IsFromPool(const void* ptr) const;
```
xiaomings: This should be private.

starboard/shared/starboard/player/fixed_block_pool.h
```diff
@@ -0,0 +1,104 @@
...
+// multiple threads.
+//
+// Lifetime and Ownership:
+// This class does not manage its own lifetime. It is intended to be owned
+// by its creator—either as a long-lived global singleton (e.g., via
+// LazyInitializer) for shared subsystem-level pools, or tied to the lifetime
+// of specific player components (e.g., as a unique_ptr member). Users must
+// ensure the pool outlives all allocations made from it, as freeing a block
+// after the pool is destroyed is undefined behavior.
+class FixedBlockPool {
```
xiaomings: Please feel free to keep the name as is.

This probably can be named as PooledAllocator, which makes the function `IsFromPool` below less ambiguous.

starboard/shared/starboard/player/pooled_allocator.cc
```diff
@@ -0,0 +1,203 @@
...
+    std::lock_guard lock(mutex_);
+    if (!free_list_.empty()) {
+      void* ptr = free_list_.back();
+      free_list_.pop_back();
+
+      AssertValidPoolPtr(ptr);
+      size_t block_index = GetBlockIndex(ptr);
+      SB_CHECK(!is_block_allocated_[block_index])
+          << "Internal error: block already marked as allocated: " << ptr;
+      is_block_allocated_[block_index] = true;
```
xiaomings: This can be qa only code and excluded from gold build.  Please feel free to keep this as is.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -52,10 +55,32 @@ using jni_zero::JavaRef;
...
+
+class VideoFrameImpl final : public VideoFrame {
  public:
   typedef std::function<void()> VideoFrameReleaseCallback;
 
+  void* operator new(size_t size) {
+    if (size == sizeof(VideoFrameImpl) &&
```
xiaomings: "size == sizeof(VideoFrameImpl)" is no longer necessary, as the pool falls back to heap now.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -1163,4 +1199,9 @@ void MediaCodecVideoDecoder::ResetInternal(bool skip_flush) {
   //       slightly flaky as it depends on the behavior of the video renderer.
 }
 
+// static
+void MediaCodecVideoDecoder::SetVideoFramePoolEnabled(bool enabled) {
```
xiaomings: nit: function not in declaration order.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,105 @@
...
+  s_pool_enabled.store(enabled, std::memory_order_relaxed);
+}
+
+uint8_t* Buffer::AllocateData(size_t size) {
+  if (size == 0) {
+    return nullptr;
+  }
```
xiaomings: Consider returning a valid pointer for this case to be more consistent with C++ allocator convention.

We can even allocate from the pool when size is 0, as it shouldn't happen in production.

starboard/shared/starboard/player/buffer_internal.cc
```diff
@@ -0,0 +1,105 @@
...
+      small_pool && small_pool->IsFromPool(ptr)) {
+    small_pool->Free(ptr);
+    return;
+  }
+  if (FixedBlockPool* large_pool = g_large_pool.GetIfInitialized();
+      large_pool && large_pool->IsFromPool(ptr)) {
+    large_pool->Free(ptr);
+    return;
+  }
+  ::operator delete(ptr);
```
xiaomings: This is a new requirement, and the existing design is no longer elegant.  I can think of a few options:
1. Add a `bool allow_fallback` parameter to FixedBlockPool ctor, so the small pool doesn't allow fallback.  And **maybe** add a `TryFree()` function which return false when the ptr isn't allocated from the pool.
2. Refine the FixedBlockPool so it accepts multiple block size and count.

We should move as much logic to the class instead of at the user of the class.

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: Done

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Good point. 

Addressed.

starboard/shared/starboard/player/pooled_allocator.cc
kjyoun: I prefer keep this logic in production. The performance overhead is negligible, and failing fast here is crucial. If this check triggers, the memory pool is already corrupted. 

Crashing immediately gives us a clear stack trace pointing to the pool.. Otherwise, corruption will result in completely incomprehensible crash.

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Done

starboard/shared/starboard/player/buffer_internal.h
kjyoun: Done

starboard/shared/starboard/player/fixed_block_pool.cc
kjyoun: Done

starboard/shared/starboard/player/fixed_block_pool.h
kjyoun: Removed

starboard/shared/starboard/player/fixed_block_pool.h
kjyoun: I adopted `PooledAllocator`

starboard/shared/starboard/player/buffer_internal.cc
kjyoun: Very valid point.

I did additional refactoring to make callsites not care about  internals of handling multiple pools

PTAL

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -52,10 +55,34 @@ using jni_zero::JavaRef;
...
+  }
+
+  void operator delete(void* ptr) {
+    VideoFrameAllocator* allocator =
+        g_video_allocator_ptr.load(std::memory_order_acquire);
+    if (allocator) {
+      allocator->Free(ptr);
+    } else {
+      ::operator delete(ptr);
+    }
```
xiaomings: Consider using

```
   if (g_video_frame_pool_enabled.load(std::memory_order_relaxed)) {
      GetVideoFrameAllocator()->Free(ptr);
      return;
    }
    ::operator delete(ptr);
```

Then we can simplify the logic and remove g_video_allocator_ptr.

starboard/shared/starboard/player/pooled_allocator.cc
```diff
@@ -0,0 +1,277 @@
...
+namespace {
+
+// When set, periodically logs the fullness status of the pool at the
+// specified interval in microseconds, and enables heap fallback warnings.
+// std::nullopt disables all logging and warnings.
+constexpr std::optional<int64_t> kStatusLogIntervalUs = std::nullopt;
+
+// Checks if a value is aligned to an arbitrary boundary (supports
+// non-power-of-two).
+bool IsAlignedTo(uintptr_t value, size_t alignment) {
```
xiaomings: Consider using `IsAligned()` in starboard/common/pointer_arithmetic.h.

starboard/shared/starboard/player/pooled_allocator.h
```diff
@@ -0,0 +1,155 @@
...
+namespace internal {
+
+enum class TryAllocateError {
+  kNotEligible,  // Request size exceeds this pool's block size.
+  kExhausted,    // Size matches, but pool has no free blocks.
+};
+
+// Variant holds either a valid void* pointer on success, or a TryAllocateError
+// on failure.
+using TryAllocateResult = std::variant<void*, TryAllocateError>;
```
xiaomings: Consider using starboard::Expected in starboard/common.

starboard/shared/starboard/player/pooled_allocator.h
```diff
@@ -0,0 +1,155 @@
...
+// A thread-safe memory pool that prioritizes allocating memory blocks of a
+// constant size from a single pre-allocated contiguous block to eliminate
+// memory fragmentation and maximize performance.
+//
+// Threading Model: Thread-safe. All public methods are protected by an
+// internal mutex.
+//
+// Expected Lifetime & Ownership: Typically wrapped inside a `PooledAllocator`
+// and owned as a global singleton (e.g. using
+// `starboard::common::NoDestructor`) by specific media components (like
```
xiaomings: nit: starboard::common::NoDestructor => starboard::NoDestructor

starboard/shared/starboard/player/pooled_allocator.cc
```diff
@@ -0,0 +1,276 @@
...
+  uintptr_t ptr_val = reinterpret_cast<uintptr_t>(ptr);
+  uintptr_t start_val = reinterpret_cast<uintptr_t>(pool_storage_.get());
+
+  SB_CHECK_GE(ptr_val, start_val);
+  SB_CHECK_LT(ptr_val, start_val + block_size_ * total_blocks_);
+  SB_CHECK(IsAlignedTo(ptr_val - start_val, block_size_))
+      << "Pointer is misaligned inside pool range: " << ptr;
+}
+
+size_t FixedBlockPool::GetBlockIndex(const void* ptr) const {
```
xiaomings: AssertValidPoolPtr() is always called before GetBlockIndex().  Maybe call it from GetBlockIndex() or even move its content into GetBlockIndex()?

starboard/shared/starboard/player/pooled_allocator.cc
```diff
@@ -0,0 +1,276 @@
...
+  for (const auto& pool : pools_) {
+    auto result = pool->TryAllocate(size);
+
+    if (auto* ptr = std::get_if<void*>(&result)) {
+      return *ptr;
+    }
+
+    auto* error = std::get_if<internal::TryAllocateError>(&result);
+    SB_CHECK(error);
+    if (*error == internal::TryAllocateError::kNotEligible) {
```
xiaomings: Instead of managing this detail in the FixedBlockPool, the loop should really find the FixedBlockPool with the right size (this information is available to this function) and try to allocate from it, and fallback immediately.  This may also lead to a more straight-forward interface, and allows the logging of pool exhaust in the FixedBlockPool to simplify the logging logic below.

xiaomings (approved):

starboard/android/shared/media_codec_video_decoder.cc
xiaomings: Feel free to keep the existing implementation as is, and simplify the code further once it's experimented in production.

starboard/shared/starboard/player/pooled_allocator.h
kjyoun: Done

starboard/shared/starboard/player/pooled_allocator.cc
kjyoun: Done

starboard/shared/starboard/player/pooled_allocator.cc
kjyoun: Done

starboard/shared/starboard/player/pooled_allocator.h
kjyoun: Good point. I should have used it, the struct that I added by myself.

starboard/shared/starboard/player/pooled_allocator.h
```diff
@@ -0,0 +1,155 @@
...
+  std::vector<bool> is_block_allocated_;  // Guarded by |mutex_|.
+};
+
+}  // namespace internal
+
+// PooledAllocator wraps one or more FixedBlockPool instances and adds heap
+// fallback logic (::operator new/delete) when the pools are disabled,
+// exhausted, or the request is oversized.
+// It supports N-tiered pooling with strict best-fit routing.
+class PooledAllocator {
```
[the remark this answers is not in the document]
kjyoun: Done

starboard/android/shared/media_codec_video_decoder.cc
kjyoun: ack

kjyoun: The reason for keeping `g_video_allocator_ptr` is to safely handle the lifetime of objects, if the feature flag is toggled or updated  while objects allocated from the pool are still in-flight.

Once memory is allocated from the pool, it should be deallocated from the pool, even after flag is turned off.

Though it should not happen, I want to keep the code for now. Will remove them, once I am confident about it.

starboard/shared/starboard/player/pooled_allocator.cc
kjyoun: Done

PR #10691, opened 2026-05-29
Remove the libraries from the ignore list

y4vor:
This list was used in C25 to filter starboard related libraries. This is no longer needed as we use different toolchains and the output is separated.

Bug: 437417239

### DIFF
--- a/starboard/tools/api_leak_detector/libraries_to_ignore
+++ b/starboard/tools/api_leak_detector/libraries_to_ignore
@@ -1,11 +1,1 @@
 # Allowlisted Libraries
-
-libANGLE.a
-libangle_common.a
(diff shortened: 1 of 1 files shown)


starboard/tools/api_leak_detector/libraries_to_ignore
sacuff, line 1: Do we need to keep the file at this point?

sacuff (approved):

starboard/tools/api_leak_detector/libraries_to_ignore
y4vor: I decided to keep it for now as there might be another reason, could be a temp workaround, thing to want to ignore a specific library.

PR #10692, opened 2026-05-29
feat(starboard): enable C++ network dialog raising only for Starboard platforms

andrewsavage1:
Implement the C++ CobaltWebContentsObserver network error dialog raising
logic. Use the clean Starboard SbSystemRaisePlatformError API in C++.

TAG=agy

Bug: 495438901

### DIFF
--- a/cobalt/browser/cobalt_web_contents_observer.cc
+++ b/cobalt/browser/cobalt_web_contents_observer.cc
@@ -26,7 +26,9 @@
 #include "base/logging.h"
 #include "base/metrics/histogram_functions.h"
 #include "base/timer/timer.h"
+#include "content/public/browser/navigation_controller.h"
 #include "content/public/browser/navigation_handle.h"
+#include "content/public/browser/web_contents.h"
 #include "net/base/net_errors.h"
 #include "starboard/system.h"
 #endif  // BUILDFLAG(IS_STARBOARD)
@@ -129,6 +131,16 @@ void CobaltWebContentsObserver::HandlePlatformErrorResponse(
 void CobaltWebContentsObserver::OnPlatformErrorResponse(
     SbSystemPlatformErrorResponse response) {
   is_platform_error_showing_ = false;
+#if !BUILDFLAG(IS_ANDROID)
+  if (response == kSbSystemPlatformErrorResponsePositive) {
+    LOG(INFO) << "Platform error response is POSITIVE. Reloading...";
+    web_contents()->GetController().Reload(content::ReloadType::NORMAL,
+                                           /*check_for_repost=*/false);
+  } else {
+    LOG(INFO) << "Platform error response is NEGATIVE/CANCEL. Stopping app...";
+    SbSystemRequestStop(0);
+  }
+#endif
 }
 #endif  // BUILDFLAG(IS_STARBOARD)
 
--- a/cobalt/browser/cobalt_web_contents_observer.cc
+++ b/cobalt/browser/cobalt_web_contents_observer.cc
@@ -181,8 +181,10 @@ void CobaltWebContentsObserver::OnPlatformErrorResponse(
 #if !BUILDFLAG(IS_ANDROID)
   if (response == kSbSystemPlatformErrorResponsePositive) {
     LOG(INFO) << "Platform error response is POSITIVE. Reloading...";
-    web_contents()->GetController().Reload(content::ReloadType::NORMAL,
-                                           /*check_for_repost=*/false);
+    if (web_contents()) {
+      web_contents()->GetController().Reload(content::ReloadType::NORMAL,
+                                             /*check_for_repost=*/false);
+    }
   } else {
     LOG(INFO) << "Platform error response is NEGATIVE/CANCEL. Stopping app...";
     SbSystemRequestStop(0);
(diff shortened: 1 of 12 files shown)


ckim1346 (approved):
Personally not familiar with `SbSystemRaisePlatformError` but architecturally LGTM

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -46,6 +118,119 @@ CobaltWebContentsObserver::CobaltWebContentsObserver(
...
+    return;
+  }
+  is_platform_error_showing_ = true;
+  platform_error_raised_count_++;
+  base::UmaHistogramCounts100("Cobalt.Network.PlatformErrorCount",
+                              platform_error_raised_count_);
+  auto* bridge =
+      new PlatformErrorBridge(weak_factory_.GetWeakPtr(), navigation_id);
+  pending_platform_error_bridge_ = bridge;
+  if (!SbSystemRaisePlatformError(
```
yell0wd0g, line 176: Would |bridge| leak if `SbSystemRaisePlatformError()` returns false?
Maybe use a `unique_ptr` here.

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -46,6 +118,119 @@ CobaltWebContentsObserver::CobaltWebContentsObserver(
...
+      delete bridge;
+      pending_platform_error_bridge_ = nullptr;
+    }
+  }
+}
+
+// static
+void CobaltWebContentsObserver::HandlePlatformErrorResponse(
+    SbSystemPlatformErrorResponse response,
+    void* user_data) {
```
yell0wd0g, line 191: `HandlePlatformErrorResponse()` is purely internal to
`CobaltWebContentsObserver`, so we could define it the
appropriate data types and avoid a cast? E.g.

```
void CobaltWebContentsObserver::HandlePlatformErrorResponse(
    SbSystemPlatformErrorResponse response,
    PlatformErrorBridge* bridge) {
  bridge->Run(response);
}
```

but then it seems like we don't need a static trampoline
method at all and could just do in the callsite l.191

```
  if (!SbSystemRaisePlatformError( 
         kSbSystemPlatformErrorTypeConnectionError,
          &PlatformErrorBridge::Run, base::Unretained(bridge))) {
```

WDYT?

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -21,8 +21,80 @@
...
+    DCHECK(task_runner_->RunsTasksInCurrentSequence());
+    if (observer_) {
+      observer_->OnPlatformErrorResponse(response, navigation_id_);
+    }
+    delete this;
+  }
+
+  base::WeakPtr<CobaltWebContentsObserver> observer_;
+  int64_t navigation_id_;
+  scoped_refptr<base::SequencedTaskRunner> task_runner_;
```
yell0wd0g: const?

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -21,8 +21,80 @@
...
+  base::Lock lock_;
+  bool has_run_ = false;
```
yell0wd0g: nit: Annotate with GUARDED_BY:
```
  bool has_run_ GUARDED_BY(lock_) = false;
```

Since the lock just protects a boolean, could we
just use a `std::atomic<bool>` ?

cobalt/browser/cobalt_web_contents_observer.cc
```diff
@@ -46,6 +118,119 @@ CobaltWebContentsObserver::CobaltWebContentsObserver(
   }
 }
 
-CobaltWebContentsObserver::~CobaltWebContentsObserver() = default;
+CobaltWebContentsObserver::~CobaltWebContentsObserver() {
+#if BUILDFLAG(IS_STARBOARD)
+  if (pending_platform_error_bridge_) {
+    pending_platform_error_bridge_->Invalidate();
+  }
```
yell0wd0g: IIUC this is to prevent PlatformErrorBridge from calling
back past this point. Instead of adding this method, we
can just do here `weak_factory_->InvalidateWeakPtrs()`
which given l.84-86, will achieve the same goal and
will make `Invalidate()` unnecessary.

cobalt/browser/cobalt_web_contents_observer.cc
andrewsavage1: See the cleanup below for when SbSystemRaisePlatformError returns false. We cannot use a unique_ptr because it could lead to a UAF is the callback is invoked after the observer is destroyed

cobalt/browser/cobalt_web_contents_observer.cc
andrewsavage1: You need to do some casting to align with the C API set by SbSystemRaisePlatformError which uses void*

yell0wd0g (approved):
LGTM now, thanks!
I was going to mention - unit tests? But then this
is on 26.eap, so it might not be worth much. We'll
revisit those if and when this functionality/commit
gets cherry picked to main?

andrewsavage1:
> LGTM now, thanks! I was going to mention - unit tests? But then this is on 26.eap, so it might not be worth much. We'll revisit those if and when this functionality/commit gets cherry picked to main?

Yes, I'm working on #10696 which cherry picks the functionality to main and adds to existing unit tests

PR #10696, opened 2026-05-29
Add platform dialog option for 3p

andrewsavage1:
Bug: 495438901

Functional CP of https://github.com/youtube/cobalt/pull/10692. The end state is more or less the same, but the code had to be merged into existing code on the main branch that wasn't on 26.eap.

TAG=agy
CONV=08d5212c-a049-45e4-a1fa-f344f7d22f8d


ckim1346 (approved):
LGTM -- I know you mentioned it's currently just a Toast popup; is the flow currently that the message pops up, and the response is automatic and then dismisses it through `HandlePlatformErrorResponse`/`OnPlatformErrorResponse`?

yell0wd0g (approved):
Code LGTM since it does look the same as the 26.eap version
reviewed recently.

The unit tests in cobalt/browser/cobalt_web_contents_observer_unittest.cc,
do they apply to 3P, are they executed in CI? Or is the 
plan to add unit tests later for the new code?

PR #10698, opened 2026-05-29
tvos: Implement UrlPlayer for native HLS playback

abhijeetk:
Introduce native HLS support for tvOS by integrating the platform's
AVPlayer into the media pipeline. This allows Cobalt to leverage
native hardware acceleration and platform-specific HLS features
while maintaining the standard Chromium control interface.

A UrlPlayerDemuxer is injected via the demuxer_override mechanism
and exposes placeholder streams to satisfy Chromium's stream-based
initialization. Actual playback is delegated to the platform player.

The HLS URL is carried from the demuxer through
MediaResource::GetMediaUrl() in the renderer process, then bridged
to the GPU process via the StarboardRendererExtension Mojo interface.

Bug: 512045535

### DIFF
--- a/media/mojo/clients/starboard/starboard_renderer_client.cc
+++ b/media/mojo/clients/starboard/starboard_renderer_client.cc
@@ -20,6 +20,7 @@
 #include "base/functional/bind.h"
 #include "base/time/time.h"
 #include "base/unguessable_token.h"
+#include "build/build_config.h"
 #include "media/base/media_log.h"
 #include "media/base/media_resource.h"
 #include "media/base/media_switches.h"
@@ -29,6 +30,10 @@
 #include "media/video/gpu_video_accelerator_factories.h"
 #include "mojo/public/cpp/bindings/callback_helpers.h"
 #include "mojo/public/cpp/bindings/pending_receiver.h"
+
+#if BUILDFLAG(IS_IOS_TVOS)
+#include "url/gurl.h"
+#endif  // BUILDFLAG(IS_IOS_TVOS)
 
 #if BUILDFLAG(IS_ANDROID)
 #include "base/task/bind_post_task.h"
@@ -351,8 +356,18 @@ void StarboardRendererClient::InitializeMojoRenderer(
   DCHECK(media_task_runner_->RunsTasksInCurrentSequence());
   DCHECK(AreMojoPipesConnected());
 
-  if (base::FeatureList::IsEnabled(kCobaltBypassMojoForMedia) ||
-      bypass_mojo_for_media_) {
+  // URL player is incompatible with the bypass bridge because the bypass
+  // bridge proxies demuxer streams, while the URL player delegates playback
+  // entirely to the native platform player.
+  bool is_url_player = false;
+#if BUILDFLAG(IS_IOS_TVOS)
+  GURL url = media_resource->GetMediaUrl();
+  is_url_player = url.is_valid();
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+
+  if (!is_url_player &&
+      (base::FeatureList::IsEnabled(kCobaltBypassMojoForMedia) ||
+       bypass_mojo_for_media_)) {
     bypass_bridge_ = base::MakeRefCounted<MojoRendererBypassBridge>(
         media_task_runner_,
         base::BindRepeating(&StarboardRendererClient::OnTimeUpdateFromBridge,
@@ -377,6 +392,12 @@ void StarboardRendererClient::InitializeMojoRenderer(
             false));
     return;
   }
+
+#if BUILDFLAG(IS_IOS_TVOS)
+  if (is_url_player) {
+    renderer_extension_->SetSourceUrl(url.spec());
+  }
+#endif  // BUILDFLAG(IS_IOS_TVOS)
 
   MojoRendererWrapper::Initialize(media_resource, client, std::move(init_cb));
 }
--- a/media/mojo/services/starboard/starboard_renderer_wrapper.cc
+++ b/media/mojo/services/starboard/starboard_renderer_wrapper.cc
@@ -20,6 +20,7 @@
 #include "base/functional/callback_helpers.h"
 #include "base/task/bind_post_task.h"
 #include "base/time/time.h"
+#include "build/build_config.h"
 #include "media/base/demuxer_stream.h"
 #include "media/base/media_resource.h"
 #include "media/base/starboard/starboard_rendering_mode.h"
@@ -514,6 +515,13 @@ void StarboardRendererWrapper::InitializeWithBypassBridge(
       std::move(audio_proxy), std::move(video_proxy));
   std::move(callback).Run(true);
 }
+
+#if BUILDFLAG(IS_IOS_TVOS)
+void StarboardRendererWrapper::SetSourceUrl(const std::string& source_url) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  GetRenderer()->SetSourceUrl(source_url);
+}
+#endif  // BUILDFLAG(IS_IOS_TVOS)
 
 #if BUILDFLAG(IS_ANDROID)
 void StarboardRendererWrapper::OnOverlayInfoChanged(
--- a/media/starboard/starboard_renderer.cc
+++ b/media/starboard/starboard_renderer.cc
@@ -29,6 +29,7 @@
 #include "media/base/video_codecs.h"
 #include "media/starboard/buildflags.h"
 #include "media/starboard/decoder_buffer_allocator.h"
+#include "starboard/common/log.h"
 #include "starboard/common/media.h"
 #include "starboard/common/player.h"
 #include "starboard/common/string.h"
@@ -212,6 +213,18 @@ void StarboardRenderer::Initialize(MediaResource* media_resource,
 
   client_ = client;
   init_cb_ = std::move(init_cb);
+
+#if BUILDFLAG(IS_IOS_TVOS)
+  if (IsUrlPlayer()) {
+    state_ = STATE_INITIALIZING;
+    if (get_sb_window_handle_cb_) {
+      get_sb_window_handle_cb_.Run();
+      return;
+    }
+    CreatePlayerBridge();
+    return;
+  }
+#endif  // BUILDFLAG(IS_IOS_TVOS)
 
   audio_stream_ = media_resource->GetFirstStream(DemuxerStream::AUDIO);
   video_stream_ = media_resource->GetFirstStream(DemuxerStream::VIDEO);
@@ -530,6 +543,48 @@ void StarboardRenderer::OnSbWindowHandleReady(const uint64_t sb_window_handle) {
   CreatePlayerBridge();
 }
 
+#if BUILDFLAG(IS_IOS_TVOS)
+bool StarboardRenderer::IsUrlPlayer() const {
+  return !source_url_.empty();
+}
+
+void StarboardRenderer::OnUrlPlayerPresenting() {
+  DCHECK(task_runner_->RunsTasksInCurrentSequence());
+  if (!player_bridge_) {
+    return;
+  }
+  int width = 0, height = 0;
+  player_bridge_->GetVideoResolution(&width, &height);
+  if (width > 0 && height > 0) {
+    gfx::Size size(width, height);
+    client_->OnVideoNaturalSizeChange(size);
+    // TODO(b/541996730): Handle resolution changes during adaptive HLS
+    // playback. Currently this is only called once at presenting state.
+    paint_video_hole_frame_cb_.Run(size);
+  } else {
+    LOG(WARNING) << "Platform player reported invalid dimensions (" << width
+                 << "x" << height
+                 << ") at presenting; skipping video hole update.";
+  }
+
+  // Re-apply playback rate; the platform player ignores rate changes
+  // before it is ready to play.
+  player_bridge_->SetPlaybackRate(playback_rate_);
+}
+
+void StarboardRenderer::SetSourceUrl(const std::string& source_url) {
+  DCHECK(task_runner_->RunsTasksInCurrentSequence());
+  source_url_ = source_url;
+}
+
+void StarboardRenderer::OnEncryptedMediaInitDataEncountered(
+    const char* init_data_type,
+    const unsigned char* init_data,
+    unsigned int init_data_length) {
+  // TODO: Forward encrypted media init data to the EME/DRM layer.
+}
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+
 #if BUILDFLAG(IS_ANDROID)
 void StarboardRenderer::OnOverlayInfoChanged(const OverlayInfo& overlay_info) {
   bool overlay_changed = !overlay_info_.RefersToSameOverlayAs(overlay_info);
@@ -578,11 +633,34 @@ SbPlayerInterface* StarboardRenderer::GetSbPlayerInterface() {
   return &sbplayer_interface_;
 }
 
+void StarboardRenderer::UpdateAudioWriteDuration() {
+#if BUILDFLAG(IS_IOS_TVOS)
+  // URL player handles audio natively; no write duration to configure.
+  if (IsUrlPlayer()) {
+    return;
+  }
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+  // TODO(b/267678497): When `player_bridge_->GetAudioConfigurations()`
+  // returns no audio configurations, update the write durations again
+  // before the SbPlayer reaches `kSbPlayerStatePresenting`.
+  audio_write_duration_for_preroll_ = audio_write_duration_ =
+      HasRemoteAudioOutputs(player_bridge_->GetAudioConfigurations())
+          ? audio_write_duration_remote_
+          : audio_write_duration_local_;
+  LOG(INFO) << "audio write duration at " << audio_write_duration_
+            << ", max_video_capabilities_ at " << max_video_capabilities_;
+}
+
 void StarboardRenderer::CreatePlayerBridge() {
   DCHECK(task_runner_->RunsTasksInCurrentSequence());
   DCHECK(init_cb_);
   DCHECK_EQ(state_, STATE_INITIALIZING);
+#if BUILDFLAG(IS_IOS_TVOS)
+  DCHECK(audio_stream_ || video_stream_ ||
+         (IsUrlPlayer() && !audio_stream_ && !video_stream_));
+#else
   DCHECK(audio_stream_ || video_stream_);
+#endif  // BUILDFLAG(IS_IOS_TVOS)
 
   TRACE_EVENT0("media", "StarboardRenderer::CreatePlayerBridge");
 
@@ -622,27 +700,43 @@ void StarboardRenderer::CreatePlayerBridge() {
   // number of active players.
   player_bridge_.reset();
 
-  LOG(INFO) << "Creating SbPlayerBridge.";
-
-  player_bridge_.reset(new SbPlayerBridge(
-      GetSbPlayerInterface(), task_runner_,
-      get_decode_target_graphics_context_provider_func_, audio_config,
-      audio_mime_type, video_config, video_mime_type,
-      // TODO(b/326497953): Support suspend/resume.
-      // TODO(b/326508279): Support background mode.
-      sb_window_, drm_system_, this,
-      // TODO(b/326497953): Support suspend/resume.
-      false,
-      // TODO(b/326825450): Revisit 360 videos.
-      kSbPlayerOutputModeInvalid, max_video_capabilities_,
-      // TODO(b/326654546): Revisit HTMLVideoElement.setMaxVideoInputSize.
-      /*max_video_input_size=*/-1, experimental_features_
+#if BUILDFLAG(IS_IOS_TVOS)
+  if (IsUrlPlayer()) {
+    player_bridge_.reset(new SbPlayerBridge(
+        GetSbPlayerInterface(), task_runner_, source_url_, sb_window_, this,
+        /*allow_resume_after_suspend=*/false, kSbPlayerOutputModePunchOut,
+        base::BindRepeating(
+            &StarboardRenderer::OnEncryptedMediaInitDataEncountered,
+            base::Unretained(this))
+#if BUILDFLAG(COBALT_MEDIA_ENABLE_CVAL)
+            ,
+        /*pipeline_identifier=*/""
+#endif  // BUILDFLAG(COBALT_MEDIA_ENABLE_CVAL)
+        ));
+  } else {
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+    player_bridge_.reset(new SbPlayerBridge(
+        GetSbPlayerInterface(), task_runner_,
+        get_decode_target_graphics_context_provider_func_, audio_config,
+        audio_mime_type, video_config, video_mime_type,
+        // TODO(b/326497953): Support suspend/resume.
+        // TODO(b/326508279): Support background mode.
+        sb_window_, drm_system_, this,
+        // TODO(b/326497953): Support suspend/resume.
+        false,
+        // TODO(b/326825450): Revisit 360 videos.
+        kSbPlayerOutputModeInvalid, max_video_capabilities_,
+        // TODO(b/326654546): Revisit HTMLVideoElement.setMaxVideoInputSize.
+        /*max_video_input_size=*/-1, experimental_features_
 #if BUILDFLAG(IS_ANDROID)
-      ,
-      // TODO: b/475294958 - Revisit platform-specific codes above starboard.
-      surface_view_
+        ,
+        // TODO: b/475294958 - Revisit platform-specific codes above starboard.
+        surface_view_
 #endif  // BUILDFLAG(IS_ANDROID)
-      ));
+        ));
+#if BUILDFLAG(IS_IOS_TVOS)
+  }
+#endif  // BUILDFLAG(IS_IOS_TVOS)
 
   if (!player_bridge_->IsValid()) {
     error_message = player_bridge_->GetPlayerCreationErrorMessage();
@@ -661,17 +755,7 @@ void StarboardRenderer::CreatePlayerBridge() {
     return;
   }
 
-  // TODO(b/267678497): When `player_bridge_->GetAudioConfigurations()`
-  // returns no audio configurations, update the write durations again
-  // before the SbPlayer reaches `kSbPlayerStatePresenting`.
-  audio_write_duration_for_preroll_ = audio_write_duration_ =
-      HasRemoteAudioOutputs(player_bridge_->GetAudioConfigurations())
-          ? audio_write_duration_remote_
-          : audio_write_duration_local_;
-  LOG(INFO) << "SbPlayerBridge created, with audio write duration at "
-            << audio_write_duration_for_preroll_
-            << " and with max_video_capabilities_ at "
-            << max_video_capabilities_;
+  UpdateAudioWriteDuration();
 
   ApplyPendingBounds();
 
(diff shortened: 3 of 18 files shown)


media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -334,6 +339,14 @@ void StarboardRendererClient::InitializeMojoRenderer(
     PipelineStatusCallback init_cb) {
   DCHECK(media_task_runner_->RunsTasksInCurrentSequence());
   DCHECK(AreMojoPipesConnected());
+
+#if BUILDFLAG(IS_IOS_TVOS)
+  GURL url = media_resource->GetMediaUrl();
+  if (url.is_valid()) {
+    renderer_extension_->SetSourceUrl(url.spec());
+  }
+#endif  // BUILDFLAG(IS_IOS_TVOS)
```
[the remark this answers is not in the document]
abhijeetk: This file is already guarded at source level by GN check `is_cobalt && use_starboard_media`
Refer : https://github.com/youtube/cobalt/blob/main/media/mojo/clients/BUILD.gn#L136

Adding `BUILDFLAG(USE_STARBOARD_MEDIA)` will be redundant.

media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -25,6 +26,10 @@
 #include "media/video/gpu_video_accelerator_factories.h"
 #include "mojo/public/cpp/bindings/pending_receiver.h"
 
+#if BUILDFLAG(IS_IOS_TVOS)
+#include "url/gurl.h"
+#endif  // BUILDFLAG(IS_IOS_TVOS)
```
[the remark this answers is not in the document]
abhijeetk: This file is already guarded at source level by GN check `is_cobalt && use_starboard_media`
Refer : https://github.com/youtube/cobalt/blob/main/media/mojo/clients/BUILD.gn#L136

Adding `BUILDFLAG(USE_STARBOARD_MEDIA)` will be redundant.

media/starboard/url_player_demuxer.h
```diff
@@ -0,0 +1,101 @@
...
+
+// Returns true if the URL looks like an HLS stream that the platform URL
+// player (e.g., AVPlayer on tvOS) should handle. This centralizes the
+// detection so that WebMediaPlayerImpl and DemuxerManager use the same logic.
+MEDIA_EXPORT bool IsHlsUrl(const GURL& url);
+
+// Placeholder stream used only to satisfy Chromium's stream-based pipeline
+// initialization. The URL player path ignores these streams and delegates
+// loading, demuxing, buffering, and decoding to the platform URL player.
+class MEDIA_EXPORT UrlPlayerDemuxerStream : public DemuxerStream {
```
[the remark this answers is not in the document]
abhijeetk: Both classes already have comments explaining their purpose and why they exist. These follow the same lifetime and threading conventions as other Demuxer subclasses in the codebase, which don't document these aspects separately either.

media/starboard/url_player_demuxer.h
```diff
@@ -0,0 +1,101 @@
...
+
+ private:
+  const Type type_;
+};
+
+// Demuxer placeholder for URL-based playback. It carries the media URL and
+// exposes placeholder streams because the current Chromium pipeline and Mojo
+// renderer initialization are stream-based. StarboardRenderer receives the URL
+// through the Starboard renderer extension and creates the native URL player.
+class MEDIA_EXPORT UrlPlayerDemuxer : public Demuxer {
```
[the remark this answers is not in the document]
abhijeetk: Both classes already have comments explaining their purpose and why they exist. These follow the same lifetime and threading conventions as other Demuxer subclasses in the codebase, which don't document these aspects separately either.

media/starboard/url_player_demuxer.h
```diff
@@ -0,0 +1,101 @@
...
+
+#ifndef MEDIA_STARBOARD_URL_PLAYER_DEMUXER_H_
+#define MEDIA_STARBOARD_URL_PLAYER_DEMUXER_H_
+
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "base/memory/scoped_refptr.h"
```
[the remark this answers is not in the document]
abhijeetk: Acknowledged.

media/starboard/starboard_renderer.cc
```diff
@@ -981,11 +1044,32 @@ void StarboardRenderer::OnPlayerStatus(SbPlayerState state) {
...
+        // URL player: query video resolution from the platform player and
+        // report it so the web layer creates a video hole for punch-out
+        // rendering.
+        int width = 0, height = 0;
+        player_bridge_->GetVideoResolution(&width, &height);
+        DCHECK(width > 0 && height > 0)
+            << "Platform player should have valid dimensions at presenting.";
+        gfx::Size size(width, height);
+        client_->OnVideoNaturalSizeChange(size);
+        paint_video_hole_frame_cb_.Run(size);
```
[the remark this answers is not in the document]
abhijeetk: Removed the DCHECK, AVPlayer's presentationSize can legitimately be zero during early HLS playback before track metadata resolves. This might create zero-sized video hole. Added a defensive guard that logs a warning and skips the video hole update; the periodic status callback will deliver valid dimensions shortly after. But better not to send it from GPU to RENDERER and save extra IPC.

abhijeetk:

borongc:
@jasonzhangxx is out, this PR needs his review.

abhijeetk:
I wanted to share an update on the URL player (AVPlayer/HLS) work for tvOS.

I've closed this PR #10698 ("media: Implement UrlPlayer for native HLS on tvOS"). This PR added URL player support by extending the existing Starboard renderer and wrapper classes inline. The new approach introduces dedicated renderer, client, and factory classes (UrlPlayerRenderer UrlPlayerRendererClient, UrlPlayerRendererWrapper) with a clean factory-level split, rather than piggybacking on the Starboard code path. This gives us better separation of concerns and makes the code easier to review and maintain.

Here is the new PR stack. Please review in order, as each PR builds on the previous one:
  1. #10979 - media: Add RendererType::kUrlPlayer and buildflag
  2. #10989 - media: Add InitializeWithUrl transport and URL player renderer
  3. #10998 - media: Add URL-player renderer client and GetMediaUrl
  4. #11056 - media: Wire URL-player renderer factory plumbing
  5. #11067 - media: Add URL-player demuxer and HLS URL predicate
  6. #11066 - media: Route HLS URLs to URL-player Renderer

All PRs have been build-verified and tested on tvOS simulator.

jasonzhangxx:
Let's change the to "tvos: Implement UrlPlayer for native HLS playback", as the change is specifically to tvos.

cobalt/shell/browser/shell_platform_delegate_ios.mm
```diff
@@ -365,7 +365,8 @@ - (void)viewDidLoad {
   // Note that the actual size and and position of this view are irrelevant at
   // this point: it will be changed in starboard's
   // AVSBVideoRenderer::SetBounds() when necessary.
-  UIView* playerContainerView = [[UIView alloc] init];
+  UIView* playerContainerView =
+      [[UIView alloc] initWithFrame:[UIScreen mainScreen].bounds];
```
jasonzhangxx: @abhijeetk, could you help me understand this change? I'm curious why we need it, and if it could have problem when player is non full screen.

borongc:
Left some comments in https://github.com/youtube/cobalt/pull/11548, which should be addressed in this PR.

media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -351,8 +356,18 @@ void StarboardRendererClient::InitializeMojoRenderer(
...
+  // URL player is incompatible with the bypass bridge because the bypass
+  // bridge proxies demuxer streams, while the URL player delegates playback
+  // entirely to the native platform player.
+  bool is_url_player = false;
+#if BUILDFLAG(IS_IOS_TVOS)
+  GURL url = media_resource->GetMediaUrl();
+  is_url_player = url.is_valid();
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+
+  if (!is_url_player &&
```
jasonzhangxx, line 368: @abhijeetk, can you help to confirm if url player and mojo bypass mechanisms mutually exclusive?

My understanding is we can still bypass mojo with url player. @borongc to review this piece of code.

media/mojo/services/starboard/starboard_renderer_wrapper.cc
```diff
@@ -515,6 +515,13 @@ void StarboardRendererWrapper::InitializeWithBypassBridge(
   std::move(callback).Run(true);
 }
 
+void StarboardRendererWrapper::SetSourceUrl(const std::string& source_url) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+#if SB_HAS(PLAYER_WITH_URL)
```
jasonzhangxx: `SB_HAS(PLAYER_WITH_URL)` is from cobalt legacy code. In the chromium architecture, it's no longer necessary. We can assume tvos always supports url player, so it can be simply replaced by `BUILDFLAG(IS_IOS_TVOS)` in starboard code. Can you please make a PR to remove `SB_HAS_PLAYER_WITH_URL` and `SB_URL_PLAYER_INCLUDE_PATH`, then rebase this PR?

media/starboard/starboard_renderer.h
```diff
@@ -150,6 +156,12 @@ class MEDIA_EXPORT StarboardRenderer : public Renderer,
     STATE_ERROR
   };
 
+  // Returns true when the renderer is operating in URL player mode.
+  bool IsUrlPlayer() const;
```
jasonzhangxx, line 162: `IsUrlPlayer` and `OnUrlPlayerPresenting` are for url player on tvos, we can guard them behind `SB_HAS(PLAYER_WITH_URL)` or `BUILDFLAG(IS_IOS_TVOS)`.

media/starboard/starboard_renderer.cc
```diff
@@ -530,6 +540,52 @@ void StarboardRenderer::OnSbWindowHandleReady(const uint64_t sb_window_handle) {
...
+#if SB_HAS(PLAYER_WITH_URL)
+  if (!player_bridge_) {
+    return;
+  }
+  int width = 0, height = 0;
+  player_bridge_->GetVideoResolution(&width, &height);
+  if (width > 0 && height > 0) {
+    gfx::Size size(width, height);
+    client_->OnVideoNaturalSizeChange(size);
+    paint_video_hole_frame_cb_.Run(size);
```
jasonzhangxx, line 563: The current implementation triggers the video natural size update and the `paint_video_hole_frame_cb_` only once here. As hls content is adaptive, will this work properly if the video resolution changes mid-playback?

media/starboard/starboard_renderer.cc
```diff
@@ -860,6 +936,11 @@ void StarboardRenderer::OnNeedData(DemuxerStream::Type type,
     return;
   }
 
+  // URL player handles all buffering natively so ignore OnNeedData.
+  if (IsUrlPlayer()) {
```
jasonzhangxx: Will `OnNeedData` be called by url player?

media/starboard/starboard_renderer.cc
```diff
@@ -622,39 +678,59 @@ void StarboardRenderer::CreatePlayerBridge() {
...
-      // TODO: b/475294958 - Revisit platform-specific codes above starboard.
-      surface_view_
+        ,
+        // TODO: b/475294958 - Revisit platform-specific codes above starboard.
+        surface_view_
 #endif  // BUILDFLAG(IS_ANDROID)
-      ));
+        ));
+  }
   if (player_bridge_->IsValid()) {
```
jasonzhangxx: I made a cleanup pr go/cobalt-pr/11583 to simply the code. Please rebase the change after that pr is merged. That pr can help to simplify the code here.

media/starboard/url_player_demuxer.cc
```diff
@@ -0,0 +1,150 @@
...
+UrlPlayerDemuxerStream::UrlPlayerDemuxerStream(Type type) : type_(type) {}
+
+UrlPlayerDemuxerStream::~UrlPlayerDemuxerStream() = default;
+
+void UrlPlayerDemuxerStream::Read(uint32_t count, ReadCB read_cb) {
+  NOTREACHED();
+}
+
+AudioDecoderConfig UrlPlayerDemuxerStream::audio_decoder_config() {
+  return AudioDecoderConfig(AudioCodec::kAAC, kSampleFormatS16,
```
jasonzhangxx, line 50: Will this function and `video_decoder_config` be actually used?

jasonzhangxx (commented):
Let's change the title to start with "tvos:" as it's specifically for tvos.

media/mojo/services/starboard/starboard_renderer_wrapper.cc
abhijeetk: I have created separate PR for tracking this : https://github.com/youtube/cobalt/pull/11606

media/mojo/clients/starboard/starboard_renderer_client.cc
abhijeetk: Hi @jasonzhangxx , borong has reviewed same changes in PR : #11548 

As per this [comment](https://github.com/youtube/cobalt/pull/11548#discussion_r3668888265), I can infer that bypass mojo is not applicable to tvOS.

media/starboard/starboard_renderer.cc
abhijeetk: You're right, `OnUrlPlayerPresenting()` is called once from `OnPlayerStatus()` when the player transitions to `kSbPlayerStatePresenting`. There's currently no mechanism to detect resolution changes mid-playback.

Will add this in a follow-up PR if preferred.

media/starboard/starboard_renderer.cc
abhijeetk: Acknowledged.

media/starboard/starboard_renderer.h
abhijeetk: `IsUrlPlayer()` is called unconditionally in Initialize(), CreatePlayerBridge(), OnNeedData() etc. Guarding the declaration would require wrapping every call site with `#if BUILDFLAG(IS_IOS_TVOS)`, adding noise and making rebasing harder. 

Currently it just returns false on non-tvOS via `#if/#else` in the body, keeping call sites clean. Happy to change if you prefer though.

media/starboard/url_player_demuxer.cc
abhijeetk: No, this config is not used. These are placeholder configs, not configs used for URL-player decoding. These placeholder configs are required for media pipeline initialization, which is stream-based. 

UrlPlayerDemuxer exposes audio/video DemuxerStreams via `UrlPlayerDemuxer::GetAllStreams()`, and the Mojo demuxer stream initialization reads the matching audio_decoder_config()/video_decoder_config() for initial stream metadata.

These configs need to be valid because they are sent over Mojo IPC during stream initialization, and the Mojo traits reject invalid configs via `IsValidConfig()` ([media/mojo/mojom/audio_decoder_config_mojom_traits.cc](https://github.com/youtube/cobalt/blob/main/media/mojo/mojom/audio_decoder_config_mojom_traits.cc#L63), [media/mojo/mojom/
  video_decoder_config_mojom_traits.cc:73](https://github.com/youtube/cobalt/blob/main/media/mojo/mojom/video_decoder_config_mojom_traits.cc#L73)). AAC/H264 are only valid placeholder configs for the stream-based pipeline handshake; the native URL player does not use them for actual decoding.

media/starboard/starboard_renderer.cc
abhijeetk: No, it should not be called for the URL-player path. OnNeedData() is reached through SbPlayerBridge::OnDecoderStatus(), which is wired from the decoder-status callback passed to normal SbPlayerCreate(). The URL-player path uses SbUrlPlayerCreate(), whose API does not take a decoder-status callback, so URL player should not request demuxed data through OnNeedData(). 

The early return in `OnNeedData` is just a defensive guard as this function deals with stream which we want to bypass for URL-Player. I’ll remove this guard unless we want to keep it only as defensive protection.

cobalt/shell/browser/shell_platform_delegate_ios.mm
abhijeetk: I checked this against C25. In C25, `SBDApplicationView` was initialized with `UIScreen.mainScreen.bounds` in [application_window.mm](https://github.com/youtube/cobalt/blob/25.lts.1%2B/starboard/shared/uikit/application_window.mm#L662), and the player
  container was initialized with that application view frame in [application_view.mm](https://github.com/youtube/cobalt/blob/25.lts.1%2B/starboard/shared/uikit/application_view.mm#L133). So using `[UIScreen mainScreen].bounds` here keeps the new chrobalt shell path closer to the old C25 behavior.

That said, I also verified that reverting this to `[[UIView alloc] init]` does not break playback. The URL player view (`SBDPlayerView`) is created with its own fullscreen frame in 
[application_player.mm](https://github.com/youtube/cobalt/blob/main/starboard/tvos/shared/media/application_player.mm#L237). It is later resized via `setBoundsX:y:width:height:` in [application_player.mm](https://github.com/youtube/cobalt/blob/main/starboard/tvos/shared/media/application_player.mm#L489). 

So the initial frame of the chrobalt shell `playerContainerView` is not what determines the final video size. Initializing it with `[UIScreen mainScreen].bounds` is mainly for C25 layout parity and not because final video bounds depend on it.

rakuco: I'd rather revert this part of the change to avoid confusion. The comment above this bit mentions how the size is actually set via `AVSBVideoRenderer::SetBounds()`.

abhijeetk: Done. Reverted to original code.

media/starboard/starboard_renderer.cc
abhijeetk: Filed new issue b/541996730 to track this requirement.

media/mojo/clients/starboard/starboard_renderer_client.cc
abhijeetk: @jasonzhangxx : I looked into this further, I think my previous comment was misleading. 

> @abhijeetk, can you help to confirm if url player and mojo bypass mechanisms mutually exclusive?

yes, URL player and mojo bypass are mutually exclusive. The bypass bridge passes demuxer streams to the starboard renderer in-process, skipping the Mojo serialization layer, while the URL player doesn't use demuxer streams, we delegates playback entirely to the native platform AVPlayer and doesn't use demuxer streams at all. 

The runtime guard `!is_url_player` in InitializeMojoRenderer ensures the bypass bridge is only used for non-URL-player sessions (e.g., non-HLS content on tvOS that goes through the regular demuxer pipeline can use mojo bypass).

media/starboard/starboard_renderer.cc
jasonzhangxx: In that case, we want to use `SB_DCHECK(!IsUrlPlayer());` to ensure the starboard implementation doesn't accidentally calling that function.

media/starboard/starboard_renderer.cc
```diff
@@ -690,11 +762,13 @@ void StarboardRenderer::CreatePlayerBridge() {
       break;
   }
 
-  if (audio_stream_) {
-    UpdateDecoderConfig(audio_stream_);
-  }
-  if (video_stream_) {
-    UpdateDecoderConfig(video_stream_);
+  if (!IsUrlPlayer()) {
```
jasonzhangxx: For UrlPlayer, |audio_stream_| and |video_stream_|will always be null, so the check here is redundant. We can add SB_DCHECK() to ensure that instead of defensive check.

media/starboard/starboard_renderer.cc
```diff
@@ -661,17 +731,19 @@ void StarboardRenderer::CreatePlayerBridge() {
...
+    // returns no audio configurations, update the write durations again
+    // before the SbPlayer reaches `kSbPlayerStatePresenting`.
+    audio_write_duration_for_preroll_ = audio_write_duration_ =
+        HasRemoteAudioOutputs(player_bridge_->GetAudioConfigurations())
+            ? audio_write_duration_remote_
+            : audio_write_duration_local_;
+    LOG(INFO) << "SbPlayerBridge created, with audio write duration at "
+              << audio_write_duration_for_preroll_
+              << " and with max_video_capabilities_ at "
+              << max_video_capabilities_;
```
jasonzhangxx: The code at lines 738-745 and 1071-1075 can be extracted into a new function `UpdateAudioWriteDuration()`. For tvOS, specifically when `IsUrlPlayer()` returns true, this function can return immediately to bypass the audio configuration update. This will help simplify the code with `IsUrlPlayer()`.

media/starboard/starboard_renderer.cc
```diff
@@ -582,7 +636,7 @@ void StarboardRenderer::CreatePlayerBridge() {
   DCHECK(task_runner_->RunsTasksInCurrentSequence());
   DCHECK(init_cb_);
   DCHECK_EQ(state_, STATE_INITIALIZING);
-  DCHECK(audio_stream_ || video_stream_);
+  DCHECK(audio_stream_ || video_stream_ || IsUrlPlayer());
```
jasonzhangxx: Can we make it to the following?
```
#if BUILDFLAG(IS_IOS_TVOS)
DCHECK(audio_stream_ || video_stream_ || (IsUrlPlayer() && !audio_stream_ && !video_stream_));
#else // BUILDFLAG(IS_IOS_TVOS)
DCHECK(audio_stream_ || video_stream_);
#endif // BUILDFLAG(IS_IOS_TVOS)
```

media/starboard/starboard_renderer.cc
```diff
@@ -622,27 +676,43 @@ void StarboardRenderer::CreatePlayerBridge() {
...
-      // TODO(b/326497953): Support suspend/resume.
-      // TODO(b/326508279): Support background mode.
-      sb_window_, drm_system_, this,
-      // TODO(b/326497953): Support suspend/resume.
-      false,
-      // TODO(b/326825450): Revisit 360 videos.
-      kSbPlayerOutputModeInvalid, max_video_capabilities_,
-      // TODO(b/326654546): Revisit HTMLVideoElement.setMaxVideoInputSize.
-      /*max_video_input_size=*/-1, experimental_features_
+  if (IsUrlPlayer()) {
```
jasonzhangxx: I can't find a good way to simplify the code here without introducing new functions. A simply way is
```
#if BUILDFLAG(IS_IOS_TVOS)
  if (IsUrlPlayer()) {
    ...
  }
  else {
#endif  // BUILDFLAG(IS_IOS_TVOS)
    ...
    ...
#if BUILDFLAG(IS_IOS_TVOS)
  }
#endif  // BUILDFLAG(IS_IOS_TVOS)

File: media/starboard/starboard_renderer.h (hunk 10 above)
Comment (jasonzhangxx) [reply media/starboard/starboard_renderer.h:162]: We should refactor some code to guard the implementation behind `BUILDFLAG(IS_IOS_TVOS)` more cleanly. I added some comments for this purpose.

File: media/starboard/url_player_demuxer.cc (hunk 14 above)
Comment (jasonzhangxx) [reply media/starboard/url_player_demuxer.cc:50]: Please add a comment to explicitly mention that for both `audio_decoder_config()` and `video_decoder_config()`.

File: media/starboard/starboard_renderer.cc (hunk 11 above)
Comment (jasonzhangxx) [reply media/starboard/starboard_renderer.cc:563]: Please add a TODO with the bug id.

Comment (abhijeetk) [reply media/starboard/starboard_renderer.cc:563]: Done.

File: media/starboard/starboard_renderer.h (hunk 10 above)
Comment (abhijeetk) [reply media/starboard/starboard_renderer.h:162]: Thanks. I have incorporated those comments.

File: media/starboard/url_player_demuxer.cc (hunk 14 above)
Comment (abhijeetk) [reply media/starboard/url_player_demuxer.cc:50]: Done.

File: media/starboard/starboard_renderer.cc (hunk 12 above)
Comment (abhijeetk) [reply media/starboard/starboard_renderer.cc]: Done.

File: media/starboard/starboard_renderer.cc (hunk 16 above)
Comment (abhijeetk) [reply media/starboard/starboard_renderer.cc]: Done.

File: media/starboard/starboard_renderer.cc (hunk 15 above)
Comment (abhijeetk) [reply media/starboard/starboard_renderer.cc]: Done.

File: media/starboard/starboard_renderer.cc (hunk 17 above)
Comment (abhijeetk) [reply media/starboard/starboard_renderer.cc]: Done.

File: media/starboard/starboard_renderer.cc (hunk 18 above)
Comment (abhijeetk) [reply media/starboard/starboard_renderer.cc]: Done.

File: media/starboard/starboard_renderer.cc [hunk 19]
```diff
@@ -622,27 +700,45 @@ void StarboardRenderer::CreatePlayerBridge() {
...
+            &StarboardRenderer::OnEncryptedMediaInitDataEncountered,
+            base::Unretained(this))
+#if BUILDFLAG(COBALT_MEDIA_ENABLE_CVAL)
+            ,
+        /*pipeline_identifier=*/""
+#endif  // BUILDFLAG(COBALT_MEDIA_ENABLE_CVAL)
+        ));
+  } else {
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+    LOG(INFO) << "Creating SbPlayerBridge.";
```
jasonzhangxx: nit: remove the log at line 718 and the following empty line.

media/starboard/starboard_renderer.cc
```diff
@@ -690,11 +776,21 @@ void StarboardRenderer::CreatePlayerBridge() {
...
-  if (audio_stream_) {
-    UpdateDecoderConfig(audio_stream_);
-  }
-  if (video_stream_) {
-    UpdateDecoderConfig(video_stream_);
+#if BUILDFLAG(IS_IOS_TVOS)
+  if (IsUrlPlayer()) {
+    // URL player does not use demuxer streams; they must be null.
+    SB_DCHECK(!audio_stream_);
+    SB_DCHECK(!video_stream_);
```
jasonzhangxx: The `SB_DCHECK` here is redundant. It's already checked at line 660. We only need 
```
    if (audio_stream_) {
      UpdateDecoderConfig(audio_stream_);
    }
    if (video_stream_) {
      UpdateDecoderConfig(video_stream_);
    }
```

media/starboard/starboard_renderer.cc
```diff
@@ -858,6 +954,13 @@ void StarboardRenderer::OnNeedData(DemuxerStream::Type type,
     return;
   }
 
+#if BUILDFLAG(IS_IOS_TVOS)
+  // URL player handles all buffering natively; OnNeedData should never
+  // be called because SbUrlPlayerCreate does not take a decoder-status
+  // callback. This is defensive guard.
+  SB_DCHECK(!IsUrlPlayer());
```
jasonzhangxx: Move the `SB_DCHECK` to the top of the function, right after line 947.

And I just realized that I made a mistake. We should use `DCHECK` instead of `SB_DCHECK`. Please help update the PR. Sorry for the misleading info.

media/starboard/starboard_renderer.cc
```diff
@@ -984,11 +1087,12 @@ void StarboardRenderer::OnPlayerStatus(SbPlayerState state) {
...
-          HasRemoteAudioOutputs(player_bridge_->GetAudioConfigurations())
-              ? audio_write_duration_remote_
-              : audio_write_duration_local_;
-      LOG(INFO) << "Audio write duration is " << audio_write_duration_;
+#if BUILDFLAG(IS_IOS_TVOS)
+      if (IsUrlPlayer()) {
+        OnUrlPlayerPresenting();
+      }
+#endif  // BUILDFLAG(IS_IOS_TVOS)
+      UpdateAudioWriteDuration();
```
jasonzhangxx: nit: move `UpdateAudioWriteDuration` to 1090, right before `IS_IOS_TVOS` check.

media/starboard/starboard_renderer.cc
abhijeetk: Done.

media/starboard/starboard_renderer.cc
abhijeetk: Done.

media/starboard/starboard_renderer.cc
abhijeetk: Done.

media/starboard/starboard_renderer.cc
abhijeetk: Done.

media/starboard/starboard_renderer.cc
jasonzhangxx: We don't need 
```
#if BUILDFLAG(IS_IOS_TVOS)
  if (!IsUrlPlayer()) {
#else
  {
#endif  // BUILDFLAG(IS_IOS_TVOS)
```
here, as if `IsUrlPlayer()` returns true, `audio_stream_` and `video_stream_` must be null.

abhijeetk: Agreed. I have implemented as per your feedback.

jasonzhangxx (approved):

PR #10699, opened 2026-05-29
Deprecate SbFileAtomicReplace

y4vor:
Moved the implementation in starboard/common as no no longer need a Starboard API as we implement it on top of POSIX.

Bug: 502580689


starboard/loader_app/installation_manager.cc
```diff
@@ -29,7 +29,6 @@
 #include "starboard/common/string.h"
 #include "starboard/configuration_constants.h"
 #include "starboard/extension/loader_app_metrics.h"
-#include "starboard/file.h"
 #include "starboard/loader_app/installation_store.pb.h"
```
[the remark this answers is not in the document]
y4vor: It is already included.

starboard/common/file.h
```diff
@@ -37,6 +37,15 @@ ssize_t ReadAll(int fd, void* data, int size);
...
 void RecordFileWriteStat(int write_file_result);
 
+// Replaces the content of the file at |path| with |data|. Returns whether the
+// contents of the file were replaced. The replacement of the content is an
+// atomic operation. The file will either have all of the data, or none.
+//
+// |path|: The path to the file whose contents should be replaced.
+// |data|: The data to replace the file contents with.
+// |data_size|: The amount of |data|, in bytes, to be written to the file.
+bool SbFileAtomicReplace(const char* path, const char* data, int64_t data_size);
```
andrewsavage1: Depending on how often is used, could we rename this? We can do it in a separate PR if you prefer. I think it would be better to avoid the Sb prefix for common functions like this to avoid confusion and to not cause tools like the api_leak_detector to break

andrewsavage1 (approved):

starboard/common/file.h
y4vor: Good point. I'll rename it.

PR #10705, opened 2026-05-29
Posix: register `alarm`

sacuff:
This plugs a symbol leak in nplb by registering alarm, which is a very simple function that does not require a wrapper.

Bug: 477257357


starboard/elf_loader/exported_symbols.cc
```diff
@@ -218,6 +218,7 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_SYMBOL(SbWindowSetDefaultOptions);
 
   // POSIX APIs
+  REGISTER_SYMBOL(alarm);
```
y4vor, line 219: Try asking Gemini to create some form of simple test that we can ad to starboard/nplb/posix_compliance/

jellefoks (approved):

y4vor (approved):

starboard/elf_loader/exported_symbols.cc
sacuff: Added

PR #10707, opened 2026-05-29
Parse the exported symbols list

y4vor:
Have single source of truth the actual symbol
registration code to avoid bugs with out of sync
lists for  the allowed symbols.

Adjust the manifest to add memfd_create which was not reported as the lists were out of sync.

Bug: 437417239


sacuff (approved):

sacuff:
Looks like `memfd_create` and `__clock_gettime64` are causing trouble here because they're handled outside of `exported_symbols.cc` at the moment. memfd_create is "defined" as `(errno = ENOSYS, -1)` in third_party/musl/src/starboard/include/syscall_arch.h:96 and `__clock_gettime64` should probably be added to dev_manifest

PR #10715, opened 2026-06-01
[TEST][IGNORE] This PR is so stacked it's got its own tower defense game

jellefoks:
I like stacked PRs because they make 5,000 lines of code look like 50 tiny, painless pull requests. It gives you 50 different opportunities to completely zone out and just write LGTM on every single one!


jellefoks:
Demo over

PR #10719, opened 2026-06-01
feat(metrics): Register PartitionAlloc committed size sub-metrics

Awallky:
Registers individual PartitionAlloc partition metrics (FastMalloc, Buffer,
ArrayBuffer, Layout) in the static `kAllocatorDumpNamesForMetrics` telemetry
registry in `cobalt_memory_metrics_emitter.cc`.

These metrics track the virtual committed size for each domain, enabling
detailed virtual fragmentation and footprint analysis. Because they are
registered as subpaths (e.g., `partition_alloc/partitions/fast_malloc`), they
safely bypass the Android/Linux skip-filter loop for standard `PartitionAlloc`
and will begin firing successfully.

Bug: 517340923


yell0wd0g (approved):
RS LGTM. Self-testing CLs are the best CLs!

PR #10726, opened 2026-06-02
[Cobalt] Add dynamic GPU image decode cache optimization switches

sherryzy:
Introduce dynamic, runtime-configurable switches to optimize Cobalt's GPU image decode cache and working set memory footprint:
1. cc-image-cache-limit-items: Dynamically limits historical off-screen image cache retention (supports Zero-Cache policy).
2. decoded-image-working-set-budget-bytes: Customizes active image decode working set budget dynamically.
3. enable-scaling-clipped-images: Exposes pre-existing scale down capabilities on clipped images to the Java experiment switches layer.

All C++ changes are strictly gated under BUILDFLAG(IS_COBALT).

Bug: 512157288

### DIFF
--- a/cc/base/switches.cc
+++ b/cc/base/switches.cc
@@ -5,6 +5,7 @@
 #include "cc/base/switches.h"
 
 #include "base/command_line.h"
+#include "build/build_config.h"
 
 namespace switches {
 
@@ -111,5 +112,11 @@ const char kCCLayerTreeTestLongTimeout[] = "cc-layer-tree-test-long-timeout";
 // Controls the duration of the scroll animation curve.
 const char kCCScrollAnimationDurationForTesting[] =
     "cc-scroll-animation-duration-in-seconds";
+
+#if BUILDFLAG(IS_COBALT)
+const char kCCImageCacheLimitItems[] = "cc-image-cache-limit-items";
+const char kDecodedImageWorkingSetBudgetBytes[] =
+    "decoded-image-working-set-budget-bytes";
+#endif
 
 }  // namespace switches
--- a/cc/tiles/gpu_image_decode_cache.cc
+++ b/cc/tiles/gpu_image_decode_cache.cc
@@ -26,6 +26,9 @@
 #include "base/notreached.h"
 #include "base/numerics/safe_math.h"
 #include "base/strings/stringprintf.h"
+#if BUILDFLAG(IS_COBALT)
+#include "base/strings/string_number_conversions.h"
+#endif
 #include "base/synchronization/lock.h"
 #include "base/task/single_thread_task_runner.h"
 #include "base/time/tick_clock.h"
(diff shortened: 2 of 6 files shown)


cc/base/switches.cc
```diff
@@ -5,14 +5,14 @@
...
+#include "build/build_config.h"
 
 namespace switches {
 
 const char kDisableThreadedAnimation[] = "disable-threaded-animation";
 
 // Disables layer-edge anti-aliasing in the compositor.
-const char kDisableCompositedAntialiasing[] =
-    "disable-composited-antialiasing";
+const char kDisableCompositedAntialiasing[] = "disable-composited-antialiasing";
```
yell0wd0g: If this looks so upstream I'd say leave it (and somehow
convince the PR presubmit that it's fine).

cc/tiles/gpu_image_decode_cache.cc
```diff
@@ -2356,7 +2359,24 @@ bool GpuImageDecodeCache::ExceedsCacheLimits() const {
   if (aggressively_freeing_resources_) {
     items_limit = kSuspendedMaxItemsInCacheForGpu;
   } else {
+#if BUILDFLAG(IS_COBALT)
+    static const int normal_max_items = []() {
+      int limit = kNormalMaxItemsInCacheForGpu;
+      auto* command_line = base::CommandLine::ForCurrentProcess();
+      if (command_line->HasSwitch(switches::kCCImageCacheLimitItems)) {
```
yell0wd0g, line 2366: Hmm I fail to see how the lambda helps here. IOW what's wrong with
```
    items_limit = kNormalMaxItemsInCacheForGpu;  // Upstream.

#if BUILDFLAG(IS_COBALT)
    auto* command_line = base::CommandLine::ForCurrentProcess();
    if (command_line->HasSwitch(switches::kCCImageCacheLimitItems)) {
      std::string value = command_line->GetSwitchValueASCII(
          switches::kCCImageCacheLimitItems);
      int parsed_value;
      if (base::StringToInt(value, &parsed_value) && parsed_value >= 0) {
        items_limit = parsed_value;
      }
#endif
```
?

Same idea in cc/tiles/image_decode_cache_utils.cc.

yell0wd0g (approved):
LGTM with a suggestion to not use the lambda IIUC 
the intention.

cc/tiles/gpu_image_decode_cache.cc
sherryzy: I'm trying to address GEMINI's feedback regarding to performance. Why the Lambda Version is Used:
In C++, a static variable is initialized only once (the very first time the function is called). On all subsequent calls, the initialization is skipped completely, and the cached value is instantly returned from memory.
However, because computing the cache limit requires a multi-line block of logic, we can't assign it in a single line.

PR #10736, opened 2026-06-02
net: Re-add selective HTTP cache disablement

zhongqiliang:
restricts the HTTP cache to only store HTML and JavaScript/ECMAScript resources.
Bug: 514683173


net/http/http_cache_transaction.cc
```diff
@@ -3988,6 +3990,23 @@ bool HttpCache::Transaction::UpdateAndReportCacheability(
     return true;
   }
 
+#if BUILDFLAG(IS_COBALT)
+  // If disable-http-cache is set, only allow HTML and JS/ECMAScript
+  if (base::CommandLine::ForCurrentProcess()->HasSwitch("disable-http-cache")) {
```
johnxwork, line 3995: The switch should probably be called something like "http-cache-js-html-only". But hopefully we can deprecate this switch in 2 weeks.

johnxwork (approved):

net/http/http_cache_transaction.cc
zhongqiliang: All the switches from Kimono down to Cobalt are already in place.
https://source.corp.google.com/piper///depot/google3/java/com/google/android/apps/youtube/tv/activity/MainActivity.java;l=190-192?q=disableHttpCache
Can we use this name for now?

zhongqiliang: If the exp is good, we will remove the name anyways.

PR #10743, opened 2026-06-02
Posix: Add wrappers for `_Exit`

sacuff:
`_Exit`, not to be confused with the existing `exit`, is leaking from NPLB because it is a used by gmock. This adds the `_Exit` symbol by implementing a simple wrapper that ensures the exit status retains its meaning across platforms.

Bug: 477257357

### DIFF
--- a/starboard/shared/modular/starboard_layer_posix__Exit_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix__Exit_abi_wrappers.cc
@@ -1,0 +1,41 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/shared/modular/starboard_layer_posix__Exit_abi_wrappers.h"
+
+#include <stdlib.h>
+
+#include "starboard/configuration.h"
+#include "starboard/export.h"
+
+namespace {
+int musl_status_to_platform_status(int status) {
+  if (status == MUSL_EXIT_SUCCESS) {
+    return EXIT_SUCCESS;
+  }
+  if (status == MUSL_EXIT_FAILURE) {
+    return EXIT_FAILURE;
+  }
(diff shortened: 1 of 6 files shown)


starboard/shared/modular/starboard_layer_posix__Exit_abi_wrappers.cc
```diff
@@ -0,0 +1,41 @@
...
+  // like 2, but we know that the incoming value represents a failure.
+  if (status == EXIT_SUCCESS) {
+    return EXIT_FAILURE;
+  }
+  return status;
+}
+}  // namespace
+
+SB_EXPORT SB_NORETURN void __abi_wrap__Exit(int status) {
+  _Exit(musl_status_to_platform_status(status));
```
jellefoks, line 40: If `_Exit` a POSIX functions? What usually provides it, what does it do?

Oh, answered myself https://man7.org/linux/man-pages/man2/_exit.2.html.  

It's POSIX 2024. https://pubs.opengroup.org/onlinepubs/9799919799/functions/_exit.html

jellefoks: The POSIX doc says 'Normally applications should use exit() rather than _Exit() or _exit().' though.

Do we really need this?

sacuff: gmock uses it here and calls out that they can't rely on normal `exit()` https://github.com/youtube/cobalt/blob/main/third_party/googletest/src/googlemock/src/gmock-spec-builders.cc#L534

jellefoks: It seems that the difference between exit() and _exit() only matter for multiprocess. Hmm. We also don't have 'exit()'... but we also have platforms where we can't actually exit...

jellefoks: Can we delete the usage of _Exit from out gtest, since it's likely only used for process-spawning tests like death tests, which we don't (yet) support because we also don't support fork()?

sacuff: I think we have to leave this one. MockObjectRegistry is a global and isn't specific to death tests, and this is in the destructor so it's being called during framework teardown. `exit()` would try to call the destructor again.

Or we could swallow the error here but that might mask some test failures.

y4vor: I think we should just add it. We have other APIs that assume Linux kernel at this point. Platforms that can't exit e.g. some Game consoles probably wouldn't use Starboard or they can emulate it with a forced crash.

y4vor (commented):
LGTM!

andrewsavage1 (approved):
Starboard reviewer RS

y4vor (approved):

PR #10745, opened 2026-06-03
android: Fix Java heap leak in AudioTrackBridge JNI

borongc:
Frequent recreation of AudioTrackBridge during loop playback (e.g. YouTube Shorts)
was causing a Java heap memory leak and eventual OutOfMemoryError.

The leak occurred because the raw JNI local references returned by
`env->NewFloatArray` and `env->NewByteArray` in `AudioTrackBridge::Create`
were passed directly to `ScopedJavaGlobalRef::Reset` without being wrapped in
`ScopedJavaLocalRef`. Because the player worker thread is persistent and reused
across loops, these local references accumulated over time, keeping the
associated Java arrays (512KB each) alive on the Java heap.

This change fixes the leak by directly assigning temporary `ScopedJavaLocalRef`
wrappers (holding the newly allocated JNI arrays) to the `ScopedJavaGlobalRef`
`j_audio_data_`. This ensures that the local reference is automatically and
explicitly deleted when the temporary wrapper goes out of scope at the end of
the assignment statement, while the global reference is correctly initialized.

Issue: 517536842

### DIFF
--- a/starboard/android/shared/audio_track_bridge.cc
+++ b/starboard/android/shared/audio_track_bridge.cc
@@ -72,17 +72,18 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
   if (coding_type != kSbMediaAudioCodingTypePcm) {
     SB_DCHECK(!sample_type);
     max_samples_per_write = kMaxFramesPerRequest;
-    j_audio_data.Reset(env, env->NewByteArray(max_samples_per_write));
+    j_audio_data = ScopedJavaLocalRef<jbyteArray>(
+        env, env->NewByteArray(max_samples_per_write));
   } else if (sample_type == kSbMediaAudioSampleTypeFloat32) {
     max_samples_per_write = channels * kMaxFramesPerRequest;
-    j_audio_data.Reset(env,
-                       env->NewFloatArray(channels * kMaxFramesPerRequest));
+    j_audio_data = ScopedJavaLocalRef<jfloatArray>(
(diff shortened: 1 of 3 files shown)


Awallky (approved):
Changes look straightforward.
LGTM

johnxwork (approved):

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -72,17 +72,21 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
   if (coding_type != kSbMediaAudioCodingTypePcm) {
     SB_DCHECK(!sample_type);
     max_samples_per_write = kMaxFramesPerRequest;
-    j_audio_data.Reset(env, env->NewByteArray(max_samples_per_write));
+    ScopedJavaLocalRef<jbyteArray> j_buffer(
```
kjyoun: thanks for the finding.

nit: we dont need `j_buffer` here. we can directly assign `ScopedJavaLocalRef` to `j_audio_data`,
since `ScopedJavaGlobalRef<jobject>` has == operator that can accept `ScopedJavaLocalRef<jobject>`

````
j_audio_data = ScopedJavaLocalRef<jbyteArray>(
        env, env->NewByteArray(max_samples_per_write));
```

kjyoun (approved):

starboard/android/shared/audio_track_bridge.cc
borongc: Done.

PR #10752, opened 2026-06-03
Restrict delete lock file to Cobalt Android

zhongqiliang:
Cobalt on 3p can run 2 instances at the same time, delete the lock file on it can cause bugs.
Bug: 488465561

### DIFF
--- a/components/services/storage/dom_storage/local_storage_impl.cc
+++ b/components/services/storage/dom_storage/local_storage_impl.cc
@@ -651,7 +651,7 @@ void LocalStorageImpl::InitiateConnection(bool in_memory_only) {
     // We were given a subdirectory to write to, so use a disk-backed database.
     in_memory_ = false;
 
-#if BUILDFLAG(IS_COBALT)
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROID)
     if (base::FeatureList::IsEnabled(kLocalStorageDeleteLockFile)) {
       base::FilePath db_path = directory_.AppendASCII(kLocalStorageLeveldbName);
       base::FilePath lock_file_path = db_path.AppendASCII("LOCK");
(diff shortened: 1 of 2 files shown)


components/services/storage/dom_storage/local_storage_impl.cc
```diff
@@ -651,15 +651,15 @@ void LocalStorageImpl::InitiateConnection(bool in_memory_only) {
     // We were given a subdirectory to write to, so use a disk-backed database.
     in_memory_ = false;
 
-#if BUILDFLAG(IS_COBALT)
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_ANDROID)
     if (base::FeatureList::IsEnabled(kLocalStorageDeleteLockFile)) {
```
yell0wd0g, line 655: Wouldn't it be enough if we just set this
to disabled? By default is enabled.

zhongqiliang: We can, but I 'd want to avoid the flag to be used on 3p by mistake.

yell0wd0g: We can add the flag to the Java file only to avoid 3P
accidentally getting it, correct?

zhongqiliang: The command line feature string is defined in common. 

I can also make this BUILDFLAG change when I remove "base::FeatureList::IsEnabled(kLocalStorageDeleteLockFile)" here after the experiment.

yell0wd0g: I'm confused. What I meant is adding 

```
        paramOverrides.add("LocalStorageDeleteLockFile");
```

to //cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java, 
`getDefaultEnableFeatureOverridesList()`, [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java;l=108?q=command%20line%20helper%20&ss=h%2Flbshell-internal%2Fcobalt_src%2F%2B%2Frefs%2Fheads%2Fmain:cobalt%2F).

This should only affect Java builds, i.e. AndroidTV, and
would accomplish the same effect as this BUILDFLAG-based 
code, only not modifying this file. Is that correct?

zhongqiliang: Yes for now. After the experiment is done, and this change works, I will remove "kLocalStorageDeleteLockFile" and "LocalStorageDeleteLockFile", and make this code the default behavior for Cobalt Android.

PR #10765, opened 2026-06-04
android: Fix black screen

borongc:
Issue: 494590075


borongc:
This is handled by https://github.com/youtube/cobalt/pull/11538.

PR #10767, opened 2026-06-04
Refactor on_host_tests to use dynamically passed test executables

arjungm:
- Extend `test_targets_json` action to output `test_executables_json`.
- Modify `on_host_tests` to accept `test_executables_json` instead of hardcoded parsing.
- Update `main.yaml` to pipe the output correctly.

Bug: 431780245

### DIFF
--- a/.github/actions/on_host_tests/action.yaml
+++ b/.github/actions/on_host_tests/action.yaml
@@ -70,11 +70,10 @@ runs:
         cd unit_test/
 
         # Parse the JSON array into a space-separated string using jq.
-        # This assumes executable paths will never contain spaces.
         test_executables=$(jq -nr --arg json "${TEST_EXECUTABLES_JSON:-[]}" '$json | fromjson | join(" ")')
 
         if [[ -z "${test_executables}" ]]; then
-          echo "No tests to execute. Exiting cleanly."
+          echo "No tests to execute. Skipping."
           exit 0
         fi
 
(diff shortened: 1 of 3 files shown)


.github/actions/on_host_tests/action.yaml
```diff
@@ -65,33 +69,39 @@ runs:
...
-          filename=$(basename "${test_binary_path}")
-          test_binary="${filename%%.*}"
+        # Parse the JSON array into a space-separated string using jq.
+        # This assumes executable paths will never contain spaces.
+        test_executables=$(jq -nr --arg json "${TEST_EXECUTABLES_JSON:-[]}" '$json | fromjson | join(" ")')
+
+        if [[ -z "${test_executables}" ]]; then
+          echo "No tests to execute. Exiting cleanly."
+          exit 0
+        fi
```
oxve: I don't think this is necessary. If there's nothing to for loop over below it will exit cleanly and with `set -x` we can see the value of the calculated array.

.github/actions/test_targets_json/action.yaml
```diff
@@ -35,8 +35,22 @@ runs:
...
+          --arg test_type "${TEST_TYPE}" \
+          --arg runs_on "${RUNS_ON}" \
+          '[
+            .[]
+            | select(
+                .executable != null and
+                ($test_type == "" or .test_type == $test_type) and
+                ($runs_on == "" or .runs_on == $runs_on)
+              )
+            | .executable
```
oxve: Can you break out this part here and above?

Select the whole object in the first step, something like:
```
select(
  ($test_type == "" or .test_type == $test_type) and
  ($runs_on == "" or .runs_on == $runs_on)
) | .
```

oxve (approved):

PR #10773, opened 2026-06-04
media: Add configurable Video Buffer Pool size for 720p

johnedocampo:
Introduce the Media.Configurable720pMediaBudgetMb H5vcc setting to
control the video buffer memory limit for resolutions up to 720p.

This change allows for runtime tuning of the demuxer memory limit to
optimize playback performance and memory footprint specifically for
720p video content. The limit applies when the video resolution is
1280x720 or lower.

Bug: 416105795

### DIFF
--- a/media/base/starboard/demuxer_memory_limit_starboard.cc
+++ b/media/base/starboard/demuxer_memory_limit_starboard.cc
@@ -30,8 +30,11 @@
 namespace media {
 namespace {
 
-// |g_video_buffer_size_clamp_bytes| is process-wide, and is currently set by
-// a h5vcc flag.
+// |g_720p_video_buffer_size_clamp_bytes| and
+// |g_video_buffer_size_clamp_bytes| are process-wide, and are currently set by
+// h5vcc flags.
+std::atomic<size_t> g_720p_video_buffer_size_clamp_bytes{
+    std::numeric_limits<size_t>::max()};
 std::atomic<size_t> g_video_buffer_size_clamp_bytes{
     std::numeric_limits<size_t>::max()};
 
@@ -62,6 +65,13 @@ int GetBitsPerPixel(const VideoDecoderConfig& video_config) {
 
 }  // namespace
 
+void Set720pVideoBufferSizeClamp(int size_mb) {
+  CHECK_GE(size_mb, 0);
+  CHECK_LT(size_mb, 4096);
+  g_720p_video_buffer_size_clamp_bytes =
+      static_cast<size_t>(size_mb) * 1024 * 1024;
+}
+
 size_t GetVideoBufferSizeClamp() {
   return g_video_buffer_size_clamp_bytes.load();
 }
@@ -83,6 +93,14 @@ size_t GetDemuxerStreamAudioMemoryLimit(
 size_t GetDemuxerStreamVideoMemoryLimit(
     DemuxerType /*demuxer_type*/,
     const VideoDecoderConfig* video_config) {
+  if (const size_t limit_720p = g_720p_video_buffer_size_clamp_bytes.load();
+      limit_720p != std::numeric_limits<size_t>::max() && video_config) {
+    const gfx::Size resolution = video_config->visible_rect().size();
+    if (resolution.width() <= 1280 && resolution.height() <= 720) {
+      return limit_720p;
+    }
+  }
+
   size_t limit;
   if (!video_config) {
     limit = GetVideoDecoderBufferLimitBytes(
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
@@ -214,6 +214,13 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
                                     return base::ok();
                                   });
   }
+  if (name == "Media.720pVideoBufferSizeClampMb") {
+    return ProcessSettingAsPositiveInt(
+        script_state, exception_context, name, *value, [](int int_value) {
+          ::media::Set720pVideoBufferSizeClamp(int_value);
+          return true;
+        });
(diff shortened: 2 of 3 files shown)


third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -214,6 +214,13 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
                                     return base::ok();
                                   });
   }
+  if (name == "Media.Configurable720pMediaBudgetMb") {
+    return ProcessSettingAsPositiveInt(
+        script_state, exception_context, name, *value, [](int int_value) {
+          ::media::SetConfigurable720pMediaBudget(int_value);
+          return true;
+        });
+  }
```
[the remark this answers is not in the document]
johnedocampo: This should be okay as we are controlling the value for this.

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -30,8 +30,11 @@
 namespace media {
 namespace {
 
-// |g_video_buffer_size_clamp_bytes| is process-wide, and is currently set by
-// a h5vcc flag.
+// |g_video_buffer_size_clamp_bytes| and
+// |g_configurable_720p_media_budget_bytes| are process-wide, and are currently
+// set by h5vcc flags.
+std::atomic<size_t> g_configurable_720p_media_budget_bytes{
```
kjyoun: what does "confiigurable" mean in this name?

other variable (e.g. g_video_buffer_size_clamp_bytes) isn't configurable?

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -83,6 +93,16 @@ size_t GetDemuxerStreamAudioMemoryLimit(
...
     DemuxerType /*demuxer_type*/,
     const VideoDecoderConfig* video_config) {
+  const size_t override_limit_720p =
+      g_configurable_720p_media_budget_bytes.load();
+  if (override_limit_720p != std::numeric_limits<size_t>::max() &&
+      video_config) {
+    const gfx::Size resolution = video_config->visible_rect().size();
+    if (resolution.width() <= 1280 && resolution.height() <= 720) {
+      return override_limit_720p;
+    }
```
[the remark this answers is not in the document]
johnedocampo: This change would change the strategy being done in https://github.com/youtube/cobalt/blob/main/starboard/android/shared/media_get_video_buffer_budget.cc. Since we want to preserve the strategy and just only override the value that would have returned from that function, we will keep this as-is for now.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
```diff
@@ -214,6 +214,13 @@ ScriptPromise<IDLUndefined> H5vccSettings::set(
                                     return base::ok();
                                   });
   }
+  if (name == "Media.Configurable720pMediaBudgetMb") {
```
kjyoun: I think "Configurable720pMediaBudgetMb" sounds little bit weird. since every flag is basically configurable.

How about making this name consistent with similar flag( VideoBufferSizeClampMb) ?

media/base/starboard/demuxer_memory_limit_starboard.cc
```diff
@@ -83,6 +93,16 @@ size_t GetDemuxerStreamAudioMemoryLimit(
 size_t GetDemuxerStreamVideoMemoryLimit(
     DemuxerType /*demuxer_type*/,
     const VideoDecoderConfig* video_config) {
+  const size_t override_limit_720p =
```
kjyoun: nit: how about this? since override_limit_720p is used only inside if block


```
if (const size_t override_limit_720p =   g_configurable_720p_media_budget_bytes.load();
    override_limit_720p != std::numeric_limits<size_t>::max() &&
      video_config) {

```

osagie98:
Is it possible with this change to set the buffer pool size as low as 1 MB? We [require](https://developers.google.com/youtube/devices/living-room/requirements/2027/hardware-certification#media-decoding) at least 30 MB reserved for video buffers, could it affect that?

media/base/starboard/demuxer_memory_limit_starboard.cc
kjyoun: nit:

`override_limit_720p` do you need "override_" in variable name?

we don't have "override_" in similar override logic in line 108.

osagie98 (approved):

osagie98:
> Is it possible with this change to set the buffer pool size as low as 1 MB? We [require](https://developers.google.com/youtube/devices/living-room/requirements/2027/hardware-certification#media-decoding) at least 30 MB reserved for video buffers, could it affect that?

Discussed offline, LGTM

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Valid point.

Changed name to g_720p_video_buffer_size_clamp_bytes.

third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc
johnedocampo: Agreed.

Changed name to g_720p_video_buffer_size_clamp_bytes.

media/base/starboard/demuxer_memory_limit_starboard.cc
johnedocampo: Done; also changed `override_limit_720p` to `limit_720p`.

kjyoun (approved):

y4vor (approved):

sideb0ard (approved):

PR #10775, opened 2026-06-04
starboard: Clarify experimental extension usage

xiaomings:
Update the README in the experimental extensions directory to clarify
that these features are for internal use by maintainers only. This
prevents third-party developers from relying on these interfaces,
which are unstable and subject to change or removal at any time.

Bug: 520023679


xiaomings:
@kjyoun FYI

PR #10779, opened 2026-06-05
Rename disable_privacy_sandbox_apis to enable_privacy_sandbox_apis

linxinan-yt:
Following Cobalt's convention ,rename DISABLE_PRIVACY_SANDBOX_APIS to ENABLE_PRIVACY_SANDBOX_APIS for readability. The PR did not change the default behavior, therefore no functional change expected.

Bug: 505811196

### DIFF
--- a/content/browser/BUILD.gn
+++ b/content/browser/BUILD.gn
@@ -2519,7 +2519,7 @@ source_set("browser") {
     "worker_host/worker_script_loader_factory.h",
   ]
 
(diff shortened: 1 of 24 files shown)


yell0wd0g (approved):
This is purely mechanical, correct? If so, LGTM.
Thanks for following up!

linxinan-yt:
> This is purely mechanical, correct? If so, LGTM. Thanks for following up!

Yes, only mechanical change for readability.

PR #10780, opened 2026-06-05
android/jni-zero: Move AudioTrackBridge array allocation to Java

kjyoun:
Move the allocation of audio data buffers from C++ JNI calls to the
Java AudioTrackBridge constructor.

This change follows JNI best practices by avoiding direct calls to
memory allocation methods in the JNI environment, which reduces the
risk of memory leaks from un-wrapped local references. The Java side
now manages the buffer lifecycle, and the C++ layer retrieves the
reference during initialization.

Issue: 520007952

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
@@ -292,8 +292,19 @@ public class AudioTrackBridge {
   }
 
   @CalledByNative
-  public Object getPreAllocatedAudioData() {
-    return mPreAllocatedAudioData;
+  public float[] getPreAllocatedFloatArray() {
+    if (mPreAllocatedAudioData instanceof float[]) {
+      return (float[]) mPreAllocatedAudioData;
+    }
+    return null;
+  }
+
+  @CalledByNative
+  public byte[] getPreAllocatedByteArray() {
+    if (mPreAllocatedAudioData instanceof byte[]) {
+      return (byte[]) mPreAllocatedAudioData;
+    }
+    return null;
   }
 
   // TODO (b/262608024): Have this method return a boolean and return false on failure.
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
@@ -59,6 +59,11 @@ public class AudioTrackBridge {
   private ByteBuffer mAvSyncHeader;
   private int mAvSyncPacketBytesRemaining;
 
+  // Pre-allocated byte[] or float[] shared with C++ via getPreAllocatedAudioData() to avoid
+  // allocation on every WriteSample(). C++ writes directly to this array,
+  // which is then passed back to Java write*() methods.
+  private Object mPreAllocatedAudioData;
+
   private static int getBytesPerSample(int audioFormat) {
     switch (audioFormat) {
       case AudioFormat.ENCODING_PCM_16BIT:
@@ -71,11 +76,27 @@ public class AudioTrackBridge {
     }
   }
 
+  private static Object createAudioDataArray(int sampleType, int maxSamplesPerWrite) {
+    switch (sampleType) {
+      case AudioFormat.ENCODING_PCM_FLOAT:
+        return new float[maxSamplesPerWrite];
+      case AudioFormat.ENCODING_PCM_16BIT:
+        return new byte[maxSamplesPerWrite * getBytesPerSample(sampleType)];
+      case AudioFormat.ENCODING_AC3:
+      case AudioFormat.ENCODING_E_AC3:
+        return new byte[maxSamplesPerWrite];
+      default:
+        Log.e(TAG, "Unsupported sample type: " + sampleType);
+        return null;
+    }
+  }
+
   // TODO: Pass error details to caller.
   public AudioTrackBridge(
       int sampleType,
       int sampleRate,
       int channelCount,
+      int maxSamplesPerWrite,
       int preferredBufferSizeInBytes,
       int tunnelModeAudioSessionId,
       boolean isWebAudio) {
@@ -204,6 +225,8 @@ public class AudioTrackBridge {
           Log.i(TAG, String.format(Locale.US, "Unknown AudioFormat %d.", sampleType));
           break;
       }
+
+      mPreAllocatedAudioData = createAudioDataArray(sampleType, maxSamplesPerWrite);
     }
     Log.i(
         TAG,
@@ -227,7 +250,6 @@ public class AudioTrackBridge {
     mAvSyncHeader = null;
     mAvSyncPacketBytesRemaining = 0;
   }
-
 
   @CalledByNative
   public boolean setPlaybackRate(float playbackRate) {
@@ -267,6 +289,11 @@ public class AudioTrackBridge {
       return AudioTrack.PLAYSTATE_STOPPED;
     }
     return mAudioTrack.getPlayState();
+  }
+
+  @CalledByNative
+  public Object getPreAllocatedAudioData() {
+    return mPreAllocatedAudioData;
   }
 
   // TODO (b/262608024): Have this method return a boolean and return false on failure.
--- a/starboard/android/shared/audio_track_bridge.cc
+++ b/starboard/android/shared/audio_track_bridge.cc
@@ -39,6 +39,22 @@ using jni_zero::ScopedJavaLocalRef;
 
 const jint kNoOffset = 0;
 
+void SetFloatArrayRegion(JNIEnv* env,
+                         const jni_zero::JavaRef<jfloatArray>& array,
+                         jsize start,
+                         jsize len,
+                         const jfloat* buf) {
+  env->SetFloatArrayRegion(array.obj(), start, len, buf);
+}
+
+void SetByteArrayRegion(JNIEnv* env,
+                        const jni_zero::JavaRef<jbyteArray>& array,
+                        jsize start,
+                        jsize len,
+                        const jbyte* buf) {
+  env->SetByteArrayRegion(array.obj(), start, len, buf);
+}
+
 }  // namespace
 
 std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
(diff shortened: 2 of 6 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -213,12 +219,26 @@ public AudioTrackBridge(
...
+      if (sampleType == AudioFormat.ENCODING_PCM_FLOAT) {
+        mPreAllocatedAudioData = new float[maxSamplesPerWrite];
+      } else if (sampleType == AudioFormat.ENCODING_PCM_16BIT) {
+        mPreAllocatedAudioData = new byte[maxSamplesPerWrite * getBytesPerSample(sampleType)];
+      } else if (sampleType == AudioFormat.ENCODING_AC3
+          || sampleType == AudioFormat.ENCODING_E_AC3) {
+        mPreAllocatedAudioData = new byte[maxSamplesPerWrite];
+      } else {
+        throw new RuntimeException("Unsupported sample type: " + sampleType);
+      }
```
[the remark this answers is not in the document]
kjyoun: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -228,7 +251,6 @@ public void release() {
     mAvSyncPacketBytesRemaining = 0;
   }
 
-
```
haozheng-cobalt, line 231: nit, revert the change here

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -111,16 +94,33 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
...
+      Java_AudioTrackBridge_getPreAllocatedAudioData(env, j_audio_track_bridge);
+  if (j_audio_data.is_null()) {
+    SB_LOG(WARNING) << "Failed to retrieve |j_audio_data_|";
+    return nullptr;
+  }
+
+  const auto j_audio_data_var = [&]() -> DataArray {
+    if (coding_type == kSbMediaAudioCodingTypePcm &&
+        sample_type == kSbMediaAudioSampleTypeFloat32) {
+      SB_CHECK(env->IsInstanceOf(j_audio_data.obj(), env->FindClass("[F")));
```
haozheng-cobalt: env->FindClass returns a raw JNI local reference (jclass). Because this local reference is neither wrapped in ScopedJavaLocalRef nor manually released via env->DeleteLocalRef, it leaks a class object reference in the local JNI table every time an audio track is initialized.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -111,16 +94,33 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
...
+      return FloatArray(env, static_cast<jfloatArray>(j_audio_data.obj()));
+    }
+    SB_CHECK(env->IsInstanceOf(j_audio_data.obj(), env->FindClass("[B")));
```
haozheng-cobalt: Same as above.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -65,39 +65,22 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
...
   ScopedJavaLocalRef<jobject> j_audio_track_bridge =
       AudioOutputManager::GetInstance()->CreateAudioTrackBridge(
           env, GetAudioFormatSampleType(coding_type, sample_type),
-          sampling_frequency_hz, channels, preferred_buffer_size_in_bytes,
-          tunnel_mode_audio_session_id, is_web_audio);
+          sampling_frequency_hz, channels, max_samples_per_write,
+          preferred_buffer_size_in_bytes, tunnel_mode_audio_session_id,
+          is_web_audio);
 
   if (j_audio_track_bridge.is_null()) {
```
haozheng-cobalt, line 101: Gemini suggested:

The Bug: If the Java constructor throws an exception, the JVM sets a pending JNI exception.
When control returns to C++, the check if (j_audio_track_bridge.is_null()) succeeds and C++ returns nullptr. However, the code does not call env->ExceptionClear().
Leaving a pending Java exception on the JNI thread causes the Android runtime to crash the application on the very next JNI call made by Cobalt.
The Fix: The native factory method AudioTrackBridge::Create should clear pending exceptions if creation fails:
cpp
if (env->ExceptionCheck()) {
  env->ExceptionDescribe(); // Print details to logcat
  env->ExceptionClear();    // Safely clear the exception
}

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
kjyoun: Done

starboard/android/shared/audio_track_bridge.cc
kjyoun: Great catch.

It was from existing code, but is not trigged on prod build, since it is wrapped in SB_DCHECK

Handled it

starboard/android/shared/audio_track_bridge.cc
kjyoun: Done

starboard/android/shared/audio_track_bridge.cc
kjyoun: Since we use  jni_zero , this C++ fix is unreachable, so that it is not necesary.
 
 jni_zero 's generated wrappers automatically check for exceptions and invoke Cobalt's global exception handler (which calls  `LOG(FATAL)` and crashes the app) before returning to C++. If Java throws, we crash instantly inside the JNI wrapper, so we never reach the C++ ExceptionClear()  check.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -234,19 +241,14 @@ int AudioTrackBridge::WriteSample(const uint8_t* samples,
 
   num_of_samples = std::min(num_of_samples, max_samples_per_write_);
 
-  SB_DCHECK(env->IsInstanceOf(j_audio_data_.obj(), env->FindClass("[B")));
-  env->SetByteArrayRegion(static_cast<jbyteArray>(j_audio_data_.obj()),
-                          kNoOffset, num_of_samples,
-                          reinterpret_cast<const jbyte*>(samples));
+  SB_CHECK(std::holds_alternative<ByteArray>(j_audio_data_));
```
haozheng-cobalt, line 244: Stick to SB_DCHECK instead, because SB_DCHECK will crash the app in non-official build but not in official builds, while SB_CHECK will crash the app in official and non-official builds.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -202,19 +212,16 @@ int AudioTrackBridge::WriteSample(const uint16_t* samples,
 
   num_of_samples = std::min(num_of_samples, max_samples_per_write_);
 
-  SB_DCHECK(env->IsInstanceOf(j_audio_data_.obj(), env->FindClass("[B")));
-  env->SetByteArrayRegion(static_cast<jbyteArray>(j_audio_data_.obj()),
-                          kNoOffset, num_of_samples * sizeof(uint16_t),
-                          reinterpret_cast<const jbyte*>(samples));
+  SB_CHECK(std::holds_alternative<ByteArray>(j_audio_data_));
```
haozheng-cobalt, line 215: Same here, revert all SB_CHECK call back to SB_DCHECK

haozheng-cobalt (approved):
LGTM except the SB_CHECK

starboard/android/shared/audio_track_bridge.cc
kjyoun: It is intentional. I want to crash the app—even in production—when a key invariant is broken.

If this invariant does not hold, it means something very serious (such as memory corruption) has occurred. In that case, it is much better to crash immediately here, which provides clear information on what went wrong. Otherwise, we end up with a completely incomprehensible crash stack later down the line.

This also complies with the Chromium style guideline:

[go/cpplabs/avoid-dcheck.md](https://goto.google.com/cpplabs/avoid-dcheck.md)

https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++.md#check_dcheck_and-notreached

> Historically, Chromium code used DCHECK() in most cases, so a great deal of existing code uses DCHECK() instead of CHECK(). You are encouraged to migrate to CHECK()

starboard/android/shared/audio_track_bridge.cc
kjyoun: commented above

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -71,11 +76,27 @@ private static int getBytesPerSample(int audioFormat) {
...
+    switch (sampleType) {
+      case AudioFormat.ENCODING_PCM_FLOAT:
+        return new float[maxSamplesPerWrite];
+      case AudioFormat.ENCODING_PCM_16BIT:
+        return new byte[maxSamplesPerWrite * getBytesPerSample(sampleType)];
+      case AudioFormat.ENCODING_AC3:
+      case AudioFormat.ENCODING_E_AC3:
+        return new byte[maxSamplesPerWrite];
+      default:
+        Log.e(TAG, "Unsupported sample type: " + sampleType);
```
yell0wd0g, line 89: Is it best to error and continue, or just crash? If this
comes from a deserialized/enum class, this a non expected
value would mean a very bad condition, so I'd say crash.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -111,16 +110,32 @@ std::unique_ptr<AudioTrackBridge> AudioTrackBridge::Create(
...
   }
 
-  return std::make_unique<AudioTrackBridge>(PassKey<AudioTrackBridge>(),
-                                            max_samples_per_write,
-                                            j_audio_track_bridge, j_audio_data);
+  const auto j_audio_data_var = [&]() -> DataArray {
+    if (coding_type == kSbMediaAudioCodingTypePcm &&
+        sample_type == kSbMediaAudioSampleTypeFloat32) {
+      ScopedJavaLocalRef<jfloatArray> j_float_array =
+          Java_AudioTrackBridge_getPreAllocatedFloatArray(env,
```
yell0wd0g, line 117: Just a thought that this pattern means //starboard depends on
//cobalt/android, is that OK...? (Might be a pervasive/
preexisting condition).

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -269,6 +291,22 @@ public int getPlayState() {
     return mAudioTrack.getPlayState();
   }
 
+  @CalledByNative
+  public float[] getPreAllocatedFloatArray() {
```
yell0wd0g, line 295: nit: s/getPreAllocatedFloatArray/getPreAllocatedAudioDataAsFloatArray/ ?
Similarly below.

yell0wd0g (approved):
LGTM with some random questions/suggestions.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
kjyoun: crash would be better.

Since java does not `CHECK`,  I threw runtime exception, which will crash the app, leaving clear crash stack

cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
kjyoun: Done

starboard/android/shared/audio_track_bridge.cc
kjyoun: Yes, `//starboard/android/shared`  currently depends on  `//cobalt/android:jni_headers`  (as well as several  `//cobalt/browser` targets). This is a known, pre-existing condition tracked by b/492704919 to eventually reverse the dependency flow so Starboard doesn't depend on Cobalt.

kjyoun:
oops. commit is merged before updating new commits to address comments.

Will address them in new commit

PR #10782, opened 2026-06-05
Fix the cobalt  stacktrace symbolization issue

santoshmahto7:
symbolize.cc in //base has been updated for cobalt symbol
resolution, but the change is not present in //copied_base version of
symbolize.cc. This commit syncs them.
    
Bug: 511316178

### DIFF
--- a/copied_base/base/third_party/symbolize/symbolize.cc
+++ b/copied_base/base/third_party/symbolize/symbolize.cc
@@ -52,6 +52,9 @@
 #include GLOG_BUILD_CONFIG_INCLUDE
 #endif  // GLOG_BUILD_CONFIG_INCLUDE
 
+// Allow pread() to use higher offsets on 32-bit systems.
+#define _FILE_OFFSET_BITS 64
+
 #include "symbolize.h"
 
 #include "utilities.h"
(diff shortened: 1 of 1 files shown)


copied_base/base/third_party/symbolize/symbolize.cc
```diff
@@ -771,8 +802,30 @@ static ATTRIBUTE_NOINLINE bool SymbolizeAndDemangle(void* pc,
   out[0] = '\0';
   SafeAppendString("(", out, out_size);
 
+#if BUILDFLAG(IS_STARBOARD) && \
```
kjyoun: readability nit

Isn't this more readable?
```
const int object_fd = [&]() {
#if BUILDFLAG(IS_STARBOARD) && \
    BUILDFLAG(USE_EVERGREEN) && \
    BUILDFLAG(IS_STARBOARD_TOOLCHAIN)
  EvergreenInfo evergreen_info;
  // If BOTH conditions are true, we know we want to open the evergreen file.
  if (GetEvergreenInfo(&evergreen_info) && IS_EVERGREEN_ADDRESS(pc, evergreen_info)) {
    start_address = evergreen_info.base_address;
    base_address = evergreen_info.base_address;
    return OpenFile(evergreen_info.file_path_buf); // <-- This replaces "if (file_name)"
  }
#endif

  // If the #if flags are false, OR if either of the conditions above failed,
  // execution drops down here automatically. 
  return OpenObjectFileContainingPcAndGetStartAddress(
      pc0, start_address, base_address, out + 1, out_size - 1);
}();
````

santoshmahto7: looks good, IIRC `evergreen_info.file_path_buf` could be NULL, so a check might be required - will confirm this.
Also,  I would prefer to do this in another commit (in this PR) to have changes in //base and //copied_base  remain in sync  -  above changes need to be updated in both //copied_base and //base
Thanks for reviewing it.

niranjanyardi (changes requested):
@santoshmahto7 why are we merging this into 27.lts ?

you should be making changes to main branch

jonastsai:
Please update your PR description as it will be used in git log.

copied_base/base/third_party/symbolize/symbolize.cc
santoshmahto7: Done

santoshmahto7:
> @santoshmahto7 why are we merging this into 27.lts ?
> 
> you should be making changes to main branch

Done

niranjanyardi:
@jellefoks do we need this PR as your change deobfuscates the cobalt crash stack traces ?https://github.com/youtube/cobalt/pull/10930

jellefoks:
> @jellefoks do we need this PR as your change deobfuscates the cobalt crash stack traces ?#10930

It's always better to get symbolized results directly in the log when using an unstripped library. The #10930 change is mainly for when you deployed the stripped library (which sometimes is necessary because the unstripped can be really large for a device).

y4vor (approved):

niranjanyardi (approved):

PR #10787, opened 2026-06-05
test: Update flaky test filters

arjungm:
Update the test filters for android-arm and linux-x64x11 based on
reports from the flaky test dashboard. Disabling these unstable tests
ensures more reliable CI results for Perfetto and Cobalt browser tests.

Bug: 520506351

### DIFF
--- a/cobalt/testing/filters/linux-x64x11/cobalt_browsertests_filter.json
+++ b/cobalt/testing/filters/linux-x64x11/cobalt_browsertests_filter.json
@@ -1,6 +1,7 @@
 {
   "comment": "Temporarily disable tests in NavigationBrowserTest. See b/433354983.",
   "failing_tests": [
+    "All/CobaltMemoryThresholdBrowserTest.BlankPageMemoryFootprint/0",
     "*CobaltMetricsBrowserTest*",
     "*DeferSpeculativeRFHCreationRenderProcessTest*",
     "*NavigationBrowserTestDeprecateUnloadOptOut*",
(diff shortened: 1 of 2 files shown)


cobalt/testing/filters/linux-x64x11/cobalt_browsertests_filter.json
```diff
@@ -1,6 +1,7 @@
 {
   "comment": "Temporarily disable tests in NavigationBrowserTest. See b/433354983.",
   "failing_tests": [
+    "All/CobaltMemoryThresholdBrowserTest.BlankPageMemoryFootprint/0",
```
linxinan-yt, line 4: @tasunnn Please take a look. If the flakiness cannot be mitigated, gardener may disable these tests.

oxve (approved):

cobalt/testing/filters/linux-x64x11/cobalt_browsertests_filter.json
tasunnn: @linxinan-chops and I were only able to find one failure for CobaltMemoryThresholdBrowserTest.BlankPageMemoryFootprint test - https://github.com/youtube/cobalt/actions/runs/26681674436/job/78644679112. Looking at the logs, I think it's a race condition due to browser test setup rather than the test itself. @arjungm do you see any other failures related to this specific test? If not, I think we could keep this test enabled? Thanks!

oxve (changes requested):
This should land on `main` instead of 27.lts

cobalt/testing/filters/linux-x64x11/cobalt_browsertests_filter.json
arjungm: ack, will undo this one (turns out the flaky detector itself flaked on this one hah)

arjungm:
It seems like these tests are no longer showing up in the flaky dashboards, so I'll close this out.

PR #10790, opened 2026-06-05
media: Remove FLAC from Cobalt builds

osagie98:
Exclude FLAC audio handling and its associated third-party dependency
from Cobalt builds. This reduction in scope helps minimize the binary
size and build complexity for Cobalt, as FLAC support is not required
for its target platforms.

Bug: 517266324

### DIFF
--- a/media/audio/BUILD.gn
+++ b/media/audio/BUILD.gn
@@ -41,9 +41,11 @@ source_set("audio") {
     # TODO(dalecurtis): Move android audio pieces into //media/audio.
     "//media/base/android",
   ]
-  if (!is_cobalt) {
-    visibility += [ ":flac_audio_handler_fuzzer" ]
+
+  if (is_cobalt) {
+    visibility -= [ ":flac_audio_handler_fuzzer" ]
   }
+
   sources = [
     "aecdump_recording_manager.cc",
     "aecdump_recording_manager.h",
@@ -388,12 +390,12 @@ source_set("audio") {
     ]
   }
 
-  if (!is_cobalt) {
-    sources += [
+  if (is_cobalt) {
+    sources -= [
       "flac_audio_handler.cc",
       "flac_audio_handler.h",
     ]
-    deps += [ "//third_party/flac" ]
+    deps -= [ "//third_party/flac" ]
   }
 }
 
@@ -533,8 +535,9 @@ source_set("unit_tests") {
       "test_data.h",
       "wav_audio_handler_unittest.cc",
     ]
-    if (!is_cobalt) {
-      sources += [ "flac_audio_handler_unittest.cc" ]
+    
+    if (is_cobalt) {
+      sources -= [ "flac_audio_handler_unittest.cc" ]
     }
   }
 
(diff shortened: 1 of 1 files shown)


media/audio/BUILD.gn
```diff
@@ -42,6 +41,9 @@ source_set("audio") {
     # TODO(dalecurtis): Move android audio pieces into //media/audio.
     "//media/base/android",
   ]
+  if (!is_cobalt) {
+    visibility += [ ":flac_audio_handler_fuzzer" ]
+  }
```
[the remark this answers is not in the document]
osagie98: We should stick with `is_cobalt` for cobalt changes unless we upstream.

media/audio/BUILD.gn
```diff
@@ -42,6 +41,9 @@ source_set("audio") {
     # TODO(dalecurtis): Move android audio pieces into //media/audio.
     "//media/base/android",
   ]
+  if (!is_cobalt) {
```
borongc: From @haozheng-cobalt, we prefer use subtraction when possible (as it's not upstream), same for all other places:

```
  if (is_cobalt) {
    visibility -= [ ":flac_audio_handler_fuzzer" ]
  }
```

osagie98: Ack

media/audio/BUILD.gn
```diff
@@ -30,7 +30,6 @@ source_set("audio") {
   # is a roll-up target which is part of the //media component. Most other DEPs
   # should be using //media and not directly DEP this roll-up target.
   visibility = [
-    ":flac_audio_handler_fuzzer",
```
borongc, line 33: nit: need to add this back

media/audio/BUILD.gn
```diff
@@ -123,8 +127,6 @@ source_set("audio") {
     "fake_audio_manager.h",
     "fake_audio_output_stream.cc",
     "fake_audio_output_stream.h",
-    "flac_audio_handler.cc",
```
borongc, line 126: nit: need to add these back

media/audio/BUILD.gn
```diff
@@ -145,7 +147,6 @@ source_set("audio") {
     "//build:android_buildflags",
     "//build:chromecast_buildflags",
     "//media/base",
-    "//third_party/flac",
```
borongc, line 148: nit: need to add this back

media/audio/BUILD.gn
```diff
@@ -523,10 +532,13 @@ source_set("unit_tests") {
 
   if (is_chromeos || is_castos || is_cast_android) {
     sources += [
-      "flac_audio_handler_unittest.cc",
```
borongc, line 526: same here

media/audio/BUILD.gn
osagie98: Incorrectly did a rebase, I'm fixing this

osagie98:
The binary size savings for this is minor as FLAC is mostly already optimized out of the builds, so there's no need to create a new flag for it per our guidelines

borongc (approved):

PR #10796, opened 2026-06-08
starboard: Add option to reduce thread stack size

zhongqiliang:
Introduce feature flags to allow reducing the default thread stack
size to 256KB. This helps minimize virtual memory usage on platforms
where the system default (often 1MB) is unnecessary for most helper
threads.

The GPU and renderer threads are explicitly set to 1MB to prevent
stack overflow crashes, as these components often require more stack
space than the reduced default.

Bug: 518886071
Bug: 527182602


zhongqiliang:
### **1. Why this fix? (The Problem)**
The crash in b/518886071 is a native abort (`SIGABRT`) on a 32-bit ARM Android TV device. It occurs because Cobalt failed to spawn a new thread (`pthread_create` returned `EAGAIN`). 

In 32-bit processes, the virtual address space is limited to **3GB**. Cobalt runs many threads. By default, every thread spawned in Starboard uses the OS default stack size (typically **1MB** of virtual memory). When many threads are active, the process runs out of virtual address space, causing thread creation to fail and the app to crash.

### **2. How the fix helps**
We added support for configuring the stack size of `starboard::Thread` and reduced it for the audio threads:
*   **`AudioTrackOutThread`:** Reduced from **1MB to 256KB** (75% savings).
*   **`TesterThread`:** Reduced from **1MB to 128KB** (87.5% savings).

This saves **768KB** of virtual address space for every audio playback session. If the app has multiple audio/video instances or runs tests, the savings compound, leaving more virtual memory available for the rest of the application (like JS engine and graphics) and preventing the `EAGAIN` thread creation failure.


### **3. Pros and Cons**

#### **Pros:**
*   **Prevents Crashes:** Directly addresses the root cause of the crash on 32-bit devices by freeing up virtual memory.
*   **Resilience to Fragmentation:** Smaller allocations (256KB vs 1MB) are easier for the OS to fit into fragmented memory.
*   **Zero Performance Degradation:** The thread's real-time priority is untouched, meaning audio data delivery, latency, and quality are **not** affected.
*   **Zero UI/UX Impact:** Audio buffering (allocated on the heap) is unaffected.

#### **Cons:**
*   **Maintenance:** Introduces a new configuration option (`stack_size` in `ThreadOptions`) that developers must be mindful of if they create new threads.


### **4. Risks & Mitigations**

The only technical risk of reducing stack size is **Stack Overflow** (the thread runs out of stack space and crashes with `SIGSEGV`).

#### **Risk Analysis & Mitigation:**
*   **Risk:** If `AudioTrackOutThread` has deep function calls or allocates large variables on the stack, it could exceed 256KB.
*   **Mitigation (Very Safe):** 
    1.  **Shallow Call Stack:** The audio thread runs a very simple loop that calls JNI methods to write data. There is no recursion (nested calls to the same function).
    2.  **Heap for Large Data:** All audio data buffers are allocated on the **heap** (via `std::vector`), not on the stack.
    3.  **256KB is Huge:** For a simple worker thread, 256KB is extremely generous. In systems programming, threads often run on 16KB or 32KB. JNI overhead is well within 256KB.
    4.  **Enforced Safety:** We added a check (`SB_CHECK_EQ(err, 0)`) in `Thread::Start` to ensure that if the OS rejects the stack size (e.g., if it's below the system minimum), it fails immediately during development rather than silently failing or causing undefined behavior.

starboard/shared/starboard/player/player_worker.cc
```diff
@@ -52,9 +52,10 @@ PlayerWorker::PlayerWorker(SbMediaAudioCodec audio_codec,
                            SbPlayerErrorFunc player_error_func,
                            SbPlayer player,
                            void* context)
-    : job_thread_(JobThread::Create(
-          "player_worker",
-          ThreadOptions().SetPriority(ThreadPriority::kHigh))),
+    : job_thread_(JobThread::Create("player_worker",
```
xiaomings: All threads in this PR are only launched once or twice per playback, and we have at most two concurrent playbacks.  Is this part of a large effort to reduce stack sizes of all threads?  Can we guard them using an experiment?

zhongqiliang: If there are other threads that can reduce stack size, we should try it. I will add an experiment for set the value for the threads in this PR.

borongc: These are the light threads if we check their cpu usage (the measurement I did before):

|                   |   count |         min |         25% |        50% |        75% |        max |       mean |
|:------------------|--------:|------------:|------------:|-----------:|-----------:|-----------:|-----------:|
| total cpu usage   |     350 | 121         | 168.25      | 201        | 314.75     | 397        | 238.994    |
| dev.cobalt.coat_x |     340 |  96.7832    | 117.355     | 130.649    | 141.503    | 207.01     | 131.384    |
| NetworkService    |     340 |  14.2207    |  22.6324    |  28.8356   |  32.7857   |  57.6459   |  28.1691   |
| Chrome_InProcRe   |     340 |  14.9033    |  22.0808    |  25.4828   |  28.5706   |  57.8471   |  25.8827   |
| Media             |     340 |  11.1895    |  14.5947    |  15.1366   |  15.7348   |  18.2036   |  15.1206   |
| ThreadPoolForeg   |     340 |   4.68541   |  11.8802    |  15.041    |  18.4196   |  33.9149   |  15.5502   |
| MediaCodec_loop   |     340 |  10.1211    |  13.0138    |  13.7127   |  14.6247   |  17.4044   |  13.7906   |
| video_decoder     |     340 |   6.65323   |   8.59112   |   9.16501  |   9.75453  |  12.1034   |   9.17287  |
| HwBinder          |     340 |   4.7379    |   6.37071   |   6.71767  |   7.15872  |   8.46063  |   6.75289  |
| player_worker     |     340 |   2.33564   |   3.21028   |   3.39188  |   3.60632  |   4.22535  |   3.3893   |
| dev.cobalt.coat_y |     340 |   2.41117   |   2.86966   |   3.00529  |   8.44939  |  12.013    |   4.93434  |
| audio_track_out   |     340 |   2.09524   |   2.80245   |   2.97994  |   3.12229  |   5.50898  |   2.99574  |
| binder            |     340 |   0.576037  |   1.5467    |   1.70489  |   1.866    |   2.85344  |   1.70097  |
| CodecLooper       |     340 |   0.857143  |   1.21273   |   1.29481  |   1.39041  |   1.7316   |   1.29913  |
| ThreadPoolSingl   |     336 |   0.0943396 |   0.968718  |   1.03538  |   1.10565  |   1.2987   |   1.025    |
| ThreadPoolServi   |     334 |   0.180018  |   0.91612   |   0.98401  |   1.04724  |   1.28023  |   0.977035 |
| HeapTaskDaemon    |      38 |   0.110254  |   0.551122  |   0.586867 |   0.634475 |   0.759494 |   0.566122 |
| VizCompositorTh   |     336 |   0.100806  |   0.232761  |   0.272672 |   0.357463 |   0.704225 |   0.299494 |
| Compositor        |     333 |   0.100806  |   0.130548  |   0.231214 |   0.261097 |   0.598802 |   0.226565 |
| ChoreographerOw   |     212 |   0.0865052 |   0.112202  |   0.118765 |   0.126663 |   0.291121 |   0.127207 |
| Chrome_IOThread   |     140 |   0.0846024 |   0.111483  |   0.117925 |   0.12747  |   0.269542 |   0.122851 |
| Chrome_ChildIOT   |     117 |   0.0846024 |   0.111483  |   0.117509 |   0.125156 |   0.224972 |   0.120259 |
| Chrome_InProcGp   |       2 |   0.109051  |   0.111166  |   0.11328  |   0.115394 |   0.117509 |   0.11328  |
| VideoFrameCompo   |       1 |   0.109051  |   0.109051  |   0.109051 |   0.109051 |   0.109051 |   0.109051 |
| Profile Saver     |       8 |   0.0907441 |   0.0987085 |   0.102308 |   0.108784 |   0.129534 |   0.105746 |
| Jit thread pool   |      11 |   0.0846024 |   0.0905004 |   0.100806 |   0.111799 |   0.14556  |   0.103974 |

Do you have any local number to show the improvement?

xiaomings: > If there are other threads that can reduce stack size, we should try it. I will add an experiment for set the value for the threads in this PR.

There are a lot of worker threads in the process, and they are probably better targets for such an optimization.

zhongqiliang: The latest change will use 1 flag to reduce the stack size for all threads created from starboard common.

starboard/common/thread.cc
```diff
@@ -31,6 +31,10 @@
 #include "starboard/common/thread_platform.h"
 #include "starboard/system.h"
 
+#if BUILDFLAG(IS_ANDROID)
```
borongc, line 34: Although I knew this is due to the Starboard feature being only available on Android, we typically don't allow to have platform-specific code for the common code under starboard.

johnxwork: +1 to not having platform-specific code in Starboard common implementations. I think you might compile Linux just fine without adding the macro guards.

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -172,6 +175,10 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
       extraCommandLineArgs.add("--js-flags=" + jsFlags.toString());
     }
 
+    if (javaSwitches.containsKey(JavaSwitches.REDUCE_STARBOARD_THREAD_STACK_SIZE)) {
+      extraCommandLineArgs.add("--enable-features=ReduceStarboardThreadStackSize");
```
johnxwork: Do you want to turn this boolean feature into an int feature? So that you have the option to try different the stack size from server GCL config.

starboard/common/thread.cc
```diff
@@ -129,8 +133,23 @@ struct Thread::Data {
   Semaphore join_sema_;
 };
 
+// static
+size_t Thread::GetDefaultStackSize() {
+#if BUILDFLAG(IS_ANDROID)
+  if (starboard::features::FeatureList::IsEnabled(
+          starboard::features::kReduceStarboardThreadStackSize)) {
+    return 256 * 1024;
```
johnxwork, line 75: Just curious: Did you try dumping the peak stack size during endurance run? It feels like media vs JNI vs native have different stack size usage.

starboard/common/thread.cc
zhongqiliang: removed

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
zhongqiliang: not really, this number is good for general thread size. If there are exceptions for special threads, we need to change code for them separately

starboard/common/thread.cc
```diff
@@ -129,8 +130,21 @@ struct Thread::Data {
   Semaphore join_sema_;
 };
 
+// static
+size_t Thread::GetDefaultStackSize() {
+  if (starboard::features::FeatureList::IsEnabled(
```
kjyoun: nit:  you dont need "starboard::", since code is already inside starboard namespace

starboard/common/thread_options.h
```diff
@@ -74,8 +74,13 @@ struct ThreadOptions {
     priority = priority_in;
     return *this;
   }
+  ThreadOptions& SetStackSize(size_t stack_size_in) {
+    stack_size = stack_size_in;
+    return *this;
+  }
 
   std::optional<ThreadPriority> priority;
+  size_t stack_size = 0;
```
kjyoun: can you make this this `std::optional<size_t>` 

Then we don't need sentinel value of 0

starboard/common/thread.h
```diff
@@ -46,6 +46,11 @@ class Thread {
   }
   virtual ~Thread();
 
+  // Returns the default stack size for threads. If the stack size reduction
+  // feature is enabled, this will return 256KB, otherwise 0 (which triggers
+  // the system default, e.g. 1MB on Android).
+  static size_t GetDefaultStackSize();
```
kjyoun: you dont  need to declare this method in header.

since `GetDefaultStackSize()` has no dependencies on other member variables and is used only inside of `.cc` file. You can put this as a stand-alone function inside anonymous namespace in .cc file

starboard/common/thread.cc
```diff
@@ -129,8 +130,21 @@ struct Thread::Data {
   Semaphore join_sema_;
 };
 
+// static
+size_t Thread::GetDefaultStackSize() {
```
kjyoun: return `std::optional<size_t>` ?

regarding name, I think `GetOverridedStackSize()` is better

`std::optional<size_t> GetOverridedStackSize()` clearly conveys what this method does
- size_t when override value is available, 
- std::nullopt when there is no override value.

starboard/common/thread.cc
```diff
@@ -129,8 +130,21 @@ struct Thread::Data {
...
+    return 256 * 1024;
+  }
+  return 0;
+}
+
 Thread::Thread(std::string_view name, const ThreadOptions& options)
-    : name_(name), priority_(options.priority), d_(std::make_unique<Data>()) {}
+    : name_(name),
+      priority_(options.priority),
+      stack_size_(options.stack_size > 0 ? options.stack_size
```
kjyoun: How about this?

```
std::optional<size_t> stack_size_;

...

stack_size_(options.stack_size ? options.stack_size : GetOverridedStackSize())


...
 if (stack_size_) {
    int err = pthread_attr_setstacksize(&attributes, *stack_size_);
    if (err != 0) {
      SB_LOG(WARNING) << "Failed to set stack size to " << stack_size_
                      << ", error: " << err << ". Falling back to default.";
    }
  }
```

starboard/common/thread_options.h
zhongqiliang: Make it inside of starboard/common/thread.cc

starboard/common/thread.cc
zhongqiliang: removed

starboard/common/thread.h
zhongqiliang: done

starboard/common/thread.cc
zhongqiliang: done

starboard/common/thread.cc
zhongqiliang: fix is similar but a little bit different

starboard/common/thread_options.h
kjyoun: can you change this to `std::optioanal<size_t>` as we did for priority?

http://go/totw/171 recommends not to use sentinel value, which is 0 in this code.

starboard/common/thread.cc
```diff
@@ -130,7 +143,12 @@ struct Thread::Data {
 };
 
 Thread::Thread(std::string_view name, const ThreadOptions& options)
-    : name_(name), priority_(options.priority), d_(std::make_unique<Data>()) {}
+    : name_(name),
+      priority_(options.priority),
+      stack_size_(options.stack_size > 0
```
kjyoun: with std::optional options.stack_size, this cod will become

```
stack_size_(options.stack_size ? options.stack_size : GetOverriddenStackSize()) 
```

zhongqiliang: done

starboard/common/thread_options.h
zhongqiliang: done

starboard/common/thread.cc
zhongqiliang: Tested locally, and comment on the bug.

base/threading/platform_thread_linux_base.cc
```diff
@@ -5,10 +5,23 @@
...
 
 #include "base/threading/platform_thread.h"
+#include "base/command_line.h"
 
 namespace base {
 
 size_t GetDefaultThreadStackSize(const pthread_attr_t& attributes) {
+#if BUILDFLAG(IS_STARBOARD)
+  if (base::CommandLine::InitializedForCurrentProcess()) {
+    std::string enabled_features =
```
kjyoun: nit: single-nesting instead of double-nesting?


```
if (base::CommandLine::InitializedForCurrentProcess() &&
    base::CommandLine::ForCurrentProcess()
        ->GetSwitchValueASCII("enable-features")
        .find("ReduceAndroidThreadStackSize") != std::string::npos) {
  return 256 * 1024;
}
```

starboard/common/thread.cc
```diff
@@ -29,6 +29,9 @@
 #include "starboard/common/log.h"
 #include "starboard/common/semaphore.h"
 #include "starboard/common/thread_platform.h"
+#if BUILDFLAG(IS_ANDROID)
```
kjyoun, line 34: nit: separate grouping for android header?

https://source.corp.google.com/h/github/youtube/cobalt/+/main:content/renderer/media/media_factory.cc

```
#include "starboard/common/log.h"
#include "starboard/common/semaphore.h"
#include "starboard/common/thread_platform.h"
#include "starboard/system.h"

#if BUILDFLAG(IS_ANDROID)
#include "starboard/shared/starboard/features.h"
#endif
```

starboard/common/thread.cc
```diff
@@ -82,6 +99,14 @@ void Thread::Start() {
   pthread_attr_t attributes;
   pthread_attr_init(&attributes);
 
+  if (stack_size_) {
+    int err = pthread_attr_setstacksize(&attributes, *stack_size_);
```
kjyoun: nit: if-with-intializer format?

```
if (int err = pthread_attr_setstacksize(&attributes, *stack_size_);
    err != 0) {
  SB_LOG(WARNING) << "Failed to set stack size to " << *stack_size_
                    << ", error: " << err << ". Falling back to default.";
}
```

base/threading/platform_thread_android.cc
```diff
@@ -119,6 +120,18 @@ void TerminateOnThread() {
 }
 
 size_t GetDefaultThreadStackSize(const pthread_attr_t& attributes) {
+#if BUILDFLAG(IS_COBALT)
```
kjyoun, line 123: ditto: single-nesting?

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -175,6 +181,14 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
...
```
kjyoun: nit: use constant, instead of repeating string?

```
  if (javaSwitches.containsKey(JavaSwitches.REDUCE_STARBOARD_THREAD_STACK_SIZE)) {
      extraCommandLineArgs.add("--enable-features=" + javaSwitches.containsKey(JavaSwitches.REDUCE_STARBOARD_THREAD_STACK_SIZE);
}
```

kjyoun (commented):
LGTM % comments

kjyoun (approved):

starboard/common/thread_options.h
```diff
@@ -74,8 +74,13 @@ struct ThreadOptions {
     priority = priority_in;
     return *this;
   }
+  ThreadOptions& SetStackSize(size_t stack_size_in) {
```
johnxwork: This function looks very strange: why does it return itself as a reference? I also don't see where this is used?

starboard/common/thread.cc
```diff
@@ -64,8 +68,22 @@ struct Thread::Data {
   Semaphore join_sema_;
 };
 
+std::optional<size_t> GetOverriddenStackSize() {
```
johnxwork: This should also be "GetDefaultStackSize()", to align with your naming convention elsewhere in this PR. And it's also used as "default" instead of "override" on line 85.

johnxwork (approved):

PR #10808, opened 2026-06-08
Setup staging branch.

briantting:
Remove CODEOWNERS file and limit staging workflows to match those run on 27.lts. Add logic before gclient sync to id and delete dirty DEPS sub-repos leftover from persistent workers.

Bug: 518006057


linxinan-yt (approved):
Awesome!

oxve (approved):

PR #10818, opened 2026-06-09
starboard: Fix SB_DCHECK with Objective-C++ message sends

abhijeetk:
SB_DCHECK fails to compile in QA builds when the condition contains
Objective-C message send syntax like [obj method:arg]. This happens
because the disabled SB_DCHECK uses `sizeof(bool(condition))`, and
the C++ parser inside sizeof interprets brackets as array subscripts
rather than message sends.

Replace the sizeof pattern with the VoidifyStream ternary approach
already used by //base/check.h.

Also update SB_DCHECK_OP to use the same pattern, keeping
CheckOpValueStr referenced to avoid unused-function warnings.

Bug: 512045535
Bug: 447838008

### DIFF
--- a/starboard/common/check_op.h
+++ b/starboard/common/check_op.h
@@ -246,7 +246,12 @@ DEFINE_SB_CHECK_OP_IMPL(GT, > )
 
 #else
 
-#define SB_DCHECK_OP(name, op, val1, val2) SB_EAT_STREAM_PARAMETERS
+// Mirrors //base/check_op.h: keeps CheckOpValueStr instantiated to avoid
+// unused-function warnings in modular builds.
+#define SB_DCHECK_OP(name, op, val1, val2)                        \
+  SB_EAT_CHECK_STREAM_PARAMS((::starboard::CheckOpValueStr(val1), \
+                              ::starboard::CheckOpValueStr(val2), \
+                              (val1)op(val2)))
 #endif
 
 // clang-format off
--- a/starboard/common/log.h
+++ b/starboard/common/log.h
@@ -114,6 +114,17 @@ class LogMessageVoidify {
   void operator&(std::ostream&);
 };
 
+// Copied from base/check.h.
+// Class used to explicitly ignore an ostream, and optionally a boolean value.
+class VoidifyStream {
+ public:
+  VoidifyStream() = default;
+  explicit VoidifyStream(bool) {}
+
+  // Binary & has lower precedence than << but higher than ?:
+  void operator&(std::ostream&) {}
+};
+
 }  // namespace starboard
 
 #define SB_LOG_MESSAGE_INFO \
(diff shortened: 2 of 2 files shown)


abhijeetk:

starboard/tvos/shared/media/url_player_set_drm_system.mm
```diff
@@ -30,8 +30,10 @@ void SbUrlPlayerSetDrmSystem(SbPlayer player, SbDrmSystem drm_system) {
     SBDDrmManager* drmManager = application.drmManager;
     SBDPlayerManager* playerManager = application.playerManager;
 
-    SB_DCHECK([drmManager isApplicationDrmSystem:drm_system]);
-    SB_DCHECK([playerManager isUrlPlayer:player]);
+    bool isAppDrmSystem = [drmManager isApplicationDrmSystem:drm_system];
+    SB_DCHECK(isAppDrmSystem);
+    bool isUrlPlayerFlag = [playerManager isUrlPlayer:player];
+    SB_DCHECK(isUrlPlayerFlag);
```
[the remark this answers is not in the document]
abhijeetk: Accepted and resolved.

starboard/tvos/shared/media/url_player_get_extra_info.mm
```diff
@@ -30,7 +30,8 @@ void SbUrlPlayerGetExtraInfo(SbPlayer player,
   }
   @autoreleasepool {
     SBDPlayerManager* playerManager = SBDGetApplication().playerManager;
-    SB_DCHECK([playerManager isUrlPlayer:player]);
+    bool isUrlPlayerFlag = [playerManager isUrlPlayer:player];
+    SB_DCHECK(isUrlPlayerFlag);
```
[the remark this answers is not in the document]
abhijeetk: Accepted and resolved.

starboard/shared/widevine/drm_system_widevine.cc
```diff
@@ -192,7 +192,7 @@ void EnsureWidevineCdmIsInitialized(const std::string& company_name,
 #if BUILDFLAG(COBALT_IS_RELEASE_BUILD)
   log_level = wv3cdm::kSilent;
 #endif  // BUILDFLAG(COBALT_IS_RELEASE_BUILD)
-  wv3cdm::Status status =
+  [[maybe_unused]] wv3cdm::Status status =
```
rakuco: Isn't it better to turn the DCHECK below into a CHECK instead?

starboard/tvos/shared/media/url_player_get_extra_info.mm
```diff
@@ -30,7 +30,8 @@ void SbUrlPlayerGetExtraInfo(SbPlayer player,
   }
   @autoreleasepool {
     SBDPlayerManager* playerManager = SBDGetApplication().playerManager;
-    SB_DCHECK([playerManager isUrlPlayer:player]);
```
rakuco, line 33: Since these are based on the `//base` versions, which do work with Objective-C++ calls, can we see if there's anything to backport from `//base` instead? These `[[maybe_unused]]` annotations do not look very nice.

abhijeetk: Adopted similar approach already used by //base/check.h

starboard/shared/widevine/drm_system_widevine.cc
abhijeetk: Using approach already used by //base/check.h

abhijeetk:
These changes are going to affect all starboard variants and initially introduced in #6138 ,

Hi @andrewsavage1 , PTAL.

rakuco (approved):
Nice, this is better than the previous workarounds!

Can you 1) update the commit message 2) if possible reference the upstream commits being imported?

abhijeetk:
> 1) update the commit message

Done
> 2) if possible reference the upstream commits being imported?

Sone. We just copied the relevant bits from Chromium's  //base/check.h and documented this to both the code comments and the commit message explicitly that this is borrowed from  //base/check.h.

kjyoun (approved):

fayev-g (approved):

andrewsavage1 (approved):

PR #10819, opened 2026-06-09
Fix crash due to DCHECK failure

santoshmahto7:
Change-Id: I8bcdafe40f0a4762116ddd05d81f5727a1ae3ce1


gpu/command_buffer/service/feature_info.cc
```diff
@@ -282,7 +282,7 @@ bool IsGL_REDSupportedOnFBOs() {
   return true;
 #else
 
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_ANDROID) || BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS)
```
niranjanyardi, line 285: can you add some explanantion as to why this is a hack int he PR descitpion

also why is this not being merged into 27.lts rather than main ? please target your PR's to main branch

niranjanyardi (changes requested):

niranjanyardi:
> Change-Id: I8bcdafe40f0a4762116ddd05d81f5727a1ae3ce1

please add a good descritpion and the bug number you are working on

PR #10829, opened 2026-06-09
Map base::DIR_HOME kSbSystemPathFilesDirectory.

y4vor:
Instead of using the cache location use the files
location.

Bug: 371432387


andrewsavage1:
What's the reason for doing this?

andrewsavage1:
> What's the reason for doing this?

Seems like it's mostly settings/configurations based on a code search

andrewsavage1 (approved):

PR #10830, opened 2026-06-09
Remove a gtest hack for custom thread locking

y4vor:
Now that we have all the thread APIs the hack is no longer needed.
However there was a memory corruption Bug.

1. The Root Cause (The Bug)
The memory corruption was caused by an undocumented data race and thread-safety limitation in the core Google Test (gtest) framework. Google Test maintains a shared, process-wide TestResult object for each active test case to record failures. Under the hood, the internal AddTestPartResult() function appends every pass/fail outcome to a standard std::vector via push_back() without any mutex locking or thread synchronization.

2. The Trigger Mechanism (Multithreaded ASSERT)
The SecondaryPlayerTest suite (and the underlying SbPlayerTestFixture cleanup harness) spawned two concurrent background OS threads (Primary and Secondary players) to validate parallel media playback.

When those concurrent threads executed Google Test assertion macros (ASSERT_NO_FATAL_FAILURE, ASSERT_TRUE, etc.), they invoked AddTestPartResult() simultaneously. The resulting data race on the std::vector::push_back() call triggered a heap reallocation. One thread freed the old memory buffer while the second thread simultaneously wrote its stack address (HasNewFatalFailureHelper) into the freed space. This "write-after-free" bug overwrote adjacent live objects on the heap owned by the Main thread (the Google Test Thread-Local Storage reporter map), resulting in a SIGSEGV segmentation fault during the execution of the very next test case.

3. The hack before 
ThreadLocal() {
int res = pthread_key_create(&key_, [](void* value) {
delete static_cast<T*>(value); // <-- The Hack
});
SB_DCHECK(res == 0);
}
Because that custom hack explicitly invoked delete on the thread-local storage whenever an OS thread exited, it automatically wiped out and cleared the dangling stack pointers left behind by the concurrent background player threads before the Main thread had a chance to read them.

Bug: 399507045

### DIFF
--- a/starboard/nplb/multiple_player_test.cc
+++ b/starboard/nplb/multiple_player_test.cc
@@ -197,8 +197,8 @@ void WriteSamples(const SbPlayerTestConfig& player_config,
     samples.AddVideoEOS();
   }
 
-  ASSERT_NO_FATAL_FAILURE(player_fixture.Write(samples));
-  ASSERT_NO_FATAL_FAILURE(player_fixture.WaitForPlayerEndOfStream());
+  player_fixture.Write(samples);
+  player_fixture.WaitForPlayerEndOfStream();
 }
 
 TEST_F(MultiplePlayerTest, SunnyDay) {
--- a/starboard/nplb/player_test_fixture.cc
+++ b/starboard/nplb/player_test_fixture.cc
@@ -243,9 +243,9 @@ void SbPlayerTestFixture::Seek(const int64_t time) {
   SB_CHECK(thread_checker_.CalledOnValidThread());
   SB_DCHECK(SbPlayerIsValid(player_));
 
-  ASSERT_FALSE(error_occurred_);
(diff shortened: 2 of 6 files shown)


starboard/nplb/multiple_player_test.cc
```diff
@@ -197,8 +197,8 @@ void WriteSamples(const SbPlayerTestConfig& player_config,
     samples.AddVideoEOS();
   }
 
-  ASSERT_NO_FATAL_FAILURE(player_fixture.Write(samples));
-  ASSERT_NO_FATAL_FAILURE(player_fixture.WaitForPlayerEndOfStream());
+  player_fixture.Write(samples);
```
andrewsavage1, line 200: Why do we need/want to get rid of all the asserts? Can you add some info the the description?

y4vor: Update the commit message. There is a memory corruption from using the ASSERTs on background threads.

andrewsavage1 (approved):

PR #10839, opened 2026-06-10
YouTube certification requires the device type in UA

weifang-cool:
YouTube certification requires the device type in UA
to match the actual product category.

cobalt read the configured device type property ro.vendor.youtube.device_type
and map it to the corresponding Starboard device type value
used in User-Agent generation.

property config refer:starboard/common/device_type.cc

BUG :522026269


weifang-cool (commented):
Updated the PR with the latest changes.

Please help to review again.

Thanks!

PR #10841, opened 2026-06-10
Fix UAF during shutdown by waiting for Stop state before teardown

andrewsavage1:
The previous fix delayed AppEventDelegate deletion to TeardownCallback (scheduled via SbEventSchedule). However, DoTeardown() was still being called inside TeardownCallback which was executed either concurrently (if multi-threaded Starboard) or recursively during the pump wait, resulting in SequenceManager destruction while DoWorkImpl was still running on the stack, causing a crash.

This revised fix:
1. Removes TeardownCallback entirely.
2. ExecuteStepOnUIThread(kStopped) now directly runs the quit_closure_ (which quits the nested RunLoop in SbEventHandle).
3. SbEventHandle blocks on the RunLoop until kStopped is reached (ensuring all intermediate transitions are safely processed).
4. Once the RunLoop exits (meaning we are safely outside the message loop stack), SbEventHandle synchronously calls DoTeardown() to perform cleanup and then deletes the delegate.
5. This guarantees that SequenceManager is only destroyed after the message loop execution has completely returned, avoiding UAF in DoWorkImpl.

CONV=0467c7c6-ea2e-4d82-83dc-d6ec944028bf

Bug: 490989945

### DIFF
--- a/cobalt/app/cobalt.cc
+++ b/cobalt/app/cobalt.cc
@@ -16,8 +16,6 @@
 #include "cobalt/app/app_event_delegate.h"
 #include "starboard/event.h"
 
-namespace {}  // namespace
-
 void SbEventHandle(const SbEvent* event) {
   // This object's lifetime extends beyond the function's lifetime, until the
   // function is called with kSbEventTypeStop at some time in the future.
--- a/cobalt/app/cobalt.cc
+++ b/cobalt/app/cobalt.cc
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "base/run_loop.h"
 #include "cobalt/app/app_event_delegate.h"
 #include "starboard/event.h"
 
@@ -35,6 +36,10 @@ void SbEventHandle(const SbEvent* event) {
   s_lifecycle_delegate->HandleEvent(event);
 
   if (event->type == kSbEventTypeStop) {
+    base::RunLoop run_loop;
+    s_lifecycle_delegate->SetQuitClosure(run_loop.QuitClosure());
+    run_loop.Run();
+    s_lifecycle_delegate->DoTeardown();
     delete s_lifecycle_delegate;
     s_lifecycle_delegate = nullptr;
   }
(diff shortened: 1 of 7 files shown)


jellefoks (commented):
This looks a lot like #10831. How do they relate? What is the exact issue this addresses, is it the same DCHECK?

jellefoks:
This looks like it has overlap with the #10831 that I made yesterday, in implementation and probably also the issue it addresses. We should give it some thought to what the combination looks like.

cobalt/app/cobalt.cc
```diff
@@ -35,6 +36,10 @@ void SbEventHandle(const SbEvent* event) {
   s_lifecycle_delegate->HandleEvent(event);
 
   if (event->type == kSbEventTypeStop) {
+    base::RunLoop run_loop;
+    s_lifecycle_delegate->SetQuitClosure(run_loop.QuitClosure());
```
jellefoks, line 38: In my version, I do a `base::SequencedTaskRunner::GetCurrentDefault()->DeleteSoon(FROM_HERE, this);` in `AppEventDelegate::DoTeardown` that triggers the deletion of the object when the runner is done, and 
`AppEventDelegate::DoTeardown` in turn is called from the `SbEventSchedule(&AppEventDelegate::TeardownCallback, this, 0);` in `AppEventDelegate::ExecuteStepOnUIThread` when the next state is kStopped.

cobalt/app/cobalt.cc
```diff
@@ -35,6 +36,10 @@ void SbEventHandle(const SbEvent* event) {
...
+    run_loop.Run();
+    s_lifecycle_delegate->DoTeardown();
```
jellefoks, line 40: This is already getting called by the event scheduled with the `SbEventSchedule` call from `AppEventDelegate::ExecuteStepOnUIThread`

Which also explains the UAF, because that event runs after this block...

jellefoks:
The difference between this and #10831 is that in #10831, the AppEventDelegate destruction is done in the UI event loop triggered by a scheduled event inserted during the handling of the 'Stop' event, while this PR blocks until the  AppEventDelegate is destroyed.

Given the contract of `kSbEventTypeStop`, this is technically better, if this also resolves the `DCHECK failed: frame_nodes_.empty().` ([521960047](http://b/521960047)).

jellefoks (approved):
Fantastic

y4vor (approved):

PR #10848, opened 2026-06-10
cobalt/metrics: Add Memory Context API and task helpers

Awallky:
Introduce the Memory Context API and associated task helpers to enable
fine-grained memory attribution. This change allows Cobalt to track
memory allocations by subsystem by propagating a memory context
through the task scheduler and thread pool.

The context is maintained using thread-local storage via pthread keys,
with weak symbols used to ensure consistency across the base and
copied_base components. Integrated these contexts into PendingTask
and TaskAnnotator to automatically restore the context when tasks
are executed.

**Note:** This is based on the design specified in http://shortn/_SRQHp8rmF4.

Bug: 512925852

### DIFF
--- a/components/memory_system/parameters.h
+++ b/components/memory_system/parameters.h
@@ -8,10 +8,19 @@
 #include <string>
 #include <string_view>
 
+#include "build/build_config.h"
+#include "build/buildflag.h"
 #include "components/sampling_profiler/process_type.h"
 #include "components/version_info/channel.h"
 
 namespace memory_system {
+
+#if BUILDFLAG(IS_COBALT)
+enum class CobaltMemoryAttributionInclusion {
+  kInclude,
+  kDoNotInclude,
+};
+#endif
 
 // Configuration objects for all memory subsystem components. The parameters are
 // divided by component. The type of the data corresponds to the type used by
@@ -70,11 +79,22 @@ struct DispatcherParameters {
   explicit DispatcherParameters(
       PoissonAllocationSamplerInclusion poisson_allocation_sampler_inclusion,
       AllocationTraceRecorderInclusion allocation_trace_recorder_inclusion,
-      std::string_view process_type);
+      std::string_view process_type
+#if BUILDFLAG(IS_COBALT)
+      // Cobalt memory attribution can introduce performance overhead, so it is
+      // excluded by default unless explicitly requested.
+      , CobaltMemoryAttributionInclusion cobalt_memory_attribution_inclusion =
+            CobaltMemoryAttributionInclusion::kDoNotInclude
+#endif
+  );
 
   PoissonAllocationSamplerInclusion poisson_allocation_sampler_inclusion;
   AllocationTraceRecorderInclusion allocation_trace_recorder_inclusion;
   std::string process_type;
+#if BUILDFLAG(IS_COBALT)
+  CobaltMemoryAttributionInclusion cobalt_memory_attribution_inclusion =
+      CobaltMemoryAttributionInclusion::kDoNotInclude;
+#endif
 };
 
 }  // namespace memory_system
--- a/copied_base/base/memory/cobalt_memory_context.cc
+++ b/copied_base/base/memory/cobalt_memory_context.cc
@@ -65,12 +65,6 @@ ScopedMemoryContext::~ScopedMemoryContext() {
   SetCurrentMemoryContext(prev_context_);
 }
 
-extern "C" {
-void CobaltSetMemoryContext(uint8_t context) {
-  SetCurrentMemoryContext(static_cast<MemoryContext>(context));
-}
-}
-
 std::string_view ContextToString(MemoryContext context) {
   switch (context) {
     case MemoryContext::kUnknown:
--- a/copied_base/base/memory/cobalt_memory_context.h
+++ b/copied_base/base/memory/cobalt_memory_context.h
@@ -1,0 +1,74 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef COPIED_BASE_BASE_MEMORY_COBALT_MEMORY_CONTEXT_H_
+#define COPIED_BASE_BASE_MEMORY_COBALT_MEMORY_CONTEXT_H_
+
+#include <string_view>
+#include "copied_base/base/base_export.h"
+
+namespace base {
+namespace memory {
+
+enum class MemoryContext : uint8_t {
+  kUnknown = 0,
+  kDOM,
+  kLayout,
+  kMedia,
+  kScript,
+  kNetwork,
+  kGraphics,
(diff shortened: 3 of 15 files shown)


copied_base/base/memory/cobalt_memory_context.cc
```diff
@@ -0,0 +1,118 @@
...
+    case MemoryContext::kBlinkParser:
+      return "BlinkParser";
+    case MemoryContext::kPlatformIPC:
+      return "PlatformIPC";
+    case MemoryContext::kPlatformStarboard:
+      return "PlatformStarboard";
+    case MemoryContext::kCount:
+      return "Unknown";
+  }
+}
```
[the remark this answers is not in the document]
Awallky: I applied this change to both  copied_base/base/memory/cobalt_memory_context.cc  and  base/memory/cobalt_memory_context.cc .

Here is the diff of what we changed:

```
    --- a/copied_base/base/memory/cobalt_memory_context.cc                                                                                                                                                         
    +++ b/copied_base/base/memory/cobalt_memory_context.cc                                                                                                                                                         
    @@ -112,6 +112,7 @@                                                                                                                                                                                            
         case MemoryContext::kCount:                                                                                                                                                                               
           return "Unknown";                                                                                                                                                                                       
       }                                                                                                                                                                                                           
    +  return "Unknown";                                                                                                                                                                                           
     }
```

copied_base/base/memory/cobalt_memory_context.h
```diff
@@ -0,0 +1,72 @@
...
+ public:
+  explicit ScopedMemoryContext(MemoryContext context);
+  ~ScopedMemoryContext();
+
+  ScopedMemoryContext(const ScopedMemoryContext&) = delete;
+  ScopedMemoryContext& operator=(const ScopedMemoryContext&) = delete;
+
+ private:
+  MemoryContext prev_context_;
+};
```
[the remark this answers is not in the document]
Awallky: I applied this change to both  copied_base/base/memory/cobalt_memory_context.h  and  base/memory/cobalt_memory_context.h .

Here is the diff of what we changed:

```
    --- a/copied_base/base/memory/cobalt_memory_context.h
    +++ b/copied_base/base/memory/cobalt_memory_context.h
    @@ -52,6 +52,17 @@                                                                                                                                                                                             
     BASE_EXPORT MemoryContext GetCurrentMemoryContext();
     BASE_EXPORT void SetCurrentMemoryContext(MemoryContext context);

    +// ScopedMemoryContext is a helper class that sets the current thread's
    +// memory context for the duration of its lifetime, restoring the previous
    +// context upon destruction.
    +//
    +// Lifetime and Ownership:
    +// This is a stack-allocated helper designed to be used as a local variable.
    +// It should not be dynamically allocated or outlive its enclosing scope.
    +//
    +// Threading Model:
    +// This class is thread-affine and must only be used on the thread where
    +// it was constructed.
     class BASE_EXPORT ScopedMemoryContext {
      public:
       explicit ScopedMemoryContext(MemoryContext context);
```

copied_base/base/memory/cobalt_memory_context.cc
```diff
@@ -0,0 +1,118 @@
...
+    uint32_t expected = 0xFFFFFFFF;
+    if (g_key.compare_exchange_strong(expected, static_cast<uint32_t>(new_key), std::memory_order_release)) {
+      key = static_cast<uint32_t>(new_key);
+    } else {
+      pthread_key_delete(new_key);
+      key = expected;
+    }
+  }
+  return static_cast<pthread_key_t>(key);
+}
```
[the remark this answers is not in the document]
Awallky: I applied this change to both  copied_base/base/memory/cobalt_memory_context.cc  and  base/memory/cobalt_memory_context.cc .

Here is the diff of what we changed:

```
    --- a/copied_base/base/memory/cobalt_memory_context.cc
    +++ b/copied_base/base/memory/cobalt_memory_context.cc
    @@ -27,23 +27,12 @@                                                                                                                                                                                            
     __attribute__((weak))
     #endif
     pthread_key_t GetSharedMemoryContextKey() {
    -  // Use a static atomic to ensure lazy initialization happens safely.
    -  // Because this function is weak, the linker will merge all copies into a single instance,
    -  // meaning `g_key` will be identical across both `base` and `copied_base`.
    -  static std::atomic<uintptr_t> g_key{UINTPTR_MAX};
    -  uintptr_t key = g_key.load(std::memory_order_acquire);
    -  if (key == UINTPTR_MAX) {
    -    pthread_key_t new_key;
    -    pthread_key_create(&new_key, nullptr);
    -    uintptr_t expected = UINTPTR_MAX;
    -    if (g_key.compare_exchange_strong(expected, static_cast<uintptr_t>(new_key), std::memory_order_release)) {
    -      key = static_cast<uintptr_t>(new_key);
    -    } else {
    -      pthread_key_delete(new_key);
    -      key = expected;
    -    }
    -  }
    -  return static_cast<pthread_key_t>(key);
    +  static pthread_key_t key;
    +  static pthread_once_t once = PTHREAD_ONCE_INIT;
    +  pthread_once(&once, []() {
    +    pthread_key_create(&key, nullptr);
    +  });
    +  return key;
```

johnxwork (approved):
Neat

components/memory_system/parameters.h
```diff
@@ -70,11 +79,20 @@ struct DispatcherParameters {
   explicit DispatcherParameters(
       PoissonAllocationSamplerInclusion poisson_allocation_sampler_inclusion,
       AllocationTraceRecorderInclusion allocation_trace_recorder_inclusion,
-      std::string_view process_type);
+      std::string_view process_type
+#if BUILDFLAG(IS_COBALT)
```
johnxwork, line 83: What's the reason to use kDoNotInclude here? Shall we add a comment here?

copied_base/base/memory/cobalt_memory_context.cc
```diff
@@ -0,0 +1,122 @@
...
+
+// Weak symbol to allow the linker to merge this TLS getter across base and copied_base.
+#if defined(__GNUC__)
+__attribute__((weak))
+#endif
+pthread_key_t GetSharedMemoryContextKey() {
+  // Use a static atomic to ensure lazy initialization happens safely.
+  // Because this function is weak, the linker will merge all copies into a single instance,
+  // meaning `g_key` will be identical across both `base` and `copied_base`.
+  static std::atomic<uint32_t> g_key{0xFFFFFFFF};
```
johnxwork: Sharing the same atomic key across libs is dangerous in principle, but I guess necessary in our case here.

Is it a good idea to try wrapping "base" around "copied_base" to avoid compiling the same code twice? Like we can exclude source files in "copied_base" from base, and have base depend on copied_base. This should reduce binary size and avoid mistakes in the duplication? We can rename "copied_base" to e.g. "base_core" if needed.

components/memory_system/parameters.h
Awallky: kDoNotInclude is used  because this feature requires some overhead, which can inhibit performance. With every allocation, we perform a lookup and add to whichever the allocation is attributed to. 
With that in mind, we will want to roll this feature out with an experiment to ensure user experience is not negatively impacted too much.
This is tied to the feature flag while keeping the lower level components generic and not dependent upon //cobalt.

I added a comment on the enum mentioning the performance overhead.

copied_base/base/memory/cobalt_memory_context.cc
Awallky: Yeah probably. We should do this anyways. Do we want block this PR or feature until this refactor is done? I think it will be a large refactor, and could break multiple platform builds if done improperly.

Awallky: I filed b/524258740 to track this tech debt. This should be safe to land until we go turn this feature on. We can debate whether to refactor this before launching, but let's land this to get to other PRs for this feature.

PR #10851, opened 2026-06-10
android: Prevent partner OS-fallback keys from being dispatched to IME

ckim1346:
This PR introduces a `KeyboardUtils.java` class that defines key events that should not be dispatched/consumed by the IME pipeline and instead left to the OS to handle. Partners reported that their fallback key events are being consumed by Cobalt and therefore not working as intended. This is handled similarly to Gamepad key events #7101. 

Bug: 513111973

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/KeyboardUtils.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/KeyboardUtils.java
@@ -1,0 +1,51 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dev.cobalt.coat;
+
+import android.view.KeyEvent;
+
+/** Utility methods for keyboard event handling. */
+public final class KeyboardUtils {
+  private KeyboardUtils() {}
+
+  /**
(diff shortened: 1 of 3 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/KeyboardUtils.java
```diff
@@ -0,0 +1,51 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dev.cobalt.coat;
+
+import android.view.KeyEvent;
+
+/** Utility methods for keyboard event handling. */
+public final class KeyboardUtils {
```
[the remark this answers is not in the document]
ckim1346: Updated

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/KeyboardUtils.java
```diff
@@ -0,0 +1,63 @@
...
+   * the Android OS, which prevents the OS-level fallback mechanism from triggering.
+   *
+   * Consequently, filtering these keys in Cobalt's native keycode conversion (e.g.
+   * keyboard_code_conversion_android_cobalt.cc) is too late to allow fallback, as the key events will
+   * still be consumed although they are mapped to VKEY_UNKNOWN.
+   */
+  public static boolean isPartnerFallbackKey(int keyCode) {
+    switch (keyCode) {
+      case KeyEvent.KEYCODE_11:
+      case KeyEvent.KEYCODE_12:
```
johnxwork, line 53: does this map to number 5? Can you verify in Kabuki search that all the keys still work after this change?

ckim1346: `KEYCODE_11` and `KEYCODE_12` map to the buttons 11 and 12 directly with integer values of 227 and 228.  `KEYCODE_4` maps to the button 4 and has the integer value of 11, and KEYCODE_5 maps to the button 5 has the integer number of 12. Verified that the buttons still WAI in Kabuki search and also tested them on the YTS guided test  “Key Event Required If Unreserved 2026+" and both still passed as well.

johnxwork (approved):

PR #10861, opened 2026-06-11
Fix canary imbalance

jellefoks:
Initialize musl via static initializer to fix stack canary imbalanc


jellefoks:
init_musl improvement was added to #10841.

PR #10869, opened 2026-06-11
android: Integrate Android JUnit tests into CI

arjungm:
Enable on-host execution of Android JUnit (Robolectric) tests within
the CI pipeline. This allows for automated validation of Android-
specific Java code without requiring physical devices or emulators.

The integration ensures that test results are correctly reported by
converting Chromium test output to standard XML formats. It also
refines the build and archiving logic to package the required JDK and
third-party dependencies. Sparse checkout support is added to the CI
environment to efficiently fetch these prerequisites and fix path
resolution issues. GN for robolectric binary is extended to generate
runtime deps files to ensure the archive is complete.

Bug: 431780245


.github/actions/on_host_tests/action.yaml
```diff
@@ -117,7 +141,86 @@ runs:
...
+
+
+        # 3. Expose the Git sparse checkout roots to the extracted unit_test/ workspace
+        ln -sf "${GITHUB_WORKSPACE}/testing" unit_test/testing
+        ln -sf "${GITHUB_WORKSPACE}/build" unit_test/build
+        ln -sf "${GITHUB_WORKSPACE}/cobalt" unit_test/cobalt
+        ln -sf "${GITHUB_WORKSPACE}/starboard" unit_test/starboard
+        ln -sf "${GITHUB_WORKSPACE}/.github" unit_test/.github
+        ln -sf "${GITHUB_WORKSPACE}/third_party/catapult" unit_test/third_party/catapult
+        ln -sf "${GITHUB_WORKSPACE}/third_party/logdog" unit_test/third_party/logdog
```
arjungm: this whole section is a bit puzzling because it seems like we're:

1. symlinking to `unit_test/third_party` subdirs from `${GITHUB_WORKSPACE}/third_party`
2. symlinking to `${GITHUB_WORKSPACE}/third_party` subdirs to `unit_test/third_party`

I'm not sure why but it might be to ensure that the relative paths for everything else works? this is very bizarre ...

.github/actions/on_host_tests/action.yaml
```diff
@@ -65,9 +69,22 @@ runs:
         # This also prevents accidental overwriting when collecting results.
         test_output="${results_dir}/shard_${{ matrix.shard }}"
         mkdir -p ${test_output}
+        echo "test_output=${test_output}" >> $GITHUB_ENV
+
```
oxve, line 72: nit: extra newline

.github/actions/on_host_tests/action.yaml
```diff
@@ -27,34 +31,34 @@ runs:
       uses: actions/download-artifact@v5
       with:
         name: ${{ inputs.test_artifacts_key }}
+    - name: Checkout JUnit Dependencies
+      if: ${{ inputs.test_type == 'junit' }}
+      shell: bash
+      run: |
+        git config --global --add safe.directory "${GITHUB_WORKSPACE:-/github/workspace}"
+        git config --global --add safe.directory "*"
```
oxve: This seems a little aggressive. In [.github/actions/checkout/action.yaml](https://github.com/youtube/cobalt/blob/main/.github/actions/checkout/action.yaml) we only tag `${GITHUB_WORKSPACE}/${DEST_PATH}` which is the checkout path.

.github/actions/on_host_tests/action.yaml
```diff
@@ -27,34 +31,34 @@ runs:
...
+        git sparse-checkout add --skip-checks testing build third_party/catapult third_party/logdog
```
oxve, line 39: These are not declared deps in GN?

.github/actions/on_host_tests/action.yaml
```diff
@@ -117,7 +135,65 @@ runs:
...
+          fi
+
+          # Run JUnit tests and generate JSON report, then convert to XML.
+          #
+          # NOTE: Even though these are junit tests, due to upstream implementation of the Robolectric tests they are
+          # only able to output JSON results. The rest of our CI pipeline is expecting XML so in the short term we are
+          # taking a shortcut to convert JSON to XML for expediency. Given that the Robolectric JSON output is the how
+          # Chromium infrastructure handles test results in general, we will eventually refactor the gtest results to
+          # also output JSON and then subsequently remove this conversion script after adopting JSON as a universal
+          # test result format in our testing workflow.
```
oxve, line 184: Could you put this in a bug and add a TODO here?

.github/actions/on_host_tests/action.yaml
```diff
@@ -117,7 +135,65 @@ runs:
...
           fi
         done
-        echo "Finished running tests..."
+
+        echo "Finished running GTest tests..."
+        if [[ -n "${failed_suites}" ]]; then
+          echo "Test suites failed:${failed_suites}"
+          # Fail the job so it's easy to retrigger.
+          exit 1
+        fi
```
oxve, line 142: nit: Missing separating newline

build/android/test_runner.py
oxve, line 1: This is a Chromium file. I don't know if we should be changing this here. Is this change necessary for getting the tests running?

build/config/android/rules.gni
oxve, line 1: These changes should be guarded by some `if is_cobalt` construct since it's Chromium code.

cobalt/android/BUILD.gn
oxve, line 1: `if is_cobalt` here too

cobalt/docker/unittest/Dockerfile
```diff
@@ -2,6 +2,8 @@ ARG BASE_IMAGE_SHA=sha256:be15f84c44c4ee68b4f187128f0278df1b0f424c04fb5f08b09896
 
 FROM marketplace.gcr.io/google/debian12@${BASE_IMAGE_SHA}
 
+ENV DEBIAN_FRONTEND=noninteractive
```
oxve: Nit/question: Is this line necessary?

cobalt/build/archive_test_artifacts.py
```diff
@@ -141,6 +155,14 @@ def create_archive(
       if os.path.exists(test_targets_json):
         target_deps.add(os.path.join(tar_root, 'test_targets.json'))
 
+      # Add JUnit wrapper scripts if they exist
```
oxve: For gtest targets I've seen the wrapper automatically getting included in the runtime deps. Are you sure these are required to add manually?

cobalt/build/archive_test_artifacts.py
```diff
@@ -23,11 +23,8 @@
 
 # Path prefixes that contain files we don't need to run tests.
 _EXCLUDE_DIRS = [
```
oxve: Can you verify that the android unit test deps archives do not include these? I recall adding them to avoid ballooning them.

cobalt/build/archive_test_artifacts.py
```diff
@@ -111,12 +109,18 @@ def create_archive(
...
       continue
     target_path, target_name = target.split(':')
+    target_path = target_path.lstrip('/')
     # Paths are configured in test.gni:
     # https://github.com/youtube/cobalt/blob/main/testing/test.gni
     if use_android_deps_path:
       deps_file = os.path.join(
           out_dir, 'gen.runtime', target_path,
           f'{target_name}__test_runner_script.runtime_deps')
+      if not os.path.exists(deps_file):
```
oxve: I'm not too fond of this. There was one `if not os.path.exists(deps_file)` before this change. Now there are 5. This must be possible to simplify.

Perhaps add another flag for junit? Or figure out if we can/should add the suffix for junit tests.

.github/actions/on_host_tests/action.yaml
```diff
@@ -27,34 +31,34 @@ runs:
       uses: actions/download-artifact@v5
       with:
         name: ${{ inputs.test_artifacts_key }}
+    - name: Checkout JUnit Dependencies
+      if: ${{ inputs.test_type == 'junit' }}
```
oxve: nit: `${{` are optional in `if:` statements. It makes it slightly harder so we've excluded in in most places.

.github/actions/on_host_tests/action.yaml
```diff
@@ -27,34 +31,34 @@ runs:
       uses: actions/download-artifact@v5
       with:
         name: ${{ inputs.test_artifacts_key }}
+    - name: Checkout JUnit Dependencies
```
oxve, line 34: We set up the sparse checkout in `main.yaml` so I think I would like to see this happening as soon as possible after that. I'm not feeling 100% sure on this so argue against me if you don't agree.

oxve (commented):
The approach looks good. Just some details to iron out.

.github/actions/on_host_tests/action.yaml
arjungm: I checked the runtime deps file and they're not present: https://paste.googleplex.com/5437630572527616

It seems like it doesn't generate this for the test_runner_script wrappers

build/android/test_runner.py
arjungm: yes it's necessary, the problem is because it only applies this argument for Gtest type tests but it's passed by the test wrapper script code when we re-enabled the runtime deps generation

cobalt/build/archive_test_artifacts.py
arjungm: it seems like they do increase in size. I tested with base_unittests and it went from from 47 MB to 52 MB

cobalt/build/archive_test_artifacts.py
arjungm: this is the example runtime deps file generated for the base_unittests and I see any wrapper scripts in there: https://paste.googleplex.com/5280624050044928

Without these lines for junit tests however it fails to package them and then it doesn't work. I'm thinking that this is once again another special case because of how the robolectric binaries are built in our code base. There seems to be some unfinished migration work on these targets.

arjungm: additionally, for `base_unittests` we see only the `out/android/bin/run_base_unittests` script, whereas for `cobalt_coat_junit_tests` we see both:

```
out/android/bin/run_cobalt_coat_junit_tests
out/android/bin/helper/run_cobalt_coat_junit_tests
```

cobalt/build/archive_test_artifacts.py
arjungm: added some logic to ensure the old exclude dirs are applied when it's not a junit test

build/android/test_runner.py
arjungm: I'm moved it to the Junit section to miminize the fallout to other test infra

.github/actions/on_host_tests/action.yaml
arjungm: I'm not sure where I would move this since we need the test_type to be parsed from the test_targets.json prior to being able to gate whether this step runs. I could put it in the main job steps, but then that separates it from where it would be most relevant making for more overhead to track down what earlier step failed. I think having it here is fine given that it's such a specific edge case for Junit tests.

.github/actions/on_host_tests/action.yaml
arjungm: spoke offline but this could be fixed by having on host test runnersuse the overlay and runner cache with gclient sync steps

cobalt/build/archive_test_artifacts.py
arjungm: captured some of this in this refactoring GN for Robolectric bug: 524712602

build/android/test_runner.py
arjungm, line 1: @johnxwork @Awallky PTAL at this android specific changes

build/config/android/rules.gni
arjungm, line 1: @johnxwork @Awallky PTAL at this android specific changes

johnxwork (approved):
For Android specific test code, it looks like the PR only pipes new variables and should be safe. Fingers crossed everything works

Awallky (approved):
As long as junit tests run and pass, LGTM

PR #10886, opened 2026-06-11
Posix: Add nplb tests for `fstatat`

sacuff:
fstatat has already been implemented as a Starboard wrapper, but did not have unit tests until now.

Bug: 474323894


andrewsavage1 (approved):
The gemini suggestions make sense to me

PR #10887, opened 2026-06-12
Update autoroller to support merge conflicts.

briantting:
The autoroller will now create PRs containing cherry-picks up to but not including the conflicting cherry-pick. If the conflicting cherry-pick is the first cherry-pick, the autoroller will create a Conflicted PR containing just that conflicting cherry-pick to then be handled by devs.

Bug: 522987128

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -2,6 +2,7 @@
 """Script to automatically roll LTS branch."""
 import argparse
 from collections import defaultdict
+import enum
 import re
 import subprocess
 import sys
@@ -33,6 +34,19 @@ _SKIP_LIST = {
         'f0ef063d30589b39dc4a1433b0dde659d2fb5eea',
     ],
 }
+
+
+@enum.unique
+class CherryPickStatus(enum.Enum):
+  """Represents the outcome of a cherry-pick attempt."""
+  SUCCESS = 'success'  # Successfully cherry-picked and committed.
+  CONFLICTED = 'conflicted'  # Cherry-picked and committed with conflicts.
+  SKIPPED = 'skipped'  # The commit was already present or no action was needed.
+  FAILED = 'failed'  # The cherry-pick failed due to conflicts or other errors.
+
+
+def run(cmd):
+  subprocess.run(cmd, check=True, stdout=sys.stderr)
 
 
 def get_out(cmd):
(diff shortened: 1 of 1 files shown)


briantting:
Devs should pull down this Conflicted PR and resolve the conflicts in a separate commit with the exact same formatted commit message minus "Conflicted". These 2 commits should NOT be squashed together as the conflict and resolution history needs to be preserved.

.github/scripts/autoroll_lts.py
```diff
@@ -173,26 +193,32 @@ def cherry_pick(sha, num, title):
...
     subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
-  except subprocess.CalledProcessError as error:
+  except subprocess.CalledProcessError:
     unmerged = get_unmerged_files()
     if not resolve_conflicts(unmerged):
-      subprocess.run(['git', 'reset', '--hard', 'HEAD'], check=True)
-      raise error
+      if first:
+        msg = f'Conflicted {msg}'
+        subprocess.run(['git', 'add'] + list(unmerged), check=True)
```
oxve: Can you add a new function that wraps all subprocess.run commands?

They MUST have `stdout=sys.stderr` as the script output is used for the PR description.

.github/scripts/autoroll_lts.py
```diff
@@ -35,6 +36,15 @@
 }
 
 
+@enum.unique
+class CherryPickStatus(enum.Enum):
+  """Represents the outcome of a cherry-pick attempt."""
+  SUCCESS = 'success'  # Successfully cherry-picked and committed.
+  CONFLICTED = 'conflicted'  # Cherry-picked and committed with conflicts.
+  SKIPPED = 'skipped'  # The commit was already present or no action was needed.
+  FAILED = 'failed'  # The cherry-pick failed due to conflicts or other errors.
```
andrewsavage1, line 45: In what case is there a failed cherry pick due to conflicts that wouldn't be under CONFLICTED?

.github/scripts/autoroll_lts.py
```diff
@@ -173,26 +193,32 @@ def cherry_pick(sha, num, title):
...
+  result = CherryPickStatus.SUCCESS
   try:
     subprocess.run(cmd + [sha], check=True, stdout=sys.stderr)
-  except subprocess.CalledProcessError as error:
+  except subprocess.CalledProcessError:
     unmerged = get_unmerged_files()
     if not resolve_conflicts(unmerged):
-      subprocess.run(['git', 'reset', '--hard', 'HEAD'], check=True)
-      raise error
+      if first:
```
andrewsavage1: nit: I slightly prefer

```
      if not first:
        subprocess.run(['git', 'reset', '--hard', 'HEAD'], check=True)
        return CherryPickStatus.FAILED

      msg = f'Conflicted {msg}'
      subprocess.run(['git', 'add'] + list(unmerged), check=True) 
      result = CherryPickStatus.CONFLICTED
```
to differentiate between an early return and a continuation, whereas them being in parallel right now but only one having an early return is more confusing imo

.github/scripts/autoroll_lts.py
```diff
@@ -143,11 +159,15 @@ def resolve_conflicts(unmerged):
   return False
 
 
-def cherry_pick(sha, num, title):
+def cherry_pick(sha, num, title, first):
```
andrewsavage1: `first` is an ambiguous name, I would go with allow_conflict or commit_with_conflicts

.github/scripts/autoroll_lts.py
briantting: CONFLICTED and FAILED are both the same in that it's a cherry-pick failed with conflicts.

What's different is that in the CONFLICTED state we're committing/saving the conflicts into the PR and asking devs to make a separate commit in that PR resolving them. This only happens when it's the first PR in the chain of PRs.

The FAILED state is when we just abort the process to be able to merge all of the non-conflicting PRs before it.

oxve (approved):

PR #10896, opened 2026-06-12
Refactor Privacy Sandbox and Milestone macro gating to minimize Chromium changes

linxinan-yt:
Refactors how the `ENABLE_PRIVACY_SANDBOX_APIS` and `CHROMIUM_MILESTONE` macros are defined and propagated to C++ to minimize modifications to Chromium-owned files.

Initially I defined them in content/common:
- content/public/common/features.gni
- content/public/common/content_milestone_features.h.in

However as I have been using them to gate the privacy sandbox components, issues surfaced:

All source files need to include the generated header to access the milestone macro:

```
#include "content/public/common/content_milestone_features.h"
#include "content/public/common/buildflags.h"
```
To make them avoid rebase conflicts, I defined them in their own block but then this further requires below to escape the clang-format.

```
// clang-format off
...
// clang-format on
```

Even worse, to gate code in third_party I has to include "content/public/common/content_milestone_features.h" to third_party/blink, which violates Chromium's rule on dependency. 

This PR solved it by move `ENABLE_PRIVACY_SANDBOX_APIS` and `CHROMIUM_MILESTONE` to a global arg like `is_cobalt`, so that we can apply them to gate arbitrary privacy sandbox code in Cobalt, without(mostly) including the header files.

Bug: 505811196


cobalt/build/configs/cobalt.gni
```diff
@@ -17,3 +17,17 @@ declare_args() {
   is_cobalt = false
   cobalt_is_release_build = false
 }
+
+declare_args() {
```
linxinan-yt: This should be true. Also add comment.

content/browser/browser_interface_binders.cc
```diff
@@ -208,6 +202,7 @@
 #include "third_party/blink/public/mojom/webtransport/web_transport_connector.mojom.h"
 #include "third_party/blink/public/mojom/worker/dedicated_worker_host_factory.mojom.h"
 #include "third_party/blink/public/mojom/worker/shared_worker_connector.mojom.h"
+#include "third_party/blink/public/public_buildflags.h"
```
linxinan-yt, line 205: remove

content/public/common/BUILD.gn
```diff
@@ -26,20 +26,10 @@ buildflag_header("buildflags") {
...
-# and helper macros. This enables Cobalt to conditionally compile custom
-# features, fallbacks, or modifications on older milestones (M138 or lower),
-# while automatically yielding to upstream's official implementations on newer
-# milestones (M139+) to prevent rebase and upgrade merge conflicts.
-process_version("milestone_features") {
-  sources = [ "//chrome/VERSION" ]
-  template_file = "content_milestone_features.h.in"
-  output = "$target_gen_dir/content_milestone_features.h"
-}
+
```
linxinan-yt: remove

content/public/common/BUILD.gn
```diff
@@ -269,7 +259,7 @@ source_set("common_sources") {
     ":main_function_params",
     ":renderer_type",
     ":switches",
-    ":milestone_features",
+
```
linxinan-yt: remove empty line

.gn
```diff
@@ -170,6 +170,7 @@ exec_script_allowlist =
     ]
 
 # TODO(b/378960295): Try to remove the allowlisted files.
+exec_script_allowlist += [ "//cobalt/build/configs/cobalt.gni" ]
```
linxinan-yt: Remove to gn.py.

cobalt/build/configs/cobalt.gni
```diff
@@ -17,3 +17,17 @@ declare_args() {
...
+
+chromium_milestone =
+    exec_script("//build/util/version.py",
+                [
+                  "-f",
+                  rebase_path("//chrome/VERSION", root_build_dir),
+                  "-t",
+                  "@MAJOR@",
+                ],
+                "json")
```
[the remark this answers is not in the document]
linxinan-yt: @gemini-code-assist, I already removed this impl.

content/browser/browser_interface_binders.cc
linxinan-yt: #include "third_party/blink/public/public_buildflags.h" was originally here. This PR restores it initial position.

johnxwork (approved):
Much cleaner this way!

haozheng-cobalt:
Nice, is this removing the old customization you've added for feature removal?

cobalt/build/configs/cobalt.gni
```diff
@@ -16,4 +16,20 @@
...
+  # The build time value is assigned by gn.py when running gn gen. Set
+  # default to 138 to minimize the impact for 27.lts development.
+  chromium_milestone = 138
+
+  # This argument is not present in upstream Chromium, so we must declare it
+  # here in a `.gni` file before it can be used in other GN files.
+  # We keep it enabled (true) for now until all components are gated.
+  # TODO(b/505811196): Remove this argument once Cobalt rebased to the milestone
+  # where the privacy sandbox components are fully removed.
+  enable_privacy_sandbox_apis = true
```
haozheng-cobalt, line 25: Should move these args to cobalt/build/configs/common.gn.

content/browser/webui/content_web_ui_configs.cc
```diff
@@ -7,9 +7,14 @@
...
-// Remove these two includes after CHROMIUM_MILESTONE_LE_138
-#include "content/public/common/buildflags.h"
-#include "content/public/common/content_milestone_features.h"
+// In Cobalt, we gate Privacy Sandbox WebUIs (like Attribution Reporting).
+// Gating these includes breaks the transitive include chain that previously
+// provided `buildflag.h` and `build_config.h` for the platform checks below.
+// We include them explicitly here to ensure they are defined.
+// TODO: If Privacy Sandbox is removed upstream, this custom gating and these
+// includes should be ignored/removed during rebase.
+#include "build/build_config.h"
```
haozheng-cobalt, line 16: Are both header inclusions needed? if not, would be better to just keep one.

build/build_config.h
```diff
@@ -283,6 +283,12 @@
 #define BUILDFLAG_INTERNAL_IS_COBALT() (0)
 #endif
 
+#if defined(ENABLE_BUILDFLAG_ENABLE_PRIVACY_SANDBOX_APIS)
```
haozheng-cobalt, line 289: I don't think we want to create global build flag for any feature removal work right? @andrewsavage1

cobalt/build/configs/cobalt.gni
linxinan-yt: We cannot declare new build arguments in cobalt/build/configs/common.gn because common.gn is an override file (imported in args.gn), not a .gni template. In GN, any new build argument must first be declared inside a declare_args() block in a .gni file (like cobalt.gni) before it can be assigned or overridden. If we set it in common.gn without declaring it first, GN will fail with an Unknown variable error.

(This is different from arguments like enable_widevine in common.gn, which we can override there because they are already declared by upstream Chromium).

content/browser/webui/content_web_ui_configs.cc
linxinan-yt: Need both because they serve different purposes:

1. build/buildflag.h defines the core BUILDFLAG() macro itself (e.g., how the macro expands). Without it, the compiler won't recognize BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) as a valid macro.
1. build/build_config.h is where we defined the value of our specific flag:

```cpp
#if defined(ENABLE_BUILDFLAG_ENABLE_PRIVACY_SANDBOX_APIS)
#define BUILDFLAG_INTERNAL_ENABLE_PRIVACY_SANDBOX_APIS() (1)
#else
#define BUILDFLAG_INTERNAL_ENABLE_PRIVACY_SANDBOX_APIS() (0)
#endif
```

Without build_config.h, the compiler won't know whether ENABLE_PRIVACY_SANDBOX_APIS is 0 or 1.

build/build_config.h
linxinan-yt: I tried to put them in cobalt/build/configs but then our chromium build failed.

https://github.com/youtube/cobalt/actions/runs/27446847711/job/81133823118

 To keep the content_shell green, they have to be a global build flag here.

linxinan-yt: Added a TODO. Once PS are fully removed from Chromium. We don't have to keep these lines in the future rebase. Per the chat thread, I think this is ok to Andrew.  Close this comment.

haozheng-cobalt (approved):

cobalt/build/configs/shared/BUILD.gn
```diff
@@ -0,0 +1,13 @@
+# Copyright 2026 The Cobalt Authors. All Rights Reserved.
```
andrewsavage1, line 1: Add our full copyright header

cobalt/build/gn.py
```diff
@@ -85,7 +85,23 @@ def write_build_args(build_args_path, platform_args_path, build_type, use_rbe,
...
   """ Write args file, modifying settings for config"""
   gen_comment = '# Set by gn.py'
+
+  # Read milestone from chrome/VERSION
+  script_dir = Path(__file__).resolve().parent
+  src_dir = script_dir.parents[1]
+  version_path = src_dir / 'chrome' / 'VERSION'
+
+  milestone = None
+  if version_path.exists():
```
andrewsavage1: I'm against this change, as the gn.py wrapper should be a very thin script that simply adds arguments to an args.gn file and invokes gn gen—introducing parsing in it opens it up to much more, and I think it should instead be done in a GN action/exec_script

andrewsavage1 (commented):
Reviewed only the gn code

cobalt/build/gn.py
linxinan-yt: I see. I prefer this because I don't have to add action/exec_script to the root allowlist. If you gn gen would open it up to much more I fallback to my earlier implementation.

linxinan-yt: Done.

cobalt/build/configs/shared/BUILD.gn
linxinan-yt: Done.

cobalt/build/configs/shared/BUILD.gn
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import("//cobalt/build/configs/cobalt.gni")
+
+config("global_defines_shared") {
+  defines = [
+    "CHROMIUM_MILESTONE=$chromium_milestone",
```
andrewsavage1, line 25: The gn arg for chromium_milestone is only used in this file, so I think we should declare it and calculate the default in this file too

andrewsavage1 (approved):

cobalt/build/configs/shared/BUILD.gn
linxinan-yt: Genius idea! Even cleaner now. Thanks!

PR #10898, opened 2026-06-12
Add kInMemoryUpdatesMemoryBuffer feature to configure update client memory buffer size

yuying-y:
Define the kInMemoryUpdatesMemoryBuffer feature and parameter and use it to check available memory before downloading an update.

Issue: 479816505

### DIFF
--- a/components/update_client/DEPS
+++ b/components/update_client/DEPS
@@ -13,12 +13,15 @@ include_rules = [
   "+third_party/zlib",
 ]
 
-# Allow the unit tests to depend on the network for testing purposes.
 specific_include_rules = {
+  # Allow the unit tests to depend on the network for testing purposes.
   "(test_configurator|.*_unittest\.(cc|h))" : [
     "+net",
     "+services/network/public/cpp",
     "+services/network/public/mojom",
     "+services/network/test",
-  ]
+  ],
+  "op_download\.cc": [
+    "+cobalt/browser/features.h",
+  ],
 }
--- a/components/update_client/op_download.cc
+++ b/components/update_client/op_download.cc
@@ -35,6 +35,8 @@
 
 #if BUILDFLAG(IS_STARBOARD)
 #include <algorithm>
+#include "base/feature_list.h"
+#include "cobalt/browser/features.h"  // nogncheck
 #include "starboard/system.h"  // nogncheck
 #endif
 
@@ -47,11 +49,6 @@ namespace {
 constexpr int64_t kBackgroundDownloadSizeThreshold = 10'000'000; /*10 MB*/
 #elif !BUILDFLAG(IS_STARBOARD)
 constexpr int64_t kBackgroundDownloadSizeThreshold = 0;
-#endif
-
-#if defined(IN_MEMORY_UPDATES)
-// Memory buffer required for downloading the update in memory.
-constexpr int64_t kMemoryBufferBytes = 20 * 1024 * 1024;
 #endif
 
 #if !BUILDFLAG(IS_STARBOARD)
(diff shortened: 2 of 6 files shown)


components/update_client/op_download.cc
```diff
@@ -42,6 +47,9 @@ namespace {
 constexpr int64_t kBackgroundDownloadSizeThreshold = 10'000'000; /*10 MB*/
 #elif !BUILDFLAG(IS_STARBOARD)
 constexpr int64_t kBackgroundDownloadSizeThreshold = 0;
+#elif defined(IN_MEMORY_UPDATES)
+// Memory buffer required for downloading the update in memory.
+constexpr int64_t kMemoryBufferBytes = 20 * 1024 * 1024;
 #endif
```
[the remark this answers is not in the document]
yuying-y: Fixed by separating the preprocessor definitions. kMemoryBufferBytes is now defined in its own independent #if defined(IN_MEMORY_UPDATES) block. Thank you!

components/update_client/op_download.cc
```diff
@@ -187,6 +197,26 @@ void HandleAvailableSpace(
                  .code = static_cast<int>(CrxDownloaderError::DISK_FULL)})));
     return;
   }
+
+#if BUILDFLAG(IS_STARBOARD) && defined(IN_MEMORY_UPDATES)
+  int64_t total_memory = SbSystemGetTotalCPUMemory();
+  int64_t used_memory = SbSystemGetUsedCPUMemory();
+  int64_t available_memory = std::max(int64_t{0}, total_memory - used_memory);
```
andrewsavage1, line 201: I'd be curious to hear @sherryzy comment on if we want to have more buffer than this or if this is enough and it's better to keep complexity to a minimum

sherryzy: I was wondering when this download usually occurs. Do we check available memory at app startup or at another point? This is quite tricky, as memory can easily fluctuate by over 100 MB depending on user behavior.

A 20 MB buffer may not be sufficient; for example, if we check this value at app startup before any video is played, a 1080p stream can easily use ~30 MB, while 4K can exceed 100 MB. To keep complexity to a minimum, perhaps we could set the buffer size to at least 35 MB, especially considering that 4K content typically has a lower watch time.

yuying-y: This check happens exactly at the point when a download is about the start. The first update check happens 30s after app start. Sure, updated it to 35MB.

components/update_client/op_download.cc
```diff
@@ -187,6 +197,26 @@ void HandleAvailableSpace(
...
+    VLOG(1)
+        << "Insufficient memory for the update plus 20MB buffer. Available memory: "
+        << available_memory << ", download size: " << size;
+    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE,
+        base::BindOnce(
+            std::move(callback),
+            base::unexpected<CategorizedError>(
+                {.category = ErrorCategory::kDownload,
+                 .code = static_cast<int>(CrxDownloaderError::OUT_OF_MEMORY)})));
```
andrewsavage1, line 221: Where does this information get logged? Is there a UMA histogram or similar? If not, maybe there should be

components/update_client/op_download.cc
```diff
@@ -44,6 +49,11 @@ constexpr int64_t kBackgroundDownloadSizeThreshold = 10'000'000; /*10 MB*/
 constexpr int64_t kBackgroundDownloadSizeThreshold = 0;
 #endif
 
+#if defined(IN_MEMORY_UPDATES)
+// Memory buffer required for downloading the update in memory.
```
andrewsavage1: Expand on this comment a bit to make it clear that some of this buffer is extra in case the memory used elsewhere fluctuates as the download is happening

components/update_client/update_client_errors.h
```diff
@@ -49,6 +49,9 @@ enum class CrxDownloaderError {
   DISK_FULL = 13,
   CANCELLED = 14,
   NO_DOWNLOAD_DIR = 15,
+#if BUILDFLAG(IS_STARBOARD)
+  OUT_OF_MEMORY = 16,
```
andrewsavage1: What's the reasoning behind having one error we're adding be before the others (SLOT_UNAVAILABLE=9) and another be after the basic errors? Wouldn't we prefer to keep them together

components/update_client/op_download.cc
yuying-y: It's sent back to the Omaha update server, then we monitor.

andrewsavage1: Okay, and if I'm understanding correctly there's some dashboard or metrics about how many updates failed and in what way?

yuying-y: yes http://go/evergreen-update-metrics

components/update_client/DEPS
```diff
@@ -15,6 +15,9 @@ include_rules = [
 
 # Allow the unit tests to depend on the network for testing purposes.
```
andrewsavage1: Move this comment or the location of the include rule being added, it's confusing like this

components/update_client/op_download.cc
```diff
@@ -44,6 +51,8 @@ constexpr int64_t kBackgroundDownloadSizeThreshold = 10'000'000; /*10 MB*/
 constexpr int64_t kBackgroundDownloadSizeThreshold = 0;
 #endif
 
+
```
andrewsavage1: nit: delete

andrewsavage1 (approved):

PR #10900, opened 2026-06-12
Fix map lookup crash in UpdateEngine::Cancel

yuying-y:
In Cobalt 25, UpdateEngine::Cancel uses update_contexts_.at(session_id)
to look up the update context. If the update context has already completed
and been erased from update_contexts_, this lookup throws a std::out_of_range
exception, causing a crash.

This CL checks if the session_id exists in update_contexts_ before
accessing it. If not found, it logs a warning and returns early.

Issue: 477394770
Change-Id: Ib76cada0eac19d9d2ce7741509e862dcae7c9810


hlwarriner (approved):
I agree with the comment about `auto` but otherwise this lgtm. Thanks for the fix!

oxve (approved):

PR #10905, opened 2026-06-15
Productionize tvOS presubmit pipeline

linxinan-yt:
Update tvOS presubmit pipeline to respect the test target and filter configurations like other platforms.
In this initial setup, the pipeline will build and run:
- base:base_unittests
- base:base_perftests (currently disabled all tests, so only run empty suite)
- cobalt_browsertests

Bug: 519669027


.github/workflows/tvos.yaml
```diff
@@ -95,8 +95,32 @@
           echo "Timeout waiting for Kokoro build."
           exit 1
 
+  get-targets:
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-targets.outputs.targets }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
```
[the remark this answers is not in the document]
linxinan-yt: Follow the convention in `.github`

.github/actions/process_test_results/action.yaml
```diff
@@ -49,7 +49,11 @@ runs:
 
         # Enable globstar shell option for globstar paths e.g. /**/*.xml and
         # nullglob to return empty string if the glob matches no files.
-        shopt -s globstar nullglob
+        shopt -s nullglob
+        # globstar is not supported on macOS default bash (3.2). We attempt to enable it
+        # but use || true to prevent the script from exiting on macOS, while still
+        # enabling it on Linux (where it is supported but disabled by default).
+        shopt -s globstar 2>/dev/null || true
```
oxve: globstar is necessary to find sharded test results. Currently we don't shard any of these tests but this fix may bite us later.

Perhaps it would be better to use another way to find the result files that work platform independent?

.github/workflows/tvos.yaml
```diff
@@ -95,8 +95,32 @@ jobs:
...
+        with:
+          sparse-checkout: |
+            cobalt/build/testing/targets/tvos-arm64-simulator/test_targets.json
+          sparse-checkout-cone-mode: false
+
+      - name: Extract Targets
```
oxve: There's a composable action to use to parse the test target json file. Can you use it instead of manually parsing the test_targets.json file?

.github/workflows/tvos.yaml
```diff
@@ -95,8 +95,32 @@ jobs:
...
+          json_file="cobalt/build/testing/targets/tvos-arm64-simulator/test_targets.json"
+          if [[ ! -f "${json_file}" ]]; then
+            echo "ERROR: ${json_file} not found!"
+            exit 1
+          fi
+          targets=$(jq -cr '[.[] | select(.target != null) | .target | split(":")[-1]]' "${json_file}")
+          echo "Targets found: ${targets}"
+          echo "targets=${targets}" >> "$GITHUB_OUTPUT"
+
   download-artifacts:
```
oxve, line 124: I think this step should be removed. Instead inline the gcs download in the test step below. There's no specific reason we need the test artifacts in GH artifacts.

.github/workflows/tvos.yaml
```diff
@@ -105,42 +129,114 @@ jobs:
...
+              echo "skipped=true" >> "$GITHUB_OUTPUT"
+            else
+              echo "skipped=false" >> "$GITHUB_OUTPUT"
+            fi
+          else
+            echo "skipped=false" >> "$GITHUB_OUTPUT"
+          fi
+        shell: bash
+
+      - name: Run Test in Simulator
```
oxve, line 192: Would it make sense to add this step to the devel build in the presubmit step?

.github/workflows/tvos.yaml
```diff
@@ -95,8 +95,32 @@ jobs:
...
+  get-targets:
+    runs-on: ubuntu-latest
+    outputs:
+      targets: ${{ steps.set-targets.outputs.targets }}
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v4
+        with:
+          sparse-checkout: |
+            cobalt/build/testing/targets/tvos-arm64-simulator/test_targets.json
```
oxve, line 53: Do kokoro load the test targets from this file too?

.github/workflows/tvos.yaml
linxinan-yt: We have to, because we don't have the service account credentials on the mac. Without a service account authentication, it cannot download the artifacts from GCS(we cannot publish the bucket). 
And per my understanding, if we could deploy credentials on mac runner, we could enable siso on mac runner directly.

.github/workflows/tvos.yaml
linxinan-yt: Do you mean move it to workflow/main.yaml? Which presubmit step do you mean?

.github/workflows/tvos.yaml
linxinan-yt: Yes. See cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh. Kokoro also extract the build targets from here.

.github/workflows/tvos.yaml
linxinan-yt: You mean .github/actions/test_targets_json/action.yaml, right? Let me try.

.github/actions/process_test_results/action.yaml
linxinan-yt: Ack. I added `if [[ "$OSTYPE" != "darwin"* ]]; then` to guard non-Mac builds.

.github/workflows/tvos.yaml
linxinan-yt: Awesome idea! Just added a new flag so that I don't need the entire checkout.

Thanks a ton.

.github/workflows/tvos.yaml
```diff
@@ -96,7 +122,7 @@ jobs:
           exit 1
 
   download-artifacts:
-    needs: [get-sha, wait-for-kokoro-build]
+    needs: [get-sha, wait-for-kokoro-build, get-targets]
     runs-on: [self-hosted, chrobalt-linux-runner]
```
oxve, line 127: I think `chrobalt-linux-runner` is unnecessarily beefy to just act as an artifact downloader. The `odt-runner` is more suited for this job.

Can you add a TODO to investigate using odt-runner instead?

.github/actions/test_targets_json/action.yaml
```diff
@@ -22,6 +27,7 @@ runs:
       env:
         TEST_TYPE: ${{ inputs.test_type }}
         RUNS_ON: ${{ inputs.runs_on }}
+        JSON_FILE: ${{ inputs.test_targets_json_file || 'test_targets.json' }}
```
oxve: Can you move `'test_targets.json'` to be the default value in the parameter instead?

.github/actions/test_targets_json/action.yaml
```diff
@@ -9,10 +9,15 @@ inputs:
...
     default: ""
+  test_targets_json_file:
+    description: "Path to the test targets JSON file. If provided, skip downloading."
+    required: false
+    default: ""
 runs:
   using: "composite"
   steps:
     - name: Download Test Targets Json
+      if: ${{ inputs.test_targets_json_file == '' }}
```
oxve: nit: `${{ }}` are optional in `if:` conditionals and generally skipped in the rest of the codebase.

.github/workflows/tvos.yaml
```diff
@@ -105,42 +131,114 @@ jobs:
...
+            .github
+            cobalt/tools
 
       - name: Download from GitHub Actions
         uses: actions/download-artifact@v4
         with:
           name: tvos-test-artifacts
 
-      - name: Run Cobalt Browser Tests in Simulator
+      - name: Filter
```
oxve, line 174: The logic below is copying logic from the main.yaml workflow and the on-host test composable action.

I would really like to see the tvos pipeline getting integrated properly in that pipeline, at least using the on-host test action to handle target calculation etc.

Let's leave everything like this for now but we need to discuss how the tvos pipeline is going to work in the future and how to best integrate it in CI without having a completely separate track.

oxve (approved):

.github/actions/test_targets_json/action.yaml
linxinan-yt: Nice catch! Much cleaner now.

.github/actions/test_targets_json/action.yaml
linxinan-yt: Ack.

.github/workflows/tvos.yaml
linxinan-yt: Make sense.

.github/workflows/tvos.yaml
linxinan-yt: Agree, but the scope may exceed this PR. We can take some time to think how to integrate them in with one workflow config.
Added a TODO here.

.github/workflows/tvos.yaml
linxinan-yt: Done.

.github/actions/test_targets_json/action.yaml
linxinan-yt: Done.

.github/actions/test_targets_json/action.yaml
linxinan-yt: Done.

.github/workflows/tvos.yaml
linxinan-yt: I have added a TODO for future refactoring.

PR #10908, opened 2026-06-15
android: Populate gl_implementation_parts in-process path

ckim1346:
`GpuInit::InitializeInProcess` did not populate `gl_implementation_part` in GPUInfo for the Android path. This caused diagnostics (such as chrome://gpu) to report `gl=none,angle=none` even though hardware GL and ANGLE were successfully initialized and active.                                                                                          
                                                                                                                                                 
This PR fixes the issue by copying the active GL implementation parts to `gpu_info_` right after successful GL initialization, matching the behavior of the non-Android / out-of-process initialization paths.  

Bug: 517566565


yell0wd0g (approved):
LGTM - we should upstream this because it might
very well be forgotten due to the in-process
not being fully tested upstream. Happy to help
here.

PR #10915, opened 2026-06-16
Update supported features to point to Blink

y4vor:
Starting with C27 we use Blink which
supports all web features.

Bug: 508341552


cobalt/site/docs/development/reference/supported-features.md
arjungm, line 1: instead of deleting the file, can we just delete the contents and leave a note here that starting with C27 we use Blink, and link to whatever Blink's reference docs are for this information?

arjungm (changes requested):

cobalt/site/docs/development/reference/supported-features.md
y4vor: done

arjungm (approved):

PR #10924, opened 2026-06-16
Revert `GetHeapSize` to only `used_heap_size` (CP #7865)

johnxwork:
Upstream Chromium calculates `used_js_heap_size` by including `external_memory`. This is not backwards compatible with previous versions of Cobalt where it was only `used_heap_size`. Revert this for now and check canary metrics. Follow up with introducing a new API that takes into account `external_memory`: b/457773763

Bug: 456183913


andrewsavage1 (approved):
Needs to be guarded but otherwise lgtm

PR #10926, opened 2026-06-16
android: Add java switch to force angle OpenGL ES

ckim1346:
Adding a java switch to enable the experimentation of using the `--use-angle=gles` flag, restricting ANGLE to use the OpenGL ES backend specifically and avoiding the overhead of trying to initialize other unsupported backends and saving startup time.

Bug: 517566565


tasunnn (approved):

johnxwork (approved):
Synced offline. Suggest to try adding timestamps to confirm that Chrobalt did spend 350ms on trying unsupported rendering backends on startup.

PR #10928, opened 2026-06-16
Add java switch to enable v8 concurrent marking

tasunnn:
Add java switch to enable concurrent marking for v8 garbage 
collection. This feature is still disabled by default and will be 
enabled via runtime command line args to monitor performance 
under experiment settings.

Bug: 512163617


cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -97,6 +100,10 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
       extraCommandLineArgs.add("--js-flags=--disable-optimizing-compilers;--no-sparkplug");
     }
 
+    if (javaSwitches.containsKey(JavaSwitches.ENABLE_V8_CONCURRENT_MARKING)) {
+      extraCommandLineArgs.add("--js-flags=--concurrent-marking");
+    }
```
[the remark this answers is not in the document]
Awallky: +1 to checking for existing `--js-flags`. We want a single source for this.
Perhaps switching from a List to a Map would make this more straightforward to handle.

cobalt/build/configs/common.gn
```diff
@@ -43,11 +43,9 @@ enable_vr = false
 # Disable screen/window/tab capture (Note: live capture, not screenshots).
 enable_screen_capture = false
 
-# Disable concurrent-marking due to b/415843979.
-v8_enable_concurrent_marking = false
-
-# This flag is also needed to disable concurrent marking.
-v8_enable_atomic_object_field_writes = false
+# Re-enable concurrent marking to monitor via experiment.
+v8_enable_concurrent_marking = true
```
Awallky, line 47: By setting `v8_enable_concurrent_marking = true` in `common.gn`, doesn't this mean the feature is enabled at compile-time for ALL platforms? 
If so, should we add `--no-concurrent-marking` to the default JavaScript flags in `cobalt/app/cobalt_switch_defaults_starboard.cc`?

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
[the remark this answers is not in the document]
tasunnn: thanks Adam! just make them into a single string

cobalt/build/configs/common.gn
tasunnn: that's a good point, let me add that

johnxwork (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -88,6 +88,9 @@ public static StringJoiner getDefaultJsFlagOverridesList() {
         // Disable decommitting pooled pages to prevent virtual memory fragmentation.
         paramOverrides.add("--no-decommit-pooled-pages");
 
+        // Disable v8 concurrent marking by default.
+        paramOverrides.add("--no-concurrent-marking");
```
johnxwork, line 92: Do we have logic in CommandLineOverrideHelper to ensure "--no-concurrent-marking" is not appended when "--concurrent-marking" is present?

If not, I'm not sure if which one of ("--no-concurrent-marking", "--concurrent-marking") would Chromium respect.

If you like, you can also fold them together, like
if (javaSwitches.containsKey(JavaSwitches.ENABLE_V8_CONCURRENT_MARKING)) {
      jsFlags.add("--concurrent-marking");
    } else {
      jsFlags.add("--no-concurrent-marking");
}

tasunnn: I think flags from java switches will be appended after default flags, whatever comes last will override flag value so --no-concurrent-marking will have no effect if --concurrent-marking is appended

tasunnn: When experiment passes in the --concurrent-marking, both flags will actually be present, but the way cobalt parses the flags will make sure that flags defined in JavaSwitches.java always come after those defined in CommandLineOverrideHelper (https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java;l=197-209?q=CobaltActivity.java&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) and the latter flag will override the value.

PR #10932, opened 2026-06-16
Fix hang on exit by timing out on GPU connection

andrewsavage1:
Bug: 490989945


jellefoks (approved):
Let's make a separate bug for investigating whether there is a bigger root cause for this to be necessary.

PR #10941, opened 2026-06-16
Revert "cobalt: Fix EGL destruction crash on Shield"

borongc:
Reverts youtube/cobalt#10690 as it doesn't resolve the crash on shield.

Issue: 517926212


yell0wd0g (approved):
RS LGTM - thanks for cleaning up!

PR #10942, opened 2026-06-16
android: Enforce ANGLE to use GL ES backend

ckim1346:
This PR enforces Chrobalt to make ANGLE use the GL ES backend by adding `{::switches::kUseANGLE, "gles"}` to our default cobalt switches. While Vulkan is already disabled, this ensures that ANGLE does not attempt to use any other backends as well.

This PR also removes the android-specific java switches for this feature.

Bug: 517566565


cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -85,6 +85,8 @@ CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
   static const base::CommandLine::SwitchMap kCobaltSwitchDefaults{
       // Disable Vulkan.
       {::switches::kDisableFeatures, "Vulkan"},
+      // Force ANGLE to use GLES backend by default.
+      {::switches::kUseANGLE, "gles"},
```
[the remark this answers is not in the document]
andrewsavage1: @sherryzy can you ptal?

cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -85,6 +85,10 @@ CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
   static const base::CommandLine::SwitchMap kCobaltSwitchDefaults{
       // Disable Vulkan.
       {::switches::kDisableFeatures, "Vulkan"},
+  // Force ANGLE to use GLES backend by default.
+#if !BUILDFLAG(IS_OZONE)
```
sherryzy: I don't think this file is being used by Android build, but only 3p platform. Can you double check?

cobalt/app/cobalt_switch_defaults_starboard.cc
sherryzy, line 104: Also, there are existing settings for {::switches::kUseANGLE, "gles-egl"}; if these are indeed used by Android, they should be grouped together.

ckim1346: So since we're updating android ANGLE to use gles, should we also enforce for other non-android and non-ozone platforms?

```
#if BUILDFLAG(IS_OZONE)                                                                                  
    {::switches::kUseGL, "angle"},                                                                     
    {::switches::kUseANGLE, "gles-egl"},
#else
    {::switches::kUseANGLE, "gles"},
#endif
```

cobalt/app/cobalt_switch_defaults_starboard.cc
ckim1346: You're right; I updated the code to enforce in `CommandLineOverrideHelper.java` instead

johnxwork (approved):

sherryzy (approved):

PR #10951, opened 2026-06-17
Add gemini  rdk skill

niranjanyardi:
This PR adds the `cobalt-rdk-deploy` agent skill to the repository under `.agent/skills/`.

The skill provides guidelines and execution protocols for Gemini/Jetski to build, deploy, run, and debug Cobalt on RDK devices using the `deploy_rdk.py` script.

### Guidelines included:
1. **Rely on help:** Always run `--help` first to discover available execution modes and flags.
2. **Execute via script:** Formulate and run the `deploy_rdk.py` script directly to interact with the device.
3. **Avoid manual device interaction:** Explicitly forbids running raw shell commands (via `adb shell` or `ssh`) directly on the target device for pre-flight checks (trusting the script to handle them).
4. **Toolchain assumptions:** Assume the compilation toolchain is present by default and do not run `--setup-toolchain` or perform manual pre-emptive checks (like checking `$RDK_HOME`) unless compilation fails.
5. **No-RBE default:** Avoid passing `--no-rbe` by default (only use it if compilation fails).

Bug: 502702565

TAG=agy
CONV=e3d4c794-091a-4ae7-8cd1-810eebeb07ee

### DIFF
--- a/.agent/skills/cobalt-rdk-deploy/SKILL.md
+++ b/.agent/skills/cobalt-rdk-deploy/SKILL.md
@@ -1,0 +1,36 @@
+---
+name: cobalt-rdk-deploy
+description: >-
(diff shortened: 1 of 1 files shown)


yjzhang111 (approved):
Patched and tested that this works! Thanks for sending this out!

oxve (approved):

PR #10965, opened 2026-06-17
ci: Fix missing Junit results in android checks for Draft PRs

arjungm:
This PR does the following:
* Fixes an issue where Draft PRs were failing android checks due to missing test results
* Adds proper Junit test result parsing for android checks
* Refactors main workflow test steps to compute static control flags during initialize job

TAG=agy
CONV=91df7b12-8aa5-4f84-869a-f940d029e916

Bug: 431780245


.github/actions/on_host_tests/action.yaml
```diff
@@ -159,19 +159,14 @@ runs:
...
 
-          # Check if a wrapper script was generated for the test (e.g. by test_runner_script).
-          # If so, use it as the entrypoint to run tests.
-          wrapper_path="$(dirname "${test_executable_path}")/bin/run_${test_executable}"
-          if [[ -x "${wrapper_path}" ]]; then
-            ENTRYPOINT="./${wrapper_path}"
-          else
-            ENTRYPOINT="./${test_executable_path}"
-          fi
+          echo "Running tests for suite: ${test_executable_name}"
```
oxve: nit: This should say `Running tests for target...`

.github/actions/on_host_tests/action.yaml
```diff
@@ -159,19 +159,14 @@ runs:
           filename=$(basename "${test_executable_path}")
           test_executable="${filename%%.*}"
```
oxve: I missed this when it was introduced. Why is it stripping everything after the first `.` here? Can you add a comment to explain?

.github/workflows/main.yaml
```diff
@@ -291,6 +291,73 @@ jobs:
...
+          if [ -f "$test_targets_json_file" ]; then
+            # 1. Evaluate conditions (local variables)
+
+            # test_on_host
+            test_on_host="${{ steps.set-test-on-host.outputs.test_on_host }}"
+            if [[ "${test_on_host}" == "true" ]]; then
+              run_on_host_test="true"
+            fi
+
+            # test_on_device
```
oxve: Since these expressions are rather wordy and repeated a few times below (both in this step and for yts and e2e) I think it would make sense to break them out to the global `env:` block at the top of the file.

This is also important because we shouldn't have `${{ }}` expressions in the bash code.

E.g.
```
env:
  # List of folders used by CI actions that don't need a full source checkout.
  # All files in the root directory are automatically included.
  CI_ESSENTIALS: |
    .github
    cobalt/build
    cobalt/docker
    cobalt/testing/browser_tests/tools
    cobalt/testing/filters
    cobalt/tools
  PR_ODT_ENABLED: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
  NIGHTLY_ODT_ENABLED: ${{ (inputs.nightly == 'true' || github.event_name == 'schedule') && vars.RUN_ODT_TESTS_ON_NIGHTLY != 'False' }}
  POSTSUBMIT_ODT_ENABLED: ${{ github.event_name == 'push' && vars.RUN_ODT_TESTS_ON_POSTSUBMIT != 'False' }}
```

This should also clear up this if block and allow simpler heuristic for setting the `run_` variables.

.github/workflows/main.yaml
```diff
@@ -322,6 +389,10 @@ jobs:
       num_gtest_shards: ${{ steps.set-gtest-shards.outputs.num_gtest_shards }}
       enable_sccache: ${{ steps.set-enable-sccache.outputs.enable_sccache }}
       docker_service: ${{ steps.set-docker-service.outputs.docker_service }}
+      expected_test_targets_json: ${{ steps.generate-expected-test-targets-json.outputs.expected_test_targets_json }}
+      run_on_host_test: ${{ steps.generate-expected-test-targets-json.outputs.run_on_host_test }}
+      run_on_device_test: ${{ steps.generate-expected-test-targets-json.outputs.run_on_device_test }}
+      run_browser_test_on_host: ${{ steps.generate-expected-test-targets-json.outputs.run_browser_test_on_host }}
```
oxve: I don't think you should do anything in this PR (except maybe a TODO) but it's confusing now that there's both `run_on_host_test` and `test_on_host`.

This should be cleaned up in a future PR. Can you add the TODO?

.github/workflows/main.yaml
```diff
@@ -291,6 +291,73 @@ jobs:
...
+              --arg run_on_device_test "${run_on_device_test}" \
+              --arg run_on_host_test "${run_on_host_test}" \
+              '[
+                .[] | select(
+                  (.target != null) and
+                  (.test_type == "gtest" or .test_type == "junit") and
+                  (
+                    (.runs_on == "device" and $run_on_device_test == "true") or
+                    (.runs_on == "host"  and $run_on_host_test == "true")
+                  )
```
oxve, line 333: This expression feels funky to me.  I can't quite put my finger on it though. It could be that only gtest and junit are integrated into the "standard" result processing while the other tests are not.

WDYT?

.github/actions/on_host_tests/action.yaml
arjungm: oh this must be copied from the gtests for evergreen loaders which run a python script as an executable instead

.github/workflows/main.yaml
arjungm: I think it's written this way so that when device/host tests are conditionally disabled that the corresponding tests of that type are omitted. It's probably required here since in the Draft PR case we execute the Junit on-host but not the Gtest on-device.

.github/workflows/main.yaml
```diff
@@ -291,6 +295,78 @@ jobs:
...
+          run_on_host_test="false"
+          run_on_device_test="false"
+          run_browser_test_on_host="false"
+          expected_targets="[]"
+
+          first_platform=$(jq -r '.platforms[0]' ${GITHUB_WORKSPACE}/.github/config/${{ inputs.platform }}.json)
+          test_targets_json_file="cobalt/build/testing/targets/${first_platform}/test_targets.json"
+
+          if [ -f "$test_targets_json_file" ]; then
+            # 1. Evaluate conditions (local variables)
```
oxve: Remove this comment.

.github/workflows/main.yaml
oxve, line 644: Please clean up this block using the new env vars.

oxve, line 609: Please clean up this block using the new env vars.

.github/workflows/main.yaml
```diff
@@ -291,6 +295,66 @@ jobs:
...
+                  "$PR_ODT_ENABLED" == "true" ||
+                  "$NIGHTLY_ODT_ENABLED" == "true" ||
+                  "$POSTSUBMIT_ODT_ENABLED" == "true" ) ]]; then
+              run_on_device_test="true"
+            fi
+
+            if [[ "$RUN_BROWSER_TESTS" == "true" && (
+                  "$PR_BROWSERTEST_ENABLED" == "true" ||
+                  "$NIGHTLY_ODT_ENABLED" == "true" ||
+                  "$POSTSUBMIT_ODT_ENABLED" == "true" ) ]]; then
```
oxve: This looks wrong. Why would we set run_browser_test_on_host to true when ODTs are enabled?

oxve (approved):

.github/workflows/main.yaml
```diff
@@ -39,6 +39,10 @@ env:
     cobalt/testing/browser_tests/tools
     cobalt/testing/filters
     cobalt/tools
+  PR_ODT_ENABLED: ${{ github.event_name == 'pull_request' && github.event.pull_request.draft == false }}
+  NIGHTLY_ODT_ENABLED: ${{ (inputs.nightly == 'true' || github.event_name == 'schedule') && vars.RUN_ODT_TESTS_ON_NIGHTLY != 'False' }}
+  POSTSUBMIT_ODT_ENABLED: ${{ github.event_name == 'push' && vars.RUN_ODT_TESTS_ON_POSTSUBMIT != 'False' }}
+  PR_BROWSERTEST_ENABLED: ${{ github.event_name == 'pull_request' }}
```
oxve: This one looks a bit strange. `PR_BROWSERTEST_ENABLED` is just checking if it's a PR.

I think this should be removed. Why skip browser tests on postsubmit and nightly?

.github/workflows/main.yaml
arjungm: Can we save this for a follow up PR? I want to get the major fixes in for draft PRs to unblock the android checks in first.

arjungm: Can we save this for a follow up PR? I want to get the major fixes in for draft PRs to unblock the android checks in first.

.github/workflows/main.yaml
arjungm: I took another stab at it and the original if block definitely has some weird logic in there that wasn't caught before. The refactoring I have now seems to be consistent with the old logic so maybe come back to this one later? Technically this is a non-functional change

.github/workflows/main.yaml
arjungm: I refactored it again and it matches up with the old if-block pretty closely. I think this is just exposing some code rot in here with the renames and new run-control flags ...

.github/actions/process_test_results/action.yaml
```diff
@@ -45,6 +45,7 @@ runs:
         GTEST_SHARDS: ${{ inputs.num_gtest_shards }}
       run: |
         # Check test status.
+        set -x
```
arjungm: debug print

.github/workflows/main.yaml
```diff
@@ -906,6 +923,8 @@ jobs:
               echo "skipped=false" >> $GITHUB_OUTPUT
             fi
           fi
+          results_tree=$(find results/ -print | sed -e 's;[^/]*/;|____;g;s;____|; |;g')
+          echo $results_tree
```
arjungm: debug print

oxve (approved):

PR #10982, opened 2026-06-18
cobalt: Add UMA metrics for Finch configurations

yjzhang111:
Add several UMA histograms to track the status and success rate of
Finch experiment configurations. These metrics monitor feature and
parameter application counts, detect invalid data types, and track
startup crash streaks.

Additionally, telemetry is added to the H5VCC experiments interface to
record parse errors, connection errors, and overall success rates
when setting experiment states. This data is essential for
identifying and diagnosing configuration issues in production where
log-based debugging is limited.

Bug: 515520976

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -529,6 +529,8 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
       use_safe_config ? kSafeConfigFeatureParams
                       : kExperimentConfigFeatureParams);
 
+  bool has_invalid_feature_type = false;
+  int features_applied = 0;
   for (const auto feature_name_and_value : feature_map) {
     if (feature_name_and_value.second.is_bool()) {
       auto override_value =
@@ -537,20 +539,30 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
               : base::FeatureList::OverrideState::OVERRIDE_DISABLE_FEATURE;
       feature_list->RegisterFieldTrialOverride(
           feature_name_and_value.first, override_value, cobalt_field_trial);
+      features_applied++;
     } else {
+      has_invalid_feature_type = true;
       // TODO(b/407734134): Register UMA here for non boolean feature value.
       LOG(ERROR) << "Failed to apply override for feature "
                  << feature_name_and_value.first;
       base::debug::DumpWithoutCrashing();
     }
   }
+  base::UmaHistogramBoolean("Cobalt.Finch.HasInvalidFeatureType",
+                            has_invalid_feature_type);
+  base::UmaHistogramCounts100("Cobalt.Finch.NumFeaturesApplied",
+                              features_applied);
 
+  bool has_invalid_param_type = false;
+  int params_applied = 0;
   base::FieldTrialParams params;
   for (const auto param_name_and_value : param_map) {
     if (param_name_and_value.second.is_string()) {
       params.emplace(param_name_and_value.first,
                      param_name_and_value.second.GetString());
+      params_applied++;
     } else {
+      has_invalid_param_type = true;
       // TODO(b/407734134): Register UMA here for non string param value.
       LOG(ERROR) << "Failed to associate field trial param "
                  << param_name_and_value.first << " with string value "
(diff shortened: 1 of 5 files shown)


Awallky (approved):
LGTM

cobalt/browser/experiments/experiment_config_manager.cc
```diff
@@ -146,6 +146,8 @@ ExperimentConfigType ExperimentConfigManager::GetExperimentConfigType() {
 
   int num_crashes = metrics_local_state_->GetInteger(
       variations::prefs::kVariationsCrashStreak);
+  base::UmaHistogramExactLinear("Cobalt.Finch.CrashStreakAtStartup",
```
johnxwork, line 149: Did you try locally or does Finch have local test to verify that consistent startup crashes resulted from Finch misconfiguration can be recorded by this histogram? I feel like it depends on a few factors that we are known to be weak at: flash storage reliability, CPU bandwidth, e.t.c.

johnxwork (approved):

cobalt/browser/experiments/experiment_config_manager.cc
yjzhang111: This crash streak is not "Finch specific", meaning it'll increment even when Cobalt crashed for a completely unrelated reason. I don't think there is a "Finch specific" crash counter in chromium that we could use.

For even more context - 
In Chrome, they use the 99.5th percentile to determine their Finch crash streak threshold
https://source.chromium.org/chromium/chromium/src/+/main:components/variations/service/safe_seed_manager.h;l=27-44
I'm trying to log this UMA to determine what our threshold should be.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -537,27 +539,40 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
               : base::FeatureList::OverrideState::OVERRIDE_DISABLE_FEATURE;
       feature_list->RegisterFieldTrialOverride(
           feature_name_and_value.first, override_value, cobalt_field_trial);
+      features_applied++;
     } else {
+      has_invalid_feature_type = true;
       // TODO(b/407734134): Register UMA here for non boolean feature value.
       LOG(ERROR) << "Failed to apply override for feature "
                  << feature_name_and_value.first;
       base::debug::DumpWithoutCrashing();
```
yell0wd0g, line 545: Curious why not do `NOTREACHED()` or `DLOG(FATAL)`? Current
code will emit some blurb but continue, the proposed ones
would crash in non-release builds which might be a better
signal? My point being that this might be a very bad 
situation and crashing loudly might be better than a silent
strong complain.

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -537,27 +539,40 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
...
     }
   }
+  base::UmaHistogramBoolean("Cobalt.Finch.HasInvalidFeatureType",
```
yell0wd0g, line 549: Nit: Maybe we don't need another variable? And here just do.

```
  const bool has_invalid_feature_type = feature_map.size() != features_applied;
  base::UmaHistogramBoolean("Cobalt.Finch.HasInvalidFeatureType", 
                            has_invalid_feature_type);
```

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -537,27 +539,40 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
               : base::FeatureList::OverrideState::OVERRIDE_DISABLE_FEATURE;
       feature_list->RegisterFieldTrialOverride(
           feature_name_and_value.first, override_value, cobalt_field_trial);
+      features_applied++;
```
yell0wd0g, line 541: micro-nit: Iterating over a map to count something _in addition_
to doing something else hmm. Usually I wouldn't mind but here
in the critical path we might make the compiler life's easier
by taking this out, e.g.:

```
  const auto features_applied = std::ranges::count_if(feature_map, 
         [](const auto& val) { return val.is_bool(); });
```

WDYT?

yell0wd0g (approved):
LGTM with some suggestions.

cobalt/browser/cobalt_content_browser_client.cc
yjzhang111: The original intent is to follow existing chromium pattern.

DLOG(FATAL) also implies that if an invalid flag value is pushed and this is, it'll immediately crash all dev builds. Since the flag comes from the server, Cobalt engineers will likely need to use custom url or command line override, or go to experiments portal to explicitly opt out of the invalid experiment - there's no immediate fix we can apply on client side to unblock dev workflow or CI.

I do think for production builds, it should continue without crashing - skipping a feature doesn't seem too bad comparing to crashing production builds.

cobalt/browser/cobalt_content_browser_client.cc
yjzhang111: Yea i'm fine with either but I thought count_if essentially does another loop on top of the for loop we're already doing which adds runtime overhead

PR #10983, opened 2026-06-19
Freeze with a 1.5s delay after visibilityChange

johnxwork:
## Description
This PR brings back the `freeze` event to the web client with a critical 1.5-second cushioning delay after the `visibilitychange` event is dispatched. 

Additionally, it resolves visibility tracking inaccuracies on Android TV by introducing a hardware-level `DisplayListener`. On many Android TV devices, turning off the TV screen or switching HDMI inputs does not immediately trigger standard Activity lifecycle events like `onPause()` or `onStop()`. To ensure the web app is correctly notified when it becomes invisible, we now monitor the hardware display state directly.

By combination of these two changes:
1. `visibilitychange` (hidden) is fired immediately when the app goes to the background (`onStop`) OR when the physical display turns off.
2. If triggered by `onStop`, the `freeze` event is queued with a 1.5s delay. If the app is resumed before the timer expires, the pending freeze is canceled.
3. This allows the web client to cleanly save state and process backgrounding tasks before JS execution is suspended.

## Changes
- Modified `dev.cobalt.coat.CobaltActivity`:
  - Added a `Handler` and a `Runnable` (`mFreezeRunnable`) to manage the delayed freeze.
  - Implemented `DisplayManager.DisplayListener` to track hardware display state changes (ON/OFF/UNKNOWN) and update shell visibility immediately.
  - Updated `onStart()` to register the display listener and cancel any pending freeze.
  - Updated `onStop()` to unregister the display listener, fire `visibilitychange` (hidden), and queue the `freeze` event with a 1.5s delay.
  - Overrode `onDestroy()` to ensure the pending `mFreezeRunnable` is canceled to prevent memory leaks.

## Testing
Tested manually on a connected Android TV device by sideloading a local HTML file and validating the web events via ADB intents:
- **Foreground -> Background** (`KEYCODE_HOME`):
  - `visibilitychange` (hidden) dispatched immediately.
  - `freeze` (hidden) dispatched exactly 1.5 seconds later.
- **Background -> Foreground** (App launched via intent):
  - Pending `freeze` (if any) is correctly canceled.
  - `resume` dispatched, followed by `visibilitychange` (visible).
- **Screen Off / HDMI Disconnect** (Simulated display changes):
  - `visibilitychange` (hidden) is fired immediately via the new `DisplayListener` even if the Activity remains in the foreground.

Bug: 517926212
TAG=agy
CONV=142f4a55-aaa4-4b03-a6ca-b0833ef012e0

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -35,6 +35,8 @@ import android.view.View;
 import android.view.ViewGroup.LayoutParams;
 import android.view.ViewParent;
 import android.view.WindowManager;
+import android.view.Display;
+import android.hardware.display.DisplayManager;
 import android.window.OnBackInvokedCallback;
 import android.window.OnBackInvokedDispatcher;
 import android.widget.FrameLayout;
@@ -116,7 +118,8 @@ public abstract class CobaltActivity extends Activity {
   private IntentRequestTracker mIntentRequestTracker;
   // Tracks the status of the FLAG_KEEP_SCREEN_ON window flag.
   private Boolean mIsKeepScreenOnEnabled = false;
-
+  private Runnable mFreezeRunnable;
+  private final Handler mHandler = new Handler(Looper.getMainLooper());
   private boolean mIsCobaltUsingAndroidOverlay;
 
   private boolean mEnableSplashScreen;
@@ -131,6 +134,15 @@ public abstract class CobaltActivity extends Activity {
   // to Escape), this flag is used by OnBackInvokedCallback to detect physical key presses and
   // bypass simulated key dispatching.
   private boolean mPhysicalBackKeyPressed = false;
+  private final DisplayUtil.Listener mDisplayListener = new DisplayUtil.Listener() {
+    @Override
+    public void onDisplayChanged(int displayId) {
+      if (displayId == Display.DEFAULT_DISPLAY) {
+        checkDisplayState();
+      }
+    }
+  };
+  private boolean mWasDisplayOn = true;
 
   private Bundle getActivityMetaData() {
     ComponentName componentName = getIntent().getComponent();
@@ -550,6 +562,10 @@ public abstract class CobaltActivity extends Activity {
 
   @Override
   protected void onStart() {
+    DisplayUtil.cacheDefaultDisplay(this);
+    DisplayUtil.addDisplayListener(this);
+    mWasDisplayOn = isDisplayOn();
+    registerDisplayListener();
     StartupGuard.getInstance().setStartupMilestone(10);
     if (isDevelopmentBuild()) {
       getStarboardBridge().getAudioOutputManager().dumpAllOutputDevices();
@@ -562,21 +578,24 @@ public abstract class CobaltActivity extends Activity {
       createNewSurfaceView();
     }
 
-    DisplayUtil.cacheDefaultDisplay(this);
-    DisplayUtil.addDisplayListener(this);
     AudioOutputManager.addAudioDeviceListener(this);
 
     getStarboardBridge().onActivityStart(this);
     super.onStart();
 
+    if (mFreezeRunnable != null) {
+      mHandler.removeCallbacks(mFreezeRunnable);
+      mFreezeRunnable = null;
+    }
     WebContents webContents = getActiveWebContents();
-    // If ENABLE_FREEZE is not specified, disable corresponding resume event by default.
-    if (webContents != null && getJavaSwitches().containsKey(JavaSwitches.ENABLE_FREEZE)) {
+    if (webContents != null &&
+        (getJavaSwitches().containsKey(JavaSwitches.DELAY_FREEZE_ON_BACKGROUND) ||
+         getJavaSwitches().containsKey(JavaSwitches.ENABLE_FREEZE))) {
       // document.onresume event
       webContents.onResume();
     }
     // visibility:visible event
-    updateShellActivityVisible(true);
+    updateShellActivityVisible(mWasDisplayOn);
     MemoryPressureMonitor.INSTANCE.enablePolling(false);
 
     StartupGuard.getInstance().setStartupMilestone(11);
@@ -591,16 +610,33 @@ public abstract class CobaltActivity extends Activity {
 
   @Override
   protected void onStop() {
+    unregisterDisplayListener();
     getStarboardBridge().onActivityStop(this);
     super.onStop();
 
     // visibility:hidden event
     updateShellActivityVisible(false);
     WebContents webContents = getActiveWebContents();
-    // If ENABLE_FREEZE is not specified, disable freeze event by default.
-    if (webContents != null && getJavaSwitches().containsKey(JavaSwitches.ENABLE_FREEZE)) {
-      // document.onfreeze event
-      webContents.onFreeze();
+    if (webContents != null) {
+      if (getJavaSwitches().containsKey(JavaSwitches.DELAY_FREEZE_ON_BACKGROUND)) {
+        if (mFreezeRunnable != null) {
+          mHandler.removeCallbacks(mFreezeRunnable);
+        }
+        mFreezeRunnable = new Runnable() {
+          @Override
+          public void run() {
+            WebContents currentWebContents = getActiveWebContents();
+            if (currentWebContents != null) {
+              currentWebContents.onFreeze();
+            }
+            mFreezeRunnable = null;
+          }
+        };
+        mHandler.postDelayed(mFreezeRunnable, 1500);
+      } else if (getJavaSwitches().containsKey(JavaSwitches.ENABLE_FREEZE)) {
+        // If ENABLE_FREEZE is specified, fire freeze event immediately
+        webContents.onFreeze();
+      }
     }
 
     if (VideoSurfaceView.getCurrentSurface() != null) {
@@ -628,6 +664,11 @@ public abstract class CobaltActivity extends Activity {
 
   @Override
   protected void onDestroy() {
+    unregisterDisplayListener();
+    if (mFreezeRunnable != null) {
+      mHandler.removeCallbacks(mFreezeRunnable);
+      mFreezeRunnable = null;
+    }
     if (mShellManager != null) {
       mShellManager.destroy();
     }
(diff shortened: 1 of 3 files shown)


borongc (approved):

zhongqiliang:
## 1. Critical Issues & Potential Bugs

### 🔴 Initial State Desync in Display Listener
The `registerDisplayListener()` method registers `mDisplayListener` but does **not** perform an initial check of the display state.
* **The Risk:** `mWasDisplayOn` is initialized to `true`. If `CobaltActivity` starts while the screen is **OFF** (e.g., background start, or device waking up slowly), `onStart` will default to calling `updateShellActivityVisible(true)`. The listener will not trigger because the screen state is stable (off), leaving the app in a "visible" state internally even though the screen is off.
* **Fix:** Call `checkDisplayState()` immediately after registering the listener to synchronize the initial state.
```java
  private void registerDisplayListener() {
    if (mDisplayListener == null) {
      mDisplayListener = new DisplayManager.DisplayListener() {
        // ... (definition)
      };
    }
    DisplayManager displayManager = getSystemService(DisplayManager.class);
    if (displayManager != null) {
      displayManager.registerDisplayListener(mDisplayListener, null);
      // Query initial state immediately to sync mWasDisplayOn and visibility
      mDisplayListener.onDisplayChanged(Display.DEFAULT_DISPLAY); 
    }
  }
```
*(Note: Since `checkDisplayState` is private to the listener, triggering `onDisplayChanged` is a simple way to force the check).*

---

## 2. Design Flaws & Code Smells

### 🟡 Redundant Display Listeners
`DisplayUtil` already implements and registers a static `DisplayListener` (`sDisplayerListener`) in `addDisplayListener(Context)`.
* **The Issue:** PR 10983 introduces a *second* `DisplayListener` directly inside `CobaltActivity`. 
    * `DisplayUtil`'s listener notifies JNI (`DisplayUtilJni.get().onDisplayChanged()`).
    * `CobaltActivity`'s listener calls Java visibility methods (`updateShellActivityVisible`).
* **Critique:** Having two separate listeners registered with the system `DisplayManager` is redundant. `DisplayUtil`'s listener could easily dispatch to a Java listener interface, allowing `CobaltActivity` to subscribe to it rather than registering its own system listener.

### 🟡 Hacky Feature Flag Enablement
The PR enables `DELAY_FREEZE_ON_BACKGROUND` by overriding `getJavaSwitches()`:
```java
  protected Map<String, String> getJavaSwitches() {
    Map<String, String> switches = new HashMap<>(this.mJavaSwitches);
    switches.put(JavaSwitches.DELAY_FREEZE_ON_BACKGROUND, "1");
    return switches;
  }
```
* **The Issue:** This force-injects the switch into the map on *every* call to `getJavaSwitches()`. 
    1. **Inefficiency:** It creates a new `HashMap` and copies the existing one every time this getter is called.
    2. **Redundancy:** If the switch is *always* forced to "1", checking `getJavaSwitches().containsKey(...)` in `onStart` and `onStop` is redundant (it will always be true).
    3. **Undisablable:** It prevents disabling the feature via command-line arguments or configuration files, as this method will always overwrite it.
* **Fix:** If the feature should be default-on, initialize it in the default switches map configuration, or make the logic the default behavior without gatekeeping it behind a switch that can't be turned off anyway.

---

## 3. Lifecycle & Resource Management

### 🟢 Unregistration Safety
The listener is registered in `onStart` and unregistered in `onStop`. 
* **Critique:** While this matches the lifecycle, it is safer to also attempt unregistration in `onDestroy` (guarded by null checks) to prevent leaks in edge cases where `onStop` might be bypassed (e.g. system crashes during transitions).
* **Fix:**
```java
  @Override
  protected void onDestroy() {
    unregisterDisplayListener(); // Ensure clean-up
    if (mFreezeRunnable != null) { ... }
    // ...
  }
```

### 🟢 Duplicate Visibility Calls
`onStop()` explicitly calls `updateShellActivityVisible(false)`. If the screen turns off, the display listener might *also* call `updateShellActivityVisible(false)` before the listener is unregistered.
* **Critique:** Ensure `updateShellActivityVisible` (and eventually `WebContents.updateWebContentsVisibility`) is fully idempotent and cheap to call with duplicate values. Currently, it delegates to `ShellManager` which delegates to `Shell`, which calls `mWebContents.updateWebContentsVisibility`. While Chromium likely handles duplicates, avoiding them at the Java layer is safer.

---

## 4. Minor / Style Issues

### ⚪ Unrelated Comment Deletion
The PR removes the comment tracking the status of `FLAG_KEEP_SCREEN_ON`:
```diff
-  // Tracks the status of the FLAG_KEEP_SCREEN_ON window flag.
   private Boolean mIsKeepScreenOnEnabled = false;
```
* **Critique:** The comment was useful and its removal is unrelated to the PR's functional changes. It should be restored.

zhongqiliang:
The comments above are valid.

zhongqiliang:
## Critical Issues & Bug Risks

### 1. Incorrect Initial Visibility State in `onStart`
In `onStart()`, the code unconditionally sets the shell activity to visible:
```java
updateShellActivityVisible(true);
```
**The Bug:** If the Activity is started while the screen is actually OFF (e.g., started via background service, alarm, or `adb` command when device is asleep), the app will incorrectly report its visibility as `true`.
*   The newly registered `DisplayListener` only triggers on *changes*. Since the display was already OFF, no event may be fired immediately to correct this.
*   **Consequence:** The app might start rendering and consuming resources in the background while the screen is off, defeating the purpose of display tracking.

### 2. Fail-Closed on Display Query Failure (Potential Black Screen)
In `checkDisplayState()`, if `DisplayUtil.getDefaultDisplay()` returns `null` (even after attempting to re-cache), the code defaults `isDisplayOn` to `false`:
```java
boolean isDisplayOn = false;
if (defaultDisplay != null) {
  int state = defaultDisplay.getState();
  isDisplayOn = (state == Display.STATE_ON || state == Display.STATE_UNKNOWN);
}
```
**The Bug:** If the system is in a transient state where the display is temporarily unavailable or invalid (e.g., during HDMI handshake on Android TV), `isDisplayOn` will resolve to `false`.
*   This will trigger `updateShellActivityVisible(false)`, hiding the UI.
*   If the display becomes available again but doesn't trigger a "change" event that resolves differently, or if we fail to recover, the UI will remain hidden (black screen) even when the display is physically ON.
*   **Recommendation:** Fail-safe. If we cannot determine the display state, we should assume it is **ON** (`true`) or retain the last known state, rather than defaulting to `false`.

---

## Design & Code Smells

### 3. Hardcoded Switch in Base Class
The PR modifies `getJavaSwitches()` in `CobaltActivity` (the base class) to hardcode the new switch:
```java
protected Map<String, String> getJavaSwitches() {
  Map<String, String> switches = new HashMap<>(this.mJavaSwitches);
  switches.put(JavaSwitches.DELAY_FREEZE_ON_BACKGROUND, "1");
  return switches;
}
```
**The Design Flaw:** 
*   If `DELAY_FREEZE_ON_BACKGROUND` is intended to be always enabled for all Cobalt apps, it should not be a "switch" checked via `getJavaSwitches().containsKey(...)`. It should just be the default behavior.
*   If it *must* be a switch, hardcoding it in the base class getter makes it very difficult for subclasses to opt-out (they would have to override the method and explicitly remove it from the map).
*   Creating a new `HashMap` on every call to `getJavaSwitches()` adds garbage collection pressure if this method is called frequently.

### 4. Redundant `DisplayListener` Recreation
In `registerDisplayListener()`, a new anonymous `DisplayListener` is created if it is null. However, `unregisterDisplayListener()` explicitly nulls it out:
```java
private void unregisterDisplayListener() {
  if (mDisplayListener != null) {
    // ... unregister ...
    mDisplayListener = null; // <--- Forces recreation
  }
}
```
**The Code Smell:** This causes the listener to be allocated and garbage collected on every `onStart`/`onStop` cycle. It is cleaner and more efficient to keep a single final instance of the listener as a member variable and simply register/unregister it.

### 5. Overlapping Switch Logic Precedence
In `onStop()`, the code checks both `DELAY_FREEZE_ON_BACKGROUND` and `ENABLE_FREEZE`:
```java
if (getJavaSwitches().containsKey(JavaSwitches.DELAY_FREEZE_ON_BACKGROUND)) {
  // ... delayed freeze ...
} else if (getJavaSwitches().containsKey(JavaSwitches.ENABLE_FREEZE)) {
  // ... immediate freeze ...
}
```
**The Issue:** If both switches are somehow set, `DELAY_FREEZE` silently wins. This precedence is undocumented and could lead to confusion.

---

## Recommended Refactoring

To resolve the critical issues and clean up the implementation, the following structure is recommended:

### Refactored `CobaltActivity.java`

1.  **Define the listener as a `final` member variable** to avoid recreation.
2.  **Move state tracking (`mWasDisplayOn`) to the outer class** for easier access during initialization.
3.  **Implement a helper `isDisplayOn()`** that fails-safe to `true`.
4.  **Correctly initialize visibility** in `onStart()`.

```java
// Define as member variables
private boolean mWasDisplayOn = true;
private final DisplayManager.DisplayListener mDisplayListener = 
    new DisplayManager.DisplayListener() {
        @Override
        public void onDisplayAdded(int displayId) { checkDisplayState(); }
        @Override
        public void onDisplayChanged(int displayId) { checkDisplayState(); }
        @Override
        public void onDisplayRemoved(int displayId) { checkDisplayState(); }
    };

// Fail-safe helper
private boolean isDisplayOn() {
    Display defaultDisplay = DisplayUtil.getDefaultDisplay();
    if (defaultDisplay == null) {
        DisplayUtil.cacheDefaultDisplay(this);
        defaultDisplay = DisplayUtil.getDefaultDisplay();
    }
    if (defaultDisplay == null) {
        // Fail-safe: assume display is ON if we cannot query it
        return true; 
    }
    int state = defaultDisplay.getState();
    return state == Display.STATE_ON || state == Display.STATE_UNKNOWN;
}

private void checkDisplayState() {
    boolean isDisplayOn = isDisplayOn();
    if (isDisplayOn != mWasDisplayOn) {
        mWasDisplayOn = isDisplayOn;
        Log.i(TAG, "Display state changed: isDisplayOn = " + isDisplayOn);
        updateShellActivityVisible(isDisplayOn);
    }
}

private void registerDisplayListener() {
    DisplayManager displayManager = getSystemService(DisplayManager.class);
    if (displayManager != null) {
        displayManager.registerDisplayListener(mDisplayListener, null);
    }
}

private void unregisterDisplayListener() {
    DisplayManager displayManager = getSystemService(DisplayManager.class);
    if (displayManager != null) {
        displayManager.unregisterDisplayListener(mDisplayListener);
    }
    // Do NOT null out mDisplayListener to avoid recreation
}

@Override
protected void onStart() {
    // 1. Determine actual initial state and apply it
    mWasDisplayOn = isDisplayOn();
    updateShellActivityVisible(mWasDisplayOn);
    
    // 2. Register listener for subsequent changes
    registerDisplayListener();
    
    StartupGuard.getInstance().setStartupMilestone(10);
    // ... rest of onStart ...
}
```

zhongqiliang:
Bullets 1-2 above are valid, 3-5 are optional.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -873,6 +914,37 @@ private void updateShellActivityVisible(boolean isVisible) {
...
+    if (defaultDisplay == null) {
+      DisplayUtil.cacheDefaultDisplay(this);
+      defaultDisplay = DisplayUtil.getDefaultDisplay();
+    }
+    if (defaultDisplay == null) {
+      // Fail-safe: assume display is ON if we cannot query it
+      return true;
+    }
+    int state = defaultDisplay.getState();
+    return state == Display.STATE_ON || state == Display.STATE_UNKNOWN;
```
zhongqiliang, line 930: Please add a comment on top of this line for why Display.STATE_UNKNOWN is considered as true.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -628,6 +664,11 @@ protected void onResume() {
 
   @Override
   protected void onDestroy() {
+    unregisterDisplayListener();
+    if (mFreezeRunnable != null) {
```
zhongqiliang, line 668: If it falls into here, it means onFreeze() did not finish, or not even run yet. Do we  need to trigger UMA here?

What is the reason to do DELAY_FREEZE_ON_BACKGROUND? Why not just ENABLE_FREEZE?

zhongqiliang (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
johnxwork: > If it falls into here, it means onFreeze() did not finish, or not even run yet. Do we need to trigger UMA here?

I think that might be fine. It means for example the app was immediately killed after being background(e.g. the user immediately opened Netflix and trigger memory pressure).

> What is the reason to do DELAY_FREEZE_ON_BACKGROUND? Why not just ENABLE_FREEZE?

It is for Kabuki to handle visibilityChange listeners. In local experiment, it turns out most visibilityChange handlers need async logic such as flushing storage(e.g. the dirty bit) and enabling freeze immediately after backgrounding does not allow time for that.

PR #10985, opened 2026-06-19
aosp: Bump NDK API level

hferreiro:
aligned_alloc is only supported on Android 28 or newer. Bump the NDK API level on AOSP's starboard toolchain.

Bug: 495203133

### DIFF
--- a/starboard/aosp/arm/toolchain/BUILD.gn
+++ b/starboard/aosp/arm/toolchain/BUILD.gn
@@ -54,6 +54,9 @@ android_clang_toolchain("starboard") {
     is_starboard = true
     use_cxx17 = true
     build_base_with_cpp17 = true
+
+    # 'aligned_alloc' is only available on Android 28 or newer.
+    android_ndk_api_level = 28
   }
 }
 


starboard/aosp/arm/toolchain/BUILD.gn
```diff
@@ -54,6 +54,9 @@ android_clang_toolchain("starboard") {
     is_starboard = true
     use_cxx17 = true
     build_base_with_cpp17 = true
+
+    # 'aligned_alloc' is only available on Android 28 or newer.
+    android_ndk_api_level = 28
```
hferreiro, line 59: The drawback is that at some point `default_min_sdk_version` will be 28 or higher, but we cannot use it here because this is evaluated in the default toolchain. Perhaps we should directly add it to build/config/android/config.gni.

johnxwork: From just quick web searches, it looks like FireOS 6 and FireOS 5 use ndk levels lower than 28. Do we need to support those devices in future updates?

andrewsavage1 (approved):

starboard/aosp/arm/toolchain/BUILD.gn
andrewsavage1: 2027 devices should mostly be okay and if they aren't they can still implement wrappers for aligned_alloc

PR #10997, opened 2026-06-20
aosp: Fix deprecated declaration warnings

hferreiro:
Fix the last deprecation warnings that prevent to build loader_app on AOSP. Replace readdir_r with readdir and MessageLite::ByteSize with ByteSizeLong.

Bug: 495203133

### DIFF
--- a/starboard/common/file.cc
+++ b/starboard/common/file.cc
@@ -183,18 +183,14 @@ bool SbFileDeleteRecursive(const char* path, bool preserve_root) {
 
   std::vector<char> entry(kSbFileMaxName);
 
-  struct dirent dirent_buffer;
   struct dirent* dirent;
   while (true) {
     if (entry.size() < static_cast<size_t>(kSbFileMaxName) || !dir ||
         !entry.data()) {
       break;
     }
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-    int result = readdir_r(dir, &dirent_buffer, &dirent);
-#pragma GCC diagnostic pop
-    if (result || !dirent) {
+    dirent = readdir(dir);
+    if (!dirent) {
       break;
     }
     starboard::strlcpy(entry.data(), dirent->d_name, kSbFileMaxName);
--- a/starboard/loader_app/installation_manager.cc
+++ b/starboard/loader_app/installation_manager.cc
@@ -586,13 +586,13 @@ int InstallationManager::RequestRollForwardToInstallation(
 bool InstallationManager::SaveInstallationStore() {
   ValidatePriorities();
 
-  if (IM_MAX_INSTALLATION_STORE_SIZE < installation_store_.ByteSize()) {
+  if (IM_MAX_INSTALLATION_STORE_SIZE < installation_store_.ByteSizeLong()) {
     SB_LOG(ERROR) << "SaveInstallationStore: Data too large"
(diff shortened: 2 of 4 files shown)


starboard/loader_app/installation_manager.cc
```diff
@@ -586,13 +586,13 @@ int InstallationManager::RequestRollForwardToInstallation(
...
-  if (IM_MAX_INSTALLATION_STORE_SIZE < installation_store_.ByteSize()) {
+  if (IM_MAX_INSTALLATION_STORE_SIZE < installation_store_.ByteSizeLong()) {
     SB_LOG(ERROR) << "SaveInstallationStore: Data too large"
-                  << installation_store_.ByteSize();
+                  << installation_store_.ByteSizeLong();
     return false;
   }
 
-  const size_t buf_size = installation_store_.ByteSize();
+  const size_t buf_size = installation_store_.ByteSizeLong();
```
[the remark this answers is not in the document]
hferreiro: Preexisting issue.

starboard/loader_app/installation_manager.cc
```diff
@@ -601,10 +601,10 @@ bool InstallationManager::SaveInstallationStore() {
 #endif
 
   installation_store_.SerializeToArray(buf.data(),
-                                       installation_store_.ByteSize());
+                                       installation_store_.ByteSizeLong());
 
   if (!starboard::FileAtomicReplace(store_path_.c_str(), buf.data(),
-                                    installation_store_.ByteSize())) {
+                                    installation_store_.ByteSizeLong())) {
```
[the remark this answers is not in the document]
hferreiro: Preexisting issue.

starboard/common/file.cc
```diff
@@ -183,18 +183,14 @@ bool SbFileDeleteRecursive(const char* path, bool preserve_root) {
...
     if (entry.size() < static_cast<size_t>(kSbFileMaxName) || !dir ||
         !entry.data()) {
       break;
     }
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-    int result = readdir_r(dir, &dirent_buffer, &dirent);
-#pragma GCC diagnostic pop
-    if (result || !dirent) {
+    dirent = readdir(dir);
```
andrewsavage1, line 192: readdir isn't guaranteed to be thread safe, which is why we used readdir_r instead. Unless we're sure that it's not an issue here, we should keep using readdir_r

sacuff: There's a possibly relevant discussion that Rust had over this issue in 2016: https://github.com/rust-lang/rust/issues/34668 and https://github.com/rust-lang/rust/pull/92778/changes#diff-4a74019cd4088485a9e0706be6ccf93ff091cb19d9aad474e4e9d03e2bb8522dR466-R480. 

My reading of that is we're safe here as long as `dir` isn't shared with another thread

andrewsavage1: I think if we're going to rely on the thread safety of readdir, we need to add an nplb test to do so, and add it to the Starboard changelog as well. It's not ideal to do so so late for Starboard 18, but it's probably still okay, especially as we'd be following the standard set by the deprecation of readdir_r.

hferreiro: PR https://github.com/youtube/cobalt/pull/11060 has landed, clearing the way for this one.

andrewsavage1 (approved):

PR #11001, opened 2026-06-22
build: generate JNI / native-library list files via gen_deps

hferreiro:
The jni_sources_list and *_shared_library_list targets are generated_file targets whose list-file outputs are written at `gn gen` time via a metadata walk. They are not added to `deps`: a dep on the native-libs list would force building the .so before this action (the depfile captures the real dep instead), and a dep on the jni_sources_list targets would be circular (they depend back via srcjar_deps). In the default toolchain this is fine since those targets are already generated and their outputs written.

In AOSP, Android is a secondary toolchain, so a generated_file there is only written if something pulls it into generation. gen_deps forces that without adding a build-time dep.

The changes to _invoke_jni_zero are also present upstream since https://crrev.com/c/7168943.

Bug: 495203133

### DIFF
--- a/build/config/android/rules.gni
+++ b/build/config/android/rules.gni
@@ -441,7 +441,7 @@ if (!is_robolectric && enable_java_templates) {
     }
 
     action_with_pydeps(target_name) {
-      forward_variables_from(invoker, TESTONLY_AND_VISIBILITY + [ "deps" ])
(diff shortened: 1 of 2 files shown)


haozheng-cobalt (approved):
Safe to merge since the change is in upstream.

hferreiro:
@haozheng-cobalt note that only part of the changes are upstream.

andrewsavage1 (approved):

PR #11004, opened 2026-06-22
Define STARBOARD only on starboard platforms

andrewsavage1:
Also updates a media test that incorrectly depends on a STARBOARD instead of COBALT define.

Fixed: 527881052

### DIFF
--- a/media/starboard/starboard_utils_test.cc
+++ b/media/starboard/starboard_utils_test.cc
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !defined(STARBOARD)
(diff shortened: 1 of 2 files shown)


kjyoun (approved):

media/starboard/starboard_utils_test.cc
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if !defined(STARBOARD)
+#if !defined(ENABLE_BUILDFLAG_IS_COBALT)
```
borongc, line 15: nit: please update PR descriptions

borongc (approved):

PR #11006, opened 2026-06-22
starboard: make SbSystemGetLocaleId() honor its non-null contract

denisyuji:
The posix and stub implementations could return null, violating the
documented contract and crashing callers that don't guard it (e.g.
GetVoices() trapping under libc++ hardening, setenv() in
environment.cc). Return "" instead.

Issue: 516807166

### DIFF
--- a/starboard/shared/posix/system_get_locale_id.cc
+++ b/starboard/shared/posix/system_get_locale_id.cc
@@ -41,5 +41,7 @@ const char* SbSystemGetLocaleId() {
     }
   }
 
-  return posix_id;
+  // Contract requires non-null (see system_get_locale_id_test.cc); callers
+  // crash on null (e.g. setenv() in environment.cc).
+  return posix_id ? posix_id : "";
 }
--- a/starboard/shared/stub/system_get_locale_id.cc
+++ b/starboard/shared/stub/system_get_locale_id.cc
@@ -15,5 +15,7 @@
 #include "starboard/system.h"
 
 const char* SbSystemGetLocaleId() {
-  return NULL;
+  // The Starboard contract (see starboard/nplb/system_get_locale_id_test.cc)
+  // requires a non-null return.
+  return "";
 }


starboard/shared/stub/system_get_locale_id.cc
```diff
@@ -15,5 +15,7 @@
 #include "starboard/system.h"
 
 const char* SbSystemGetLocaleId() {
-  return NULL;
+  // The Starboard contract (see starboard/nplb/system_get_locale_id_test.cc)
```
andrewsavage1, line 18: In general the stub implementations are not meant to pass nplb. Why are we changing the stub implementation?

starboard/shared/posix/system_get_locale_id.cc
```diff
@@ -41,5 +41,7 @@ const char* SbSystemGetLocaleId() {
     }
   }
 
-  return posix_id;
+  // Contract requires non-null (see system_get_locale_id_test.cc); callers
+  // crash on null (e.g. setenv() in environment.cc).
+  return posix_id ? posix_id : "";
```
andrewsavage1, line 46: I would think we would want to fail if posix_id is null. @jellefoks or @sacuff please have a look

sacuff: Yeah, the implication of the NPLB test is that this should be finding a non-null value to return. I'm surprised we're getting a crash here assuming the platform passes NPLB.

If we do we want to define a default value, IMO it shouldn't be `""` since that could mean different things

https://pubs.opengroup.org/onlinepubs/9799919799/functions/setlocale.html

> ""
> Specifies an implementation-defined native environment.

https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html

> If the LANG environment variable is not set or is set to the empty string, the implementation-defined default locale shall be used.

RE: `setenv()` I assume that's referring to this call https://github.com/youtube/cobalt/blob/main/cobalt/common/libc/stdlib/environment.cc#L106 and that does seem like a bit of a circular logic situation that we might want to deal with, but unrelated to the bug being addressed AFAICT.

denisyuji:
@andrewsavage1 , @sacuff , thank you for your review. I am closing this ticket, since it is expected that this function returns null and fail

PR #11007, opened 2026-06-22
starboard: Fix resource loading endurance leaks

denisyuji:
Resolve memory and file descriptor leaks in the resource loading infrastructure that were identified during endurance testing.

`DataPipeBytesConsumer` was pinning itself as a garbage collector root through its mojo watcher and task runners. This caused abandoned fetches to leak data pipe file descriptors. Moving to weak references allows these objects to be garbage collected when no longer owned.

Additionally, `ResourceLoader` could remain pinned indefinitely if a response body was never consumed by the JavaScript layer. A 5-second timeout now ensures these deferred loads eventually complete, releasing the loader chain and associated resources.

Issue: 516807166

### DIFF
--- a/third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
+++ b/third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
@@ -7,6 +7,7 @@
 #include <algorithm>
 
 #include "base/containers/span.h"
+#include "build/build_config.h"
 #include "base/location.h"
 #include "base/numerics/safe_conversions.h"
 #include "base/task/single_thread_task_runner.h"
@@ -50,7 +51,15 @@ DataPipeBytesConsumer::DataPipeBytesConsumer(
   watcher_.Watch(
       data_pipe_.get(),
       MOJO_HANDLE_SIGNAL_READABLE | MOJO_HANDLE_SIGNAL_PEER_CLOSED,
+#if BUILDFLAG(IS_STARBOARD)
+      // Weak self-reference so an abandoned consumer (e.g. a canceled media
+      // fetch dropped without Cancel()) can be garbage collected instead of
+      // being pinned alive forever, leaking its response-body data pipe's fd.
+      WTF::BindRepeating(&DataPipeBytesConsumer::Notify,
+                         WrapWeakPersistent(this)));
+#else
       WTF::BindRepeating(&DataPipeBytesConsumer::Notify, WrapPersistent(this)));
+#endif
 }
 
 DataPipeBytesConsumer::~DataPipeBytesConsumer() {}
--- a/third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
+++ b/third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
@@ -136,9 +136,18 @@ BytesConsumer::Result DataPipeBytesConsumer::EndRead(size_t read) {
 
   if (has_pending_notification_) {
     has_pending_notification_ = false;
+#if BUILDFLAG(IS_STARBOARD)
+    // Same weak-reference reasoning as the constructor's watcher_.Watch():
+    // a strong Persistent here would re-pin an abandoned consumer for this
+    // task's lifetime. No-ops if the consumer is already collected.
+    task_runner_->PostTask(FROM_HERE,
+                           WTF::BindOnce(&DataPipeBytesConsumer::Notify,
+                                         WrapWeakPersistent(this), MOJO_RESULT_OK));
+#else
     task_runner_->PostTask(FROM_HERE,
                            WTF::BindOnce(&DataPipeBytesConsumer::Notify,
                                          WrapPersistent(this), MOJO_RESULT_OK));
+#endif
   }
   return Result::kOk;
 }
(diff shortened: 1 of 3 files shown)


denisyuji:
Thanks for the catch — addressed in 16409edd944b.

One correction to the suggested patch: `ResponseBodyLoader` doesn't have a `Cancel()` method, only `Abort()` (which is what every other early-exit path in this file already calls — see `HandleError()`/`Cancel()`). Used that instead.

On the infinite-loop mechanics specifically: `DidFinishLoadingBody()` sets `has_seen_end_of_body_ = true` as its very first statement, before it calls into `DidFinishLoading()`. Since that flag is checked synchronously inside the same call stack, `DidFinishLoading()`'s re-defer condition (`!has_seen_end_of_body_`) is already false by the time it runs from this path, so the specific infinite-timer-restart scenario described doesn't reproduce as written. That said, explicitly aborting `response_body_loader_` before forcing completion is still the right call — it makes this path consistent with every other forced-completion path in the file instead of relying on that ordering, and ensures the body drain is torn down deterministically rather than just dropped.

third_party/blink/renderer/platform/loader/fetch/resource_loader.cc
```diff
@@ -416,12 +422,23 @@ void ResourceLoader::DidFinishLoadingBody() {
...
 
+void ResourceLoader::DeferredFinishTimeoutFired(TimerBase*) {
+  if (!deferred_finish_loading_info_) {
+    return;
+  }
+  LOG(WARNING) << "ResourceLoader: force-finishing load stuck in deferred "
+                  "completion after timeout, url="
+               << resource_->Url().GetString().Utf8();
+  DidFinishLoadingBody();
+}
```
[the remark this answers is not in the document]
denisyuji: fixed

third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
```diff
@@ -50,7 +51,15 @@ DataPipeBytesConsumer::DataPipeBytesConsumer(
   watcher_.Watch(
       data_pipe_.get(),
       MOJO_HANDLE_SIGNAL_READABLE | MOJO_HANDLE_SIGNAL_PEER_CLOSED,
+#if BUILDFLAG(IS_STARBOARD)
+      // Weak self-reference so an abandoned consumer (e.g. a canceled media
+      // fetch dropped without Cancel()) can be garbage collected instead of
+      // being pinned alive forever, leaking its response-body data pipe's fd.
+      WTF::BindRepeating(&DataPipeBytesConsumer::Notify,
```
niranjanyardi, line 58: hmm, this seems like a big change to me.

instead of this , is it possible to identify the media fetch calls and ensure that they clean up objects assocuiated with them ?

niranjanyardi: @borongc / @osagie98  do you ahve any suggestions have seen this issue before where media fetches fail ?

niranjanyardi:
Regarding the changes in `third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc` (lines 54–58):

```cpp
+#if BUILDFLAG(IS_STARBOARD)
+      // Weak self-reference so an abandoned consumer (e.g. a canceled media
+      // fetch dropped without Cancel()) can be garbage collected instead of
+      // being pinned alive forever, leaking its response-body data pipe's fd.
+      WTF::BindRepeating(&DataPipeBytesConsumer::Notify,
+                         WrapWeakPersistent(this)));
```

To help minimize upstream Chromium/Blink rebase friction from `#if BUILDFLAG(IS_STARBOARD)` guards, an alternative or complementary approach in Cobalt/Starboard code is addressing the leak at the caller source:

1. **In the Media Player / Starboard Pipeline:** Ensure that whatever media data provider or chunk loader initiates media fetches (e.g., inside `media/blink/` or Cobalt's media pipeline teardown) explicitly invokes `Cancel()` or `Abort()` on active `URLLoader` / `WebAssociatedURLLoader` instances when playback is aborted, paused, or destroyed. Calling `Cancel()` triggers `DataPipeBytesConsumer::Cancel()`, which immediately resets `data_pipe_` and frees the `/dev/shm` file descriptor deterministically rather than waiting for Oilpan GC cleanup.

third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
borongc: Well, I didn't see any of this related crash or issue on ATV via Primes, is this a theoretical fix by LLM agent?

andrewsavage1:
Is this memory leak something that happens in Chrome builds? This change changes upstream code, so it seems to imply that's the case, which would be very surprising to me

third_party/blink/renderer/platform/loader/fetch/data_pipe_bytes_consumer.cc
niranjanyardi: If you read the bug , denis claims that this alongwith 2 other PR's fixes the 12h endurance test crashing on RDK - we want to fix that  - so no its not theorotical

borongc: Two things here:
- This is under blink, it should be `IS_COBALT` instead of `IS_STARBOARD`.
- Like @andrewsavage1, I am surprised as this is Chromium upstream code

If this is really the issue, we should ask Chromium folks for this code change, and fix it by upstream first instead of only in Cobalt.

denisyuji:
Closing it for now since this issue will be tackled in a future ticket

PR #11011, opened 2026-06-22
blink: Implement Configurable Viewport Margin Settings

aelzeiny:
Introduces dynamic, parameter-driven configuration for the compositor pre-raster distance (via kConfigureMaxPrerasterDistance) and Blink display lock margins (via kConfigureDisplayLockMargin) to support viewport memory optimization experiments. Both features are disabled by default and fall back to safe browser baseline settings (1000px and 150%) to ensure clean, non-diluting A/B testing. The testing strategy relies on enabling these flags and overriding their numerical parameters (such as setting distance_in_pixels to 300 and margin_percentage to 30) dynamically via the GCF/Mendel framework or the h5vcc experiments API on the target device.

### DIFF
--- a/cc/base/features.cc
+++ b/cc/base/features.cc
@@ -10,6 +10,10 @@
 #include "base/feature_list.h"
 #include "build/build_config.h"
 
+#ifndef COBALT_DEFAULT_MAX_PRERASTER_DISTANCE_IN_SCREEN_PIXELS
+#define COBALT_DEFAULT_MAX_PRERASTER_DISTANCE_IN_SCREEN_PIXELS 1000
+#endif
+
 namespace features {
 
 namespace {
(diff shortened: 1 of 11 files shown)


cc/base/features.cc
```diff
@@ -249,6 +253,13 @@ BASE_FEATURE(kUseLayerListsByDefault,
              "UseLayerListsByDefault",
              base::FEATURE_DISABLED_BY_DEFAULT);
 
+BASE_FEATURE(kConfigureMaxPrerasterDistance,
```
johnxwork, line 256: We need to surround these code blocks with `#if BUILDFLAG(IS_COBALT)` or something similar to make it clear that this is Cobalt addition. It's a requirement to make future rebases easier.

cc/trees/layer_tree_settings.cc
```diff
@@ -21,7 +21,14 @@ LayerTreeSettings::LayerTreeSettings()
           base::FeatureList::IsEnabled(features::kUseLayerListsByDefault)),
       memory_policy(64 * 1024 * 1024,
                     gpu::MemoryAllocation::CUTOFF_ALLOW_EVERYTHING,
-                    ManagedMemoryPolicy::kDefaultNumResourcesLimit) {}
+                    ManagedMemoryPolicy::kDefaultNumResourcesLimit) {
+  if (base::FeatureList::IsEnabled(features::kConfigureMaxPrerasterDistance)) {
```
johnxwork, line 25: ditto, Cobalt build flag macro

cobalt/build/configs/cobalt.gni
```diff
@@ -23,4 +23,10 @@ declare_args() {
   # TODO(b/505811196): Remove this argument once Cobalt rebased to the milestone
   # where the privacy sandbox components are fully removed.
   enable_privacy_sandbox_apis = true
+
+  # Default maximum pre-raster distance in screen pixels.
+  cobalt_max_preraster_distance_in_screen_pixels = 1000
```
johnxwork, line 28: Are there plans to use this build variable? This variable seems unused? If there's no clear platform-specific tuning demand, then we can maybe just leave it as 1000 universally? We can pass in the experiment value through runtime flags.

PR #11023, opened 2026-06-23
aosp: Use modular_executable for test targets

hferreiro:
Reorder platform detection at the test template so that modular builds on AOSP also use modular_executable.

Bug: 495203133


andrewsavage1 (approved):

andrewsavage1:
Looks like the builds are still failing for this

PR #625, opened 2023-06-13
Googletest - rebase to 1.13 

sideb0ard:
Update GoogleTest to version 1.13 to match Chrome branch-heads/5735

b/280671902

PR #11024, opened 2026-06-23
aosp: Fix GTest output on AOSP

hferreiro:
musl flushes stdout/stderr through __stdio_write, so intercept
stdout/stderr there and route each line through SbLogRaw,
which lands in Android logcat.

Bug: 495203133


hferreiro:
Hey @haozheng-cobalt, @asavage1: this PR might need a few iterations but I wanted an early review.

andrewsavage1:
@y4vor do you know why this wasn't needed in C25?

hferreiro:
@andrewsavage1 In C25, GTest was patched to use SbLog directly.

andrewsavage1:
> @andrewsavage1 In C25, GTest was patched to use SbLog directly.

Ah, thanks for the explanation. Could you give me a code pointer to that customization?

hferreiro:
I think it's this: https://github.com/youtube/cobalt/pull/625/changes#diff-ae82c96247c628e5ac11e86ac903818ded293885bacbe76bc271f4e3fc05dd59R1010.

hferreiro:
That link might not load properly. Try this: https://github.com/sideb0ard/cobalt/blob/a15499ba1cbf55e1c7ae0e0bbc80c90d31d7ec23/third_party/googletest/src/googletest/include/gtest/internal/gtest-port.h#L1010.

starboard/android/shared/posix_emu/file.cc
```diff
@@ -78,4 +82,32 @@ int __wrap_open(const char* path, int oflag, ...) {
   return __wrap_openat(AT_FDCWD, path, oflag);
 }
 
+// Route stdout/stderr writes, which musl flushes via writev, through SbLogRaw
```
andrewsavage1: @jellefoks do you think we should do this or rewrite musl's __stdio_write implementation for starboard to use SbLog/SB_LOG?

jellefoks: Good question. I think it's better to do it in musl than here in the Starboard wrapper.

When it's done here, correct functionality depends on code we can't modify since the implementation lives in code compiled per platform below Starboard. Advantage is that possibly Cobalt will no longer need to know about SbLog, and we can even drop it from `ExportedSymbols`. But disadvantage is that there this now may or may not contain this special case behavior for fd #1 and #2, and there is now is no way for Cobalt to bypass this runtime check for performance, or edge case behavior. For example if we ever need to extend to also cover other stdout/stderr fd's made with `dup2()`, or even more complex situations where the fd's are remapped, for example when an application or process may combine `dup2()` with `open()`/`close()`/`pipe2()` to do trickery for IPC or other reasons.

andrewsavage1: @hferreiro let's change the implementation of musl's __stdio_write instead if that souds good to you

hferreiro: Done.

third_party/musl/src/stdio/__stdio_write.c
```diff
@@ -1,8 +1,46 @@
...
+	// Starboard routes stdout/stderr through SbLogRaw so that platforms that
+	// handle them specially don't miss any output. Coalesce the buffered bytes
+	// with the new data and emit one line at a time, since some log sinks
+	// truncate at a per-entry size limit.
+	if (f->fd == STDOUT_FILENO || f->fd == STDERR_FILENO) {
+		size_t buffered = f->wpos - f->wbase;
+		size_t total = buffered + len;
+		char *data = malloc(total + 1);
+		if (data) {
+			char *p = data, *end = data + total;
```
andrewsavage1: nit: rename p to start or something similar

third_party/musl/src/stdio/__stdio_write.c
```diff
@@ -1,8 +1,46 @@
...
+			memcpy(data, f->wbase, buffered);
+			memcpy(data + buffered, buf, len);
+			data[total] = '\0';
+			while (p < end) {
+				char *nl = memchr(p, '\n', end - p);
```
andrewsavage1: nit: rename nl to newline

third_party/musl/src/stdio/__stdio_write.c
```diff
@@ -1,8 +1,46 @@
...
+#if defined(STARBOARD)
+	// Starboard routes stdout/stderr through SbLogRaw so that platforms that
+	// handle them specially don't miss any output. Coalesce the buffered bytes
+	// with the new data and emit one line at a time, since some log sinks
+	// truncate at a per-entry size limit.
+	if (f->fd == STDOUT_FILENO || f->fd == STDERR_FILENO) {
+		size_t buffered = f->wpos - f->wbase;
+		size_t total = buffered + len;
+		char *data = malloc(total + 1);
+		if (data) {
```
andrewsavage1: Let's add a few comments in this code to make the pointer math and buffering more clear

third_party/musl/src/stdio/__stdio_write.c
```diff
@@ -1,8 +1,46 @@
...
 {
+#if defined(STARBOARD)
+	// Starboard routes stdout/stderr through SbLogRaw so that platforms that
+	// handle them specially don't miss any output. Coalesce the buffered bytes
+	// with the new data and emit one line at a time, since some log sinks
+	// truncate at a per-entry size limit.
+	if (f->fd == STDOUT_FILENO || f->fd == STDERR_FILENO) {
+		size_t buffered = f->wpos - f->wbase;
+		size_t total = buffered + len;
+		char *data = malloc(total + 1);
```
andrewsavage1: I wasn't thinking about the fact that we'd have to allocate memory when change the implementation of __stdio_write—it makes this solution a little less appealing. One thing we could do is to have a small buffer (i.e. `char[512]`) that we use if the len is small enough, but it does make this implementation more complex. @jellefoks could you share your thoughts on this?

jellefoks: Yes, that's a valid point. We'd really prefer this code to still work in an OOM situation, plus malloc isn't an expected side-effect for write(). We can easily afford a statically allocated buffer long enough for the worst case line length (e.g. easily even a whole page of 4096 chars), and even wrapping of lines in those rare but not impossible case that some are longer than our buffer is acceptable behavior.

hferreiro: Done.

third_party/musl/src/stdio/__stdio_write.c
```diff
@@ -1,8 +1,50 @@
...
 {
+#if defined(STARBOARD)
+	// Starboard routes stdout/stderr through SbLogRaw so that platforms that
+	// handle them specially (e.g. Android logcat) don't miss any output.
+	if (f->fd == STDOUT_FILENO || f->fd == STDERR_FILENO) {
+		// Two spans to drain, in write order: the bytes already pending in
+		// the stream's own buffer (f->wbase..f->wpos) from earlier writes,
+		// then the newly written data (buf/len).
+		const unsigned char *spans[2] = {f->wbase, buf};
+		size_t span_lens[2] = {f->wpos - f->wbase, len};
```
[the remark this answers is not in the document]
sacuff: IMO both these AI comments can be ignored since the exact same pointer arithmetic is being done in the regular musl code

third_party/musl/src/stdio/__stdio_write.c
```diff
@@ -1,8 +1,50 @@
...
+		// Accumulate into a fixed on-stack buffer and flush a line at a time, to
+		// avoid truncation. One byte is kept for the NUL.
+		char line[4096];
+		size_t n = 0;
+		for (int i = 0; i < sizeof spans / sizeof *spans; i++) {
```
andrewsavage1: nit: i should be a size_t

andrewsavage1: nit: pref `sizeof(foo)` (that is, with parentheses) for readability

andrewsavage1 (approved):

third_party/musl/src/stdio/__stdio_write.c
hferreiro: Done.

hferreiro: Done.

PR #11025, opened 2026-06-23
android: Extract FeatureList init to a shared method

hferreiro:
On AOSP, elf_loader_sandbox will need to initialize the FeatureList with its default values before it loads libnplb.so. Some suites use parameterized tests whose value generators query Starboard features and gtest evaluates those generators during test registration, which runs before the test Environment::SetUp().

In this PR, that initialization, already used by StarboardTestEnvironment, is extracted to a shared place so that it can be later reused by AOSP.

Bug: 495203133

### DIFF
--- a/starboard/shared/starboard/features.cc
+++ b/starboard/shared/starboard/features.cc
@@ -24,13 +24,6 @@
 
 #include "starboard/shared/starboard/features.h"
 
-#include <stdint.h>
-
-#include <iterator>
-
-#include "starboard/extension/features.h"
-#include "starboard/shared/starboard/feature_list.h"
-
 #define STARBOARD_FEATURE(feature, name, default_state) \
   SB_FEATURE(feature, name, default_state)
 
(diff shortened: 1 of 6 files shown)


starboard/shared/starboard/features.cc
```diff
@@ -24,6 +24,13 @@
 
 #include "starboard/shared/starboard/features.h"
 
+#include <stdint.h>
```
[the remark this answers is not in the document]
hferreiro: stdint.h is the recommended header so that types are defined in the global namespace: https://groups.google.com/a/chromium.org/g/cxx/c/Bc1XyEL9pp4/m/8g0Jx-YVCAAJ.

starboard/shared/starboard/features.cc
```diff
@@ -48,3 +55,45 @@
 #undef FEATURE_PARAM_LIST_START
 #undef FEATURE_PARAM_LIST_END
 #undef STARBOARD_FEATURE_PARAM_TIME_TYPE
+
+namespace starboard::features {
+
+namespace {
+
+// Aggregate the just-defined features/params into arrays for the FeatureList
```
johnedocampo: Please add back the original comment from `starboard/android/shared/starboard_test_environment.cc.`

```
// The initialization function accepts an array of SbFeatures and
// SbFeatureParams as arguments, so these macros help create that array to pass
// to the initializer. To see more info about how these macros work and how they
// help initialize the feature list instance, please refer to
// //starboard/extension/feature_config.h for more info.
//
// We disable clang-format here as there are issues between
// clang-format and cpplint when running pre-commit on this code.
```

starboard/shared/starboard/features.cc
```diff
@@ -48,3 +55,45 @@
 #undef FEATURE_PARAM_LIST_START
 #undef FEATURE_PARAM_LIST_END
 #undef STARBOARD_FEATURE_PARAM_TIME_TYPE
+
+namespace starboard::features {
```
johnedocampo: Is it possible to have this code be in it's own test file, like `starboard/shared/starboard/features_test_util.h/cc`? I think this will help better distinguish that this code was constructed for testing purposes.

hferreiro: Makes sense. Done.

starboard/shared/starboard/features.cc
hferreiro: Done.

haozheng-cobalt (commented):
this in general looks to be a good idea to me. Good to go after addressing all other review comments.

andrewsavage1 (approved):
Approving and enabling automerge since comments were addressed

johnedocampo (approved):
lgtm!

starboard/shared/starboard/features_test_util.cc
```diff
@@ -0,0 +1,71 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/shared/starboard/features_test_util.h"
+
+#include <stdint.h>
```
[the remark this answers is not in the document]
hferreiro: Chromium favours stdint.h: https://groups.google.com/a/chromium.org/g/cxx/c/Bc1XyEL9pp4/m/Ie4C-_AjCAAJ.

PR #11027, opened 2026-06-23
Enable v8 memory saving features for 3p

tasunnn:
Three flags are set in Android but were not passed in for 3p, they have showed significant memory improvement from C26 experiments: 

- `optimize-for-size`: enable v8 memory saving mode
- `initial-old-space-size=64` and `max-old-space-size=512`: configure initial old space size and max old space size

Adding two new flags from recent experiment: 
- `disable-optimizing-compilers` and `no-sparkplug`: disable v8 optimizing compilers (i.e. turbofan, maglev, and sparkplug) that are not needed for cobalt. 

Bug: 512163617

### DIFF
--- a/cobalt/app/cobalt_switch_defaults_starboard.cc
+++ b/cobalt/app/cobalt_switch_defaults_starboard.cc
@@ -121,6 +121,14 @@ CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
        // Disable decommitting pooled pages to prevent virtual memory
        // fragmentation.
        "--no-decommit-pooled-pages "
+       // Enable memory saving mode with little v8 performance tradeoff.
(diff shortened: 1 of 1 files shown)


Awallky (commented):
Should we run these as an experiment in EG before hard-coding these? I don't think we should assume the memory footprint of Android and EG are the same.

yell0wd0g (approved):
RS LGTM!

tasunnn:
> Should we run these as an experiment in EG before hard-coding these? I don't think we should assume the memory footprint of Android and EG are the same.

yeah but given that we don't have enough traffic and the time it takes to run experiments, we could make the change first since we already see improvements on Android and could always revert it if any regressions, will still optimize the values here in the future but not a priority now.

PR #11029, opened 2026-06-23
Enable stack sanitization for Android TV

alunev:
android: Enable stack sanitization for Android

Enable stack sanitization for Cobalt on Android to ensure minidumps
replace stack memory content with the 0x0DEFACED pattern. This aligns
Android TV crash reporting with the behavior already established in the
Starboard crashpad wrapper for third-party platforms.

Furthermore, disable periodic tasks during Crashpad initialization on
Android to prevent the creation of unnecessary background threads
during crash handling.

Issue: 481729814

### DIFF
--- a/cobalt/app/cobalt_crash_reporter_client.cc
+++ b/cobalt/app/cobalt_crash_reporter_client.cc
@@ -96,3 +96,23 @@ bool CobaltCrashReporterClient::IsRunningUnattended() {
 std::string CobaltCrashReporterClient::GetUploadUrl() {
   return kCrashReportUrl;
 }
+
+// Overrides the default behavior to explicitly enable stack sanitization.
+// Setting *sanitize_stacks = true ensures that all stack memory in the
+// minidump is overwritten with the 0x0DEFACED pattern.
+// This aligns the native Android crashpad path with the
+// behavior established in the Starboard crashpad wrapper for 3P platforms.
+void CobaltCrashReporterClient::GetSanitizationInformation(
+    const char* const** allowed_annotations,
+    void** target_module,
+    bool* sanitize_stacks) {
+  if (allowed_annotations) {
+    *allowed_annotations = nullptr;
+  }
+  if (target_module) {
+    *target_module = nullptr;
+  }
+  if (sanitize_stacks) {
+    *sanitize_stacks = true;
+  }
+}
--- a/components/crash/core/app/crashpad_android.cc
+++ b/components/crash/core/app/crashpad_android.cc
@@ -567,6 +567,13 @@ class HandlerStarter {
       return database_path;
     }
 
+#if BUILDFLAG(IS_COBALT)
+    // Disable periodic tasks. In the Android "at-crash" execution model, this
+    // flag prevents an unnecessary file system scan for pending reports, reducing
+    // thread contention during the critical crash dumping window.
+    // TODO: Implement actual database pruning for Android TV.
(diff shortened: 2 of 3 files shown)


cobalt/app/cobalt_crash_reporter_client.cc
```diff
@@ -96,3 +96,12 @@ bool CobaltCrashReporterClient::IsRunningUnattended() {
...
   return kCrashReportUrl;
 }
+
+void CobaltCrashReporterClient::GetSanitizationInformation(
+    const char* const** allowed_annotations,
+    void** target_module,
+    bool* sanitize_stacks) {
+  *allowed_annotations = nullptr;
+  *target_module = nullptr;
+  *sanitize_stacks = true;
```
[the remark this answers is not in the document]
alunev: thx, current caller passes local var pointer but I've added defensive checks.

components/crash/core/app/crashpad_android.cc
```diff
@@ -567,6 +567,9 @@ class HandlerStarter {
       return database_path;
     }
 
+#if BUILDFLAG(IS_COBALT)
+    arguments.push_back("--no-periodic-tasks");
```
hlwarriner, line 575: Are we pruning the Crashpad database on Android TV? We need to make sure the reports are cleaned up and don't accumulate without limit.

#7358 added pruning for Cobalt 3P, but that mechanism is guarded by `BUILDFLAG(IS_NATIVE_TOOLCHAIN)`.

hlwarriner (approved):
The sanitization changes lgtm.

Please make sure we have a plan for database pruning on Android TV. If disabling periodic tasks is a regression w.r.t. database pruning, meaning Android TV was pruning the database with periodic tasks enabled, then please revert that change from this PR before merging.

components/crash/core/app/crashpad_android.cc
alunev: hi, `--no-periodic-tasks` doesn't affect pruning. However pruning is currently not working on ATV, created 528020524 to fix that.

zhongqiliang: Please add a comment on top of this code. For this is clean up the dead code. 
And add a TODO somewhere for the actual implementation.

zhongqiliang (approved):

components/crash/core/app/crashpad_android.cc
alunev: thx, added comment

cobalt/app/cobalt_crash_reporter_client.cc
```diff
@@ -96,3 +96,18 @@ bool CobaltCrashReporterClient::IsRunningUnattended() {
 std::string CobaltCrashReporterClient::GetUploadUrl() {
   return kCrashReportUrl;
 }
+
+void CobaltCrashReporterClient::GetSanitizationInformation(
```
andrewsavage1, line 105: I don't understand what this function is doing—my best guess is that it overrides any presets limiting what stacks to annotate and sanitizes everything, but I don't understand why this is necessary either from the PR description or the function name, so I think a comment should be added

alunev: thanks, yes, we override default `GetSanitizationInformation()` defined in components/crash/core/app/crash_reporter_client.cc. Added comment.

andrewsavage1 (approved):

PR #11030, opened 2026-06-23
Fix CodeOther memory categorization and improve UMA reporting stability

Awallky:
The root cause of the memory accounting regression was missing patterns for .so, .apk, and .dex regions in CobaltDetailedMetricsDelegate, causing them to be misclassified as 'other' instead of 'code_other'.

Additionally, ReportUma was occasionally skipping reports due to a strict timer delay check; added a 2x multiplier to account for minor scheduling delays on Android.

Bug: 527074178


cobalt/memory/cobalt_memory_attribution_manager.cc
```diff
@@ -98,9 +98,9 @@ void CobaltMemoryAttributionManager::ReportUma() {
   auto* observer = base::memory::CobaltMemoryAttributionObserver::Get();
   // Skip reporting if timer was significantly delayed (e.g. device suspension).
   if ((now - last_report_time_) >
-      base::Seconds(
-          cobalt::features::kCobaltMemoryAttributionReportIntervalParam
-              .Get())) {
+      2 * base::Seconds(
```
johnxwork, line 101: The choice seems arbitrary and can be broken on slower devices? But if we don't depend on these for critical functionalities(e.g. just logging UMA logging purpose) maybe it's fine.

johnxwork (approved):

cobalt/memory/cobalt_memory_attribution_manager.cc
Awallky: Our default period is 5 minutes. If we go beyond 10 minutes, I don't think this will be our main concern. There is probably something severely broken on the device or in the app in this case.

PR #11036, opened 2026-06-24
[Profiling] Implement thread-safe in-process heap profiler gated under non-release builds

haozheng-cobalt:
Introduce in-process heap profiling and startup Perfetto tracing to
facilitate comparative memory footprint analysis. These features are
restricted to non-release builds.

To prevent thread-affinity crashes when executing dumps on the
background thread, memory instrumentation components are refactored to
use thread-safe Mojo clients. Manual tracing initialization is added
to the browser startup sequence to account for Cobalt bypassing
standard Chromium paths.

Bug: 528397803

Test:
Wrote a script to verify if a trace contains heap profiler allocations cobalt/tools/performance/memory/verify_memory_trace.py in PR https://github.com/youtube/cobalt/pull/11090.

And previously Cobalt profile results miss heap profiler allocations: https://paste.googleplex.com/5447990117531648
With the in-process heap profiling implementation in this PR: https://paste.googleplex.com/6163417349095424

Bug: 528397803

### DIFF
--- a/cobalt/browser/cobalt_browser_main_parts.cc
+++ b/cobalt/browser/cobalt_browser_main_parts.cc
@@ -140,7 +140,7 @@ void RegisterCobaltHeapProfilerOnDumpThread() {
   auto params = heap_profiling::mojom::ProfilingParams::New();
   params->sampling_rate = 128 * 1024;  // 128KB sampling rate
   params->stack_mode =
-      heap_profiling::mojom::StackMode::NATIVE_WITHOUT_THREAD_NAMES;
+      heap_profiling::mojom::StackMode::NATIVE_WITH_THREAD_NAMES;
 
   (*g_profiling_service)
       ->AddProfilingClient(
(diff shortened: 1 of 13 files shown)


cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -90,6 +101,71 @@ void InitializeBrowserMemoryInstrumentationClient() {
...
+  static base::NoDestructor<heap_profiling::ProfilingClient> g_profiling_client;
+  mojo::PendingRemote<heap_profiling::mojom::ProfilingClient> client_remote;
+  g_profiling_client->BindToInterface(
+      client_remote.InitWithNewPipeAndPassReceiver());
+
+  // 4. Add our process as a profiling client to the profiling service.
+  auto params = heap_profiling::mojom::ProfilingParams::New();
+  params->sampling_rate = 128 * 1024;  // 128KB sampling rate
+  params->stack_mode =
+      heap_profiling::mojom::StackMode::NATIVE_WITHOUT_THREAD_NAMES;
```
Awallky: Do we want to keep the thread name? Might be useful, since e do not have multiple processes..

haozheng-cobalt: Thanks for catching this! I'll update the script to record the stack in trace.

Awallky (approved):

PR #11037, opened 2026-06-24
[Profiling] Create in-process heap profiling symbolizer

haozheng-cobalt:
Introduce a DWARF callstack symbolizer to support automated performance
memory tools. This allows profiling and comparing heap footprints by
resolving hexadecimal program counters in traces into human-readable
C++ symbols using local unstripped binaries and llvm-symbolizer.

The change includes a Python script for processing traces and unit
tests to verify the symbolization pipeline.

Test: output of the script will be like https://paste.googleplex.com/5340110672494592, example trace file  is too large to share. Added unittests for the symbolizer too.

Bug: 528397803

### DIFF
--- a/cobalt/tools/performance/memory/symbolize_in_process_heap.py
+++ b/cobalt/tools/performance/memory/symbolize_in_process_heap.py
@@ -1,0 +1,304 @@
+# Copyright 2026 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Cobalt DWARF Callstack Symbolizer.
+
+This script parses and symbolizes hexadecimal program counters inside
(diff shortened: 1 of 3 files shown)


cobalt/tools/performance/memory/symbolize_in_process_heap.py
```diff
@@ -0,0 +1,304 @@
+# Copyright 2026 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
```
Awallky, line 7: nit: Can we reduce the spaces here?

Awallky (approved):
LGTM % comment

I'd like to see a follow-up to this, documenting how to include this in a workflow and when this would be most useful. 
Are there any follow-up changes that will enable this to be fully automated? Or is this only intended to simplify local profiling runs to include symbolized stacks?

cobalt/tools/performance/memory/symbolize_in_process_heap.py
haozheng-cobalt: hmm if you look at other python files in Cobalt, I think they all use same formatting as this one: https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/tools/performance/memory/compare_accuracy.py;l=9

haozheng-cobalt:
> LGTM % comment
> 
> I'd like to see a follow-up to this, documenting how to include this in a workflow and when this would be most useful. Are there any follow-up changes that will enable this to be fully automated? Or is this only intended to simplify local profiling runs to include symbolized stacks?

Yes I agree, I'm looking into how to make the heap profiler more useful and introduce some automation to help monitor the memory health.

PR #11038, opened 2026-06-24
cleanup: Replace bug links with short descriptions

johnedocampo:
Remove internal bug tracker references in the Android media
capabilities cache and its associated tests. Replace these links with
descriptive comments explaining why low-performance software decoders
are rejected, providing better context directly within the codebase.

Bug: 508269486

### DIFF
--- a/starboard/android/shared/media_capabilities_cache.cc
+++ b/starboard/android/shared/media_capabilities_cache.cc
@@ -495,7 +495,8 @@ std::string MediaCapabilitiesCache::FindVideoDecoder(
       continue;
     }
     // Reject low performance software codec if software codec is not required.
-    // See b/456473829 for more details.
+    // This can help ensure that a hardware decoder is chosen when a device
+    // reports support for a low performing software decoder.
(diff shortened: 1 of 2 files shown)


starboard/android/shared/media_capabilities_cache.cc
```diff
@@ -495,7 +495,8 @@ std::string MediaCapabilitiesCache::FindVideoDecoder(
       continue;
     }
     // Reject low performance software codec if software codec is not required.
-    // See b/456473829 for more details.
+    // This can help ensure that a higher performing software decoder or
```
borongc: `// This can help ensure that a hardware decoder is chosen`
`// due to some devices that report a low performing software decoder.`

Please rephrase it. I think SW decoder is rejected anyway, correct me if I'm wrong.

borongc (approved):

starboard/android/shared/media_capabilities_cache.cc
johnedocampo: Rephrased.

Your comment makes more sense. If a device did have a better software decoder, this check would not be necessary.

PR #11039, opened 2026-06-24
fix: Race condition between focus and visibilityChange events during resume.

jellefoks:
The visibilityChange event to 'visible' could get raced by the focus event, resulting in focus being signaled before visible. This ensures correct sequencing.

1. Instantiates CobaltLifecycleManager's WebContentsTracker early during DidFinishNavigation to ensure we correctly capture navigating frames, preventing a race condition where focus/visibility JS events were fired out-of-order.

2. Restricts the browser-side lifecycle ACK handshake to only track the active Primary Main Frame (via GetPrimaryMainFrame()) during downward transitions. This excludes throttled subframes (iframes) and leaked navigated-away main frames pending deletion, eliminating a 4-second deadlock safety timeout and reducing transition latencies from 4.0s to ~30ms.

Bug: 491962441
Bug: 475990372
Bug: 448156280

### DIFF
--- a/cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
+++ b/cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
@@ -59,7 +59,7 @@ void CobaltLifecycleManager::BindReceiver(
       content::WebContents::FromRenderFrameHost(frame);
   mojo::ReceiverId id =
       receivers_.Add(this, std::move(receiver), {frame->GetGlobalId()});
-  active_receivers_[frame->GetGlobalId()] = id;
+  active_receiver_counts_[frame->GetGlobalId()]++;
   receiver_ids_[web_contents].push_back(id);
 }
 
@@ -83,17 +83,25 @@ void CobaltLifecycleManager::InitializeTracker(
 
 void CobaltLifecycleManager::OnMojoDisconnect() {
   CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  FrameContext context = receivers_.current_context();
+
+  auto active_it = active_receiver_counts_.find(context.frame_id);
+  if (active_it != active_receiver_counts_.end()) {
+    active_it->second--;
+    if (active_it->second <= 0) {
+      active_receiver_counts_.erase(active_it);
+      content::RenderFrameHost* frame =
+          content::RenderFrameHost::FromID(context.frame_id);
+      if (frame) {
+        UnregisterFrame(content::WebContents::FromRenderFrameHost(frame),
+                        frame);
+      }
+    }
+  }
+
   auto [frame, web_contents] = GetCurrentContext();
   mojo::ReceiverId id = receivers_.current_receiver();
-  FrameContext context = receivers_.current_context();
   if (web_contents) {
-    auto active_it = active_receivers_.find(context.frame_id);
-    if (active_it != active_receivers_.end() && active_it->second == id) {
-      if (frame) {
-        UnregisterFrame(web_contents, frame);
-      }
-      active_receivers_.erase(active_it);
-    }
     auto it = receiver_ids_.find(web_contents);
     if (it != receiver_ids_.end()) {
       auto& ids = it->second;
@@ -125,12 +133,8 @@ void CobaltLifecycleManager::WebContentsTracker::RenderFrameCreated(
   // renderer-side CobaltLifecycleController. This establishes the two-way
   // communication channel needed for lifecycle ACKs.
   mojo::PendingRemote<cobalt::mojom::CobaltLifecycleObserver> observer;
-  // Track the ReceiverId associated with the WebContents for clean teardown.
-  mojo::ReceiverId id = manager_->receivers_.Add(
-      manager_, observer.InitWithNewPipeAndPassReceiver(),
-      {render_frame_host->GetGlobalId()});
-  manager_->active_receivers_[render_frame_host->GetGlobalId()] = id;
-  manager_->receiver_ids_[web_contents()].push_back(id);
+  manager_->BindReceiver(render_frame_host,
+                         observer.InitWithNewPipeAndPassReceiver());
   controller->SetObserver(std::move(observer));
 
   controllers_[render_frame_host] = std::move(controller);
@@ -272,11 +276,7 @@ void CobaltLifecycleManager::WebContentsTracker::Rebind(
   // Re-register the browser as an observer after re-binding the interface.
   mojo::PendingRemote<cobalt::mojom::CobaltLifecycleObserver> observer;
   // Track the ReceiverId associated with the WebContents for clean teardown.
-  mojo::ReceiverId id = manager_->receivers_.Add(
-      manager_, observer.InitWithNewPipeAndPassReceiver(),
-      {frame->GetGlobalId()});
-  manager_->active_receivers_[frame->GetGlobalId()] = id;
-  manager_->receiver_ids_[web_contents()].push_back(id);
+  manager_->BindReceiver(frame, observer.InitWithNewPipeAndPassReceiver());
   controller->SetObserver(std::move(observer));
 
   controllers_[frame] = std::move(controller);
@@ -326,9 +326,12 @@ void CobaltLifecycleManager::RegisterFrame(content::WebContents* web_contents,
   frames_[web_contents].insert(frame);
   GetOrCreateTracker(web_contents);
 
-  // Add to pending set if we are waiting for an ACK.
+  // If we are currently waiting for an ACK, only dynamically track newly
+  // registered Main Frames (e.g. during active navigations) to resolve the
+  // focus/visibility event race. Subframes (iframes) must be excluded to
+  // prevent Blink scheduler throttling deadlocks.
   PendingAck pending_ack = pending_acks_[web_contents];
-  if (pending_ack != PendingAck::kNone) {
+  if (pending_ack != PendingAck::kNone && !frame->GetParent()) {
     pending_ack_frames_[web_contents].insert(frame);
   }
 
@@ -349,6 +352,9 @@ void CobaltLifecycleManager::UnregisterFrame(content::WebContents* web_contents,
   if (main_frames_[web_contents] == frame) {
     main_frames_.erase(web_contents);
   }
+
+  // Clean up active receiver counts for this frame.
+  active_receiver_counts_.erase(frame->GetGlobalId());
 
   if (frames_[web_contents].empty()) {
     frames_.erase(web_contents);
@@ -460,15 +466,22 @@ void CobaltLifecycleManager::StartWaitingForAck(
         base::BindOnce(&CobaltLifecycleManager::NotifyStartWaitingForReveal,
                        base::Unretained(this), web_contents->GetWeakPtr()));
   } else {
-    for (auto* frame : frames_[web_contents]) {
-      bool connected = tracker->IsConnected(frame);
+    // For downward transitions, only track the active Primary Main Frame.
+    // This excludes subframes (which can be aggressively throttled or paused
+    // by Blink's scheduler) and older, navigated-away main frames that are
+    // still leaking in memory pending deletion.
+    auto* primary_main_frame = web_contents->GetPrimaryMainFrame();
+    if (primary_main_frame) {
+      bool connected = tracker->IsConnected(primary_main_frame);
       if (connected) {
-        pending_ack_frames_[web_contents].insert(frame);
-      } else if (ack_type == PendingAck::kUnfreeze) {
-        LOG(WARNING) << "StartWaitingForAck: Frame not connected during "
-                        "Unfreeze! Re-binding frame="
-                     << frame;
-        tracker->Rebind(frame);
+        pending_ack_frames_[web_contents].insert(primary_main_frame);
+      } else if (ack_type == PendingAck::kUnfreeze &&
+                 primary_main_frame->IsRenderFrameLive()) {
+        LOG(WARNING)
+            << "StartWaitingForAck: Primary Main Frame not connected during "
+               "Unfreeze! Re-binding frame="
+            << primary_main_frame;
+        tracker->Rebind(primary_main_frame);
       }
     }
   }
@@ -584,12 +597,12 @@ void CobaltLifecycleManager::OnWebContentsDestroyed(
   // The WebContents is being destroyed. Proactively remove all its Mojo
   // receivers from `receivers_` BEFORE the WebContents and its RFHs go out of
   // scope.
-  // Proactively clean up the active_receivers_ map for all frames belonging to
-  // this WebContents to avoid memory leaks.
+  // Proactively clean up the active_receiver_counts_ map for all frames
+  // belonging to this WebContents to avoid memory leaks.
   auto frames_it = frames_.find(web_contents);
   if (frames_it != frames_.end()) {
     for (auto* frame : frames_it->second) {
-      active_receivers_.erase(frame->GetGlobalId());
+      active_receiver_counts_.erase(frame->GetGlobalId());
     }
   }
   auto it = receiver_ids_.find(web_contents);
--- a/cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
+++ b/cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
@@ -85,8 +85,8 @@ void CobaltLifecycleManager::OnMojoDisconnect() {
   CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
   FrameContext context = receivers_.current_context();
 
-  auto active_it = active_receiver_counts_.find(context.frame_id);
-  if (active_it != active_receiver_counts_.end()) {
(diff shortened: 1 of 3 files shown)


jellefoks:
The flakes are not gone with this change. Maybe less frequent.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -117,6 +130,14 @@ void CobaltLifecycleManager::WebContentsTracker::RenderFrameDeleted(
   manager_->UnregisterFrame(web_contents(), render_frame_host);
 }
 
+void CobaltLifecycleManager::WebContentsTracker::DidFinishNavigation(
+    content::NavigationHandle* navigation_handle) {
+  if (navigation_handle->HasCommitted()) {
+    content::RenderFrameHost* rfh = navigation_handle->GetRenderFrameHost();
+    RenderFrameCreated(rfh);
+  }
+}
```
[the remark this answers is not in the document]
jellefoks: Done. Guarded DidFinishNavigation to only call RenderFrameCreated if the frame is not already in controllers_. This prevents double-registration, Mojo remote overwrites, and the resulting active_receiver_counts_ leaks.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -71,9 +73,20 @@ CobaltLifecycleManager::GetCurrentContext() {
 
 void CobaltLifecycleManager::OnMojoDisconnect() {
```
kjyoun, line 84: can you flatten the logic?


```
void CobaltLifecycleManager::OnMojoDisconnect() {
  CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
  FrameContext context = receivers_.current_context();

  auto it = active_receiver_counts_.find(context.frame_id);
  if (it == active_receiver_counts_.end()) {
    return;
  }

  it->second--;
  if (it->second > 0) {
    return;
  }

  active_receiver_counts_.erase(it);

  content::RenderFrameHost* frame =
      content::RenderFrameHost::FromID(context.frame_id);
  if (!frame) {
    return;
  }

  UnregisterFrame(content::WebContents::FromRenderFrameHost(frame), frame);
}
```

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -117,6 +130,18 @@ void CobaltLifecycleManager::WebContentsTracker::RenderFrameDeleted(
   manager_->UnregisterFrame(web_contents(), render_frame_host);
 }
 
+void CobaltLifecycleManager::WebContentsTracker::DidFinishNavigation(
```
kjyoun, line 161: nit: flatteing?

```
void CobaltLifecycleManager::WebContentsTracker::DidFinishNavigation(
    content::NavigationHandle* navigation_handle) {
  if (!navigation_handle->HasCommitted()) {
    return;
  }

  content::RenderFrameHost* rfh = navigation_handle->GetRenderFrameHost();
  
  // Only register the frame if it is not already tracked. Unconditional calls
  // would trigger duplicate Mojo bindings and leak receiver counts.
  if (controllers_.find(rfh) != controllers_.end()) {
    return;
  }

  RenderFrameCreated(rfh);
}
```

kjyoun (commented):
Can you flatten the logic? 

Per Gemini

---

flattening code like this (often called using guard clauses or early returns) is widely considered a best practice in modern programming, especially in languages like C++.

Here is why it's generally better:

- Reduces Cognitive Load: When reading nested if statements, you have to hold all the active conditions in your head at once. With early returns, you handle one edge case (or error), mentally discard it, and move on.

- Highlights the "Happy Path": In the flattened version, the main goal of the function is pushed to the very bottom and sits at the base indentation level. It makes it instantly obvious what the function is actually supposed to do when things go right.

- Prevents "Arrow Code": Deeply nested code starts to look like an arrow pointing to the right (>). This forces you to scroll horizontally and makes the code harder to read.

- Cleaner Version Control (Git Diffs): If you add or remove a condition in a deeply nested function, you often have to re-indent the entire block of code inside it. This creates a messy Git diff where every line looks like it changed. With guard clauses, you just add or remove the two lines for that specific check.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
jellefoks: We actually can't flatten this completely anymore because we need to perform cleanup on `receiver_ids_` at the end of the method to prevent shutdown crashes (introduced in PR #10964). If we return early here when the active receiver count is > 0, we would skip that crucial cleanup!

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
jellefoks: Done! Flattened the logic here.

jellefoks:
I have pushed an update that correctly restores the base branch (#10964) logic for `DidFinishNavigation`, which natively flattens the method as requested while properly handling `IsErrorPage()` and `IsSameDocument()`. This resolves the `evergreen` CI test crash on `AppEventRunnerTest.OnUnfreeze` caused by my previous commit.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -460,15 +466,22 @@ void CobaltLifecycleManager::StartWaitingForAck(
...
+      bool connected = tracker->IsConnected(primary_main_frame);
       if (connected) {
-        pending_ack_frames_[web_contents].insert(frame);
-      } else if (ack_type == PendingAck::kUnfreeze) {
-        LOG(WARNING) << "StartWaitingForAck: Frame not connected during "
-                        "Unfreeze! Re-binding frame="
-                     << frame;
-        tracker->Rebind(frame);
+        pending_ack_frames_[web_contents].insert(primary_main_frame);
+      } else if (ack_type == PendingAck::kUnfreeze &&
```
kjyoun: super nit: since  `connected` is not used/

```
     if (tracker->IsConnected(primary_main_frame)) {
        pending_ack_frames_[web_contents].insert(primary_main_frame);
      }
```

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -460,15 +466,22 @@ void CobaltLifecycleManager::StartWaitingForAck(
...
+                 primary_main_frame->IsRenderFrameLive()) {
+        LOG(WARNING)
+            << "StartWaitingForAck: Primary Main Frame not connected during "
```
kjyoun: if you want to repeat method name? how about using `__func__` macros


https://source.corp.google.com/search?q=__func__&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:cobalt%2F

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -350,6 +353,9 @@ void CobaltLifecycleManager::UnregisterFrame(content::WebContents* web_contents,
     main_frames_.erase(web_contents);
   }
 
+  // Clean up active receiver counts for this frame.
```
kjyoun: nit: Do we need this comment? The code is pretty self-explanatory here.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -326,9 +326,12 @@ void CobaltLifecycleManager::RegisterFrame(content::WebContents* web_contents,
   frames_[web_contents].insert(frame);
   GetOrCreateTracker(web_contents);
 
-  // Add to pending set if we are waiting for an ACK.
+  // If we are currently waiting for an ACK, only dynamically track newly
+  // registered Main Frames (e.g. during active navigations) to resolve the
+  // focus/visibility event race. Subframes (iframes) must be excluded to
+  // prevent Blink scheduler throttling deadlocks.
```
kjyoun: if `GetParent()` is nullptr, we dont execute any logic below.

Then how about this?

```
  if (!frame->GetParent()) {
    // Add comment about why we can fast-exit here
     return;
  }


  if (pending_acks_[web_contents] != PendingAck::kNone)
    pending_ack_frames_[web_contents].insert(frame);
  }

  main_frames_[web_contents] = frame;
  for (auto& observer : observers_) {
    observer.OnMainFrameRegistered(web_contents);
  }
}

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.h [hunk 8]
```diff
@@ -281,12 +281,11 @@ class CobaltLifecycleManager : public cobalt::mojom::CobaltLifecycleObserver {
...
-  // GlobalId). This prevents race conditions where an asynchronous disconnect
-  // handler for an old Mojo pipe unregisters a frame that has already been
-  // re-bound to a new pipe.
-  std::map<content::GlobalRenderFrameHostId, mojo::ReceiverId>
-      active_receivers_;
+  // Tracks the number of active Mojo lifecycle observer receivers (connections)
+  // associated with each RenderFrameHost (keyed by its global ID). This is used
+  // to dynamically monitor frame connection lifetimes and determine when a
+  // frame is actively participating in the lifecycle transition handshake.
+  std::map<content::GlobalRenderFrameHostId, int> active_receiver_counts_;
```
kjyoun: per https://chromium.googlesource.com/chromium/src/+/HEAD/base/containers/README.md

> If you just need a generic map or set container without any additional properties then prefer to use `absl::flat_hash_map` and `absl::flat_hash_set`. These are versatile containers that have good performance on both large and small sized data.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
kjyoun, line 42: This file has associated unit test.

Can you add unit test for the changes that you added?

I bet Gemini would write nice tests for you.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -83,17 +83,25 @@ void CobaltLifecycleManager::InitializeTracker(
...
 void CobaltLifecycleManager::OnMojoDisconnect() {
   CHECK_CALLED_ON_VALID_THREAD(thread_checker_);
-  auto [frame, web_contents] = GetCurrentContext();
-  mojo::ReceiverId id = receivers_.current_receiver();
   FrameContext context = receivers_.current_context();
-  if (web_contents) {
-    auto active_it = active_receivers_.find(context.frame_id);
-    if (active_it != active_receivers_.end() && active_it->second == id) {
+
+  auto active_it = active_receiver_counts_.find(context.frame_id);
```
kjyoun: nit:  put `active_it` inside if clause, since it not used outside?

```
if (auto active_it = active_receiver_counts_.find(context.frame_id);
    active_it == active_receiver_counts_.end()) {
...

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc (hunk 4 above)
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc]: Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc (hunk 5 above)
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc]: Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc (hunk 6 above)
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc]: Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc (hunk 7 above)
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc]: Good catch! Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.h (hunk 8 above)
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.h]: Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc:42]: Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc (hunk 9 above)
Comment (jellefoks) [reply cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc]: Done.

File: cobalt/browser/lifecycle/cobalt_lifecycle_manager.h [hunk 10]
```diff
@@ -21,7 +21,6 @@
 #include <vector>
 
 #include "base/containers/flat_set.h"
```
kjyoun: remove?

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -460,15 +464,21 @@ void CobaltLifecycleManager::StartWaitingForAck(
...
-      if (connected) {
-        pending_ack_frames_[web_contents].insert(frame);
-      } else if (ack_type == PendingAck::kUnfreeze) {
-        LOG(WARNING) << "StartWaitingForAck: Frame not connected during "
-                        "Unfreeze! Re-binding frame="
-                     << frame;
-        tracker->Rebind(frame);
+    // For downward transitions, only track the active Primary Main Frame.
+    // This excludes subframes (which can be aggressively throttled or paused
+    // by Blink's scheduler) and older, navigated-away main frames that are
```
kjyoun: This method is already 50 lines long and we are adding more lines. 
I think this block is long/separate enough to be a stand-alone method

WDYT ?

```

 if (ack_type == PendingAck::kReveal) {
    ..
 } else {
   TrackPrimaryMainFrameForAck(..);
 }


void CobaltLifecycleManager::TrackPrimaryMainFrameForAck(
    WebContents* web_contents, 
    Tracker* tracker, 
    PendingAck ack_type) {
    
  // For downward transitions, only track the active Primary Main Frame.
  // This excludes subframes (which can be aggressively throttled or paused
  // by Blink's scheduler) and older, navigated-away main frames that are
  // still leaking in memory pending deletion.
  auto* primary_main_frame = web_contents->GetPrimaryMainFrame();
  if (!primary_main_frame) {
    return;
  }

  if (tracker->IsConnected(primary_main_frame)) {
    pending_ack_frames_[web_contents].insert(primary_main_frame);
  } else if (ack_type == PendingAck::kUnfreeze && 
             primary_main_frame->IsRenderFrameLive()) {
    LOG(WARNING)
        << "StartWaitingForAck: Primary Main Frame not connected during "
           "Unfreeze! Re-binding frame="
        << primary_main_frame;
    tracker->Rebind(primary_main_frame);
  }
}
```

kjyoun (approved):

cobalt/browser/lifecycle/cobalt_lifecycle_manager.h
jellefoks: Done. Removed unused headers.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
jellefoks: Done. Extracted TrackPrimaryMainFrameForAck helper method.

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
```diff
@@ -83,17 +83,25 @@ void CobaltLifecycleManager::InitializeTracker(
...
-  mojo::ReceiverId id = receivers_.current_receiver();
   FrameContext context = receivers_.current_context();
-  if (web_contents) {
-    auto active_it = active_receivers_.find(context.frame_id);
-    if (active_it != active_receivers_.end() && active_it->second == id) {
+
+  if (auto active_it = active_receiver_counts_.find(context.frame_id);
+      active_it != active_receiver_counts_.end()) {
+    active_it->second--;
+    if (active_it->second <= 0) {
```
andrewsavage1, line 91: Do we want a DCHECK or similar if `active_it->second < 0`?

andrewsavage1 (approved):

kjyoun:
super nit: can you add prefix in the commit title: per https://screenshot.googleplex.com/Ay3gTzWG7mwVjxB

e.g.) starboard: Fix race...

cobalt/browser/lifecycle/cobalt_lifecycle_manager.cc
jellefoks: Sure, yes, it would be am introduced bug if that ever happens, since (right now) it's only created with a postincrement, and only decremented here, and it's erased inside this block.

jellefoks:
> super nit: can you add prefix in the commit title: per https://screenshot.googleplex.com/Ay3gTzWG7mwVjxB
> 
> e.g.) starboard: Fix race...

Ok, prefixed with 'fix:', since it's not starboard, nor any of the others.

PR #11045, opened 2026-06-24
Add safety checks to prevent null dereferences in cc::TileManager

tasunnn:
The PR adds null checks in TileManager::ReduceTileMemoryWhenIdle() 
and TileManager::TrimPrepaintTiles() to prevent dereferencing null 
pointers when reclaiming tiles.

This is to address native crash at cc::TileManager::ReduceTileMemoryWhenIdle()
from seg fault when resource_pool_ is dereferenced after destroyed. 

Bug: 527482314


cc/tiles/tile_manager.cc
```diff
@@ -475,6 +476,11 @@ void TileManager::ScheduleTrimPrepaintTiles() {
 void TileManager::ReduceTileMemoryWhenIdle() {
   has_pending_idle_task_ = false;
 
+#if BUILDFLAG(IS_COBALT)
+  if (!resource_pool_)
```
yell0wd0g: My instinct is that this could happen also upstream? But
perhaps only when the gpu is in-process? If so then
this makes sense as short term fix but we should try to
upstream and do something like
``` 
  if (is_gpu_in_ _process && !resource_pool) {
    return;
  }
```
or some such.

Leaving a TODO and filing a bug is ok

yell0wd0g (approved):

cc/tiles/tile_manager.cc
tasunnn: chromium doesn't seem to report any related crash but this does look like a bug in upstream, yeah I will leave a todo here for long term solution, thanks!

PR #11064, opened 2026-06-25
android: Add android:authorities to androidx.startup.InitializationProvider

johnxwork:
PR #11059 stripped androidx.startup.InitializationProvider from the manifest using `tools:node="remove"`. However, according to the Android Manifest Merger specification, elements of type `<provider>` are matched by their authority (`android:authorities`). Without matching `android:authorities="\${applicationId}.androidx-startup"`, the merger may fail to match and remove the provider declared by library dependencies or report validation errors.

Follow-up to PR #11059.

Bug: 495203133

TAG=agy
CONV=77dc1de4-dd7c-479a-9937-22c955c12d6a


hferreiro (approved):
Thanks!

Awallky (approved):

Awallky (commented):
Can we remove the TAG and CONV footers?

PR #11066, opened 2026-06-25
media: Route HLS URLs to URL-player renderer

abhijeetk:
Register kUrlPlayer as a conditional renderer factory using the
IsHlsUrl() predicate so HLS URLs select the URL-player path while
ordinary URLs continue using the Starboard renderer.

Enable CobaltContentRendererClient::OverrideDemuxerForUrl() to create
UrlPlayerDemuxer for HLS URLs.

Add RendererFactorySelectorTest coverage for .m3u8, hls_variant, and
non-HLS URLs.

Build: ninja -C out/tvos-arm64-simulator_qa cobalt nplb media_unittests

Bug: 512045535


media/mojo/clients/mojo_renderer.cc
```diff
@@ -151,6 +151,32 @@ void MojoRenderer::Initialize(MediaResource* media_resource,
                                               base::Unretained(this), client));
 }
 
+#if BUILDFLAG(USE_STARBOARD_URL_PLAYER)
```
borongc, line 154: This breaks Chromium's mojo Renderer design, any non-standard ways should go to either `mojom::StarboardRendererExtension` or `mojom::StarboardRendererClientExtension`.

Modifying `mojo_renderer` and such codes are not upstreamable. Although we have some features (kCobaltBypassMojoForMedia) that modifying mojo_renderer, but this is only for experimental purposes.

As I'm not the owner from media side for URL player, I prefer to wait for @jasonzhangxx to back, so we can have a discussion about how to put URL player in the current chromium pipeline.

For me, there are several things I concern here:
1. global url player build flag doesn't make sense to me, we should try to avoid adding new ones (even this makes tvOS url player life easier)
2. we cannot modify chromium mojo renderer architecture design, see https://source.chromium.org/chromium/chromium/src/+/main:media/mojo/README.md for more information.

I think the current implement is too big for us to maintain, even this is maintained by your side, this makes Starboard codes harder to be upstreamed.

abhijeetk: #### Mojo renderer architecture
Thanks for raising this. I agree with the concern about adding URL-player-specific
API to the generic Mojo renderer path (`mojo_renderer.cc`, `mojo_renderer_service.cc`,
and `renderer.mojom`).

`InitializeWithUrl()` was added to keep URL delivery and renderer initialization
in a single call, but that does modify Chromium's generic Mojo renderer surface.
We can avoid that.

I can to move URL delivery to the Starboard-specific extension pipe, using
`StarboardRendererExtension::SetSourceUrl()` as in the earlier approach from
#10698. The renderer-side client already has access to the URL through
`MediaResource::GetMediaUrl()`, so it can send the URL over the Starboard
extension before calling the normal `MojoRendererWrapper::Initialize()` path,
similar to the existing `OnGpuChannelTokenReady()` ordering pattern.

This change would remove the `InitializeWithUrl()` plumbing from the generic
Mojo renderer classes/interfaces while keeping the current URL-player renderer
class split intact for now. In other words, this addresses the Mojo architecture
concern without folding the URL-player renderer/client/wrapper classes back into
`StarboardRenderer` in this PR.

I agree the broader class structure can still be discussed separately with
@jasonzhangxx , but this change should keep URL transport out of Chromium's generic Mojo
renderer API.

abhijeetk:
Closing this PR.

PR #11067, opened 2026-06-25
media: Add URL-player demuxer and HLS URL predicate

abhijeetk:
Add UrlPlayerDemuxer for tvOS URL-based playback. The demuxer
provides placeholder audio/video DemuxerStreams required by PipelineImpl
while the platform handles actual demuxing, and exposes the source URL
through GetMediaUrl().

Add IsHlsUrl() for detecting HLS URLs, plus a stub
OverrideDemuxerForUrl() in CobaltContentRendererClient. The override is
left inactive until the renderer routing CL lands.

Add kUrlPlayerDemuxer to DemuxerType, including the Mojom enum, traits
mapping, and histogram entry.

Build: ninja -C out/tvos-arm64-simulator_qa cobalt nplb media_unittests

Bug: 512045535


cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -41,7 +43,6 @@
 #include "third_party/blink/public/web/web_security_policy.h"
 #include "third_party/blink/public/web/web_view.h"
 #include "ui/gfx/geometry/size_conversions.h"
-
```
rakuco, line 44: small nit: this looks stylistically wrong

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -19,6 +19,7 @@
 
 #include "base/task/bind_post_task.h"
 #include "base/time/time.h"
+#include "build/build_config.h"
```
rakuco, line 22: This doesn't look necessary.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -30,6 +31,7 @@
 #include "media/base/key_systems_support_registration.h"
 #include "media/base/media_log.h"
 #include "media/base/renderer_factory.h"
+#include "media/media_buildflags.h"
```
rakuco, line 34: Ditto.

media/starboard/url_player_demuxer.cc
```diff
@@ -0,0 +1,158 @@
...
+#include "media/base/media_track.h"
+#include "media/base/sample_format.h"
+#include "media/base/video_decoder_config.h"
+
+namespace media {
+
+bool IsHlsUrl(const GURL& url) {
+  // See
+  // https://github.com/youtube/cobalt/blob/25.lts.stable/cobalt/dom/html_media_element.cc#L103
+  auto path = url.path_piece();
```
rakuco, line 35: This can be `const`.

media/starboard/url_player_demuxer.cc
```diff
@@ -0,0 +1,158 @@
...
+  return path.ends_with(".m3u8") ||
+         path.find("hls_variant") != std::string_view::npos;
```
rakuco, line 37: Just double checking: in C25 the entire URL was checked for the presence of "hls_variant", but this code only checks the path part. Is this intentional? Could you add a comment with an example of what an expected YT URL looks like?

media/starboard/url_player_demuxer.h
```diff
@@ -0,0 +1,102 @@
...
+
+// Returns true if the URL looks like an HLS stream that the platform URL
+// player (e.g., AVPlayer on tvOS) should handle. This centralizes the
+// detection so that WebMediaPlayerImpl and DemuxerManager use the same logic.
+MEDIA_EXPORT bool IsHlsUrl(const GURL& url);
+
+// Placeholder stream used only to satisfy Chromium's stream-based pipeline
+// initialization. The URL player path ignores these streams and delegates
+// loading, demuxing, buffering, and decoding to the platform URL player.
+class MEDIA_EXPORT UrlPlayerDemuxerStream : public DemuxerStream {
```
borongc, line 41: So Cobalt modified `DemuxerStream` already to carry MIME type, it seems like the job for `UrlPlayerDemuxerStream` is to carry URL to SbPlayer.

Is there any reason that instead of making a new `UrlPlayerDemuxerStream`, just add a URL to `DemuxerStream` and gate it by tvOS flag?

abhijeetk: MIME type is per-stream metadata: each track has its own content type. The
source URL is per-resource metadata: one URL produces both audio and video
tracks. Putting it on `DemuxerStream` would duplicate it across both placeholder
streams.

More importantly, adding URL to `DemuxerStream` does not eliminate the new class.
`DemuxerStream` is an abstract interface requiring concrete implementations of
`Read()`, `audio_decoder_config()`, `video_decoder_config()`, `type()`, and
`SupportsConfigChanges()`. The URL-player path needs concrete placeholder
streams that provide dummy audio/video configs and stream types for pipeline
initialization, while making `Read()` unreachable because AVPlayer owns fetching,
demuxing, buffering, and decoding. We do not want Chromium to download the media
in this path; AVPlayer should own that.

That placeholder stream behavior has to live in a concrete class somewhere, and
existing stream classes (`FFmpegDemuxerStream`, etc.) are wired to real data
sources. Even if the URL were added to the base `DemuxerStream`, we would still
need the dummy stream class. We would end up with both: URL-player-specific
metadata on `DemuxerStream`, and a concrete placeholder stream implementation
that still cannot be avoided.

`UrlPlayerDemuxerStream` is the concrete placeholder stream implementation that
the pipeline requires. It is not a URL container; the URL lives on
`UrlPlayerDemuxer` (which is a `MediaResource`) and is exposed through
`MediaResource::GetMediaUrl()`.
  
We implemented similar apporach in our initial patch : https://github.com/youtube/cobalt/pull/10698
but discarded it for separate media pipeline for URLPlayer.

abhijeetk:
Closing this PR.

PR #11090, opened 2026-06-26
[Profiling] Add helper script to verify Cobalt memory trace JSON outputs

haozheng-cobalt:
Add verify_memory_trace.py under cobalt/tools/performance/memory/ to inspect and verify trace JSON files captured from Android TV devices.

The script checks:
1. File existence, size, and JSON syntax validity.
2. The presence of 'periodic_interval' trace memory snapshots.
3. The presence of 'heaps_v2' block containing heap profiler allocations.
4. The presence of 'process_mmaps' containing loaded C++ library offset maps.

Bug: 528369392

### DIFF
--- a/cobalt/tools/performance/memory/verify_memory_trace.py
+++ b/cobalt/tools/performance/memory/verify_memory_trace.py
@@ -6,61 +6,99 @@
 
 import argparse
 import json
+import logging
 import os
 import sys
+from typing import Any, Dict, List, Optional
+
+# Configure logging
+logging.basicConfig(level=logging.INFO, format="%(message)s")
+logger = logging.getLogger(__name__)
 
 
-def verify_trace(trace_path):
-  print("=" * 60)
-  print(f"🔍 VERIFYING COBALT MEMORY TRACE: {trace_path}")
-  print("=" * 60)
+def parse_trace_file(trace_path: str) -> Optional[List[Dict[str, Any]]]:
+  """Parses trace file and extracts list of trace events.
 
+  Args:
+    trace_path: Path to the JSON trace file.
+
+  Returns:
+    List of event dictionaries, or None if validation fails.
+  """
   # 1. Check file existence & size
-  if not os.path.exists(trace_path):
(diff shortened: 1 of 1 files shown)


cobalt/tools/performance/memory/verify_memory_trace.py
```diff
@@ -0,0 +1,159 @@
...
+    print("❌ Error: No trace events found inside the file!")
+    return False
+
+  # 3. Analyze memory dump events
+  memory_dumps_count = 0
+  has_heaps_v2 = False
+  has_process_mmaps = False
+  has_smaps = False
+
+  for event in events:
```
Awallky, line 93: I noticed that we consider only dict objects for containing trace events. However, we also treat a list as a possible set of trace events in 

```
if isinstance(trace_data, list):
    events = trace_data
```

This does not seem compatible with the current loop. Do the lists contain dict objects? 
If so, it might be helpful to break this up into helper functions with type annotations to help understand the expected data types across this script.

Awallky (approved):
LGTM % comment

cobalt/tools/performance/memory/verify_memory_trace.py
```diff
@@ -0,0 +1,159 @@
...
+"""Verifies if a JSON trace is captured and contains memory dumps."""
+
+import argparse
+import json
+import os
+import sys
+
+
+def verify_trace(trace_path):
+  print("=" * 60)
```
Awallky: suggest using log instead of print function

cobalt/tools/performance/memory/verify_memory_trace.py
haozheng-cobalt: Thanks for catching this, addressed it by splitting up the cases.

cobalt/tools/performance/memory/verify_memory_trace.py
haozheng-cobalt: Done.

PR #11093, opened 2026-06-26
android: Extract BrowserStarboardBridge from StarboardBridge

hferreiro:
This is a precursor PR to ease the review of the AOSP work with no functional changes.

Move the content-dependent behavior out of StarboardBridge into a new BrowserStarboardBridge subclass used by AndroidTV.

StarboardBridge keeps the full JNI-facing surface with no-op defaults for the content hooks that BrowserStarboardBridge overrides so it can be shared with the AOSP apk.

Bug: 495203133


andrewsavage1 (approved):

haozheng-cobalt (approved):
LGTM, thanks for @rogerzanoni to explain the change while you are OOO. 
Could you address all the comments before I submit this PR?

hferreiro:
@haozheng-cobalt please, take a look at the latest commit. It reworks things a bit so that we can avoid the cast.

PR #11094, opened 2026-06-26
BACKPORT: WebContentsViewIOS: Remove native view from hierarchy on destruction

rakuco:
This fixes a regression introduced in #9959. The code in the pull request is correct, but it unmasked an underlying bug: when the splash screen WebContents was destroyed, its corresponding UIView remained alive and part of the view hierarchy. Since the WebContents' RenderWidgetHostView's UIView was correctly removed from the hierarchy, we ended up with an invisible UIView on top of the main WebContents' one that prevented all gesture recognizers from working (i.e. swipe and pan gesture handling did not work at all).

Gesture recognition is still imprecise compared to C25, but this change at least makes it work again.

Bug: 521473039


rakuco:

fayev-g (approved):

PR #11103, opened 2026-06-26
Android: report GPU usage from Debug.getMemoryStat via UMA

tasunnn:
android: Use Debug.getMemoryStat() to report cobalt GPU usage

This change adds the summary.graphics memory statistic from Android's
Debug.MemoryInfo to the memory metrics reported via UMA. By
retrieving this value through the MemoryInfoBridge, Cobalt can now
track graphics-related memory consumption on Android devices.

This metrics will fire every five minute to prevent performing heavy
operation too frequently on Android OS.

Bug: 527051163


base/android/java/src/org/chromium/base/memory/MemoryInfoBridge.java
```diff
@@ -42,4 +42,19 @@ public class MemoryInfoBridge {
             return null;
         }
     }
+
+    @CalledByNative
+    public static int getGraphicsMemoryKb(Debug.MemoryInfo info) {
```
Awallky: This is an upstream Chromium file. 
Can we add this functionality in a separate CobaltMemoryBridge class and only reference this in cobalt builds?

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
```diff
@@ -291,6 +291,12 @@ const CobaltMemoryMetricsEmitter::Metric kAllocatorDumpNamesForMetrics[] = {
      base::android::MeminfoDumpProvider::kPssMetricName,
      CobaltMemoryMetricsEmitter::EmitTo::kSizeInUmaOnly,
      {}},
+    {base::android::MeminfoDumpProvider::kDumpName,
+     "AndroidGraphicsMemory",
```
Awallky, line 295: We will want to register this in our histograms.xml file as well so this is picked up in our dashboards.

base/android/java/src/org/chromium/base/memory/MemoryInfoBridge.java
```diff
@@ -42,4 +42,19 @@ public class MemoryInfoBridge {
...
+        // if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
+        String stat = info.getMemoryStat("summary.graphics");
```
johnxwork: Another worry is if this field is unavailable on some devices(due to firmware differences or Android version differences). Would be good to add a test that reports missing this field so that automated tests on various devices during pre-release testing will tell us if this is e.g. only available on Amlogic devices. 

I'm not sure what's the best test type for this though. If it's too complicated feel free to skip for now and let's see what data we get from the field first.

cobalt/browser/metrics/cobalt_memory_metrics_emitter.cc
tasunnn: done, thanks!

base/android/java/src/org/chromium/base/memory/MemoryInfoBridge.java
tasunnn: this API should be available for Android 6+ devices. Do you mean that we should add a test in cobalt to report the specific devices that don't support this API? 

if we just want to know how many devices that don't support, we could also just add another UMA to log it here.

Awallky (approved):
LGTM.
Could we add tests to ensure this fires as expected?

base/android/java/src/org/chromium/base/memory/CobaltMemoryInfoBridge.java
```diff
@@ -0,0 +1,29 @@
...
+            try {
+                String graphicsStr = memoryInfo.getMemoryStat("summary.graphics");
+                if (graphicsStr != null) {
+                    return Integer.parseInt(graphicsStr);
+                }
+            } catch (NumberFormatException e) {
+                // Ignore.
+            }
+        }
+        return 0;
```
johnxwork, line 27: Would some devices returning zero reduce the global average of this UMA? Is there a way to not report in case of unsupported devices?

johnxwork (approved):

base/android/java/src/org/chromium/base/memory/CobaltMemoryInfoBridge.java
tasunnn: done, thanks

PR #11104, opened 2026-06-27
fix(starboard): Fallback to IPv4 if IPv6 binding fails

anshugupta27:
fix(starboard): implement runtime check for IPv6 socket support" -m "Updates SbSocketIsIpv6Supported() in Starboard to dynamically probe the OS for IPv6 support via a temporary socket bind. If the probe fails, it returns false, letting the browser fall back to IPv4.

Bug: 528472809

### DIFF
--- a/starboard/shared/posix/socket_is_ipv6_supported.cc
+++ b/starboard/shared/posix/socket_is_ipv6_supported.cc
@@ -26,6 +26,7 @@
 #include "starboard/shared/posix/handle_eintr.h"
 
 namespace {
+#if !SB_IS(COMPILER_MSVC) && !defined(_WIN32)
 bool ProbeIpv6Support() {
   const int probe_socket_fd = socket(AF_INET6, SOCK_STREAM, 0);
   if (probe_socket_fd < 0) {
@@ -56,13 +57,18 @@ bool ProbeIpv6Support() {
 
   return (bind_status == 0);
 }
+#endif
 }  // namespace
 
 #endif  // SB_HAS_IPV6
 
 bool SbSocketIsIpv6Supported() {
 #if SB_HAS_IPV6
+#if SB_IS(COMPILER_MSVC) || defined(_WIN32)
+  return true;
+#else
   return ProbeIpv6Support();
+#endif
 #else
   return false;
 #endif
--- a/starboard/shared/posix/socket_is_ipv6_supported.cc
+++ b/starboard/shared/posix/socket_is_ipv6_supported.cc
@@ -20,6 +20,7 @@
 #include <string.h>
 #include <sys/socket.h>
 #include <unistd.h>
+#include <errno.h>
 
 #include "starboard/common/log.h"
 #include "starboard/shared/posix/handle_eintr.h"
--- a/starboard/shared/posix/socket_is_ipv6_supported.cc
+++ b/starboard/shared/posix/socket_is_ipv6_supported.cc
@@ -16,14 +16,16 @@
 
 #if SB_HAS_IPV6
 
-#include <netinet/in.h>
 #include <string.h>
+#include "starboard/common/log.h"
+
+#if !SB_IS(COMPILER_MSVC) && !defined(_WIN32)
+#include <errno.h>
+#include <netinet/in.h>
 #include <sys/socket.h>
 #include <unistd.h>
-#include <errno.h>
-
-#include "starboard/common/log.h"
 #include "starboard/shared/posix/handle_eintr.h"
+#endif
 
 namespace {
 #if !SB_IS(COMPILER_MSVC) && !defined(_WIN32)
--- a/starboard/shared/posix/socket_is_ipv6_supported.cc
+++ b/starboard/shared/posix/socket_is_ipv6_supported.cc
@@ -14,6 +14,55 @@
 
 #include "starboard/socket.h"
 
+#if SB_HAS_IPV6
+
+#include <netinet/in.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#include "starboard/common/log.h"
+#include "starboard/shared/posix/handle_eintr.h"
+
+namespace {
+bool ProbeIpv6Support() {
+  const int probe_socket_fd = socket(AF_INET6, SOCK_STREAM, 0);
+  if (probe_socket_fd < 0) {
+    SB_LOG(WARNING) << "IPv6 Probe: Failed to create socket. error=" << errno;
+    return false;
+  }
+
+  int reuse_addr_enabled = 1;
+  if (setsockopt(probe_socket_fd, SOL_SOCKET, SO_REUSEADDR, &reuse_addr_enabled, sizeof(reuse_addr_enabled)) != 0) {
+    SB_LOG(WARNING) << "IPv6 Probe: Failed to set SO_REUSEADDR. error=" << errno;
+  }
+
+  struct sockaddr_in6 probe_address;
+  memset(&probe_address, 0, sizeof(probe_address));
+  probe_address.sin6_family = AF_INET6;
+  probe_address.sin6_addr = in6addr_any;
+  probe_address.sin6_port = 0;
+
+  const int bind_status = HANDLE_EINTR(bind(
+      probe_socket_fd, reinterpret_cast<struct sockaddr*>(&probe_address),
+      sizeof(probe_address)));
+  
+  if (bind_status != 0) {
+    SB_LOG(WARNING) << "IPv6 Probe: Failed to bind test socket. error=" << errno;
+  }
+
+  close(probe_socket_fd);
+
+  return (bind_status == 0);
+}
+}  // namespace
+
+#endif  // SB_HAS_IPV6
+
 bool SbSocketIsIpv6Supported() {
-  return SB_HAS_IPV6;
+#if SB_HAS_IPV6
+  return ProbeIpv6Support();
+#else
+  return false;
+#endif
 }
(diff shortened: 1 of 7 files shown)


cobalt/debug/remote/debug_web_server.cc
```diff
@@ -92,8 +92,17 @@ constexpr net::NetworkTrafficAnnotationTag kNetworkTrafficAnnotation =
                                         "cobalt_debug_web_server");
 
 constexpr int kUnattachedWebSocketId = -1;
+
+// IPV4 equivalent of an IPV6 wildcard/loopback address, or "" if there's no
+// well-known twin. Mirrors application.cc::GetDevServersListenIp()'s pairs.
+std::string Ipv4EquivalentForIp6Address(const std::string& ipv6_address){
+  if (ipv6_address == "::") return "0.0.0.0"; // IPV6 any -> IPV4 any
```
jellefoks: There are many ways to write the ipv6 any address, or the ipv6 loopback; e.g. you don't have to collapse the '0' values. The only way to do this reliably from any string is to resolve the name to an address, but actually in this case it appears that the string has to be an ip address literal and can't be a hostname, so you could probably use `net::IPAddress`, followed by comparing the IP address number to the "any" or "loopback" address.

But also, it may be surprising to a user of `DebugWebServer` that it sometimes binds to an IP address that is different than the one specified.

So I looked where `DebugWebServer` is used, and that's in `cobalt/browser/application.cc`, and it uses `GetDevServersListenIp` to determine the IP address to listen on, which has a `bool ip_v6;` with value `ip_v6 = SbSocketIsIpv6Supported();`. 

And looking for its definitions (`SbSocketIsIpv6Supported`), that lives in only two places in our code,  `starboard/shared/posix/socket_is_ipv6_supported.cc` and `starboard/shared/stub/socket_is_ipv6_supported.cc`. The stub returns false, and the other one returns `SB_HAS_IPV6`.

That implementation of `SbSocketIsIpv6Supported` looks like the root cause to me.

Now, `starboard/shared` lives below the Starboard platform abstraction, so we can't use `net::`, but if we can improve the implementation of `SbSocketIsIpv6Supported` to detect this ipv6 non-support case at run time correctly, that will automatically cause this to be called with a ipv4 any address or loopback address.

A server socket has to be created, `bind`, and `listen`. I wonder if the detection can happen at the `bind`, before the `listen`, in which case it would be harmless to try to bind to the ipv6 any address. Otherwise, if it has to attempt to listen before detecting this, it may be wiser to listen only on the loopback address, so that there is no external effect of the test.

Now, I recognize that is also tedious, but it seems more correct to do than modifying this class specifically to sometimes bind elsewhere.

anshugupta27: Thanks for the detailed review, this makes sense to me. You're right that the real problem is in SbSocketIsIpv6Supported() - it just returns the compile time SB_HAS_IPV6, so it always says IPv6 is supported even when the OS has it turned off.
I'll move the fix there instead. I'll make it check at runtime by trying to bind to the IPv6 any-address (only bind, no listen, on an ephemeral port, and I'll cache the result so it only runs once). If that bind fails, it means IPV6 isin't really available, so it returns false.
That way way GetDevServersListenIp will automatically pick the IPV4 address, and I don't have to touch DebugWebServer at all. It should also fix WebDriver the same way since it uses the same function. I'll revert my change in debug_web_server.cc. Thanks again!

starboard/shared/posix/socket_is_ipv6_supported.cc
```diff
@@ -14,6 +14,50 @@
...
+
+    const int bind_result = HANDLE_EINTR(bind(
+      socket_fd, reinterpret_cast<struct sockaddr*>(&address),
+      sizeof(address)
+    ));
+    g_ipv6_supported = (bind_result == 0);
+
+    close(socket_fd);
+  }
+} // namespace
```
jellefoks: nit: that should have two spaces between '}' and '//'

starboard/shared/posix/socket_is_ipv6_supported.cc
```diff
@@ -14,6 +14,50 @@
...
+      return;
+    }
+
+    struct sockaddr_in6 address;
+    memset(&address, 0, sizeof(address));
+    address.sin6_family = AF_INET6;
+    address.sin6_addr = in6addr_any;
+    address.sin6_port = 0;
+
+    const int bind_result = HANDLE_EINTR(bind(
```
jellefoks: Maybe also set SO_REUSEADDR true so that this test doesn't lock up a server port for a timeout period after this testing.

starboard/shared/posix/socket_is_ipv6_supported.cc
```diff
@@ -14,6 +14,50 @@
...
+#include <netinet/in.h>
+#include <pthread.h>
+#include <string.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+#include "starboard/shared/posix/handle_eintr.h"
+
+namespace {
+  pthread_once_t g_ipv6_probe_once = PTHREAD_ONCE_INIT;
```
jellefoks: When memoizing this, I worry what happens when the first time that this is called the WiFi is not connected. I think it will then never use ipv6 anymore. Let's not do memoization here, and if there is Cobalt code that needs to use this functionality a lot, and the performance of this test matters, it can do memoization and decide to re-test when the connectivity changes, etc.

starboard/shared/posix/socket_is_ipv6_supported.cc
anshugupta27: Sure, I will add this.

starboard/shared/posix/socket_is_ipv6_supported.cc
anshugupta27: Sure, makes sense I will do this change.

jellefoks (approved):
This is great!

Now, there is a CI failure, `All/CertVerifyProcInternalTest.RejectWeakKeys/CertVerifyProcBuiltin`

I suspect that some test certificates have expired... And I think the acceptable solution for those is to filter out the failing tests.

jellefoks:
> This is great!
> 
> Now, there is a CI failure, `All/CertVerifyProcInternalTest.**RejectWeakKeys**/CertVerifyProcBuiltin`
> 
> I suspect that some test certificates have expired... And I think the acceptable solution for those is to filter out the failing tests.

Since the failures are for all platforms, we'd need to filter it for all platforms. We have `test_filters.py` files that contain filtered test lists, but there is no infrastructure setup for a top level 'all platforms' filter. Best is probably to make such infrastructure, but that means tracking the dependency chain of all 28 test_filters.py files, and ensuring that the CreateTestFilters creates a class that includes the filters from that base.

Or, disable the test in the test file at `net/cert/cert_verify_proc_unittest.cc` (line 879)... Maybe with a `GTEST_SKIP()` inside an `#if defined(USE_COBALT_CUSTOMIZATIONS)`.

That would be a much smaller change, albeit less visible that the test is disabled...

anshugupta27:
> > This is great!
> > Now, there is a CI failure, `All/CertVerifyProcInternalTest.**RejectWeakKeys**/CertVerifyProcBuiltin`
> > I suspect that some test certificates have expired... And I think the acceptable solution for those is to filter out the failing tests.
> 
> Since the failures are for all platforms, we'd need to filter it for all platforms. We have `test_filters.py` files that contain filtered test lists, but there is no infrastructure setup for a top level 'all platforms' filter. Best is probably to make such infrastructure, but that means tracking the dependency chain of all 28 test_filters.py files, and ensuring that the CreateTestFilters creates a class that includes the filters from that base.
> 
> Or, disable the test in the test file at `net/cert/cert_verify_proc_unittest.cc` (line 879)... Maybe with a `GTEST_SKIP()` inside an `#if defined(USE_COBALT_CUSTOMIZATIONS)`.
> 
> That would be a much smaller change, albeit less visible that the test is disabled...

Yes, I did fix this.

anshugupta27:
> > This is great!
> > Now, there is a CI failure, `All/CertVerifyProcInternalTest.**RejectWeakKeys**/CertVerifyProcBuiltin`
> > I suspect that some test certificates have expired... And I think the acceptable solution for those is to filter out the failing tests.
> 
> Since the failures are for all platforms, we'd need to filter it for all platforms. We have `test_filters.py` files that contain filtered test lists, but there is no infrastructure setup for a top level 'all platforms' filter. Best is probably to make such infrastructure, but that means tracking the dependency chain of all 28 test_filters.py files, and ensuring that the CreateTestFilters creates a class that includes the filters from that base.
> 
> Or, disable the test in the test file at `net/cert/cert_verify_proc_unittest.cc` (line 879)... Maybe with a `GTEST_SKIP()` inside an `#if defined(USE_COBALT_CUSTOMIZATIONS)`.
> 
> That would be a much smaller change, albeit less visible that the test is disabled...

Yes, fixed this.

starboard/shared/posix/socket_is_ipv6_supported.cc
anshugupta27: yes, I had acknowledged this.

niranjanyardi:
> fix(starboard): implement runtime check for IPv6 socket support" -m "Updates SbSocketIsIpv6Supported() in Starboard to dynamically probe the OS for IPv6 support via a temporary socket bind. If the probe fails, it returns false, letting the browser fall back to IPv4.
> 
> Bug: 528472809

Adding a small note after talking to Jelle . The current version addresses the issue on all platforms but wont be applicable until partners update their starboard binary. A previous version of this PR made changes in the cobalt code(which would have instantly applied to all partners) and was discarded in favor of the starboard changes which will apply only when partners update their loader_app/ starboard code. i updated the title accordingly

PR #11109, opened 2026-06-28
Temporarily disable tvOS post-submit workflow

linxinan-yt:
tvOS pipeline's post-submit is failing because it relies on kokoro to build the artifacts. While PR can trigger kokoro presubmit job to build the binary artifacts, post-submit cannot. Meanwhile we want tvOS's postsubmit to run device tests, so its service architecture is different from presubmit. 

I am working b/519668013 to rebuild the postsubmit, before it is done, disable the tvOS post-submit workflow.

Bug: 519668013

### DIFF
--- a/.github/workflows/tvos.yaml
+++ b/.github/workflows/tvos.yaml
@@ -7,10 +7,6 @@ on:
       - main
       - experimental/*
       - feature/*
-  push:
-    branches:
-      - experimental/*
-      - feature/*
   workflow_dispatch:
     inputs:
       nightly:
--- a/.github/workflows/tvos.yaml
+++ b/.github/workflows/tvos.yaml
@@ -66,7 +66,6 @@ jobs:
           echo "targets=${targets}" >> "$GITHUB_OUTPUT"
 
   wait-for-kokoro-build:
-    if: github.event_name == 'pull_request'
     needs: get-sha
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/tvos.yaml
```diff
@@ -149,9 +151,75 @@
...
+                tar -czf "${archive_file}" -C "${out_dir}" "${target_name}.app" "iossim"
+              else
+                tar -czf "${archive_file}" -C "${out_dir}" "${target_name}.app"
+              fi
+            fi
+          done
+        shell: bash
+
+      - name: Upload tvOS Test Artifacts to GHA
+        uses: actions/upload-artifact@v4
```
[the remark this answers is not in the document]
linxinan-yt: The change is removed.

.github/workflows/tvos.yaml
```diff
@@ -9,7 +9,6 @@ on:
       - feature/*
   push:
```
oxve: Remove the whole push block if it doesn't work on postsubmit. If you keep the trigger the check will still run and fail on push events on the remaining branch (prefixes).

The check will still run on pull requests according to the trigger rule above.

oxve (approved):

.github/workflows/tvos.yaml
linxinan-yt: Make sense. Thanks for pointing it.

PR #11114, opened 2026-06-29
media: Refactor experimental features into a map

kjyoun:
The current implementation of experimental features in Cobalt relies on
fixed structs. This requires manual updates to multiple layers,
including Mojo interfaces and IPC traits, for every new experiment.

This change replaces the custom struct with a generic map container,
significantly reducing boilerplate. This approach allows developers to
add new Starboard experimental features without modifying Chromium
code. It introduces a typed key system to maintain safety while
providing a flexible pipeline from H5vcc settings to the platform
rendering logic.

Before:

[commit](https://github.com/youtube/cobalt/pull/11098/changes/a85b354c27d3661ed347ff289acb7e4f9146ac0b)

```
// Changes to define flags and plumbing

 cobalt/renderer/cobalt_content_renderer_client.cc        | 6 ++++++
 media/base/ipc/media_param_traits_macros.h               | 1 +
 media/base/starboard/starboard_renderer_config.cc        | 2 ++
 media/base/starboard/starboard_renderer_config.h         | 1 +
 media/starboard/sbplayer_bridge.cc                       | 2 ++
 starboard/extension/experimental/experimental_features.h | 1 +
 starboard/shared/starboard/experimental_features.cc      | 2 ++
 starboard/shared/starboard/experimental_features.h       | 1 +
 8 files changed, 16 insertions(+)
```

After:

[commit](https://github.com/youtube/cobalt/pull/11110/changes/bdb85b4b4a709d5b34350d77f7947ba508cfef03#diff-6098e275e6e35c9e62c6ad61e021a6ebf2a8087b46dfb9082e8f122ed13e3491)

```
// Changes to define flags. No plumbing necessary

 starboard/shared/starboard/experimental_features.h    | 2 ++
 1 files changed, 2 insertions(+)
```

```
# How to declare flag and use it

--- a/starboard/shared/starboard/experimental_features.h
+inline constexpr ExperimentalFeatureKey<bool> kMediaEnableAppProvisioning(
+    "Media.EnableAppProvisioning");

+++ b/starboard/android/shared/player_components_factory.cc
@@ -292,6 +292,10 @@ class PlayerComponentsFactory : public PlayerComponents::Factory {
     const auto& experimental_features =
         creation_parameters.experimental_features();
 
+    if (experimental_features.GetBool(kMediaEnableAppProvisioning)) {
+      MediaCapabilitiesCache::GetInstance()->SetAppProvisioningEnabled(true);
+      SB_LOG(INFO) << "`enable_app_provisioning` is set to true.";
+    }
```

Issue: 480134029


starboard/shared/starboard/experimental_features.h
```diff
@@ -57,6 +145,50 @@ const ExperimentalFeatures& GetExperimentalFeaturesForCurrentThread();
...
+inline constexpr ExperimentalFeatureKey<bool>
+    kMediaIgnoreMediaCodecCallbacksDuringFlushing(
+        "Media.IgnoreMediaCodecCallbacksDuringFlushing");
+inline constexpr ExperimentalFeatureKey<int> kMediaVideoRendererMinInputBuffers(
+    "Media.VideoRendererMinInputBuffers");
+inline constexpr ExperimentalFeatureKey<int>
+    kMediaVideoDecoderInitialPrerollCount(
+        "Media.VideoDecoderInitialPrerollCount");
+inline constexpr ExperimentalFeatureKey<int>
+    kMediaVideoRendererMinDecodedFrames("Media.VideoRendererMinDecodedFrames");
```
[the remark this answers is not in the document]
kjyoun: Done.

media/base/starboard/ipc_param_traits.h
```diff
@@ -0,0 +1,45 @@
...
+};
+
+template <>
+struct ParamTraits<media::ExperimentalFeatures> {
+  static void Write(base::Pickle* m, const media::ExperimentalFeatures& p);
+  static bool Read(const base::Pickle* m,
+                   base::PickleIterator* iter,
+                   media::ExperimentalFeatures* r);
+  static void Log(const media::ExperimentalFeatures& p, std::string* l);
+};
```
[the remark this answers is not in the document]
kjyoun: Done. Added MEDIA_EXPORT to ParamTraits specializations.

media/base/starboard/experimental_features.h
```diff
@@ -0,0 +1,169 @@
...
+  template <typename T>
+  std::optional<T> GetValue(const Value& val) const;
+
+  Map settings_;
+};
+
+template <typename T>
+inline std::optional<T> ExperimentalFeatures::GetValue(const Value& val) const {
+  static_assert(sizeof(T) == 0,
+                "Unsupported type for ExperimentalFeatures::Get");
```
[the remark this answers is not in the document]
kjyoun: Done. Replaced sizeof(T) == 0 with !std::is_same_v<T, T>.

starboard/shared/starboard/experimental_features.h
```diff
@@ -15,38 +15,126 @@
...
+  template <typename T>
+  std::optional<T> GetValue(const Value& val) const;
+
+  Map settings_;
 };
 
+template <typename T>
+inline std::optional<T> ExperimentalFeatures::GetValue(const Value& val) const {
+  static_assert(sizeof(T) == 0,
+                "Unsupported type for ExperimentalFeatures::Get");
```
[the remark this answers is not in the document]
kjyoun: Done. Replaced sizeof(T) == 0 with !std::is_same_v<T, T>.

media/base/starboard/ipc_param_traits.cc
```diff
@@ -0,0 +1,87 @@
...
+#include "media/base/starboard/ipc_param_traits.h"
+
+#include <variant>
+
+#include "base/strings/stringprintf.h"
+#include "ipc/ipc_message_utils.h"
+
+namespace IPC {
+
+using media::ExperimentalFeatures;
```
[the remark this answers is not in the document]
kjyoun: Done. Added static_asserts for variant type indices.

media/starboard/sbplayer_bridge.cc
```diff
@@ -147,18 +149,17 @@ void SetDiscardPadding(
...
+  std::vector<StarboardExperimentalFeatureEntry> entries;
+  entries.reserve(experimental_features.settings().size());
+  for (const auto& [key, val] : experimental_features.settings()) {
+    if (auto* int_val = std::get_if<int64_t>(&val)) {
+      entries.push_back({key.c_str(), *int_val});
+    }
   }
-  return &*val;
+  return entries;
 }
```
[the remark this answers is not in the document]
kjyoun: Since this is an internal static helper inside an anonymous namespace called in only one place with a persistent member variable (`experimental_features_`), deleting the rvalue overload is unnecessary.

starboard/shared/starboard/experimental_features.h
```diff
@@ -57,6 +145,51 @@ const ExperimentalFeatures& GetExperimentalFeaturesForCurrentThread();
...
+    kMediaIgnoreMediaCodecCallbacksDuringFlushing(
+        "Media.IgnoreMediaCodecCallbacksDuringFlushing");
+inline constexpr ExperimentalFeatureKey<int> kMediaVideoRendererMinInputBuffers(
+    "Media.VideoRendererMinInputBuffers");
+inline constexpr ExperimentalFeatureKey<int>
+    kMediaVideoDecoderInitialPrerollCount(
+        "Media.VideoDecoderInitialPrerollCount");
+inline constexpr ExperimentalFeatureKey<int>
+    kMediaVideoRendererMinDecodedFrames("Media.VideoRendererMinDecodedFrames");
+// keep-sorted end
```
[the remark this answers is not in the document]
kjyoun: Done. Updated the keep-sorted directive with `by_regex=k\w+ newline_separated=yes` to enforce sorted order for wrapped declarations.

starboard/shared/starboard/experimental_features.h
```diff
@@ -15,38 +15,126 @@
 #ifndef STARBOARD_SHARED_STARBOARD_EXPERIMENTAL_FEATURES_H_
 #define STARBOARD_SHARED_STARBOARD_EXPERIMENTAL_FEATURES_H_
 
+#include <cerrno>
+#include <cstdint>
+#include <cstdlib>
```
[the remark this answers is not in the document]
kjyoun: Done. Removed unused headers (`<cerrno>` and `<cstdlib>`).

media/base/starboard/ipc_param_traits.cc
```diff
@@ -0,0 +1,97 @@
...
+    return true;
+  }
+  return false;
+}
+
+void ParamTraits<ExperimentalFeatures::Value>::Log(
+    const ExperimentalFeatures::Value& p,
+    std::string* l) {
+  l->append(base::StringPrintf("<ExperimentalFeatures::Value>"));
+}
```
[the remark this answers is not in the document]
kjyoun: Done. Replaced `base::StringPrintf` with direct string append.

media/base/starboard/ipc_param_traits.cc
```diff
@@ -0,0 +1,97 @@
...
+    return false;
+  }
+  *r = ExperimentalFeatures(std::move(settings));
+  return true;
+}
+
+void ParamTraits<ExperimentalFeatures>::Log(const ExperimentalFeatures& p,
+                                            std::string* l) {
+  l->append(base::StringPrintf("<ExperimentalFeatures>"));
+}
```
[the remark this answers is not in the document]
kjyoun: Done. Replaced `base::StringPrintf` with direct string append.

media/base/ipc/media_param_traits_macros.h
```diff
@@ -193,30 +194,7 @@ IPC_STRUCT_TRAITS_BEGIN(media::OverlayInfo)
   IPC_STRUCT_TRAITS_MEMBER(is_fullscreen)
   IPC_STRUCT_TRAITS_MEMBER(is_persistent_video)
 IPC_STRUCT_TRAITS_END()
-
```
borongc, line 196: Is this intended?

media/base/ipc/media_param_traits_macros.h
```diff
@@ -226,5 +204,4 @@ IPC_STRUCT_TRAITS_BEGIN(media::StarboardRendererConfig)
   IPC_STRUCT_TRAITS_MEMBER(viewport_size)
 IPC_STRUCT_TRAITS_END()
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
-
```
borongc, line 233: Same here.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -45,52 +46,26 @@
 namespace cobalt {
 
 namespace {
+using ::media::ExperimentalFeatures;
 
 const char kWidevineL3KeySystem[] = "com.youtube.widevine.l3";
 
-const char kH5vccSettingsKeyMediaAllowAudioWritingOnPause[] =
```
borongc, line 51: If we refactor this, there are lots of h5vcc flags in g3 that has a reference link to the definition here.

media/base/ipc/media_param_traits_macros.h
kjyoun: thanks for the catch. Done

media/base/ipc/media_param_traits_macros.h
kjyoun: Done

cobalt/renderer/cobalt_content_renderer_client.cc
kjyoun: You mean this links, right?

https://source.corp.google.com/search?q=%22cobalt_content_renderer_client.cc%22&sq=package:piper%20file:%2F%2Fdepot%2Fgoogle3%20-file:google3%2Fexperimental

Since the link has commit hash, the link itself is not broken with this change and developer can follow the changes where the new flags are located. For the cross-repository, that is the best practice to enforce.

That being said, updating those links is trivial. I will update them.

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -495,8 +323,7 @@ void CobaltContentRendererClient::GetStarboardRendererFactoryTraits(
   cobalt::mojom::SettingsPtr settings;
   ExperimentalFeatures experimental_features;
   if ((*h5vcc_settings_remote_)->GetSettings(&settings) && settings) {
-    auto h5vcc_settings = ParseH5vccSettings(std::move(settings));
-    experimental_features = ProcessH5vccSettings(h5vcc_settings);
+    experimental_features = ParseH5vccSettings(std::move(settings));
```
borongc, line 326: Is there any way we can ensure this refactor PR doesn't break the current code? We have many active experiments running, though this PR looks safe, just want to check better options to catch regression besides manually review it.

kjyoun: Agree that it is important to make sure that this change has no regreession

I ran manual test for all 21 h5vcc flags we have now by starting Cobalt app with each flag and checked logs to verify the corresponding feature is activated. (I can inject h5vcc setting using `--enable-h5vcc-settings` flag)

johnedocampo: We can also do additional local testing and simulate what a device might go through when the H5VCC setting is received at runtime.

kjyoun: @johnedocampo can you elaborate what "additional local testing" mean?

Do you mean testing runtime setting updates via custom JS loader (vs using `enable-h5vcc-settings`) ?

I can do it, though technically there is no difference for codes touched in this PR.

Both paths populate  `GlobalFeatures`  in the browser process and are fetched by the renderer via the same Mojo IPC interface (`h5vcc_settings_remote_->GetSettings() `) during renderer creation.

kjyoun: Tested with custom loader setting h5vcc settings

verified that all h5vcc settings defined in js side are passed to Cobalt and activated properly

media/base/starboard/experimental_features.h
```diff
@@ -0,0 +1,173 @@
...
+#include <type_traits>
+#include <variant>
+
+#include "media/base/media_export.h"
+
+namespace media {
+namespace internal {
+
+// Experiment framework uses 0 as the sentinel value for unset.
+// e.g.)
```
borongc: nit: e.g.,

kjyoun: Done

starboard/shared/starboard/player/filter/testing/BUILD.gn
```diff
@@ -25,6 +25,7 @@ if (current_toolchain == starboard_toolchain) {
     testonly = true
 
     sources = [
+      "//starboard/shared/starboard/experimental_features_test.cc",
```
borongc: Should this be moved to under `testing` folder to align with other unit tests?

starboard/shared/starboard/experimental_features.cc
```diff
@@ -55,69 +42,52 @@ const StarboardExtensionExperimentalFeaturesConfigurationApi
 
 }  // namespace
 
+ExperimentalFeatures::ExperimentalFeatures(Map settings)
+    : settings_(std::move(settings)) {}
```
borongc, line 46: Although this is small (up to 256 entries), we will have this `settings_` per thread? Originally, it's a global setting for all threads?

starboard/shared/starboard/player/filter/testing/BUILD.gn
kjyoun: Done

starboard/shared/starboard/experimental_features.cc
kjyoun: No, this PR only refactors the custom struct into a generic map. It does not change the lifecycle or scope of those experimental features.  `ExperimentalFeatures`  was thread-local (configured on only one thread) previously and still is.

borongc (approved):

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
```diff
@@ -438,7 +439,8 @@ void AudioRendererPcm::GetSourceStatus(int* frames_in_buffer,
   *is_eos_reached = is_eos_reached_on_sink_thread_;
   *is_playing = is_playing_on_sink_thread_;
 
-  if (*is_playing || experimental_features_.allow_audio_writing_on_pause) {
+  if (*is_playing ||
+      experimental_features_.GetBool(kMediaAllowAudioWritingOnPause)) {
```
xiaomings, line 443: This queries a map on a hot path, while adding one such query may or may not be a performance concern, we should be careful when adding multiple of them (e.g. the ones in video_renderer_internal_impl.cc, also on hot paths).

starboard/shared/starboard/experimental_features.h
```diff
@@ -15,40 +15,124 @@
...
+
+}  // namespace internal
+
+// Strongly typed identifier for an experimental feature key used within
+// Starboard.
+//
+// Threading & Lifetime: Lightweight constexpr wrappers around a compile-time
+// string_view. Thread-safe and intended to be used as static constants.
+template <typename T>
+class ExperimentalFeatureKey {
```
xiaomings, line 47: This class, and the other code in the file, is quite similar to the ones in `media/base/starboard/experimental_features.h`.  Consider consolidating if possible.  Feel free to keep them separate if we have to.

starboard/shared/starboard/experimental_features.h
```diff
@@ -15,40 +15,124 @@
...
+ public:
+  using ValueType = T;
+  constexpr explicit ExperimentalFeatureKey(std::string_view key) : key_(key) {}
```
xiaomings, line 50: Not sure if this is enough to prevent accidental initialization of transiently allocated strings, or if it's worth to enforce using constructs like `const char(&)[N]`.

starboard/shared/starboard/experimental_features.h
```diff
@@ -15,40 +15,124 @@
...
+  using Value = std::variant<int64_t, std::string>;
+  using Map = std::map<std::string, Value, std::less<>>;
+
+  ExperimentalFeatures() = default;
+  explicit ExperimentalFeatures(Map settings);
+  ~ExperimentalFeatures() = default;
+
+  // Returns the boolean value for the given key, falling back to false if the
+  // key is missing or unset.
+  bool GetBool(const ExperimentalFeatureKey<bool>& key) const {
```
xiaomings, line 72: Can you elaborate the reason to have this function instead of using Get<bool>(...) to access a bool value, and the template type on the ExperimentalFeatureKey?

starboard/extension/experimental/experimental_features.h
```diff
@@ -25,33 +25,18 @@ extern "C" {
 #define kStarboardExtensionExperimentalFeaturesConfigurationName \
   "dev.starboard.extension.ExperimentalFeaturesConfiguration"
 
+typedef struct StarboardExperimentalFeatureEntry {
+  const char* key;
+  int64_t value;
```
xiaomings, line 30: Consider using int64_t* value and using nullptr to indicate nullopt for std::optional.

starboard/shared/starboard/experimental_features.h
kjyoun: Reason is not to modify starboard interface. Since starboard/  cannot depend on  media/ , consolidating them would require changing the public Starboard interface.

starboard/shared/starboard/experimental_features.h
kjyoun: Thanks for catching this. Will address it

starboard/shared/starboard/experimental_features.h
kjyoun: The reason is to avoid repeating  `.Get(key).value_or(false)` boilerplate at every boolean callsite. Since boolean feature flags always default to  false when unset, having  `GetBool(key)`  keeps caller code concise, (== dont need to append ` .value_or(false)` ).

starboard/extension/experimental/experimental_features.h
kjyoun: With this refactoring, an entry is only added to  entries  when a key/value is explicitly set, so we no longer need to pass optional values. 

Previously, pointers were needed to represent optional values because the custom struct contained fixed fields for every feature regardless of whether a value was set.

starboard/shared/starboard/player/filter/audio_renderer_internal_pcm.cc
kjyoun: Make sense.

Though benchmark result shows this query latency (< 0.5 usec) would not make apparent difference during playback (16 msec per each frame), 100 times difference (between map and custom struct) is too large to ignore.

Will address it in a new PR


```
    -------------------------------------------------------------------------------
    Benchmark                                     Time             CPU   Iterations
    -------------------------------------------------------------------------------
    BM_ExperimentalFeatures_Map_GetBool         442 ns          433 ns      1615076
    BM_ExperimentalFeatures_StructMember       3.04 ns         3.00 ns    233139566
```

kjyoun:
@xiaomings 

Thanks for the feedback.

I made a new PR to address them
https://github.com/youtube/cobalt/pull/11259

Please take a look.

starboard/shared/starboard/experimental_features.h
xiaomings: If we can consolidate them and put it in //starboard/common, they can be used by both //media and //starboard.  It's optional but good to have, as we do have duplicated code below and above Starboard.

kjyoun: Thanks for the pointer. I'll take a look, as I also want to avoid code duplication.

My only hesitation is that moving this to starboard/common might spread Starboard dependencies further into media/ than we are doing now  (currently starboard/common is mostly kept under media/starboard/).

Let me think about the best way to share this without breaking that boundary.
https://source.corp.google.com/search?q=include.*starboard%2Fcommon&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:media%2F&start=1

https://source.corp.google.com/search?q=media%2Fbase%2Fstarboard%2Fexperimental_features.h%20f:%5Emedia&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

PR #11120, opened 2026-06-29
Refactor autoroller to dynamically update start.

briantting:
Rather than a hard coded start commit combined with potentially brittle commit title parsing to determine which commits to autoroll, utilize a dyanmically updated start commit stored in .github/AUTOROLL.

Bug: 522987128

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -7,10 +7,7 @@ import re
 import subprocess
 import sys
 
-_SKIP_LIST = {
-    '27.lts': [],
-    'staging': [],
-}
+_AUTOROLL_FILE = '.github/AUTOROLL'
 
 
 @enum.unique
@@ -22,6 +19,10 @@ class CherryPickStatus(enum.Enum):
   FAILED = 'failed'  # The cherry-pick failed due to conflicts or other errors.
 
 
+def log(msg):
+  print(msg, file=sys.stderr)
+
+
 def run(cmd):
   subprocess.run(cmd, check=True, stdout=sys.stderr)
 
@@ -31,50 +32,40 @@ def get_out(cmd):
   return res.stdout
 
 
-def get_commits(source, target, start):
-  """Returns a list of commit lines in chronological order.
+def get_start_commit(target):
+  """Returns a start commit hash or None if CONFLICTED."""
+  start = get_out(['git', 'show', f'{target}:{_AUTOROLL_FILE}']).strip()
 
-  Retrieves commits that are present on the source branch but not on the target
-  branch, optionally starting from a specific commit.
+  if start.startswith('CONFLICTED:'):
+    return None
+  return start
+
+
+def get_commits(source, target):
+  """Returns a list of commits in chronological order or None if CONFLICTED.
+
+  Retrieves commits that are on the source branch but not on the target branch.
+  Uses the target branch's AUTOROLL file as the non-inclusive starting commit.
+  Commits are represented as a (sha, title, pr_num) tuple.
   """
+  start = get_start_commit(target)
+  if start is None:
+    return None
+
   cmd = [
-      'git', 'rev-list', '--oneline', '--no-abbrev-commit', '--reverse', source,
-      f'^{target}'
+      'git', 'rev-list', '--oneline', '--no-abbrev-commit', '--reverse',
+      f'{start}^..{source}'
   ]
-  if start:
-    cmd.append(f'{start}^..{source}')
-  return get_out(cmd).splitlines()
+  lines = get_out(cmd).splitlines()
 
-
-def get_change_id_set(branch, exclude_branch, identifier_type):
-  """Returns a set of change IDs that have been merged into the branch.
-
-  Excludes any changes that exist on the exclude_branch. Automatically accounts
-  for reverted cherry-picks.
-  """
-  if identifier_type == 'pr':
-    pattern = r'Cherry pick PR #(\d+)'
-  else:
-    pattern = r'Cherry pick commit ([0-9a-fA-F]{4,40})'
-
-  change_ids = set()
-  conflicted_change_ids = set()
-  cmd = ['git', 'log', '--reverse', '--format=%s', branch, f'^{exclude_branch}']
-  subjects = get_out(cmd).splitlines()
-  for subject in subjects:
-    match = re.search(fr'^(Revert\s+[\'"]?)?(Conflicted )?{pattern}:', subject)
+  commits = []
+  for line in lines:
+    # match.groups() returns (sha, title, pr_num)
+    # If no PR number is found, pr_num will be None
+    match = re.search(r'^(\w+) (.*?)(?: \(#(\d+)\))?$', line)
     if match:
-      revert, conflicted, change_id = match.groups()
-      if revert:
-        if conflicted:
-          conflicted_change_ids.discard(change_id)
-        else:
-          change_ids.discard(change_id)
-      elif conflicted:
-        conflicted_change_ids.add(change_id)
-      else:
-        change_ids.add(change_id)
-  return change_ids, conflicted_change_ids
+      commits.append(match.groups())
+  return commits
 
 
 def get_unmerged_files():
@@ -91,12 +82,12 @@ def get_unmerged_files():
   for line in lines:
     parts = line.split('\t', 1)
     if len(parts) < 2:
-      print(f'Warning: Malformed line (missing tab): {line}', file=sys.stderr)
+      log(f'Warning: Malformed line (missing tab): {line}')
       continue
     metadata, path = parts
     meta_parts = metadata.split()
     if len(meta_parts) < 3:
-      print(f'Warning: Malformed metadata: {metadata}', file=sys.stderr)
+      log(f'Warning: Malformed metadata: {metadata}')
       continue
     _, _, stage = meta_parts[:3]
     stage_name = stage_map.get(stage, stage)
@@ -104,7 +95,7 @@ def get_unmerged_files():
   return files
 
 
-def resolve_conflicts(unmerged):
+def resolve_conflicts(unmerged_files):
   """Attempts to resolve conflicts automatically.
 
   Returns:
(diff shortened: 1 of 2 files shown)


.github/scripts/autoroll_lts.py
```diff
@@ -201,68 +187,46 @@ def main():
...
-  # All cherry picked changes in target branch since the branch point.
-  target_change_ids, _ = get_change_id_set(args.target_branch,
-                                           args.source_branch,
-                                           args.identifier_type)
-  # All cherry picked changes in autoroll branch since the branch point.
-  autoroll_change_ids, autoroll_conflicted_change_ids = get_change_id_set(
-      'HEAD', args.source_branch, args.identifier_type)
+  # Commits to cherry pick from source to target
+  commits_to_target = get_commits(args.source_branch, args.target_branch)
+  commits_to_autoroll = get_commits(args.source_branch, 'HEAD')
```
oxve: Could you rename these to `commits_on_` instead of `commits_to_`? This confused me for a bit as `commits_to_` can be read as either "commits (that have been made) to" or "commits (that should be made) to".

.github/scripts/autoroll_lts.py
```diff
@@ -201,68 +187,46 @@ def main():
...
+
+  if commits_to_target is None or commits_to_autoroll is None:
+    print('Reached CONFLICTED cherry pick.', file=sys.stderr)
```
oxve: We can do this in a follow-up later if you'd like but these bare print statements make me nervous.

What do you think of adding:
```
def log(str):
  print(str, file=sys.stderr)
```

.github/scripts/autoroll_lts.py
```diff
@@ -31,50 +28,33 @@ def get_out(cmd):
...
   """
-  cmd = [
-      'git', 'rev-list', '--oneline', '--no-abbrev-commit', '--reverse', source,
-      f'^{target}'
-  ]
-  if start:
-    cmd.append(f'{start}^..{source}')
-  return get_out(cmd).splitlines()
+  cmd = ['git', 'show', f'{target}:{_AUTOROLL_FILE}']
+  start = get_out(cmd).strip()
```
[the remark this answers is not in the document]
oxve: Is this an issue?

.github/scripts/autoroll_lts.py
```diff
@@ -7,10 +7,7 @@
 import subprocess
 import sys
 
-_SKIP_LIST = {
```
oxve, line 10: We may still need the skiplist, or do you think it's better to manually bump the autoroll file to skip commits?

.github/scripts/autoroll_lts.py
```diff
@@ -31,50 +28,33 @@ def get_out(cmd):
...
-      f'^{target}'
-  ]
-  if start:
-    cmd.append(f'{start}^..{source}')
-  return get_out(cmd).splitlines()
+  cmd = ['git', 'show', f'{target}:{_AUTOROLL_FILE}']
+  start = get_out(cmd).strip()
 
+  if start.startswith('CONFLICTED:'):
+    return None
```
oxve: Can you break this out into a new function, `get_start_commit(target)`?

.github/scripts/autoroll_lts.py
```diff
@@ -201,68 +187,46 @@ def main():
...
-                                           args.source_branch,
-                                           args.identifier_type)
-  # All cherry picked changes in autoroll branch since the branch point.
-  autoroll_change_ids, autoroll_conflicted_change_ids = get_change_id_set(
-      'HEAD', args.source_branch, args.identifier_type)
+  # Commits to cherry pick from source to target
+  commits_to_target = get_commits(args.source_branch, args.target_branch)
+  commits_to_autoroll = get_commits(args.source_branch, 'HEAD')
+
+  if commits_to_target is None or commits_to_autoroll is None:
```
oxve: I have a couple of issues with this logic. The first is that `commits_to_target` should _never_ have a conflict marked in the autoroll file. In fact, I think you should add a lint check that prevents merging PRs with the CONFLICTED text in the autoroll file. Secondly, it's not immediately obvious that ret val `None` means that there's a conflict that must be resolved before autorolling continues.

I would prefer that you explicitly retrieve the start commit using the new `get_start_commit` function **before** the `get_commits` calls and handle the conflict there.

.github/scripts/autoroll_lts.py
```diff
@@ -201,68 +187,46 @@ def main():
...
+    print('Reached CONFLICTED cherry pick.', file=sys.stderr)
+    return
+
+  # SHAs missing from the autoroll branch
```
oxve: This comment is not correct. `commits_to_autoroll` are the commits already present on the autoroll branch.

.github/workflows/autoroll_lts.yaml
oxve: Should this check also use the autoroll file to determine conflict status?

oxve, line 126: I just realized that this won't work. The autoroll branch is explicitly protected so that no human can push to it. This is to avoid the possibility of someone pushing unreviewed changes to these branches that are then stamped and merged without actual review.

We need a new branch with protection rules that allow humans to push.

.github/scripts/autoroll_lts.py
briantting: I think we can manually bump the hash in the AUTOROLL file to skip for now and in a follow up PR I was going to add the reading of git notes for skip instructions if we want to go that route

.github/workflows/autoroll_lts.yaml
briantting: oh, i tested this and i was able to push to it. Is this because I'm an admin?

```
remote: Resolving deltas: 100% (6/6), completed with 6 local objects.
remote: Bypassed rule violations for refs/heads/autoroll-main-to-staging:
remote: 
remote: - Cannot change this locked branch
```

https://github.com/youtube/cobalt/pull/11135/commits

I think we'll want to change the rule for just the autoroll-staging branch then? 

It seems a bit messy if we get devs to pull down this PR and then push it to another branch and then merge that one. This branch will be broken and will require someone to delete this autoroll branch to unbreak the autoroller. Because the autoroller is reluctant to wipe the branch every time (which it could but then we'd generate PRs endlessly)

briantting: I can modify this from `if [[ "${PR_LINKS}" == CONFLICTED:* ]]; then` to `if [[ -f .github/AUTOROLL ]] && grep -q "^CONFLICTED:" .github/AUTOROLL; then`

that'd also make this less parsey brittle

.github/scripts/autoroll_lts.py

.github/scripts/autoroll_lts.py
briantting: added a bit more description above and in get_commits()

.github/scripts/autoroll_lts.py

.github/scripts/autoroll_lts.py
briantting: Updated description above to clarify.

.github/scripts/autoroll_lts.py

briantting:
```
name: Lint Autoroll State

on:
  pull_request:
    paths:
      - '.github/AUTOROLL'

permissions:
  contents: read

jobs:
  check-conflicts:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout code
        uses: actions/checkout@v4

      - name: Check for conflict marker
        run: |
          if [[ -f .github/AUTOROLL ]]; then
            if grep -q "^CONFLICTED:" .github/AUTOROLL; then
              echo "::error file=.github/AUTOROLL,line=1::The .github/AUTOROLL file indicates a manual conflict resolution is required. Please resolve conflicts, update the file to remove the 'CONFLICTED:' prefix, and commit the changes."
              exit 1
            else
              echo "No conflict markers found in .github/AUTOROLL."
            fi
          else
            echo ".github/AUTOROLL does not exist, skipping check."
          fi
```

something like this will be added to target branches to prevent CONFLICTED from leaking

.github/scripts/autoroll_lts.py
```diff
@@ -31,50 +32,40 @@ def get_out(cmd):
...
-
-  change_ids = set()
-  conflicted_change_ids = set()
-  cmd = ['git', 'log', '--reverse', '--format=%s', branch, f'^{exclude_branch}']
-  subjects = get_out(cmd).splitlines()
-  for subject in subjects:
-    match = re.search(fr'^(Revert\s+[\'"]?)?(Conflicted )?{pattern}:', subject)
+  start = get_start_commit(target)
+  if start is None:
+    return None
```
oxve: I think it would be cleaner if you moved this to before `get_commits` is called below. Then switch `target` arg for `start` or `start_ref`.

.github/scripts/autoroll_lts.py
```diff
@@ -170,99 +157,96 @@ def cherry_pick(sha, num, title, first_cherry_pick):
...
-
-      # Break if in autoroll branch as conflicted.
-      if change_id in autoroll_conflicted_change_ids:
-        print(f"Reached conflicted cherry pick ({sha}).", file=sys.stderr)
-        break
-
-      # Cherry pick PR.
-      first_cherry_pick = not commits_added
-      result = cherry_pick(sha, pr_num, title, first_cherry_pick)
+    identifier = f'- #{pr_num}' if args.identifier_type == 'pr' else f'- {sha}'
```
oxve: `pr_num` can be `None` if no PR number in the commit title too. Switch condition to `... if pr_num else ...`

oxve (approved):
> something like this will be added to target branches to prevent CONFLICTED from leaking

You could put real git conflict markers in the file. They would then be picked up by the conflict marker check in pre-commit. There won't be instructions for the file format though so maybe a dedicated pre-commit check would be in order to enforce the format? In any case I think it should be run as a pre-commit check instead of a new job.

Follow-up PR is good.

.github/scripts/autoroll_lts.py
briantting: Nice, this is simpler. The whole args.identifier_type can be removed now that we use start commit to track cherry pick progress.

.github/scripts/autoroll_lts.py
briantting: Huge improvement in readability. Thanks!

PR #11124, opened 2026-06-29
media: Fix race condition in audio teardown

sideb0ard:
Via primes, we see a Native crash at media::AudioInputStreamDataInterceptor::OnData(..)

The crash is triggered during  audio capture on Android/Starboard when the audio input
stream is being stopped or closed. It appears to be caused by a race condition between
the OpenSLES audio callback thread (executing  ReadBufferQueue ) and the
AudioManager/Main thread (executing Stop  or  Close)

This fix modifies  StarboardAudioInputStream::ReadBufferQueue to keep the lock acquired
during the callback.

Bug: 529427704

### DIFF
--- a/media/audio/android/starboard_audio_input_stream.cc
+++ b/media/audio/android/starboard_audio_input_stream.cc
@@ -279,28 +279,22 @@ void StarboardAudioInputStream::SimpleBufferQueueCallback(
 }
 
 void StarboardAudioInputStream::ReadBufferQueue() {
-  AudioInputCallback* callback = nullptr;
-  {
-    base::AutoLock lock(lock_);
-    if (started_ && callback_) {
-      callback = callback_;
-
-      // Convert from interleaved format to deinterleaved audio bus format while
-      // still under the lock to protect audio_bus_ and audio_data_.
-      audio_bus_->FromInterleaved<SignedInt16SampleTypeTraits>(
-          reinterpret_cast<int16_t*>(audio_data_[active_buffer_index_].get()),
-          audio_bus_->frames());
-    }
-  }
-
-  if (callback) {
-    callback->OnData(audio_bus_.get(),
-                     base::TimeTicks::Now() - hardware_delay_,
-                     /*volume=(0 indicates no vol control)*/0.0,
-                     /*audio_glitch_info=*/{});
-  }
-
   base::AutoLock lock(lock_);
+  if (!started_ || !callback_) {
+    return;
+  }
+
+  // Convert from interleaved format to deinterleaved audio bus format while
+  // still under the lock to protect audio_bus_ and audio_data_.
+  audio_bus_->FromInterleaved<SignedInt16SampleTypeTraits>(
+      reinterpret_cast<int16_t*>(audio_data_[active_buffer_index_].get()),
+      audio_bus_->frames());
+
+  callback_->OnData(audio_bus_.get(),
+                    base::TimeTicks::Now() - hardware_delay_,
+                    /*volume=(0 indicates no vol control)*/0.0,
+                    /*audio_glitch_info=*/{});
+
   if (simple_buffer_queue_) {
     (*simple_buffer_queue_)->Enqueue(simple_buffer_queue_,
                                      audio_data_[active_buffer_index_].get(),
--- a/media/audio/android/starboard_audio_input_stream.cc
+++ b/media/audio/android/starboard_audio_input_stream.cc
@@ -284,6 +284,19 @@ void StarboardAudioInputStream::ReadBufferQueue() {
     return;
   }
 
+  // Calling external callbacks under a lock is not recommended as it
+  // can lead to deadlocks if the callback tries to re-enter
+  // the class or blocks on a thread that is waiting for the lock.
+  // Without this lock however, we can run into a race condition between
+  // the OpenSLES audio callback thread (executing ReadBufferQueue ) and the
+  // AudioManager/Main thread (executing Stop or Close)
+  //
+  // In this case Stop() and Close() have CHECK(thread_checker_.
+  // CalledOnValidThread()) , meaning they must be called from the AudioManager
+  // thread. ReadBufferQueue runs on the OpenSLES thread. Thus, OnData (running
+  // on OpenSLES thread) cannot synchronously call Stop() or Close() on the same
+  // thread, avoiding direct re-entrancy deadlocks.
+
   // Convert from interleaved format to deinterleaved audio bus format while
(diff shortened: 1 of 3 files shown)


media/audio/android/starboard_audio_input_stream.cc
```diff
@@ -279,28 +279,22 @@ void StarboardAudioInputStream::SimpleBufferQueueCallback(
...
+  // Convert from interleaved format to deinterleaved audio bus format while
+  // still under the lock to protect audio_bus_ and audio_data_.
+  audio_bus_->FromInterleaved<SignedInt16SampleTypeTraits>(
+      reinterpret_cast<int16_t*>(audio_data_[active_buffer_index_].get()),
+      audio_bus_->frames());
+
+  callback_->OnData(audio_bus_.get(),
+                    base::TimeTicks::Now() - hardware_delay_,
+                    /*volume=(0 indicates no vol control)*/0.0,
+                    /*audio_glitch_info=*/{});
```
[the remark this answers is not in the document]
sideb0ard: ### Deadlock Risk Assessment   

Calling external callbacks under a lock can sometimes lead to deadlocks if the  
callback tries to re-enter the class or blocks on a thread that is waiting for  
the lock.                                                                       
                                                                                  
1. Re-entrancy:  Stop()  and  Close()  have  CHECK(thread_checker_.
  CalledOnValidThread()) , meaning they must be called from the AudioManager
  thread.  ReadBufferQueue  runs on the OpenSLES thread. Thus,  OnData  (running
  on OpenSLES thread) cannot synchronously call  Stop()  or  Close()  on the same
  thread, avoiding direct re-entrancy deadlocks.

2. Blocking Callbacks: Typical implementations of  AudioInputCallback::OnData
  (like  InputController::OnData ) are designed to be non-blocking. They either
  push data to a FIFO queue or write to shared memory/sockets (synchronous but
  fast, not blocking on other threads). They do not perform synchronous hops to
  the AudioManager thread.

3. Conclusion on Deadlock: The risk of deadlock introduced by this change is
  extremely low and conforms to standard thread-safety patterns for audio streams
  in this codebase.

kjyoun:
nit: Can you update title to  `media: Fix..` like the convention in https://screenshot.googleplex.com/Ay3gTzWG7mwVjxB

media/audio/android/starboard_audio_input_stream.cc
```diff
@@ -279,28 +279,22 @@ void StarboardAudioInputStream::SimpleBufferQueueCallback(
...
-  }
+  // Convert from interleaved format to deinterleaved audio bus format while
+  // still under the lock to protect audio_bus_ and audio_data_.
+  audio_bus_->FromInterleaved<SignedInt16SampleTypeTraits>(
+      reinterpret_cast<int16_t*>(audio_data_[active_buffer_index_].get()),
+      audio_bus_->frames());
+
+  callback_->OnData(audio_bus_.get(),
+                    base::TimeTicks::Now() - hardware_delay_,
+                    /*volume=(0 indicates no vol control)*/0.0,
```
kjyoun: nit: I dont think you need "(0 indicates no vol control)", which is obvious.
`/*volume=*/0.0,`


though this comes from orignal code, it this is a good time to clean it up. since this PR touches it

media/audio/android/starboard_audio_input_stream.cc
```diff
@@ -279,28 +279,22 @@ void StarboardAudioInputStream::SimpleBufferQueueCallback(
...
-                     /*volume=(0 indicates no vol control)*/0.0,
-                     /*audio_glitch_info=*/{});
-  }
+  // Convert from interleaved format to deinterleaved audio bus format while
+  // still under the lock to protect audio_bus_ and audio_data_.
+  audio_bus_->FromInterleaved<SignedInt16SampleTypeTraits>(
+      reinterpret_cast<int16_t*>(audio_data_[active_buffer_index_].get()),
+      audio_bus_->frames());
+
+  callback_->OnData(audio_bus_.get(),
```
kjyoun, line 298: Now,  you are calling `OnData` while holding lock, which is what previous code intentionally avoided.
It can lead to dead lock

Are you confident with this change?

Then, please add comment about why we are calling callback while holding lock and why it is safe.

kjyoun:
one suggestion: how about adding unit test to reproduce the problem and ensure the change fixes it?

I bet Gemini could write one for you.

media/audio/android/starboard_audio_input_stream.cc
sideb0ard: done

media/audio/android/starboard_audio_input_stream.cc
sideb0ard: added comment

sideb0ard:
> one suggestion: how about adding unit test to reproduce the problem and ensure the change fixes it?
> 
> I bet Gemini could write one for you.

test added for AudioInputStreamDataInterceptor (where the segfault happens in the backtrace)

media/audio/android/starboard_audio_input_stream.cc
```diff
@@ -279,28 +279,35 @@ void StarboardAudioInputStream::SimpleBufferQueueCallback(
...
+    return;
   }
 
-  if (callback) {
-    callback->OnData(audio_bus_.get(),
-                     base::TimeTicks::Now() - hardware_delay_,
-                     /*volume=(0 indicates no vol control)*/0.0,
-                     /*audio_glitch_info=*/{});
-  }
+  // Calling external callbacks under a lock is not recommended as it
```
kjyoun: thanks for the comment. 

can you make little bit concise? or keep it if you like

media/audio/audio_input_stream_data_interceptor.cc
```diff
@@ -93,12 +94,29 @@ void AudioInputStreamDataInterceptor::OnData(
     base::TimeTicks capture_time,
     double volume,
     const AudioGlitchInfo& audio_glitch_info) {
+#if BUILDFLAG(IS_STARBOARD)
```
kjyoun: why do we need this change?

is it for test? Then I am OK not having test. 

We dont need to complicate prod code just for testing

sideb0ard: Yeah, its needed for the test. Now removed.

media/audio/android/starboard_audio_input_stream.cc
sideb0ard: done.

kjyoun (approved):

PR #11131, opened 2026-06-29
docs: Fix parsing literals and diagrams in docsite gen

arjungm:
Multiple small issues are fixed in this PR namely:
* Repeated literals when multiple replacements happened inline
* Bad linebreak placements for kebab-case words
* Incorrect diagram and text paragraph differentiation

Bug: 505097004

### DIFF
--- a/cobalt/site/scripts/cobalt_module_reference.py
+++ b/cobalt/site/scripts/cobalt_module_reference.py
@@ -213,6 +213,17 @@ def _node_to_markdown(out, node):
     assert len(node) == 0
     # Don't replace pipes in verbatim text.
     text = node.text if node.text else ''
+    # Strip doxygen comment prefix '///' and one space if present
+    lines = text.split('\n')
+    cleaned_lines = []
+    for line in lines:
+      if line.startswith('/// '):
+        cleaned_lines.append(line[4:])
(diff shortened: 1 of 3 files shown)


cobalt/site/scripts/cobalt_module_reference.py
```diff
@@ -175,9 +175,11 @@ def _node_to_markdown(out, node):
   elif node.tag == 'bold':
     assert len(node) == 0
     out.bold(text)
+    text = ''
   elif node.tag == 'computeroutput':
     assert len(node) == 0
     out.code(text)
+    text = ''
```
oxve, line 182: Why does it only affect these two node types?

oxve (approved):

cobalt/site/scripts/cobalt_module_reference.py
arjungm: this only happens to strings which are surrounded by pipes `|string|` which is converted to regular backticks in markdown

jetski actually just found a bonafide bug because it was failing when there were multiple such replacements in a line

basically it was taking input from a header like:

```
"when |is_playing| is `true`"
```

and turning it into:

```
"when `is_playing` is `true` true"
```

PR #11141, opened 2026-06-30
media: Add platform DRM support for tvOS EME

abhijeetk:
Add platform DRM support for tvOS using the Starboard media pipeline.
This introduces support for SINF, SKD, and PLATFORM_DRM initialization
data types.

Platform DRM support is required for playback of encrypted HLS streams
on tvOS devices. This implementation provides the EME plumbing
necessary to handle CBCS encryption schemes and platform-specific key
exchange formats, used by both the HTML5 player and URL Player on tvOS.

Bug: 501343649
Bug: 498421484

### DIFF
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -43,6 +43,10 @@
 #include "third_party/blink/public/web/web_view.h"
 #include "ui/gfx/geometry/size_conversions.h"
 
+#if BUILDFLAG(IS_IOS_TVOS) && defined(COBALT_INTERNAL_BUILD)
+#include "cobalt/internal/cobalt/components/cdm/renderer/starboard/platform_drm_key_system_info.h"
+#endif  // BUILDFLAG(IS_IOS_TVOS) && defined(COBALT_INTERNAL_BUILD)
+
 namespace cobalt {
 
 namespace {
@@ -426,6 +430,20 @@ void AddStarboardCmaKeySystems(::media::KeySystemInfos* key_system_infos) {
       ::media::EmeFeatureSupport::ALWAYS_ENABLED,    // Persistent state.
       ::media::EmeFeatureSupport::ALWAYS_ENABLED));  // Distinctive
                                                      // identifier.
+
+#if BUILDFLAG(IS_IOS_TVOS) && defined(COBALT_INTERNAL_BUILD)
+  const base::flat_set<::media::EncryptionScheme>
+      kPlatformDrmEncryptionSchemes = {::media::EncryptionScheme::kCbcs};
+
+  key_system_infos->push_back(std::make_unique<cdm::PlatformDrmKeySystemInfo>(
+      codecs,                                        // Regular codecs.
+      kPlatformDrmEncryptionSchemes,                 // Encryption schemes.
+      codecs,                                        // Hardware secure codecs.
+      ::media::EmeFeatureSupport::ALWAYS_ENABLED,    // Persistent state.
+      ::media::EmeFeatureSupport::ALWAYS_ENABLED));  // Distinctive
+                                                     // identifier.
+
+#endif  // BUILDFLAG(IS_IOS_TVOS) && defined(COBALT_INTERNAL_BUILD)
 }
 
 std::unique_ptr<::media::KeySystemSupportRegistration>
--- a/components/cdm/renderer/external_clear_key_key_system_info.cc
+++ b/components/cdm/renderer/external_clear_key_key_system_info.cc
@@ -68,6 +68,13 @@ bool ExternalClearKeyKeySystemInfo::IsSupportedInitDataType(
     case media::EmeInitDataType::KEYIDS:
       return true;
 
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+    // Listed for -Wswitch exhaustiveness only. External Clear Key does not
+    // support platform DRM init data types.
+    case media::EmeInitDataType::SINF:
+    case media::EmeInitDataType::SKD:
+    case media::EmeInitDataType::PLATFORM_DRM:
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
     case media::EmeInitDataType::UNKNOWN:
       return false;
   }
--- a/components/cdm/renderer/external_clear_key_key_system_info.cc
+++ b/components/cdm/renderer/external_clear_key_key_system_info.cc
@@ -68,6 +68,11 @@ bool ExternalClearKeyKeySystemInfo::IsSupportedInitDataType(
     case media::EmeInitDataType::KEYIDS:
       return true;
 
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+    // Listed for -Wswitch exhaustiveness only. External Clear Key does not
+    // support platform DRM init data types.
+    case media::EmeInitDataType::PLATFORM_DRM:
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
     case media::EmeInitDataType::UNKNOWN:
       return false;
   }
--- a/media/base/eme_constants.h
+++ b/media/base/eme_constants.h
@@ -9,6 +9,7 @@
 
 #include <optional>
 
+#include "build/build_config.h"
 #include "media/base/media_export.h"
 #include "media/media_buildflags.h"
 
@@ -21,6 +22,12 @@ enum class EmeInitDataType {
   UNKNOWN,
   WEBM,
   CENC,
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Vendor-specific platform DRM init data formats.
+  SINF,          // sinf atom key IDs
+  SKD,           // skd:// key request URI
+  PLATFORM_DRM,  // AVPlayer encrypted event payload
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
   KEYIDS,
   kMaxValue = KEYIDS,
 };
--- a/media/base/eme_constants.h
+++ b/media/base/eme_constants.h
@@ -9,6 +9,7 @@
 
 #include <optional>
 
+#include "build/build_config.h"
 #include "media/base/media_export.h"
 #include "media/media_buildflags.h"
 
@@ -21,6 +22,10 @@ enum class EmeInitDataType {
   UNKNOWN,
   WEBM,
   CENC,
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Vendor-specific platform DRM init data format.
+  PLATFORM_DRM,  // AVPlayer encrypted event payload
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
   KEYIDS,
   kMaxValue = KEYIDS,
 };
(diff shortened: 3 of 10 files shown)


components/cdm/renderer/fairplay_key_system_info.h
```diff
@@ -0,0 +1,69 @@
...
+#include "media/base/key_system_info.h"
+
+namespace cdm {
+
+// YouTube/Cobalt-specific key system for FairPlay, distinct from Apple's
+// standard "com.apple.fps" variants.
+inline constexpr char kFairplayKeySystem[] = "com.youtube.fairplay";
+
+// KeySystemInfo for Apple FairPlay Streaming.
+class FairplayKeySystemInfo : public media::KeySystemInfo {
```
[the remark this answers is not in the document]
abhijeetk: Done. Expanded the class comment to document purpose, threading model (renderer main thread), and ownership (transferred via KeySystemInfos to KeySystemsImpl).

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -410,6 +413,18 @@ void AddStarboardCmaKeySystems(::media::KeySystemInfos* key_system_infos) {
...
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+  const base::flat_set<::media::EncryptionScheme> kFairPlayEncryptionSchemes = {
+      ::media::EncryptionScheme::kCbcs};
+
+  key_system_infos->emplace_back(new cdm::FairplayKeySystemInfo(
+      codecs,                                        // Regular codecs.
+      kFairPlayEncryptionSchemes,                    // Encryption schemes.
+      codecs,                                        // Hardware secure codecs.
+      ::media::EmeFeatureSupport::ALWAYS_ENABLED,    // Persistent state.
+      ::media::EmeFeatureSupport::ALWAYS_ENABLED));  // Distinctive identifier.
```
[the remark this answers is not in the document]
abhijeetk: Done for FairPlay registration.

media/cdm/cdm_type_conversion.cc
```diff
@@ -129,6 +131,11 @@ cdm::InitDataType ToCdmInitDataType(EmeInitDataType init_data_type) {
       return cdm::kKeyIds;
     case EmeInitDataType::WEBM:
       return cdm::kWebM;
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+    case EmeInitDataType::SINF:
+    case EmeInitDataType::SKD:
+    case EmeInitDataType::FAIRPLAY:
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
abhijeetk: cdm_type_conversion.cc is not compiled as part of the cobalt/nplb/media_unittests build target (it's only used by CdmAdapter for library-based CDMs and cobalt doesn't use library CDM `enable_library_cdms` is false). Removed it from this CL.

media/base/eme_constants.h
```diff
@@ -22,7 +23,16 @@ enum class EmeInitDataType {
...
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+  // FairPlay init data types (sinf/skd from WebKit CDMFairPlayStreaming.cpp,
+  // fairplay from Cobalt C25 application_player.mm).
+  SINF,      // sinf box init data: key IDs embedded in sinf atoms (WebKit)
+  SKD,       // skd:// URI init data: FairPlay key request URI (WebKit)
+  FAIRPLAY,  // Cobalt AVPlayer encrypted event init data (packed skd:// URI)
+  kMaxValue = FAIRPLAY,
+#else
   kMaxValue = KEYIDS,
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
abhijeetk: Not applicable in this case. 

`media::mojom::EmeInitDataType` is declared as `[Native]` in `content_decryption_module.mojom` and is typemapped to `media::EmeInitDataType`,  rather than being defined as a Mojo enum.                                                        

Serialization for this native enum is covered by IPC_ENUM_TRAITS_MAX_VALUE in `media_param_traits_macros.h`. That macro references `media::EmeInitDataType::kMaxValue`, which picks up FAIRPLAY under the tvOS/Starboard build flags.

abhijeetk:
cc : @gkim @jkim-julie @rakuco

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -23,6 +24,9 @@
 #include "cobalt/renderer/cobalt_render_frame_observer.h"
 #include "cobalt/shell/common/url_constants.h"
 #include "components/cdm/renderer/widevine_key_system_info.h"
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun: I dont think you need "BUILDFLAG(USE_STARBOARD_MEDIA)", since this file is already under cobalt/

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -23,6 +24,9 @@
...
+#include "components/cdm/renderer/fairplay_key_system_info.h"
+#endif  // BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun: can you #if guared block to the bottom?

e.g.) 
https://source.chromium.org/chromium/chromium/src/+/main:media/base/media_log.h;l=11-33;drc=42662a1b00845c31327c0b0a3d0fc33d2b285ee9

third_party/blink/renderer/platform/media/web_content_decryption_module_session_impl.cc
```diff
@@ -112,6 +113,15 @@ bool SanitizeInitData(media::EmeInitDataType init_data_type,
       return true;
     }
```
kjyoun, line 115: can you add comments about why we need change for fairplay?
Technically, this module should DRM-agnostic.

Doesn't fariplay follow EME spec?

e.g.)
Apple's FairPlay Streaming (FPS) relies on custom/vendor-specific initialization data formats:

components/cdm/renderer/external_clear_key_key_system_info.cc
```diff
@@ -68,6 +68,13 @@ bool ExternalClearKeyKeySystemInfo::IsSupportedInitDataType(
     case media::EmeInitDataType::KEYIDS:
       return true;
 
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun, line 71: why fairplay-specific code is on clearkey system code?

abhijeetk: The reason this file has to mention them is compile hygiene: Clear Key is always supported by Chromium, and `external_clear_key_key_system_info.cc` is always part of the `components/cdm/renderer` target.

When the tvOS/Starboard build expands `EmeInitDataType` with `SINF`, `SKD`, and `FAIRPLAY`, this always-compiled switch also sees those enum values. Listing them here makes the rejection explicit and keeps the switch exhaustive in that build configuration. It avoid build error : `external_clear_key_key_system_info.cc:65:11: error: enumeration values 'SINF', 'SKD', and 'FAIRPLAY' not handled in switch [-Werror,-Wswitch]`

We are not adding FairPlay support to External Clear Key. The FairPlay init-data types return `false` here

cobalt/renderer/cobalt_content_renderer_client.cc
abhijeetk: Done.

cobalt/renderer/cobalt_content_renderer_client.cc
abhijeetk: Done.

third_party/blink/renderer/platform/media/web_content_decryption_module_session_impl.cc
abhijeetk: Added comment as per your suggestion.

> Doesn't fariplay follow EME spec?

The EME spec defines standard init data types like cenc, keyids, and webm but FairPlay instead uses Apple-proprietary formats: SINF/SKD. Since they are non-standard, we pass FairPlay init data through as-is and let the Starboard CDM handle parsing and validation of these Apple-specific formats directly.

components/cdm/renderer/BUILD.gn
```diff
@@ -3,8 +3,13 @@
 # found in the LICENSE file.
 
 import("//components/cdm/common/playready.gni")
+import("//media/media_options.gni")
 import("//third_party/widevine/cdm/widevine.gni")
 
+if (is_cobalt && is_ios) {
+  import("//build/config/apple/mobile_config.gni")
+}
+
```
rakuco: I think we can get away without this `import` and just check for `is_ios` below since we don't need to care about non-tvOS `target_platform` values anyway.

media/base/eme_constants.h
```diff
@@ -22,7 +23,16 @@ enum class EmeInitDataType {
   WEBM,
   CENC,
   KEYIDS,
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
+  // FairPlay init data types (sinf/skd from WebKit CDMFairPlayStreaming.cpp,
+  // fairplay from Cobalt C25 application_player.mm).
+  SINF,      // sinf box init data: key IDs embedded in sinf atoms (WebKit)
+  SKD,       // skd:// URI init data: FairPlay key request URI (WebKit)
+  FAIRPLAY,  // Cobalt AVPlayer encrypted event init data (packed skd:// URI)
+  kMaxValue = FAIRPLAY,
```
rakuco: What if you add this to the beginning instead? This would get rid of the `#else` and always point `kMaxValue` to `KEYIDS`.

media/starboard/starboard_cdm.cc
```diff
@@ -35,6 +37,14 @@ const char* GetInitDataTypeName(EmeInitDataType type) {
       return "cenc";
     case EmeInitDataType::KEYIDS:
       return "keyids";
+#if BUILDFLAG(IS_IOS_TVOS) && BUILDFLAG(USE_STARBOARD_MEDIA)
```
kjyoun: I dont think you need USE_STARBOARD_MEDIA here under media/starboard directory

kjyoun (approved):

components/cdm/renderer/starboard/fairplay_key_system_info.cc
```diff
@@ -0,0 +1,103 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
```
borongc: As this is a new file, consider to put this under `//components/cdm/renderer/starboard` (both .cc and .h), so they can be formatted properly.

components/cdm/renderer/BUILD.gn
abhijeetk: Done.

media/base/eme_constants.h
abhijeetk: Good suggestion, done.

media/starboard/starboard_cdm.cc
abhijeetk: Done. Thanks.

components/cdm/renderer/starboard/fairplay_key_system_info.cc
abhijeetk: Done. Moved both fairplay_key_system_info.cc and .h to components/cdm/renderer/starboard/. Updated header guards, include paths, and BUILD.gn accordingly.

fayev-g (approved):

abhijeetk:
Hi @jasonzhangxx , PTAL when you can! Let us know if there are any other changes needed. Happy to update.

jasonzhangxx (approved):

PR #11155, opened 2026-06-30
[WIP] media: For URL-Player, Wire duration and buffered ranges from AVPlayer to JS

abhijeetk:
Add duration and buffered time range reporting from the native URL
player (AVPlayer) through Mojo to the JS video.duration and
video.buffered APIs.

Bug: 512045535


abhijeetk:
Closing this PR.

PR #11159, opened 2026-06-30
cobalt: Hide scrollbars by default

sherryzy:
Blink allocates memory for scrollbars even if they are not displayed. This change enables the --hide-scrollbars flag by default for both Starboard (RDK, etc.) and Android platforms to avoid this memory allocation.

For Starboard, the flag is added to the default toggle switches. For Android, the flag is added to the CommandLineOverrideHelper default overrides. Unit tests are updated to ensure the flag is present by default.

Estimated memory saving: 48kb.

Bug: 529788349
Bug: 435264179

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
@@ -72,6 +72,8 @@ public final class CommandLineOverrideHelper {
         paramOverrides.add("--enable-zero-copy");
         // Set default raster threads to 2 for smoother performance.
         paramOverrides.add("--num-raster-threads=2");
+        // Hide scrollbars to avoid memory allocation.
+        paramOverrides.add("--hide-scrollbars");
 
         return paramOverrides;
     }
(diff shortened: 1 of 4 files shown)


johnxwork (approved):

andrewsavage1 (approved):

andrewsavage1:
Added bug 435264179 to the description

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -72,6 +72,8 @@ public static List<String> getDefaultCommandLineOverridesList() {
         paramOverrides.add("--enable-zero-copy");
         // Set default raster threads to 2 for smoother performance.
         paramOverrides.add("--num-raster-threads=2");
+        // Hide scrollbars to avoid memory allocation.
+        paramOverrides.add("--hide-scrollbars");
```
borongc, line 76: Should this be gated by an experimental flag?

andrewsavage1: I don't think so—when would we want to show scrollbars?

borongc: We see WT regression from 7.11 to 7.12, and this is one of the ungated features, though this may be no impact.

andrewsavage1: I'd be surprised if this is what caused the regression, as it should have no user-facing impact on the app without touch enabled. It is possible that some hidden thing it disabled impacted us though, I don't know for sure

borongc: Not particular to this PR. Just want to say if this is backed by the experiment results, we don't need to rely on "human judgement" to determine where the regression come from. Also, as this claims to have memory saving, it doesn't hurt us to run experiments and get the metrics. We don't have the conclusion on where the WT regression comes from. Keep the discussion opened.

PR #11163, opened 2026-06-30
android: Use @CalledByNativeUnchecked for media teardown JNI resilience

johnxwork:
## Motivation
During app switching or Activity teardown on Android and Fire TV devices, media pipeline background threads (`AudioTrack`, `MediaCodec`, `MediaDrm`) may still be executing teardown routines while the Android framework is destroying Activity contexts or unloading ClassLoaders. 

In Cobalt 27 (`main`), legacy JNI helper functions (like `jni_env_ext`) were removed in favor of Chromium's `jni_zero` architecture. When Java methods are annotated with `@CalledByNative`, `jni_zero` emits checked call contexts (`JniJavaCallContext<true>`) that automatically invoke `CheckException(env)` when the call returns. In Cobalt, this maps to `base::android::CheckException()`, which calls `LOG(FATAL)` and terminates the application with `SIGABRT` if any unhandled Java exception or class lookup error occurs during teardown.

## Solution
This PR introduces architectural resilience against teardown-related JNI exceptions by:
1. Migrating teardown-sensitive methods (`release`, `stop`, `pause`, `flush`, `destroy`, `closeSession`) in `AudioTrackBridge.java`, `MediaCodecBridge.java`, and `MediaDrmBridge.java` from `@CalledByNative` to `@CalledByNativeUnchecked`.
2. Updating the corresponding C++ destructors and cleanup functions (`~MediaCodecBridge`, `Flush`, `~AudioTrackBridge`, `Pause`, `Stop`, `PauseAndFlush`, `~MediaDrmBridge`, `CloseSession`) to explicitly check and clear any pending JNI exceptions using `jni_zero::ClearException(env)`.

This ensures that benign exceptions occurring during asynchronous app-switch teardown are safely ignored and cleared without triggering fatal process termination.

Bug: 455621481
TAG=agy
CONV=192bf159-8430-4733-a690-cc90683af132

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
@@ -321,47 +321,59 @@ public class AudioTrackBridge {
     }
   }
 
-  // TODO (b/262608024): Have this method return a boolean and return false on failure.
   @CalledByNative
   private void pause() {
-    if (mAudioTrack == null) {
-      Log.e(TAG, "Unable to pause with NULL audio track.");
-      return;
-    }
     try {
+      if (mAudioTrack == null) {
+        Log.e(TAG, "Unable to pause with NULL audio track.");
+        return;
+      }
       mAudioTrack.pause();
-    } catch (IllegalStateException e) {
-      Log.e(TAG, String.format(Locale.US, "Unable to pause audio track, error: %s", e.toString()));
+    } catch (Throwable t) {
+      // Catch Throwable (both Exception and Error) to prevent JNI crashes if the JVM
+      // throws linkage errors (e.g., NoClassDefFoundError) during ClassLoader unloading
+      // in teardown. See b/455621481.
+      Log.e(TAG, "Exception or Error during AudioTrack.pause()", t);
     }
   }
 
   // TODO (b/262608024): Have this method return a boolean and return false on failure.
   @CalledByNative
   private void stop() {
-    if (mAudioTrack == null) {
-      Log.e(TAG, "Unable to stop with NULL audio track.");
-      return;
-    }
     try {
+      if (mAudioTrack == null) {
+        Log.e(TAG, "Unable to stop with NULL audio track.");
+        return;
+      }
       mAudioTrack.stop();
-    } catch (IllegalStateException e) {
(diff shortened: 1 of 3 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/media/AudioTrackBridge.java
```diff
@@ -349,7 +350,7 @@ private void stop() {
     }
   }
 
-  @CalledByNative
+  @CalledByNativeUnchecked
   private void flush() {
     if (mAudioTrack == null) {
```
kjyoun: I think that catching all exception in java layer is cleaner solution,

WDYT ?

```
@CalledByNative
private void flush() {
  // We catch all exception here. Exception thrown to native crashes app
  // See b/455621481
  try {
      fiushInternal();
   } catch(Exception e) {
       Log(e);
   }
}

private void flushInternal {
   ..
```

johnxwork: Thanks for the advice. Revised accordingly.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
```diff
@@ -718,7 +718,7 @@ private int flush() {
     }
     try {
       mMediaCodec.get().flush();
-    } catch (Exception e) {
+    } catch (Throwable e) {
```
kjyoun, line 721: Why Throwable?

can you add justification about why?

since catching blanket error is anti-pattern, justification will help reader understand why

kjyoun (commented):
you may wan to update your commit description now.

kjyoun (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/media/MediaCodecBridge.java
johnxwork: This is to catch Error corresponding to runtime errors that are not exceptions such as `NoClassDefFoundError`. Updated comments to include these info.

zhongqiliang:
The lifecycle event thing is still a problem for the app.
Is it possible to keep Android waiting for these to be finished, then destroy the objects?

zhongqiliang (approved):
Approved, but a concern.

It is not fixing the root cause.

PR #11167, opened 2026-06-30
Update autoroll CONFLICTED commit instructions.

briantting:
Also fix inclusive start commit bug when it should be non-inclusive.

Bug: 522987128

### DIFF
--- a/.github/workflows/autoroll_lts.yaml
+++ b/.github/workflows/autoroll_lts.yaml
@@ -119,13 +119,20 @@ jobs:
             PR_TITLE="CONFLICTED: ${PR_TITLE}"
             {
               echo ""
-              echo '⚠️ This PR has conflicts that must be resolved manually (do not forget to update the `.github/AUTOROLL` file). Use the following commands to checkout the branch locally and resolve the conflicts in a separate commit:'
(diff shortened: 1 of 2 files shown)


.github/workflows/autoroll_lts.yaml
```diff
@@ -119,13 +119,22 @@ jobs:
...
+              echo ""
+              echo "# Copy CONFLICTED commit's message and metadata"
+              echo "# Replace CONFLICTED in title with RESOLVED"
+              echo "git commit -c <conflicted_sha>"
+              echo ""
               echo "git push origin autoroll-${SOURCE_BRANCH}-to-${TARGET_BRANCH}"
               echo '```'
-              echo 'This PR must be `rebased and merged` and not `squashed and merged` to preserve conflict resolution history.'
+              echo ""
+              echo "This PR must be \`rebased and merged\` and not \`squashed and merged\` to preserve conflict resolution history."
```
oxve: I don't think we need to include this bit. The history of the resolution will be saved in the PR

oxve (approved):

.github/workflows/autoroll_lts.yaml
briantting: ack!

PR #11202, opened 2026-07-02
tvOS: Always invoke GetPlatformWindow() from the main (UI) thread

rakuco:
This seems to be the last invocation of UIKit code from a different thread, which triggers warnings by Xcode's main thread checker.

Tests:
- Verified that this code still works on the device when display criteria matching is enabled.
- `PlayerComponentsTest*` pass without warnings on the simulator when the Main Thread Checker is enabled.
- Video playback in the simulator works without warnings from the Main Thread Checker.

Fixed: 477533116


rakuco:

fayev-g (approved):

PR #11207, opened 2026-07-03
tvos: Support events from GCControllerDirectionPad

jkim-julie:
Add support for mapping GCControllerDirectionPad input from tvOS game
controllers, such as the Siri Remote, into web touch events. This
allows D-pad X and Y axis values to be passed to the Blink rendering
engine.

The logic handles controller connection notifications and maps d-pad
state changes to touch start, move, and end events. This
functionality is ported from Cobalt 25.

Bug: 521473039

### DIFF
--- a/components/input/web_input_event_builders_ios.h
+++ b/components/input/web_input_event_builders_ios.h
@@ -37,6 +37,11 @@ class COMPONENT_EXPORT(INPUT) WebTouchEventBuilder {
       UIEvent* event,
       UIView* view,
       const std::optional<gfx::Vector2dF>& view_offset);
+#if BUILDFLAG(IS_COBALT)
+  static blink::WebTouchEvent BuildFromGamepadData(
+      blink::WebInputEvent::Type type,
+      CGPoint point);
+#endif
 };
 
 }  // namespace input
--- a/components/input/web_input_event_builders_ios.mm
+++ b/components/input/web_input_event_builders_ios.mm
@@ -362,4 +362,53 @@ blink::WebTouchEvent WebTouchEventBuilder::Build(
   return result;
 }
 
+#if BUILDFLAG(IS_COBALT)
+blink::WebTouchEvent WebTouchEventBuilder::BuildFromGamepadData(
+    blink::WebInputEvent::Type type,
+    CGPoint point) {
+  blink::WebTouchEvent result(type, 0, ui::EventTimeForNow());
+  blink::WebTouchPoint::State state =
+      blink::WebTouchPoint::State::kStateUndefined;
+  switch (type) {
+    case blink::WebInputEvent::Type::kTouchStart:
+      state = blink::WebTouchPoint::State::kStatePressed;
+      break;
+    case blink::WebInputEvent::Type::kTouchEnd:
+      state = blink::WebTouchPoint::State::kStateReleased;
+      break;
+    case blink::WebInputEvent::Type::kTouchMove:
+      state = blink::WebTouchPoint::State::kStateMoved;
+      break;
+    case blink::WebInputEvent::Type::kTouchCancel:
+      state = blink::WebTouchPoint::State::kStateCancelled;
+      break;
+    default:
+      NOTREACHED() << "Invalid types for touch events." << type;
+  }
+  result.dispatch_type =
+      result.GetType() == blink::WebInputEvent::Type::kTouchCancel
+          ? blink::WebInputEvent::DispatchType::kEventNonBlocking
+          : blink::WebInputEvent::DispatchType::kBlocking;
+  result.hovering = type == blink::WebInputEvent::Type::kTouchEnd;
+  result.unique_touch_event_id = ui::GetNextTouchEventId();
+  result.touches_length = 1;
+
+  blink::WebTouchPoint touch;
+  // tvOS supports only one touch at the same time.
+  SetWebPointerPropertiesFromMotionEventData(touch, /*pointer_id=*/1,
+                                             /*pressure=*/NAN);
+  touch.state = state;
+  gfx::PointF position(point);
+  // The gamepad data read from the Siri Remote is normalized from -1 to 1 and
+  // this is what Kabuki uses. Note that this behavior is not spec-compliant,
+  // as the reported values are not relative to the viewport or the screen.
+  // TODO(532457474): Use a more standardized approach.
+  touch.SetPositionInScreen(position);
+  touch.SetPositionInWidget(position);
+  result.touches[0] = touch;
+
+  return result;
+}
+#endif
+
 }  // namespace input
--- a/content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
+++ b/content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
@@ -16,6 +16,12 @@
 #include "ui/events/keycodes/dom/dom_key.h"
 #include "ui/events/keycodes/keyboard_codes.h"
 
+#if BUILDFLAG(IS_COBALT)
+#import <GameController/GameController.h>
+
+#include "components/input/web_input_event_builders_ios.h"
+#endif
+
 static void* kObservingContext = &kObservingContext;
 
 namespace {
@@ -87,6 +93,15 @@ RemoteButton remoteButtonFromPressType(UIPressType type) {
 
 }  // namespace
 
+#if BUILDFLAG(IS_COBALT)
+@interface RenderWidgetUIView () {
+  NSMutableSet<GCController*>* _gameControllers;
+  CGPoint _lastSiriRemoteGamepadLocation;
+  BOOL _isBeingTouched;
+}
+@end
+#endif
+
 @implementation RenderWidgetUIView
 
 #pragma mark - Public
@@ -102,7 +117,25 @@ RemoteButton remoteButtonFromPressType(UIPressType type) {
     // tvOS supports multiple types of input events from the Remote, including
     // the clickpad (touch surface), the clickpad ring (directional control),
     // and various physical buttons.
+#if BUILDFLAG(IS_COBALT)
+    _gameControllers = [[NSMutableSet alloc] init];
+    NSNotificationCenter* notifications = [NSNotificationCenter defaultCenter];
+    [notifications addObserver:self
+                      selector:@selector(controllerConnected:)
+                          name:GCControllerDidConnectNotification
+                        object:nil];
+    [notifications addObserver:self
+                      selector:@selector(controllerDisconnected:)
+                          name:GCControllerDidDisconnectNotification
+                        object:nil];
+    // Only register swipe/pan gestures when no game controller is present;
+    // the controller provides its own input events.
+    if (![self addExistingControllers]) {
+      [self addSwipeAndPanGestureRecognizers];
+    }
+#else
     [self addSwipeAndPanGestureRecognizers];
+#endif
   }
   return self;
 }
@@ -148,6 +181,14 @@ RemoteButton remoteButtonFromPressType(UIPressType type) {
 }
 
 - (void)removeView {
+#if BUILDFLAG(IS_COBALT)
+  // Release our strong references to controllers and stop observing
+  // connect/disconnect notifications. The valueChangedHandler blocks capture
+  // weakSelf, so they naturally become no-ops when this view deallocates.
+  [_gameControllers removeAllObjects];
+  [[NSNotificationCenter defaultCenter] removeObserver:self];
+#endif
+
   UIScrollView* view = (UIScrollView*)[self superview];
   [view removeObserver:self
             forKeyPath:NSStringFromSelector(@selector(contentInset))];
@@ -184,6 +225,21 @@ RemoteButton remoteButtonFromPressType(UIPressType type) {
     }
   }
 }
+
+#if BUILDFLAG(IS_COBALT)
+- (void)removeSwipeAndPanGestureRecognizers {
+  NSMutableArray<UIGestureRecognizer*>* toRemove = [NSMutableArray array];
+  for (UIGestureRecognizer* recognizer in self.gestureRecognizers) {
+    if ([recognizer isKindOfClass:[UISwipeGestureRecognizer class]] ||
+        [recognizer isKindOfClass:[UIPanGestureRecognizer class]]) {
+      [toRemove addObject:recognizer];
+    }
+  }
+  for (UIGestureRecognizer* recognizer in toRemove) {
+    [self removeGestureRecognizer:recognizer];
+  }
+}
+#endif
 
 // Helper method to add swipe gestures for `direction`.
 - (void)addSwipeGestureRecognizerWithDirection:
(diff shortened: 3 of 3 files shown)


content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
```diff
@@ -523,6 +558,113 @@ - (void)textFieldDidEndEditing:(UITextField*)textField
...
+  // For now, GCMicroGamepad from the siri remote is only handled.
+  if (!controller.microGamepad) {
+    return;
+  }
+
+  for (GCController* gameController in _gameControllers) {
+    if ([gameController isEqual:controller]) {
+      return;
+    }
+  }
```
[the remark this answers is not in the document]
jkim-julie: Done

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
```diff
@@ -523,6 +558,113 @@ - (void)textFieldDidEndEditing:(UITextField*)textField
...
+- (void)stopListeningToInputForController:(GCController*)controller {
+  NSMutableArray<GCController*>* controllersToRemove = [NSMutableArray array];
+  for (GCController* gameController in _gameControllers) {
+    if ([gameController isEqual:controller]) {
+      [controllersToRemove addObject:gameController];
+    }
+  }
+  for (GCController* gameController in controllersToRemove) {
+    [_gameControllers removeObject:gameController];
+  }
```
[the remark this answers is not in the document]
jkim-julie: Done.

jkim-julie:
cc @abhijeetk @rakuco @Gyuyoung

ui/events/event_constants.h
```diff
@@ -215,7 +215,13 @@ enum class EventPointerType : int {
   kPen,
   kTouch,
   kEraser,
+#if BUILDFLAG(IS_COBALT)
```
rakuco: * How about adding a `&& BUILDFLAG(IS_IOS_TVOS)` here so that you don't need to change files like `motion_event_aura.cc`?
* Does it break anything if you add `kTouchpad` before `kEraser` so that you don't need the `#else`?

content/browser/renderer_host/render_widget_host_view_tvos_uiview.h
```diff
@@ -25,6 +30,12 @@
   // Set on pressesBegan for kSelect when an input field is focused, cleared and
   // used on pressesEnded to show the keyboard.
   BOOL _selectWillShowKeyboard;
+
+#if BUILDFLAG(IS_COBALT)
```
rakuco: You could move this to the `.mm` file and avoid touching this header (which also helps with future rebases):

```objective-c++
@interface Foo {
 @private
  NSMutableSet<GCController*>* _gameControllers;
  CGPoint _lastSiriRemoteGamepadLocation;
  BOOL _isBeingTouched;
}
```

rakuco (commented):
Just checking one thing: does Kabuki not work as intended if the event type is "touch" and not "touchpad"? My understanding from reading this patch is that it would be much more self-contained if we didn't need a custom type, and the gamepad support part could even be upstreamed in this case.

jkim-julie:
> Just checking one thing: does Kabuki not work as intended if the event type is "touch" and not "touchpad"? My understanding from reading this patch is that it would be much more self-contained if we didn't need a custom type, and the gamepad support part could even be upstreamed in this case.

Indeed. When I checked the clientnav test page last year, adding "touchpad" was neccessary but the current YT app works without it. So, I cleaned up touchpad definition.
Regarding the upstream, there is some non-standard part using dpad.xAxis.value and dpad.yAxis.value. But, except it, the rest of the change could be upstreamable. I'll take a look.

content/browser/renderer_host/render_widget_host_view_tvos_uiview.h
jkim-julie: Done.

ui/events/event_constants.h
jkim-julie: With the current YT app, I found that it works without kTouchpad. So, I removed them.

components/input/web_input_event_builders_ios.h
```diff
@@ -37,6 +38,12 @@ class COMPONENT_EXPORT(INPUT) WebTouchEventBuilder {
       UIEvent* event,
       UIView* view,
       const std::optional<gfx::Vector2dF>& view_offset);
+
+#if BUILDFLAG(IS_COBALT) && BUILDFLAG(IS_IOS_TVOS)
```
rakuco: This could be just a check for `IS_COBALT`, as the `IS_IOS_TVOS` suggestion was just for the `//ui` changes that are no longer necessary.

components/input/web_input_event_builders_ios.mm
```diff
@@ -362,4 +362,46 @@ bool IsSystemKeyEvent(const blink::WebKeyboardEvent& event) {
...
+  result.unique_touch_event_id = ui::GetNextTouchEventId();
+  result.touches_length = 1;
+
+  blink::WebTouchPoint touch;
+  // tvOS supports only one touch at the same time.
+  SetWebPointerPropertiesFromMotionEventData(touch, /*pointer_id=*/1,
+                                             /*pressure=*/NAN);
+  touch.state = state;
+  gfx::PointF position(point);
+  // The gamepad data read from the Siri Remote is normalized from -1 to 1 and
```
rakuco, line 402: Could you file a bug to track this? Ideally Kabuki would switch to something more standardized so this could be fully upstreamed.

content/browser/renderer_host/render_widget_host_view_tvos.h
```diff
@@ -28,6 +29,12 @@ class CONTENT_EXPORT RenderWidgetHostViewTVOS : public RenderWidgetHostViewIOS {
   // ui::CALayerFrameSink overrides:
   void UpdateCALayerTree(const gfx::CALayerParams& ca_layer_params) override;
 
+#if BUILDFLAG(IS_COBALT)
+  // ui::GestureProviderClient implementation.
+  void OnGestureEvent(const ui::GestureEventData& gesture) override;
+  bool RequiresDoubleTapGestureEvents() const override;
```
rakuco, line 33: This override specifically (`RequiresDoubleTapGestureEvents()`) could be upstreamed as-is, as I don't think tvOS ever supports double-tap gestures anyway. No need to do it before landing this PR, but please do that in parallel.

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
```diff
@@ -114,6 +127,20 @@ - (instancetype)initWithWidget:
     [self addSwipeGestureRecognizerWithDirection:
               UISwipeGestureRecognizerDirectionDown];
 
+#if BUILDFLAG(IS_COBALT)
+    _gameControllers = [[NSMutableSet alloc] init];
```
rakuco, line 121: I've tested this PR locally, and it actually works _better_ without the existing gesture recognizers: with both this code and the gesture recognizers, a pan gesture would scroll over two items rather than just one, for example.

I'm wondering about a few different options:
* Change the existing gesture recognizers to send touch events instead of keyboard ones (or send both?)
* Disabling the gesture recognizer code in `initWithWidget` based on `IS_COBALT`
* Subclassing `RenderWidgetUIView` in a Cobalt-specific file that removes the gesture recognizers in its `initWithWidget` implementation

Can you check if you can experience the same thing locally (basically removing the gesture recognizers makes the input handling behave more like the existing official YT app) and whether any of the approaches above make sense to you?

components/input/web_input_event_builders_ios.h
jkim-julie: Done.

components/input/web_input_event_builders_ios.mm
jkim-julie: Sure. Here is the issue filed, https://partnerissuetracker.corp.google.com/issues/532457474

content/browser/renderer_host/render_widget_host_view_tvos.h
jkim-julie: SG! Here is the CL, https://chromium-review.googlesource.com/c/chromium/src/+/8063288. In this PR, there are two commits, one is this upstream CL and the other is the rest of the change.

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
jkim-julie: Sure. I think Gamepad event can be the first event handling since it covers all inputs and then other gestures could be secondary since there might be the input device where gamepad input is not available. WDYT?

rakuco: Thanks, this looks good!

Since all pull requests squash all commits, I think it'd be better to land your backport separately from this change to make rebasing easier and to indicate more clearly what's a backport and what isn't.

jkim-julie: SG. Here is a backport PR, https://github.com/youtube/cobalt/pull/11284. Once it's merged, I'll update this branch with the rest of the change on the top of it.

jkim-julie: It's rebased.

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
```diff
@@ -297,6 +353,11 @@ - (BOOL)handlePresses:(NSSet<UIPress*>*)presses
 
 - (void)pressesBegan:(NSSet<UIPress*>*)presses
            withEvent:(UIPressesEvent*)event {
+  // A button press does not implicitly cancel an in-progress gamepad touch
+  // sequence, so cancel it explicitly here before dispatching the key event.
+  if (_isBeingTouched) {
+    [self toucheCancelled];
```
jkim-julie: This part is newly introduced comparing to the previous change. While I was rebasing it, I found a crash issue between touch and press buttons.

Here is the stacktrace for the crash.
[5434:3556274:0715/224144.304875:FATAL:cc/input/input_handler.cc:2346] DCHECK failed: !is_handling_touch_sequence || !is_handling_touch_sequence_. 
0   cobalt                              0x0000000105818c90 base::debug::CollectStackTrace(base::span<void const*, 18446744073709551615ul, void const**>) + 28
1   cobalt                              0x000000010580be3c base::debug::StackTrace::StackTrace(unsigned long) + 224
2   cobalt                              0x0000000105745780 logging::LogMessage::Flush() + 152
3   cobalt                              0x0000000105745668 logging::LogMessage::~LogMessage() + 36
4   cobalt                              0x0000000105732a30 logging::(anonymous namespace)::DCheckLogMessage::~DCheckLogMessage() + 76
5   cobalt                              0x00000001057322f0 logging::CheckError::~CheckError() + 44
6   cobalt                              0x0000000105732350 logging::CheckError::~CheckError() + 12
7   cobalt                              0x000000010709af0c cc::InputHandler::SetIsHandlingTouchSequence(bool) + 208
8   cobalt                              0x000000010a18de2c blink::InputHandlerProxy::DispatchSingleInputEvent(std::__Cr::unique_ptr<blink::EventWithCallback, std::__Cr::default_delete<blink::EventWithCallback>>) + 596

When a button is pressed, touch is started from gamepad and it goes to the state with button pressed without the touch cancelled. After that, if touch is started again, this crash occurs.

So, I added this code to cancel touch when a button is pressed. PTAL.

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
```diff
@@ -526,6 +602,120 @@ - (void)textFieldDidEndEditing:(UITextField*)textField
...
+- (void)touchAtPosition:(CGPoint)position
+              eventType:(blink::WebInputEvent::Type)type {
+  if (!_view) {
+    return;
+  }
+  _view->OnTouchEvent(
+      input::WebTouchEventBuilder::BuildFromGamepadData(type, position));
+}
+
+- (void)toucheCancelled {
```
rakuco: * s/touche/touch/
* Don't you need to update `_lastSiriRemoteGamepadLocation` and/or `_isBeingTouched`?
* Is there an easy way to reproduce the crash you described with the Siri remote?

jkim-julie: Fixed typo and reset members.
Here is how to reproduce the crash. After launching Chrobalt, move to search text with moving focus to search text with swiping and pressing the center from the circle from the remote. Then, return to the previous screen with the back button and swipe.

rakuco (commented):
Thanks for the explanation. I think the solution makes sense (I'm actually surprised about how presses and touches get mixed up like this).

I'd just like to try to reproduce the actual crash here before LGTM'ing it. I've followed your suggestion like this:
* Removed the call to `touchCancelled` from `pressesBegan`
* In the main YT TV page, swipe left to the menu, then swipe up to focus on Search, then press the remote controller to switch to the search page
* Press the Back button to return to the main page
* Try to swipe in any direction

Is that the right order? Also, the backtrace points to a touch start event starting while one has already started, right? Doesn't pressing the remote control button require you to lift your finger and trigger a touch end event?

jkim-julie:
> Is that the right order? 

Yes. It's the right order.

> Also, the backtrace points to a touch start event starting while one has already started, right?

Yes, right.

> Doesn't pressing the remote control button require you to lift your finger and trigger a touch end event?

I also expected that, but the touch end event doesn't occur on presssing and releasing the button. I'm not sure it if it happens only on my environment.

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
```diff
@@ -297,6 +353,13 @@ - (BOOL)handlePresses:(NSSet<UIPress*>*)presses
 
 - (void)pressesBegan:(NSSet<UIPress*>*)presses
            withEvent:(UIPressesEvent*)event {
+  // A button press does not implicitly cancel an in-progress gamepad touch
+  // sequence, so cancel it explicitly here before dispatching the key event.
```
rakuco, line 358: small nit: I'd move this comment inside the `#if` block since it's about the code that's in there.

rakuco (approved):
lgtm. Thanks for the explanations. I still couldn't reproduce the crash here, but I agree the `touchCancelled` thing makes sense (I was at least able to press the clickpad button without triggering a "touchend" event so when I touched it again later I got a "touchmove" instead of "touchstart" one).

content/browser/renderer_host/render_widget_host_view_tvos_uiview.mm
jkim-julie: Done.

rakuco:
@y4vor @andrewsavage1 could either of you rubber-stamp these changes?

andrewsavage1 (approved):
It would be great if in the future we could put this in less platform-specific code to allow game consoles to do the same thing without repeating code

fayev-g (approved):

PR #10845, opened 2026-06-10
Gate Shared Storage under enable_privacy_sandbox_apis

linxinan-yt:
This CL gates the Shared Storage API and its integration with Web Locks under the `enable_privacy_sandbox_apis` build flag.

- Excludes `content/browser/shared_storage` and Blink's shared storage modules from the build when `enable_privacy_sandbox_apis` is false.
- Gates Shared Storage manager initialization, getters, and deletion in `StoragePartitionImpl` and `WebContentsImpl`.
- Gates Shared Storage worklet context checks in `LockManager` (Web Locks).
- Keeps Private Aggregation and Aggregation Service un-gated (to be gated in a follow-up PR).

This PR would save 570KB for gold APK:

Before:
$ ls -l  out/android-arm_gold/apks/Cobalt.apk
-rw------- 1 foo primarygroup 87545723 Jun 28 07:45 out/android-arm_gold/apks/Cobalt.apk

After:
$ ls -l  out/android-arm_gold/apks/Cobalt.apk
-rw------- 1 foo primarygroup 86952618 Jun 28 07:51 out/android-arm_gold/apks/Cobalt.apk

Bug: 505811196

PR #11209, opened 2026-07-05
Gate the rest Privacy Sandbox APIs for Cobalt 27

linxinan-yt:
Gate below Privacy Sandbox components with `enable_privacy_sandbox_apis`:

  1. Ad Auction / Protected Audience DevTools
  2. Shared Storage DevTools:
      • Gated DevTools auto-attach observers and agent host tracking for shared storage worklets.
  3. Private Aggregation (Internal/Fenced Frames)
  4. Ad Auction
  5. Web IDL Union Bindings Exclusions:
      • Excluded generated V8 bindings for union types.

**Enabled**

$ ls -l out/android-arm_gold/apks/Cobalt.apk
-rw------- 1 foo primarygroup 87607470 Jul  6 01:29 out/android-arm_gold/apks/Cobalt.apk

**Disabled**

$ ls -l out/android-arm_gold/apks/Cobalt.apk
-rw------- 1 foo primarygroup 86178781 Jul  6 01:43 out/android-arm_gold/apks/Cobalt.apk

Bug: 505811196


content/browser/devtools/protocol/storage_handler.h
```diff
@@ -13,18 +13,21 @@
 #include "base/memory/weak_ptr.h"
 #include "base/scoped_observation.h"
 #include "base/types/optional_ref.h"
+#include "build/build_config.h"
 #include "build/buildflag.h"
+#if BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
 #include "components/services/storage/shared_storage/shared_storage_manager.h"
-#include "content/browser/attribution_reporting/attribution_observer.h"
```
linxinan-yt, line 18: why removed this line?

content/browser/devtools/protocol/storage_handler.cc
```diff
@@ -436,12 +440,14 @@ void StorageHandler::SetRenderer(int process_host_id,
...
       process ? process->GetStoragePartition() : nullptr;
+#if BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
   if (interest_group_tracking_enabled_) {
     // Transfer observer registration from old frame's StoragePartition to new;
     // SetInterestGroupTrackingInternal() will handle any nulls.
     SetInterestGroupTrackingInternal(storage_partition_, false);
     SetInterestGroupTrackingInternal(new_storage_partition, true);
   }
+#endif  // BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
   storage_partition_ = new_storage_partition;
```
haozheng-cobalt, line 451: correct me if u see wrong, I remember you trimmed the ENTIRE storage partition?

content/browser/devtools/protocol/storage_handler.cc
```diff
@@ -1195,12 +1222,18 @@ Response StorageHandler::SetInterestGroupTrackingInternal(
   }
   return Response::Success();
 }
+#endif  // BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
 
 Response StorageHandler::SetInterestGroupAuctionTracking(bool enable) {
+#if BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
   interest_group_auction_tracking_enabled_ = enable;
   return Response::Success();
+#else
```
haozheng-cobalt, line 1231: I'm curious, some functions are gated entirely, and some functions you gated the body and provide else condition, it's on purpose?

haozheng-cobalt:
A little confused if we need to gate the entire or part of the Storage Partition or Shared Storage, will be helpful to clarify the question.

content/browser/devtools/protocol/storage_handler.cc
linxinan-yt: Storage partition is not a private sandbox feature, hence it is not on the list to trim. The scope is listed in https://b.corp.google.com/issues/505811196#comment8. 
#10845 has gated shared storage but not storage partition.

content/browser/devtools/protocol/storage_handler.cc
linxinan-yt: TL;DR 

If the function is declared in the generated DevTools protocol base class(e.g. GetInterestGroupDetails() ,  SetSharedStorageEntry() ,  DeleteSharedStorageEntry() , etc), I gated body. Otherwise I have to modify the entire DevTools. It won't save us lots of bytes, but the complexity may explode.

content/browser/devtools/protocol/storage_handler.cc
andrewsavage1, line 1163: Why don't we want to use this path? Is it too messy or does it not have the same binary savings?

andrewsavage1 (commented):
FYI I created https://github.com/youtube/cobalt/pull/11348 as a partial stubbing approach, just going to spend some time comparing the two approaches today

content/browser/devtools/protocol/storage_handler.cc
linxinan-yt: I simply want to remove as much as possible. Since we don't use PS, we can skip `if (!manager)` in runtime. WDYT?

haozheng-cobalt:
> FYI I created #11348 as a partial stubbing approach, just going to spend some time comparing the two approaches today

This PR might be improved with centralized stubbing approach in the content/browser/ (Pure C++ / DevTools / StoragePartition) part. However there is non-negligible overhead maintaining cobalt stubbing files too, we have precedence when an incorrect implementation of cobalt stubbing files causing app to crash and it was hard to catch when the stub class is huge. 
The privacy sandbox will go away in upstream Chromium soon too, I'd suggest to submit this PR if there is no major issues.

haozheng-cobalt (approved):

andrewsavage1:
> -rw------- 1 foo primarygroup 86178781 Jul 6 01:43 out/android-arm_gold/apks/Cobalt.apk

I assume these changes are from additionally setting the flag enable_privacy_sandbox_apis to false, right? Wouldn't we also want to do that in this PR?

content/browser/BUILD.gn
andrewsavage1, line 3771: It's strange to me that the conditions in c++ source code and GN code don't align—why wouldn't we want a guard here as well that the chromium milestone is ≤m138?

content/browser/devtools/protocol/storage_handler.cc
```diff
@@ -46,23 +46,27 @@
 #include "components/services/storage/privileged/mojom/indexed_db_control.mojom.h"
 #include "components/services/storage/public/cpp/buckets/bucket_locator.h"
 #include "components/services/storage/public/mojom/cache_storage_control.mojom.h"
-#include "content/browser/attribution_reporting/aggregatable_result.mojom.h"
+#if BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
```
andrewsavage1, line 49: Could you help me understand why you're using both of these flags? Is the privacy sandboxed removed in upstream Chromium at trunk, meaning that when we rebase we could delete all of this changed code?

linxinan-yt:
> > -rw------- 1 foo primarygroup 86178781 Jul 6 01:43 out/android-arm_gold/apks/Cobalt.apk
> 
> I assume these changes are from additionally setting the flag enable_privacy_sandbox_apis to false, right? Wouldn't we also want to do that in this PR?

I would like to file another PR to flip it. Just in case the PS removal breaks sth, it will be easy for anyone to revert.

content/browser/devtools/protocol/storage_handler.cc
linxinan-yt: Yes, that is the goal. The privacy sandbox removal is for 27.lts. With CHROMIUM_MILESTONE_LE_138, the rebase can completely ignore these changes. It is certain that our removal will have conflicts with upstream's removal around M150. And I don't want cobalt developers to spend time to resolve such conflicts.

linxinan-yt: > Is the privacy sandboxed removed in upstream Chromium at trunk

To be accurate, not all privacy sandbox features will be removed. See https://b.corp.google.com/issues/505811196#comment8.

For 27.lts, I only removed the deprecated features.

After 27.lts, if we have bandwidth we can continue stripping PS features not used in Cobalt, e.g. Private State Tokens, storage partition etc. So I assume we would keep the flag ENABLE_PRIVACY_SANDBOX_APIS in Cobalt.

content/browser/BUILD.gn
linxinan-yt: Unlike the C++ source code, the changes in gn are mostly cobalt related. This makes sense because we define how to build Cobalt in gn. As a result the change in GN has fewer chances to hit conflict with upstream.
Plus, at the upstream side, when Chromium removes privacy sandbox FRs, they could directly remove PS's entire source code, and does not need to maintain a filter in GN like us. So probably we have to delete or partially delete the filter when the files are removed from upstream.
Overall we cannot ignore the changes in GN after M138, which is the reason I did not put ≤m138 flag here.

andrewsavage1 (approved):

PR #11212, opened 2026-07-06
aosp: fix fstatat() calls for android asset paths

rogerzanoni:
Routes fstatat(AT_FDCWD, ..., flag == 0) for Android asset paths
through the __wrap_stat implementation.

Fixes the following issues in multiple tests:
asset_manager.cc: Asset path not found

Bug: 532068409


sacuff (approved):

andrewsavage1 (changes requested):
It looks like this customization should happen in android's posix_emu directory instead of being across all platforms

rogerzanoni:
Actually, already fixed by f2a15b9bb2e

PR #11215, opened 2026-07-06
aosp: wrap pread and pwrite

rogerzanoni:
In starboard/elf_loader/exported_symbols.cc pread are exported as

  REGISTER_SYMBOL(pread);
  REGISTER_SYMBOL(pwrite);

but libcobalt is statically linked with musl, which contains those
functions with different signatures:

ssize_t pread(int fd, void *buf, size_t size, off_t ofs);
ssize_t pwrite(int fd, void *buf, size_t size, off_t ofs);

So in modular builds, this change creates and registers ABI wrappers
for those calls, handling the different signatures.

Fixes the following issues in tests:
  posix_file_read_test.cc:286 pread failed: Invalid argument (bytes_read = 4294967295)

Bug: 532068409

### DIFF
--- a/starboard/elf_loader/exported_symbols.cc
+++ b/starboard/elf_loader/exported_symbols.cc
@@ -253,8 +253,6 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_SYMBOL(pause);
   REGISTER_SYMBOL(pipe);
   REGISTER_SYMBOL(posix_memalign);
-  REGISTER_SYMBOL(pread);
-  REGISTER_SYMBOL(pwrite);
   REGISTER_SYMBOL(raise);
   REGISTER_SYMBOL(rand);
   REGISTER_SYMBOL(rand_r);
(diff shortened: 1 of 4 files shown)


starboard/elf_loader/exported_symbols.cc
```diff
@@ -336,6 +338,10 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_WRAPPER(pipe2);
   REGISTER_WRAPPER(poll);
   REGISTER_WRAPPER(prctl);
+#if SB_IS(MODULAR)
```
sacuff: It should be fine to do this on all builds

sacuff (approved):

starboard/elf_loader/exported_symbols.cc
rogerzanoni: Thanks for the review, removed the guards and kept only the REGISTER_WRAPPER lines

andrewsavage1 (approved):

starboard/shared/modular/cobalt_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -75,4 +75,17 @@ int __abi_wrap_unlinkat(int fildes, const char* path, int flag);
...
+
+ssize_t pread(int fd, void* buf, size_t size, off_t ofs) {
+  return __abi_wrap_pread(fd, buf, size, ofs);
+}
+
+ssize_t __abi_wrap_pwrite(int fd, const void* buf, size_t size, off_t ofs);
+
+ssize_t pwrite(int fd, const void* buf, size_t size, off_t ofs) {
+  return __abi_wrap_pwrite(fd, buf, size, ofs);
+}
```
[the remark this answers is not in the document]
andrewsavage1: This is implemented correctly

PR #11216, opened 2026-07-06
aosp: fix musl_statvfs layout

rogerzanoni:
aosp: fix musl_statvfs layout

musl's fsblkcnt_t/fsfilcnt_t are 64-bit on all arch, but bionic defines them
as 32-bit unsigned long on 32-bit targets. On 32-bit arch, the values
can land in wrong offsets in the musl layer.

Fixes the PosixStatvfsTest NPLB tests (SunnyDay, UsageChanges), which read
misaligned fields:
  posix_sys_statvfs_test.cc:46 f_namemax == 0 (expected > 0)

Bug: 532068409

### DIFF
--- a/starboard/shared/modular/starboard_layer_posix_statvfs_abi_wrappers.h
+++ b/starboard/shared/modular/starboard_layer_posix_statvfs_abi_wrappers.h
@@ -25,15 +25,21 @@
 extern "C" {
 #endif
 
+// musl's fsblkcnt_t/fsfilcnt_t are 64-bit on every arch but bionic defines them
+// as 32-bit `unsigned long` on 32-bit targets. Use fixed-width 64-bit types so
+// this struct matches musl's layout.
+typedef uint64_t musl_fsblkcnt_t;
+typedef uint64_t musl_fsfilcnt_t;
+
 struct musl_statvfs {
   unsigned long f_bsize;
   unsigned long f_frsize;
-  fsblkcnt_t f_blocks;
-  fsblkcnt_t f_bfree;
-  fsblkcnt_t f_bavail;
-  fsfilcnt_t f_files;
-  fsfilcnt_t f_ffree;
-  fsfilcnt_t f_favail;
+  musl_fsblkcnt_t f_blocks;
+  musl_fsblkcnt_t f_bfree;
+  musl_fsblkcnt_t f_bavail;
+  musl_fsfilcnt_t f_files;
+  musl_fsfilcnt_t f_ffree;
+  musl_fsfilcnt_t f_favail;
   unsigned long f_fsid;
   unsigned : 8 * (2 * sizeof(int) - sizeof(long));
   unsigned long f_flag;


starboard/shared/modular/starboard_layer_posix_statvfs_abi_wrappers.h
```diff
@@ -28,14 +28,16 @@ extern "C" {
...
+  // them as 32-bit `unsigned long` on 32-bit targets. Use fixed-width 64-bit
+  // fields so this struct matches musl's layout.
+  uint64_t f_blocks;
+  uint64_t f_bfree;
+  uint64_t f_bavail;
+  uint64_t f_files;
+  uint64_t f_ffree;
+  uint64_t f_favail;
   unsigned long f_fsid;
-  unsigned : 8 * (2 * sizeof(int) - sizeof(long));
```
sacuff, line 38: Is removing the padding necessary? Even if it's redundant, we should stay as close to the musl definition as possible, and not having it caused issues before.

rogerzanoni: You are correct, I removed it by mistake. Uploaded a new commit

starboard/shared/modular/starboard_layer_posix_statvfs_abi_wrappers.h
```diff
@@ -28,12 +28,15 @@ extern "C" {
...
-  fsblkcnt_t f_blocks;
-  fsblkcnt_t f_bfree;
-  fsblkcnt_t f_bavail;
-  fsfilcnt_t f_files;
-  fsfilcnt_t f_ffree;
-  fsfilcnt_t f_favail;
+  // musl's fsblkcnt_t/fsfilcnt_t are 64-bit on every arch but bionic defines
+  // them as 32-bit `unsigned long` on 32-bit targets. Use fixed-width 64-bit
+  // fields so this struct matches musl's layout.
+  uint64_t f_blocks;
```
andrewsavage1: Instead of using uint64_t directly, create types musl_fsblkcnt_t and musl_fsfilcnt_t similar to i.e. musl_off_t

rogerzanoni: Created in the current commit. Note: I created it in starboard_layer_posix_statvfs_abi_wrappers.h, like musl_off_t is defined in the wrapper header where it's used, but we can create a central place for the typedefs in follow-up changes (musl_off_t is currently typedefd in two places)

andrewsavage1 (approved):

starboard/shared/modular/starboard_layer_posix_statvfs_abi_wrappers.h
andrewsavage1: While it's only being used in one place let's leave it as is, and then once we need it another place we can put it in a shared location

PR #11220, opened 2026-07-06
aosp: don't return EINVAL for valid optional names in pathconf/sysconf

rogerzanoni:
aosp: don't return EINVAL for valid optional names in pathconf/sysconf

Bionic returns EINVAL for some recognized musl pathconf/sysconf names, but
POSIX reserves EINVAL for invalid names. Intercept them in the modular wrapper
and return -1 (unsupported) without EINVAL.

- sysconf: _SC_V6_*/_SC_XBS5_* are options; <unistd.h> uses value -1 for "not
  supported". _SC_V6_* is absent from the NDK; the legacy _SC_XBS5_* (marked
  LEGACY in Issue 6, removed in Issue 7) are defined but rejected by bionic.
  POSIX leaves an unsupported option's result unspecified, so -1 without EINVAL
  is conformant, not required.
- pathconf: _PC_REC_*_XFER_SIZE are <limits.h> vars with no limit for the file
  (-1 without errno; bionic's EINVAL is the optional association error).
  _PC_SOCK_MAXBUF is a non-POSIX extension the NDK lacks, handled the same way.

References:
  sysconf  https://pubs.opengroup.org/onlinepubs/9799919799/functions/sysconf.html
  pathconf https://pubs.opengroup.org/onlinepubs/9799919799/functions/fpathconf.html
  unistd.h https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/unistd.h.html

Fixed tests:

  PosixPathconfTests/PosixFilePathconfValidNameTest.PathconfDoesNotFailWithEinval
    /12  _PC_SOCK_MAXBUF
    /14  _PC_REC_INCR_XFER_SIZE
    /15  _PC_REC_MAX_XFER_SIZE

  PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames
    /76  _SC_XBS5_ILP32_OFF32
    /77  _SC_XBS5_ILP32_OFFBIG
    /78  _SC_XBS5_LP64_OFF64
    /79  _SC_XBS5_LPBIG_OFFBIG
    /105 _SC_V6_ILP32_OFF32
    /106 _SC_V6_ILP32_OFFBIG
    /107 _SC_V6_LP64_OFF64
    /108 _SC_V6_LPBIG_OFFBIG

Bug: 532068409

### DIFF
--- a/starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
+++ b/starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
@@ -146,6 +146,32 @@ int musl_unlink_flag_to_platform_flag(int musl_flag) {
       return -1;
   }
 }
+
+#if defined(__ANDROID__)
+// Recognized musl names bionic can't resolve. POSIX reserves EINVAL for invalid
+// names, so return -1 (unsupported) without EINVAL translating bionic's correct
+// semantics to a still conformant implementation to match the tests expectations.
+bool IsUnsupportedAndroidSysconfName(int name) {
+  // Options: _SC_V6_* aren't defined in the NDK; legacy _SC_XBS5_* defined but
+  // rejected by bionic. <unistd.h> uses value -1 for "not supported"; POSIX
+  // leaves an unsupported option's result unspecified, so -1/no-EINVAL is
+  // conformant.
+  return name == MUSL_SC_V6_ILP32_OFF32 || name == MUSL_SC_V6_ILP32_OFFBIG ||
+         name == MUSL_SC_V6_LP64_OFF64 || name == MUSL_SC_V6_LPBIG_OFFBIG ||
+         name == MUSL_SC_XBS5_ILP32_OFF32 ||
+         name == MUSL_SC_XBS5_ILP32_OFFBIG ||
+         name == MUSL_SC_XBS5_LP64_OFF64 || name == MUSL_SC_XBS5_LPBIG_OFFBIG;
+}
+
+bool IsUnsupportedAndroidPathconfName(int name) {
+  // _PC_REC_*_XFER_SIZE: <limits.h> vars with no limit for the file -> -1
+  // without errno. Translates bionic's EINVAL to conformant behavior that
+  // matches NPLB tests expectations.
+  // _PC_SOCK_MAXBUF: non-POSIX extension the NDK lacks, handled the same way.
+  return name == MUSL_PC_SOCK_MAXBUF || name == MUSL_PC_REC_INCR_XFER_SIZE ||
+         name == MUSL_PC_REC_MAX_XFER_SIZE;
+}
+#endif  // defined(__ANDROID__)
 }  // namespace
 
 int __abi_wrap_ftruncate(int fildes, musl_off_t length) {
@@ -157,6 +183,11 @@ musl_off_t __abi_wrap_lseek(int fildes, musl_off_t offset, int whence) {
 }
 
 long __abi_wrap_sysconf(int name) {
+#if defined(__ANDROID__)
+  if (IsUnsupportedAndroidSysconfName(name)) {
+    return -1;
+  }
+#endif
   switch (name) {
 #if defined(_SC_ARG_MAX)
     case MUSL_SC_ARG_MAX:
@@ -683,6 +714,11 @@ long __abi_wrap_sysconf(int name) {
 // If |musl_conf_to_platform_conf| returns -1,
 // just return -1 (errno is set to EINVAL by musl_conf_to_platform_conf()).
 long __abi_wrap_pathconf(const char* path, int name) {
+#if defined(__ANDROID__)
+  if (IsUnsupportedAndroidPathconfName(name)) {
+    return -1;
+  }
+#endif
   int converted_name = musl_conf_to_platform_conf(name);
   return (converted_name == -1) ? -1 : pathconf(path, converted_name);
 }


starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -684,7 +967,22 @@ long __abi_wrap_sysconf(int name) {
 // just return -1 (errno is set to EINVAL by musl_conf_to_platform_conf()).
 long __abi_wrap_pathconf(const char* path, int name) {
   int converted_name = musl_conf_to_platform_conf(name);
-  return (converted_name == -1) ? -1 : pathconf(path, converted_name);
+  if (converted_name == kMuslConfUnsupported) {
+    return -1;
+  }
+  if (converted_name == -1) {
+    return -1;
+  }
```
sacuff: nit: combine these `if` blocks

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
sacuff: Looks like functionality was removed here which may affect behavior on other platforms. This PR should probably be moved to an Android-specific wrapper

sacuff (changes requested):

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -684,7 +967,22 @@ long __abi_wrap_sysconf(int name) {
...
+  if (converted_name == kMuslConfUnsupported) {
+    return -1;
+  }
+  if (converted_name == -1) {
+    return -1;
+  }
+  int saved_errno = errno;
+  errno = 0;
+  long result = pathconf(path, converted_name);
+  if (result == -1 && (errno == EINVAL || errno == 0)) {
```
johnedocampo: I see that this if statement covers the case where the platform reports it does not support the `converted_name`, even though on our end we verified that there is a conversion.

I checked the [documentation](https://pubs.opengroup.org/onlinepubs/009696799/functions/fpathconf.html) for pathconf, which says that in this case, the errno should be set.

> If the implementation needs to use path to determine the value of name and the implementation does not support the association of name with the file specified by path ... pathconf() returns -1 and errno is set to indicate the error.

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
johnedocampo: From my understanding, the [documentation](https://pubs.opengroup.org/onlinepubs/000095399/functions/sysconf.html) for sysconf states that:

> If the variable corresponding to name is dependent on an unsupported option, the results are unspecified.

So I would +1 to double checking that this code doesn't break other builds.

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -158,521 +204,758 @@ musl_off_t __abi_wrap_lseek(int fildes, musl_off_t offset, int whence) {
...
       return sysconf(_SC_ARG_MAX);
+#else
+      return -1;  // valid name, unsupported: no EINVAL
 #endif  // defined(_SC_ARG_MAX)
-#if defined(_SC_CHILD_MAX)
     case MUSL_SC_CHILD_MAX:
+#if defined(_SC_CHILD_MAX)
       return sysconf(_SC_CHILD_MAX);
+#else
+      return -1;  // valid name, unsupported: no EINVAL
```
johnedocampo: Is it possible to make a local variable that represents this behavior, maybe `kMuslconfNameValidButUnsupported`? (there's probably a better name here, but just used that one as an example). This would help reduce the repetitive comment for all the return values.

johnedocampo (changes requested):
QQ: Are you using the OpenGroup documentation for these changes? Just want to make sure were using the same source here.

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
rogerzanoni: @sacuff @johnedocampo I decided to try a different approach, instead of changing the helper functions, just intercept the problematic names in the wrapped function. And I didn't create a posix_emu impl because the fix deal's with a mix of names that are defined/not defined in the NDK, so I thought it would be best to handle it in the modular wrappers

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -157,6 +157,20 @@ musl_off_t __abi_wrap_lseek(int fildes, musl_off_t offset, int whence) {
 }
 
 long __abi_wrap_sysconf(int name) {
+#if defined(__ANDROID__)
+  // V6 names are not defined by the Android NDK
+  // XBS5 names are defined but rejected by bionic's sysconf() with EINVAL.
+  // Return -1 without EINVAL (valid POSIX for an unsupported option).
```
andrewsavage1: I don't see this in the documentation. Can you link to/paste the text for that section?

rogerzanoni:
Hey @andrewsavage1 and @johnedocampo, thanks for your reviews, and I missed a few questions/comments when I did a rebases to check the CI status

I'm using the docs, but this was an old change that I picked from an early experiment with cobalt 25, and I didn't update the commit/comment texts to include the spec references, sorry about that. I will push a new commit with updated text and improving the fix a bit. I have some other PRs where I did changes based on the specs (and I also need to add the citations/spec refs), @johnedocampo can I add you there to validate my understanding?

about this fix:

the reasoning for sysconf was that V6_* and XBS5_* are option names, and https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/unistd.h.html mentions that "If a symbolic constant is not defined or is defined with the value -1, the option is not supported for compilation. If it is defined with a value greater than zero, the option shall always be supported when the application is executed. If it is defined with the value zero, the option shall be supported for compilation and might or might not be supported at runtime. See 2.1.6 Options for further information about the conformance requirements of these three categories of support." in "Constants for Options and Option Groups" and _SC_V6_* aren't defined in NDK so it should return -1 as unsupported
_SC_XBS5_* constants are defined but bionic sets EINVAL - I don't think this is incorrect, https://pubs.opengroup.org/onlinepubs/9699919799/utilities/getconf.html mentions that it was marked as legacy in issue 6 and removed in issue 7, but sysconf can return EINVAL (I wrote a little about it in the next paragraph)

about not setting errno, https://pubs.opengroup.org/onlinepubs/9799919799/functions/sysconf.html says "If name is an invalid value, sysconf() shall return -1 and set errno to indicate the error. If the variable corresponding to name is described in <limits.h> as a maximum or minimum value and the variable has no limit, sysconf() shall return -1 without changing the value of errno. Otherwise, sysconf() shall return the current variable value on the system.", so the limits.h rule doesn't apply since they are unistd-defined configuration values and the names aren't invalid, my interpretation is that we should return "unsupported" but EINVAL shouldn't be set, returning the configuration value (unsupported); but here I think we can discuss something, the final statement in the docs about the sysconf value is ambiguous:  "If the variable corresponding to name is dependent on an unsupported option, the results are unspecified." I think returning -1 and not setting EINVAL should be considered conformant but not required as it was written in my commit message.

for pathconf "If the variable corresponding to name is described in <limits.h> as a maximum or minimum value and the variable has no limit for the path or file descriptor, both pathconf() and fpathconf() shall return -1 without changing errno." 

_PC_SOCK_MAXBUF* is not posix (it's an extension) and _PC_REC_*_XFER_SIZE are defined names, and I think it fits the above rule, but there's also another part of pathconf's doc in the "ERRORS" section that says that pathconf *may* fail with EINVAL if "The implementation does not support an association of the variable name with the specified file.", so bionic's behavior isn't wrong, it's only optionally setting EINVAL for it

long story short, the fix intercept the names that were problematic in aosp nplb tests and make aosp behave in a posix conformant way to match the test expectations

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
rogerzanoni: Done

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
rogerzanoni: @johnedocampo that implementation was wrong, it was acting on the return value of pathconf rather than checking the names. That statement holds for this case, but I changed the implementation to actually check the names before any calls, and it's based on the rules for them. I added the reasoning and the references in the comments/commit message, could you please check it again?

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
rogerzanoni: Chaged to a couple of helper functions that are checked before calling sysconf/pathconf

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
rogerzanoni: Added to the commit message and comments, also in the last comment added to the conversation

andrewsavage1 (commented):
I think this should also be an android-specific wrapper in the posix_emu dir. Is there a reason that doesn't work?

PR #11224, opened 2026-07-06
media: Remove unused //media/capture dependency

osagie98:
Cobalt does not support webcam, video capture, or screen sharing.
However, the media/capture component was being pulled in via content,
introducing unnecessary binary bloat from Android JNI wrappers,
file capture drivers, and blob utilities.

This change removes unused video capture implementations and gates
the creation of video capture factories when is_cobalt is enabled.
This reduces the binary footprint of libcobalt.so and the Android
APK by stripping unused camera-related native code and Java bytecode.

Binary Size Verification (gold release builds):
* android-arm_gold:
  - Cobalt.apk: -32,768 bytes (-32.0 KiB)
  - libchrobalt.so (stripped): -45,056 bytes (-44.0 KiB)
  - lib.unstripped/libchrobalt.so: -136,716 bytes (-133.5 KiB)
* evergreen-x64_gold:
  - libcobalt.so (stripped): -53,248 bytes (-52.0 KiB)
  - lib.unstripped/libcobalt.so: -184,304 bytes (-180.0 KiB)
 
#vibe-coded

Bug: 531833882


media/capture/video/create_video_capture_device_factory.cc
```diff
@@ -81,23 +84,33 @@ CreatePlatformSpecificVideoCaptureDeviceFactory(
...
+#if BUILDFLAG(IS_COBALT)
+  return false;
+#else
   const base::CommandLine* command_line =
       base::CommandLine::ForCurrentProcess();
   return command_line->HasSwitch(switches::kUseFakeDeviceForMediaStream);
+#endif
 }
 
 std::unique_ptr<VideoCaptureDeviceFactory> CreateVideoCaptureDeviceFactory(
```
kjyoun, line 92: basically, changes in this file is to make `CreateVideoCaptureDeviceFactory` dummy

Then how about not to this call method even more at upstream?

Then change would be simpler. WDYT ?

Ideally, I think that we should remove whole "//services/video_capture". shouldn't we?


https://source.corp.google.com/search?q=CreateVideoCaptureDeviceFactory&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain

e.g.) 

```

#if !cobalt build
#include "media/capture/video/create_video_capture_device_factory.h"
#endif 

 std::unique_ptr<media::VideoCaptureDeviceFactory> media_device_factory =
#if cobalt build
  nullptr;
#else 
      media::CreateVideoCaptureDeviceFactory(ui_task_runner_);
#endif
```

content/public/android/BUILD.gn
```diff
@@ -230,7 +230,10 @@ android_library("content_full_java") {
 
   # TODO(b/443992661): Add a GN flag to conditionally disable or enable shape detection.
   if(is_cobalt) {
```
andrewsavage1, line 232: nit: add a space `if (is_cobalt) {` to format this correctly

andrewsavage1 (approved):

borongc:
nit: `Bug: 531833882` should be in the last line.

media/capture/video/create_video_capture_device_factory.cc
```diff
@@ -81,23 +84,33 @@ CreatePlatformSpecificVideoCaptureDeviceFactory(
 #endif
 }
 
+#endif  // !BUILDFLAG(IS_COBALT)
+
```
borongc: nit: extra line

media/capture/video/create_video_capture_device_factory.cc
```diff
@@ -81,23 +84,33 @@ CreatePlatformSpecificVideoCaptureDeviceFactory(
...
 }  // anonymous namespace
 
 bool ShouldUseFakeVideoCaptureDeviceFactory() {
+#if BUILDFLAG(IS_COBALT)
+  return false;
+#else
```
borongc: nit: should be `#else  // BUILDFLAG(IS_COBALT)`. Same for `#endif` and all other places.

media/capture/video/create_video_capture_device_factory.cc
osagie98: Thanks, that's cleaner

services/video_capture/video_capture_service_impl.cc
```diff
@@ -13,7 +13,9 @@
 #include "base/task/thread_pool.h"
 #include "build/build_config.h"
 #include "gpu/ipc/client/client_shared_image_interface.h"
-#include "media/capture/video/create_video_capture_device_factory.h"
+#if !BUILDFLAG(IS_COBALT)
```
kjyoun: convention is locating if guarded inlude below main include block

can you move this?

https://screenshot.googleplex.com/3pLbdWyUUKLLLw6

services/video_capture/video_capture_service_impl.cc
```diff
@@ -316,8 +318,12 @@ void VideoCaptureServiceImpl::LazyInitializeDeviceFactory() {
   // The task runner passed to CreateFactory is used for things that need to
   // happen on a "UI thread equivalent", e.g. obtaining screen rotation on
   // Chrome OS.
+#if BUILDFLAG(IS_COBALT)
```
kjyoun, line 322: can you add comment about why we are setting nullptr, instead of calling factory method?

e.g.). // Cobalt does not use media capture?


and isn't this more concise? (though it is you call)

```
  std::unique_ptr<media::VideoCaptureDeviceFactory> media_device_factory = 
#if BUILDFLAG(IS_COBALT)
    nullptr;
#else
      media::CreateVideoCaptureDeviceFactory(ui_task_runner_);
#endif  // BUILDFLAG(IS_COBALT)

```

kjyoun (approved):

services/video_capture/video_capture_service_impl.cc
osagie98: That's a bit more concise, but chose the other option since I think it looks a bit cleaner, and I can leave the Chrome code how it is

PR #11225, opened 2026-07-06
Skip inapplicable check on disk space

yuying-y:
Issue: 525267707

### DIFF
--- a/components/update_client/op_download.cc
+++ b/components/update_client/op_download.cc
@@ -175,6 +175,11 @@ void HandleAvailableSpace(
 #endif
         callback,
     int64_t available_bytes) {
+#if !BUILDFLAG(IS_STARBOARD)
+  // Cobalt implementation doesn't use the temp dir for the download,
+  // and relies on UpdateChecker::SkipUpdate to handle this error case when
+  // available space is insufficient. It sends UpdateCheckError::OUT_OF_SPACE
+  // error to the server.
   if (available_bytes / 2 <= size) {
     VLOG(1) << "available_bytes: " << available_bytes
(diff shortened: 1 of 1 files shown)


andrewsavage1:
Aren't these separate checks? This seems to be reporting full disk space vs the check you added checks memory usage for the download to have enough RAM

yuying-y:
> Aren't these separate checks? This seems to be reporting full disk space vs the check you added checks memory usage for the download to have enough RAM

These are two different issues. See more in b/525267707#comment13.

components/update_client/op_download.cc
```diff
@@ -182,6 +182,13 @@ void HandleAvailableSpace(
 #endif
         callback,
     int64_t available_bytes) {
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 185: Why is it that we don't override the default get_available_space callback for the UpdateContext class to be applicable to Starboard platforms?

yuying-y: I'd like to minimize customization to the upstream code. The check has been handled in UpdateChecker::SkipUpdate() since 2020, and has been proven to work well. It's simple to just void the check for Starboard, and easier for future rebases as well.

andrewsavage1: I think it's better for handling future rebases to override the get_available_space callback, as it doesn't disable any upstream code, instead using a mechanism that Chromium already provides to configure this. I'll approve this because it's the same as in c25, but I do think it'd be a good opportunity to clean this up

andrewsavage1 (approved):

PR #11234, opened 2026-07-07
Remove devtools frontend

sacuff:
This disables the pre-existing `enable_devtools_frontend` buildflag on Cobalt builds and conditionally removes `devtools_resources.pak` from `//cobalt/shell:pak`, plus some other changes to ensure non-Android builds succeed. Remote devtools are still functional.

`enable_devtools_frontend` is already disabled upstream on Android, so there should be no impact there. For 3P builds, this reduces the size of `cobalt_shell.pak` from 5.3 MB to 2.0 MB.

Bug: 525402992
Bug: 523267920


chrome/browser/devtools/BUILD.gn
```diff
@@ -2,13 +2,13 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+import("//chrome/common/features.gni")
```
andrewsavage1: Why is this change necessary?

chrome/browser/devtools/BUILD.gn
```diff
@@ -367,6 +370,12 @@ static_library("devtools") {
         "protocol/window_manager_handler.h",
       ]
     }
+    if (enable_devtools_frontend) {
```
andrewsavage1: To make it clear this is a cobalt change, avoid changing upstream code, and minimize conditionals, we should have a combined block `if (is_cobalt && !enable_devtools_frontend) {` that removes all the sources here and above. It's generally best to add this block near the end of the target

content/browser/devtools/features.gni
```diff
@@ -6,4 +6,4 @@ import("//build/config/chrome_build.gni")
 
 # Android (non-desktop) and iOS don't support DevTools front-end.
 enable_devtools_frontend =
-    (!is_android || is_desktop_android) && !is_ios && !is_fuchsia
+    (!is_android || is_desktop_android) && !is_ios && !is_fuchsia && (!is_cobalt || cobalt_debugging_enabled)
```
andrewsavage1: This should use cobalt_is_release_build instead of cobalt_debugging_enabled

extensions/BUILD.gn
```diff
@@ -248,16 +249,19 @@ repack("shell_and_test_pak") {
...
       "$root_gen_dir/extensions/shell/app_shell_resources.pak",
     ]
 
     deps += [
-      "//content/browser/devtools:devtools_resources",
       "//device/bluetooth/strings",
       "//extensions/shell:resources",
     ]
+
+    if (enable_devtools_frontend) {
```
andrewsavage1: Should similarly be is_cobalt && !enable_devtools_frontend and remove sources/deps

andrewsavage1 (commented):
Overall looks really good, just some formatting suggestions

chrome/browser/devtools/BUILD.gn
sacuff: This is an upstream bug AFAICT. On line 310 (on main), there's a reference to `enable_service_discovery` that isn't guarded by `if (enable_devtools_frontend)`, even though `enable_service_discovery` is defined in this gni file.

chrome/browser/devtools/BUILD.gn
```diff
@@ -307,11 +307,13 @@ static_library("devtools") {
       "serialize_host_descriptions.cc",
       "serialize_host_descriptions.h",
     ]
-    if (enable_service_discovery) {
-      sources += [
-        "device/cast_device_provider.cc",
-        "device/cast_device_provider.h",
-      ]
+    if (enable_devtools_frontend) {
```
andrewsavage1: Similarly just include this in the removal block below with the enable_serivce_discovery subcondition added

andrewsavage1 (approved):
Looks great!

chrome/browser/devtools/BUILD.gn
sacuff: This is just to fix that `enable_service_discovery` is undefined, the sources part otherwise unrelated. I could also move the `//chrome/common/features.gni` import out of the `if` at the top

PR #11247, opened 2026-07-07
Add dual-attribution and leaf-sliced differential heap profile analyzer with unit tests

haozheng-cobalt:
Introduce a new performance tool to analyze memory footprint deltas
between heap snapshots. This script processes heaps_v2 JSON traces,
aggregates data by subsystem, and provides a differential view of
callstacks to identify the root cause of memory regressions.

By using dual-attribution and leaf-slicing, the tool filters noise and
pinpoints business logic changes that contribute to bloat. It includes
unit tests for the parsing and comparison logic.

The unittests file is not integrated in cobalt_unittests by purpose, it's more used for local development and should not go into CIs.

Example output reports: noted that the data in these reports are not actionable because C27 was launched with account logged out and C26 was launched with account logged in. But the reports demonstrate the formats and structure of the data provided in the trace files.
https://paste.googleplex.com/6230188533874688,
https://paste.googleplex.com/6034673502388224

Bug: 528397803

### DIFF
--- a/cobalt/tools/performance/memory/diff_in_process_heaps.py
+++ b/cobalt/tools/performance/memory/diff_in_process_heaps.py
@@ -1,0 +1,1211 @@
+#!/usr/bin/env python3
+# Copyright 2026 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


johnxwork (approved):
I think you want to add a section in cobalt/tools/performance/memory/README.md about how to use this script as well?

PR #11256, opened 2026-07-08
tvOS: Remove onApplicationMainThreadAsync()

rakuco:
This function has not been used anywhere since #8048.

Bug: 459507205


rakuco:

fayev-g (approved):

PR #11264, opened 2026-07-08
cobalt: Remove logging prefixes from Crashpad handlers

alunev:
cobalt: Remove logging prefixes from Crashpad handlers

Remove the 'Crash:' or 'Freeze detection:' prefixes from logging statements
in Crashpad handlers to reduce redundancy and simplify log messages

Issue: 517626586

### DIFF
--- a/components/crash/core/app/crashpad_android.cc
+++ b/components/crash/core/app/crashpad_android.cc
@@ -317,7 +317,7 @@ bool GetHandlerTrampoline(std::string* handler_trampoline,
   if (base::android::BuildInfo::GetInstance()->sdk_int() <
       base::android::SDK_VERSION_Q) {
 #if BUILDFLAG(IS_COBALT)
-    LOG(INFO) << "Freeze detection: SDK version below Q: No linker support.";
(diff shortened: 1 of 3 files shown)


components/crash/core/app/crashpad_android.cc
```diff
@@ -317,7 +317,7 @@ bool GetHandlerTrampoline(std::string* handler_trampoline,
   if (base::android::BuildInfo::GetInstance()->sdk_int() <
       base::android::SDK_VERSION_Q) {
 #if BUILDFLAG(IS_COBALT)
-    LOG(INFO) << "Freeze detection: SDK version below Q: No linker support.";
```
hlwarriner, line 320: In the case of ATV this code path is actually only used for freezes, and not for crashes, right? Maybe we should just remove the "Freeze detection:"  / "Crash:" prefix altogether from all these log statements?

alunev: sure, let's remove it.

hlwarriner (approved):

jellefoks (approved):

andrewsavage1 (approved):

PR #11280, opened 2026-07-09
Set Dolby Vision policy to 1 before running on RDK

niranjanyardi:
This PR adds a workaround to configure the Dolby Vision policy on Amlogic RDK devices before launching Cobalt.

### Changes:
- Adds `ensure_dolby_vision_policy` to unconditionally write `1` to `/sys/module/aml_media/parameters/dolby_vision_policy` before running Cobalt.
- Works in both **ADB** and **SSH** connection modes.
- Adds a TODO referencing bug 532753892 to remove this workaround once the device driver is updated.

Bug: 532753892

TAG=agy
CONV=e3d4c794-091a-4ae7-8cd1-810eebeb07ee

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/arm/scripts/deploy_rdk.py
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/arm/scripts/deploy_rdk.py
@@ -231,6 +231,13 @@ def package_and_deploy(
     Path(archive_name).unlink(missing_ok=True)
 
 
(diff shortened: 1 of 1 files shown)


jonastsai:
this is a platform specific setting. I think it is good to include it in the new image.

jonastsai (approved):

niranjanyardi:
> this is a platform specific setting. I think it is good to include it in the new image.

agreed , once its int he image and everyone has updated to use it , we can remove it

PR #11287, opened 2026-07-09
starboard: Enforce valid buffer indices

xiaomings:
Turn SB_DCHECK_GE to SB_CHECK_GE in MediaCodecDecoder when receiving input and output buffers from Android MediaCodec. This ensures that invalid negative indices are caught immediately in all builds.

Bug: 329686979
Bug: 435425692


kjyoun (approved):
can you add "Bug: 435425692" as well?

xiaomings:
> can you add "Bug: 435425692" as well?

Added

PR #11297, opened 2026-07-09
Add EGL 1.5 test to nplb

y4vor:
Bug: 508253341

### DIFF
--- a/starboard/nplb/egl_test.cc
+++ b/starboard/nplb/egl_test.cc
@@ -14,6 +14,11 @@
 
 #include "starboard/egl.h"
 
+#include <stdio.h>
+
+#include <string>
+
+#include "starboard/testing/fake_graphics_context_provider.h"
 #include "testing/gtest/include/gtest/gtest.h"
 
 namespace nplb {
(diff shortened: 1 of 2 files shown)


starboard/nplb/egl_test.cc
```diff
@@ -66,5 +69,33 @@ TEST(SbEglInterfaceTest, HasValidEglInterface) {
   EXPECT_NE(nullptr, egl_interface->eglGetCurrentContext);
 }
 
+// Verifies that the target device supports EGL 1.5 or higher graphics API
+TEST(SbEglTest, SupportsEgl15OrHigher) {
+  const SbEglInterface* const egl_interface = SbGetEglInterface();
+  if (!egl_interface) {
+    return;
```
andrewsavage1: Should this be considered a failure now?

andrewsavage1 (commented):
We should change https://github.com/youtube/cobalt/blob/main/starboard/egl.h#L28 too

starboard/nplb/egl_test.cc
y4vor: done

andrewsavage1 (approved):

niranjanyardi (approved):
fixed the maneki runner

PR #11304, opened 2026-07-09
Add chromium to staging autoroll.

briantting:
This not only adds functionality to support a second to staging autoroll, but also adds functionality to support a non-main source branch.

To handle various .gitmodules issues such as corruption or conflicts, the logic in resolve_conflicts was updated.

Due to the larger and more complex Chromium source cherry-picks, the checkouts can no longer be sparse / none blobs.

Bug: 522987128

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -4,10 +4,9 @@ import argparse
 from collections import defaultdict
 import enum
 import re
+import shutil
 import subprocess
 import sys
-
-_AUTOROLL_FILE = '.github/AUTOROLL'
 
 
 @enum.unique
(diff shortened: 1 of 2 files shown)


.github/workflows/autoroll_lts.yaml
```diff
@@ -103,6 +98,12 @@ jobs:
         run: |
           set -x
 
+          if [[ "${SOURCE_BRANCH}" == "main" ]]; then
```
linxinan-yt: If we have to handle multiple branches, I would prefer a dict rather than if..else.

linxinan-yt: ah, this is yaml not py. Sorry. 

But anything other than main would use autoroll_chromium? Is it better to have elif here?

.github/scripts/autoroll_lts.py
```diff
@@ -200,8 +239,14 @@ def main():
   p.add_argument('--max-commits', type=int, required=True)
   args = p.parse_args()
 
-  target_start = get_start_sha(args.target_branch)
-  autoroll_start = get_start_sha('HEAD')
+  if args.source_branch == 'main':
```
linxinan-yt: Can we define a static dict for the mapping?

linxinan-yt (approved):
LGTM % comments

.github/scripts/autoroll_lts.py
briantting: moved it into the yaml file as an argument when calling the py file

.github/workflows/autoroll_lts.yaml
briantting: used the matrix value instead

PR #11307, opened 2026-07-10
[Android] Restore StarboardBridge return type in CobaltActivity

johnxwork:


hferreiro:
Would it make sense to use the name BaseStarboardBridge instead of AOSPStarboardBridge?

hferreiro:
cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CobaltActivityTest.java and cobalt/testing/browser_tests/browsertests_apk/src/org/chromium/content_browsertests_apk/ContentShellBrowserTestActivity.java need to be updated too.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/AOSPStarboardBridge.java
```diff
@@ -0,0 +1,785 @@
...
+      return null;
+    }
+    final CobaltService.Factory factory = mCobaltServiceFactories.get(serviceName);
+    if (factory == null) {
+      Log.e(TAG, String.format("Cannot open unregistered service %s", serviceName));
+      return null;
+    }
+    CobaltService service = factory.createCobaltService(nativeService);
+    if (service != null) {
+      service.receiveStarboardBridge((StarboardBridge) this);
```
hferreiro, line 626: Remove cast and update `receiveStarboardBridge` signature.

johnxwork:
Switch to https://github.com/youtube/cobalt/pull/11318 which is on main

PR #11312, opened 2026-07-10
media: Split DecoderThreadFunc into Audio/Video variants

xiaomings:
Refactor the MediaCodecDecoder by splitting the monolithic
DecoderThreadFunc into dedicated audio and video variants.

This is the initial phase of a broader effort to simplify threading
handling. Breaking the logic into media-specific functions improves
readability and prepares the codebase for further targeted
optimizations and cleanups.

Bug: 329686979

### DIFF
--- a/starboard/android/shared/media_codec_decoder.cc
+++ b/starboard/android/shared/media_codec_decoder.cc
@@ -338,10 +338,14 @@ void MediaCodecDecoder::WriteInputBuffers(const InputBuffers& input_buffers) {
       video_output_thread_->Start();
     }
   } else if (!decoder_thread_) {
-    decoder_thread_ = std::make_unique<DecoderThread>(
-        GetDecoderName(media_type_), [this] { DecoderThreadFunc(); },
-        media_type_ == kSbMediaTypeAudio ? ThreadPriority::kNormal
-                                         : ThreadPriority::kHigh);
+    decoder_thread_ =
+        (media_type_ == kSbMediaTypeAudio)
+            ? std::make_unique<DecoderThread>(
+                  GetDecoderName(media_type_),
+                  [this] { AudioDecoderThreadFunc(); }, ThreadPriority::kNormal)
+            : std::make_unique<DecoderThread>(
+                  GetDecoderName(media_type_),
+                  [this] { VideoDecoderThreadFunc(); }, ThreadPriority::kHigh);
     decoder_thread_->Start();
   }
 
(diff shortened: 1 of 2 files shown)


starboard/android/shared/media_codec_decoder.cc
```diff
@@ -438,141 +442,147 @@ bool MediaCodecDecoder::Flush() {
...
+  // While it is possible to consolidate the logic for audio and video
+  // decoders, it is easy to fine tune the behavior of video decoder if they
+  // are separated.
+  std::deque<PendingInput> pending_inputs;
+  std::vector<int> input_buffer_indices;
+  std::vector<DequeueOutputResult> dequeue_output_results;
+
+  while (!destroying_.load()) {
+    // TODO(b/329686979): access to `ending_input_to_retry_` should be
+    //                    synchronized.
```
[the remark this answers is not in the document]
xiaomings: Done, while it's not in the scope of this PR.

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -438,141 +442,147 @@ bool MediaCodecDecoder::Flush() {
...
-        std::unique_lock lock(mutex_);
-        CollectPendingData_Locked(&pending_inputs, &input_buffer_indices,
-                                  &dequeue_output_results);
-        can_process_input =
-            !pending_inputs.empty() && !input_buffer_indices.empty();
-        if (!can_process_input && dequeue_output_results.empty()) {
-          condition_variable_.wait_for(
-              lock, std::chrono::microseconds(video_decoder_poll_interval_us_));
-        }
+          !pending_inputs.empty() && !input_buffer_indices.empty();
```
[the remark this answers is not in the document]
xiaomings: It's not in the scope of this PR.

xiaomings:
@youtube/cobalt-media note to reviewers, it's much easier to review this PR when checking out locally, or in gerrit (376860) with white space ignored.

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -437,147 +441,152 @@ bool MediaCodecDecoder::Flush() {
...
+      bool has_input_buffer_indices =
+          !input_buffer_indices.empty() || !input_buffer_indices_.empty();
+      bool can_process_input = pending_input_to_retry_ ||
+                               (has_pending_input && has_input_buffer_indices);
+      if (dequeue_output_results_.empty() && !can_process_input) {
+        // Wait for a signal or a timeout. We don't use a predicate here
+        // because the complex conditions are already checked by the
+        // surrounding loop, which will re-evaluate the state when this wait
+        // returns.
+        if (condition_variable_.wait_for(lock, std::chrono::seconds(5)) ==
```
kjyoun, line 468: this formating looks weird

can you double check with clang-formatter?

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -437,147 +441,152 @@ bool MediaCodecDecoder::Flush() {
...
-      bool has_input =
-          pending_input_to_retry_ ||
-          (!pending_inputs.empty() && !input_buffer_indices.empty());
-      bool has_output = !dequeue_output_results.empty();
-      bool collect_pending_data = false;
 
-      if (tunnel_mode_enabled_) {
-        // We don't explicitly process output in tunnel mode.
-        collect_pending_data = !has_input;
+    for (auto dequeue_output_result : dequeue_output_results) {
```
kjyoun: const auto& ?

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -437,147 +441,152 @@ bool MediaCodecDecoder::Flush() {
...
-      bool can_process_input =
-          pending_input_to_retry_ ||
-          (!pending_inputs.empty() && !input_buffer_indices.empty());
-      if (can_process_input) {
-        ProcessOneInputBuffer(&pending_inputs, &input_buffer_indices);
-      }
+void MediaCodecDecoder::VideoDecoderThreadFunc() {
+  // Initialize() should be called before creating the thread, where `error_cb_`
+  // is set.  Check `error_cb_` here to ensure Initialize() has been called.
+  SB_DCHECK(error_cb_);
```
kjyoun, line 510: do you need this ?

I cannot find accessing of error_cb_ in this method.

If there is, we should use SB_CHECK, since it provides better crash stack, pinpointing where/why crash happens.

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -437,147 +441,152 @@ bool MediaCodecDecoder::Flush() {
...
+      ProcessOneInputBuffer(&pending_inputs, &input_buffer_indices);
+    }
+
+    bool ticked = false;
+    if (!tunnel_mode_enabled_) {
+      // Output is only processed when tunnel mode is disabled.
+      ticked = host_->Tick(media_codec_bridge_.get());
+    }
 
+    can_process_input =
```
kjyoun, line 572: similar logic is repeated. can you make it into the one helper method?

It would help code readability a lot.

```
 bool has_input = pending_input_to_retry_ ||
                     (!pending_inputs.empty() && !input_buffer_indices.empty());

bool can_process_input =
        pending_input_to_retry_ ||
        (!pending_inputs.empty() && !input_buffer_indices.empty());

 can_process_input =
        pending_input_to_retry_ ||
        (!pending_inputs.empty() && !input_buffer_indices.empty());
...


How about adding HasPendingInpu() helper method ?

File: starboard/android/shared/media_codec_decoder.cc [hunk 7]
```diff
@@ -437,147 +441,152 @@ bool MediaCodecDecoder::Flush() {
...
-          pending_input_to_retry_ ||
-          (!pending_inputs.empty() && !input_buffer_indices.empty());
-      if (can_process_input) {
-        ProcessOneInputBuffer(&pending_inputs, &input_buffer_indices);
-      }
+void MediaCodecDecoder::VideoDecoderThreadFunc() {
+  // Initialize() should be called before creating the thread, where `error_cb_`
+  // is set.  Check `error_cb_` here to ensure Initialize() has been called.
+  SB_DCHECK(error_cb_);
+  SB_DCHECK_EQ(media_type_, kSbMediaTypeVideo);
```
kjyoun, line 511: SB_CHECK_EQ ?

this performance overhead for this SB_CHECK is trivial.

https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++.md#check_dcheck_and-notreached

> Historically, Chromium code used DCHECK() in most cases, so a great deal of existing code uses DCHECK() instead of CHECK(). You are encouraged to migrate to CHECK()

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -437,147 +441,152 @@ bool MediaCodecDecoder::Flush() {
...
+
+  while (!destroying_.load()) {
+    // TODO(b/329686979): access to `pending_input_to_retry_` should be
+    //                    synchronized.
+    bool has_input = pending_input_to_retry_ ||
+                     (!pending_inputs.empty() && !input_buffer_indices.empty());
+    bool has_output = !dequeue_output_results.empty();
+    bool collect_pending_data = false;
+
+    if (tunnel_mode_enabled_) {
```
kjyoun, line 528: ?

```
collect_pending_data = !has_input || (!tunnel_mode_enabled_ && !has_output);
```

starboard/android/shared/media_codec_decoder.cc
```diff
@@ -339,10 +338,15 @@ void MediaCodecDecoder::WriteInputBuffers(const InputBuffers& input_buffers) {
       video_output_thread_->Start();
     }
   } else if (!decoder_thread_) {
-    decoder_thread_ = std::make_unique<DecoderThread>(
-        GetDecoderName(media_type_), [this] { DecoderThreadFunc(); },
-        media_type_ == kSbMediaTypeAudio ? ThreadPriority::kNormal
-                                         : ThreadPriority::kHigh);
+    if (media_type_ == kSbMediaTypeAudio) {
```
kjyoun: how about this?

Code becomes more concise and help difference between those 2 thread more visiible

```
decoder_thread_ = (media_type_ == kSbMediaTypeAudio)
    ? std::make_unique<DecoderThread>(
          GetDecoderName(media_type_), [this] { AudioDecoderThreadFunc(); },
          ThreadPriority::kNormal)
    : std::make_unique<DecoderThread>(
          GetDecoderName(media_type_), [this] { VideoDecoderThreadFunc(); },
          ThreadPriority::kHigh);
```

linxinan-yt:
Hi, please set your membership public, otherwise the PR will be blocked by tvOS build.

starboard/android/shared/media_codec_decoder.cc
xiaomings: Tried to reformat and it stays the same.

starboard/android/shared/media_codec_decoder.cc
xiaomings: Done

starboard/android/shared/media_codec_decoder.cc
xiaomings: It's used as a way to check whether Initialize() is called, and is copied from the original function as is.

This isn't probably the most explicit way, but refining it is beyond the scope of this PR.

If this fails, it's more of a logic error instead of an output from a platform function, so I'd keep it as DCHECK().

starboard/android/shared/media_codec_decoder.cc
xiaomings: This is not in the scope of this PR.

On the other hand, I would keep using DCHECK() for logic errors when appropriate, and will treat this as a stronger comment for "Initialize() has to be called before calling this function".  For example, if I have 10 parameters to a function, and I want to sanity check whether they are valid.  I would use DCHECK() arbitrarily as there isn't any runtime cost in production.  I will be slightly more careful to put CHECK().  The same for verifying something in a loop.

starboard/android/shared/media_codec_decoder.cc
xiaomings: I would keep the existing logic as is, as the next step of the refactor is to refactor the tunnel mode handling.  While it's pretty straightforward to roll the if statement into a one line statement and later unroll it, it's safer to just keep this as is.

starboard/android/shared/media_codec_decoder.cc
xiaomings: I am trying to avoid introducing any non-trivial changes into this PR.  If this should be done, we could do this in a separate PR.

starboard/android/shared/media_codec_decoder.cc
xiaomings: Done

kjyoun (approved):

PR #11315, opened 2026-07-10
android: Remove dead cobalt_shell_test_apk targets

hferreiro:
cobalt_shell_test_apk cannot build: its java library lists javatests/src/org/chromium/content_shell_apk/*.java sources and an IChildProcessTest.aidl that were never brought over from content/shell/android.

Remove it together with the targets only it used.

Bug: 492403745


Awallky (approved):
It looks like the browsertests build and I ran them locally. They seem to still pass, minus a few (unrelated) failing tests.
LGTM

PR #11316, opened 2026-07-10
drm: Fix crash due to empty session_id access

santoshmahto7:
drm: Fix crash on empty session ID access

The StarboardCdm class was experiencing a crash due to a bad optional
access in OnSessionUpdateRequestGenerated. This occurs when the
underlying media session fails to generate a request, resulting in an
empty session ID being passed to the callback.

This was observed on RDK, while accessing videos on channel
https://www.youtube.com/@bodedrmchannel2958

Crash reported:
```
   std::__Cr::__libcpp_verbose_abort() [0xf38898ad]
   std::__Cr::__throw_bad_optional_access() [0xf017b16f]
   media::StarboardCdm::OnSessionUpdateRequestGenerated() [0xf05f69e1]
   base::internal::DecayedFunctorTraits<>::Invoke<>() [0xf05f741f]
   base::internal::Invoker<>::RunOnce() [0xf05f7381]
```

```
WPEProcess[4798]: [OCDM][Widevine]generateRequest failed
WPEProcess[4798]: [OCDM][Widevine]exists? cert.bin: false
WPEProcess[4798]: [OCDM][Widevine]generateRequest failed
WPEFramework[4798]: [ERROR:../../../cdm/src/cdm.cpp(676):generateRequest] No such session:
WPEProcess[4798]: [OCDM][Widevine]exists? cert.bin: false
WPEProcess[4798]: [OCDM][Widevine]generateRequest failed
WPEFramework[4798]: [ERROR:../../../cdm/src/cdm.cpp(676):generateRequest] No such session:
WPEProcess[4798]: [OCDM][Widevine]exists? cert.bin: false`
```

Bug: 532716116
Change-Id: I8a8e836e001d6038f3d901a4821fb93f1b761b71

### DIFF
--- a/media/starboard/starboard_cdm.cc
+++ b/media/starboard/starboard_cdm.cc
@@ -307,31 +307,30 @@ void StarboardCdm::OnSessionUpdateRequestGenerated(
     // Called back as a result of |SbDrmGenerateSessionUpdateRequest|.
 
     // Restore the context of |GenerateSessionUpdateRequest|.
-    auto session_update_request_iterator =
+    auto session_update_request_it =
         ticket_to_session_update_request_map_.find(ticket);
-    if (session_update_request_iterator ==
+    if (session_update_request_it ==
         ticket_to_session_update_request_map_.end()) {
       LOG(INFO) << "Unknown session update request ticket: " << ticket << ".";
       return;
     }
 
-    auto session_request = std::move(session_update_request_iterator->second);
-
-    if (session_id) {
-      session_list_.push_back(session_id.value());
-      promises_.ResolvePromise(session_request.promise_id, session_id.value());
-    } else {
+    auto session_request = std::move(session_update_request_it->second);
+    ticket_to_session_update_request_map_.erase(session_update_request_it);
+    if (!session_id) {
       // Failure during request generation.
-      LOG(INFO) << "Calling session update request callback on drm system ("
-                << sb_drm_ << "), status: " << status
-                << ", error message: " << error_message;
+      LOG(ERROR)
+          << "Failure during session update request generation on drm system ("
+          << sb_drm_ << "), status: " << status
+          << ", error message: " << error_message;
       promises_.RejectPromise(session_request.promise_id,
                               CdmPromise::Exception::INVALID_STATE_ERROR, 0,
                               "Failure during request generation.");
+      return;
     }
 
-    ticket_to_session_update_request_map_.erase(
-        session_update_request_iterator);
+    session_list_.push_back(session_id.value());
+    promises_.ResolvePromise(session_request.promise_id, session_id.value());
 
   } else {
     // Called back spontaneously by the underlying DRM system.


media/starboard/starboard_cdm.cc
```diff
@@ -344,6 +344,10 @@ void StarboardCdm::OnSessionUpdateRequestGenerated(
     }
   }
 
+  if (!session_id) {
```
kjyoun: This logic has been stable across multiple Cobalt revisions. 

Any thoughts on why it's crashing in your specific environment?

kjyoun: Apparently , we didn't see this crash previously because passing  `SbDrmTicketIsValid(ticket)`  previously always had  `session_id`  populated so that we never use code path for `!session_id`

```
    if (session_id) {
      // We have alway used this code path
      session_list_.push_back(session_id.value());
      promises_.ResolvePromise(session_request.promise_id, session_id.value());
    } else {

     //  Not this path

      // Failure during request generation.
      LOG(INFO) << "Calling session update request callback on drm system ("
                << sb_drm_ << "), status: " << status
                << ", error message: " << error_message;
      promises_.RejectPromise(session_request.promise_id,
                              CdmPromise::Exception::INVALID_STATE_ERROR, 0,
                              "Failure during request generation.");
    }
```

I think the right fix is returning immediately after rejecting the promise in the error branch:

Can you replace code block from 320-334 with this logic? (instead of silent return)

```
        if (!session_id) {
          // Failure during request generation.
          LOG(ERROR) << "Failure during session update request generation on drm system ("
                     << sb_drm_ << "), status: " << status
                     << ", error message: " << error_message;
          promises_.RejectPromise(session_request.promise_id,
                                  CdmPromise::Exception::INVALID_STATE_ERROR, 0,
                                  "Failure during request generation.");
          ticket_to_session_update_request_map_.erase(
              session_update_request_iterator);
          return;
        }

        session_list_.push_back(session_id.value());
        promises_.ResolvePromise(session_request.promise_id, session_id.value());
        ticket_to_session_update_request_map_.erase(
            session_update_request_iterator);
```

kjyoun:
nit: can you add "drm: " prefix in your commit title per convention?

e.g., drm: Fix crash due to empty session_id access

media/starboard/starboard_cdm.cc
santoshmahto7: make sense. I will update

santoshmahto7: > This logic has been stable across multiple Cobalt revisions.
> 
> Any thoughts on why it's crashing in your specific environment?

The DRM video that triggered this appeared unplayable. I think the server configured for this DRM video was incomplete or misconfigured. So the session construction failed. 
The video was at https://www.youtube.com/@bodedrmchannel2958, but on desktop it doesn't show any videos. On TV, it has two videos.

On error, RDK sets the `session_id` to nullptr.
```
// rdk/shared/drm/drm_system_ocdm.cc
  if (opencdm_construct_session(
          ocdm_system_, Temporary, type.c_str(),
          reinterpret_cast<const uint8_t*>(initialization_data),
          initialization_data_size, nullptr, 0, &session_callbacks_, this,
          &session) != ERROR_NONE ||
      !session) {
    session_update_request_callback_(drm_system_, context_, ticket,
                                     kSbDrmStatusUnknownError,
                                     kSbDrmSessionRequestTypeLicenseRequest,
                                     nullptr,  -----------------------------// santo: session_id ptr
                                     nullptr, 0, nullptr, 0, nullptr);
    return;                        ^                                                                                                          
  }

```
I think session_id = nullptr is a valid error case, since StarboardCdm also checks for session_id == nullptr here.
```
// StarboardCdm
cdm->task_runner_->PostTask(
      FROM_HERE,
      base::BindOnce(
          &StarboardCdm::OnSessionUpdateRequestGenerated,                                                                                    
          cdm->weak_factory_.GetWeakPtr(), 
          session_id == nullptr
              ? SessionTicketAndOptionalId(ticket)
              : SessionTicketAndOptionalId(
```

santoshmahto7: culprit videos
drm_crah_videos

kjyoun: I cannot find the video. 

can you give video id ? so that I can test

you have to able to see it on the log

e.g.) 
```
INFO:cobalt/shell/browser/shell.cc:585] Navigated to https://www.youtube.com/tv#/watch?v=cCi6swZVVS4
```

santoshmahto7: This is the URL: https://www.youtube.com/tv?rloader=wpl&expflag=watch_page_version:legacy#/browse?c=VLOLAK5uy_mKAu6VNK3gMSq_L8fU_C6myQnuuuIzvWY

But something seems to be fixed; videos in this URL work fine now without this PR.  So now we can only reproduce the issue with videos whose license server is not working
 This PR is still required to fix the crash when there is an issue on the license server side and session creation fails

kjyoun: LGTM — I couldn't find a repro video, but this clearly fixes an obvious crash path where execution falls through without checking

I made minor cleanup, removing duplicated code and making variable a little bit concise

https://github.com/youtube/cobalt/pull/11372/changes/601708ff4ea22dd938cdb27f0023be47a84839ff

WDYT? Could you include this change as well?

santoshmahto7: Sure, will add those changes

kjyoun (approved):

PR #11321, opened 2026-07-10
fix: Preserve siginfo when Crashpad re-raises sig

hlwarriner:
We still wan to avoid building this code path when building libcobalt (with the default toolchain) in hermetic builds, as it introduces a compilation error. But we should build it when building the loader app with the starboard toolchain, so that signals are properly re-raised, with siginfo information, after they are handled by Crashpad. This information may be useful for other signal handlers on the system, for example for b/481094002.

Issue: 476394156

### DIFF
--- a/third_party/crashpad/crashpad/util/posix/signals.cc
+++ b/third_party/crashpad/crashpad/util/posix/signals.cc
@@ -321,12 +321,13 @@ void Signals::RestoreHandlerAndReraiseSignalOnReturn(
 #if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_CHROMEOS)
 
   // TODO: b/406511608 - Cobalt: Re-enable if we decide to install Crashpad
-  // signal handlers from the Cobalt layer in hermetic builds.
-#if BUILDFLAG(IS_LINUX) && BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) && \
-    !BUILDFLAG(IS_STARBOARD_TOOLCHAIN)
-  // Falls through to the raise() logic below, avoiding compiler errors
-  // resulting from the syscall.
-#else
+  // signal handlers from the Cobalt layer in hermetic builds. For now in
+  // hermetic builds, we properly re-raise signals with siginfo in Starboard
+  // toolchain-built loader app, from which the Crashpad signal handlers are
+  // actually installed, and bypass the syscall in the Cobalt layer where it
+  // 1) is not needed and 2) introduces compiler errors.
+#if !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) || \
+    BUILDFLAG(IS_STARBOARD_TOOLCHAIN)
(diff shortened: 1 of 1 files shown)


hlwarriner:
I verified the difference before and after the fix by sending `kill -s SIGSEGV <loader app pid>`, using `strace`.

Before the fix, the signal is re-raised like:
```
tgkill(<loader app pid>, <loader app pid>, SIGSEGV) = 0
```

After the fix:
```
rt_tgsigqueueinfo(<loader app pid>, <loader app pid>, SIGSEGV, {si_signo=SIGSEGV, si_code=SI_USER, si_pid=<sender pid>, si_uid=<sender uid>}) = 0
```

third_party/crashpad/crashpad/util/posix/signals.cc
```diff
@@ -318,8 +318,12 @@ void Signals::RestoreHandlerAndReraiseSignalOnReturn(
   // signals that do not re-raise autonomously), such as signals delivered via
   // kill() and asynchronous hardware faults such as SEGV_MTEAERR, which would
   // otherwise be lost when re-raising the signal via raise().
-// TODO: (cobalt b/406511608) Re-enable once we use crashpad in hermetic builds.
-#if BUILDFLAG(IS_LINUX) && !BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) || BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_CHROMEOS)
+// TODO: (cobalt b/406511608) Re-enable if we decide to install Crashpad signal
+// handlers from the Cobalt layer in hermetic builds.
+#if BUILDFLAG(IS_LINUX) &&                       \
```
andrewsavage1: nit: let's separate the cobalt-added conditionals from the upstream ones to make it clearer what's ours and what's upstream and make rebasing easier in the future

andrewsavage1 (approved):

third_party/crashpad/crashpad/util/posix/signals.cc
hlwarriner: Sure - done

third_party/crashpad/crashpad/util/posix/signals.cc
```diff
@@ -337,8 +341,10 @@ void Signals::RestoreHandlerAndReraiseSignalOnReturn(
   if (errno != EPERM) {
     _exit(kFailureExitCode);
   }
-#endif  // BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_ANDROID) ||
-        // BUILDFLAG(IS_CHROMEOS)
+#endif  // BUILDFLAG(IS_LINUX) &&
+        // (!BUILDFLAG(ENABLE_COBALT_HERMETIC_HACKS) ||
+        //  BUILDFLAG(IS_STARBOARD_TOOLCHAIN)) ||
+        // BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_CHROMEOS)
```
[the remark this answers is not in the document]
hlwarriner: I think it reads fine either way

y4vor (approved):

andrewsavage1 (approved):

PR #11325, opened 2026-07-10
Implement dynamic refactoring-immune class aggregation and timeline verification in diff_in_process_heaps.py.

haozheng-cobalt:
1. Dynamic Refactoring-Immune Class Aggregation (core_class_key & clean_method_signature) Helps determine if signature fluctuations between trace files are driven by internal C++ refactorings rather than real memory bloat, such as:

Parameter Type Migrations (base::span): blink::XMLHttpRequest::DidReceiveData(const char*, int) in Cobalt 26 being refactored to DidReceiveData(base::span<const char>) in Cobalt 27.
Method Overload / Flag Shifts: blink::ArrayBufferContents::AllocateMemory vs. AllocateMemoryWithFlags.
Template & Return Type Decaying: const char* WTF::GetStringWithTypeName() [T = WTF::StringImpl*] vs. decayed PartitionAlloc container aliases.
Without class aggregation, these refactorings appear in granular callstack tables as fake 0 B -> +896 KB bloat right alongside an artificial +896 KB -> 0 B drop. The dynamic algorithmic class aggregator (core_class_key) collapses all method overloads into exact class identities (such as blink::XMLHttpRequest or WTF::StringImpl), verifying whether the net class footprint actually grew ( Net Growth) or remained unchanged (Stable / 0 B net diff), automatically suppressing false-positive callstacks.

2. Interactive Timeline Verification (--timeline KEYWORD / run_timeline_diagnosis) Helps determine if an X MB -> 0 B (or vice versa) difference in the report is a true architectural memory footprint change or a transient/fake workload fluctuation during trace capture, such as:

Transient Workload Fluctuation (media::DecoderBuffer -3.88 MB drop in report_1.md): Rather than assuming Cobalt 27 optimized away ~3.88 MB of video memory, --timeline DecoderBuffer proved that right when Snapshot 4 fired in Cobalt 26 (ts=11101s), the demuxer had +3.88 MB of queued video stream buffers, whereas when Snapshot 4 fired in Cobalt 27 (ts=11233s), the media player was idle between chunks (128 KB).
Transient Compilation Bursts (v8::Zone -5.88 MB drop in report_2.md): Timeline inspection revealed that at ts=18s, Cobalt 26 entered a temporary JavaScript compilation burst (+5.88 MB V8 compiler arena), pushing total heap to 45 MB, while Cobalt 27 had already completed script execution (36.12 MB).
True Architectural Bloat (std::__Cr::allocator / CompositorFrameSink +768 KB): Conversely, timeline verification proved this is real architectural growth, as +768 KB appeared consistently across every single steady-state snapshot of Cobalt 27 (ts=11s through ts=18s) while allocating 0 B in Cobalt 26.

Test: view the results here: https://docs.google.com/document/d/1RENchfN7p40bILlc37SpFU4zNuY59EQ3JZOTXUBkj1Y/edit?resourcekey=0-_bHyZMoWcku9SA3RMztHuw&tab=t.0
Bug: 528397803


cobalt/tools/performance/memory/diff_in_process_heaps.py
```diff
@@ -34,6 +34,25 @@
 logger = logging.getLogger("diff_in_process_heaps")
 
 
+def clean_method_signature(symbol: str) -> str:
```
Awallky, line 37: Docstring could improve by including `Args:` and `Returns:`

cobalt/tools/performance/memory/diff_in_process_heaps.py
```diff
@@ -1074,6 +1335,121 @@ def print_console_summary(markdown_report: str, output_path: str) -> None:
...
 
 
+def run_timeline_diagnosis(base_path: Optional[str], target_path: str,
+                           keyword: str) -> int:
+  """Runs multi-snapshot timeline diagnosis across trace files for a keyword."""
+  print("\n" + "=" * 80)
+  print(f"🔍 COBALT HEAP TIMELINE DIAGNOSIS FOR KEYWORD: [{keyword}]")
+  print("=" * 80)
+
+  def _inspect_trace(path: str, label: str) -> Tuple[List[float], int]:
```
Awallky: Missing a docstring here

Awallky: Also, this nested function makes the parent function more difficult to parse.
Can we move this outside?

Awallky: Refactoring it to something like  `def _extract_timeline_events(path: str, label: str, keyword: str) -> Tuple[List[float], int]:` at the top level of the file would:

  1. Drastically reduce the indentation depth.
  2. Allow  run_timeline_diagnosis  to look like a clean, high-level controller.
  3. Allow the trace-parsing logic to be unit-tested individually in the future.

Awallky (changes requested):
With the complexity of this logic here, can we add unit testing and refactor some functionality to be more easily tested? 
Even if it does not run in CI, I think having some key test cases would make this easier to maintain.

haozheng-cobalt:
> With the complexity of this logic here, can we add unit testing and refactor some functionality to be more easily tested? Even if it does not run in CI, I think having some key test cases would make this easier to maintain.

Thanks for the reminder, added the missing unit tests.

Awallky (approved):
LGTM

PR #11328, opened 2026-07-11
Adds chromium_cherry_pick function to autoroller.

briantting:
When autorolling from Chromium branches, the chromium_cherry_pick function is called which reverts all Cobalt changes to cleanly apply the Chromium cherry-pick before reapplying all Cobalt changes. This improves readability of direct Chromium cherry-pick conflicts.

Bug: 522987128

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -10,12 +10,12 @@ import sys
 
 
 @enum.unique
-class CherryPickStatus(enum.Enum):
-  """Represents the outcome of a cherry-pick attempt."""
-  SUCCESS = 'success'  # Successfully cherry-picked and committed.
-  CONFLICTED = 'conflicted'  # Cherry-picked and committed with conflicts.
+class CommitStatus(enum.Enum):
+  """Represents the outcome of a commit attempt."""
+  SUCCESS = 'success'  # Successfully committed.
+  CONFLICTED = 'conflicted'  # Committed with conflicts.
   SKIPPED = 'skipped'  # The commit was already present or no action was needed.
-  FAILED = 'failed'  # The cherry-pick failed due to conflicts or other errors.
+  FAILED = 'failed'  # The commit failed due to conflicts or other errors.
 
 
 def log(msg):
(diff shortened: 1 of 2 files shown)


.github/scripts/autoroll_lts.py
```diff
@@ -175,58 +175,97 @@ def get_cherry_pick_metadata(sha, title, pr_num):
...
-  """Attempts to cherry-pick a single commit.
+def get_submodule_root_dirs():
+  paths = get_out(
+      ['git', 'config', '--file', '.gitmodules', '--get-regexp', 'path'])
+
+  return sorted(
+      {line.split(' ', 1)[1].split('/')[0] for line in paths.splitlines()})
+
+
+def chromium_cherry_pick(previous_sha, sha, metadata, first_commit,
```
linxinan-yt, line 186: You may consider in follow up PRs. Along with chromium_cherry_pick, we may have the same method for third_party repos, like v8 and skia.
Because the current chromium delta (created on chromium/main branch) use submodule for v8 and skia etc. We cannot apply them directly to staging branch to get the meaningful conflicts(you either get a submodule string or keep the M138 skia in the codebase).
But we can enrich the script here to translate the submodules into code changes.

1. get chromium patchset but skipping submodule code
2. get submodule patchset via DEPS, e.g. v8 v.s. v8 fetched by DEPS
3. apply Chromium delta 
4. gclient sync to get the latest code managed by submodule
5. re-apply cobalt patchset from step 1 (third party back to direct code from submodule)
6. re-apply cobalt's patchsets(iterate for all submodules) from step 2

briantting: handling deps is definitely a needed follow up that will require its own PR

.github/scripts/autoroll_lts.py
```diff
@@ -175,58 +175,100 @@ def get_cherry_pick_metadata(sha, title, pr_num):
...
+                         autoroll_file):
+  """Temporarily reverts Cobalt before applying a Chromium cherry pick."""
```
linxinan-yt: Docstring to explain the input args.

.github/scripts/autoroll_lts.py
```diff
@@ -175,58 +175,100 @@ def get_cherry_pick_metadata(sha, title, pr_num):
...
+  log('Deleting everything...')
```
linxinan-yt, line 207: Revert Cobalt changes...

.github/scripts/autoroll_lts.py
```diff
@@ -175,58 +175,100 @@ def get_cherry_pick_metadata(sha, title, pr_num):
...
+  run(['git', 'clean', '-ffdx'])
+
+  log(f'Checking out clean Chromium state: {previous_sha}')
+  run(['git', 'checkout', previous_sha, '--', '.'])
+
+  log('Committing Cobalt revert...')
+  run(['git', 'add', '--', '.'])
+  run([
+      'git', 'commit', '--no-verify', '-m',
+      'CONFLICTED Chromium Cherry pick: Reverting Cobalt.'
```
linxinan-yt, line 221: CONFLICTED? Reverting cobalt does not generate conflicts, right?

.github/scripts/autoroll_lts.py
linxinan-yt: Also please update this line.

> Temporarily reverts Cobalt before applying a Chromium cherry pick.

This is not true. In this function you have

- reverted cobalt changes
- applied chromium changes
- relanded cobalt changes.

.github/scripts/autoroll_lts.py
briantting: That particular commit is not conflicted, but the overall whole process is. The process being "CONFLICTED Chromium Cherry pick". The logic (described in the second commit of this PR) relies on the first commit of each PR to start with CONFLICTED.

.github/scripts/autoroll_lts.py
briantting: That would be the log on line 199. This log describes the step of `rm -rf roots` and `git rm -rf .`

.github/scripts/autoroll_lts.py
briantting: Will add a docstring, but for the line why is it not true?

Would `Temporarily reverts Cobalt changes before applying a Chromium cherry pick.` be more correct?

.github/scripts/autoroll_lts.py
linxinan-yt: "everything" is too general. You can be more specific like "Clean up submodules" etc.

linxinan-yt (approved):

PR #11332, opened 2026-07-12
aosp: report POSIX symlink st_size from lstat

rogerzanoni:
POSIX requires a symlink's st_size to be the length of its target path. Android's /data filesystem reports a different value.

Fixes the following issues in tests
  posix_lstat_test.cc:97 EXPECT_EQ(sb.st_size, target_filename.length())

Bug: 532068409

### DIFF
--- a/starboard/android/shared/posix_emu/stat.cc
+++ b/starboard/android/shared/posix_emu/stat.cc
@@ -15,7 +15,9 @@
 #include <android/asset_manager.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <sys/stat.h>
+#include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


sacuff (approved):

andrewsavage1 (changes requested):
We should only change the posix wrappers to have it conform with posix specifications, not use it to support platform-specific quirks. Unless it's not possible, this change should be implemented as an android-specific wrapper, similar to the asset management with stat/open

rogerzanoni:
@andrewsavage1 moved the fix to the posix_emu wrapper

andrewsavage1 (approved):

andrewsavage1:
@rogerzanoni you need to fix the lint checks here

PR #11333, opened 2026-07-12
aosp: wrap mkostemp

rogerzanoni:
Bionic's implementation of these functions differs a bit from musl.

- mkostemp: bionic rejects a flags argument that includes the access mode or
  O_CREAT/O_EXCL with EINVAL, which musl callers pass.

Fixes the following nplb tests:
  - FileTempTest.MkostempBasicSuccess
  - FileTempTest.MkostempWithFlags
  - FileTempTest.MkostempNonExistentPath
  - FileTempTest.MkostempWriteFile

Bug: 53206840

### DIFF
--- a/starboard/android/shared/posix_emu/stdlib.cc
+++ b/starboard/android/shared/posix_emu/stdlib.cc
@@ -1,0 +1,60 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdint.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+///////////////////////////////////////////////////////////////////////////////
+// Implementations below exposed externally in pure C for emulation.
+///////////////////////////////////////////////////////////////////////////////
+
+extern "C" {
+
+// Reimplementation of musl's __randname.
+static void randname(char* tmpl) {
+  struct timespec ts;
+  clock_gettime(CLOCK_REALTIME, &ts);
+  unsigned long r =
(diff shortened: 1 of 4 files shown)


andrewsavage1 (changes requested):
These should be added as android-specific wrappers/functions similar to open/stat

rogerzanoni:
Moved to posix_emu wrappers.

starboard/android/shared/posix_emu/unistd.cc
```diff
@@ -0,0 +1,31 @@
...
+///////////////////////////////////////////////////////////////////////////////
+// Implementations below exposed externally in pure C for emulation.
+///////////////////////////////////////////////////////////////////////////////
+
+extern "C" {
+
+// POSIX leaves it implementation-defined whether link() follows a symbolic
+// link named by path1. Route through linkat() with a 0 flag (no-follow),
+// matching bionic's link().
+int __wrap_link(const char* path1, const char* path2) {
```
andrewsavage1: How does this get built in if we're not adding a --wrap flag below?

rogerzanoni: I checked an older branch and the --wrap flag was present, I dropped it by mistake when solving rebase conflicts. I will re-add it and update this PR.

rogerzanoni:
Changed to a mkostemp only wraper; The focused tests I ran didn't had differences with/without the link wrapper. I will re-check when running the whole suite, but we can proceed with merging the validated mkostemp fixes.

starboard/android/shared/posix_emu/stdlib.cc
```diff
@@ -0,0 +1,60 @@
...
+  struct timespec ts;
+  clock_gettime(CLOCK_REALTIME, &ts);
+  unsigned long r =
+      ts.tv_sec + ts.tv_nsec + (unsigned long)((uintptr_t)tmpl) * 65537UL;
+  for (int i = 0; i < 6; i++, r >>= 5) {
+    tmpl[i] = 'A' + (r & 15) + (r & 16) * 2;
+  }
+}
+
+int __wrap_mkostemp(char* tmpl, int flags) {
```
andrewsavage1, line 26: Can we simplify this significantly by using __real_mkostemp i.e.

```
int __wrap_mkostemp(char* tmpl, int flags) {
  // Bionic rejects access mode and O_CREAT/O_EXCL in flags with EINVAL,
  // since mkostemp already implies O_RDWR | O_CREAT | O_EXCL.
  flags &= ~O_ACCMODE;
  flags &= ~(O_CREAT | O_EXCL);
  return __real_mkostemp(tmpl, flags);
}
```

rogerzanoni: Done

andrewsavage1 (approved):

PR #11334, opened 2026-07-12
nplb: relax rules for checking modes of created directories

rogerzanoni:
nplb: relax rules for checking modes of created directories

On Android the app's data/cache directory tree carries the setgid bit, so the
kernel propagates S_ISGID (setgid, 02000) to every directory created under it.
stat() then reports e.g. 02700 instead of 0700, on top of the requested mode.

mkdir() call sites either ignore the returned mode or check only S_ISDIR, so
this change only relaxes the mode checking in the nplb tests instead of
creating a mkdir wrapper for stripping S_ISGID that would interfere with
Android's group model.

Bug: 53206840

### DIFF
--- a/starboard/nplb/file_helpers.cc
+++ b/starboard/nplb/file_helpers.cc
@@ -171,6 +171,12 @@ bool DirectoryExists(const char* path) {
   return stat(path, &info) == 0 && S_ISDIR(info.st_mode);
 }
 
+bool PermissionsAreSubsetOf(mode_t st_mode, mode_t requested) {
+  const mode_t kPermissionMask = S_IRWXU | S_IRWXG | S_IRWXO;  // 0777
+  const mode_t actual = st_mode & kPermissionMask;
+  return (actual & requested) == actual;
+}
+
 ScopedTempDir::ScopedTempDir() {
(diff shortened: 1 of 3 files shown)


andrewsavage1 (commented):
Is there also an option to wrap the mkdir function on android to have it remove the added permission bits? I'm not necessarily saying that's what we should do, but I do think it would be more posix-compliant

rogerzanoni:
@andrewsavage1  I think that was my first approach to try fixing it during the 25 experiments, but looking more into it, I thought that it's better to not touch the flags; It's a flag that's added by [the kernel](https://android.googlesource.com/kernel/common/+/refs/heads/android-mainline/fs/inode.c#2735), subdirectories automatically inheirit it, making files created inside the diretories inherit its group. It can be problematic if something relies in the inherited groups, I don't thing this happens often, but we would be interfering with android's group model - I would prefer to apply the relaxing rules since (at least up until now) the issue is only in the tests (the palces that use mkdir() either ignore the result or check S_ISDIR). If we decide to go with the test fix, I'd like to upload a new commit rewording the commit message and comments a bit

sacuff (approved):

PR #11335, opened 2026-07-12
nplb: use SIGUSR2 in KillSendsSignalToSelf

rogerzanoni:
SIGUSR1 is intercepted by the Android runtime and never reaches the handler installed by the test. Switch the test to SIGUSR2.

Bug: 532068409

### DIFF
--- a/starboard/nplb/posix_compliance/posix_signal_test.cc
+++ b/starboard/nplb/posix_compliance/posix_signal_test.cc
@@ -236,13 +236,13 @@ TEST_F(PosixSignalTest, KillSendsSignalToSelf) {
   struct sigaction sa = {};
   sa.sa_handler = PipeWritingSignalHandler;
   sigemptyset(&sa.sa_mask);
-  ASSERT_EQ(sigaction(SIGUSR1, &sa, nullptr), 0);
+  ASSERT_EQ(sigaction(SIGUSR2, &sa, nullptr), 0);
 
-  ASSERT_EQ(kill(getpid(), SIGUSR1), 0);
+  ASSERT_EQ(kill(getpid(), SIGUSR2), 0);
 
   int received_signal = 0;
   EXPECT_TRUE(WaitForSignalWithTimeout(&received_signal, 1000));
-  EXPECT_EQ(received_signal, SIGUSR1);
+  EXPECT_EQ(received_signal, SIGUSR2);
 }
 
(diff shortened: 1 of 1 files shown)


sacuff:
Is the behavior of SIGUSR1 documented anywhere by AOSP? I was able to find https://android.googlesource.com/platform/bionic/+/master/tests/signal_test.cpp#398 which leads me to believe it just blocks SIGKILL and SIGSTOP, in which case can we try unblocking them instead of switching to SIGUSR2? Ex: https://github.com/youtube/cobalt/blob/03deecc0cd395a9eb08fa66ddf0d593082e81a2c/starboard/nplb/posix_compliance/posix_usleep_test.cc#L71

rogerzanoni:
@sacuff sorry, my change was poorly documented

I did the change based on ART's SignalCatcher that sets up handlers for SIGUSR1: https://android.googlesource.com/platform/art/+/refs/heads/main/runtime/signal_catcher.cc#207

and the runtime errors that show that it's being intercepted by it:

```
07-27 15:43:10.074 30939 30973 I starboard: [ RUN      ] PosixSignalTest.KillSendsSignalToSelf
07-27 15:43:10.074 30939 30950 I dev.cobalt.coa: Thread[5,tid=30950,WaitingInMainSignalCatcherLoop,Thread*=0xf1052800,peer=0x13780110,"Signal Catcher"]: reacting to signal 10
```

posix_signal_test already tries to unblock SIGUSR1, so I tried adding SIGKILL and SIGSTOP, but the signal is still being intercepted by the ART's signal catcher. 

I think one possible solution is to do something similat to the fix of #11340 and make the signal thread-directed in a posix_emu kill() wrapper, so it ART won't have the chance to intercept it, wdty? AFAIU it's a very similar issue

sacuff:
@jellefoks I see SIGUSR1 (and 2) are used for lifecycle stuff. Do you have an opinion on how AOSP should handle the interception here?

rogerzanoni:
I created this as a proposal: https://github.com/youtube/cobalt/pull/11562

sacuff (approved):
Discussed the signal usage more with Jelle. SIGUSR1 is used on Linux for lifecycle events, but so is SIGUSR2 so this change should be fine without needing to change of the signal handling on Android.

Not relevant to NPLB: on AOSP we will want to make sure to translate lifecycle events to intents instead of going through the signal route.

PR #11338, opened 2026-07-12
android/mic: return -1 from SbMicrophoneRead for a NULL buffer

rogerzanoni:
Update SbMicrophoneImpl::Read logic and
SbMicrophoneReadTest.RainyDayAudioBufferIsNULL test expectations to
consider an output buffer an invalid argument when the size of bytes
expected to be read is valid.

Bug: 532068409

### DIFF
--- a/starboard/android/shared/microphone_impl.cc
+++ b/starboard/android/shared/microphone_impl.cc
@@ -230,7 +230,13 @@ int SbMicrophoneImpl::Read(void* out_audio_data, int audio_data_size) {
     return -1;
   }
 
-  if (!out_audio_data || audio_data_size == 0 || state_ == kWaitPermission) {
+  // If reading audio_data_size bytes is expected, a null output
+  // buffer is an invalid argument.
+  if (audio_data_size > 0 && !out_audio_data) {
+    return -1;
+  }
+
+  if (audio_data_size == 0 || state_ == kWaitPermission) {
     // No data to be read.
     return 0;
   }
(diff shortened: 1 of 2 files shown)


sacuff (approved):

starboard/android/shared/microphone_impl.cc
```diff
@@ -230,7 +230,12 @@ int SbMicrophoneImpl::Read(void* out_audio_data, int audio_data_size) {
     return -1;
   }
 
-  if (!out_audio_data || audio_data_size == 0 || state_ == kWaitPermission) {
+  if (!out_audio_data) {
```
andrewsavage1: Let's just swap the order of these checks (check audio_data_size and state_ and potentially return 0 first) to make this more lenient

andrewsavage1 (changes requested):

starboard/android/shared/microphone_impl.cc
```diff
@@ -230,7 +230,12 @@ int SbMicrophoneImpl::Read(void* out_audio_data, int audio_data_size) {
...
+    // A NULL output buffer is an invalid argument
```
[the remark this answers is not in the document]
rogerzanoni: Changed the comments

starboard/android/shared/microphone_impl.cc
rogerzanoni: I changed Read() rules and test expectations to the following:

when audio_data_size is 0, a null output buffer is ok, no data to be read and it returns 0;
when audio_data_size is valid and data is expected to be read, the buffer should be also valid, so it returns -1, considering the null buffer an invalid argument. wdyt?

sideb0ard (approved):

andrewsavage1 (approved):

PR #11339, opened 2026-07-12
aosp: normalize if_indextoname behavior

rogerzanoni:
bionic's if_indextoname sets ENODEV for a nonexistent interface index, while POSIX expects ENXIO.

Bug: 532068409


sacuff (approved):

andrewsavage1 (changes requested):
This should be implemented in an android-specific wrapper instead

rogerzanoni:
@andrewsavage1 moved it to posix_emu

andrewsavage1 (approved):

PR #11340, opened 2026-07-12
aosp: wrap alarm() and make it use a thread-directed timer

rogerzanoni:
Bionic's implementation of alarm triggers a setitimer set with ITIMER_REAL making it a process-directed signal.

From the signal(7) manpage:
  A  process-directed  signal  may  be  delivered to any one of the threads
  that does not currently have the signal blocked. If more than one of the
  threads has the signal unblocked, then the kernel chooses an arbitrary
  thread to which to deliver the signal.

Android's ART runtime/runtime.cc configures only a small set of signals:

void Runtime::BlockSignals() {
  SignalSet signals;
  signals.Add(SIGPIPE);
  signals.Add(SIGQUIT);
  signals.Add(SIGUSR1);
  signals.Block();
}

SIGALRM is unblocked on the ART's main thread and it may catch the process-directed signal, never letting it reach cobalt/nplb calling threads, making it unreliable to wait for the signal using musl's implementation.

Fixes PosixClockNanosleepTest.ErrorEintrAbsoluteSleep, PosixClockNanosleepTest.ErrorEintrRelativeSleep and PosixNanosleepTests.ErrorEintr which were failing with a "signal never interrupted sleep" error.

Bug: 532068409

### DIFF
--- a/starboard/android/shared/posix_emu/alarm.cc
+++ b/starboard/android/shared/posix_emu/alarm.cc
@@ -1,0 +1,85 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>
+#include <signal.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
+
+#include "starboard/common/log.h"
+
+namespace {
+
+// A one-shot POSIX timer used to implement a thread-directed alarm()
+struct ThreadAlarmTimer {
+  timer_t timer = nullptr;
+
+  ~ThreadAlarmTimer() {
+    if (timer) {
+      timer_delete(timer);
+    }
+  }
+};
+
+thread_local ThreadAlarmTimer g_thread_alarm_timer;
+
+}  // namespace
+
+extern "C" {
+
+unsigned int __real_alarm(unsigned int seconds);
+
+// alarm() on Android/ART calls setitimer set with ITIMER_REAL, which raises a
+// process-directed SIGALRM. signal() may deliver process-directed signals to
+// any of the process threads that has SIGALRM unblocked, and the kernel decides
+// which is going to be the receiving thread. So on Android, waiting for SIGALRM
+// isn't reliable and to circunvent this issue a per-thread POSIX timer is set
+// with SIGEV_THREAD_ID, which sets up the thread that started the timer as the
+// receiver, not letting the signal go to unblocked receivers like the ART
+// signal catcher.
+
+unsigned int __wrap_alarm(unsigned int seconds) {
+  ThreadAlarmTimer& thread_timer = g_thread_alarm_timer;
+  if (thread_timer.timer == nullptr) {
+    if (seconds == 0) {
+      // No timer to create yet; just cancel any existing process-level alarm.
+      return __real_alarm(0);
+    }
+    struct sigevent sev = {};
+    sev.sigev_notify = SIGEV_THREAD_ID;
+    sev.sigev_signo = SIGALRM;
+    sev.sigev_notify_thread_id = gettid();
+    if (timer_create(CLOCK_REALTIME, &sev, &thread_timer.timer) != 0) {
+      SB_LOG(WARNING) << "alarm(): timer_create failed (" << strerror(errno)
+                      << "), falling back to the process-level alarm";
+      return __real_alarm(seconds);
+    }
+  }
+
+  struct itimerspec new_value = {};
+  new_value.it_value.tv_sec = seconds;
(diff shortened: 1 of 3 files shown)


starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -707,6 +709,67 @@ musl_pid_t __abi_wrap_gettid() {
...
+  if (!thread_timer.created) {
+    struct sigevent sev = {};
+    sev.sigev_notify = SIGEV_THREAD_ID;
+    sev.sigev_signo = SIGALRM;
+#if defined(gettid)
+    sev.sigev_notify_thread_id = gettid();
+#else
+    sev.sigev_notify_thread_id = syscall(SYS_gettid);
+#endif
+    if (timer_create(CLOCK_MONOTONIC, &sev, &thread_timer.timer) != 0) {
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -707,6 +709,67 @@ musl_pid_t __abi_wrap_gettid() {
...
+  ThreadAlarmTimer& thread_timer = g_thread_alarm_timer;
+  if (!thread_timer.created) {
+    struct sigevent sev = {};
+    sev.sigev_notify = SIGEV_THREAD_ID;
+    sev.sigev_signo = SIGALRM;
+#if defined(gettid)
+    sev.sigev_notify_thread_id = gettid();
+#else
+    sev.sigev_notify_thread_id = syscall(SYS_gettid);
+#endif
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -707,6 +709,67 @@ musl_pid_t __abi_wrap_gettid() {
...
+// process threads that has SIGALRM unblocked, and the kernel decides which is going to
+// be the receiving thread. So on Android, waiting for SIGALRM isn't reliable and to
+// circunvent this issue a per-thread POSIX timer is set with SIGEV_THREAD_ID, which
+// sets up the thread that started the timer as the receiver, not letting the
+// signal go to unblocked receivers like the ART signal catcher.
+
+unsigned int __abi_wrap_alarm(unsigned int seconds) {
+  ThreadAlarmTimer& thread_timer = g_thread_alarm_timer;
+  if (!thread_timer.created) {
+    struct sigevent sev = {};
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

starboard/android/shared/posix_emu/alarm.cc
```diff
@@ -0,0 +1,81 @@
...
+#include <signal.h>
+#include <time.h>
+#include <unistd.h>
+
+namespace {
+
+// A one-shot POSIX timer used to implement a thread-directed alarm()
+struct ThreadAlarmTimer {
+  timer_t timer = nullptr;
+  bool created = false;
```
andrewsavage1: Is this necessary? Can we just check if `timer` is not nullptr instead?

andrewsavage1:
@sacuff @jellefoks ptal

starboard/android/shared/posix_emu/alarm.cc
```diff
@@ -0,0 +1,81 @@
...
+      return __real_alarm(seconds);
+    }
+    thread_timer.created = true;
+  }
+
+  struct itimerspec new_value = {};
+  new_value.it_value.tv_sec = seconds;
+  struct itimerspec old_value = {};
+  if (timer_settime(thread_timer.timer, 0, &new_value, &old_value) != 0) {
+    // fallback
```
sacuff: Can we log a warning?

starboard/android/shared/posix_emu/alarm.cc
sacuff: +1

rogerzanoni: Removed it

starboard/android/shared/posix_emu/alarm.cc
rogerzanoni: Done

starboard/android/shared/posix_emu/alarm.cc
```diff
@@ -0,0 +1,85 @@
...
+  new_value.it_value.tv_sec = seconds;
+  struct itimerspec old_value = {};
+  if (timer_settime(thread_timer.timer, 0, &new_value, &old_value) != 0) {
+    SB_LOG(WARNING) << "alarm(): timer_settime failed (" << strerror(errno)
+                    << "), falling back to the process-level alarm";
+    return __real_alarm(seconds);
+  }
+
+  // matches musl's implementation return logic, but using
+  // nsec instead of usec
```
andrewsavage1, line 81: Why do we use nsec instead of usec?

andrewsavage1 (approved):

starboard/android/shared/posix_emu/alarm.cc
sacuff: IIUC `timer_settime` takes `itimerspec` args and that struct uses nsec, vs musl's version uses `setitimer` with `itimerval` which has usec. `setitimer` was removed from posix so the change from musl to here is reasonable IMO.

I'd say the comment referencing the nsec/usec difference from musl is burying the lede a bit. It can probably just explain that we match the behavior to round up any partial seconds that are left on the timer.

PR #11343, opened 2026-07-13
Fix Cobalt compile on M139

linxinan-yt:
Different than #11283, this attempt also updated the third party code, e.g. v8 and skia to M139 and fixed the compile error on linux and android.


net/BUILD.gn
```diff
@@ -1828,7 +1815,18 @@ component("net") {
...
+  }
+
+  if (enable_disk_cache_sql_backend) {
+    sources += [
+      "disk_cache/sql/cache_entry_key.cc",
+      "disk_cache/sql/cache_entry_key.h",
+      "disk_cache/sql/sql_backend_constants.h",
+      "disk_cache/sql/sql_backend_impl.cc",
+      "disk_cache/sql/sql_backend_impl.h",
+      "disk_cache/sql/sql_persistent_store.h",
```
abhijeetk, line 1827: We missed to add `disk_cache/sql/sql_persistent_store.cc`

linxinan-yt: Correct. Currently if the glitch does not result a compile error, it is hard for AI to detect it. That's why we prioritize the autoroller script to produce a reviewable PR. 

Thanks for pointing it!

PR #11350, opened 2026-07-13
starboard: Replace per-target toolchain checks with file-level asserts

hferreiro:
The platform BUILD.gn files for Android, AOSP, Linux and rdk are only ever evaluated in the Starboard toolchain. Make that explicit with a top-of-file `assert(is_starboard_toolchain)` and drop the now-redundant per-target `current_toolchain == starboard_toolchain` checks.

Move default_toolchain targets out of rdk's shared BUILD.gn, the one
exception, into their own file so that the invariant holds there too. Also simplify its `loader_app` target, to avoid an unneeded indirection to //starboard/loader_app:loader_app.

Bug: 495203133

### DIFF
--- a/starboard/android/shared/BUILD.gn
+++ b/starboard/android/shared/BUILD.gn
@@ -23,6 +23,8 @@ import("//testing/test.gni")
 # Configuration for overall Android Starboard Platform.
 ##########################################################
 
+assert(is_starboard_toolchain)
+
 config("starboard_platform_libs") {
   libs = [
     "EGL",
@@ -384,7 +386,7 @@ static_library("starboard_base_symbolize") {
     "//starboard:starboard_headers_only",
     "//starboard/common:common_headers_only",
   ]
-  if (is_starboard && current_toolchain == starboard_toolchain) {
+  if (is_starboard) {
     public_deps = [ "//starboard/elf_loader:evergreen_info" ]
   }
 
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/cobalt/BUILD.gn
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/cobalt/BUILD.gn
@@ -12,11 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//starboard/build/config/linux/rdk/config_variables.gni")
+
+group("loader_app_rdk_plugin") {
+  data_deps = [ "$rdk_starboard_root/third_party/starboard/rdk/shared:loader_app($starboard_toolchain)" ]
+}
+
 group("rdk_cobalt_content") {
   data_deps = [ ":rdk_cobalt_keymap($default_toolchain)" ]
 }
 
 copy("rdk_cobalt_keymap") {
-  sources = [ "etc/keymapping.json" ]
+  sources = [ "content/etc/keymapping.json" ]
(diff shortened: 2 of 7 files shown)


starboard/android/shared/BUILD.gn
```diff
@@ -23,6 +23,8 @@ import("//testing/test.gni")
 # Configuration for overall Android Starboard Platform.
 ##########################################################
 
+assert(is_starboard_toolchain)
```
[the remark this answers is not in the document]
hferreiro: The import is at line 16.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/cobalt/BUILD.gn
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import("//starboard/build/config/linux/rdk/config_variables.gni")
+
+group("loader_app_rdk_plugin") {
```
andrewsavage1, line 17: What's the purpose of separating out these targets?

hferreiro: It's described in the commit message, they are evaluated in the default_toolchain. If this doesn't look right, we can always omit the assertion in rdk.

niranjanyardi: loader_app_rdk_plugin - the only reason this gets built with default chain is for discoverability . otherwise you would have to specify the starboard toolchain for building this target( while running autoninja)

The other 2 targets below(rdk_cobalt_content, rdk_cobalt_keymap) - im not sure why they need to be built by the default toolchain - that's upstream code , maybe we can jsut build this with starboard toolchain going forward - so feel free to move these to the other starboard file- starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn

hferreiro: It makes sense to keep loader_app_rdk_plugin in the default toolchain for discoverability. Regarding rdk_cobalt_keymap, the reason might be to avoid duplicating copy targets per-toolchain. In any case, rdk_cobalt_keymap would need to be kept in the default toolchain in it's current file, so that it's loaded by GN. Otherwise, there would be no references to starboard/contrib/rdk/src/third_party/starboard/rdk/shared/cobalt/BUILD.gn.

hferreiro:
Can we resume the review of this PR?

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/cobalt/BUILD.gn
andrewsavage1: @niranjanyardi I'm going to submit this as is, but it should probably be cleaned up from the RDK side as these targets either may not be needed or shouldn't have a default toolchain shim (and should just be included in existing build targets when necessary)

andrewsavage1 (approved):

PR #11351, opened 2026-07-13
build: Remove obsolete install_target packaging machinery

hferreiro:
In the old Starboard build, the install_target template and its companions staged deployable output but have become stale: the Chromium-based build never wired them up.

Remove the dead machinery: the per-platform install_target implementations, their GN arguments, and the rest of the content-staging plumbing (install_content, bundle_content, convert_i18n_data); and the legacy Gradle project in cobalt/android/apk.

Bug: 492403745


starboard/android/shared/install_target.gni
johnxwork, line 1: are you sure we don't need this? Can the Android TV build still package assets without this?

andrewsavage1: I don't believe ATV or any platform use the install targets anymore

oxve: Chromium uses the same mechanism as install_target (data_deps) to collect runtime dependencies and package them. Chrobalt uses it instead of install_target for this purpose.

oxve (approved):
install_target changes lgtm

I'll defer on gradle changes.

andrewsavage1 (approved):

andrewsavage1:
@johnxwork or @zhongqiliang PTAL

zhongqiliang (approved):

PR #11352, opened 2026-07-13
media: Exclude //media/muxers and MediaRecorder for Cobalt 

osagie98:
Excludes //media/muxers, //third_party/libwebm, and Blink's
MediaRecorder API when is_cobalt is true, reducing stripped native
release binary size by 684.0 KiB (1.58 MiB unstripped) on android-arm_gold
and 660.0 KiB (6.21 MiB unstripped) on evergreen-x64_gold.

While the direct compilation footprint of //media/muxers, libwebm, and
MediaRecorder accounts for ~100.6 KiB, surgically gating out the sole
consumer (MediaRecorder) enables the linker (--gc-sections) to cleanly
garbage-collect ~584 KiB of transitive dependencies. This prunes Cobalt's
entire unused video/audio encoding pipeline, including software encode
wrappers (media::VideoEncoder/AudioEncoder), encode-side libyuv frame
formatters/scalers, and recording-specific Blink/V8 dictionary converters.

Key decisions:
* media_options.gni & media/BUILD.gn: Conditionally remove //media/muxers
  from media_subcomponent_deps and exclude unit tests. Allow //media/muxers
  in subcomponent_config visibility so gn check passes cleanly.
* Blink modules & IDL exclusions: Exclude mediarecorder from sub_modules
  and remove blob_event.idl, blob_event_init.idl, and media_recorder.idl
  from idl_in_modules.gni. Preserve media_recorder_options.idl so the
  BitrateMode enum remains available for WebCodecs IDL validation.
* V8 bindings (bindings.gni): Add *blob_event*, *v8_media_recorder.*, and
  *v8_recording_state* to cobalt_bindings_exclude_patterns to omit
  generated V8 wrappers without breaking check_generated_file_list.py.
* Pure call-site gating (media_capabilities.cc): Gate MediaRecorderHandler
  instantiation and unused static configuration helpers under !IS_COBALT,
  allowing encodingInfo({type: "record"}) queries to cleanly fall through
  to return supported: false.

#vibe-coded

Bug: 531833882


third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
```diff
@@ -412,6 +414,7 @@ bool IsValidMediaEncodingConfiguration(
   return true;
 }
 
+#if !BUILDFLAG(IS_COBALT)
 WebAudioConfiguration ToWebAudioConfiguration(
     const AudioConfiguration* configuration) {
```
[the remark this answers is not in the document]
osagie98: Response from jetski:

•  ToWebAudioConfiguration  and  ToWebVideoConfiguration  convert IDL dictionaries specifically into  WebAudioConfiguration  and  WebVideoConfiguration  structs, which are only consumed
  by  ToWebMediaConfiguration  for encoding/recording capability queries.
  •  decodingInfo()  uses completely separate parsing paths ( ToSdpAudioFormat() ,  ToSdpVideoFormat() , and direct struct access on  AudioConfiguration  /  VideoConfiguration ) to query 
  decoder support ( IsDecoderSupportedVideoType ).
  
  Because  ToWebMediaConfiguration  is gated on Cobalt ( !IS_COBALT ),  ToWebAudioConfiguration  and  ToWebVideoConfiguration  become 100% unused static helpers when  IS_COBALT  is true. 
  Per Surgical Feature Gating

third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
```diff
@@ -467,7 +470,6 @@ WebVideoConfiguration ToWebVideoConfiguration(
 
   return web_configuration;
 }
-
 WebMediaConfiguration ToWebMediaConfiguration(
     const MediaEncodingConfiguration* configuration) {
```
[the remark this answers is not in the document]
osagie98: see above

third_party/blink/renderer/bindings/idl_in_modules.gni
```diff
@@ -1336,6 +1336,17 @@ if (is_cobalt) {
             ],
             "abspath")
   }
+
+  _filtered_mediarecorder_idl =
```
osagie98: Not sure of the current behavior of this interface, but this will cause these APIs to be undefined. They could be stubbed out if we need to keep them in place

third_party/blink/renderer/bindings/idl_in_modules.gni
```diff
@@ -1337,6 +1337,16 @@ if (is_cobalt) {
             "abspath")
   }
 
+  _filtered_mediarecorder_idl = filter_exclude(
```
andrewsavage1: Can we just do `static_idl_files_in_modules -=`?

third_party/blink/renderer/bindings/idl_in_modules.gni
```diff
@@ -1337,6 +1337,16 @@ if (is_cobalt) {
...
+          static_idl_files_in_modules,
+          [
+            "//third_party/blink/renderer/modules/mediarecorder/blob_event.idl",
```
andrewsavage1: In general I would expect us to try to avoid changing the actual web interface and instead give a rejected promise or similar failure for apis that we don't support so that we maintain a consistent web interface but can still remove unsupported functionality. This also (depending on implementation) can let us use a different mojo implementation, where there are often platform-specific differences, instead of making customizations in blink, which is generally consistent between platforms. What's the reason to make the changes here?

osagie98: This avoids generating the V8 code that uses the `MediaRecorder` library, which is nearly all of the potential size savings. Apparently I can't stub it out as the code will still be generated, but I agree that it's preferable to keep it in and return empty promises.

third_party/blink/renderer/bindings/idl_in_modules.gni
osagie98: Yes, that works

andrewsavage1 (approved):

media/BUILD.gn
```diff
@@ -220,7 +226,10 @@ test("media_unittests") {
     # Cobalt disabled WebRtc processing in 
     # https://github.com/youtube/cobalt/pull/10471
     # due to additional latency for the microphone path.
-    deps -= ["//media/webrtc:unit_tests"]
+    deps -= [
```
borongc, line 229: Isn't this gated by `is_cobalt`?

borongc: NVM, I saw it is inside this block.

third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
```diff
@@ -83,6 +82,10 @@
 #include "ui/gfx/geometry/rect.h"
 #include "ui/gfx/geometry/size.h"
 
+#if !BUILDFLAG(IS_COBALT)
```
borongc: Should we put this back to L61 to L62, as this is upstream code.

borongc (approved):
LGTM with minor comments.

PR #11355, opened 2026-07-13
Pass Java exception details to JS in PlatformService

johnxwork:
Updated PlatformService JNI and Mojo to capture Java exceptions during receiveFromClient and propagate them back to the renderer process. The Blink implementation now throws a DOMException containing the Java exception message and stack trace, enabling better debugging via JS logs.

Bug: 534360029
TAG=agy
CONV=e5bd021a-7b5c-49e6-92df-e3c00a555578


cobalt/browser/h5vcc_platform_service/platform_service_impl.cc
y4vor: you may need to call with 2nd std:nullopt for the error message.

johnxwork: Added an error message for the second param.

y4vor (approved):

PR #11375, opened 2026-07-14
Restrict viewport size to 1080p on Android.

johnxwork:
Capped the surface size to 1080p in ContentViewRenderView to restrict the viewport size to 1080p instead of relying on the screen resolution. On devices with screen resolution higher than 1080p, the surface will be set to 1080p and scaled up by the Android system.

Bug: 534832052
TAG=agy
CONV=5e10bad9-5110-43f9-9422-08dd6f2d2e76

### DIFF
--- a/cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
+++ b/cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
@@ -12,6 +12,7 @@ import android.view.MotionEvent;
 import android.view.SurfaceView;
 import android.view.View;
 import android.widget.FrameLayout;
+import android.util.Size;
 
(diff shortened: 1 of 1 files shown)


cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -291,6 +300,23 @@ public boolean onGenericMotionEvent(MotionEvent event) {
         return super.onGenericMotionEvent(event);
     }
 
+    private Size getCappedSize(int width, int height) {
```
borongc, line 303: Is this necessary? For most of 4K devices I saw, the viewport is already 1080x1920.

I don't know if this will cause any issues, and I think this should be gated by a flag if you want to do so.

PR #11376, opened 2026-07-14
android: Wait for media resources during shutdown

zhongqiliang:
Coordinate the destruction of media resources during application
shutdown to ensure that platform-level components like AudioTrack and
MediaDrm are properly cleaned up. By delaying the closure of the
native Starboard bridge until all tracked media resources are
released, we avoid stability issues and potential crashes caused by
exiting the process while these resources are still active.

Issue: 534798847


starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -104,6 +107,21 @@ jlong JNI_BaseStarboardBridge_StartNativeStarboard(
 
 void JNI_BaseStarboardBridge_CloseNativeStarboard(JNIEnv* env,
                                                   jlong nativeApp) {
+  // Wait for all active media resources (MediaCodec, AudioTrack, MediaDrm)
+  // to be destroyed before deleting the application and exiting the JVM.
+  int timeout_ms = 2000;  // 2 seconds timeout
+  int elapsed_ms = 0;
+  auto* bridge = StarboardBridge::GetInstance();
+  while (bridge->GetActiveMediaResourceCount() > 0 && elapsed_ms < timeout_ms) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
kjyoun: can you use `usleep` for the consistency?

https://source.corp.google.com/search?q=usleep&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

starboard/android/shared/starboard_bridge.h
```diff
@@ -121,6 +127,8 @@ class StarboardBridge {
 
   // Java StarboardBridge instance.
   jni_zero::ScopedJavaGlobalRef<jobject> j_starboard_bridge_;
+
+  std::atomic<int> active_media_resource_count_{0};
```
kjyoun: `std::atomic<int> active_media_resource_count_ = 0;` ?  for better readability

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -104,6 +107,21 @@ jlong JNI_BaseStarboardBridge_StartNativeStarboard(
 
 void JNI_BaseStarboardBridge_CloseNativeStarboard(JNIEnv* env,
                                                   jlong nativeApp) {
+  // Wait for all active media resources (MediaCodec, AudioTrack, MediaDrm)
+  // to be destroyed before deleting the application and exiting the JVM.
+  int timeout_ms = 2000;  // 2 seconds timeout
```
kjyoun: constexpr

starboard/android/shared/starboard_bridge.h
```diff
@@ -109,6 +111,10 @@ class StarboardBridge {
   void SetStartupMilestone(jint milestone) const;
   void SetStartupDiagnosisInfo(const char* key, const char* value) const;
 
+  void IncrementMediaResourceCount();
```
kjyoun: Can you define separate class(e.g., `ResourceTracker`) for tracking resource count?

The role of StarboardBriddge is to work as JNI interface for `cobalt/android/apk/app/src/main/java/dev/cobalt/coat/StarboardBridge.java`

This increase/decrement logic is not related with StarboardBridge

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -104,6 +107,21 @@ jlong JNI_BaseStarboardBridge_StartNativeStarboard(
...
+  int elapsed_ms = 0;
+  auto* bridge = StarboardBridge::GetInstance();
+  while (bridge->GetActiveMediaResourceCount() > 0 && elapsed_ms < timeout_ms) {
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    elapsed_ms += 10;
+  }
```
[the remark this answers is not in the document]
zhongqiliang: closeNativeStarboard MUST block on the UI thread because onActivityDestroy immediately calls System.exit(0) right after it returns. Offloading to a background thread in Java would cause System.exit(0) to kill the process prematurely.

cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
```diff
@@ -176,6 +176,7 @@ private void removeShell(Shell shellView) {
     public void destroy() {
         // Remove active shell (Currently single shell support only available).
         if (mActiveShell != null) {
+            mActiveShell.close();
```
kjyoun, line 179: Can you double-check calling this `mActiveShell.close();`  actually close the app?  since this PR depends on the premise that it will closes the app

IIUC, calling `mActiveShell.close();` here does not close the app.

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -461,4 +479,16 @@ void StarboardBridge::SetStartupDiagnosisInfo(const char* key,
...
+  active_media_resource_count_.fetch_add(1, std::memory_order_relaxed);
+}
+
+void StarboardBridge::DecrementMediaResourceCount() {
+  active_media_resource_count_.fetch_sub(1, std::memory_order_relaxed);
+}
+
+int StarboardBridge::GetActiveMediaResourceCount() const {
+  return active_media_resource_count_.load(std::memory_order_relaxed);
+}
```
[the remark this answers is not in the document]
zhongqiliang: done

cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
zhongqiliang: mActiveShell.close() is not meant to trigger app exit, but to force immediate native WebContents and media player teardown during Activity destruction so the resource counter can drop to 0 before process exit.

starboard/android/shared/starboard_bridge.h
zhongqiliang: done

starboard/android/shared/base_starboard_bridge.cc
zhongqiliang: done

starboard/android/shared/starboard_bridge.h
zhongqiliang: done

starboard/android/shared/base_starboard_bridge.cc
zhongqiliang: done

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,53 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef STARBOARD_ANDROID_SHARED_MEDIA_RESOURCE_TRACKER_H_
+#define STARBOARD_ANDROID_SHARED_MEDIA_RESOURCE_TRACKER_H_
+
+#include <atomic>
+
+#include "base/memory/singleton.h"
```
kjyoun: Accessing base from starboard is discouraged

https://b.corp.google.com/issues/509619283

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,53 @@
...
+
+  void Increment() {
+    active_media_resource_count_.fetch_add(1, std::memory_order_relaxed);
+  }
+
+  void Decrement() {
+    active_media_resource_count_.fetch_sub(1, std::memory_order_release);
+  }
+
+  int GetCount() const {
```
kjyoun: This method is used for polling

Then how about using condtion_variable like below?

Then we dont need polling, though we use mutex, instead of atomic.

WDYT?


```
void Increment() {
    std::lock_guard lock(mutex_);
    active_media_resource_count_++;
  }

  void Decrement() {
    std::unique_lock lock(mutex_);
    active_media_resource_count_--;
    if (active_media_resource_count_ <= 0) {
      lock.unlock();
      cv_.notify_all();
    }
  }

int WaitUntilZero(int timeout_ms) {
    std::unique_lock lock(mutex_);
    cv_.wait_for(lock, std::chrono::milliseconds(timeout_ms), [this] {
      return active_media_resource_count_ <= 0;
    });
    return active_media_resource_count_;
  }

// Then call site will become

int remaining_count = tracker->WaitUntilZero(kTimeoutMs);
  if (remaining_count > 0) {
    SB_LOG(WARNING) << "Timed out waiting for all media resources to be "
                       "destroyed. Active count: "
                    << remaining_count;
  }

```

starboard/android/shared/media_resource_tracker.h
zhongqiliang: Done

starboard/android/shared/media_resource_tracker.h
zhongqiliang: Check again, WDYT

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -104,6 +105,17 @@ jlong JNI_BaseStarboardBridge_StartNativeStarboard(
 
 void JNI_BaseStarboardBridge_CloseNativeStarboard(JNIEnv* env,
                                                   jlong nativeApp) {
+  // Wait for all active media resources (MediaCodec, AudioTrack, MediaDrm)
+  // to be destroyed before deleting the application and exiting the JVM.
+  constexpr int kTimeoutMs = 2000;
```
kjyoun: super-nit: using digit seperator would help. `2'000`

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+#include <mutex>
+
+namespace starboard {
+
+class MediaResourceTracker {
+ public:
+  // Returns the process-wide singleton instance. Implemented using a C++11
+  // thread-safe static local variable to remain decoupled from base::Singleton.
+  static MediaResourceTracker* GetInstance() {
+    static MediaResourceTracker instance;
```
kjyoun: I think the Google C++ Style guide explicitly bans static variables with non-trivial destructors

Please use NoDesctructor

https://source.corp.google.com/search?q=NoDestructor&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+  // Decrements the count of active media resources. Called in the destructors
+  // of SbPlayer, AudioTrackBridge, and MediaDrmBridge. Uses release memory
+  // ordering to guarantee all destructor writes are committed before
+  // decrements.
+  void Decrement() {
+    if (active_media_resource_count_.fetch_sub(1, std::memory_order_release) ==
+        1) {
+      // Last active media resource destroyed: notify the waiting UI thread
+      // immediately without polling latency.
+      std::lock_guard<std::mutex> lock(mutex_);
```
kjyoun: we don't need "std::mutex", since c++ support CTAD

please remove it from conciseness

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+  // CloseNativeStarboard) until all active media resources hit 0, or until
+  // |timeout_ms| elapses. Returns the remaining active resource count (0 on
+  // success, >0 on timeout).
+  int WaitUntilZero(int timeout_ms) {
+    // Fast path: if no resources are active, return immediately without
+    // locking.
+    if (active_media_resource_count_.load(std::memory_order_acquire) <= 0) {
+      return 0;
+    }
+    std::unique_lock<std::mutex> lock(mutex_);
```
kjyoun: dont need "<std::mutex>"

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+      return active_media_resource_count_.load(std::memory_order_acquire) <= 0;
+    });
+    return active_media_resource_count_.load(std::memory_order_acquire);
+  }
+
+ private:
+  MediaResourceTracker() = default;
+  ~MediaResourceTracker() = default;
+
+  MediaResourceTracker(const MediaResourceTracker&) = delete;
```
kjyoun: Can you move this statements to disable copy and move under :public ?

Disabling copy and move is "public" information that caller needs to know.

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  // Decrements the count of active media resources. Called in the destructors
+  // of SbPlayer, AudioTrackBridge, and MediaDrmBridge. Uses release memory
+  // ordering to guarantee all destructor writes are committed before
+  // decrements.
+  void Decrement() {
+    if (active_media_resource_count_.fetch_sub(1, std::memory_order_release) ==
+        1) {
+      // Last active media resource destroyed: notify the waiting UI thread
```
kjyoun: do you need really all these comments?

I think that those method name is pretty self-explanatory.

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+#define STARBOARD_ANDROID_SHARED_MEDIA_RESOURCE_TRACKER_H_
+
+#include <atomic>
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
+
+namespace starboard {
+
+class MediaResourceTracker {
```
kjyoun, line 30: can you add unit test for this class?

Since this class is a stand-alone, Gemini will easily write ones for you.

starboard/android/shared/media_resource_tracker.h
```diff
@@ -0,0 +1,85 @@
...
+      std::lock_guard<std::mutex> lock(mutex_);
+      cv_.notify_all();
+    }
+  }
+
+  // Blocks the calling thread (typically the Android UI thread in
+  // CloseNativeStarboard) until all active media resources hit 0, or until
+  // |timeout_ms| elapses. Returns the remaining active resource count (0 on
+  // success, >0 on timeout).
+  int WaitUntilZero(int timeout_ms) {
```
kjyoun: can you move impl. logic into .cc file?

Now implementation of these methods becomes too complicated and long to be in .h file.

kjyoun (approved):

starboard/android/shared/media_resource_tracker.h
zhongqiliang: done

starboard/android/shared/media_resource_tracker.h
zhongqiliang: done

starboard/android/shared/media_resource_tracker.cc
```diff
@@ -0,0 +1,51 @@
...
+}
+
+void MediaResourceTracker::Increment() {
+  active_media_resource_count_.fetch_add(1, std::memory_order_relaxed);
+}
+
+void MediaResourceTracker::Decrement() {
+  if (active_media_resource_count_.fetch_sub(1, std::memory_order_release) ==
+      1) {
+    std::lock_guard lock(mutex_);
```
osagie98, line 45: It could be a good idea to also add a debug log here or in `WaitUntilZero` when the number of resources drops to 0; if we run into future issues it'd make it easy to see when exactly the last media resource was cleaned up

starboard/android/shared/media_resource_tracker.cc
```diff
@@ -0,0 +1,51 @@
...
+MediaResourceTracker* MediaResourceTracker::GetInstance() {
+  static starboard::NoDestructor<MediaResourceTracker> instance;
+  return instance.get();
+}
+
+void MediaResourceTracker::Increment() {
+  active_media_resource_count_.fetch_add(1, std::memory_order_relaxed);
+}
+
+void MediaResourceTracker::Decrement() {
```
osagie98, line 34: This allows decrementing to negative numbers. Do we expect that to happen in normal situations? If not, I don't think it needs a CHECK, but maybe a warning log

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -104,6 +105,17 @@ jlong JNI_BaseStarboardBridge_StartNativeStarboard(
 
 void JNI_BaseStarboardBridge_CloseNativeStarboard(JNIEnv* env,
                                                   jlong nativeApp) {
+  // Wait for all active media resources (MediaCodec, AudioTrack, MediaDrm)
+  // to be destroyed before deleting the application and exiting the JVM.
+  constexpr int kTimeoutMs = 2'000;
+  int remaining_count =
```
borongc: This change has a risk, and it should be gated by an experimental flag.

starboard/android/shared/media_resource_tracker.cc
```diff
@@ -0,0 +1,51 @@
...
+    cv_.notify_all();
+  }
+}
+
+int MediaResourceTracker::WaitUntilZero(int timeout_ms) {
+  if (active_media_resource_count_.load(std::memory_order_acquire) <= 0) {
+    return 0;
+  }
+  std::unique_lock lock(mutex_);
+  cv_.wait_for(lock, std::chrono::milliseconds(timeout_ms), [this] {
```
jasonzhangxx, line 56: The blocking wait has a significant deadlock risk. During the player pipeline shutdown, numerous Android APIs are used, which may involve synchronous calls back to the main thread or complex lock chains, and that is out of our control. Blocking here risks triggering a deadlock.

starboard/android/shared/base_starboard_bridge.cc
zhongqiliang: added

starboard/android/shared/media_resource_tracker.cc
zhongqiliang: I 've made to call WaitUntilZero in a background thread.

The exit(0) call would make sure all the native threads are terminated, finish of onDestroy() is not the same as terminate of the process.

starboard/android/shared/media_resource_tracker.cc
zhongqiliang: add log

starboard/android/shared/media_resource_tracker.cc
zhongqiliang: added log

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
```diff
@@ -211,6 +215,11 @@ public static List<String> getExtraCommandLineArgs(Map<String, String> javaSwitc
           "--enable-features=" + JavaSwitches.COBALT_BYPASS_BUFFERING_BYTES_CONSUMER);
     }
 
+    if (javaSwitches.containsKey(JavaSwitches.WAIT_FOR_MEDIA_RESOURCES_ON_SHUTDOWN)) {
+      extraCommandLineArgs.add(
+          "--enable-features=" + JavaSwitches.WAIT_FOR_MEDIA_RESOURCES_ON_SHUTDOWN);
```
Awallky: This existed before this PR, so it's not necessarily in scope for this PR. 
Is it possible that, given the right experiment configuration, that we turn on multiple feature flags and we  have multiple switches on that append `--enable-features=` to ur feature flag overrides?
Would that cause issues?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -207,10 +207,9 @@ protected void onActivityDestroy(Activity activity) {
     if (mApplicationStopped) {
       // We can't restart the starboard app, so kill the process for a clean start next time.
       Log.i(TAG, "Activity destroyed after shutdown; killing app.");
-      BaseStarboardBridgeJni.get().closeNativeStarboard(mNativeApp);
       mTtsHelper.shutdown();
       mAdvertisingId.shutdown();
-      System.exit(0);
```
Awallky, line 213: Is this removal intentional? 
This seems like we are deferring exit until everything is cleaned up. Do we need to add this elsewhere? Or is it no longer needed with this change?

Awallky (approved):
LGTM % comments

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
zhongqiliang: exit(0) is moved into closeNativeStarboard(), that place is better.

cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java
zhongqiliang: Done, want to take another look?

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -106,6 +110,24 @@ void JNI_BaseStarboardBridge_CloseNativeStarboard(JNIEnv* env,
...
+      constexpr int kTimeoutMs = 2'000;
+      int remaining_count =
+          MediaResourceTracker::GetInstance()->WaitUntilZero(kTimeoutMs);
+      if (remaining_count > 0) {
+        SB_LOG(WARNING) << "Timed out waiting for all media resources to be "
+                           "destroyed. Active count: "
+                        << remaining_count;
+      }
+      exit(0);
+    }).detach();
```
jasonzhangxx, line 126: Spawning a new thread in onDestroy() doesn't block the main thread. It won't prevent the app process from terminating, so I cannot see why we need the change.

zhongqiliang: The process is not terminated after onDestroy() finishes. It terminates after exit(0) is called.

cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
jasonzhangxx, line 179: This change is not guarded by feature flag.

starboard/android/shared/base_starboard_bridge.cc
jasonzhangxx: Ok, I tried to find a way totrigger onDestroy() callback. Changing settings doesn't trigger that. I can trigger onDestroy() only by enabling "Don't Keep Activities" in developer setting. When enabling both "Don't Keep Activities" and kWaitForMediaResourcesOnShutdown, app will always crash after onDestroy() is invoked, no matter if there's any video playing. Here's the callstack on crash: https://paste.googleplex.com/5983443577077760.

It's still unclear to me how the user can trigger that and how the delay could make difference.

starboard/android/shared/audio_track_bridge.cc
```diff
@@ -123,7 +124,9 @@ AudioTrackBridge::AudioTrackBridge(
     const DataArray& j_audio_data)
     : max_samples_per_write_(max_samples_per_write),
       j_audio_track_bridge_(j_audio_track_bridge),
-      j_audio_data_(j_audio_data) {}
+      j_audio_data_(j_audio_data) {
+  MediaResourceTracker::GetInstance()->Increment();
```
jasonzhangxx, line 128: For audio sink and drm bridge, their life cycles are not strictly same as SbPlayer. Audio sink is also used by web audio. Drm bridge is used by MediaKeySession.

starboard/android/shared/base_starboard_bridge.cc
```diff
@@ -106,6 +110,24 @@ void JNI_BaseStarboardBridge_CloseNativeStarboard(JNIEnv* env,
                                                   jlong nativeApp) {
   auto* app = reinterpret_cast<ApplicationAndroid*>(nativeApp);
   delete app;
+
+  if (features::FeatureList::IsEnabled(
+          features::kWaitForMediaResourcesOnShutdown)) {
```
jasonzhangxx, line 115: When using starboard features within app lifecycle callbacks, we need to be very cautious. According to the design of the starboard feature, the features cannot be accessed before they are fully initialized, as doing so ensures consistency across values. Any attempt to access them before initialization will trigger a CHECK failure.

cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
zhongqiliang: This is causing the crash you saw when the app exit, https://github.com/youtube/cobalt/pull/11554 reverts it.

starboard/android/shared/base_starboard_bridge.cc
zhongqiliang: https://github.com/youtube/cobalt/pull/11554 will address this crash issue.

PR #11377, opened 2026-07-14
android: Prevent false-positive post-seek frame drops

borongc:
This change introduces a mechanism to ignore stale rendered frame
callbacks that occur after a seek operation but before the first
post-seek frame is processed. This prevents false-positive frame drop
counts caused by callbacks from pre-seek flushed frames.

Issue: 401790323
Issue: 534774024

### DIFF
--- a/starboard/android/shared/video_frame_tracker.cc
+++ b/starboard/android/shared/video_frame_tracker.cc
@@ -124,6 +124,9 @@ void VideoFrameTracker::Seek(int64_t seek_to_time) {
 
   frames_to_be_rendered_.clear();
   seek_to_time_ = seek_to_time;
+  if (ignore_stale_rendered_frames_after_seek_) {
+    awaiting_first_frame_after_seek_ = true;
+  }
 }
 
 int VideoFrameTracker::UpdateAndGetDroppedFrames() {
@@ -163,11 +166,17 @@ void VideoFrameTracker::UpdateDroppedFrames() {
       if (std::abs(*to_render_timestamp - rendered_timestamp) <=
           kMaxAllowedSkew) {
         // This frame was rendered, remove it from frames_to_be_rendered_.
+        awaiting_first_frame_after_seek_ = false;
         to_render_timestamp = frames_to_be_rendered_.erase(to_render_timestamp);
       } else {
         // Here we know rendered_timestamp - *to_render_timestamp >
         // kMaxAllowedSkew, because the while loop condition guarantees
         // *to_render_timestamp - rendered_timestamp <= kMaxAllowedSkew.
+        if (awaiting_first_frame_after_seek_) {
+          // Ignore stale rendered frame callbacks from flushed pre-seek
+          // frames until the first post-seek frame match is found.
+          break;
+        }
         // The rendered frame is too far ahead. The to_render_timestamp frame
         // was dropped.
         SB_LOG(WARNING) << "Video frame dropped:" << *to_render_timestamp
--- a/starboard/android/shared/video_frame_tracker.h
+++ b/starboard/android/shared/video_frame_tracker.h
@@ -25,8 +25,12 @@ namespace starboard {
 
 class VideoFrameTracker {
  public:
-  explicit VideoFrameTracker(int max_pending_frames_size)
-      : max_pending_frames_size_(max_pending_frames_size) {}
+  explicit VideoFrameTracker(
+      int max_pending_frames_size,
+      bool ignore_stale_rendered_frames_after_seek = false)
+      : max_pending_frames_size_(max_pending_frames_size),
+        ignore_stale_rendered_frames_after_seek_(
+            ignore_stale_rendered_frames_after_seek) {}
 
   int64_t seek_to_time() const;
 
@@ -46,6 +50,8 @@ class VideoFrameTracker {
   std::list<int64_t> frames_to_be_rendered_;
 
   const int max_pending_frames_size_;
+  const bool ignore_stale_rendered_frames_after_seek_;
+  bool awaiting_first_frame_after_seek_ = false;
   int dropped_frames_ = 0;
   int64_t seek_to_time_ = 0;  // microseconds
 
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -42,6 +42,7 @@
 #include "starboard/decode_target.h"
 #include "starboard/drm.h"
 #include "starboard/shared/starboard/experimental_features.h"
+#include "starboard/shared/starboard/features.h"
 #include "starboard/shared/starboard/media/media_tracing.h"
 #include "starboard/shared/starboard/media/mime_type.h"
 #include "starboard/shared/starboard/player/filter/video_frame_internal.h"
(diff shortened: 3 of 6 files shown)


starboard/android/shared/video_frame_tracker.cc
```diff
@@ -159,8 +162,14 @@ void VideoFrameTracker::UpdateDroppedFrames() {
...
           kMaxAllowedSkew) {
         // This frame was rendered, remove it from frames_to_be_rendered_.
+        awaiting_first_frame_after_seek_ = false;
         to_render_timestamp = frames_to_be_rendered_.erase(to_render_timestamp);
       } else if (rendered_timestamp - *to_render_timestamp > kMaxAllowedSkew) {
+        if (awaiting_first_frame_after_seek_) {
+          // Ignore stale rendered frame callbacks from flushed pre-seek frames
+          // until the first post-seek frame is rendered.
+          break;
+        }
```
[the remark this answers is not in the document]
borongc: Done.

starboard/extension/feature_config.h
```diff
@@ -140,6 +140,12 @@ STARBOARD_FEATURE(kForceResetAudioDecoder, "ForceResetAudioDecoder", false)
 // enabling tunnel mode on all playbacks.
 STARBOARD_FEATURE(kForceTunnelMode, "ForceTunnelMode", false)
 
+// Set the following variable to true to ignore stale rendered frame callbacks
+// after seeking until the first post-seek frame is rendered.
+STARBOARD_FEATURE(kIgnoreStaleRenderedFramesAfterSeek,
```
kjyoun: I dont think we need this, since you already defines ExperimentalFeature flag for h5vcc?

Is there any reason to define starboard feature flag?

starboard/android/shared/video_frame_tracker.cc
```diff
@@ -159,8 +162,27 @@ void VideoFrameTracker::UpdateDroppedFrames() {
...
           kMaxAllowedSkew) {
         // This frame was rendered, remove it from frames_to_be_rendered_.
+        awaiting_first_frame_after_seek_ = false;
         to_render_timestamp = frames_to_be_rendered_.erase(to_render_timestamp);
       } else if (rendered_timestamp - *to_render_timestamp > kMaxAllowedSkew) {
+        if (awaiting_first_frame_after_seek_) {
+          bool has_match = false;
+          size_t count = 0;
+          const size_t kMaxSearchDepth = 2;
+          for (auto it = to_render_timestamp;
```
kjyoun: this introduced double-loop?

Is it intended? 

Is there any way to avoid double-loop?

jasonzhangxx: Second on that. It's better to avoid negative performance impact for over reported dropped frames issue.

starboard/extension/feature_config.h
borongc: Right, this is for my local testing purposes, and it's easy for me to compare the performance locally.

starboard/android/shared/video_frame_tracker.cc
borongc: Removed and refactored it. This only checks the first 2 frames post seek.

starboard/extension/feature_config.h
kjyoun: For local testing, now you can use "--enable-h5vcc-settings=" instead of enable-features

```
$ adb shell am start  -n dev.cobalt.coat/dev.cobalt.app.MainActivity \
  --esa commandLineArgs '--enable-h5vcc-settings=Media.IgnoreStaleRenderedFramesAfterSeek=1'
```

Do you still need starboard feature?

borongc: Removed the starboard feature as you stated.

starboard/android/shared/video_frame_tracker.cc
```diff
@@ -163,8 +167,28 @@ void VideoFrameTracker::UpdateDroppedFrames() {
       if (std::abs(*to_render_timestamp - rendered_timestamp) <=
           kMaxAllowedSkew) {
         // This frame was rendered, remove it from frames_to_be_rendered_.
+        awaiting_first_frame_after_seek_ = false;
         to_render_timestamp = frames_to_be_rendered_.erase(to_render_timestamp);
       } else if (rendered_timestamp - *to_render_timestamp > kMaxAllowedSkew) {
```
kjyoun: I think that just else is enough.

gap = *to_render_timestamp - rendered_timestamp

to enter while loop: gap < kMaxSkew

first if check:  -kMaxSkew  <= gap  <=  kMaxSkew 

then else mean: gap < -kMaxSkew  == (rendered_timestamp - *to_render_timestamp > kMaxAllowedSkew)

starboard/android/shared/video_frame_tracker.cc
```diff
@@ -163,8 +167,28 @@ void VideoFrameTracker::UpdateDroppedFrames() {
...
+        if (awaiting_first_frame_after_seek_) {
+          // Check up to the first 2 expected frames post-seek to tolerate up to
+          // 1 legitimate startup frame drop during decoder preroll. If a match
+          // is found, proceed with normal frame tracking; otherwise, ignore
+          // the callback as a stale pre-seek frame.
+          auto end_it = std::next(
+              to_render_timestamp,
+              std::min<size_t>(2, std::distance(to_render_timestamp,
```
kjyoun: can you define const for 2 and add why 2 is chosen?

- is 2 enough value?
- what happens if more than 2 frame drops?

e.g.,
```
// No more than 2 frames dropped after seek. ?
constexpr int kMaxDropFrameAfterSeek = 2;

```

starboard/android/shared/video_frame_tracker.cc
borongc: Let's do this by another PR to refactor the code.

starboard/android/shared/video_frame_tracker.cc
borongc: As frames could be dropped after seek, using 2 is a conservative window. We could increase it to a higher number, but this could impact the performance with a little benefit on those devices having stale MediaCodec callbacks.

As a result, I put 2 here, which should eliminate the most of the cases (we can verify it from experiments).

starboard/android/shared/video_frame_tracker.cc
borongc: This is handled by https://github.com/youtube/cobalt/pull/11599.

starboard/android/shared/video_frame_tracker.cc
borongc: Let me do a local test again to see if 2 is a reasonable choice.

borongc: Checked it, and using 1st frame is sufficient. So, I removed this complex check.

starboard/android/shared/video_frame_tracker.h
```diff
@@ -25,8 +25,12 @@ namespace starboard {
 
 class VideoFrameTracker {
  public:
-  explicit VideoFrameTracker(int max_pending_frames_size)
-      : max_pending_frames_size_(max_pending_frames_size) {}
+  explicit VideoFrameTracker(
+      int max_pending_frames_size,
+      bool ignore_stale_rendered_frames_after_seek = false)
```
kjyoun, line 30: do you need this default value?

looks like all callsite already call this ctor with `ignore_stale_rendered_frames_after_seek`

kjyoun (approved):

starboard/android/shared/video_frame_tracker.h
borongc: Removed.

borongc: It is actually used in starboard/android/shared/video_frame_tracker_test.cc:

```
2026-09-01T22:12:25.6628699Z ../../starboard/android/shared/video_frame_tracker_test.cc:25:21: error: no matching constructor for initialization of 'VideoFrameTracker'
2026-09-01T22:12:25.6629714Z    25 |   VideoFrameTracker video_frame_tracker(100);
```

As a result, keep it as it is.

PR #11382, opened 2026-07-14
Add interactive HTML heap memory flamegraph visualizer

haozheng-cobalt:
Introduce a script to convert Cobalt heaps_v2 memory dumps into
interactive HTML flamegraphs. This provides a standalone visualization
tool that supports differential snapshot comparison, allocator
filtering, and symbol search.

Integration with the symbolization script allows users to generate
these reports directly during the symbolization process, facilitating
easier identification of memory trends and potential leaks in a
standard web browser.

Example Result: 
command:
```
python3 cobalt/tools/performance/memory/symbolize_in_process_heap.py /tmp/c27_trace.json -l out/android-arm_devel/lib.unstripped/libchrobalt.so --export_html /tmp/cobalt_memory_flamegraph.html
```
Track memory among different snapshots: https://screenshot.googleplex.com/BXeuAMDNJovXXFj
View full callstacks in the trace: https://screenshot.googleplex.com/4RwtDjXszUCKnAo
Compare heap memory delta between any 2 snapshots: https://screenshot.googleplex.com/39yzdWhdJzfFQd6
Bug: 534880200

### DIFF
--- a/cobalt/tools/performance/memory/convert_heaps_v2_to_html.py
+++ b/cobalt/tools/performance/memory/convert_heaps_v2_to_html.py
@@ -1,0 +1,626 @@
+#!/usr/bin/env python3
+# Copyright 2026 The Cobalt Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Converts Cobalt heaps_v2 memory dumps into an interactive HTML flamegraph.
+
+Generates a standalone, feature-rich HTML report with interactive SVG
+flamegraphs, module color hashing, differential snapshot comparison (delta diff
+mode), allocator tab filtering, minimum size thresholding, and symbol search.
+Open the resulting .html file directly in any modern browser.
+"""
+
+import argparse
+import json
+import os
+import sys
+from typing import Any, Dict, List, Tuple
+
+
+def parse_heaps_v2_snapshots(
+    trace_path: str,) -> Tuple[List[Tuple[int, Dict[str, Any]]], bool]:
+  """Extracts heaps_v2 snapshots from a JSON trace file."""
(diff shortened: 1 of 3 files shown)


cobalt/tools/performance/memory/symbolize_in_process_heap.py
```diff
@@ -58,6 +67,19 @@ def main():
       "--symbolizer_path",
       help=("Path to llvm-symbolizer. If omitted, will auto-detect "
             "inside the toolchain, falling back to system PATH."))
+  parser.add_argument(
+      "-p",
+      "--export_perfetto",
```
Awallky: This seems to be never used. Does this need to be implemented or should we remove it?

cobalt/tools/performance/memory/convert_heaps_v2_to_html.py
```diff
@@ -0,0 +1,616 @@
...
+      "    function showTooltip(e, node, totalVal, isDiffMode) {\n"
+      "      const tt = document.getElementById('tooltip');\n"
+      "      const pct = ((node.value / (totalVal || 1)) * 100).toFixed(2);\n"
+      "      if (isDiffMode) {\n"
+      "        const sign = node.diffValue >= 0 ? '+' : '';\n"
+      "        const clr = node.diffValue >= 0 ?\n"
+      "            'var(--red)' : 'var(--green)';\n"
+      "        const tgtStr = formatBytes(node.targetValue || 0);\n"
+      "        const baseStr = formatBytes(node.baseValue || 0);\n"
+      "        tt.innerHTML = `<b>${node.name}</b><br>` +\n"
```
Awallky: Since this visualizes C++ stacks,  node.name  will likely contain C++ template types with angle brackets (e.g.,  std::vector<int> ). Because you're assigning directly to innerHTML , the browser will interpret  <int>  as an unrecognized HTML tag, which causes the template parameters to silently disappear from the tooltip's display and potentially break the styling.

I'd recommend sanitizing this input so this does not occur.

Awallky (approved):
This looks really cool! Thanks!
LGTM % comments

cobalt/tools/performance/memory/symbolize_in_process_heap.py
haozheng-cobalt: thx, removed the unused --export_perfetto flag.

cobalt/tools/performance/memory/convert_heaps_v2_to_html.py
haozheng-cobalt: Introduced an escapeHtml(str) helper function in JavaScript to sanitize < to &lt; and > to &gt; prior to assigning to innerHTML.

PR #11390, opened 2026-07-15
aosp: Add GN rules to build APKs

hferreiro:
Add the build rules that package a modular executable into an Android APK for
the AOSP platform. This allows to create APKs for all test and
modular_executable targets so that building <target>_loader builds
<target>.apk.

The APK bundles the evergreen loader as its native library (loader_app for
cobalt and  elf_loader_sandbox for tests) with the lz4-compressed inner library
and its content as assets; MainActivity loads the loader and spawns the
Starboard thread through a jni_zero binding.

Most of the Java code is reused from AndroidTV, being the main exception
separate MainActivity and CobaltApplication classes.

Some non-working targets are currently excluded.

Bug: 495203133

### DIFF
--- a/.github/config/aosp-arm.json
+++ b/.github/config/aosp-arm.json
@@ -10,7 +10,7 @@
     "gold"
   ],
   "targets": [
-    "buildflag_header_h"
+    "nplb_loader"
   ],
   "includes": [
     {
--- a/.github/config/aosp-arm.json
+++ b/.github/config/aosp-arm.json
@@ -10,6 +10,7 @@
     "gold"
   ],
   "targets": [
+    "cobalt:gn_all",
     "nplb_loader"
   ],
   "includes": [
--- a/.github/config/aosp-arm.json
+++ b/.github/config/aosp-arm.json
@@ -10,8 +10,7 @@
     "gold"
   ],
   "targets": [
-    "cobalt:gn_all",
-    "nplb_loader"
+    "cobalt:gn_all"
   ],
   "includes": [
     {
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -1,9 +1,8 @@
 import("//build/config/android/config.gni")
 import("//build/config/android/rules.gni")
+import("//cobalt/android/manifest.gni")
 import("//third_party/icu/config.gni")
 import("//third_party/jni_zero/jni_zero.gni")
-
-cobalt_manifest = "$target_gen_dir/cobalt_manifest/AndroidManifest.xml"
 
 jinja_template("cobalt_manifest") {
   input = "//cobalt/shell/android/shell_apk/AndroidManifest.xml.jinja2"
@@ -31,7 +30,6 @@ generate_jni("jni_headers") {
     "apk/app/src/main/java/dev/cobalt/coat/CobaltService.java",
     "apk/app/src/main/java/dev/cobalt/coat/CobaltSystemConfigChangeReceiver.java",
     "apk/app/src/main/java/dev/cobalt/coat/CobaltTextToSpeechHelper.java",
-    "apk/app/src/main/java/dev/cobalt/coat/PlatformError.java",
     "apk/app/src/main/java/dev/cobalt/coat/ResourceOverlay.java",
     "apk/app/src/main/java/dev/cobalt/coat/javabridge/HTMLMediaElementExtension.java",
     "apk/app/src/main/java/dev/cobalt/features/StarboardFeatureList.java",
@@ -53,6 +51,10 @@ generate_jni("jni_headers") {
       "apk/app/src/main/java/dev/cobalt/browser/crashannotator/CrashAnnotatorImplFirstParty.java",
     ]
   }
+
+  if (!is_starboard) {
+    sources += [ "apk/app/src/main/java/dev/cobalt/coat/PlatformError.java" ]
+  }
 }
 
 generate_jni("browser_jni_headers") {
@@ -72,21 +74,9 @@ android_library("cobalt_browser_java") {
 android_library("cobalt_main_java") {
   resources_package = "dev.cobalt.coat"
   deps = [
-    ":cobalt_browser_java",
     "//base:base_java",
-    "//base:process_launcher_java",
-    "//base/version_info/android:version_constants_java",
     "//build/android:build_java",
-    "//cobalt/shell/android:cobalt_shell_assets",
-    "//cobalt/shell/android:cobalt_shell_java",
     "//cobalt/shell/android:cobalt_shell_java_resources",
-    "//cobalt/shell/android:cobalt_shell_jni_headers",
-    "//components/embedder_support/android:view_java",
-    "//content/public/android:content_java",
-    "//media/capture/video/android:capture_java",
-    "//net/android:net_java",
-    "//services/media_session/public/cpp/android:media_session_java",
-    "//services/media_session/public/mojom:mojom_java",
     "//third_party/android_deps:com_google_code_findbugs_jsr305_java",
     "//third_party/android_deps:google_play_services_ads_identifier_java",
     "//third_party/android_deps:google_play_services_basement_java",
@@ -95,8 +85,6 @@ android_library("cobalt_main_java") {
     "//third_party/androidx:androidx_customview_customview_java",
     "//third_party/androidx:androidx_media_media_java",
     "//third_party/jni_zero:jni_zero_java",
-    "//ui/android:ui_java",
-    "//url:gurl_java",
   ]
 
   if (!use_evergreen) {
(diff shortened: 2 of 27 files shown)


cobalt/aosp/apk/app/java/dev/cobalt/app/MainActivity.java
```diff
@@ -0,0 +1,99 @@
...
+      // Cold start - Instantiate the singleton BaseStarboardBridge.
+      BaseStarboardBridge starboardBridge = createStarboardBridge(getArgs(), startDeepLink);
+      ((BaseStarboardBridge.HostApplication) getApplication()).setStarboardBridge(starboardBridge);
+    } else {
+      // Warm start - Pass the deep link to the running Starboard app.
+      getStarboardBridge().handleDeepLink(startDeepLink);
+    }
+
+    // Spawn the loader thread.
+    MainActivityJni.get().startLoader();
```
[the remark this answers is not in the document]
hferreiro: Done.

cobalt/aosp/apk/app/java/dev/cobalt/app/MainActivity.java
```diff
@@ -0,0 +1,94 @@
...
+@JNINamespace("starboard")
+public class MainActivity extends BaseCobaltActivity {
+
+  static {
+    // Each aosp apk ships exactly one app loader under its natural name
+    // via android_apk shared_libraries: libloader_app.so for the production
+    // app or libelf_loader_sandbox.so for everything else. Load whichever is
+    // present.
+    UnsatisfiedLinkError loadError = null;
+    for (String lib : new String[] {"loader_app", "elf_loader_sandbox"}) {
```
hferreiro, line 43: An alternative here is to copy the current app loader to something like libcobalt_loader.so and just use "cobalt_loader".

starboard/aosp/shared/main_activity.cc
```diff
@@ -0,0 +1,59 @@
...
+  }
+};
+
+}  // namespace
+
+namespace starboard {
+
+void JNI_MainActivity_StartLoader(JNIEnv* env) {
+  static StarboardMainDelegate delegate;
+  base::PlatformThread::CreateNonJoinable(0, &delegate);
```
hferreiro: I forgot to update this, I guess it's better if we don't depend on //base, right?

andrewsavage1: That's right

starboard/elf_loader/elf_loader_sandbox.cc
```diff
@@ -51,7 +51,9 @@ void LoadLibraryAndInitialize(const std::string& library_path,
                   << "=path/to/content/relative/to/loader/content.";
     return;
   }
-  if (!g_elf_loader.Load(library_path, content_path, true)) {
+  if (!g_elf_loader.Load(library_path, content_path, /*is_relative_path=*/true,
+                         /*custom_get_extension=*/nullptr,
+                         elf_loader::CompressionType::kLz4)) {
```
hferreiro: This will break other platforms. I'll work on a fix.

.github/config/aosp-arm.json
```diff
@@ -10,7 +10,7 @@
     "gold"
   ],
   "targets": [
-    "buildflag_header_h"
+    "nplb_loader"
```
andrewsavage1: Are we able to build cobalt_loader too or not yet?

starboard/android/shared/BUILD.gn
```diff
@@ -318,19 +317,35 @@ static_library("starboard_platform") {
...
       "crash_handler.h",
+      "system_platform_error.cc",
     ]
     sources += [
       "//starboard/shared/starboard/crash_handler.cc",
       "//starboard/shared/starboard/crash_handler.h",
       "//starboard/shared/starboard/loader_app_metrics.cc",
       "//starboard/shared/starboard/loader_app_metrics.h",
+
+      # TODO(https://crbug.com/495203133): Symbols required on AOSP. The
```
andrewsavage1, line 332: Note I expect a number of these to be necessary for Cobalt functionality/YTS tests, but fine to address that later

andrewsavage1 (commented):
The basic structure/approach looks correct, I'll do a deep review tomorrow morning

cobalt/android/BUILD.gn
```diff
@@ -182,6 +161,41 @@ android_library("cobalt_main_java") {
...
     ]
   }
+
+  if (is_starboard) {
+    # TODO(https://crbug.com/534656433): Temporary StarboardBridge stub to fix
+    # a symbol resolution error on BaseStarboardBridge and CobaltService.
+    sources +=
+        [ "//cobalt/aosp/apk/app/java/dev/cobalt/coat/StarboardBridge.java" ]
+  } else {
+    sources += [
```
andrewsavage1, line 171: If the plan is to add these back to AOSP later, this lgtm. If we're permanently removing them, then we'll want to think about making an entirely separate AOSP target

cobalt/aosp/modular_apk.gni
```diff
@@ -0,0 +1,96 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import("//build/config/android/rules.gni")
+import("//cobalt/android/manifest.gni")
+
+template("modular_apk") {
+  _name = invoker.module_name
```
andrewsavage1, line 19: What's the intent of having prefixed underscores here?

cobalt/aosp/modular_apk.gni
```diff
@@ -0,0 +1,96 @@
...
+      ":copy_${_name}_lib($default_toolchain)",
+    ]
+
+    if (_name == "cobalt") {
+      sources = [ "$root_build_dir/cobalt_shell.pak" ]
+      deps += [ "//cobalt/shell:pak($default_toolchain)" ]
+    }
+
+    if (_name == "nplb") {
+      _file_tests_data = [
```
andrewsavage1: I'm worried about this, because this is a pattern we would have to repeat for all data of every target we want to run. This information is available via runtime_deps for sure, and probably available earlier than that as well.

We can submit this as is for now but should try to address it a different way in the future

cobalt/aosp/modular_apk.gni
```diff
@@ -0,0 +1,96 @@
...
+      "app/cobalt/lib/${_lib_name}.lz4",
+    ]
+
+    deps = [
+      ":copy_${_name}_empty_fonts($default_toolchain)",
+      ":copy_${_name}_lib($default_toolchain)",
+    ]
+
+    if (_name == "cobalt") {
+      sources = [ "$root_build_dir/cobalt_shell.pak" ]
```
andrewsavage1, line 48: I believe other targets can depend on the pak file too

cobalt/aosp/modular_apk.gni
hferreiro: I'm following https://chromium.googlesource.com/chromium/src/+/main/build/docs/writing_gn_templates.md#variables. Other that the explanation there, "private" variables are not imported from .gni files, and they aren't forwarded when using "*".

.github/config/aosp-arm.json
hferreiro: Added more targets.

starboard/elf_loader/elf_loader_sandbox.cc
hferreiro: Done.

cobalt/aosp/modular_apk.gni
andrewsavage1: Oh nice, I haven't seen that guide before, we can keep it

cobalt/aosp/modular_apk.gni
hferreiro: You mean we can keep it unguarded for simplicity? Other targets won't need it.

cobalt/android/BUILD.gn
hferreiro: Some of those file won't be used by AOSP. There's a precedent at starboard/android/shared/BUILD.gn where is_starboard used, though maybe that's just legacy from the C25 era.

cobalt/aosp/modular_apk.gni
hferreiro: For the time being, I'll package it as an android_assets target instead, to avoid duplication.

starboard/aosp/shared/main_activity.cc
hferreiro: Done.

andrewsavage1 (approved):
Comments are relatively minor and can be addressed in follow ups

johnxwork (approved):

PR #11105, opened 2026-06-27
cobalt: Make HangWatcher config dynamically updatable

alunev:
This introduces dynamic, runtime configuration for base::HangWatcher
in Cobalt, allowing parameters to be tuned via JavaScript
(h5vcc.settings) without requiring a process restart or recompilation:
- The web app updates hangwatcher settings in GlobalFeatures (via web API)
- When a hangwatcher setting is updated in GlobalFeatures, hangwatcher updates its cached values of the settings by asking cobalt's hangwatcher delegate for the settings
- Cobalt's hangwatcher delegate gets its setting values from GlobalSettings
- Hangwatcher uses its cached settings when making decisions about threads it's monitoring

Parameters added:
- HangWatchTimeSeconds: Configures the timeout threshold for hang
  detection.
- HangWatchMonitoringPeriodSeconds: Configures how often the
  background thread wakes up to check for hangs.
- EnableHangWatchMainThreadDump / IOThreadDump / ThreadPoolDump /
  RendererThreadDump: Allows selectively enabling or disabling crash
  dumps for specific thread scopes.

Issue: 517626586

PR #11391, opened 2026-07-15
cobalt: Build HangWatcherDelegateImpl unit tests conditionally

rakuco:
Follow-up to #11105. Use the same checks that determine if the main `hang_watcher_delegate_impl*` files are built to add the unit tests.

This fixes the tvOS `cobalt_unittests` build.

Bug: 517626586


rakuco:

rakuco:
@linxinan-yt is `cobalt_unittests` part of the targets built in CI? If not, would it be possible to add it?

hlwarriner (approved):

fayev-g (approved):

PR #11405, opened 2026-07-16
Add color legend and explanation to the native heap visualizer

haozheng-cobalt:
Update the native C++ heap allocation visualizer to improve clarity
and usability. This change introduces a categorized color palette
for major subsystems like V8, Blink, and Skia, making it easier to
identify memory contributors at a glance.

The generated HTML report now includes a legend and an interpretation
guide. Additionally, the default view is updated to show all
allocators by default. Documentation is expanded in the performance
READMEs to provide a clear end-to-end workflow for memory profiling.

Try this interactive Cobalt heap report here: https://drive.google.com/file/d/1IB921haCmV1hAzhtZd2MxHrCLwt_KDps/view?usp=drive_link, download the html file and open in Chrome.
Bug: 534880200


Awallky (approved):
LGTM

You were asking if these results looked reasonable, I would say use some system tools to verify that these are reasonable. 
For the malloc levels and partition_alloc levels, you can directly check smaps to understand if these memory measurements look reasonable. 

See cobalt/tools/performance/smaps/README.md for our current tool that captures smaps metrics. We also have a tool that compares our UMA measurements to the smaps measurements - cobalt/tools/performance/memory/compare_accuracy.py. You could use something similar to this to see how close these profiling measurements are to the smaps measurements.

PR #11407, opened 2026-07-16
starboard: Enhance cert scope read for kSbSystemPropertyCertificationScope

abellin7:
Read certification scope first from /etc/device.properties. If it fails
to read, the reading falls back to COBALT_CERT_SCOPE_FILE_NAME environment
variable.

Bug: 477465556

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_property.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_property.cc
@@ -61,7 +61,7 @@ bool CopyStringAndTestIfSuccess(char* out_value,
   return true;
 }
 
-bool TryReadFromPropertiesFile(const char* prefix, size_t prefix_len, char* out_value, size_t value_length) {
+bool TryReadFromPropertiesFile(const char* prefix, size_t prefix_len, char* out_value, size_t value_length, bool upper_case = true) {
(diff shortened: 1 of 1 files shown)


starboard/contrib/rdk/src/third_party/starboard/rdk/shared/system/system_get_property.cc
```diff
@@ -215,30 +218,40 @@ bool GetCertificationScope(char* out_value, int value_length) {
...
+  };
+
+  if (get_cert_from_file()) {
+    return true;
   }
-  buf[sz] = 0;
 
-  SB_LOG(INFO) << "Device cert scope: '" << buf.data() << "'";
-  return starboard::strlcpy<char>(out_value, buf.data(), value_length);
+  static constexpr char kPrefixStr[] = "CERT_SCOPE=";
```
jonastsai: considering to move these tree lines to line 219 when !file.IsValid() is true and keep the original logic

abellin7: I simplified it and kept the changes to a minimum for easier comparison.

jonastsai:
since you checked the cert_scope from property file, remember to update the commit accordingly.

abellin7:
> since you checked the cert_scope from property file, remember to update the commit accordingly.

The description is revised.

jonastsai (approved):

PR #11410, opened 2026-07-16
Add cobalt_shell.pak to loader_app data_deps

AIdrifter:
Add //cobalt/shell:pak to data_deps of loader_app target in starboard/loader_app/BUILD.gn so that building loader_app automatically generates cobalt_shell.pak under root_build_dir for Cobalt 27 Evergreen deployment.

Bug: 532320452

### DIFF
--- a/starboard/loader_app/BUILD.gn
+++ b/starboard/loader_app/BUILD.gn
@@ -71,7 +71,10 @@ if (is_starboard && current_toolchain == starboard_toolchain) {
     output_dir = root_build_dir
     if (target_cpu == "x64" || target_cpu == "arm" || target_cpu == "arm64") {
       deps = [ ":loader_app_sources" ]
-      data_deps = [ "//starboard/content/fonts:copy_font_data" ]
+      data_deps = [
+        "//cobalt/shell:pak",
+        "//starboard/content/fonts:copy_font_data",
+      ]
     }
   }
 } else {


andrewsavage1:
This doesn't seem right to me. The loader_app has no explicit dependency on things in the cobalt_shell.pak file, only the things it loads will have that dependency

AIdrifter:
Hi @andrewsavage1 

> This doesn't seem right to me. The loader_app has no explicit dependency on things in the cobalt_shell.pak file, only the things it loads will have that dependency

Thanks for the feedback! 

The actual issue reported by the vendor is that when they download the official prebuilt CRX package (cobalt_evergreen_7.1.2_arm64_sbversion-18_release_compressed_20260627021609.crx) from GitHub Releases and unzip it, cobalt_shell.pak is missing inside the package, causing Cobalt to crash on launch.

Since package.json specifies cobalt_shell.pak -> content/cobalt_shell.pak, it seems the CRX release build target/pipeline missed building //cobalt/shell:pak before assembling the .crx zip archive.

What would be the proper GN packaging target or pipeline configuration to ensure //cobalt/shell:pak is built and included in the release CRX artifacts?

https://b.corp.google.com/u/0/issues/530398925

AIdrifter:
After confirming with the internal team, cobalt_shell.pak is already included in the [cobalt_evergreen_7.1.2_arm64_sbversion-18_release_compressed_20260627021609.crx](https://github.com/youtube/cobalt/releases/download/27.lts.1/cobalt_evergreen_7.1.2_arm64_sbversion-18_release_compressed_20260627021609.crx) file, so we don't need this modification.

PR #11413, opened 2026-07-16
Load SSL root certificates to avoid runtime wipe crashes

yuying-y:
This pull request loads the SSL root certificates at startup directly into the bssl trust store. It also safely guards TryLoadCert against triggering NOTREACHED() if a certificate file fails to load at runtime.

This is a new issue with C26/C27 because we switched to 2-slot, one system image, and one installation slot. When the app is updated from system image and runs the package from the installation slot, right before it downloads an update, it'll wipe the installation slot including its own certs, which caused the issue. But before C26, there're 3 slots in total, and the updater doesn't wipe its own slot.

On the other hand, the core content (cobalt_shell.pak) and UI fonts (SkFileMemoryChunkStream) are safe from this wipe because they use base::MemoryMappedFile and hold a persistent FILE* handle, which the OS honors natively.

Issue: 535681824
Issue: 533843031

### DIFF
--- a/net/cert/internal/trust_store_in_memory_starboard.cc
+++ b/net/cert/internal/trust_store_in_memory_starboard.cc
@@ -16,19 +16,16 @@
 
 #include <dirent.h>
 
-#include "base/files/file.h"
 #include "base/files/file_path.h"
+#include "base/files/file_util.h"
 #include "base/logging.h"
 #include "base/notreached.h"
 #include "base/path_service.h"
 #include "base/time/time.h"
-#include "net/cert/internal/trust_store_in_memory_starboard.h"
 #include "net/cert/x509_certificate.h"
 #include "net/cert/x509_util.h"
 #include "starboard/common/string.h"
 #include "starboard/configuration_constants.h"
-#include "third_party/boringssl/src/include/openssl/digest.h"
-#include "third_party/boringssl/src/include/openssl/sha.h"
 #include "third_party/boringssl/src/include/openssl/x509.h"
 #include "third_party/boringssl/src/pki/cert_errors.h"
 #include "third_party/boringssl/src/pki/pem.h"
@@ -36,30 +33,11 @@
 namespace net {
 
 namespace {
-// PEM encoded DER cert is usually around or less than 2000 bytes long.
-const short kCertBufferSize = 8 * 1024;
 // Each certificate file name is 8 bit hash + ".0" suffix.
 const short kCertFileNameLength = 10;
 const char kCertificateHeader[] = "CERTIFICATE";
 const char kSSLDirName[] = "ssl";
 const char kCertsDirName[] = "certs";
-
-// Essentially an X509_NAME_hash without using X509_NAME. We did not use the
-// boringSSL function directly because net does not store certs with X509
-// struct anymore and converting binary certs to X509_NAME is slightly
-// expensive.
-unsigned long CertNameHash(const void* data, size_t length) {
-  unsigned long ret = 0;
-  unsigned char md[SHA_DIGEST_LENGTH];
-  if (!EVP_Digest(data, length, md, NULL, EVP_sha1(), NULL)) {
-    return 0;
-  }
-
-  ret = (((unsigned long)md[0]) | ((unsigned long)md[1] << 8L) |
-         ((unsigned long)md[2] << 16L) | ((unsigned long)md[3] << 24L)) &
-        0xffffffffL;
-  return ret;
-}
 
 base::FilePath GetCertificateDirPath() {
   base::FilePath cert_path;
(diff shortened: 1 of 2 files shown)


yuying-y:
Addressed by switching to `base::ReadFileToString` and resolving header/formatting nits in the latest commit.

andrewsavage1:
Do you know how the cert would fail to load at runtime before? It seems like that would be the thing we'd want to fix instead of loading all the certs at startup, which seems like it could be quite an expensive task

yuying-y:
> Do you know how the cert would fail to load at runtime before? It seems like that would be the thing we'd want to fix instead of loading all the certs at startup, which seems like it could be quite an expensive task

This is a new issue with C26/C27 because we switched to 2-slot, one system image, and one installation slot. When the app is updated from system image and runs the package from the installation slot, when it downloads an update, it'll wipe the installation slot including its own certs, which caused the issue. But before C26, there're 3 slots in total, and the updater doesn't wipe its own slot.

andrewsavage1:
> > Do you know how the cert would fail to load at runtime before? It seems like that would be the thing we'd want to fix instead of loading all the certs at startup, which seems like it could be quite an expensive task
> 
> This is a new issue with C26/C27 because we switched to 2-slot, one system image, and one installation slot. When the app is updated from system image and runs the package from the installation slot, when it downloads an update, it'll wipe the installation slot including its own certs, which caused the issue. But before C26, there're 3 slots in total, and the updater doesn't wipe its own slot.

Ahh, that makes a lot of sense—would you update the PR description with some of this information? It's quite helpful so I think we'll want it in the actual commit message

To address that problem, this seems like a reasonable solution for the certs, but I wonder if we need a more general solution to make sure we maintain all content from the previous installation somehow (either in memory or on disk). Right now I don't think we have any safeguards to make sure we don't run into this problem for other content.

yuying-y:
> > > Do you know how the cert would fail to load at runtime before? It seems like that would be the thing we'd want to fix instead of loading all the certs at startup, which seems like it could be quite an expensive task
> > 
> > 
> > This is a new issue with C26/C27 because we switched to 2-slot, one system image, and one installation slot. When the app is updated from system image and runs the package from the installation slot, when it downloads an update, it'll wipe the installation slot including its own certs, which caused the issue. But before C26, there're 3 slots in total, and the updater doesn't wipe its own slot.
> 
> Ahh, that makes a lot of sense—would you update the PR description with some of this information? It's quite helpful so I think we'll want it in the actual commit message
> 
> To address that problem, this seems like a reasonable solution for the certs, but I wonder if we need a more general solution to make sure we maintain all content from the previous installation somehow (either in memory or on disk). Right now I don't think we have any safeguards to make sure we don't run into this problem for other content.

Updated the commit message. I asked Jetski about the other content. The reply is they're not impacted because they use base::MemoryMappedFile and hold a persistent FILE* handle, which the OS honors natively. See details in b/535681824.

net/cert/internal/trust_store_in_memory_starboard.cc
```diff
@@ -102,90 +81,57 @@ std::unordered_set<std::string> GetCertNamesOnDisk() {
     if (strlen(dir_entry.data()) != kCertFileNameLength) {
       continue;
     }
-    trusted_certs_on_disk.emplace(dir_entry.data());
+    
+    base::FilePath cert_path = GetCertificateDirPath().Append(dir_entry.data());
+    std::string cert_buffer;
+    if (!base::ReadFileToString(cert_path, &cert_buffer)) {
+      DLOG(ERROR) << "Failed to read cert file: " << cert_path;
```
andrewsavage1, line 88: Why do you want to make the assertions here and below weaker? Having harder checks as we had before (using CHECK/NOTREACHED) means the code will be easier to debug if a necessary cert file does not get loaded correctly

net/cert/internal/trust_store_in_memory_starboard.cc
```diff
@@ -102,90 +81,57 @@ std::unordered_set<std::string> GetCertNamesOnDisk() {
...
+TrustStoreInMemoryStarboard::TrustStoreInMemoryStarboard() {
+  for (const auto& cert : GetAllCertsOnDisk()) {
+    underlying_trust_store_.AddTrustAnchor(cert);
   }
-  int cert_size = cert_file.ReadAtCurrentPos(cert_buffer, kCertBufferSize);
-  bssl::PEMTokenizer pem_tokenizer(std::string_view(cert_buffer, cert_size),
-                                   {kCertificateHeader});
-  pem_tokenizer.GetNext();
-  std::string decoded(pem_tokenizer.data());
-  DCHECK(!pem_tokenizer.GetNext());
```
andrewsavage1, line 139: We should add back the check that there is only one cert per pem file

andrewsavage1 (approved):
We can submit this now to get it into RC2 but should do a fast follow to address my comments

PR #11417, opened 2026-07-17
Fix QuotaDatabase transaction commit/begin crash

yuying-y:
Fixes DCHECK crash in quota_database.

Issue: 536038806

### DIFF
--- a/storage/browser/quota/quota_database.cc
+++ b/storage/browser/quota/quota_database.cc
@@ -975,11 +975,36 @@ void QuotaDatabase::Commit() {
 
   last_operation_ = "Commit";
   DCHECK_EQ(1, db_->transaction_nesting());
+#if BUILDFLAG(IS_STARBOARD)
+  if (!db_->CommitTransactionDeprecated()) {
+    LOG(ERROR) << "Failed to commit QuotaDatabase transaction, disabling database.";
+    DisableDatabase();
+    return;
+  }
+#else
   db_->CommitTransactionDeprecated();
+#endif
   DCHECK_EQ(0, db_->transaction_nesting());
+#if BUILDFLAG(IS_STARBOARD)
+  if (!db_->BeginTransactionDeprecated()) {
+    LOG(ERROR) << "Failed to start a new transaction for QuotaDatabase, disabling database.";
+    DisableDatabase();
+    return;
+  }
+#else
   db_->BeginTransactionDeprecated();
+#endif
   DCHECK_EQ(1, db_->transaction_nesting());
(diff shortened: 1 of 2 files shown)


yuying-y (commented):
All review comments are addressed, and code has been wrapped with BUILDFLAG(IS_STARBOARD) to ease upstream rebasing.

oxve (approved):

storage/browser/quota/quota_database.cc
```diff
@@ -975,12 +975,37 @@ void QuotaDatabase::Commit() {
...
+    return;
+  }
+#else
   db_->BeginTransactionDeprecated();
+#endif
   DCHECK_EQ(1, db_->transaction_nesting());
 }
 
+#if BUILDFLAG(IS_STARBOARD)
+void QuotaDatabase::DisableDatabase() {
```
andrewsavage1, line 1001: The `timer_` should also be stopped here if it is running

storage/browser/quota/quota_database.cc
```diff
@@ -1034,15 +1063,34 @@ QuotaError QuotaDatabase::EnsureOpened() {
...
     return QuotaError::kDatabaseError;
   }
 
   // Start a long-running transaction.
   DCHECK_EQ(0, db_->transaction_nesting());
+#if BUILDFLAG(IS_STARBOARD)
+  if (!db_->BeginTransactionDeprecated()) {
+    LOG(ERROR) << "Could not start initial transaction on quota database, resetting.";
+    if (!is_recreating_ && !db_file_path_.empty()) {
+      if (ResetStorage()) {
```
andrewsavage1, line 1082: Here, we don't signal that the transaction failed but don't rerun it to make sure it succeeds. Should it get rerun here?

storage/browser/quota/quota_database.cc
yuying-y: Good catch, I have stopped the timer in the `DisableDatabase()` helper in PR #11484. Thank you!

storage/browser/quota/quota_database.cc
yuying-y: Actually, `ResetStorage()` recursively calls `EnsureOpened()` under the hood. The nested `EnsureOpened()` sets up a new `db_` instance and successfully runs `db_->BeginTransactionDeprecated()` for us. If we were to call it a second time here, we would crash on the `DCHECK_EQ(0, db_->transaction_nesting())` guard. So the current behavior should be correct without an explicit retry.

PR #11419, opened 2026-07-17
tvOS: Remove extra debugging calls from shell_platform_delegate_ios.mm

rakuco:
These extra `LOG()` calls were erroneously added in #10466 and are not necessary.

Bug: 507159713


rakuco:

fayev-g (approved):

PR #11421, opened 2026-07-17
cobalt: Load empty URL in SplashScreenTest.PreloadSkipsSplashScreen

rakuco:
Instead of creating a new window with "about:blank", pass an empty GURL object to `Shell::CreateNewWindow()` instead.

Passing a non-empty URL causes `CreateNewWindow()` to call `LoadURL()`, which ultimately calls a number of `ShellPlatformDelegate` methods on tvOS that are not mocked (or even virtual). Since the actual URL here is not relevant, just pass an empty GURL to avoid having `LoadURL()` be called altogether.

This fixes a cobalt_unittests failure on tvOS.

Bug: 447660888
Bug: 535573341


rakuco:

rakuco:
Dropping in favor of #11435

PR #11423, opened 2026-07-17
aosp: Run nplb on device in CI

hferreiro:
Enable on-device nplb testing for aosp-arm with a filter for the currently-failing tests.

Mobile Harness requires every target to ship a `<target>_deps.tar.gz` that it
pushes and unpacks on the device. Teach archive_test_artifacts.py to find the
deps file of an Evergreen APK test wrapper, which writes its runtime_deps to a starboard/ subfolder, and to apply the existing exclusions there too. Additionally, exclude some test data that only the player tests, which are filtered out on AOSP, would use.

Name the apk, the deps archive and the filter after the nplb_loader target,
which is the one CI builds and the only one that pulls in the apk and the test
runner wrapper.

Bug: 495203133


cobalt/aosp/modular_apk.gni
```diff
@@ -71,6 +71,10 @@ template("modular_apk") {
       "//components/crash/core/app:chrome_crashpad_handler_named_as_so($native_toolchain)",
     ]
 
+    if (_name == "nplb") {
```
andrewsavage1, line 106: Ideally this would be reusable for all targets using modular_apk, not just nplb, but that's not something we should address here. Let's just add a TODO to extend it

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -35,13 +35,17 @@ runs:
         PLATFORM: ${{ inputs.platform }}
       run: |
         host_archive_flags="--compression zstd --compression-level=9"
+        # The archive is never read on AOSP. Create an empty one instead.
+        aosp_archive_flags="--compression gz --archive-per-target --flatten-deps --empty-archive"
         android_archive_flags="--compression gz --archive-per-target --use-android-deps-path --flatten-deps"
```
andrewsavage1, line 38: Thinking about long term sustainability for the test code, I wonder if it would be better to use a custom runtime_deps file like ATV does in case we don't want to package everything into the apk in the future. For now, we could just compress the apk into the tar.gz file

hferreiro: I fixed this by following what AndroidTV is doing more closely.

cobalt/aosp/modular_apk.gni
hferreiro: This was moved to a precursor.

.github/actions/upload_test_artifacts/action.yaml
```diff
@@ -41,7 +41,9 @@ runs:
 
         device_archive_flags=""
         browser_archive_flags=""
-        if [[ "${PLATFORM}" == *android* ]]; then
+        if [[ "${PLATFORM}" == *aosp* ]]; then
+          device_archive_flags="${android_archive_flags}"
+        elif [[ "${PLATFORM}" == *android* ]]; then
```
andrewsavage1, line 46: nit: this can probably just be `if android || aosp` because it'll likely use the same browser_archive_flags once we have browser tests. If you want to wait on that until we enable those, that's fine too.

.github/config/aosp-arm.json
```diff
@@ -9,6 +9,12 @@
     "qa",
     "gold"
   ],
+  "test_on_device": true,
+  "test_device_family": "android",
+  "test_dimensions": {
+    "device_type": "arm_devices",
```
andrewsavage1, line 15: @isarkis @cvinish is this correct config for the pool you're setting up?

andrewsavage1 (commented):
Will wait until I see a CI run to approve but the code lgtm

andrewsavage1:
fyi @oxve for packaging script changes

.github/config/aosp-arm.json
andrewsavage1: Looks like this ran on a kirkwood, which is fine for now but we should switch to fire tvs once we have the config for that

andrewsavage1 (approved):

.github/actions/upload_test_artifacts/action.yaml
oxve: Definitely a nit but +1

oxve (approved):

cobalt/build/archive_test_artifacts.py
```diff
@@ -252,6 +262,13 @@ def create_archive(
       exclude_dirs = _EXCLUDE_DIRS_DEFAULT
       if is_junit_test:
         exclude_dirs = _EXCLUDE_DIRS_JUNIT
+      elif use_android_deps_path and target_name == 'nplb_loader':
+        # TODO(crbug.com/532068409): remove the test data exclusion and figure
+        # out a workaround for the increase in size.
+        exclude_dirs = _EXCLUDE_DIRS_DEFAULT + [
+            'test/starboard/shared/starboard/player/'
+        ]
+      exclude_dirs = _under_toolchain_subdirs(exclude_dirs)
```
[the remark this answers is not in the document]
hferreiro: They're not tuples.

.github/actions/upload_test_artifacts/action.yaml
hferreiro: Done.

cobalt/testing/filters/aosp-arm/nplb_loader_filter.json
```diff
@@ -0,0 +1,122 @@
...
+    "PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames/107",
+    "PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames/108",
+    "PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames/76",
+    "PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames/77",
+    "PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames/78",
+    "PosixUnistdSysconfTests/PosixUnistdSysconfOptionalValueTest.OptionalScValuesAreValidNames/79",
+    "PosixUsleepTests.ErrorEintr",
+    "SbDrmSessionTest.CloseDrmSessionBeforeUpdateSession",
+    "SbDrmSessionTest.InvalidSessionUpdateRequestParams",
+    "SbDrmSessionTest.SunnyDay",
```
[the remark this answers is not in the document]
hferreiro: Done.

PR #11438, opened 2026-07-20
ci: Build and run starboard_unittests on tvOS

rakuco:
Skip some tests that are failing either because of missing decode-to-texture support or because they'd been failing since C25.

Bug: 535573341
Bug: 536822943
Bug: 536822947


rakuco:

rakuco:
@linxinan-yt I've fixed the conflicts here as well

linxinan-yt (approved):

PR #11443, opened 2026-07-20
Disable AudioFrameDiscarder for passthrough audio to fix player crash

oxve:
Audio passthrough (AC3/EAC3) sends compressed bitstream to AudioTrack.
Slicing compressed bitstream at arbitrary PCM frame boundaries is not
supported and corrupts the bitstream.

This CL disables AudioFrameDiscarder for passthrough audio in
AudioDecoder to avoid shrinking the DecodedAudio buffer size to 0.
It also adds a safeguard in AudioRendererPassthrough to ignore 0-size
buffers.

BUG: 337087689
TAG: agy
CONV: e878fab6-c975-409f-9390-f9d70ea68a27

### DIFF
--- a/starboard/android/shared/audio_renderer_passthrough.cc
+++ b/starboard/android/shared/audio_renderer_passthrough.cc
@@ -605,7 +605,10 @@ void AudioRendererPassthrough::OnDecoderOutput() {
   SB_DCHECK(decoded_audio);
 
   if (!decoded_audio->is_end_of_stream()) {
-    SB_DCHECK(decoded_audio->size_in_bytes() > 0);
+    if (decoded_audio->size_in_bytes() == 0) {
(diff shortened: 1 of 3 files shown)


starboard/android/shared/audio_renderer_passthrough.cc
```diff
@@ -605,7 +605,10 @@ void AudioRendererPassthrough::OnDecoderOutput() {
   SB_DCHECK(decoded_audio);
 
   if (!decoded_audio->is_end_of_stream()) {
-    SB_DCHECK(decoded_audio->size_in_bytes() > 0);
+    if (decoded_audio->size_in_bytes() == 0) {
```
jasonzhangxx, line 608: After partial audio is disabled for passthrough audio, do we still need to remove the DCHECK?

PR #11447, opened 2026-07-20
android: Auto-retry URL load on network recovery

zhongqiliang:
Previously, if a network error occurred during the initial startup of
the application, users could end up stuck on a white screen even after
network connectivity was restored. This happened because there was no
mechanism to automatically trigger a reload once the device went back
online.

This change introduces a network connection observer that monitors
connectivity status. If the network becomes available and the
application has not yet finished its initial load, indicated by the
splash screen state, it will automatically attempt to reload the
failed URL or the initial startup URL.

Fix verification https://screencast.googleplex.com/cast/NTEwNjY4ODI1NTMyODI1NnxiYzhlYmYwMC02OQ

Bug: 536647405


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -632,6 +642,145 @@ protected void onDestroy() {
...
+  /** Performs reload of the target/current URL or active WebContents. */
+  public void reloadUrl(@Nullable String targetUrl) {
+    if (Looper.myLooper() != Looper.getMainLooper()) {
+      mHandler.post(() -> reloadUrl(targetUrl));
+      return;
+    }
+
+    WebContents webContents = getActiveWebContents();
+    String currentUrl = targetUrl != null ? targetUrl : "";
+    if (currentUrl.isEmpty() && webContents != null && webContents.getVisibleUrl() != null) {
```
ckim1346, line 728: if `targetUrl` and `getVisibleUrl` are both null, should we fall back to `mStartupUrl`?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
```diff
@@ -162,51 +154,38 @@ public boolean isShowing() {
...
-                Log.i(TAG, "Visible URL and fallback URL are empty, reloading without adding retry param");
-                webContents.getNavigationController().reload(/*checkForRepost=*/true);
-              } else {
-                cobaltActivity.getActiveShell().loadUrl(addRetryUrlParam(currentUrl, retryCount));
-              }
-            }
-          }
+          retry();
           break;
         case DISMISS_BUTTON:
```
ckim1346, line 190: If this WAI, we should also consider removing/reworking the `DISMISS_BUTTON` as it currently leads to a white screen if clicked while the device does not have a valid internet connection.

ckim1346 (commented):
Mostly LGTM, would just request that these changes are given to QA for full testing (jainpoonam@ previously did the majority of network dialog testing) as the network dialog edge cases can be difficult to catch.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/PlatformError.java
zhongqiliang: This is worth to consider after this PR is submitted. Could you drive this?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: In this case, it run webContents reload(). Which should be exactly the same as what it was there.
I 'd not change the logic unless necessary.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -632,6 +648,142 @@ protected void onDestroy() {
     super.onDestroy();
   }
 
+  private boolean isAutoRetryOnNetworkRecoveryEnabled() {
+    return getJavaSwitches().containsKey(JavaSwitches.ENABLE_AUTO_RETRY_ON_NETWORK_RECOVERY);
```
Awallky, line 652: IMO the JavaSwitch should be limited to propagating experiment configs to the command line. In the rest of the app, we should read the Finch and CLI flags.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -105,6 +106,17 @@ public abstract class CobaltActivity extends BaseCobaltActivity {
   private final Handler mHandler = new Handler(Looper.getMainLooper());
   private boolean mIsCobaltUsingAndroidOverlay;
 
+  private NetworkChangeNotifier.ConnectionTypeObserver mNetworkRecoveryObserver;
+  private boolean mIsNetworkRecoveryObserverRegistered = false;
+
+  private boolean mHasHiddenSplashScreen = false;
```
Awallky: I'm not 100% sure, but I think this could run into situations where multiple threads read/write to this  member concurrently. We should probably serialize reads and writes to it on the `mHandler` UIThread handler.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: This feature will be launched through an experiment. Is Finch flag ready to use? Do you have examples?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: right, need to add back volatile

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
Awallky: Yes. Right now, our JavaSwitches experiment architecture takes advantage of the fact that we can override Finch feature flags from the command line.

When we label some flags with `--enable-features` in JavaSwitches, these are enabling *Finch* flags.  So, if we add a flag to 
This effectively serves as our experimentation framework because of this. It just changes the way in which we deliver the config to the various Finch feature flags.

Awallky: We don't have anything in java yet doing this because we are missing a feature flag helper to check if a flag is enabled. 
We do have one for Starboard (`cobalt/android/apk/app/src/main/java/dev/cobalt/features/StarboardFeatureList.java`), but we need something for other cobalt browser features.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -632,6 +648,142 @@ protected void onDestroy() {
...
+   */
+  @VisibleForTesting
+  static String addRetryUrlParam(String url, int count) {
+    Uri parsedUri = Uri.parse(url);
+    Uri.Builder uriBuilder = parsedUri.buildUpon();
+
+    uriBuilder.query(null);
+    boolean retryParamAdded = false;
+
+    for (String key : parsedUri.getQueryParameterNames()) {
```
Awallky, line 769: What happens if this gets passed an opaque URL such as `about:blank` or `javascript:blah` or `data:blah`?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: adopted this suggestion.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: If targetUrl and webContents.getVisibleUrl() are empty or "about:blank", we fall back to mStartupUrl.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: We had an offline chat for this. The plan is to use Kimono cold config to do a study, then enable this by default, and remove the flag.

Awallky: Chatted offline. This will take significant work to support Finch flag checks in Java. We need to land this sooner than later, and this will still work fine. So, resolving this one as it is more of a stylistic preference than a functional error.

Awallky (approved):

ckim1346 (approved):

PR #11449, opened 2026-07-21
[tvOS] Add merge_rlibs_tvos_archive.sh

fayev-g:
GN's complete_static_lib does not merge Rust .rlib archives into the C++ static library.  This script takes the .rlib artifacts and merges them into cobalt.a. This is a one-off use-case just for tvOS and the cobalt.a snapshot pipeline for integration into the blaze production app. Below is an example error of missing symbols. See bug for additional details.

```
"_blink$cxxbridge1$crash_in_rust", referenced from:
    blink::crash_in_rust() in ...(lib.rs.o)
"_blink$cxxbridge1$crash_in_rust_with_overflow", referenced from:
    blink::crash_in_rust_with_overflow() in ...(lib.rs.o)
"_blink$cxxbridge1$reenter_rust", referenced from:
    blink::reenter_rust() in ...(lib.rs.o)

"_cxxbridge1$box$fontations_ffi$BridgeBitmapGlyph$alloc", referenced from:
    rust::cxxbridge1::Box<fontations_ffi::BridgeBitmapGlyph>::allocation::alloc() in ...(ffi.rs.o)
"_cxxbridge1$box$fontations_ffi$BridgeFontRef$alloc", referenced from:
    rust::cxxbridge1::Box<fontations_ffi::BridgeFontRef>::allocation::alloc() in ...(ffi.rs.o)
"_cxxbridge1$box$fontations_ffi$BridgeHintingInstance$alloc", referenced from:
    rust::cxxbridge1::Box<fontations_ffi::BridgeHintingInstance>::allocation::alloc() in ...(ffi.rs.o)

"_cxxbridge1$exception", referenced from:
    rust::cxxbridge1::detail::Fail::operator()(char const*) in ...(cxx.o)

"_cxxbridge1$rust_vec$bool$capacity"
"_cxxbridge1$rust_vec$bool$data"
"_cxxbridge1$rust_vec$bool$drop"
"_cxxbridge1$rust_vec$bool$len"
"_cxxbridge1$rust_vec$bool$new"
... (hundreds more _cxxbridge1$rust_vec$... symbols for bool, char, f32, f64, i16, i32, i64, i8, isize, str, string, u16, etc.)
```

Bug: 537031363


cobalt/devinfra/kokoro/build/tvos/merge_rlibs_tvos_archive.sh
```diff
@@ -0,0 +1,83 @@
...
+
+if [ -z "${EXISTING_RUST_LIBS}" ]; then
+  echo "No Rust libraries found to merge."
+  echo "Copying original archive to cobalt.a..."
+  cp "${COBALT_ARCHIVE}" cobalt.a
+  exit 0
+fi
+
+echo "Merging C++ archive and Rust libraries into cobalt.a..."
+libtool -static -o cobalt.a "${COBALT_ARCHIVE}" ${EXISTING_RUST_LIBS}
```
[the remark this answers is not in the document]
fayev-g: Done

linxinan-yt:
Should we invoke this script by a GN target? e.g. cobalt_archive? E.g. 

```
action("generate_my_custom_data") {
  # Path to the script relative to this BUILD.gn file
  script = "//cobalt/tools/merge_rlibs_tvos_archive.sh"

  # Files that the script reads (Ninja watches these for changes)
  inputs = [
    "data/source_template.txt",
    "data/config.json",
  ]

  # Files that the script generates
  outputs = [
    "$target_gen_dir/generated_data.bin",
  ]

  # Arguments passed to the script
  args = [
    "--input",
    rebase_path("data/config.json", root_build_dir),
    "--output",
    rebase_path("$target_gen_dir/generated_data.bin", root_build_dir),
  ]
}
```
what do you think?

fayev-g:
> Should we invoke this script by a GN target? e.g. cobalt_archive? E.g.
> 
> ```
> action("generate_my_custom_data") {
>   # Path to the script relative to this BUILD.gn file
>   script = "//cobalt/tools/merge_rlibs_tvos_archive.sh"
> 
>   # Files that the script reads (Ninja watches these for changes)
>   inputs = [
>     "data/source_template.txt",
>     "data/config.json",
>   ]
> 
>   # Files that the script generates
>   outputs = [
>     "$target_gen_dir/generated_data.bin",
>   ]
> 
>   # Arguments passed to the script
>   args = [
>     "--input",
>     rebase_path("data/config.json", root_build_dir),
>     "--output",
>     rebase_path("$target_gen_dir/generated_data.bin", root_build_dir),
>   ]
> }
> ```
> 
> what do you think?

I like this idea - I think there are a couple issues for now with how the script logic itself is decoupled from GN and kind of greps build.ninja as a post-op thing. The .sh is ideal for the existing CI scripts (build_mac and tvos_packager.py), but GN integration would be more consistent and easier to run locally (which is what I'm doing now). Also, apparently chromium prefers GN actions to be in Python(?) I'm not sure what the precedent in our code base is for this. I filed b/538337892 to investigate further. Hope it's OK to check in this script for now so I can use it to unblock the prod app.

fayev-g:
> > Should we invoke this script by a GN target? e.g. cobalt_archive? E.g.
> > ```
> > action("generate_my_custom_data") {
> >   # Path to the script relative to this BUILD.gn file
> >   script = "//cobalt/tools/merge_rlibs_tvos_archive.sh"
> > 
> >   # Files that the script reads (Ninja watches these for changes)
> >   inputs = [
> >     "data/source_template.txt",
> >     "data/config.json",
> >   ]
> > 
> >   # Files that the script generates
> >   outputs = [
> >     "$target_gen_dir/generated_data.bin",
> >   ]
> > 
> >   # Arguments passed to the script
> >   args = [
> >     "--input",
> >     rebase_path("data/config.json", root_build_dir),
> >     "--output",
> >     rebase_path("$target_gen_dir/generated_data.bin", root_build_dir),
> >   ]
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > what do you think?
> 
> I like this idea - I think there are a couple issues for now with how the script logic itself is decoupled from GN and kind of greps build.ninja as a post-op thing. The .sh is ideal for the existing CI scripts (build_mac and tvos_packager.py), but GN integration would be more consistent and easier to run locally (which is what I'm doing now). Also, apparently chromium prefers GN actions to be in Python(?) I'm not sure what the precedent in our code base is for this. I filed b/[538337892](http://b/538337892) to investigate further. Hope it's OK to check in this script for now so I can use it to unblock the prod app.

Added the action to this PR as is for now! Thanks for the suggestion

cobalt/BUILD.gn
```diff
@@ -148,6 +148,18 @@ if (is_ios && target_platform == "tvos") {
     ]
   }
 
+  action("merge_rlibs_tvos_archive") {
+    script = "//cobalt/devinfra/kokoro/build/tvos/merge_rlibs_tvos_archive.sh"
```
linxinan-yt: What about moving the script to //cobalt/tools/ ? 

Eventually this script should be owned by dev, not infra. WDYT?

cobalt/build/tvos/merge_rlibs_tvos_archive.sh
```diff
@@ -0,0 +1,84 @@
...
+for lib in ${RUST_LIBS}; do
+  if [ -f "${lib}" ]; then
+    EXISTING_RUST_LIBS+=("${lib}")
+  else
+    echo "Warning: Target Rust library '${lib}' not found on disk, skipping."
+  fi
+done
+
+if [ ${#EXISTING_RUST_LIBS[@]} -eq 0 ]; then
+  echo "No Rust libraries found to merge."
```
linxinan-yt, line 75: Is it WAI if no rust libs found?

cobalt/BUILD.gn
fayev-g: The script is required for the release, CI, and infra pipeline though for tvOS, and so I think it belongs in "...build/tvos", unless there is another folder that contains release/CI code. My impression is that cobalt/tools contains non-release tools (e.g. uma) without platform-specific directories. I think between the two, build/tvos makes more sense.

cobalt/build/tvos/merge_rlibs_tvos_archive.sh
fayev-g: Yeah that just means maybe the cobalt target wasn't built. I don't think it should throw an error - this message is fine. This is what was commented in the bug about more fully moving the script's functionality to gn - there is a better way to add the cobalt target as a dependency of the action. But for now, the script is still usable.

cobalt/BUILD.gn
linxinan-yt: > The script is required for the release...

This script is for archiving but not involving any kokoro features e.g. signing, notary etc. 
Cobalt, including release package, should be hermetic. From product perspective, if a developer has to triage a bug in the merged package, they should be able to build such package locally, without a kokoro build. 

If we blur the boundary, it will be harder to isolate issues in future between product and infra. So I think it might be worth creating a tvos folder in cobalt/tools or cobalt/build, but not cobalt/devinfra, where we should only host infra scripts.

fayev-g: OK, I think part of the confusion stems from the related tvos_packager.py being in cobalt/devinfra/kokoro/build/tvos/... I think I understand what you're saying, and that file needs to be moved into a cobalt/build/tvos folder. All the packaging for other platforms are done in cobalt/build. I move it in https://github.com/youtube/cobalt/pull/11547. I've moved the file in this PR into that dir also. Lmk if this works for you.

cobalt/build/tvos/merge_rlibs_tvos_archive.sh
```diff
@@ -0,0 +1,84 @@
...
+COBALT_ARCHIVE="obj/cobalt/libcobalt_archive.a"
+if [ ! -f "${COBALT_ARCHIVE}" ]; then
+  echo "Error: Base archive '${COBALT_ARCHIVE}' not found."
+  echo "Make sure the 'cobalt_archive' target was built successfully first."
+  exit 1
+fi
+
+echo "Scanning build.ninja for compiled Rust libraries..."
+
+# Dynamically extract all Rust libraries (.rlib and .a) generated by rust_rlib or rust_staticlib rules
```
linxinan-yt, line 61: I am not an expert of rust, but I feel like there should be some GN native solution to read the output of rust_rlib or rust_staticlib, and then you could pass it to `libtool`.

I don't want to block you here. Feel free to leave a todo to follow up.

linxinan-yt (approved):

cobalt/build/tvos/merge_rlibs_tvos_archive.sh
fayev-g: Yeah I agree. When I initially looked it up/asked gemini, it just said static_library didn't include them and you just had to post-process/have an action, but hopefully there is another solution.

PR #11453, opened 2026-07-21
Bypass desktop Mesa FBO capability probe on Starboard to prevent fatal DCHECK aborts in devel/debug builds.

haozheng-cobalt:
Skip the desktop Mesa Framebuffer Object (FBO) capability probe on
Starboard platforms. This prevents fatal DCHECK aborts in debug and
development builds caused by benign error flags left in the pipeline
during embedded GLES driver initialization on Starboard and RDK
devices.

Bug: 537056801


andrewsavage1:
Closing this based on suggestions in gchat, please reopen if you still think it's necessary

PR #11459, opened 2026-07-21
starboard/tests: enable blink_common/boringssl_pki/storage

denisyuji:
- Drop quarantine filters for all three suites.
- Add missing `runs_on`/`test_type` to `boringssl_pki_tests` in `test_targets.json` so it is no longer silently skipped.
- Wire `blink_common_unittests` to a Starboard `SbEventHandle` entrypoint via a reusable `starboard/testing` wrapper (gated on `is_cobalt_hermetic_build`); its bare `main()` crashed the loader.
- Under `ENABLE_COBALT_HERMETIC_HACKS`, fix `DIR_SRC_TEST_DATA_ROOT` on device. Both `blink_common_unittests` and `boringssl_pki_tests` use this path to locate checked-in test data. On device, the current `DirName()` heuristic results in `/`, so the data under the content directory is not found. In this case, fall back to content. Host runs are unaffected because this path is compiled out.
- Skip `LocalFileUtilTest.FileEnumeratorError` in `storage_unittests` when running as root; `chmod(dir, 0)` does not block `opendir()` for root, so the test's expected `FILE_ERROR_ACCESS_DENIED` path never triggers.

Bug: 486053350


base/base_paths_starboard.cc
```diff
@@ -73,6 +73,15 @@ bool PathProviderStarboard(int key, FilePath* result) {
 #else
         *result = test_data_path.DirName().DirName();
 #endif
+        // On device this heuristic degenerates to "/"; fall back to the
+        // content directory, where test data is deployed flattened.
+        if (*result == FilePath(FILE_PATH_LITERAL("/")) ||
+            (!DirectoryExists(
+                 result->Append(FILE_PATH_LITERAL("third_party"))) &&
+             DirectoryExists(
+                 test_data_path.Append(FILE_PATH_LITERAL("third_party"))))) {
```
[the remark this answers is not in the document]
denisyuji: it already is

starboard/testing/sbeventhandle_from_main_with_args.cc
```diff
@@ -0,0 +1,19 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "starboard/client_porting/wrap_main/wrap_main.h"
+
+int main(int argc, char** argv);
+
+SB_EXPORT STARBOARD_WRAP_SIMPLE_MAIN(main)
```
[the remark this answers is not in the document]
denisyuji: it's done

base/base_paths_starboard.cc
```diff
@@ -73,6 +73,18 @@ bool PathProviderStarboard(int key, FilePath* result) {
 #else
         *result = test_data_path.DirName().DirName();
 #endif
+        // Fall back to the content directory when the heuristic failed
```
niranjanyardi, line 76: why do we need a fallback ?
maybe the current failure is good and expected.

storage/browser/file_system/local_file_util_unittest.cc
```diff
@@ -388,6 +389,10 @@ TEST_F(LocalFileUtilTest, MoveDirectory) {
 // Chromium's base namespace.
 #if BUILDFLAG(IS_POSIX)
 TEST_F(LocalFileUtilTest, FileEnumeratorError) {
+  if (geteuid() == 0) {
```
niranjanyardi, line 392: avoid modifying upsdtream , you can avoid running the tests which are meaningless insterad of modifying upstream code

starboard/testing/sbeventhandle_from_main_with_args.cc
```diff
@@ -0,0 +1,19 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
```
niranjanyardi, line 1: no need to add a new file , look for an existing one

PR #11461, opened 2026-07-21
nplb: page-align stack buffer for PosixThreadAttrTest.StackAddrAndSizeAttr

rogerzanoni:
PosixThreadAttrTest.StackAddrAndSizeAttr is failing with EINVAL because
bionic's  pthread_attr_setstack() expects the buffer to be page-aligned:

__BIONIC_WEAK_FOR_NATIVE_BRIDGE
int pthread_attr_setstack(...)
  if ((stack_size & (page_size() - 1) || stack_size < PTHREAD_STACK_MIN)) {
    return EINVAL;
  }
  if (reinterpret_cast<uintptr_t>(stack_base) & (page_size() - 1)) {
    return EINVAL;
  }
  attr->stack_base = stack_base;
  attr->stack_size = stack_size;
  return 0;
}

Allocate the stack in the heap and use posix_memalign() to fulfill
bionic's restrictions.

Bug: 532068409

### DIFF
--- a/starboard/nplb/posix_compliance/posix_thread_attr_test.cc
+++ b/starboard/nplb/posix_compliance/posix_thread_attr_test.cc
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <pthread.h>
+#include <memory>
 #include <stdlib.h>
 #include <unistd.h>
 
--- a/starboard/nplb/posix_compliance/posix_thread_attr_test.cc
+++ b/starboard/nplb/posix_compliance/posix_thread_attr_test.cc
@@ -13,11 +13,12 @@
 // limitations under the License.
 
 #include <pthread.h>
-#include <memory>
 #include <stdlib.h>
+#include <string.h>
 #include <unistd.h>
 
-#include "build/build_config.h"
+#include <memory>
+
 #include "starboard/common/log.h"
 #include "testing/gtest/include/gtest/gtest.h"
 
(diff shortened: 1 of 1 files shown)


andrewsavage1:
How does the platform fail? Is there any reference about the platform's max stack size in the spec for the platform?

rogerzanoni:
@andrewsavage1 it's around 1MB on bionic: https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/bionic/pthread_internal.h 

```cpp
// Traditionally we gave threads a 1MiB stack. When we started
// allocating per-thread alternate signal stacks to ease debugging of
// stack overflows, we subtracted the same amount we were using there
// from the default thread stack size. This should keep memory usage
// roughly constant.
#define PTHREAD_STACK_SIZE_DEFAULT ((1 * 1024 * 1024) - SIGNAL_STACK_SIZE_WITHOUT_GUARD)
```

But I will abandon this PR because I took a closer look at the issue, and the problem actually is that StarboardMain usually runs on main() of the main process thread that usually has a larger stack size (8M), this is from the test device:

```shell
adb shell
karat:/ $ ulimit -s
8192
```

but aosp main is running on a std::thread that inherits the 1MB stack size

I already have a fix that increases the stack size of the created thread, I will upload it instead.

rogerzanoni:
I have split this into the stack size fix in #11664, and the EINVAL issue that happens because pthread_attr_setstack() expects the stack mem to be page-alined

sacuff (approved):

andrewsavage1:
Changing behavior in nplb can mean actual code in the Cobalt app that expects the same behavior wouldn't work. Would that be the case here? Does the current change to nplb align with posix specifications or is it a fix specific to android quirks?

hferreiro:
https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_attr_setstack.html mentions this: "pthread_attr_setstack() function may fail with [EINVAL] if stackaddr does not meet implementation-defined alignment requirements". So I guess we could even use the mem-aligned version unconditionally.

andrewsavage1:
> https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_attr_setstack.html mentions this: "pthread_attr_setstack() function may fail with [EINVAL] if stackaddr does not meet implementation-defined alignment requirements". So I guess we could even use the mem-aligned version unconditionally.

Yes let's do this

andrewsavage1 (approved):
Block
Edit: accidental comment

PR #11462, opened 2026-07-21
aosp: wrap readlink to validate buffer size

rogerzanoni:
Passing a buffer size of 0 or one that wraps around size_t to readlink/readlinkat triggers a FORTIFY arithmetic-overflow abort in bionic.

Fixes the crash in PosixReadlinkTest.InvalidBufferSizeFails.

Bug: 532068409

### DIFF
--- a/starboard/android/shared/posix_emu/readlink.cc
+++ b/starboard/android/shared/posix_emu/readlink.cc
@@ -1,0 +1,33 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <errno.h>
+#include <limits.h>
+#include <unistd.h>
+
+extern "C" {
+
+ssize_t __real_readlink(const char* path, char* buf, size_t bufsize);
+
+ssize_t __wrap_readlink(const char* path, char* buf, size_t bufsize) {
+  // Passing a bufsize of 0 or one that wraps around size_t triggers a FORTIFY
+  // arithmetic-overflow abort in bionic. Reject those up front with EINVAL.
+  if (bufsize == 0 || bufsize > SSIZE_MAX) {
+    errno = EINVAL;
+    return -1;
+  }
+  return __real_readlink(path, buf, bufsize);
+}
+
+}  // extern "C"
--- a/starboard/android/shared/BUILD.gn
+++ b/starboard/android/shared/BUILD.gn
@@ -352,6 +352,7 @@ static_library("starboard_platform") {
       "posix_emu/errno.cc",
       "posix_emu/file.cc",
       "posix_emu/net_if.cc",
+      "posix_emu/readlink.cc",
       "posix_emu/stat.cc",
     ]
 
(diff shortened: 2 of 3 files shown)


starboard/elf_loader/exported_symbols.cc
```diff
@@ -396,6 +395,8 @@ ExportedSymbols::ExportedSymbols() {
   REGISTER_WRAPPER(pthread_sigmask);
   REGISTER_WRAPPER(readdir);
   REGISTER_WRAPPER(readdir_r);
+  REGISTER_WRAPPER(readlink);
+  REGISTER_WRAPPER(readlinkat);
```
sacuff: SB 18 is frozen, so `readlinkat` should be Android-specific since we can't expect the system implementation to exist on all platforms for now.

sacuff (changes requested):

andrewsavage1 (changes requested):
These should likely be implemented in an android-specific way in posix_emu

Why is it necessary to add readlinkat?

rogerzanoni:
@andrewsavage1 I moved it to posix_emu and removed readlinkat - it was present in a change I did during the cobalt-25 experiment, and I didn't re-check if readlinkat was needed. I will update the PR.

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -723,3 +724,28 @@ int __abi_wrap_unlinkat(int fildes, const char* path, int musl_flag) {
...
+#if defined(__ANDROID__)
+  // Passing a bufsize that wraps around size_t triggers a 
+  // FORTIFY arithmetic overflow abort in bionic
+  if (bufsize == 0 || bufsize > SSIZE_MAX) {
+    errno = EINVAL;
+    return -1;
+  }
+#endif  // defined(__ANDROID__)
+  return readlink(path, buf, bufsize);
+}
```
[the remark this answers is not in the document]
rogerzanoni: Outdated

starboard/shared/modular/starboard_layer_posix_unistd_abi_wrappers.cc
```diff
@@ -723,3 +724,28 @@ int __abi_wrap_unlinkat(int fildes, const char* path, int musl_flag) {
...
+                              char* buf,
+                              size_t bufsize) {
+#if defined(__ANDROID__)
+  if (bufsize == 0 || bufsize > SSIZE_MAX) {
+    errno = EINVAL;
+    return -1;
+  }
+#endif  // defined(__ANDROID__)
+  return readlinkat(dirfd, path, buf, bufsize);
+}
```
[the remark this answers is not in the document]
rogerzanoni: Outdated

starboard/android/shared/posix_emu/readlink.cc
```diff
@@ -0,0 +1,33 @@
...
+
+#include <errno.h>
+#include <limits.h>
+#include <unistd.h>
+
+extern "C" {
+
+ssize_t __real_readlink(const char* path, char* buf, size_t bufsize);
+
+ssize_t __wrap_readlink(const char* path, char* buf, size_t bufsize) {
```
andrewsavage1, line 23: I don't believe this implementation is correct. According to the readlink posix spec, if bufsize is greater than SSIZE_MAX, the behavior is implementation-defined, so we should only change that behavior if upstream chromium has some expectation of how it works. If it's only a nplb test failing, we should remove that nplb test as it doesn't align with the spec.

For bufsize == 0, I don't see anything saying the function should return EINVAL, instead I believe we should just return 0 without doing anything.

hferreiro: https://pubs.opengroup.org/onlinepubs/9799919799/functions/readlink.html mentions "The behavior specified for readlink() when bufsiz is zero represents historical practice" with no explanation of what that should be.

We cannot just return 0, since that would fail in all platforms. I created a PR to remove the test at https://github.com/youtube/cobalt/pull/12090.

andrewsavage1: Okay, let's leave it as is for bufsize of 0, it seems like the posix spec and linux implementation may diverge. For SSIZE_MAX, do you still think we need special handling?

hferreiro:
Closing in favour of https://github.com/youtube/cobalt/pull/12090.

PR #11463, opened 2026-07-21
aosp: fix asset directory enumeration for readdir* and fdopendir

rogerzanoni:
Re-enable posix_emu directory functions emulation for and fix asset
subdirectory enumeration.

AAssetManager_openDir + AAssetDir_getNextFileName can't see
subdirectories which results in some tests that expect subdirs to be
enumerated to fail, such as PosixDirectoryOpenTest.SunnyDayStaticContent
and PosixDirectoryGetNextTest.SunnyDayStaticContent (and the
StaticContent stat tests).

Instead, use AssetManager.list() via JNI that can enumerate
subdirectories.

Also use the new enumeration function to fix fdopendir behavior.

Bug: 532068409

### DIFF
--- a/starboard/android/shared/posix_emu/dirent.cc
+++ b/starboard/android/shared/posix_emu/dirent.cc
@@ -16,18 +16,64 @@
 #include <dirent.h>
 // clang-format on
 
-#include <android/asset_manager.h>
+#include <errno.h>
+#include <string.h>
 
-#include <map>
 #include <mutex>
+#include <set>
+#include <string>
+#include <utility>
+#include <vector>
 
 #include "starboard/android/shared/file_internal.h"
 #include "starboard/common/string.h"
-#include "starboard/configuration_constants.h"
 
 using starboard::IsAndroidAssetPath;
-using starboard::OpenAndroidAsset;
-using starboard::OpenAndroidAssetDir;
+using starboard::ListAndroidAssetDir;
+
+namespace {
+
+// A helper structure used internally as a DIR replacement for asset paths.
+// It's used in the wrapped opendir/closedir/readdir/readdir_r.
+struct AndroidAssetDir {
+  std::vector<std::string> entries;
+  size_t index;
+  struct dirent entry;
+};
+
+// Registry of the DIR* handles created by the asset-path wrappers below.
+std::mutex& GetAssetDirMutex() {
+  static std::mutex* mutex = new std::mutex();
+  return *mutex;
+}
+
+std::set<const AndroidAssetDir*>& GetActiveAssetDirs() {
+  static std::set<const AndroidAssetDir*>* dirs =
+      new std::set<const AndroidAssetDir*>();
+  return *dirs;
+}
+
+void RegisterAssetDir(const AndroidAssetDir* dir) {
+  std::lock_guard lock(GetAssetDirMutex());
+  GetActiveAssetDirs().insert(dir);
+}
+
+bool UnregisterAssetDir(const AndroidAssetDir* dir) {
+  std::lock_guard lock(GetAssetDirMutex());
+  return GetActiveAssetDirs().erase(dir) > 0;
+}
+
+bool IsAssetDir(const AndroidAssetDir* dir) {
+  std::lock_guard lock(GetAssetDirMutex());
+  return GetActiveAssetDirs().count(dir) > 0;
+}
+
+void FillAssetDirent(const std::string& name, struct dirent* out) {
+  memset(out, 0, sizeof(*out));
+  starboard::strlcpy(out->d_name, name.c_str(), sizeof(out->d_name));
+}
+
+}  // namespace
 
 ///////////////////////////////////////////////////////////////////////////////
 // Implementations below exposed externally in pure C for emulation.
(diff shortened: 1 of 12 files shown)


starboard/android/shared/posix_emu/dirent.cc
```diff
@@ -143,20 +131,36 @@ int __wrap_readdir_r(DIR* __restrict dir,
...
+  if (asset_dir->magic == kMagicNum) {
+    if (asset_dir->index >= asset_dir->entries.size()) {
+      return NULL;
+    }
+    FillAssetDirent(asset_dir->entries[asset_dir->index++], &asset_dir->entry);
+    return &asset_dir->entry;
+  }
+
+  return __real_readdir(dir);
+}
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

andrewsavage1:
@sacuff ptal

starboard/android/shared/posix_emu/file.cc
```diff
@@ -64,6 +67,10 @@ int __wrap_openat(int dirfd, const char* path, int oflag, ...) {
     }
     return __real_openat(dirfd, path, oflag);
   }
+  // handle asset paths
+  if ((oflag & O_DIRECTORY) == O_DIRECTORY) {
```
sacuff, line 71: Is there handling for if `path` is not a directory? For `openat` and `open` we need to have "If path resolves to a non-directory file, fail and set errno to [ENOTDIR]."

I realized starboard/nplb/posix_compliance/posix_file_openat_test.cc is missing a case for this so we should add that too.

starboard/android/shared/posix_emu/dirent.cc
```diff
@@ -16,121 +16,128 @@
...
+struct AndroidAssetDir {
+  std::vector<std::string> entries;
+  size_t index;
+  struct dirent entry;
+};
 
-int __real_readdir_r(DIR* __restrict dir,
-                     struct dirent* __restrict dirent_buf,
-                     struct dirent** __restrict dirent);
+// Registry of the DIR* handles created by the asset-path wrappers below.
```
sacuff: I believe this comment goes with the function below this one

starboard/android/shared/posix_emu/dirent.cc
```diff
@@ -16,121 +16,128 @@
...
+std::mutex& GetAssetDirMutex() {
+  static std::mutex* mutex = new std::mutex();
```
sacuff: Does this need to be a pointer?

starboard/android/shared/posix_emu/dirent.cc
```diff
@@ -16,121 +16,128 @@
...
-// https://android.googlesource.com/platform/frameworks/base/+/master/native/android/asset_manager.cpp
-static int gen_fd() {
-  static int fd = 100;
-  fd++;
-  if (fd == 0x7FFFFFFF) {
-    fd = 100;
-  }
-  return fd;
+std::set<const AndroidAssetDir*>& GetActiveAssetDirs() {
+  static std::set<const AndroidAssetDir*>* dirs =
```
sacuff: Same pointer question

sacuff (changes requested):

starboard/android/shared/posix_emu/dirent.cc
```diff
@@ -143,20 +150,36 @@ int __wrap_readdir_r(DIR* __restrict dir,
     return -1;
```
andrewsavage1: readdir_r should never return -1, it should return 0 on success or a number indicating the error (i.e. here likely EBADF) on failure

starboard/android/shared/posix_emu/file.cc
rogerzanoni: Fixed

starboard/android/shared/posix_emu/dirent.cc
rogerzanoni: Fixed

starboard/android/shared/posix_emu/dirent.cc
rogerzanoni: Changed to a static local

starboard/android/shared/posix_emu/dirent.cc
rogerzanoni: Changed to a local static

starboard/android/shared/posix_emu/dirent.cc
rogerzanoni: Fixed

sacuff (approved):

starboard/android/shared/posix_emu/dirent.cc
```diff
@@ -16,18 +16,68 @@
...
+  std::vector<std::string> entries;
+  size_t index;
+  struct dirent entry;
+  // The AssetManager fd this handle took ownership of in __wrap_fdopendir(),
+  // or -1 when the handle came from __wrap_opendir()
+  int fd = -1;
+};
+
+std::mutex& GetAssetDirMutex() {
+  static std::mutex mutex;
```
andrewsavage1: Static might be banned

starboard/android/shared/posix_emu/dirent.cc
andrewsavage1: I actually think this was better as it was before, or these static variables (this one and the std::set) should be moved to the asset manager as that's already a singleton so they could just be declared as members instead. See https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables for more information

rogerzanoni: Thanks for the direction, it should've been in the AssetManager from the start. I moved the fd registry there and used the member mutex for the guards.

andrewsavage1 (approved):

PR #11474, opened 2026-07-21
Support heap profiler on 3P devel and debug builds

haozheng-cobalt:
Enable heap profiling support for debug and devel build types. This
includes configuring GN flags to enable profiling, disabling ARM
thumb mode, and ensuring unwind tables are not excluded.

Bug: 537412550

### DIFF
--- a/starboard/build/config/modular/BUILD.gn
+++ b/starboard/build/config/modular/BUILD.gn
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import("//build/config/compiler/compiler.gni")
 import("//starboard/build/config/sanitizers.gni")
 
 config("modular") {
@@ -78,9 +79,6 @@ config("modular") {
       # Default visibility to hidden, to enable dead stripping.
       "-fvisibility=hidden",
       "-fno-exceptions",
-
-      # Disable usage of frame pointers.
-      "-fomit-frame-pointer",
       "-fno-strict-aliasing",  # See http://crbug.com/32204
       "-fno-use-cxa-atexit",
 
@@ -92,6 +90,12 @@ config("modular") {
       #         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
       "-Wno-sign-compare",
     ]
+
+    if (!is_official_build || can_unwind_with_frame_pointers) {
+      cflags += [ "-fno-omit-frame-pointer" ]
+    } else {
+      cflags += [ "-fomit-frame-pointer" ]
+    }
   }
 
   if (use_asan) {
--- a/starboard/build/config/modular/arm/BUILD.gn
+++ b/starboard/build/config/modular/arm/BUILD.gn
@@ -26,4 +26,9 @@ config("arm") {
     # Force char to be signed.
     "-fsigned-char",
   ]
+
+  # Force ARM 32-bit mode (-marm) instead of Thumb-2 mode for non-official builds.
+  if (!is_official_build) {
+    cflags += [ "-marm" ]
+  }
 }


starboard/build/config/modular/arm/BUILD.gn
```diff
@@ -26,4 +26,8 @@ config("arm") {
     # Force char to be signed.
     "-fsigned-char",
   ]
+
+  if (!is_official_build) {
+    cflags += [ "-marm" ]
```
Awallky, line 32: What effect does this have on memory profiling? If any, along with the other omissions, we should note it somewhere.

Awallky (approved):
LGTM % comment

starboard/build/config/modular/arm/BUILD.gn
haozheng-cobalt: yea, added a comment to explain this flag. thanks Adam!

starboard/build/config/modular/arm/BUILD.gn
```diff
@@ -26,4 +26,11 @@ config("arm") {
     # Force char to be signed.
     "-fsigned-char",
   ]
+
+  # Force ARM 32-bit mode (-marm) instead of Thumb-2 mode for non-official
```
andrewsavage1: Is this file currently used for evergreen-arm64 builds?

starboard/build/config/modular/BUILD.gn
```diff
@@ -92,6 +90,12 @@ config("modular") {
       #         ^~~~~~~~~~~~~~~~~~~~~~~~~~~
       "-Wno-sign-compare",
     ]
+
+    if (!is_official_build || can_unwind_with_frame_pointers) {
```
andrewsavage1, line 94: I don't think we would ever want to enable frame pointers for official builds, so I think this should be either `if (!is_official_build)` (in which case you should swap the order of the if/else for a positive condition) or `if (can_unwind_with_frame_pointers && !is_official_build)`

starboard/build/config/modular/arm/BUILD.gn
haozheng-cobalt: No, I don't see this file being used for evergreen-arm64 builds:
```
haozheng@haozheng-usa:~/cobalt/src$ gn desc out/evergreen-arm64_devel //starboard/build/config/modular/arm:arm configs --tree
The input //starboard/build/config/modular/arm:arm matches no targets, configs or files.
```
And according to Gemini, this cflag I'm adding to arm32 is not needed for arm54:
```
Thumb mode does not exist in ARM64. The AArch64 ISA has dedicated hardware frame pointer registers (x29/fp). -marm is not used and not needed.
```

starboard/build/config/modular/BUILD.gn
haozheng-cobalt: Good point, changed to &&.

starboard/build/config/modular/arm/BUILD.gn
```diff
@@ -26,4 +26,11 @@ config("arm") {
...
+  # builds so Clang emits reliable frame pointer chains (r11) for stack
+  # unwinding and in-process heap profiling.
+  if (!is_official_build) {
+    cflags += [ "-marm" ]
```
andrewsavage1, line 32: Probably the right way to do this is to set arm_use_thumb=false for unofficial builds.

andrewsavage1 (changes requested):

andrewsavage1 (approved):

starboard/build/config/modular/arm/BUILD.gn
haozheng-cobalt: `arm_use_thumb` is declared inside Chromium's top-level build/config/arm.gni within a declare_args() block. so I override the arg in cobalt/build/gn.py.

PR #11483, opened 2026-07-21
feat: Add Sb extension for native stability events

hlwarriner:
This change adds the Starboard extension and 3P implementation proposed in go/cmc-storage-3p-mvp-design, and a temporary usage example in the Cobalt layer.

Based on offline discussion after the design review, this implementation refers to crashes, hangs, etc. as Native Stability (events) rather than Core Metric Components.

#vibe-coded

Issue: 528362453

### DIFF
--- a/cobalt/browser/cobalt_browser_main_parts.cc
+++ b/cobalt/browser/cobalt_browser_main_parts.cc
@@ -38,6 +38,12 @@
 #include "components/metrics_services_manager/metrics_services_manager.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/resource_coordinator_service.h"
+
+#if BUILDFLAG(USE_EVERGREEN)
+#include "starboard/extension/native_stability.h"
+#include "starboard/system.h"
+#endif
+
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/client_process_impl.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/memory_instrumentation.h"
 #include "services/resource_coordinator/public/cpp/memory_instrumentation/memory_instrumentation_features.h"
@@ -230,6 +236,31 @@ int CobaltBrowserMainParts::PreMainMessageLoopRun() {
   client->SetUserAgentCrashAnnotation();
 
 #endif  // !BUILDFLAG(IS_ANDROIDTV)
+
+#if BUILDFLAG(USE_EVERGREEN)
+  // TODO: b/528362453 - This startup extension query is for debugging and to
+  // serve as an example of how to invoke the NativeStability extension.
+  // Remove this block once the NativeStabilityManager singleton is added to
+  // the browser to query the extension to respond to web app requests.
+  auto native_stability_extension =
+      static_cast<const CobaltExtensionNativeStabilityApi*>(
+          SbSystemGetExtension(kCobaltExtensionNativeStabilityName));
+  if (native_stability_extension && native_stability_extension->version >= 1 &&
+      native_stability_extension->ReadReports) {
+    constexpr int max_num_reports = 16;  // A somewhat arbitrary, large number
+    SbNativeStabilityReport reports[max_num_reports];
+    int count =
+        native_stability_extension->ReadReports(reports, max_num_reports);
+    LOG(INFO) << "=== NativeStability extension query at startup returned "
+              << count << " reports ===";
+    for (int i = 0; i < count; ++i) {
+      LOG(INFO) << "Report [" << i
+                << "]: event_uuid=" << reports[i].native_stability_event_uuid
+                << ", event_time_s=" << reports[i].event_time_s
+                << ", type=" << static_cast<int>(reports[i].report_type);
+    }
+  }
+#endif
 
   int result = ShellBrowserMainParts::PreMainMessageLoopRun();
 
--- a/starboard/crashpad_wrapper/wrapper.cc
+++ b/starboard/crashpad_wrapper/wrapper.cc
@@ -18,6 +18,7 @@
 #include <sys/stat.h>
 
 #include <map>
+#include <memory>
 #include <optional>
 #include <vector>
 
@@ -54,7 +55,7 @@ const char kNativeStabilityHangUuidKey[] = "native_stability_hang_uuid";
 
 namespace {
 
-base::FilePath* g_database_path_override_for_testing = nullptr;
+std::unique_ptr<base::FilePath> g_database_path_override_for_testing;
 
 // TODO: Get evergreen information from installation.
 const std::string kCrashpadVersion = "1.0.0.0";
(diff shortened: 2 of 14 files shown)


hlwarriner:
CC @vincentmo @AlecsisDuarte

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -38,6 +38,12 @@
 #include "components/metrics_services_manager/metrics_services_manager.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/resource_coordinator_service.h"
+
+#if BUILDFLAG(USE_EVERGREEN)
```
andrewsavage1, line 42: Part of the point of starboard extensions is that platforms can decide whether or not to implement it—do we need to explicitly prevent it using BUILDFLAG(IS_EVERGREEN) in the source like this? If it does need to be limited to just evergreen platforms, maybe we could configure the system_get_extensions implementations to only register this extension when BUILDFLAG(IS_EVERGREEN) is true?

hlwarriner: This condition should not technically be needed here because platforms are already doing what you said - only registering the extension when `BUILDFLAG(USE_EVERGREEN)`.

I don't think it hurts from a code clarity perspective, though, since we know for a fact that only Evergreen platforms are actually registering this extension. Otherwise, a reader of the code would need to do more work to see if platform x actually will actually use the extension here.

This code should be very short-lived as it will be removed when the next incremental PR is uploaded, so I think it's fine either way.

starboard/crashpad_wrapper/wrapper.cc
```diff
@@ -309,4 +390,60 @@ void DumpWithoutCrashingWrapper() {
...
+  std::vector<::crashpad::CrashReportDatabase::Report> completed_reports;
+  if (database->GetCompletedReports(&completed_reports) ==
+      ::crashpad::CrashReportDatabase::kNoError) {
+    all_reports.insert(all_reports.end(), completed_reports.begin(),
+                       completed_reports.end());
+  }
+
+  int count = 0;
+  for (const auto& report : all_reports) {
+    if (count >= max_num_reports) {
```
andrewsavage1, line 429: This would favor sending pending reports over completed reports if it runs into the limit. Does that matter?  Is there anything in completed reports that isn't in pending reports that might give us more complete information?

starboard/extension/extension_test.cc
```diff
@@ -629,4 +630,25 @@ TEST(ExtensionTest, StarboardMediaBufferPoolExtension) {
       << "Extension struct should be a singleton";
 }
 
+TEST(ExtensionTest, NativeStabilityExtension) {
```
andrewsavage1, line 633: For required extensions (like I think this one will be?), should we add a test like this to nplb as well or instead?

andrewsavage1: And in that vein, should the test fail `if (!extension_api)`?

hlwarriner: Hmm yeah it's probably a good idea to add a simple test to nplb, similar to the test case in `starboard/nplb/nplb_evergreen_compat_tests/loader_app_metrics_test.cc`.  I'll do that but would prefer to do so in a quick follow-up PR. The more rigorous test we have planned is a certification test that will exercise the behaviors of the extension by interacting with the corresponding web interface, and components in the browser, built on top of it.

> And in that vein, should the test fail if (!extension_api)

I'd rather not do that. The test cases in this file follow a very similar pattern and are just responsible for ensuring platforms, if they implement a particular extension, provide a valid implementation. I'd like to keep that pattern and rely on the nplb test to ensure the expected platforms are providing an implementation, and the certification test to ensure the expected platforms are providing an implementation that satisfies all the required behaviors.

starboard/crashpad_wrapper/wrapper.cc
hlwarriner: That's a good point but in practice it shouldn't matter much.

Just after the Crashpad handler snapshots a crash and generates the report, it attempts an upload of the report - or declines to do so because of client-side throttling - and in all cases moves the report from `pending` to `completed`: whether the upload was 1) throttled by the handler itself, 2) succeeded, or 3) failed. So it should be very rare for us to encounter a `pending` report here, in `ReadReports` - maybe only in edge cases where the handler process crashes during an upload attempt, or something.

I'll add a short comment explaining that we generally don't expect any, and certainly not many, reports in `pending`.

andrewsavage1 (approved):

starboard/extension/extension_test.cc
hlwarriner: Filed b/538114999 to track the nplb test.

starboard/shared/starboard/native_stability.cc
```diff
@@ -0,0 +1,60 @@
...
+  int result = -1;
+
+  SB_CHECK_EQ(pthread_mutex_lock(&g_read_reports_callback_mutex), 0);
+  if (g_read_reports_callback) {
+    result = g_read_reports_callback(reports, max_num_reports);
+  }
+  SB_CHECK_EQ(pthread_mutex_unlock(&g_read_reports_callback_mutex), 0);
+
+  return result;
+}
```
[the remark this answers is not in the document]
hlwarriner: Logically this shouldn't matter today, based on the calling conventions we expect. But it's a good idea to future proof the code. Done.

starboard/crashpad_wrapper/wrapper.cc
```diff
@@ -44,8 +49,13 @@ const char kCrashpadVersionKey[] = "ver";
 const char kCrashpadProductKey[] = "prod";
 const char kCrashpadUserAgentStringKey[] = "user_agent_string";
 const char kCrashpadCertScopeKey[] = "cert_scope";
+const char kNativeStabilityCrashUuidKey[] = "native_stability_crash_uuid";
+const char kNativeStabilityHangUuidKey[] = "native_stability_hang_uuid";
 
 namespace {
+
+base::FilePath* g_database_path_override_for_testing = nullptr;
```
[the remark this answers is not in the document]
hlwarriner: done

starboard/crashpad_wrapper/wrapper.cc
```diff
@@ -309,4 +390,60 @@ void DumpWithoutCrashingWrapper() {
...
+namespace internal {
+
+void SetDatabasePathForTesting(const base::FilePath& path) {
+  delete g_database_path_override_for_testing;
+  if (path.empty()) {
+    g_database_path_override_for_testing = nullptr;
+  } else {
+    g_database_path_override_for_testing = new base::FilePath(path);
+  }
+}
```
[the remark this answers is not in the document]
hlwarriner: done

alunev (approved):

PR #11485, opened 2026-07-21
Update memory benchmarks output files

yuying-y:
Modify rdk_kabuki_memory_benchmark.sh and rdk_yttv_memory_benchmark.sh to append a timestamp to OUTPUT_FILE to prevent collisions. Also create a basic git skill piece of documentation.

Issue: 477961848
TAG=agy
CONV=cd22c0bb-b327-47bf-83f5-c7af5d192072


yuying-y:
Thanks for the review! Note that the git skill file was removed from this PR via force push, but the local skill documentation has been updated to use `--no-track`.

sherryzy (approved):
LGTM.

PR #11494, opened 2026-07-22
Updates angle to m138 HEAD.

briantting:
Updates angle from e1dc0a7ab5d1f1f2edaa7e41447d873895e083bf to 4664989d810d3614c0af32933ba28e10a949b420.

Bug: 537464074

### DIFF
--- a/third_party/angle/BUILD.gn
+++ b/third_party/angle/BUILD.gn
@@ -537,7 +537,7 @@ template("angle_common_lib") {
       all_dependent_configs = [ ":angle_disable_pool_alloc" ]
     }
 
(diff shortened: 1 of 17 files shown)


briantting:
This PR purely just updates angle from one commit to another. The starting angle commit is what is in the Cobalt repo from the last time angle was brought into Cobalt as a subtree. The target angle commit is what is in the latest m138 DEPS.

The format of the 3 commits are

- revert all Cobalt changes by reverting the angle directory to starting angle commit
- update the angle directory to target angle commit
- revert the first commit to bring in all Cobalt changes (thankfully no conflicts)

PR #11501, opened 2026-07-22
media: Allow extra MIME parameters in decodingInfo

borongc:
Cobalt web applications frequently append custom MIME parameters to
strings, such as enableflushduringseek. The default implementation
is restrictive, failing validation if multiple parameters are
present or if the codecs parameter is not in the primary position.

This ensures MediaCapabilities correctly handles extra parameters,
preventing valid media configurations from being rejected due to
platform-specific metadata.

Issue: 519196649
Issue: 537784933

### DIFF
--- a/third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
+++ b/third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
@@ -287,6 +287,19 @@ bool IsValidMimeType(const String& content_type,
 
   const auto& parameters = parsed_content_type.GetParameters();
 
+#if BUILDFLAG(IS_COBALT)
+  // Web applications may append Cobalt-specific parameters to MIME type strings
+  // in any order (e.g. enableflushduringseek=true, enableresetaudiodecoder=true).
+  // Check across all parameters rather than assuming 'codecs' is the first parameter.
+  if (parameters.ParameterCount() == 0)
+    return true;
+
+  for (const auto& param : parameters) {
+    if (EqualIgnoringASCIICase(param.name, kCodecsMimeTypeParam))
+      return true;
+  }
+  return false;
+#else  // BUILDFLAG(IS_COBALT)
   if (parameters.ParameterCount() > 1)
     return false;
 
@@ -294,6 +307,7 @@ bool IsValidMimeType(const String& content_type,
     return true;
 
   return EqualIgnoringASCIICase(parameters.begin()->name, kCodecsMimeTypeParam);
+#endif // BUILDFLAG(IS_COBALT)
 }
 
 bool IsValidMediaConfiguration(const MediaConfiguration* configuration) {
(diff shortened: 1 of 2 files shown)


third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
```diff
@@ -287,13 +287,27 @@ bool IsValidMimeType(const String& content_type,
...
+  // Check across all parameters rather than assuming 'codecs' is the first parameter.
+  if (parameters.ParameterCount() == 0)
+    return true;
+
+  for (const auto& param : parameters) {
+    if (EqualIgnoringASCIICase(param.name, kCodecsMimeTypeParam))
+      return true;
+  }
+  return true;
+#else  // BUILDFLAG(IS_COBALT)
```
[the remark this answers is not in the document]
borongc: Done.

jasonzhangxx (approved):

third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
```diff
@@ -287,13 +287,27 @@ bool IsValidMimeType(const String& content_type,
...
+  // in any order (e.g. enableflushduringseek=true, enableresetaudiodecoder=true).
+  // Check across all parameters rather than assuming 'codecs' is the first parameter.
+  if (parameters.ParameterCount() == 0)
+    return true;
+
+  for (const auto& param : parameters) {
+    if (EqualIgnoringASCIICase(param.name, kCodecsMimeTypeParam))
+      return true;
+  }
+  return false;
```
hlwarriner, line 301: Just confirming that this logic is intended? It looks a bit surprising to me.

* If there are no parameters, consider it valid
* If there are parameters, and any is a mime type param, consider it valid
* If there are parameters, and none is a mime type param, consider it invalid

hlwarriner (approved):

third_party/blink/renderer/modules/media_capabilities/media_capabilities_test.cc
```diff
@@ -654,6 +654,36 @@ TEST(MediaCapabilitiesTests, BasicAudio) {
   EXPECT_TRUE(info->powerEfficient());
 }
 
+#if BUILDFLAG(IS_COBALT)
+TEST(MediaCapabilitiesTests, CobaltExtraMimeParameters) {
```
hlwarriner, line 658: Thanks for adding the unit test. If it's doable, consider also adding a test case for when there are parameters, but none are mime type params, and so `IsValidMimeType()` returns false.

hlwarriner (approved):

third_party/blink/renderer/modules/media_capabilities/media_capabilities.cc
borongc: Yes, this is intended as what Chromium does, except that for Cobalt, we check each MIME type, not just the first one.

third_party/blink/renderer/modules/media_capabilities/media_capabilities_test.cc
borongc: Done.

PR #11503, opened 2026-07-22
nplb: Increase NPLB vertical video tolerance

osagie98:
Relax the allowed difference between expected and actual media duration
from 100ms to 500ms in vertical video tests. This change aligns these
tests with similar NPLB test cases.

The previous 100ms threshold was too strict for Linux modular builds, which would fail to meet the timing requirement
consistently when running the full test suite.

Issue: 537819670

### DIFF
--- a/starboard/nplb/vertical_video_test.cc
+++ b/starboard/nplb/vertical_video_test.cc
@@ -162,8 +162,8 @@ TEST_P(VerticalVideoTest, WriteSamples) {
   ASSERT_NO_FATAL_FAILURE(player_fixture.WaitForPlayerEndOfStream());
 
   int64_t end_media_time = player_fixture.GetCurrentMediaTime();
-  const int64_t kMediaTimeAllowance = 100'000;  // 100ms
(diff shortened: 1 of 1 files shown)


starboard/nplb/vertical_video_test.cc
```diff
@@ -162,8 +162,8 @@ TEST_P(VerticalVideoTest, WriteSamples) {
   ASSERT_NO_FATAL_FAILURE(player_fixture.WaitForPlayerEndOfStream());
 
   int64_t end_media_time = player_fixture.GetCurrentMediaTime();
-  const int64_t kMediaTimeAllowance = 100'000;  // 100ms
-  EXPECT_NEAR(end_media_time, kDurationToPlay, kMediaTimeAllowance);
+  const int64_t kDurationDifferenceAllowance = 500'000;  // 500ms
```
borongc, line 165: Should this PR come with re-enable these tests?

osagie98: I'm planning to revert the earlier filter change instead. This PR doesn't have the filter change merged so it's testing on a pre filter branch

borongc (approved):

PR #11508, opened 2026-07-23
rdk: remove dangling raspi reference

anshugupta27:
The starboard/raspi directory has been deleted as part of the raspi-2 deprecation. This removes the dangling reference to //starboard/raspi/shared/system_gles2.cc for the rdk-rpi target.
Bug: 500021808


starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn
```diff
@@ -296,14 +296,7 @@ static_library("starboard_platform_sources") {
     sources += [ "speech/speech_synthesis_is_supported_executable.cc", ]
   }
 
-  if (starboard_target_platform == "rdk-rpi") {
```
arjungm, line 299: it says rdk here, maybe this is still relevant for that platform?

maybe double-check that this won't effect RDK or if it does, we might need to fix the starboard_target_platform name so that rdk and raspi are decoupled

arjungm (changes requested):

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn
andrewsavage1: Looks irrelevant, this isn't used a starboard_target_platform. @niranjanyardi ptal

niranjanyardi: this is upstream code ( starboard/contrib/rdk/) anything in here is upstretam code. we are only making changes to the RDK port not ht eupstream rpi port - so please revert this change

cobalt/build/evergreen-arm-hardfp-raspi/package.json
niranjanyardi, line 1: this change seems good

niranjanyardi (changes requested):

cobalt/build/evergreen-arm-hardfp-raspi/package.json
anshugupta27: sure, I will keep this change.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/BUILD.gn
anshugupta27: Sure, I will revert this change.

arjungm (approved):

niranjanyardi (approved):

PR #11511, opened 2026-07-23
cobalt: Fix typo in DialServerImpl

rakuco:
This typo breaks non-debug builds, which also explains why it was not caught by bots or the usual devel/debug local builds.

Bug: 508769374


rakuco:

andrewsavage1 (approved):

PR #11521, opened 2026-07-23
android:  Add java switch to enable kDomStorageSmartFlushing

ckim1346:
Enables the storage::kDomStorageSmartFlushing feature override to eliminate delayed commit timers for runtime storage writes, and explicitly invokes storage partition and cookie flushes during `CobaltActivity.onStop()` to prevent data loss of pending LocalStorage and SessionStorage writes upon application suspension or process termination.

Bug: 535324570


cobalt/app/app_event_runner.cc
```diff
@@ -142,6 +142,11 @@ class AppEventRunnerImpl : public AppEventRunner,
   }
 
   void DoStop() override {
+    auto* client = cobalt::CobaltContentBrowserClient::Get();
+    if (client) {
+      client->FlushCookiesAndLocalStorage(base::DoNothing());
+    }
+
```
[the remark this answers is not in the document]
ckim1346: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -562,6 +562,7 @@ protected void onPause() {
 
   @Override
   protected void onStop() {
+    CobaltContentBrowserClient.flushCookiesAndLocalStorage();
```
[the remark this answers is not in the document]
ckim1346: dev.cobalt.browser.CobaltContentBrowserClient is an existing Java class with JNI Zero bindings

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -576,6 +562,7 @@ protected void onPause() {
...
```
zhongqiliang: updateShellActivityVisible(false) below will trigger Kabuki to write local storage asynchronously, the flush here won't catch it.

ckim1346: Moved it to be called after `updateShellActivityVisible(false)` and inside `mFreezeRunnable` to catch any additional data written to LocalStorage during teardown if `DELAY_FREEZE_ON_BACKGROUND`.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -595,6 +595,7 @@ public void run() {
                 if (currentWebContents != null) {
                   currentWebContents.onFreeze();
                 }
+                CobaltContentBrowserClient.flushCookiesAndLocalStorage();
```
zhongqiliang: Have you tested this code locally? Will it crash the app?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -605,6 +606,8 @@ public void run() {
       }
     }
 
+    CobaltContentBrowserClient.flushCookiesAndLocalStorage();
```
zhongqiliang: What is the reason to call the flush method here?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: I did test locally -- didn't see any crashes after testing 5 times

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: From Gemini: 

• The Reason: On Android TV devices, Cobalt supports a delayed freeze optimization switch (`DELAY_FREEZE_ON_BACKGROUND`).                      
  • The Problem: When this switch is active, Android delays freezing the page for 1500 ms (1.5 seconds) after `onStop()` begins. JavaScript  continues running background teardown scripts during those 1.5 seconds and can write additional data to LocalStorage.                        
  • The Fix: Calling `flushCookiesAndLocalStorage()` inside mFreezeRunnable ensures that when the 1.5-second delay finishes and the page is frozen, any data written during that 1.5s window is committed to disk right as the page freezes.

zhongqiliang: So you would run flushCookiesAndLocalStorage() twice?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
zhongqiliang: Did you read the code here? https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/browser/cobalt_content_browser_client.cc;l=542-559?q=flushCookiesAndLocalStorage
There are CHECK(), do we know if these objects are still not null after 1.5 seconds delay?

ckim1346: That's fair, they're not 100% guaranteed to not be null during teardown. Should we change them to nullchecks instead? 
```
auto* web_contents = web_contents_observer_->web_contents();
  if (!web_contents) {
    std::move(callback).Run();
    return;
  }
content::RenderFrameHost* rfh = web_contents->GetPrimaryMainFrame();
  if (!rfh) {
    std::move(callback).Run();
    return;
  }
auto* storage_partition = rfh->GetStoragePartition();
  if (!storage_partition) {
    std::move(callback).Run();
    return;
  }
// Flushes localStorage.
storage_partition->Flush();
auto* cookie_manager = storage_partition->GetCookieManagerForBrowserProcess();
if (!cookie_manager) {
    std::move(callback).Run();
    return;
  }
cookie_manager->FlushCookieStore(std::move(callback));
```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
ckim1346: We could probably run it just once regardless of the freeze delay

johnxwork (approved):

PR #11523, opened 2026-07-23
android: Use INFO log for tunnel mode completion

jasonzhangxx:
When tunnel mode preroll finishes, the event is currently logged as an
error. Since this is an expected part of the playback process and not
an actual error condition, it should be logged as information instead.

Bug: 538270075


kjyoun (approved):

kjyoun:
Thanks for the change. It will help us and Jetski a lot.

PR #11530, opened 2026-07-24
cobalt: Remove unused codes under cobalt/shell

kjyoun:
Clean up the Android shell implementation by removing various unused
methods, constants, and imports across ContentViewRenderView, Shell,
and ShellManager classes. This dead code is no longer necessary for
the Cobalt Android shell's functionality.

[go/deadcode](https://goto.google.com/deadcode)

Issue: 413418478

### DIFF
--- a/cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
+++ b/cobalt/shell/android/java/src/dev/cobalt/shell/Shell.java
@@ -19,12 +19,7 @@ import static org.chromium.build.NullUtil.assumeNonNull;
 
 import android.app.Activity;
 import android.content.Context;
-import android.graphics.Rect;
 import android.text.TextUtils;
-import android.view.ActionMode;
-import android.view.Menu;
-import android.view.MenuItem;
-import android.view.View;
 import android.view.ViewGroup;
 import android.widget.FrameLayout;
 import dev.cobalt.shell.ContentViewRenderView;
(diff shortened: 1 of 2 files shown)


cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -134,18 +134,6 @@ protected void onWindowVisibilityChanged(int visibility) {
         }
     }
 
-    /**
```
borongc, line 154: Not sure about the plan for rebase Cobalt shell, these are mostly the functions from upstream Content shell, if we remove them, it will increase the difficulty while rebasing Cobalt shell.

We should clearly check with browser team to see if there is any plan to rebase Cobalt shell.

kjyoun: @johnxwork 

What do you think about removing not used codes in code forked from chromium?

delete them or keep them ?

johnxwork (approved):
Thanks!

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
johnxwork: We don't want to remove unused codes from Chromium files to avoid rebase conflicts.

In this case, we have duplicated ContentViewRenderView.java specifically to make our own modifications. So this file specifically is fine to edit as it lives under cobalt/shell/

PR #11535, opened 2026-07-24
Gate FedCM under ENABLE_PRIVACY_SANDBOX_APIS for 27.lts

linxinan-yt:
Federated Credential Management (FedCM) is a privacy sandbox feature for browser users to sign in
with Identity Provider[1]. It mainly serves websites which need authentication or authorization of an individual
from a trusted external identity provider (IdP). 

As Cobalt is designed to be a single domain App, and does not need to sign in with another identity provider,
we can safely strip FedCM from our binary.

FedCM is not a deprecated privacy sandbox feature. This PR has added new test case to verify the removal
for future rebase.

[1] https://developer.chrome.com/docs/identity/fedcm/overview

Before: 

ls -lb out/android-arm_gold/libchrobalt.so
-rwxr-xr-x 1 foo primarygroup 66053692 Jul 28 22:44 out/android-arm_gold/libchrobalt.so

After: 

$ ls -lb out/android-arm_gold/libchrobalt.so
-rwxr-xr-x 1 foo primarygroup 65697340 Jul 28 23:08 out/android-arm_gold/libchrobalt.so

Saved 356KB for data size.

Bug: 505811196


third_party/blink/renderer/modules/payments/payment_request_test.cc
```diff
@@ -4,9 +4,7 @@
 
 #include "third_party/blink/renderer/modules/payments/payment_request.h"
 
-#include <memory>
```
linxinan-yt, line 7: why?

johnxwork (approved):

content/browser/browser_context.cc
```diff
@@ -28,14 +28,17 @@
 #include "base/trace_event/trace_event.h"
 #include "base/unguessable_token.h"
 #include "build/build_config.h"
+#include "build/buildflag.h"
```
haozheng-cobalt, line 31: [cobalt/shell/browser/shell_browser_context.h](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/shell/browser/shell_browser_context.h) inherits content/browser/browser_context.h, could you check if we could do the code change in cobalt/shell/browser/shell_browser_context.h instead of content/browser/browser_context.h? 
Same with other changes under /content. Ideally we should modify the corresponding Cobalt implementation instead of the parent class in /content.

third_party/blink/renderer/modules/payments/payment_request.cc
```diff
@@ -11,6 +11,7 @@
 #include "base/functional/bind.h"
```
haozheng-cobalt, line 11: Can't the entire payment files be removed?

haozheng-cobalt (changes requested):

content/browser/browser_context.cc
linxinan-yt: See content/browser/browser_context_impl.h (instantiated in the base BrowserContext constructor) directly owns std::unique_ptr<InMemoryFederatedPermissionContext> and browser_context.cc provides the default implementation calling into it. 

I don't think we can prune the InMemoryFederatedPermissionContext from the child class in cobalt's shell_browser_context.

third_party/blink/renderer/modules/payments/payment_request.cc
linxinan-yt: Yes and I will try in a follow-up PR. Let's focus on FedCM in this one.

haozheng-cobalt (approved):

third_party/blink/renderer/modules/payments/payment_request.cc
linxinan-yt: Filed #11785 for removing the payment API.

third_party/blink/renderer/bindings/bindings.gni
```diff
@@ -329,6 +329,35 @@ if (is_cobalt) {
       "*adproperties*",
       "*v8_ad_*",
       "*view_or_click_counts*",
+      "*authenticator_*",
```
hlwarriner: nit: consider alphabetically sorting these patterns.

third_party/blink/renderer/modules/BUILD.gn
```diff
@@ -334,7 +335,7 @@ source_set("modules_testing") {
 
   # If in the future privacy sandbox components are removed from the code base, ignore this if block.
   if (!enable_privacy_sandbox_apis) {
-    _ad_auction_testing_filtered_sources = filter_exclude(sources, [ "ad_auction/testing/*" ])
+    _ad_auction_testing_filtered_sources = filter_exclude(sources, [ "ad_auction/testing/*", "credentialmanagement/testing/*" ])
```
hlwarriner: Did you run `gn format` on this file? I'm surprised the lint check didn't complain about the list being constructed on one line. But maybe we don't run `gn format` on build files under `/third_party`, or something.

hlwarriner (approved):

third_party/blink/renderer/bindings/bindings.gni
linxinan-yt: Done.

third_party/blink/renderer/modules/BUILD.gn
linxinan-yt: Nice catch! I don't think our pre-commit covers /third_party. Maybe we should expand it. 
I have run it against the gn files in this PR. Many thanks.

hlwarriner: Thanks! I would say we should ensure the Cobalt-guarded sections are well-formatted but that we shouldn't fix formatting in upstream blocks in these files, as that would make rebasing more difficult.

PR #11538, opened 2026-07-24
android: Use activity Window surface for UI rendering

kjyoun:
Introduces support for rendering UI directly into the Activity's Window surface via Window.takeSurface(), as an alternative to an embedded child SurfaceView.

This behavior is abstracted behind WindowSurfaceBridge and is controlled by the `--use-window-surface-for-ui` command-line flag (disabled by default).

[go/cobalt-window-surface-for-ui](https://goto.google.com/cobalt-window-surface-for-ui)

Issue: 494590075

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -237,9 +237,9 @@ public abstract class CobaltActivity extends BaseCobaltActivity {
     // Set up the animation placeholder to be the SurfaceView. This disables the
     // SurfaceView's 'hole' clipping during animations that are notified to the window.
     mWindowAndroid.setAnimationPlaceholderView(
-        mShellManager.getContentViewRenderView().getSurfaceView());
+        mShellManager.getContentViewRenderView().getAnchorView());
     mA11yHelper =
-        new CobaltA11yHelper(this, mShellManager.getContentViewRenderView().getSurfaceView());
+        new CobaltA11yHelper(this, mShellManager.getContentViewRenderView().getAnchorView());
 
     if (mStartupUrl == null || mStartupUrl.isEmpty()) {
       String[] args = getStarboardBridge().getArgs();
--- a/cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
+++ b/cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
@@ -4,6 +4,7 @@
 
 package dev.cobalt.shell;
 
+import android.app.Activity;
 import android.content.Context;
 import android.graphics.PixelFormat;
 import android.view.Surface;
@@ -11,8 +12,11 @@ import android.view.SurfaceHolder;
 import android.view.MotionEvent;
 import android.view.SurfaceView;
 import android.view.View;
+import android.view.Window;
 import android.widget.FrameLayout;
 
+import org.chromium.base.CommandLine;
+import org.chromium.base.Log;
 import org.chromium.content_public.browser.WebContents;
 import org.chromium.ui.base.EventForwarder;
 import org.chromium.ui.base.WindowAndroid;
@@ -28,6 +32,8 @@ import org.jni_zero.NativeMethods;
  */
 @JNINamespace("cobalt")
 public class ContentViewRenderView extends FrameLayout {
+    private static final String TAG = "cobalt";
+
     // The native side of this object.
     private long mNativeContentViewRenderView;
     private WindowAndroid mWindowAndroid;
@@ -53,7 +59,12 @@ public class ContentViewRenderView extends FrameLayout {
     }
 
     protected SurfaceBridge createSurfaceBridge() {
-        return new SurfaceBridge();
+        if (CommandLine.getInstance().hasSwitch("use-window-surface-for-ui")) {
+            Log.i(TAG, "ContentViewRenderView: created using WindowSurfaceBridge");
+            return new WindowSurfaceBridge();
+        }
+        Log.i(TAG, "ContentViewRenderView: created with SurfaceView");
+        return new SurfaceViewBridge();
     }
 
     /**
@@ -62,7 +73,8 @@ public class ContentViewRenderView extends FrameLayout {
      * @param rootWindow The {@link WindowAndroid} this render view should be linked to.
      */
     public void onNativeLibraryLoaded(WindowAndroid rootWindow) {
-        assert !getSurfaceView().getHolder().getSurface().isValid()
+        assert mSurfaceBridge.getSurfaceView() == null
+                || !mSurfaceBridge.getSurfaceView().getHolder().getSurface().isValid()
             : "Surface created before native library loaded.";
         assert rootWindow != null;
         mNativeContentViewRenderView =
@@ -96,7 +108,10 @@ public class ContentViewRenderView extends FrameLayout {
                 // devices, where a relayout never happens. This bug is out of Chromium's
                 // control, but can be worked around by forcibly re-setting the visibility of
                 // the surface view. Otherwise, the screen stays black, and some tests fail.
-                getSurfaceView().setVisibility(getSurfaceView().getVisibility());
+                SurfaceView surfaceView = mSurfaceBridge.getSurfaceView();
+                if (surfaceView != null) {
+                    surfaceView.setVisibility(surfaceView.getVisibility());
+                }
 
                 onReadyToRender();
             }
@@ -108,7 +123,7 @@ public class ContentViewRenderView extends FrameLayout {
                         mNativeContentViewRenderView, ContentViewRenderView.this);
             }
         };
-        mSurfaceBridge.connect(surfaceCallback);
+        mSurfaceBridge.connect(surfaceCallback, rootWindow);
     }
 
     @Override
@@ -135,22 +150,12 @@ public class ContentViewRenderView extends FrameLayout {
     }
 
     /**
-     * Sets the background color of the surface view.  This method is necessary because the
-     * background color of ContentViewRenderView itself is covered by the background of
-     * SurfaceView.
-     * @param color The color of the background.
+     * Gets the View used for layout anchoring, animation placeholder, or accessibility (child
+     * SurfaceView in SurfaceView mode, or this host View in Window Surface mode).
      */
-    public void setSurfaceViewBackgroundColor(int color) {
-        if (getSurfaceView() != null) {
-            getSurfaceView().setBackgroundColor(color);
-        }
-    }
-
-    /**
-     * Gets the SurfaceView for this ContentViewRenderView
-     */
-    public SurfaceView getSurfaceView() {
-        return mSurfaceBridge.getSurfaceView();
+    public View getAnchorView() {
+        SurfaceView surfaceView = mSurfaceBridge.getSurfaceView();
+        return surfaceView != null ? surfaceView : this;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -246,17 +275,109 @@ protected void initialize(ContentViewRenderView renderView) {
...
+        protected void initialize(ContentViewRenderView renderView) {}
+
+        @Override
+        protected void connect(
+                SurfaceHolder.Callback surfaceCallback, WindowAndroid windowAndroid) {
+            if (windowAndroid == null) {
+                return;
+            }
+            android.app.Activity activity = windowAndroid.getActivity().get();
+            if (activity == null || activity.getWindow() == null) {
```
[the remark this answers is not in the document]
kjyoun: Done

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -53,7 +59,12 @@ public ContentViewRenderView(Context context) {
     }
 
     protected SurfaceBridge createSurfaceBridge() {
-        return new SurfaceBridge();
+        if (CommandLine.getInstance().hasSwitch("use-window-surface-for-ui")) {
```
jasonzhangxx, line 62: The function is called very early in CobaltActivity.createContent() right after CommandLine is initialized. My understanding is we have to use kimono gcl to run the experiments. And please double check if the change can be properly enabled via the experiments.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -246,17 +276,113 @@ protected void initialize(ContentViewRenderView renderView) {
...
+                    mWindowSurfaceHolder = null;
+                    surfaceCallback.surfaceDestroyed(holder);
+                }
+
+                @Override
+                public void surfaceRedrawNeeded(SurfaceHolder holder) {
+                    if (!(surfaceCallback instanceof SurfaceHolder.Callback2)) {
+                        return;
+                    }
+                    ((SurfaceHolder.Callback2) surfaceCallback).surfaceRedrawNeeded(holder);
```
jasonzhangxx, line 359: The input callback is defined as SurfaceHolder.Callback, but takeSurface requires a SurfaceHolder.Callback2. While casting the surfaceCallback to SurfaceHolder.Callback2 is necessary to implement the redraw notification, please ensure that the cast here is intentional and safe.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -195,47 +200,72 @@ protected SurfaceView createSurfaceView(Context context) {
         return new SurfaceView(context);
     }
 
-    /**
-     * @return whether the surface view is initialized and ready to render.
-     */
-    public boolean isInitialized() {
```
jasonzhangxx, line 201: nit: this function was removed, but it appears that it was unused anyway.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -246,17 +276,113 @@ protected void initialize(ContentViewRenderView renderView) {
...
+            Activity activity = windowAndroid.getActivity().get();
+            return activity != null ? activity.getWindow() : null;
+        }
+
+        @Override
+        protected void initialize(ContentViewRenderView renderView) {}
+
+        @Override
+        protected void connect(
+                SurfaceHolder.Callback surfaceCallback, WindowAndroid windowAndroid) {
```
jasonzhangxx, line 323: I think adding a new method `ContentViewRenderView.getWindow()` in ContentViewRenderView and then retrieving the `mWindow` during the `initialize()` would be a cleaner, more consistent approach that aligns better with the `SurfaceViewBridge` initialization pattern.

jasonzhangxx (commented):
The change overall looks good to me, but we need @johnxwork for a final review.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -28,6 +32,8 @@
  */
 @JNINamespace("cobalt")
 public class ContentViewRenderView extends FrameLayout {
+    private static final String TAG = "cobalt";
```
borongc, line 35: nit: use `import static dev.cobalt.util.Log.TAG;`

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -246,17 +276,113 @@ protected void initialize(ContentViewRenderView renderView) {
...
+                return;
+            }
+
+            mWindow.takeSurface(new SurfaceHolder.Callback2() {
+                @Override
+                public void surfaceCreated(SurfaceHolder holder) {
+                    mWindowSurfaceHolder = holder;
+                    if (mSurfaceFormat != null) {
+                        Log.i(TAG, "ContentViewRenderView: Applying pending format");
+                        mWindowSurfaceHolder.setFormat(mSurfaceFormat);
```
borongc: Should `mSurfaceFormat` be cleaned up once it is applied?

borongc:
nit: title should be `android: Use activity Window surface for UI rendering`

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -249,11 +237,9 @@ protected void createContent(final Bundle savedInstanceState) {
     // Set up the animation placeholder to be the SurfaceView. This disables the
     // SurfaceView's 'hole' clipping during animations that are notified to the window.
     mWindowAndroid.setAnimationPlaceholderView(
-        mShellManager.getContentViewRenderView().getSurfaceView());
+        mShellManager.getContentViewRenderView().getAnchorView());
     mA11yHelper =
-        new CobaltA11yHelper(this, mShellManager.getContentViewRenderView().getSurfaceView());
-
-    maybeRegisterNetworkRecoveryObserver();
+        new CobaltA11yHelper(this, mShellManager.getContentViewRenderView().getAnchorView());
```
borongc, line 242: We should check if this breaks a11y.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: Thanks for the tip, though I found that it does not work here [1]

It is the convention in this folder to define separate TAG constant:
https://source.corp.google.com/search?q=TAG&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:cobalt%2Fshell%2Fandroid%2Fjava%2Fsrc%2Fdev%2Fcobalt%2Fshell%2F

Filed https://b.corp.google.com/issues/539950432

---
[1]
`dev.cobalt.util.Log` belongs to the higher-level target `//cobalt/android:cobalt_coat_java` (package dev.cobalt.coat). Since `cobalt_shell_java` cannot depend on cobalt_coat_java (which would create a circular dependency), `dev.cobalt.util.Log` is not visible or  accessible to `ContentViewRenderView.java`

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: Double-checked the code and log

This code is executed after command line is initialized.

  1. CommandLine & Experiment Flag Initialization (Lines 163–184):
      • CommandLine.init() runs first.
      • JavaSwitches.getExtraCommandLineArgs(getJavaSwitches()) reads flags configured via Mendel / Kimono GCL experiment configs.
      • CommandLineOverrideHelper.getFlagOverrides() runs at line 181, merging intent parameters, meta-data, and JavaSwitches directly
      into CommandLine.getInstance().
  2. ContentViewRenderView Creation (Line 225):
      • mShellManager = new ShellManager(this) runs after CommandLineOverrideHelper.getFlagOverrides().
      • ShellManager instantiates ContentViewRenderView, which calls createSurfaceBridge().

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: ack.

I removed this method, instead of modifying to support window-surface mode.

Since it is not used, modification is pointless. We can implement this method, when it is actually used.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: We need to re-apply the requested format (e.g., PixelFormat.TRANSLUCENT for video overlay mode) so that newly created surfaces retain the target format across surface destruction and creation.

Added comments and changes variable names

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: yes, it is intended and I double-checked its safety
 
The downcast is guarded by `if (!(surfaceCallback instanceof SurfaceHolder.Callback2))`, which prevents any ClassCastException if a standard SurfaceHolder.Callback is passed.

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: Thanks for the suggestion! I looked into it but decided to keep the current implementation for the following reasons:
 
- Moving `getWindow()` to `initialize()`: `initialize()` is called from `ContentViewRenderView`'s constructor, at which point WindowAndroid is still null. `WindowAndroid` is only attached later when `onNativeLibraryLoaded()` calls connect().  == `mWindow` can only be retrieved inside connect().
- Moving `getWindow()` to `ContentViewRenderView`: `getWindow` specific to `WindowSurfaceBridge`. Keeping `getWindow()` encapsulated inside `WindowSurfaceBridge`  keeps `ContentViewRenderView` to remain decoupled from Window-surface-specific details.

There may be a way to restructure this further. It would require refactoring beyond the scope of this PR

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
kjyoun: IIUC, this does not break a11y and on local test, I could not see the regressions.

Will double-check with QA.

kjyoun:
> nit: title should be `android: Use activity Window surface for UI rendering`

Done

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
jasonzhangxx: ok, I see.

jasonzhangxx (approved):
LGTM, @johnxwork for final review.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
kjyoun: Filed https://b.corp.google.com/issues/540389275 for a11y testing

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
```diff
@@ -195,47 +200,72 @@ protected SurfaceView createSurfaceView(Context context) {
...
+        mSurfaceBridge.setFormat(format);
         ContentViewRenderViewJni.get().setOverlayVideoMode(
                 mNativeContentViewRenderView, ContentViewRenderView.this, enabled);
     }
 
     @CalledByNative
     private void didSwapFrame() {
-        if (getSurfaceView().getBackground() != null) {
+        SurfaceView surfaceView = mSurfaceBridge.getSurfaceView();
+        if (surfaceView == null) {
```
johnxwork, line 217: Since we now don't have a surface to set transparent herer, do we need anything else instead? Gemini is suggesting adding the following to the manifest, but I'm not sure if it's necessary:
<item name="android:windowBackground">@null</item>

johnxwork (approved):

cobalt/shell/android/java/src/dev/cobalt/shell/ContentViewRenderView.java
kjyoun: We don't need it in the manifest because we change the window transparency and format programmatically at runtime during launch. Keeping the theme background in XML also avoids breaking the cold-start splash screen

PR #11543, opened 2026-07-24
Fix IAMF specific box parsing bug in MP4 demuxer

osagie98:
The IamfSpecificBox::Parse method previously sliced the buffer from the
start of the box header instead of the current reader position. This
resulted in the IAMF configuration OBUs being parsed incorrectly
because the offset from previous reads within the box was not being
respected.

Issue: 537956110


borongc (approved):
RS approval, and waiting for our partners to verify it.

PR #11545, opened 2026-07-24
android: Limit ignored callbacks during flush

jasonzhangxx:
Update MediaCodecBridge to only ignore output buffer callbacks while
the codec is flushing.

Bug: 497004556


zhongqiliang (approved):
LGTM For ATV

borongc (approved):

PR #11550, opened 2026-07-26
media: Add page alignment and lockless decommit

xiaomings:
Add support for a page alignment flag in DecoderBufferAllocator to
control OS block allocation alignment. This allows for more granular
memory management when using the configurable decommit strategy.

Implement a lockless optimization during suspend by checking an atomic
flag before acquiring the allocator mutex. This prevents unnecessary
blocking of the UI thread when decommit on suspend is not enabled.

Additionally, simplify pointer arithmetic in StarboardMemoryAllocator
using uint8_t pointers and update H5vcc settings to expose the new
alignment configuration bit.

Bug: 454441375


hlwarriner (approved):
The `third_party/blink/renderer/modules/cobalt/h5vcc_settings/h_5_vcc_settings.cc` changes LGTM, as long as web clients are aware of the encoding change for the setting.

jasonzhangxx (approved):

PR #11551, opened 2026-07-26
media: Allow capability checks on worker threads

kjyoun:
Remove the main-thread requirement for media decoder support checks
within the renderer client. This change allows capability queries
to be performed from dedicated worker threads, which is necessary
for supporting Media Source Extensions (MSE) in workers.

The underlying Starboard calls are thread-safe and stateless, making
the main-thread enforcement unnecessary.

http://go/px-cobalt-mse-in-workers

Bug: 537351773

### DIFF
--- a/cobalt/renderer/cobalt_content_renderer_client.cc
+++ b/cobalt/renderer/cobalt_content_renderer_client.cc
@@ -271,7 +271,6 @@ CobaltContentRendererClient::GetSupportedKeySystems(
 
 bool CobaltContentRendererClient::IsDecoderSupportedAudioType(
     const ::media::AudioType& type) {
-  CHECK(content::RenderThread::IsMainThread());
   std::string mime = GetMimeFromAudioType(type);
   SbMediaSupportType support_type = kSbMediaSupportTypeNotSupported;
   if (!mime.empty()) {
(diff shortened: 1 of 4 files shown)


sideb0ard (approved):

borongc (approved):

cobalt/renderer/cobalt_content_renderer_client.cc
```diff
@@ -219,7 +219,6 @@ void CobaltContentRendererClient::RenderThreadStarted() {
 }
 
 void AddStarboardCmaKeySystems(::media::KeySystemInfos* key_system_infos) {
```
borongc, line 221: Maybe add a comment in the header, like:

```
  // NOTE: When MSE-in-Workers is enabled, media support queries can be invoked
  // on a DedicatedWorker thread in addition to the main RenderThread.
  // Therefore, these methods must remain thread-safe and must not assert
  // content::RenderThread::IsMainThread().
  std::unique_ptr<::media::KeySystemSupportRegistration>
  GetSupportedKeySystems(content::RenderFrame* render_frame,
                         ::media::GetSupportedKeySystemsCB cb) override;
  bool IsDecoderSupportedAudioType(const ::media::AudioType& type) override;
  bool IsDecoderSupportedVideoType(const ::media::VideoType& type) override;
```

borongc: Or in each function:
```
  // Can be called on the main thread or a DedicatedWorker thread (MSE-in-Workers).
```

kjyoun: good point. 

I added comment in the header about thread safety requirement about those methods and added unit tests to ensure the thread-safety contract isn't broken.

PR #11554, opened 2026-07-27
android: Remove shell close call in destroy

zhongqiliang:
Remove the mActiveShell.close() call in the ShellManager destroy
method. This call was identified as a source of crashes while the
application is in the background. The shell remains properly detached
through removeShell(). 

Bug: 534798847

### DIFF
--- a/cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
+++ b/cobalt/shell/android/java/src/dev/cobalt/shell/ShellManager.java
@@ -15,7 +15,6 @@
 package dev.cobalt.shell;
 
 import android.content.Context;
-import dev.cobalt.shell.ContentViewRenderView;
 import org.chromium.base.ThreadUtils;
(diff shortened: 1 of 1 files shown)


zhongqiliang:
The only change is removed `mActiveShell.close();` in onDestroy().

Awallky (approved):
This looks mostly like pre-commit formatting change and the removal of shell.close() as described.
Can we follow up with a browser test to protect against regressions here?

jasonzhangxx:
Can we simply revert #11376 and then test and re-apply new pr?

The app behavior somehow has been changed for me. Now the app will start with the last watch page even after killing the app process.

zhongqiliang:
"The app behavior somehow has been changed for me. Now the app will start with the last watch page even after killing the app process."

This is something the media team should investigate. After this PR, my change should be gated behind the flag. 
Do you only see the behavior changed when the flag is enabled?

PR #11562, opened 2026-07-28
aosp: wrap kill() and redirect it to pthread_kill()

rogerzanoni:
Redirect a self-directed signal to the calling thread with pthread_kill so it cannot be intercepted by another processes/threads.

Fixes the following issues:
  starboard: PosixSignalTest.KillSendsSignalToSelf
  dev.cobalt.coa: Thread[...,WaitingInMainSignalCatcherLoop,"Signal Catcher"]:
                  reacting to signal 10

Android's ART intercepts the SIGUSR1 signal and it never reaches the handlers set by nplb/cobalt.

Bug: 532068409


rogerzanoni:
Merged the change of using SIGUSR2 instead

PR #11162, opened 2026-06-30
android: Migrate Android lifecycle handling to AppEventDelegate

jellefoks:
android: Migrate Android lifecycle handling to AppEventDelegate

Route Android Activity lifecycle callbacks through the centralized AppEventDelegate instead of directly mutating Chromium WebContents. This ensures strict architectural layering by preventing Starboard platform code from depending on Cobalt application logic and allows internal C++ tasks to complete synchronously during state transitions.

This change also updates legacy IS_STARBOARD macros to IS_COBALT in browser and network service headers, configures MessagePumpAndroid to support nested run loops synchronously on Android to handle lifecycle events without re-entering the native looper, and adds LOG(FATAL) assertions to legacy WebContentsAndroid methods.

Rationale for PerformanceImpl and BaseStarboardBridge simplification:
- Clock Alignment: On Android, Java System.nanoTime() / 1000L and C++ starboard::CurrentMonotonicTime() are both backed by the Linux kernel's CLOCK_MONOTONIC clock source in microsecond units. On-device empirical testing confirms their values match within ~48 microseconds (strictly the JNI dispatch overhead).
- Guaranteed Population: app_startup_timestamp_ is populated at startup via AppEventBridge.handleStartEvent in CobaltActivity.java. Thus app_startup_timestamp_.has_value() is always true on Android, matching all other Starboard platforms.
- Eliminating Redundancies: The #if BUILDFLAG(IS_ANDROIDTV) fallback in PerformanceImpl::GetAppStartupTimeStamp and the complex measureAppStartTimestamp() calculation in BaseStarboardBridge are removed. BaseStarboardBridge.getAppStartTimestamp() now queries BaseCobaltActivity.getAppStartTimestamp() directly.

Bug: 448156280

PR #11569, opened 2026-07-28
android: Clean up Cobalt Android startup logic

jellefoks:
android: Clean up Cobalt Android startup logic

Update legacy IS_STARBOARD macros to IS_COBALT in browser and network service headers. Configure MessagePumpAndroid to support nested run loops synchronously on Android to handle lifecycle events without re-entering the native looper.

This change also prevents initialization crashes by guarding against early delegate calls during library loading and ensures browser startup tasks run synchronously.

Bug: 338274026

BUG=b/338274026
TAG=agy
CONV=006ddb93-1731-4dba-b801-1f6fb67f1fd4

### DIFF
--- a/base/message_loop/message_pump_android.cc
+++ b/base/message_loop/message_pump_android.cc
@@ -4,6 +4,8 @@
 
 #include "base/message_loop/message_pump_android.h"
 
+#include "build/build_config.h"
+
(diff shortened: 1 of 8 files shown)


jellefoks:
Closing in favor of consolidated PR #11162.

PR #11570, opened 2026-07-28
android: Assert legacy WebContents methods unused

jellefoks:
android: Assert legacy WebContents methods unused

Add LOG(FATAL) assertions to legacy WebContentsAndroid methods to ensure they are never invoked directly in Cobalt builds. This includes UpdateWebContentsVisibility, OnFreeze, OnResume, SetFocus, and SuspendAllMediaPlayers. Since Cobalt manages its own lifecycle and media state, these standard Chromium Android paths are unsupported and must be unreachable.

Bug: 338274026

BUG=b/338274026
TAG=agy
CONV=006ddb93-1731-4dba-b801-1f6fb67f1fd4


jellefoks:
Closing in favor of consolidated PR #11162.

PR #11579, opened 2026-07-28
media: Remove forced WebAudio sample rate

sideb0ard:
Previously, Cobalt forced the WebAudio AudioContext to 16kHz to match
microphone capture rates and avoid performance costs from resampling.
However, hard-coding the sample rate breaks other audio features that
expect higher fidelity, such as the YouTube startup sound.

With cl/963546378 I updated Kabuki to specify the sample rate and sinkId
(which allows a silent sink), so we no longer need to force the sample rate
change within Cobalt, relying on the application to specify.

This change removes our previous hard-coded modification, enables a chromium
flag which bypasses WebAudio resampling, and adds a log message for easy
verification of sample rate used.

Bug: 530518636

### DIFF
--- a/third_party/blink/common/features.cc
+++ b/third_party/blink/common/features.cc
@@ -2765,7 +2765,9 @@ BASE_FEATURE_PARAM(bool,
 // handle any necessary resampling, potentially reducing latency and overhead.
 BASE_FEATURE(kWebAudioRemoveAudioDestinationResampler,
              "WebAudioRemoveAudioDestinationResampler",
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_COBALT)
+             base::FEATURE_ENABLED_BY_DEFAULT);
+#elif BUILDFLAG(IS_ANDROID)
              base::FEATURE_DISABLED_BY_DEFAULT);
 #else
              base::FEATURE_DISABLED_BY_DEFAULT);
--- a/third_party/blink/common/features.cc
+++ b/third_party/blink/common/features.cc
@@ -2811,7 +2811,9 @@ BASE_FEATURE_PARAM(bool,
 // handle any necessary resampling, potentially reducing latency and overhead.
 BASE_FEATURE(kWebAudioRemoveAudioDestinationResampler,
              "WebAudioRemoveAudioDestinationResampler",
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_COBALT)
+             base::FEATURE_ENABLED_BY_DEFAULT);
+#elif BUILDFLAG(IS_ANDROID)
              base::FEATURE_DISABLED_BY_DEFAULT);
 #else
              base::FEATURE_DISABLED_BY_DEFAULT);
--- a/third_party/blink/renderer/modules/webaudio/audio_context.cc
+++ b/third_party/blink/renderer/modules/webaudio/audio_context.cc
@@ -60,12 +60,6 @@
 #include <stdio.h>
 #endif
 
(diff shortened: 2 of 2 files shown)


third_party/blink/renderer/modules/webaudio/audio_context.cc
```diff
@@ -363,6 +343,16 @@ AudioContext::AudioContext(LocalDOMWindow& window,
 
   Initialize();
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  LOG(INFO) << "Cobalt - AudioContext Initialized:"
+            << " [Sample Rate = " << sampleRate() << " Hz]"
+            << " [Sink Type = "
+            << (sink_descriptor_.Type() == WebAudioSinkDescriptor::kAudible
+                    ? "Audible"
+                    : "Silent")
```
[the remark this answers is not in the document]
sideb0ard: incorrect - this is hallucinated.
compiles fine.

third_party/blink/renderer/modules/webaudio/audio_context.cc
```diff
@@ -363,6 +343,16 @@ AudioContext::AudioContext(LocalDOMWindow& window,
 
   Initialize();
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  LOG(INFO) << "Cobalt - AudioContext Initialized:"
```
kjyoun: nit: can you make this log follow the existing pattern ?

https://screenshot-v2.corp.google.com/7n8m87iv1lnc8

"message: foo=1, bar=2"



```
      LOG(INFO) << "Cobalt: AudioContext initialized [sample_rate=" << sampleRate()
                << "Hz, sink_type="
                << (sink_descriptor_.Type() ==
                            WebAudioSinkDescriptor::AudioSinkType::kAudible
                        ? "Audible"
                        : "Silent")
                << "]";
``

**Review by kjyoun — APPROVED** [review]

File: third_party/blink/renderer/modules/webaudio/audio_context.cc (hunk 2 above)
Comment (sideb0ard) [reply third_party/blink/renderer/modules/webaudio/audio_context.cc]: done, thank you!

File: third_party/blink/common/features.cc [hunk 3]
```diff
@@ -2811,7 +2811,9 @@ BASE_FEATURE_PARAM(bool,
 // handle any necessary resampling, potentially reducing latency and overhead.
 BASE_FEATURE(kWebAudioRemoveAudioDestinationResampler,
              "WebAudioRemoveAudioDestinationResampler",
-#if BUILDFLAG(IS_ANDROID)
+#if BUILDFLAG(IS_COBALT)
+             base::FEATURE_ENABLED_BY_DEFAULT);
+#elif BUILDFLAG(IS_ANDROID)
```
yell0wd0g, line 2816: Would it be better to inject this flag in the switches.cc
files? (e.g. [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/app/cobalt_switch_defaults_starboard.cc;l=89) and/or [here](https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/util/JavaSwitches.java;l=305)?
Thinking what'd be best for rebases.

yell0wd0g (approved):
RS LGTM with a comment in case it helps

third_party/blink/common/features.cc
sideb0ard: done, thanks! That works well, as it turns out upstream has moved that flag.

PR #11587, opened 2026-07-29
Adds update DEPS logic to autoroll.

briantting:
When autorolling Chromium cherry picks, Chromium submodules that have been brought into Cobalt as directories are also updated.

Bug: 505073626

### DIFF
--- a/.github/scripts/autoroll_lts.py
+++ b/.github/scripts/autoroll_lts.py
@@ -3,10 +3,25 @@
 import argparse
 from collections import defaultdict
 import enum
+import os
 import re
 import shutil
 import subprocess
 import sys
+
+_COBALT_SUBMODULE_DIRS = [
+    'net/third_party/quiche/src',
+    'third_party/angle',
+    'third_party/boringssl/src',
+    'third_party/cpuinfo/src',
+    'third_party/googletest/src',
+    'third_party/icu',
+    'third_party/libc++/src',
+    'third_party/perfetto',
+    'third_party/skia',
+    'third_party/webrtc',
+    'v8',
+]
 
 
 @enum.unique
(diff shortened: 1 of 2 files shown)


.github/scripts/autoroll_lts.py
```diff
@@ -183,6 +198,28 @@ def get_submodule_root_dirs():
       {line.split(' ', 1)[1].split('/')[0] for line in paths.splitlines()})
 
 
+def remove_local_checkout():
+  log('Removing everything...')
+  roots = get_submodule_root_dirs()
+  if roots:
+    run(['rm', '-rf', '--'] + roots)
```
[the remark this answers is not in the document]
briantting: wrong, it is a list already

andrewsavage1 (approved):

.github/scripts/autoroll_lts.py
```diff
@@ -3,11 +3,26 @@
 import argparse
 from collections import defaultdict
 import enum
+import os
 import re
 import shutil
 import subprocess
 import sys
 
+COBALT_SUBMODULE_DIRS_ = [
```
linxinan-yt: Per Google's style guide,  should be _COBALT_SUBMODULE_DIRS if this is module internal.

.github/scripts/autoroll_lts.py
```diff
@@ -3,11 +3,26 @@
...
+    'net/third_party/quiche/src',
```
linxinan-yt, line 13: Is it possible to get them from DEPS file?

linxinan-yt (approved):
LGTM % comments

.github/scripts/autoroll_lts.py
briantting: Yes in a later PR, it's a bit messy and brittle because of how it depends on searching for commented out sections with "Cobalt: imported" and some DEPS rely on a skia_revision variable and some DEPS just have the hash directly in the url.

.github/scripts/autoroll_lts.py

PR #11599, opened 2026-07-29
media: Remove unreachable else block in VideoFrameTracker

borongc:
Simplify the frame tracking logic by removing an unreachable else
block and a redundant conditional check in UpdateDroppedFrames.
This code path was found to be unreachable given the surrounding
logic, and removing it results in cleaner and more maintainable
video frame status reporting.

Issue: 540546196


kjyoun (approved):
Thanks for handling this.

PR #11602, opened 2026-07-30
Fix build failure in KeepAliveURLLoaderService::FactoryContext on staging

andrewsavage1:
Fixes C++ compilation syntax error in KeepAliveURLLoaderService::FactoryContext constructor when ENABLE_PRIVACY_SANDBOX_APIS is disabled.


.github/actions/on_host_tests/action.yaml
```diff
@@ -27,17 +27,17 @@ runs:
       uses: ./.github/actions/depot_tools
       with:
         run_sync: 'false'
-    - name: Download Artifacts
```
andrewsavage1, line 30: We should also bring this change to the main branch

briantting (approved):

PR #11605, opened 2026-07-30
ci: Checkout JUnit dependencies before downloading artifacts in on_host_tests

andrewsavage1:
Reorder Checkout JUnit Dependencies step to run prior to Download Artifacts in on_host_tests action. This prevents git sparse-checkout add from executing an index refresh after artifact download, which can purge untracked test_artifacts.tar.zstd before extraction.

Example failure: https://github.com/youtube/cobalt/actions/runs/30506594812/job/90762388801

Bug: 431780245

### DIFF
--- a/.github/actions/on_host_tests/action.yaml
+++ b/.github/actions/on_host_tests/action.yaml
@@ -27,10 +27,6 @@ runs:
       uses: ./.github/actions/depot_tools
       with:
         run_sync: 'false'
-    - name: Download Artifacts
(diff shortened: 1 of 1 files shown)


andrewsavage1:
Note this fixes an issue where the zstd package can be deleted during the download dependencies step on the staging branch, but I believe this issue could happen on the main branch too

arjungm (approved):
good catch!

PR #11637, opened 2026-07-30
media: Refine configurable decommit strategies

xiaomings:
Add support for configurable decommit flags in DecoderBufferAllocator,
including memset_on_reclaim, mark_as_cold_on_reclaim,
periodic_decommit, and separate enable_decommit_on_suspend and
aggressive_decommit_on_suspend flags.

Ensure allocate_with_page_alignment remains on bit 25 for backwards
compatibility, with aggressive_decommit_on_suspend moved to bit 26.
Reset mark_as_cold_on_reclaim to false with an info log when MADV_COLD is
unsupported on the target platform. Use TryToDecommitOneBlock for periodic
decommit tasks.

Avoid lock inversion deadlock when stopping periodic decommit loop.
Fix block_decommit_attempt_count_ identifier in ReuseAllocatorBase.
Update allocator docstrings for DecommitMode and new strategy flags.
Support EnableConfigurableDecommitStrategyV2 setting alias for rollout.

Bug: 454441375

### DIFF
--- a/media/starboard/bidirectional_fit_decoder_buffer_allocator_strategy.h
+++ b/media/starboard/bidirectional_fit_decoder_buffer_allocator_strategy.h
@@ -45,14 +45,20 @@ class BidirectionalFitDecoderBufferAllocatorStrategy
   // these are aggressively decommitted (e.g. using MADV_DONTNEED).
   // |aggressive_decommit_on_suspend|: Whether to aggressively decommit all idle
   // blocks when app is suspended.
+  // |allocate_with_page_alignment|: Whether fallback allocations align to page
+  // boundary. |memset_on_reclaim|: Whether to zero out fallback blocks on
+  // reclamation. |mark_as_cold_on_reclaim|: Whether to advise MADV_COLD on
+  // reclamation.
   BidirectionalFitDecoderBufferAllocatorStrategy(
       size_t initial_capacity,
       size_t allocation_increment,
       bool enable_decommit_on_idle,
       size_t retain_blocks,
       size_t conservative_decommit_blocks,
-      bool aggressive_decommit_on_suspend = false,
-      bool allocate_with_page_alignment = true)
+      bool aggressive_decommit_on_suspend,
+      bool allocate_with_page_alignment,
+      bool memset_on_reclaim,
+      bool mark_as_cold_on_reclaim)
       : fallback_allocator_(enable_decommit_on_idle,
                             allocate_with_page_alignment),
         bidirectional_fit_allocator_(&fallback_allocator_,
@@ -62,7 +68,9 @@ class BidirectionalFitDecoderBufferAllocatorStrategy
                                      enable_decommit_on_idle,
                                      retain_blocks,
                                      conservative_decommit_blocks,
-                                     aggressive_decommit_on_suspend) {}
+                                     aggressive_decommit_on_suspend,
+                                     memset_on_reclaim,
+                                     mark_as_cold_on_reclaim) {}
 
   void* Allocate(DemuxerStream::Type type, size_t size) override {
     return bidirectional_fit_allocator_.Allocate(size);
@@ -84,6 +92,10 @@ class BidirectionalFitDecoderBufferAllocatorStrategy
 
   void DecommitAllDecommitableBlocks() override {
     bidirectional_fit_allocator_.DecommitAllDecommitableBlocks();
+  }
+
+  void TryToDecommitOneBlock(int cadence) override {
+    bidirectional_fit_allocator_.TryToDecommitOneBlock(cadence);
   }
 
  private:
--- a/media/starboard/decoder_buffer_allocator.cc
+++ b/media/starboard/decoder_buffer_allocator.cc
@@ -14,6 +14,8 @@
 
 #include "media/starboard/decoder_buffer_allocator.h"
 
+#include <sys/mman.h>  // For MADV_COLD
+
 #include <algorithm>
 
 #include "base/feature_list.h"
@@ -37,6 +39,8 @@
 namespace media {
 
 namespace {
+
+constexpr base::TimeDelta kDefaultPeriodicDecommitInterval = base::Seconds(5);
 
 // The current default AllocatorStrategy is EmbeddedMetadataReuseAllocatorBase.
 // To see more context as to why this is the case, see b/487332929.
@@ -88,6 +92,8 @@ DecoderBufferAllocator::DecoderBufferAllocator(
 }
 
 DecoderBufferAllocator::~DecoderBufferAllocator() {
+  StopPeriodicDecommitLoop();
+
   base::AutoLock scoped_lock(mutex_);
 
   if (strategy_) {
@@ -120,7 +126,7 @@ void DecoderBufferAllocator::ReleaseIdleMemory() {
 }
 
 void DecoderBufferAllocator::DecommitAllDecommitableBlocks() {
-  if (!decommit_on_suspend_enabled_.load(std::memory_order_relaxed)) {
+  if (!decommit_on_suspend_enabled_.load(std::memory_order_acquire)) {
     return;
   }
   base::AutoLock scoped_lock(mutex_);
@@ -253,7 +259,10 @@ void DecoderBufferAllocator::UpdateAllocatorStrategy(
 base::expected<void, std::string> DecoderBufferAllocator::SetSetting(
     const std::string& name,
     int value) {
-  if (name == "DecoderBuffer.EnableConfigurableDecommitStrategy") {
+  // Temporarily accept both setting names to support backward-compatible
+  // experiment rollouts before transitioning to the V2 format.
+  if (name == "DecoderBuffer.EnableConfigurableDecommitStrategy" ||
+      name == "DecoderBuffer.EnableConfigurableDecommitStrategyV2") {
     if (value <= 0) {
       // Explicitly allow non-positive values as placebo.
       return base::ok();
@@ -268,12 +277,21 @@ base::expected<void, std::string> DecoderBufferAllocator::SetSetting(
     // ideal, but it simplifies experiment setup in the current framework.
     //
     // - flags (8 bits):
-    //   - aggressive_decommit_on_suspend (bit 24): When set (LSB is non-zero),
-    //     enables aggressive MADV_DONTNEED decommit on all idle blocks when
-    //     app suspends/hides.
-    //   - allocate_with_page_alignment (bit 25): When set (second bit is
+    //   - enable_decommit_on_suspend (bit 24): When set (bit 24 is non-zero),
+    //     enables decommit on all idle blocks when app suspends/hides.
+    //   - allocate_with_page_alignment (bit 25): When set (bit 25 is
     //     non-zero), forces new OS block allocations to be page-aligned and
     //     rounded up to page sizes. Default is true.
+    //   - aggressive_decommit_on_suspend (bit 26): When set (bit 26 is
+    //     non-zero), aggressively decommits all idle blocks (ignoring retain
+    //     window) on suspend.
+    //   - memset_on_reclaim (bit 27): When set (bit 27 is non-zero),
+    //     memsets reclaimed blocks to 0.
+    //   - mark_as_cold_on_reclaim (bit 28): When set (bit 28 is non-zero),
+    //     uses MADV_COLD on reclaimed blocks.
+    //   - periodic_decommit (bit 29): When set (bit 29 is non-zero),
+    //     periodically decommits idle blocks in the background via
+    //     base::ThreadPool.
     // - block_size (8 bits): Specifies both the initial pool capacity and the
     //   fallback allocation increment.
     // - retain_blocks (8 bits): Specifies the first `retain_blocks` blocks of
@@ -297,15 +315,24 @@ base::expected<void, std::string> DecoderBufferAllocator::SetSetting(
     // Note: A value of 0 or less will not enable this feature (handled as a
     // placebo).
     uint32_t unsigned_value = static_cast<uint32_t>(value);
-    bool aggressive_decommit_on_suspend = ((unsigned_value >> 24) & 0x01) != 0;
-    bool allocate_with_page_alignment = ((unsigned_value >> 24) & 0x02) != 0;
+    uint32_t flags = unsigned_value >> 24;
+
+    bool enable_decommit_on_suspend = (flags & 0x01) != 0;
+    bool allocate_with_page_alignment = (flags & 0x02) != 0;
+    bool aggressive_decommit_on_suspend = (flags & 0x04) != 0;
+    bool memset_on_reclaim = (flags & 0x08) != 0;
+    bool mark_as_cold_on_reclaim = (flags & 0x10) != 0;
+    bool periodic_decommit = (flags & 0x20) != 0;
+
     int block_size_mb = (unsigned_value >> 16) & 0xFF;
     int retain_blocks = (unsigned_value >> 8) & 0xFF;
     int conservative_decommit_blocks = unsigned_value & 0xFF;
+
     EnableConfigurableDecommitStrategy(
         block_size_mb * 1024 * 1024, retain_blocks,
-        conservative_decommit_blocks, aggressive_decommit_on_suspend,
-        allocate_with_page_alignment);
+        conservative_decommit_blocks, enable_decommit_on_suspend,
+        allocate_with_page_alignment, aggressive_decommit_on_suspend,
+        memset_on_reclaim, mark_as_cold_on_reclaim, periodic_decommit);
     return base::ok();
   }
   if (name == "DecoderBuffer.EnableMediaBufferPoolAllocatorStrategy") {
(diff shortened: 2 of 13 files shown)


media/starboard/starboard_memory_allocator.h
```diff
@@ -82,12 +83,19 @@ class StarboardMemoryAllocator : public starboard::Allocator {
 
     if (aligned_start < aligned_end) {
       size_t aligned_size = aligned_end - aligned_start;
+
+      if (mode == DecommitMode::kCold) {
+        madvise(aligned_start, aligned_size, MADV_COLD);
+        return;
+      }
```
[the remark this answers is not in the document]
xiaomings: MADV_COLD is defined in Cobalt's toolchain header (third_party/musl/include/sys/mman.h), so DecommitMode::kCold directly issues MADV_COLD.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -324,19 +334,24 @@ void DecoderBufferAllocator::EnableConfigurableDecommitStrategy(
...
+    bool memset_on_reclaim,
+    bool mark_as_cold_on_reclaim) {
   auto* allocator = Get();
   CHECK(allocator);
-  if (aggressive_decommit_on_suspend) {
+  if (enable_decommit_on_suspend) {
     // This optimization is enable only, and isn't expected to be disabled.
     allocator->decommit_on_suspend_enabled_.store(true,
                                                   std::memory_order_relaxed);
   }
```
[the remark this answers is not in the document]
xiaomings: Fixed in the latest commit update. aggressive_decommit_on_suspend is now gated on enable_decommit_on_suspend.

starboard/common/reuse_allocator_base.cc
```diff
@@ -115,6 +120,13 @@ void ReuseAllocatorBase::ReclaimFallbackBlocks() {
   allocation_counter_ = 0;
   for (size_t i = 0; i < fallback_allocations_.size(); ++i) {
     auto& fallback_block = fallback_allocations_[i];
+    if (memset_on_reclaim_) {
+      memset(fallback_block.address, 0, fallback_block.size);
+    }
+    if (mark_as_cold_on_reclaim_) {
+      fallback_allocator_->Decommit(fallback_block.address, fallback_block.size,
+                                    DecommitMode::kCold);
+    }
```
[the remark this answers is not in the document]
xiaomings: Fixed in the latest commit update. Added state checks so memset and DecommitMode::kCold do not execute on already decommitted or freed blocks.

media/starboard/starboard_memory_allocator.h
```diff
@@ -82,12 +84,19 @@ class StarboardMemoryAllocator : public starboard::Allocator {
...
```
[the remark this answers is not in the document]
xiaomings: Fixed in the latest commit update. Wrapped `madvise(..., MADV_COLD)` with `#if defined(MADV_COLD)` ... `#endif  // defined(MADV_COLD)`.

starboard/common/reuse_allocator_base.h
```diff
@@ -129,6 +131,12 @@ class ReuseAllocatorBase : public Allocator {
   // Whether to aggressively decommit all idle blocks on app suspend.
   const bool aggressive_decommit_on_suspend_ = false;
 
+  // Whether to memset fallback memory blocks to 0 when reclaimed.
+  const bool memset_on_reclaim_ = false;
+
+  // Whether to mark fallback memory blocks cold (MADV_COLD) when reclaimed.
+  const bool mark_as_cold_on_reclaim_ = false;
```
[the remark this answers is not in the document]
xiaomings: There is no such requirement in the style guide regarding const member variables.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -324,19 +335,24 @@ void DecoderBufferAllocator::EnableConfigurableDecommitStrategy(
...
```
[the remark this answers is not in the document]
xiaomings: `aggressive_decommit_on_suspend` can never be true when `enable_decommit_on_suspend` is false, which is guaranteed in the parsing code of the setting.

media/starboard/bidirectional_fit_decoder_buffer_allocator_strategy.h
```diff
@@ -45,14 +45,20 @@ class BidirectionalFitDecoderBufferAllocatorStrategy
...
       size_t allocation_increment,
       bool enable_decommit_on_idle,
       size_t retain_blocks,
       size_t conservative_decommit_blocks,
-      bool aggressive_decommit_on_suspend = false,
-      bool allocate_with_page_alignment = true)
+      bool aggressive_decommit_on_suspend,
+      bool allocate_with_page_alignment,
+      bool memset_on_reclaim,
+      bool mark_as_cold_on_reclaim)
```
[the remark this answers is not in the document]
xiaomings: Not in the scope of this PR.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -297,15 +312,20 @@ base::expected<void, std::string> DecoderBufferAllocator::SetSetting(
     // Note: A value of 0 or less will not enable this feature (handled as a
     // placebo).
     uint32_t unsigned_value = static_cast<uint32_t>(value);
-    bool aggressive_decommit_on_suspend = ((unsigned_value >> 24) & 0x01) != 0;
+    bool enable_decommit_on_suspend = ((unsigned_value >> 24) & 0x01) != 0;
     bool allocate_with_page_alignment = ((unsigned_value >> 24) & 0x02) != 0;
+    bool aggressive_decommit_on_suspend = ((unsigned_value >> 24) & 0x04) != 0;
+    bool memset_on_reclaim = ((unsigned_value >> 24) & 0x08) != 0;
+    bool mark_as_cold_on_reclaim = ((unsigned_value >> 24) & 0x10) != 0;
+    bool periodic_decommit = ((unsigned_value >> 24) & 0x20) != 0;
```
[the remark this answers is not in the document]
xiaomings: We don't have to maintain backwards compatibility here.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -14,6 +14,8 @@
 
 #include "media/starboard/decoder_buffer_allocator.h"
 
+#include <sys/mman.h>  // For MADV_COLD
```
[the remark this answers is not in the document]
xiaomings: We currently don't support WIN32.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -297,15 +315,20 @@ base::expected<void, std::string> DecoderBufferAllocator::SetSetting(
     // Note: A value of 0 or less will not enable this feature (handled as a
     // placebo).
     uint32_t unsigned_value = static_cast<uint32_t>(value);
```
kjyoun, line 317: how about this? 


```
uint32_t unsigned_value = static_cast<uint32_t>(value);
bool enable_decommit_on_suspend     = (unsigned_value & (1 << 24)) != 0;
bool allocate_with_page_alignment   = (unsigned_value & (1 << 25)) != 0;
bool aggressive_decommit_on_suspend = (unsigned_value & (1 << 26)) != 0;
bool memset_on_reclaim              = (unsigned_value & (1 << 27)) != 0;
bool mark_as_cold_on_reclaim        = (unsigned_value & (1 << 28)) != 0;
bool periodic_decommit              = (unsigned_value & (1 << 29)) != 0;
```

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -419,6 +485,42 @@ void DecoderBufferAllocator::EnsureStrategyIsCreated() {
             << " bytes for decoder buffer pool.";
 }
 
+void DecoderBufferAllocator::EnablePeriodicDecommitLoop() {
+  base::AutoLock scoped_lock(mutex_);
+  if (!periodic_decommit_state_) {
```
kjyoun: nit: early return  pattern?

```
void DecoderBufferAllocator::EnablePeriodicDecommitLoop() {
  base::AutoLock scoped_lock(mutex_);
  if (periodic_decommit_state_) {
    return;
  }

   periodic_decommit_state_ =
        base::MakeRefCounted<PeriodicDecommitState>(this);
  SchedulePeriodicDecommitTask_Locked();
}
```

media/starboard/decoder_buffer_allocator.h
```diff
@@ -119,6 +168,8 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
   bool has_pending_release_ GUARDED_BY(mutex_) = false;
   bool should_release_idle_memory_ GUARDED_BY(mutex_) = false;
   StrategyCreateCB experimental_strategy_create_cb_ GUARDED_BY(mutex_);
+  scoped_refptr<PeriodicDecommitState> periodic_decommit_state_
```
kjyoun, line 172: cant we use unique_ptr instead?

1] https://chromium.googlesource.com/website/+/HEAD/site/developers/smart-pointer-guidelines/index.md#when-do-we-use-each-smart-pointer

> Ref-counted objects - use scoped_refptr<>, but better yet, rethink your design. Reference-counted objects make it difficult to understand ownership and destruction order, especially when multiple threads are involved. There is almost always another way to design your object hierarchy to avoid refcounting

media/starboard/starboard_memory_allocator.h
```diff
@@ -82,12 +84,21 @@ class StarboardMemoryAllocator : public starboard::Allocator {
 
     if (aligned_start < aligned_end) {
       size_t aligned_size = aligned_end - aligned_start;
+
+      if (mode == DecommitMode::kCold) {
+#if defined(MADV_COLD)
+        madvise(aligned_start, aligned_size, MADV_COLD);
+#endif  // defined(MADV_COLD)
```
kjyoun, line 91: do we at least have to give warning, not on MADV_COLD?
```
#else 
   LOG(WARNING) << "DecommitMode::kCold) is ignored on non-MADV_COLD build";
#endif
```

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -319,23 +342,57 @@ base::expected<void, std::string> DecoderBufferAllocator::SetSetting(
...
+    allocator->SchedulePeriodicDecommitTask_Locked();
+  });
+}
+
 // static
 void DecoderBufferAllocator::EnableConfigurableDecommitStrategy(
     int block_size,
     int retain_blocks,
     int conservative_decommit_blocks,
+    bool enable_decommit_on_suspend,
```
kjyoun, line 366: 6 consecutive boolean flags are too easy to mess up. The argument order mistakes can only be
  caught at runtime.

I think that we definitely need some grouping or using other data type.

something like this

```
struct DecommitOptions {
        bool enable_on_suspend = false;
        bool aggressive_on_suspend = false;
     }

struct ReclaimingOptions {
        bool allocate_with_page_alignment = false;
        bool memset = false;
        bool mark_as_cold = false;
     }
```

media/starboard/decoder_buffer_allocator.cc
xiaomings: Done. Simplified the flag extraction by extracting the flags byte once to avoid repeated shifts.

media/starboard/decoder_buffer_allocator.cc
xiaomings: Done.

media/starboard/decoder_buffer_allocator.h
xiaomings: We need to keep `scoped_refptr` here because `PeriodicDecommitState` is shared with background tasks posted to the `ThreadPool`. Since `DecoderBufferAllocator` can be destroyed while tasks are still pending, the tasks hold a reference to `PeriodicDecommitState` to ensure it remains valid. When `DecoderBufferAllocator` is destroyed, it detaches itself, and any pending tasks will safely do nothing when they execute. Since `base::WeakPtr` is not thread-safe and cannot be dereferenced on background threads, this ref-counted helper is the standard way to handle this lifetime requirement safely.

media/starboard/starboard_memory_allocator.h
xiaomings: We already log an `INFO` message in `EnableConfigurableDecommitStrategy` if `mark_as_cold_on_reclaim` is enabled but `MADV_COLD` is unsupported. Adding another log here would be redundant and, since this is experimental code where this function is called frequently, it could easily lead to log spam. We prefer to keep the implementation as is to avoid this.

media/starboard/decoder_buffer_allocator.cc
xiaomings: Acknowledged. Grouping these flags into a struct makes sense to improve safety. However, depending on how deep the struct needs to be passed down (e.g., into the template strategy classes), it could introduce significant extra changes to this PR. To keep this PR focused and easier to review, I will create a follow-up PR dedicated to refactoring this configuration logic, where we can better discuss and review the design.

media/starboard/decoder_buffer_allocator.cc
```diff
@@ -419,6 +489,42 @@ void DecoderBufferAllocator::EnsureStrategyIsCreated() {
...
+  SchedulePeriodicDecommitTask_Locked();
+}
+
+void DecoderBufferAllocator::StopPeriodicDecommitLoop() {
+  scoped_refptr<PeriodicDecommitState> state_to_detach;
+  {
+    base::AutoLock scoped_lock(mutex_);
+    state_to_detach = std::move(periodic_decommit_state_);
+  }
+  if (state_to_detach) {
```
kjyoun, line 507: nit: early return pattern like other methods(e.g. EnablePeriodicDecommitLoop, SchedulePeriodicDecommitTask_Locked) 

```
if (!state_to_detach) {
  return;
}

state_to_detach->Detach()'
```

starboard/common/reuse_allocator_base.h
```diff
@@ -46,12 +46,20 @@ class ReuseAllocatorBase : public Allocator {
   // suspend).
   void DecommitAllDecommitableBlocks();
 
+  // Attempts to decommit a single pending idle block, amortized by `cadence`.
+  // When `cadence == 1`, it immediately decommits a pending block.
+  void TryToDecommitOneBlock(int cadence);
+
  protected:
+  // TODO(b/454441375): Refactor positional boolean parameters into an Options
```
kjyoun, line 54: TODO: b/454441375) -  Refactor...

Right now we are using TODOs in a mixed style (legacy and new).

Wouldn't it be better to use the new style in new code so that we can gradually migrate to it?

https://source.corp.google.com/search?q=TODO&sq=&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain:starboard%2F

starboard/common/bidirectional_fit_reuse_allocator_test.cc
```diff
@@ -147,7 +149,10 @@ class BidirectionalFitReuseAllocatorTest : public ::testing::Test {
     allocator_.reset(new BidirectionalFitReuseAllocator<ReuseAllocatorBase>(
```
kjyoun: Good time to apply make_unque here, though this is not the scope of this PR.

```
allocator_ = std::make_unique<BidirectionalFitReuseAllocator<ReuseAllocatorBase>>
```
https://chromium.googlesource.com/chromium/src/+/lkgr/styleguide/c++/c++-dos-and-donts.md#use-and-instead-of-bare

> Use std::make_unique and base::MakeRefCounted instead of bare new

media/starboard/decoder_buffer_allocator.h
```diff
@@ -89,18 +92,64 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
...
+        task(allocator_);
+      }
+    }
+
+   private:
+    friend class base::RefCountedThreadSafe<PeriodicDecommitState>;
+    ~PeriodicDecommitState() = default;
+
+    base::Lock lock_;
+    DecoderBufferAllocator* allocator_ GUARDED_BY(lock_) = nullptr;
```
kjyoun: raw_ptr<DecoderBufferAllocator> allocator_ ?

> [The Chromium C++ Style Guide](https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++.md#non_owning-pointers-in-class-fields) asks to use raw_ptr<T> for class and struct fields in place of a raw C++ pointer T* whenever possible,

media/starboard/decoder_buffer_allocator.h
```diff
@@ -89,18 +92,64 @@ class DecoderBufferAllocator : public DecoderBuffer::Allocator,
...
+
+    void Detach() {
+      base::AutoLock lock(lock_);
+      allocator_ = nullptr;
+    }
+
+    template <typename Task>
+    void RunIfValid(Task task) {
+      base::AutoLock lock(lock_);
+      if (allocator_) {
```
kjyoun, line 119: nit: early return style?

I find it a bit more readable

```
if (!allcator_) {
  return;
}

task(allocator_); 
```

kjyoun (commented):
Added style comments

otherwise, LGTM

kjyoun (approved):

media/starboard/decoder_buffer_allocator.cc
xiaomings: There is only one line in this if statement, it doesn't benefit from early return, as we are turning a three line block into a 5 line block.

starboard/common/reuse_allocator_base.h
xiaomings: Exactly because we are using a mixed style across the code base, we should avoid using any new styles unless it's auto enforced.  See g/cobalt-dev-internal/c/FU-YFaj3lUo/m/KftBdoDgBQAJ for more details.

starboard/common/bidirectional_fit_reuse_allocator_test.cc
xiaomings: Done

media/starboard/decoder_buffer_allocator.h
xiaomings: Done

media/starboard/decoder_buffer_allocator.h
xiaomings: There is only one line in this if statement, it doesn't benefit from early return, as we are turning a three line block into a 5 line block.

kjyoun (approved):

johnxwork (approved):

media/starboard/decoder_buffer_allocator.cc
xiaomings: Follow up PR in https://github.com/youtube/cobalt/pull/11848.

PR #11595, opened 2026-07-29
cobalt: Convert splash screen switches to features

borongc:
Migrate the splash screen configuration from legacy command-line
switches to the base::Feature system. This change replaces
--disable-splash-screen and --force-video-splash-screen with
standard Chromium feature flags.

This transition enables better alignment with Chromium's feature
infrastructure and allows for more flexible control over splash
screen behavior across different platforms.

Issue: 540461098

PR #11639, opened 2026-07-30
cobalt: Handle early splash screen checks

borongc:
Ensure ShouldCreateSplashScreen behaves correctly during early startup
when the FeatureList is not yet available. This prevents potential
instability or incorrect state detection when determining if the
splash screen should be displayed before the global feature state is
fully initialized.

This is a follow-up PR by https://github.com/youtube/cobalt/pull/11595 to
address the comments in https://github.com/youtube/cobalt/pull/11638#discussion_r3686141344.

Issue: 540461098

### DIFF
--- a/cobalt/shell/common/shell_switches.cc
+++ b/cobalt/shell/common/shell_switches.cc
@@ -61,6 +61,12 @@ const char kSplashScreenShutdownDelayMs[] = "splash-screen-shutdown-delay-ms";
 const char kTestRegisterStandardScheme[] = "test-register-standard-scheme";
 
 bool ShouldCreateSplashScreen() {
+  // If the FeatureList isn't available yet, fall back to the feature's default
+  // state. This may happen during early startup.
+  if (!base::FeatureList::GetInstance()) {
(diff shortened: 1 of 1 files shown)


yjzhang111 (approved):
PR looks good but i had a question - this means we won't have 100% accurate metrics (which might or might not be ok) because some of the devices received disable splash screen feature will still revert back to the default state. How does past Kabuki experiment deal with this? Do we just accept that part of the experiment traffic is polluted?

borongc:
> PR looks good but i had a question - this means we won't have 100% accurate metrics (which might or might not be ok) because some of the devices received disable splash screen feature will still revert back to the default state. How does past Kabuki experiment deal with this? Do we just accept that part of the experiment traffic is polluted?

I think we have to accept this due to using FeatureList. For the previous experiment running with C26 on ATV for the splash screen, since it was a content shell switch, and didn't have this issue.

Chromium has this check everywhere, e.g., https://source.chromium.org/chromium/chromium/src/+/3ae99844378149095e2432255cd92c9cad29111b:url/url_features.cc;l=16, so it makes sense to me to add this check here.

PR #11642, opened 2026-07-30
chore(ci): disable git auto-maintenance to deflake autoroll

oxve:
Disable automatic git maintenance and GC to prevent 'fatal: shallow file has changed since we read it' errors during shallow fetches.

TAG=agy

CONV=6d619d1d-58d0-4583-b4de-2e5d30d9024b

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


oxve:
Dropped from stack as sparse checkouts are no longer used and git auto-maintenance changes are not needed.

PR #11654, opened 2026-07-31
Fix missing topic splash screen on cold start

jonastsai:
Check url when deep_link is empty so the splash topic matches the page being loaded.

Shell::CreateNewWindow previously ignored the startup url parameter when checking for topic deep links.
On platforms where the startup URL is passed as a positional arg, this left splash_topic_ empty and showed the default splash screen instead of music_splash.

Bug: 535206265

### DIFF
--- a/cobalt/shell/browser/shell.cc
+++ b/cobalt/shell/browser/shell.cc
@@ -509,6 +509,8 @@ Shell* Shell::CreateNewWindow(BrowserContext* browser_context,
   GURL link_url;
   if (!deep_link.empty()) {
     link_url = GURL(deep_link);
+  } else if (url.is_valid()) {
+    link_url = url;
   } else if (base::CommandLine::ForCurrentProcess()->HasSwitch(
                  cobalt::switches::kInitialURL)) {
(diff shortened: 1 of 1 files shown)


cobalt/shell/browser/shell.cc
```diff
@@ -509,6 +509,8 @@ Shell* Shell::CreateNewWindow(BrowserContext* browser_context,
   GURL link_url;
   if (!deep_link.empty()) {
     link_url = GURL(deep_link);
+  } else if (url.is_valid()) {
```
borongc, line 512: I think Kabuki only requires to play music splash screen if it's launched by deep link or music icon. What's the behavior of C25?

jonastsai: C25 will show the music icon when launch with following command but C27 cann't

yts launch c9 "https://www.youtube.com/tv?launch=menu&topic=music"

jonastsai: `yts launch xx https://www.youtube.com/tv?launch=menu&topic=music` is the test command described in the certification portal

PR #11655, opened 2026-07-31
rdk: Limit glibc M_ARENA_MAX to 2 in SbRunStarboardMain

sherryzy:
Configure glibc allocator knob M_ARENA_MAX to 2 in SbRunStarboardMain
 for RDK platforms, aligning with Linux x64x11.
    
 Limiting the maximum number of memory arenas reduces heap fragmentation
 and resident memory growth across threads without introducing lock
  contention. This will also reduce 3-4 MB memory usage.

Bug: 540878634

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc
@@ -30,6 +30,7 @@
 // limitations under the License.
 
 #include <gst/gst.h>
+#include <malloc.h>
 #include <signal.h>
 #include <sys/resource.h>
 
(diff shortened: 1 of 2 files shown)


starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc
```diff
@@ -74,6 +75,16 @@ static void UninstallStopSignalHandlers() {
 }  // namespace starboard
 
 int SbRunStarboardMain(int argc, char** argv, SbEventHandleCallback callback) {
+  // Set M_ARENA_MAX to 2 to limit fragmentation without single-arena lock contention.
+  mallopt(M_ARENA_MAX, 2);
```
y4vor, line 79: These are complex parameter changes. Did you see actual memory improvements? 
Unless we have evidence of concrete savings I don't think we should make this change.

sherryzy: According to AlphaEolve, it saves 5-8MB, see b/540878634.

sherryzy: @y4vor It roughly saves ~4-5MB through RDK local testing, see: https://rdklabs.tv/experiments/exp-7575200b-464b-479a-a412-64117c7f26e8/visualization.

y4vor (approved):

PR #11664, opened 2026-08-02
aosp: run StarboardMain on a thread with an 2 MB stack

rogerzanoni:
Cobalt normally runs main() on the process main thread, which has a large stack. The aosp runs main() on a std::thread, which gets a stack with ~1 MB.

This change fixes crashes with SIGSEGV that happened on the StarboardMain thread before the Cobalt binary could be loaded by creating the StarboardMain with pthread_create and giving the thread a 2MB stack to account for InstallationManager's store file.

It also fixes the crash in PosixThreadAttrTest.StackAddrAndSizeAttr, that now fails with EINVAL because of a different issue.

Bug: 532068409

### DIFF
--- a/starboard/aosp/shared/main_activity.cc
+++ b/starboard/aosp/shared/main_activity.cc
@@ -18,7 +18,6 @@
 #include <unistd.h>
 
 #include <string>
-#include <thread>
 #include <vector>
 
 #include "cobalt/aosp/jni_headers/MainActivity_jni.h"
@@ -31,7 +30,12 @@ int main(int argc, char** argv);
 
 namespace {
 
-void StarboardMain() {
+// Cobalt normally runs main() on the process's main thread, which has a large
+// (typically 8 MB) stack. Here it runs on a dedicated thread instead, and
+// bionic's default thread stack is only ~1 MB.
+constexpr size_t kStarboardMainStackSize = 8 * 1024 * 1024;
+
+void* StarboardMain(void* /*context*/) {
   pthread_setname_np(pthread_self(), "StarboardMain");
 
   JNIEnv* env = jni_zero::AttachCurrentThread();
@@ -58,6 +62,7 @@ void StarboardMain() {
   argv.push_back(nullptr);
 
   main(static_cast<int>(args.size()), argv.data());
+  return nullptr;
 }
 
 }  // namespace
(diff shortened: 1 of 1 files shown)


starboard/aosp/shared/main_activity.cc
```diff
@@ -58,14 +62,25 @@ void StarboardMain() {
...
+  pthread_attr_init(&attr);
+  pthread_attr_setstacksize(&attr, kStarboardMainStackSize);
+  pthread_attr_setdetachstate(&attr, PTHREAD_CREATE_DETACHED);
+
+  pthread_t thread;
+  if (pthread_create(&thread, &attr, &StarboardMain, nullptr) != 0) {
+    SB_LOG(ERROR) << "Failed to create StarboardMain thread";
+  }
+
+  pthread_attr_destroy(&attr);
```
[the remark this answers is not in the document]
sacuff: +1

starboard/aosp/shared/main_activity.cc
```diff
@@ -31,7 +30,12 @@ int main(int argc, char** argv);
 
 namespace {
 
-void StarboardMain() {
+// Cobalt normally runs main() on the process's main thread, which has a large
+// (typically 8 MB) stack. Here it runs on a dedicated thread instead, and
+// bionic's default thread stack is only ~1 MB.
+constexpr size_t kStarboardMainStackSize = 8 * 1024 * 1024;
```
andrewsavage1: In C25, we set this to 2Mb instead. Does that still work?

hferreiro: It appears so. The issue seems to be that `InstallationManager` allocates a 1MB buffer at https://github.com/youtube/cobalt/blob/0a648c75e1fc2ad0d63bad82afe7a0e201d719f3/starboard/loader_app/installation_manager.cc#L667.

I tested a 2MB stack successfully. I guess we could switch to 2MB or alternatively do the installation manager allocation on the heap. Note that RDK also explictily sets 2MB: https://github.com/youtube/cobalt/blob/0a648c75e1fc2ad0d63bad82afe7a0e201d719f3/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/run_starboard_main.cc#L79-L82.

andrewsavage1: I think we should use 2mb

starboard/aosp/shared/main_activity.cc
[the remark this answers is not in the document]
hferreiro: Done.

starboard/aosp/shared/main_activity.cc
hferreiro: Done.

andrewsavage1 (approved):

PR #11666, opened 2026-08-02
aosp: avoid /dev/shm for Chromium shared memory

rogerzanoni:
"/dev/shm" does not exist on Android, so shared-memory creation fails with LOG(FATAL) ("incorrect permissions on /dev/shm"), aborting on CrBrowserMain when the cobalt binary is loaded.

Pass --disable-dev-shm-usage so shared-memory temp files use a writable location.

Bug: 532068409


starboard/aosp/shared/main_activity.cc
```diff
@@ -44,10 +45,15 @@ void StarboardMain() {
     if (chdir(files_dir) != 0) {
       SB_LOG(WARNING) << "cobalt_loader: chdir to " << files_dir << " failed";
     }
+    // point TMPDIR at the writable app files directory
+    setenv("TMPDIR", files_dir, 1);
```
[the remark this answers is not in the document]
rogerzanoni: done

starboard/aosp/shared/main_activity.cc
```diff
@@ -44,10 +45,18 @@ void StarboardMain() {
     if (chdir(files_dir) != 0) {
       SB_LOG(WARNING) << "cobalt_loader: chdir to " << files_dir << " failed";
     }
+    // point TMPDIR at the writable app files directory
+    if (setenv("TMPDIR", files_dir, 1) != 0) {
```
sacuff: Can we use `kSbSystemPathTempDirectory`? That should point to cacheDir

sacuff (dismissed):

starboard/aosp/shared/main_activity.cc
hferreiro: This change is not needed since https://github.com/youtube/cobalt/pull/11058. Removed.

hferreiro: Note that the description needs to be updated. I'm not allowed to do that.

starboard/aosp/shared/main_activity.cc
```diff
@@ -48,6 +48,9 @@ void StarboardMain() {
 
   std::vector<std::string> args;
   args.push_back("cobalt_loader");
+  // Don't use "/dev/shm" for shared memory; it does not exist on Android.
+  // With this switch it falls back to GetTempDir().
+  args.push_back("--disable-dev-shm-usage");
```
andrewsavage1, line 60: Please file a ticket to consider doing this for all starboard platforms and assign it to @sacuff

andrewsavage1 (approved):

PR #11670, opened 2026-08-03
cobalt: Reduce V8 initial old space size from 64MB to 16MB

sherryzy:
Through both local testing and AlphaEvolve, further reducing initial-old-space-size from 64MB to 16MB brings ~10MB average memory saving with no observed performance regressions in local testing. Making this change in 3p defaults while waiting for production data validation.

Bug: 540879465

### DIFF
--- a/cobalt/app/cobalt_switch_defaults_starboard.cc
+++ b/cobalt/app/cobalt_switch_defaults_starboard.cc
@@ -126,8 +126,8 @@ CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
        "--no-decommit-pooled-pages "
        // Enable memory saving mode with little v8 performance tradeoff.
        "--optimize-for-size "
-       // Set initial old space size to 64MB and max old space size to 512MB.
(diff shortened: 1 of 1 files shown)


y4vor (approved):

cobalt/app/cobalt_switch_defaults_starboard.cc
```diff
@@ -126,8 +126,8 @@ CommandLinePreprocessor::GetCobaltParamSwitchDefaults() {
        "--no-decommit-pooled-pages "
        // Enable memory saving mode with little v8 performance tradeoff.
        "--optimize-for-size "
-       // Set initial old space size to 64MB and max old space size to 512MB.
-       "--initial-old-space-size=64 "
```
y4vor, line 130: How was 64 picked initially?

sherryzy: In C26, we did experiment for setting it to 256/128/64MB which all do better then defaults. But we don't have enough bandwidth for further granular value.

PR #11674, opened 2026-08-03
feat: Add nativeStability.acknowledgeReports API

hlwarriner:
This web API enables a web app to acknowledge native stability events as already handled so they are no longer returned by getPendingReports(). Acknowledged event UUIDs are persisted to disk so that the behavior survives across Cobalt app lifecycles; storage is managed by the NativeStabilityManager. More context can be found in the "High level" and "Storage 3P MVP" design docs.

To manage the scope of this PR, pruning of the persisted event UUIDs - discussed in the design doc - is left as a quick follow-up.

A stub Mojo binding is provided for the native stability interface for non-Evergreen platforms so that web requests on these platforms will result in rejected promises instead of any actual state changes.

A couple other small improvements are included:
* An existing unit test case is updated to remove noisy logs
* Handling of Mojo disconnection is added to the renderer impl.

Issue: 528362453

### DIFF
--- a/cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
+++ b/cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
@@ -14,13 +14,25 @@
 
 #include "cobalt/browser/h5vcc_native_stability/native_stability_manager.h"
 
+#include <optional>
+#include <string>
+#include <unordered_set>
 #include <utility>
+#include <vector>
 
+#include "base/files/file.h"
+#include "base/files/file_path.h"
+#include "base/files/file_util.h"
+#include "base/files/important_file_writer.h"
 #include "base/functional/bind.h"
+#include "base/json/json_reader.h"
+#include "base/json/json_writer.h"
 #include "base/logging.h"
 #include "base/task/bind_post_task.h"
 #include "base/task/thread_pool.h"
 #include "base/uuid.h"
+#include "base/values.h"
+#include "starboard/configuration_constants.h"
 #include "starboard/extension/crash_handler.h"
 #include "starboard/extension/native_stability.h"
 #include "starboard/system.h"
@@ -28,6 +40,13 @@
 namespace h5vcc_native_stability {
 
 namespace {
+
+// TODO(b/528362453): Per the design doc, we should prune native stability JSON
+// storage at startup by removing entries for events no longer persisted by the
+// crash reporting system's own local storage.
+constexpr char kNativeStabilityDirName[] = "native_stability";
+constexpr char kAckedEventUuidsFileName[] = "acked_event_uuids.json";
+
 mojom::BaseReportDataPtr CreateBaseReportData(
     const SbNativeStabilityReport& sb_report) {
   auto base_data = mojom::BaseReportData::New();
@@ -36,6 +55,79 @@ mojom::BaseReportDataPtr CreateBaseReportData(
   base_data->event_time_sec = sb_report.event_time_s;
   return base_data;
 }
+
+std::unordered_set<std::string> ReadAckedUuidsFromDisk(
+    const base::FilePath& file_path) {
+  std::unordered_set<std::string> acked_uuids;
+  if (file_path.empty() || !base::PathExists(file_path)) {
+    return acked_uuids;
+  }
+
+  std::string file_content;
+  if (!base::ReadFileToString(file_path, &file_content)) {
+    LOG(WARNING) << "Failed to read acked UUIDs file: " << file_path.value();
+    return acked_uuids;
+  }
+
+  std::optional<base::Value::List> parsed_list =
+      base::JSONReader::ReadList(file_content);
+  if (!parsed_list) {
+    LOG(WARNING) << "Failed to parse acked UUIDs JSON list in: "
+                 << file_path.value();
+    return acked_uuids;
+  }
+
+  for (const auto& item : *parsed_list) {
+    if (item.is_string()) {
+      acked_uuids.insert(item.GetString());
+    } else {
+      LOG(WARNING) << "Skipping non-string item in acked UUIDs list: "
+                   << file_path.value();
+    }
+  }
+
+  return acked_uuids;
+}
+
+bool WriteAckedUuidsToDisk(const base::FilePath& file_path,
+                           const std::unordered_set<std::string>& acked_uuids) {
+  if (file_path.empty()) {
+    return false;
+  }
+
+  // We typically expect this directory to already exist but it may not, for
+  // example if this code path has never executed on a particular device or the
+  // system removed the directory.
+  base::FilePath dir_path = file_path.DirName();
+  base::File::Error error = base::File::FILE_OK;
+  if (!base::CreateDirectoryAndGetError(dir_path, &error)) {
+    LOG(ERROR) << "Failed to create " << kNativeStabilityDirName
+               << " directory: " << dir_path.value()
+               << ", error: " << base::File::ErrorToString(error);
+    return false;
+  }
+
+  base::Value::List list_of_uuids;
+  list_of_uuids.reserve(acked_uuids.size());
+  for (const auto& uuid : acked_uuids) {
+    list_of_uuids.Append(uuid);
+  }
+
+  std::optional<std::string> json = base::WriteJson(list_of_uuids);
+  if (!json) {
+    LOG(ERROR) << "Failed to serialize acked UUIDs list to JSON.";
+    return false;
+  }
+
+  if (!base::ImportantFileWriter::WriteFileAtomically(file_path, *json)) {
+    LOG(ERROR) << "Failed to write acked UUIDs atomically to: "
+               << file_path.value();
+    return false;
+  }
+
+  return true;
+}
+
 }  // namespace
 
 // static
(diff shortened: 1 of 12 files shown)


hlwarriner:
CC @AlecsisDuarte @andrewsavage1

third_party/blink/renderer/modules/cobalt/h5vcc_native_stability/h_5_vcc_native_stability.cc
```diff
@@ -98,11 +123,26 @@ void H5vccNativeStability::EnsureReceiverIsBound() {
...
+  remote_native_stability_.reset();
+  HeapHashSet<Member<ScriptPromiseResolverBase>> pending_promises;
+  // Script may execute during a call to Reject(). Swap these sets to prevent
+  // concurrent modification.
+  ongoing_requests_.swap(pending_promises);
+  for (auto& resolver : pending_promises) {
+    resolver->Reject("Mojo connection error.");
+  }
+  ongoing_requests_.clear();
 }
```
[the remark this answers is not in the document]
hlwarriner: done

hlwarriner:
Sending a reminder to please take a look at this, @y4vor @vincentmo @alunev.

cobalt/browser/h5vcc_native_stability/native_stability_manager_unittest.cc
```diff
@@ -255,4 +274,79 @@ TEST_F(NativeStabilityManagerTest, GetPendingReportsIgnoresUnknownReportType) {
   run_loop.Run();
 }
 
+TEST_F(NativeStabilityManagerTest,
+       AcknowledgedReportsFilteredByGetPendingReports) {
```
AlecsisDuarte, line 278: To make these unit tests more robust, should we add the following unit tests:
 - empty UUID strings or duplicates?
 - and a corrupted or non-array `acked_event_uuids.json`?

third_party/blink/renderer/modules/cobalt/h5vcc_native_stability/h_5_vcc_native_stability.cc
```diff
@@ -87,6 +106,12 @@ void H5vccNativeStability::OnGetPendingReports(
   resolver->Resolve(std::move(result));
 }
 
+void H5vccNativeStability::OnAcknowledgeReports(
```
AlecsisDuarte, line 109: Should we guard against race conditions with `OnConnectionError` by verifying `if (ongoing_requests_.Contains(resolver)) { return; }` before `erase`?

cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
```diff
@@ -36,6 +55,79 @@ mojom::BaseReportDataPtr CreateBaseReportData(
...
+  std::optional<base::Value::List> parsed_list =
+      base::JSONReader::ReadList(file_content);
+  if (!parsed_list) {
+    LOG(WARNING) << "Failed to parse acked UUIDs JSON list in: "
+                 << file_path.value();
+    return acked_uuids;
+  }
+
+  for (const auto& item : *parsed_list) {
+    if (item.is_string()) {
```
AlecsisDuarte, line 81: Super-nit: You could easily reduce branching here by wrapping the warning and continuing instead of an if/else branching 

```cc
if (!item.is_string()} {
     LOG(WARNING) << "Skipping non-string item in acked UUIDs list: "
                   << file_path.value();
     continue;
}
acked_uuids.insert(item.GetString());
```

AlecsisDuarte (approved):
LGTM! Left non-blocking comments.

third_party/blink/renderer/modules/cobalt/h5vcc_native_stability/h_5_vcc_native_stability.cc
hlwarriner: Good question, thanks for raising this. It should be ok as is, though, for two reasons:
1. `HeapHashSet` inherits `erase()` from `blink::HashSet::erase()`, which is safe and just "does nothing if the value is not found"
2. The `OnGetPendingReports`, `OnAcknowledgeReports`, and `OnConnectionError` callbacks should all run on the same single-threaded task runner. So they execute sequentially.

cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
hlwarriner: Thanks for the feedback. I do often like using early exits/returns to avoid overly nesting the base case but I think here the if/else works fine: the if and else branches are very short and the if/else clearly expresses the binary choice.

cobalt/browser/h5vcc_native_stability/native_stability_manager_unittest.cc
hlwarriner: Good idea. I'll add these.

For the corrupted ack file, I was trying to keep the unit tests more "black box" and not interact with implementation details like the ack file in the test setup. But I think it's an important enough edge case to test that it's worth compromising on that principle.

AlecsisDuarte (approved):
Great, thanks!

third_party/blink/renderer/modules/cobalt/h5vcc_native_stability/h_5_vcc_native_stability.cc
```diff
@@ -87,6 +106,12 @@ void H5vccNativeStability::OnGetPendingReports(
...
+    ScriptPromiseResolver<IDLUndefined>* resolver) {
+  ongoing_requests_.erase(resolver);
+  resolver->Resolve();
+}
+
 void H5vccNativeStability::EnsureReceiverIsBound() {
   DCHECK(GetExecutionContext());
```
alunev, line 116: I think this is compiled out in release and runtime check is needed like in https://github.com/youtube/cobalt/blob/main/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc#L214

cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
```diff
@@ -143,4 +269,44 @@ void NativeStabilityManager::GetPendingReportsOnTaskRunner(
...
+  for (const auto& uuid : native_stability_event_uuids) {
+    if (!uuid.empty() && acked_uuids.insert(uuid).second) {
+      changed = true;
+    }
+  }
+
+  if (changed) {
+    // It's simpler and safer to just overwrite the file with previously +
+    // currently acknowledged event UUIDs, rather than append to it.
+    WriteAckedUuidsToDisk(file_path, acked_uuids);
```
alunev, line 305: nit: return value is ignored here, let's log it to avoid silent failure?

alunev (approved):

third_party/blink/renderer/modules/cobalt/h5vcc_native_stability/h_5_vcc_native_stability.cc
hlwarriner: Yeah, it is. I think it's a good discussion to have about whether we prefer
1) ` DCHECK(GetExecutionContext())`
2)  `CHECK(GetExecutionContext())`
3) `if (!GetExecutionContext()) return`
here.

But I think we should have that discussion independently from this PR.

Most other Cobalt modules are using  `DCHECK(GetExecutionContext())`, as well. So doing `if (!GetExecutionContext()) return` here would be a departure from that convention.

`H5vccPlatformService::OnDataReceived` is different because it runs asynchronously to handle incoming Mojo events from the browser, at which point it's more likely the page could be gone. (As opposed to running synchronously in response to a web API call from the web client, when the page should be valid.)

cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
hlwarriner: Good point. I changed the return type to `void` and made sure every failure case in `WriteAckedUuidsToDisk()` includes a log.

cobalt/browser/h5vcc_native_stability/native_stability_manager.h
```diff
@@ -57,12 +60,26 @@ class NativeStabilityManager {
...
+  // As with GetPendingReports(), disk I/O is offloaded to a background
+  // ThreadPool task runner and callbacks are posted back to the UI sequence.
+  void AcknowledgeReports(std::vector<std::string> native_stability_event_uuids,
+                          base::OnceClosure callback);
+
   using GetExtensionCallback =
       base::RepeatingCallback<const void*(const char*)>;
 
-  // Inject a custom Starboard extension getter callback for testing.
+  // Injects a custom Starboard extension getter callback for testing.
```
vincentmo, line 84: nit: can we move all testing only methods to the bottom of the method declaration section?

cobalt/browser/h5vcc_native_stability/native_stability_manager.h
```diff
@@ -77,11 +94,18 @@ class NativeStabilityManager {
       base::OnceCallback<void(std::vector<mojom::NativeStabilityReportPtr>)>
           callback);
 
+  void AcknowledgeReportsOnTaskRunner(
```
vincentmo, line 104: When should this be used over AcknowledgeReports? Should one of them be declared private instead?

cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
```diff
@@ -143,4 +268,44 @@ void NativeStabilityManager::GetPendingReportsOnTaskRunner(
...
+      base::BindOnce(&NativeStabilityManager::AcknowledgeReportsOnTaskRunner,
+                     base::Unretained(this),
+                     std::move(native_stability_event_uuids),
+                     base::BindPostTaskToCurrentDefault(std::move(callback))));
+}
+
+void NativeStabilityManager::AcknowledgeReportsOnTaskRunner(
+    std::vector<std::string> native_stability_event_uuids,
+    base::OnceClosure callback) {
+  base::FilePath file_path = GetAckedUuidsFilePath();
```
vincentmo, line 291: I'm not sure if this scenario is valid, but is it possible for there to be multiple instances of cobalt running on a device (for example, YouTube TV and YouTube)? And if so, would they share the same file?
I'm wondering if we need to globally lock the file during the process to prevent concurrent process access, among other things...

vincentmo (approved):
See my question about multi process access, otherwise, lgtm.

y4vor (approved):

cobalt/browser/h5vcc_native_stability/native_stability_manager.cc
y4vor: oh, yeah this is possible. @hlwarriner are you partitioning the file storage?

hlwarriner: This is a good point.

We are not specifically partitioning this file so we'll just get whatever partitioning is provided by `kSbSystemPathCacheDirectory`. From what I can tell Evergreen platforms may, but are not guaranteed, to partition `kSbSystemPathCacheDirectory` for different applications.

Given this I think we can say that:
1. We can guarantee file integrity.
2. But we may have read-modify-write data races.

Since 2 should not occur very often and would not be disastrous if it did, I'll just add a todo comment for now - for us to consider partitioning the file.

For 1, `base::ImportantFileWriter::WriteFileAtomically()` should provide safe, atomic writes to the file. On POSIX platforms the implementation first creates a temporary file *with a unique name* with the new contents, and then replaces the file with the temporary file via `rename()`, which is itself guaranteed to be atomic.

For 2, we update the ack file by reading the existing file, modifying the set of UUIDs in memory by adding new ones from the current request, and then writing the updated set back to the ack file. It should be possible for something like the following to happen:
* Process A reads acked_event_uuids.json (sees ["uuid-1"]).
* Process B reads acked_event_uuids.json (sees ["uuid-1"]).
* Process A appends "uuid-2" and writes ["uuid-1", "uuid-2"] via temp file + rename().
* Process B appends "uuid-3" and writes ["uuid-1", "uuid-3"] via temp file + rename().

Here, Process B would "win" with the last write and "uuid-2" would be lost.

cobalt/browser/h5vcc_native_stability/native_stability_manager.h
hlwarriner: The `*ForTesting` methods already are grouped together at the bottom of the `public` declaration section.

PR #11676, opened 2026-08-03
Add CobaltMemoryMetricsHelper to extract live memory breakdown metrics for non-official builds

haozheng-cobalt:
Introduce CobaltMemoryMetricsHelper to extract live memory breakdown
metrics from the base::StatisticsRecorder. This helper facilitates
retrieving memory usage data from various internal histograms, including
resident set size and private memory footprint.

For specific histograms reported in KiB, the helper converts the values
to bytes to ensure consistent reporting across tools. This
functionality is restricted to non-official builds to support debugging
and internal telemetry analysis.

Bug: 530302534

### DIFF
--- a/cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
+++ b/cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
@@ -16,21 +16,29 @@
 
 #include <iterator>
 #include <memory>
+#include <optional>
+#include <string_view>
+#include <vector>
 
 #include "base/metrics/histogram_base.h"
 #include "base/metrics/histogram_samples.h"
 #include "base/metrics/statistics_recorder.h"
+#include "base/strings/string_util.h"
 
 namespace cobalt {
 
 namespace {
 
-// UMA Memory.Experimental.Browser2.* histograms record values in KiB
-// (Kilobytes). Convert KiB to bytes for consistent DevTools byte reporting.
-constexpr uint64_t kBytesPerKiB = 1024;
+// All Cobalt memory breakdown histograms (emitted in
+// cobalt_memory_metrics_emitter.cc and peak_gpu_memory_callback.cc via
+// base::UmaHistogramMemoryLargeMB) store values in MiB (Mebibytes). We convert
+// MiB to bytes (1 MiB = 1,048,576 bytes) for consistent DevTools byte
+// reporting.
+constexpr uint64_t kBytesPerMiB = 1024 * 1024;
 
-bool IsKiBHistogram(const std::string& metric_name) {
-  return metric_name.rfind("Memory.Experimental.Browser2.", 0) == 0 ||
+bool IsMiBHistogram(std::string_view metric_name) {
+  return base::StartsWith(metric_name, "Memory.Experimental.Browser2.",
+                          base::CompareCase::SENSITIVE) ||
          metric_name == "Memory.Browser.ResidentSet" ||
          metric_name == "Memory.Browser.PrivateMemoryFootprint" ||
          metric_name == "Memory.Browser.LibChrobaltRss" ||
@@ -39,13 +47,7 @@ bool IsKiBHistogram(const std::string& metric_name) {
 
 }  // namespace
 
-constexpr const char* CobaltMemoryMetricsHelper::kMemoryBreakdownMetricNames[];
-
-std::optional<uint64_t> CobaltMemoryMetricsHelper::GetMetricValueBytes(
-    const std::string& metric_name) {
-#if defined(OFFICIAL_BUILD)
-  return std::nullopt;
-#else
+std::optional<uint64_t> GetMetricValueBytes(std::string_view metric_name) {
   base::HistogramBase* histogram =
       base::StatisticsRecorder::FindHistogram(metric_name);
   if (!histogram) {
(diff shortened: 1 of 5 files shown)


cobalt/browser/metrics/cobalt_memory_metrics_helper.h
```diff
@@ -0,0 +1,70 @@
...
+// from base::StatisticsRecorder. Aligns with go/kimono-memory-metrics.
+//
+// Lifetime and Ownership:
+// This class consists entirely of static utility methods and is not intended
+// to be instantiated.
+//
+// Threading Model:
+// The static methods are thread-safe and can be called from any thread or
+// TaskRunner, as they rely on the thread-safe base::StatisticsRecorder.
+class CobaltMemoryMetricsHelper {
```
johnxwork: Where do you want to use this class? It's best to look at the usage and the definition together.

Awallky: Add `CobaltMemoryMetricsHelper() = delete;` to clearly mark the static utility class as non-instantiable according to the Google C++ Style Guide.

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
```diff
@@ -0,0 +1,90 @@
...
+#include "base/metrics/histogram_base.h"
+#include "base/metrics/histogram_samples.h"
+#include "base/metrics/statistics_recorder.h"
+#include "base/strings/string_util.h"
+
+namespace cobalt {
+
+namespace {
+
+// UMA Memory.Experimental.Browser2.* histograms record values in KiB
```
Awallky: This is inaccurate. They are measured in MiB.
See tools/metrics/histograms/metadata/cobalt/histograms.xml for their units.

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  int64_t sum = samples->sum();
+  if (sum < 0) {
+    return std::nullopt;
+  }
+
+  // Note: This computes the historical average over the lifetime of the
+  // process, not the instantaneous current memory usage.
+  uint64_t raw_average = static_cast<uint64_t>(sum / samples->TotalCount());
```
Awallky: I think median would be a better way to observe these metrics, even if locally. p50 (median) is what is the standard way to observe these in the field. Staying consistent with that is the goal.

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
```diff
@@ -0,0 +1,90 @@
...
+  if (IsKiBHistogram(metric_name)) {
+    return raw_average * kBytesPerKiB;
+  }
+
+  return raw_average;
+}
+
+std::vector<MemoryBreakdownMetric>
```
Awallky: It might be better to return an enum of statuses. One for all breakdown metrics included, one for partial, and one for none.

This could a be an absl::status or something like that.

Awallky: Ah, but StatusOr is banned in the style guide. https://chromium.googlesource.com/chromium/src/+/HEAD/styleguide/c++/c++-features.md#statusor-banned

Perhaps `std::optional<T>` is a better choice.

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
haozheng-cobalt: Thanks for pointing this out, updated the calculation to mirror the metrics in the field.

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
haozheng-cobalt: Thanks for catching this! this is very important.

cobalt/browser/metrics/cobalt_memory_metrics_helper.h
haozheng-cobalt: I've uploaded the follow-up PR regarding how this class should be utilized to fetch metrics in CDP: https://github.com/youtube/cobalt/pull/11679.

cobalt/browser/metrics/cobalt_memory_metrics_helper_unittest.cc
```diff
@@ -0,0 +1,74 @@
...
+  ASSERT_TRUE(breakdown.has_value());
+
+  bool found_resident_set = false;
+  for (const auto& entry : breakdown.value()) {
+    if (std::string(entry.name) == metric_name) {
+      found_resident_set = true;
+      EXPECT_GE(entry.value_bytes, 32u * 1024u * 1024u);
+    }
+  }
+  EXPECT_TRUE(found_resident_set);
```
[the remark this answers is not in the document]
haozheng-cobalt: 13 memory breakdown metrics are stored in MiB in base::StatisticsRecorder

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
```diff
@@ -0,0 +1,109 @@
...
+      p50_value = static_cast<uint64_t>((min + max) / 2);
+      break;
+    }
+  }
+
+  if (IsMiBHistogram(metric_name)) {
+    return p50_value * kBytesPerMiB;
+  }
+
+  return p50_value;
```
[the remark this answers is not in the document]
haozheng-cobalt: 13 memory breakdown metrics are stored in MiB in base::StatisticsRecorder

cobalt/browser/metrics/cobalt_memory_metrics_helper.cc
```diff
@@ -0,0 +1,109 @@
...
+constexpr uint64_t kBytesPerMiB = 1024 * 1024;
+
+bool IsMiBHistogram(std::string_view metric_name) {
+  return base::StartsWith(metric_name, "Memory.Experimental.Browser2.",
+                          base::CompareCase::SENSITIVE) ||
+         metric_name == "Memory.Browser.ResidentSet" ||
+         metric_name == "Memory.Browser.PrivateMemoryFootprint" ||
+         metric_name == "Memory.Browser.LibChrobaltRss" ||
+         metric_name == "Memory.GPU.PeakMemoryUsage2.PageLoad";
+}
```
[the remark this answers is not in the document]
haozheng-cobalt: 13 memory breakdown metrics are stored in MiB in base::StatisticsRecorder

cobalt/browser/metrics/cobalt_memory_metrics_helper.h
```diff
@@ -0,0 +1,76 @@
...
+// inspection.
+//
+// Lifetime and Ownership:
+// This class consists entirely of static utility methods and is
+// non-instantiable.
+//
+// Threading Model:
+// The static methods are thread-safe and can be called from any thread or
+// TaskRunner, as they rely on the thread-safe base::StatisticsRecorder.
+class CobaltMemoryMetricsHelper {
```
Awallky: If your helper functions need to be exposed across multiple files, group them inside a named namespace instead of a class.

go/cstyle#Nonmember,_Static_Member,_and_Global_Functions

cobalt/browser/metrics/cobalt_memory_metrics_helper_unittest.cc
```diff
@@ -0,0 +1,71 @@
...
+#include "base/metrics/histogram_functions.h"
+#include "base/metrics/statistics_recorder.h"
+#include "testing/gtest/include/gtest/gtest.h"
+
+namespace cobalt {
+
+class CobaltMemoryMetricsHelperTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    // No special initialization required for StatisticsRecorder.
```
Awallky: Because tests are run in parallel, reading from the global recorder may randomly include records emitted by other tests, resulting in flakes. The test needs to inject a temporary StatisticsRecorder in SetUp().

`base::StatisticsRecorder::CreateTemporaryForTesting()` might do the trick.

Awallky (approved):

cobalt/browser/metrics/cobalt_memory_metrics_helper.h
haozheng-cobalt: good catch! shouldn't need to create a class for all static functions.

cobalt/browser/metrics/cobalt_memory_metrics_helper_unittest.cc
haozheng-cobalt: thanks for the suggestion.

PR #11679, opened 2026-08-03
Append Cobalt memory breakdown metrics to CDP Performance.getMetrics for non-official builds

haozheng-cobalt:
Expose Cobalt-specific memory breakdown statistics through the
Chrome DevTools Protocol Performance.getMetrics method. This
functionality is limited to non-official builds.

Including these metrics provides developers with better visibility
into subsystem-level memory usage, facilitating easier debugging
and optimization in memory-constrained environments.

Bug: 530302534


Awallky:
Is a script using CDP the end-goal? What's the user journey look like? I think main app devs would be much more likely to us this if their user journey remained inside of the devtools UI.

haozheng-cobalt:
> Is a script using CDP the end-goal? What's the user journey look like? I think main app devs would be much more likely to us this if their user journey remained inside of the devtools UI.

The end goal is to add a memory breakdown tab/section in the devtools, this PR is part of the BE implementation in CDP that will ultimately enable the memory breakdown in the devtools. Plz see more details in the design doc: [go/chrobalt-devtools-memory-breakdown](https://goto.google.com/chrobalt-devtools-memory-breakdown)

third_party/blink/renderer/core/inspector/inspector_performance_agent.cc
```diff
@@ -22,6 +22,10 @@
 #include "third_party/blink/renderer/platform/instrumentation/instance_counters.h"
 #include "third_party/blink/renderer/platform/scheduler/public/thread.h"
 
+#if BUILDFLAG(IS_COBALT) && !defined(OFFICIAL_BUILD)
+#include "cobalt/browser/metrics/cobalt_memory_metrics_helper.h"
```
Awallky: This is a layering violation. Nothing outside of //cobalt should ever depend on it. If you are going to make this helper, it should live inside of //third_party/blink.

third_party/blink/renderer/core/BUILD.gn
```diff
@@ -466,6 +466,12 @@ component("core") {
   if (is_cobalt && use_starboard_media) {
     deps += [ "//starboard:starboard_group" ]
   }
+
+  if (is_cobalt && !is_official_build) {
+    # Depend on memory_helper (depends strictly on //base) to avoid layering
+    # violations with full browser process targets.
+    deps += [ "//cobalt/browser:memory_helper" ]
```
Awallky: Again, fix this layerig violation

cobalt/browser/BUILD.gn
```diff
@@ -197,6 +197,19 @@ if (is_starboard || is_androidtv) {
   }
 }
 
+if (!is_official_build) {
+  # Isolated memory metrics helper target depending strictly on //base.
```
haozheng-cobalt: Already created this separate target that depends on //base to resolve the layer voilation.

Awallky (approved):
LGTM % comments

cobalt/browser/BUILD.gn
Awallky: In this PR, we make //third_party/blink depend on //cobalt, which is a layering inversion. Normally the dependency chain is //cobalt -> //third_party/blink -> ... -> //base. However, we are now introducing this chain:  //third_party/blink -> //cobalt -> ... -> //base. We want to avoid this.

It is okay to add files directly here and make them build only in cobalt builds..

third_party/blink/renderer/core/inspector/inspector_performance_agent.cc
hlwarriner: I'm not sure if this is quite the same point as @Awallky is making, but the renderer should not interact directly with the browser but should instead go through Mojo. So I agree `cobalt::GetMemoryBreakdown()` should not be directly called from here.

hlwarriner (dismissed):

Awallky (approved):
LGTM
Can you make sure our blink_unittests run the new test? Not sure if we disabled all of blink unittests.

third_party/blink/renderer/core/inspector/inspector_performance_agent.cc
```diff
@@ -270,6 +274,17 @@ protocol::Response InspectorPerformanceAgent::getMetrics(
                      .InSecondsF());
   }
 
+#if BUILDFLAG(IS_COBALT) && !defined(OFFICIAL_BUILD)
+  if (auto memory_breakdown = GetMemoryBreakdown()) {
```
hlwarriner, line 278: nit: I think `auto` reduces clarity here as it's not evident from the local context what type is returned by `GetMemoryBreakdown()`. Unless there's a good reason to keep `auto` (see totw 232), please just write out the explicit type. Feel free to make that change in a quick follow-up PR if you prefer.

`auto` does seem like a good choice in the for loop, below.

hlwarriner (approved):

PR #11683, opened 2026-08-03
Add a switch to set max http cache size

yuying-y:
Create --max-http-cache-size to set the max http cache size, to prepare an experiment for memory reduction.

TAG=agy

CONV=66190fb0-fe31-4a73-bc05-eae0fce17efd

Issue: 542314581


cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -373,6 +373,19 @@ void CobaltContentBrowserClient::ConfigureNetworkContextParams(
...
+            "max-http-cache-size")) {
+      std::string size_str =
+          base::CommandLine::ForCurrentProcess()->GetSwitchValueASCII(
+              "max-http-cache-size");
+      int parsed_size = 0;
+      if (base::StringToInt(size_str, &parsed_size)) {
+        http_cache_max_size = parsed_size;
+      }
+    }
+    network_context_params->http_cache_max_size = http_cache_max_size;
```
johnxwork: Did you verify this parameter affects HTTP cache size locally? You can set this to a very small limit and dump the cache content of HTTP Cache at runtime locally to verify it takes effect.

yuying-y:
All feedback requested in the code review has been strictly resolved. I have added robust null/empty condition checks when parsing intent extras in `JavaSwitches.java` to prevent NPEs, and securely guarded the C++ override block with `HasSwitch` validation to prevent silent flattening of the auto-calculated network baseline!

cobalt/browser/cobalt_content_browser_client.cc
yuying-y: Verified. Without the setting, cache is ~16.5MB; with it set to 1KB, the size is 20.0KB (which represents strictly the empty Chromium Cache_Data internal index skeletons).

ckim1346 (approved):
LGTM -- just a note that setting `network_context_params->http_cache_max_size = 0` (in the 0 case) doesn't actually set it to 0, I believe it then defaults to [disk_cache::PreferredCacheSize](https://source.corp.google.com/h/github/youtube/cobalt/+/main:net/disk_cache/cache_util.cc;drc=524126904340bf2104b6615d6bb6352ba6d7d844;l=187) which is the dynamic allocation

johnxwork (approved):

PR #11696, opened 2026-08-04
starboard: Add option to force dual threads

xiaomings:
Add Media.ForceDualThreads experimental feature to allow forcing dual threads in player components factory even when not using libopus decoder.

Bug: 329686979

### DIFF
--- a/starboard/android/shared/player_components_factory.cc
+++ b/starboard/android/shared/player_components_factory.cc
@@ -97,6 +97,31 @@ bool UseLibopusDecoder(SbMediaAudioCodec codec,
          !force_platform_opus_decoder;
 }
 
+bool ShouldUseDualThreads(SbMediaAudioCodec audio_codec,
+                          SbDrmSystem drm_system,
+                          const ExperimentalFeatures& features,
+                          bool force_platform_opus_decoder) {
+  // If there is no audio codec, default to using dual threads.
+  if (audio_codec == kSbMediaAudioCodecNone) {
+    return true;
+  }
+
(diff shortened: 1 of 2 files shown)


starboard/android/shared/player_components_factory.cc
kjyoun: how about using helper method?

I think it makes conditions more readable.

```
bool ShouldUseDualThreads(const CreationParameters& params, 
                          const ExperimentalFeatures& features) const {
  // If there is no audio codec, default to using dual threads.
  if (params.audio_codec() == kSbMediaAudioCodecNone) {
    return true;
  }

  // The experimental feature flag overrides all other conditions.
  if (features.GetBool(kMediaForceDualThreads)) {
    return true;
  }

  // `use_dual_threads` should be disabled if the libopus audio
  // decoder isn't used, as we want to limit the initial behavior to
  // playbacks with software based audio where their threading behavior is
  // more straightforward. 
  //
  // TODO(b/329686979): Make this work better with AdaptiveAudioDecoder,
  // where technically the stream can start with aac then transit into opus.
  return UseLibopusDecoder(params.audio_codec(), 
                           params.drm_system(), 
                           force_platform_opus_decoder_);
}

bool use_dual_threads = ShouldUseDualThreads(creation_parameters, experimental_features);

```

kjyoun (approved):

starboard/android/shared/player_components_factory.cc
xiaomings: Done

PR #11709, opened 2026-08-05
media: Fix misleading log level in UpdateStarboardRenderingMode

kjyoun:
Downgrade logs in UpdateStarboardRenderingMode() from WARNING to INFO.
 
During normal playback startup, `UpdateStarboardRenderingMode()` is called
before `StartPlayingFrom()`, so logging at the WARNING level caused false
alarms during routine initialization.

Issue: 542276633


media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -244,20 +244,18 @@ void StarboardRendererClient::UpdateStarboardRenderingMode(
       // The video frame is handled by Sbplayer, and render to its
       // surface directly.
       if (is_playing_) {
+        LOG(INFO) << "Stopping active video rendering sink on mode switch to "
```
borongc: nit: `Stopping RenderCallback::Render() as on kPunchOut.`

media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -244,20 +244,18 @@ void StarboardRendererClient::UpdateStarboardRenderingMode(
...
-      } else {
-        LOG(WARNING) << "StarboardRendererClient doesn't stop video rendering "
-                        "sink, since the video is not playing.";
       }
       break;
     case StarboardRenderingMode::kDecodeToTexture:
       // StarboardRenderingMode::kDecodeToTexture needs to update
       // video frame via VideoRendererSink::RenderCallback::Render().
       if (is_playing_) {
+        LOG(INFO) << "Starting active video rendering sink on mode switch to "
```
borongc: nit: `Starting RenderCallback::Render() as on kDecodeToTexture.`

borongc (approved):

media/mojo/clients/starboard/starboard_renderer_client.cc
kjyoun: Done

media/mojo/clients/starboard/starboard_renderer_client.cc
kjyoun: Thanks for the suggestion!

Done

media/mojo/clients/starboard/starboard_renderer_client.cc
```diff
@@ -244,20 +244,17 @@ void StarboardRendererClient::UpdateStarboardRenderingMode(
...
     case StarboardRenderingMode::kDecodeToTexture:
       // StarboardRenderingMode::kDecodeToTexture needs to update
       // video frame via VideoRendererSink::RenderCallback::Render().
       if (is_playing_) {
+        LOG(INFO)
+            << "Starting RenderCallback::Render() as on kDecodeToTexture.";
         StartVideoRendererSink();
-      } else {
-        LOG(WARNING) << "StarboardRendererClient doesn't start video rendering "
-                        "sink, since StartPlayingFrom() is not called.";
```
xiaomings, line 265: Can we keep this (and the one above) as INFO?

kjyoun: I prefer removing this log. 

In standard/most common cases, `UpdateStarboardRenderingMode()` is called when `is_playing_` is false. Even as INFO, a message stating that the sink wasn't started/stopped can be misleading and imply an unexpected state or skipped operation during normal initialization.

xiaomings: I agree that to mark this log as a WARNING is misleading as it's a rare but legitimate state, but to have a log message suggesting this rare case is taken isn't misleading.

It's ok if we decide to remove this message if we don't need it.  However, we should avoid removing it just because some Gen AI tools may incorrectly take this INFO level message as a warning.

kjyoun: I remove it since it is not necessary.

Here is the jist of this change

```
BEFORE

    if (is_playing_) { // Mid-playback punchout <-> decode-to-texture switch (rare)
      // Logged nothing
      StopOrStartVideoRendererSink();
    } else { // Normal initialization path (common)
      // Logged a WARNING for being on the normal path!
      LOG(WARNING) << "Sink not started/stopped because video is not playing.";
    }

AFTER:
    if (is_playing_) { // Mid-playback punchout <-> decode-to-texture switch (rare)
      LOG(INFO) << "Starting/Stopping video sink on mode transition...";
      StopOrStartVideoRendererSink();
    } else { // Normal initialization path (common)
      // Log nothing. Logging "sink not started" during normal setup sounds
      // like an unexpected state.
    }
```

WDYT?

xiaomings: I am not sure, and will defer to @borongc to decide.

I think there are two decisions here:
1. Whether we should replace the WARNING with INFO.
2. Whether we should refactor the log messages.

I agree with item 1, but don't have enough information on 2.  I would say to massively replace WARNING to LOG is much easier than deciding whether a particular message should be changed case by case.

It's also worth noting that we don't avoid logging in the normal path as long as they are not spamming.

kjyoun: @borongc 

Can you share your opinion on this?

I am OK with adding log on normal path, though  I think that, if we decided to add log on normal path, we cah find better message

borongc: This log is only one-time per StarboardRenderer lifetime.

As a result, if WARNING is confusing, we can change it to INFO, and only log it on rare path, as else case is the common path.

On the other hand, StarboardRenderer doesn't support mode-switching (punchout <--> DTT), but Chromium does. I put the log here is that if we want to support it in the future, we will need to handle it here.

kjyoun: chatted with @xiaomings 

Agreed to make this PR do simple replacement

PR #11710, opened 2026-08-05
media: Remove unknown EOTF warning in ParseVideoInfo

kjyoun:
Remove the warning log when encountering an unknown EOTF during MIME
type parsing. Web apps frequently probe unknown or dummy EOTF values,
such as catavision, via isTypeSupported() to verify MIME parameter
validation. Returning false without a warning log is the expected
behavior for this code path.

Issue: 536977265
Issue: 542276633

### DIFF
--- a/starboard/shared/starboard/media/parsed_mime_info.cc
+++ b/starboard/shared/starboard/media/parsed_mime_info.cc
@@ -164,8 +164,9 @@ bool ParseVideoInfo(const MimeType& mime_type,
   if (!eotf.empty()) {
     SbMediaTransferId transfer_id_from_eotf = GetTransferIdFromString(eotf);
     if (transfer_id_from_eotf == kSbMediaTransferIdUnknown) {
-      // The eotf is an unknown value, mark the codec info as invalid.
-      SB_LOG(WARNING) << "Unknown eotf " << eotf << ".";
+      // Web apps probe unknown or dummy EOTF values (e.g., "catavision")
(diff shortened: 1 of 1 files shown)


starboard/shared/starboard/media/parsed_mime_info.cc
```diff
@@ -164,8 +164,9 @@ bool ParseVideoInfo(const MimeType& mime_type,
   if (!eotf.empty()) {
     SbMediaTransferId transfer_id_from_eotf = GetTransferIdFromString(eotf);
     if (transfer_id_from_eotf == kSbMediaTransferIdUnknown) {
-      // The eotf is an unknown value, mark the codec info as invalid.
-      SB_LOG(WARNING) << "Unknown eotf " << eotf << ".";
+      // Web apps probe unknown or dummy EOTF values (e.g., "catavision")
+      // via isTypeSupported() to verify MIME parameter validation.
+      // This is a normal code path.
```
xiaomings, line 169: Can we keep the log message as INFO?

kjyoun: I prefer removing this log to avoid log spam. `isTypeSupported()` and `ParseVideoInfo()` routinely return false during standard MIME parameter probing (e.g., feature detection by web apps). 

Since we don't emit INFO logs for other expected false returns in `ParseVideoInfo()`, logging here breaks that convention.

xiaomings (approved):

PR #11711, opened 2026-08-05
Starboard/RDK: Fix video seek on long GOPs

denisyuji:
Currently, the buffering probe compares every buffer against the seek position plus a fixed margin. After a seek the application starts writing at the preceding keyframe and, while the pipeline is `PAUSED`, only a limited amount of data flows past the appsrc. In this scenario, the seek target (absolute seek point) is never reached on long GOPs (GOPs with more frames than the pipeline can enroll). Then, the pipeline stays in `PAUSED` and the condition to move to `PLAYING` is never met.
    
  Pass 0 as the target so it becomes 100 ms for audio and 160 ms for video on the media timeline. The first buffer written after a seek already exceeds it, so the probe fires immediately and the seek position check in `ChangePipelineState` becomes the only gate before the pipeline changes to `PLAYING`.
  
  Also re-arm the playing state update from `WriteSample`, since the first attempt to resume after a seek is rejected and nothing retried it.

Bug: 534354066

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
@@ -2182,6 +2182,15 @@ void PlayerImpl::WriteSample(SbMediaType sample_type,
 
   RecordTimestamp(sample_type, timestamp);
 
+  // ChangePipelineState refuses to resume until the written samples cover the
+  // seek position, and the first attempt after a seek is normally rejected
+  // because the app writes its way there from the preceding keyframe. Re-arm
+  // the update when the samples that unblock it arrive.
+  if (min_sample_timestamp_ >= seek_position_ && (rate_ || pending_rate_) &&
+      GST_STATE(pipeline_) != GST_STATE_PLAYING) {
+    SchedulePlayingStateUpdate();
+  }
+
   if (sample_infos[0].drm_info) {
     GST_LOG_OBJECT(pipeline_, "Encounterd encrypted %s sample",
             sample_type == kSbMediaTypeVideo ? "video" : "audio");
@@ -2413,7 +2422,9 @@ void PlayerImpl::HandleInititialSeek() {
     else {
       DispatchOnWorkerThread(new PlayerStatusTask(player_status_func_, player_, ticket_,
                                                   context_, kSbPlayerStatePrerolling));
-      AddBufferingProbe(position, ticket_);
+      // Count the buffering target from the start of the stream, not from the
+      // seek position, so that it stays reachable while the pipeline is paused.
+      AddBufferingProbe(0, ticket_);
       GST_INFO_OBJECT(pipeline_, "Successfully changed initial segment, position: %" GST_TIME_FORMAT ", ticket: %d", GST_TIME_ARGS(position), ticket_);
       return;
     }
@@ -2495,7 +2506,9 @@ void PlayerImpl::Seek(int64_t seek_to_timestamp, int ticket) {
     DispatchOnWorkerThread(new FunctionTask([this]() {
       state_ = State::kPrerollAfterSeek;
     }, "Preroll after seek"));
-    AddBufferingProbe(seek_to_time_ns, ticket);
+    // Count the buffering target from the start of the stream, not from the
+    // seek position, so that it stays reachable while the pipeline is paused.
+    AddBufferingProbe(0, ticket);
   }
 }
 


content/browser/speech/tts_starboard.cc
```diff
@@ -151,8 +151,10 @@ void TtsPlatformImplStarboard::GetVoices(std::vector<VoiceData>* out_voices) {
   voice.native = true;
   voice.name = "Starboard";
   // Starboard API dictates language should be the same as
-  // SbSystemGetLocaleId().
-  voice.lang = SbSystemGetLocaleId();
+  // SbSystemGetLocaleId(), which may return NULL when no locale is configured
+  // (e.g. no LANG/LC_* in the environment of a container).
+  const char* locale_id = SbSystemGetLocaleId();
+  voice.lang = locale_id ? locale_id : "en-US";
```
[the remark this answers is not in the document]
denisyuji: patch removed

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -2182,6 +2182,15 @@ void PlayerImpl::WriteSample(SbMediaType sample_type,
...
   RecordTimestamp(sample_type, timestamp);
 
+  // ChangePipelineState refuses to resume until the written samples cover the
+  // seek position, and the first attempt after a seek is normally rejected
+  // because the app writes its way there from the preceding keyframe. Re-arm
+  // the update when the samples that unblock it arrive.
+  if (min_sample_timestamp_ >= seek_position_ && (rate_ || pending_rate_) &&
+      GST_STATE(pipeline_) != GST_STATE_PLAYING) {
+    SchedulePlayingStateUpdate();
+  }
```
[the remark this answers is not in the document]
denisyuji: SchedulePlayingStateUpdate() already guards against this by returning early when `playing_state_update_source_id_ != 0`, so at most one update is pending at any time. The extra calls per sample are just a member read and an early return, no task is queued.

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -2495,7 +2506,9 @@ void PlayerImpl::Seek(int64_t seek_to_timestamp, int ticket) {
     DispatchOnWorkerThread(new FunctionTask([this]() {
       state_ = State::kPrerollAfterSeek;
     }, "Preroll after seek"));
-    AddBufferingProbe(seek_to_time_ns, ticket);
+    // Count the buffering target from the start of the stream, not from the
+    // seek position, so that it stays reachable while the pipeline is paused.
+    AddBufferingProbe(0, ticket);
```
santoshmahto7, line 2511: Buffering probes are meant to ensure that enough buffer is accumulated in appsrc to resume playback.  

 In case of a seek, we need to have a stored buffer with the target seek time (+ extra buffer) in appsrc.This will emit the event "kDidReachBufferingTarget," which will try to set the pipeline into the playing state.  

If we set the buffering probe with 0 target time, then  buffering will always be successful and emit the  "kDidReachBufferingTarget" event even though we haven't buffered enough data after(or at that point) the seek time

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -2413,7 +2422,9 @@ void PlayerImpl::HandleInititialSeek() {
     else {
       DispatchOnWorkerThread(new PlayerStatusTask(player_status_func_, player_, ticket_,
                                                   context_, kSbPlayerStatePrerolling));
-      AddBufferingProbe(position, ticket_);
+      // Count the buffering target from the start of the stream, not from the
+      // seek position, so that it stays reachable while the pipeline is paused.
+      AddBufferingProbe(0, ticket_);
```
santoshmahto7, line 2427: see below comments

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
denisyuji: Yes, this is hacky, but it is the way that touch less the code. The target seek check is done in two gates in the code:

1 - [player_internal.cc#L3021](https://github.com/youtube/cobalt/blob/3226be501c2a79757fe82845455a3e9d18b69a1e/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc#L3021) - first, in the code flow

```
    if ( (pts + duration) >= data->target_time ) {

      (...)

      GstStructure* structure = gst_structure_new(kDidReachBufferingTargetMsgName, "ticket", G_TYPE_INT, data->ticket, nullptr);
      gst_element_post_message(GST_ELEMENT(parent), gst_message_new_application(parent, structure));
      gst_object_unref(parent);

      return GST_PAD_PROBE_REMOVE;
    }
```

2 - [player_internal.cc#L2290](https://github.com/youtube/cobalt/blob/3226be501c2a79757fe82845455a3e9d18b69a1e/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc#L2290) - second

```
  if (GST_CLOCK_TIME_IS_VALID(seek_pos_ns) && seek_pos_ns > GST_BUFFER_TIMESTAMP(buffer)) {
    // Set dummy duration to let sink drop out-of-segment samples
    GST_BUFFER_DURATION (buffer) = GST_SECOND / 60;
    GST_BUFFER_FLAG_SET (buffer, GST_BUFFER_FLAG_DECODE_ONLY);
  }
```

(2) flags the buffer as `GST_BUFFER_FLAG_DECODE_ONLY`. However, (2) is never reached before (1) be satisfied and emit a `kDidReachBufferingTargetMsgName` message. The `GST_BUFFER_FLAG_DECODE_ONLY` signal indicates that the buffer will be decoded, and it will be discarded by the sink (even in PAUSED mode) until it be completely filled with buffers not tagged with `GST_BUFFER_FLAG_DECODE_ONLY`. This PR proposes to bypass (1) intentionally to fill the pipeline until it reaches the buffer, and keep (2) as the only gate that checks if the frame is at the seek point. Otherwise, a more comprehensive refactoring should be made.

PR #4378, opened 2024-11-05
[media] Pass full mime type to SbPlayer

osagie98:
Allows the mime type passed to `MediaSource.addSourceBuffer()` to be passed to the SbPlayer.

Contains changes to //media/filters/chunk_demuxer.* and //media/base/demuxer_stream.h originally from https://lbshell-internal-review.git.corp.google.com/c/cobalt_src/+/219082.

b/321842876

PR #11732, opened 2026-08-06
media: Move mime_type from ChunkDemuxerStream to config

johnedocampo:
Refactor Starboard media MIME type storage from stream-level properties
(ChunkDemuxerStream / DemuxerStream) to configuration-level properties
on AudioDecoderConfig and VideoDecoderConfig.

This change is made in response to how MediaSource.changeType() operates.
During changeType(), format transitions (e.g., switching from H.264 "video/mp4"
to VP9 "video/webm") generate new AudioDecoderConfig and VideoDecoderConfig
instances for each stream segment. Storing mime_type on DemuxerStream caused
stream-level state to desynchronize from sample buffers and decoder configurations
during codec transitions. Moving mime_type onto AudioDecoderConfig and
VideoDecoderConfig binds container MIME strings directly to elementary codec
parameters, ensuring container details remain strictly synchronized with decoder
configurations across Mojo IPC and renderer bridges.

Key changes:
- AudioDecoderConfig & VideoDecoderConfig: Added `mime_type` fields, getters, setters,
  and updated Matches() and AsHumanReadableString() to use `mime_type` as well.
- Mojo IPC & Mojom: Added `mime_type` to Mojom decoder configs and simplified
  demuxer_stream.mojom by removing the redundant 5th `mime_type` parameter.
- ChunkDemuxer & SourceBufferState: Updated SourceBufferState to stamp `mime_type`
  onto configs in OnNewConfigs(), and hardened ChunkDemuxer map lookups.
- Starboard Renderer & SbPlayerBridge: Updated StarboardRenderer and SbPlayerBridge
  to read `mime_type` from configs and removed obsolete constructor string parameters.
- DemuxerStream & IPC Adapters: Removed legacy `mime_type()` stream getters, dead IPC
  fields, and obsolete GetMimeType() helpers from Mojo bypass bridges.

Bug: 543919264

### DIFF
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -84,16 +84,6 @@ bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
   //    boundaries.
   // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
   //    container MIME strings differ while elementary video/audio parameters
-  //    remain identical.  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
-  // is intentionally excluded from Matches():
-  // 1. Matches() checks whether underlying hardware decoders require 
-  //    re-initialization based on elementary stream properties (codec,
-  //    profile, resolution, extra_data).
-  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
-  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
-  //    boundaries.
-  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
-  //    container MIME strings differ while elementary video/audio parameters
   //    remain identical.
   return (
       (codec() == config.codec()) &&
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -74,6 +74,27 @@ bool AudioDecoderConfig::IsValidConfig() const {
 }
 
 bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
+  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require 
+  //    re-initialization based on elementary stream properties (codec,
+  //    profile, resolution, extra_data).
+  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
+  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
+  //    boundaries.
+  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
+  //    container MIME strings differ while elementary video/audio parameters
+  //    remain identical.  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require 
+  //    re-initialization based on elementary stream properties (codec,
+  //    profile, resolution, extra_data).
+  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
+  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
+  //    boundaries.
+  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
+  //    container MIME strings differ while elementary video/audio parameters
+  //    remain identical.
   return (
       (codec() == config.codec()) &&
       (bytes_per_channel() == config.bytes_per_channel()) &&
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -73,26 +73,11 @@ bool AudioDecoderConfig::IsValidConfig() const {
          seek_preroll_ >= base::TimeDelta() && codec_delay_ >= 0;
 }
 
-bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
-  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)),
-  // `mime_type()` is intentionally omitted from Matches() to minimize risk to
-  // playback stability:
-  // 1. Matches() evaluates whether a stream config change requires tearing
-  //    down and re-initializing the underlying hardware decoder (SbPlayer /
-  //    MediaCodec).
-  // 2. The existing elementary stream fields compared below (codec, profile,
-  //    resolution, extra_data) already provide sufficient information to
-  //    detect any hardware-relevant format changes. Real format switches
-  //    (e.g., H.264 -> VP9) alter these elementary fields and correctly
-  //    trigger decoder re-initialization, making an additional `mime_type()`
-  //    comparison unnecessary.
-  // 3. Omitting `mime_type()` minimizes the risk of false-positive decoder
-  //    flushes: adding `mime_type()` introduces brand new behavior that is
-  //    risky. It's now possible that mime_type strings that didn't cause
-  //    Matches() to fail now do with this change, introducing new
-  //    regressions.
+// TODO(b/545325130): Explicitly signal SbPlayer of a changeType() call when
+// the config/mime_type is identical to the current config/mime_type.
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
   return (
       (codec() == config.codec()) &&
       (bytes_per_channel() == config.bytes_per_channel()) &&
@@ -108,6 +93,9 @@ bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
        config.should_discard_decoder_delay()) &&
       (target_output_channel_layout() ==
        config.target_output_channel_layout()) &&
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+      (mime_type() == config.mime_type()) &&
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
       (target_output_sample_format() == config.target_output_sample_format()));
 }
 
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -74,17 +74,25 @@ bool AudioDecoderConfig::IsValidConfig() const {
 }
 
 bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
-  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
-  // is intentionally excluded from Matches():
-  // 1. Matches() checks whether underlying hardware decoders require 
-  //    re-initialization based on elementary stream properties (codec,
-  //    profile, resolution, extra_data).
-  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
-  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
-  //    boundaries.
-  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
-  //    container MIME strings differ while elementary video/audio parameters
-  //    remain identical.
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)),
+  // `mime_type()` is intentionally omitted from Matches() to minimize risk to
+  // playback stability:
+  // 1. Matches() evaluates whether a stream config change requires tearing
+  //    down and re-initializing the underlying hardware decoder (SbPlayer /
+  //    MediaCodec).
+  // 2. The existing elementary stream fields compared below (codec, profile,
+  //    resolution, extra_data) already provide sufficient information to
+  //    detect any hardware-relevant format changes. Real format switches
+  //    (e.g., H.264 -> VP9) alter these elementary fields and correctly
+  //    trigger decoder re-initialization, making an additional `mime_type()`
+  //    comparison unnecessary.
+  // 3. Omitting `mime_type()` minimizes the risk of false-positive decoder
+  //    flushes: adding `mime_type()` introduces brand new behavior that is
+  //    risky. It's now possible that mime_type strings that didn't cause
+  //    Matches() to fail now do with this change, introducing new
+  //    regressions.
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
   return (
       (codec() == config.codec()) &&
       (bytes_per_channel() == config.bytes_per_channel()) &&
--- a/media/base/audio_decoder_config.cc
+++ b/media/base/audio_decoder_config.cc
@@ -110,6 +110,9 @@ std::string AudioDecoderConfig::AsHumanReadableString() const {
     << base::ToString(should_discard_decoder_delay())
     << ", target_output_channel_layout: "
     << ChannelLayoutToString(target_output_channel_layout())
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+    << ", mime_type: " << mime_type()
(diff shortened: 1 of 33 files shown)


media/filters/chunk_demuxer.cc
```diff
@@ -870,8 +870,15 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(*insert_result.first == id);
   CHECK(insert_result.second);  // Only true if insertion succeeded.
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  const std::string& mime_type = id_to_mime_map_[id];
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/mojo/services/starboard/starboard_renderer_wrapper_unittest.cc
```diff
@@ -427,15 +427,14 @@ TEST_F(StarboardRendererWrapperTest, ProxyDemuxerStreamDelegation) {
...
+  // Verify delegation of EnableBitstreamConverter() through proxy streams.
   EXPECT_CALL(*streams_[0], EnableBitstreamConverter());
   EXPECT_CALL(*streams_[1], EnableBitstreamConverter());
 
   auto proxy_streams = captured_media_resource->GetAllStreams();
   ASSERT_EQ(proxy_streams.size(), 2u);
-  EXPECT_EQ(proxy_streams[0]->mime_type(), "audio/mp4");
-  EXPECT_EQ(proxy_streams[1]->mime_type(), "video/mp4");
+  EXPECT_EQ(proxy_streams[0]->audio_decoder_config().mime_type(), "audio/mp4");
+  EXPECT_EQ(proxy_streams[1]->video_decoder_config().mime_type(), "video/mp4");
```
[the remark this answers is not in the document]
johnedocampo: Updated the test.

media/filters/source_buffer_state.h
```diff
@@ -50,6 +50,9 @@ class MEDIA_EXPORT SourceBufferState {
   ~SourceBufferState();
 
   void Init(StreamParser::InitCB init_cb,
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+            const std::string& mime_type,
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
             std::optional<std::string_view> expected_codecs,
             const StreamParser::EncryptedMediaInitDataCB&
                 encrypted_media_init_data_cb);
```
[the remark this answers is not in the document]
johnedocampo: This will be addressed in a follow-up PR, which focuses on adding support for ChangeType.

media/filters/source_buffer_state.cc
```diff
@@ -154,12 +154,18 @@ SourceBufferState::~SourceBufferState() {
 }
 
 void SourceBufferState::Init(StreamParser::InitCB init_cb,
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+                             const std::string& mime_type,
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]

media/filters/source_buffer_state.h
```diff
@@ -50,6 +50,9 @@ class MEDIA_EXPORT SourceBufferState {
   ~SourceBufferState();
 
   void Init(StreamParser::InitCB init_cb,
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+            const std::string& mime_type,
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: Done.

media/base/audio_decoder_config.cc
```diff
@@ -74,6 +74,17 @@ bool AudioDecoderConfig::IsValidConfig() const {
...
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require 
+  //    re-initialization based on elementary stream properties (codec,
+  //    profile, resolution, extra_data).
+  // 2. Format transitions (e.g., MediaSource.changeType) are explicitly
+  //    signaled down the pipeline via DemuxerStream::kConfigChanged buffer
+  //    boundaries.
+  // 3. Omitting `mime_type()` prevents redundant hardware decoder teardowns if
+  //    container MIME strings differ while elementary video/audio parameters
+  //    remain identical.
```
xiaomings: Can you elaborate why this will cause an unnecessary hardware decoder tear down?

media/base/audio_decoder_config.cc
```diff
@@ -74,6 +74,17 @@ bool AudioDecoderConfig::IsValidConfig() const {
 }
 
 bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
+  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
+  // is intentionally excluded from Matches():
+  // 1. Matches() checks whether underlying hardware decoders require
```
xiaomings: nit: remove trailing space after require, and the same to video_decoder_config.cc

media/base/audio_decoder_config.cc
```diff
@@ -74,6 +74,17 @@ bool AudioDecoderConfig::IsValidConfig() const {
 }
 
 bool AudioDecoderConfig::Matches(const AudioDecoderConfig& config) const {
+  // Note: On Starboard builds (BUILDFLAG(USE_STARBOARD_MEDIA)), `mime_type()`
```
xiaomings: nit: guard this comment under USE_STARBOARD_MEDIA macro.

media/base/audio_decoder_config.h
```diff
@@ -122,6 +122,10 @@ class MEDIA_EXPORT AudioDecoderConfig {
     return target_output_sample_format_;
   }
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  void set_mime_type(std::string_view mime_type) { mime_type_ = mime_type; }
+  const std::string& mime_type() const { return mime_type_; }
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
xiaomings: nit: empty line below

media/base/demuxer_stream.cc
```diff
@@ -45,12 +45,6 @@ const char* DemuxerStream::GetStatusName(Status status) {
 
 DemuxerStream::~DemuxerStream() = default;
 
-#if BUILDFLAG(USE_STARBOARD_MEDIA)
```
xiaomings, line 48: I saw a lot of code removed, can we find the original PR adding mime_type to DemuxerStream and create a revert PR for it?  This will create an explicit link between the two.

johnedocampo: The original PR that added mime_type to DemuxerStream is from https://github.com/youtube/cobalt/pull/4378.

We still need some of the functionality from this change (like updating web_media_sorce_impl.cc to accept a mime_type string), so this would only be a partial revert of the original PR. We could still do a full revert (and have the changes we need from the original PR made again here), but I feel that this is a little redundant.

As an alternative for an explicit link, would making an explicit comment in the commit message about how this PR partially reverts https://github.com/youtube/cobalt/pull/4378 work? I'm happy to go either way, please let me know what you think.

xiaomings: In that case feel free to just keep this PR as is, without creating a revert PR explicitly.

media/base/audio_decoder_config.cc
johnedocampo: Done.

media/base/audio_decoder_config.cc
johnedocampo: Done.

media/base/audio_decoder_config.h
johnedocampo: Done; added an empty line.

media/base/audio_decoder_config.cc
johnedocampo: My bad, the last comment did not exactly articulate what I meant.

After thinking about it, I think the risk vs. reward of adding mime_type to Matches() is not worth it.

Mainly for two reasons:

- Important changes in mime_type should be already reflected in the other fields. The existing fields in matches already provide enough information to detect if we really do need to update the decoder configs. If two mime type strings had two different codecs, we would already see that in the codecs field (similar to other fields as well). This makes matching mime_type here redundant.
- Reduce unneeded risk: Building off the first point, adding mime_type() here would be adding potential new behavior to the existing code. For me, this is a bit more risky of a change when compared to the rest of the changes. I think if we want to add mime_type() to Matches(), it would be best to run an experiment for it.

I've updated the comment here to better reflect my reasonings as to why mime_type is not part of Matches().

media/filters/chunk_demuxer.cc
```diff
@@ -870,8 +852,16 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(*insert_result.first == id);
   CHECK(insert_result.second);  // Only true if insertion succeeded.
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto it = id_to_mime_map_.find(id);
+  CHECK(it != id_to_mime_map_.end());
+  const std::string& mime_type = it->second;
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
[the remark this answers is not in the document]
johnedocampo: We will keep it CHECK for now.

media/base/audio_decoder_config.cc
xiaomings: The risk is minimum, as changeType() is a completely new feature, and mime_type() should stay the same if changeType() isn't called, thus doesn't change the return value of Matches() for existing playbacks.

I am not sure how we catch config changes later in the pipeline, but the manual comparisons later in StarboardRenderer related code could be eliminated if Matches() returns false when mime_type doesn't match, in which case SourceBufferStream (or a related class) may automatically handle this.

One thing might be still missing even if mime_type() is considered in Matches(), is that when changeType() is called with the same config/mime_type.  We may want to explicitly pass this signal to the SbPlayer.

Also "whether a stream config change requires tearing down and re-initializing the underlying hardware decoder (SbPlayer / MediaCodec)" is solely a decision of the underlying SbPlayer implementation.  The implementation may (or may not) decide that it won't recreate the hardware decoder if the codec and bit depth isn't changed, even if the mime type has changed.

johnedocampo: Discussed offline, updated Matches() to check mime_type. Also added a TODO for explicitly passing a signal to SbPlayer on changeType() calls with the same config/mime_type.

media/starboard/sbplayer_bridge.cc
```diff
@@ -327,40 +325,38 @@ SbPlayerBridge::~SbPlayerBridge() {
...
   DCHECK(audio_config.IsValidConfig());
 
   LOG(INFO) << "Updated AudioDecoderConfig -- "
             << audio_config.AsHumanReadableString();
 
   audio_config_ = audio_config;
-  audio_mime_type_ = mime_type;
+  audio_mime_type_ = audio_config.mime_type();
   audio_stream_info_ = MediaAudioConfigToSbMediaAudioStreamInfo(
       audio_config_, audio_mime_type_.c_str());
```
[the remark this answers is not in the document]
johnedocampo: Done. Removed `audio_mime_type_` as it is no longer needed.

media/starboard/sbplayer_bridge.cc
```diff
@@ -327,40 +325,38 @@ SbPlayerBridge::~SbPlayerBridge() {
...
   video_stream_info_.codec =
       MediaVideoCodecToSbMediaVideoCodec(video_config_.codec());
-  video_stream_info_.color_metadata =
-      MediaToSbMediaColorMetadata(video_config_.color_space_info(),
-                                  video_config_.hdr_metadata(), mime_type);
-  video_mime_type_ = mime_type;
+  video_stream_info_.color_metadata = MediaToSbMediaColorMetadata(
+      video_config_.color_space_info(), video_config_.hdr_metadata(),
+      video_mime_type_);
   video_stream_info_.mime = video_mime_type_.c_str();
```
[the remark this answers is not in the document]
johnedocampo: Done; removed `video_mime_type_` as it is no longer needed.

media/filters/chunk_demuxer.cc
```diff
@@ -870,8 +852,16 @@ ChunkDemuxer::Status ChunkDemuxer::AddIdInternal(
   CHECK(*insert_result.first == id);
   CHECK(insert_result.second);  // Only true if insertion succeeded.
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  auto it = id_to_mime_map_.find(id);
```
xiaomings: If we pass mime_type from `AddId()` to this function, we can remove `id_to_mime_map_`.

media/mojo/mojom/audio_decoder_config_mojom_traits.h
```diff
@@ -72,6 +72,12 @@ struct StructTraits<media::mojom::AudioDecoderConfigDataView,
 
   static bool Read(media::mojom::AudioDecoderConfigDataView input,
                    media::AudioDecoderConfig* output);
+
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  static const std::string& mime_type(const media::AudioDecoderConfig& input) {
+    return input.mime_type();
+  }
+#endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
```
xiaomings, line 77: Let's move this somewhere above Read(), to be closer to other attributes, and also in sync with how this is defined in media/mojo/mojom/video_decoder_config_mojom_traits.h.

xiaomings (commented):
@borongc please help review the Mojo related code.

media/starboard/starboard_renderer.cc
```diff
@@ -828,12 +823,12 @@ void StarboardRenderer::UpdateDecoderConfig(DemuxerStream* stream) {
 
   if (stream->type() == DemuxerStream::AUDIO) {
     const AudioDecoderConfig& decoder_config = stream->audio_decoder_config();
-    player_bridge_->UpdateAudioConfig(decoder_config, stream->mime_type());
+    player_bridge_->UpdateAudioConfig(decoder_config);
```
borongc, line 826: Mojo parts look good to me, but plz help to verify mime_type is passed via Mojo Renderer here, including when `CobaltBypassMojoForMedia` is enabled.

borongc:
nit: title should be `media: Move mime_type from ChunkDemuxerStream to config` as this change is across platforms.

media/filters/chunk_demuxer.cc
johnedocampo: Done; changed AddIdInternal to accept an optional `std::string_view mime_type`. Also refactored the custom Starboard implementation of AddID to directly go to AddIdInternal.

Also removed id_to_mime_map_.

media/mojo/mojom/audio_decoder_config_mojom_traits.h
johnedocampo: Done; moved mime_type() to above read, which is now synced with how it is defined in media/mojo/mojom/video_decoder_config_mojom_traits.h.

media/starboard/starboard_renderer.cc
johnedocampo: Verified locally through playing videos that mime type is passed, including with `CobaltBypassMojoForMedia` being enabled.

johnedocampo:
> nit: title should be `media: Move mime_type from ChunkDemuxerStream to config` as this change is across platforms.

Done.

media/filters/chunk_demuxer.cc
```diff
@@ -834,16 +816,40 @@ ChunkDemuxer::Status ChunkDemuxer::AddId(const std::string& id,
   if (!ParseMimeType(mime_type, &type, &codecs)) {
     return kNotSupported;
   }
-  DCHECK(!base::Contains(id_to_mime_map_, id));
-  id_to_mime_map_[id] = mime_type;
-  return AddId(id, type, codecs);
+  base::AutoLock auto_lock(lock_);
```
xiaomings: Consider adding a default parameter to the default `AddId()` and calling it from here, instead of duplicating the code below.  This helps with future rebase.

xiaomings (approved):

media/filters/chunk_demuxer.cc
johnedocampo: Done; updated the default AddId() to have an additional default parameter of mime_type. Simplified the code in the Starboard-specific AddId to call this special version.

PR #11747, opened 2026-08-06
Support --proxy-server command line flag

Awallky:
Cherrypicked from https://github.com/youtube/cobalt/compare/main...aelzeiny:cobalt-browser:filter_proxy_by_build_type

Bug: 543518257

### DIFF
--- a/cobalt/browser/BUILD.gn
+++ b/cobalt/browser/BUILD.gn
@@ -142,6 +142,24 @@ source_set("browser") {
       "//cobalt/browser/dial/public/mojom",
     ]
   }
+
+  if (!cobalt_is_release_build) {
+    deps += [ ":proxy_server_support" ]
+  }
+}
+
+if (!cobalt_is_release_build) {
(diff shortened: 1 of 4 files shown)


cobalt/browser/BUILD.gn
```diff
@@ -142,6 +142,24 @@ source_set("browser") {
       "//cobalt/browser/dial/public/mojom",
     ]
   }
+
+  if (!is_official_build) {
```
niranjanyardi: are we skipping qa builds ?

my understanding is we would like to run memory benchamarkks for qa builds . 
you would use is_cobalt_release_build - to avoid gold builds - wdyt ?

Awallky: Yeah good idea. Done!

niranjanyardi (approved):
I'm not sure if we could reuse any existing chromium functionality insterad of doing this - feel free to get another review if you want more feedback

PR #11748, opened 2026-08-06
starboard: Consolidate player settings extensions

borongc:
Consolidate single-purpose player settings extensions into a unified
PlayerSettings extension to reduce extension lookup overhead.

Refactor Android thread-local storage in video_max_video_input_size.cc
and video_surface_view.cc to use C++ thread_local variables within
anonymous namespaces, eliminating POSIX pthread_key_t / pthread_once_t
boilerplate, symbol pollution, and pointer-integer type casts.

Remove legacy PlayerSetMaxVideoInputSize and PlayerSetVideoSurfaceView
extensions and update sbplayer_bridge to use the new extension.

Issue: 438002792


starboard/extension/thread_local_player_settings.h
```diff
@@ -0,0 +1,47 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define kStarboardExtensionThreadLocalPlayerSettingsName \
+  "dev.starboard.extension.ThreadLocalPlayerSettings"
+
+typedef struct StarboardExtensionThreadLocalPlayerSettingsApi {
```
kjyoun: Do we need put "ThreadLocal" in the name?

I think that it is more implementation details and method alone already conveys meaning with  `ForCurrentThread` suffix

Then `PlayerSettings` would be enough.
For example, `ExperimentalFeaturesConfiguration` does not have it its name, though it is using thread local.

WDYT?

borongc: Done.

kjyoun (approved):

PR #11749, opened 2026-08-06
starboard: Remove unused PlayerConfiguration extension

kjyoun:
Remove starboard/extension/player_configuration.h and its associated methods and references. This extension is not used in Cobalt:
- SetDecodeToTexturePreferred is not called anywhere in Cobalt.
- SetTunnelModePreferred is not implemented or used anywhere.

Instead:
- Decode-to-texture mode is configured and negotiated per-player
  directly through the standard Starboard API (SbPlayerCreationParam
  and SbPlayerGetPreferredOutputMode).
- Tunnel mode is controlled per-stream through MIME attributes and
  Starboard feature flags.

Issue: 150410605


media/starboard/sbplayer_interface.cc
```diff
@@ -17,34 +17,10 @@
 #include <string>
 
 #include "base/logging.h"
-#include "starboard/extension/player_configuration.h"
 #include "starboard/system.h"
 
 namespace media {
 
-bool SbPlayerInterface::SetDecodeToTexturePreferred(bool preferred) {
```
borongc, line 25: I remember this is for tvOS. @jasonzhangxx to confirm.

kjyoun: per gemini

---

  Yes, Bo-Rong's memory is historically correct, but the extension is currently dead and completely disabled on tvOS.

  Here is the exact context and history from the codebase:
  ──────
  ### 1. Historical Context (Why Bo-Rong remembered tvOS)

  • Back in legacy Cobalt (cobalt/media/), MediaModule::SetConfiguration had a string hook for "PlayerConfiguration.DecodeToTexturePreferred"
  that called sbplayer_interface_->SetDecodeToTexturePreferred(value).
  • It was primarily intended for tvOS where decode-to-texture mode was being experimented with.
  ──────
  ### 2. Why it is completely unused today

  1. Dead on tvOS Backend:
      • In system_get_extensions.cc, tvOS explicitly returns nullptr for this extension:
	if (strcmp(name, kStarboardExtensionPlayerConfigurationName) == 0) {
	  // TODO: b/447334535 - This depends on decode to texture mode being
	  // implemented.
	  // return starboard::GetPlayerConfigurationApi();
	  return nullptr;
	}

      • Furthermore, starboard/tvos/shared/media/player_configuration.cc isn't even compiled in BUILD.gn.
  2. Dead on Cobalt Client:
      • With the migration to Chromium's media pipeline (media/starboard/), MediaModule::SetConfiguration was removed.
      • SbPlayerInterface::SetDecodeToTexturePreferred was carried over into media/starboard/sbplayer_interface.cc, but nothing in Cobalt or
      Chromium media ever calls it (0 callers).
  3. Replaced by Standard Starboard API:
      • Decode-to-texture mode is now negotiated per-player through creation_param.output_mode = kSbPlayerOutputModeDecodeToTexture and
      SbPlayerGetPreferredOutputMode(&creation_param).

jasonzhangxx: We can remove it. I don't think it's in use.

jasonzhangxx: To be cautious, let's confirm with @abhijeetk @rakuco.

kjyoun: @rakuco 

IIUC, tvos does not use this extension

Can you confirm it?

rakuco: Confirmed. Decode to texture is completely disabled at the moment (see b/447334535) and it'll need to be reimplemented differently in the future.

rakuco (approved):

jasonzhangxx (approved):

fayev-g (approved):

PR #11751, opened 2026-08-06
Implement non-blocking CI tests for update_client

arjungm:
Bug: 446745795


.github/actions/on_device_tests/action.yaml
oxve, line 1: I think all the changes to this file should be reverted. There should be minimal decision making in this job.

In fact, IMO this job _should_ fail even if the failing tests are optional. I think it's analogous to  similar to how we unrequire e.g. e2e tests when they are unstable.

Making this job pass would make it harder to retry optional tests.

.github/actions/test_targets_json/action.yaml
```diff
@@ -44,16 +44,16 @@ runs:
...
       shell: bash
 outputs:
   test_targets_count:
     description: The number of matched test targets.
     value: ${{ steps.parse-json.outputs.test_targets_count }}
+  test_objects_json:
+    description: The matched test targets in a JSON array of objects.
+    value: '${{ steps.parse-json.outputs.test_objects_json }}'
   test_targets_json:
     description: The matched test targets in a JSON array.
```
oxve, line 55: Can you update this description to make it clear that this is an array of target names (strings) as opposed to the objects in the other json?

.github/actions/test_targets_json/action.yaml
oxve, line 1: Can we improve the grokability of this further?

I think something like `test_target_configs_json` would be a better descriptor, perhaps even renaming `test_targets_json` to something like `test_targets_list_json` to further differentiate them.

.github/actions/on_host_tests/action.yaml
oxve, line 1: Same here as in the odt action above; Minimal decision making and make it easy to retry failing optional tests.

.github/workflows/main.yaml
oxve, line 1: Mind fixing the `${{ inputs.platform }}` template variable violations here?

cobalt/build/testing/targets/evergreen-arm-hardfp-rdk/test_targets.json
```diff
@@ -112,5 +112,12 @@
   {
     "target": "third_party/opus:test_opus_padding_loader",
     "executable": "out/evergreen-arm-hardfp-rdk_devel/test_opus_padding_loader.py"
+  },
+  {
+    "target": "components/update_client:update_client_unittests_loader",
+    "executable": "out/evergreen-arm-hardfp-rdk_devel/update_client_unittests_loader.py",
+    "runs_on": "device",
+    "test_type": "gtest",
+    "required_passing": false
```
oxve, line 127: Simply put "required"? The passing suffix seems superfluous.

cobalt/build/testing/targets/evergreen-x64/test_targets.json
```diff
@@ -201,5 +201,12 @@
     "executable": "out/evergreen-x64_devel/url_unittests_loader.py",
     "runs_on": "host",
     "test_type": "gtest"
+  },
+  {
+    "target": "components/update_client:update_client_unittests_loader",
+    "executable": "out/evergreen-x64_devel/update_client_unittests_loader.py",
+    "runs_on": "host",
+    "test_type": "gtest",
+    "required_passing": false
```
oxve, line 210: Same.

cobalt/tools/junit_mini_parser.py
oxve, line 1: I don't think these changes are necessary. The call point in `process_test_results/action.yaml` can be amended to handle optional tests without additional complexity here. This is after all the mini parser, not the fully featured one. :)

.github/workflows/main.yaml
oxve: `name: ${{ matrix.name }}_tests_passing`

Here is the test_passing job. It's the only one that actually required to pass to merge a PR. This is where you need to ensure that the optional test is handled as such.

I think this check should be made smarter and use test_targets.json to verify that all tests have valid results instead.

.github/actions/process_test_results/action.yaml
oxve, line 1: I thought some more about it and I think these changes should be reverted as well. For the same reason as stated elsewhere.

PR #11757, opened 2026-08-07
Implement HangWatcher UUID tracking and NSE recovery reporting

alunev:
cobalt: Implement hang UUIDs and recovery tracking

Introduce UUID-based tracking for thread hangs within HangWatcher to
better identify unique hang incidents. This enables the
NativeStabilityManager to correlate hang reports and track whether a
process successfully recovers from a frozen state.

HangWatcher now generates a unique ID upon detecting a hang and
injects it into Crashpad annotations via Starboard. It also monitors
thread deadlines to detect recovery, notifying the delegate when
all threads have resumed execution.

Bug: 528364110

### DIFF
--- a/base/threading/hang_watcher.cc
+++ b/base/threading/hang_watcher.cc
@@ -29,12 +29,14 @@
 #include "base/time/default_tick_clock.h"
 #include "base/time/time.h"
 #include "base/trace_event/base_tracing.h"
+#include "base/uuid.h"
 #include "build/build_config.h"
 
 #if BUILDFLAG(IS_STARBOARD)
 #include <inttypes.h>
 
 #include "starboard/extension/crash_handler.h"
+#include "starboard/extension/native_stability.h"
 #include "starboard/system.h"
 #endif
 
@@ -607,6 +609,9 @@ void HangWatcher::InitializeOnMainThread(ProcessType process_type,
 }
 
 void HangWatcher::UninitializeOnMainThreadForTesting() {
+#if BUILDFLAG(IS_COBALT)
+  g_hang_reporting_enabled.store(true, std::memory_order_relaxed);
+#endif
   g_use_hang_watcher.store(false, std::memory_order_relaxed);
   g_threadpool_log_level.store(LoggingLevel::kNone, std::memory_order_relaxed);
   g_io_thread_log_level.store(LoggingLevel::kNone, std::memory_order_relaxed);
@@ -897,7 +902,14 @@ void HangWatcher::Run() {
 #endif
     Wait();
 
-    if (!IsWatchListEmpty() &&
+    bool has_work = !IsWatchListEmpty();
+#if BUILDFLAG(IS_COBALT)
+    // If the watch list is empty but we have an active hang UUID, we still
+    // have work to do (cleaning up the recovery state).
+    has_work = has_work || !active_hang_uuid_.empty();
+#endif
+
+    if (has_work &&
         g_keep_monitoring.load(std::memory_order_relaxed)) {
       Monitor();
       if (after_monitor_closure_for_testing_) {
@@ -1157,6 +1169,83 @@ HangWatcher::WatchStateSnapShot HangWatcher::GrabWatchStateSnapshotForTesting()
   return snapshot;
 }
 
+#if BUILDFLAG(IS_COBALT)
+void HangWatcher::RecordHangStarted() {
+  DCHECK_CALLED_ON_VALID_THREAD(hang_watcher_thread_checker_);
+  if (!g_hang_reporting_enabled.load(std::memory_order_relaxed)) {
+    return;
+  }
+  if (!active_hang_uuid_.empty()) {
+    return;
+  }
+
+  active_hang_uuid_ = base::Uuid::GenerateRandomV4().AsLowercaseString();
+  std::string uuid_to_report = active_hang_uuid_;
+
+  {
+    base::AutoUnlock auto_unlock(watch_state_lock_);
+    auto* delegate = g_hang_watcher_delegate.load(std::memory_order_acquire);
+    if (delegate) {
+      delegate->RecordHangStarted(uuid_to_report);
+    }
+  }
+
+#if BUILDFLAG(IS_STARBOARD)
+  auto* crash_ext = static_cast<const CobaltExtensionCrashHandlerApi*>(
+      SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+  if (crash_ext && crash_ext->version >= 2 && crash_ext->SetString) {
+    crash_ext->SetString(kNativeStabilityHangUuidKey, uuid_to_report.c_str());
+  }
+#endif
+}
+
+void HangWatcher::CheckAndRecordHangRecovered() {
+  DCHECK_CALLED_ON_VALID_THREAD(hang_watcher_thread_checker_);
+
+  if (active_hang_uuid_.empty()) {
+    return;
+  }
+
+  bool any_thread_hung = false;
+  base::TimeTicks now = base::TimeTicks::Now();
+
+  // We iterate over the watch states to determine if ANY thread is currently
+  // physically hung. Note: We are ALREADY inside HangWatcher::Monitor(), so
+  // watch_state_lock_ is currently held by the caller. We do not need to
+  // acquire or release it here for the iteration.
+  for (const auto& watch_state : watch_states_) {
+    if (watch_state->GetDeadline() <= now) {
+      any_thread_hung = true;
+      break;
+    }
+  }
+
+  if (!any_thread_hung) {
+    std::string uuid_to_report = active_hang_uuid_;
+    active_hang_uuid_.clear();
+
+    {
+      // Unlock before calling out to the delegate to prevent deadlocks with
+      // embedder code.
+      base::AutoUnlock auto_unlock(watch_state_lock_);
+
+      auto* delegate = g_hang_watcher_delegate.load(std::memory_order_acquire);
+      if (delegate) {
+        delegate->RecordHangRecovered(uuid_to_report);
+      }
+    }
+
+#if BUILDFLAG(IS_STARBOARD)
+    auto* crash_ext = static_cast<const CobaltExtensionCrashHandlerApi*>(
+        SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+    if (crash_ext && crash_ext->version >= 2 && crash_ext->SetString) {
+      crash_ext->SetString(kNativeStabilityHangUuidKey, "");
+    }
+#endif
+  }
+}
+#endif
+
 void HangWatcher::Monitor() {
   DCHECK_CALLED_ON_VALID_THREAD(hang_watcher_thread_checker_);
 
@@ -1172,6 +1261,9 @@ void HangWatcher::Monitor() {
   // If all threads unregistered since this function was invoked there's
   // nothing to do anymore.
   if (watch_states_.empty()) {
+#if BUILDFLAG(IS_COBALT)
+    CheckAndRecordHangRecovered();
+#endif
     return;
   }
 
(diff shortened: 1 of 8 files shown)


base/threading/hang_watcher.h
```diff
@@ -396,6 +398,12 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
   // invokes the appropriate closure if so.
   void Monitor() LOCKS_EXCLUDED(watch_state_lock_);
 
+#if BUILDFLAG(IS_COBALT)
+  void CheckAndRecordHangRecovered()
+      EXCLUSIVE_LOCKS_REQUIRED(watch_state_lock_);
+  void RecordHangStarted();
+#endif
```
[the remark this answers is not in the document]
alunev: the lock is needed for checking state but I've added `base::AutoUnlock auto_unlock(watch_state_lock_);` for calling the delegate

base/threading/hang_watcher.cc
```diff
@@ -1157,6 +1162,66 @@ HangWatcher::WatchStateSnapShot HangWatcher::GrabWatchStateSnapshotForTesting()
...
+      auto* crash_ext = static_cast<const CobaltExtensionCrashHandlerApi*>(
+          SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+      if (crash_ext && crash_ext->version >= 2 && crash_ext->SetString) {
+        crash_ext->SetString(kNativeStabilityHangUuidKey, "");
+      }
+#endif
+      active_hang_uuid_.clear();
+    }
+  }
+}
```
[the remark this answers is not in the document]
alunev: thx, the lock is held by the caller. And for calling delegate I've added `base::AutoUnlock auto_unlock(watch_state_lock_);`

base/threading/hang_watcher.cc
```diff
@@ -29,12 +29,14 @@
 #include "base/time/default_tick_clock.h"
 #include "base/time/time.h"
 #include "base/trace_event/base_tracing.h"
+#include "base/uuid.h"
```
hlwarriner, line 32: nit: this `#include` should be guarded.

base/threading/hang_watcher.h
```diff
@@ -144,6 +144,8 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
   class BASE_EXPORT Delegate {
    public:
     virtual ~Delegate() = default;
+    virtual void RecordHangStarted(const std::string& hang_uuid) {}
```
hlwarriner, line 147: Please add function comments for all methods added to the HangWatcher and Delegate classes. Without comments it may not be intuitive to readers how `RecordHangStarted` differs from `RecordHang`, for example.

base/threading/hang_watcher.cc
```diff
@@ -1157,6 +1162,76 @@ HangWatcher::WatchStateSnapShot HangWatcher::GrabWatchStateSnapshotForTesting()
...
+    return;
+  }
+  if (!active_hang_uuid_.empty()) {
+    return;
+  }
+
+  active_hang_uuid_ = base::Uuid::GenerateRandomV4().AsLowercaseString();
+  std::string uuid_to_report = active_hang_uuid_;
+
+  LOG(INFO) << "Freeze detection: start reporting";
```
hlwarriner: nit: this is identical to the statement logged in `RecordHang()`. Is that intentional? It may help to include the method name or something to differentiate the logs.

base/threading/hang_watcher.cc
```diff
@@ -1157,6 +1162,76 @@ HangWatcher::WatchStateSnapShot HangWatcher::GrabWatchStateSnapshotForTesting()
...
+
+  {
+    base::AutoUnlock auto_unlock(watch_state_lock_);
+    auto* delegate = g_hang_watcher_delegate.load(std::memory_order_acquire);
+    if (delegate) {
+      delegate->RecordHangStarted(uuid_to_report);
+    }
+
+#if BUILDFLAG(IS_STARBOARD)
```
hlwarriner, line 1193: I recommend moving this `#if` block outside of the `{}` scope created for the `auto_unlock`.

For two reasons:
1. It's better to keep that scope as narrow as possible.
2. I think it's more readable, at least to me (for example finding the closing `}`).

base/threading/hang_watcher.cc
```diff
@@ -1157,6 +1162,76 @@ HangWatcher::WatchStateSnapShot HangWatcher::GrabWatchStateSnapshotForTesting()
...
+    if (crash_ext && crash_ext->version >= 2 && crash_ext->SetString) {
+      crash_ext->SetString(kNativeStabilityHangUuidKey, uuid_to_report.c_str());
+    }
+#endif
+  }
+}
+
+void HangWatcher::CheckAndRecordHangRecovered() {
+  DCHECK_CALLED_ON_VALID_THREAD(hang_watcher_thread_checker_);
+  if (!active_hang_uuid_.empty()) {
```
hlwarriner: nit: an early exit `if (active_hang_uuid_.empty())` would help here by reducing the (significant) nesting.

base/threading/hang_watcher.cc
```diff
@@ -1157,6 +1162,76 @@ HangWatcher::WatchStateSnapShot HangWatcher::GrabWatchStateSnapshotForTesting()
...
+        auto* delegate =
+            g_hang_watcher_delegate.load(std::memory_order_acquire);
+        if (delegate) {
+          delegate->RecordHangRecovered(uuid_to_report);
+        }
+#if BUILDFLAG(IS_STARBOARD)
+        auto* crash_ext = static_cast<const CobaltExtensionCrashHandlerApi*>(
+            SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+        if (crash_ext && crash_ext->version >= 2 && crash_ext->SetString) {
+          crash_ext->SetString(kNativeStabilityHangUuidKey, "");
```
hlwarriner: IIUC this is to reset the crashpad annotation. I get that we want to do this after `DumpWithoutCrashing()` returns, but it's surprising to me that we would do it in `CheckAndRecordHangRecovered()`. It seems unrelated to `CheckAndRecordHangRecovered()`'s other responsibilitites.

base/threading/hang_watcher.h
```diff
@@ -433,6 +441,9 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
   WatchStateSnapShot watch_state_snapshot_
       GUARDED_BY_CONTEXT(hang_watcher_thread_checker_);
 
+  std::string active_hang_uuid_
```
hlwarriner, line 457: Optional: a comment like this may help: `Communicates the unique hang identifier from RecordHangStarted() to a subsequent execution of CheckAndRecordHangRecovered().`

base/threading/hang_watcher.cc
```diff
@@ -1180,6 +1255,10 @@ void HangWatcher::Monitor() {
 
   if (watch_state_snapshot_.IsActionable()) {
     DoDumpWithoutCrashing(watch_state_snapshot_);
+  } else {
```
hlwarriner, line 1275: Is there an edge case we're missing where there could be both an `IsActionable()` `watch_state_snapshot_` *and* a recovered hang?

Before looking at this PR I didn't realize that `DoDumpWithoutCrashing()` and `RecordHang()`, and therefore a "hang" NSE, can batch multiple hung threads. So I'm thinking about the case where on the first run of `Monitor()`, the snapshot is actionable because thread x is hung but then on the second run of `Monitor()`, the snapshot is actionable because (only) thread y is hung.

Maybe you would say we're not missing an edge case because our definition of a "hang" is not exactly a hung thread?

base/threading/hang_watcher_unittest.cc
```diff
@@ -1171,6 +1171,387 @@ TEST_F(WatchHangsInScopeBlockingTest, MAYBE_NewScopeDoesNotBlockDuringCapture) {
...
+
+  EXPECT_NE(first_uuid, second_uuid);
+}
+
+// Scenario: Staggered / Overlapping Hangs
+// 1. Thread A hangs. Monitor generates ONE UUID and calls RecordHangStarted.
+// 2. Time passes. Thread B hangs.
+// 3. Thread A recovers. Thread B is still hung.
+// 4. Monitor detects Thread B has a newer deadline than Thread A's snapshot.
+//    It triggers a *second* Crashpad dump, but retains the existing UUID and
```
hlwarriner, line 1472: So we may not have a 1:1 mapping from hang NSE to hang Crashpad report? It's interesting that Crashpad considers there to be two "hangs" in this scenario but our Native Stability reporting considers there to be only one. We should at least capture this design decision in the hangs design doc.

base/threading/hang_watcher.cc
hlwarriner: edit: this is related to my comment about the `OverlappingHangsSingleUuid` test case. I'm sure the decision you've made here is reasonable but we should document the behavior in the code and raise it in the design doc so we can align on a decision if we haven't already.

hlwarriner (commented):
The changes look good overall but I do have a few questions about design decisions made around the scope of hangs and opportunies to mark them as recovered.

Feel free to address the nits and optional comments in follow-up changes.

base/threading/hang_watcher_unittest.cc
alunev: yes, that's intentional. Added comment about this and in the doc. In short, the NSE hang is considered recovered when there's no hanging threads left in `Monitor()` check.

base/threading/hang_watcher.cc
alunev: added comment about this. In short, the NSE hang is considered recovered when there's no hanging threads left  in `Monitor()` check.

base/threading/hang_watcher.h
alunev: thx, added.

base/threading/hang_watcher.cc
alunev: yes, dump is async and also logically `RecordHangStarted()` pairs with `CheckAndRecordHangRecovered()`, so we set uuid in first and clear in 2nd. uuids exist only in NSM context so it seems logical this way.

base/threading/hang_watcher.cc
alunev: makes sense

base/threading/hang_watcher.cc
alunev: thx, done

base/threading/hang_watcher.cc
alunev: I think I'll just remove this one for clarity.

base/threading/hang_watcher.h
alunev: done

base/threading/hang_watcher.cc
alunev: done

base/threading/hang_watcher.cc
hlwarriner: Ok, thanks. I'm not convinced this is the right design decision but it seems reasonable and I don't have a clearly better alternative to suggest. It sounds fine to me to stick with this implementation in the MVP and continue to discuss comparisons with alternatives in the design doc.

base/threading/hang_watcher_unittest.cc
hlwarriner: Ok, thanks. Please see https://github.com/youtube/cobalt/pull/11757#discussion_r3776312900 but sounds good to go with this approach for now.

base/threading/hang_watcher.cc
hlwarriner: Are you sure the dump is async?

I was looking through the call stack and found that (at least for linux)
* `CrashpadClient::DumpWithoutCrash()` calls `SignalHandler::Get()->HandleCrash()`
* `LaunchAtCrashHandler.HandleCrashImpl()` calls `waitpid()` from the parent process, which I think should block until the handler process is done snapshotting the hung process.

hlwarriner (approved):

base/threading/hang_watcher.cc
alunev: thanks, I take that back, it runs blocking on HW thread.

PR #11760, opened 2026-08-07
Allow paramaterization of Highest PMF functionality

Awallky:
Enabled highest PMF reporter parameterization interval configs. This aligns with A/B experiments by decoupling the static constants and mapping them to Finch flags safely when IS_COBALT is enabled.

Bug: 543085512


johnxwork (approved):

cobalt/browser/metrics/highest_pmf_reporter_browsertest.cc
```diff
@@ -0,0 +1,156 @@
+// Copyright 2025 The Cobalt Authors. All Rights Reserved.
```
y4vor: 2026?

y4vor (approved):

cobalt/browser/metrics/highest_pmf_reporter_browsertest.cc
Awallky: Thanks! Fixed.

PR #11766, opened 2026-08-07
[Starboard] Fix mouse wheel event handling

jellefoks:
Previously, mouse wheel events from Starboard were ignored in the PlatformEventSource to avoid crashes from constructing generic mouse events. This caused a regression in mouse scrolling functionality. In addition, mouse button drag events and chorded/multiple button presses were not fully tracked, and X11 side mouse buttons were partially unmapped.

This change includes:
1. **Mouse Wheel Event Handling**: Implements proper construction of `ui::MouseWheelEvent` calculating horizontal and vertical scroll offsets with standard wheel deltas (`ui::MouseWheelEvent::kWheelDelta`).
2. **Mouse Button State Tracking & Drag Support**: Tracks active pressed buttons via `current_pressed_mouse_buttons_` to preserve button flags during mouse moves/drags across platforms, handles chorded/multiple button presses, provides an unpress fallback for omitted key codes to prevent stuck buttons, and clears pressed state on window blur.
3. **X11 Side & Navigation Button Mapping**: Maps standard X11 buttons 8 & 9 and auxiliary side buttons 10 & 11 to `kSbKeyMouse4` (Back) and `kSbKeyMouse5` (Forward), routing them as standard DOM mouse events (`ui::EF_BACK_MOUSE_BUTTON` / `ui::EF_FORWARD_MOUSE_BUTTON`, buttons 3 and 4) rather than TV remote key events.
4. **Unified Modifiers & Code Size Optimizations**: Streamlines modifier mapping into a unified table and optimizes button lookup with a compact switch statement for minimal code footprint.
5. **Unit Tests**: Adds a comprehensive suite of 36 unit tests in `platform_event_source_starboard_unittest.cc` covering mouse moves, drags, chorded clicks, wheel offsets, keyboard modifiers, and edge cases.

Bug: 543203262

### DIFF
--- a/starboard/shared/x11/application_x11.cc
+++ b/starboard/shared/x11/application_x11.cc
@@ -541,6 +541,8 @@ enum {
   kWheelRightButton = 7,
   kPointerBackButton = 8,
   kPointerForwardButton = 9,
+  kPointerForwardButtonAlt = 10,
+  kPointerBackButtonAlt = 11,
 };
 
 SbKey XButtonEventToSbKey(XButtonEvent* event) {
@@ -561,9 +563,11 @@ SbKey XButtonEventToSbKey(XButtonEvent* event) {
     case kWheelRightButton:
       return kSbKeyRight;
     case kPointerBackButton:
-      return kSbKeyBrowserBack;
+    case kPointerBackButtonAlt:
+      return kSbKeyMouse4;
     case kPointerForwardButton:
-      return kSbKeyBrowserForward;
+    case kPointerForwardButtonAlt:
+      return kSbKeyMouse5;
     default:
       return kSbKeyUnknown;
   }
--- a/ui/ozone/platform/starboard/platform_event_source_starboard.cc
+++ b/ui/ozone/platform/starboard/platform_event_source_starboard.cc
@@ -14,6 +14,8 @@
 
 #include "ui/ozone/platform/starboard/platform_event_source_starboard.h"
 
+#include <cmath>
+
 #include "base/containers/fixed_flat_map.h"
 #include "base/logging.h"
 #include "base/task/single_thread_task_runner.h"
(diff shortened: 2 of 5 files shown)


ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -143,20 +143,40 @@ void PlatformEventSourceStarboard::HandleEvent(const SbEvent* event) {
       case kSbInputEventTypeInput:
         break;
     }
-    int flag = ui::EF_NONE;
-    if (kSbInputEventTypePress) {
-      flag = ui::EF_LEFT_MOUSE_BUTTON;
+    int flags = ui::EF_NONE;
+    if (input_data->type == kSbInputEventTypePress) {
+      flags |= ui::EF_LEFT_MOUSE_BUTTON;
+    }
```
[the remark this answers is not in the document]
jellefoks: Done. Added support for pointer button modifiers (`kSbKeyModifiersPointerButtonLeft`, `kSbKeyModifiersPointerButtonRight`, `kSbKeyModifiersPointerButtonMiddle`, `kSbKeyModifiersPointerButtonBack`, `kSbKeyModifiersPointerButtonForward`) in `GetEventFlags` and `CreateMouseInputEvent`. Also added unit tests verifying that normal move events have no button flags, drag events preserve button flags and produce `kMouseDragged`, and release events have clean flags.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -143,20 +143,40 @@ void PlatformEventSourceStarboard::HandleEvent(const SbEvent* event) {
...
     }
 
-    // Mouse wheel scrolls are separate from MouseEvent and will crash here. We
-    // need to handle them properly.
     if (event_type == ui::EventType::kMousewheel) {
-      return;
+      gfx::Vector2d offset(static_cast<int>(input_data->delta.x *
+                                            ui::MouseWheelEvent::kWheelDelta),
+                           static_cast<int>(input_data->delta.y *
+                                            ui::MouseWheelEvent::kWheelDelta));
```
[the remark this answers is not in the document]
jellefoks: Done. Updated to use `std::round` for computing wheel delta offsets and added unit tests testing fractional smooth scroll deltas.

ui/ozone/platform/starboard/test/platform_event_source_starboard_unittest.cc
```diff
@@ -0,0 +1,99 @@
...
+  std::optional<gfx::Vector2d> last_wheel_offset_;
+  std::optional<gfx::PointF> last_location_;
+};
+
+class PlatformEventSourceStarboardTest : public ::testing::Test {
+ protected:
+  base::test::SingleThreadTaskEnvironment task_environment_;
+};
+
+TEST_F(PlatformEventSourceStarboardTest, HandleMouseWheelEvent) {
```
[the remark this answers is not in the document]
jellefoks: Done. Added comprehensive parameterized test suites (`PlatformEventSourceStarboardKeyboardModifierTest` and `PlatformEventSourceStarboardPointerButtonTest`) with `TEST_P` covering all modifier flags and pointer button combinations.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -64,114 +224,63 @@ std::unique_ptr<ui::Event> CreateTouchInputEvent(const SbEvent* event) {
...
 void DeliverEventHandler(std::unique_ptr<ui::Event> ui_event) {
-  CHECK(ui::PlatformEventSource::GetInstance());
   static_cast<PlatformEventSourceStarboard*>(
       ui::PlatformEventSource::GetInstance())
       ->DeliverEvent(std::move(ui_event));
 }
 
+}  // namespace
+
 void PlatformEventSourceStarboard::HandleEvent(const SbEvent* event) {
```
jellefoks, line 237: This long function is split into separate `CreateKeyboardRemoteInputEvent`, `CreateMouseWheelInputEvent`, `CreateMouseInputEvent`, and  `CreateTouchInputEvent`

starboard/shared/x11/application_x11.cc
```diff
@@ -561,9 +563,11 @@ SbKey XButtonEventToSbKey(XButtonEvent* event) {
     case kWheelRightButton:
       return kSbKeyRight;
     case kPointerBackButton:
-      return kSbKeyBrowserBack;
+    case kPointerBackButtonAlt:
```
andrewsavage1, line 566: Is this just to be able to test on linux or would partners have to make similar changes?

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -64,114 +224,59 @@ std::unique_ptr<ui::Event> CreateTouchInputEvent(const SbEvent* event) {
...
-    int flag = ui::EF_NONE;
-    if (kSbInputEventTypePress) {
-      flag = ui::EF_LEFT_MOUSE_BUTTON;
-    }
+  }
 
-    // Mouse wheel scrolls are separate from MouseEvent and will crash here. We
-    // need to handle them properly.
-    if (event_type == ui::EventType::kMousewheel) {
-      return;
```
andrewsavage1, line 154: This is quite a large change to apply to 26.eap, >300 LOC changed without the tests. Could we apply a more targeted change to not have to worry about affecting other input types that we can cherry pick back to 26.eap then follow up on main with this broader cleanup to ozone event handling for starboard platforms?

jellefoks: This diff is large here because the function is too long and I split it in subfunctions to make it more maintainable.

Note that this PR is against main, not 26.eap nor 27.lts. For maintenance branches, a minimized change for only reported issue can be extracted if that's preferred.

See the comment above: This long function is split into separate CreateKeyboardRemoteInputEvent, CreateMouseWheelInputEvent, CreateMouseInputEvent, and CreateTouchInputEvent

starboard/shared/x11/application_x11.cc
jellefoks: These are two previously unmapped buttons. Partners making these change would improve their mouse support allowing mice that have those buttons to back/forward instead of default behavior of unknown buttons to play/pause.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -33,10 +35,168 @@ namespace ui {
...
+PlatformEventSourceStarboard::CreateMouseWheelInputEvent(const SbEvent* event) {
+  DCHECK_CALLED_ON_VALID_SEQUENCE(sequence_checker_);
+  if (!event || !event->data) {
+    return nullptr;
+  }
+  const auto* input_data = static_cast<const SbInputData*>(event->data);
+  int flags = GetEventFlags(*input_data);
+  gfx::Vector2d offset(
+      static_cast<int>(
+          std::round(input_data->delta.x * ui::MouseWheelEvent::kWheelDelta)),
```
jellefoks, line 126: The sign is wrong here, this scrolls in the opposite direction as C25.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
lamdbui: Let's keep it minimal for 26.eap as we want to get this resolved sooner, but keep it more isolated, if possible

ui/ozone/platform/starboard/platform_event_source_starboard.cc
jellefoks: Fixed in latest push

ui/ozone/platform/starboard/platform_event_source_starboard.cc
jellefoks: 26.eap got its own minimal PR

starboard/shared/x11/application_x11.cc
andrewsavage1: To clarify—are these breaking changes for partners or would their current implementations continue to work? It sounds like they would only need these changes if they wanted additional behavior?

starboard/shared/x11/application_x11.cc
```diff
@@ -561,9 +563,11 @@ SbKey XButtonEventToSbKey(XButtonEvent* event) {
     case kWheelRightButton:
       return kSbKeyRight;
     case kPointerBackButton:
-      return kSbKeyBrowserBack;
```
andrewsavage1, line 564: Why do we want to map kPointerBackButton to kSbKeyMouse4 instead of kSbKeyBrowserBack?

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -33,10 +35,171 @@ namespace ui {
...
+      static_cast<int>(
+          std::round(-input_data->delta.y * ui::MouseWheelEvent::kWheelDelta)));
+  return std::make_unique<ui::MouseWheelEvent>(
+      offset, gfx::PointF(input_data->position.x, input_data->position.y),
+      gfx::PointF(input_data->position.x, input_data->position.y),
+      base::TimeTicks() + base::Microseconds(event->timestamp), flags,
+      /*changed_button_flags=*/0);
+}
+
+std::unique_ptr<ui::Event> PlatformEventSourceStarboard::CreateMouseInputEvent(
```
andrewsavage1, line 145: I don't feel very strongly about this, but I wonder if it's confusing to call this function CreateMouseInputEvent when the wheel being used is usually on a remote

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -33,10 +35,171 @@ namespace ui {
...
+      event_type = ui::EventType::kMouseReleased;
+      break;
+    default:
+      return nullptr;
+  }
+
+  int flags = GetEventFlags(*input_data);
+  int changed_button_flags = 0;
+
+  if (input_data->type == kSbInputEventTypePress) {
```
andrewsavage1: nit: can't we combine this with the switch statement above?

ui/ozone/platform/starboard/platform_event_source_starboard.cc
```diff
@@ -64,114 +227,59 @@ std::unique_ptr<ui::Event> CreateTouchInputEvent(const SbEvent* event) {
                          data.size.x, data.size.y, pressure));
 }
 
-}  // namespace
+namespace {
```
andrewsavage1: Can we combine this with the existing anonymous namespace at the top of the file?

starboard/shared/x11/application_x11.cc
jellefoks: Before this PR, the `SbKey` output value of this function is effectively ignored inside Cobalt: Every mouse button press is sent into Cobalt as a left mouse button press: The code maps all mouse buttons to `ui::EF_LEFT_MOUSE_BUTTON` with the wonderfully always true if statement quoted below, completely ignoring the value returned by this function (which is stored in the 'key' member of the input ).

```
    if (kSbInputEventTypePress) {
      flag = ui::EF_LEFT_MOUSE_BUTTON;
    }
```
See https://github.com/youtube/cobalt/blob/main/ui/ozone/platform/starboard/platform_event_source_starboard.cc#L147-L149

The new code mirrors the mouse button mapping that chrome uses when reading from evdev (we get it from the X server so the numbers are different, but the mapping is (now) the same), here: https://github.com/youtube/cobalt/blob/main/ui/events/ozone/evdev/event_factory_evdev.cc#L319-L339

With this PR, the function `GetMouseButtonFromKey` added to `platform_event_source_starboard.cc` ensures that the mouse buttons are mapped correctly to Chrome's `ui::EF_FOO_MOUSE_BUTTON` button values.

And to ensure that the mapping is correct, the `PlatformEventSourceStarboardMouseButtonKeyTest` added in this PR tests that`kSbKeyMouseN` is mapped to the corresponding Chrome value for  `ui::EF_FOO_MOUSE_BUTTON` (where FOO is LEFT, MIDDLE, RIGHT, BACK, or FORWARD).

(manual testing also confirms with this PR  Cobalt behavior matches Chrome for those buttons).

starboard/shared/x11/application_x11.cc
jellefoks: Current implementations would continue to work. Even middle and right click are broken before this PR.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
jellefoks: This function is called purely for `device_type` of `kSbInputDeviceTypeMouse` input events. That already uses the term mouse. 

The `CreateKeyboardRemoteInputEvent` handles both `kSbInputDeviceTypeRemote` and `kSbInputDeviceTypeKeyboard`.

That said, this current PR does only handle `type = kSbInputEventTypeWheel` from `device_type` of  `kSbInputDeviceTypeMouse`.

That does match your smaller 'wheel only' fix merged to C26, but inspecting `system_window.cc` in C25, that still is a regression from C25 because that handled `type = kSbInputEventTypeWheel` on touchpad and touchscreen inputs and this current PR does not.

There is probablty no harm in treating all `type = kSbInputEventTypeWheel` the same independent from the `device_type`. That would make the behavior different from <=C25 and C26, but it would allow platforms to then send wheel events from the remote device type.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
jellefoks: Yes, that would clean it up, done.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
jellefoks: Yes, that would clean it up, done.

ui/ozone/platform/starboard/platform_event_source_starboard.cc
jellefoks: Changed accordingly: Now `type = kSbInputEventTypeWheel` is handled for all device_type values.

andrewsavage1 (approved):

PR #11768, opened 2026-08-07
Separate main and chromium autorollers.

briantting:
This way each autoroller can be triggered independently and run fewer steps so that it may run faster. The yamls are the same but with the main yaml missing depot_tools setup and with a leaner blobless checkout.

Bug: 543555380

### DIFF
--- a/.github/workflows/autoroll_chromium.yaml
+++ b/.github/workflows/autoroll_chromium.yaml
@@ -1,0 +1,164 @@
+name: Autoroll Chromium
+
+on:
+  schedule:
+    - cron: '0 */4 * * *' # Every 4th hour.
+  workflow_dispatch:
+    inputs:
+      max_commits:
+        description: 'Max number of commits to include on a single PR'
+        type: number
+        default: 10
+
+permissions:
+  contents: read
+
+jobs:
+  autoroll:
+    runs-on: ubuntu-latest
+    env:
+      DEPOT_TOOLS_METRICS: 0
+      GIT_AUTHOR_NAME: "cobalt-github-releaser-bot"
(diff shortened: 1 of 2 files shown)


.github/workflows/autoroll_chromium.yaml
```diff
@@ -1,8 +1,8 @@
-name: Autoroll LTS
+name: Autoroll Chromium
 
 on:
   schedule:
-    - cron: '0 * * * *' # Hourly.
+    - cron: '0 */4 * * *' # Every 4th hour.
```
linxinan-yt, line 5: Do we need to schedule cronjob for it? I think it is ok to manually trigger. If a rebase PR merged, I would rather triggering the aurotller to create the next one than waiting hours for the cronjob. WDYT?

linxinan-yt (approved):
LGTM % comment

.github/workflows/autoroll_chromium.yaml
briantting: i made it intentionally slow at every 4 hours with the expectation thatll wed manually trigger it when we want it to create a new one

even with the current cron of every hour, im constantly manually retriggering it, i think in history you can see me trigger 4-6 in 30mins

i wanted to leave it on a slow cron kind of as a check in case
- we just forgot to manually trigger
- we dont find outselves in a weird spot where only manual custom runs will work and the autoroller breaks on auto runs

briantting: if we find accommodating the auto runs is too messy we can remove that code completely, but right now i think its harmless to leave it in

it ensures that the manual runs and PR merges are will autoroll compliant and nothing breaks

PR #11769, opened 2026-08-07
Android: Add java switch for memory handling in foreground under critical pressure

tasunnn:
android: Add foreground critical memory switch

Introduce a Java switch and command-line flag to allow critical memory
pressure handling while the renderer is in the foreground.

By default, critical memory pressure is downgraded to moderate level
for foreground processes to avoid performance impact. This change
provides a runtime mechanism to bypass this downgrade, allowing for
experimentation with V8 memory management under critical pressure
without requiring a specific build configuration.

Bug: 521848285


content/renderer/render_thread_impl.cc
```diff
@@ -1821,8 +1821,12 @@ void RenderThreadImpl::OnSyncMemoryPressure(
 
 #if !BUILDFLAG(ALLOW_CRITICAL_MEMORY_PRESSURE_HANDLING_IN_FOREGROUND)
   // In order to reduce performance impact, translate critical level to
-  // moderate level for foreground renderer.
-  if (!RendererIsHidden() &&
```
Awallky, line 1825: Should this include cobalt BUILDFLAG guards?

content/renderer/render_thread_impl.cc
```diff
@@ -1821,8 +1821,12 @@ void RenderThreadImpl::OnSyncMemoryPressure(
...
+  // moderate level for foreground renderer unless for experiment.
+  const bool allow_critical_in_foreground =
```
Awallky: Since we only need to read this once, consider adding a getter method, with a static variable inside, which ensures we lazily initialize this once. From there on, we simply get the value determined after the one-time init.

e.g
```cpp
const bool getAllowCriticalInFg() {
    static bool allow_critical_in_foreground =
        base::CommandLine::ForCurrentProcess()->HasSwitch(
            "allow-critical-memory-pressure-handling-in-foreground");
    return allow_critical_in_foreground;
}
```

tasunnn: great point, just made it static, thanks

content/renderer/render_thread_impl.cc
tasunnn: fixed, thanks!

content/renderer/render_thread_impl.cc
```diff
@@ -1821,10 +1821,21 @@ void RenderThreadImpl::OnSyncMemoryPressure(
 
 #if !BUILDFLAG(ALLOW_CRITICAL_MEMORY_PRESSURE_HANDLING_IN_FOREGROUND)
```
Awallky, line 1831: Do we have this build flag enabled in our release builds?

content/renderer/render_thread_impl.cc
```diff
@@ -1821,10 +1821,21 @@ void RenderThreadImpl::OnSyncMemoryPressure(
...
   // In order to reduce performance impact, translate critical level to
-  // moderate level for foreground renderer.
+  // moderate level for foreground renderer for non cobalt builds.
+#if BUILDFLAG(IS_COBALT)
+  static const bool allow_critical_in_foreground =
```
Awallky: Recommend wrapping it in a function-local getter in anonymous namespace to defer initialization until the function is first called and keep the static var scope to this file.

Following Google Style guidelines, constants fixed for the duration of the program should use the k prefix followed by mixed case:

```cpp
bool IsCriticalAllowedInForeground() {
  static const bool kAllowCriticalInForeground =
      base::CommandLine::ForCurrentProcess()->HasSwitch(
          "allow-critical-memory-pressure-handling-in-foreground");
  return kAllowCriticalInForeground;
}
```

Awallky (approved):
LGTM % comments?

content/renderer/render_thread_impl.cc
tasunnn: this flag is enabled if either is_cast_android or is_castos is true, both of the flags are default to [false](https://source.corp.google.com/h/github/youtube/cobalt/+/main:build/config/cast.gni;l=28-32?q=build%2Fconfig%2Fcast.gni&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain) and I don't seem to find any platform overrides for them in cobalt, so the flag should be disabled in all cobalt builds

content/renderer/render_thread_impl.cc
tasunnn: directly initializing the static var as is right now should be functionally the same (defer initialization until called and fix scope), but yeah wrapping them inside a function would be better for style, will make the change, thanks

PR #11770, opened 2026-08-07
Disable UpdaterModule in Evergreen Lite mode (b/541294441)

yuying-y:
For Evergreen Lite modes, the Cobalt Updater module should not be initialized. We check for the availability of the Installation Manager extension (kCobaltExtensionInstallationManagerName); if it's missing, it indicates that Evergreen Lite is active, and therefore the UpdaterModule should not be instantiated.

TAG=agy
CONV=082e83d2-b5c5-4d2e-84b9-ec8300020c42

Bug: 541294441

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -85,7 +85,8 @@
 
 #if BUILDFLAG(USE_EVERGREEN)
 #include "cobalt/updater/updater_module.h"  //nogncheck
-#include "content/public/browser/storage_partition.h"
+#include "starboard/extension/installation_manager.h"
+#include "starboard/system.h"
 #endif  // BUILDFLAG(USE_EVERGREEN)
 
 #if BUILDFLAG(IS_ANDROID)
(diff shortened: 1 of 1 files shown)


yuying-y:
@gemini-code-assist The inclusion of storage_partition.h was duplicated at lines 72 and 88, so we only removed the redundant include from line 88. Compilations and CI have passed successfully since line 72 satisfies the IWYU requirements!

cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -441,10 +432,14 @@ void CobaltContentBrowserClient::OnWebContentsCreated(
...
-    LOG(INFO) << "Creating UpdaterModule singleton.";
-    updater::UpdaterModule::CreateInstance(
-        storage_partition->GetURLLoaderFactoryForBrowserProcess(),
-        GetUserAgent(), updater::kDefaultUpdateCheckDelay);
+    if (SbSystemGetExtension(kCobaltExtensionInstallationManagerName)) {
+      LOG(INFO) << "Creating UpdaterModule singleton.";
+      updater::UpdaterModule::CreateInstance(
+          storage_partition->GetURLLoaderFactoryForBrowserProcess(),
+          GetUserAgent(), updater::kDefaultUpdateCheckDelay);
+    } else {
```
hlwarriner, line 440: I shared this perspective in the bug:
> When deciding whether to create the updater module instance I do think we should check the --evergreen_lite switch value directly instead of only looking at the presence of the InstallationManager Starboard extension. This provides a cleaner API contract with partners, since --evergreen_lite should be all that's required in order to turn off Evergreen-Full behaviors.

Can you please reply to that either here or in the bug?

yuying-y: I prefer to keep the original approach, because it's consistent with what we've  been doing in the code base, to difference EG-full vs lite, e.g. https://source.corp.google.com/h/lbshell-internal/cobalt_src/+/main:cobalt/browser/user_agent/user_agent_platform_info.cc;l=282.

For the kEvergreenLite switch, we only checks it in starboard/loader_app/loader_app.cc

hlwarriner: To elaborate on my concern, I really don't think this provides a clean interface for partners. We tell partners they should run Evergreen Lite by providing the `--evergreen_lite` switch - https://github.com/youtube/cobalt/blob/main/starboard/doc/evergreen/cobalt_evergreen_lite.md#how-to-run-evergreen-lite - and I feel we should respect that.

For example if they don't set `--evergreen_lite` and they also don't provide an implementation of the IM extension, I think it would be better for us to raise an error on startup rather than assume the partner wanted EG Lite. They may have wanted to use EG Full and made a mistake by not implementing the IM extension.

I get your point about this approach existing in the codebase, though. I'm fine with this PR getting merged as is and I'll file a bug for us to consider improving the approach later.

hlwarriner (approved):

PR #11781, opened 2026-08-10
android: Allow all origins for injected Java bridge objects

abhijeetk:
The M138 rebase (2b6e73fd8a78) adapted to the new fourth parameter of
JavascriptInjector.addPossiblyUnsafeInterface() by passing an empty
origin allowlist.

An empty allowlist matches no origin, so every Java object injected
into the WebContents became unreachable from JavaScript. That includes
the 2-argument HTMLMediaElement.canPlayType(mimeType, keySystem)
override, which checks DRM capability via
SbMediaCanPlayMimeAndKeySystem().

The breakage is silent. The polyfill installs the override only when 
the HTMLMediaElementExtension global is defined, so the standard
1-argument canPlayType() stays in place with no error.

Pass "*" to allow all origins again, as before M138.

google-java-format also corrected pre-existing formatting in this file
that is unrelated to the fix.

Bug: 543808733

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -18,7 +18,6 @@ import static dev.cobalt.util.Log.TAG;
 
 import android.content.Intent;
 import android.net.Uri;
-import java.util.Locale;
 import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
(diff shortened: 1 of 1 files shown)


abhijeetk:
cc : @Gyuyoung @jkim-julie @rakuco @Sazabi1976

johnxwork:
We just need the media team to give approval here. We can rubber stamp. Jason seems to be the eng that introduced the HTMLMediaElementExtension java injection

jasonzhangxx:
> We just need the media team to give approval here. We can rubber stamp. Jason seems to be the eng that introduced the HTMLMediaElementExtension java injection

@johnxwork, I think the change makes all origin allowed with originAllowlist of "*"". It looks ok to me, but need your opinion on the origin allowlist change.

jasonzhangxx (approved):

PR #11796, opened 2026-08-11
cobalt: Release all GPU resources and EGL contexts on background suspend

jellefoks:
When Cobalt transitions to the background (concealed or suspended), all GPU resources are now released to minimize memory footprint. This includes marking client renderer contexts lost, tearing down GPU channels, and destroying shared EGL contexts and Skia states.

Offscreen surfaces and native platform window handles are also destroyed and lazily re-created upon returning to the foreground. This change provides a significant reduction in Private Dirty and PSS memory usage, as verified by automated memory benchmark tools.

### Background & Architecture
This change is non-trivial because upstream Chromium does not have a pre-existing path to release all GPU resources without exiting the application or killing the dedicated GPU process. In Chromium multi-process mode, full GPU resource release is handled by terminating the separate GPU process and spawning a new one on demand. In Cobalt's single-process in-process GPU architecture on Starboard, all GPU services run on an in-process thread; completely tearing down the EGL display via `eglTerminate()`, destroying shared context states and textures, and safely re-initializing the display and offscreen surfaces on resume requires coordinating lifecycle transitions across UI, GPU, and renderer threads.

### Summary of Changes
- **Release all GPU resources on conceal**: Synchronously tears down all GPU channels, marks client contexts lost, destroys `SharedContextState` (Skia GrDirectContext), offscreen pbuffer surfaces, native EGL window surfaces, and invokes `eglTerminate()` via `GLDisplayEGL::Shutdown()`.
- **Restore GPU subsystem and offscreen surfaces on reveal**: Proactively re-initializes `GLDisplayEGL`, creates the default offscreen surface, and flushes any pending `EstablishGpuChannel` requests queued while backgrounded.
- **Clean up lifecycle transition logging**: Streamlines lifecycle transition logs across `cobalt/app/app_event_delegate.cc` and `sb_event_handle.cc` into atomic, inline log messages without redundant state tracking.
- **Notify memory pressure and reclaim memory on conceal**: Proactively invokes `base::MemoryPressureListener::NotifyMemoryPressure(kCritical)` during conceal to purge discardable memory, V8 heap caches, font caches, and Skia resource caches.
- **Scope HangWatcher pausing to frozen state**: Restricts `base::HangWatcher::Suspend()` and `Resume()` strictly to `OnFreeze()` / `OnUnfreeze()` (where the OS halts CPU execution with `SIGSTOP`), preserving active hang detection while running backgrounded in `kConcealed`.
- **Fix HangWatcher startup initialization race**: Reorders initialization in `CobaltMainDelegate::PostEarlyInitialization` so base HangWatcher globals are initialized on the main thread before the delegate applies Finch/GlobalSettings configurations, preventing a `DCHECK` failure on startup when Finch enables renderer thread dumping.
- **Direct Starboard Stop event dispatch**: Resolves event starvation in `starboard/shared/starboard/application.cc` by directly executing `HandleEventAndUpdateState(kSbEventTypeStop)` after intermediate synthetic transitions instead of re-injecting into the event queue.
- **Single-process teardown stabilization**: Prevents teardown deadlocks in single-process mode by releasing `InProcessRendererThread` during process termination and handling expected IPC channel shutdown errors.

Bug: 525810025


jellefoks:
### Benchmark Methodology & Reproduction Steps

To ensure empirical reproducibility, the 25-iteration memory benchmark compares a clean upstream `origin/main` baseline binary against this branch under identical execution parameters.

---

#### 1. Test Environment & Setup
* **Platform**: Linux Evergreen QA (`evergreen-x64` / `qa`)
* **Display Server**: X11 (`DISPLAY=:100` via Xvfb or active desktop display)
* **Target WebApp**: `https://www.youtube.com/tv`
* **Metrics Source**: Linux Kernel `/proc/$PID/smaps_rollup` (`Private_Dirty`, `Pss`, `Rss`) and `/proc/$PID/status` (`VmData`)
* **Iterations**: 25 independent cold-start runs per build

---

#### 2. Lifecycle Execution Flow per Iteration
Each iteration is automated via `cobalt/tools/performance/memory/run_multi_benchmark.py`:
1. **Cold Start & Warmup (15s)**:
   - Launches `${BUILD_DIR}/loader_app --url=https://www.youtube.com/tv`.
   - Waits 15 seconds for initial DOM rendering, splash screen transition, and recommendation shelf asset streaming.
2. **Foreground State Sampling (5s)**:
   - Samples `/proc/$PID/smaps_rollup` at 1-second intervals and records median foreground metrics.
3. **Conceal / Background Transition (SIGUSR2)**:
   - Injects `SIGUSR2` (signal 12) into `loader_app` to trigger `CobaltLifecycleManager` conceal, tearing down `gpu_channels_`, shared EGL contexts, and offscreen surfaces.
   - Waits 2 seconds for event loop draining.
4. **Background State Sampling (5s)**:
   - Samples `/proc/$PID/smaps_rollup` at 1-second intervals and records median background metrics.
5. **Reveal / Foreground Restore (SIGCONT)**:
   - Injects `SIGCONT` (signal 18) into `loader_app` to trigger reveal and verify lazy display/surface re-initialization.
6. **Process Termination**:
   - Gracefully terminates the application process and waits for socket/IPC cleanup before starting the next iteration.

---

#### 3. Reproduction Commands

```bash
# 1. Build authentic clean baseline (origin/main)
git checkout origin/main
COBALT_OUT_DIR=out/evergreen-x64_qa_baseline \
  node .gemini/config/skills/cobalt-build-run-test/scripts/build.cjs evergreen-x64 qa cobalt

# 2. Build patched branch
git checkout feature/gpu-resource-teardown-on-suspend
node .gemini/config/skills/cobalt-build-run-test/scripts/build.cjs evergreen-x64 qa cobalt

# 3. Execute 25 iterations for baseline
python3 cobalt/tools/performance/memory/run_multi_benchmark.py \
  --build_dir=out/evergreen-x64_qa_baseline \
  --iterations=25 \
  --warmup=15 \
  --measure=5 \
  --output_csv=/tmp/true_baseline_25runs.csv

# 4. Execute 25 iterations for patched build
python3 cobalt/tools/performance/memory/run_multi_benchmark.py \
  --build_dir=out/evergreen-x64_qa \
  --iterations=25 \
  --warmup=15 \
  --measure=5 \
  --output_csv=/tmp/true_patched_25runs.csv
```

jellefoks:
## 📊 Linux Evergreen Empirical Lifecycle Metrics

(Verified using continuous lifecycle transitions against https://www.youtube.com/tv on `evergreen-x64_qa`)

### Process Memory Comparison: Before vs. After Feature Branch (`/proc/<pid>/status` & `/proc/<pid>/smaps_rollup`)

| Metric | Source | Before (Stock `main`)<br>Foreground → Background (Delta) | After (This Branch)<br>Foreground → Background (Delta) | Improvement with Feature Branch |
| :--- | :--- | :--- | :--- | :--- |
| **`VmRSS`** | `/proc/<pid>/status` | 362.33 MB → 358.37 MB (**-3.95 MB**, -1.09%) | 355.87 MB → 340.09 MB (**-15.78 MB**, -4.43%) | **4.0x more VmRSS freed** |
| **`RssShmem`** | `/proc/<pid>/status` | 18.92 MB → 15.24 MB (**-3.69 MB**, -19.49%) | 12.52 MB → 0.19 MB (**-12.33 MB**, **-98.47%**) | **GPU memory buffers 100% reclaimed** |
| **`PSS`** | `/proc/<pid>/smaps_rollup` | 309.90 MB → 304.64 MB (**-5.26 MB**, -1.70%) | 302.56 MB → 289.47 MB (**-13.09 MB**, -4.33%) | **2.5x more PSS reduction** |
| **`Private_Dirty`** | `/proc/<pid>/smaps_rollup` | 160.56 MB → 153.32 MB (**-7.24 MB**, -4.51%) | 150.13 MB → 140.79 MB (**-9.33 MB** to **-21.95 MB**) | **Substantial dirty memory reduction** |
| **`RssAnon`** | `/proc/<pid>/status` | 265.02 MB → 264.76 MB (**-0.27 MB**, -0.10%) | 263.00 MB → 259.94 MB (**-3.06 MB**, -1.16%) | **11x more anonymous memory freed** |

> **Net Impact Summary:**
> - **4x VmRSS Reduction:** Background conceal frees **~15.8 MB total VmRSS** (and up to **-23.7 MB VmRSS** on initial settled transition) compared to only -3.95 MB on stock `main`.
> - **Complete GPU Memory Reclaim:** **`RssShmem`** drops by **-12.33 MB (-98.5%)** from `12.52 MB` down to `0.19 MB`, proving that GPU channels, native EGL contexts, and shared buffer allocations are fully released upon background conceal.
> - **Seamless Resume:** Memory recovers cleanly upon reveal/resume with zero regressions or memory leaks across continuous cycles.

cobalt/app/app_event_runner.cc
```diff
@@ -213,6 +213,9 @@ class AppEventRunnerImpl : public AppEventRunner,
   void DoConceal() override {
     content::Shell::OnConceal();
     WaitForAck(PendingAck::kConceal);
+    base::MemoryPressureListener::NotifyMemoryPressure(
+        base::MemoryPressureListener::MEMORY_PRESSURE_LEVEL_CRITICAL);
+    ::partition_alloc::MemoryReclaimer::Instance()->ReclaimAll();
```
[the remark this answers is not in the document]
jellefoks: Done. Deferred ReclaimAll() by posting it to the task runner so observers across threads have dispatched their critical memory pressure handlers and released caches prior to PartitionAlloc purging.

gpu/ipc/service/gpu_channel_manager.cc
```diff
@@ -880,6 +920,27 @@ void GpuChannelManager::HandleMemoryPressure(
...
+          /*fallback_to_software_gl=*/false,
+          /*disable_gl_drawing=*/false,
+          /*init_extensions=*/true,
+          gl::GpuPreference::kDefault);
+    }
+    default_offscreen_surface_ = gl::init::CreateOffscreenGLSurface(
+        gl::GetDefaultDisplayEGL(), gfx::Size());
+  }
+  return default_offscreen_surface_.get();
+}
```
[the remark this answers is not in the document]
jellefoks: Done. Cached the display pointer, added a null check guard, and reused the cached display pointer in CreateOffscreenGLSurface.

ui/ozone/platform/starboard/platform_window_starboard.cc
```diff
@@ -180,6 +180,7 @@ void PlatformWindowStarboard::Hide() {
     widget_available_ = false;
     delegate_->OnAcceleratedWidgetDestroyed();
   }
+  DestroySbWindowInstance();
```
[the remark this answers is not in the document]
jellefoks: A lazy recreation mechanism is already in place in PlatformWindowStarboard::Show(). When returning to the foreground, Show() checks !SbWindowIsValid(sb_window_), calls SbWindowCreate(&options), invokes the global creation callback, and re-notifies OnAcceleratedWidgetAvailable() with the new platform handle.

jellefoks:
Converted to draft for now, adding eglTerminate/eglInitialize roundtrip for resume.

cobalt/app/app_event_delegate.cc
```diff
@@ -348,6 +348,14 @@ void AppEventDelegate::ExecuteStepOnUIThread(ApplicationState next_state,
...
     base::AutoLock lock(lock_);
     SetApplicationState(next_state);
+    if (application_state_ != target_state_) {
+      LOG(INFO) << "Transition to " << GetStateString(next_state)
+                << " complete (target: " << GetStateString(target_state_)
+                << ")";
+    } else {
+      LOG(INFO) << "Transition to " << GetStateString(next_state)
+                << " complete";
+    }
```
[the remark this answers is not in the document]
jellefoks: In AppEventDelegate, ExecuteNextStep() already checks if (application_state_ == target_state_) to reset is_transitioning_ = false and execute quit_closure_ upon reaching the target state. For all intermediate and terminating lifecycle transitions other than kStopped, ExecuteStepOnUIThread sets should_execute_next_step = true which dispatches ExecuteNextStep() to finalize the state progression and unblock run_loop.Run(). ExecuteStepOnUIThread only executes quit_closure_ directly for kStopped because kStopped is a terminal state that does not invoke ExecuteNextStep().

components/viz/service/gl/gpu_service_impl.cc
```diff
@@ -1258,6 +1284,41 @@ void GpuServiceImpl::OnForegrounded() {
...
+          /*fallback_to_software_gl=*/false,
+          /*disable_gl_drawing=*/false,
+          /*init_extensions=*/true,
+          gl::GpuPreference::kDefault);
+    }
+    if (display->IsInitialized()) {
+      scoped_refptr<gl::GLSurface> surface =
+          gl::init::CreateOffscreenGLSurface(display, gfx::Size());
+      gpu_channel_manager_->SetDefaultOffscreenSurface(std::move(surface));
+    }
```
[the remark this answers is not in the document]
jellefoks: Done. Added a check for !gpu_channel_manager_->default_offscreen_surface() before creating and setting the default offscreen surface so repeated foreground calls during resume do not redundantly recreate it.

gpu/ipc/service/gpu_channel_manager.cc
```diff
@@ -799,6 +799,43 @@ void GpuChannelManager::OnBackgroundCleanup() {
...
+  if (shared_context_state_) {
+    shared_context_state_->ReleaseCurrent(nullptr);
+    shared_context_state_->MarkContextLost();
+    if (shared_context_state_->real_context()) {
+      shared_context_state_->real_context()->Destroy();
+    } else if (shared_context_state_->context()) {
+      shared_context_state_->context()->Destroy();
+    }
+    shared_context_state_.reset();
+  }
```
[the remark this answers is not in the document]
jellefoks: Done. Removed the redundant manual context destruction calls from OnBackgroundCleanup() since SharedContextState::MarkContextLost() already explicitly destroys the context handles under BUILDFLAG(IS_COBALT).

ui/gl/gl_surface_egl.cc
```diff
@@ -534,12 +564,12 @@ void NativeViewGLSurfaceEGL::Destroy() {
   vsync_provider_internal_ = nullptr;
 
   if (surface_) {
-    if (!eglDestroySurface(display_->GetDisplay(), surface_)) {
```
y4vor, line 537: lets keep the original code and do the IS_COBALT guard.

ui/gl/gl_surface_egl.cc
```diff
@@ -57,6 +61,32 @@ struct TraceSwapEventsInitializer {
 static base::LazyInstance<TraceSwapEventsInitializer>::Leaky
     g_trace_swap_enabled = LAZY_INSTANCE_INITIALIZER;
 
+void SafeDestroyEGLSurface(GLDisplayEGL* display, EGLSurface surface) {
```
y4vor, line 65: lets not change the existing code. All of our customizations need to be guarded.

third_party/blink/renderer/platform/mediastream/webrtc_uma_histograms.cc
```diff
@@ -26,7 +26,9 @@ void UpdateWebRTCMethodCount(RTCAPIName api_name) {
 }
 
 PerSessionWebRTCAPIMetrics::~PerSessionWebRTCAPIMetrics() {
+#if !BUILDFLAG(IS_STARBOARD)
```
y4vor: Something wrong with our threads? We should at least document what is happing and file ticket to investigate further.

y4vor (changes requested):
Also these I think are legitimate suggestions: 
https://paste.googleplex.com/5541097196093440

components/viz/service/gl/gpu_service_impl.cc
```diff
@@ -76,7 +76,13 @@
 #include "ui/base/ozone_buildflags.h"
 #include "ui/base/ui_base_features.h"
 #include "ui/gl/gl_context.h"
+#if BUILDFLAG(IS_COBALT)
```
andrewsavage1: nit: it's better to put the conditional includes below (and in a single block) like the rest of them

content/public/browser/gpu_utils.cc
```diff
@@ -26,6 +26,7 @@
 #include "media/base/media_switches.h"
 #include "media/gpu/buildflags.h"
 #include "media/media_buildflags.h"
+#include "mojo/public/cpp/bindings/callback_helpers.h"
```
andrewsavage1: nit: guard this with a buildflag at the bottom of the includes

components/viz/service/gl/gpu_service_impl.cc
```diff
@@ -1258,6 +1307,42 @@ void GpuServiceImpl::OnForegrounded() {
...
+      gpu_channel_manager_->SetDefaultOffscreenSurface(std::move(surface));
+    }
+  }
+
+  // Fulfill all channel establishment requests that arrived while backgrounded,
+  // now that the display and default offscreen surface are re-initialized.
+  auto pending_requests = std::move(pending_establish_gpu_channel_requests_);
+  pending_establish_gpu_channel_requests_.clear();
+  for (auto& request : pending_requests) {
+    EstablishGpuChannel(request.client_id, request.client_tracing_id,
```
andrewsavage1: If the display is not initialized correctly on foregrounding, is it possible that these requests are never fulfilled and end up hanging?

gpu/ipc/service/gles2_command_buffer_stub.cc
andrewsavage1, line 199: Would this create a surface while the app is backgrouded?

cobalt/shell/browser/shell_platform_delegate.cc
```diff
@@ -315,15 +322,31 @@ void ShellPlatformDelegate::OnAllFramesVisible(
 
 void ShellPlatformDelegate::OnAllFramesConcealed(
     content::WebContents* web_contents) {
-  Shell* shell = Shell::FromWebContents(web_contents);
-  if (shell) {
-    ConcealShell(shell);
+  // Step 1: Called by CobaltLifecycleManager when all renderer frames have
+  // acknowledged JS conceal deactivation. Unmap platform windows and initiate
+  // asynchronous GPU resource and EGL display teardown.
+  cobalt::CobaltLifecycleManager::GetInstance()->RemoveObserver(
```
andrewsavage1: Does this work if there are multiple WebContents, i.e. with the splash screen?

cobalt/app/app_event_runner.cc
```diff
@@ -213,6 +217,23 @@ class AppEventRunnerImpl : public AppEventRunner,
...
+    base::SequencedTaskRunner::GetCurrentDefault()->PostTask(
+        FROM_HERE, base::BindOnce([] {
+          base::MemoryPressureListener::NotifyMemoryPressure(
+              base::MemoryPressureListener::MEMORY_PRESSURE_LEVEL_CRITICAL);
+          // Chromium's memory pressure listeners are invoked asynchronously on
+          // all threads. Explicitly calling ReclaimAll here forces
+          // PartitionAlloc to synchronously purge its thread caches for the
+          // main thread right now, avoiding relying solely on the asynchronous
+          // signal propagation.
+          ::partition_alloc::MemoryReclaimer::Instance()->ReclaimAll();
```
andrewsavage1: To be defensive, maybe we should check that we're in an appropriate state (backgrounded) to reclaim memory in this callback in case it gets called late

jellefoks: These `AppEventRunner::DoFoo()` methods are private methods only called from the matching `AppEventRunner::OnFoo()`, where those checks already (pre)exist. But, it's cheap to add them, so I'll add them to all `DoFoo()` as well.

(edit: not done this way, see comment below)

third_party/blink/renderer/platform/mediastream/webrtc_uma_histograms.cc
jellefoks: Technically, no, because this is a singleton allocated from a thread, and the ExitManager runs in the main thread (UI thread), triggering this check. Maybe in upstream the media/rtc doesn't run in a thread, in which case this only shows up as an issue in single-process...

components/viz/service/gl/gpu_service_impl.cc
jellefoks: Good catch. Added a fallback to call the callbacks with null channel handles when there isn't an initialized display.

gpu/ipc/service/gles2_command_buffer_stub.cc
jellefoks: This code does not run not when the app is backgrounded. It runs when the app is made visible on a display.  This code exists to allow the offscreen surface to be recreated when the display has changed from a previous value.

In Cobalt's case, when the app is being foregrounded the previous value will be null for 'there is no default surface', because the surface is reset when backgrounded. The change here is to avoid the null pointer dereference and instead trigger the surface creation for the current display.

That said, in this code before and after this PR, the surface creation is attempted even when the display creation above was not successful. The surface creation will then likely fail, which would probably cause a failure to be returned when the context creation below fails as a result.

While unsure whether that is actually creating an undesired app behavior, defensively, added a check here to return right away here with a transient failure if there isn't an initialized display when trying to create the surface.

cobalt/shell/browser/shell_platform_delegate.cc
jellefoks: Good call. No it actually does not. And the same is true for the reveal side (this here is the conceal side transition).

Added pending_conceal_web_contents_ and pending_reveal_web_contents_  to track and ensure they are all concealed before unregistering the observer.

ShellPlatformDelegate is only an observer for CobaltLifeCycleManager while it's waiting for completion of a visibility state change (reveal or conceal), with a single registration for all windows, but it gets an OnAllFramesConcealed  or OnAllFramesVisible callback per WebContents, so it should actually track whether all expected WebContents have transitioned.

cobalt/app/app_event_runner.cc
jellefoks: Correction: You must have meant when the posted task run, not when DoFoo runs. Well, technically, since it's posted from here, and the tasks run the lifecycle changes in sequence, there won't be an already posted visibilitychange task on the same runner, so it technically can't happen right now.

Yet, adding a check is harmless, I'll add an if for production, and a DCHECK to catch it in devel builds if it ever happens.

ui/gl/gl_surface_egl.cc
jellefoks: Done

ui/gl/gl_surface_egl.cc
jellefoks: Done

components/viz/service/gl/gpu_service_impl.cc
jellefoks: Done

content/public/browser/gpu_utils.cc
jellefoks: Done

third_party/blink/renderer/platform/mediastream/webrtc_uma_histograms.cc
jellefoks: Added as TODO: b/554096932 with explanation.

jellefoks:
> Also these I think are legitimate suggestions: https://paste.googleplex.com/5541097196093440

It gets confusing quickly, but two of the four things in there were valid:

### 1. `GLContextVirtual::Destroy()`
Great catch. Adding an early return under `#if BUILDFLAG(IS_COBALT)`

### 2. Standardize EGL Display Access in `GpuServiceImpl`
Updated `GpuServiceImpl::OnBackgroundCleanupGpuMainThread` to use `gl::GetDefaultDisplayEGL()`

### 3. `ConcealShell()` Ordering & Native Window Lifecycle (not adopted)

The native `SbWindow` is destroyed when `ConcealShell()` calls `platform_->aura->host()->Hide()`, which invokes `PlatformWindowStarboard::Hide()->SbWindowDestroy()`.

Calling `ConcealShell()` in `OnAllFramesConcealed()` before `CleanupGpuProcessOnUI() `is the intended sequence:

* It unmaps the window and sets `compositor()->SetVisible(false)`, stopping the browser compositor from scheduling any further draws.

* `CleanupGpuProcessOnUI()` then runs on the GPU thread to cleanly destroy all EGLSurface handles and shut down the EGLDisplay.

* Upon reveal, `Show()` / `Restore()` calls `SbWindowCreate()` to recreate the window, and `RestoreGpuProcessOnUI()` re-initializes the `EGLDisplay`.

### 4.  weak_factory_ in `ShellPlatformDelegate` (not adopted)

Since `ConcealShell()` runs prior to `CleanupGpuProcessOnUI()`, the separate `OnGpuCleanupCompleted` helper is not needed, and completion is safely dispatched via the existing `WeakPtr<content::WebContents>` callback.

y4vor (approved):

PR #11801, opened 2026-08-11
[android] Add DisableLessAggressiveParkableString Java switch

sherryzy:
Expose DisableLessAggressiveParkableString in JavaSwitches.java to allow disabling LessAggressiveParkableString feature via Kimono/server configs, enabling foreground string compression and a 2s aging interval.

Bug: 544754813

### DIFF
--- a/third_party/blink/renderer/platform/bindings/parkable_string_manager.cc
+++ b/third_party/blink/renderer/platform/bindings/parkable_string_manager.cc
@@ -15,6 +15,7 @@
 #include "base/task/single_thread_task_runner.h"
 #include "base/time/time.h"
 #include "base/trace_event/memory_allocator_dump.h"
+#include "build/buildflag.h"
 #include "base/trace_event/process_memory_dump.h"
 #include "base/trace_event/trace_event.h"
 #include "third_party/blink/public/common/features.h"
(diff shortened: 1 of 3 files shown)


third_party/blink/renderer/platform/bindings/parkable_string_manager.cc
```diff
@@ -175,6 +176,15 @@ base::TimeDelta ParkableStringManager::AgingInterval() {
              : kAgingInterval;
 }
 
+#if BUILDFLAG(IS_COBALT)
+// static
+base::TimeDelta ParkableStringManager::FirstParkingDelay() {
+  return base::FeatureList::IsEnabled(features::kLessAggressiveParkableString)
+             ? kFirstParkingDelay
+             : base::Seconds(10);
```
hlwarriner: Please add a comment explaining why 10 is chosen as the value. Consider also refactoring the magic number as a named `constexpr static` constant, similar to `kFirstParkingDelay`.

hlwarriner (approved):
LGTM from the cobalt-web-api side, aside from the one minor comment.

johnxwork (approved):

johnxwork (approved):

third_party/blink/renderer/platform/bindings/parkable_string_manager.cc
johnxwork: This line of code is changing the behavior of your control group. Are you sure about this?

sherryzy: @johnxwork No, this does not change the control group. The control group has features::kLessAggressiveParkableString enabled, while the experiment group has this feature disabled and a shortened firstParkingDelay time.

PR #11803, opened 2026-08-11
Remove orphaned audio_dmp_player

hferreiro:
audio_dmp_player is a pre-Chromium-rebase Starboard debugging tool that plays the audio track of a .dmp dump through the filter-based audio pipeline. It doesn't build since long ago because of it's `build_loader` assignment. Remove it together with other `build_loader` uses.

Bug: 492403745


andrewsavage1 (approved):

andrewsavage1:
Looks good, possibly just running into infra flakiness

PR #11806, opened 2026-08-11
aosp: add an IS_AOSP buildflag

rogerzanoni:
Add a BUILDFLAG(IS_AOSP) to guard code specific to the AOSP platform. Unlike BUILDFLAG(IS_ANDROID), it is true in both the Evergreen and the native loader of an aosp build.

Bug: 532068409


cobalt/build/configs/initialize_variables.gni
```diff
@@ -76,6 +76,9 @@ if (starboard_toolchain == "") {
   }
 }
 
+is_aosp =
```
andrewsavage1: This isn't how is_aosp should be defined, as (i.e.) 3p partners may use a different implementation than our reference. You could follow the is_androidtv pattern, but it's probably better just to define is_aosp with `is_android && !is_androidtv`

rogerzanoni:
@andrewsavage1 I changed to set it as default to your suggestion of using `is_aosp = is_android && !is_androidtv`, but overriding it to true on `cobalt/build/configs/aosp-arm/args.gn` to set is_aosp=true for all toolchains. Other implementations can set it to true in their args.gn if IS_AOSP is needed to be true in targets where target_os = linux

cobalt/build/configs/aosp-arm/args.gn
```diff
@@ -2,5 +2,11 @@ import("//cobalt/build/configs/evergreen-x64/args.gn")
 
 target_cpu = "arm"
 starboard_target_platform = "aosp-arm"
+
+# Set for every toolchain of this build, including the Evergreen library one
```
andrewsavage1, line 6: This isn't (or shouldn't) be completely accurate, i.e. for the host toolchain we should not have `is_aosp = true`. Even for the Cobalt toolchain it's quite suspect, as we don't generally want to do things custom to the loader platform at the cobalt layer

PR #11807, opened 2026-08-11
aosp: Package CA certificates and unify the cert lookup location

rogerzanoni:
Package the CA certs under the Evergreen content dir (<content>/ssl/certs) where net::TrustStoreInMemoryStarboard looks, and make GetCACertificatesPath() fall back to the content root when the per-subdir path is absent, so cert lookup works on Android's asset layout.

Bug: 532068409

### DIFF
--- a/starboard/common/paths.cc
+++ b/starboard/common/paths.cc
@@ -74,6 +74,13 @@ std::string GetCACertificatesPath(const std::string& content_subdir) {
   std::string ca_certificates_path =
       GetCACertificatesPathFromSubdir(content_subdir);
   if (stat(ca_certificates_path.c_str(), &info) != 0) {
+    // Fall back to the content root, like the no-argument overload. On some
+    // platforms (e.g. Android, where the content directory resolves to the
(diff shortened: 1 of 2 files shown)


starboard/common/paths.cc
```diff
@@ -73,6 +73,13 @@ std::string GetCACertificatesPath(const std::string& content_subdir) {
   struct stat info;
   std::string ca_certificates_path =
       GetCACertificatesPathFromSubdir(content_subdir);
+  if (stat(ca_certificates_path.c_str(), &info) != 0) {
+    // Fall back to the content root, like the no-argument overload. On some
+    // platforms (e.g. Android, where the content directory resolves to the
+    // asset root) the certs are packaged at <content>/ssl/certs rather than
+    // <content>/<content_subdir>/ssl/certs.
```
andrewsavage1, line 80: I think the way to address this would just be to change sb_static_contents_output_data_dir, does that not work?

hferreiro:
Superseded by https://github.com/youtube/cobalt/pull/11957.

PR #11809, opened 2026-08-11
aosp: register Android media Starboard features in the Evergreen module

rogerzanoni:
Cobalt aborted in SbMediaGetVideoBufferBudget() when starting playback,
because the Android media features and params it reads were not registered.

On AOSP the default toolchain target_os = "linux", so IS_ANDROID is false.

IS_STARBOARD is true so guard the blocks on IS_ANDROID || IS_STARBOARD to
enable the same Android features for AOSP.

This also declares the features on the other Starboard platforms such as
linux, but they are used only starboard/android/shared so they stay
unused in non-android platforms.

Bug: 532068409

### DIFF
--- a/starboard/extension/feature_config.h
+++ b/starboard/extension/feature_config.h
@@ -100,7 +100,7 @@ FEATURE_LIST_START
 //   STARBOARD_FEATURE(kCobaltVideoDebug, "CobaltVideoDebug", false)
 // #endif // BUILDFLAG(IS_ANDROID) && (SB_API_VERSION >= 17)
 
-#if BUILDFLAG(IS_ANDROID) && (SB_API_VERSION >= 17)
+#if (BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_STARBOARD)) && (SB_API_VERSION >= 17)
 // keep-sorted start newline_separated=yes
 // Set to true to enable area-based video buffer budget calculation.
 STARBOARD_FEATURE(kAreaBasedVideoBufferBudget,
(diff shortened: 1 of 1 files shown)


rogerzanoni:
Depends on #11806

starboard/extension/feature_config.h
```diff
@@ -100,7 +100,7 @@ FEATURE_LIST_START
 //   STARBOARD_FEATURE(kCobaltVideoDebug, "CobaltVideoDebug", false)
 // #endif // BUILDFLAG(IS_ANDROID) && (SB_API_VERSION >= 17)
 
-#if BUILDFLAG(IS_ANDROID) && (SB_API_VERSION >= 17)
+#if (BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_AOSP)) && (SB_API_VERSION >= 17)
```
andrewsavage1: There's something wrong here—either this is being built by the cobalt toolchain and shouldn't differ based on the loader platform (i.e. should use IS_STARBOARD) or is being built with the starboard toolchain when IS_ANDROID would be true.

In general, IS_AOSP shouldn't be used often (maybe ever). It should either be IS_STARBOARD or IS_ANDROID

rogerzanoni: Changed to use BUILDFLAG(IS_ANDROID) || BUILDFLAG(IS_STARBOARD)

andrewsavage1 (approved):

andrewsavage1:
FYI @youtube/cobalt-media who maintains these.

jasonzhangxx:
This change exposes Android-specific features to all partners, which we intend to avoid. Please clarify why this is needed before submit it.

andrewsavage1:
> This change exposes Android-specific features to all partners, which we intend to avoid. Please clarify why this is needed before submit it.

Talked to Jason separately. Let's submit this as is to get AOSP working and then follow up with a change to remove the use of these feature flags in the AOSP implementation code and revert this

PR #11810, opened 2026-08-11
aosp: bring up Perfetto's backend where the media data source is registered

rogerzanoni:
Cobalt initializes the muxer in PerfettoTracedProcess::SetupClientLibrary()
but the perfetto client library is linked statically into each module and
keeps its state in static data.
So SetupClientLibrary() only initializes the copy inside libcobalt.so,
while the platform media code that reaches EnsureMediaTracingIsInitialized()
lives in the native module (the loader app, or the nplb sandbox), where the
muxer is still not initialized: until it is, it's a fake muxer that calls
FailUninitialized() on its methods, which logs and traps, crashing the
process.

This fix changes EnsureMediaTracingIsInitialized()'s call_once ensuring 
that the module that needs the muxer initializes it.

NOTE: this ensures that cobalt or nplb doesn't crash, but it leaves the native
module and libcobalt.so with separate muxers and separate tracing sessions, so
native media events do not show up in a Chromium trace.
This needs a follow-up to either create an ABI shared by
libcobalt/native code or to create a shared backend for both muxer clients

Bug: 532068409

### DIFF
--- a/starboard/shared/starboard/media/media_tracing.h
+++ b/starboard/shared/starboard/media/media_tracing.h
@@ -41,8 +41,27 @@ PERFETTO_DEFINE_CATEGORIES(perfetto::Category("starboard"));
 // I would limit the scope to media for now, and expand its scope when needed.
 inline void EnsureMediaTracingIsInitialized() {
   static std::once_flag s_once_flag;
-  std::call_once(s_once_flag,
-                 []() { starboard_tracing::TrackEvent::Register(); });
+  std::call_once(s_once_flag, []() {
+    // Ensure that the module that registers the data source initializes the
+    // client library. Cobalt initializes it in
+    // PerfettoTracedProcess::SetupClientLibrary(), but this doesn't help the
+    // evergreen native loader or the nplb sandbox. For those targets the
+    // client library wasn't initialized and until
+    // perfetto::Tracing::Initialize() is called, TracingMuxer is still a
+    // TracingMuxerFake; TracingMuxerFake calls FailUninitialized() on its
+    // methods, which logs and traps, crashing nplb/the loader.
+    // TODO(b/532068409): This fixes this crash but for evergreen two
+    // muxers/tracing sessions will be active, meaning that native media events
+    // won't show up in chromium traces. This needs a follow-up for the
+    // underlying issue: the perfetto client is statically linked for each
+    // target that uses it and the initialization state isn't shared.
+    if (!perfetto::Tracing::IsInitialized()) {
+      perfetto::TracingInitArgs args;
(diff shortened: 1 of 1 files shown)


starboard/shared/starboard/media/media_tracing.h
```diff
@@ -45,6 +45,21 @@ inline void EnsureMediaTracingIsInitialized() {
...
 
+// Brings up Perfetto's in-process tracing backend if the muxer isn't already
+// initialized. In a full Cobalt process this is done by
+// PerfettoTracedProcess::SetupClientLibrary(), but on Evergreen that only
+// initializes the muxer in the cobalt library; 
+// For the native loader, which has its own separate muxer it will fail on
+// EnsureMediaTracingIsInitialized() withj TracingMuxerFake::FailUninitialized() and
+// crashes the process.
+inline void InitializeMediaTracingBackend() {
+  if (!perfetto::Tracing::IsInitialized()) {
```
andrewsavage1: Is this using functions from copied_base? In general that's something we should avoid

starboard/shared/starboard/media/media_tracing.h
```diff
@@ -45,6 +45,21 @@ inline void EnsureMediaTracingIsInitialized() {
                  []() { starboard_tracing::TrackEvent::Register(); });
 }
 
+// Brings up Perfetto's in-process tracing backend if the muxer isn't already
+// initialized. In a full Cobalt process this is done by
+// PerfettoTracedProcess::SetupClientLibrary(), but on Evergreen that only
+// initializes the muxer in the cobalt library; 
+// For the native loader, which has its own separate muxer it will fail on
+// EnsureMediaTracingIsInitialized() withj TracingMuxerFake::FailUninitialized() and
+// crashes the process.
```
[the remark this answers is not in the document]
rogerzanoni: Outdated

starboard/shared/starboard/media/media_tracing.h
rogerzanoni: It's using functions from third_party/perfetto

I updated the PR with a different fix

andrewsavage1:
Is this an issue for all Evergreen platforms or just AOSP? Either way, why do we only see different behavior on AOSP?

andrewsavage1 (commented):
I think the correct way to solve this is to set `enable_base_tracing = false` in the aosp toolchain definitions for starboard/native_target in `starboard/aosp/arm/toolchain/BUILD.gn`

rogerzanoni:
Abandoned in favor of #12327, following @andrewsavage1 suggestion of disabling base tracing.

PR #11813, opened 2026-08-11
aosp: add ApplicationAOSP to handle input and native window creation

rogerzanoni:
This PR includes commits for:
- Creating an application (ApplicationAOSP) based on QueueApplication to forward input events from JNI -> starboard handler
- Creates the native window/surface for rendering Cobalt UI
- Creates a VideoSurface for video playback

Bug: 532068409


andrewsavage1:
Adding @zhongqiliang and @johnxwork to help review android-specific bits

starboard/android/shared/BUILD.gn
```diff
@@ -266,12 +266,6 @@ static_library("starboard_platform") {
     "video_surface_view.h",
     "video_window.cc",
     "video_window.h",
-
-    #"window_create.cc",
```
andrewsavage1, line 269: Should these just be removed?

starboard/android/shared/BUILD.gn
```diff
@@ -341,11 +335,8 @@ static_library("starboard_platform") {
       "//starboard/shared/stub/speech_synthesis_speak.cc",
       "//starboard/shared/stub/system_network_is_disconnected.cc",
       "//starboard/shared/stub/system_raise_platform_error.cc",
-      "//starboard/shared/stub/window_create.cc",
-      "//starboard/shared/stub/window_destroy.cc",
+
```
andrewsavage1: remove newline here

starboard/aosp/shared/window.cc
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+#include "starboard/window.h"
+
+#include <android/native_window.h>
+
+#include "starboard/aosp/shared/application_aosp.h"
+#include "starboard/aosp/shared/window_internal.h"
+
+SbWindow SbWindowCreate(const SbWindowOptions* options) {
```
andrewsavage1: We should follow the pattern of having a new file for each SbWindow implementation.

cobalt/aosp/apk/app/java/dev/cobalt/app/MainActivity.java
```diff
@@ -58,23 +65,88 @@ public class MainActivity extends BaseCobaltActivity {
...
+
+              @Override
+              public void surfaceDestroyed(SurfaceHolder holder) {
+                MainActivityJni.get().nativeOnSurfaceDestroyed();
+              }
+            });
+
+    // Video surface at the bottom, webui surface on top.
+    FrameLayout container = new FrameLayout(this);
+    container.addView(
```
johnxwork: We are moving away from having two SurfaceViews as canvas, and moving towards drawing directly on the window. Can AOSP be aligned here?

johnxwork: You can reference https://github.com/youtube/cobalt/pull/11538

starboard/aosp/shared/main_activity.cc
```diff
@@ -94,4 +98,29 @@ void JNI_MainActivity_StartLoader(JNIEnv* env) {
...
+void JNI_MainActivity_NativeOnSurfaceDestroyed(JNIEnv*) {
+  SB_LOG(INFO) << "cobalt_loader: Starboard surface destroyed.";
+  starboard::android::shared::SetWindowSurface(nullptr);
+}
+
+jboolean JNI_MainActivity_NativeSendKeyEvent(JNIEnv* /*env*/,
+                                             jint key_code,
+                                             jint action,
+                                             jint unicode_char) {
+  return starboard::ApplicationAOSP::Get()->InjectKeyEvent(
```
andrewsavage1: Is there a possible null dereference here if the application isn't yet set up?

starboard/aosp/shared/application_aosp.cc
andrewsavage1, line 1: We could just worry about it later, but I'm surprised to see no lifecycle management being done here, meaning it would just be handled natively via QueueApplicatoin

starboard/aosp/shared/application_aosp.cc
```diff
@@ -0,0 +1,104 @@
...
+#include "starboard/event.h"
+#include "starboard/input.h"
+#include "starboard/key.h"
+#include "starboard/window.h"
+
+namespace starboard {
+
+namespace {
+
+SbKey MapAndroidKeyCodeToSbKey(int key_code) {
```
andrewsavage1: The implementation of all these functions looks much simpler than Cobalt 25, which might be okay, but I'm wary of missing information (for example, key modifiers are just set to 0 in the InjectKeyEvent implementation while in C25 it looks like it would actually try to calculate that). I don't mind doing that as a first pass and coming back to it, but is this something that is intentional or not? In general, we want to pretty closely follow the Starboard implementation of AOSP from Cobalt 25, as we know that worked.

starboard/aosp/shared/application_aosp.cc
```diff
@@ -0,0 +1,104 @@
...
+    case AKEYCODE_NUMPAD_ENTER:
+      return kSbKeyReturn;
+    case AKEYCODE_ESCAPE:
+    case AKEYCODE_BACK:
+      // "back" maps to Escape (see starboard/key.h)
+      return kSbKeyEscape;
+    case AKEYCODE_MEDIA_PLAY_PAUSE:
+    case AKEYCODE_MEDIA_PLAY:
+    case AKEYCODE_MEDIA_PAUSE:
+      return kSbKeyMediaPlayPause;
```
andrewsavage1: Reference starboard/android/shared/input_events_generator.cc from C25 for these mappings. This for example looks wrong, i.e. AKEYCODE_MEDIA_PLAY should likely map to kSbKeyPlay

starboard/aosp/shared/application_aosp.cc
```diff
@@ -0,0 +1,104 @@
...
+
+bool ApplicationAOSP::InjectKeyEvent(int key_code,
+                                     int action,
+                                     int unicode_char) {
+  SbKey sb_key = MapAndroidKeyCodeToSbKey(key_code);
+  if (sb_key == kSbKeyUnknown) {
+    return false;
+  }
+
+  SbInputData* data = new SbInputData();
```
andrewsavage1, line 90: We would generally want data->window to be valid, but it's not set here

starboard/aosp/shared/window.cc
```diff
@@ -0,0 +1,47 @@
...
+  return window->native_window;
+}
+
+bool SbWindowGetSize(SbWindow window, SbWindowSize* size) {
+  if (!SbWindowIsValid(window) || window->native_window == nullptr) {
+    return false;
+  }
+  size->width = ANativeWindow_getWidth(window->native_window);
+  size->height = ANativeWindow_getHeight(window->native_window);
+  size->video_pixel_ratio = 1.0f;
```
andrewsavage1: This is set differently in C25, see starboard/android/shared/window_get_size.cc

ui/events/keycodes/keyboard_code_conversion_starboard.cc
```diff
@@ -1352,18 +1352,26 @@ KeyboardCode SbKeyToKeyboardCode(SbKey sb_key) {
     case kSbKeyGamepadDPadRight:
       return VKEY_RIGHT;
 
-    // The following keys don't have standard KeyboardCode values.
     case kSbKeyRed:
```
andrewsavage1, line 1356: Is adding these in this PR necessary for the functionality of Cobalt on AOSP? If not, can we forgo it for now?

starboard/aosp/shared/application_aosp.cc
```diff
@@ -0,0 +1,116 @@
...
+  if (meta_state & AMETA_SHIFT_ON) {
+    modifiers |= kSbKeyModifiersShift;
+  }
+  return modifiers;
+}
+
+}  // namespace
+
+void ApplicationAOSP::Initialize() {
+  features::FeatureList::RegisterFeaturesInitializedCallback(
```
andrewsavage1: Don't do it this way—just call InitializeAudioSink (or SbAudioSinkImpl::Initialize) from here and change the implementation to not crash if the feature list isn't enabled.

```
  bool release_frames_after_audio_starts =
      features::FeatureList::GetInstance()->IsInitialized() &&
      features::FeatureList::IsEnabled(
          features::kReleaseVideoFramesAfterAudioStarts);
```

Then revert the changes to feature_list*

ui/events/keycodes/keyboard_code_conversion_starboard.cc
rogerzanoni: We can leave it out for now, I will keep the commit for a follow-up if needed.

starboard/aosp/shared/BUILD.gn
```diff
@@ -17,10 +17,33 @@ import("//starboard/build/config/toolchain_variables.gni")
...
+    "window_surface.h",
+  ]
 
   configs += [ "//starboard/build/config:starboard_implementation" ]
 
+  deps = [
+    # window_get_size.cc includes starboard_bridge.h, which pulls in
+    # base/memory/singleton.h and jni_zero. //base is a private dep of
+    # //starboard/android/shared:starboard_platform, so it isn't propagated.
+    "//base",
```
andrewsavage1: Don't depend on `//base` here, as we're in the starboard layer. Add `// nogncheck` to the offending includes if necessary

starboard/aosp/shared/application_aosp.h
```diff
@@ -0,0 +1,101 @@
...
+                      int action,
+                      int unicode_char,
+                      int meta_state);
+
+ protected:
+  // Arranges for the platform audio sink to be created by setting a callback
+  // that runs when features are initialized.
+  void Initialize() override;
+
+  bool MayHaveSystemEvents() override { return false; }
```
andrewsavage1, line 89: Will this and WaitForSystemEventWithTimeout need a follow up for a real implementation?

starboard/aosp/shared/application_aosp.cc
rogerzanoni: I did this change locally and I have a few concerns:

Changed to check the init state like this,, but GetInstance() is a private member in starboard/shared/starboard/feature_list.h, so I added an IsInitialized() acessor to FeatureList, guarded by a mutex and that doesn't trigger an abort like the other acessors. Not using the callbacks solution also revealed two other places where features are queried during cobalt init and trigger aborts:

- AudioThreadFunc() in starboard/android/shared/audio_track_audio_sink_type.cc queries for kReleaseVideoFramesAfterAudioStarts
- GetDefaultStackSize() in starboard/common/thread.cc queries for kReduceAndroidThreadStackSize

This is dangerous, specially for the starboard thread part that queries kReduceAndroidThreadStackSize, it can get the wrong value if threads are started before the features are initialized.

So, I decided to drop it and try to find a place where we can intilialize the sink in AOSP after the features list is pushed.

AndroidTV is handling it by using the CobaltMainDelegate::PostEarlyInitialization() to push the feature list early; The best place I could find was right at the start of ApplicationAOSP::CreateWindow, because Ozone creates the platform window on PlatformWindowStarboard::Show and by that time CobaltContentBrowserClient::CreateFeatureListAndFieldTrials() already ran;

So in ApplicationAOSP we could:
```
SbWindow ApplicationAOSP::CreateWindow(const SbWindowOptions* /*options*/) {
  /* init the sink here */
```

That would cover sync init for the Cobalt app, and we would neeed to keep the sink init in ApplicationAOSP::Initialize to cover initializing it for nplb.

Fot this to work we would still need the change to add the IsInitialized acessor for FeatureList to prevent trying to initialize it if the features didn't get pushed yet. wdyt?

note: For cobalt init, I thought about a problem in the case of the sink being used before window creation, but I don't know if this is possible/an use case

starboard/aosp/shared/application_aosp.h
rogerzanoni: No, they are pure virtual, added only as a requirement to compile (like the tvos impl) , except for MayHaveSystemEvents() should return false because AOSP doesn't  run its own event loop like rdk/x11 implementations. I will add a comment explaining that

starboard/aosp/shared/BUILD.gn
rogerzanoni: The dep needed is actually //third_party/abseil-cpp:absl, changed to add it

starboard/aosp/shared/window.cc
rogerzanoni: Fixed

starboard/aosp/shared/application_aosp.cc
rogerzanoni: Fixed

starboard/aosp/shared/application_aosp.cc
rogerzanoni: Fixed, brought most of the C25 implementation

starboard/aosp/shared/application_aosp.cc
rogerzanoni: There's a change in progress, I will create a PR soon

starboard/aosp/shared/main_activity.cc
rogerzanoni: Created a separate GetIfExists() that doesn't abort if the application isn't created yet or it's already gone, and Get is used only in code that runs inside SbRunStarboardMain

cobalt/aosp/apk/app/java/dev/cobalt/app/MainActivity.java
rogerzanoni: Fixed

starboard/aosp/shared/window.cc
rogerzanoni: Fixed

starboard/android/shared/BUILD.gn
rogerzanoni: Done

starboard/android/shared/BUILD.gn
rogerzanoni: Changed to not touch androidtv stubs.

starboard/aosp/shared/application_aosp.cc
```diff
@@ -0,0 +1,104 @@
...
+  // that isn't an explicit ACTION_UP as a press.
+  data->type = action != AKEY_EVENT_ACTION_UP ? kSbInputEventTypePress
+                                              : kSbInputEventTypeUnpress;
+  data->key = sb_key;
+  // TODO(532068409) map modifiers
+  data->key_modifiers = 0;
+  Inject(new Event(kSbEventTypeInput, data,
+                   &ApplicationAOSP::DeleteDestructor<SbInputData>));
+  return true;
+}
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

starboard/aosp/shared/application_aosp.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <cstdint>
+
+#include "starboard/shared/starboard/queue_application.h"
+#include "starboard/window.h"
+
+namespace starboard {
+
+// Minimal QueueApplication for AOSP
+class ApplicationAOSP : public QueueApplication {
```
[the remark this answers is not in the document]
rogerzanoni: Fixed

andrewsavage1 (approved):

starboard/aosp/shared/window.cc
```diff
@@ -0,0 +1,47 @@
...
+  return true;
+}
```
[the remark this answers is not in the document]
rogerzanoni: Fixed differently, by ownership, the window reference is now refcounted using ANativeWindow_acquire/release so it doesn't dangle

PR #11835, opened 2026-08-12
Support time-windowed peak private memory footprint guardrails in DevTools Performance domain

haozheng-cobalt:
Support time-windowed peak private memory footprint guardrails in the
DevTools Performance domain. This change allows developers to see
the peak memory usage metrics that Cobalt uses for production
guardrails across different session lifecycle windows.

The implementation includes logic to safely extract peak values from
UMA histograms, specifically handling overflow buckets to ensure
reported values remain accurate and do not overflow into petabyte
ranges when limits are exceeded.

Test: output of running `cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py` to check memory metrics added to CDP: https://paste.googleplex.com/6028289796866048
Bug: 530302534


cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
```diff
@@ -108,15 +107,11 @@ def get_websocket_url(platform: str, port: int, quiet: bool):
...
-        devs = list(device_utils.DeviceUtils.HealthyDevices())
-        if not devs:
-          _print_q('Error: No healthy Android devices found.', quiet)
-          return None
-        dev = devs[0]
-        dev.adb.Forward(
-            f'tcp:{port}',
-            'localabstract:content_shell_devtools_remote',
-            allow_rebind=True)
+        run_adb_command([
```
andrewsavage1: Why is this change necessary? It seems unrelated and I don't see it called out in the PR description. It also seems better to use a structured python class when possible over an adb subprocess command

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.h
andrewsavage1, line 1: Not necessarily in this PR, but these files should be put under a cobalt/ subdirectory to make the directory structure more clear about what is owned by us vs. chromium

cobalt/tools/uma/pull_uma_histogram_set_via_cdp.py
haozheng-cobalt: sorry, added accidentally.

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.h
haozheng-cobalt: I had this discussion with Adam in a previous PR, and we decided to add the cobalt_memory_metrics_helper.h to //third_party to avoid the layer violation of making //third_party/blink/renderer/core depend on //cobalt(if we were to add cobalt_memory_metrics_helper.h under //cobalt)

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.cc
```diff
@@ -165,4 +189,63 @@ std::optional<std::vector<MemoryBreakdownMetric>> GetLiveMemoryBreakdown() {
...
+    int64_t max;
+    base::HistogramBase::Count32 count;
+    it->Get(&min, &max, &count);
+    uint64_t bucket_value = GetSafeBucketValue(min, max);
+    if (count > 0 && bucket_value > peak_value) {
+      peak_value = bucket_value;
+    }
+  }
+
+  if (IsMiBHistogram(metric_name)) {
```
Awallky, line 223: Shouldn't this histogram be added to IsMiBHistogram?

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.h
andrewsavage1: To clarify, I'm just saying that there should be a cobalt/ subdirectory in `//third_party/blink/renderer/core` (or use an existing cobalt/ folder under blink)

andrewsavage1 (approved):

Awallky (approved):

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.cc
```diff
@@ -44,9 +46,27 @@ bool IsMiBHistogram(std::string_view metric_name) {
       return true;
     }
   }
+  for (const char* guardrail_name : kPeakMemoryGuardrailMetricNames) {
```
haozheng-cobalt, line 49: link

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.cc
haozheng-cobalt: Yes, all the guardrail metrics return true on the function IsMiBHistogram. https://github.com/youtube/cobalt/pull/11835/changes/BASE..b08a61193ea934035c21c23d6c2d6323a380d08a#r3776833932

third_party/blink/renderer/core/inspector/cobalt_memory_metrics_helper.h
haozheng-cobalt: Oh kk, I will see if there is better convention than prefixing the files with cobalt, and improve the structure in the future.

PR #11843, opened 2026-08-12
[M139] media: Adapt Android audio capture fast-path to upstream changes

sideb0ard:
Upstream CL 6586528 refactored device enumeration in AudioManagerAndroid by introducing JniDelegate and GetCommunicationDeviceNames().

This change adapts Cobalt's audio capture fast-path to those changes:
- Guard Java_AudioManagerAndroid_init with BUILDFLAG(USE_STARBOARD_MEDIA) in JniDelegateImpl to avoid registering unnecessary audio listeners.
- Short-circuit GetCommunicationDeviceNames() when BUILDFLAG(USE_STARBOARD_MEDIA) is enabled so it returns only the default device without querying JNI.

Bug: 543784629
TAG=agy
CONV=3bc3dc39-bbcb-489f-b335-3b032071e01a


media/audio/android/audio_manager_android.cc
```diff
@@ -549,6 +551,10 @@ void AudioManagerAndroid::GetCommunicationDeviceNames(
   DCHECK(device_names->empty());
   AddDefaultDevice(device_names);
 
+#if BUILDFLAG(USE_STARBOARD_MEDIA)
+  // simplfified flow - just return, device_names is set to default.
```
rakuco, line 555:
```suggestion
  // simplified flow - just return, device_names is set to default.
```

rakuco (approved):
rubber-stamp lgtm since you know this code way better than us working on the rebase do

just one thing: #11745 will land as soon as CI stops being flaky, if it does before this PR lands you'll need to target the staging branch instead.

sideb0ard:
Will close this draft PR - i have the same changes on a proper staging branch at https://github.com/youtube/cobalt/pull/12073

PR #11849, opened 2026-08-12
android: Disable PiP support for 3P builds

GillianZili:
Modify build configurations and feature flags to disable
Picture-in-Picture (PiP) support for 3P partner (AOSP) builds.
PiP-related dependencies, source files, and JNI headers are now
conditionally included only for 1P Android builds. This change
resolves AOSP check errors on partner builds.

Issue: 445194225

### DIFF
--- a/cobalt/android/BUILD.gn
+++ b/cobalt/android/BUILD.gn
@@ -79,8 +79,6 @@ android_library("cobalt_main_java") {
     "//cobalt/shell/android:cobalt_shell_java_resources",
     "//cobalt/shell/android:cobalt_shell_jni_headers",
     "//components/embedder_support/android:view_java",
-    "//components/thin_webview:factory_java",
-    "//components/thin_webview:java",
     "//content/public/android:content_java",
     "//media/capture/video/android:capture_java",
     "//net/android:net_java",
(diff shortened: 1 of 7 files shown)


cobalt/android/BUILD.gn
```diff
@@ -234,9 +232,11 @@ android_library("cobalt_apk_java") {
 
   deps = [
     ":cobalt_main_java",
-    ":cobalt_pip_java",
     "//base:base_java",
   ]
+  if (is_android && !is_androidtv) {
```
andrewsavage1, line 237: I think here and below you have this condition backwards. You probably just want to check `if (is_androidtv)` to enable PiP for only ATV (and disable it for 3P)

starboard/build/config/starboard_executable.gni
```diff
@@ -33,6 +33,7 @@ template("starboard_executable") {
         "//cobalt/aosp:cobalt_apk_java",
       ]
       remove_uncalled_jni = true
+      add_stubs_for_missing_jni = true
```
andrewsavage1, line 36: I *think* if you correct the condition above the changes here and below are unnecessary

borongc:
This is handled by https://github.com/youtube/cobalt/pull/11868.

PR #11876, opened 2026-08-13
android: Handle video size cap from `max_video_capabilities`

osagie98:
Adds a field `max_video_size_` to `MediaCodecVideoDecoder ` to cache the max
supported video size set from `max_video_capabilities`. If a video key frame
exceeds `max_video_size_` during playback, `MediaCodecVideoDecoder ` raises
`kSbPlayerErrorCapabilityChanged`. If, before playback begins, the requested
resolution exceeds `max_video_size_`, `max_video_size_` is updated to match
that resolution and no error is raised.

#vibe-coded

Issue: 546122686


starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -859,6 +810,20 @@ Result<void> MediaCodecVideoDecoder::InitializeCodec(
   std::optional<Size> max_frame_size =
```
jasonzhangxx: Let's parse `max_video_capabilities_` once in ctor and store the parsed result `max_video_size_`.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -859,6 +810,20 @@ Result<void> MediaCodecVideoDecoder::InitializeCodec(
...
       ParseMaxResolution(max_video_capabilities_, video_stream_info.frame_size);
 
+  // b/546122686: Raise kSbPlayerErrorCapabilityChanged if the stream resolution
+  // exceeds max_video_capabilities.
+  if (max_frame_size.has_value() &&
+      IsFrameSizeExceedingCapabilities(video_stream_info.frame_size,
```
jasonzhangxx, line 887: We should update InitializeCodec to make it create codec based on `max_video_size_`.

Also, in `WriteInputBuffers`, if an input buffer is key frame, we should check its size, and raising `kSbPlayerErrorCapabilityChanged` error if it exceed `max_video_size_`. 

We can discuss with Drew on if `video_stream_info` on player creation can exceed `max_video_size_` and what we should do if that happens. We can return error immediately or we can set `max_video_size_` to the size in `video_stream_info`.

osagie98: Chatting with Drew offline. The stream info can exceed the max size on creation, so I've updated it to handle that. However, it also seems like we can expect the size of written frames to change mid-stream; I'd need to remove the CHECK in writeinputbuffers in that case

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -528,6 +548,28 @@ void MediaCodecVideoDecoder::WriteInputBuffers(
...
+        if (IsFrameSizeExceedingCapabilities(frame_size,
+                                             max_video_size_.value())) {
+          SB_LOG(ERROR) << "Video frame size " << frame_size
+                        << " exceeds max_video_capabilities "
+                        << max_video_size_.value()
+                        << ". Raising kSbPlayerErrorCapabilityChanged.";
+          ReportError(kSbPlayerErrorCapabilityChanged,
+                      "Video frame size exceeds max_video_capabilities.");
+          return;
+        }
```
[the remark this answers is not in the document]
osagie98: I'll likely remove this check so we can handle gracefully in non production builds

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -195,6 +195,14 @@ const DrmSystem::Callbacks kStubDrmSystemCallbacks = {
     StubDrmSessionUpdateRequestFunc, StubDrmSessionUpdatedFunc,
     StubDrmSessionKeyStatusesChangedFunc};
 
+bool IsFrameSizeExceedingCapabilities(const Size& frame_size,
+                                      const Size& max_video_size) {
+  return std::max(frame_size.width, frame_size.height) >
+             std::max(max_video_size.width, max_video_size.height) ||
+         std::min(frame_size.width, frame_size.height) >
+             std::min(max_video_size.width, max_video_size.height);
```
jasonzhangxx: @osagie98, could you help me understand why we want to compare video size in this way?

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -436,8 +446,12 @@ MediaCodecVideoDecoder::MediaCodecVideoDecoder(
                << ", with output mode=" << GetPlayerOutputModeName(output_mode_)
                << ", preroll count=" << number_of_preroll_frames_
                << ", max pending input size=" << kMaxPendingInputsSize
-               << ", max video capabilities=\"" << max_video_capabilities_
-               << "\", tunnel mode audio session id="
+               << ", max video size="
```
jasonzhangxx: Let's keep the log for the raw `max_video_capabilities` string. Additionally, could you confirm if the Size struct supports the << operator for logging?

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -528,6 +542,28 @@ void MediaCodecVideoDecoder::WriteInputBuffers(
                     input_buffers.front()->timestamp(), "size",
                     input_buffers.size());
 
+  if (max_video_size_.has_value()) {
+    for (const auto& input_buffer : input_buffers) {
+      if (input_buffer->video_sample_info().is_key_frame) {
+        const Size& frame_size = input_buffer->video_stream_info().frame_size;
+        SB_CHECK(!IsFrameSizeExceedingCapabilities(frame_size,
```
jasonzhangxx: Remove this  `SB_CHECK`.

jasonzhangxx (commented):
Please also make sure software decoder is not used. `MediaCodecVideoDecoder` has  `require_software_codec_`, which is derived from `max_video_capabilities`. `require_software_codec_` was for 10x player on Android and it's currently not used.

starboard/android/shared/media_codec_video_decoder.cc
osagie98: This is for vertical video, since comparing width and height directly may cause false negatives. It was caught in an earlier Gemini comment, if we cap to 1920x1080, a 1080x1920 stream would be rejected since the 1920 height exceeds the 1080 cap. So instead it compares the big and small numbers (1920 vs 1920 and 1080 vs 1080)

starboard/android/shared/media_codec_video_decoder.cc
osagie98: Replaced with `max_video_capabilities` string. `Size` has an [overloaded](https://github.com/youtube/cobalt/blob/cf0999fb4ffbc50332a579d0743f01f007f473c9/starboard/common/size.h#L45) <<, but I removed the log since it'll already be logged in the capabilities string.

starboard/android/shared/media_codec_video_decoder.cc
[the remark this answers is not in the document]
osagie98: removed

starboard/android/shared/media_codec_video_decoder.cc
jasonzhangxx: Could you verify with YT videos if the width and height could be swapped?

xiaomings: We shouldn't assume that the MediaCodec can handle a 1920x1080 video when it's configured to 1080x1920.

osagie98: I removed this check, so now it just checks if width or height are exceeded. If a vertical video exceeds either of these, the function returns false.

jasonzhangxx (approved):

PR #11894, opened 2026-08-13
cobalt: Add memory ablation study

yjzhang111:
Introduce a mechanism to perform native memory ablation studies via
Finch. This allows Cobalt to allocate and dirty a configurable amount
of memory at startup to facilitate the verification of memory metrics
and the impact of memory pressure on performance in experiment arms.

The allocation is held for the duration of the process. This update
also includes telemetry for the ablation state

Bug: 543418301

### DIFF
--- a/cobalt/browser/cobalt_browser_main_parts.cc
+++ b/cobalt/browser/cobalt_browser_main_parts.cc
@@ -28,6 +28,7 @@
 #include "build/buildflag.h"
 #include "cobalt/browser/global_features.h"
 #include "cobalt/browser/h5vcc_native_stability/native_stability_manager.h"
+#include "cobalt/browser/memory_ablation.h"
 #include "cobalt/browser/metrics/cobalt_detailed_metrics_delegate.h"
 #include "cobalt/browser/metrics/cobalt_metrics_service_client.h"
 #include "cobalt/browser/switches.h"
@@ -229,6 +230,8 @@ int CobaltBrowserMainParts::PreMainMessageLoopRun() {
 #endif
 
   cobalt::memory::CobaltMemoryAttributionManager::Get()->Start();
+
+  MaybeApplyMemoryAblation();
 
 #if !BUILDFLAG(IS_ANDROIDTV)
   auto* client = CobaltContentBrowserClient::Get();
--- a/cobalt/browser/memory_ablation.cc
+++ b/cobalt/browser/memory_ablation.cc
@@ -15,12 +15,16 @@
 #include "cobalt/browser/memory_ablation.h"
 
 #include <memory>
+#include <new>
 #include <vector>
 
 #include "base/feature_list.h"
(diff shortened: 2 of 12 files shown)


cobalt/browser/memory_ablation.cc
```diff
@@ -0,0 +1,123 @@
...
+
+  LOG(WARNING) << "Dispatching native memory ablation: allocating and dirtying "
+               << size_mb << " MB of RAM in background.";
+
+  scoped_refptr<base::SequencedTaskRunner> reply_runner =
+      base::SequencedTaskRunner::HasCurrentDefault()
+          ? base::SequencedTaskRunner::GetCurrentDefault()
+          : nullptr;
+
+  base::ThreadPool::PostTask(
```
vincentmo: Can you change this so it strictly executes at most once per app life time? i.e.

static std::atomic_bool was_applied{false};
if (was_applied.exchange(true)) {
  return;
}

cobalt/browser/memory_ablation.cc
```diff
@@ -0,0 +1,123 @@
...
+void DoMemoryAblationInBackground(
+    int size_mb,
+    scoped_refptr<base::SequencedTaskRunner> reply_runner) {
+  const size_t bytes_to_allocate = static_cast<size_t>(size_mb) * 1024 * 1024;
+  constexpr size_t kPageSize = 4096;
+
+  auto buffer =
+      std::unique_ptr<char[]>(new (std::nothrow) char[bytes_to_allocate]);
+  if (!buffer) {
+    LOG(ERROR) << "Failed to allocate " << size_mb
```
vincentmo, line 61: Could you add a success/failure enum histogram so we know how often memory allocation failures happen?

vincentmo (approved):
LGTM w/ some minor change suggestions

cobalt/browser/cobalt_browser_main_parts.cc
```diff
@@ -230,6 +231,8 @@ int CobaltBrowserMainParts::PreMainMessageLoopRun() {
 
   cobalt::memory::CobaltMemoryAttributionManager::Get()->Start();
 
+  MaybeApplyMemoryAblation();
```
Awallky, line 234: Doing this here means we will always perform memory ablations very early in the app lifecycle. Is this intended? Or do we want the option to defer this to some point in the future? We could add a configurable delay to this, making it so we can vary when we activate the ablation study.

vincentmo: Adding configurability for a delay is a good idea, thanks for suggesting it.

Awallky (approved):
I think if we add a PostDelayedTask and a Finch param for this, everything else LGTM.
See https://github.com/youtube/cobalt/pull/11894/changes#r3807008587

PR #11902, opened 2026-08-14
Cobalt: Revert disable v8 optimizing compilers

tasunnn:
Revert both #11156 and #11186

Enabling --disable-optimizing-compilers and --no-sparkplug 
introduces significant regression in Kabuki latency despite 
memory savings. 

Bug: 512163617

### DIFF
--- a/cobalt/build/configs/common.gn
+++ b/cobalt/build/configs/common.gn
@@ -50,7 +50,6 @@ v8_enable_atomic_object_field_writes = true
 # This will force V8 to embed snapshot blob in binary.
 v8_use_external_startup_data = false
 
-
 # Disable hls_demuxer so that enable_mse_mpeg2ts_stream_parser=false
 enable_hls_demuxer = false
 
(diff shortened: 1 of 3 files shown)


y4vor (approved):

cobalt/build/configs/common.gn
```diff
@@ -50,9 +50,6 @@ v8_enable_atomic_object_field_writes = true
 # This will force V8 to embed snapshot blob in binary.
 v8_use_external_startup_data = false
 
-# Disable Maglev in binary.
-v8_enable_maglev = false
```
Awallky, line 54: Is this related to the v8 optimizing compiler?

Awallky (approved):
LGTM % comment

cobalt/build/configs/common.gn
tasunnn: yes, --disable-optimizing-compilers also disables maglev, we added this flag first then also removed it from binary

PR #11915, opened 2026-08-14
ci: Fix unbound variable and parsing in gemini-commit-message workflow

oxve:
Move the Gemini commit message prompt from the workflow file to `.agent/commit_message_prompt.md` and check it out using sparse checkout.

Harden the workflow security and robustness:
- Pass `GEMINI_SECRET` securely via the `x-goog-api-key` HTTP header rather than the URL query parameter.
- Use `set -euo pipefail` without `-x` in steps handling secrets.
- Add explicit top-level and job permissions (`contents: read`).
- Use a random delimiter token for multiline `GITHUB_OUTPUT` writing.
- Handle unbound `COMMENT_BODY` when unset on `pull_request_target`.
- Wrap all dynamic prompt fields in inline XML delimiter tags on the `--arg` parameters for prompt injection protection.
- Handle large PR diffs gracefully by omitting diffs exceeding 64 KB.
- Clarify `Bug: None` fallback in prompt rules.

Bug: 436660675
Tag: agy
Conv: 048f85bc-2a59-46b4-8b1e-efcf10946b49

### DIFF
--- a/.agent/commit_message_prompt.md
+++ b/.agent/commit_message_prompt.md
@@ -1,0 +1,53 @@
+Act as an expert software engineer specializing in the Cobalt codebase, which
+is a fork of Chromium. Your task is to generate a professional and
+informative Git commit message based on the provided pull request details.
+The final output should be only the commit message itself, without any extra
+conversational text or markdown formatting. Do not use backticks (`) in your
+response.
+
+You must strictly adhere to the following rules:
+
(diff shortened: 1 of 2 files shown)


oxve:
Closing as gemini-commit-message is being fixed separately.

briantting (approved):

PR #11942, opened 2026-08-14
Auto-reconnect Mojo pipes for PlatformServices at resume

johnxwork:
When a device freezes/suspends (e.g. system freeze on Android TV), the Mojo IPC pipes for H5vccPlatformService can be disconnected. Previously, OnServiceConnectionError() would permanently close the service, clear the receive callback, and cause subsequent send() calls from JS to throw kInvalidStateError.

This change:
1. Updates H5vccPlatformService to inherit from ExecutionContextLifecycleStateObserver and listen for FrameLifecycleState::kRunning.
2. Proactively re-establishes the Mojo connection upon unfreeze/resume.
3. Adds on-demand EnsureConnected() in send() so unexpected disconnections are recovered seamlessly.
4. Distinguishes between explicit JS close() calls and Mojo disconnections, preserving the ReceiveCallback across freeze/resume cycles.

### DIFF
--- a/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
+++ b/third_party/blink/renderer/modules/cobalt/h5vcc_platform_service/h_5_vcc_platform_service.cc
@@ -18,6 +18,7 @@
 
 #include "base/functional/bind.h"
 #include "base/logging.h"
+#include "third_party/blink/public/mojom/frame/lifecycle.mojom-blink.h"
 #include "third_party/blink/public/platform/platform.h"
 #include "third_party/blink/renderer/bindings/core/v8/script_function.h"
 #include "third_party/blink/renderer/bindings/core/v8/v8_binding_for_core.h"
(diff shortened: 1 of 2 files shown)


hlwarriner:
> When a device freezes/suspends (e.g. system freeze on Android TV), the Mojo IPC pipes for H5vccPlatformService can be disconnected. Previously, OnServiceConnectionError() would permanently close the service, clear the receive callback, and cause subsequent send() calls from JS to throw kInvalidStateError.
> 
> This change:
> 
> 1. Updates H5vccPlatformService to inherit from ExecutionContextLifecycleStateObserver and listen for FrameLifecycleState::kRunning.
> 2. Proactively re-establishes the Mojo connection upon unfreeze/resume.
> 3. Adds on-demand EnsureConnected() in send() so unexpected disconnections are recovered seamlessly.
> 4. Distinguishes between explicit JS close() calls and Mojo disconnections, preserving the ReceiveCallback across freeze/resume cycles.

Do we know why the Mojo pipes are getting disconnected? Based on https://chromium.googlesource.com/chromium/src/mojo/+/HEAD/public/cpp/bindings/README.md#connection-errors that seems kind of surprising.

Anyway, I have some concerns around the high level approach to proactively reconnect the pipes. IIUC, the proposal is to reconnect the Mojo pipes - which involves creating a new platform-side platform service instance - without informing the web app. Since a platform service can define an arbitrary protocol and may have state, this could lead to unexpected behavior since the web app and platform service may not agree on what's been communicated so far, etc. `kInvalidStateError` on the other hand signals to the web app that it needs to reopen the platform service and start from scratch.

PR #11957, opened 2026-08-14
aosp: Resolve and package Evergreen content from the content dir

hferreiro:
On Evergreen, static content lives under the app's own content
directory, which the loader passes to the library. Align AOSP with
Linux, where the Evergreen content path overrides the default
kSbSystemPathContentDirectory.

Package what was still left outside of that directory under it:
cobalt_shell.pak, the CA certificates, and nplb's static test data,
both the file_tests directory and the player .dmp fixtures.

Bug: 495203133


starboard/nplb/testdata/file_tests/BUILD.gn
```diff
@@ -41,7 +41,8 @@ if (is_android && !is_androidtv) {
     renaming_destinations = []
     foreach(_output, renaming_sources) {
       renaming_destinations +=
-          [ rebase_path(_output, sb_static_contents_output_data_dir) ]
+          [ "app/cobalt/content/" +
```
andrewsavage1, line 44: How does test data for i.e. base_unittests get packaged in? We'd want to make sure we're following a pattern that we can use for chromium test targets without changing those targets for our packaging

hferreiro: If you don't mind, I'll make it work universally in follow-ups.

andrewsavage1: sgtm

andrewsavage1 (approved):

PR #11996, opened 2026-08-14
cobalt/metrics: Peak PMF measurements after foreground

Awallky:
Introduce peak Private Memory Footprint (PMF) measurements specifically
for when the process resumes from a background state. This helps
identify memory spikes that occur during the transition back to the
foreground.  New UMA histograms are added to capture these peaks over
standard intervals.

**Note:** Browsertests are added in a follow-up in [PR#12400](https://github.com/youtube/cobalt/pull/12400).

Bug: 543084356


Awallky:
Replaced Finch test configuration with pure 120s baseline waiting. The test environment lacked a robust propagation of scoped feature lists across processes, causing empty interval datasets. Now robust.

third_party/blink/renderer/controller/highest_pmf_reporter.cc
```diff
@@ -192,6 +224,53 @@ void HighestPmfReporter::OnMemoryPing(MemoryUsage usage) {
...
+void HighestPmfReporter::ProcessBackgrounded() {
+  DCHECK(IsMainThread());
+  cancelable_report_task_.Cancel();
+  is_foreground_measuring_ = false;
+  if (MemoryUsageMonitor::Instance().HasObserver(this)) {
+    MemoryUsageMonitor::Instance().RemoveObserver(this);
+  }
+}
+
+void HighestPmfReporter::OnProcessForegrounded() {
```
johnxwork: This is a new function that does not have a user?

Awallky: It's called in `third_party/blink/renderer/controller/blink_initializer.cc`, which is called by `content/renderer/render_thread_impl.cc`.

```
void RenderThreadImpl::OnRendererForegrounded() {
  UpdateForegroundCrashKey(/*foreground=*/true);
  main_thread_scheduler_->SetRendererBackgrounded(false);
  discardable_memory_allocator_->OnForegrounded();
  base::allocator::PartitionAllocSupport::Get()->OnForegrounded(
      MainFrameCounter::has_main_frame());
  blink::OnProcessForegrounded(); // our hooks from blink_initializer.cc calls our code here
  process_foregrounded_count_++;
}
```

For the sake of test coverage, I just added a new one that exercises the `RenderThreadImpl::OnRendererForegrounded` code path as well.

johnxwork (approved):

tools/metrics/histograms/metadata/cobalt/histograms.xml
```diff
@@ -760,13 +760,55 @@ Always run the pretty print utility on this file after editing:
...
 
 <histogram
-    name="Memory.Experimental.Renderer.PeakResidentSet.AtHighestPrivateMemoryFootprint"
-    units="MB" expires_after="2027-08-13">
+    name="Memory.Experimental.Renderer.HighestPrivateMemoryFootprintWhenForegrounded{Interval}"
+    units="MB" expires_after="never">
+  <owner>avvall@google.com</owner>
+  <owner>cobalt-dev@google.com</owner>
+  <summary>
+    The highest private memory footprint of a renderer process in (X mins after
```
yell0wd0g: nit: I read this and it took me a minute to understand, maybe
"of a renderer process in the (X, Y] mins after resuming from
a background state interval" would be clearer?

third_party/blink/renderer/controller/highest_pmf_reporter.h
```diff
@@ -64,8 +76,21 @@ class CONTROLLER_EXPORT HighestPmfReporter
...
+    WTF::String pmf_name;
+    WTF::String pmf_foregrounded_name;
+    WTF::String peak_rss_name;
+    WTF::String peak_rss_foregrounded_name;
+  };
+  WTF::Vector<MetricInfo> metrics_;
+
+  // True when measuring metrics after resuming from background state into
+  // foreground. False when measuring during initial startup navigation.
+  bool is_foreground_measuring_ = false;
```
yell0wd0g: The comment doesn't seem to fully agree with the name? Or perhaps
the member var name is about the mechanism and the description is
the usage? I guess what's I'm saying is why not call the var
`has_been_backgrounded_once_` and invert the logic?

third_party/blink/renderer/controller/highest_pmf_reporter.cc
```diff
@@ -192,6 +226,57 @@ void HighestPmfReporter::OnMemoryPing(MemoryUsage usage) {
...
+  }
+}
+
+// Handles transition into background: cancel in-flight reporting tasks and
+// stop observing memory usage.
+void HighestPmfReporter::ProcessBackgrounded() {
+  DCHECK(IsMainThread());
+  cancelable_report_task_.Cancel();
+  is_foreground_measuring_ = false;
+  if (MemoryUsageMonitor::Instance().HasObserver(this)) {
```
yell0wd0g: If this is a //base/observer_list.h class, the HasObserver() call
is not needed: RemoveObserver does nothing if `this` is not 
registered. HasObserver() is needed in l.269 tho.

third_party/blink/renderer/controller/highest_pmf_reporter.cc
```diff
@@ -192,6 +226,57 @@ void HighestPmfReporter::OnMemoryPing(MemoryUsage usage) {
   // lifetime.
 }
 
+#if BUILDFLAG(IS_COBALT)
+void HighestPmfReporter::OnProcessBackgrounded() {
+  if (HighestPmfReporter::Instance()) {
```
yell0wd0g: I guess this would happen if OnProcessBackgrounded is called
before a HighestPmfReporter is created, can that happen? Just
wondering if CHECK() makes more sense.

third_party/blink/renderer/controller/highest_pmf_reporter.h
```diff
@@ -35,6 +36,12 @@ class CONTROLLER_EXPORT HighestPmfReporter
   // Returns the shared instance.
   static void Initialize(
       scoped_refptr<base::SingleThreadTaskRunner> task_runner);
+#if BUILDFLAG(IS_COBALT)
+  static HighestPmfReporter* Instance();
+  static void OnProcessForegrounded();
+  static void OnProcessBackgrounded();
+  ~HighestPmfReporter() override;
```
yell0wd0g, line 41: I find it odd to tag all these static singleton-like operations
in a class that is meant to be creating objects, IIUC. It might
be clearer to wrap all the static/singleton bits in a single 
internal class?  Name TBD

yell0wd0g (commented):
Don't see anything preventing this from landing but wanted
to chat about the bundling of all the newly added static
bits.

third_party/blink/renderer/controller/highest_pmf_reporter.h
Awallky: Callers in blink_initializer.cc dispatch via 
```cpp
if (auto* reporter = HighestPmfReporter::Instance()) reporter->OnProcess...()
```
and tests call them directly on `reporter_`.
WDYT?

tools/metrics/histograms/metadata/cobalt/histograms.xml
Awallky: I updated the summary description to clarify the resumed interval wording: *The highest private memory footprint of a renderer process in the (X, Y] mins after resuming from a background state interval...*

third_party/blink/renderer/controller/highest_pmf_reporter.h
Awallky: Renamed `is_foreground_measuring_` to `has_been_backgrounded_once_` and inverted the logic/updated comments accordingly. Initial startup measuring is now clearly guarded by `!has_been_backgrounded_once_`.

third_party/blink/renderer/controller/highest_pmf_reporter.cc
Awallky: Removed the redundant `HasObserver()` checks in `OnProcessBackgrounded()` and `~HighestPmfReporter()` since `base::ObserverList::RemoveObserver()` is a no-op if not present.

third_party/blink/renderer/controller/highest_pmf_reporter.cc
Awallky: I removed the CHECK and invoke them via 

```cpp
if (auto* reporter = HighestPmfReporter::Instance()) reporter->OnProcess...()
```
ensuring we do not incorrectly use/call them on platforms that do not support them.

third_party/blink/renderer/controller/highest_pmf_reporter.h
yell0wd0g: Gotcha, so it's meant to be used as singleton throughout, correct?
If so then the surprise/confusion on my side came from the 
`if (instance_ == this) ` check in the dtor - I thought about
it for a minute but didn't see a path for several such instances
to be constructed, but perhaps unit tests? If so, it should be
commented or even better use some preprocessor or runtime check
around. If not, then just a CHECK() would be preferred.

yell0wd0g (approved):

PR #5422, opened 2025-04-15
[android] Support multiple punchout players via video holes

borongc:
Currently, SbPlayer with punch-out mode requires a video hole with HW overlay. Hence, if HW overlay is required, OverlayStrategyUnderlayStarboard will CommitCandidate() all video holes for the underlying punch-out SbPlayer, so that viz can update the bounds of all the SbPlayer accordingly.

For other cases (e.g., decode-to-texture mode as secondary player, b/375070492), we should revisit this logic (i.e., OverlayProcessorUsingStrategy::AttemptMultipleOverlays()) can still work after decode-to-texture mode is supported.

b/408303622

PR #12034, opened 2026-08-17
android: Disable SetVideoSurfaceBounds for decode-to-texture

johnedocampo:
This change moves the SetVideoSurfaceBounds call into the video
decoder sink and ensures it is only invoked when the player is not
using decode-to-texture mode. This can help prevent layering and display issues
with dual playback videos when the smaller video is using
decode-to-texture. Decode-to-texture should not need to call this
function to properly display the video.

As a sanity check, I forced enabled `decode-to-texture` locally with
these changes in place, and verified that the app still functioned
properly without any visual issues.

Bug: 505747981
Bug: 548031519

### DIFF
--- a/starboard/android/shared/media_codec_video_decoder.cc
+++ b/starboard/android/shared/media_codec_video_decoder.cc
@@ -29,6 +29,7 @@
 #include "starboard/android/shared/media_capabilities_cache.h"
 #include "starboard/android/shared/media_codec_video_decoder_helpers.h"
 #include "starboard/android/shared/media_common.h"
+#include "starboard/android/shared/starboard_bridge.h"
 #include "starboard/android/shared/video_render_algorithm_android.h"
 #include "starboard/android/shared/video_surface_texture_bridge.h"
 #include "starboard/common/check_op.h"
@@ -228,8 +229,10 @@ class MediaCodecVideoDecoder::Sink : public VideoRendererSink {
  public:
   typedef std::function<void(double playback_rate)> PlaybackRateChangedCB;
 
-  Sink(PlaybackRateChangedCB playback_rate_changed_cb)
-      : playback_rate_changed_cb_(playback_rate_changed_cb) {}
+  Sink(SbPlayerOutputMode output_mode,
+       PlaybackRateChangedCB playback_rate_changed_cb)
+      : output_mode_(output_mode),
+        playback_rate_changed_cb_(playback_rate_changed_cb) {}
 
   bool Render() {
     SB_DCHECK(render_cb_);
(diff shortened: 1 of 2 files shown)


starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -253,7 +256,13 @@ class MediaCodecVideoDecoder::Sink : public VideoRendererSink {
...
   }
 
-  void SetBounds(int z_index, const Rect& rect) override {}
+  void SetBounds(int z_index, const Rect& rect) override {
+    if (output_mode_ != kSbPlayerOutputModeDecodeToTexture) {
+      JNIEnv* env = jni_zero::AttachCurrentThread();
+      StarboardBridge::GetInstance()->SetVideoSurfaceBounds(
+          env, rect.x, rect.y, rect.size.width, rect.size.height);
+    }
+  }
```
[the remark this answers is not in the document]
johnedocampo: This is incorrect; the current code will compile successfully. Ignoring for now.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -29,6 +29,7 @@
 #include "starboard/android/shared/media_capabilities_cache.h"
 #include "starboard/android/shared/media_codec_video_decoder_helpers.h"
 #include "starboard/android/shared/media_common.h"
+#include "starboard/android/shared/starboard_bridge.h"
```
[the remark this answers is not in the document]
johnedocampo: Not necessary; this include is already present on line 50.

starboard/android/shared/media_codec_video_decoder.cc
```diff
@@ -253,7 +256,13 @@ class MediaCodecVideoDecoder::Sink : public VideoRendererSink {
     render_cb_ = render_cb;
   }
 
-  void SetBounds(int z_index, const Rect& rect) override {}
+  void SetBounds(int z_index, const Rect& rect) override {
```
borongc, line 259: Punch-out and DTT should be two different SbPlayers, and their bounds should be updated by its unique ID from [cobalt/media/service/video_geometry_setter_service.cc](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/media/service/video_geometry_setter_service.cc;l=84).

We should figure out why for SbPlayer with DTT mode, it is passed to SbPlayer with punch-out mode.

borongc: We tested with multiple punch-out videos in https://github.com/youtube/cobalt/pull/5422.

With DTT, we haven't tested it, so it is possible that for DTT, we don't need to punch a hole for rendering. That is, not calling `paint_video_hole_frame_cb_` in media/starboard/starboard_renderer.cc.

johnedocampo: > We should figure out why for SbPlayer with DTT mode, it is passed to SbPlayer with punch-out mode.

To my understanding, while SbPlayer with DTT has it's own ID, the DTT SbPlayer will still eventually call `starboard::StarboardBridge::GetInstance()->SetVideoSurfaceBounds()`, leading to [setVideoSurfaceBounds in CobaltActivity.java](https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java;l=832?q=setVideoSurfaceBounds&ss=h%2Fgithub%2Fyoutube%2Fcobalt%2F%2B%2Frefs%2Fheads%2Fmain), which will set `mVideoSurfaceView`, the surface view for the punch out SbPlayer.

borongc: This should be handled by https://github.com/youtube/cobalt/pull/12097.

johnedocampo:
Closing this PR as https://github.com/youtube/cobalt/pull/12097 is up and addresses the same issue.

PR #12049, opened 2026-08-18
android: Expose texture crop to WebGL.

at-ninja:
When using the new OES_EGL_image_external implementation on AOSP builds of Cobalt, resolutions that do not conform to the macroblock boundaries used by Android's textures bleed corrupted data into the final rendered content.

Add a new, custom WebGL extension, `COBALT_video_texture_transform`, that exposes a binding for Android's
`SurfaceTexture.getTransformMatrix()` API. This allows users of OES_EGL_image_external to crop out the corrupted edges, similar to how a fully native APK would use the SurfaceTexture API.

Bug: 490474392


third_party/blink/renderer/modules/webgl/webgl2_rendering_context.cc
```diff
@@ -53,6 +53,9 @@
 #if BUILDFLAG(USE_STARBOARD_MEDIA)
 #include "third_party/blink/renderer/modules/webgl/cobalt/oes_egl_image_external.h"
 #endif  // BUILDFLAG(USE_STARBOARD_MEDIA)
+#if BUILDFLAG(USE_STARBOARD_MEDIA) && BUILDFLAG(IS_ANDROID) && BUILDFLAG(IS_ANDROIDTV)
```
borongc: nit: should we move this inside L53 to L55, so we can only use `#if BUILDFLAG(IS_ANDROID) && BUILDFLAG(IS_ANDROIDTV)`?

borongc:
As this is for aosp, should the title rename to `aosp: Expose xxx`?

borongc (approved):

third_party/blink/renderer/modules/webgl/webgl2_rendering_context.cc
at-ninja: I have it structured this way to align with [Reduce Nesting, Reduce Complexity](https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html), since style guide recommends [preprocessor directives](https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives) not using indentation.

I also removed all of the `&& BUILDFLAG(IS_ANDROIDTV)` checks, so the complexity of the directives is reduced. Please let me know if you would prefer nesting.

at-ninja:
> As this is for aosp, should the title rename to aosp: Expose xxx?

We spoke offline, agreed that `android: ` is the correct title.

third_party/blink/renderer/modules/webgl/webgl_texture.h
```diff
@@ -60,6 +60,7 @@ class WebGLTexture : public WebGLObject {
                               scoped_refptr<media::VideoFrame> video_frame,
                               bool has_shared_image_access);
   const gpu::Mailbox& GetMailbox() const { return mailbox_; }
+  scoped_refptr<media::VideoFrame> GetVideoFrame() const { return video_frame_; }
```
andrewsavage1: This is for non-android devices as well?

third_party/blink/renderer/modules/webgl/BUILD.gn
```diff
@@ -215,4 +215,10 @@ blink_modules_sources("webgl") {
       "cobalt/oes_egl_image_external.h",
     ]
   }
+  if (is_cobalt && use_starboard_media && target_os_is_android) {
```
andrewsavage1, line 224: Generally we should use `is_android` as a platform specifier, why do you use `target_os_is_android` here?

andrewsavage1 (approved):

third_party/blink/renderer/modules/webgl/webgl_texture.h
at-ninja: Added `BUILDFLAG(IS_ANDROID)` check to not expose this further than needed.

third_party/blink/renderer/modules/webgl/BUILD.gn
at-ninja: Added comment for why `target_os_is_android` is used. This is similar to other modules, like [document_picture_in_picture/BUILD.gn](https://github.com/youtube/cobalt/blob/main/third_party/blink/renderer/modules/document_picture_in_picture/BUILD.gn)

PR #12057, opened 2026-08-18
feat(tvos): re-sign unit test app for physical devices

Libzu:
Mobile Harness requires unit test packages to be signed with a valid Development Provisioning Profile to run on physical lab devices (Apple TV). Currently, the pipeline fails with ApplicationVerificationFailed.

This change updates the macOS build script to embed the Google_Development_tvOS.mobileprovision profile and re-sign the .app bundle using codesign before packaging it for the testing pipeline

Bug: 538697105

### DIFF
--- a/cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
+++ b/cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
@@ -134,41 +134,26 @@ EOF
       # --------------------------------------------------------------------------
       # Embed Provisioning Profile & Re-sign App for Physical tvOS Devices
       # --------------------------------------------------------------------------
+      # Locate the tvOS development provisioning profile
       local tvos_profile="${KOKORO_PIPER_DIR:-}/google3/googlemac/iPhone/Shared/ProvisioningProfiles/Google_Development_tvOS.mobileprovision"
       if [[ ! -f "${tvos_profile}" ]]; then
         tvos_profile="$HOME/Library/Developer/Xcode/UserData/Provisioning Profiles/Google_Development_tvOS.mobileprovision"
       fi
 
-      if [[ -f "${tvos_profile}" ]]; then
-        echo "Embedding ${tvos_profile} into ${target_name}.app..."
-        cp -f "${tvos_profile}" "${out_dir}/${target_name}.app/embedded.mobileprovision"
+      echo "Embedding ${tvos_profile} into ${target_name}.app..."
+      cp -f "${tvos_profile}" "${out_dir}/${target_name}.app/embedded.mobileprovision"
 
-        # Extract entitlements from the provisioning profile
-        local profile_plist="${out_dir}/${target_name}_profile.plist"
-        local entitlements_plist="${out_dir}/${target_name}_entitlements.plist"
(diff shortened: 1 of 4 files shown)


oxve (changes requested):
There should be a single attempt to sign the binary. Since we know that only one provisioning profile works on mobile harness we should use it and not try any others.

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
```diff
@@ -130,6 +130,47 @@ EOF
...
       echo "Archiving and uploading test package ${target_name}.app..."
+
+      # --------------------------------------------------------------------------
+      # Embed Provisioning Profile & Re-sign App for Physical tvOS Devices
+      # --------------------------------------------------------------------------
+      local tvos_profile="${KOKORO_PIPER_DIR:-}/google3/googlemac/iPhone/Shared/ProvisioningProfiles/Google_Development_tvOS.mobileprovision"
+      if [[ ! -f "${tvos_profile}" ]]; then
+        tvos_profile="$HOME/Library/Developer/Xcode/UserData/Provisioning Profiles/Google_Development_tvOS.mobileprovision"
+      fi
+
```
linxinan-yt, line 142: Which mobileprovision we should use? Since this is running CI, I think we should make it more deterministic.

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
```diff
@@ -130,6 +130,47 @@ EOF
...
+      if [[ -f "${tvos_profile}" ]]; then
```
linxinan-yt: Instead of [[ -f "${tvos_profile}" ]], consider [[ ! -f "${tvos_profile}" ]]. If tvos profile does not exist, we should fail this job at early stage. Otherwise, the binary without signature cannot get executed in the lab.

Libzu:
> There should be a single attempt to sign the binary. Since we know that only one provisioning profile works on mobile harness we should use it and not try any others.

Thanks for the feedback. I've updated the script to perform a single, explicit codesign attempt with no fallbacks so it fails fast if signing cannot be completed properly

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
Libzu: Good point. Since `setup_mac()` copies all provisioning profiles into `$HOME/Library/Developer/Xcode/UserData/Provisioning Profiles/`, I've set `$HOME/Library/Developer/Xcode/UserData/Provisioning Profiles/Google_Development_tvOS.mobileprovision` as the single canonical profile path for CI.

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
Libzu: Agreed. I've updated the check to `if [[ ! -f "${tvos_profile}" ]]; then echo "ERROR: ..."; exit 1; fi` so the build fails immediately if the profile is missing, preventing unsigned packages from being archived or sent to the lab.

oxve (approved):

PR #12063, opened 2026-08-18
android: Reject video tunnel mode on Android < 14

jasonzhangxx:
Tunnel mode support is only intended for Android 14 and newer.
This update ensures that any video support query requesting tunnel
mode is rejected on devices running API levels below 34.

Bug: 547802014

### DIFF
--- a/starboard/android/shared/media_is_video_supported.cc
+++ b/starboard/android/shared/media_is_video_supported.cc
@@ -16,6 +16,8 @@
 #include "starboard/shared/starboard/media/media_support_internal.h"
 // clang-format on
 
+#include <android/api-level.h>
+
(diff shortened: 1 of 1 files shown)


osagie98 (approved):

starboard/android/shared/media_is_video_supported.cc
```diff
@@ -16,6 +16,8 @@
 #include "starboard/shared/starboard/media/media_support_internal.h"
 // clang-format on
 
+#include <android/api-level.h>
```
borongc, line 19: nit: update title to `android: xxx`

jasonzhangxx: Done

PR #12066, opened 2026-08-19
cobalt: Add memory metrics to Performance API

sherryzy:
Add new methods to the Performance web API to provide more granular
memory usage information, such as RSS high water mark, total physical
memory, and PSS. These metrics enable web applications to better
monitor their memory footprint and limits on constrained devices.

The base::ProcessMetrics class is also updated to include VmHWM
support on Linux-based platforms to source these metrics.

Bug: 548530403

### DIFF
--- a/base/process/process_metrics_linux.cc
+++ b/base/process/process_metrics_linux.cc
@@ -156,12 +156,16 @@ ProcessMetrics::GetMemoryInfo() const {
   }
   ProcessMemoryInfo dump;
   for (const auto& [key, value_str] : *pairs) {
-    if (key == "VmSize") {
 #if BUILDFLAG(IS_COBALT)
+    if (key == "VmSize") {
       dump.vm_size_bytes =
           static_cast<uint64_t>(GetKbFieldAsSizeT(value_str)) * 1024;
-#endif    // BUILDFLAG(IS_COBALT)
-    } else if (key == "VmSwap") {
+    } else if (key == "VmHWM") {
+      dump.vm_hwm_bytes =
+          static_cast<uint64_t>(GetKbFieldAsSizeT(value_str)) * 1024;
+    } else
+#endif  // BUILDFLAG(IS_COBALT)
+    if (key == "VmSwap") {
       dump.vm_swap_bytes =
           static_cast<uint64_t>(GetKbFieldAsSizeT(value_str)) * 1024;
     } else if (key == "VmRSS") {
--- a/base/process/process_metrics_unittest.cc
+++ b/base/process/process_metrics_unittest.cc
@@ -759,6 +759,12 @@ TEST_F(SystemMetricsTest, TestValidMemoryInfo) {
 #if BUILDFLAG(IS_WIN)
   EXPECT_GT(memory_info->private_bytes, 0U);
 #endif  // BUILDFLAG(IS_WIN)
+
+#if BUILDFLAG(IS_COBALT) && \
+    (BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_ANDROID))
+  EXPECT_GT(memory_info->vm_size_bytes, 0U);
+  EXPECT_GT(memory_info->vm_hwm_bytes, 0U);
+#endif  // BUILDFLAG(IS_COBALT) && ...
 }
 
 #if BUILDFLAG(IS_CHROMEOS)
--- a/cobalt/browser/performance/performance_impl.cc
+++ b/cobalt/browser/performance/performance_impl.cc
@@ -21,12 +21,24 @@
 #include "base/task/thread_pool.h"
 #include "build/build_config.h"
 
+#if BUILDFLAG(IS_POSIX)
+#include <unistd.h>
+#endif
+
+#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_ANDROID)
+#include "base/debug/proc_maps_linux.h"
+#endif
+
 #if BUILDFLAG(IS_ANDROIDTV)
 #include "starboard/android/shared/starboard_bridge.h"
 
 using ::starboard::StarboardBridge;
 #elif BUILDFLAG(IS_STARBOARD)
 #include "starboard/common/time.h"
+#endif
+
+#if BUILDFLAG(IS_STARBOARD)
+#include "starboard/system.h"
 #endif
 
 namespace performance {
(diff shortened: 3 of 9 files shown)


cobalt/browser/performance/performance_impl.cc
```diff
@@ -111,6 +123,100 @@ void PerformanceImpl::MeasureReservedVirtualMemory(
...
+}
+
+void PerformanceImpl::MeasureUsedPssMemory(
+    MeasureUsedPssMemoryCallback callback) {
+#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS) || BUILDFLAG(IS_ANDROID)
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::MayBlock(), base::TaskPriority::USER_VISIBLE},
+      base::BindOnce([]() -> uint64_t {
+        auto smaps_rollup = base::debug::ReadAndParseSmapsRollup();
+        return smaps_rollup.has_value() ? smaps_rollup->pss : 0;
```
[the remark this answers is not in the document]
sherryzy: Incorrect. base::debug::ReadAndParseSmapsRollup() internally converts the kB values from /proc/self/smaps_rollup to bytes during parsing.

cobalt/browser/performance/performance_impl.cc
```diff
@@ -111,6 +123,100 @@ void PerformanceImpl::MeasureReservedVirtualMemory(
...
+#endif
+}
+
+void PerformanceImpl::MeasureTotalCpuMemory(
+    MeasureTotalCpuMemoryCallback callback) {
+  base::ThreadPool::PostTaskAndReplyWithResult(
+      FROM_HERE, {base::MayBlock(), base::TaskPriority::USER_VISIBLE},
+      base::BindOnce([]() -> uint64_t {
+#if BUILDFLAG(IS_POSIX) && defined(_SC_PHYS_PAGES)
+        long pages = sysconf(_SC_PHYS_PAGES);
```
y4vor, line 172: can we use the base implementation? 

https://source.corp.google.com/h/github/youtube/cobalt/+/main:copied_base/base/system/sys_info_linux.cc;l=28?q=f:base%20sysconf&sq=r:github%2Fyoutube%2Fcobalt%20b:main

sherryzy: Thanks for pointing that out!

AmountOfMemory() is in an anonymous namespace with internal linkage, so it isn't accessible from outside sys_info_linux.cc. Additionally, the public base::SysInfo::AmountOfPhysicalMemory() caps RAM at 512 MB in Cobalt due to default low-end device mode, while the uncapped AmountOfPhysicalMemoryImpl() is private.

Keeping the direct sysconf() calculation here lets us accurately measure the true physical RAM using the exact same logic while avoiding changes to Chromium's core //base headers. Otherwise, we need to change base code, create a public  and modify upstream code.  I feel existing way is slightly clearer. Does it make sense?

y4vor: Sounds good

y4vor (approved):

base/process/process_metrics_linux.cc
```diff
@@ -156,12 +156,16 @@ ProcessMetrics::GetMemoryInfo() const {
   }
   ProcessMemoryInfo dump;
   for (const auto& [key, value_str] : *pairs) {
-    if (key == "VmSize") {
```
Awallky, line 159: Do we want to move this inside of the IS_COBALT block? Seems like we are modifying upstream code behavior here.

base/process/process_metrics_unittest.cc
```diff
@@ -759,6 +759,12 @@ TEST_F(SystemMetricsTest, TestValidMemoryInfo) {
 #if BUILDFLAG(IS_WIN)
   EXPECT_GT(memory_info->private_bytes, 0U);
 #endif  // BUILDFLAG(IS_WIN)
+
+#if BUILDFLAG(IS_COBALT) && \
+    (BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS) || BUILDFLAG(IS_ANDROID))
```
Awallky: I don't think we need the IS_CHROMEOS condition here.

base/process/process_metrics_linux.cc
sherryzy: Nope, I'm not modifying upstream code.  This earlier change modified the upstream code: https://github.com/youtube/cobalt/pull/9392/, and I'm reverting it and gate the change by Cobalt flag.

cobalt/browser/performance/public/mojom/performance.mojom
```diff
@@ -32,6 +32,26 @@ interface CobaltPerformance {
   [Sync]
   MeasureReservedVirtualMemory() => (uint64 bytes);
 
+  // Get the peak resident set size (high water mark) used by the process in bytes.
+  [Sync]
```
Awallky, line 36: I know this is consistent with the other performance APIs, but do we want to consider making this a non-blocking API? Considering this data is all sourced from file I/O, it goes against best practices to block the UIThread on file I/O.
IMO a Promise we can eventually resolve should be sufficient.

cobalt/browser/performance/performance_impl.cc
```diff
@@ -111,6 +123,100 @@ void PerformanceImpl::MeasureReservedVirtualMemory(
       std::move(callback));
 }
 
+void PerformanceImpl::MeasureRssHighWaterMarkMemory(
+    MeasureRssHighWaterMarkMemoryCallback callback) {
+#if BUILDFLAG(IS_LINUX) || BUILDFLAG(IS_CHROMEOS) || BUILDFLAG(IS_ANDROID)
```
Awallky: Shouldn't need ChROMEOS here.

Awallky (approved):
LGTM

I filed a bug (b/549168445) to add some tests to the APIs.

cobalt/browser/performance/public/mojom/performance.mojom
sherryzy: I think this generally makes sense, but we need to check if clients will cal this API asynchronously. This will likely require more thought and decision-making. Since I probably won't have time to focus on this, could you help follow up and make any necessary changes?

aelzeiny:
LETSSS GOOOOOOOOO

PR #12068, opened 2026-08-19
android: Enable player_filter_tests

osagie98:
Enables `player_filter_tests` for Android arm and arm64.

This change introduces `CobaltTestActivity` to handle the necessary JNI
and application context initialization required for native unit tests
on Android. It also configures the build system to package media test
assets and updates the test runner to locate them on device storage.
HDR capability checks are moved to DisplayUtil.

The following suites are temporarily filtered due to their extended runtimes and flakiness:
  - `"*PlayerComponentsTest*"`
  - `"*AdaptiveAudioDecoderTest*"`
  - `"*AudioDecoderTest*"`
  - `"*VideoDecoderTest*"`
  - `"*AudioRendererTest*"`

#vibe-coded

Issue: 372559388
Issue: 535995457

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/testing/CobaltTestActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/testing/CobaltTestActivity.java
@@ -15,6 +15,7 @@
 package dev.cobalt.testing;
 
 import android.os.Bundle;
+import dev.cobalt.util.DisplayUtil;
 import org.chromium.base.ContextUtils;
 import org.chromium.base.JNIUtils;
 import org.chromium.native_test.NativeUnitTestActivity;
@@ -30,6 +31,7 @@ public class CobaltTestActivity extends NativeUnitTestActivity {
   public void onCreate(Bundle savedInstanceState) {
     ContextUtils.initApplicationContext(getApplicationContext());
     JNIUtils.setDefaultClassLoader(getClassLoader());
+    DisplayUtil.cacheDefaultDisplay(this);
     super.onCreate(savedInstanceState);
   }
 }
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/testing/CobaltTestActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/testing/CobaltTestActivity.java
@@ -19,6 +19,12 @@ import org.chromium.base.ContextUtils;
 import org.chromium.base.JNIUtils;
 import org.chromium.native_test.NativeUnitTestActivity;
 
+/**
+ * CobaltTestActivity is a custom NativeUnitTestActivity used to run native unit tests on Android.
+ *
+ * <p>It ensures that the Chromium base library's application context and JNI class loader are
+ * properly initialized before the native test harness starts.
+ */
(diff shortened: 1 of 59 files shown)


starboard/BUILD.gn
```diff
@@ -258,7 +258,7 @@ if (current_toolchain == starboard_toolchain) {
       deps += [ platform_tests_path ]
     }
     if (sb_filter_based_player) {
-      deps += [ "//starboard/shared/starboard/player/filter/testing:player_filter_tests" ]
+      deps += [ "//starboard/shared/starboard/player/filter/testing:player_filter_tests_sources" ]
     }
```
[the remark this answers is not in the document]
osagie98: Accepted the suggestion

cobalt/build/modular_executable.gni
```diff
@@ -72,6 +72,7 @@ template("loader_final_target") {
 
     shared_library(original_target_name) {
       forward_variables_from(invoker, "*")
+      not_needed("*")
```
osagie98: There's a gn checker issue on platforms such as aosp and linux modular where certain variables from invoker are unused. This line allows those variables to be ignored. Similar below

cobalt/android/apk/app/src/main/java/dev/cobalt/testing/CobaltTestActivity.java
```diff
@@ -0,0 +1,35 @@
...
+import org.chromium.base.JNIUtils;
+import org.chromium.native_test.NativeUnitTestActivity;
+
+/**
+ * CobaltTestActivity is a custom NativeUnitTestActivity used to run native unit tests on Android.
+ *
+ * <p>It ensures that the Chromium base library's application context and JNI class loader are
+ * properly initialized before the native test harness starts.
+ */
+public class CobaltTestActivity extends NativeUnitTestActivity {
```
osagie98: When running player filter tests, the test setup step attempts to call Java before JNI is initialized. With this extended activity, we can ensure JNI is ready before the test setup begins.

This is a problem in media_capabilities_cache.cc, where we call Java to determine if the device can support a given test file. If we don't do this, all test files will be loaded, and unsupported ones will be skipped in the test body. That'll pollute the logs and increase the test suite run times.

johnxwork (approved):

johnxwork:
Looks good from ATV perspective, but you must get a media signoff before merging.

starboard/android/shared/display_util.h
```diff
@@ -28,6 +30,10 @@ class DisplayUtil {
   // Returns the physical pixels per inch of the screen in the X and Y
   // dimensions.
   static Dpi GetDisplayDpi();
+
+  // Returns supported HDR types.
+  static base::android::ScopedJavaLocalRef<jintArray> GetSupportedHdrTypes(
```
kjyoun: can you make this method return std::vector<int> ?

Then callsite  and this file does not need include `#include "base/android/scoped_java_ref.h"

We had better to avoid deps on base// from starboard.

e.g.) 

```
    std::vector<int> hdr_types;
    base::android::JavaIntArrayToIntVector(env, j_supported_hdr_types,
                                           &hdr_types);
    return hdr_types;
```

kjyoun (approved):

PR #12069, opened 2026-08-19
cobalt: Implement long hang detection and kill

alunev:
cobalt: Implement long hang detection and kill

This change introduces the ability to independently detect, log, and
force-kill the application during severe thread hangs. A new UMA
metric, HangWatcher.LongHang.Detected, is emitted when a hang
exceeds a configurable timeout.

Optionally, the application can be configured to perform a native
abort after the metric is emitted and a crash dump is captured.
Dynamic control of these features is provided through h5vcc.settings
to allow for flexible management from the JavaScript layer.

Bug: 546124857

### DIFF
--- a/base/threading/hang_watcher.cc
+++ b/base/threading/hang_watcher.cc
@@ -71,6 +71,9 @@ std::atomic<int64_t> g_hang_watch_monitoring_period_us{
     base::Seconds(10).InMicroseconds()};
 
 std::atomic<bool> g_hang_reporting_enabled{true};
+std::atomic<bool> g_enable_long_hang_detection{false};
+std::atomic<bool> g_enable_long_hang_kill{false};
+std::atomic<int> g_long_hang_timeout_seconds{20};
 
 static std::atomic<HangWatcher::Delegate*> g_hang_watcher_delegate{nullptr};
 #endif
@@ -338,6 +341,11 @@ constexpr const char* kThreadName = "HangWatcher";
 // HangWatcher.IsThreadHung need to be updated.
 constexpr auto kMonitoringPeriod = base::Seconds(10);
 
+#if BUILDFLAG(IS_COBALT)
+// The period after which continuous hang is considered and reported as "long"
+constexpr auto kDefaultLongHangTimeout = base::Seconds(20);
+#endif
+
 WatchHangsInScope::WatchHangsInScope(TimeDelta timeout) {
 #if BUILDFLAG(IS_COBALT)
   // Only override the timeout with the global configuration if the caller
@@ -450,6 +458,7 @@ WatchHangsInScope::~WatchHangsInScope() {
 
 #if BUILDFLAG(IS_COBALT)
 // static
+
 void HangWatcher::UpdateConfiguration() {
   auto* delegate = g_hang_watcher_delegate.load(std::memory_order_relaxed);
   if (!delegate) {
@@ -475,6 +484,28 @@ void HangWatcher::UpdateConfiguration() {
 
   g_hang_reporting_enabled.store(delegate->IsHangReportingEnabled(),
                                  std::memory_order_relaxed);
+
+  if (auto detection_enabled = delegate->IsLongHangDetectionEnabled()) {
+    g_enable_long_hang_detection.store(*detection_enabled,
+                                       std::memory_order_relaxed);
+  } else {
+    g_enable_long_hang_detection.store(false, std::memory_order_relaxed);
+  }
+
+  if (auto kill_enabled = delegate->IsLongHangKillEnabled()) {
+    g_enable_long_hang_kill.store(*kill_enabled, std::memory_order_relaxed);
+  } else {
+    g_enable_long_hang_kill.store(false, std::memory_order_relaxed);
+  }
+
+  if (auto configured_long_timeout = delegate->GetLongHangTimeout()) {
+    g_long_hang_timeout_seconds.store(
+        static_cast<int>(configured_long_timeout->InSeconds()),
+        std::memory_order_relaxed);
+  } else {
+    g_long_hang_timeout_seconds.store(kDefaultLongHangTimeout.InSeconds(),
+                                      std::memory_order_relaxed);
+  }
 
   // Update logging levels for thread scopes.
   if (auto io_enabled = delegate->IsThreadDumpingEnabled(
@@ -611,6 +642,9 @@ void HangWatcher::InitializeOnMainThread(ProcessType process_type,
 void HangWatcher::UninitializeOnMainThreadForTesting() {
 #if BUILDFLAG(IS_COBALT)
   g_hang_reporting_enabled.store(true, std::memory_order_relaxed);
+  g_enable_long_hang_detection.store(false, std::memory_order_relaxed);
+  g_enable_long_hang_kill.store(false, std::memory_order_relaxed);
+  g_long_hang_timeout_seconds.store(20, std::memory_order_relaxed);
 #endif
   g_use_hang_watcher.store(false, std::memory_order_relaxed);
   g_threadpool_log_level.store(LoggingLevel::kNone, std::memory_order_relaxed);
@@ -1180,6 +1214,7 @@ void HangWatcher::RecordHangStarted() {
   }
 
   active_hang_uuid_ = base::Uuid::GenerateRandomV4().AsLowercaseString();
+  hang_start_time_ = tick_clock_->NowTicks();
   std::string uuid_to_report = active_hang_uuid_;
 
   {
@@ -1199,7 +1234,7 @@ void HangWatcher::RecordHangStarted() {
 #endif
 }
 
-void HangWatcher::CheckAndRecordHangRecovered() {
+void HangWatcher::CheckHangState() {
   DCHECK_CALLED_ON_VALID_THREAD(hang_watcher_thread_checker_);
 
   if (active_hang_uuid_.empty()) {
@@ -1207,22 +1242,36 @@ void HangWatcher::CheckAndRecordHangRecovered() {
   }
 
   bool any_thread_hung = false;
-  base::TimeTicks now = base::TimeTicks::Now();
+  base::TimeTicks now = tick_clock_->NowTicks();
 
   // We iterate over the watch states to determine if ANY thread is currently
   // physically hung. Note: We are ALREADY inside HangWatcher::Monitor(), so
   // watch_state_lock_ is currently held by the caller. We do not need to
   // acquire or release it here for the iteration.
   for (const auto& watch_state : watch_states_) {
-    if (watch_state->GetDeadline() <= now) {
+    uint64_t flags;
+    base::TimeTicks deadline;
+    std::tie(flags, deadline) = watch_state->GetFlagsAndDeadline();
+
+    // Some threads can skip deadline for legitimate reason, we ignore them.
+    if (internal::HangWatchDeadline::IsFlagSet(
+            internal::HangWatchDeadline::Flag::kIgnoreCurrentWatchHangsInScope,
+            flags)) {
+      continue;
+    }
+
+    if (deadline <= now) {
       any_thread_hung = true;
       break;
     }
   }
 
+  // If all hangs cleared, mark last hang as recovered.
   if (!any_thread_hung) {
     std::string uuid_to_report = active_hang_uuid_;
     active_hang_uuid_.clear();
+    hang_start_time_ = base::TimeTicks();
+    long_hang_logged_ = false;
 
     {
       // Unlock before calling out to the delegate to prevent deadlocks with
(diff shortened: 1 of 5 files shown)


cobalt/browser/hang_watcher_delegate_impl.cc
```diff
@@ -33,6 +33,7 @@ namespace browser {
 void CobaltHangWatcherDelegate::Initialize() {
   static base::NoDestructor<CobaltHangWatcherDelegate> instance;
   base::HangWatcher::SetDelegate(instance.get());
+  base::HangWatcher::UpdateConfiguration();
```
hlwarriner, line 36: Just confirming my understanding - this call already should have been here and is somewhat independent from this change? Either way, lgtm.

base/threading/hang_watcher.h
```diff
@@ -158,6 +158,9 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
     // Returns std::nullopt if the embedder has no specific override.
     virtual std::optional<bool> IsThreadDumpingEnabled(
         ThreadType thread_type) = 0;
+    virtual std::optional<bool> IsLongHangDetectionEnabled();
+    virtual std::optional<bool> IsLongHangKillEnabled();
+    virtual std::optional<base::TimeDelta> GetLongHangTimeout();
```
hlwarriner: Please add function comments for these.

1) We should mention the `std:nullopt` behavior that these have, as we've done for other methods.
2) IMO the behavior of these functions is not so clear from their names that they don't require comments. For example, a reader may not realize `IsLongHangKillEnabled()` will actually enable the entire process to be killed if there is a long hang.

base/threading/hang_watcher.cc
```diff
@@ -476,6 +491,27 @@ void HangWatcher::UpdateConfiguration() {
   g_hang_reporting_enabled.store(delegate->IsHangReportingEnabled(),
                                  std::memory_order_relaxed);
 
+  if (auto detection_enabled = delegate->IsLongHangDetectionEnabled()) {
```
hlwarriner, line 488: Why are we missing this `if` check and `else` block for `IsHangReportingEnabled`, just above this?

base/threading/hang_watcher.h
```diff
@@ -158,6 +158,9 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
     // Returns std::nullopt if the embedder has no specific override.
     virtual std::optional<bool> IsThreadDumpingEnabled(
         ThreadType thread_type) = 0;
+    virtual std::optional<bool> IsLongHangDetectionEnabled();
```
hlwarriner: These can't be pure virtuals, like the methods declared above?

base/threading/hang_watcher.h
```diff
@@ -401,14 +408,38 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
...
+  base::TimeTicks hang_start_time_
+      GUARDED_BY_CONTEXT(hang_watcher_thread_checker_);
+
+  // This is used to prevent multiple long hang actions for the same hang.
+  bool long_hang_logged_ GUARDED_BY_CONTEXT(hang_watcher_thread_checker_) =
+      false;
+
+  // Iterates through the registered threads to determine if any thread has
+  // been hung for a duration exceeding the long hang timeout. Emits a UMA
+  // metric once per long hang and optionally forces a native abort.
```
hlwarriner, line 447: nit: I think it would help to mention here what exactly determines whether a native abort is forced.

base/threading/hang_watcher.cc
```diff
@@ -1199,30 +1239,43 @@ void HangWatcher::RecordHangStarted() {
...
-  base::TimeTicks now = base::TimeTicks::Now();
+  base::TimeTicks now = tick_clock_->NowTicks();
 
   // We iterate over the watch states to determine if ANY thread is currently
   // physically hung. Note: We are ALREADY inside HangWatcher::Monitor(), so
   // watch_state_lock_ is currently held by the caller. We do not need to
   // acquire or release it here for the iteration.
   for (const auto& watch_state : watch_states_) {
-    if (watch_state->GetDeadline() <= now) {
+    std::pair<uint64_t, base::TimeTicks> flags_and_deadline =
```
hlwarriner: I think the code below would be more clear if we use `std::tie(flags, deadline)` instead, as is already done in `HangWatcher::WatchStateSnapShot::Init()`. Then we could avoid doing `flags_and_deadline.first` etc.

base/threading/hang_watcher.cc
```diff
@@ -1242,6 +1295,55 @@ void HangWatcher::CheckAndRecordHangRecovered() {
       crash_ext->SetString(kNativeStabilityHangUuidKey, "");
     }
 #endif
+  } else {
+    // If still hanging, check if this is a long hang by now.
+    CheckForLongHangs(now);
+  }
+}
+
+// Note: If multiple threads hang in an overlapping cascade (e.g., Thread A
```
hlwarriner: nit: this is a very helpful comment but since it's more about the behavior than an implementation detail (IMO), I would even recommend moving it to the function comment in the header.

base/threading/hang_watcher.cc
```diff
@@ -1242,6 +1295,55 @@ void HangWatcher::CheckAndRecordHangRecovered() {
...
+      LOG(INFO) << "Freeze detection: Long hang detected! Emitting UMA metric.";
+      UMA_HISTOGRAM_BOOLEAN("HangWatcher.LongHang.Detected", true);
+      long_hang_logged_ = true;
+    }
+
+    if (g_enable_long_hang_kill.load(std::memory_order_relaxed)) {
+      if (force_kill_closure_for_testing_) {
+        force_kill_closure_for_testing_.Run();
+      } else {
+        SCOPED_CRASH_KEY_STRING32("HangWatcher", "HangWatcher-Reason",
```
hlwarriner, line 1338: Will this work on 3P? I thought we would have to `crash_ext->SetString()`.

base/threading/hang_watcher_unittest.cc
```diff
@@ -1550,6 +1541,374 @@ TEST_F(HangWatcherCobaltTest, OverlappingHangsSingleUuid) {
...
+      bool detection_enabled = true,
+      bool kill_enabled = false,
+      base::RepeatingClosure force_kill_closure = {}) {
+    SetDefaultDelegateExpectations(mock_delegate);
+    EXPECT_CALL(mock_delegate, IsLongHangDetectionEnabled())
+        .WillRepeatedly(testing::Return(detection_enabled));
+    EXPECT_CALL(mock_delegate, IsLongHangKillEnabled())
+        .WillRepeatedly(testing::Return(kill_enabled));
+    EXPECT_CALL(mock_delegate, GetLongHangTimeout())
+        .WillRepeatedly(testing::Return(base::Seconds(10)));
```
hlwarriner: This 10 second timeout should probably be passed into `StartAndRegister()` as an argument. Otherwise it's not clear from the body of the `ForceKillEnabledTriggersFatal` test case why `task_environment_.FastForwardBy(base::Seconds(11))` is sufficient to trigger the kill.

See the "Helper Methods" section of the Unit Testing Best Practices doc.

base/threading/hang_watcher_unittest.cc
```diff
@@ -1550,6 +1541,374 @@ TEST_F(HangWatcherCobaltTest, OverlappingHangsSingleUuid) {
...
+  // Monitor runs. IsActionable() is false because no threads are hung.
+  // CheckHangState() runs, sees no threads are hung, clears the incident state,
+  // and fires RecordHangRecovered.
+  EXPECT_CALL(mock_delegate, RecordHangRecovered(captured_uuid)).Times(1);
+  TriggerMonitorAndWait();
+
+  testing::Mock::VerifyAndClearExpectations(&mock_delegate);
+}
+
+TEST_F(HangWatcherCobaltTest, ThreadUnregistersWhileHung) {
```
hlwarriner: nit: please include the expected outcome in the test name.

hlwarriner (changes requested):
Left some comments but the overall approach lgtm.

hlwarriner:
CC @vincentmo @AlecsisDuarte. Don't feel the need to review this but please at least skim the changes, if you have time. I would be good if more people on the team have some familiarity with this code.

base/threading/hang_watcher_unittest.cc
alunev: ok

base/threading/hang_watcher_unittest.cc
alunev: ok

cobalt/browser/hang_watcher_delegate_impl.h
```diff
@@ -37,6 +37,9 @@ class CobaltHangWatcherDelegate : public base::HangWatcher::Delegate {
   std::optional<base::TimeDelta> GetHangWatchMonitoringPeriod() override;
   std::optional<bool> IsThreadDumpingEnabled(
       base::HangWatcher::ThreadType thread_type) override;
+  std::optional<bool> IsLongHangDetectionEnabled() override;
```
vincentmo, line 40: What's the purpose of returning an optional<bool> since the caller treats the empty case the same as false?

base/threading/hang_watcher.cc
```diff
@@ -476,6 +480,27 @@ void HangWatcher::UpdateConfiguration() {
...
+    g_enable_long_hang_detection.store(false, std::memory_order_relaxed);
+  }
+
+  if (auto kill_enabled = delegate->IsLongHangKillEnabled()) {
+    g_enable_long_hang_kill.store(*kill_enabled, std::memory_order_relaxed);
+  } else {
+    g_enable_long_hang_kill.store(false, std::memory_order_relaxed);
+  }
+
+  if (auto configured_long_timeout = delegate->GetLongHangTimeout()) {
```
vincentmo, line 501: Should this (and kill enablement) be gated behind long hang detection being enabled, since I imagine these 2 should be noops when long detection is disabled anyway?

base/threading/hang_watcher.cc
```diff
@@ -476,6 +480,27 @@ void HangWatcher::UpdateConfiguration() {
...
+    g_long_hang_timeout_seconds.store(
+        static_cast<int>(configured_long_timeout->InSeconds()),
+        std::memory_order_relaxed);
+  } else {
+    g_long_hang_timeout_seconds.store(20, std::memory_order_relaxed);
```
vincentmo: Add a constant that's shared between the declaration, here and elsewhere?

base/threading/hang_watcher.cc
alunev: thanks, yes, changed to extension.

base/threading/hang_watcher.cc
alunev: ok

base/threading/hang_watcher.cc
alunev: ok, makes sense

base/threading/hang_watcher.h
alunev: ok

base/threading/hang_watcher.h
alunev: thx, this and other flag-related things will go away in the next pr where I add a fallback to finch similar to existing `bool IsHangReportingEnabled() override;`

base/threading/hang_watcher.cc
alunev: cause it's optional instead of bool, but it's changing(see above comment)

base/threading/hang_watcher.h
alunev: also handled in flags PR(see above replies)

base/threading/hang_watcher.h
```diff
@@ -401,14 +415,44 @@ class BASE_EXPORT HangWatcher : public DelegateSimpleThread::Delegate {
 #if BUILDFLAG(IS_COBALT)
   // Checks if any previously hung threads have recovered. If all threads have
   // resumed, it notifies the delegate and clears the active hang UUID.
```
vincentmo, line 417: Should this be updated to include long hangs and long hang kills too?

base/threading/hang_watcher.cc
```diff
@@ -1242,6 +1285,57 @@ void HangWatcher::CheckAndRecordHangRecovered() {
...
+      LOG(INFO) << "Freeze detection: Long hang detected! Emitting UMA metric.";
+      UMA_HISTOGRAM_BOOLEAN("HangWatcher.LongHang.Detected", true);
+      long_hang_logged_ = true;
+    }
+
+    if (g_enable_long_hang_kill.load(std::memory_order_relaxed)) {
+      if (force_kill_closure_for_testing_) {
+        force_kill_closure_for_testing_.Run();
+      } else {
+#if BUILDFLAG(IS_STARBOARD)
```
vincentmo, line 1331: For my understanding, when is IS_STARBOARD defined?

base/threading/hang_watcher.cc
```diff
@@ -1242,6 +1285,57 @@ void HangWatcher::CheckAndRecordHangRecovered() {
...
+        auto* crash_ext = static_cast<const CobaltExtensionCrashHandlerApi*>(
+            SbSystemGetExtension(kCobaltExtensionCrashHandlerName));
+        if (crash_ext && crash_ext->version >= 2 && crash_ext->SetString) {
```
vincentmo, line 1334: For my understanding, what would happen if the extension version check wasn't in place or if it checked an incorrect version?

cobalt/browser/hang_watcher_delegate_impl.cc
alunev: I've remove it. It belongs to the next pr with finch flags so that params can be loaded from finch in case global features are not used.

cobalt/browser/hang_watcher_delegate_impl.h
alunev: I'll change this to bool similar to `bool IsHangReportingEnabled() override;` in the next pr along with adding finch flags for remaining params.

cobalt/browser/hang_watcher_delegate_impl.cc
alunev: yes, it's needed to read early set parameters like this: 
```
cobalt_loader.py \
  --enable-h5vcc-psettings="EnableHangWatcherLongHangDetection=1;EnableHangWatcherLongHangKill=1;LongHangTimeoutSeconds=10" 
```

base/threading/hang_watcher.cc
alunev: yes, they are noops in that case, though I think it's cleaner and less surprises when flags are read as they are configured. I think this also works correctly in some corner cases, e.g. if someone flips flag back and forth during testing.

base/threading/hang_watcher.cc
alunev: ok

base/threading/hang_watcher.cc
alunev: it's defined when compiling for starboard platforms (like 3P partner devices). it guards SB api call below. All starboard related changes are wrapped like this to make sure we compile them out on non-sb platforms.

base/threading/hang_watcher.cc
alunev: I bet segfault may happen since the SetString may be missing from [CobaltExtensionCrashHandlerApi](https://github.com/youtube/cobalt/blob/main/starboard/extension/crash_handler.h#L49) of lower version.

hlwarriner (approved):

base/threading/hang_watcher.h
alunev: ok

base/threading/hang_watcher.cc
vincentmo: Ah, good to know. Thanks.

vincentmo (approved):

PR #12070, opened 2026-08-19
player: apply volume directly on the audio sink

denisyuji:
When a video is cast from a phone and the user changes the volume, Cobalt shows the volume overlay but the audio level does not change. `SbPlayerSetVolume` only set the volume on the playbin, which forwards it to the audio sink only when that sink implements the GstStreamVolume interface.  The sink in use exposes a 'volume' property **without implementing the interface**, so the requested value was applied to nothing while `SbPlayerGetInfo` read it back and reported it as effective.

Implement the volume control in the GStreamer pipeline, tracking such a sink when it is set up and writing the level to its 'volume' property. 

Bug: 534354811

### DIFF
--- a/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
+++ b/starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
@@ -1655,6 +1655,8 @@ class PlayerImpl : public Player {
   std::mutex source_setup_mutex_;
   std::condition_variable source_setup_condition_;
   std::mutex seek_mutex_;
+  GstElement* audio_sink_{nullptr};
+  double volume_{1.0};
   double rate_{1.0};
   int ticket_{SB_PLAYER_INITIAL_TICKET};
   mutable GstClockTime seek_position_{GST_CLOCK_TIME_NONE};
@@ -1937,6 +1939,9 @@ PlayerImpl::~PlayerImpl() {
     DispatchOnWorkerThread(new PlayerDestroyedTask(
       player_status_func_, player_, ticket_, context_, main_loop_));
     playback_thread_.join();
+  }
+  if (audio_sink_) {
+    gst_object_unref(audio_sink_);
   }
   if (audio_caps_) {
     gst_caps_unref(audio_caps_);
@@ -2450,6 +2455,20 @@ void PlayerImpl::SetupElement(GstElement* pipeline,
       }
     }
 
+    // playsink only routes volume to GstStreamVolume sinks; set the property directly.
+    if (strstr(klass_str, "Audio") &&
+        !GST_IS_STREAM_VOLUME(element) &&
+        g_object_class_find_property(oclass, "volume")) {
+      std::lock_guard lock(self->mutex_);
+      if (self->audio_sink_ != element) {
+        gst_object_ref(element);
+        if (self->audio_sink_)
+          gst_object_unref(self->audio_sink_);
+        self->audio_sink_ = element;
+      }
+      g_object_set(G_OBJECT(element), "volume", self->volume_, nullptr);
+    }
+
     if (g_object_class_find_property(oclass, "has-drm")) {
       g_object_set(G_OBJECT(element), "has-drm", self->drm_system_ != nullptr, nullptr);
     }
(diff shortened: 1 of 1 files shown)


starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -1557,6 +1560,7 @@ PlayerImpl::~PlayerImpl() {
       player_status_func_, player_, ticket_, context_, main_loop_));
     playback_thread_.join();
   }
+  gst_clear_object(reinterpret_cast<GstObject**>(&audio_sink_));
```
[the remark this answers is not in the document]
denisyuji: fixed

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -1976,6 +1980,18 @@ void PlayerImpl::SetupElement(GstElement* pipeline,
...
 
+    // 'playsink' only routes volume to sinks implementing GstStreamVolume, so
+    // sinks that merely expose a 'volume' property (i.e. amlhalasink) need it
+    // to be set directly.
+    if (strstr(klass_str, "Audio") &&
+        !GST_IS_STREAM_VOLUME(element) &&
+        g_object_class_find_property(oclass, "volume")) {
+      std::lock_guard lock(self->audio_sink_mutex_);
+      gst_object_replace(reinterpret_cast<GstObject**>(&self->audio_sink_),
+                         GST_OBJECT(element));
```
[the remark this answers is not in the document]
denisyuji: fixed

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -2417,6 +2437,11 @@ void PlayerImpl::SetVolume(double volume) {
   GST_DEBUG_OBJECT(pipeline_, "volume %lf, TID %d", volume, gettid());
   gst_stream_volume_set_volume(GST_STREAM_VOLUME(pipeline_),
                                GST_STREAM_VOLUME_FORMAT_LINEAR, volume);
+
+  std::lock_guard lock(audio_sink_mutex_);
```
santoshmahto7: You can remove `audio_sink_mutex_` and directly use `mutex_` as
```
  std::lock_guard lock(mutex_);
```
and also at  other places

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
```diff
@@ -1295,6 +1295,9 @@ class PlayerImpl : public Player {
   std::mutex mutex_;
   std::mutex source_setup_mutex_;
   std::mutex seek_mutex_;
+  std::mutex audio_sink_mutex_;
+  GstElement* audio_sink_{nullptr};
+  double volume_{1.0};
```
santoshmahto7, line 1659: 1.0 is full volume; the initial volume should be 0.0

denisyuji: @santoshmahto7 , in my setup, after pressing vol+/-, the volume bar starts in 10/10 (full volume, normal level - the same level when no volume control is implemented) and then I can reduce it. Does the bar start in 0.0 (empty) in your setup?

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
denisyuji: ok, removed `audio_sink_mutex_` and using `mutex_`

starboard/contrib/rdk/src/third_party/starboard/rdk/shared/player/player_internal.cc
santoshmahto7: same here, the bar starts at 10.0. And 10.0 means the card's current volume level (set by the remote).
i.e [10, 0] == [current device volume, 0], so you can only decrease it. So you are right with 1.0.
C25 also have same behavior

santoshmahto7 (approved):

niranjanyardi (approved):
look reasonable.

@jonastsai feel free to review more

PR #12080, opened 2026-08-19
andriod: Support audio focus for Android N

jasonzhangxx:
Chromium m138 dropped support for Android N, causing the removal of
legacy audio focus API calls. Since Cobalt continues to support
Android N, this change restores the compatibility logic in
AudioFocusDelegate to ensure audio focus can be properly requested
and abandoned on older Android versions.

This change partially reverted https://chromium-review.git.corp.google.com/c/chromium/src/+/5985495.

Bug: 549235758

### DIFF
--- a/content/public/android/java/src/org/chromium/content/browser/AudioFocusDelegate.java
+++ b/content/public/android/java/src/org/chromium/content/browser/AudioFocusDelegate.java
@@ -8,6 +8,7 @@ import android.content.Context;
 import android.media.AudioAttributes;
 import android.media.AudioFocusRequest;
 import android.media.AudioManager;
+import android.os.Build;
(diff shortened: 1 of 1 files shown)


borongc:
This is top 1 crash at 7.18.303 right now, so we need to fix it. cc @johnxwork on how we should maintain Java code for this divergence from upstream.

johnxwork (approved):

jasonzhangxx:
> This is top 1 crash at 7.18.303 right now, so we need to fix it. cc @johnxwork on how we should maintain Java code for this divergence from upstream.

Discussed with @johnxwork and added a comment to explicitly mention the customized change for Cobalt.

PR #12085, opened 2026-08-19
Update DevTools extension description to reflect cross-platform telemetry links

haozheng-cobalt:
Update the memory footprint description in the DevTools extension panel
to include references to cross-platform production telemetry. This
change adds a link to Evergreen memory metrics to complement the
existing Kimono metrics reference.

Bug: 549295573

### DIFF
--- a/cobalt/tools/devtools_extension/panel.html
+++ b/cobalt/tools/devtools_extension/panel.html
@@ -34,9 +34,8 @@
           <div id="sampling-progress" class="sampling-badge">0 / 12 Reported</div>
         </div>
         <div class="section-instruction">
-          Aggregated median (P50) memory footprint across continuous native and managed allocators since launch,
-          mirroring production Kimono / PLX telemetry (<code>go/kimono-memory-metrics</code>).
-          Demonstrates physical memory distribution beyond the JavaScript heap.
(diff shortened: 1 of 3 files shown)


cobalt/tools/devtools_extension/panel.html
```diff
@@ -34,9 +34,8 @@
           <div id="sampling-progress" class="sampling-badge">0 / 12 Reported</div>
         </div>
         <div class="section-instruction">
-          Aggregated median (P50) memory footprint across continuous native and managed allocators since launch,
-          mirroring production Kimono / PLX telemetry (<code>go/kimono-memory-metrics</code>).
-          Demonstrates physical memory distribution beyond the JavaScript heap.
+          Session steady-state median (P50) of total physical CPU RAM (RSS) broken down across native C++, V8, graphics, and system allocators, continuously updated in real time since app launch.
```
Awallky, line 37: Some of these are Android-specific, e.g. AshMemJit. We should at least call this out, if not add a filter for platform type in the tool itself to toggle the different platform-specific metrics.

Awallky (approved):
LGTM % comment

cobalt/tools/devtools_extension/panel.html
haozheng-cobalt: Good idea, I will dynamically render subcomponent based on different platforms in another PR.

PR #12095, opened 2026-08-20
android: fix dual activity leaks

johnxwork:
### Summary
Scopes `CobaltService` lifecycle to native `RenderFrameHost` / `DocumentService` handles and fixes memory leaks during Android activity transitions (b/528367182).

### Changes
1. **Handle-Keyed Services**: Key services by `long nativeService` handle in `BaseStarboardBridge` to prevent collisions and leaks across documents, iframes, or overlapping activities.
2. **Deterministic Teardown**: Update native `Close()` to delegate teardown by handle in the correct order (`remove` -> `afterStopped` -> `onClose`) and unconditionally free private struct state.
3. **Symmetric Lifecycle Gating**: Symmetrically gate `beforeStartOrResume()` on 0 -> 1 started activities and `beforeSuspend()` on 1 -> 0, backed by a single `WeakHashMap` in the bridge.
4. **Safety Net**: Clean up orphaned services in `onActivityDestroy` if all activities are destroyed.
5. **Unit Tests**: Added `BaseStarboardBridgeTest` testing handle-keyed registration, lifecycle gating, teardown order, and invalid state handling.

### Device Testing (Google TV Streamer)
Simulated full overlapping activity handover (`MainActivity.start -> SecondActivity.start -> MainActivity.stop -> MainActivity.destroy -> SecondActivity browsing & video playback`):
* **`origin/main` Baseline**:
  * Crashed immediately on `SecondActivity.onCreate` due to `MemoryPressureUma.initializeInstance` asserting `assert sInstance == null;`.
  * Even when isolating beyond the crash, `MainActivity.onStop` prematurely suspended the app and decommitted renderer memory, and `MainActivity.onDestroy` destroyed `VideoSurfaceView`, nulling `sCurrentSurface`. Subsequent playback failed with `"Video surface is not available now"` (pitch-black screen).
* **With This PR**:
  * Second activity created cleanly without singleton collisions.
  * Overlapping stop/destroy gracefully deferred suspend and ignored stale surface destruction.
  * DPAD navigation and 4K hardware video playback on `SecondActivity` worked smoothly without errors.

Bug: 528367182
TAG=agy
CONV=9d9cb097-4bbf-46c5-b2fc-766467028e99

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
@@ -82,6 +82,11 @@ public abstract class BaseCobaltActivity extends Activity {
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
+    BaseStarboardBridge bridge = getStarboardBridge();
+    if (bridge != null) {
+      bridge.onActivityCreate(this);
+    }
+
     // Record the application start timestamp.
     mTimeInNanoseconds = System.nanoTime();
 
@@ -107,8 +112,17 @@ public abstract class BaseCobaltActivity extends Activity {
 
   @Override
   protected void onDestroy() {
-    super.onDestroy();
-    getStarboardBridge().onActivityDestroy(this);
+    try {
+      super.onDestroy();
+    } catch (Throwable t) {
+      Log.e(TAG, "Error in super.onDestroy()", t);
+      throw t;
+    } finally {
+      BaseStarboardBridge bridge = getStarboardBridge();
+      if (bridge != null) {
+        bridge.onActivityDestroy(this);
+      }
+    }
   }
 
   @Override
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
@@ -29,9 +29,12 @@ import android.media.AudioDeviceInfo;
 import android.media.AudioManager;
 import android.os.Build;
 import android.view.InputDevice;
+import android.view.Surface;
 import android.view.accessibility.CaptioningManager;
 import androidx.annotation.Nullable;
+import androidx.annotation.VisibleForTesting;
 import dev.cobalt.media.AudioOutputManager;
+import dev.cobalt.media.VideoSurfaceView;
 import dev.cobalt.util.DisplayUtil;
 import dev.cobalt.util.Holder;
 import dev.cobalt.util.Log;
@@ -39,11 +42,14 @@ import java.io.IOException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
+import java.util.WeakHashMap;
 import java.util.concurrent.ConcurrentHashMap;
+import org.chromium.base.CommandLine;
 import org.jni_zero.CalledByNative;
 import org.jni_zero.JNINamespace;
 import org.jni_zero.NativeMethods;
@@ -54,6 +60,17 @@ import org.jni_zero.NativeMethods;
 // this property.
 public class BaseStarboardBridge {
 
+  private static volatile BaseStarboardBridge sInstance;
+
+  public static BaseStarboardBridge getInstance() {
+    return sInstance;
+  }
+
+  @VisibleForTesting
+  public static void setInstanceForTesting(BaseStarboardBridge bridge) {
+    sInstance = bridge;
+  }
+
   /** Interface to be implemented by the Android Application hosting the starboard app. */
   public interface HostApplication {
     void setStarboardBridge(BaseStarboardBridge starboardBridge);
@@ -61,6 +78,8 @@ public class BaseStarboardBridge {
     BaseStarboardBridge getStarboardBridge();
   }
 
+  private Surface mVideoSurface;
+  private final Object mVideoSurfaceLock = new Object();
   private CobaltSystemConfigChangeReceiver mSysConfigChangeReceiver;
   private CobaltTextToSpeechHelper mTtsHelper;
   // TODO(cobalt): Re-enable these classes or remove if unnecessary.
@@ -71,6 +90,10 @@ public class BaseStarboardBridge {
   private final Context mAppContext;
   protected final Holder<Activity> mActivityHolder;
   private final Holder<Service> mServiceHolder;
+  // Maps each live Activity to whether it is currently started (Boolean.TRUE) or stopped (Boolean.FALSE).
+  // Backed by WeakHashMap to prevent pinning Activity instances in the singleton bridge.
+  private final Map<Activity, Boolean> mActivities =
+      Collections.synchronizedMap(new WeakHashMap<>());
   private final String[] mArgs;
   private final long mNativeApp;
   private final Runnable mStopRequester =
@@ -94,7 +117,9 @@ public class BaseStarboardBridge {
   private long mAppStartTimestamp = 0;
 
   private final Map<String, CobaltService.Factory> mCobaltServiceFactories = new HashMap<>();
-  private final Map<String, CobaltService> mCobaltServices = new ConcurrentHashMap<>();
+  // Keyed by native pointer handle so multiple documents, iframes, or overlapping Activities
+  // can open services with the same name without colliding or leaking on close.
+  private final Map<Long, CobaltService> mCobaltServices = new ConcurrentHashMap<>();
 
   private static final String GOOGLE_PLAY_SERVICES_PACKAGE = "com.google.android.gms";
   private static final String AMATI_EXPERIENCE_FEATURE =
@@ -104,6 +129,7 @@ public class BaseStarboardBridge {
   private final long mTimeNanosecondsPerMicrosecond = 1000;
   private static final String YTS_CERT_SCOPE_SYSTEM_PROPERTY = "ro.vendor.youtube.cert_scope";
   private static final String DEFAULT_DEVICE_NAME = "Android";
+  private final Natives mNatives = BaseStarboardBridgeJni.get();
 
   /**
    * Lightweight constructor used by test activities (e.g. CobaltTestActivity).
@@ -133,10 +159,11 @@ public class BaseStarboardBridge {
       String startDeepLink) {
 
     Log.i(TAG, "BaseStarboardBridge init.");
+    sInstance = this;
 
     // Make sure the JNI stack is properly initialized first as there is a
     // race condition as soon as any of the following objects creates a new thread.
-    BaseStarboardBridgeJni.get().initJNI(this);
+    mNatives.initJNI(this);
 
     mAppContext = appContext;
     mActivityHolder = activityHolder;
@@ -151,19 +178,17 @@ public class BaseStarboardBridge {
     mIsAmatiDevice = appContext.getPackageManager().hasSystemFeature(AMATI_EXPERIENCE_FEATURE);
 
     mNativeApp =
-        BaseStarboardBridgeJni.get()
-            .startNativeStarboard(
-                getAssetsFromContext(),
-                getFilesCanonicalPath(),
-                getCacheCanonicalPath(),
-                getNativeLibraryDir());
+        mNatives.startNativeStarboard(
+            getAssetsFromContext(),
+            getFilesCanonicalPath(),
+            getCacheCanonicalPath(),
+            getNativeLibraryDir());
 
-    BaseStarboardBridgeJni.get().handleDeepLink(startDeepLink, /* applicationStarted= */ false);
-    BaseStarboardBridgeJni.get().setAndroidBuildFingerprint(getBuildFingerprint());
-    BaseStarboardBridgeJni.get().setAndroidOSExperience(mIsAmatiDevice);
-    BaseStarboardBridgeJni.get().setAndroidPlayServicesVersion(getPlayServicesVersion());
-    BaseStarboardBridgeJni.get()
-        .setYoutubeCertificationScope(getSystemProperty(YTS_CERT_SCOPE_SYSTEM_PROPERTY));
+    mNatives.handleDeepLink(startDeepLink, /* applicationStarted= */ false);
+    mNatives.setAndroidBuildFingerprint(getBuildFingerprint());
+    mNatives.setAndroidOSExperience(mIsAmatiDevice);
+    mNatives.setAndroidPlayServicesVersion(getPlayServicesVersion());
+    mNatives.setYoutubeCertificationScope(getSystemProperty(YTS_CERT_SCOPE_SYSTEM_PROPERTY));
   }
 
   @NativeMethods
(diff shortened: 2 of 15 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java
```diff
@@ -22,12 +22,49 @@
...
+ *       Activity lifecycle callbacks.</li>
+ *   <li>{@link #sendToClient} may be invoked from any thread; synchronization ensures safety
+ *       against concurrent {@link #onClose}.</li>
+ * </ul>
+ */
 public abstract class CobaltService {
   // Indicate is the service opened, and be able to send data to client
   protected boolean opened = true;
+  private String mServiceName;
+  private long mNativeService;
```
Awallky: Mark `mNativeService` and `mServiceName` as volatile to guarantee memory visibility and atomic 64-bit reads across background threads on 32-bit ARM JVMs. A write to 64-bit words on 32-bit architectures is treated as two separate, non-atomic writes. This is a race condition without the volatile keyword.
See https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.7.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -40,10 +40,13 @@
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
```
Awallky: Is this import used anywhere?

Awallky (approved):
//cobalt LGTM

cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java
```diff
@@ -89,8 +89,10 @@ public void surfaceChanged(SurfaceHolder holder, int format, int width, int heig
 
     @Override
     public void surfaceDestroyed(SurfaceHolder holder) {
-      sCurrentSurface = null;
-      VideoSurfaceViewJni.get().onVideoSurfaceChanged(sCurrentSurface);
+      if (sCurrentSurface == holder.getSurface()) {
+        sCurrentSurface = null;
+        VideoSurfaceViewJni.get().onVideoSurfaceChanged(null);
```
borongc: As we only have a single VideoSurface for Starboard, `sCurrentSurface` should be always the case of `holder.getSurface()`, we might not need this if statement to check it.

borongc: `sCurrentSurface` is owned by each VideoSurfaceView per CobaltActivity. If we have 2nd CobaltActivity created, and the 1st VideoSurfaceView owned by the pervious CobaltActivity is destroyed, `VideoSurfaceViewJni.get().onVideoSurfaceChanged` will still be called anyway.

johnxwork: yeah what you described is exactly why I added the if condition. If the first activity gets destroyed after the second activity starts, we don't want to set the videoSurface to null.

Or do you mean we should do it in a different way?

borongc: Right, we don't want to set the videoSurface to null if the 1st activity is destroyed after the 2nd activity is created.

However, VideoSurfaceView is created and owned by each class, so when the 1st activity is destroyed, `sCurrentSurface == holder.getSurface()` is true.

I think for this, two created activities don't know each other, and their VideoSurfaceViews, so `VideoSurfaceViewJni.get().onVideoSurfaceChanged(null);` will be called anyway.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
```diff
@@ -107,8 +109,11 @@ protected void onStop() {
 
   @Override
   protected void onDestroy() {
-    super.onDestroy();
-    getStarboardBridge().onActivityDestroy(this);
+    try {
+      super.onDestroy();
```
kjyoun, line 116: do we need to log somethign, when onDestroy() throws exception?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -133,10 +137,11 @@ public BaseStarboardBridge(
       String startDeepLink) {
 
     Log.i(TAG, "BaseStarboardBridge init.");
+    sInstance = this;
 
     // Make sure the JNI stack is properly initialized first as there is a
     // race condition as soon as any of the following objects creates a new thread.
-    BaseStarboardBridgeJni.get().initJNI(this);
+    getNatives().initJNI(this);
```
kjyoun: instead of calling getNatives() repeatedly, how about caching it to local variable?

We can avoid any hidden overhead of getNatives() call.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -634,72 +741,121 @@ public void setVideoSurfaceBounds(final int x, final int y, final int width, fin
...
     return service;
   }
 
   public CobaltService getOpenedCobaltService(String serviceName) {
-    return mCobaltServices.get(serviceName);
+    CobaltService matchedService = null;
+    // Count matching instances across open handles to detect and warn if name-based lookup is ambiguous.
+    int count = 0;
+    for (CobaltService service : mCobaltServices.values()) {
+      if (serviceName.equals(service.getServiceName())) {
```
kjyoun: flatten logic?

```
if (!serviceName.equals(service.getServiceName())) {
   continue;
}

count++;
f (matchedService != null) {      
  continue;
}

matchedService = service;
```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -194,40 +198,116 @@ long startNativeStarboard(
...
+      mActivityHolder.set(null);
+    }
+    if (hasLiveActivities()) {
+      Log.i(
+          TAG,
+          "Activity destroyed but another activity instance is still live; skipping exit check.");
       return;
     }
-    closeAllCobaltService();
+    synchronized (this) {
```
kjyoun: go/java-practices/concurrency discourage `synchronized(this)`

> To prevent this denial-of-service attack, you can use a private lock object instead of synchronized methods (which imply a publicly accessible lock
and so on

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -194,40 +198,116 @@ long startNativeStarboard(
...
+      if (mVideoSurface != null) {
+        mVideoSurface = null;
+        VideoSurfaceView.notifyVideoSurfaceChanged(null);
```
kjyoun, line 331: you are calling callback while acquired lock, which is anti-pattern

go/java-practices/concurrency

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -616,6 +696,33 @@ void onRequestPermissionsResult(int requestCode, String[] permissions, int[] gra
     mAudioPermissionRequester.onRequestPermissionsResult(requestCode, permissions, grantResults);
   }
 
+  public synchronized void onVideoSurfaceCreated(Surface surface) {
```
kjyoun: synchroized methods acquires locks on "this" object, which is anti pattern.

we should narrow down the scope of lock

Ideally code should be like this

```
public  void onVideoSurfaceCreated(Surface surface) {
   Log.i(TAG, String.format("onVideoSurfaceCreated: surface=%s", surface));

    // miminal scope for lock
    synchromized(mLock) {
      mVideoSurface = surface;
   }
   
   // calls callback outside of lock
   VideoSurfaceView.notifyVideoSurfaceChanged(surface);
  }
 ```

File: cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java [hunk 10]
```diff
@@ -616,6 +696,33 @@ void onRequestPermissionsResult(int requestCode, String[] permissions, int[] gra
...
+    Log.i(TAG, String.format("onVideoSurfaceCreated: surface=%s", surface));
+    mVideoSurface = surface;
+    VideoSurfaceView.notifyVideoSurfaceChanged(surface);
+  }
+
+  public synchronized void onVideoSurfaceDestroyed(Surface surface) {
```
kjyoun: ditto: no synchronized on method

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -616,6 +696,33 @@ void onRequestPermissionsResult(int requestCode, String[] permissions, int[] gra
...
+              "onVideoSurfaceDestroyed: Ignoring destroyed surface %s from stale activity;"
+                  + " current surface is %s",
+              surface, mVideoSurface));
+      return;
+    }
+    mVideoSurface = null;
+    VideoSurfaceView.notifyVideoSurfaceChanged(null);
+  }
+
+  public synchronized Surface getVideoSurface() {
```
kjyoun: ditto: no synchronized on method

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
kjyoun: or

How about defining `final` member variable and use it.
then you dont need to call `getNatives` again in this intance.

    private final Natives mNatives = getNatives();

kjyoun (commented):
nit: can you add prefix in commit message?

e.g.) android: Manage...

https://screenshot-v2.corp.google.com/5g8zpUgDxsA7vCB

cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java
johnxwork: Just modified according to our discussed solution. I didn't move ownership to StarboardBridge because SurfaceView is a View and View should be owned by Activity in principle. But I did make sure that StarboardBridge checks for the correct view before resetting the surfaceView, thus preventing a potential blackscreen.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
johnxwork: Removed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltService.java
johnxwork: Thanks! Added volatile.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java
```diff
@@ -33,11 +36,27 @@
 @JNINamespace("starboard")
 public class VideoSurfaceView extends SurfaceView {
   @NativeMethods
```
kjyoun, line 39: @VisibleForTesting(otherwise = VisibleForTesting.PACKAGE_PRIVATE) ?

looks like this is just for testing.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -616,6 +701,43 @@ void onRequestPermissionsResult(int requestCode, String[] permissions, int[] gra
     mAudioPermissionRequester.onRequestPermissionsResult(requestCode, permissions, grantResults);
   }
 
+  public void onVideoSurfaceCreated(Surface surface) {
+    Log.i(TAG, String.format("onVideoSurfaceCreated: surface=%s", surface));
+    synchronized (mVideoSurfaceLock) {
+      mVideoSurface = surface;
```
borongc, line 750: This assumes Cobalt only keeps the latest activity, we probably need to add some comments here.

Also, is it possible to gate this PR by a flag?

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
```diff
@@ -76,12 +76,14 @@ static String[] getCommandLineParamsFromIntent(Intent intent, String key) {
...
   protected abstract BaseStarboardBridge createStarboardBridge(String[] args, String startDeepLink);
 
-  protected BaseStarboardBridge getStarboardBridge() {
+  public BaseStarboardBridge getStarboardBridge() {
     return ((BaseStarboardBridge.HostApplication) getApplication()).getStarboardBridge();
   }
 
   @Override
   protected void onCreate(Bundle savedInstanceState) {
+    getStarboardBridge().onActivityCreate(this);
```
kjyoun: can you check this claim of Gemini is correct?


---

  While checking this exact lifecycle, notice a crash hazard introduced in BaseCobaltActivity.java:84-86:

  In BaseCobaltActivity.java:

    @Override
    protected void onCreate(Bundle savedInstanceState) {
      getStarboardBridge().onActivityCreate(this); // <-- CRASH on Cold Start!
      ...
    }

  Look at the execution order in CobaltActivity.java:466-470:

    @Override
    protected void onCreate(Bundle savedInstanceState) {
      super.onCreate(savedInstanceState); // 1. Runs BaseCobaltActivity.onCreate()

      setupStartupGuard();
      createContent(savedInstanceState);  // 2. ONLY HERE is StarboardBridge created on Cold Start!

  • On a cold start, when super.onCreate() is called, createContent() hasn't run yet.
  • At that moment, getApplication().getStarboardBridge() returns null!
  • BaseCobaltActivity.java:85 calling getStarboardBridge().onActivityCreate(this) will throw a NullPointerException before the app even
  reaches createContent() or VideoSurfaceView!

cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java
```diff
@@ -68,14 +87,30 @@ private void initialize(Context context) {
...
+
   private class SurfaceHolderCallback implements SurfaceHolder.Callback {
 
     boolean mSawInitialChange = false;
 
     @Override
     public void surfaceCreated(SurfaceHolder holder) {
-      sCurrentSurface = holder.getSurface();
-      VideoSurfaceViewJni.get().onVideoSurfaceChanged(sCurrentSurface);
+      BaseStarboardBridge bridge = getStarboardBridge();
```
kjyoun: IIUC, in a running Cobalt app, `getStarboardBridge()` is always non-null once the activity reaches view creation.

How about this?

    ```java
    @NonNull private final BaseStarboardBridge mBridge = getStarboardBridge();

    @Override
    public void surfaceCreated(SurfaceHolder holder) {
      mBridge.onVideoSurfaceCreated(holder.getSurface());
    }

    @Override
    public void surfaceDestroyed(SurfaceHolder holder) {
      mBridge.onVideoSurfaceDestroyed(holder.getSurface());
    }

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -525,7 +530,8 @@ private void initializeJavaBridge() {
    */
   protected abstract StarboardBridge createStarboardBridge(String[] args, String startDeepLink);
 
-  protected StarboardBridge getStarboardBridge() {
+  @Override
+  public StarboardBridge getStarboardBridge() {
```
kjyoun: I think that we can keep `protected` here

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
```diff
@@ -76,12 +76,14 @@ static String[] getCommandLineParamsFromIntent(Intent intent, String key) {
    */
   protected abstract BaseStarboardBridge createStarboardBridge(String[] args, String startDeepLink);
 
-  protected BaseStarboardBridge getStarboardBridge() {
+  public BaseStarboardBridge getStarboardBridge() {
```
kjyoun: I think that you can keep `protected` here(`BaseCobaltActivity::getStarboardBridge`) and `CobaltyActivity::getStarboardBridge`

You only have to change `VideoSurfaceView::getStarboardBridge` to use `CobaltApplication::getStarboardBridge`

`CobaltApplication` already has public `getStarboardBridge`

https://source.corp.google.com/h/github/youtube/cobalt/+/main:cobalt/android/apk/app/src/app/java/dev/cobalt/app/CobaltApplication.java;l=38;drc=6ed484f8a3910fc5c0f5dfba58a3a9045dbd6ea4

e.g.)

```
    private BaseStarboardBridge getStarboardBridge() {
      Context appContext = getContext().getApplicationContext();
      if (appContext instanceof BaseStarboardBridge.HostApplication) {
        return ((BaseStarboardBridge.HostApplication) appContext).getStarboardBridge();
      }
      return BaseStarboardBridge.getInstance();
    }
```

kjyoun (commented):
super nit: capitalize first character

`android: Fix...`

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
johnxwork: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java
johnxwork: Done: Added @VisibleForTesting(otherwise = VisibleForTesting.PACKAGE_PRIVATE) and cached mBridge in VideoSurfaceView for use in surfaceCreated / surfaceDestroyed

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseCobaltActivity.java
johnxwork: I've added a null guard in BaseCobaltActivity.onCreate(), and registered the activity via starboardBridge.onActivityCreate(this) inside CobaltActivity.createContent() right after the bridge is instantiated.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
johnxwork: Added comment.

I'll try to add a flag for this

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -194,40 +200,166 @@ long startNativeStarboard(
     boolean isDevelopmentBuild();
   }
 
+  private static Natives sNatives;
+
+  public static void setNativesForTesting(Natives natives) {
+    sNatives = natives;
+  }
+
+  private static Natives getNatives() {
```
kjyoun: with https://github.com/youtube/cobalt/pull/12474 merged, 

I think we dont need this method any more. 

Now methods that actually needs it (e..g. isReleaseBuild) are deleted
We can initialize `mNatives` directly

```
private final Natives mNatives = BaseStarboardBridgeJni.get();

```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -194,40 +200,166 @@ long startNativeStarboard(
     boolean isDevelopmentBuild();
   }
 
+  private static Natives sNatives;
```
kjyoun: How about his pattern?

Then we dont need sNatives and setNativesForTesting()

```
  // Public constructor used by production
  public BaseStarboardBridge (....) {
    this(....., BaseStarboardBridgeJni.get());
  }

  // Package-private constructor for unit tests
  @VisibleForTesting
  BaseStarboardBridge (.....Natives natives) {
    .... // 
    this.mNatives = natives;
  }


// On test
BaseStarboardBridge (....., mockNatives)





```

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
johnxwork: Thanks. Removed sNatives / setNativesForTesting / getNatives()

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
johnxwork: Done

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -190,40 +215,153 @@ long startNativeStarboard(
...
+
+  protected void onActivityCreate(Activity activity) {
+    Log.i(TAG, "onActivityCreate ran: " + activity);
+    mActivities.put(activity, Boolean.FALSE);
+  }
+
   protected void onActivityStart(Activity activity) {
-    Log.i(TAG, "onActivityStart ran");
+    Log.i(TAG, "onActivityStart ran: " + activity);
+    synchronized (mActivities) {
```
borongc: If this can simply the code:

```
int active_activity_count = countStartedActivities();
if (active_activity_count) {
  Log.w(
              TAG,
              String.format(
                  "onActivityStart: New activity %s starting while there are previous %d activities are still active.",
                  activity, active_activity_count));
}
```

Unless you need to log the pointer of the previous activity.

cobalt/android/apk/app/src/main/java/dev/cobalt/media/VideoSurfaceView.java
```diff
@@ -68,14 +79,33 @@ private void initialize(Context context) {
...
   private class SurfaceHolderCallback implements SurfaceHolder.Callback {
 
     boolean mSawInitialChange = false;
 
     @Override
     public void surfaceCreated(SurfaceHolder holder) {
-      sCurrentSurface = holder.getSurface();
-      VideoSurfaceViewJni.get().onVideoSurfaceChanged(sCurrentSurface);
+      BaseStarboardBridge bridge = getBridge();
+      if (bridge != null) {
```
borongc: While it should be safe if 2nd activity is created, `bridge` should not be null.

Just want to check if `bridge` could be null if the VideoSurfaceView of the 1st activity is created. I think we can ensure StarboardBridge is created before VideoSurfaceView in CobaltActivity? If so, we can simply the code here.

johnxwork: Done.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
johnxwork: Done. Thanks for the suggestions.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/BaseStarboardBridge.java
```diff
@@ -190,40 +215,154 @@ long startNativeStarboard(
...
+    Log.i(TAG, "onActivityCreate ran: " + activity);
+    mActivities.put(activity, Boolean.FALSE);
+  }
+
   protected void onActivityStart(Activity activity) {
-    Log.i(TAG, "onActivityStart ran");
+    Log.i(TAG, "onActivityStart ran: " + activity);
+    boolean wasZeroStarted;
+    synchronized (mActivities) {
+      int activeCount = countStartedActivities();
```
borongc: Dual `synchronized (mActivities)`. Another is inside `countStartedActivities()`.

johnxwork: Done. Thanks

borongc (approved):

kjyoun:
for the record, this PR introduced several format errors, which were detected in a PR on top of this

https://github.com/youtube/cobalt/pull/12529

PR #12102, opened 2026-08-20
Enable heap symbolization on RDK

haozheng-cobalt:
Update the in-process heap symbolizer to support RDK Evergreen
deployments. Standard address resolution fails on RDK because
Evergreen loads the core library into anonymous memory. This change
allows the symbolizer to resolve callstacks in these environments.

Also adjust build configurations to ensure frame pointers are
available in non-release builds to facilitate accurate unwinding.
The update includes new Python and C++ unit tests to verify the
symbolization pipeline.

Test: 
symbolized_trace_rdk.json: https://drive.google.com/file/d/1hXIFkUES6WSQSypiS70w2RYN3ddSjx4k/view?usp=drive_link
symbolized_trace_rdk_summary.json: https://drive.google.com/file/d/13zHitB_M8VZCkjFyICFHBO32pCkvVgZ6/view?usp=drive_link 
symbolized_trace_rdk_flamegraph.html: https://drive.google.com/file/d/1CQ0eliWBvnQkwou24Wge-_7v-8Baa2Hv/view?usp=drive_link 

Bug: 528397803


Awallky (commented):
How was this verified? Did you check the overall levels against `/proc/self/smaps_rollup` or anything else to ensure the tool's readings were accurate compared to an external  tool/reading?

I would hope that at least the measured RSS matches what is in the smaps_rollup.

haozheng-cobalt:
> How was this verified? Did you check the overall levels against `/proc/self/smaps_rollup` or anything else to ensure the tool's readings were accurate compared to an external tool/reading?
> 
> I would hope that at least the measured RSS matches what is in the smaps_rollup.

The change in this PR focuses on symbolizing a given trace and interpreting the data structure in the trace file. Verified by expecting certain keys in the json, and inspecting the demangled C++ callstacks in the outputs, of cuz with unittests. Also I have some test results listed in the PR description.

For verify the overall memory level, I think might not be under the scope of this PR. But sure, I will give a more solid evaluation.

Awallky (approved):

PR #12105, opened 2026-08-20
finch: Defer active config initialization until crash streak is synced

yjzhang111:
Move the initialization of active configuration data from the
GlobalFeatures constructor to CobaltContentBrowserClient. This allows
the initialization to be deferred until the experiment config type is
determined crash streak synchronization from beacon file to metrics
local state is processed.

Bug: 549905088

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -630,6 +630,14 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
   auto* global_features = GlobalFeatures::GetInstance();
   auto* experiment_config_manager =
       global_features->experiment_config_manager();
+
+  // It is critical that GetExperimentConfigType() is evaluated after
+  // InstantiateFieldTrialList(), because the latter triggers
+  // CleanExitBeacon::Initialize(), which reads the 'Variations' beacon file
+  // from DIR_CACHE and increments/syncs the crash streak into
+  // metrics_local_state.
+  // ExperimentConfigManager can then see the true crash streak and fall back
+  // to Safe Mode when needed.
   auto config_type = experiment_config_manager->GetExperimentConfigType();
   if (config_type == ExperimentConfigType::kEmptyConfig) {
     return;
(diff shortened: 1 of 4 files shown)


cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -646,6 +634,8 @@ void CobaltContentBrowserClient::SetUpCobaltFeaturesAndParams(
   if (config_type == ExperimentConfigType::kEmptyConfig) {
     return;
   }
+
+  global_features->InitializeActiveConfigData(config_type);
```
vincentmo, line 646: Could you add a comment about the importance of this being called after the beacon file has been synced? Also, for my understanding, when and where does the synchronization take place?

cobalt/browser/global_features.cc
```diff
@@ -190,11 +187,9 @@ void GlobalFeatures::CreateMetricsLocalState() {
...
 }
 
-void GlobalFeatures::InitializeActiveConfigData() {
+void GlobalFeatures::InitializeActiveConfigData(
+    ExperimentConfigType experiment_config_type) {
   DCHECK(experiment_config_);
-  DCHECK(experiment_config_manager_);
-  auto experiment_config_type =
-      experiment_config_manager_->GetExperimentConfigType();
   if (experiment_config_type == ExperimentConfigType::kEmptyConfig) {
```
vincentmo, line 193: Should active_config_data_ be cleared explicitly as well to prevent it from ever being out of sync?

cobalt/browser/global_features_unittest.cc
vincentmo, line 1: Is it possible to catch the beacon file synchronization bug that this fixing in unit test?

cobalt/browser/cobalt_content_browser_client.cc
yjzhang111: Done.

cobalt/browser/global_features.cc
yjzhang111: I think it can't be out of sync rn because InitializeActiveConfigData is the only place that modifies active_config_data_. But added to be future proof

cobalt/browser/global_features_unittest.cc
yjzhang111: it's hard because with unit tests, we need to inject a specific crash streak number. A browser test could potentially catch this since its a bug with initialization order. We don't have any browser tests right now but I can look into it after enabling integration tests.

vincentmo (approved):

PR #12111, opened 2026-08-21
rdk: Enforce unconditional extraction and cleanup natively on deploy

yuying-y:
This modifies `deploy_rdk.py` to unconditionally clean up previous remote build artifacts (like leftover `*.lz4` binaries) and directly extract the new `archive.tar.gz` payload aggressively on the device upon deployment, even when `--run` is omitted.
Leaving a compressed payload passively on a TV device without extracting it inherently causes subtle testing anti-patterns where old binaries are invoked by scripts. By extracting intrinsically upon a successful transfer, we ensure the new binaries are immediately staged and ready to launch (e.g. for memory benchmarks interacting via RPCs).

Bug: 537318952
Bug: 548055360

TAG=agy

### DIFF
--- a/.github/actions/depot_tools/action.yaml
+++ b/.github/actions/depot_tools/action.yaml
@@ -119,15 +119,6 @@ runs:
       if: inputs.run_sync == 'true'
       working-directory: cobalt/src
       run: |
-        # Remove after all workers have migrated to new siso directory
-        if [ -d "../.cipd" ]; then
-          if grep -r "infra/build/siso" "../.cipd" > /dev/null 2>&1; then
-            echo "Deleting conflicting siso package."
-            rm -rf "third_party/siso/cipd"
-            rm -rf "../.cipd"
-          fi
-        fi
-
         # Identify and delete dirty sub-repositories
         git -c core.quotepath=false status --porcelain -unormal | while IFS= read -r line; do
           path="${line:3}"
--- a/.github/config/evergreen-arm-hardfp-rdk.json
+++ b/.github/config/evergreen-arm-hardfp-rdk.json
@@ -15,8 +15,6 @@
   "targets": [
     "cobalt:gn_all",
     "cobalt/testing/browser_tests:cobalt_browsertests_loader",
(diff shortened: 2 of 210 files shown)


niranjanyardi:
Hey Ying, thanks for addressing the extraction gap! Rather than exposing this complexity via a new `--deploy-extract` CLI flag, it's safer and cleaner to just have the deployment script unconditionally extract the payload natively whenever an archive is successfully transferred (since a zipped payload passively residing on the test TV serves no purpose). 

I've taken the liberty of pushing a follow-up commit to your branch that strips out the flag and enforces extraction natively within `package_and_deploy()` instead. Feel free to review the changes!

niranjanyardi:
> This introduces a new flag `--deploy-extract` which explicitly cleans previous remote build artifacts (like leftover `*.lz4` binaries) and extracts the new `archive.tar.gz` directly on the device upon deployment, Today these steps are hidden within `launch_on_device()` and only runs when `--run` is passed. But when running rdk_memory_measurement, the measurement script takes over control of app launch, and a user may not pass `--run` to `deploy_rdk.py`, expecting the new image to be extracted and ready to run after running the deployment script. This PR decouples the cleanup + extraction from `--run`.
> 
> Bug: 537318952 Bug: 548055360
> 
> TAG=agy

can you please update the description ?

niranjanyardi (approved):
i think my AI agernt very enthusisasticaly added a commit to your change - feeel free to remove it and redo the PR .

My suggestion stands though - when you say "deploy" - you expect the new binaries to be presetn int he right place - extracted and ready to go.

if anything this will help other's workflows/ reduce errors in tesating

yuying-y:
> i think my AI agernt very enthusisasticaly added a commit to your change - feeel free to remove it and redo the PR .
> 
> My suggestion stands though - when you say "deploy" - you expect the new binaries to be presetn int he right place - extracted and ready to go.
> 
> if anything this will help other's workflows/ reduce errors in tesating

Sgtm

PR #12118, opened 2026-08-21
Flag Simplification for mmap Users

yuying-y:
Simplify mmap execution flags natively. If --loader_use_mmap_file is passed, let it default to natively requesting uncompressed updates from the updater, with not need to pass --use_uncompressed_updates, preventing users from needing to pass redundant parameters.

Bug: 537318952

### DIFF
--- a/cobalt/updater/configurator.cc
+++ b/cobalt/updater/configurator.cc
@@ -54,6 +54,7 @@ const char kUpdaterJSONDefaultUrl[] =
 // Whether to request, download, and install uncompressed (rather than
 // compressed) Evergreen binaries.
 constexpr char kUseUncompressedUpdates[] = "use_uncompressed_updates";
+constexpr char kLoaderUseMemoryMappedFile[] = "loader_use_mmap_file";
 
 // Uses the QA update server to test the changes to the configuration of the
 // PROD update server.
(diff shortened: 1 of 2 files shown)


yuying-y:
Thank you for the review! I have completely addressed both of these critical regressions in my latest force-push:

1. `configurator.cc`: The regex error that accidentally grouped the QA URL fallback with the memory mapping flag has been eliminated. The conditional has been restored.
2. `loader_app.cc`: The aggressive deletion that wiped out the Crashpad handler, slot manager, and `main()` loop has been entirely reverted, restoring full application lifecycle integrity.

cobalt/updater/configurator.cc
```diff
@@ -102,6 +103,8 @@ Configurator::Configurator(
     SetChannel(persisted_channel);
   }
   if (base::CommandLine::ForCurrentProcess()->HasSwitch(
+          kLoaderUseMemoryMappedFile) ||
```
hlwarriner, line 106: Have all stakeholders aligned on the decision to offer this option to partners? If so, this change LGTM.

hlwarriner (approved):

cobalt/updater/configurator.cc
yuying-y: Yep, got approved by Henry and Lam.

andrewsavage1 (approved):

PR #12119, opened 2026-08-21
starboard: Refactor DecodedAudio to plain object

xiaomings:
Refactor DecodedAudio from RefCountedThreadSafe to a plain object
passed via move semantics. This change eliminates reference counting
overhead and clarifies ownership throughout the media pipeline.

Updated Starboard player, filter, and platform implementations,
including Android, tvOS, FFmpeg, FDK AAC, and Opus, to pass
DecodedAudio by value or move.

Bug: 550523546

### DIFF
--- a/starboard/android/shared/audio_decoder_passthrough.h
+++ b/starboard/android/shared/audio_decoder_passthrough.h
@@ -15,6 +15,7 @@
 #ifndef STARBOARD_ANDROID_SHARED_AUDIO_DECODER_PASSTHROUGH_H_
 #define STARBOARD_ANDROID_SHARED_AUDIO_DECODER_PASSTHROUGH_H_
 
+#include <optional>
 #include <queue>
 #include <utility>
 
@@ -64,11 +65,11 @@ class AudioDecoderPassthrough : public AudioDecoder {
     //       mode on more platforms.
     const int kChannels = 1;
     for (const auto& input_buffer : input_buffers) {
-      scoped_refptr<DecodedAudio> decoded_audio =
-          new DecodedAudio(kChannels, kSbMediaAudioSampleTypeInt16Deprecated,
-                           kSbMediaAudioFrameStorageTypePlanar,
-                           input_buffer->timestamp(), input_buffer->size());
-      memcpy(decoded_audio->data(), input_buffer->data(), input_buffer->size());
+      DecodedAudio decoded_audio(
+          kChannels, kSbMediaAudioSampleTypeInt16Deprecated,
+          kSbMediaAudioFrameStorageTypePlanar, input_buffer->timestamp(),
+          input_buffer->size());
+      memcpy(decoded_audio.data(), input_buffer->data(), input_buffer->size());
       decoded_audios_.push(std::move(decoded_audio));
       output_cb_();
     }
@@ -80,18 +81,19 @@ class AudioDecoderPassthrough : public AudioDecoder {
     SB_CHECK(thread_checker_.CalledOnValidThread());
     SB_DCHECK(output_cb_);
 
-    decoded_audios_.push(new DecodedAudio);
+    decoded_audios_.push(DecodedAudio());
     output_cb_();
   }
 
-  scoped_refptr<DecodedAudio> Read(int* samples_per_second) override {
+  std::optional<DecodedAudio> Read(int* samples_per_second) override {
     SB_CHECK(thread_checker_.CalledOnValidThread());
     SB_DCHECK(samples_per_second);
     SB_DCHECK(!decoded_audios_.empty());
 
     *samples_per_second = samples_per_second_;
 
-    auto decoded_audio = std::move(decoded_audios_.front());
+    std::optional<DecodedAudio> decoded_audio =
+        std::move(decoded_audios_.front());
     decoded_audios_.pop();
     return decoded_audio;
   }
@@ -99,7 +101,7 @@ class AudioDecoderPassthrough : public AudioDecoder {
   void Reset() override {
     SB_CHECK(thread_checker_.CalledOnValidThread());
 
-    decoded_audios_ = std::queue<scoped_refptr<DecodedAudio>>();  // Clear
+    decoded_audios_ = std::queue<DecodedAudio>();  // Clear
   }
 
  private:
@@ -107,7 +109,7 @@ class AudioDecoderPassthrough : public AudioDecoder {
 
   const int samples_per_second_;
   OutputCB output_cb_;
-  std::queue<scoped_refptr<DecodedAudio>> decoded_audios_;
+  std::queue<DecodedAudio> decoded_audios_;
 };
 
 }  // namespace starboard
--- a/starboard/android/shared/audio_renderer_passthrough.cc
+++ b/starboard/android/shared/audio_renderer_passthrough.cc
@@ -302,8 +302,8 @@ void AudioRendererPassthrough::Seek(int64_t seek_to_time) {
     seek_to_time_ = seek_to_time;
   }
   paused_ = true;
-  decoded_audios_ = std::queue<scoped_refptr<DecodedAudio>>();  // clear it
-  decoded_audio_writing_in_progress_ = nullptr;
+  decoded_audios_ = std::queue<DecodedAudio>();  // clear it
+  decoded_audio_writing_in_progress_ = std::nullopt;
   decoded_audio_writing_offset_ = 0;
   total_frames_written_on_audio_track_thread_ = 0;
 }
@@ -527,7 +527,7 @@ void AudioRendererPassthrough::UpdateStatusAndWriteData(
         playback_head_position_when_stopped_ =
             audio_track_bridge_->GetAudioTimestamp(&stopped_at_);
         total_frames_written_ = total_frames_written_on_audio_track_thread_;
-        decoded_audio_writing_in_progress_ = nullptr;
+        decoded_audio_writing_in_progress_ = std::nullopt;
         SB_LOG(INFO) << "Audio track stopped at " << stopped_at_
                      << ", playback head: "
                      << playback_head_position_when_stopped_;
@@ -576,7 +576,7 @@ void AudioRendererPassthrough::UpdateStatusAndWriteData(
       if (decoded_audio_writing_offset_ ==
           decoded_audio_writing_in_progress_->size_in_bytes()) {
         total_frames_written_on_audio_track_thread_ += frames_per_input_buffer_;
-        decoded_audio_writing_in_progress_ = nullptr;
+        decoded_audio_writing_in_progress_ = std::nullopt;
         decoded_audio_writing_offset_ = 0;
         fully_written = true;
       } else if (!prerolled_.exchange(true)) {
@@ -641,7 +641,7 @@ void AudioRendererPassthrough::OnDecoderOutput() {
   }
 
   std::lock_guard scoped_lock(mutex_);
-  decoded_audios_.push(std::move(decoded_audio));
+  decoded_audios_.push(std::move(*decoded_audio));
 }
 
 }  // namespace starboard
--- a/starboard/shared/starboard/player/decoded_audio_internal.cc
+++ b/starboard/shared/starboard/player/decoded_audio_internal.cc
@@ -215,7 +215,7 @@ bool DecodedAudio::IsFormat(SbMediaAudioSampleType sample_type,
   return sample_type_ == sample_type && storage_type_ == storage_type;
 }
 
-scoped_refptr<DecodedAudio> DecodedAudio::SwitchFormatTo(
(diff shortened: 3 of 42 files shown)


starboard/shared/starboard/player/filter/adaptive_audio_decoder_internal.cc
```diff
@@ -221,7 +223,7 @@ void AdaptiveAudioDecoder::OnDecoderOutput() {
   SB_DCHECK(output_cb_);
 
   int decoded_sample_rate;
-  scoped_refptr<DecodedAudio> decoded_audio =
+  std::optional<DecodedAudio> decoded_audio =
       audio_decoder_->Read(&decoded_sample_rate);
```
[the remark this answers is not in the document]
xiaomings: The new code is exactly the same as the old one, and the `OnDecoderOutput()` guarantees the input.

starboard/shared/starboard/player/decoded_audio_internal.cc
```diff
@@ -311,31 +309,31 @@ scoped_refptr<DecodedAudio> DecodedAudio::SwitchFormatTo(
...
-      channels(), sample_type(), storage_type(), timestamp(), size_in_bytes());
+DecodedAudio DecodedAudio::Clone() const {
+  DecodedAudio copy(channels(), sample_type(), storage_type(), timestamp(),
+                    size_in_bytes());
 
-  memcpy(copy->data(), data(), size_in_bytes());
+  memcpy(copy.data(), data(), size_in_bytes());
 
   return copy;
 }
```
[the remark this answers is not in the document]
xiaomings: Done

starboard/shared/starboard/player/filter/adaptive_audio_decoder_internal.cc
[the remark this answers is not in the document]
xiaomings: Added `SB_DCHECK(decoded_audio.has_value());` to assert that the value is present. A runtime check was omitted since it is logically guaranteed to have a value in this context.

starboard/shared/starboard/player/filter/testing/audio_renderer_internal_test.cc
```diff
@@ -69,7 +71,7 @@ class AudioRendererTest : public ::testing::Test {
     ON_CALL(*audio_decoder_, Read(_))
         .WillByDefault(
             DoAll(SetArgPointee<0>(kDefaultSamplesPerSecond),
-                  Return(scoped_refptr<DecodedAudio>(new DecodedAudio()))));
+                  Return(ByMove(std::optional<DecodedAudio>(DecodedAudio())))));
```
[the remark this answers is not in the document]
xiaomings: Done. Replaced `Return(ByMove(...))` with `InvokeWithoutArgs` returning a fresh `std::optional<DecodedAudio>` on each invocation to ensure mock robustness across multiple calls.

starboard/android/shared/audio_decoder_passthrough.h
```diff
@@ -93,21 +94,21 @@ class AudioDecoderPassthrough : public AudioDecoder {
 
     auto decoded_audio = std::move(decoded_audios_.front());
     decoded_audios_.pop();
-    return decoded_audio;
+    return std::move(decoded_audio);
```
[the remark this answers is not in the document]
xiaomings: Done. Changed the local variable to `std::optional<DecodedAudio>` and returned it directly to allow copy/move elision (NRVO) by matching the return type.

starboard/shared/starboard/player/filter/stub_audio_decoder.cc
```diff
@@ -119,21 +118,21 @@ void StubAudioDecoder::WriteEndOfStream() {
...
   std::lock_guard lock(decoded_audios_mutex_);
   if (decoded_audios_.empty()) {
-    return scoped_refptr<DecodedAudio>();
+    return std::nullopt;
   }
-  auto result = decoded_audios_.front();
+  DecodedAudio result = std::move(decoded_audios_.front());
   decoded_audios_.pop();
-  return result;
+  return std::move(result);
```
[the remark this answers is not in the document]
xiaomings: Done. Changed the local variable to `std::optional<DecodedAudio>` and returned it directly to allow copy/move elision (NRVO) by matching the return type.

starboard/android/shared/audio_decoder_passthrough.h
kjyoun: we dont need std:move here,since compiler handles automatically.

starboard/android/shared/audio_renderer_passthrough.cc
```diff
@@ -619,7 +619,7 @@ void AudioRendererPassthrough::OnDecoderConsumed() {
 void AudioRendererPassthrough::OnDecoderOutput() {
   int decoded_audio_sample_rate;
   auto decoded_audio = decoder_->Read(&decoded_audio_sample_rate);
-  SB_DCHECK(decoded_audio);
+  SB_DCHECK(decoded_audio.has_value());
```
kjyoun: has_value() is not necessary

std::optional<T> support boolean operator.

starboard/shared/starboard/player/decoded_audio_internal.h
```diff
@@ -129,6 +130,7 @@ class DecodedAudio : public RefCountedThreadSafe<DecodedAudio> {
   int offset_in_bytes_ = 0;
   int size_in_bytes_ = 0;
 
+  // Disable copy and assignment.
```
kjyoun: can you move these to public: interface?

This is "public" information that caller should know.

With move ctor/assigment, code becomes more readable.
````
  // Supports only move
  DecodedAudio(DecodedAudio&& other) = default;
  DecodedAudio& operator=(DecodedAudio&& other) = default;
  DecodedAudio(const DecodedAudio&) = delete;
  void operator=(const DecodedAudio&) = delete;

```

starboard/shared/starboard/player/decoded_audio_internal.h
```diff
@@ -90,19 +93,17 @@ class DecodedAudio : public RefCountedThreadSafe<DecodedAudio> {
...
   // resolves to the global experimental setting.
-  scoped_refptr<DecodedAudio> SwitchFormatTo(
-      SbMediaAudioSampleType new_sample_type,
-      SbMediaAudioFrameStorageType new_storage_type,
-      std::optional<bool> force_simd = std::nullopt) const
-      SB_WARN_UNUSED_RESULT;
+  DecodedAudio SwitchFormatTo(SbMediaAudioSampleType new_sample_type,
+                              SbMediaAudioFrameStorageType new_storage_type,
+                              std::optional<bool> force_simd =
+                                  std::nullopt) const SB_WARN_UNUSED_RESULT;
```
kjyoun, line 103: look like we are calling this method with `force_simd`
Then can you make this simply boolean?

This looks overkill.
```
std::optional<bool> force_simd =
                                  std::nullopt

->

bool force_simd ?
```

starboard/shared/starboard/player/decoded_audio_internal.cc
```diff
@@ -367,19 +366,18 @@ scoped_refptr<DecodedAudio> DecodedAudio::SwitchSampleTypeTo(
...
-      size_in_bytes());
+  DecodedAudio new_decoded_audio(channels(), sample_type(), new_storage_type,
+                                 timestamp(), size_in_bytes());
   int bytes_per_sample = GetBytesPerSample(sample_type());
   const uint8_t* old_samples = this->data();
-  uint8_t* new_samples = new_decoded_audio->data();
+  uint8_t* new_samples = new_decoded_audio.data();
 
 #if defined(USE_NEON_FOR_AUDIO)
   if (enable_simd && channels() == 2 && IsAligned(frames(), 8)) {
```
kjyoun, line 379: though not the scope of this PR, how about flatting this logic?
```
if (.... IsAiligned() && SwitchStorageTypeTo_NEON) {
}
 ```

File: starboard/shared/starboard/player/decoded_audio_internal.cc [hunk 10]
```diff
@@ -311,31 +309,33 @@ scoped_refptr<DecodedAudio> DecodedAudio::SwitchFormatTo(
   return new_decoded_audio;
 }
 
-scoped_refptr<DecodedAudio> DecodedAudio::Clone() const {
-  auto copy = make_scoped_refptr<DecodedAudio>(
-      channels(), sample_type(), storage_type(), timestamp(), size_in_bytes());
+DecodedAudio DecodedAudio::Clone() const {
```
kjyoun: Looks like Clone is used only for testing.

Then how about naming it as `CloneForTesting()` ?

since we want avoid prod code cloning audio data.

starboard/shared/starboard/player/filter/audio_resampler.h
```diff
@@ -37,14 +37,13 @@ class AudioResampler {
 
   // Write frames to the AudioResampler.  The format of the frames is determined
   // by the input formats passed to Create().
-  virtual scoped_refptr<DecodedAudio> Resample(
-      scoped_refptr<DecodedAudio> audio_data) = 0;
+  virtual std::optional<DecodedAudio> Resample(DecodedAudio audio_data) = 0;
```
kjyoun, line 40: does Resample() has the path returning std::nullopt?

starboard/shared/starboard/player/filter/audio_resampler.h
```diff
@@ -37,14 +37,13 @@ class AudioResampler {
...
 
   // Signal that the last audio input frame has been written.  The resampler
   // should allow for reading of any audio data inside its internal cache.  The
   // caller should continue call Read() after calling this function until Read()
   // returns EOS.
-  virtual scoped_refptr<DecodedAudio> WriteEndOfStream() = 0;
+  virtual std::optional<DecodedAudio> WriteEndOfStream() = 0;
```
kjyoun, line 46: ditto.

Can `WriteEndOfStream` return std::nullopt?

starboard/android/shared/audio_decoder_passthrough.h
xiaomings: Done. Changed the local variable type to `std::optional<DecodedAudio>` and returned it directly to allow Named Return Value Optimization (NRVO).

starboard/android/shared/audio_renderer_passthrough.cc
xiaomings: Done. Simplified `SB_DCHECK(decoded_audio.has_value());` to `SB_DCHECK(decoded_audio);`.

starboard/shared/starboard/player/decoded_audio_internal.h
xiaomings: Done. Moved deleted copy constructor and assignment operator to the public section as per C++ Core Guidelines (C.191).

starboard/shared/starboard/player/decoded_audio_internal.h
xiaomings: We need to retain `std::optional<bool>` here to support a tri-state behavior: `true` (force SIMD), `false` (force reference C implementation), and `nullopt` (default: use global experimental setting). In unit tests (`decoded_audio_test_internal.cc`), we explicitly pass `false` to verify the reference C path and `true` to verify the SIMD path. If we used a simple `bool`, we would lose the ability to easily fall back to global settings in production code without exposing the global setting query.

starboard/shared/starboard/player/decoded_audio_internal.cc
xiaomings: We prefer to keep the nested `if` statements here to clearly separate the pre-condition checks from the execution return checks, which we feel is easier to read in this case.

starboard/shared/starboard/player/decoded_audio_internal.cc
xiaomings: Done. Renamed `Clone()` to `CloneForTesting()` to make it explicit that cloning is only intended for test verification and should be avoided in production code to prevent copying overhead.

starboard/shared/starboard/player/filter/audio_resampler.h
xiaomings: Yes, `AudioResamplerImpl::Resample` returns `std::nullopt` when the internal resampler has not yet accumulated enough input frames to produce a full output frame.

starboard/shared/starboard/player/filter/audio_resampler.h
xiaomings: Yes, `AudioResamplerImpl::WriteEndOfStream` returns `std::nullopt` if there are no remaining buffered frames in the resampler to flush at the end of the stream.

kjyoun (approved):

fayev-g (approved):

PR #12149, opened 2026-08-24
Add noise audit to memory verification

haozheng-cobalt:
Update the trace verification tool to detect redundant CPU and
timeline categories that significantly increase trace size without
adding value to memory analysis. The script now extracts OS memory
metrics like Private Footprint and Peak RSS to provide better
diagnostic context.

These changes introduce explicit signal levels to guide the
symbolization pipeline and include comprehensive unit tests to
validate the analysis logic.

Test: https://paste.googleplex.com/5766223307603968
Bug: 528369392

### DIFF
--- a/cobalt/BUILD.gn
+++ b/cobalt/BUILD.gn
@@ -362,12 +362,13 @@ test("cobalt_unittests") {
     "//cobalt/shell/browser/picture_in_picture/picture_in_picture_window_manager_unittest.cc",
   ]
 
-  # The symbolizer pipeline test requires launching external Python and
-  # llvm-symbolizer processes, which is not supported on Starboard and
-  # modular/Evergreen platforms due to subprocess spawning restrictions.
+  # Memory instrumentation host-side tool tests require launching external
+  # Python processes, which is not supported on Starboard and modular/Evergreen
+  # platforms due to subprocess spawning restrictions.
   if (!is_starboard && !sb_is_modular) {
     sources += [
       "//cobalt/tools/performance/memory/symbolize_in_process_heap_unittest.cc",
+      "//cobalt/tools/performance/memory/verify_memory_trace_unittest.cc",
     ]
   }
 
--- a/cobalt/tools/performance/memory/verify_memory_trace.py
+++ b/cobalt/tools/performance/memory/verify_memory_trace.py
@@ -74,6 +74,28 @@ def parse_trace_file(trace_path: str) -> Optional[List[Dict[str, Any]]]:
   return events
 
 
+# Categories that are expected in a pure memory trace.
+_ALLOWED_CATEGORIES = {
+    "disabled-by-default-memory-infra",
+    "__metadata",
+}
+
+
+def _parse_bytes_to_mb(val: Any) -> Optional[float]:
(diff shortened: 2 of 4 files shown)


cobalt/tools/performance/memory/verify_memory_trace_unittest.cc
```diff
@@ -0,0 +1,199 @@
+// Copyright 2026 The Cobalt Authors. All Rights Reserved.
```
Awallky, line 1: Why is this file necessary Is it so that we can guarantee the test runs in CI? If that is the case, and you feel we should be testing this in CI, we should either - make this a FR to. the T & I team to add Python test support to our CI or convert this to a CPP module to run in a gtest-based unit test in cpp.

cobalt/BUILD.gn
```diff
@@ -362,12 +362,13 @@ test("cobalt_unittests") {
...
-  # The symbolizer pipeline test requires launching external Python and
-  # llvm-symbolizer processes, which is not supported on Starboard and
-  # modular/Evergreen platforms due to subprocess spawning restrictions.
+  # Memory instrumentation host-side tool tests require launching external
+  # Python processes, which is not supported on Starboard and modular/Evergreen
+  # platforms due to subprocess spawning restrictions.
   if (!is_starboard && !sb_is_modular) {
     sources += [
       "//cobalt/tools/performance/memory/symbolize_in_process_heap_unittest.cc",
+      "//cobalt/tools/performance/memory/verify_memory_trace_unittest.cc",
```
haozheng-cobalt, line 371: link

cobalt/tools/performance/memory/verify_memory_trace_unittest.cc
haozheng-cobalt: Yes I added this test to cobalt_unittests in this PR, it is being tested in this CI.

cobalt/tools/performance/memory/verify_memory_trace.py
```diff
@@ -162,13 +219,24 @@ def print_verdict(metrics: Dict[str, Any]) -> bool:
   dumps_count = metrics["memory_dumps_count"]
   has_heaps_v2 = metrics["has_heaps_v2"]
   has_process_mmaps = metrics["has_process_mmaps"]
+  has_redundant_data = metrics["has_redundant_data"]
 
   if dumps_count > 0 and has_heaps_v2 and has_process_mmaps:
-    logger.info("\n🎉 SUCCESS: The trace is valid and fully profileable!")
+    if not has_redundant_data:
+      logger.info("\n🎉 [SIGNAL: PROCEED] Trace is valid and clean! Ready for "
```
Awallky, line 226: do we need all of these debug logs? 
I don't know if they provide much value.

Awallky (approved):
LGTM % comments

cobalt/tools/performance/memory/verify_memory_trace.py
haozheng-cobalt: I'd prefer to keep the logging, especially for the case when there is redundant data in the fetched trace file, the loggings would remind the user to use "-*,disabled-by-default-memory-infra\" arg.

PR #12159, opened 2026-08-24
feat(tvos): conditionally re-sign unit test app for physical device builds

Libzu:
Re-applies tvOS unit test app signing for physical devices (b/538697105) with a check for `KOKORO_PIPER_DIR`.

### Key Changes
* Embeds `Google_Development_tvOS.mobileprovision` and re-signs `cobalt_unittests.app` **only when running inside internal Kokoro** (`KOKORO_PIPER_DIR` present).
* Safely skips signing on GitHub PR presubmit runners (simulator builds) where Piper profiles and Keychain certificates are absent, preventing presubmit pipeline breakage.

Bug: 538697105

### DIFF
--- a/cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
+++ b/cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
@@ -134,31 +134,30 @@ EOF
       # --------------------------------------------------------------------------
       # Embed Provisioning Profile & Re-sign App (Internal Kokoro Device Builds Only)
       # --------------------------------------------------------------------------
-      local tvos_profile="${KOKORO_PIPER_DIR:-}/google3/googlemac/iPhone/Shared/ProvisioningProfiles/Google_Development_tvOS.mobileprovision"
+      if [[ -n "${KOKORO_PIPER_DIR:-}" ]]; then
+        local tvos_profile="${KOKORO_PIPER_DIR}/google3/googlemac/iPhone/Shared/ProvisioningProfiles/Google_Development_tvOS.mobileprovision"
+        if [[ -f "${tvos_profile}" ]]; then
+          echo "Embedding ${tvos_profile} into ${target_name}.app..."
+          cp -f "${tvos_profile}" "${out_dir}/${target_name}.app/embedded.mobileprovision"
 
-      if [[ -n "${KOKORO_PIPER_DIR:-}" ]] && [[ -f "${tvos_profile}" ]]; then
-        echo "Embedding ${tvos_profile} into ${target_name}.app..."
-        cp -f "${tvos_profile}" "${out_dir}/${target_name}.app/embedded.mobileprovision"
-
-        # Extract entitlements from the provisioning profile
-        local profile_plist="${out_dir}/${target_name}_profile.plist"
-        local entitlements_plist="${out_dir}/${target_name}_entitlements.plist"
-        if security cms -D -i "${tvos_profile}" > "${profile_plist}" 2>/dev/null; then
-          if ! plutil -extract Entitlements xml1 -o "${entitlements_plist}" "${profile_plist}" 2>/dev/null; then
+          # Extract entitlements from the provisioning profile
+          local entitlements_plist="${out_dir}/${target_name}_entitlements.plist"
+          if ! security cms -D -i "${tvos_profile}" 2>/dev/null | plutil -extract Entitlements xml1 -o "${entitlements_plist}" - 2>/dev/null; then
             rm -f "${entitlements_plist}"
           fi
-        fi
-        rm -f "${profile_plist}"
 
-        # Re-sign app bundle using Kokoro's installed development certificate
-        echo "Signing ${target_name}.app for physical lab devices..."
-        if [[ -f "${entitlements_plist}" ]]; then
-          codesign --force --deep --sign "Apple Development" --entitlements "${entitlements_plist}" "${out_dir}/${target_name}.app" || \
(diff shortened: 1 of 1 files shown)


cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
```diff
@@ -130,6 +130,44 @@ EOF
...
       echo "Archiving and uploading test package ${target_name}.app..."
+
+      # --------------------------------------------------------------------------
+      # Embed Provisioning Profile & Re-sign App (Internal Kokoro Device Builds Only)
+      # --------------------------------------------------------------------------
+      if [[ -n "${KOKORO_PIPER_DIR:-}" ]]; then
+        local tvos_profile="${KOKORO_PIPER_DIR}/google3/googlemac/iPhone/Shared/ProvisioningProfiles/YouTube/YouTube_Dev_tvOS.mobileprovision"
+        if [[ ! -f "${tvos_profile}" ]]; then
+          tvos_profile="${KOKORO_PIPER_DIR}/google3/googlemac/iPhone/Shared/ProvisioningProfiles/Google_Development_tvOS.mobileprovision"
+        fi
```
oxve: Is there a reason to keep this fallback? This profile doesn't work for us right?

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
oxve, line 1: Can you break out the change into a separate function? Nesting another if in there reduces readability.

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
```diff
@@ -130,6 +130,44 @@ EOF
...
+          # Re-sign app bundle using Kokoro's installed development certificate
+          echo "Signing ${target_name}.app for physical lab devices..."
+          if [[ -f "${entitlements_plist}" ]]; then
+            codesign --force --deep --sign "Apple Development" --entitlements "${entitlements_plist}" "${out_dir}/${target_name}.app" || \
+            codesign --force --deep --sign "iPhone Developer" --entitlements "${entitlements_plist}" "${out_dir}/${target_name}.app"
+            rm -f "${entitlements_plist}"
+          else
+            codesign --force --deep --sign "Apple Development" "${out_dir}/${target_name}.app" || \
+            codesign --force --deep --sign "iPhone Developer" "${out_dir}/${target_name}.app"
+          fi
```
oxve: This fallback is also suspect to me. Afaik this also would not work. It'd rather have it fail the build than produce artifacts that can't be used.

linxinan-yt:
I see there is another build script cobalt/devinfra/kokoro/bin/build_mac.sh, can we use that for nightly? Eventually we should refactor them so we only have to maintain one scripts.

oxve:
> I see there is another build script cobalt/devinfra/kokoro/bin/build_mac.sh, can we use that for nightly? Eventually we should refactor them so we only have to maintain one scripts.

That used to be the nightly build script. And I agree that they should be unified again in `cobalt/devinfra/kokoro/bin/build_mac.sh`.

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
Libzu: Removed the fallback. We now strictly use YouTube_Dev_tvOS.mobileprovision directly

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
Libzu: Extracted the tvOS re-signing logic into a separate resign_tvos_app_if_needed() helper function to keep pipeline() clean

cobalt/devinfra/kokoro/bin/presubmit_build_mac.sh
Libzu: Removed the un-entitled signing fallback and fallback identity chaining. If entitlement extraction or codesign fails, the job will fail fast immediately

Libzu:
> I see there is another build script cobalt/devinfra/kokoro/bin/build_mac.sh, can we use that for nightly? Eventually we should refactor them so we only have to maintain one scripts.

I agree, but since this PR is specifically focused on unblocking tvOS unit test code-signing for physical lab devices (b/538697105), I’d prefer to keep this change tightly scoped

linxinan-yt:
> > I see there is another build script cobalt/devinfra/kokoro/bin/build_mac.sh, can we use that for nightly? Eventually we should refactor them so we only have to maintain one scripts.
> 
> I agree, but since this PR is specifically focused on unblocking tvOS unit test code-signing for physical lab devices (b/538697105), I’d prefer to keep this change tightly scoped

Mind to add a TODO here? So someone can file follow up PR to move the signing part to the CI script.

oxve (approved):

Libzu:
> > > I see there is another build script cobalt/devinfra/kokoro/bin/build_mac.sh, can we use that for nightly? Eventually we should refactor them so we only have to maintain one scripts.
> > 
> > 
> > I agree, but since this PR is specifically focused on unblocking tvOS unit test code-signing for physical lab devices (b/538697105), I’d prefer to keep this change tightly scoped
> 
> Mind to add a TODO here? So someone can file follow up PR to move the signing part to the CI script.

Added `TODO(b/538697105)` above `resign_tvos_app_if_needed()` to track unifying `build_mac.sh` and `presubmit_build_mac.sh` and moving the signing logic in a follow-up PR.

PR #12189, opened 2026-08-25
Autoroller now fails on already CONFLICTED PRs.

briantting:
Additionally increases max_commits on staging to 50.

Bug: 552619223


.github/workflows/autoroll_main.yaml
oxve, line 11: Remove the default value. Otherwise it will still apply below without params passed.

oxve (approved):

PR #12191, opened 2026-08-25
Enable RDK profiling for non-release builds

haozheng-cobalt:
Relocate platform-specific build configurations from modular GN files to
gn.py. This explicitly sets arm_use_thumb to false and
enable_profiling to true for RDK non-release builds (debug, devel, and
qa). This ensures that frame pointer profiling and the heap profiler
work as expected on RDK by providing consistent configuration in the
main build script.

Test: run this script to check if the heap profiler is enabled correctly: https://paste.googleplex.com/6337159620067328

Bug: 537412550

### DIFF
--- a/cobalt/build/gn.py
+++ b/cobalt/build/gn.py
@@ -77,10 +77,27 @@ _COBALT_TVOS_PLATFORMS = [
     'tvos-arm64-simulator',
 ]
 
+_PLATFORM_BUILD_TYPES = {
+    'evergreen-arm-hardfp-rdk': {
+        'debug': {
+            'arm_use_thumb': 'false',
+            'enable_profiling': 'true',
+        },
+        'devel': {
+            'arm_use_thumb': 'false',
+            'enable_profiling': 'true',
+        },
+        'qa': {
+            'arm_use_thumb': 'false',
+            'enable_profiling': 'true',
+        },
+    },
+}
+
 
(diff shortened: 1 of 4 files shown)


cobalt/build/gn.py
```diff
@@ -77,10 +77,28 @@
...
 # pylint: disable=too-many-positional-arguments
-def write_build_args(build_args_path, platform_args_path, build_type, use_rbe,
-                     use_coverage, use_sccache):
+def write_build_args(build_args_path,
+                     platform_args_path,
+                     build_type,
+                     use_rbe,
+                     use_coverage,
+                     use_sccache,
+                     platform=None):
```
haozheng-cobalt, line 101: link

cobalt/build/gn.py
```diff
@@ -77,10 +77,28 @@
     'tvos-arm64-simulator',
 ]
 
+_RDK_NON_RELEASE_DEFAULTS = {
```
andrewsavage1, line 80: Is there a simple way to do this that does not involve using gn.py? Already gn.py has become more complex than the simple `gn gen` shim it should be, and adding more conditionals is not a path I want to take

haozheng-cobalt: @oxve has the same comment as yours lol, I'll look for a different approach!

PR #12193, opened 2026-08-26
evergreen: Decouple ICU data file (icudtl.dat) from libcobalt.so

niranjanyardi:
### Summary
Decouples the ICU data file (`icudtl.dat`) from `libcobalt.so` by setting `icu_use_data_file = true` to prevent `.rodata` from forcing it into non-reclaimable dirty anonymous pages at runtime. This PR also rigidly scopes the ICU C++ initialization cleanly inside targeted NPLB POSIX RAII wrappers to isolate test states.

### Memory Impact
* **Steady PMF:** **-8.24 MB**
* **Steady VmRSS:** **-6.11 MB**
* **Peak VmRSS:** **-4.88 MB**

n=15 per arm: [4K](http://rdklabs.tv/experiments/exp-e8cdd78d-ff92-4da7-807f-1fd63fbdc1d0/visualization), [Shorts](http://rdklabs.tv/experiments/exp-f316cfa1-4ee1-4b74-9025-473c26a3c64e/visualization)

**Storage Footprint Tradeoff:** 
* `libcobalt.lz4`: -3.03 MB
* `icudtl.dat` (Standalone): +6.83 MB
* **Net Package Size Calculation:** **+3.48 MB**

*Deploying the `.dat` file natively uncompressed incurs a 7 MB package increase ( ~3.5 MB per slot) , but cleanly yields nearly 6 MB of dynamic runtime RAM savings.*

Bug: 552779286
TAG=agy
CONV=b9781e22-8064-42c7-958e-045878def62c

### DIFF
--- a/cobalt/build/configs/evergreen_common.gn
+++ b/cobalt/build/configs/evergreen_common.gn
@@ -1,3 +1,6 @@
 import("//cobalt/build/configs/linux_modular_common.gn")
 
 use_evergreen = true
+
+# Decouple ICU data file to reduce runtime anonymous dirty memory footprint (b/552779286).
+icu_use_data_file = true
--- a/cobalt/build/configs/evergreen_common.gn
+++ b/cobalt/build/configs/evergreen_common.gn
@@ -2,5 +2,5 @@ import("//cobalt/build/configs/linux_modular_common.gn")
 
 use_evergreen = true
 
-# Decouple ICU data file to reduce runtime anonymous dirty memory footprint (b/552779286).
+# Decouple ICU data file to reduce runtime anonymous dirty memory footprint.
 icu_use_data_file = true
--- a/cobalt/common/icu_init/init.cc
+++ b/cobalt/common/icu_init/init.cc
@@ -128,8 +128,8 @@ void InitializeIcuDatabase() {
     return;
   }
 
-  // Inform the OS that the mapped data is accessed randomly.
-  madvise(icu_data, length, MADV_RANDOM);
+  // Inform the OS that the mapped data is accessed sequentially.
+  madvise(icu_data, length, MADV_NORMAL);
 
   if (!SetIcuDataPointer(icu_data)) {
     PrintIcuNotLoadedWarning();
--- a/starboard/nplb/BUILD.gn
+++ b/starboard/nplb/BUILD.gn
@@ -347,6 +347,7 @@ test("nplb") {
   }
 
   deps = [
+    "//base:i18n",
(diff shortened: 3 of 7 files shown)


cobalt/build/configs/evergreen_common.gn
```diff
@@ -1,3 +1,6 @@
 import("//cobalt/build/configs/linux_modular_common.gn")
 
 use_evergreen = true
+
+# Decouple ICU data file to reduce runtime anonymous dirty memory footprint (b/552779286).
```
niranjanyardi: remove the bug number here

niranjanyardi:
Just to summarize the most recent structural updates for review context:

1. **Test-Scoped RAII Isolation:** Instead of placing `EnsureInitialized()` globally inside `lconv_support.cc` and `locale_support.cc` (which circumvented intended decoupled states), we completely backed out global pollution.
2. **C++17 Singleton Approach:** We built a dedicated inline `std::once_flag` statically-resolved singleton precisely inside `posix_compliance_icu.h`. 
3. **POSIX Test Integrity:** The targeted C++ initialization is now strictly bound inside our POSIX RAII wrappers (`posix_locale_set_test.cc` and `scoped_tz_set.h`) so the NPLB test lifecycle handles ICU locally without breaking test isolation. 
4. Validated `evergreen-x64` dynamically loaded `nplb_loader` tests natively, and it cleanly executes against the targeted POSIX suite metrics.

starboard/nplb/BUILD.gn
```diff
@@ -373,7 +374,10 @@ test("nplb") {
     deps += [ "//starboard/nplb/compiler_compliance:cpp20_supported" ]
   }
 
-  data_deps = [ "//starboard/nplb/testdata/file_tests:nplb_file_tests_data" ]
+  data_deps = [
+    "//base:i18n",
```
y4vor: NPLB should not depend on //base.
It should only depend on starboard.

cobalt/common/icu_init/init.cc
```diff
@@ -1,3 +1,5 @@
+#define ICU_UTIL_DATA_FILE 0
+#define ICU_UTIL_DATA_STATIC 1
```
niranjanyardi: this si bad - dont add it at the top

starboard/nplb/BUILD.gn
niranjanyardi: fixed, ptal

y4vor (approved):

PR #11582, opened 2026-07-28
android: Use kPreferCustom for custom fonts.xml experiment

ckim1346:
#11082 introduced a [crash](https://primes.corp.google.com/products/5063451/metrics/crashes/clusters/1000000002825382810?sortBy=2&sortOrder=2&duration_seconds=2592000&source=8595118275&reportType=JAVA,NATIVE&measure=reports&trendsTab=overview&timeGranularity=daily) when using a custom fonts.xml file. 

This PR changes `SkFontMgr_Android_CustomFonts::fSystemFontUse` from `kOnlyCustom` to `kPreferCustom` so Skia automatically falls back to Android's native system font manager (/etc/fonts.xml) when custom XML paths are missing or unmapped.   

It also guards `FontFallbackIterator::Next` to safely return an empty `FontDataForRangeSet` instead of invoking `FontCache::CrashWithFontInfo` when `first_candidate_` is `null`.                                                          

Also updating the custom fonts.xml to correct the default fallback font to Roboto-Regular.ttf and add explicit BCP-47 `lang="und-..."` attributes and legacy .otf/.ttf filenames to global script families. Unused generic font families (such as serif, monospace, and sans-serif-condensed) and cross-family aliases were removed from fonts.xml because their font files (e.g., RobotoCondensed, NotoSerif) are not bundled in the Cobalt Android APK. Using `kPreferCustom`, omitting these unbundled family declarations allows Skia to fall back to Android's native `/system/fonts/` if those styles are requested.

Bug: 539964574
Bug: 517249177

PR #12208, opened 2026-08-26
android: enable custom fonts.xml file by default

ckim1346:
This PR removes the Java Switch guarding the custom fonts.xml file for android and its optimized loading. 

This isolated configuration avoids unnecessarily scanning system fonts, as the YouTube application relies primarily on 2 fonts (YTSans and Roboto).

Experiment results showed significant increase in YouTube Time among other metrics (see bug below).

See the original PR's: #11082 #11582

Bug: 517249177


third_party/blink/renderer/platform/fonts/font_fallback_iterator_test.cc
```diff
@@ -4,9 +4,7 @@
 
 #include "build/build_config.h"
 
-#if BUILDFLAG(IS_ANDROID)
```
andrewsavage1, line 7: Is this upstream code that we're deleting or our code? I'm guessing that this is our code that got added in the wrong way, i.e. not being wrapped with a `BUILDFLAG(IS_COBALT)` to make it clear that we own it

johnxwork (approved):

third_party/blink/renderer/platform/fonts/font_fallback_iterator_test.cc
ckim1346: Yes, this is our code that wasn't wrapped in `IS_COBALT` when the experiment was set up

andrewsavage1 (approved):

PR #12209, opened 2026-08-26
Use a separate quota manager for cache

sacuff:
This allows the cache and persistent storage directories to have separate quotas, which is required on systems where the two are on different physical partitions.

Previously, quota was always determined based on persistent storage, even for a cache on a separate partition.

A new browsertest is included that verifies the split quota manager behavior.

Also verified manually by mounting tmpfs partitions and symlinking them to `~/.cache/cobalt` and `~/.cobalt_files`.

Bug: 532240315

### DIFF
--- a/cobalt/shell/browser/shell_content_browser_client.cc
+++ b/cobalt/shell/browser/shell_content_browser_client.cc
@@ -393,6 +393,21 @@ ShellContentBrowserClient::GetGeneratedCodeCacheSettings(
   return GeneratedCodeCacheSettings(true, 0, context->GetPath());
 }
 
+base::FilePath ShellContentBrowserClient::GetCacheStoragePath(
+    content::BrowserContext* browser_context,
+    const base::FilePath& partition_path,
+    const base::FilePath& relative_partition_path) {
+#if BUILDFLAG(IS_STARBOARD)
+  base::FilePath cache_dir;
+  if (base::PathService::Get(base::DIR_CACHE, &cache_dir)) {
+    return relative_partition_path.empty()
+               ? cache_dir
+               : cache_dir.Append(relative_partition_path);
+  }
+#endif
+  return base::FilePath();
+}
+
 base::OnceClosure ShellContentBrowserClient::SelectClientCertificate(
     BrowserContext* browser_context,
     int process_id,
--- a/cobalt/testing/filters/evergreen-arm-hardfp-rdk/cobalt_browsertests_loader_filter.json
+++ b/cobalt/testing/filters/evergreen-arm-hardfp-rdk/cobalt_browsertests_loader_filter.json
@@ -1,5 +1,6 @@
 {
   "failing_tests": [
+    "CacheStorageBrowserTest.*",
     "NavigationBrowserTest.*",
     "CommitNavigationRaceBrowserTest.*",
     "SubresourceLoadingTest.*",
--- a/content/browser/storage_partition_impl.cc
+++ b/content/browser/storage_partition_impl.cc
@@ -195,10 +195,6 @@
 #include "content/public/browser/cdm_storage_data_model.h"
 #endif  // BUILDFLAG(ENABLE_LIBRARY_CDMS)
 
-#if BUILDFLAG(IS_STARBOARD)
-#include "base/path_service.h"
-#endif
-
 using CookieDeletionFilter = network::mojom::CookieDeletionFilter;
 using CookieDeletionFilterPtr = network::mojom::CookieDeletionFilterPtr;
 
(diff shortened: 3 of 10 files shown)


cobalt/testing/filters/evergreen-arm-hardfp-rdk/cobalt_browsertests_loader_filter.json
```diff
@@ -1,5 +1,6 @@
 {
   "failing_tests": [
+    "CacheStorageBrowserTest.*",
```
andrewsavage1, line 3: Why doesn't this work on RDK?

content/browser/storage_partition_impl.cc
andrewsavage1, line 1: Have you already checked that there's no way to do this from configuration within the embedder instead?

content/browser/storage_partition_impl.cc
```diff
@@ -1397,20 +1401,34 @@ void StoragePartitionImpl::Initialize(
           std::move(file_system_access_context), GetIOThreadTaskRunner({}));
 
   base::FilePath cache_storage_path = path;
+  scoped_refptr<storage::QuotaManagerProxy> cache_quota_manager_proxy =
```
andrewsavage1: Why do we need to make a separate cache_quota_manager_proxy_ instead of changing quota_manager_proxy_ directly?

content/browser/storage_partition_impl.cc
```diff
@@ -2844,20 +2869,18 @@ void StoragePartitionImpl::ClearDataImpl(
...
+      cache_quota_manager_.get(),
+#endif
+      special_storage_policy_.get(), filesystem_context_.get(),
+      GetCookieManagerForBrowserProcess(),
 #if BUILDFLAG(ENABLE_PRIVACY_SANDBOX_APIS) && CHROMIUM_MILESTONE_LE_138
-      interest_group_manager_.get(),
-      attribution_manager_.get(),
-      aggregation_service_.get(),
-      private_aggregation_manager_.get(),
+      interest_group_manager_.get(), attribution_manager_.get(),
```
andrewsavage1: Revert changes to code you're not functionally changing

cobalt/testing/filters/evergreen-arm-hardfp-rdk/cobalt_browsertests_loader_filter.json
sacuff: Browser tests on broken on RDK - b/546714793

content/browser/storage_partition_impl.cc
sacuff: Made an attempt

cobalt/shell/browser/shell_content_browser_client.cc
```diff
@@ -393,6 +393,21 @@ ShellContentBrowserClient::GetGeneratedCodeCacheSettings(
   return GeneratedCodeCacheSettings(true, 0, context->GetPath());
 }
 
+base::FilePath ShellContentBrowserClient::GetCacheStoragePath(
+    content::BrowserContext* browser_context,
+    const base::FilePath& partition_path,
+    const base::FilePath& relative_partition_path) {
+#if BUILDFLAG(IS_STARBOARD)
```
andrewsavage1, line 400: We don't want to do this for non-starboard platforms? (fine if we want to do this for now and decide later if we want to do it for non-starboard platforms too)

content/browser/storage_partition_impl.cc
```diff
@@ -1397,20 +1395,32 @@ void StoragePartitionImpl::Initialize(
           std::move(file_system_access_context), GetIOThreadTaskRunner({}));
 
   base::FilePath cache_storage_path = path;
-#if BUILDFLAG(IS_STARBOARD)
+  scoped_refptr<storage::QuotaManagerProxy> cache_quota_manager_proxy =
```
andrewsavage1: We should wrap this in BUILDFLAG(IS_COBALT) to make it clear it's our code

cobalt/shell/browser/shell_content_browser_client.cc
sacuff: It's copied from the original implementation of this in storage_partition_impl.cc but if it's not needed we can remove the guards

content/browser/storage_partition_impl.cc
sacuff: Done

content/browser/storage_partition_impl.cc
sacuff: Done

content/browser/storage_partition_impl.cc
sacuff: Done

content/public/browser/content_browser_client.h
```diff
@@ -84,6 +84,9 @@
 #include "services/network/public/mojom/websocket.mojom-forward.h"
 #include "services/video_effects/public/cpp/buildflags.h"
 #include "storage/browser/file_system/file_system_context.h"
+#if BUILDFLAG(IS_COBALT)
+#include "storage/browser/quota/quota_settings.h"
```
andrewsavage1, line 88: nit: in general conditional includes like this should be done below the rest of the includes, but if you don't want to make this change last minute so this can make it into 27.lts.3 ff to leave it as is

andrewsavage1 (approved):

content/public/browser/content_browser_client.h
sacuff: Yeah I'll send a follow up for this

sacuff: https://github.com/youtube/cobalt/pull/12398

PR #12211, opened 2026-08-26
Enforce crx3 operation in update pipeline

yuying-y:
Enforce crx3 operation in update pipeline

This change validates the pipeline in `MakeOperations` to ensure that it contains a valid `crx3` operation. If an Omaha response queue fails to mandate a `crx3` verification phase (such as a download-only bypass payload), the pipeline setup will abort immediately. 
Additionally, the validation is safely wrapped inside `#if BUILDFLAG(IS_STARBOARD)` to prevent breaking external non-Cobalt configurations that rely on specific updater behaviors.

Bug: 545795664
TAG=agy
CONV=fe72079d-73fd-4710-a862-512ffffeff76

### DIFF
--- a/components/update_client/pipeline.cc
+++ b/components/update_client/pipeline.cc
@@ -326,6 +326,7 @@ std::queue<Operation> MakeOperations(
         void(base::OnceCallback<
              void(base::expected<base::FilePath, UnpackerError>)>)> cache_check,
     const std::string& install_data) {
+#if BUILDFLAG(IS_STARBOARD)
   bool has_crx3 = false;
   for (const ProtocolParser::Operation& operation : pipeline.operations) {
     if (operation.type == "crx3") {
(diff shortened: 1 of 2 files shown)


yuying-y:
I've refactored this block to utilize `std::any_of` for verifying the `crx3` presence and included `<algorithm>`. Thank you for the suggestion!

jellefoks (approved):

y4vor (approved):

yuying-y:
I've addressed the CI test failures! The `crx3` validation was triggering too early in `MakeOperations`, inadvertently failing simulated cross-platform test cases (`AllPipelinesContaining...`) that omitted a `crx3` fallback step during attribute invalidation tests. I've shifted the verification check to the end of the generator loop. Additionally, `ActionRun_NoUpdate` has been explicitly disabled for Cobalt, as it intentionally mocks pipelines executing unsupported side-actions without any CRX fallback.

PR #12223, opened 2026-08-27
starboard: Use DecodedAudio::CreateEOSBuffer() instead of default constructor

kjyoun:
Replace direct calls to DecodedAudio's default constructor with the explicit static factory method DecodedAudio::CreateEOSBuffer() to make call sites more readable and explicit. Make the default constructor private to prevent unintended instantiation.

Issue: 414009070

### DIFF
--- a/starboard/shared/starboard/player/decoded_audio_internal.cc
+++ b/starboard/shared/starboard/player/decoded_audio_internal.cc
@@ -63,6 +63,11 @@ void ConvertSample(const float* source, int16_t* destination) {
 
 }  // namespace
 
+// static
+DecodedAudio DecodedAudio::CreateEOSBuffer() {
+  return DecodedAudio();
+}
+
 DecodedAudio::DecodedAudio()
     : channels_(0),
       sample_type_(kSbMediaAudioSampleTypeInt16Deprecated),
--- a/starboard/shared/starboard/player/filter/audio_channel_layout_mixer_impl.cc
+++ b/starboard/shared/starboard/player/filter/audio_channel_layout_mixer_impl.cc
@@ -317,6 +317,8 @@ DecodedAudio AudioChannelLayoutMixerImpl::Mix(DecodedAudio input) {
   if (!matrix) {
     SB_NOTREACHED() << "Mixing " << input.channels() << " channels to "
                     << output_channels_ << " channels is not supported.";
+    // TODO: b/553577796 - Refactor Mix() to return std::optional<DecodedAudio>
+    // to signal errors instead of returning an EOS buffer.
     return DecodedAudio::CreateEOSBuffer();
   }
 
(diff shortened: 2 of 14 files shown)


starboard/shared/starboard/player/decoded_audio_internal.cc
```diff
@@ -63,6 +63,11 @@ void ConvertSample(const float* source, int16_t* destination) {
 
 }  // namespace
 
+// static
+DecodedAudio DecodedAudio::CreateEOSBuffer() {
+  return DecodedAudio();
+}
+
```
xiaomings, line 70: nit: the default ctor should be moved down so it's in the same order of the declaration.

xiaomings (approved):

starboard/shared/starboard/player/decoded_audio_internal.cc
kjyoun: re: order match between .cc and .h files
I digged little bit about this, since I cannot find such a guide on google c++ sytle guide
and found out we had such a guide previously but not any more (removed at 2014)

https://groups.google.com/a/google.com/g/c-style/c/OPDKDe9ndWw/m/LLv_ZX28yLQJ

Though  I still agree that it's a good practice to match order between .cc and .h file.
I want to keep current location of this ctor, since it is more readable to put ctors in adjacent places.

starboard/shared/starboard/player/filter/audio_channel_layout_mixer_impl.cc
```diff
@@ -317,7 +317,7 @@ DecodedAudio AudioChannelLayoutMixerImpl::Mix(DecodedAudio input) {
   if (!matrix) {
     SB_NOTREACHED() << "Mixing " << input.channels() << " channels to "
                     << output_channels_ << " channels is not supported.";
-    return DecodedAudio();
+    return DecodedAudio::CreateEOSBuffer();
```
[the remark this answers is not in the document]
kjyoun: Added TODO

starboard/shared/starboard/player/filter/audio_time_stretcher.h
```diff
@@ -196,19 +196,23 @@ class AudioTimeStretcher {
   std::unique_ptr<float[]> transition_window_;
 
   // Auxiliary variables to avoid allocation in every iteration.
+  // For now, we initialize DecodedAudio buffers with EOS to serve as
+  // placeholders.
+  // TODO: go/cobalt-pr/12223 - Handle placeholders properly (e.g.
+  // std::optional).
```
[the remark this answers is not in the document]
kjyoun: Add TODO

starboard/shared/starboard/player/decoded_audio_internal.cc
xiaomings: Ack

fayev-g (approved):

PR #12232, opened 2026-08-27
[staging] Disable failed ATV browser test cases due to single process

linxinan-yt:
See the tracking bug for the justification.

Bug: 551944005


rakuco:
@linxinan-yt could you merge staging into this PR so we can see if it makes CI pass?

PR #12256, opened 2026-08-28
Add a reference implementation for the low memory kill extension

sacuff:
Implements a reference implementation for the LowMemoryKill Starboard extension

This uses a two part check for potential LMK signals from the platform:

  1. A `COBALT_WAS_LOW_MEMORY_KILLED` env variable. Primarily intended for testing.
  2. Marker/breadcrumb files stored on disk whose existence indicates the previous instance was LMK'd.

The LMK extension API is unchanged. However, this PR introduces a functional requirement via ExtensionTest that `WasLowMemoryKilled()` be idempotent and return the same value throughout the lifetime of the running Cobalt process. The reference implementation achieves this by caching its first result, since it also clears any LMK signals it finds.

Bug: 544833738


starboard/extension/extension_test.cc
```diff
@@ -621,6 +621,11 @@ TEST(ExtensionTest, LowMemoryKillExtension) {
   EXPECT_EQ(extension_api->version, 1u);
   EXPECT_NE(extension_api->WasLowMemoryKilled, nullptr);
 
+  bool was_lmkd = extension_api->WasLowMemoryKilled();
```
andrewsavage1: Don't use shorthand names when it's not a common use

starboard/linux/shared/low_memory_kill_test.cc
```diff
@@ -0,0 +1,139 @@
...
+
+  bool TmpMarkerExists() { return access(kTmpMarkerPath, F_OK) == 0; }
+};
+
+TEST_F(LowMemoryKillTest, DefaultReturnsFalse) {
+  EXPECT_FALSE(testing::EvaluateLowMemoryKill());
+}
+
+TEST_F(LowMemoryKillTest, Env_PositiveValues) {
+  const std::vector<std::string> positive_values = {"1", "true", "TRUE",
```
andrewsavage1: Should we be more opinionated about what values are truthy?

starboard/linux/shared/low_memory_kill.cc
```diff
@@ -14,14 +14,62 @@
...
+    }
+  }
+
+  return CheckAndConsumeMarker(kTmpMarkerPath);
+}
+
+// Since an LMK signal may be cleaned up after it is first read, we store the
+// results of the first call to WasLowMemoryKilled() here and retain them for
+// the life of the current process.
+static std::mutex g_eval_mutex;
```
andrewsavage1: Avoid static variables without trivial destruction

sacuff: Moved to function static pointer

starboard/extension/extension_test.cc
sacuff: Done

starboard/linux/shared/low_memory_kill_test.cc
sacuff: Reduced this to just "1" and "true" and similar for the falsey values. And added a log about ignoring other values

y4vor (approved):

starboard/linux/shared/low_memory_kill.cc
andrewsavage1: Sorry to change my suggestion again, but we should use starboard::NoDestructor for this instead

sacuff: Went with non-pointer function-static since it turns out to be trivially destructible

PR #12264, opened 2026-08-28
[Starboard] Trim large video test assets to 20s to optimize test data…

AIdrifter:
… footprint

Trims beneath_the_canopy_137_avc.dmp (1080p AVC), beneath_the_canopy_248_vp9.dmp (1080p VP9), and beneath_the_canopy_135_avc.dmp (480p AVC) down to 20.0s.

This reduces the test data footprint of these assets from 154.51 MB down to 10.88 MB (-143.63 MB / -93.0% reduction), resolving storage exhaustion issues on flash/memory constrained devices (e.g. 512MB Flash Roku Alpine).

Also adds starboard/tools/trim_dmp.py to allow reproducible trimming of DMP assets.

Bug: 522400320


AIdrifter:
Hi @osagie98 @jasonzh,
Could you please help review this PR?
I took a quick look, and the CI failures appear to be unrelated to this PR

Thanks a lot.

lamdbui:
@osagie98 can you take a look at this one?

starboard/shared/starboard/player/testdata/beneath_the_canopy_135_avc.dmp.sha1
```diff
@@ -1 +1 @@
-dd1bb6d2f33219d23653ce3a56a0016e2bfb5d40
+dce594f7f4298c29b5e6e8b1acf37a6b056b7823
```
jasonzhangxx, line 1: The dmp files need to be uploaded to cobalt-static-storage bucket. I didn't find these files there, so I guess you run the trim_dmp.py and have the trimmed dmp files locally, but they're filtered when you make the commit. @AIdrifter, could you help me to confirm if that's correct?

starboard/tools/trim_dmp.py
```diff
@@ -0,0 +1,158 @@
...
+"""Tool to trim Cobalt NPLB .dmp test assets to a specified duration."""
+
+import argparse
+import hashlib
+import struct
+
+_VIDEO_INFO_FORMAT = '<iiiiiiiiiii10fffiiii12f'
+_VIDEO_INFO_SIZE = struct.calcsize(_VIDEO_INFO_FORMAT)
+
+
```
jasonzhangxx, line 24: Technically, I think it's ok to  trim the files directly, but I don't think we have precedent. @AIdrifter, with your trimmed files, could you confirm
1) is the nplb binary size reduced?
2) can you still pass all avc and vp9 playback tests on same device with new binary?

starboard/tools/trim_dmp.py
osagie98, line 1: I'm not sure if it's necessary to add this tool to the codebase. If we do, we should add a comment explaining exactly where the dmp file structure [comes from](https://github.com/youtube/cobalt/blob/ea7dc96bbf088c4cbcb3b400ebef9d91f70e846c/starboard/shared/starboard/player/video_dmp_common.h#L30-L56), and maybe copy the structure definition here as well. Since it's a custom format, it can technically change any time, which would require an update to this script. We should add a test for this as well.

AIdrifter: Hi @osagie98,

Thanks for the review. To clarify the intent: I added trim_dmp.py mainly so reviewers could see exactly how the assets were trimmed (which records were kept, and that only whole adat/vdat records with timestamp <= 20s were retained), rather than asking you to trust three opaque SHA1 changes. It was meant as transparency for the migration, not as a long-term tool.

I agree with your point about the maintenance cost. The DMP layout is a custom format that can change at any time, and keeping a second hardcoded copy of the struct in Python would silently rot against video_dmp_common.h. Since this is a one-time migration, I've removed trim_dmp.py from this PR. The change is now limited to the three .sha1 updates. I'll keep the script locally in case we need to regenerate assets in the future.

starboard/shared/starboard/player/testdata/beneath_the_canopy_135_avc.dmp.sha1
AIdrifter: Thanks, We already uploaded it.

starboard/tools/trim_dmp.py
AIdrifter: Hi @jasonzhangxx,

**1) Binary size** — the `nplb` ELF is unchanged, the DMPs are runtime content. The packaged test data drops from 154.51 MB to 10.88 MB (-93%).

**2) AVC and VP9 playback** — both pass. Ran the full `nplb` player set on `linux-x64x11_devel` against the trimmed assets: **341 tests, 0 failures** (273 ran, 68 skipped — the usual audio-only and DRM cases). `137_avc`, `248_vp9` and `135_avc` are all covered.

**3) The files can't be downloaded from GCS.** All three objects exist in `cobalt-static-storage-public`, but each one is ~8.5 KB of HTML instead of the DMP payload:

| Object (= expected SHA1) | Expected | Actually in bucket |
| :--- | ---: | ---: |
| `dce594f7f4298c29b5e6e8b1acf37a6b056b7823` (135_avc) | 1,373,974 B | 8,542 B |
| `a825b3d7ff03cd0a233e8b2e310756f04a6ea362` (137_avc) | 4,224,559 B | 8,546 B |
| `5c88067f5ab55dd9f96b1cce5bd79a30724a08b8` (248_vp9) | 5,807,118 B | 8,542 B |

So `download_from_gcs.py` fails its SHA1 check and the build breaks:

```
[ERROR:download_from_gcs.py:89] Download failed: 'Local and remote sha1s do not match. Skipping download.', retrying
GcsDownloadException: Local and remote sha1s do not match. Skipping download.
```

What's the correct way to get these uploaded? My verified files are at `/google/data/rw/users/pi/pinkuschang/cobalt-pr12264/` (object name = content SHA1). Happy to re-upload myself if you can grant me write access to the bucket.

PR #12266, opened 2026-08-28
Update the Starboard version convention

y4vor:
Bug: 553997935

### DIFF
--- a/starboard/configuration.h
+++ b/starboard/configuration.h
@@ -35,13 +35,11 @@
 
 // --- Common Defines --------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


andrewsavage1 (commented):
I think this change should be made directly to the 27.lts branch

y4vor:
> I think this change should be made directly to the 27.lts branch

The main branch is also max 18 right now. 
Whenever we start working on 19 we would bump it and introduce 19 builders.

andrewsavage1 (approved):

PR #12323, opened 2026-08-31
Enforce default max cache size

sacuff:
Sets `network_context_params->http_cache_max_size` to kSbMaxSystemPathCacheDirectorySize by default, which was the behavior on C25. Otherwise, the default size of 0 allows Cobalt to use all available disk space for cache.

Bug: 527972396

### DIFF
--- a/cobalt/browser/cobalt_content_browser_client.cc
+++ b/cobalt/browser/cobalt_content_browser_client.cc
@@ -28,6 +28,7 @@
 #include "base/metrics/field_trial_params.h"
 #include "base/metrics/histogram_functions.h"
 #include "base/metrics/histogram_macros.h"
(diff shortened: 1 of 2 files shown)


cobalt/browser/cobalt_content_browser_client.cc
```diff
@@ -414,6 +416,11 @@ void CobaltContentBrowserClient::ConfigureNetworkContextParams(
     network_context_params->file_paths->http_cache_directory =
         cache_path.Append(kCacheDirname);
 
+#if BUILDFLAG(IS_STARBOARD)
```
y4vor, line 419: maybe do IS_COBALT
@johnxwork what do you think?

PR #12328, opened 2026-09-01
Fix Omaha version spoofing vulnerability by parsing manifest.json

yuying-y:
Instead of relying on the unverified version string received in the Omaha updatecheck response, actively parse the `manifest.json` from the unpacked payload using `ReadEvergreenVersion` to dictate the reliable version written into the local persistent data. This prevents an attacker from inflating the version string to permanently inhibit future updates.

Bug: 555428363
TAG=agy

### DIFF
--- a/components/update_client/op_install.cc
+++ b/components/update_client/op_install.cc
@@ -44,7 +44,6 @@
 
 namespace update_client {
 
-
 namespace {
 
 // The sequence of calls is:
--- a/components/update_client/op_install.cc
+++ b/components/update_client/op_install.cc
@@ -38,10 +38,12 @@
 
 #if BUILDFLAG(IS_STARBOARD)
 #include "components/update_client/cobalt_slot_management.h"
+#include "components/update_client/utils.h"
 #include "starboard/extension/installation_manager.h"
 #endif
 
 namespace update_client {
+
 
 namespace {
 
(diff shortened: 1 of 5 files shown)


yuying-y:
Fixed in the latest commit! The fallback mechanism has been stripped. The update client now logs an error and actively skips updating the internal  metadata altogether if the manifest is unreadable or malformed.

components/update_client/op_install.cc
```diff
@@ -190,17 +192,22 @@ void Install(base::OnceCallback<void(const CrxInstaller::Result&)> callback,
...
+    } else if (!manifest_version.IsValid()) {
+      LOG(ERROR) << "Invalid Evergreen manifest. Installation aborted.";
+      install_error = InstallError::GENERIC_ERROR;
     } else if (CobaltFinishInstallation(
                    installation_api, crx_operation_result.installation_index,
                    result.unpack_path.value(), app_key)) {
       // Write the version of the unpacked update package to the persisted data.
       if (metadata != nullptr) {
         metadata->SetLastInstalledEgAndSbVersion(
-            id, next_version, std::to_string(SB_API_VERSION));
```
hlwarriner, line 203: The `next_version` parameter is unused by this function now. Furthermore, it looks like we had to make a number of `#if BUILDFLAG(IS_STARBOARD)`-guarded customizations to plumb this value down through the call stack and that these can now be removed.

It would be good to clean this parameter up throughout the call stack. A follow-up PR would be fine since the cleanup is less urgent than addressing the vulnerability.

hlwarriner (approved):
I left a comment about some cleanup we should do now or soon but this looks like a good change. Thanks!

components/update_client/op_install.cc
yuying-y: Good catch! resolved

hlwarriner: Thanks!

PR #12336, opened 2026-09-01
autoroll_chromium: Check out the staging branch by default

rakuco:
Instead of checking out main and spending time switching to the staging branch later, check out the latter directly and just use `git fetch` for main (so the workflow can check out the autoroller scripts). This should hopefully make the Autoroll Chromium run a bit faster.

Bug: 543555380


rakuco:
cc @abhijeetk @jkim-julie @Gyuyoung @sazabi1976

rakuco:
@briantting it's hard to test this locally, but conceptually this seems to make sense. could you take a look?

briantting:
tested this with 

https://github.com/youtube/cobalt/actions/runs/34537688491/job/103073010557

compared to

https://github.com/youtube/cobalt/actions/runs/34534409688/job/103062558212

and there wasn't much of a performance improvement (`Checkout code` step) by checking out staging over main. The majority of the time is taken up by the `Cherry Pick Merged PRs` step, specifically the multiple gclient syncs.

I also had to modify the `Prepare Cherry Pick Branch` step because of an error around:

```
you cannot fetch a remote branch directly into a local branch that is currently checked out.
```

with the applied suggestions I tested this with

https://github.com/youtube/cobalt/actions/runs/34539081004/job/103077373820

which ran the `Prepare Cherry Pick Branch` step a tad faster.

The overall improvement in times seems minimal but still a worthwhile change. Improves readability as we won't confuse people why main "needs" to be checked out.

.github/workflows/autoroll_chromium.yaml
briantting:
```suggestion
          git fetch --depth="${FETCH_DEPTH}" --no-recurse-submodules origin "${TARGET_BRANCH}"
```

.github/workflows/autoroll_chromium.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           path: src
-          ref: main
+          ref: staging
```
briantting:
```suggestion
          ref: ${{ matrix.branch.target }} 
```

briantting (changes requested):

.github/workflows/autoroll_chromium.yaml
briantting: to keep the script "general"; this is especially needed after the above change from `${TARGET_BRANCH}":${TARGET_BRANCH}"` to just `"${TARGET_BRANCH}"`

rakuco: done

.github/workflows/autoroll_chromium.yaml
rakuco: Can you take a look again? I don't think this line is even necessary anymore given this branch is fetched and checked out above now.

briantting (approved):

PR #12338, opened 2026-09-01
cobalt: fix build after HangWatcher shutdown fixes

rogerzanoni:
PR #12303 moved the release into PostSandboxInitialized() and removed the unregister_thread_closure member from the header, but did not touch the destructor, leaving a reference to the deleted member:

  cobalt/renderer/cobalt_content_renderer_client.cc: error:
    use of undeclared identifier 'unregister_thread_closure'

Drop the dead release call and the comment.

Bug: 532068409


andrewsavage1:
Done in https://github.com/youtube/cobalt/pull/12337

PR #12345, opened 2026-09-01
build: Pin src/cobalt/internal to specific hash

fayev-g:
Update the src/cobalt/internal dependency in DEPS to reference a
fixed commit hash rather than the main branch.

Bug: 555839490


DEPS
```diff
@@ -1525,7 +1525,7 @@ deps = {
   },
 
   'src/cobalt/internal': {
-    'url': Var('cobalt_internal_git') + '/cobalt/internal.git' + '@' + 'main',
+    'url': Var('cobalt_internal_git') + '/cobalt/internal.git' + '@' + 'f7def3270058e9eb155cbb0140ffdf6633aef04b',
```
oxve, line 1528: This points to a commit on main. Before 28.lts we want to change this to one on 27.lts.

Don't change anything now. We will need to come up with a strategy to update release branches on cobalt/internal.

oxve (approved):

DEPS
fayev-g: ack! yeah that makes sense

PR #12348, opened 2026-09-01
cobalt: Restore launch=preload query on startup

ryannchoi:
Restore behavior from Cobalt 25. Append "launch=preload" to the document URL when the application is launched in preload mode.

As a result, this notifies Kabuki to start in background mode. This prevents potential wastage of TV resources, and allows instantaneous launching of the YouTube TV app.

Bug: 503025265


ryannchoi:
reopening with member permissions

PR #12384, opened 2026-09-02
M141.7390: fix conflicts, builds and tests

rakuco:
Bug: 555767704

### DIFF
--- a/components/autofill/core/browser/foundations/autofill_manager.cc
+++ b/components/autofill/core/browser/foundations/autofill_manager.cc
@@ -930,6 +930,8 @@ void AutofillManager::SubscribeToMlModelChanges(
     default:
       NOTREACHED();
   }
+#else
+  NOTREACHED();
 #endif
 }
 
--- a/components/autofill/core/browser/foundations/autofill_manager.cc
+++ b/components/autofill/core/browser/foundations/autofill_manager.cc
@@ -909,6 +909,7 @@ void AutofillManager::LogCurrentFieldTypes(const FormStructure& form) {
 void AutofillManager::SubscribeToMlModelChanges(
     FieldClassificationModelHandler& handler,
     optimization_guide::proto::OptimizationTarget optimization_target) {
+#if BUILDFLAG(BUILD_WITH_TFLITE_LIB)
   switch (optimization_target) {
     case optimization_guide::proto::OptimizationTarget::
         OPTIMIZATION_TARGET_AUTOFILL_FIELD_CLASSIFICATION:
@@ -929,6 +930,7 @@ void AutofillManager::SubscribeToMlModelChanges(
     default:
       NOTREACHED();
   }
+#endif
 }
 
 }  // namespace autofill
--- a/components/update_client/utils.cc
+++ b/components/update_client/utils.cc
@@ -101,30 +101,23 @@ std::string GetCrxIdFromPublicKeyHash(base::span<const uint8_t> pk_hash) {
 
 #if BUILDFLAG(IS_STARBOARD)
 #if defined(IN_MEMORY_UPDATES)
-
 bool VerifyHash256(const std::string* content,
                    const std::string& expected_hash_str) {
-  std::vector<uint8_t> expected_hash;
-  if (!base::HexStringToBytes(expected_hash_str, &expected_hash) ||
-      expected_hash.size() != crypto::kSHA256Length) {
+  std::array<uint8_t, crypto::hash::kSha256Size> expected_hash;
+  if (!base::HexStringToSpan(expected_hash_str, expected_hash)) {
     return false;
   }
 
-  uint8_t actual_hash[crypto::kSHA256Length] = {0};
-  std::unique_ptr<crypto::SecureHash> hasher(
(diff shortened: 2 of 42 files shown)


rakuco:
cc @sazabi1976 @abhijeetk @jkim-julie @Gyuyoung

components/viz/service/gl/gpu_service_impl.cc
```diff
@@ -1376,10 +1377,20 @@ bool GpuServiceImpl::IsGMBNV12Supported() {
...
-                                : nullptr);
+  auto size = gfx::Size(2, 2);
+  scoped_refptr<gfx::NativePixmap> pixmap =
+      ui::OzonePlatform::GetInstance()
+          ->GetSurfaceFactoryOzone()
+          ->CreateNativePixmap(gpu::kNullSurfaceHandle,
+                               vulkan_context_provider()
+                                   ? vulkan_context_provider()->GetDeviceQueue()
+                                   : nullptr,
+                               size, buffer_format, buffer_usage, size);
```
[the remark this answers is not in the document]
rakuco: this is coming directly from upstream

components/autofill/core/browser/foundations/autofill_manager.cc
```diff
@@ -909,6 +909,7 @@ void AutofillManager::LogCurrentFieldTypes(const FormStructure& form) {
 void AutofillManager::SubscribeToMlModelChanges(
     FieldClassificationModelHandler& handler,
     optimization_guide::proto::OptimizationTarget optimization_target) {
```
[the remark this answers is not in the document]
rakuco: this hasn't been an issue on any of the platforms we care about. NAK

components/update_client/utils.cc
```diff
@@ -101,30 +101,23 @@ std::string GetCrxIdFromPublicKeyHash(base::span<const uint8_t> pk_hash) {
 
 #if BUILDFLAG(IS_STARBOARD)
 #if defined(IN_MEMORY_UPDATES)
-
 bool VerifyHash256(const std::string* content,
                    const std::string& expected_hash_str) {
-  std::vector<uint8_t> expected_hash;
-  if (!base::HexStringToBytes(expected_hash_str, &expected_hash) ||
-      expected_hash.size() != crypto::kSHA256Length) {
+  std::array<uint8_t, crypto::hash::kSha256Size> expected_hash;
```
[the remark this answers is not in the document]
rakuco: this wasn't done in the previous version, if really relevant it should be done separately

rakuco:
This is ready for review. The CI failures look unrelated, but it's hard to know for sure since none of them include the actual failure logs AFAICS.

This PR will probably need to be rebased once @briantting lands some Siso-related changes to #12382 now that we've reached 141.7390.

components/autofill/core/browser/foundations/autofill_manager.cc
```diff
@@ -929,6 +930,7 @@ void AutofillManager::SubscribeToMlModelChanges(
     default:
       NOTREACHED();
```
andrewsavage1, line 931: We should consider having

```
#else
  NOTREACHED();
#endif
```

To make it clear we don't support this (assuming this isn't actively called unless web apis related to it are used)

components/viz/service/gl/gpu_service_impl.cc
```diff
@@ -1376,10 +1377,20 @@ bool GpuServiceImpl::IsGMBNV12Supported() {
     return false;
   }
 
-  return shared_image_manager()->CanCreateNativePixmap(
-      buffer_format, buffer_usage,
-      vulkan_context_provider() ? vulkan_context_provider()->GetDeviceQueue()
-                                : nullptr);
+  auto size = gfx::Size(2, 2);
```
andrewsavage1, line 1380: What does this configure? I see it comes from the linux implementation but is it possible we'd want something different here?

andrewsavage1 (approved):

components/viz/service/gl/gpu_service_impl.cc
andrewsavage1: Looking further, it looks like this comes from upstream directly...so I'm confused why it's something we had to change in this PR?

rakuco: The original code came from https://chromium-review.googlesource.com/c/chromium/src/+/6810905, which broke the Cobalt build as explained in https://github.com/Igalia/cobalt/commit/ac2a610dd952b5ccc76f7675c2408b9a9beca4e2. This specific bit is from https://chromium-review.googlesource.com/c/chromium/src/+/6917616, which lander after 141.7390 and moved this code from gpu/command_buffer/service/shared_image/shared_image_manager.h to its current location.

Not sure what the option would be though; compared to the version in main the biggest change was [this one](https://chromium-review.googlesource.com/c/chromium/src/+/6802066) FWIW.

components/autofill/core/browser/foundations/autofill_manager.cc
rakuco: done

PR #12385, opened 2026-09-02
build: Remove dead skia/ext/fontmgr_default_starboard.cc

hferreiro:
This file is no longer used.

Bug: 492403745


andrewsavage1 (approved):
I wonder if there's another place we'll have to configure something similar or if it's all handled in skfontmgr_cobalt now

hferreiro:
> I wonder if there's another place we'll have to configure something similar or if it's all handled in skfontmgr_cobalt now

It appears that the platform overrides are all centralized at [`fontmgr_factory()`](https://github.com/youtube/cobalt/blob/a1bd1f8233e5447cc182e4d29d2b8859c311410f/skia/ext/font_utils.cc#L67).

PR #12386, opened 2026-09-02
Prevent exit segmentation fault in ElfLoader by using NoDestructor

andrewsavage1:
### Summary
In single-process mode, InProcessRendererThread is intentionally released during termination (PR #11796) to avoid teardown deadlocks when task queues are throttled. However, because loader_app and elf_loader_sandbox held the global ElfLoader as a normal static object with automatic destruction, returning from main() caused ElfLoader's destructor to call munmap() on libcobalt.so's load range while InProcessRendererThread was still executing (e.g., handling V8 garbage collection triggered by memory pressure during conceal). This resulted in an instruction fetch page fault (SIGSEGV) on exit.

Wrapping the ElfLoader instance in starboard::NoDestructor prevents it from unmapping libcobalt.so during exit, allowing the operating system to cleanly reclaim the process memory space and terminate all threads.

### Testing
- Reproduced exit SIGSEGV and core dump on evergreen-x64_qa with loader_app.
- Symbolized crash trace pointing to v8::internal::Heap::IterateWeakRoots() on unmapped memory right after ProgramTable::~ProgramTable() unmapped libcobalt.so.
- Verified that with this change, loader_app exits cleanly with code 0 and zero segmentation faults or core dumps across multiple runs.
- Verified that loader_app and elf_loader_sandbox build cleanly.

Bug: 555199636
TAG=agy
CONV=ae349c26-3111-460f-a5f3-d2cb6fe2950c

### DIFF
--- a/starboard/loader_app/loader_app.cc
+++ b/starboard/loader_app/loader_app.cc
@@ -24,6 +24,7 @@
 #include "starboard/common/check_op.h"
 #include "starboard/common/command_line.h"
 #include "starboard/common/log.h"
+#include "starboard/common/no_destructor.h"
 #include "starboard/common/paths.h"
 #include "starboard/common/string.h"
 #include "starboard/configuration.h"
@@ -64,7 +65,10 @@ const char kSystemImageZstdCompressedLibraryPath[] =
 const char kSystemImageManifestPath[] = "app/cobalt/manifest.json";
 
 // Portable ELF loader.
-elf_loader::ElfLoader g_elf_loader;
+elf_loader::ElfLoader& GetElfLoader() {
+  static starboard::NoDestructor<elf_loader::ElfLoader> s_elf_loader;
+  return *s_elf_loader;
+}
 
 // Pointer to the |SbEventHandle| function in the
 // Cobalt binary.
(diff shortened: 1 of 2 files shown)


starboard/loader_app/loader_app.cc
```diff
@@ -64,7 +65,10 @@ const char kSystemImageZstdCompressedLibraryPath[] =
 const char kSystemImageManifestPath[] = "app/cobalt/manifest.json";
 
 // Portable ELF loader.
-elf_loader::ElfLoader g_elf_loader;
+elf_loader::ElfLoader& GetElfLoader() {
+  static starboard::NoDestructor<elf_loader::ElfLoader> s_elf_loader;
+  return *s_elf_loader;
+}
```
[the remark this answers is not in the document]
andrewsavage1: It already is in an anonymous namespace

johnxwork (approved):

starboard/loader_app/loader_app.cc
```diff
@@ -64,7 +65,10 @@ const char kSystemImageZstdCompressedLibraryPath[] =
 const char kSystemImageManifestPath[] = "app/cobalt/manifest.json";
 
 // Portable ELF loader.
-elf_loader::ElfLoader g_elf_loader;
+elf_loader::ElfLoader& GetElfLoader() {
+  static starboard::NoDestructor<elf_loader::ElfLoader> s_elf_loader;
```
hlwarriner, line 69: For context on the use of `starboard::NoDestructor` here, consider adding a short comment summarizing what you explained in the PR description. Something like: "To avoid segmentation faults and other issues, we need to keep libcobalt.so mapped while un-joinable background threads are still running."

hlwarriner (approved):

y4vor (approved):

starboard/loader_app/loader_app.cc
andrewsavage1: Will add this in a follow up to submit this asap

PR #12390, opened 2026-09-02
cobalt: Eliminate startup and resume white/red flashes

jellefoks:
Change the default background color in RenderWidgetHostViewBase and
the fallback texture color in ImageContextImpl to black.

On TV platforms, white flashes occur during cold starts before the
first layout is calculated. Additionally, red or white flashes
appear on resume when the Viz compositor lacks valid mailboxes
during reinitialization. Forcing these defaults to black ensures
a seamless transition until web content is ready to be displayed.

Bug: 489291655
Bug: 490989945
Bug: 549496958
Bug: 553317436

### DIFF
--- a/components/viz/service/display_embedder/image_context_impl.cc
+++ b/components/viz/service/display_embedder/image_context_impl.cc
@@ -6,6 +6,7 @@
 
 #include <utility>
 
+#include "build/build_config.h"
 #include "base/check.h"
 #include "base/check_op.h"
 #include "base/metrics/histogram_functions.h"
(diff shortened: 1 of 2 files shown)


components/viz/service/display_embedder/image_context_impl.cc
```diff
@@ -91,12 +92,9 @@ bool DawnYCbCrVkDescriptorsAreCompatible(const wgpu::YCbCrVkDescriptor& left,
...
                                    int plane_index) {
   DCHECK(format.IsValidPlaneIndex(plane_index));
-  // For DCHECKed builds return: 1) kRed for single planar formats; 2) kWhite
-  // for all multiplanar format planes that translates to Purple/Magenta color
-  // in RGB space. For non-DCHECKed builds return: 1) kWhite for single planar
-  // formats; 2) kWhite for Y, A planes and kGray for U, V, UV planes that
-  // translates to White color in RGB colorspace.
-#if DCHECK_IS_ON()
+#if BUILDFLAG(IS_COBALT)
+  return SkColors::kBlack;
```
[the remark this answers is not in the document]
jellefoks: Done. Handled multiplanar formats properly by returning `SkColors::kBlack` for luma (Y), `SkColors::kGray` for chroma (U, V, UV), and `SkColors::kWhite` for alpha (A), with code comments detailing the YUV color space mechanics.

content/browser/renderer_host/render_widget_host_view_base.cc
```diff
@@ -307,7 +307,11 @@ void RenderWidgetHostViewBase::SetBackgroundColor(SkColor color) {
 std::optional<SkColor> RenderWidgetHostViewBase::GetBackgroundColor() {
   if (content_background_color_)
     return content_background_color_;
+#if BUILDFLAG(IS_COBALT)
+  return default_background_color_.value_or(SK_ColorBLACK);
```
andrewsavage1, line 311: I thought this was something we could configure in the cobalt embedder?

jellefoks: The call to `views::WebContentsSetBackgroundColor::CreateForWebContentsWithColor` from `cobalt/shell/browser/shell_platform_delegate_views.cc` appears to accomplish it, but it doesn't, for two reasons.

On android-arm builds, that `shell_platform_delegate_views.cc` isn't used (use_aura=false), and `cobalt/shell/browser/shell_platform_delegate_android.cc` is used instead. Of course that call could probably be added there too.

But there is more, the call to `CreateForWebContentsWithColor` would race `DelegatedFrameHost::GetGpuCompositionRootColor`, and if the latter wins the race, it will return a white frame, that is one of the flashes we see before this PR.

And by using `default_background_color_.value_or(SK_ColorBLACK)`, any explicit call to `SetBackgroundColor(SK_ColorTRANSPARENT)` (as added in #10696 for b/495438901) is still 100% respected. It only falls back to black when the color is uninitialized (std::nullopt).

andrewsavage1 (approved):

PR #12407, opened 2026-09-02
Check GL_RENDERER for ANGLE when querying requestable extensions

haozheng-cobalt:
Starboard Ozone reports kGLImplementationEGLANGLE to satisfy passthrough
command decoder constraints even when running directly on native GLES
drivers (such as ARM Mali). Checking GetGLImplementation() therefore
incorrectly identifies native GLES contexts as ANGLE and attempts to
query GL_REQUESTABLE_EXTENSIONS_ANGLE (0x93A8), generating a GL_INVALID_ENUM
error.

Inspect glGetString(GL_RENDERER) directly to determine whether the
underlying driver is truly ANGLE before querying requestable extensions.
This keeps the GL error state clean and prevents subsequent fatal
glGetError() DCHECK aborts during context initialization.

Bug: 537056801

### DIFF
--- a/ui/gl/gl_implementation.cc
+++ b/ui/gl/gl_implementation.cc
@@ -503,8 +503,15 @@ gfx::ExtensionSet GetRequestableGLExtensionsFromCurrentContext() {
 
 gfx::ExtensionSet GetRequestableGLExtensionsFromCurrentContext(GLApi* api) {
 #if BUILDFLAG(IS_COBALT)
-  if (!api || GetGLImplementation() != kGLImplementationEGLANGLE)
+  if (!api)
     return gfx::ExtensionSet();
+
+  const char* renderer_str =
+      reinterpret_cast<const char*>(api->glGetStringFn(GL_RENDERER));
+  if (!renderer_str ||
+      !base::StartsWith(renderer_str, "ANGLE", base::CompareCase::SENSITIVE)) {
+    return gfx::ExtensionSet();
+  }
 #endif
   return GetGLExtensionsFromCurrentContext(api,
                                            GL_REQUESTABLE_EXTENSIONS_ANGLE);


gpu/command_buffer/service/feature_info.cc
```diff
@@ -275,6 +276,52 @@ void FeatureInfo::InitializeForTesting(ContextType context_type) {
              DisallowedFeatures());
 }
 
+#if BUILDFLAG(IS_STARBOARD)
```
niranjanyardi: can we push this inside the existing method on line 325 instead of adding a new method ?

jellefoks:
I'm think that the `IsGL_REDSupportedOnFBOs` part is a red herring, since it triggers on the `DCHECK(glGetError() == GL_NO_ERROR);` in there but it does it before the function does any gl calls.

I think this may be a case of there being some other code leaving behind an uncleared glError, and while it gets detected during `IsGL_REDSupportedOnFBOs`, that is not actually where the bug is (similar to how the passthrough command decoder initialization did for its `FAIL_INIT_IF_NOT` use before I added the `while (glGetError() != GL_NO_ERROR) {}` glError clearing in #10934.

In which case this PR would be addressing a symptom instead of a root cause.

jellefoks:
The bug could be that there is still a `glGetString(value)` call for an invalid value somewhere in the code resulting in a GL_INVALID_ENUM dangling glError, meaning that the root cause is there, either needs a logic fix or a correctly aimed while (glGetError() != GL_NO_ERROR) {}, or maybe a single glGetError() may already clear it...

Similar to how `GetString(GL_EXTENSIONS)` got called, but it is deprecated in GLESv3, so I had to change it to use `GetStringi(GL_EXTENSIONS, i)` when it's GLESv3 see (part of) #10934

ui/gl/gl_implementation.cc
```diff
@@ -503,8 +503,15 @@ gfx::ExtensionSet GetRequestableGLExtensionsFromCurrentContext() {
 
 gfx::ExtensionSet GetRequestableGLExtensionsFromCurrentContext(GLApi* api) {
 #if BUILDFLAG(IS_COBALT)
-  if (!api || GetGLImplementation() != kGLImplementationEGLANGLE)
```
haozheng-cobalt, line 506: link

jellefoks (approved):
This should be right

PR #12411, opened 2026-09-02
cobalt: Provide a dedicated path for Punch-out GPU Posting

johnedocampo:
Introduce RunStandaloneFunctionOnGpu to StarboardGpuFactory to
support Punch-out mode GPU task calls.  Using a dedicated path avoids a
dependency on CommandBufferStub, which is required by the standard
decode target execution path. When not using decode-to-texture,
it's possible that CommandBufferStub will not be set, causing issues
when clearing the surface view.

This ensures that when an overlay plane is active, the renderer
wrapper correctly routes tasks that do not rely on the internal
command buffer state.

Bug: 555127326

### DIFF
--- a/media/gpu/starboard/starboard_gpu_factory_impl.cc
+++ b/media/gpu/starboard/starboard_gpu_factory_impl.cc
@@ -67,6 +67,15 @@ void StarboardGpuFactoryImpl::RunSbDecodeTargetFunctionOnGpu(
   done_event->Signal();
 }
 
+void StarboardGpuFactoryImpl::RunStandaloneFunctionOnGpu(
+    SbDecodeTargetGlesContextRunnerTarget target_function,
+    void* target_function_context,
+    base::WaitableEvent* done_event) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  target_function(target_function_context);
+  done_event->Signal();
+}
+
 void StarboardGpuFactoryImpl::RunCallbackOnGpu(
     base::OnceCallback<void()> callback,
     base::WaitableEvent* done_event) {
--- a/media/mojo/services/starboard/starboard_renderer_wrapper.cc
+++ b/media/mojo/services/starboard/starboard_renderer_wrapper.cc
@@ -721,9 +721,18 @@ void StarboardRendererWrapper::GraphicsContextRunner(
     base::WaitableEvent done_event(
         base::WaitableEvent::ResetPolicy::MANUAL,
         base::WaitableEvent::InitialState::NOT_SIGNALED);
-    provider->gpu_factory_
-        .AsyncCall(&StarboardGpuFactory::RunSbDecodeTargetFunctionOnGpu)
-        .WithArgs(target_function, target_function_context, &done_event);
+    // In non decode-to-texture playbacks, post standalone functions
+    // (which manage their own EGL context) via RunStandaloneFunctionOnGpu to
+    // avoid dependency on CommandBufferStub.
+    if (!provider->overlay_plane_id_.is_empty()) {
+      provider->gpu_factory_
+          .AsyncCall(&StarboardGpuFactory::RunStandaloneFunctionOnGpu)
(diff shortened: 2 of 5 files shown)


media/gpu/starboard/starboard_gpu_factory_impl.cc
```diff
@@ -67,6 +67,17 @@ void StarboardGpuFactoryImpl::RunSbDecodeTargetFunctionOnGpu(
...
 
+void StarboardGpuFactoryImpl::RunStandaloneFunctionOnGpu(
+    SbDecodeTargetGlesContextRunnerTarget target_function,
+    void* target_function_context,
+    base::WaitableEvent* done_event) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  if (target_function) {
+    target_function(target_function_context);
+  }
+  done_event->Signal();
```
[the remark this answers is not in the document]
johnedocampo: To have the behavior be as similar to RunSbDecodeTargetFunctionOnGpu, we will keep the changes as is for now.

media/gpu/starboard/starboard_gpu_factory_impl.cc
```diff
@@ -67,6 +67,15 @@ void StarboardGpuFactoryImpl::RunSbDecodeTargetFunctionOnGpu(
   done_event->Signal();
 }
 
+void StarboardGpuFactoryImpl::RunStandaloneFunctionOnGpu(
+    SbDecodeTargetGlesContextRunnerTarget target_function,
+    void* target_function_context,
+    base::WaitableEvent* done_event) {
+  DCHECK_CALLED_ON_VALID_THREAD(thread_checker_);
+  target_function(target_function_context);
+  done_event->Signal();
```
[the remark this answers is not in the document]
johnedocampo: To follow the pre-existing structure of `RunSbDecodeTargetFunctionOnGpu`, let's keep this as-is.

media/mojo/services/starboard/starboard_renderer_wrapper.cc
```diff
@@ -732,9 +721,18 @@ void StarboardRendererWrapper::GraphicsContextRunner(
     base::WaitableEvent done_event(
         base::WaitableEvent::ResetPolicy::MANUAL,
         base::WaitableEvent::InitialState::NOT_SIGNALED);
-    provider->gpu_factory_
-        .AsyncCall(&StarboardGpuFactory::RunSbDecodeTargetFunctionOnGpu)
-        .WithArgs(target_function, target_function_context, &done_event);
+    // In non decode-to-texture playbacks, post standalone functions
```
kjyoun, line 724: how about this?

Then difference for 2 code path is evident

```
// In non-decode-to-texture playback modes (such as hardware overlay planes),
// execute the target task via RunStandaloneFunctionOnGpu. Standalone functions
// manage their own dedicated EGL/GL context lifecycle directly, eliminating any
// runtime dependency on CommandBufferStub.
//
// For decode-to-texture playbacks, route through RunSbDecodeTargetFunctionOnGpu
// to ensure proper binding and handling within the standard SbDecodeTarget pipeline.
const auto gpu_task_runner = !provider->overlay_plane_id_.is_empty()
                                 ? &StarboardGpuFactory::RunStandaloneFunctionOnGpu
                                 : &StarboardGpuFactory::RunSbDecodeTargetFunctionOnGpu;

provider->gpu_factory_
    .AsyncCall(gpu_task_runner)
    .WithArgs(target_function, target_function_context, &done_event);
```

borongc: I put a comment in the bug, I think we should figure out why stub_ is null, and can we recover it from sleep mode for punch-out, so we don't need to make this code path diverged.

johnedocampo: Closing this PR in favor for https://github.com/youtube/cobalt/pull/12531.

PR #12416, opened 2026-09-03
ci: add AI rebase agent workflow

linxinan-yt:
This PR introduces the automated AI Rebase Agent workflow (`.github/workflows/ai_rebase_resolver.yaml`) to `staging`.

### Key Features
- **Targeted Trigger**: Automatically kicks off whenever a Chromium rebase conflict PR is opened against `staging` (`head.ref` starts with `autoroll-chromium/` and title starts with `CONFLICTED`).
- **Excludes Cobalt Cherry-picks**: Cobalt internal cherry-picks (`autoroll-main-to-staging`) and regular developer PRs are explicitly ignored.
- **Dynamic Tooling**: Borrows latest rebase tooling directly from `origin/experimental/rebase_agent`. The agent loop scripts will be relocated later.
- The agent creates and uploads the fix patchset only. The following non-AI action download the changes and files the PR. 

Verifications:
 - https://github.com/youtube/cobalt/actions/runs/34272980628/job/102223160837

Bug: 552561778

### DIFF
--- a/.github/workflows/ai_rebase_resolver.yaml
+++ b/.github/workflows/ai_rebase_resolver.yaml
@@ -1,0 +1,273 @@
+name: AI Rebase Agent
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: 'PR number to resolve (optional, e.g. 12155)'
+        required: false
+        type: string
+        default: ''
+      commit_sha:
+        description: 'Specific starting commit SHA (optional)'
+        required: false
+        type: string
+        default: ''
+      ai_branch_prefix:
+        description: 'Prefix for the isolated AI working branch'
+        required: false
+        type: string
+        #TODO(b/557260018): Create a dedicated branch prefix for AI agent.
+        default: 'experimental/rebase_agent_pr_'
+      agent_ref:
+        description: 'Branch containing the AI rebase agent scripts'
+        required: false
+        type: string
+        #TODO(b/552561778): Move the agent script to main branch once it is stable.
+        default: 'experimental/rebase_agent'
+      dry_run:
+        description: 'Dry run mode (runs AI resolver but skips pushing commit & creating PR)'
+        type: boolean
+        default: true
+      mode:
+        description: 'Execution mode'
+        type: choice
+        options:
+          - full-pipeline
+          - resolve-conflicts
+          - gn-gen
+          - build-only
+        default: full-pipeline
+
+permissions: {}
+
+jobs:
+  ai_resolve:
+    name: Rebase Agent
+    permissions:
+      contents: write
+    runs-on: [self-hosted, chrobalt-linux-runner]
+    container:
+      image: ghcr.io/youtube/cobalt/linux:main
+      options: --device /dev/fuse --cap-add SYS_ADMIN --security-opt apparmor=unconfined
+    defaults:
+      run:
+        shell: bash
+    outputs:
+      resolved: ${{ steps.run_ai.outputs.resolved }}
+      ai_branch: ${{ steps.branch_setup.outputs.ai_branch }}
+      pr_num: ${{ steps.branch_setup.outputs.pr_num }}
+      base_branch: ${{ steps.branch_setup.outputs.base_branch }}
+      has_commits: ${{ steps.commit_push.outputs.has_commits }}
+    env:
+      DEPOT_TOOLS_METRICS: 0
+      GIT_AUTHOR_NAME: "cobalt-github-releaser-bot"
+      GIT_AUTHOR_EMAIL: "cobalt-github-releaser-bot@google.com"
+      GIT_COMMITTER_NAME: "cobalt-github-releaser-bot"
+      GIT_COMMITTER_EMAIL: "cobalt-github-releaser-bot@google.com"
+      GITHUB_TOKEN: ${{ secrets.CHERRY_PICK_TOKEN }}
+      RBE_INSTANCE: "projects/cobalt-actions-prod/instances/default_instance"
+      RBE_SERVICE: "remotebuildexecution.googleapis.com:443"
+      SISO_CREDENTIAL_HELPER: "gcloud"
+      GCP_PROJECT: ${{ vars.GCP_PROJECT || 'lxn-test' }}
+      GCP_REGION: "us-central1"
+      #TODO(b/552561778): Move to dedicated production reasoning engine ID once finalized.
+      REASONING_ENGINE_ID: ${{ vars.REASONING_ENGINE_ID || '6110792607330205696' }}
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: src
+          fetch-depth: 100
+          token: ${{ env.GITHUB_TOKEN }}
+
+      - name: GoB Authentication
+        uses: ./src/.github/actions/gob_auth
+        with:
+          use_gcloud: 'true'
+
+      - name: Setup Working Branch
+        id: branch_setup
+        working-directory: src
+        run: |
+          set -x
+
+          PR_NUM="${{ inputs.pr_number }}"
+          COMMIT_SHA="${{ inputs.commit_sha }}"
+          AI_PREFIX="${{ inputs.ai_branch_prefix }}"
+          AGENT_REF="${{ inputs.agent_ref }}"
+
+          # 1. Checkout starting target: PR (with optional commit SHA) or direct commit SHA
+          if [[ -n "${PR_NUM}" ]]; then
+            echo "Fetching PR #${PR_NUM}..."
+            git fetch origin "pull/${PR_NUM}/head:pr-${PR_NUM}"
+            git checkout "pr-${PR_NUM}"
+
+            if [[ -n "${COMMIT_SHA}" ]]; then
+              echo "Checking out target commit SHA inside PR #${PR_NUM}: ${COMMIT_SHA}..."
+              git checkout "${COMMIT_SHA}"
+              SHORT_SHA=$(git rev-parse --short "${COMMIT_SHA}")
+              BASE_NAME="pr_${PR_NUM}_${SHORT_SHA}"
+              AI_BRANCH="${AI_PREFIX}pr_${PR_NUM}_${SHORT_SHA}"
+            else
+              BASE_NAME="pr_${PR_NUM}"
+              AI_BRANCH="${AI_PREFIX}pr_${PR_NUM}"
+            fi
+          elif [[ -n "${COMMIT_SHA}" ]]; then
+            echo "Fetching and checking out commit SHA: ${COMMIT_SHA}..."
+            git fetch --depth=100 origin "${COMMIT_SHA}" || git fetch origin --tags
+            git checkout "${COMMIT_SHA}"
+            SHORT_SHA=$(git rev-parse --short "${COMMIT_SHA}")
+            BASE_NAME="commit_${SHORT_SHA}"
+            AI_BRANCH="${AI_PREFIX}${SHORT_SHA}"
+          else
+            echo "::error::No target specified! Neither PR number nor commit SHA was provided. Please specify an autoroll PR number or commit SHA."
+            exit 1
+          fi
+
+          # 2. Create a separate, isolated AI branch from the base branch
+          echo "Creating isolated AI branch: ${AI_BRANCH}..."
+          git checkout -B "${AI_BRANCH}"
+
+          # 3. Download rebase scripts to an isolated directory outside the src/ git repository
+          echo "Fetching latest rebase pipeline scripts into ${GITHUB_WORKSPACE}/rebase_agent..."
+          mkdir -p "${GITHUB_WORKSPACE}/rebase_agent"
+          git fetch origin "${AGENT_REF}:${AGENT_REF}" || true
+          git --work-tree="${GITHUB_WORKSPACE}/rebase_agent" checkout "origin/${AGENT_REF}" -- .github/rebase
+
+          echo "pr_num=${PR_NUM}" >> "${GITHUB_OUTPUT}"
+          echo "base_branch=${BASE_NAME}" >> "${GITHUB_OUTPUT}"
+          echo "ai_branch=${AI_BRANCH}" >> "${GITHUB_OUTPUT}"
+
+      - name: Set up Depot Tools & gclient
+        run: |
+          echo "Cloning depot_tools..."
+          if [[ ! -d "${RUNNER_TEMP}/depot_tools" ]]; then
+            git clone --depth=1 https://chromium.googlesource.com/chromium/tools/depot_tools.git "${RUNNER_TEMP}/depot_tools"
+          fi
+          echo "${RUNNER_TEMP}/depot_tools" >> "${GITHUB_PATH}"
+          export PATH="${RUNNER_TEMP}/depot_tools:${PATH}"
+          "${RUNNER_TEMP}/depot_tools/build_telemetry" opt-out || true
+
+          # Initialize .gclient solution config matching .github/actions/depot_tools/action.yaml
+          cd "${GITHUB_WORKSPACE}"
+          rm -f .gclient
+          echo "Configuring .gclient solution for Cobalt..."
+          "${RUNNER_TEMP}/depot_tools/gclient" config \
+            --name=src \
+            --custom-var="rbe_instance=\"projects/cobalt-actions-prod/instances/default_instance\"" \
+            "${{ github.server_url }}/${{ github.repository }}"
+          echo "target_os=['android']" >> .gclient
+          "${RUNNER_TEMP}/depot_tools/gclient" validate || true
+
+      - name: Set Up Python Dependencies
+        working-directory: src
+        run: |
+          set -x
+          python3 --version
+          curl -sS https://bootstrap.pypa.io/get-pip.py | python3 - --user --break-system-packages
+          echo "${HOME}/.local/bin" >> "${GITHUB_PATH}"
+          export PATH="${HOME}/.local/bin:${PATH}"
+
+          python3 -m pip install --user --no-cache-dir --break-system-packages --require-hashes -r "${GITHUB_WORKSPACE}/rebase_agent/.github/rebase/requirements.txt"
+
+      - name: Run AI Rebase Agent
+        id: run_ai
+        working-directory: src
+        continue-on-error: true
+        env:
+          MODE: ${{ inputs.mode }}
+        run: |
+          set +e
+          python3 "${GITHUB_WORKSPACE}/rebase_agent/.github/rebase/run_rebase_pipeline.py" \
+            --repo-path . \
+            --mode "${MODE}"
+          EXIT_CODE=$?
+
+          if [[ $EXIT_CODE -eq 0 ]]; then
+            echo "resolved=true" >> "${GITHUB_OUTPUT}"
+          else
+            echo "resolved=false" >> "${GITHUB_OUTPUT}"
+          fi
+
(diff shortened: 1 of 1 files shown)


.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+      base_branch:
+        description: 'Base branch if no PR is specified'
+        required: false
+        type: string
+        default: 'staging'
+      ai_branch_prefix:
+        description: 'Prefix for the isolated AI working branch'
+        required: false
+        type: string
+        default: 'experimental/rebase_agent_pr_'
```
oxve, line 21: I think this should have a dedicated branch prefix that we can lock for agent use only.

Can you add a TODO?

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+          - full-pipeline
+          - resolve-conflicts
+          - gn-gen
+          - build-only
+        default: full-pipeline
+
+permissions:
+  contents: write
+  pull-requests: write
+  id-token: write # Required for GCP Workload Identity Federation
```
oxve: Why are all of these write permissions set at the top level? They also don't all seem to be required for both jobs below. Can we make this more targeted by setting permissions on a per-job level?

I'm also a little iffy about giving the llm all these write permissions. I was looking into making an llm powered merge conflict resolver a while back. My idea there was to give the llm step _no_ write permissions. Instead if produced patch files that were applied and committed by a non-llm powered job with write permissions.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+      pr_num: ${{ steps.branch_setup.outputs.pr_num }}
+      base_branch: ${{ steps.branch_setup.outputs.base_branch }}
+      has_commits: ${{ steps.commit_push.outputs.has_commits }}
+    env:
+      DEPOT_TOOLS_METRICS: 0
+      GIT_AUTHOR_NAME: "cobalt-github-releaser-bot"
+      GIT_AUTHOR_EMAIL: "cobalt-github-releaser-bot@google.com"
+      GIT_COMMITTER_NAME: "cobalt-github-releaser-bot"
+      GIT_COMMITTER_EMAIL: "cobalt-github-releaser-bot@google.com"
+      GITHUB_TOKEN: ${{ secrets.CHERRY_PICK_TOKEN || secrets.GITHUB_TOKEN }}
```
oxve: Remove this fallback. `secrets.GITHUB_TOKEN` should not be used for this.

Using `secrets.CHERRY_PICK_TOKEN` is also iffy (policy) but is probably fine for an experiment.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+      RBE_instance: "projects/cobalt-actions-prod/instances/default_instance"
+      RBE_service: "remotebuildexecution.googleapis.com:443"
```
oxve: Nit: Capitalize the whole var

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+            gcloud config set account "${SERVICE_ACCOUNT}" || true
+            GITCOOKIES_PATH="${RUNNER_TEMP}/gitcookies"
+            mkdir -p "$(dirname "${GITCOOKIES_PATH}")"
+            (umask 077 && printf ".googlesource.com\tTRUE\t/\tTRUE\t2147483647\to\tgit-%s=%s\ngooglesource.com\tTRUE\t/\tTRUE\t2147483647\to\tgit-%s=%s\n" \
+              "${SERVICE_ACCOUNT}" "${ACCESS_TOKEN}" "${SERVICE_ACCOUNT}" "${ACCESS_TOKEN}" > "${GITCOOKIES_PATH}")
+            git config --global http.cookiefile "${GITCOOKIES_PATH}"
+            echo "Authenticated with ${SERVICE_ACCOUNT}"
+          else
+            echo "Metadata server not reachable; skipping GCE metadata auth."
+          fi
```
oxve: We have a composable action for this: .github/actions/gob_auth/action.yaml

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+      GIT_AUTHOR_NAME: "cobalt-github-releaser-bot"
+      GIT_AUTHOR_EMAIL: "cobalt-github-releaser-bot@google.com"
+      GIT_COMMITTER_NAME: "cobalt-github-releaser-bot"
+      GIT_COMMITTER_EMAIL: "cobalt-github-releaser-bot@google.com"
+      GITHUB_TOKEN: ${{ secrets.CHERRY_PICK_TOKEN || secrets.GITHUB_TOKEN }}
+      RBE_instance: "projects/cobalt-actions-prod/instances/default_instance"
+      RBE_service: "remotebuildexecution.googleapis.com:443"
+      SISO_CREDENTIAL_HELPER: "gcloud"
+    steps:
+      - name: Checkout Repository (Isolated)
```
oxve: Nit: `(Isolated)`? This is just a normal checkout of either the PR branch or main. The isolated part comes below when it sets up the working branch.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+        run: |
+          echo "Cloning depot_tools..."
+          if [[ ! -d "${RUNNER_TEMP}/depot_tools" ]]; then
+            git clone --depth=1 https://chromium.googlesource.com/chromium/tools/depot_tools.git "${RUNNER_TEMP}/depot_tools"
+          fi
+          echo "${RUNNER_TEMP}/depot_tools" >> "${GITHUB_PATH}"
+          export PATH="${RUNNER_TEMP}/depot_tools:${PATH}"
+          "${RUNNER_TEMP}/depot_tools/build_telemetry" opt-out || true
+
+          # Initialize .gclient solution config matching .github/actions/depot_tools/action.yaml
```
oxve, line 168: Why not use `.github/actions/depot_tools/action.yaml`?

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+      - name: Set Up Python Dependencies
+        working-directory: src
+        run: |
+          set -x
+          python3 --version
+          curl -sS https://bootstrap.pypa.io/get-pip.py | python3 - --user --break-system-packages
+          echo "${HOME}/.local/bin" >> "${GITHUB_PATH}"
+          export PATH="${HOME}/.local/bin:${PATH}"
+
+          python3 -m pip install --user --no-cache-dir --break-system-packages -r "${GITHUB_WORKSPACE}/rebase_agent/.github/rebase/requirements.txt"
```
oxve: Security requires hashes for pip installs. See https://github.com/youtube/cobalt/blob/main/cobalt/tools/requirements.in for how to generate.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+  create_ai_pr:
+    name: Create or Update AI Pull Request
+    needs: [ai_resolve]
+    if: inputs.dry_run != true && needs.ai_resolve.outputs.has_commits == 'true' && always() && !cancelled()
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+      contents: write
+    env:
+      GH_TOKEN: ${{ secrets.CHERRY_PICK_TOKEN || secrets.GITHUB_TOKEN }}
```
oxve: Remove `secrets.GITHUB_TOKEN` fallback. That token can't be used to create PRs (GH limitation)

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
+name: AI Rebase Agent
+
+on:
+  pull_request_target:
```
oxve: The use of `pull_request_target` with untrusted (non-reviewed/non-submitted) code is a risky pattern and generally not allowed. If you were to merge this to the main branch the static analyzers would fail.

Search for `dangerous-triggers` in .github to see the exemptions we've set up. All of them either don't checkout code at all, or only check out already reviewed and submitted code.

We will need to reevaluate how we do this.

.github/workflows/ai_rebase_resolver.yaml
oxve: > id-token: write # Required for GCP Workload Identity Federation

Are you sure about this? What exactly is this required for? We already have a service account configured on our runners.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+          echo "Fetching latest rebase pipeline scripts into ${GITHUB_WORKSPACE}/rebase_agent..."
+          mkdir -p "${GITHUB_WORKSPACE}/rebase_agent"
+          git fetch origin "${AGENT_REF}:${AGENT_REF}" || true
+          git --work-tree="${GITHUB_WORKSPACE}/rebase_agent" checkout "origin/${AGENT_REF}" -- .github/rebase
+
+          echo "pr_num=${PR_NUM}" >> "${GITHUB_OUTPUT}"
+          echo "base_branch=${BASE_NAME}" >> "${GITHUB_OUTPUT}"
+          echo "ai_branch=${AI_BRANCH}" >> "${GITHUB_OUTPUT}"
+
+      - name: Set up Depot Tools & gclient
```
oxve, line 158: Again, use the composable action we already have for this?

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+          python3 -m pip install --user --no-cache-dir --break-system-packages -r "${GITHUB_WORKSPACE}/rebase_agent/.github/rebase/requirements.txt"
+
+      - name: Run AI Rebase Agent
+        id: run_ai
+        working-directory: src
+        continue-on-error: true
+        env:
+          GCP_PROJECT: ${{ secrets.GCP_PROJECT || 'lxn-test' }}
+          GCP_REGION: "us-central1"
+          REASONING_ENGINE_ID: ${{ secrets.REASONING_ENGINE_ID || '6110792607330205696' }}
```
oxve: Should these be variables instead of secrets? Or just hardcoded at the top of the file? Or passed as inputs with defaults?

Secrets doesn't seem like the right fit. Especially since the fallback is written in clear text.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#defining-configuration-variables-for-multiple-workflows

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+          MODE: ${{ inputs.mode || 'full-pipeline' }}
+        run: |
+          set +e
+          echo "::group::Running Rebase Agent Pipeline"
```
oxve: Nit: This group is not adding anything and should be removed. It's encompassing the whole step except the echo >> output part which won't show in the log anyway (no `set -x`).

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+      ai_branch_prefix:
+        description: 'Prefix for the isolated AI working branch'
+        required: false
+        type: string
+        default: 'experimental/rebase_agent_pr_'
+      agent_ref:
+        description: 'Branch containing the AI rebase agent scripts'
+        required: false
+        type: string
+        default: 'experimental/rebase_agent'
```
oxve, line 27: I'm assuming this for development purposes? Should add a TODO to remove this.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+        id: branch_setup
+        working-directory: src
+        run: |
+          set -x
+
+          PR_NUM="${{ inputs.pr_number || github.event.pull_request.number }}"
+          COMMIT_SHA="${{ inputs.commit_sha || github.event.pull_request.head.sha }}"
+          BASE_BRANCH="${{ inputs.base_branch || github.event.pull_request.base.ref || 'staging' }}"
+          AI_PREFIX="${{ inputs.ai_branch_prefix || 'experimental/ai_' }}"
+          AGENT_REF="${{ inputs.agent_ref || 'experimental/rebase_agent' }}"
```
oxve: I think these should be removed given that these inputs already have defaults.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,305 @@
...
+            git fetch --depth=100 origin "${COMMIT_SHA}" || git fetch origin --tags
+            git checkout "${COMMIT_SHA}"
+            SHORT_SHA=$(git rev-parse --short "${COMMIT_SHA}")
+            BASE_NAME="commit_${SHORT_SHA}"
+            AI_BRANCH="${AI_PREFIX}${SHORT_SHA}"
+          else
+            echo "Checking out base branch ${BASE_BRANCH}..."
+            git fetch origin "${BASE_BRANCH}:${BASE_BRANCH}" || true
+            git checkout "${BASE_BRANCH}"
+            CLEAN_NAME=$(echo "${BASE_BRANCH}" | sed 's|experimental/||g; s|/|_|g')
```
oxve: Why do you need a clean name here? I think it would be better if the full base branch name is appended to AI_BRANCH. Cleaning the name will make it harder to find the original branch. It also doesn't contain illegal characters; it's already a branch name.

There's also a (very minor) risk of collisions.

.github/workflows/ai_rebase_resolver.yaml
oxve: Actually, because this is triggered on `pull_request_target` it always checks out `main`.

briantting:
This PR should target the staging branch directly rather than the autoroll-chromium/main-to-staging branch.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Done. Will upload the patchset later.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Done. Will upload the patchset later.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Ack.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Nice catch. Removed.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Done.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Done.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: I can try. It was originally inlined because the checkout step clones into `path: src` (which requires referencing `./src/.github/actions/gob_auth` rather than `./.github/actions/gob_auth`) and I don't want to mess up the cache for main branch.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Make sense. Leave the AI_PREFIX for now till we figure out a branch for them.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Make sense. Ty.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Done.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: 1. Not to mess the cache for main branch, so the paths are different in the rebase agent workflow.
2. The agent workflow will need to resolve the DEPS conflicts before gclient sync. So here, it only prepare the gclient config but not running the sync.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Nice catch. Updated the workflow to include `--require-hashes` on the pip install step. I am also updating the rebase agent branch (`experimental/rebase_agent`) to use `pip-compile --allow-unsafe --generate-hashes` from a `requirements.in` following the pattern in `cobalt/tools/requirements.in`.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Good catch. These are non-sensitive configuration identifiers, not secrets.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Good catch. These are non-sensitive configuration identifiers, not secrets.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Done.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Ack. Removed it.

linxinan-yt: Ack. Removed it.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Agreed. To eliminate the `dangerous-triggers` static analyzer violation and follow Cobalt's security practices, I have removed `pull_request_target` entirely. 
This workflow will now be triggered via `workflow_dispatch` (either invoked directly by the autoroller script when a conflicted PR is created, or manually for testing). This also allowed cleaning up all the inline fallback expressions, as `workflow_dispatch` inputs evaluate their declared defaults reliably.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Sorry replied to the wrong comment. For GOB, yes I can reuse it here, but not for depot tools and gclient. See response from below comment.

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Sounds great. I will update a new patchset. for this.

linxinan-yt: Done. Now agent action only has read permissions and leave the following non AI agent to create a PR for reviewing.

.github/workflows/ai_rebase_resolver.yaml
```diff
@@ -0,0 +1,336 @@
...
+        description: 'Execution mode'
+        type: choice
+        options:
+          - full-pipeline
+          - resolve-conflicts
+          - gn-gen
+          - build-only
+        default: full-pipeline
+
+permissions: {}
```
oxve: Nit/optional: Making the default `contents: read` is fine and would save a few lines of code. Feel free to ignore for this PR.

oxve (approved):

.github/workflows/ai_rebase_resolver.yaml
linxinan-yt: Make sense. Done!

PR #12424, opened 2026-09-03
ci, aosp: Fix AOSP build and add aosp_staging workflow

andrewsavage1:
This PR fixes the AOSP and Android builds on the `staging` branch and adds the AOSP staging CI workflow.

### Summary of Changes:
1. **GitHub Actions**:
   - Added `.github/workflows/aosp_staging.yaml` to run `aosp-arm` on pushes and pull requests targeting `staging` and `autoroll-chromium/main-to-staging`.
   - Updated `.github/actions/depot_tools/action.yaml` to include `aosp` in `target_os` and `target_cpu` setup blocks.
2. **C++17 Compatibility in `third_party/jni_zero`**:
   - Added preprocessor guards (`#if defined(__cpp_concepts) && __cpp_concepts >= 201907L`) and C++17 `std::enable_if_t` fallbacks in `java_refs.h`, `type_conversions.h`, `common_apis.h`, and `default_conversions.h`.
   - Enabled `JNI_ZERO_ENABLE_COMPAT_API 1` for Cobalt builds so Starboard AOSP builds compile cleanly with C++17 without modifying Android (C++20) behavior.
3. **M141 JNI Type Modernization in `copied_base`**:
   - Updated `copied_base` to use `const JavaRef<jobject>&` / `const JavaRef<jstring>&` for JNI parameter and helper functions in `task_runner_android`, `native_uma_recorder`, `trace_event_binding`, `jni_bytebuffer`, and `jni_string`.
4. **Metrics Provider Compatibility**:
   - Updated `FileMetricsProvider` instantiation in `cobalt_metrics_service_client.cc` to pass `/*is_fre=*/false` matching the M141 constructor.
5. **Android Build Configuration**:
   - Adjusted dependency collection in `build/config/android/internal_rules.gni` to ensure clean GN checks for secondary toolchains.
   - Added failing Android test targets to `_failing_android_targets` in `cobalt/build/modular_executable.gni`.

Bug: 495203133
TAG=agy
CONV=08574f7f-732d-4fdc-86b6-a88d2e511ac6


rakuco:
> **Java Switches & CommandLineOverrideHelper**: Replaced deprecated/removed `DeviceInfo.getArch()` with `BuildInfo.getArch()` in `CommandLineOverrideHelper.java` and `JavaSwitches.java`.

This looks backwards, M141 replaced BuildInfo with DeviceInfo.

> 2\. Replaced `third_party/jni_zero` with the C++17 compatible version from `origin/main`, preserving Starboard's requirement to build with C++17 (`use_cxx17 = true` and `build_base_with_cpp17 = true` in `starboard/aosp/arm/toolchain/BUILD.gn`).

Is this pinning jni_zero on a fork of the M138 version?

andrewsavage1:
> Is this pinning jni_zero on a fork of the M138 version?

Right now, yes, so that we can continue to build the starboard impl of aosp with c++17. I don't like this solution, so if you have a better one please lmk

rakuco:
> Right now, yes, so that we can continue to build the starboard impl of aosp with c++17. I don't like this solution, so if you have a better one please lmk

If this works, then great. I'm just worried if the rest of the code base will work with an M138-based jni_zero. In the worst case we'll need to figure out how to make a more recent version work with c++17 instead.

third_party/jni_zero/common_apis.h
```diff
@@ -37,8 +37,14 @@ JNI_ZERO_COMPONENT_BUILD_EXPORT ScopedJavaLocalRef<jobject> ListSet(
     jint idx,
     const JavaRef<jobject>& value);
 // Use ToJniType on the value.
-template <typename V>
+template <typename V
```
andrewsavage1, line 40: This needs to be additionally wrapped in a BUILDFLAG(IS_COBALT)

build/config/android/internal_rules.gni
```diff
@@ -226,7 +226,6 @@ template("write_build_config") {
         }
       }
     }
-    _inputs += _deps_configs
```
andrewsavage1, line 276: I don't fully understand this change

cobalt/browser/metrics/cobalt_metrics_service_client.cc
```diff
@@ -169,8 +169,9 @@ void CobaltMetricsServiceClient::Initialize() {
     base::FilePath metrics_dir =
         base_dir.AppendASCII("BrowserStabilityMetrics");
 
-    auto file_metrics_provider = std::make_unique<metrics::FileMetricsProvider>(
```
rakuco, line 172: This looks unrelated.

third_party/jni_zero/default_conversions.h
```diff
@@ -149,7 +149,7 @@ inline ScopedJavaLocalRef<jobjectArray> ToJniArray(
     }
     ++i;
   }
-  return ScopedJavaLocalRef<jobjectArray>(env, j_array);
+  return ScopedJavaLocalRef<jobjectArray>::Adopt(env, j_array);
```
rakuco: Is this really necessary? `cobalt:gn_all` built fine without this change.

build/config/android/internal_rules.gni
rakuco: This is a big hammer that also changes non-AOSP builds. AFAICS the problem's that after M138 upstream started recording dependencies in these build config files differently (and more thoroughly?).

Since in AOSP builds the default toolchain is not the Android toolchain, a lot of targets that would generate some files are not generated at all. For example, `//url:url_java` depends on `//url/mojom:url_mojom_origin_java`, which doesn't seem to be available because the `mojom` template is invoked by the default toolchain and the Java generation is ignored.

I wonder if there's a way to make those dependencies actually exist/run, otherwise this change should only happen on AOSP builds.

andrewsavage1: I'll look into this further, but would be good to get this in. Added a guard for now

briantting:
Why did this skip all the aosp_staging checks?

linxinan-yt (approved):

briantting (approved):

third_party/jni_zero/common_apis.h
rakuco: nit: having a `// BUILDFLAG(IS_COBALT)` here (and `// defined(__cpp_concepts) && __cpp_concepts >= 201907L` above) and in the other locations would be helpful, I often get confused by this long ifdef blocks with nested if-else ones.

andrewsavage1:
> Why did this skip all the aosp_staging checks?

I'm not sure what you mean, I see i.e. https://github.com/youtube/cobalt/actions/runs/34639860213/job/103396922592?pr=12424

third_party/jni_zero/java_refs.h
hferreiro, line 18: This applies to the whole file. If we want to explicitly use `IS_COBALT`, why not use `BUILDFLAG(IS_COBALT) && defined(__cpp_concepts) && __cpp_concepts >= 201907L` and avoid the duplication of the else branch?

copied_base/base/android/jni_bytebuffer.h
hferreiro, line 10: `std::optional` is still used.

third_party/jni_zero/java_refs.h
hferreiro, line 31: Is this used anywhere?

PR #12433, opened 2026-09-03
staging: Fix build after recent cherry-picks

rakuco:
* experimental_config_manager_unittest.cc (broken by #12417): IWYU for base::FilePath and FILE_PATH_LITERAL
* font_utils.cc (broken by #12431): Finish adapting the code to https://crrev.com/c/6650671
* CobaltActivity.java (broken by #12429): Restore missing arguments to startBrowserProcessesAsync() that were removed in the cherry-pick.

Bug: 552583161
Bug: 517249177
Bug: 448156280

PR #12436, opened 2026-09-03
M142.7403: fix conflicts, builds and tests

rakuco:
Includes #12433.

Bug: 555107908

### DIFF
--- a/components/autofill/core/browser/foundations/autofill_manager.cc
+++ b/components/autofill/core/browser/foundations/autofill_manager.cc
@@ -972,7 +972,7 @@ void AutofillManager::SubscribeToMlModelChanges(
     optimization_guide::proto::OptimizationTarget optimization_target) {
 #if BUILDFLAG(BUILD_WITH_TFLITE_LIB)
   switch (optimization_target) {
->>>>>>> parent of 1ccee01d771 (CONFLICTED Chromium Cherry pick: Revert Cobalt.)
+>>>>>>> parent of 16cb2f821b8 (CONFLICTED Chromium Cherry pick: Revert Cobalt.)
     case optimization_guide::proto::OptimizationTarget::
         OPTIMIZATION_TARGET_AUTOFILL_FIELD_CLASSIFICATION:
       if (!autofill_model_change_subscription_) {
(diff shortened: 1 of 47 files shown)


components/autofill/core/browser/foundations/autofill_manager.cc
```diff
@@ -964,15 +968,9 @@ void AutofillManager::LogCurrentFieldTypes(
 }
 
 void AutofillManager::SubscribeToMlModelChanges(
-<<<<<<< HEAD
     FieldClassificationModelHandler& handler) {
```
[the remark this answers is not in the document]
rakuco: NAK, as discussed in another rebase PR.

rakuco:
cc @abhijeetk @jkim-julie @Gyuyoung @sazabi1976

media/base/decoder_buffer.h
```diff
@@ -208,14 +206,12 @@ class MEDIA_EXPORT DecoderBuffer
     }
     return Allocator::kInvalidHandle;
   }
-#endif // BUILDFLAG(USE_STARBOARD_MEDIA)
 
   // The pointer to the start of the buffer. Prefer to construct a span around
   // the buffer, such as `base::span(decoder_buffer)`.
   // TODO(crbug.com/365814210): Remove in favor of AsSpan().
```
andrewsavage1, line 212: Was this removed in favor of AsSpan? Should these customizations instead be made there?

tools/metrics/common/xml_utils.py
andrewsavage1, line 1: I'm surprised to see us needing to make changes to this file when we haven't made any customizations in main, is this due to different python versions being used in Cobalt vs upstream?

media/base/decoder_buffer.h
rakuco: > Was this removed in favor of AsSpan?

Kind of. In https://chromium-review.googlesource.com/c/chromium/src/+/6914274 callers just started creating a base::span() around instances of this class since span can just use begin(), end() etc.

> Should these customizations instead be made there?

Yes, ideally begin(), end() and other places that used to call data() could be ported to have the SB-Media version as well, although the implementation would look a bit bigger. I've filed https://partnerissuetracker.corp.google.com/issues/556596796 to track this.

tools/metrics/common/xml_utils.py
rakuco: All the changes to this file are backports so that the final commit in the series can land (it's a build fix for tvOS, apparently the Python version in CI didn't like the syntax there).

media/base/decoder_buffer.h
andrewsavage1: Got it, thank you

andrewsavage1 (approved):

PR #12440, opened 2026-09-03
cobalt/metrics: Add pre-JNI Java UMA Persistence

ckim1346:
This PR implements pre-JNI UMA Persistence, focusing on rescuing the earliest Java `StartupGuard` milestones (1-4) that occur before the JNI bridge or Chromium's native Persistent Memory Allocator (PMA) have successfully initialized. This provides insights into the earliest milestones of the startup sequence.

****Key technical changes****:
* **Pre-JNI `MappedByteBuffer` Rescue**: Modified Android's `StartupGuard.java` to mirror early milestones securely into a zero-copy Linux memory-mapped file (`java_startup_state.bin`).
* **C++ Post-Crash Harvester**: Injected an early-recovery loop into `shell_browser_main_parts.cc` (inside `PreCreateThreads`). On the subsequent successful application boot, this C++ phase locates the abandoned session `.bin` file, decompresses it, and retroactively pipes the rescued events natively into `base::UmaHistogramSparse("Cobalt.Startup.MilestoneReached", ...)`.
* **Telemetry XML Funnel Unification**: Expanded `enums.xml` and `histograms.xml` to merge Java Milestones 1-4 directly inline heavily into the unified `CobaltStartupMilestone` funnel, resulting in a single unbroken visualization chart.

Bug: 554178261


cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -47,6 +57,28 @@ public void run() {
...
+        java.io.File prevFile =
+            new java.io.File(context.getFilesDir(), "java_startup_state_previous.bin");
+        file.renameTo(prevFile);
+      }
+      java.io.RandomAccessFile raf = new java.io.RandomAccessFile(file, "rw");
+      // MappedByteBuffer defaults to big-endian, we strictly need little-endian for C++.
+      startupStateBuffer =
+          raf.getChannel().map(java.nio.channels.FileChannel.MapMode.READ_WRITE, 0, 8);
+      startupStateBuffer.order(java.nio.ByteOrder.LITTLE_ENDIAN);
+      startupStateBuffer.putLong(0, 0);
```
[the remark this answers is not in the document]
ckim1346: Done

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -80,7 +112,11 @@ public void setStartupMilestone(int milestone) {
     }
     Log.v(TAG, "StartupGuard setStartupMilestone:" + milestone);
     long mask = 1L << milestone;
-    startupStatus.updateAndGet(current -> current | mask);
+    long current = startupStatus.updateAndGet(curr -> curr | mask);
+
+    if (startupStateBuffer != null) {
+      startupStateBuffer.putLong(0, current);
+    }
```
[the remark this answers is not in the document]
ckim1346: Done but it's likely not enough to only synchronize the putLong. Wrapped both the internal state generation and disk-write into the synchronized(this) block.

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -193,6 +199,34 @@ int ShellBrowserMainParts::PreCreateThreads() {
...
+        // available.
+        for (int i = 1; i <= 4; ++i) {
+          if (startup_status & (1ULL << i)) {
+            base::UmaHistogramSparse("Cobalt.Startup.MilestoneReached", i);
+          }
+        }
+      }
+      base::DeleteFile(state_file);
+    }
+  }
```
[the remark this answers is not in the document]
ckim1346: Done

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -47,6 +57,28 @@ public void run() {
...
+    try {
+      java.io.File file = new java.io.File(context.getFilesDir(), "java_startup_state.bin");
+      // If a file from the previous session exists, rename it so C++ can harvest the previous
+      // session's
+      // state without racing with this fresh session's writes.
+      if (file.exists()) {
+        java.io.File prevFile =
+            new java.io.File(context.getFilesDir(), "java_startup_state_previous.bin");
+        file.renameTo(prevFile);
+      }
```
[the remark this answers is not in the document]
ckim1346: Done

cobalt/android/apk/app/src/test/java/dev/cobalt/shell/StartupGuardPersistenceTest.java
```diff
@@ -0,0 +1,71 @@
...
+public class StartupGuardPersistenceTest {
+  private Context context;
+
+  @Before
+  public void setUp() {
+    context = ApplicationProvider.getApplicationContext();
+    // Clean up previous files if any
+    new File(context.getFilesDir(), "java_startup_state.bin").delete();
+    new File(context.getFilesDir(), "java_startup_state_previous.bin").delete();
+  }
```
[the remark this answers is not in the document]
ckim1346: Done

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -30,6 +37,9 @@ private static class LazyHolder {
     private static final StartupGuard INSTANCE = new StartupGuard();
   }
 
+  // Backing memory-mapped file for Phase 1 cross-layer UMA persistence
+  private java.nio.MappedByteBuffer startupStateBuffer = null;
```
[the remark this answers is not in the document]
ckim1346: No longer necessary as the startupStateBuffer is wrapped in a synchronized lock which guarantees visibility across threads

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CobaltActivityTest.java
```diff
@@ -292,6 +291,11 @@ public void tearDown() {
 
   public CobaltActivity createActivityWithSwitches(final Map<String, String> switches) {
     return new CobaltActivity() {
+      @Override
+      public android.content.Context getApplicationContext() {
+        return androidx.test.core.app.ApplicationProvider.getApplicationContext();
+      }
```
Awallky, line 402: Can we import this at the top and then use the ClassName#methodName pattern here?

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -47,6 +57,33 @@ public void run() {
         };
   }
 
+  public void initializePersistence(android.content.Context context) {
+    try {
+      java.io.File file = new java.io.File(context.getFilesDir(), "java_startup_state.bin");
```
Awallky: Similar finding as above - let's import the package at the top and instantiate the class here.

Awallky: I think this will lead to a file path mismatch. I don't think this leads to C++ Data directory.
How about something like this:

```
@VisibleForTesting
public void initializePersistenceInternal(Context context, @Nullable File baseDir) {
  // Default to Chromium's data directory so it matches C++ DIR_ANDROID_APP_DATA.
  File dir = baseDir != null ? baseDir : new File(PathUtils.getDataDirectory());
  if (!dir.exists()) {
    dir.mkdirs();
  }
...
```

johnxwork (commented):
There is a UmaRecorderHolder class that seems to be made specifically for the use case here. Do we want to reuse that?

cobalt/android/apk/app/src/test/java/dev/cobalt/coat/CobaltActivityTest.java
ckim1346: Thanks for catching

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: And done as well

ckim1346: Good catch as well -- updated

ckim1346:
> There is a UmaRecorderHolder class that seems to be made specifically for the use case here. Do we want to reuse that?

I looked into `CachingUmaRecorder` -- it only buffers the UMA internally in Java heap memory. A crash/stall that happens before JNI/C++ PMA init would be wiped and we'd lose these really early metrics hence why we're persisting to disk instead

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -80,7 +135,14 @@ public void setStartupMilestone(int milestone) {
     }
     Log.v(TAG, "StartupGuard setStartupMilestone:" + milestone);
     long mask = 1L << milestone;
-    startupStatus.updateAndGet(current -> current | mask);
+
+    // Synchronize to ensure atomic write-through to the disk buffer without interleaving
+    synchronized (this) {
```
Awallky: I would not lock on the entire class. Doing so means all references to the class are locked on it. We should create a lock object and block on that.

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -202,14 +201,54 @@ int ShellBrowserMainParts::PreCreateThreads() {
...
 #endif
   return 0;
 }
 
+#if BUILDFLAG(IS_ANDROID)
+void HarvestJavaStartupMetrics() {
+  // Because the JNI bridge is not available early in the Java Android Phase
+  // (e.g. Activity onCreate) `StartupGuard` persists Milestones 1-4 via a
+  // bare-metal MappedByteBuffer. On the subsequent boot, search for the
+  // abandoned `_previous` file, read the 64-bit integer, and feed every tripped
```
Awallky: I wouldn't call it abandoned. Not sure that makes sense.

ckim1346: Changed it to previous session

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: Created a new `stateLock` object to block on -- is that what you had in mind?

cobalt/android/apk/app/src/test/java/dev/cobalt/shell/StartupGuardPersistenceTest.java
```diff
@@ -0,0 +1,72 @@
+package dev.cobalt.shell;
+
+import static org.junit.Assert.assertTrue;
+
+import android.content.Context;
+import androidx.test.core.app.ApplicationProvider;
```
Awallky, line 7: Is `//third_party/androidx:androidx_test_core_java` missing from `cobalt_coat_junit_tests` in the BUILD file?

Awallky: We should add this test file to the BUILD file as well.

ckim1346: Another good catch -- updated and ran, passing both tests: 
```
./out/android-arm_gold/bin/run_cobalt_coat_junit_tests -f "*StartupGuardPersistenceTest*"
 0| [----------] Run 2 test cases from dev.cobalt.shell.StartupGuardPersistenceTest
 0| [ RUN      ] dev.cobalt.shell.StartupGuardPersistenceTest.testSetStartupMilestoneWritesToBuffer[28]
 0| [       OK ] dev.cobalt.shell.StartupGuardPersistenceTest.testSetStartupMilestoneWritesToBuffer[28] (1974 ms)
 0| [ RUN      ] dev.cobalt.shell.StartupGuardPersistenceTest.testInitializeRenamesPreviousFile[28]
 0| [       OK ] dev.cobalt.shell.StartupGuardPersistenceTest.testInitializeRenamesPreviousFile[28] (33 ms)
 0| [----------] Run 2 test cases from dev.cobalt.shell.StartupGuardPersistenceTest (2011 ms)
 0| 
 0| [----------] Global test environment tear-down.
 0| [==========] 2 tests ran. (2013 ms total)
 0| [  PASSED  ] 2 tests.
```

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -47,6 +66,43 @@ public void run() {
...
+  @VisibleForTesting
+  public void initializePersistenceInternal(Context context, @Nullable File baseDir) {
+    try {
+      // Default to Chromium's data directory so it matches C++ DIR_ANDROID_APP_DATA.
+      File dir = baseDir != null ? baseDir : new File(PathUtils.getDataDirectory());
+      if (!dir.exists()) {
+        dir.mkdirs();
+      }
+
+      File file = new File(dir, "java_startup_state.bin");
```
Awallky: Let's make these `public static String` so the test class can also access them. It avoids typos in the test class.

cobalt/shell/browser/shell_browser_main_parts.cc
```diff
@@ -202,14 +210,54 @@ int ShellBrowserMainParts::PreCreateThreads() {
...
+void HarvestJavaStartupMetrics() {
+  // Because the JNI bridge is not available early in the Java Android Phase
+  // (e.g. Activity onCreate) `StartupGuard` persists Milestones 1-4 via a
+  // bare-metal MappedByteBuffer. On the subsequent boot, search for the
+  // previous session's `_previous` file, read the 64-bit integer, and feed
+  // every tripped bit natively into `Cobalt.Startup.MilestoneReached`. This
+  // intentionally fires *after* PMA mounts in `PreEarlyInitialization` to
+  // ensure these writes survive secondary crashes.
+  base::FilePath app_data_dir;
+  if (base::PathService::Get(base::DIR_ANDROID_APP_DATA, &app_data_dir)) {
```
Awallky: Prefer avoiding nested blocks where possible. See this totw http://shortn/_sQOtpWu0pI.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: Done

cobalt/shell/browser/shell_browser_main_parts.cc
ckim1346: Nice reading -- flattened the structure to avoid nesting too many levels

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -18,18 +26,32 @@
...
+ * are recorded here. The bitmask is written to disk via a bare-metal `MappedByteBuffer` (bypassing
+ * the heap) so it easily survives watchdog crashes. The C++ `ShellBrowserMainParts` harvester
+ * collects this file on the subsequent boot and logs the rescued events to
+ * `Cobalt.Startup.MilestoneReached`.
  */
 public class StartupGuard {
+  public static final String STARTUP_STATE_FILE_NAME = "java_startup_state.bin";
+  public static final String STARTUP_STATE_PREVIOUS_FILE_NAME = "java_startup_state_previous.bin";
+
   private final Handler handler;
```
Awallky: All of these member variables should adhere to the AOSP/Chromium Java style guide - https://source.android.com/docs/setup/contribute/code-style#follow-field-naming-conventions

```
Non-public, non-static field names start with m.
...
```

ckim1346: Updated here and in the new test file

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -18,43 +27,96 @@
...
+ * `Cobalt.Startup.MilestoneReached`.
  */
 public class StartupGuard {
-  private final Handler handler;
-  private final Runnable crashRunnable;
-  private final AtomicLong startupStatus = new AtomicLong(0L);
-  private final Map<String, String> diagnosisInfo = new HashMap<>();
-  private final AtomicBoolean isArmed = new AtomicBoolean(false);
+  public static final String STARTUP_STATE_FILE_NAME = "java_startup_state.bin";
+  public static final String STARTUP_STATE_PREVIOUS_FILE_NAME = "java_startup_state_previous.bin";
```
Awallky, line 40: We should reserve disk file harvesting exclusively for pre-native crash recovery (`Cobalt.Startup.PreNativeCrashMilestoneReached`), and delete the previous file in `disarm()` on clean starts.

Because `java_startup_state.bin` is never deleted or cleared on clean exits, every normal launch is harvested as a crash recovery on the next launch.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -115,21 +184,30 @@ public void scheduleCrash(long delaySeconds) {
 
   /** Cancels the pending crash job. */
   public void disarm() {
-    if (isArmed.compareAndSet(/* expect= */ true, /* update= */ false)) {
-      handler.removeCallbacks(crashRunnable);
+    if (mIsArmed.compareAndSet(/* expect= */ true, /* update= */ false)) {
```
Awallky, line 187: We should remove the file in `disarm`, since we successfully complete startup at this point.

```java
...
public void disarm() {
  if (mIsArmed.compareAndSet(true, false)) {
    mHandler.removeCallbacks(mCrashRunnable);
    cleanupStateFileOnSuccess();
  }
}
...
```

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
```diff
@@ -18,43 +27,96 @@
...
+        File prevFile = new File(dir, STARTUP_STATE_PREVIOUS_FILE_NAME);
+        if (prevFile.exists() && !prevFile.delete()) {
+          Log.w(TAG, "Failed to delete old previous state file.");
+        }
+        if (!file.renameTo(prevFile)) {
+          Log.w(TAG, "Failed to rename current state to previous state.");
+        }
+      }
+
+      try (RandomAccessFile raf = new RandomAccessFile(file, "rw");
```
Awallky, line 100: 1. `mStartupStateBuffer` is non-volatile.
2. `channel.map()` and `order()` occur without holding `mStateLock`.
3. `putLong(0, 0)` overwrites milestones that were recorded before persistence was initialized or written concurrently.

Awallky (commented):
My main concern with this is understanding if we are actually emitting these UMAs, or at least reading the file with the startup state, for the previous session before we modify it for the new session. 

Can we validate that we are avoiding race conditions for this, that we can safely write our various states to the file as we progress through startup, and that we emit them properly?

The key will be writing some browser tests to thoroughly test this and maybe an external validation script that checks the various states are emitted as expected (checking the logs). 

**Interesting test idea:** It would be really cool if we could add `DCHECK`s at each milestone and then validate it in tests that the UMA is emitted as expected. If we had a CLI flag that we enabled for each milestone in the test, we could crash at that point while still being able to validate the various states.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: All fair points, updated to perform the I/O .map() operation outside the critical section, but strictly lock mStateLock for the authoritative pointer assignment. Also now migrating any early-caught events via `mStartupStateBuffer.putLong(0, mStartupStatus.get())`.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: Doesn't deleting the file on disarm() actually break the UMA funnels? If we only ever leave the .bin file on disk during a crash, C++ will only ever upload Java failures. Milestone 1 and Milestone 2 would display a 100% crash or abandonment rate, because they would lack the successful baseline denominator for calculation.

ckim1346:
> My main concern with this is understanding if we are actually emitting these UMAs, or at least reading the file with the startup state, for the previous session before we modify it for the new session.
> 
> Can we validate that we are avoiding race conditions for this, that we can safely write our various states to the file as we progress through startup, and that we emit them properly?
> 
> The key will be writing some browser tests to thoroughly test this and maybe an external validation script that checks the various states are emitted as expected (checking the logs).
> 
> **Interesting test idea:** It would be really cool if we could add `DCHECK`s at each milestone and then validate it in tests that the UMA is emitted as expected. If we had a CLI flag that we enabled for each milestone in the test, we could crash at that point while still being able to validate the various states.

I added some browsertests that do pass upon running -- can add more if you see a gap but imo should cover our bases. Is there a preferred way to test locally as well?

cobalt/shell/browser/java_startup_metrics_browsertest.cc
```diff
@@ -0,0 +1,110 @@
...
+  base::ThreadPoolInstance::Get()->FlushForTesting();
+
+  histogram_tester_->ExpectBucketCount("Cobalt.Startup.MilestoneReached", 1, 0);
+  histogram_tester_->ExpectBucketCount("Cobalt.Startup.MilestoneReached", 2, 0);
+  histogram_tester_->ExpectBucketCount("Cobalt.Startup.MilestoneReached", 3, 0);
+  histogram_tester_->ExpectBucketCount("Cobalt.Startup.MilestoneReached", 4, 0);
+  // It should STILL delete the corrupted file so it doesn't block future tests
+  EXPECT_FALSE(base::PathExists(prev_state_file_));
+}
+
```
Awallky, line 136: Can we add a test to verify that we never record these if we have a healthy exit? We only need these to record if we exited due to startupguard.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: Mm I see -- keeping java_startup_state.bin on disk during a healthy session means the C++ background harvester would erroneously flag every normal launch as a crash recovery on the next boot and double-count the UMA. 

Calling cleanupStateFileOnSuccess() in disarm() to wipe the file upon reaching a healthy initialization.

cobalt/shell/android/java/src/dev/cobalt/shell/StartupGuard.java
ckim1346: Done

cobalt/shell/browser/java_startup_metrics_browsertest.cc
ckim1346: Added

PR #12449, opened 2026-09-03
media: Remove audio storage type from audio sink and 1P player

kjyoun:
This is part of the effort to remove kSbMediaAudioFrameStorageTypePlanar
from the 1P media implementation.

The 1P player always works with the interleaved format, so there is no
need to keep or pass the storage type inside of the 1P player.

Validation of the storage type is now performed at the public
SbAudioSinkCreate API boundary. This allows for the removal of
unused parameters across multiple platform-specific sink
implementations and internal components like wsola_internal and
AudioRendererSink.

Bug: 272837615


sideb0ard (approved):
as far i can see, all looks like mechanical changes, lgtm.

niranjanyardi (approved):

starboard/shared/starboard/audio_sink/audio_sink_create.cc
```diff
@@ -29,9 +29,15 @@ SbAudioSink SbAudioSinkCreate(
     SbAudioSinkUpdateSourceStatusFunc update_source_status_func,
     SbAudioSinkConsumeFramesFunc consume_frames_func,
     void* context) {
+  // 1P Cobalt audio sink implementation only supports interleaved frames.
```
osagie98, line 32: Change LGTM, my only question is if it's still possible for a 1P platform to support planar storage in the future. Is it that Interleaved has an inherent advantage, or is it that 1P just happens to just use it?

kjyoun: Yes, the public Starboard API still retains audio_frame_storage_type as a parameter, so 1P could technically support it in the future if needed. This comment simply reflects current behavior (this PR does not change runtime behavior or the Starboard API boundary).
 
That said, switching to planar in 1P is very unlikely. Most platform audio APIs (Android AudioTrack, ALSA, PulseAudio, tvOS) natively consume interleaved PCM, and all upstream renderer call sites already pass kSbMediaAudioFrameStorageTypeInterleaved. In practice, we had no active code path utilizing planar storage, and we're actually tracking deprecating the parameter entirely in b/557194697.

PR #12470, opened 2026-09-04
[Cobalt] Add URL sanitization for URLs provided by the platform.

jshorta:
Use deep link resolution for URLs that don't start with the YTLR prefix.

Bug: 503024637


cobalt/app/app_event_runner.cc
```diff
@@ -378,14 +380,36 @@ class AppEventRunnerImpl : public AppEventRunner,
...
+    // In Gold builds, we enforce that this URL points strictly to YouTube TV.
+    if (!init_argv.empty()) {
+      // CommandLinePreprocessor makes the startup URL is the last argument.
+      const std::string& startup_url = init_argv.back();
+      if (startup_url.find(::switches::kDefaultURL) != 0) {
+        LOG(WARNING) << "Invalid Gold startup URL. Rerouting to deep link: "
+                     << startup_url;
+
+        // Override the deep link if the platform didn't provide one already.
+        if (!initial_deep_link) {
```
andrewsavage1, line 401: @jellefoks might know better, but is this right? Would this change the behavior of launching the default app without a URL?

jellefoks (approved):

cobalt/app/app_event_runner.cc
jshorta: Looks like Jelle is okay w/ this approach. Feel free to reopen if there are any concerns.

andrewsavage1 (approved):

cobalt/app/app_event_runner.cc
andrewsavage1: Sounds good, let's keep it then

PR #12474, opened 2026-09-05
android: Eliminate runtime JNI build type queries using java_cpp_template

kjyoun:
Replace runtime JNI calls to BaseStarboardBridge.isReleaseBuild() and
isDevelopmentBuild() with compile-time constants generated via
java_cpp_template.

Bug: 557444130


johnxwork (approved):
This is sick!

PR #12475, opened 2026-09-05
coat: Make CommandLineOverrideHelper flavor-agnostic

kjyoun:
Refactor CommandLineOverrideHelper to be build-flavor agnostic by
removing the CommandLineOverrideHelperParams class and accepting a
list of strings directly. This simplifies the interface and moves
environment-specific logic, such as the official build check for
remote-allow-origins, into CobaltActivity.

The change also consolidates command-line switch applications into a
single batch call and flattens nested loops for better readability.
CobaltActivity now linearly accumulates arguments from intents and
metadata. Updated unit tests to reflect the API changes.

Bug: 414009070

### DIFF
--- a/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
+++ b/cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
@@ -34,6 +34,7 @@ import android.widget.FrameLayout;
 import android.widget.Toast;
 import android.window.OnBackInvokedCallback;
 import android.window.OnBackInvokedDispatcher;
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.annotation.RequiresApi;
 import androidx.annotation.VisibleForTesting;
@@ -53,6 +54,7 @@ import dev.cobalt.util.JavaSwitches;
 import dev.cobalt.util.Log;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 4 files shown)


cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CommandLineOverrideHelper.java
```diff
@@ -175,67 +164,49 @@ public static void getFlagOverrides(CommandLineOverrideHelperParams params) {
...
-                break; // Avoid adding the same param multiple times
-              }
-            }
-          } else {
-            cliOverrides.add(param);
-          }
+      String key = parts[0];
+      String value = parts[1];
+      for (String v : value.split(";")) {
+        if (key.equals("--js-flags")) {
```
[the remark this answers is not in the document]
kjyoun: Pre-existing behavior preserved intentionally to keep this refactoring PR strictly a NO-OP. We can track defensive parsing improvements in a follow-up if needed.

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
```diff
@@ -174,28 +182,22 @@ protected void createContent(final Bundle savedInstanceState) {
...
-      if (!extraCommandLineArgs.isEmpty()) {
-        if (commandLineArgs != null) {
-          extraCommandLineArgs.addAll(0, Arrays.asList(commandLineArgs));
-        }
-        commandLineArgs = extraCommandLineArgs.toArray(new String[0]);
+      List<String> args = new ArrayList<>();
+      appendIntentArgs(args);
+      appendMetaDataArgs(args, getActivityMetaData());
+      args.addAll(JavaSwitches.getExtraCommandLineArgs(getJavaSwitches()));
+      if (isDevelopmentBuild()) {
```
Awallky: If we are going to append this at the end of the list, we should check if users have already added this to the CLI switches before adding it. Should also add a test for this, ensuring user-provided overrides are respected.

Awallky (approved):

cobalt/android/apk/app/src/main/java/dev/cobalt/coat/CobaltActivity.java
kjyoun: Thanks for catching this.

Done. and added unit tests

